PR #3, opened 2020-01-24
Suggested changes from tile rendering code review

michaeljb:


spec/lib/engine/tile_spec.rb
```diff
@@ -88,12 +88,12 @@ module Engine
           '1',
           color: :yellow,
           parts: [
-            townA,
-            Path.new(edge0, townA),
-            Path.new(townA, edge2),
-            townB,
-            Path.new(edge3, townB),
-            Path.new(townB, edge5),
+            town_A,
```
tobymao: lower case

PR #4, opened 2020-03-12
Allow for purchasing the cheapest private if it has no bids

hcarver:
Right now, you can bid on the cheapest private rather than buy it outright. The existing bid action is already designed to let you purchase the private in that situation, so this change is really just changing the way the action gets rendered.


tobymao:
can you add parenthesis to keep styling consistent? other than that looks good

hcarver:
Ah! I misunderstand a rubocop warning and removed them. They should be back now.

PR #7, opened 2020-03-27
add Travis-CI config

michaeljb:
@tobymao  you'll need to setup the repo to build on Travis, but once you do, I think this config will be enough to automatically run `rake` on each push, which will run the unit tests and then rubocop

https://travis-ci.org/github/tobymao/18xx

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+---
+sudo: false
+language: ruby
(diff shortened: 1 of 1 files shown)


michaeljb:
Rejecting this in favor GitHub Actions - https://github.com/features/actions

PR #8, opened 2020-04-05
render yellow town tiles as curvilinear

michaeljb:
55, 56, and 69 don't look right, but also aren't in 1889. For those we'd need to calculate a different spot on the curve for where to put the town. Right now it's in the exact midpoint, and those three tiles have track crossings at or near the midpoint.

### DIFF
--- a/assets/js/view/part/base.rb
+++ b/assets/js/view/part/base.rb
@@ -37,6 +37,33 @@ module View
       TOP_ROW = [0, 1, 2, 3, 4].freeze
       BOTTOM_ROW = [19, 20, 21, 22, 23].freeze
 
+      REGIONS_ROTATED_ONCE = {
+        0 => 3,
+        1 => 4,
+        2 => 10,
+        3 => 11,
+        4 => 18,
+        5 => 1,
+        6 => 2,
+        7 => 8,
+        8 => 9,
+        9 => 16,
+        10 => 17,
+        11 => 23,
+        12 => 0,
+        13 => 6,
+        14 => 7,
+        15 => 14,
+        16 => 15,
+        17 => 21,
+        18 => 22,
+        19 => 5,
+        20 => 12,
+        21 => 13,
+        22 => 19,
+        23 => 20,
+      }.freeze
+
       def preferred_render_locations
         [
           {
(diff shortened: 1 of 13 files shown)


assets/js/view/part/base.rb
```diff
@@ -133,6 +133,41 @@ def combined_cost(region_weights)
           weight * regions.sum { |region| @region_use[region] }
         end
       end
+
+      def rotate_region(region, rotations: nil, degrees: nil)
+        rotations ||= (degrees / 60).to_i
+
+        regions = {
```
tobymao, line 140: turn this into a constant

assets/js/view/part/towns.rb
```diff
@@ -3,17 +3,38 @@
...
+      needs :routes
 
       def render
-        @tile.towns.map do |_|
-          h(Part::TownDot, region_use: @region_use) if @tile.lawson?
+        @tile.towns.map do |town|
+          if @tile.lawson?
+            h(Part::TownDot, region_use: @region_use, color: color_for(town))
+          else
+            paths = @tile.paths.select { |p| p.town == town }
```
tobymao, line 22: maybe we should extract this pattern to be more efficient

assets/js/view/part/track_curvilinear_path.rb
```diff
@@ -32,38 +32,61 @@ def compute_track_rotation_degrees(edge_a, edge_b)
...
+
+        @curvilinear_type = compute_curvilinear_type(edge_a, edge_b)
+        @rotation = compute_track_rotation_degrees(edge_a, edge_b)
+      end
+
       def preferred_render_locations
+        regions =
+          case @curvilinear_type
+          when SHARP
+            [13, 14, 15, 21]
```
tobymao, line 46: maybe we can turn all these things into constants so they aren't reallocated every time

assets/js/view/part/base.rb
michaeljb: cf0fd10

assets/js/view/part/towns.rb
michaeljb: Something like this? ecb2dde

michaeljb: (added 1e13ec1)

tobymao: hmm a bit fancier than i expected.

tobymao: i guess the caching is nice

assets/js/view/part/track_curvilinear_path.rb
michaeljb: 40b75db

tobymao:
looks good

assets/js/view/part/towns.rb
michaeljb: What approach were you thinking of for improving the efficiency? It would seem I didn't come up with the same approach you had in mind--I'm pretty sure apart from the caching, what I did was just as efficient :grinning:

tobymao: yea let's keep for now, maybe i'll come up with something later

PR #21, opened 2020-04-10
Only allow laying track when it's legal to. Fixes #13.

### LINKED ISSUE #13: Operating round : Placed a second tile instead of buying train
First Operating Round : Awa is floated in Stock round 1
I placed a yellow tile on home tile, then placed a second tile next to it.
Train buying step was passed by doing this

(Log is inverted)
-- Stock Round 2 --
Awa Railroadspends $80 and lays tile 7 with rotation 0 on K6
Awa Railroad's share price changes from $65 to $60
Awa Railroad does not run
Awa Railroad lays tile 5 with rotation 3 on K8

hcarver:


hcarver:
Sure, I think that's done now @tobymao. I had problems testing properly because tile rotation isn't working for me, but I was definitely able to lay a port tile + a single normal lay in the same turn.

PR #53, opened 2020-04-16
Format currency based on game title

dfannius:
Currency values are now printed as `$<val>` by default but can be customized per title (e.g., `¥<val>` for 1889).


tobymao:
thank you!

Issue #54, opened 2020-04-16
Change references to corporations to the Symbol


tobymao:
Change views to use the symbol to save space and also I think most people reference by symbol anyways. Hover text can be full corp name like in rollingstock.net

Issue #50, opened 2020-04-15
Condense player info boxes


sparr:
Current player info box:

```
Players
name: RJ_E                                  name: Vorrt                                 name: EisVrouw81
cash: 295                                   cash: 174                                   cash: 170
companies:                                  companies:                                  companies: South Iyo Railway
Iyo Railway *50%                            Awa Railroad 0%                             Awa Railroad 0%
Awa Railroad 0%                             Tosa Electric Railway *50%                  Uwajima Railway *50%
Sanuki Railway 0%                           Iyo Railway 0%                              Iyo Railway 0%
Takamatsu & Kotohira Electric Railway 0%    Sanuki Railway 0%                           Sanuki Railway 0%
Tosa Electric Railway 10%                   Takamatsu & Kotohira Electric Railway 0%    Takamatsu & Kotohira Electric Railway 0%
Tosa Kuroshio Railway 0%                    Tosa Kuroshio Railway 0%                    Tosa Electric Railway 10%
Uwajima Railway 0%                          Uwajima Railway 10%                         Tosa Kuroshio Railway 0%
```

Proposed layout:
```
Players
name        RJ_E       Vorrt  EisVrouw81
cash         295         174         170
companies                               
AR            0%          0%          0%
IR          *50%          0%          0%
SR            0%          0%          0%
TER          10%        *50%         10%
TKER          0%          0%          0%
TKR           0%          0%          0%
UR            0%         10%        *50%
```

Changes:

1. Switch to a table with just one label per row. This makes the info denser and also ensures the companies are in the same order for each player, alphabetical here but any consistent order is better than random.
1. Company names condensed to acronyms / abbreviations. Not visible above is a tooltip on each one that would expand to the full name. The color(s) of the company could also appear around or alongside the abbreviation and/or full name.
1. Holding amounts right aligned for quicker visual scanning

tobymao:
done

![Screen Shot 2020-04-17 at 10 52 27 PM](https://user-images.githubusercontent.com/8205034/79629427-98184f80-80fe-11ea-96fd-7a35856fbd45.png)

PR #65, opened 2020-04-17
Make player cards a little more pleasant

dfannius:
Starts to address #54, though hover text is not implemented, and there's a lot of design work yet to do.


dfannius:
Also starts to address #50.

tobymao:
closing in favor of @Talbatross pr coming tonight

PR #70, opened 2020-04-17
Deselect corporation when player changes, fixes #66

### LINKED ISSUE #66: Unfocus previous player's corporation choice in stock rounds
In a stock round, if Player A selects Corporation X and chooses to buy a share, the corporation is still selected at the beginning of Player B's turn. Player B should start with a clean slate.

dfannius:
I don't have a lot of faith that this code is idiomatic but it does induce the correct behavior. Improvements welcome.

### DIFF
--- a/assets/js/view/stock_round.rb
+++ b/assets/js/view/stock_round.rb
@@ -15,9 +15,14 @@ module View
 
     needs :round
     needs :selected_corporation, default: nil, store: true
+    needs :last_player, default: nil, store: true
 
     def render
       @current_entity = @round.current_entity
+      if @last_player.nil? || !@last_player.equal?(@current_entity)
+        store(:selected_corporation, nil, skip: true)
+        store(:last_player, @current_entity, skip: true)
+      end
 
       children = [
         h(UndoAndPass),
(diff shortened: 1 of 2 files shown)


assets/js/view/stock_round.rb
```diff
@@ -15,9 +15,14 @@ class StockRound < Snabberb::Component
 
     needs :round
     needs :selected_corporation, default: nil, store: true
+    needs :last_player, default: nil, store: true
 
     def render
       @current_entity = @round.current_entity
+      if @last_player.nil? || !@last_player.equal?(@current_entity)
```
tobymao, line 22: if @last_player != @current_entity

dfannius:
> my only other comment is do you only want this to happen in the stock round?

The stock round was the only place where I observed the problem. In the private auction the issue doesn't appear, and in the operating round there's nothing to accidentally leave selected (that I could tell).

dfannius:
I got a notification of a suggestion that I use `if @last_player != @current_entity` but I don't see it on this page; I don't know if you removed it or if my Github skills aren't up to the task. Anyway, that was what I originally wrote, but my recollection is that it crashed inside an equality test when one of the items being compared was `nil`. I even tried `equal` too. If you think it really should work I'll go back and investigate.

PR #81, opened 2020-04-18
Add vertical space between elements on game screen

dfannius:
I also got rid of the line that just said "Players" since it mostly acted as clutter and nothing else (corporations, stock market) is labeled anyway.

### DIFF
--- a/assets/js/view/game.rb
+++ b/assets/js/view/game.rb
@@ -158,23 +158,13 @@ module View
     def render_game
       @round = @game.round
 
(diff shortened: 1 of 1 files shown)


dfannius:
I'll try that.

dfannius:
Just did players for now. The margin on the container has to be smaller on top than on bottom, to take the top margin of the player cards into account. I couldn't find a way to make the CSS collapsing margin algorithm handle it automatically.

PR #85, opened 2020-04-18
Stock market view improvements (colors, font size)

dfannius:
Muted colors a bit (especially the red), and reduced font size a little to make more room for tokens and to be less obtrusive.


assets/js/view/stock_market.rb
```diff
@@ -23,10 +23,17 @@ def render
         border: 'solid 1px rgba(0,0,0,0.2)',
       )
 
+      color_map = {
```
tobymao, line 26: make this a constant

PR #94, opened 2020-04-19
Add brief 'about' page

hcarver:


api.rb
```diff
@@ -127,6 +127,10 @@ def on_close(room, q)
       r.halt 404 unless (game = Game[id])
       render(game_data: game.to_h(include_actions: true))
     end
+
+    r.on 'about' do
```
tobymao, line 131: can you instead add this to line 126 in the array?

assets/js/view/about.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module View
+  class About < Snabberb::Component
+    def render
+      about_string = <<~ABOUT
+        18xx.games is created and maintained by Toby Mao. It is an open source project, and you can find the
+        code at https://github.com/tobymao/18xx
+      ABOUT
+      h(:div,
```
tobymao, line 10: put the array on the same line as 10

assets/js/view/about.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module View
+  class About < Snabberb::Component
+    def render
+      about_string = <<~ABOUT
+        18xx.games is created and maintained by Toby Mao. It is an open source project, and you can find the
+        code at https://github.com/tobymao/18xx
```
tobymao, line 8: this should be an href, is this clickable?

Issue #43, opened 2020-04-15
No train roster visible 


jwgeist00:
Train roster only appears to be viewable during train buying step. Would be useful information to have on hand.

tobymao:
fixed

PR #95, opened 2020-04-19
Train roster tab

dfannius:
Addresses #43.

I would like to add more info, as is found in the per-title data from the JSON file, but we don't appear to read the train data from that file yet, so for now this is pretty bare-bones (both in data and in presentation).


dfannius:
I forgot about trains in the bank pool. Now should be good to go.

Issue #2, opened 2020-01-24
Tile Rendering


michaeljb:
required for a playable 1889:

- [x] private companies that block building
- [x] upgrade cost
- [x] icons (mountain, river)
- [x] definitions for 1889 preprinted tiles
- [x] offboard / red tiles
- [x] render reserved spaces (start with text, eventually corporation's logo)

required for playable 1830, 1846, and 1861:

- [x] add definitions for 1830
- [x] add definitions for 1846
- [ ] add definitions for 1861
- [ ] render multiple cities (e.g., OO, NY, Chi)
- [x] multiple revenue centers: render just one revenue circle, or render two separately (currently 2 rendered, right on top of each other)
- [ ] more icons (e.g., meat packing, port)
- [ ] E/W label and bonus
- [ ] render 1830's preprinted tile for PRR's home

other (mainly aesthetic features):

- [ ] three-digit revenues need to render without bleeding over the circle
- [x] elements should not render outside bounds of hex (e.g., Uwajima)
- [x] make location names part of the hexes instead of tiles
- [x] smart positioning for circles showing the revenue
- [x] "white background" / box for 4+ city slots (implemented for 2-3)
- [ ] clearly show crossing track (white outline)
- [x] fix rendering gentle curve on top of other curves (see tile 16)
- [x] center tile name when rendering just the tile name due to error?
- [x] 1889 preprinted gray town names should be rendered (Yawatahama, Muroto)

michaeljb:
svg code that was used for white outline for straight track:

```
          h(:path, attrs: { d: 'm -4 86 L -4 -86', stroke: 'white', 'stroke-width' => 2 }),
          h(:path, attrs: { d: 'm 4 86 L 4 -86', stroke: 'white', 'stroke-width' => 2 }),
```

michaeljb:
Location names should probably be moved to the hexes, and be assigned in a game's init_map

tobymao:
jcl suggests boxing grouped city circles

sparr:
Tile definitions for additional games are coming as @kelsin implements automatic export from 18xx-maker.

kelsin:
I implemented the boxes for cities of size 4-6, that's live.

kelsin:
Double city rendering is like... 1/2 done. They show up but stupid right now. Going to abstract the code I wrote for double dits, and get that into cities as well. It's in my plan for this weekend unless someone else wants it of course.

kelsin:
[...]

PR #109, opened 2020-04-20
City boxes, "all tiles" page, show company blocker on tile even before it is bought in the private auction

michaeljb:
* Cities with 2 or 3 slots have boxes now (per https://github.com/tobymao/18xx/issues/2#issuecomment-614425795)
  * the 3-slot solution is lazily scaling down the full-size hexagon
  * cities with 4+ slots still don't have a box, but 1889 doesn't use those
* add new route/page at `/all_tiles`, to show all defined tiles
  * also shows all custom starting hexes defined for 1889 (including corporations and company blockers)
* company blocker is now rendered on map even before it is bought during the private auction

### DIFF
--- a/api.rb
+++ b/api.rb
@@ -106,6 +106,10 @@ class Api < Roda
     on 'api' do |hr|
       hr.hash_routes :api
     end
+
+    on 'all_tiles' do
+      render(app_route: 'all_tiles')
+    end
   end
 
   route do |r|
@@ -126,10 +130,6 @@ class Api < Roda
     r.on 'game', Integer do |id|
       r.halt 404 unless (game = Game[id])
       render(game_data: game.to_h(include_actions: true))
-    end
-
-    r.on 'all_tiles' do
-      render(app_route: 'all_tiles')
     end
   end
 
--- a/api.rb
+++ b/api.rb
@@ -106,10 +106,6 @@ class Api < Roda
     on 'api' do |hr|
       hr.hash_routes :api
     end
-
-    on 'all_tiles' do
-      render(app_route: 'all_tiles')
-    end
   end
 
   route do |r|
@@ -123,7 +119,7 @@ class Api < Roda
       render_with_games
     end
 
-    r.on %w[/ about signup login profile] do
+    r.on %w[/ about signup login profile all_tiles] do
       render_with_games
     end
 
--- a/api.rb
+++ b/api.rb
@@ -127,6 +127,10 @@ class Api < Roda
       r.halt 404 unless (game = Game[id])
       render(game_data: game.to_h(include_actions: true))
     end
+
+    r.on 'all_tiles' do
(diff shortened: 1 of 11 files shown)


api.rb
```diff
@@ -127,6 +127,10 @@ def on_close(room, q)
       r.halt 404 unless (game = Game[id])
       render(game_data: game.to_h(include_actions: true))
     end
+
+    r.on 'all_tiles' do
+      render(app_route: 'all_tiles')
```
tobymao, line 132: you can just put this in the main  on line 94

assets/js/view/part/city.rb
```diff
@@ -61,13 +61,53 @@ def render_part
...
         end
 
-        h(:g, { attrs: { class: 'city' } }, slots)
+        children = []
+
+        children << render_box(slots.size) if slots.size > 1
+
+        children += slots
+
+        h(:g, { attrs: { class: 'city' } }, children)
```
tobymao, line 76: could do h('g.city')

assets/js/view/part/city.rb
```diff
@@ -61,13 +61,53 @@ def render_part
...
+              :polygon,
+              attrs: {
+                points: '100,0 50,-87 -50,-87 -100,-0 -50,87 50,87',
+                fill: 'white',
+                transform: 'scale(0.458)',
+              }
+            )
+          end
+
+        h(:g, { attrs: { class: 'city-box' } }, [
```
tobymao, line 108: h('g.city_box') let's use all underscores for classes and ids

assets/js/view/part/city.rb
```diff
@@ -61,13 +61,53 @@ def render_part
...
+      #   scaling the full-size hexagon
+      # - implement for 4, 5, and 6 slot cities
+      def render_box(slots)
+        box =
+          case slots
+          when 2
+            h(
+              :rect,
+              attrs: {
+                width: 2 * SLOT_RADIUS,
```
tobymao, line 90: these polygon attrs can be constants

assets/js/view/part/city.rb
michaeljb: 9ad0115

assets/js/view/part/city.rb
michaeljb: 9ad0115

assets/js/view/part/city.rb
michaeljb: 5c6d36d

api.rb
michaeljb: 3c5cb1b

michaeljb:
Huh, something in 5c6d36d0ba35125b6d77f81ccc66f1119604559e went haywire...
Screenshot from 2020-04-20 20-17-52

michaeljb:
Fixed in b0593e961138a4dfcf57372fff49aedcca4653e0

PR #151, opened 2020-04-25
Move Private Auction Actions to a better place

Talbatross:


assets/js/view/auction_round.rb
```diff
@@ -20,7 +20,6 @@ def render
       h(:div, [
         h(UndoAndPass, undo: @game.actions.size.positive?),
         *render_companies,
-        render_input,
         h(View::Players, game: @game),
       ].compact)
     end
```
tobymao, line 25: clean up action_round, remov selected_company, actionable and imports as they aren't used

assets/js/view/company.rb
```diff
@@ -2,6 +2,8 @@
 
 module View
   class Company < Snabberb::Component
+    include Actionable
```
tobymao, line 5: require this to be safe

assets/js/view/company.rb
```diff
@@ -86,14 +87,62 @@ def render
         ]),
       ]
 
-      if @bids
+      if @bids&.length&.positive?
```
tobymao, line 90: @bids&.any?

assets/js/view/company.rb
```diff
@@ -86,14 +87,62 @@ def render
...
         children << h(:div, { style: bidders_style }, 'Bidders:')
         children << render_bidders
       end
 
       children << h(:div, { style: bidders_style }, "Owner: #{@company.owner.name}") if @company.owner
 
-      h(:div, props, children)
+      h(:div, [
```
tobymao, line 97: why do this, add it to the children instead

assets/js/view/company.rb
```diff
@@ -86,14 +87,62 @@ def render
...
+        h(:div, props, children),
+        render_actions
+      ])
+    end
+
+    def render_actions
+      return h(:div) unless selected? && @game.round.auction?
```
tobymao, line 104: don't need this if you add it to the children

also don't use unless with multiple statements, change it to an if

assets/js/view/company.rb
Talbatross: so that it's outside of the card.

assets/js/view/company.rb
Talbatross: you mean add require 'view/actionable' on top?

assets/js/view/company.rb
Talbatross: if I add to the children the action will appear within the card

assets/js/view/company.rb
tobymao: yes

assets/js/view/company.rb
tobymao: i see

assets/js/view/auction_round.rb
```diff
@@ -1,17 +1,12 @@
...
 require 'view/undo_and_pass'
 
-require 'engine/action/bid'
-
 module View
   class AuctionRound < Snabberb::Component
-    include Actionable
-
-    needs :selected_company, default: nil, store: true
+    needs :game, store: true
```
tobymao, line 9: does this need to be stored?

Issue #179, opened 2020-04-27
While running trains, a 1-city route was allowed


yoyoson:
However according to rule 8.2 from t[he deepthought rules](http://www.deepthoughtgames.com/games/1889/Rules.pdf), 

> A train may not run on a route shorter than length two, therefore a 2-train must run routes of exactly length two.

For instance, a yellow straight was laid WNW out of the KO home tile, but no tile yet in Marugame (I2)

I was allowed to run K4 to J3, and it paid 30 for having the KO home tile in there.

The JSON for the game lists it as a route including from K4 to J3. 
```

{
      "type": "run_routes",
      "entity": "KO",
      "entity_type": "corporation",
      "id": 62,
      "routes": [
        {
          "train": "2-2",
          "hexes": [
            "K4",
            "J3"
          ]
        },
        {
          "train": "2-3",
          "hexes": [
            "K4",
            "K6",
            "L7"
          ]
        }
      ]
    }

```

Full blob here: 

[SingleCityRouteAllowed.txt](https://github.com/tobymao/18xx/files/4537616/SingleCityRouteAllowed.txt)
This was in a hotseat game.

jenf:
Should be resolved

PR #200, opened 2020-04-28
Enforce rule 8.2 making routes be at least 2 stops long (fixes #179)

jenf:
Sent the migration script separately

PR #205, opened 2020-04-29
Require route length of at least 2.

srlindemann:


srlindemann:
Fixes #179 
Probably should be closed in favor of #200 ?

tobymao:
this is already done in a pr https://github.com/tobymao/18xx/pull/200

PR #208, opened 2020-04-29
WIP Initial attempt to change Enter button to a link

### LINKED ISSUE #156: Ability to open game in new tab
Not sure if you intentionally don't support this, but the current JS to enter a game isn't a real link, so you can't directly shift+click into another tab

Not huge, but also easy to fix. Thanks for the great work!

sparr:
Changing it to a link was easy. Styling that link to look just like the buttons, not so much. It might be easier to define styles for both of them together, and override all the default styles on the buttons at the same time.

RESOLVES #156

### DIFF
--- a/assets/js/view/game_card.rb
+++ b/assets/js/view/game_card.rb
@@ -66,7 +66,7 @@ module View
           end
           JOIN_YELLOW
         when 'active'
-          buttons << render_button('Enter', -> { enter_game(@gdata) })
+          buttons << h('a.button', { style: { top: '1rem', float: 'right', 'text-decoration': 'none', }, attrs: { href: url(@gdata) } }, 'Enter')
           acting?(@user) ? YOUR_TURN_ORANGE : ENTER_GREEN
         when 'finished'
           buttons << render_button('Review', -> { enter_game(@gdata) })


jcoveney:
Jen Freeman (not sure of github id) noted that if you run ` docker-compose exec rack rubocop` (while you have the docker instance up) it will run rubocop locally, which is what is giving you the errors.

assets/js/view/game_card.rb
```diff
@@ -66,7 +66,7 @@ def render_header
           end
           JOIN_YELLOW
         when 'active'
-          buttons << render_button('Enter', -> { enter_game(@gdata) })
+          buttons << h('a.button', { style: { top: '1rem', float: 'right', 'text-decoration': 'none', }, attrs: { href: url(@gdata) } }, 'Enter')
```
tobymao, line 69: let's make this a function so we can do it properly

override the onclick handler to return false and provide a custom on click. that way it doesn't actually do a refresh

sparr: I am not aware of an accessible cross-browser compatible way to have a button that behaves like the user would expect with ctrl/cmd/middle/etc-click to open a new tab, but also does a page load without refresh if they click/enter the button normally.

Issue #193, opened 2020-04-28
Spreadsheet view


jcoveney:
This has come up in slack. I'm actually planning to give it a try (despite my non-existent front-end skills), and figure there should be an issue to track the feature (slack is fairly ephemeral).

gchambers419:
Agree about tracking the feature request here vs the "chat room"
I suggest logging design intent as well, so everyone knows what's coming.
My theory was that there would be a player listing as one axis, with columns for name, privates owned, money and net worth, then a set of columns, one per corporation, with names/symbols across the top, and rows at the bottom for cash on hand, trains owned, and privates owned.
As always the "working rule" applies: The person doing the work makes the rules.

jcoveney:
Yes, that was more or less my intent. If you want to give it a stab, feel free of course, though I will probably try at some point soon. I've already spent a bit of time with the code and toby graciously provided a lot of guidance. My original thought was to have a switch which turned on a spreadsheet view in the main tab, but for now I'm actually thinking of adding a spreadsheet tab. The reason is the view code right now is quite clean in how it renders everything...adding a switch based on that setting that then renders things in one of two ways is kind of messy. If it is in its own tab, then it can be completely separate and I think will be less ugly as a maintenance burden. Plus it's easier to implement initially and we can see how it looks. If we want, we can integrate it with the summary easily.

Something I would like as a nice to have would be for the spreadsheet to expend out and have the companies' income during each OR. But I'm not sure if that information is currently available in the client.

gchambers419:
Given my lack in the following areas, you are much more likely to get to it than I:
1. Ruby programming.  It's just another language, but it **is** a new language
1. Anything beyond the most rudimentary knowledge of CSS or modern web coding in general
1. Time
1. Attention span

I prefer a separate tab as well.

dfannius:
One more vote for a separate tab.

jcoveney:
Initial PR is up. Feel free to discuss specifics there. Would love for anyone good at CSS to help out...

PR #210, opened 2020-04-29
Add spreadsheet tab with a spreadsheet view of the game state

jcoveney:
This is my first stab at the spreadsheet. There are improvements that could be made, but I think this is a usable MVP.

First off, the formatting isn't great. I would love to work with someone better at formatting than I am, because I am quite bad at it!

Second, the rendering of "companies" (privates) is not great, because there is no canonical shortname. One option would be to have a legend where the privates are given a numerical index...but I think the current solution (initials) isn't too bad

There is more information that could go in here for sure...train information, bank information. I'd love to get OR information for companies and value changes for players.

Ideally, we can set parameters for what is necessary to get the PR in, and then what can be done in the future.

Thanks for the help getting ramped up, Toby!

Here's a screenshot of what it currently looks like (def needs more formatting, but usable)
spreadsheet_screenshot

Resolves https://github.com/tobymao/18xx/issues/193


jcoveney:
Hrm, it failed and I'm not sure why. Not sure how to reproduce locally. Not sure why rake would be affected...

jcoveney:
Many thanks to Jen for explaining how to run rubocop locally

jenf:
Could we add the currency symbol where appropriate?

jcoveney:
Great catch -- I dug in and realized that there were already facilities for proper formatting :) nice

jenf:
That build failure is due to my PR being merged, I think @tobymao said he'd fix it.

jcoveney:
Ah, ok. Thanks :) I tried merging master as I imagined it might be related, but I'll let him do his thing!

assets/js/view/spreadsheet.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    def render_title
+      [
+        h(:tr, [
+          h(:th, { style: { width: '20px' } }, ''),
+          h(:th, { attrs: { colspan: @game.players.length } }, 'Players'),
+          h(:th, { attrs: { colspan: 2 } }, 'Bank shares'),
+          h(:th, { attrs: { colspan: 2 } }, 'Prices'),
+          h(:th, { attrs: { colspan: 4 } }, 'Company'),
```
tobymao, line 35: let's change this to corporation

assets/js/view/spreadsheet.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    def render_player_privates
+      h(:tr, [
+        h(:td, 'Privates'),
+        *@game.players.map { |p| render_companies(p) }
+      ])
+    end
+
+    def render_player_cash
```
tobymao, line 102: how come the player section isn't styled?

assets/js/view/spreadsheet.rb
```diff
@@ -0,0 +1,123 @@
+# frozen_string_literal: true
+
+module View
+  class Spreadsheet < Snabberb::Component
+    needs :game
+
+    def render
+      props = {
+         style: {
```
tobymao, line 9: have you tested this on mobile? my guess is that it won't scroll properly, look at map / stock market for guidance on how to make thi mobile friendly

jcoveney: ah, I see what you mean by scrolling properly. Ok, I will add support for that

tobymao:
one more thing, this is a big feature, let's add a simple unit test

assets/js/view/spreadsheet.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    def render_companies(entity)
+      h(:td, entity.companies.map { |c| company_short_name(c.name) }.join(','))
+    end
+
+    # TODO: we need a better way of referring to companies...numerical index?
+    #       we could just make an ad hoc index and show it on the page
+    def company_short_name(name)
+      name.gsub('-', ' ').split(' ').map { |w| w[0] }.join
```
tobymao, line 92: add this to the company and use it if there is no sym

jcoveney: Ah, sounds good. Will add tomorrow

jcoveney:
Re: unit testing, I was thinking about that, I guess I will take a look at how you test other modules and try to put something together. Testing front-end stuff was always a dark art to me, time to open the necronomicon I guess!

michaeljb:
Was just thinking that it would be amazing if you could have this spreadsheet view during a stock round and have buy/sell buttons in the sheet. So with the screenshot at the top, on local4's turn, there would be a "-" or "sell" button in the "AR"/"local4" cell, and a "buy" or "+" button next to the "AR"/"IPO" cell (or it could be next to the par price cell).

When clicking the sell button, a small popup would let you choose how many to cell, and could even give you the numbers of how much you would get, maybe even the new share price after selling.

For unfloated companies, instead of "buy"/"+" next to the IPO share count it could say "par", and clicking on that pops up the par price options.

I could copy this into a new issue, I obviously don't want to hold up merging for this, just thinking of what would be cool to eventually have. :)

jcoveney:
@michaeljb I agree that is a great idea. Can you make a new issue and link it to this? Definitely out of scope for this change, I think. But worth discussing and I agree that something in that realm would be useful, especially in rounds where a lot of selling is going to happen

tobymao:
if it's shown on the main page it needs to be a toggle

jcoveney:
@tobymao I believe I have answered everything you've brought up up to now. I tested it on my phone and it seems to scroll properly.

There are two remaining things:
1. styling. are you ok with the current styling?
2. testing. you mentioned wanting to test this, but I'm looking in the spec and I don't see any tests of say, the corporation tabs. I'm looking at base_spec.rb and slot_spec.rb, but I'm not really seeing how you'd like this to be tested. I'm definitely fine with adding tests, but I'm not quite sure what you want to test. I could imagine if we had some games that we could hook into I could ensure that certain squares or something render in an expected way, but...yeah, I'm just not sure. This is an aspect of front-end development I don't really know well at all.

jcoveney:
I went ahead and spruced up the formatting a little on the player finances

Screen Shot 2020-04-30 at 21 28 58

dfannius:
I am really excited about the spreadsheet view, thanks for doing all this work! Some opinions about the current state:

- I prefer just "Shares" to "Bank Shares"
- It's non-intuitive to have the share counts be called "IPO" and "Market" but the corresponding share prices be called "Par" and "Current". I would change the Prices sub-columns to be "IPO" and "Market" to match the Shares terminology, or even just have a single column for IPO reading "5 @ ¥65" and one for Market reading "0 @ ¥70".
- A space between trains, please ("2 2 2 3" instead of "2223"). Once we start having more interesting train names like 3/5 this will become necessary anyway.
 - Add a little horizontal padding to the header names; right now it looks like `ParCurrentCashTrainsTokensPrivates`. (I really wish that the text here, along with everything else on the site, used a proportional font, which would make this less of an issue, but that ship has sailed.)
 - I don't really think the "Player Finances" line is necessary; it just breaks things up and makes it a little harder to follow the player columns down. (Right now it looks like there is one column of info for "Player" and one for "Finances"...)
 - Right-justify all numbers (so that the 0s line up in ¥60 and ¥230, for example)

assets/js/view/spreadsheet.rb
```diff
@@ -0,0 +1,123 @@
...
+    def render_corporation(corporation)
+      corporation_color =
+        {
+          style: {
+            background: corporation.color,
+            color: '#ffffff'
+          }
+        }
+      props = { style: {} }
+      props[:style]['background-color'] = 'rgba(220,220,220,0.4)' unless corporation.floated?
```
jenf, line 73: It'd be nice to add yellow, brown colour for when the shares are in that area. But that's future work.

jcoveney:
* I prefer just "Shares" to "Bank Shares"

I'm down with this

* It's non-intuitive to have the share counts be called "IPO" and "Market" but the corresponding share prices be called "Par" and "Current". I would change the Prices sub-columns to be "IPO" and "Market" to match the Shares terminology, or even just have a single column for IPO reading "5 @ ¥65" and one for Market reading "0 @ ¥70"

Par is pretty standard terminology in these games. That said, I'm happy making it IPO and Market to be consistent. I don't want to combine info in one cell b/c who knows where things will go, but I'm open to arguments

* A space between trains, please ("2 2 2 3" instead of "2223"). Once we start having more interesting train names like 3/5 this will become necessary anyway.

Hm, yes

* Add a little horizontal padding to the header names; right now it looks like ParCurrentCashTrainsTokensPrivates. (I really wish that the text here, along with everything else on the site, used a proportional font, which would make this less of an issue, but that ship has sailed.)

I'll look into it...my CSS is pretty non-existent but it shouldn't be too hard...

* I don't really think the "Player Finances" line is necessary; it just breaks things up and makes it a little harder to follow the player columns down. (Right now it looks like there is one column of info for "Player" and one for "Finances"...)

I like having it. I guess it's a matter of taste, I'm curious what others think. In the future I am hoping to add player financial history, which will be another subheading

* Right-justify all numbers (so that the 0s line up in ¥60 and ¥230, for example)

in practice, justifying was uglier in my experience...that would likely change with more CSS done eg tweaking the space between stuff etc


I pushed a version with some of the changes, though none of the CSS ones. I really would love to defer the CSS side of things to someone who knows CSS well and is a designer..that side of things is not my forte. That said, if we want to decide what the aesthetic requirements to get this in (and then aspiring designers can iterate on it while I add features ;))

Screen Shot 2020-04-30 at 23 26 52

PR #219, opened 2020-04-29
1846

kelsin:


lib/engine/game/base.rb
```diff
@@ -251,9 +251,17 @@ def purchasable_companies
       private
 
       def init_bank
+        return Bank.new(self.class::BANK_CASH[players.size], log: @log) if self.class::BANK_CASH.is_a?(Hash)
```
tobymao, line 254: it might just be cleaner to do 

cash = self.class::BANK_CASH
cash = cash[players.size] if cash.is_a?(Hash)

lib/engine/game/base.rb
```diff
@@ -251,9 +251,17 @@ def purchasable_companies
...
+
         Bank.new(self.class::BANK_CASH, log: @log)
       end
 
+      def init_cert_limit
+        return self.class::CERT_LIMIT[players.size] if self.class::CERT_LIMIT.is_a?(Hash)
```
tobymao, line 260: maybe same pattern here?

PR #220, opened 2020-04-29
Make Pass be more descriptive (Pass for real passes, Don't or Done otherwise)

### LINKED ISSUE #63: Use Done instead of Pass in appropriate places
Currently, the button that you push to indicate that you (player or corporation) are done performing actions is labeled Pass, but this should be reserved for cases when you are really doing nothing at all; otherwise the button should be called "Done" (or maybe something even more specific like "Lay no track") so as not to imply that you're forfeiting your entire turn.

jenf:
Pass has a specific game meaning in the initial stock round and in stock rounds.

Passing when in the operating round moves to the next step.

Use 'Pass' when this has the specific in game meaning
"Don't" when the user wishes to skip to the next action
"Done" when the user has taking an action in this step (Done Buying Trains)

Resolves #63

### DIFF
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -61,7 +61,7 @@ module Engine
         @bankrupt = false
 
         @step = self.class::STEPS.first
-        @lastactionstep = self.class::STEPS.last
+        @last_action_step = self.class::STEPS.last
         @current_routes = []
 
         payout_companies
@@ -82,7 +82,7 @@ module Engine
 
       def pass_description
         # If the user is continuing an action change the phrasing from Don't to Done
-        if @step == @lastactionstep
+        if @step == @last_action_step
           return self.class::ACTIVE_PASS_DESCRIPTION[@step] if self.class::ACTIVE_PASS_DESCRIPTION.include?(@step)
         elsif self.class::INACTIVE_PASS_DESCRIPTION.include?(@step)
           return self.class::INACTIVE_PASS_DESCRIPTION[@step]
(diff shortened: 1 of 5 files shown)


lib/engine/round/operating.rb
```diff
@@ -47,6 +61,7 @@ def initialize(entities, game:, round_num: 1)
         @bankrupt = false
 
         @step = self.class::STEPS.first
+        @lastactionstep = self.class::STEPS.last
```
tobymao, line 64: underscores

lib/engine/round/operating.rb
```diff
@@ -65,6 +80,16 @@ def description
         self.class::STEP_DESCRIPTION[@step]
       end
 
+      def pass_description
+        # If the user is continuing an action change the phrasing from Don't to Done
+        if @step == @lastactionstep
```
tobymao: i don't understand how this works.

isn't last action step always company? you're using the class definition

why would active pass descriptions have multiple things?

tobymao: ahhhh never mind you set it on line 335

jenf: step is :track and the like

tobymao: how about

self.class::ACTIVE_PASS_DESCRIPTION[@step == @last_action_step ? @step : nil] || 
  self.class::INACTIVE_PASS_DESCRIPTION[@step] ||
  'Pass'

jenf: Ahh, ruby experience shines :)

PR #221, opened 2020-04-29
More informative log during auction (fixes #191)

### LINKED ISSUE #191: When a private buy is automatic (cascading), log should reflect that
This matters because it affects priority, but in the current interface it can be hard to figure out if the buying was automatic or not.

dfannius:
This was the easiest way to implement this given the current auction architecture.

### DIFF
--- a/lib/engine/round/auction.rb
+++ b/lib/engine/round/auction.rb
@@ -159,7 +159,7 @@ module Engine
         bids = @bids[company]
         bids.reject! { |b| b.entity == entity }
         bids << bid
-        @bidders[company] << entity unless @bidders[company].include?(company)
+        @bidders[company] |= [entity]
 
         @log << "#{entity.name} bids #{@game.format_currency(price)} for #{bid.company.name}"
       end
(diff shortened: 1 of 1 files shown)


lib/engine/round/auction.rb
```diff
@@ -157,6 +159,7 @@ def add_bid(bid)
         bids = @bids[company]
         bids.reject! { |b| b.entity == entity }
         bids << bid
+        @bidders[company] << entity unless @bidders[company].include?(company)
```
tobymao, line 162: i think you can do 

@bidders[company |= entity

lib/engine/round/auction.rb
```diff
@@ -166,7 +169,9 @@ def buy_company(player, company, price)
         player.companies << company
         player.spend(price, @bank)
         @companies.delete(company)
-        @log << "#{player.name} buys #{company.name} for #{@game.format_currency(price)}"
+        only_bidder = @bidders[company].length == 1
```
tobymao, line 172: i like to always use size instead of length

PR #222, opened 2020-04-29
feat: city background boxes for sizes 4-6

kelsin:


assets/js/view/part/city.rb
```diff
@@ -87,7 +103,7 @@ def render_part
 
         children = []
 
-        children << render_box(slots.size) if (2..3).include?(slots.size)
+        children << render_box(slots.size) if (2..6).include?(slots.size)
```
tobymao, line 106: this should be

slot.size.between(2, 6)

kelsin: Right now getting: `undefined method `between' for 1`

kelsin: oh fuck, ?, fixing

PR #224, opened 2020-04-29
Offboard centering

kelsin:


assets/js/view/part/multi_revenue.rb
```diff
@@ -15,19 +15,30 @@ class MultiRevenue < Snabberb::Component
...
+          {
+            text: text,
+            width: text.size * 13,
+            color: phase == :diesel ? :gray : phase
+          }
+        end
 
-          width = text.size * 13
+        # Compute total width of rectangles so we can center
+        total_width = computed_revenues.inject(0) do |width, revenue|
```
tobymao, line 34: don't use inject, i think this is equivalent to reduce?

you can also just do copmute_revenues.sum { __ } instead

kelsin: Moving to `sum`, yeah `inject` and `reduce` are aliases :) I just thought pure ruby would want `inject` lol

PR #225, opened 2020-04-29
feat: implementing double dit tiles

kelsin:


assets/js/view/part/town_dot.rb
```diff
@@ -6,11 +6,37 @@ module View
   module Part
     class TownDot < Base
       needs :color, default: 'black'
+      needs :tile
 
       def preferred_render_locations
+        if @tile.towns.size > 1
+          return [
```
tobymao, line 13: let's make this a constant

assets/js/view/part/towns.rb
```diff
@@ -17,7 +17,10 @@ class Towns < Snabberb::Component
       def render
         @tile.towns.map do |town|
           if @tile.lawson?
-            h(Part::TownDot, region_use: @region_use, color: color_for(town))
+            h(Part::TownDot, tile: @tile, region_use: @region_use, color: color_for(town))
+          elsif @tile.paths.empty?
+            # This town is all alone, no routes to it
+            h(Part::TownDot, tile: @tile, region_use: @region_use, color: color_for(town))
```
tobymao, line 23: this looks exactly the same, how about tile.lawson? || empty

kelsin: of course, lol wasn't the same during debugging lol

assets/js/view/part/town_rect.rb
```diff
@@ -10,60 +10,146 @@ class TownRect < Base
...
-        (-height / 2.0) +
-          if diff == 1
-            43.5
-          elsif diff == 2
-            20.575
-          elsif diff == 3
-            0
-          end
+      def track_type
+        return @track_type if defined? @track_type
```
tobymao, line 45: let's do

@track_type ||= begin 
....
end

assets/js/view/part/town_rect.rb
```diff
@@ -10,60 +10,146 @@ class TownRect < Base
...
+                      else
+                        :sharp
+                      end
       end
 
-      def render
-        edge_a, edge_b = @edges
-        edge_a, edge_b = edge_b, edge_a if edge_b < edge_a
+      def position_angle
+        return @position_angle if defined? @position_angle
```
tobymao, line 60: let's use the @x ||= begin pattern

PR #226, opened 2020-04-30
Add game name and round info to game page title

### LINKED ISSUE #157: Tab title should be updated with game name (at very least)
When you enter a game, the tab doesn't change. So if I'm in 5 games in 5 tabs, they will all just say "18xx.games." It'd be ideal if it said the game name and potentially the SR/OR and current player.

Thanks for the great work!

sparr:
Changes "18xx.games" to "gamedescription - roundtype roundnum | 18xx.games" when viewing a game page

Needs testing.

RESOLVES #157

### DIFF
--- a/assets/js/view/game.rb
+++ b/assets/js/view/game.rb
@@ -157,18 +157,18 @@ module View
     end
 
     def round_info
-      name = @round.class.name.split(':').last
-      num = @round.operating? ? "#{@game.turn}.#{@round.round_num}" : @game.turn
+      name = @game.round.class.name.split(':').last
+      num = @game.round.operating? ? "#{@game.turn}.#{@game.round.round_num}" : @game.turn
       [name, num]
     end
 
     def render_round
       name, num = round_info
-      h(:div, { style: { 'font-weight': 'bold' } }, "#{name} Round #{num} - #{@round.description}")
+      h(:div, { style: { 'font-weight': 'bold' } }, "#{name} Round #{num} - #{@game.round.description}")
     end
 
     def render_action
-      case @round
+      case @game.round
       when Engine::Round::Auction
         h(AuctionRound, game: @game)
       when Engine::Round::Stock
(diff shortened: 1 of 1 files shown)


assets/js/view/game.rb
```diff
@@ -151,15 +156,19 @@ def route_anchor
...
-      description = @round.operating? ? "#{@game.turn}.#{@round.round_num}" : @game.turn
-      description = "#{description} - #{@round.description}"
-      h(:div, { style: { 'font-weight': 'bold' } }, "#{name} Round #{description}")
+      name, num = round_info
+      h(:div, { style: { 'font-weight': 'bold' } }, "#{name} Round #{num} - #{@game.round.description}")
     end
 
     def render_action
-      case @round
+      case @game.round
```
tobymao, line 171: why is this changed?

assets/js/view/game.rb
```diff
@@ -170,12 +179,10 @@ def render_action
     end
 
     def render_game
-      @round = @game.round
-
       h('div.game', [
         render_round,
         h(GameLog, user: @user),
-        h(EntityOrder, round: @round),
+        h(EntityOrder, round: @game.round),
```
tobymao, line 185: same

tobymao:
instead of this, can we have a generic views with titles, and then the app can be the only one responsible for changing the title?

assets/js/view/game.rb
sparr: @round was only set in render_game, but new code requires it elsewhere, and I didn't want to leave half the code referring to @round and half to @game.round when they are the same. The alternative is to set @round earlier/higher so it's available everywhere it's needed? (replying to this for posterity; likely this will be addressed when addressing your wider scope requirements later in the PR)

sparr:
@tobymao Good plan, but I think that's going to require more knowledge of the framework than I have right now. Happy to drop this PR and come back later when I'm better equipped.

PR #229, opened 2020-04-30
Render narrow gauge

thebamaman:
This adds the ability to denote gauge types in the tile path notation with 'p=a:*,b:*,g:*' where the types for g are 'b'road, 'n'arrow, and 'd'ual. These are represented in the code with similar ruby symbols.

Because of the need to draw two layers with dual gauge, I added a slight additional outline underneath the main track which produces a small white border around the black track.

This helps distinguish crossings, but does produce a bit of extra layers around complicated shared edges that could be cleaned up by rendering the underlayer first, then the track. That does remove the nice benefit on crossings though. It's almost like you'd need a bit of a switch on the tile definition to drive render ordering.


jenf:
Unit tests need fixing :)

assets/js/view/part/track_curvilinear_half_path.rb
```diff
@@ -84,17 +84,28 @@ def compute_direction(this_exit, other_exit, curvilinear_type)
       end
 
       def render_part
+        uprops = {
```
tobymao, line 87: change the name 

what is a u path?

assets/js/view/part/track_curvilinear_half_path.rb
```diff
@@ -84,17 +84,28 @@ def compute_direction(this_exit, other_exit, curvilinear_type)
...
+
         props = {
           attrs: {
             class: 'curvilinear_path',
             transform: "rotate(#{@rotation})",
             d: SVG_PATH_STRINGS[[@curvilinear_type, @direction]],
-            stroke: @color,
-            'stroke-width' => 8
-          }
+            stroke: @path.gauge != :dual ? @color : @ocolor,
```
tobymao, line 102: how about @path.gauge == :dual ? 

may be more clear

lib/engine/part/path.rb
```diff
@@ -5,11 +5,12 @@
 module Engine
   module Part
     class Path < Base
-      attr_reader :a, :b, :branch, :city, :edges, :junction, :offboard, :stop, :town
+      attr_reader :a, :b, :gauge, :branch, :city, :edges, :junction, :offboard, :stop, :town
 
-      def initialize(a, b)
+      def initialize(a, b, gauge)
```
tobymao, line 10: can also do gauge == :broad instead that way they don't have to pass it in

tobymao:
there's a lot of duplicate code here it seems, is there a way to make this cleaner?

thebamaman:
going to do this in parts, adding just the gauge types to the Engine::Tile, and then add the rendering code in a separate commit.

Should allow you to import tile definitions from kelsin with the gauge codes without breaking...rendering can be done later

PR #230, opened 2020-04-30
Combine multiple of the same train purchase option, with a count

### LINKED ISSUE #228: Option to buy trains from another company lists the same company twice for the same train #
Train 2 - ¥80
Train 2 - from AR
Train 2 - from AR

It should just list skip the second 'Train 2 - from AR'

sparr:
Replaces...

```
Train 2 - from AR [  ]Buy
Train 2 - from AR [  ]Buy
Train 2 - from AR [  ]Buy
Train 3 - from AR [  ]Buy
```

with

```
Train 2 - from AR (has 3) [___] [Buy]
Train 3 - from AR [___] [Buy]
```

RESOLVES #228


jenf:
Changes look good to me :)

assets/js/view/buy_trains.rb
```diff
@@ -95,7 +95,7 @@ def from_depot(depot_trains)
 
     def other_trains(other_corp_trains)
       other_corp_trains.flat_map do |other, trains|
-        trains.map do |train|
+        trains.uniq(&:name).map do |train|
```
tobymao, line 98: how about 

trains.group_by(&:name).map do |train, trains|

"Train #{train.name} ... #{trains.size}"

sparr: Got close with the latest commit

assets/js/view/buy_trains.rb
```diff
@@ -95,7 +95,7 @@ def from_depot(depot_trains)
 
     def other_trains(other_corp_trains)
       other_corp_trains.flat_map do |other, trains|
-        trains.map do |train|
+        trains.group_by(&:name).each_pair.map do |name, same_trains|
```
tobymao, line 98: i don't think you need each pair, and can you rename same_trains to group

assets/js/view/buy_trains.rb
```diff
@@ -111,11 +111,13 @@ def other_trains(other_corp_trains)
 
           buy_train = lambda do
             price = input.JS['elm'].JS['value'].to_i
-            process_action(Engine::Action::BuyTrain.new(@corporation, train, price))
+            process_action(Engine::Action::BuyTrain.new(@corporation, same_trains[0], price))
           end
 
+          count = same_trains.length
```
tobymao, line 117: let's use group.size

assets/js/view/buy_trains.rb
sparr: group_by produces a hash of {name1:trainarray1, name2:trainarray1, ...} which I couldn't find a way to iterate without first turning it into an array with each_pair.
happy to do the rename.

sparr: Well, apparently I can do `.group_by().map()` despite none of the docs I can find explaining why or how that would work. So... done.

PR #232, opened 2020-04-30
Add remaining tiles for 1861

sparr:
I biased toward existing 18xx.games tile rotations, then fwtwr for tiles without obvious existing siblings here. I put the city names as comments, for reference when the maps are laid out.


tobymao:
did you test the all_tiles view?

tobymao:
let's match the rotation here https://github.com/18xx-maker/18xx-maker/blob/master/src/data/tiles/yellow.json

tobymao:
colising in favor of kelsin automatition, thanks

PR #246, opened 2020-05-01
Make exchange of dougo railway easier

### LINKED ISSUE #58: Exchange Dougo Railway (game id 75)
I can't (seem to find out how to) exchange the Dougo Railway private for a Iyo share during SR1.

Edit: I figured it out so it's not a bug. It isn't intuitive though. When I want to exchange I need to:
- go to the companies tab, 
- select Dougo
- go to the game tab
- press the then appeared button to exchange share.

Perhaps this can be made in a more transparent way.

jenf:
Adds a button on Iyo to allow exchange of Dougo for an Iyo IPO share

If it's not a players turn it still allows the exchange but prefixes it with the players name, since you can still exchange out of turn sequence.

Resolves #58


jenf:
The 'other' way of doing it can allow you go over the 60% limit, this should probably be fixed as part of this PR.

jenf:
@tobymao now enforces the limit in the engine. Note there's a potential the someone got past the limits and so there might be an invalid game.

assets/js/view/stock_round.rb
```diff
@@ -73,6 +73,27 @@ def render_ipoed
...
+          children << if company.owner == @current_entity
+                        h(:button, { on: { click: exchange } },
+                          "Exchange #{company.name} for Share")
+                      else
+                        # This can be done outside of a players turn, but make it clear who owns it
+                        h(:button, { on: { click: exchange } },
+                          "#{company.owner.name} exchanges #{company.name} for Share")
+                      end
+        }
+        @game.companies.select(&exchangable_company).each(&add_company_exchange)
```
tobymao, line 95: can you change this to be

exchangable = @games.companies.select do |company|
  company.abilities(:exchange)&:.dig(:corporation) ...
end

exchange.each do |company|
...
end

spec/lib/engine/round/stock_spec.rb
```diff
@@ -36,6 +37,30 @@ module Engine
         expect(subject.can_buy?(corp_0.shares[0])).to be_truthy
       end
 
+      it 'can\'t buy when at cert limit when doing so would gain you the presidency' do
```
tobymao, line 40: i'm not a fan of the single quote back slash, just do ""

assets/js/view/stock_round.rb
jenf: Thanks, wasn't quite sure the best style to express it :)

lib/engine/corporation.rb
```diff
@@ -102,5 +102,10 @@ def corporation?
     def inspect
       "<#{self.class.name}: #{id}>"
     end
+
+    # Is it legal to hold percent shares in this corporation?
+    def holding_ok?(percent)
+      %i[orange brown].include?(share_price&.color) || percent <= 60
```
tobymao, line 108: i'm going to pass in the entity instead of the percent, i'll do this myself

PR #248, opened 2020-05-01
Add a little horizontal spacing to spreadsheet header

dfannius:


assets/js/view/spreadsheet.rb
```diff
@@ -32,6 +32,8 @@ def render_table
     end
 
     def render_title
+      style = { padding: '0 5px' }
```
tobymao, line 35: how about style/props = { style: { padding: '1rem' } }

then below you don't need to do it

Issue #215, opened 2020-04-29
Extend the spreadsheet view to have OR information


jcoveney:
I believe that having the OR history surfaced in the spreadsheet view will be quite useful.

Once we have resolved
https://github.com/tobymao/18xx/issues/193

And merged the eventual form of
https://github.com/tobymao/18xx/pull/210

I will make the changes necessary to get OR history

jcoveney:
I have a PR ready for this once #210 lands

PR #249, opened 2020-05-01
Add company OR history

jcoveney:
Building on the work in https://github.com/tobymao/18xx/pull/210

Closes https://github.com/tobymao/18xx/issues/215

This extends the spreadsheet to have an OR history for each company

In time, we will convert toby to the usefulness of this tab :)

Screen Shot 2020-05-01 at 23 30 12


jcoveney:
Per discussion with toby, will add information for the last set of ORs to the cards

lib/engine/corporation.rb
```diff
@@ -15,7 +15,7 @@ class Corporation
     include Spender
 
     attr_accessor :ipoed, :par_price, :share_price, :tokens
-    attr_reader :companies, :coordinates, :min_price, :name, :full_name, :logo, :trains, :color
+    attr_reader :companies, :coordinates, :min_price, :name, :full_name, :logo, :trains, :color, :or_revenue_history
```
tobymao, line 18: rename to revenue_history

lib/engine/corporation.rb
```diff
@@ -99,6 +100,11 @@ def corporation?
       true
     end
 
+    def push_revenue_history!(revenue, round_num)
```
tobymao, line 103: how about just add_revenue

lib/engine/corporation.rb
```diff
@@ -99,6 +100,11 @@ def corporation?
...
+      @or_revenue_history.push([]) if round_num == 1
```
tobymao, line 104: use << intsead of push

assets/js/view/spreadsheet.rb
```diff
@@ -31,7 +31,24 @@ def render_table
       # TODO: consider adding train availability
     end
 
+    def render_or_history_titles(arbitrary_corporation)
```
tobymao, line 34: render_history(corporation)

assets/js/view/spreadsheet.rb
```diff
@@ -40,6 +57,7 @@ def render_title
           h(:th, { attrs: { colspan: 2 } }, 'Prices'),
           h(:th, { attrs: { colspan: 4 } }, 'Corporation'),
           h(:th, { style: { width: '20px' } }, ''),
+          h(:th, { attrs: { colspan: or_history_titles.length } }, 'OR History'),
```
tobymao, line 60: size

lib/engine/game/base.rb
```diff
@@ -414,6 +414,7 @@ def new_stock_round
 
       def new_operating_round(round_num = 1)
         @log << "-- Operating Round #{@turn}.#{round_num} --"
+        @corporations.reject(&:floated?).each { |c| c.push_revenue_history!(nil, round_num) }
```
tobymao, line 417: hm i'm not sure i like this

what if revenue history was more detailed

jcoveney: Detailed in what way? I'm not sure what you mean. Do you mean if revenue history had turn and round information? Or...?

assets/js/view/spreadsheet.rb
jcoveney: ruby *shakes fist* def will do

assets/js/view/spreadsheet.rb
jcoveney: but that is not what this is. this generates the OR1.1, OR2.1, OR2.2 etc information for the title

it could be render_history_titles though, or something else

lib/engine/corporation.rb
jcoveney: ah, ruby :)

lib/engine/corporation.rb
jcoveney: will do. renaming things -- the bane of dynamically typed languages :)

lib/engine/corporation.rb
jcoveney: 2 hard problems in computer science. will do

assets/js/view/spreadsheet.rb
tobymao: sounds good

lib/engine/game/base.rb
tobymao: can we get around having to inject this fake data? what if the revenue history had more info so this didn't need to happen

jcoveney: It could be a map keyed off of turn and round_num

jcoveney:
Responded to a number of comments and added information to the cards. Note: it's not terribly attractive (I imagine you're sensing a theme). Where do you think it should go?

Screen Shot 2020-05-02 at 00 20 51

lib/engine/game/base.rb
tobymao: yes, maybe let's try that

jcoveney: ok, I'll get to that tomorrow. let me know if there are any big issues :) namely formatting of the cards lol

tobymao:
> Responded to a number of comments and added information to the cards. Note: it's not terribly attractive (I imagine you're sensing a theme). Where do you think it should go?
> 
> Screen Shot 2020-05-02 at 00 20 51

looks really bad. does it fit in the Header? or next to the president? I would label it Last Run: $...

jcoveney:
I was imagining having the information for the last OR (so, 1-3 pieces). If it's just 1 (eg the very last run), then yeah it can be "Last Run" and we can maybe put it next to the president, or right under it. I don't think there is space on the header.

jcoveney:
Current outstanding issues

- Refactor so that the information is map based, so that we don't have to inject nils
- figure out how to get the information on the cards

jcoveney:
Ok, I cleaned up the revenue history, got rid of the nils etc.

I went ahead and removed the information from the card. I really think that can/should be a different issue -- the main issue is not technical (it's not trivial to get the information), it's more of a design issue.

tobymao:
i'm going to hold off on this a bit because i'm reworking routes and revenues

tobymao:
well on second thought maybe we should add it now because people want it.

i'd like for it to be on the card, can you just put it to the left of the president, and say like 'Last Revenue'

assets/app/view/spreadsheet.rb
```diff
@@ -31,15 +31,44 @@ def render_table
...
 
+    def or_history(corporations)
+      # It looks like the set class isn't available, so we use this as a proxy
+      or_hist = {}
+      corporations.each do |c|
+        c.revenue_history.keys.each { |x| or_hist[x] = true }
+      end
+      or_hist_list = []
+      or_hist.keys.each { |x| or_hist_list << x }
+      or_hist_list.sort
```
tobymao, line 42: how does this work with unfloated companies?

assets/app/view/spreadsheet.rb
```diff
@@ -31,15 +31,44 @@ def render_table
       # TODO: consider adding train availability
     end
 
+    def or_history(corporations)
+      # It looks like the set class isn't available, so we use this as a proxy
+      or_hist = {}
+      corporations.each do |c|
+        c.revenue_history.keys.each { |x| or_hist[x] = true }
+      end
```
tobymao, line 39: ok i see what this is doing now, a bit convoluted, you can just do

corporations.flat_map { |c| c.revenue_history.keys }.uniq.sort

jcoveney:
I should be able to get it updated tomorrow (China time)

I'll try putting it on the card next to president.

assets/app/view/spreadsheet.rb
jcoveney: Ah Ruby. "uniq". Of course. My Google Fu failed me earlier -- thanks!

jcoveney:
Ok, was able to get it added to a card in what looks to be a non-hideous way. Take a look!

Screen Shot 2020-05-03 at 16 46 32

Issue #235, opened 2020-04-30
Expand Spreadsheet view to show when are in the yellow/brown


jenf:
Currently it shows grey if they are not floated it'd be nice to add yellow and brown for when they're in that area on the stack market.

jcoveney:
I went ahead and put a PR together which does this once https://github.com/tobymao/18xx/pull/210 lands. I could put it there, but I want to get that in, since it is the basis of multiple other PRs I have, and will be useful etc etc

PR #250, opened 2020-05-01
Add spreadsheet color for when a company is in the yellow, etc

jcoveney:
Builds on the work in https://github.com/tobymao/18xx/pull/210

Closes https://github.com/tobymao/18xx/issues/235

A fairly simple patch. I should say that I'm open to how we should handle coloring...I tried to pull it out, but I wasn't sure how classes in the view should refer to each other. Maybe COLOR_MAP should go into the engine stock_market class...or ideally it would be a static module method, but I wasn't sure if you like that, or how to do it.


jcoveney:
Forgot to upload a picture

Screen Shot 2020-05-01 at 23 36 34

dfannius:
I personally would prefer the coloring just on the Market Price cell.

tobymao:
yea let's do that, just coloring there

jcoveney:
Updated coloring

Screen Shot 2020-05-02 at 00 44 55

assets/js/view/spreadsheet.rb
```diff
@@ -70,7 +72,21 @@ def render_corporation(corporation)
           }
         }
       props = { style: {} }
-      props[:style]['background-color'] = 'rgba(220,220,220,0.4)' unless corporation.floated?
+
+      market_props = { style: {} }
+
+      if corporation.floated?
```
tobymao, line 78: corpoaration.floated? && (color = StockMarket::COLOR_MAP[corporation.share_price.color])

jcoveney: I don't think this is correct. If you look at the else, it greys out the line for unfloated companies. so if you add another if here, it will grey out cases where there is not a color is well..which is not correct, as certain floated companies do not have colors set yet

tobymao: ok then

unless corporation.floated?
 props[:style][bgc] = 
elsif (color = ...)

jcoveney:
Ok, refactored per toby's suggestion! :)

PR #252, opened 2020-05-01
/all_tiles page enhancements

michaeljb:
- show tile manifest for each game
- show all map hexes for each game; when a particular "tile" is used by too many
  hexes, truncate its name
- show all hexes with private blockers and corporation reservations
- if a tile is listed in a game's manifest but is not defined in Engine::Tile,
  render a blank tile with a "TODO" label (e.g., tiles 53 and 61 for 1830)
- a lot of terrain hexes are stil rendered repeatedly, e.g., the 120 cost
  mountain in 1830; this could be fixed by defining a 'mtn120' tile in
  Engine::Tile and updating the game definition to refer to that name instead of
  a tile code

update map hex view to show all hexes used in a game's map

when a single "tile" is used by many hexes, truncate the name to fit in the box;
would be nice to have some way to see the full list, punting on that for now

18Chesapeake example
1889 example
1830 example


assets/js/view/all_tiles.rb
```diff
@@ -32,42 +33,99 @@ def map_hexes_for(game_class)
...
+                  location_name: location_names[coord]
+                )
+              end
 
             # add private companies that block tile lays on this hex
             blocker = companies.find do |c|
               c.abilities(:blocks_hex)&.dig(:hex) == coord
             end
-            tile.add_blocker!(blocker) unless blocker.nil?
+            unless blocker.nil?
```
tobymao, line 55: if blocker

michaeljb: 689a46b

Issue #39, opened 2020-04-15
Make map hexes non transparent when it's not laying track phase


tobymao:
make sure to handle tile laying privates

Tin-Whisker:
Overall the map is much too transparent for my tastes.

Anvil-X:
I cannot see any value to having the map visibility reduced ever

tobymao:
@Anvil-X it helps tremendously when figuring out where you can lay your track. This is especially useful for new players.

Anvil-X:
I get that you can use technology to make a game easier to learn or play but this is not an example of that.  If you honestly think that players need help identifying the big fat black lines at the edges of tiles, then make some kind of highlighting there.  Making the map unreadable is not helpful.

Bastwood:
As a newcomer to 1889 and playing my very first game of it with this system, my absolutely biggest gripe was the "dimmed" map board. I'd never taken a close look at the game board, and initially finding just about anything on the board was an exercise in frustration and eye-squinting. I don't really care where I can lay my track if I'm having a hard time seeing the board in the first place. I'm very much in the camp that'd prefer the map to have zero transparency at all times.

PR #255, opened 2020-05-02
Developer convenience page - /map/<game_title>

michaeljb:
Map Example

Bad Game Title Example

1846 Example

1830 Example

### DIFF
--- a/assets/app/view/hex.rb
+++ b/assets/app/view/hex.rb
@@ -29,6 +29,7 @@ module View
     needs :tile_selector, default: nil, store: true
     needs :show_grid, default: false, store: true
     needs :role, default: :map
+    needs :opacity, default: nil
 
     def render
       children = [h(:polygon, attrs: { points: Lib::Hex::POINTS })]
@@ -48,7 +49,7 @@ module View
           transform: transform,
           fill: COLOR.fetch(@tile&.color, 'white'),
           stroke: 'black',
-          opacity: layable || %i[tile_selector tile_page].include?(@role) ? 1.0 : 0.3,
+          opacity: @opacity || (layable || %i[tile_selector tile_page].include?(@role) ? 1.0 : 0.3),
           cursor: clickable ? 'pointer' : nil,
         },
       }
--- a/assets/app/view/hex.rb
+++ b/assets/app/view/hex.rb
@@ -39,8 +39,8 @@ module View
 
       children << h(Tile, tile: @tile) if @tile
       children << h(View::TriangularGrid) if @show_grid
-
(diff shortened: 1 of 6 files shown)


assets/app/view/map_page.rb
```diff
@@ -0,0 +1,32 @@
...
+require 'view/map'
+
+module View
+  class MapPage < Snabberb::Component
+    needs :route
+
+    ROUTE_FORMAT = %r{/map/(.*)}.freeze
+
+    def render
+      game_title = @route.match(ROUTE_FORMAT).captures.first
```
tobymao, line 14: can't you just do match(R)[1]

assets/app/view/map_page.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      game_class = Engine::GAMES_BY_TITLE[game_title]
+
+      return h(:p, "Bad game title: #{game_title}") if game_class.nil?
```
tobymao, line 18: unless game_class

assets/app/view/map.rb
```diff
@@ -10,6 +10,7 @@ class Map < Snabberb::Component
     needs :game
     needs :tile_selector, default: nil, store: true
     needs :selected_company, default: nil, store: true
+    needs :opacity, default: nil
```
tobymao, line 13: i don't like this, but maybe it's a bigger change

while you're at it, instead of doing this

can you just fix this ? https://github.com/tobymao/18xx/issues/39

this should make the maps already non transparent and fix this issue

michaeljb: Will take a crack at that tomorrow (probably late)

michaeljb: e8c86c8 - still need to do privates

tobymao:
feel free to merge if this works for privates

assets/app/view/hex.rb
```diff
@@ -89,5 +90,16 @@ def on_hex_click(event)
         @tile_selector.tile = @tile
       end
     end
+
+    def opacity(layable)
+      corp_track_laying = @round.is_a?(Engine::Round::Operating) && (@round.step == 'track')
+      company_track_laying = @round.is_a?(Engine::Round::Special)
```
michaeljb, line 96: @tobymao admittedly I'm not sure if `company_track_laying` should have another check like `@round.can_lay_track?`, but this seemed to work correctly for all privates, i.e., it faded out most of the map after clicking on a private that could lay track, and didn't fade the map after clicking on the Dougo to exchange for a share of IR.

PR #257, opened 2020-05-02
Bold messages to make it clearer, keep messages when undoing

### LINKED ISSUE #185: Suggestion: in-game chat either bold/different colour then in-game log
Finding the in-game chat quite handy and useful. If possible it would be nice to have it different font/typeface than the in-game log text.

Just a nice to have/suggestion - definitely not a complaint! This site is coming along so nicely, great work!
Rob Thompson

jenf:
Implements two of the suggestions
* Bold messages
* Retain messages when a user presses the undo button.

resolves #185

### DIFF
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -26,7 +26,11 @@ module View
       }
 
       lines = @log.reverse.map do |line|
-        h(:div, { style: { transform: 'scaleY(-1)' } }, line)
+        if line.is_a? String
+          h(:div, { style: { transform: 'scaleY(-1)' } }, line)
+        elsif line.is_a? Engine::Action::Message
+          h(:div, { style: { 'font-weight': 'bold', transform: 'scaleY(-1)' } }, "#{line.entity.name}: #{line.message}")
+        end
       end
 
       h(:div, props, lines)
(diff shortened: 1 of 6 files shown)


assets/app/view/log.rb
```diff
@@ -26,7 +26,11 @@ def render
       }
 
       lines = @log.reverse.map do |line|
-        h(:div, { style: { transform: 'scaleY(-1)' } }, line)
+        if line.is_a? String
```
tobymao, line 29: always use parens

lib/engine/game/base.rb
```diff
@@ -225,7 +225,8 @@ def clone(actions)
       end
 
       def rollback
-        clone(@actions[0...-1])
+        action_to_remove = @actions.rindex { |x| !x.keep_on_rollback? }
+        clone(@actions[0, action_to_remove] + @actions[action_to_remove + 1..-1])
```
tobymao, line 229: @actions.slice...

tobymao:
have you tested this on a running instance? i think this will break the sorting in the database

actions have IDS and those are persisted. i think doing this is more complicated since you need to preserve relative order of the messages. not sure we want to do this.

jenf:
Your right, i'd tried undoing and it says it's the wrong turn.

But I think making messages persist is important especially in async games where you can't contact people another way.

jenf:
Closing as undo will be more complicated, separate PR for bold raised

PR #275, opened 2020-05-03
Spreadsheet OR history tweaks

dfannius:


assets/app/view/spreadsheet.rb
```diff
@@ -48,7 +48,17 @@ def render_history(corporation)
         []
       else
         or_history(@game.corporations).map do |x|
-          h(:td, hist[x].nil? ? '' : @game.format_currency(hist[x]))
+          if hist[x].nil?
```
tobymao, line 51: can we reverse this, i don't really like explicit calls to nil,

so if hist[x] ...
else

also, either in this PR or another, could you make the last RUN an ABS as well in the corporation card?

PR #277, opened 2020-05-03
Display corporation logos in OR order

dfannius:


assets/app/view/entity_order.rb
```diff
@@ -16,15 +17,38 @@ def render
...
 
-      h(:div, { style: style }, "#{entity.name}#{owner}")
+      h(:div, { style: style }, children)
     end
 
     props = {
       style: {
         margin: '1rem 0 1rem 0',
-        'font-size': 'large',
+        'font-size': '1.rem',
```
tobymao, line 50: 1 rem?

assets/app/view/entity_order.rb
```diff
@@ -8,6 +8,7 @@ def render
       style = {
         display: 'inline-block',
         'margin-right': '1rem',
+        'white-space': 'nowrap'
```
tobymao, line 11: is the entity order scrollable now? i've been wanting to do that

assets/app/view/entity_order.rb
dfannius: Stray keystroke, thanks for the catch. Fixed.

assets/app/view/entity_order.rb
dfannius: I'm not sure what you mean by scrollable; I just did this so that the browser wouldn't try to wrap at the space between the company acronym and the president.

tobymao: on an iphone, the entity order goes to multiple lines annoyingly, can you make it always one line and scroll on smaller phones, also centering the entity that's active?

dfannius: I don't know how to do that offhand. Maybe make that a new issue and see if someone with the right knowledge picks it up? I can also look into it eventually if no one grabs it.

tobymao: ok

PR #278, opened 2020-05-03
Improve train and phase information and cert limit

### LINKED ISSUE #258: Show rusting schedule of trains and which phase they cause
Extend the current trains tab to add

Rusting schedule
Phase
Number of ORs

Also add the certificate limits for the current player count.

jenf:
This adds more information about train rusting, diesel upgrades and the phase information

Screenshot from 2020-05-03 22-39-35

Fixes #258


lib/engine/game/base.rb
```diff
@@ -85,7 +85,6 @@ class Base
             '4' => 300,
             '5' => 300,
             '6' => 300,
-            'D' => 300,
```
tobymao, line 88: why is this removed? you can trade a D for a D

assets/app/view/train_roster.rb
```diff
@@ -13,7 +13,7 @@ def render
     end
```
tobymao, line 13: this class and tab should be renamed

assets/app/view/train_roster.rb
```diff
@@ -27,32 +27,115 @@ def render_body
...
+
+      current_phase = @game.phase.current
+      rows = @game.phase.phases.map do |phase|
+        phase_color = Array(phase[:tiles]).last
+        phase_props = {
+          style: {
+            padding: '0 1rem'
+          }
+        }
+        if View::Part::MultiRevenue::COLOR.include? phase_color
```
tobymao, line 59: you can probably just do Part::MultiRevenue

tobymao: also use parens

assets/app/view/train_roster.rb
```diff
@@ -27,32 +27,115 @@ def render_body
...
+          m = View::Part::MultiRevenue::COLOR[phase_color].match(/#(..)(..)(..)/)
```
tobymao, line 60: i've seen this pattern before, can this be refactored?

assets/app/view/train_roster.rb
```diff
@@ -27,32 +27,115 @@ def render_body
...
+          phase_props[:style]['background-color'] = "rgba(#{m[1].hex},#{m[2].hex},#{m[3].hex},0.4)"
+        end
+
+        buy_text = if phase[:buy_companies]
+                     'Can Buy'
```
tobymao, line 65: seems brittle

assets/app/view/train_roster.rb
```diff
@@ -27,32 +27,115 @@ def render_body
...
+
     def upcoming_trains
       td_props = {
         style: {
           padding: '0 1rem'
         }
       }
 
+      rust_schedule = {}
+      @depot.trains.group_by(&:name).each do |name, trains|
```
tobymao, line 107: i feel like this could be written better

lib/engine/game/base.rb
tobymao: i guess i'm wrong, i've removed this in master

assets/app/view/train_roster.rb
jenf: Done

assets/app/view/train_roster.rb
jenf: Yeah, i'd agree but i'm not sure the best rubyism. If it wasn't creating a list of lists .to_h would make sense, but doesn't in this.

assets/app/view/train_roster.rb
jenf: perhaps, but i'm unsure how to improve it.

assets/app/view/train_and_phase_roster.rb
```diff
@@ -0,0 +1,172 @@
...
+          margin: '0 -1rem'
+        } }, [
+        render_body
+      ])
+    end
+
+    def render_body
+      children = [h(:div, [cert_limit, upcoming_trains])]
+
+      unless @depot.discarded.empty?
```
tobymao, line 22: if @depot.discarded.any?

PR #280, opened 2020-05-03
Only tint price in spreadsheet if it affects cert count (fixes #271)

### LINKED ISSUE #271: Spreadsheet: market price shouldn't color par boxes
It's nice that yellow/orange/brown shares are now colored, but the code is so general that it colors the cell red when the token happens to be in one of the initial-par boxes, which doesn't really provide any useful information and is distracting.

dfannius:


assets/app/view/spreadsheet.rb
```diff
@@ -112,9 +112,13 @@ def render_corporation(corporation)
...
       if !corporation.floated?
         props[:style]['background-color'] = 'rgba(220,220,220,0.4)'
-      elsif (color = StockMarket::COLOR_MAP[corporation.share_price.color])
-        m = color.match(/#(..)(..)(..)/)
-        market_props[:style]['background-color'] = "rgba(#{m[1].hex},#{m[2].hex},#{m[3].hex},0.4)"
+      else
+        share_color = corporation.share_price.color
+        if @game.class::CERT_LIMIT_COLORS.include?(share_color)
+          color = StockMarket::COLOR_MAP[share_color]
+          m = color.match(/#(..)(..)(..)/)
```
tobymao, line 119: is there a better way to do this

dfannius: If you mean looking up `share_color` in `CERT_LIMIT_COLORS`, that seems like a pretty reasonable way to find out if a share counts toward the cert limit. I could abstract it into a method in `Game` or something.

If you mean parsing the `StockMarket`'s color and then recalculating it so we can change the alpha, that was there already. I could look into it though.

If you mean a third thing let me know!

tobymao: @jenf has now extracted the match method in lib::Color, so yes i meant the first

dfannius: Better now. Also added a question mark to `counts_for_limit?` since it's a predicate.

PR #307, opened 2020-05-06
New player card design

dfannius:
New format, from top to bottom:
- Just player name in top stripe (like corporation cards) and turn order
- Stats (cash, value, certs) on left, shares on right (if no shares owned, just stats, centered)
- Privates, if any

Another new feature is that the cert display is in red if the player's over the limit.

Goals of this design:
 - Remove clutter (both stats and field names) from header, so the eye is immediately drawn to the player name
 - Avoid tons of blank space once the privates have been bought in
 - Don't need to truncate private names

The one thing I'm still not totally happy with is the 1st, 2nd, 3rd denoting turn order, which is easily misinterpretable as "1st place", "2nd place", etc. If "Priority Deal" or "Priority" is not too confusing to newer players, I'd prefer to just label a single player with that. It could go either in the top stripe as now, or with the stats on the left.

### DIFF
--- a/assets/app/view/player.rb
+++ b/assets/app/view/player.rb
@@ -21,54 +21,27 @@ module View
         card_style['background-color'] = '#dfd'
       end
 
-      h(:div, { style: card_style }, [
-        render_header,
+      divs = [
+        render_title,
         render_body,
-      ])
+      ]
+
+      divs << render_companies if @player.companies.any?
+
+      h(:div, { style: card_style }, divs)
     end
 
-    def render_header
-      header_style = {
-         margin: '-0.5em',
-         'text-align': 'center',
-         'white-space': 'nowrap',
-         'background-color': 'lightgray',
+    def render_title
+      title_style = {
+        'background-color': @game.round.can_act?(@player) ? '#9b9' : 'lightgray',
+        'text-align': 'center',
+        color: 'black',
+        padding: '0.5rem 0px',
+        'font-weight': 'bold',
+        margin: '-0.5rem -0.5rem 0 -0.5rem',
       }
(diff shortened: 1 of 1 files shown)


dfannius:
Removed turn order from header and added priority marker to stats section after Slack conversation.

assets/app/view/player.rb
```diff
@@ -21,99 +21,124 @@ def render
...
+            h(:td, td_props, 'Value'),
+            h(:td, td_props, @game.format_currency(@player.value)),
+          ]),
+          h(:tr, [
+            h(:td, td_props, 'Certs'),
+            h(:td, td_cert_props, "#{num_certs}/#{cert_limit}"),
+          ]),
+        ]
+
+      if @game.players.find_index(@player).zero?
```
tobymao, line 106: indent

assets/app/view/player.rb
```diff
@@ -21,99 +21,124 @@ def render
...
+        trs << h(:tr, [
+                    h(:td, { attrs: { colspan: '2' } }, 'Priority deal'),
```
tobymao, line 108: here

assets/app/view/player.rb
```diff
@@ -21,99 +21,124 @@ def render
...
-        }
-      }
-
-      h(:table, props, [
-        h(:tr, [
-          h(:th, { style: { width: '20px' } }, ''),
-          h(:th, 'Corp'),
-          h(:th, 'Share'),
+      h(:div, div_props, [
+        h(:table, props, [
```
tobymao, line 139: just do table, props, shares

Issue #292, opened 2020-05-05
Surface who has sold what stock in a given SR


jcoveney:
While this information is in the log, it can be tough to scroll up and make sense of things, and there's no reason we can't surface this.

That said, I'm not sure what the right UX here is

dfannius:
Same as #265. I've been thinking about this but I'm happy to give @lbrande first crack at it.

lbrande:
Now I've set up the development environment and started to work on this. Could someone assign me?

dfannius:
Note that there's a redesign of the player card incoming with #307, so if you want to display this information there, you should probably wait for that to get merged in.

lbrande:
You can unassign me from this issue now. @dfannius will fix the rest of this issue.

dfannius:
I'll unassign you when the PR is merged.

lbrande:
Unassigned my self as it now works in spreadsheet view.

dfannius:
Putting this info on the corp card is implemented with c9451e1 (share amounts are written in red). I think that's sufficient.

PR #311, opened 2020-05-07
Spreadsheet now shows which players have sold shares of which companies.

lbrande:
Solves spreadsheet part of #292.

### DIFF
--- a/assets/app/view/spreadsheet.rb
+++ b/assets/app/view/spreadsheet.rb
@@ -120,7 +120,9 @@ module View
       h(:tr, props, [
         h(:th, corporation_color, corporation.name),
         *@game.players.map do |p|
-          h(:td, p.num_shares_of(corporation).to_s + (corporation.president?(p) ? '*' : ''))
+          sold_props = { style: {} }
+          sold_props[:style]['background-color'] = 'rgba(225,0,0,0.4)' if @game.round.did_sell?(corporation, p)
+          h(:td, sold_props, p.num_shares_of(corporation).to_s + (corporation.president?(p) ? '*' : ''))
         end,
         h(:td, corporation.num_shares_of(corporation)),
         h(:td, @game.share_pool.num_shares_of(corporation)),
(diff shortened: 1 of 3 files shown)


assets/app/view/spreadsheet.rb
```diff
@@ -120,7 +120,13 @@ def render_corporation(corporation)
       h(:tr, props, [
         h(:th, corporation_color, corporation.name),
         *@game.players.map do |p|
-          h(:td, p.num_shares_of(corporation).to_s + (corporation.president?(p) ? '*' : ''))
+          style_props =
+            if (@game.round.is_a? Engine::Round::Stock) && @game.round.players_sold[p][corporation] == :prev
```
tobymao, line 124: add parens around method calls

tobymao: let's not do this, instead let's add a method to game base that tells you when something was sold, then you don't need to check if it's a stock round

something like, game.round.did_sell?(p, corporation)

lib/engine/round/base.rb
```diff
@@ -217,6 +217,11 @@ def can_gain?(share, entity)
         corporation.holding_ok?(entity, share.percent) &&
         (!corporation.counts_for_limit || entity.num_certs < @game.cert_limit)
       end
+
+      def did_sell?(corporation, entity)
+        is_a?(Engine::Round::Stock) &&
```
tobymao, line 222: we should not import round stock in here as it might induce a circular dependency

intsead this should return false, and you should override did_sell? in round/stock

lib/engine/round/base.rb
```diff
@@ -217,6 +217,11 @@ def can_gain?(share, entity)
...
+        (players_sold[entity][corporation] == :now || players_sold[entity][corporation] == :prev)
```
tobymao, line 223: can't you just do player_sold[entity][corporation]

lib/engine/round/stock.rb
```diff
@@ -8,7 +8,7 @@
 module Engine
   module Round
     class Stock < Base
-      attr_reader :last_to_act, :share_pool, :stock_market
+      attr_reader :last_to_act, :share_pool, :stock_market, :players_sold
```
tobymao, line 11: delete this

PR #312, opened 2020-05-07
18xx maker import

kelsin:
WIP: Just so that @tobymao can see what I'm working towards

Do Not Merge


lib/engine/game/g_1889.rb
```diff
@@ -1,145 +1,12 @@
...
-      ].freeze
-
-      STARTING_CASH = {
-        2 => 420,
-        3 => 420,
-        4 => 420,
-        5 => 390,
-        6 => 390,
-      }.freeze
+      include ::Engine::Config::Game::G1889
```
tobymao, line 9: do you need the ::Engine, it can't just be Config::Game?

migrate/0004_modify_rotations.rb
```diff
@@ -0,0 +1,52 @@
...
+
+Sequel.migration do
+  up do
+    rotations.each do |tile, rotation|
+      actions = Action.where(json.get_text("type") => "lay_tile",
+                             json.get_text("tile") => /^#{tile}-/).all
+
+      actions.each do |action|
+        parsed = JSON.parse(action.action)
+        current = parsed["rotation"] || 0
```
tobymao, line 30: why would there be no rotation

kelsin: I don't know, got this when downgrading the migration locally, don't even know if you want a migration?

lib/engine/game/g_1889.rb
kelsin: Sure, I'll fix before anything for real. Since we were already in `Game` I thought you need the full thing to go up stack? I'll test and make sure any reference is shortened in anything final

tobymao:
don't need an actual migration, but i will need to run code

kelsin:
Do you want the script in any particular location in this PR or to just give it to you on slack?

tobymao:
slack

lib/engine/game/g_1889.rb
```diff
@@ -1,166 +1,35 @@
...
-      STARTING_CASH = {
-        2 => 420,
-        3 => 420,
-        4 => 420,
-        5 => 390,
-        6 => 390,
-      }.freeze
+      PHASES = PHASES.zip(PHASES_BEHAVIOR).map do |config, behavior|
+        config.merge(behavior)
+      end
```
kelsin, line 22: @tobymao want opinions on this. Merging in behavior with what 18xx-maker can export. Should get a good pattern going and don't know what you want it to look like.

lib/engine/game/g_1889.rb
```diff
@@ -1,166 +1,35 @@
...
-        %w[65 70 75 80p 90 100 110 125 140 155 175 200],
-        %w[60 65 70 75p 80 90 100 110 125 140],
-        %w[55 60 65 70p 75 80 90 100],
-        %w[50y 55 60 65p 70 75 80],
-        %w[45y 50y 55 60 65 70],
-        %w[40y 45y 50y 55 60],
-        %w[30o 40y 45y 50y],
-        %w[20o 30o 40y 45y],
-        %w[10o 20o 30o 40y],
+      PHASES_BEHAVIOR = [
```
tobymao, line 11: phase_behaviors

why is this split out? this is config

lib/engine/game/g_1889.rb
tobymao: i'm not sure i like this

lib/engine/game/g_1889.rb
kelsin: Yeah, for others benefit we talked online that we want all of this data to be part of the export and not have to merge some data from the game file and some data from the export. If anyone has good ideas let us know. Thinking about it myself while I clean up other parts of rotations :)

kelsin:
Ok this version has a load module that has two methods to properly convert the json into symbols at the right places.

kelsin:
The current migration script:
```Dir['./models/**/*.rb'].sort.each { |file| require file }
require './lib/engine'

def fix_rotations
  rotations = {
    '7' => 5,
    '8' => 4,
    '12' => 5,
    '16' => 4,
    '19' => 3,
    '41' => 3,
    '42' => 3,
    '45' => 3,
    '46' => 3,
    '47' => 2,
    '465' => 2,
  }.freeze

  Sequel.extension(:pg_json_ops)
  json = Sequel.pg_jsonb_op(:action)

  rotations.each do |tile, rotation|
    actions = Action.where(json.get_text("type") => "lay_tile",
                           json.get_text("tile") => /^#{tile}-/).all

    actions.each do |action|
      # puts "#{action.action["id"]}: #{action.action["tile"]} #{action.action["rotation"]} -> #{(action.action["rotation"] + rotation) % 6}"
      action.action["rotation"] = (action.action["rotation"] + rotation) % 6
      action.save
    end
  end
end
```

kelsin:
TODO: Move to reading 18xx-maker/game files raw and convert in ruby land into needed formats

PR #336, opened 2020-05-08
WIP Change payout messages to look more like arithmetic

### LINKED ISSUE #330: History messages about financial events should use = or — instead of -
"Player 1 (Teacher) receives ¥30 - ¥10 x 3"

I think this would make more sense with "=", possibly also rearranging the order of the numbers.

"TR pays out ¥110 - ¥11 per share"

This should be — or comma, not hyphen/dash.

sparr:
WIP because I need to test further, unless someone else does before I get to it.

resolves #330

### DIFF
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -351,9 +351,11 @@ module Engine
 
       def payout(revenue)
         @current_entity.add_revenue!(@game.turn, @round_num, revenue)
-        per_share = revenue / 10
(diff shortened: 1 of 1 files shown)


dfannius:
I know that some titles have shares of different sizes for a single corporation (e.g., both 10% and 5%), but that will screw up other parts of this code too if we ever get to them, so maybe it's unfair to pick on this function for not working well with that case.

PR #338, opened 2020-05-08
Change "passes" to "finishes"/"skips" for action logging

### LINKED ISSUE #329: Change messages for "passes" to match new button text
We previously decided to change a bunch of the "Pass" action buttons to "Done (STAGENAME)", but the history message for at least some of these actions still say "passes", such as "UR passes buying trains" after it already bought some trains and then clicked "Done (Trains)".

sparr:
Resolves #329

### DIFF
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -470,7 +470,7 @@ module Engine
       end
 
       def log_pass(entity)
-        verb = (@step == @last_action_step ? 'finishes' : 'skips')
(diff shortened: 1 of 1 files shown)


lib/engine/round/operating.rb
```diff
@@ -470,15 +470,16 @@ def remove_just_sold_company_abilities
       end
 
       def log_pass(entity)
+        verb = (@step == @last_action_step ? 'finishes' : 'skips')
```
tobymao, line 473: remove the parens

PR #341, opened 2020-05-08
WIP Attempt to move phase change after action

### LINKED ISSUE #332: Change order of history messages for train purchase and phase change
"-- Phase 3 (Operating Rounds: 2, Train Limit: 4, Available Tiles: Yellow, Green ) --"
"KO buys a 3 train for ¥180 from The Depot"

I think these messages should be in the opposite order. The phase changes immediately *after* the train is bought, not before.

sparr:
This is a partial implementation of moving the phase change after the processing of a round action. The new conditional method calls are needed because the current logic expects new phase rules to be in place during the processing of an action. Unfortunately there are six test failures where existing games run into a player acting out of turn, due to turn order not being appropriately advanced. Further investigation is required.

resolves #332

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -197,13 +197,15 @@ module Engine
 
         action = action_from_h(action) if action.is_a?(Hash)
         action.id = current_action_id
-        @phase.process_action(action)
         # company special power actions are processed by a different round handler
         if action.entity.is_a?(Company)
           @special.process_action(action)
         else
           @round.process_action(action)
         end
+        @phase.process_action(action)
+        @round.conditional_next_step! if round.respond_to? :conditional_next_step!
+        @round.conditional_change_entity! if round.respond_to? :conditional_change_entity!
         @actions << action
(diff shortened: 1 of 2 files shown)


tobymao:
to change the logging, this big of a change is not worth it, i'm going to close this pr

sparr:
@tobymao This isn't really about the logging. That's just a symptom. Changing the phase before applying the action is not how the game rules actually work. I expect to see some nasty hoop-jumping and side effects of this decision as more complex titles get implemented. I guess we can revisit this when we get to one of those titles, although I'm worried this will be a "frog in a pot of boiling water" sort of situation.

tobymao:
i think i agree with you. but to fix it i'd probably move the phase calculation into the operating round

daniel-sousa-me:
When a phase changes there are things that are effected "immediately", like train limit and ability to buy corps and others that only have an effect at the next OR. But in any of these cases, I think that the trigger should only run after the current action is completed and store somewhere the things that only take effect in the next OR and the things that take effect only on the next set of ORs

PR #342, opened 2020-05-08
Make undo an action and add redo capability

### LINKED ISSUE #242: Undo >>> Redo
A way to back your latest move... but some players move back too much....

### LINKED ISSUE #263: Undo shouldn't affect chat text
It's weird to me that undo gets rid of the chat text... It feels like communication should be independent of game state, especially in cases where a rules error means the game has to be rewound a bit and that deletes all of the relevant discussion!

### LINKED ISSUE #284: Undo doesn't update the active player on the main page

jenf:
Redone as per suggestion.

Note, I've had to re-add the actions to the list to keep the action.id's consistent and avoid a race
I've also added two temporary variables to see if undo and redo are possible as calculating it on the fly would be problematic.

Due to the size of the PR, i'd rather fix the advisory lock as a separate PR.

fixes #284, fixes #263, fixes #242


tobymao:
this seems overly complex, and i don't know why you need to do the change to actions, can't you just check the type of the dict and be done?

tobymao:
can the whole game utils be replaced by something like this

stack = []
undos = []

actions.each do |action|
  case action['type']
  when 'undo'
    undos << stack.pop
  when 'redo'
    stack << undos.pop
  else
    undos.clear
    stack << action
  end
end

jenf:
@tobymao hopefully that's a bit closer, could you see if it's okay in principle, i'll work on merging the 'helper' into the original class next. Then hooking up the UI

lib/engine/gameutils/undo_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+        process_actions(actions)
+      end
+
+      def process_actions(actions)
+        @undo_list = []
+        @last_known_undo = 0
+        return if actions.nil?
+
+        # List of actions that undo does not apply to
+        kept_actions = Action.constants.map do |c|
```
tobymao, line 18: hm maybe too much magic, maybe we can just stick to hardcoding

lib/engine/gameutils/undo_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+      attr_reader :undo_list
+
+      def initialize(actions = nil)
+        process_actions(actions)
+      end
+
+      def process_actions(actions)
+        @undo_list = []
+        @last_known_undo = 0
+        return if actions.nil?
```
tobymao, line 15: in generally i like do to return unless actions instead 

but i don't think we need this pattern anayws if actions is just an empty list

lib/engine/action/message.rb
```diff
@@ -19,6 +19,10 @@ def self.h_to_args(h, _)
       def args_to_h
         { 'message' => @message }
       end
+
+      def self.keep_on_undo?
```
tobymao, line 23: yea doesn't seem like you need this

lib/engine/action/base.rb
```diff
@@ -27,11 +27,15 @@ def initialize(entity)
...
       end
 
+      def [](field)
+        to_h[field]
+      end
+
       def to_h
         {
-          'type' => type_s(self),
+          'type' => self.class.type_s(self.class),
```
tobymao, line 36: doesn't seem like you needed to make these changes

lib/engine/gameutils/undo_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+        kept_actions = Action.constants.map do |c|
+          klass = Action.const_get(c)
+          klass.type_s(klass) if klass.keep_on_undo?
+        end.compact
+
+        stack = []
+        actions.each.with_index do |action, index|
+          if action['type'] == 'undo'
+            @last_known_undo = index + 1
+            @undo_list.append(*stack.pop(action['steps']))
```
tobymao, line 27: why do you need steps?

lib/engine/gameutils/undo_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+            @last_known_undo = index
+            @undo_list.pop(action['steps'])
+          elsif !kept_actions.include?(action['type'])
+            # action_id's start at 1
+            stack.append(index + 1)
+          end
+        end
+      end
+
+      def last_undoable_action(actions)
```
tobymao, line 38: i assume code under here is vestigial

tobymao:
> @tobymao hopefully that's a bit closer, could you see if it's okay in principle, i'll work on merging the 'helper' into the original class next. Then hooking up the UI

it's closer but i think it can be further simplified. happy to pair if you'd like

lib/engine/gameutils/undo_helper.rb
jenf: keep_actions=['message'] okay?

lib/engine/game/base.rb
```diff
@@ -192,11 +193,44 @@ def active_players
...
+            @undo_list.pop
+          elsif action['type'] != 'message'
+            # Adding more types of action here will break existing games
+
+            stack.append(index + 1)
+          end
+        end
+      end
+
       def process_action(action)
```
tobymao, line 217: sorry to be picky, but i feel like process_action could just handle all of this without keep track of indicies

so first pass, you remove all the undos and redos that way process_action knows nothing about undos and redos except when they get them, they must do a clone

lib/engine/game/base.rb
```diff
@@ -192,11 +193,44 @@ def active_players
         @round.active_entities.map(&:owner)
       end
 
+      def process_undos(actions)
```
tobymao, line 196: i haven't tested this orr thought it through, but this was my idea


`arr = Array.new(actions.size)
undos = []

actions.each_with_index do |action, index|
  case action['type']
  when 'undo'
    i = arr.rindex { |a| a['type'] != 'message'}
    undos << [stack.delete(i), i]
  when 'redo'
    a, i = undos.pop
    arr[i] = a
  else
    undos.clear
    arr[index] = action
  end
end

arr.compact
`

lib/engine/game/base.rb
jenf: I think the problem with this is when you come to the clone you'll need to restore the undos before being reprocessed.

tobymao: wouldn't you only pass the canonical data into the clone?

jenf: yes, there's a secondary issue i need to find a solution for, since everything is passing action ids, there's a chance that if someone undos and does an action quickly enough two browsers can diverge in what the action id is.

tobymao: would the advisory locks i added help with this?

lib/engine/game/base.rb
```diff
@@ -192,18 +188,77 @@ def active_players
...
+            filtered_actions[index] = action
+          end
+        end
+
+        @undo_possible = false
+        # replay all actions with a copy
+        filtered_actions.each.with_index do |action, index|
+          if action.nil?
+            # Restore the original action to the list to ensure action ids remain consistent but don't apply them
+            @actions << actions[index]
```
jenf, line 220: This readds it to the list

lib/engine/game/base.rb
```diff
@@ -192,18 +188,77 @@ def active_players
...
 
+      # Initialize actions respecting the undo state
+      def initialize_actions(actions)
+        active_undos = []
+        filtered_actions = Array.new(actions.size)
+
+        actions.each.with_index do |action, index|
+          case action['type']
+          when 'undo'
+            i = filtered_actions.rindex { |a| !a.nil? && a['type'] != 'message' }
```
tobymao, line 199: this can just be rindex { |a| a && a['type'] != 'message' }

lib/engine/game/base.rb
```diff
@@ -192,18 +188,77 @@ def active_players
...
+          else
+            active_undos = []
+            filtered_actions[index] = action
+          end
+        end
+
+        @undo_possible = false
+        # replay all actions with a copy
+        filtered_actions.each.with_index do |action, index|
+          if action.nil?
```
tobymao, line 218: reverse this, if action...

lib/engine/game/base.rb
```diff
@@ -192,18 +188,77 @@ def active_players
...
+            # Restore the original action to the list to ensure action ids remain consistent but don't apply them
+            @actions << actions[index]
+          else
+            action = action.copy(self) if action.is_a?(Action::Base)
+            process_action(action)
+          end
+        end
+        @redo_possible = !active_undos.empty?
```
tobymao, line 226: @redo_possible = active_undos.any?

lib/engine/game/base.rb
```diff
@@ -192,18 +188,77 @@ def active_players
...
+      end
+
+      def can_undo?
```
tobymao, line 229: i'd remove these two helpers and just make redo_possible and can_undo as readers

Issue #337, opened 2020-05-08
Smarter priority deal tag


dfannius:
Currently, during an SR the player who went first in this SR gets the Priority Deal tag on their card, which is useless information since it has no further effect on the game.

Instead, put the Priority Deal tag on the player who will go first in the following SR if everyone starts passing from here on out.

This would generalize nicely, and be very useful in, games like 1828 where player order for the following SR depends on the order in which they dropped out of this one.

dfannius:
Implemented with https://github.com/tobymao/18xx/commit/d3dbe070a841846da6f7703a5915bcb7800b4046.

PR #348, opened 2020-05-09
Separate @auction_triggerer and @last_to_act

dfannius:
The code for #337 (in an upcoming PR) wants `@last_to_act` on a round with players (auction, stock) to reliably refer to the player who most recently effectively passed the priority deal card to their left by taking a non-pass action. The auction code overloaded this instance variable in order to handle turn order both during top-level bidding and when coming out of set of limited auctions.

The state necessary to handle auction turn order has now been broken out into `@last_to_act` (which now has the desired meaning defined above) and `@auction_triggerer` (the player who triggered a set of limited auctions by buying the lowest company outright).

In passing I also fixed a bug where the "Company goes up for auction" message could get logged multiple times if there were more than two players bidding for it.

I added a bunch of comments to make it clearer to myself what was going on; if you don't like them cluttering up the code I can remove them, but I tried to be pretty terse.

I've verified that all games from the DB Toby gave me a couple of days ago are unaffected by this change.

### DIFF
--- a/lib/engine/round/auction.rb
+++ b/lib/engine/round/auction.rb
@@ -17,8 +17,16 @@ module Engine
 
         @bids = Hash.new { |h, k| h[k] = [] }
         @bidders = Hash.new { |h, k| h[k] = [] }
-        @auctioning_company = nil
+
+        # The player after this one will have priority deal next round
+        # if everyone passes.
         @last_to_act = nil
+
+        # The company currently up for limited auction, or nil.
+        @auctioning_company = nil
+
+        # The player who kicked off the limited auction, or nil.
+        @auction_triggerer = nil
       end
 
       def name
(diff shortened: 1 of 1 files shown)


lib/engine/round/auction.rb
```diff
@@ -63,24 +71,35 @@ def all_passed?
         @entities.all?(&:passed?)
       end
 
+      # Process a non-pass action.
       def _process_action(bid)
+        @last_to_act = @current_entity unless @auctioning_company
```
tobymao, line 76: can this go on line 81? so you don't need the double auction company check

lib/engine/round/auction.rb
```diff
@@ -63,24 +71,35 @@ def all_passed?
...
         if @auctioning_company
+          # We're in the middle of a limited auction.
```
tobymao, line 78: i don't think these 2 comments are necessary, the code reads pretty well here

lib/engine/round/auction.rb
```diff
@@ -121,10 +148,12 @@ def pass_processed(_action)
 
       def placement_bid(bid)
         if @companies.first == bid.company
-          @last_to_act = bid.entity
+          # We're buying the first company.
```
tobymao, line 151: these two comments are unnecssary

tobymao:
overall looks fine to me

PR #349, opened 2020-05-09
More dynamic priority deal updating (implements #337)

dfannius:


lib/engine/game/base.rb
```diff
@@ -397,7 +397,24 @@ def end_game
...
 
-      def calculate_priority_deal
+      def priority_deal_player
+        if @round.current_entity.player?
+          # We're in a round that iterates over players, so the
+          # priority deal card goes to the player who will go first if
+          # everyone passes starting now.  last_to_act is nil before
+          # anyone has gone, in which case the first player has PD.
+          last_to_act = @round.last_to_act
+          priority_idx = last_to_act ? (@players.find_index(last_to_act) + 1) % @players.size : 0
```
tobymao, line 407: could this be

@players[(@players.find_index(last_to_act) + 1) % @players.size || 0]

dfannius: No, when `last_to_act` is `nil` then `@players.find_index(last_to_act)` is `nil` and we can't add 1 to it.

tobymao: ok so it could be 

(find_index(last_to_act) || -1) + 1 % ...

dfannius: That's too golfy for my tastes (and if we ever had a bug where `last_to_act` is non-nil but not in `@players`, this would silently hide it) but it's your codebase! Change if you like.

tobymao:
i agree which is why i merged it as is

Issue #227, opened 2020-04-30
"Your games" game sorting should prioritize active games


jcoveney:
IMO, the order should be: active games, then started games, then games that have not started yet.

jcoveney:
Is this uninteresting to you? It's still an "issue" as far as I can tell
(quotes bc it's not some major bug). Not huge, but annoying. If you have a
lot of games, always having to click "next" to check for new ones seems
like there's no reason for it to be necessary. If one ever had 3 pages of
games it's be quite annoying, as each page swap can take a couple of
seconds on my connection.

tobymao:
fix is merged

dfannius:
This was implemented in PR #351, which was merged 4 hours ago as I write this. It may not have gone live yet.

jcoveney:
Ah ok I was on my phone. Nice work as always :)

tobymao:
ger contributed this one

dragon12:
@jcoveney ah I should have posted here :) I note your follow-up https://github.com/tobymao/18xx/issues/356 , and I've also just opened https://github.com/tobymao/18xx/issues/375 because the new Finished Games addition aren't sorted now.

PR #350, opened 2020-05-09
issue_227: sort user's games by ready, active, id

dragon12:
https://github.com/tobymao/18xx/issues/227

### DIFF
--- a/assets/app/view/home.rb
+++ b/assets/app/view/home.rb
@@ -21,6 +21,13 @@ module View
       your_games, other_games = @games.partition { |game| user_in_game?(@user, game) }
       grouped = other_games.group_by { |game| game['status'] }
 
+      # Ready, then active, then unstarted.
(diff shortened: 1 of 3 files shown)


assets/app/view/home.rb
```diff
@@ -21,6 +21,13 @@ def render
       your_games, other_games = @games.partition { |game| user_in_game?(@user, game) }
       grouped = other_games.group_by { |game| game['status'] }
 
+      # Ready, then active, then unstarted.
+      your_games.sort_by! do |game|
+        [user_is_acting?(@user, game) ? -2 : 2,
```
tobymao, line 26: can you reformat this to be 
[
  x,
  y,
  z,
]

also is it intentional to sort older games first as opposed to new games?

dragon12: No, that was an oversight, thanks. Fixed now.

PR #352, opened 2020-05-09
Adding decode of additional 'g' path notation for tile gauge

thebamaman:
Tile rendering with different gauges will be added in a future commit/pr


lib/engine/tile.rb
```diff
@@ -201,21 +201,34 @@ def self.from_code(name, color, code, **opts)
       Tile.new(name, color: color, parts: decode(code), **opts)
     end
 
+    def self.connection_from_code(code, cache)
```
tobymao, line 204: path_from_code

tobymao:
this merge looks bad

PR #353, opened 2020-05-09
Location names on separate lines, more developer-friendly tile routes, and more tile behind-the-scenes changes

michaeljb:
* fix an 1846 location name ("ERIE" => "Erie")
* move some of 18Chesapeake's hex definitions into the main Tile class
* add ability for game classes to define their own tiles in `TILES`; move `X_` tiles for 18Chesapeake and 1817 out of main tile class into respective game classes
* `/all_tiles` is now `/tiles/all`
* added `/tiles/<tile_name>` (large view)
* added `/tiles/<game_title>` (and moved individual game tile pages out of the "all" page)
* added `/tiles/<game_title>/<hex_coord>` (large view)
* added `/tiles/<game_title>/<tile_name>` (large view)

hex coord screenshot

18Chesapeake map before this branch:

Chessie before

18Chesapeake map after this branch (with long names rendered across multiple lines):

Chessie after


assets/app/view/part/location_name.rb
```diff
@@ -69,7 +71,48 @@ def render_part
...
+
+        case segments.size
+        when 3
+          # join the middle word with the shorter of the first and last words;
+          # prefer joining with first if the first and last words are the same
+          # length
+          if segments[0].size > segments[2].size
+            [segments[0], segments.slice(1..2).join(' ')]
+          else
+            [segments.slice(0..1).join(' '), segments[2]]
```
tobymao, line 108: rather than slice maybe just segments[0..1].join(' ')

assets/app/view/part/location_name.rb
```diff
@@ -69,7 +71,48 @@ def render_part
...
+          end
+        when 4
+          # join first two words together, and join last two words together
+          [segments.slice(0..1).join(' '), segments.slice(2..3).join(' ')]
```
tobymao, line 112: segments.each_slice(2).map { |pair| pair.join(' ') }

lib/engine/game/base.rb
```diff
@@ -42,6 +42,13 @@ class Base
 
       HEXES = {}.freeze
 
+      CUSTOM_TILES = {
```
tobymao, line 45: i feel like we could treat this like we do hexes inside tiles and have inline definition for custom tiles, i think that'd be closer to @kelsin 18xxmaker

lib/engine/game/base.rb
```diff
@@ -333,8 +340,23 @@ def init_hexes(companies, corporations)
...
+          tiles.map do |name, code|
+            tile = Tile.from_code(name, color, code)
+            [name, tile]
+          end
+        end.to_h
+
         self.class::TILES.flat_map do |name, num|
-          num.times.map { |index| Tile.for(name, index: index) }
+          num.times.map do |index|
+            tile = custom_tiles[name]
```
tobymao, line 352: if (tile = custom_tiles[name])

lib/engine/game/g_18_chesapeake.rb
```diff
@@ -22,13 +22,13 @@ class G18Chesapeake < Base
           %w[B6 B8 B10 C3 C7 C9 C11 E7 E9 E13 F6 F12 G7 I7 J8 J10 L4 F4 G5 H2 I3] => 'blank',
           %w[B12 D4 D6 D10 E5] => 'mtn80',
           %w[F10 G9 G11 H12] => 'wtr40',
-          %w[B4 K3 K5] => 't=r:0;t=r:0',
+          %w[B4 K3 K5] => 'double_town',
```
tobymao, line 25: i think we're trying to move away from these aliases, thoughts @kelsin

assets/app/view/part/location_name.rb
michaeljb: 3f2b6ff

assets/app/view/part/location_name.rb
michaeljb: 3f2b6ff

assets/app/view/part/location_name.rb
michaeljb: 3f2b6ff

assets/app/view/part/location_name.rb
michaeljb: 3f2b6ff

lib/engine/game/g_18_chesapeake.rb
michaeljb: If that's the way we want to go, I'll update the PR to ditch all of these aliases.

lib/engine/game/base.rb
michaeljb: Updated in d8f4da2 and later commits

michaeljb:
I've just realized that in game on the tiles tab, tiles now render with a rotation other than 0--I'm guessing whichever rotation the first tile of that name has, since I'm using `render_tile_block(name, tile: tiles.first, num: tiles.size)` in `tile_manifest.rb` now instead of passing in `nil` for the tile and creating it from scratch. (before I added `tile: tiles.first` there, the in-game tile tab broke if any custom tiles are defined in the game class). Probably need to make a clone of the tile and reset its rotation to 0 rather than using the game's first tile of that name.

This reminds me that on the tiles_page, I would also like to be able to click a tile to rotate it. I looked into something with that a while back but the rotation stuff I found was very integrated in the tile selector, I wasn't very persistent before giving up on that then.

lib/engine/game/g_18_chesapeake.rb
tobymao: @kelsin thoughts?

PR #367, opened 2020-05-11
Enforce blocking on hexes (which breaks fixture #164) and check games script in

### LINKED ISSUE #188: Ohzu upgradeable before Ehime bought in (game id 300)
UR was able to place a green upgrade in Ohzu while the Ehime private was still held by a player. According to 15.1 in the DTG rulebook, the hex should be blocked.

### LINKED ISSUE #243: Enforcing Private's A power
A way to block K4, until it's sold....

jenf:
fixes #243
fixes #188
there are three active games which this breaks from the db drop. None of which look easy to fix.


tobymao:
i can't reopen this, can you make a new pr?

Issue #149, opened 2020-04-25
Password reset


michaeljb:
Nearly had an issue today where I thought I had lost my password, but luckily it actually was in my password manager, just saved under the IP for the old server rather than "18xx.games".

But, sooner or later this will happen to a user who won't be able to recover their password, so we'll want to be able to automatically email them with a short-lived, one-time link to set a new password.

tobymao:
i have code to do this in rollingstock, haven't implemented it yet

cfreeman1305:
Implementing it would be great, because I am one of those schmucks who mistyped their password the first time and I need to reset it.

mkarbowiak:
> Implementing it would be great, because I am one of those schmucks who mistyped their password the first time and I need to reset it.

I second that

PR #385, opened 2020-05-14
Add support for pinned code to allow disruptive game changes without impacting existing users

### LINKED ISSUE #373: Add version system for rule implementation
Migrating older games when rules are broken is turning out to be problematic

Add a version field into the database
(Maybe add a git commit hash as well)
Add the version field into the game constructor

When we strengthen a rule that breaks existing games we put it behind a if version>x

Depending on the version (alpha, beta, production) of the rules we should display a different message to the users when creating games

Alpha:
"18Cheasepeake is currently considered 'alpha', the rules implementation is likely to not be complete and as we improve our implementation of the rules, games that are breaking the rules will be deleted. We suggest that any alpha quality game is concluded within 7 days"

Beta:
"1889 is currently considered 'beta', the rules implementation is not robust yet and can allow illegal moves, we are working to improve the quality of the rules implementation, games that are breaking the rules will be given 7 days to complete before being deleted. Because of this it may not be suitable playing games that may take months to complete"

jenf:
Add pinning support
Modify the rakefile to generate pinned versions automatically

(I'll add a migration script as part of another PR restoring PR #367)

fixes #373


api.rb
```diff
@@ -76,6 +76,8 @@ class Api < Roda
   plugin :halt
 
   ASSETS = Assets.new(precompiled: PRODUCTION)
+  # TODO: This should scan the directory
+  PIN_ASSETS = { 'deadbeef' => Assets.new(precompiled: true, pin: 'deadbeef') }.freeze
```
tobymao, line 80: we shouldn't serve these assets or render them in opal on the server, let's only support static gzip file

api.rb
```diff
@@ -123,27 +125,29 @@ class Api < Roda
...
 
   def render_with_games
     render(games: Game.home_games(user, **request.params).map(&:to_h))
   end
 
-  def render(**needs)
+  def render(pin: nil, **needs)
     return debug(**needs) if request.params['debug'] && !PRODUCTION
 
+    asset = PIN_ASSETS[pin] || ASSETS
```
tobymao, line 139: as i said above, let's have the app fetch only the gzip version of the js

migrate/0004_add_game_pinning.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    add_column :games, :pin_version, String, null: true, default: nil
```
tobymao, line 5: i don't think we need this, we can add it into settings

models/game.rb
```diff
@@ -113,6 +113,7 @@ def to_h(include_actions: false)
       actions: actions_h,
       created_at: pp_created_at,
       updated_at: pp_updated_at,
+      pin_version: pin_version
```
tobymao, line 116: let's not check in any of the assets

jenf: that's fine :)

api.rb
jenf: I'll get everything else working then try and fix this, there's a question of how to disable the prerender.

tobymao: i think in addition the static gzip, we'll need to generate the game page html

migrate/0004_add_game_pinning.rb
jenf: fixed :)

api.rb
jenf: yeah, that's why i went down the asset route since it already knew how to generate it.

tobymao: take a look at api.rb#134 and #149 
if you want it to look nice you'll need to do something like 137, else you can try 153

jenf:
Apart from the rakefile and lib/assets.rb I think this is pretty close (lib/assets.rb will be reverted), @tobymao could you have a look at the other files and see if there's any comments?

I'll add a migration script as well as a mechanism for storing pinned versions

api.rb
```diff
@@ -146,6 +147,24 @@ def render(**needs)
...
   end
 
+  def render_pin(**needs)
+    needsjs = Snabberb.wrap(app_route: request.path, **needs)
+    attach_func = "Opal.$$.App.$attach('app', #{needsjs})"
+    <<~HTML
+      <html>
+        <head>
+          <meta charset="utf-8">
+          <title>18xx.games</title>
```
tobymao, line 157: should we change the title here

tobymao: some indication of the pin? i dunno just a thought

assets/app/game_manager.rb
```diff
@@ -125,8 +131,14 @@ def url(game, path = '')
     "/game/#{game['id']}#{path}"
   end
 
-  def hs_url(game)
-    "/hotseat/#{game['id']}"
+  def hs_url(game, game_data)
```
tobymao, line 134: how can hot seats ever get pinned?

lib/engine/game/base.rb
```diff
@@ -184,7 +184,7 @@ def self.load_from_json(json)
         const_set(:HEXES, data['hexes'])
       end
 
-      def initialize(names, id: 0, actions: [])
+      def initialize(names, id: 0, actions: [], pin: nil)
```
tobymao, line 187: should we call all the other references pin as well? as opposed to pin_version

assets/app/view/actionable.rb
```diff
@@ -41,9 +41,19 @@ def process_action(action)
           @game_data[:result] = {}
           @game_data[:status] = 'active'
         end
+
         Lib::Storage[@game_data[:id]] = @game_data
       elsif participant
-        @connection.safe_post("/game/#{@game_data['id']}/action", action.to_h)
+        json = action.to_h
+        meta = {
```
tobymao, line 48: should this only be sent for pinned games?

lib/engine/game/base.rb
jenf: Can do, will change them all to pin

assets/app/game_manager.rb
jenf: Pass ?pin_version=githash

This would be used if someone reported a broken hotseat to us, we can get them to set the relevant hash.

tobymao:
gotcha

Rakefile
```diff
@@ -106,5 +106,12 @@ end
...
 task :precompile do
   require_relative 'lib/assets'
-  Assets.new(cache: false, make_map: false, compress: true, gzip: true).combine
+  bundle = Assets.new(cache: false, make_map: false, compress: true, gzip: true).combine
+
+  # Copy to the pin directory
+  git_rev = `git rev-parse --short HEAD`.strip
+  pin_dir = Assets::OUTPUT_BASE + Assets::PIN_DIR
+  FileUtils.mkdir_p(pin_dir)
+  FileUtils.cp(bundle, "#{pin_dir}/#{git_rev}.js")
```
tobymao, line 115: we only need the gz right?

tobymao:
lemme know when this is good to go, maybe we can merge tomorrow

tobymao:
the only comment i have is that it saves every deploy but there's no way to pin games in the rake.
can we make pinning and generating and shipping the asset a command as opposed to just by default? let's plan to ship this tomorrow if you have time

tobymao:
so to clarify, there should be a command that will pin select games, generate the assets and ship the file

jenf:
Yeah, I'd hope to incorporate it as a separate pr. I'm unlikely to get to that today since I'm a bit busy today

PR #386, opened 2020-05-14
Add desktop notifications to game and home pages

### LINKED ISSUE #176: Consider desktop notifications for the webapp
Such a slick notification! While I normally dislike it when sites ask to notify me, this is actually an example where it would be perfect! Could have a notification for chat updates (with the chat), a notification for moves (ideally with the move just taken), and if you click the notification it opens up a tab with the game.

sparr:
This PR adds a desktop notification (through the Notification javascript API) when it becomes the current user's turn while viewing the game or home pages. These notifications are, as with all web site desktop notifications, optional and opt-in, and can be controlled by the the user's browser and/or OS.

This may leave a notification gap. I am unclear on whether or not the player will get email notifications while viewing the About or Profile (or hidden top level views like Tiles) pages, and since those pages have no event subscriptions they will also not get these desktop notifications.

This implementation also fails to deduplicate notifications. If the player has the same game open in two tabs/windows, or the game and the home page both open, they will get two notifications. It may be possible to resolve this with local storage, such as with a simple list of the most recent action number for which each game has issued a notification, although that would fail to account for the desired notification when a game encounters an undo and redo.

Resolves #176

### DIFF
--- a/assets/app/lib/notification.rb
+++ b/assets/app/lib/notification.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Lib
+  module Notification
+    def self.notify(message)
+      %x{
+        if ("Notification" in window) {
+          if (Notification.permission === "granted") {
+            new Notification(#{message});
+          }
+          else if (Notification.permission !== "denied") {
+            Notification.requestPermission().then(function (permission) {
+              if (permission === "granted") {
+                new Notification(#{message});
+              }
+            });
+          }
+        }
+      }
+    end
+  end
+end
--- a/assets/app/view/game.rb
+++ b/assets/app/view/game.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'lib/connection'
+require 'lib/notification'
 require 'lib/params'
 require 'view/auction_round'
 require 'view/companies'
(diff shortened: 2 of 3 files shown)


tobymao:
Can this just be done with game_data['acting']? then we can have the logic in the notify. Also, should we only notify users if the tab is not in view?

you can check that with 

something like this

`  var isHidden = function() {
    if (hiddenProperty !== undefined) {
      return document[hiddenProperty];
    } else {
      return !document.hasFocus;
    }
  };
`

sparr:
By "done with game_data['acting']" you mean changing https://github.com/tobymao/18xx/pull/386/files#diff-e53ed7abda333dfc52e9c9f79db95f79R77 to work more like https://github.com/tobymao/18xx/pull/386/files#diff-f7531823748653b50a63a42bde5ab5cbR50 ?

As to active tabs... I want the notification regardless. The tab might be front and focused, but I'm not looking at the computer. Or it might be focused but another window is on top of it. I guess we could add profile options for how to filter notifications? I think it should be up to the browser + computer + user to decide if and when to surface notifications.

sparr:
Putting all of the acting logic in .notify would not be great. I expect that we will eventually want to [optionally] send more types of notifications, like a game starting, or even a chat message arriving.

tobymao:
yes, regarding the data acting. maybe we can add another method that's specific to notifying game turns in notify. 

let's add the user specific notifications settings to the profile page so that the user can choose to only show them when the page is active or not. i think the default should be only notify if the browser is not in view

PR #393, opened 2020-05-15
Json loading

kelsin:
Don't like how I make a module with ONLY a json string in it... but file reading doesn't work otherwise right?


lib/engine/game/base.rb
```diff
@@ -131,6 +131,58 @@ def self.title
         name.split('::').last.slice(1..-1)
       end
 
+      def self.load_from_json(json)
```
tobymao, line 134: i'm not sure if it's worth the complexity, but it would be theoretically possible to lead the json directly, if you want to do this instead of putting the json in a ruby file let me know

kelsin: Yep, later on let's do that whenever ready :)

PR #394, opened 2020-05-16
Add warnings about alpha and beta quality games on initialization

jenf:
I thought separating this out from the rest of the opal pinning would be sensible.

It adds the following messages to the top of the log

Screenshot from 2020-05-16 11-31-48

For dev purposes if you do http://localhost:9292/new_game?all you can see all the games not just those marked alpha and above, 1889 is the only game visible at the moment.

### DIFF
--- a/lib/engine.rb
+++ b/lib/engine.rb
@@ -15,5 +15,8 @@ module Engine
     klass
   end.compact
 
+  # Games that are alpha or above
+  VISIBLE_GAMES = GAMES.select { |game| %i[alpha beta production].include? game::DEV_STAGE }
+
   GAMES_BY_TITLE = GAMES.map { |game| [game.title, game] }.to_h
 end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -27,7 +27,7 @@ module Engine
       attr_reader :actions, :bank, :cert_limit, :cities, :companies, :corporations,
                   :depot, :finished, :hexes, :id, :log, :phase, :players, :round,
                   :share_pool, :special, :stock_market, :tiles, :turn, :undo_possible, :redo_possible
-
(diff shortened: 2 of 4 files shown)


lib/engine.rb
```diff
@@ -15,5 +15,8 @@ module Engine
     klass
   end.compact
 
+  # Games that are alpha or above
+  VISIBLE_GAMES = GAMES.select { |game| %i[alpha beta production].include? game::DEV_STAGE }
```
tobymao, line 19: maybe it's simpler to reject prealpha?

lib/engine/game/base.rb
```diff
@@ -192,6 +192,23 @@ def initialize(names, id: 0, actions: [])
...
+          ' the rules implementation is likely to not be complete.'
+          @log << 'As the implementation improves, games that are not compatible'\
+          ' with the latest version will be deleted.'
+          @log << 'We suggest that any alpha quality game is concluded within 7 days.'
+        when :beta
+          @log << "#{self.class.title} is currently considered 'beta',"\
+          ' the rules implementation may allow illegal moves.'
+          @log << 'As the implementation improves, games that are not compatible'\
+          ' with the latest version will be given 7 days to be completed before being deleted.'
+          @log << 'Because of this we suggest not playing games that may take months to complete.'
```
tobymao, line 209: but with your opal changes will this be true anymore?

jenf: Depends on how often we want to pin, are we gonna keep them all?

tobymao: yea maybe not

jenf: it's probably easier to give the users a "you have 7 days" rather than leaving them with the impression that they might have infinity time to complete.
We may of course be more generous than that :)

PR #395, opened 2020-05-16
Show info when game cannot be loaded

### LINKED ISSUE #339: Attempting to load game data that leads to an invalid state should jump to the last valid move in the history
Currently if you try to load invalid game data, the hotseat game will refuse to load/render, with a console error that the game constructor has failed. It would be nice if it would load at the correct action index to render the last valid game state, so that the error can be produced by clicking forward one step in the history, and the game state can be inspected.

jenf:
Fixes #339

(This'll be needed in the opal pinning for hotseat games)

The only thing I couldn't quite work out how to do is to be able to disable it when running in development, or a way of rethrowing the exception after the render.


lib/engine/game/base.rb
```diff
@@ -305,6 +305,10 @@ def process_action(action)
         @actions << action
         next_round! while @round.finished? && !@finished
         self
+      rescue GameError => e
+        # Fill in the action id
+        e.action_id = current_action_id
+        raise
```
tobymao, line 311: should we raise game error?

jenf: I thought "raise" bubbled the current exception up

HQ063: it does

jenf: Toby, when you have a mo, could you rereview?

tobymao: so this doesn't in dev?

tobymao: do you have access to rack env?

jenf: it's be in opal world, so no. but i could pass it in somehow.

tobymao: can you just pass it in as a need

PR #396, opened 2020-05-16
Reset password

Talbatross:


michaeljb:
This would fix #149 

(just commenting to link the PR and issue)

assets/app/mail/reset.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def render
+    h(:div, [
+      h(:div, "Hello #{@user['name']},"),
+      h(:br),
+      h(:div, "You've requested a password reset at #{Time.now}"),
+      h(:br),
+      h(:div, "Here is your temporary password: #{@hash}"),
+      h(:text, 'Please '),
+      h(:a, { attrs: { href: "#{@base_url}/reset?user_id=#{@user['id']}" } }, 'click here'),
```
tobymao, line 16: there's no reason to have the user_id in a param here since they will log in with their email anyways

assets/app/view/reset.rb
```diff
@@ -0,0 +1,26 @@
...
+      inputs = [
+        render_input('Temporary Password', id: :hash, type: :password, attrs: { autocomplete: 'temp-password' }),
+        render_input('New Password', id: :password, type: :password, attrs: { autocomplete: 'new-password' }),
+        h(:div, [render_button('Reset Password') { submit }])
+      ]
+      render_form(title, inputs)
+    end
+
+    def submit
+      compiled_params = params
```
tobymao, line 21: let's have them type in their email , temp password and password

routes/user.rb
```diff
@@ -26,6 +26,37 @@ class Api
           login_user(User.create(params))
         end
 
+        # POST '/api/user/forgot'
+        r.is 'forgot' do
+          user = User.by_email(r['email'])
+          if user
```
tobymao, line 32: just do halt(400, 'Could not find email')

routes/user.rb
```diff
@@ -26,6 +26,37 @@ class Api
...
+            puts html
+            { result: true }
+          else
+            { result: false }
+          end
+        end
+
+        # POST '/api/user/reset'
+        r.is 'reset' do
+          user = User.first(id: r['id'])
```
tobymao, line 49: you should lookup by email here

and just halt instead of if else

routes/user.rb
```diff
@@ -26,6 +26,37 @@ class Api
...
+          if !user
+            halt(400, 'Invalid user!')
+          elsif user.reset_hashes.include? r['hash']
+            user.update(password: r['password'])
+            login_user(user)
```
tobymao, line 54: can we store a reset attempt in user session so that people can't spam this?

assets/app/user_manager.rb
```diff
@@ -49,6 +49,22 @@ def logout
     store(:app_route, '/')
   end
 
+  def forgot(params)
+    @connection.safe_post('/user/forgot', params) do |data|
+      break unless data['result'] == true
```
tobymao, line 54: return unless data['result']

models/user.rb
```diff
@@ -25,6 +25,11 @@ def password=(new_password)
     super Argon2::Password.create(new_password)
   end
 
+  def can_reset
```
tobymao, line 28: can_reset?

models/user.rb
```diff
@@ -25,6 +25,11 @@ def password=(new_password)
...
+    reset_time = settings['last_password_reset'].to_i + RESET_WINDOW
+    reset_time < Time.now.to_i
```
tobymao, line 30: settings['last_password_reset']&.to_i || 0 < Time.now.to_i - RESET_WINDOW

it might be nil right?

routes/user.rb
```diff
@@ -26,13 +26,44 @@ class Api
...
+          puts html
+          { result: true }
+        end
+
+        # POST '/api/user/reset'
+        r.is 'reset' do
+          user = User.by_email(r['email'])
+
+          halt(400, 'Invalid email!') unless user
+          halt(400, 'Invalid code!') unless user.reset_hashes.include? r['hash']
```
tobymao, line 52: include?(r['hash'])

routes/user.rb
```diff
@@ -26,13 +26,44 @@ class Api
...
+          halt(400, "You've recently reset your password!  Please try again later!") unless user.can_reset
```
tobymao, line 53: do we need this extra check?

routes/user.rb
```diff
@@ -26,13 +26,44 @@ class Api
...
+
+          user.password = r['password']
+          user.settings['last_password_reset'] = Time.now.to_i
```
tobymao, line 56: shouldn't this be triggered on the forgot?

assets/app/user_manager.rb
Talbatross: rubocop didn't like return.

tobymao:
ok that's fine

Issue #264, opened 2020-05-03
Visual rendering of route should take into account town location


jcoveney:
This is a super minor issue, but still, I figure we can log these and prioritize accordingly (eg not high, but a nice little wart to remove)

<img width="171" alt="Screen Shot 2020-05-03 at 12 29 04" src="https://user-images.githubusercontent.com/830087/80898774-10efec00-8d3a-11ea-80f4-bb5b8b519d52.png">

kelsin:
I'll take a peek at this when I abstract out the town/city location stuff. Right now I don't think the track has any way of knowing where the town ended up being laid. @tobymao does that sound accurate? I think track gets rendered first and right now doesn't know enough about the town's location to divide properly. I'll keep thinking about it as I learn more about the code base. We'll find an answer I'm sure 👍

rheretic:
OP:  I would retry this to see if it's still present.  Quite a bit of code change since you opened this.

tobymao:
no this was not fixed

roseundy:
This should have been fixed with PR #1535

PR #401, opened 2020-05-17
OO changes for 18Chesapeake

### LINKED ISSUE #388: 18Chesapeake - Baltimore (H6) and Philadelphia (J4) hexes need work
Parts are overlapping incorrectly.

Screenshot from 2020-05-14 18-33-41

### LINKED ISSUE #389: 18Chesapeake - OO tiles (X4, X5) have some incorrectly overlapping parts
- X4's revenue should not be on track
- X5's label and revenue should not overlap

Screenshot from 2020-05-14 18-55-18

michaeljb:
- fix -I path for `rake irb`
- add more weight to cities rendered near edge of tiles with new constant
  EDGE_CITY_LOCATIONS (to influence rendering of other tile parts)
- add Tile method `edges_for_city_rendering`, simplify logic in
  `View::Part::City#preferred_render_locations` using the new method
    - cache the values per rotation
- render location adjustments for label, location name, revenue, upgrade
- add background box for location names to help with visiblity when the name
  intersects with track
- better warning message when multiple revenues are found on a tile (include the
  tile name)
- fix bug with upgrades part; pass `@tile` through
- change order of tile part rendering for OO-style tiles only--render the name
  earlier on
- add method `Engine::Part::Base#is?` to help identify which city is the one
  being rendered for OO-style tiles

Fixes #388 
Fixes #389


michaeljb:
Shared some screenshots in Slack, could add here later if there is interest.

michaeljb:
Fixes #388

michaeljb:
@kelsin might be interested in this

michaeljb:
imgur album - https://imgur.com/a/fHTSTlG

michaeljb:
Just added some more to the album above showing all rotations for the three green OO tiles with location name "Philadelphia"

kelsin:
Could this logic be abstracted and update the double dits with it? (town_rect)

michaeljb:
> Could this logic be abstracted and update the double dits with it? (town_rect)

I think so, shouldn't be too complicated. Should I try adding that to this or do it on a new branch/PR? I was thinking separate so this one doesn't get left hanging for too long.

Somewhat related, I think this logic could also be used to render the right part of a curvilinear path to fix #264 (the behavior in that screenshot is from my attempt to just render half the path)

kelsin:
Yeah don't hold this one for that, but want it to be done. I don't like my shitty code coexisting with this much better code.

lib/engine/part/base.rb
```diff
@@ -9,6 +9,12 @@ def id
         "#{tile.id}-#{index}"
       end
 
+      def is?(other)
```
tobymao, line 12: why do you need this can't you just do part == part?

lib/engine/tile.rb
```diff
@@ -214,6 +214,83 @@ def inspect
       "<#{self.class.name}: #{name}, hex: #{@hex&.name}>"
     end
 
+    # returns array where the value at an index is which edge that city (based
+    # on index in self.cities) should be rendered
+    def edges_for_city_rendering
+      @_edges_for_city_rendering ||= {}
+      cached_val = @_edges_for_city_rendering[@rotation]
+      return cached_val unless cached_val.nil?
```
tobymao, line 222: i prefer the @edges_for_city_rendering ||= begin 
end cache

if you want to keep this, invert it and do return cached_val if cached_val

lib/engine/tile.rb
```diff
@@ -214,6 +214,83 @@ def inspect
...
+
+      # array of hashes; index is city index, edges is nums of edges that city
+      # connects to
+      city_edges = cities.map do |city|
```
tobymao, line 226: this can just be
city_edges = Hash.new { |h, k| h[k] = [] }
paths.each  { |p| city_edges[p.city].concat(p.exits) if p.city } 

or something like this

why do you store an array of indices?

tobymao:
this is not tested, nor exactly what you've done, but i think that an approach like this could work and be much simpler. let me know if you want to chat about it

i don't see a reason to store the city index when we have access to the city itself

`
    def preferred_city_edges
      @preferred_city_edges ||=
        begin
          city_edges = Hash.new { |h, k| h[k] = [] }
          edge_count = Hash.new(0)

          paths.each do |path|
            next unless (city = path.city)

            path.exits.each do |edge|
              city_edges[city] << edge
              edge_count[edge] += 1
            end
          end

          sorted = city_edges.each { |_, v| v.sort! }.sort_by { |_, edges| edges }

          sorted.map do |city, edges|
            [city, edges.sort_by { |edge| edge_count[edge] }.first]
          end.to_h
        end
    end
`

lib/engine/part/base.rb
michaeljb: Ran into an issue that I think stemmed from `Engine::Part::City#==` where the different cities on an OO tile evaluated as equal to each other and I needed to differentiate them. May have made more sense to add the `id` (and possibly `tile.id`) check to `==` instead of creating a new method.

Anyway your merge of the new routing stuff you got rid of that `==` overload and everything seems to be working fine with c4c6302

tobymao:
that shouldn’t happen, every city is different

lib/engine/tile.rb
michaeljb: I wanted to handle the edges always starting with the lowest edge that has a path. That edge could be connected to city 0 or city 1, so after sorting, city 1 could end up before city 0, but I needed to keep track of the original order of the cities.

I didn't thoroughly test this out, but my worry was that a tile could be rendered differently depending on which city was called city 0 and I wanted the tile rendering to be consistent regardless of which city was which. (I'm willing to admit worrying about this led to over-engineering :stuck_out_tongue: )

Will check out your code snippet now...

tobymao: well you can just sort by the min edge?

michaeljb:
@tobymao I used your suggestion in 6bcca6b and then added in a new version of my logic to get my tests passing and everything rendering the way I want in 41238cf. It's definitely an improvement over what I originally wrote, thanks for the code snippet to get me started on a simpler path :+1:

lib/engine/tile.rb
michaeljb: Resolved in 6bcca6b and 41238cf

lib/engine/tile.rb
michaeljb: Resolved in 6bcca6b and 41238cf

tobymao:
wow so clear now

PR #418, opened 2020-05-20
Fix OO token teleporting issue when laying green OO tile

michaeljb:
This breaks a couple connection tests, but I don't entirely understand what's happening in those tests; added some questions in comments in aa40d57.

Still need to make green OO tiles upgradable to brown, so I'm marking this as a Draft PR. Additionally I now seem to be unable to run a route through both cities on a OO tile.

In this flie, it is C&A's turn, they cannot upgrade J4 or run through it twice, though both should be possible.
[cannot_run_or_upgrade_oo.json.txt](https://github.com/tobymao/18xx/files/4654826/cannot_run_or_upgrade_oo.json.txt)

### DIFF
--- a/lib/engine/hex.rb
+++ b/lib/engine/hex.rb
@@ -66,17 +66,14 @@ module Engine
       # map old cities to new based on edges they are connected to
       city_map =
         # if @tile is blank, map cities by index
-        if @tile.cities.flat_map(&:connected_edges).empty? && (@tile.cities.size == tile.cities.size)
+        if @tile.cities.flat_map(&:exits).empty? && (@tile.cities.size == tile.cities.size)
           @tile.cities.zip(tile.cities).to_h
-
         # if @tile is not blank, ensure connectivity is maintained
         else
           @tile.cities.map do |old_city|
             new_city = tile.cities.find do |city|
               # we want old_edges to be subset of new_edges
-              old_edges = old_city.connected_edges
-              new_edges = city.connected_edges
-              old_edges & new_edges == old_edges
+              (old_city.exits - city.exits).empty?
             end
             [old_city, new_city]
           end.to_h
--- a/lib/engine/hex.rb
+++ b/lib/engine/hex.rb
@@ -72,21 +72,13 @@ module Engine
         # if @tile is not blank, ensure connectivity is maintained
         else
           @tile.cities.map do |old_city|
-            cities = tile.cities.select do |new_city|
+            new_city = tile.cities.find do |city|
               # we want old_edges to be subset of new_edges
               old_edges = old_city.connected_edges
-              new_edges = new_city.connected_edges
+              new_edges = city.connected_edges
               old_edges & new_edges == old_edges
-            end.compact
-
-            unless cities.one?
-              err_msg = "City #{old_city.index} on old tile maps to "\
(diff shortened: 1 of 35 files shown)


lib/engine/part/city.rb
```diff
@@ -73,6 +74,11 @@ def place_token(corporation)
       def exchange_token(token)
         @tokens[get_slot(token.corporation)] = token
       end
+
+      # returns Array[Integer] of all edges that this city are connected to
+      def connected_edges
+        @connected_edges[tile.rotation] ||= tile.paths.select { |p| p.city == self }.flat_map(&:edges).map(&:num).sort
```
tobymao, line 80: you should never call edges, just call exits

lib/engine/part/city.rb
```diff
@@ -17,6 +17,7 @@ def initialize(revenue, slots = 1)
         @slots = slots.to_i
         @tokens = Array.new(@slots)
         @reservations = []
+        @connected_edges = {}
```
tobymao, line 20: intsead of doing this, let's add to part base an attr called paths

this way every part has a back reference to it's paths

you may need to grep the code a bit to refactor this

lib/engine/hex.rb
```diff
@@ -60,16 +60,46 @@ def name
...
+        else
+          @tile.cities.map do |old_city|
+            cities = tile.cities.select do |new_city|
+              # we want old_edges to be subset of new_edges
+              old_edges = old_city.connected_edges
+              new_edges = new_city.connected_edges
+              old_edges & new_edges == old_edges
+            end.compact
+
+            unless cities.one?
```
tobymao, line 82: it it possible that it maps to another city?

michaeljb: Probably not...got rid of this check in cc2a5d3

lib/engine/part/city.rb
michaeljb: d9d46a6

michaeljb:
@tobymao I fixed the tests and removed my comments, but I haven't gone through the connection code very much yet, so I'm not fully confident my "fix" makes sense - 8a0ff24

lib/engine/part/city.rb
michaeljb: Given that when the parts are created, all the paths are not necessarily created (https://github.com/tobymao/18xx/blob/3d43922ace1febb16591b609272a3cd2ea36d036/lib/engine/tile.rb#L68-L104), would you be fine with the different part classes having a `paths` method that calculates `@paths` based on `@tile`?

tobymao:
hmmm let me think about it, let's chat on slack m aybe

lib/engine/part/city.rb
michaeljb: For example, here's how `Engine::Part::City` could look:

```
def initialize
  # ...
  @paths = {}
end

# ...

def paths
  @paths[tile.rotation] ||= tile.paths.select { |p| p.city == self }
end

def connected_edges
  paths.flat_map(&:exits).sort
end

tobymao:
@michaeljb it makes sense, this error was due to kelsin migrating rotations and i didn't fix the test!

michaeljb:
> @michaeljb it makes sense, this error was due to kelsin migrating rotations and i didn't fix the test!

Perfect, that makes way more sense!

Issue #71, opened 2020-04-17
Disallow duplicate player names in solo mode


dfannius:
In solo mode, the user can create players with identical names, which can confuse code that checks for player equality by comparing names.

sparr:
Is there any other way to check player equality that might be more robust?

strathmeyer:
Reproduction steps:
- go to https://www.18xx.games/new_game
- select "Hotseat"
- change all three player names to "Player"
- click "Create"
- click "Mitsubishi Ferry" and then click "Place Bid". Repeat this multiple times.

Expected: Multiple bids show up under Mitsubishi Ferry.
Actual: Only a single bid shows up, and it is incremented by the different players.

Possible solutions: I agree that making player equality more robust could have other benefits throughout the app, but even if we did that, we shouldn't permit two players to have the same name. So, I'm going to work on a PR that shows an error if you try to submit duplicate names in Hotseat mode.

tobymao:
you can do store(:flash_opts, 'Cannot have duplicate player names') to show the error

PR #424, opened 2020-05-20
Prevent duplicate Hotseat player names

strathmeyer:
Fixes tobymao/18xx#71

After I prevented strict duplicates, I realized it was still possible to add names that would appear identical such as `<space>Player<space>1` and `Player<space><space>1`. So I added a second commit to trim whitespace out of the name. That took the longest because I couldn't figure out why `.gsub(/[[:space:]]+/, " ")` wasn't working. Turns out it's not supported by Opel! 😢 

I haven't added any tests yet. I couldn't find anything existing in the `/spec` folder that mentioned the Hotseat feature or the CreateGame class. Let me know if that's a merge blocker.

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -95,7 +95,7 @@ module View
         players = params
           .select { |k, _| k.start_with?('player_') }
           .values
-          .map { |name| { name: name } }
+          .map { |name| name.gsub(/\s+/, ' ').strip }
 
         if players.any? { |name| players.count(name) > 1 }
           return store(:flash_opts, 'Cannot have duplicate player names')
@@ -114,7 +114,7 @@ module View
         end
 
         create_hotseat(
-          players: players,
+          players: players.map { |name| { name: name } },
           title: params[:title],
           description: params[:description],
           max_players: params[:max_players],
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -97,6 +97,10 @@ module View
           .values
           .map { |name| { name: name } }
 
+        if players.any? { |name| players.count(name) > 1 }
+          return store(:flash_opts, 'Cannot have duplicate player names')
+        end
+
         game_data = params['game_data']
 
         if game_data.empty?


assets/app/view/create_game.rb
```diff
@@ -110,7 +114,7 @@ def submit
         end
 
         create_hotseat(
-          players: players,
+          players: players.map { |name| { name: name } },
```
tobymao, line 117: why is this changed? was it always broken?

assets/app/view/create_game.rb
```diff
@@ -95,7 +95,11 @@ def submit
         players = params
           .select { |k, _| k.start_with?('player_') }
           .values
-          .map { |name| { name: name } }
+          .map { |name| name.gsub(/\s+/, ' ').strip }
```
tobymao, line 98: ehhh i don't think it's necessary to do this since the names are user input

assets/app/view/create_game.rb
```diff
@@ -95,7 +95,11 @@ def submit
...
+
+        if players.any? { |name| players.count(name) > 1 }
```
tobymao, line 100: i think this could be simplified as if players.uniq.size < players.size

tobymao: return store(:flash_opts) if player.uniq.size < players.size

assets/app/view/create_game.rb
strathmeyer: Nope, but when I was doing the whitespace stripping below, I ended up with code like `.map { |name| name[:name].gsub(/\s+/, ' ').strip }` which was just too many "names" in a row.

I thought, "why not map to parameter format down inside `create_hotseat` where it's required, and let the rest of the code deal with the player names as an array of strings." Perhaps I should have renamed the array of strings to `player_names`?

assets/app/view/create_game.rb
strathmeyer: If our stance is that users should avoid Hotseat usernames that result in indistiguishable display strings, we should just close #71 as won't fix.

assets/app/view/create_game.rb
strathmeyer: Sounds like you have a one line fix for #71 in mind. I think it'll save us both time if you commit that directly to master, rather than me updating this PR.

assets/app/view/create_game.rb
tobymao: well the difference is that white space is only confusing whereas duplicate names breaks the game

assets/app/view/create_game.rb
tobymao: ah i see it now thanks, this is fine

assets/app/view/create_game.rb
tobymao: i will just merge this and update it

assets/app/view/create_game.rb
strathmeyer: You're right. I was thinking that having two indistinguishable players also effectively breaks the game for a Hotseat play.

I usually try to avoid solving a bigger problem than the one described in the issue. I'll be extra careful about that in future PRs. 👍

assets/app/view/create_game.rb
strathmeyer: Thank you Toby! 🚂

tobymao: thanks for your first PR, looking forward to your log fixes

PR #425, opened 2020-05-20
View::Log scrolls to bottom via JavaScript

### LINKED ISSUE #298: Cursor keys in log window are backwards
Up-arrow scrolls down and down-arrow scrolls up.

### LINKED ISSUE #346: Copying from log produces lines in reverse order
Possibly related to #298?

If I select a region of the log in my browser (tested in Firefox and Chrome on macOS) and paste it somewhere else, the lines are pasted in reverse order.

strathmeyer:
Log is used in both chat and gameplay logs. The existing code
implemented scroll-to-bottom through CSS, which caused
problems with keyboard scrolling and copy-paste.

This commit removes the scroll-specific CSS, and replaces it
with a scroll_to_bottom lambda that is called in the postpatch
hook.

I'm new to Opal, so I have no idea if this is a reasonable solution!

Fixes #298 
Fixes #346

### DIFF
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -11,6 +11,7 @@ module View
       end
 
       props = {
+        key: 'log',
         hook: {
           postpatch: ->(_, vnode) { scroll_to_bottom.call(vnode) },
         },
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -5,18 +5,16 @@ module View
     needs :log
 
     def render
-      reverse_scroll = lambda do |event|
-        %x{
-          var e = #{event}
-          e.preventDefault()
-          e.currentTarget.scrollTop -= e.deltaY
-        }
+      scroll_to_bottom = lambda do |vnode|
+        elm = Native(vnode)['elm']
+        elm.scrollTop = elm.scrollHeight
       end
 
       props = {
-        on: { wheel: reverse_scroll },
+        hook: {
+          postpatch: ->(_, vnode) { scroll_to_bottom.call(vnode) },
+        },
         style: {
-          transform: 'scaleY(-1)',
           overflow: 'auto',
           height: '200px',
           padding: '0.5rem',
@@ -25,11 +23,11 @@ module View
         },
       }
 
-      lines = @log.reverse.map do |line|
+      lines = @log.map do |line|
         if line.is_a?(String)
-          h(:div, { style: { transform: 'scaleY(-1)' } }, line)
+          h(:div, line)
         elsif line.is_a?(Engine::Action::Message)
-          h(:div, { style: { 'font-weight': 'bold', transform: 'scaleY(-1)' } }, "#{line.entity.name}: #{line.message}")
+          h(:div, { style: { 'font-weight': 'bold' } }, "#{line.entity.name}: #{line.message}")
         end
       end
 
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -3,6 +3,7 @@
 module View
   class Log < Snabberb::Component
     needs :log
+    needs :follow_scroll, default: true, store: true
 
     def render
       scroll_to_bottom = lambda do |vnode|
@@ -10,10 +11,17 @@ module View
         elm.scrollTop = elm.scrollHeight
       end
 
+      scroll_handler = lambda do
+        # can't seem to access the target of the event here
+      end
+
       props = {
         key: 'log',
         hook: {
-          postpatch: ->(_, vnode) { scroll_to_bottom.call(vnode) },
+          postpatch: ->(_, vnode) { scroll_to_bottom.call(vnode) if @follow_scroll },
+        },
+        on: {
+          scroll: scroll_handler
         },
         style: {
           overflow: 'auto',


assets/app/view/log.rb
```diff
@@ -5,18 +5,16 @@ class Log < Snabberb::Component
...
-          e.currentTarget.scrollTop -= e.deltaY
-        }
+      scroll_to_bottom = lambda do |vnode|
+        elm = Native(vnode)['elm']
+        elm.scrollTop = elm.scrollHeight
       end
 
       props = {
-        on: { wheel: reverse_scroll },
+        hook: {
```
tobymao, line 14: wow ok that seems simple if it works

can you add key: i think you generally need that if you're doing hooks

does this scroll to the bottom anytime we patch though? what happens if you're scrolling and you get a new event?

tobymao: for example, if you look at the logs and then someone sends a chat,

strathmeyer: If you're looking at the logs and then someone sends a chat, it does scroll to accommodate the new message, but it only really works if you're scrolled to the bottom of chat. Otherwise, it effectively scrolls down by the amount of the new message.

If you're scrolled up in the log and a new message appears, it should either stay in place or scroll to bottom, but the existing behavior isn't ideal.

I also discovered that this doesn't actually fix #346.  I'm going to close the PR for now and re-open when I find a more complete solution.

strathmeyer: Yes, it scrolls to the bottom any time we patch. If you're scrolling and you get a new event, it scrolls to the bottom. (Existing behavior on master is that a new event scrolls the window by the height of the new event.)

strathmeyer:
I have tested on Chrome, Firefox, and Safari (all OS X). I'd like to test on mobile Chrome and mobile Safari before merging.

tobymao:
ideally i’d like it so that it doesn’t scroll at all or scrolls only by the event like in master. going all the way down is not great. perhaps you can only fire the hook if they are not scrolled?

strathmeyer:
I've tried both `update` and `postpatch`, and both give the same scrollHeight for the old node and the new one, so there's no way to ask "were we already at the bottom?"  I can't find any documentation on hooks that take both an `oldVnode` and a `vnode`:  https://github.com/snabbdom/snabbdom#hooks

If this were React, I'd set state on the element when we scroll up to disable the "scroll to bottom" behavior, and re-enable it when the user scrolled back to the bottom. @tobymao Can you point me to some Snabbdom/Snabberb docs that would help me do the same thing?

tobymao:
So I would have done it the way you would have in React. You can do this by adding an on: scroll event. However, your initial suggesting of checking the scrollHeight is very clean and I like that. It should be easy to do.

Just compare vnode['elm']['scrollTop'] with vnode['elm']['scrollHeight'] if scrollTop < 5 or something for some wiggle room

strathmeyer:
I haven't found a way to compare the previous element with the updated element, so probably the best I can do is to say "if we're less than X away from the bottom after updating, jump to the bottom"

With this math, we wouldn't scroll if someone posted a really big message. I'll have to test it to see what happens when the game puts out a number of messages (e.g., when someone buys the cheapest private company and the auction starts resolving). If it's individual updates, then we should scroll between them, but if it's a single batched update, it might be too big to trigger the scroll. I'll fiddle around and see what I can do.

I do think the ideal solution is to maintain a "followScroll" state on the Log. It starts as True, in which case we use the logic we already have. If the user scrolls up, we set followScroll to false, and if they scroll to the end, we set it back to true.

Is component state possible in Snabbdom?

tobymao:
ok that's fine. i see what you're saying. if a big text comes you're saying that scrollTop would not be 0? i had thought that scrollTop would always be 0 if were were at the bottom but i did not test this

 snabbdom can't do state, but snabberb does https://github.com/tobymao/snabberb

you do this with 

```
store(:scrolled, true, skip: false)
```

strathmeyer:
Thank you for the info about snabberb's state! I'll rework this PR using that. The rest of this post is an explanation of scrollTop, etc for anyone interested.

-----

After we remove the CSS reversing, we're back to the default HTML behavior and scrollTop is 0 when we're all the way at the **top** of the Log. We compare strollTop against scrollHeight (how tall the entire scrollable area is) and clientHeight (how tall the visible area is on screen).

If we disable scrolling logic, here's some output from when the page loads:

```
scrollTop 0
scrollHeight 216
clientHeight 216

scrollTop 0
scrollHeight 1096
clientHeight 216
```

The first output is before the Log has contents. The height on the client is 216, and we only have 216 of content, so there's no scrolling. It follows that we start with a scrollTop of 0, meaning we've scrolled 0px down. The second output is after the Log has contents. The height is still 216 and we've still scrolled 0px, but now the available scroll distance is 1096.

If we type in a new message, we see this update:

```
scrollTop 0
scrollHeight 1114
clientHeight 216
```

Again, our scrollTop and clientHeight haven't changed, but the available scroll has increased by 18px, the size of a line of chat text. If we manually scroll to the bottom, and post a new message, we see:

```
scrollTop 898
scrollHeight 1132
clientHeight 216
```

898 + 216 = 1114. We are now 18px far off the bottom. This is a typical situation to see after a new message has been posted. If we define scrollBottom as the number of pixels from the maximum bottom of the scroll, we can calculate it:

```
scrollBottom = scrollHeight - clientHeight - scrollTop
```

In this PR, when we want to scroll to the bottom, we do `elm.scrollTop = elm.scrollHeight` simply because scrollHeight is a number we know is too big. If there was an easy way to reference JS's `Number.MAX_VALUE` from within Ruby, I'd probably do that.

tobymao:
scrollTop = scrollHeight is fine

assets/app/view/log.rb
```diff
@@ -3,17 +3,25 @@
...
     needs :log
+    needs :follow_scroll, default: true, store: true
 
     def render
       scroll_to_bottom = lambda do |vnode|
         elm = Native(vnode)['elm']
         elm.scrollTop = elm.scrollHeight
       end
 
+      scroll_handler = lambda do
```
tobymao, line 14: lambd do |event|

strathmeyer: I couldn't figure out what to do with the event once I had it. It didn't respond to `inspect`or to `target`.  How does one access the event target element in Snabberb?

I tried:
```
scroll_handler = lambda do |e|
  puts e
end
```

```
scroll_handler = lambda do |e|
  elm = Native(e.target)['elm']
  puts elm.scrollTop
end
```

In general, how do you debug Opal? I tried throwing a `debugger` in there, but that barfed.

tobymao: it will be much easier if you join the slack

the e is a js event so it cannot respond to inspect
you'll have to do

`console.log(e)`

tobymao: use back ticks
https://github.com/opal/opal/blob/master/docs/compiled_ruby.md#inline-javascript

tobymao:
hey @strathmeyer have you made progress on this?

tobymao:
i've decided to merge and fix this, thanks for the early effort

PR #436, opened 2020-05-21
Disallow non-positive spend amounts

### LINKED ISSUE #431: [1889] It's possible to buy a train from another corporation for 0Y
In game 973 one player managed to buy a train from another corporation for 0Y which should be forbidden. The move was undo.

sparr:
Resolves #431

### DIFF
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -433,7 +433,7 @@ module Engine
         @just_sold_company = company
 
         @current_entity.companies << company
-        @current_entity.spend(price, entity)
+        @current_entity.spend(price, entity) if price.positive?
         @log << "#{@current_entity.name} buys #{company.name} from #{entity.name} for #{@game.format_currency(price)}"
       end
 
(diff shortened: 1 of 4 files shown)


lib/engine/round/operating.rb
```diff
@@ -433,7 +433,7 @@ def buy_company(company, price)
         @just_sold_company = company
 
         @current_entity.companies << company
-        @current_entity.spend(price, entity)
+        @current_entity.spend(price, entity) if price.positive?
```
tobymao, line 436: this one shouldn't be here, i'll fix it

lib/engine/round/operating.rb
```diff
@@ -492,7 +492,7 @@ def liquidate(player)
           sell_shares(bundle)
         end
 
-        player.spend(player.cash, @bank)
+        player.spend(player.cash, @bank) if player.cash.positive?
```
tobymao, line 495: this is incorrect as well

lib/engine/round/operating.rb
sparr: $0 is not a valid price in 1889, but in a game where it is valid, this check will be needed.

lib/engine/round/operating.rb
sparr: This one is needed for when a player goes bankrupt with zero dollars

PR #449, opened 2020-05-22
Update /map render tests to check for every string that should be rendered on the map

michaeljb:
This is currently failing due to 18Chesapeake's company "blockers" being broken and not appearing on the map. Once that is fixed, these new/improved tests will pass and will be more useful.


michaeljb:
Instead of a hardcoded array of string literals, I expect it is also possible to generate that list from inspecting a new game object. That would also make it easier to add more games to this test...I may take a crack at that.

michaeljb:
This PR is a failing test for #452

PR #471, opened 2020-05-23
feat: ability to define colors for a game

kelsin:
This uses the colors names from 18xx-maker in the game definition, which WILL pass through if not defined. This means if we don't define colors the svg will just use `red` or `green` and the like. However now you can call `register_colors` to define the colors for this game in the game file. I did it for 1889 and 18Chesapeake.

For 1889 and 18Chesapeake I used the colors currently used.
For 1846 and 1830 I used the GMT color scheme from 18xx-maker.
For 1817 I used every Rails on Board token color (one for each of the 20), we will probably change this if you want to use AAG's colors of course.

At least now changing that is just on your side 👍

### DIFF
--- a/lib/engine/config/game/g_1830.rb
+++ b/lib/engine/config/game/g_1830.rb
@@ -349,7 +349,7 @@ module Engine
       "float_percent": 60,
       "sym": "B&O",
       "name": "Baltimore & Ohio Railroad",
-      "logo": "1830/B&O",
+      "logo": "1830/BO",
       "tokens": [
         0,
         40,
@@ -362,7 +362,7 @@ module Engine
       "float_percent": 60,
       "sym": "C&O",
       "name": "Chesapeake & Ohio Railroad",
-      "logo": "1830/C&O",
+      "logo": "1830/CO",
       "tokens": [
         0,
         40,
@@ -400,7 +400,7 @@ module Engine
       "float_percent": 60,
       "sym": "B&M",
       "name": "Boston & Maine Railroad",
-      "logo": "1830/B&M",
+      "logo": "1830/BM",
       "tokens": [
         0,
         40
--- a/lib/engine/config/game/g_1830.rb
+++ b/lib/engine/config/game/g_1830.rb
@@ -315,7 +315,7 @@ module Engine
         100
       ],
       "coordinates": "H12",
-      "color": "#237333"
+      "color": "green"
     },
     {
       "float_percent": 60,
@@ -329,7 +329,7 @@ module Engine
         100
       ],
       "coordinates": "E19",
-      "color": "#37383a"
+      "color": "black"
     },
     {
       "float_percent": 60,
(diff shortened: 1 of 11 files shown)


lib/engine/config/game/g_1889.rb
```diff
@@ -231,7 +231,7 @@ module G1889
       "name": "Mitsubishi Ferry",
       "value": 30,
       "revenue": 5,
-      "desc": "Player owner may place the port tile on a coastal town (B11, G10, I12, or J9) without a tile on it already, outside of the operating rounds of a corporation controlled by another player. The player need not control a corporation or have connectivity to the placed tile from one of their corporations. This does not close the company.",
+      "desc": "Player owner may place the port tile on a coastal town (B11, G10, I12, or J9) without a tile on it already, outside of the operating rounds of a corporation controlled by another player. The player need not control a corporation or have connectivity to the placed tile from one of their corporations. This does not close the corporation.",
```
kelsin, line 234: I fixed my description but looks like you missed one `company` to `corporation` so fixing.

lib/engine/config/game/g_18_chesapeake.rb
```diff
@@ -691,7 +697,8 @@ module G18Chesapeake
       "tiles": [
         "yellow",
         "green",
-        "brown"
+        "brown",
+        "gray"
```
kelsin, line 701: This was a bug pointed out in chat, doesn't look fixed yet in the code so added here.

lib/engine/game/base.rb
```diff
@@ -79,6 +79,14 @@ def self.title
         name.split('::').last.slice(1..-1)
       end
 
+      def self.register_colors(colors)
+        colors.default_proc = proc do |_, key|
+          key
+        end
```
kelsin, line 85: This makes sure that when you don't define a certain color it just passes the color name through (which works fine in SVG)

lib/engine/config/game/g_1830.rb
```diff
@@ -343,33 +343,33 @@ module G1830
...
       "coordinates": "I15",
-      "color": "#0189d1"
+      "color": "blue"
     },
     {
       "float_percent": 60,
       "sym": "C&O",
       "name": "Chesapeake & Ohio Railroad",
-      "logo": "1830/CO",
+      "logo": "1830/C&O",
```
tobymao, line 365: we can''t use & in logos anymore, i think nginx strips them out or something

lib/engine/config/game/g_1846.rb
```diff
@@ -237,34 +237,36 @@ module G1846
...
       "coordinates": "D20",
-      "color": "#37383a"
+      "color": "black"
     },
     {
       "float_percent": 20,
       "sym": "B&O",
       "name": "Baltimore & Ohio",
-      "logo": "1846/BO",
+      "logo": "1846/B&O",
```
tobymao, line 246: remove all &

lib/engine/config/game/g_1889.rb
tobymao: no this references the company as corporations don't close

lib/engine/config/game/g_1830.rb
kelsin: Done.

lib/engine/config/game/g_1889.rb
kelsin: Yep.... dumb. Fixed.

PR #476, opened 2020-05-23
[issue_474] game_card: comma-separate player names

### LINKED ISSUE #474: Player names should be comma-separated in the Games display
Originally reported in #295 but I wanted to split it out so I could fix it on its own.

dragon12:
fixes #474

Please let me know if there's a more idiomatic way of doing this. I tried post-processing it, but this resulted in an extra space between the name and the succeeding comma:
`p_elm = p_elm.zip((',' * (p_elm.size - 1)).chars).flatten.compact`

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -137,11 +137,11 @@ module View
 
       index = players.size
       p_elm = players.map do |player|
-        index = index - 1
+        index -= 1
         elm = h(
           acting?(player) ? :u : :span,
           { style: { 'margin-right': '0.5rem' } },
-          player['name'] + (index > 0 ? ',' : ''),
+          player['name'] + (index.positive? ? ',' : ''),
         )
 
         if owner? && new? && player['id'] != @user['id']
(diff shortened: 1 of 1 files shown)


dfannius:
I imagine something like `p_elm.join(', ')` (untested) would work.

dragon12:
> I imagine something like `p_elm.join(', ')` (untested) would work.

That was the very first thing I tried. Something about the elements inside p_elm mean that doesn't work:

> error	"to_str: [object Object] doesn't respond to #to_str, #to_ary or #to_s"

tobymao:
as discussed should you remove the p_elm.join since it will crash?

dragon12:
> as discussed should you remove the p_elm.join since it will crash?

heh, yeah, sorry, that got accidentally included in the commit when I was demonstrating the error message I pasted earlier. Fixed now.

tobymao:
sorry for being picky but could you change it to

`
players.each_with_index do |player, index|
`

dragon12:
> sorry for being picky but could you change it to
> 
> `players.each_with_index do |player, index|`

Please, be as picky as you can be. Ruby's relatively foreign to me so anything more idiomatic is appreciated.

I had to refactor a little because each_with_index doesn't return a new array, let me know if that's what you had in mind

tobymao:
you can do map.with_index

dragon12:
> you can do map.with_index

so you can, thanks

PR #478, opened 2020-05-23
Enforce maximum bid (fixes #460)

### LINKED ISSUE #460: Set aside money in private bids
Money bid for privates should be set aside and not made available until that private's auction is resolved. Right now a player can bid up to their entire current bankroll even if they have outstanding bids on other privates. This can also lead to stuck states where, for example, a player has placed the only bid on P6, and wins P5 with a gigantic bid that doesn't leave enough money for P6 to resolve.

dfannius:


lib/engine/round/auction.rb
```diff
@@ -75,6 +75,23 @@ def company_pending_par
         @companies_pending_par[0]
       end
 
+      def committed_cash(player)
+        bids_for_player(player).map(&:price).sum
+      end
+
+      def current_bid_amount(player, company)
```
tobymao, line 82: bids[company]&.find { |b| b.entity == player }&.price || 0

lib/engine/round/auction.rb
```diff
@@ -75,6 +75,23 @@ def company_pending_par
         @companies_pending_par[0]
       end
 
+      def committed_cash(player)
```
tobymao, line 78: bids_for_player(player).sum(&:price)

assets/app/view/auction_round.rb
```diff
@@ -57,7 +57,7 @@ def render_input
         value: @round.min_bid(@selected_company),
         step: step,
         min: @selected_company.min_bid + step,
-        max: @current_entity.cash,
+        max: @round.max_bid(@current_entity, @selected_company),
```
tobymao, line 60: looks like this a UI only fix?

that's fine for me but making sure that's your intent

dfannius: My impression is that we often enforce rules only at the UI level (e.g., not giving you a pass button if you need to sell shares), so it seemed okay to me.

tobymao: ideally it would be in both places, but sometimes i'm afraid to do it in the server because it will break too many games... but the future is here now

PR #537, opened 2020-05-27
Support for blue hexes

michaeljb:
Necessary for 1846, useful for 18Chesapeake--render impassable water hexes instead of having an apparent gap in the map.

Probably necessary for the offboard port hex in 1882 (conceivably it could be a normal red offboard with some other indicator that it's a special port thing, but that would be suboptimal).

For 1846 and 18Chesapeake, routes should not be allowed to point to blue hexes, but they will have to in 1882.

What I've pushed here only draws the blue hex, it doesn't prevent track from being pointed to it, or prevent the hex from being clicked on during the track laying step (and doing so completely breaks the game)

Screenshot from 2020-05-27 10-37-27

### DIFF
--- a/assets/app/view/hex.rb
+++ b/assets/app/view/hex.rb
@@ -25,6 +25,7 @@ module View
       brown: '#cb7745',
       gray: '#bcbdc0',
       red: '#ec232a',
+      blue: '#00f',
     }.freeze
 
     needs :hex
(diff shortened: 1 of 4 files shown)


gchambers419:
If/when this happens, can I vote to add the water hexes to 1889?  It makes "what towns can be ports" check much more intuitive.

michaeljb:
Don't think I wrote this down anywhere, but I was thinking of adding the port icon to 1889 once we have it for 1846. If that wouldn't be enough, we could definitely consider adding blue hexes on the map.

PR #553, opened 2020-05-28
Reload trains last routes when entering run routes

### LINKED ISSUE #286: Display/save previously ran routes
Somewhat related to #283.

What if the routes were depicted by large dot at the end of route and solid line between those dots. Multiple trains would be somewhat simple to differentiate with the large dots. This dot - line - dot could be the color of the company and visible / togleable during opponent turns.

Then if this route is cut with a token, the rest of the route could turn red, meaning it has to be planned again.

Other could see what you ran instead of trying to parse the log.

jenf:
fixes #286 

There's no button to reload the route, you can refresh the browser to do that if someone so desires

### DIFF
--- a/assets/app/view/route_selector.rb
+++ b/assets/app/view/route_selector.rb
@@ -17,8 +17,31 @@ module View
       @routes.select { |r| r.connections.any? }
     end
 
+    def last_run
+      trains = @game.round.current_entity.trains
+      last_run = @game.round.current_entity.last_run
+      return [] unless last_run
+
+      last_run.map do |train, connections|
+        connection_hexes = connections&.map do |ids|
+          ids.map { |id| @game.hex_by_id(id) }
+        end
+        # A future enhancement to this could be to find trains and move the routes over
+        Engine::Route.new(@game.phase, train, connection_hexes: connection_hexes) if trains.include?(train)
+      end.compact
+    end
+
     def render
       trains = @game.round.current_entity.trains
+
+      if @routes.empty?
+        @routes = last_run
+        if @routes.any?
+          @selected_route = @routes.first
+          store(:routes, @routes, skip: true)
+          store(:selected_route, @selected_route, skip: true)
+        end
+      end
 
       if !@selected_route && (first_train = trains[0])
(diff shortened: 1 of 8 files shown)


lib/engine/corporation.rb
```diff
@@ -33,6 +33,7 @@ def initialize(sym:, name:, tokens:, **opts)
       @trains = []
       @companies = []
       @revenue_history = {}
+      @last_run = nil
```
tobymao, line 36: can we store all the runs and replace revenue history with this

lib/engine/corporation.rb
```diff
@@ -102,6 +103,12 @@ def corporation?
       true
     end
 
+    def add_run!(run)
```
tobymao, line 106: probably need to do like OR, Train, Hexes

assets/app/view/route_selector.rb
```diff
@@ -17,9 +17,32 @@ def active_routes
...
+        end
+        # A future enhancement to this could be to find trains and move the routes over
+        Engine::Route.new(@game.phase, train, connection_hexes: connection_hexes) if trains.include?(train)
+      end.compact
+    end
+
     def render
       trains = @game.round.current_entity.trains
 
+      if @routes.empty?
```
tobymao, line 37: this looks awkward, the code will probably change when we move last_revenue to all revenues but if you were to keep this

it should be like

@routes = last_run if @routes.empty

...

lib/engine/corporation.rb
jenf: yeah, but we need to store the connection_hexes in textual form due to upgrades on tiles.

I'll add a new class to store this info and replace it with @operating_history

assets/app/view/route_selector.rb
jenf: You'd need to store the routes and selected routes so the map updates correctly.

lib/engine/corporation.rb
tobymao: you can store both the text form and maybe a property for revenue

jenf: yep, see the diff :)

PR #574, opened 2020-05-30
change white to tobytan and delete 3 blue hexes

scottredracecar:


scottredracecar:
Change to BDFL tobytan and delete three blue hexes from 18Chesapeake

Issue #565, opened 2020-05-29
Improve forced purchase UI


jenf:
Often users are confused on the forced purchase UI

* Text should reflect that the user must purchase a train
* Should say how much extra cash the player must get in order to buy the train
* When clicking on the stock the 'Sell Share' button should be next to the share you click on
* It'd be nice if there was an explanation of why you couldn't sell shares (although this should be separated into a different issue)

PR #579, opened 2020-05-30
Improve UI of emergency purchase

### LINKED ISSUE #551: 18Chesapeake: Able to force bankruptcy when president have enough liquid shares to buy the cheapest train
[game.txt](https://github.com/tobymao/18xx/files/4696971/game.txt)

At id 370 I can force bankruptcy with the "Declare bankruptcy button"

This should only happen when the presedent sell every legal shares and still can't buy the cheapest train in bank or bank pool

rules:

> _If the company is still unable to buy a train after
> the president has contributed all possible funds and sold all
> shares that can be legally sold and no other company is
> willing to sell one of its trains to the company for an
> agreeable price, then the player is declared bankrupt and
> the game is immediately ended._

jenf:
* Show how much the player needs to get
* Only show declare bankrupcy if the player liquidate enough
* Align the shares sell buttons under the share
* Fix seeing other trains after selling shares
* Move shares above trains if player needs to contribute

fixes #551 and #565
Screenshot from 2020-05-30 12-29-05
Screenshot from 2020-05-30 12-28-20

### DIFF
--- a/assets/app/view/buy_trains.rb
+++ b/assets/app/view/buy_trains.rb
@@ -9,6 +9,56 @@ module View
   class BuyTrains < Snabberb::Component
     include Actionable
     needs :show_other_players, default: nil, store: true
+    needs :selected_corporation, default: nil, store: true
+
+    def can_buy_others
+      !@game.actions.last.is_a?(Engine::Action::SellShares)
+    end
+
+    def render_president_contributions
+      player = @corporation.owner
+      funds_required = @depot.min_depot_price - (@corporation.cash + player.cash)
+      raisable = player.shares_by_corporation.sum do |corporation, shares|
+        next 0 if shares.empty?
+
+        last = @game.round.sellable_bundles(player, corporation).last
+        last ? last.price : 0
+      end
+
+      children = []
+
+      if raisable > funds_required
+        children << h('div.margined',
+                      'The company cannot afford the cheapest train from The Depot or The Discard,'\
+                      " president must raise #{@game.format_currency(funds_required)}")
+
+        props = {
+          style: {
+            display: 'inline-block',
+            'vertical-align': 'top',
+          }
+        }
+
+        player.shares_by_corporation.each do |corporation, shares|
+          next if shares.empty?
+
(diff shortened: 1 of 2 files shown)


assets/app/view/buy_trains.rb
```diff
@@ -9,6 +9,56 @@ module View
   class BuyTrains < Snabberb::Component
     include Actionable
     needs :show_other_players, default: nil, store: true
+    needs :selected_corporation, default: nil, store: true
+
+    def can_buy_others
```
tobymao, line 14: this is actually not true for chesapeake

assets/app/view/buy_trains.rb
```diff
@@ -9,6 +9,56 @@ module View
...
+      !@game.actions.last.is_a?(Engine::Action::SellShares)
+    end
+
+    def render_president_contributions
+      player = @corporation.owner
+      funds_required = @depot.min_depot_price - (@corporation.cash + player.cash)
+      raisable = player.shares_by_corporation.sum do |corporation, shares|
```
tobymao, line 21: i think there's a new method called liquidity

assets/app/view/buy_trains.rb
jenf: Gotcha will look and fix

assets/app/view/buy_trains.rb
jenf: Yeah, unfortunately it counts shares that would cause the presidency to shift which aren't allowed in '89

assets/app/view/buy_trains.rb
tobymao: we'd have to change the game logic too, which is why i'm wafflin

assets/app/view/buy_trains.rb
tobymao: can that be added as an option to the call?

assets/app/view/buy_trains.rb
jenf: Yep, it'll need to move down a layer then.

PR #581, opened 2020-05-30
user-selected background & font colors (aka "dark mode")

yzemaze:
**Caveat**: This is my first time coding ruby. Just let me know if there’s anything ugly, problematic etc. E. g. `'background-color': @user&.dig(:settings, :bg_color)` in `assets/app/app.rb` feels a bit wonky to me. Same goes for two methods to render bg and font color buttons in `assets/app/view/user.rb` and changes in `routes/user.rb` in general. I didn’t know better and had to resort to copy & paste. Any feedback would be appreciated.
That said the changes do work fine locally and RuboCop was happy, too – at last ;)

I’ve not implemented prefers-color-scheme because on an involved site like this it’s imho way better to let users decide.

**Features & changes:**
- background and font color buttons in user profile
- color: black on light backgrounds added where not set
- color: black on other backgrounds replaced by currentColor to display in font_color chosen by user
- logo: white bg
- About and Profile links in font_color chosen by user
- button border set to gray to work better on dark bg and don’t annoy on light/white bg [button hover bg (black) not changed]
- game card border set to text color for contrast
- explicit gray text color replaced by opacity [spreadsheet.rb]
- app min-height: 98vh to stretch background to near total height, else bg only covers parts of the height. 100vh often entails problems (on mobile, scroll bars etc.).
- stock-market cells’ bg explicitely set to white when no price.color set

**todo?**
Maybe increase opacity in `assets/app/view/hex.rb` L104 to 0.5? 0.3 is a tad to dark for my taste on dark bgs and 0.5 seems to work fine on white bg, too.
To prevent problems with differing user preferences in the future:
`color: black` on light backgrounds
`background-color: white` where desired
`color: currentColor` to get regular text-color (if not overriden by parent)


assets/app/app.rb
```diff
@@ -27,8 +27,11 @@ def render
     props = {
       props: { id: 'app' },
       style: {
-        padding: '0.5rem',
+        'background-color': @user&.dig(:settings, :bg_color),
```
tobymao, line 30: what happens when this is nil?

assets/app/view/company.rb
```diff
@@ -75,8 +76,10 @@ def render
         },
         on: { click: onclick },
       }
-
-      props[:style]['background-color'] = 'lightblue' if selected?
+      if selected?
```
tobymao, line 79: we could make selected a class, you could add it to index.rb

then do
h("div.selected_entity") to add the class or something

assets/app/app.rb
yzemaze: In theory it should be an "invalid property value" like any other => no background-color set. That’s exactly what happens if no user is logged in.

assets/app/view/company.rb
yzemaze: Well, if you’d like to use more CSS/SASS I wouldn’t mind. I took your answers yesterday as "avoid CSS whenever possible" ;)

tobymao: haha, no i mean you can use it when it's necessary or beneficial

yzemaze: Ok, got it. It’s actually ok in this case.
There are some other instances where it’s actually easier, less errorprone and more maintenance-friendly to use CSS. All those "color: black"s e. g. and other redundant styles, but I didn’t want to push "my agenda" with the first PR ;-)

tobymao: i wouldn't see it that way unless you rewrote the whole thing in css.

if it makes the code cleaner and less redundant to use css, by all means do it.
for example i have a css class called margined or button etc

yzemaze: At least it makes for cleaner and more legible HTML. CSS tends to get very convoluted for larger projects, so it’s a slippery slope for sure. So I do get why you prefer another way, even when it’s not dynamically changed.
But having more CSS in one place might motivate someone without any ruby knowledge to improve the design/presentation of the site. (Back in April I would’ve put in a PR instead of an issue – if there had been a simple .css.)
I might tackle CSS after this PR and tile/map-themes are done.

PR #582, opened 2020-05-30
Random non-president's share, plus excluded corps (for 1882)

dfannius:
Also did a pass on privates text.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -495,15 +495,13 @@ module Engine
             company.desc = "#{company.desc} The random corporation in this game is #{corporation.name}."
             @log << "#{company.name} comes with the president's share of #{corporation.name}"
           when 'random'
-            exclusions = ability[:'random-share-exclude']&.split
-            corporation = nil
-            while corporation.nil? || exclusions&.include?(corporation.name)
-              corporation = @corporations[rand % @corporations.size]
-            end
-            share = corporation.shares[1] # not president
+            corp_strs = ability[:'random-share-corps']&.split
+            corporations = @corporations.select { |c| corp_strs&.include?(c.name) } || @corporations
+            corporation = corporations[rand % corporations.size]
+            share = corporation.shares.find { |s| !s.president }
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -477,11 +477,21 @@ def init_company_abilities
...
           when 'random-president'
-            share = @corporations[rand % @corporations.size].shares[0]
+            corporation = @corporations[rand % @corporations.size]
+            share = corporation.shares[0]
             ability[:share] = share
-            corporation = share.corporation
             company.desc = "#{company.desc} The random corporation in this game is #{corporation.name}."
             @log << "#{company.name} comes with the president's share of #{corporation.name}"
+          when 'random'
+            exclusions = ability[:'random-share-exclude']&.split
```
tobymao, line 486: i think it's easier to do inclusions as opposed to exclusions, thoughts?

inclusion also follows the pattern of other privates we have like what tiles and hexes to lay

lib/engine/game/base.rb
```diff
@@ -477,11 +477,21 @@ def init_company_abilities
...
+            corporation = nil
+            while corporation.nil? || exclusions&.include?(corporation.name)
```
tobymao, line 488: we can just  selected the included corps and then do a rand instead of this while loop

lib/engine/game/base.rb
```diff
@@ -477,11 +477,21 @@ def init_company_abilities
...
+              corporation = @corporations[rand % @corporations.size]
+            end
+            share = corporation.shares[1] # not president
```
tobymao, line 491: it's probbaly safe but if you want to be more robust, corporation.shares.find { |s| !s.president }

lib/engine/game/base.rb
scottredracecar: Agree on inclusions. That also how I rewrote it in the rules text to ease setup and understanding!

dfannius:
PR updated.

Issue #591, opened 2020-05-31
Wrong value in description for Big 4


gchambers419:
https://github.com/tobymao/18xx/blob/57e6b9afca45a875dc23ffdfef91d89f7f92160e/lib/engine/config/game/g_1846.rb#L165

Company starts with $40 in treasury, not $60.

PR #592, opened 2020-06-01
Fix descriptions for Independent Railroads.

gchambers419:
1. Big 4 starting treasury is $40 not $60
2. Mention debt that must be paid off when purchasing.


gchambers419:
Fixes issue #591

PR #593, opened 2020-06-01
Support for "pointy" maps

michaeljb:
* tweak axes for flat, fix completely for pointy
* add "layout" property to game JSON (TODO: add this in 18xx-maker/export-rb)
* update 1846 map config (TODO: add this in 18xx-maker, and also include updates for Chi tiles)

Some views can use `View::Part::Base#rotation_for_layout` or `View::Part::Base#angle_for_layout`; `Label` and `LocationName` have some additional movements applied depending on the layout, but currently they each handle this a little differently. `Label` has all of the preferred locations in constants at the top of the class, `LocationName` has some methods for the locations--this is required to some extent by its use of the `delta_y` method to adjust the `y` positioning of the text based on how many lines there are in the name. 

Screenshot from 2020-05-31 18-45-36
Screenshot from 2020-05-31 18-44-56


assets/app/view/axis.rb
```diff
@@ -41,45 +41,66 @@ def col_labels
       hex_x, hex_y = hex_size
 
       labels = @cols.map do |c|
-        x = (hex_x / 2) + (hex_x * (c.to_i - 1))
+        x = hex_x * (c.to_i - 1)
+        y = 0
         label = ('A'..'Z').to_a[c - 1]
-        h(:text, { attrs: { x: x, y: @font_size } }, label)
+        h(:text, { attrs: { x: x, y: y } }, label)
```
tobymao, line 47: do you even need y if it's always 0?

assets/app/view/axis.rb
```diff
@@ -41,45 +41,66 @@ def col_labels
...
     def row_labels
       hex_x, hex_y = hex_size
 
       labels = @rows.map do |row|
-        y = (hex_y * row.to_i)
-        h(:text, { attrs: { x: @font_size, y: y } }, row)
+        x = 0
+
+        multiplier = row.to_i
+        multiplier -= (1 / 2) if @layout == :pointy
```
tobymao, line 77: you mean 0.5 ? 

1 / 2 evaluates to 0

tobymao: or do you mean multiplier * 0.5 ?

michaeljb: I did mean 0.5, got lucky with JS not evaluating it to 0 where Ruby would

assets/app/view/axis.rb
michaeljb: Apparently not

PR #595, opened 2020-06-01
1846 IRR discounts and new svgs

scottredracecar:


scottredracecar:
This applies a negative discount to the independent railroads, edits the wording of them, and revises the logos to better match the board game and colors.

PR #598, opened 2020-06-01
dark mode

yzemaze:
**Features & changes:**
- background and font color buttons in user profile (immediate change on save)
- color: black on light backgrounds added where not set
- color: black on other backgrounds replaced by currentColor to display in font_color chosen by user
- logo-component: dynamic change of polygon fill and text color dependent on prefers-color-scheme: dark or user-defined bg
- About and Profile links in font_color chosen by user
- brightness calculation
- button border set to gray to work better on dark bg and don’t annoy on light/white bg [button hover bg (black) not changed]
- game card border set to text color for contrast
- explicit gray text color replaced by opacity [spreadsheet.rb]
- app min-height: 98vh to stretch background to near total height, else bg only covers parts of the height. 100vh often entails problems (on mobile, scroll bars etc.).
- stock-market cells’ bg explicitely set to white when no price.color set

**todo?**
Maybe increase opacity in assets/app/view/hex.rb L104 to 0.5? 0.3 is a tad to dark for my taste on dark bgs and 0.5 seems to work fine on white bg, too.

To prevent problems with differing user preferences in the future:
`color: black` on light backgrounds
`background-color: white` where desired
`color: currentColor` to get regular text-color (if not overriden by parent)


assets/app/index.rb
```diff
@@ -5,6 +5,32 @@ def render
...
+          background-color: black;
+          color: #eee;
+        }
+        .nav #logo--dark path {
+          fill: #eee;
+        }
+        .nav #logo--dark polygon {
+          fill: #ec232a;
+        }
+        .nav__links a {
```
tobymao, line 19: why is this double under score?

assets/app/view/logo.rb
```diff
@@ -0,0 +1,51 @@
...
+        attrs: {
+          width: '500px',
+          height: '198px',
+          viewBox: '0 0 500 198',
+        },
+      }
+
+      bg_color = @user&.dig(:settings, :bg_color) || '#ffffff'
+      logo_id = bg_color == '#ffffff' ? 'logo--dark' : 'logo'
+      # green = '#71bf44'
```
tobymao, line 28: remove the comment

assets/app/view/logo.rb
```diff
@@ -0,0 +1,51 @@
...
+      offboard = '#ec232a'
```
tobymao, line 29: you can make these constants in the class

assets/app/index.rb
```diff
@@ -5,6 +5,32 @@ def render
     css = <<~CSS
       * { font-family: 'Inconsolata', monospace; }
 
+      @media (prefers-color-scheme: dark) {
+        body {
+          background-color: black;
+          color: #eee;
+        }
+        .nav #logo--dark path {
```
tobymao, line 13: can we get rid of the doubles and call it 
#logo_dark

assets/app/view/logo.rb
```diff
@@ -0,0 +1,51 @@
...
+        path_fill = '#eee'
+      else
+        polygon_fill = yellow
+        path_fill = '#000'
+      end
+
+      h('svg#logo', props, [
+        h(:title, '18xx.Games'),
+        h(:g, { attrs: { id: logo_id } }, [
+          h(:polygon, attrs: { id: 'logo__polygon', fill: polygon_fill, points: '250 0 335.736515 49.5 335.736515 148.5 250 198 164.263485 148.5 164.263485 49.5' }),
```
tobymao, line 42: it seems like there are only 2 colors? i thought there would be infinite many which is why we choose to make it a class, if it's just two colors, should we just make it 2 static svgs?

assets/app/index.rb
yzemaze: BEM naming convention, it’s pretty handy in regards to CSS. https://www.freecodecamp.org/news/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849/

assets/app/index.rb
yzemaze: We could, but again BEM naming convention.

assets/app/view/logo.rb
yzemaze: The polygon has 2 colors only as of now, I could switch this to infinite if you’d wish.

assets/app/index.rb
tobymao: i see, although we don't use it anywhere else, so i'd prefer to change just to snake case

assets/app/view/logo.rb
yzemaze: Will do and remove #

assets/app/view/logo.rb
tobymao: if it's only going to be 2 colors, i'd rather have it as 2 svgs. if you're going to do all this work to make it inline, might as well make use of it.

yzemaze: Sure. Polygon still as backdrop or can I move it to the left of 18xx.Games?

PR #602, opened 2020-06-02
Switch enter button to link to allow opening in a new tab, closes #156

### LINKED ISSUE #156: Ability to open game in new tab

wheresvic:
This switches the `enter` button to use a Link class with an `href` attribute which can then be used to right-click and open the game in a new window/tab.


jenf:
Initial review looks okay, but Toby will have an opinion

Issue #291, opened 2020-05-05
Spreadsheet - sort corps by Market value


8bg:
To allow for quickly identifying corp OR turn order. Or possibly offer different sort option by clicking column headers.

lbrande:
I really think this would reduce the usability of the spreadsheet view, it can be very confusing if companies move around.

8bg:
I find the current view frustrating after years of rr18xx use where the default sort is by operating order/market value. The header sort option I mentioned would accommodate different preferences.

lbrande:
As long as it's an option it should be fine. As a Board18 player i would find it very frustrating if companies moved around.

ptrinidad:
I'd also feel more comfortable if companies are sorted in operating order. However I understand other that prefer keeping a fixed order from the beginning of the game. Since this is an opinionated feature, I think that offering both options is the best choice.

sparr:
"Obviously" the one true sort order for corporations is (alphabetical|par order|float order|stock value|operating order|...)

+1 for just making the table sortable by any column, like a spreadsheet.

michaeljb:
> +1 for just making the table sortable by any column, like a spreadsheet.

My friend just echoed this request. His exact words:

> Would be cool if you could sort the spreadsheet by varying factors. High low values in different columns etc

Tin-Whisker:
I highly encourage the idea of sorting by not just market value but operating order.  Having one place where  you can see corp, money, and trains makes it easier to predict what trains are going to be bought by whom.  

So if you are to allow sorting on a column please add a column for operating order.   Thanks.

wheresvic:
I can try and tackle this one :)

PR #612, opened 2020-06-02
Add company operating order to spreadsheet

wheresvic:
So this is part of #291 where we first implement an operating order column for the spreadsheet tab. Companies that have already operated are marked with an `*`. Unfloated companies do not have an order and neither do companies in a stock/auction round.

See screenshots:

[https://postimg.cc/SYz8Lhrf](https://postimg.cc/SYz8Lhrf)
[https://postimg.cc/BLzH16xv](https://postimg.cc/BLzH16xv)

In the next iteration we could tackle sorting on the UI :)

### DIFF
--- a/assets/app/view/spreadsheet.rb
+++ b/assets/app/view/spreadsheet.rb
@@ -97,8 +97,7 @@ module View
     end
 
     def render_corporations
-      current_round = [@game.turn]
-      current_round << @game.round.round_num if @game.round.name == 'Operating Round'
+      current_round = @game.round.turn_round_num
 
       floated_corporations = @game.round.entities
       @game.corporations.map do |c|
@@ -128,7 +127,9 @@ module View
       operating_order_text = ''
       if operating_order.positive?
         operating_order_text = operating_order.to_s
-        operating_order_text += '*' if corporation_operated?(corporation, current_round)
+        corporation.operating_history.each do |history|
+          operating_order_text += '*' if history[0] == current_round
+        end
       end
 
       h(:tr, props, [
@@ -182,13 +183,6 @@ module View
         h(:th, 'Certs'),
         *@game.players.map { |p| h(:td, p.num_certs) },
       ])
-    end
-
-    def corporation_operated?(corporation, current_round)
-      corporation.operating_history.each do |history|
-        return true if history[0] == current_round
-      end
-      false
     end
   end
 end
--- a/assets/app/view/spreadsheet.rb
+++ b/assets/app/view/spreadsheet.rb
@@ -86,6 +86,7 @@ module View
           h(:th, props, 'IPO'),
           h(:th, props, 'Market'),
           h(:th, props, 'Cash'),
+          h(:th, props, 'Operating Order'),
           h(:th, props, 'Trains'),
           h(:th, props, 'Tokens'),
           h(:th, props, 'Privates'),
@@ -96,10 +97,20 @@ module View
     end
 
     def render_corporations
-      @game.corporations.map { |c| render_corporation(c) }
+      current_round = [@game.turn]
+      current_round += [@game.round.round_num] if @game.round.name == 'Operating Round'
+
+      floated_corporations = @game.round.entities
+      @game.corporations.map do |c|
+        operating_order = -1
+        operating_index = floated_corporations.find_index(c)
+        operating_order = operating_index + 1 if operating_index
+
+        render_corporation(c, operating_order, current_round)
+      end
     end
 
-    def render_corporation(corporation)
+    def render_corporation(corporation, operating_order, current_round)
       name_props =
         {
           style: {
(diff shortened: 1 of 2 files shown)


assets/app/view/spreadsheet.rb
```diff
@@ -96,10 +97,20 @@ def render_title
     end
 
     def render_corporations
-      @game.corporations.map { |c| render_corporation(c) }
+      current_round = [@game.turn]
+      current_round += [@game.round.round_num] if @game.round.name == 'Operating Round'
```
tobymao, line 101: current_round << @game.round.round_num if ...
current_round = current_round.join('.')

assets/app/view/spreadsheet.rb
```diff
@@ -96,10 +97,20 @@ def render_title
...
+
+      floated_corporations = @game.round.entities
+      @game.corporations.map do |c|
+        operating_order = -1
```
tobymao, line 105: this seems useless

assets/app/view/spreadsheet.rb
```diff
@@ -96,10 +97,20 @@ def render_title
...
+        operating_index = floated_corporations.find_index(c)
+        operating_order = operating_index + 1 if operating_index
```
tobymao, line 107: operating_order = (floated_corporations.find_index(c) || -1) + 1

assets/app/view/spreadsheet.rb
```diff
@@ -168,5 +186,17 @@ def render_player_certs
         *@game.players.map { |p| h(:td, p.num_certs) },
       ])
     end
+
+    def corporation_operated?(corporation, current_round)
+      current_round_s = current_round.join('.')
```
tobymao, line 191: corperation.operating_history.any? { |history| history[0].join('.') == current_round }

assets/app/view/spreadsheet.rb
wheresvic: for the join I would rather do it in the `corporation_operated?` function. Keep the compare logic in 1 place :)

assets/app/view/spreadsheet.rb
wheresvic: Sweet, will refactor! Gotta keep that C++ from crawling out :)

assets/app/view/spreadsheet.rb
tobymao: if you keep it as an array, do you even need to join the history here? can’t you just compare arrays?

wheresvic: My ruby is a little weak but I would like to eject from the function and return true if found. Not sure if the lamda will help me here :(

assets/app/view/spreadsheet.rb
wheresvic: Needed some sort of a marker to differentiate between floated and unfloated corps

assets/app/view/spreadsheet.rb
wheresvic: Ok I tried it as you suggested but it does not return true for the case when found - this is what I tried
```
  def corporation_operated?(corporation, current_round)
      current_round_s = current_round.join('.')
      corporation.operating_history.any? { |history| history[0].join('.') == current_round }
    end
```
also
```
  def corporation_operated?(corporation, current_round)
      current_round_s = current_round.join('.')
      corporation.operating_history.any? { |history| return true if history[0].join('.') == current_round }
    end
```

assets/app/view/spreadsheet.rb
```diff
@@ -96,10 +97,17 @@ def render_title
     end
 
     def render_corporations
-      @game.corporations.map { |c| render_corporation(c) }
+      current_round = [@game.turn]
```
tobymao, line 100: ok let's refactor this a bit

add a new method to game/base.rb

`
def turn_round_num #or better name
   [@turn, @game.round.round_num] # make round/base return round_num 1 by default
end
 `

assets/app/view/spreadsheet.rb
```diff
@@ -96,10 +97,17 @@ def render_title
...
+      current_round << @game.round.round_num if @game.round.name == 'Operating Round'
```
tobymao, line 101: maybe we don't need this

assets/app/view/spreadsheet.rb
tobymao: in corporation.rb#add_operating_info, use that method as well

wheresvic: Hmm, are you sure about it being in base? I ask because `round_num` is undefined for non-operating rounds :p.

assets/app/view/spreadsheet.rb
tobymao: you can replace the whole thing with the lambda

corperation.operating_history.any? { |history| history[0] == current_round }

this will return true if it has a match else false

assets/app/view/spreadsheet.rb
wheresvic: Ok nevermind my comment - I will default it to 1 if it does not exist!

Issue #378, opened 2020-05-13
Display number of ORs (like OR 6.2 (2/3)


PedroRCSequeira:
I know it's something basic, but while playing a few async games at the same time it could help

Not sure about the design, maybe it could be

OR6 1/3
OR6 2/3
OR6 3/3

Herbert62:
This is related to #364 (show game phase (yellow, green, brown))

scottredracecar:
I think the following syntax would be good.
`OR6.1 (1 of 3)`

If there is only one round, then don't use the parenthetical.
`OR2`

sparr:
`OR6.1 (1 of 3)` is a bit redundant. How about `OR6.1 (of 3)` or even without the parentheses?

scottredracecar:
I prefer the syntax I proposed. It is short enough already.

I think this change would appear on the game tab and in the log only currently. If it added to other places in the future, it would be good to show up there too. 

The OR name is also shown in the spreadsheet and we wouldn’t want to change it there of course.

wheresvic:
@dfannius I just had a quick look into this issue and it is actually not that straightforward. Basically the engine contains the `round_num` field but it is not populated into the db. So this issue will likely require adding a column to the `game` table, saving this info and then displaying it, etc. At least for the main page we would need to store it into the db if I understood Toby correctly.

Anyways long story short I would remove the `good first issue` label on this one :p

dfannius:
I refuse to believe that it is really that complicated! (I am probably wrong, of course.) I'll take a look.

tobymao:
it’s not complicated but it requires a db migration

sparr:
It seems like some people here are discussing putting this info on the game card on the front page (hard(ish)), while other people are discussing putting it on the game page (easy?).

scottredracecar:
No, just game page is best. Too much info for the homepage game card.

tobymao:
it’s already on the game page

tobymao:
ah i misunderstood, this is just to add the total rounds

jenf:
resolved by #614

PR #614, opened 2020-06-03
Display number of ORs in set (implements #378)

dfannius:


lib/engine/game/base.rb
```diff
@@ -617,7 +617,7 @@ def stock_round
       end
 
       def new_operating_round(round_num = 1)
-        @log << "-- Operating Round #{@turn}.#{round_num} --"
+        @log << "-- Operating Round #{@turn}.#{round_num} (of #{@operating_rounds}) --"
```
tobymao, line 620: why is this duplicated with log_new_round?

looks like i implemented log_new_round but never call it, can you clean this stuff up and just use the one in the round?

tobymao:
also check this out https://github.com/tobymao/18xx/commit/0a30c5e30afba10560363afeff754a669762fcd0

lib/engine/game/base.rb
dfannius: The easiest thing to do would be to just get rid of `log_new_round`. I spent 5 minutes trying to use `log_new_round` in a nice abstract way, but it's annoying, because by the time you have a `Round` object to call it on, you've already called the `Round` constructor, which does things like have privates pay out, so the round-change message shows up in the log late. We could call `log_new_round` in the constructors of the individual `Round` types, but then we've lost the benefit of writing it just in one place. (We could also separate out construction/initialization from the actions that take place at the start of the round, and have `new_operating_round` and `new_stock_round` call them individually, with a call to `log_new_round` in between, if we really wanted to keep that function around.)

tobymao: why don't you just call log_new_round in the base/round constructor?

dfannius: I've been trained not to call polymorphic functions on an object that hasn't been fully constructed yet. For example if we construct an operating round object and it starts by calling the base round constructor, which then calls `log_new_round`, which is overridden by the operating round, but the operating round method looks at instance variables that haven't been set yet in its own constructor, that would be bad. This can happen very easily in C++; I don't know if maybe it doesn't turn out to be an issue for some reason in Ruby, or if we happen to be able to get away with it in this particular instance. I can take a look.

tobymao: ok, that's fine just delete it from the rounds

dfannius:
Resolved conflicts that had come up in the interim.

PR #622, opened 2020-06-03
add default operating info, closes #618

### LINKED ISSUE #618: Operating order column in spreadsheet tab does not mark newly floated companies as having operated.
The order number should be marked with a `*`.

wheresvic:
This fix adds a default operating information for any company that is operating.

The default is to withhold with a revenue of 0. This allows one to identify if a company has operated its turn or not. This is especially relevant for the spreadsheet view.

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -111,6 +111,10 @@ module Engine
       @operating_history[[turn, round_num]] = info
     end
 
+    def operating_info(turn, round_num)
+      @operating_history[[turn, round_num]]
+    end
+
     def inspect
       "<#{self.class.name}: #{id}>"
     end
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -107,14 +107,6 @@ module Engine
       true
     end
 
-    def add_operating_info!(turn, round_num, info)
-      @operating_history[[turn, round_num]] = info
-    end
-
-    def operating_info(turn, round_num)
-      @operating_history[[turn, round_num]]
-    end
-
     def inspect
       "<#{self.class.name}: #{id}>"
     end
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -235,9 +235,7 @@ module Engine
 
       def _process_action(action)
         entity = action.entity
-        # default operating action is to payout 0, i.e. withhold
-        or_info = OperatingInfo.new([], Action::Dividend.new(@game.current_entity, 'withhold'), 0)
-
+        
         case action
         when Action::LayTile
           hex_id = action.hex.id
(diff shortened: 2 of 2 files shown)


lib/engine/round/operating.rb
```diff
@@ -304,6 +305,8 @@ def _process_action(action)
         when Action::Bankrupt
           liquidate(entity.owner)
         end
+
```
tobymao, line 308: is it possible for the operating info to get overidden to the wrong thing? this looks a bit unsafe

wheresvic: That is exactly the idea that each method would overwrite its own operating info. The default is to withhold with 0 revenue.

wheresvic: I think this is the simplest solution but if you have a better idea on how to do this, I am all ears :)

tobymao: if you rebase master, you'll see a new method called start_operating

in there you can initialize yoru dummy 0 value

then only in dividend, you can add that, much safer

wheresvic: perfect will do!

lib/engine/round/operating.rb
```diff
@@ -308,6 +308,12 @@ def start_operating
       def change_entity(_action)
         return unless @current_entity.passed?
 
+        # default operating action is to payout 0, i.e. withhold
+        unless @current_entity.operating_info(@game.turn, @round_num)
```
wheresvic, line 312: There is now a check to ensure that we do not overwrite any valid operating info :)

lib/engine/corporation.rb
```diff
@@ -111,6 +111,10 @@ def add_operating_info!(turn, round_num, info)
       @operating_history[[turn, round_num]] = info
     end
 
+    def operating_info(turn, round_num)
```
tobymao, line 114: let's delete this and add_operating_info! they are both unneeded

lib/engine/round/operating.rb
```diff
@@ -308,6 +308,12 @@ def start_operating
       def change_entity(_action)
         return unless @current_entity.passed?
 
+        unless @current_entity.operating_info(@game.turn, @round_num)
```
tobymao, line 311: @current_entity.operating_history[[@game.turn, @round_num]] ||= OperatingInfo.new([], Action::Dividend.new(@game.current_entity, 'withhold'), 0)

wheresvic: See above comment, I'm happy to change it if we get rid of the method call.

lib/engine/corporation.rb
tobymao: i think it's a bit confusing because when we both initially looked at it, add_operating_info! we thought that it was adding to an array, just exposing the hash will be more readable to the user

so maybe change add_operating_info to be

def operating_info(turn, round_num)=
...

wheresvic: yep perfect, let's do this

PR #623, opened 2020-06-04
Map axes: allow letters to be rows or columns

michaeljb:
screenshot album - https://imgur.com/a/zG4xN6B

by default, for pointy maps letters are rows and for flat maps letters are
columns; this can be overriden by defining AXES in the game class; for example,
this is what was used to generate the wonky 1889 screenshot:

```
AXES = { x: :number, y: :letter }.freeze
```

* update 1846 config (TODO: contribute back to 18xx-maker)
  - no aliases, this allows the blue hexes to be blue
  - rotate track appropriately

Screenshot from 2020-06-03 19-14-28

### DIFF
--- a/assets/app/view/axis.rb
+++ b/assets/app/view/axis.rb
@@ -7,6 +7,7 @@ module View
     needs :cols
     needs :rows
     needs :layout
+    needs :axes
 
     needs :font_size, default: 25
     needs :text_length, default: 25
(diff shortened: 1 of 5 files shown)


assets/app/view/axis.rb
```diff
@@ -40,9 +41,15 @@ def hex_size
...
 
-      labels = @cols.map do |c|
-        x = hex_x * (c.to_i - 1)
-        label = ('A'..'Z').to_a[c - 1]
+      labels = @cols.map do |col|
+        x = hex_x * (col.to_i - 1)
+
+        label =
+          if @axes[:x] == :letter
+            ('A'..'Z').to_a[col - 1]
```
tobymao, line 49: take this out if the hot loop same with the other (a..z).to_a

PR #633, opened 2020-06-05
Add warning that the game is ending, add what game it is to the status

### LINKED ISSUE #158: Call out end of game
It would be helpful to list why the game ended (broke the bank, etc.)

### LINKED ISSUE #320: Call out when bank has been broken
Moved out as a separate issue from a comment in #262.

### LINKED ISSUE #577: Make more visible what 18xx game the player is playing
Playing async games is confusing if I'm playing 18Chesapeake or 1889, and when the platform possibility more games it will only increase complexity. So I think that should be more obvious what game in the 18xx options I'm currently playing.

Suggestion: in the right of the 18xx.Games icon the text of the game scenario.

jenf:
fixes #158
fixes #320
fixes #577

enables the end game when a stock hits the end of the stock market for 1882 (but it's not coded yet)

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -577,8 +577,7 @@ module Engine
       end
 
       def next_round!
-        if (game_end = game_ending?)
-          _reason, after = game_end
+        if (_reason, after = game_end_reason)
           if after != :full_round || (@round.is_a?(Round::Operating) && @round.round_num == @operating_rounds)
             return end_game!
           end
@@ -605,10 +604,8 @@ module Engine
       end
 
       def game_ending_description
-        game_end = game_ending?
-        return unless game_end
-
-        reason, after = game_end
+        reason, after = game_end_reason
+        return unless after
 
         after_text = ''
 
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -577,19 +577,22 @@ def connect_hexes
       end
 
       def next_round!
+        if (game_end = game_ending?)
```
tobymao, line 580: can you do if (_reason, after = game_ending?)

lib/engine/game/base.rb
```diff
@@ -601,6 +604,38 @@ def next_round!
           end
       end
 
+      def game_ending_description
+        game_end = game_ending?
```
tobymao, line 608: reason, after = game_ending?

lib/engine/game/base.rb
```diff
@@ -601,6 +604,38 @@ def next_round!
...
+
+        reason_map = {
+                       bank: 'Bank Broken',
+                       bankrupt: 'Bankrupcy',
+                       stockmarket: 'Company hit max stock value',
+                     }
+        "#{reason_map[reason]}#{after_text}"
+      end
+
+      def game_ending?
```
tobymao, line 634: ? should really just be boolean let's change this to

def game_end_reason

lib/engine/game/base.rb
jenf: yep, forgot you can do that in ruby :)

PR #635, opened 2020-06-05
User settings: tile colors; (some) inline styles to .css

yzemaze:
As announced earlier: big, fat, messy commit. Just ask away :)


wheresvic:
If you could provide some screenshots of what it looks like that would be awesome!

yzemaze:
Some different colors:
image

Settings (colors = buttons)
image

Reset:
Peek 2020-06-05 20-35

scottredracecar:
Now we're talkin'! Thanks for doing this work

assets/app/view/hex.rb
```diff
@@ -54,7 +55,7 @@ def render
       props = {
         attrs: {
           transform: transform,
-          fill: Lib::Hex::COLOR.fetch(@tile&.color, 'white'),
+          fill: @user&.dig(:settings, @tile&.color) || (Lib::Hex::COLOR[@tile&.color || 'white']),
```
tobymao, line 58: this needs to be handled in borders too, try to find a clean way to do that

assets/vendor/polyfill.js
```diff
@@ -2,9 +2,14 @@ if (typeof setTimeout === 'undefined') { var setTimeout = function setTimeout(fu
 
 if (typeof window === 'undefined') {
   window = {
+    addEventListener: function() {},
+    matchMedia: function() {
```
tobymao, line 6: i think this can just be matchMedia: function()  { return {} },

assets/app/view/hex.rb
yzemaze: "in borders"?

routes/user.rb
```diff
@@ -68,6 +69,13 @@ class Api
           user.settings['bg_color'] = r.params['bg_color']
           user.settings['font_color'] = r.params['font_color']
           user.settings['red_logo'] = r.params['red_logo']
+          user.settings['white'] = r.params['white']
```
tobymao, line 72: let's change all of these...

in model/user.rb add in a const which is like 

SETTINGS = %w[bg_color font_color red_logo white...]

def update_settings(params)
SETTINSGS.each { |setting| settings[setting] = params[setting] }

public/assets/main.css
```diff
@@ -74,34 +108,66 @@ a.game__nav:focus {
   }
 }
 
+#profile-settings {
```
tobymao, line 111: my only other comment is this css, i don't know what it does because i don't know CSS very well.

is it necessary to put this here vs inline with functions that wrap these divs?

assets/vendor/polyfill.js
yzemaze: Fails on Chrome if removed.

tobymao: that doesn't make sense, this should never run in chrome

yzemaze: GoogleChrome the browser. It works with Firefox though.

public/assets/main.css
yzemaze: You’d have to put style-declarations in at least 8 places in the code for this to work. To me that’s way too cumbersome. As is I can view html and css side by side without scrolling which makes for easier layouting. Once that’s done you can of course move everything back to the code. But it most definitely doesn’t help future **front end** devs who’d like to or have to change CSS in the future …

assets/vendor/polyfill.js
yzemaze: Done.

assets/app/view/hex.rb
yzemaze: Define "clean". My changed part of the code should work there, too. Is there already a map with borders?

tobymao: not sure...

assets/app/view/user.rb
```diff
@@ -103,6 +138,7 @@ def submit
       when :login
         login(params)
       when :profile
+        store(:flash_opts, Native(@elms_bg))
```
tobymao, line 141: did you mean to keep this here?

routes/user.rb
```diff
@@ -20,7 +20,19 @@ class Api
...
+              red_logo: r['red_logo'],
+              white: r['white'],
+              yellow: r['yellow'],
+              green: r['green'],
+              brown: r['brown'],
+              gray: r['gray'],
+              red: r['red'],
+              blue: r['blue'],
+            },
           }.reject { |_, v| v.empty? }
```
tobymao, line 36: we should use the same update settings here so this is not hardcoded

PR #636, opened 2020-06-05
Add sortable columns to spreadsheet view, closes #291

### LINKED ISSUE #291: Spreadsheet - sort corps by Market value

wheresvic:
This PR allows sorting columns in the spreadsheet view. At the moment the following 3 columns are supported:
- Operating Order
- Cash
- Market price

The default is to sort by company symbol in ascending order (current view). The sort preferences are locally stored so they should be maintained over sessions :).

See screencast: [https://streamable.com/oobcg6](https://streamable.com/oobcg6)

### DIFF
--- a/assets/app/view/spreadsheet.rb
+++ b/assets/app/view/spreadsheet.rb
@@ -8,8 +8,6 @@ require 'lib/storage'
 module View
   class Spreadsheet < Snabberb::Component
     needs :game
-    needs :spreadsheet_sort_by, default: nil
-    needs :spreadsheet_sort_order, default: nil
 
     def render
       @spreadsheet_sort_by = Lib::Storage['spreadsheet_sort_by']
@@ -91,9 +89,9 @@ module View
           h(:th, props, 'IPO'),
           h(:th, props, 'Market'),
           h(:th, props, 'IPO'),
-          render_sort_link(props, 'Market', 'SHARE-PRICE'),
+          render_sort_link(props, 'Market', 'SHARE_PRICE'),
           render_sort_link(props, 'Cash', 'CASH'),
-          render_sort_link(props, 'Operating Order', 'OPERATING-ORDER'),
+          render_sort_link(props, 'Operating Order', 'OPERATING_ORDER'),
           h(:th, props, 'Trains'),
           h(:th, props, 'Tokens'),
           h(:th, props, 'Privates'),
@@ -138,31 +136,29 @@ module View
     def render_corporations
       current_round = @game.round.turn_round_num
 
-      ordered_corporations = sorted_corporations
-      ordered_corporations.map do |c|
+      sorted_corporations.map do |c|
         render_corporation(c[1], c[0], current_round)
       end
     end
 
     def sorted_corporations
-      result = []
       floated_corporations = @game.round.entities
 
-      @game.corporations.map do |c|
+      result = @game.corporations.map do |c|
         operating_order = (floated_corporations.find_index(c) || -1) + 1
-        result << [operating_order, c]
+        [operating_order, c]
       end
 
-      result = result.sort_by do |c|
+      result.sort_by! do |operating_order, corporation|
         case @spreadsheet_sort_by
-        when 'OPERATING-ORDER'
-          c[0]
+        when 'OPERATING_ORDER'
+          operating_order
         when 'CASH'
-          c[1].cash
-        when 'SHARE-PRICE'
-          c[1].share_price.nil? ? 0 : c[1].share_price.price
+          corporation.cash
+        when 'SHARE_PRICE'
+          corporation.share_price&.price || 0
         else
-          c[1].id
+          corporation.id
         end
       end
 
--- a/assets/app/view/spreadsheet.rb
+++ b/assets/app/view/spreadsheet.rb
@@ -8,10 +8,13 @@ require 'lib/storage'
 module View
   class Spreadsheet < Snabberb::Component
     needs :game
-    needs :spreadsheet_sort_by, default: Lib::Storage['spreadsheet_sort_by'] || 'ID', store: true
-    needs :spreadsheet_sort_order, default: Lib::Storage['spreadsheet_sort_order'] || 'ASC', store: true
+    needs :spreadsheet_sort_by, default: nil
+    needs :spreadsheet_sort_order, default: nil
 
     def render
+      @spreadsheet_sort_by = Lib::Storage['spreadsheet_sort_by']
+      @spreadsheet_sort_order = Lib::Storage['spreadsheet_sort_order']
+
       h(:div, { style: {
         overflow: 'auto',
         margin: '0 -1rem',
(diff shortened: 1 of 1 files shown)


assets/app/view/spreadsheet.rb
```diff
@@ -79,14 +83,58 @@ def render_title
           h(:th, { attrs: { colspan: or_history_titles.size } }, 'OR History'),
           ]),
         h(:tr, [
-          h(:th, { style: { width: '20px' } }, ''),
+          h(:th, { style: { width: '20px' } }, [
```
tobymao, line 86: you should be able to dry this up

`def sort_link(...)`

also lambda should be multiline do end, you don't have to inline it in the new function you create when you dry this up

assets/app/view/spreadsheet.rb
```diff
@@ -2,10 +2,14 @@
 
 require 'view/stock_market'
 require 'lib/color'
+require 'view/link'
+require 'lib/storage'
 
 module View
   class Spreadsheet < Snabberb::Component
     needs :game
+    needs :spreadsheet_sort_by, default: Lib::Storage['spreadsheet_sort_by'] || 'ID', store: true
```
tobymao, line 11: i made this framework and didn't even know you could do this lol.. probably safer to default to nil and then to reset it in the render

tobymao, line 11: you're already using local storage, so there's no reason to store it here,

change these two needs to

needs sort_by default: nil
needs sort_order default: nil

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +144,62 @@ def render_title
...
+    def sort_order_icon
+      return '(⬇️)' if @spreadsheet_sort_order == 'ASC'
+
+      '(⬆️)'
+    end
+
+    def mark_sort_column(sort_by)
+      Lib::Storage['spreadsheet_sort_by'] = sort_by
+      store(:spreadsheet_sort_by, sort_by)
+      *@spreadsheet_sort_by = sort_by
```
tobymao, line 156: store will set the variable and rerender, you should not mutate the variable

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +144,62 @@ def render_title
...
     end
 
+    def sort_order_icon
+      return '(⬇️)' if @spreadsheet_sort_order == 'ASC'
+
+      '(⬆️)'
+    end
+
+    def mark_sort_column(sort_by)
+      Lib::Storage['spreadsheet_sort_by'] = sort_by
```
tobymao, line 154: i don't think you need any needs if you're using local stoarge

so just do

Lib::Storage['spreadsheet_sort_by'] = sort_by
update

do you want this to be user global or per game? i think user global makes ssense

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +144,62 @@ def render_title
...
       floated_corporations = @game.round.entities
+
       @game.corporations.map do |c|
         operating_order = (floated_corporations.find_index(c) || -1) + 1
-        render_corporation(c, operating_order, current_round)
+        result << [operating_order, c]
+      end
+
+      result = result.sort_by do |c|
+        if @spreadsheet_sort_by == 'OPERATING-ORDER'
```
tobymao, line 190: case @sort_by
when 'OPERATING-ORDER'
...

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +144,62 @@ def render_title
...
+        elsif @spreadsheet_sort_by == 'CASH'
+          c[1].cash
+        elsif @spreadsheet_sort_by == 'SHARE-PRICE'
+          c[1].share_price.nil? ? 0 : c[1].share_price.price
+        else
+          c[1].id
+        end
       end
+
+      result = result.reverse if @spreadsheet_sort_order == 'DESC'
```
tobymao, line 201: result.reverse! if ...

assets/app/view/spreadsheet.rb
wheresvic: yes, so it was the `update` function that I was missing! Will get rid of the `store`

wheresvic: agreed.

assets/app/view/spreadsheet.rb
wheresvic: I think global also makes sense. So yeah `update` was the trick which I was missing - will clean it up!

assets/app/view/spreadsheet.rb
wheresvic: yep!

assets/app/view/spreadsheet.rb
wheresvic: cool

assets/app/view/spreadsheet.rb
wheresvic: done!

assets/app/view/spreadsheet.rb
wheresvic: done

assets/app/view/spreadsheet.rb
```diff
@@ -2,12 +2,19 @@
 
 require 'view/stock_market'
 require 'lib/color'
+require 'view/link'
+require 'lib/storage'
 
 module View
   class Spreadsheet < Snabberb::Component
     needs :game
+    needs :spreadsheet_sort_by, default: nil
```
tobymao, line 11: not needed, delete

assets/app/view/spreadsheet.rb
```diff
@@ -79,14 +86,14 @@ def render_title
...
-          h(:th, { style: { width: '20px' } }, ''),
+          render_sort_link({ style: { width: '20px' } }, 'SYM', 'ID'),
           *@game.players.map { |p| h(:th, props, p.name) },
           h(:th, props, 'IPO'),
           h(:th, props, 'Market'),
           h(:th, props, 'IPO'),
-          h(:th, props, 'Market'),
-          h(:th, props, 'Cash'),
-          h(:th, props, 'Operating Order'),
+          render_sort_link(props, 'Market', 'SHARE-PRICE'),
```
tobymao, line 94: let's call it SHARE_PRICE and OPERATING_ORDER

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +103,71 @@ def render_title
...
+      Lib::Storage['spreadsheet_sort_order'] = 'ASC' if @spreadsheet_sort_order == 'DESC'
+      Lib::Storage['spreadsheet_sort_order'] = 'DESC' unless @spreadsheet_sort_order == 'DESC'
+      update
+    end
+
     def render_corporations
       current_round = @game.round.turn_round_num
 
+      ordered_corporations = sorted_corporations
+      ordered_corporations.map do |c|
```
tobymao, line 142: sorted_corporations.map

don't need the extra var ordered_corporations

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +103,71 @@ def render_title
...
+        render_corporation(c[1], c[0], current_round)
+      end
+    end
+
+    def sorted_corporations
+      result = []
       floated_corporations = @game.round.entities
+
       @game.corporations.map do |c|
```
tobymao, line 151: if you're doing map, there's no reason to have an array

result = @game.corporations.map do |c|
[operating_order, c]
end

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +103,71 @@ def render_title
...
         operating_order = (floated_corporations.find_index(c) || -1) + 1
-        render_corporation(c, operating_order, current_round)
+        result << [operating_order, c]
       end
+
+      result = result.sort_by do |c|
```
tobymao, line 156: you can just do result.sort_by! without assignng a new var

tobymao, line 156: result.sort_by! do |order, corporation|

assets/app/view/spreadsheet.rb
```diff
@@ -96,14 +103,71 @@ def render_title
...
+        case @spreadsheet_sort_by
+        when 'OPERATING-ORDER'
+          c[0]
+        when 'CASH'
+          c[1].cash
+        when 'SHARE-PRICE'
+          c[1].share_price.nil? ? 0 : c[1].share_price.price
```
tobymao, line 163: corporation.share_price&.price || 0

assets/app/view/spreadsheet.rb
wheresvic: whoops - will remove!

assets/app/view/spreadsheet.rb
wheresvic: sure

assets/app/view/spreadsheet.rb
wheresvic: got it

PR #638, opened 2020-06-05
switch delete text to wastebasket emoji to free up ui, closes #603

### LINKED ISSUE #603: Consider a trash symbol / emoji instead of `delete` text for the button
It would save some space and liberate the UI a little bit!

wheresvic:
See screenshot

image


tobymao:
next time you see an unused method delete it

PR #646, opened 2020-06-07
special round: filter laid hexes from connected_hexes

### LINKED ISSUE #613: 1889 Mitsubishi Ferry Port tile should not upgrade tile 58
Current implementation allows placement of the Port tile as an upgrade to #58.

michaeljb:
I tried adding a check in `Engine::Round::Base#lay_tile` for
`old_tile.upgrades_to?(tile)`, but that breaks in the case of 1889's port town
-> tile 437. Will need to add a "port" part to the appropriate towns and tile
437 for that to work correctly.

This now prevents the UI from allowing a private company to lay a tile on a hex
that already has a tile laid on it, but if that action sneaks in, it is not
caught, meaning past games that did this illegally should survive for now.

[Fixes #613]

### DIFF
--- a/lib/engine/round/base.rb
+++ b/lib/engine/round/base.rb
@@ -200,6 +200,11 @@ module Engine
         rotation = action.rotation
         old_tile = hex.tile
 
+        # at time of writing, this check is invalid for tile 437 in 1889; we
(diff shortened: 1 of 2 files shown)


michaeljb:
Do #647 instead for a more proper fix

PR #648, opened 2020-06-07
forms: placeholder, label_props, for/id

yzemaze:
+ placeholder:
+ label_props
+ for on label
+ id on input

for + id connect label and input => click on label results in focus on input
dedicated placeholder to separate label/placeholder
render_input() is used in reset.rb, create_game.rb, forgot.rb and user.rb. I haven’t added placeholders there.

Without iOS device hard to test, but missing ids might have been the reason color inputs displayed standard color (black) on iOS. So this commit might fix this issue. Reports in regards to my [test-page](https://yzemaze.de/temp/18xx_profile_test.html) were positive.

### DIFF
--- a/assets/app/view/form.rb
+++ b/assets/app/view/form.rb
@@ -45,24 +45,33 @@ module View
     end
 
     # rubocop:disable Layout/LineLength
-    def render_input(label, id:, el: 'input', type: 'text', attrs: {}, container_style: {}, input_style: {}, children: [])
+    def render_input(label, placeholder: '', id:, el: 'input', type: 'text', attrs: {}, container_style: {}, label_style: {}, input_style: {}, children: [])
       # rubocop:enable Layout/LineLength
-      props = {
+      label_props = {
+        style: {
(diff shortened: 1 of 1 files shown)


assets/app/view/form.rb
```diff
@@ -45,24 +45,33 @@ def render_form(name, inputs)
     end
 
     # rubocop:disable Layout/LineLength
-    def render_input(label, id:, el: 'input', type: 'text', attrs: {}, container_style: {}, input_style: {}, children: [])
+    def render_input(label, placeholder: '', id:, el: 'input', type: 'text', attrs: {}, container_style: {}, label_style: {}, input_style: {}, children: [])
       # rubocop:enable Layout/LineLength
-      props = {
+      label_props = {
+        style: {
```
tobymao, line 51: you can just do style: label_style

yzemaze: Just copied LL51+ (now 59) ;)

Issue #649, opened 2020-06-07
List par prices in ascending order


ChrisTheCat:
It would be nice if the par values were displayed 70, 80, 95 instead of 95, 80, 75. Most people count numbers up, not down. I have accidentally chosen the wrong par price several times because of this.

<img width="350" alt="Screenshot 2020-06-07 08 32 32" src="https://user-images.githubusercontent.com/8732435/83973200-a92d4380-a899-11ea-8848-d12192db3100.png">

scottredracecar:
I'll take this

scottredracecar:
This is solved now.

PR #653, opened 2020-06-07
reverse par prices, add setup method to base

scottredracecar:


scottredracecar:
closes https://github.com/tobymao/18xx/issues/649

Issue #155, opened 2020-04-26
Clicking on someone's username while queueing for a game kicks them out of the game


jftaylor21:
I started a new game and noticed that someone joined it. I saw a box around the username and was curious about what happened when I clicked on it. It looks like it kicked them out of my game which was unintentional.

It would be nice if there was an X on the username box or some other UI element that indicates that this will kick the person.

daniel-sousa-me:
I made a PR to try to improve this. This is my first (of hopefully many) contributions. I hope everything is ok.

PR #659, opened 2020-06-08
Improved the interface of kicking users from a game (includes a extra step of confirmation)

daniel-sousa-me:
I want to start contributing to the code, so I searched a "good first issue".

This is my attempt at issue #155

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -10,6 +10,7 @@ module View
     needs :user
     needs :gdata # can't conflict with game_data
     needs :confirm_delete, store: true, default: false
+    needs :confirm_kick, store: true, default: false
 
     ENTER_GREEN = '#3CB371'
     JOIN_YELLOW = '#F0E58C'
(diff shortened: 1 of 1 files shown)


assets/app/view/game_card.rb
```diff
@@ -10,6 +10,7 @@ class GameCard < Snabberb::Component
     needs :user
     needs :gdata # can't conflict with game_data
     needs :confirm_delete, store: true, default: false
+    needs :confirm_kick, store: true, default: false
```
tobymao, line 13: let's make the types consistent, default should be nil

assets/app/view/game_card.rb
```diff
@@ -155,23 +156,52 @@ def render_body
...
+          if @confirm_kick != id
+            button_props = {
+              on: { click: -> { store(:confirm_kick, id) } },
+              style: {
+                color: 'red',
+                cursor: 'pointer',
+              },
+            }
+
+            elms << h('a', button_props, '❌')
```
tobymao, line 184: i think this could be dryed up a bit

maybe put button_props out of the block and do

button_props = { 
  style: { cursor: 'pointer' },
  on: {},
}

assets/app/view/game_card.rb
daniel-sousa-me: Good point

Should I also change it for :confirm_delete?

jenf: Yes.

PR #663, opened 2020-06-08
1882 Add neutral tokens to corps

jenf:
Add neutral token for 1882 and display it on the corporation

### DIFF
--- a/lib/engine/corporation_token.rb
+++ b/lib/engine/corporation_token.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative 'token'
+
+module Engine
+  class CorporationToken < Token
+    attr_reader :corporation, :price
+
+    def initialize(corporation, price: 0)
+      super()
+      @corporation = corporation
+      @price = price
+    end
+
+    def logo
+      @corporation.logo
+    end
+  end
+end
--- a/lib/engine/game/g_1882.rb
+++ b/lib/engine/game/g_1882.rb
@@ -2,6 +2,7 @@
 
 require_relative '../config/game/g_1882'
 require_relative 'base'
+require_relative '../neutral_token'
 
 module Engine
   module Game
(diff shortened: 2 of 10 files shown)


lib/engine/corporation_token.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative 'token'
+
+module Engine
+  class CorporationToken < Token
```
tobymao, line 6: keep corporation as token, we should mix 'markers' and tokens. 

let's have another unrelated class called a marker. 

for neutral tokens, if they act as a normal token, they should just be a token with a neutral corporation.

lib/engine/neutral_token.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require_relative 'token'
+
+module Engine
+  class NeutralToken < Token
```
tobymao, line 6: let's delete this

lib/engine/corporation_token.rb
jenf: they don't block routes but leave them permanently open. I was going to do something like blocks_routes? in the token itself, but I can probably implement that as a check against the 'neutral' corporation.

Neutral corp the best way to go?

tobymao: when a neutral corp is indtantiated it should be passed a no block property which creates non blocking tokens . just a thought

jenf: Yep, will mod as suggested :)

jenf: Just as a general discussion. My next step is to actually be able to lay the tokens, I was wanting to use the tile_selector code as a token_selector where the player has a choice of tokens.

This approach okay? It might be useful for other games as well

tobymao: that code could use a refactor, especially the placement and look of the choose and cancel,

it might be appropriate to use that for laying tokens? i'm not sure let's discuss on slack later

lib/engine/game/g_1882.rb
```diff
@@ -13,11 +13,24 @@ class G1882 < Base
...
       load_from_json(Config::Game::G1882::JSON)
 
       def stock_round
         Round::Stock.new(@players, game: self, sell_buy_order: :sell_buy_sell)
       end
+
+      def init_corporations(stock_market)
+        # Neutral corp that allows tokens that don't block other players
+        # CN runs using these tokens
+        @neutral_corp = Corporation.new(sym: 'neutral', name: 'neutral', tokens: [], logo: '1882/neutral')
```
tobymao, line 27: do you need to store @neutral corp? why use an instance var
sym :n

lib/engine/marker.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Engine
```
tobymao, line 3: do we need this now? nothing else is using it. are you planning to use it soon?

jenf: i wasn't, i assumed you'd wanted it for 46 :)

PR #682, opened 2020-06-11
WIP: Attempt to keep trains for IR's out of market

gchambers419:
First attempt to keep the trains for Big 4 and Michigan Southern out of the market.

Not yet tested; will do so after work ~7EDT

### DIFF
--- a/lib/engine/config/game/g_1846.rb
+++ b/lib/engine/config/game/g_1846.rb
@@ -333,7 +333,7 @@ module Engine
         "6",
         "7/8"
       ],
-      "num": 1
+      "num": 0
     },
     {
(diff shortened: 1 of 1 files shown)


jenf:
Generally, I don't think we're changing the config file since that comes from 18xx maker.

We should implement a way of disabling trains going into the discard (also needed for 1882), and when those privates get the trains marking them as such.

gchambers419:
If this works, the next step is to work out patches for 18xx-maker and export-rb, but I see your point.

jenf:
Fair enough :)

tobymao:
we won’t be doing it like this, it will be a code change

PR #685, opened 2020-06-12
Add ability to lay neutral tokens

jenf:
This covers the following on the 1882 todo list

- CN -  Placing these are always free and have no association to the company that laid it once it is laid. Placing is instead of a normal token.
- CN -  CN does not have board tokens, it runs neutral tokens.
- CN may not be parred until there is a neutral token on the board. If there are no neutral token on the board when it operates (such as by SC starting there and discarding it, it operates as a suitcase until someone plays a neutral and it can run from there).
- CN - Neutral tokens do not block train runs
- Track laying is permissive.

Corp sheets before laying
Screenshot from 2020-06-11 22-30-18

User clicks on city spot
Screenshot from 2020-06-11 22-30-31

Token is laid
Screenshot from 2020-06-11 22-30-51

Corp sheets after laying
Screenshot from 2020-06-11 22-30-41


assets/app/lib/radial_selector.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'view/game/hex'
```
tobymao, line 3: it doesn't seem like this import is needed

assets/app/view/game/map.rb
```diff
@@ -47,7 +48,12 @@ def render
           left = (@tile_selector.x + map_x) * SCALE
 
           selector =
-            if @tile_selector.hex.tile != @tile_selector.tile
+            if @tile_selector.is_a? Lib::TokenSelector
```
tobymao, line 51: parens

tobymao: can we add a cop for this?

assets/app/view/game/map.rb
```diff
@@ -47,7 +48,12 @@ def render
...
+              # 1882
+              h(TokenSelector,
```
tobymao, line 53: you don't need to pass in game because it's store true, maybe we should change to just pass in the selector

h(TokenSelector, tile_selector: @tile_snelector)

assets/app/view/game/part/city_slot.rb
```diff
@@ -45,13 +46,20 @@ def on_click
           return if @token
           return unless @game.round.can_place_token?
 
-          action = Engine::Action::PlaceToken.new(
-            @game.current_entity,
-            @city,
-            @slot_index
-          )
+          # If there's a choice of tokens from different corps show the selector, otherwise just place
+          token_types = @game.current_entity.tokens.reject(&:used?).group_by(&:corporation)
```
tobymao, line 50: maybe a better name is corporation_tokens

assets/app/view/game/token_selector.rb
```diff
@@ -0,0 +1,59 @@
...
+      needs :slot_index
+      TOKEN_SIZE = 40
+      SIZE = TOKEN_SIZE / 2
+      DISTANCE = TOKEN_SIZE
+
+      def render
+        tokens = list_coordinates(@tokens.keys, DISTANCE, SIZE).map do |corporation, left, bottom|
+          layable = @tokens[corporation].first
+          style = {
+            position: 'absolute',
```
tobymao, line 23: looks like this css is shared either create a class with the shared CSS in main.css or add a helper function to radial selector

assets/app/view/game/token_selector.rb
```diff
@@ -0,0 +1,59 @@
...
+      needs :tokens
+      needs :city
+      needs :slot_index
+      TOKEN_SIZE = 40
+      SIZE = TOKEN_SIZE / 2
+      DISTANCE = TOKEN_SIZE
+
+      def render
+        tokens = list_coordinates(@tokens.keys, DISTANCE, SIZE).map do |corporation, left, bottom|
+          layable = @tokens[corporation].first
```
tobymao, line 21: i think maybe we shoudn't even pass @tokens around, you can get this from the corporation

maybe let's add helper methods in corporations to handle this? just a thought

lib/engine/action/place_token.rb
```diff
@@ -5,23 +5,26 @@
 module Engine
   module Action
     class PlaceToken < Base
-      attr_reader :city, :slot
+      attr_reader :city, :slot, :tokenidx
```
tobymao, line 8: attr_reader should be token

lib/engine/action/place_token.rb
```diff
@@ -5,23 +5,26 @@
...
 
-      def initialize(entity, city, slot)
+      def initialize(entity, city, slot, tokenidx = nil)
```
tobymao, line 10: token_id

lib/engine/action/place_token.rb
```diff
@@ -5,23 +5,26 @@
...
         @entity = entity
         @city = city
         @slot = slot
+        @tokenidx = tokenidx
```
tobymao, line 14: find the token here from the token id

lib/engine/action/place_token.rb
```diff
@@ -5,23 +5,26 @@
...
       end
 
       def self.h_to_args(h, game)
-        [game.city_by_id(h['city']), h['slot']]
+        [game.city_by_id(h['city']), h['slot'], h['tokenidx']]
       end
 
       def args_to_h
-        {
+        h = {
```
tobymao, line 22: you can just add token_id in here, it will be cleared automatically if nil (please test that is true)

lib/engine/corporation.rb
```diff
@@ -43,10 +44,10 @@ def initialize(sym:, name:, tokens:, **opts)
       @float_percent = opts[:float_percent] || 60
       @coordinates = opts[:coordinates]
       @min_price = opts[:min_price]
-      @logo = "/logos/#{opts[:logo]}.svg"
       @color = opts[:color]
       @text_color = opts[:text_color] || '#ffffff'
       @always_market_price = opts[:always_market_price] || false
+      @needs_token_to_par = opts[:needs_token_to_par] || false
```
tobymao, line 50: this is super specific i don't know if i would add it here, probably an override in the game class instead

maybe this is correct, i'm not sure, we need to think about this, but i've never heard of this mechanic before except here

lib/engine/game/base.rb
```diff
@@ -90,6 +90,9 @@ class Base
       EBUY_PRES_SWAP = true # allow presidential swaps of other corps when ebuying
       EBUY_OTHER_VALUE = true # allow ebuying other corp trains for up to face
 
+      # 1882: When laying a token that is from another corp, the other corp takes ownership
+      LAY_TOKEN_OTHER_CORP_TAKES_OWNERSHIP = false
```
tobymao, line 94: again this is way too specific, we should move this ONLY to the 1882 class

lib/engine/round/operating.rb
```diff
@@ -506,14 +506,23 @@ def place_token(action)
...
+
+        price = token&.price || 0
+        action.city.place_token(entity, token, free: @teleported)
         if price.positive? && !@teleported
           entity.spend(price, @bank)
           price_log = " for #{@game.format_currency(price)}"
         end
-        @log << "#{entity.name} places a token on #{action.city.hex.name}#{price_log}"
 
+        if @game.class::LAY_TOKEN_OTHER_CORP_TAKES_OWNERSHIP && token.corporation != entity
```
tobymao, line 519: ok we need an 1882 operating class

lib/engine/round/stock.rb
```diff
@@ -173,6 +173,12 @@ def action_finalized(_action)
         end
       end
 
+      def par_prices(corporation)
```
tobymao, line 176: as discussed we may remove this not sure yet

spec/lib/engine/hex_spec.rb
```diff
@@ -200,7 +200,9 @@ module Engine
               # add initial corporation tokens
               spec[:setup][:corporations].each do |corp|
                 corporation = game.corporation_by_id(corp[:name])
-                initial_tile.cities[corp[:token]].place_token(corporation) unless corp[:token].nil?
+                unless corp[:token].nil?
```
tobymao, line 203: prefer to use if corp[:token]

lib/engine/round/stock.rb
jenf: I think there's quite a few other games that vary par prices by corporation type and by game phase, so i'd suggest it'll be needed eventually.

lib/engine/game/base.rb
jenf: Yep, I wasn't sure the best way to express it :)

lib/engine/round/stock.rb
tobymao: it may make sense then for the corporation to have it's own par prices, i already have a hack called min_value that could benefit from this. i'd suggest moving that info there

assets/app/view/game/map.rb
jenf: i'm not sure, i spent a little trying to find a cop to no success.

lib/engine/corporation.rb
jenf: Technically I don't think any corp can float without tokens, I could just specialization and make it an always check.
Just this is the only one where you can have corps without tokens afaik.

assets/app/view/game/token_selector.rb
jenf: it's dynamic, so helper function

lib/engine/action/place_token.rb
jenf: yes it is.

lib/engine/round/stock.rb
jenf: I'll fix this one tomorrow and the other remaining issue

lib/engine/corporation.rb
scottredracecar: I can't think of any other "needs token to par" games.

jenf: There are quite a few where you decide how many tokens you want... I'll try and specialize the class then for 1882

lib/engine/action/place_token.rb
```diff
@@ -5,22 +5,26 @@
...
-      attr_reader :city, :slot
+      attr_reader :city, :slot, :token
 
-      def initialize(entity, city, slot)
+      def initialize(entity, city, slot, token_id = nil)
         @entity = entity
         @city = city
         @slot = slot
+        # After processing the action the token may no longer be in the entity
+        @token_id = token_id
```
tobymao, line 15: no reason to store this, we should ensure that tokens maitain the id, add id to token and init with index,

lib/engine/action/place_token.rb
```diff
@@ -5,22 +5,26 @@
...
+        @token = @entity.tokens[token_id] unless token_id.nil?
```
tobymao, line 16: if token_id

but as you said this may not exist, so do we need to keep track in the game like the other objects?

token_by_id

lib/engine/corporation.rb
tobymao: i thought we were going to delete this?

lib/engine/corporation.rb
```diff
@@ -63,6 +64,12 @@ def counts_for_limit
       @share_price ? @share_price.counts_for_limit : true
     end
 
+    def can_par?
```
tobymao, line 67: should we just make this @tokens.any?

lib/engine/game/base.rb
```diff
@@ -164,7 +164,10 @@ def self.load_from_json(json)
           corporation.transform_keys!(&:to_sym)
 
           corporation[:color] = const_get(:COLORS)[corporation[:color]] if const_defined?(:COLORS)
-
+          if const_defined?(:CORPORATIONS_OVERRIDE)
```
tobymao, line 167: i don't think we need this

lib/engine/round/g_1882/operating.rb
```diff
@@ -0,0 +1,24 @@
...
+module Engine
+  module Round
+    module G1882
+      class Operating < Operating
+        def place_token(action)
+          entity = action.entity
+          token = action.token || entity.next_token
+          super(action)
+
+          return unless token.corporation != entity
```
tobymao, line 14: return if token.corporation == entity

lib/engine/round/g_1882/operating.rb
```diff
@@ -0,0 +1,24 @@
...
+
+          entity.tokens.delete(token)
+          token.corporation.tokens << token
+          @log.pop
```
tobymao, line 18: eek

lib/engine/game/base.rb
tobymao: delete this section

lib/engine/corporation.rb
jenf: Sorry i hadn't gotten around to it last night

lib/engine/action/place_token.rb
jenf: I was thinking about this, I'd like to add a token type (:normal and :neutral), and then instead of passing in the index, pass in the type to lay. I think it'd make it cleaner.
I'd also then group tokens by type rather than corp.

If I did it this way could remove the need for 1882 operating class by making that code specific to type :neutral in the normal operating class.

Is this a sensible approach?

jenf: Also the neutral token not blocking would be easier to code

tobymao: we can give that a shot

lib/engine/corporation.rb
jenf: As discussed on slack, this is the best available approached

lib/engine/action/place_token.rb
jenf: seems to work fine

Issue #666, opened 2020-06-08
UI color controls -- more granularity wanted


TheExcalabur:
The UI/map colour pane is a nice addition.  

Unfortunately, the current UI changes only affect the main background and text colours, which makes the light grey textboxes with black text quite bright against the UI.  

Feature request: add two more boxes for the 'secondary' background & text colour to be inherited by the various currently-grey boxes: The front page chat, the company banners, the game chat, and any other place currently using a medium-grey background.  

___
It would also be nice if the various map colour boxes were labelled with what they represent--it's not obvious which is the background/empty hex/etc. colour without paging back and forth into a game.  Perhaps a dummy map could be added to the screen?

yzemaze:
> add two more boxes for the 'secondary' background & text colour to be inherited by the various currently-grey boxes: The front page chat, the company banners, the game chat, and any other place currently using a medium-grey background.

… and buttons. Already on my todo list.

> It would also be nice if the various map colour boxes were labelled with what they represent--it's not obvious which is the background/empty hex/etc. colour without paging back and forth into a game. Perhaps a dummy map could be added to the screen?

The info is given in the hover text, albeit white instead of plain and red instead of offboard isn’t ideal. I’ll change this.
A dummy-map right in the settings is still on my todo list – with low priority.

Issue #583, opened 2020-05-30
Remove "Sym" label from corporation charters


ChrisTheCat:
The symbols are self evident and don't need labeling. Removing "Sym" would also allow you to render the symbol in a larger font if desired. At any rate, no need for clutter on the charter.

tobymao:
i don’t think they are in 1889

tobymao:
oh you mean the label sym

ChrisTheCat:
<img width="350" alt="Screenshot 2020-05-30 10 31 34" src="https://user-images.githubusercontent.com/8732435/83335308-bfffe480-a260-11ea-9d16-bd6cb4382322.png">

Not sure I understand your comment @tobymao. What I'm suggesting is that instead of:

```
TR      3,4      ¥283
Sym  Trains   Cash
```

```
TR      3,4      ¥283
         Trains   Cash
```

And the TR font could be larger and vertically centered.

ChrisTheCat:
The "Cash" label might also be unnecessary - the currency symbol is sufficient.

michaeljb:
Been thinking for a while it might be nice to just move the sym into the top of the card, e.g., "Tosa Electric Railway (TR)"

gchambers419:
I'd find it helpful to put a copy of the token symbol in there too.  It's hard to make out the ones already there when they're all greyed out because they've been placed

dfannius:
> Been thinking for a while it might be nice to just move the sym into the top of the card, e.g., "Tosa Electric Railway (TR)"

I tried this locally a while ago and ran into problems with "Takamatsu & Kotohira Electric Railway (KO)" being too long.

yzemaze:
> I tried this locally a while ago and ran into problems with "Takamatsu & Kotohira Electric Railway (KO)" being too long.
It isn’t with just the Sym, but with a token it runs a bit long. That said an ellipsis (…) for veeery long names is easy to do. So how about this?
![image](https://user-images.githubusercontent.com/33390595/83918679-83108200-a779-11ea-868e-483fa5d1914e.png)

dfannius:
That's basically the single place that the full name of a corporation is presented. I don't think we should cut it off if at all possible.

gchambers419:
Put the symbol and token in the gray area?

yzemaze:
![image](https://user-images.githubusercontent.com/33390595/83919071-2f526880-a77a-11ea-8b96-d1e7cdf9a900.png)

[...]

PR #687, opened 2020-06-12
Corporation & player cards revised

yzemaze:
Corporation cards
+ CSS-grid
+ herald in title (rounded square)
+ big symbol (#583)
+ better alignment of holdings
+ contrast of used tokens 15% => 50%
+ classes/ids where necessary/useful
+ table/tr/td alignments
+ thead & tbody (semantics! accessability/screen readers)
- render_president removed, because not used anymore (code was already incorporated/moved inline afaik)
- render_treasury dito

CSS
+ default styles for .card, .nowrap, .title, text-alignment and font-style/weight
+ default styles for tables, ths and tds within .card.

Player cards
+ synergies with corp card: a few props/styles removed, classes added
+ logo positioning in shares table changed
- table-wrapping divs removed


assets/app/view/game/corporation.rb
```diff
@@ -14,20 +14,12 @@ def render
         end
 
         card_style = {
-          display: 'inline-block',
+          border: '1px solid gainsboro',
```
tobymao, line 17: should gainsboro and line 22 be current color as opposed to hardcoded to gainsboro and black?

assets/app/view/game/corporation.rb
```diff
@@ -45,53 +37,46 @@ def render
...
-            'font-weight': 'bold',
-            'margin-left': 'auto',
-            'margin-right': 'auto',
-          },
-        }
-
         if @corporation.owner
           subchildren = (@corporation.operating_history.empty? ? [] : [render_revenue_history])
-          children << h(:table, revenue_table_props, [h(:tr, subchildren)])
+          children << h('table.revenue', [h(:tr, subchildren)])
```
tobymao, line 42: remove all of the CSS inheritance, make a method that does inline styles for the table components that you want

assets/app/view/game/corporation.rb
```diff
@@ -45,53 +37,46 @@ def render
...
         end
 
-        h(:div, { style: card_style, on: { click: onclick } }, children)
+        h('div.corp.card', { style: card_style, on: { click: onclick } }, children)
```
tobymao, line 45: there's no reason to have a corp class

assets/app/view/game/corporation.rb
yzemaze: For now it’s ok. I’ll revise this later on to fix issue #666 
currentColor only works for font color and would be wrong here. There’s no equivalent for bg – sadly.

assets/app/view/game/corporation.rb
```diff
@@ -45,53 +37,46 @@ def render
...
-          background: @corporation.color,
-          color: @corporation.text_color,
-          'text-align': 'center',
-          padding: '0.5rem 0px',
-          'font-weight': 'bold',
-          margin: '-0.5rem -0.5rem 0 -0.5rem',
+        title_row_props = {
+          style: {
+            'background-color': @corporation.color,
+            color: @corporation.text_color,
```
tobymao, line 52: move back to inline style, i don't see much benefit to having a specific styling like token outside of here and in main.css

yzemaze:
+ default style for a set, specific styles removed
+ some styles moved inline
+ main.css reorderd (a bit)

public/assets/main.css
```diff
@@ -209,3 +196,38 @@ input#chatbar {
...
+  line-height: 1.25rem;
+}
+.card th {
+  padding: 0 0.3rem;
+}
+.card td {
+  padding: 0 0.5rem;
+}
+
+.corp div {
```
tobymao, line 225: this seems leftover, should we put it back?

public/assets/main.css
```diff
@@ -209,3 +196,38 @@ input#chatbar {
...
+  display: grid;
+  align-self: center;
+  justify-self: center;
+}
+table.shareholders, table.companies, table.revenue {
```
tobymao, line 230: should we have these?

public/assets/main.css
yzemaze: .corp div { } is necessary to prevent using display: grid and alignment on every prop in corporation.rb

tobymao: ok

public/assets/main.css
yzemaze: it saves three props

tobymao: not worth it :D

yzemaze: renamed class to center, moved collapse to prop

PR #694, opened 2020-06-13
multiple jumps and carolinas update

scottredracecar:


lib/engine/game/g_18_carolinas.rb
```diff
@@ -7,6 +7,37 @@ module Engine
...
+      def init_companies(players)
+        super.take(players.size)
+      end
+
+      def setup
+        @companies.each_with_index do |company, index|
+          price = company.value
+          player = @players[index]
+          company.owner = player
+          @players[index].companies << company
```
tobymao, line 20: this these two lines to player instead of @players[index]

lib/engine/game/g_18_carolinas.rb
```diff
@@ -7,6 +7,37 @@ module Engine
...
+          @players[index].spend(price, @bank)
+          @log << "#{player.name} buys #{company.name} for #{format_currency(price)}"
+        end
+
+        @corporations[0..3].sort_by! { @game.rand }
+        @corporations[4..7].sort_by! { @game.rand }
+
+        @corporations.each_with_index do |corporation, index|
+          index -= 4 if index > 3
```
tobymao, line 29: what are you trying to do here? modulo? if so then it should be like index % 4

lib/engine/game/g_18_carolinas.rb
```diff
@@ -7,6 +7,37 @@ module Engine
...
+          @stock_market.set_par(corporation, @stock_market.par_prices[3 - index])
```
tobymao, line 30: really? variable stock price start???

lib/engine/game/g_18_carolinas.rb
```diff
@@ -7,6 +7,37 @@ module Engine
...
+
+          @log << "#{corporation.name} is set to par #{@stock_market.par_prices[3 - index].price}"
```
tobymao, line 32: reuse the 3 - index here

lib/engine/game/g_18_carolinas.rb
```diff
@@ -7,6 +7,37 @@ module Engine
...
+          corporation.tokens.pop(1) if index  == 1 || index == 2
```
tobymao, line 33: case index
when 1, 2
corporations.tokens.pop(1)
when 3
...
end

lib/engine/game/g_18_carolinas.rb
```diff
@@ -7,6 +7,37 @@ module Engine
...
+          corporation.tokens.pop(2) if index  == 3
+        end
+
+        @corporations.sort_by! { |corporation| -corporation.par_price.price }
+
+        next_round!
```
tobymao, line 39: you shouldn't call this, you should just not have an auction round

lib/engine/round/operating.rb
```diff
@@ -409,7 +409,9 @@ def payout(revenue)
           payout_entity(player, per_share)
         end
         payout_entity(@share_pool, per_share, @current_entity)
-        change_share_price(:right)
+
+        jumps = [(revenue / current_entity.share_price.price).floor, @game.class::MAX_JUMPS]
```
tobymao, line 413: instead of this we should have a function that calculates the num jumps

lib/engine/round/operating.rb
```diff
@@ -425,14 +427,14 @@ def payout_entity(holder, per_share, receiver = nil)
         @bank.spend(amount, receiver)
       end
 
-      def change_share_price(direction)
+      def change_share_price(direction, jumps)
```
tobymao, line 430: do jumps = nil so there's a default

lib/engine/stock_market.rb
```diff
@@ -23,14 +23,15 @@ def set_par(corporation, share_price)
       corporation.par_price = share_price
     end
 
-    def move_right(corporation)
+    def move_right(corporation, jumps)
```
tobymao, line 26: jumps = nil

lib/engine/stock_market.rb
```diff
@@ -23,14 +23,15 @@ def set_par(corporation, share_price)
...
       r, c = corporation.share_price.coordinates
-
-      if c + 1 < @market[r].size
-        c += 1
-        move(corporation, r, c)
-      else
-        move_up(corporation)
+      jumps.times do
```
tobymao, line 28: jumps ||= 1

lib/engine/game/g_18_carolinas.rb
scottredracecar: fixed

lib/engine/game/g_18_carolinas.rb
scottredracecar: ? where?

lib/engine/game/g_18_carolinas.rb
scottredracecar: yes!

lib/engine/game/g_18_carolinas.rb
scottredracecar: This code sets the par prices randomly to start in each of two groups (NC corps and SC corps).
First we randomized positions 0 to 3 (NC), then 4 to 7 (SC).
The index is used to set the par prices for each group 90/80/70/60. Do you think I should do it in a different way?

lib/engine/game/g_18_carolinas.rb
scottredracecar: Fixed, although I don't understand why this works. I thought player was a temporary variable and would not spend the money of give the companies. How do these changes get back into @players?

lib/engine/game/g_18_carolinas.rb
scottredracecar: fixed

lib/engine/stock_market.rb
scottredracecar: Changed this to       (jumps ||= 1).times do
I don't know if this is correct

lib/engine/stock_market.rb
scottredracecar: fixed

lib/engine/round/operating.rb
scottredracecar: fixed

lib/engine/round/operating.rb
scottredracecar: added although I am not sure if I did it right, probably not!

lib/engine/stock_market.rb
scottredracecar: I couldn't figure out what to do with this so I changed it back to jumps.times.do

PR #713, opened 2020-06-15
game cards revised

yzemaze:
+ player names truncated (max. 19 char), full name on hover
+ Max Players / Created inline
+ Updated right aligned
+ time when today, else date
+ fulldate on hover
+ kick buttons hover text
+ commas separated from name
+ no commas between buttons
+ margins/paddings optimized
+ b => strong, u => em

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -23,8 +23,7 @@ module View
           border: 'solid 1px currentColor',
           'border-radius': '10px',
           overflow: 'hidden',
-          padding: '0.5rem',
-          width: '320px',
+          width: '23rem',
           'margin': '0 0.5rem 0.5rem 0',
           'vertical-align': 'top',
         },
@@ -85,11 +84,10 @@ module View
 
       buttons << render_button('Start', -> { start_game(@gdata) }) if owner? && new? && players.size > 1
 
-      props = {
+      div_props = {
         style: {
           position: 'relative',
-          margin: '-0.5em',
-          padding: '0.5em',
+          padding: '0.3em 0.1rem 0 0.5rem',
           'background-color': color,
         },
       }
@@ -98,14 +96,15 @@ module View
         style: {
           color: 'black',
           display: 'inline-block',
-          width: '160px',
+          'max-width': '13rem',
         },
       }
+      owner_props = { attrs: { title: @gdata['user']['name'].to_s }, }
 
-      h('div', props, [
+      h('div.header', div_props, [
         h(:div, text_props, [
           h(:div, "Game: #{@gdata['title']}"),
-          h(:div, "Owner: #{@gdata['user']['name']}"),
+          h('div.nowrap', owner_props, "Owner: #{@gdata['user']['name']}"),
         ]),
         *buttons,
       ])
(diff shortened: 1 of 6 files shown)


models/base.rb
```diff
@@ -7,7 +7,23 @@ def pp_created_at
     created_at&.strftime('%Y-%m-%d')
   end
 
+  def pp_created_time
```
tobymao, line 10: let's not do this. get rid of all of these and just return created_at and updated_at as unix time stamps, time.to_i or something like that, and just render the time on the client, i updated chat to do the same

yzemaze: so just 2 defs there or scrap it altogether?

tobymao:
refactor everything to use the proper time stamps

assets/app/view/game_card.rb
```diff
@@ -148,55 +147,84 @@ def render_link(href, click, text)
...
       ]
 
       if new?
-        children << h(:div, [h(:b, 'Max Players: '), @gdata['max_players']])
-        children << h(:div, [h(:b, 'Created: '), @gdata['created_at']])
+        children << h(:div, { style: { display: 'inline' } }, [h(:strong, 'Max Players: '), @gdata['max_players']])
+        children << h(:div, { style: { display: 'inline', float: 'right' } }, [
+          h(:strong, 'Created: '),
+          h(:span, { attrs: { title: Time.at(@gdata['created_at']).strftime('%F %T') } },
+            if Time.at(@gdata['created_at']).strftime('%F') == Time.now.strftime('%F')
```
tobymao, line 193: this and the other stuff can be cleaned up

created_at = @gdata['created_at']

created_at >= Time.now - 86400 # could make this a function maybe

assets/app/view/game_card.rb
```diff
@@ -148,55 +147,84 @@ def render_link(href, click, text)
...
-        h(:div, [h(:b, 'Players: '), *p_elm]),
+        h(:div, [h(:strong, 'Id: '), @gdata['id'].to_s]),
+        h(:div, [h(:strong, 'Description: '), @gdata['description']]),
+        h(:div, [h(:strong, 'Players: '), *p_elm]),
       ]
 
       if new?
-        children << h(:div, [h(:b, 'Max Players: '), @gdata['max_players']])
-        children << h(:div, [h(:b, 'Created: '), @gdata['created_at']])
+        children << h(:div, { style: { display: 'inline' } }, [h(:strong, 'Max Players: '), @gdata['max_players']])
```
tobymao, line 189: don't you have css class inline?

assets/app/view/game_card.rb
```diff
@@ -148,55 +147,84 @@ def render_link(href, click, text)
...
           .map { |k, v| "#{k} (#{v})" }
           .join(', ')
 
         children << h(:div, [
-          h(:b, 'Result: '),
+          h(:strong, 'Result: '),
           result,
         ])
-      else
+      elsif @gdata['round'] || @gdata['updated_at']
```
tobymao, line 209: can you do elsif gdata['round'] 
elsif gdata['updated_at']

also i don't like this style of putting if end inside of an array

models/base.rb
```diff
@@ -4,10 +4,10 @@
 
 class Base
   def pp_created_at
```
tobymao, line 6: change the name of this because it's no longer pretty printed

assets/app/view/game_card.rb
yzemaze: O tempora! O mores! toby pushes CSS ;)
You did create it for inline-block (yay for blame). Will change.

assets/app/view/game_card.rb
yzemaze: If updated_at is only set when round is set, it’s even simpler.

assets/app/view/game_card.rb
yzemaze: note sure about 86400 as at least 1 hour less might be better to not run into overlap issues, but yes

public/assets/main.css
```diff
@@ -158,7 +159,12 @@ label.right {
   display: inline-block;
 }
 
-.inline { display: inline-block }
+.inline {
+  display: inline
```
tobymao, line 163: this seems like a dangerous change because all other inline elements were doing inline-block

yzemaze: It seems, until you grep for "\.inline". It returned only assets/app/view/game/issue_shares.rb. There I changed it to .inline-block. All other inline-blocks are defined inline.

PR #716, opened 2020-06-16
catch PG::UniqueViolation errors on username and email

yzemaze:
Flash pretty error message instead of
`PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "users_lower(name)_index" DETAIL: Key (lower(name))=(duplicate_username) already exists.`
when user name is already registered. Dito for email.

### DIFF
--- a/models/user.rb
+++ b/models/user.rb
@@ -22,6 +22,10 @@ class User < Base
     self[Sequel.function(:lower, :email) => email.downcase]
   end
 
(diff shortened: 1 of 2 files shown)


tobymao:
i think the proper way to do this is with this https://github.com/jeremyevans/sequel/blob/master/doc/validations.rdoc#validation_helpers-

PR #721, opened 2020-06-16
1836Jr close private on nord train buy

scottredracecar:
This is just copying the process from Chesapeake


tobymao:
instead of copy and pasting, we should generalize this as an ability

lib/engine/config/game/g_1836_jr30.rb
```diff
@@ -286,9 +286,14 @@ module G1836Jr30
             {
                "type":"tile_lay",
                "owner_type":"corporation",
+               "free_tile":"true",
```
tobymao, line 289: what is this?delete it

scottredracecar:
This is needed when the tile lay hex has a cost and this makes the tile lay
free. There is supporting code added in the tile laying code in base.

lib/engine/config/game/g_1836_jr30.rb
tobymao, line 289: we can just call this "free": true

no quotes around true

scottredracecar: okay, I fixed that

lib/engine/config/game/g_1836_jr30.rb
```diff
@@ -375,6 +380,10 @@ module G1836Jr30
                "type":"share",
                "share":"Nord_0"
             },
+            {
+               "type":"close_on_train_buy",
```
tobymao, line 384: maybe it should be "type": "close", "when": "buy_train"

lib/engine/phase.rb
```diff
@@ -21,6 +21,14 @@ def process_action(action)
         train = action.train
         next! if train.sym == @next_on
         rust_trains!(train, action.entity)
+        @game.companies.map do |company|
```
tobymao, line 24: this should not be map, it should be each because you have no use for the array, also make it into a method

lib/engine/phase.rb
```diff
@@ -21,6 +21,14 @@ def process_action(action)
...
+          next unless company.abilities(:close_on_train_buy)
```
tobymao, line 25: next unless (ability = company.abilities(:close))

PR #735, opened 2020-06-17
add tile blocking for 1846 privates (MC, OI)

### LINKED ISSUE #725: 1846 there is no indication on the map for the O&I and MC protected hexes
There is no indication on the map where the O&I (F14, F16) or MC (B10, B12) protected hexes are.

benjaminxscott:
closes #725


tobymao:
thank you!

PR #737, opened 2020-06-17
Proposed change to "Pass card" phrasing

### LINKED ISSUE #736: Consider labeling private cards "Pass" instead of just a number
In the online interface, it seems less important that "pass on buying a private" cards have a particular number, since it's not used for anything besides making sure you shuffle the correct number of them in for a given player count

This would diverge from official labeling, but may be a bit easier to parse for new players

benjaminxscott:
closes #736

### DIFF
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -50,8 +50,8 @@ module Engine
 
       def init_companies(players)
         super + @players.size.times.map do |i|
-          name = (i + 1).to_s
-          Company.new(sym: name, name: name, value: 0, desc: "Choose this card if you don't want to purchase a company")
+          name = "Pass on buying a private company this round"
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1846.rb
```diff
@@ -50,8 +50,8 @@ class G1846 < Base
 
       def init_companies(players)
         super + @players.size.times.map do |i|
-          name = (i + 1).to_s
-          Company.new(sym: name, name: name, value: 0, desc: "Choose this card if you don't want to purchase a company")
+          name = "Pass on buying a private company this round"
```
tobymao, line 53: change name to "Pass" since it's already in description.

use i as the sym in the constructor

benjaminxscott: added a commit, i'm not super familiar w/ ruby string interpolation so please don't block on me struggling with it if there's a particular approach you prefer

PR #744, opened 2020-06-17
enforce player numbers on game creation

### LINKED ISSUE #698: Tailor the Allowed Number of Players to the Player Counts Defined in the Config
18xx.games currently allows all games to be joined by 2-6 players. The latest game 1836Jr only supports 2-4 players. Starting cash and certificate limit are not defined for more than 4 players.

18xx.games should limit the number of players that can join a game to the max player count, for example by looking at the highest starting cash player count in the config.

yzemaze:
fixes #698 
+ min/max player numbers taken from config file / cert_limit
+ inputs for names get created/removed on change of player number
+ +/- player buttons removed
+ player number changes to max allowed if illegal (< min || > max) on title switch
+ player number doesn’t change when changing modes

Caveat: [Safari doesn’t respect min/max of input=number](https://caniuse.com/#feat=input-number) (yet?).

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -14,7 +14,7 @@ module View
     def render_content
       inputs = [
         mode_selector,
-        *render_buttons,
+        render_button('Create') { submit },
         render_inputs,
       ]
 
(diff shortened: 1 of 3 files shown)


assets/app/view/create_game.rb
```diff
@@ -74,28 +93,18 @@ def mode_selector
...
-    def render_buttons
-      buttons = []
-
-      buttons << render_button('Create') { submit }
-
-      if @mode == :hotseat
-        buttons << render_button('+ Player') { store(:num_players, @num_players + 1) if @num_players + 1 <= 6 }
-        buttons << render_button('- Player') { store(:num_players, @num_players - 1) if @num_players - 1 >= 2 }
+      click_handler = lambda do
+        store(:mode, mode)
```
tobymao, line 97: this does a double render you need to do skip: true

assets/app/view/form.rb
```diff
@@ -69,9 +73,15 @@ def render_input(label, placeholder: '', id:, el: 'input', type: 'text', attrs:
       )
     end
 
-    def render_button(text, &block)
+    def render_button(text, style: {}, &block)
       props = {
-        attrs: { type: :button },
+        attrs: {
+          id: text.gsub(/\s/, '-').downcase,
```
tobymao, line 79: i feel like these are just unnecessary, but if you really want them you can keep it, it's just extra code and line 60

yzemaze: As I said before: on input an id is needed to connect label and input => clicking on label to activate input won’t work otherwise (if you don’t change the html structure)
On buttons they come in handy if you’d like to address one specific button. Kill L79 if you must.

tobymao: where do you reference the id?

assets/app/view/create_game.rb
```diff
@@ -44,26 +44,45 @@ def render_content
...
+        title += " (#{game::DEV_STAGE})" if game::DEV_STAGE != :production
+
+        h(:option, { attrs: { value: game.title } }, title)
       end
-      h(:div, [
-        render_input('Game Title', id: :title, el: 'select', children: games),
-        render_input('Description', id: :description),
+
+      limit_range = lambda do
+        min = Native(@inputs[:max_players]).elm.min = @min_p[Native(@inputs[:title]).elm.value]
```
tobymao, line 62: why call Native(@inputs) n times, store that first and then you can use that

assets/app/view/create_game.rb
```diff
@@ -44,26 +44,45 @@ def render_content
...
+        max = Native(@inputs[:max_players]).elm.max = @max_p[Native(@inputs[:title]).elm.value]
+        val = Native(@inputs[:max_players]).elm.value.to_i
+        Native(@inputs[:max_players]).elm.value = (val < min || val > max ? max : val)
```
tobymao, line 65: val.between?(min, max) or (min..max).incude?(val) probably need to check the end points

assets/app/view/form.rb
yzemaze: labels: for: 
buttons: CSS e. g. (notice the *if* there)

tobymao: can you pass the id in explicitly and not do multiple gsubs?

assets/app/view/create_game.rb
yzemaze: I had used tried include, but ran into some string/integer probs (which I solved after changing, didn’t revert though). between? won’t work, will change it to include.

assets/app/view/form.rb
yzemaze: render_input: one gsub would be enough. It shouldn’t be needed under normal circumstances (who puts \s in their ids?), but it’s there to safeguard a viable id.

render_button: Of course, but that would be an additional parameter when calling. This way it gets created without the need for that. It’s just a useful default without much hassle. As I said it may come in handy (or it may not).

yzemaze: Just let me know if you want them removed, everything else is done.

tobymao: ok, let's just remove it then if it was just done as a safe guard

assets/app/view/create_game.rb
yzemaze: … because sometimes I don’t see the wood for the trees. Fixed.

assets/app/view/form.rb
tobymao: still waiting on the min_max native changes

assets/app/view/create_game.rb
```diff
@@ -44,26 +44,47 @@ def render_content
...
-        render_input('Description', id: :description),
+
+      limit_range = lambda do
+        range = Native(@inputs[:max_players]).elm
+        title = Native(@inputs[:title]).elm.value
+        min = range.min = @min_p[title]
+        max = range.max = @max_p[title]
+        val = range.value.to_i
+        range.value = (min..max).include?(val) ? val : max
+        store(:num_players, range.value.to_i)
```
tobymao, line 68: you can just use val since you already calculated that

Issue #661, opened 2020-06-08
Relabel Chesapeake Stock Round header to "Sell then Buy Shares"


rthomp9595:
suggest small tweak to header for Stock Round in Chesapeake, more aligned with rules if the header reads "Sell then Buy Shares"...just help players intuitively navigate SR

ryanespin:
Is this something for all games or specifically just for Chesapeake?

tobymao:
all games right now but could be configurable

ryanespin:
Cool, made the tweak then for all games. :)

PR #764, opened 2020-06-18
Changed description text for Stock Round.

ryanespin:
Addresses Issue #661

### DIFF
--- a/lib/engine/round/stock.rb
+++ b/lib/engine/round/stock.rb
@@ -32,7 +32,7 @@ module Engine
       end
 
       def description
-        'Buy and Sell Shares'
+        'Sell then Buy Shares'
       end
 
       def pass_description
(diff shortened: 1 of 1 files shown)


tobymao:
this is not correct for 1889 since you sell buy or buy sell

ryanespin:
Ah okay, let me change it to be configurable.  Thanks!

jenf:
Use sell_buy_order variable

ryanespin:
@jenf, used the variable.  Not sure if this is the most optimal way though.

lib/engine/round/stock.rb
```diff
@@ -32,7 +32,14 @@ def name
       end
 
       def description
-        'Buy and Sell Shares'
+        case @sell_buy_order
+        when :sell_buy_or_buy_sell
+          'Buy and Sell Shares'
```
tobymao, line 37: how about 

Buy or Sell Shares
Sell Then Buy Shares
Sell then Buy then Sell Shares
Sell Buy, Sell Shares

PR #766, opened 2020-06-18
homepage (slightly) revised

yzemaze:
+ chat height increase if > 900px width
+ chat resize only vertical
+ h1/h2
+ #/.
+ paddings/margins pruned
+ prev/next fixed position

### DIFF
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -31,9 +31,7 @@ module View
         on: { scroll: scroll_handler },
         style: {
           overflow: 'auto',
-          height: '200px',
-          width: '100%',
-          padding: '0.5rem',
+          padding: '0 0.5rem',
           'background-color': color_for(:bg2),
           color: color_for(:font2),
           'word-break': 'break-word',
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -31,7 +31,7 @@ module View
         on: { scroll: scroll_handler },
         style: {
           overflow: 'auto',
-          padding: '0 0.5rem',
+          padding: '0.5rem',
           'background-color': color_for(:bg2),
           color: color_for(:font2),
           'word-break': 'break-word',
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -7,12 +7,11 @@ module View
     def render
       props = {
         style: {
-          display: 'inline-block',
-          'margin-right': '1rem',
+          'margin-right': '1.5rem',
         },
       }
 
-      h('div.half', props, [
+      h('div#welcome.half', props, [
         render_notification,
         render_introduction,
         render_buttons,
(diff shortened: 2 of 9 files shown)


assets/app/view/log.rb
```diff
@@ -31,9 +31,7 @@ def render
         on: { scroll: scroll_handler },
         style: {
           overflow: 'auto',
-          height: '200px',
-          width: '100%',
-          padding: '0.5rem',
+          padding: '0 0.5rem',
```
tobymao, line 34: this will mess up emails

assets/app/view/welcome.rb
```diff
@@ -23,33 +22,34 @@ def render_notification
...
+        <p>You can support this project on <a href='https://www.patreon.com/18xxgames'>Patreon</a>.
         </p>
       MESSAGE
 
       props = {
         style: {
           background: 'rgb(240, 229, 140)',
           color: 'black',
-          'padding': '1em',
+          marginBottom: '1rem',
```
tobymao, line 45: should it be margin-bottom?

public/assets/main.css
```diff
@@ -12,6 +12,9 @@ a.game__nav:focus {
 button {
   margin: 0.5rem 0.5rem 0 0;
 }
+button + button {
```
tobymao, line 15: does this mess up any other views?

public/assets/main.css
```diff
@@ -241,6 +253,27 @@ table.center {
...
+#chatlog {
+  height: 200px;
+}
+@media only screen and (min-width: 900px) {
+  #homepage #chatlog {
+    height: 350px;
+  }
+}
+
+.game_row > .card_header {
```
tobymao, line 265: is this necessary to have in css?

public/assets/main.css
yzemaze: **If** there were buttons next to each other (without any el between them!) that aren’t inline and don’t use inline-styles for margin, maybe. I haven’t found any.
To me this is a useful default, hence I put it here.

public/assets/main.css
yzemaze: Not for this to work here. It would be useful if it’s needed in other places.

assets/app/view/welcome.rb
yzemaze: Either `'margin-bottom'` or `marginBottom`, cf. https://github.com/snabbdom/snabbdom/blob/d9c4abd7c5b166ff7abd705fc920eb0933b9f0a0/README.md#example
CamelCase is just easier to type.

assets/app/view/log.rb
yzemaze: Ok, will change. But mails should have a text version without HTML/CSS ;)

assets/app/view/welcome.rb
tobymao: didn't realize you could do that

PR #778, opened 2020-06-19
Allow user to choose home token in 1836Jr30/1882

jenf:
Screenshot from 2020-06-19 18-26-04
Screenshot from 2020-06-19 18-26-34

1836Jr30:
 OO yellow hex (Erie-style company) NBDS company needs to be implemented.
(Likely games will be broken)

1882:
J10 (Pile o'Bones) should have a label saying QLL starts there
QLL is kind of like Erie in 1830. Selects a home when it forms. Must leave a space open for it.
CNR doesn't have a home station

Also now hides the city reservation text when laid which made 1846 ERIE look a bit weird


lib/engine/part/city.rb
```diff
@@ -53,11 +53,18 @@ def tokenable?(corporation, free: false, tokens: nil)
           next false unless get_slot(t.corporation)
           next false if !free && t.price > corporation.cash
           next false if @tile.cities.any? { |c| c.tokened_by?(t.corporation) }
+          next true if @reservations.index(corporation.name)
```
tobymao, line 56: why do you need this one?

lib/engine/round/base.rb
```diff
@@ -167,16 +167,28 @@ def discard_train(action)
...
+        return false unless corporation.next_token # 1882
 
         hex = @game.hexes.find { |h| h.coordinates == corporation.coordinates }
-        cities = hex.tile.cities
+
+        tile = hex.tile
+        if tile.reserved_by?(corporation)
+          @log << "#{corporation.name} must choose city for home token"
+          # Needs further changes to support non-operate home token lay
+          raise GameError('Unsupported') unless @home_token_timing == :operate
```
tobymao, line 180: raise GameError, 'Unsupported'

lib/engine/round/operating.rb
```diff
@@ -213,7 +215,12 @@ def connected_paths
       end
 
       def reachable_hexes
-        @graph.reachable_hexes(@current_entity)
+        if @step == :home_token
+          homehex = @game.hexes.find { |h| h.coordinates == @current_entity.coordinates }
```
tobymao, line 219: { @game.hex_by_id(@current_entity.coordinates) => true }

lib/engine/round/operating.rb
```diff
@@ -237,6 +244,14 @@ def next_step!
         end
       end
 
+      def skip_home_token
+        if @home_token_timing == :operate
```
tobymao, line 248: return !place_home_token(@current_entity) if @home_token_timeing == operati

lib/engine/round/operating.rb
```diff
@@ -351,7 +366,7 @@ def start_operating
         @current_actions.clear
         log_operation(@current_entity)
         @current_entity.trains.each { |train| train.operated = false }
-        place_home_token(@current_entity) if @home_token_timing == :operate
+
```
tobymao, line 369: hmmmmmmm this is weird that this happens in a SKIP check

lib/engine/round/operating.rb
```diff
@@ -508,16 +523,17 @@ def buy_company(company, price)
       def place_token(action)
         entity = action.entity
         hex = action.city.hex
-        if !@game.loading && !connected_nodes[action.city] && !@teleported
+        free = @teleported || @step == :home_token
+        if !@game.loading && !connected_nodes[action.city] && !free
```
tobymao, line 527: move the free check before connected_nodes because connected nodes is an expensive call and better to short circuit

tobymao: hmmm but free refers to cost but you're overloading it here with connectivity?

lib/engine/tile.rb
```diff
@@ -222,6 +223,21 @@ def preferred_city_town_edges
       @preferred_city_town_edges ||= compute_city_town_edges
     end
 
+    def reserved_by?(corporation)
```
tobymao, line 226: do we need to store reservations or can we just get them from the cities?

lib/engine/round/operating.rb
jenf: Yeah, the other way would be to do a second check here to see if it needs laying

lib/engine/tile.rb
jenf: No, we need to store since they're disconnected from the cities

lib/engine/round/operating.rb
jenf: I've changed it to a secondary check

lib/engine/part/city.rb
jenf: I originally had

next false if !@reservations.index(corporation.name) && @tile.reservations.count { |x| corporation.name != x } >= @tile.cities.sum(&:available_slots)

but rubocop said it was too long :)

The alternative is to see if @tile.reservations.count is >=1 and then do the check on the available slots.

It gets complicated in multicity case otherwise.

PR #793, opened 2020-06-20
18EU special rules

scottredracecar:


lib/engine/round/g_18_EU/operating.rb
```diff
@@ -0,0 +1,138 @@
...
+          @log << "#{@current_entity.name} runs for #{@game.format_currency(revenue)} and pays half"
+          @log << "#{@current_entity.name} witholds #{@game.format_currency(withheld)}"
+          payout(revenue - withheld)
+        end
+
+        def change_share_price(_direction, revenue = 0)
+          return if @current_entity.minor?
+
+          price = @current_entity.share_price.price
+          @stock_market.move_left(@current_entity) if revenue.zero
```
tobymao, line 131: zero?

tobymao: you don't need this section, you can just do return super unless @current_entity.minor?

lib/engine/round/g_18_EU/operating.rb
```diff
@@ -0,0 +1,138 @@
...
+
+          amount = revenue / 2
+
+          [@current_entity, @current_entity.owner].each do |entity|
+            @log << "#{entity.name} receives #{@game.format_currency(amount)}"
+            @bank.spend(amount, entity)
+          end
+        end
+
+        def half(revenue)
```
tobymao, line 119: we should turn this into a module

PR #806, opened 2020-06-20
add icons for 1889, 1846, 1882

### LINKED ISSUE #613: 1889 Mitsubishi Ferry Port tile should not upgrade tile 58

### LINKED ISSUE #647: Icons, starting with port icon for 1889 and 1846
- 1889: render icon on port cities, replace 437's "P" with port icon, fix 437 upgrading logic
- 1846: render steamboat hexes with the same port icon, ensure the icon is still visible after tile lay (e.g., Wheeling and Toledo, also applicable for LSL in Cleveland and Toledo)

(related: #613)

michaeljb:
- when laying a tile, always verify that it's a valid upgrade
- raise error when loading game config JSON if there are multiple definitions
  for a single hex
- new "layable_by" tile part to prevent 437 from being laid by any entity
  other than Mitsubishi Ferry (MF); currently this just blocks it from appearing
  in the list of potential tiles, it doesn't raise an error meaning old games
  that violated this won't break

Icon stuff:

- 1889: port icon
  - ditch "P" label on 437, use port icon instead; add port icon to appropriate
    hexes on map
  - fix laying 437; prevent laying it on top of an already laid tile
- 1846: steamboat and meat icons (rendering only)
  - meat on Chicago, steamboat in Toledo, and steamboats in Wheeling persist
    when tile is laid on top of them ("sticky" property)
  - no bonus revenue functionality, just rendering
- 1882: NWR icon (rendering only)
  - all use the sticky property so they are visible when tiles are laid on top
    of them

[Fixes #613]
[Fixes #647]

Screenshot from 2020-06-20 15-39-45
Screenshot from 2020-06-20 15-39-14
Screenshot from 2020-06-20 15-38-49
Screenshot from 2020-06-20 15-45-26


assets/app/view/game/part/icons.rb
```diff
@@ -0,0 +1,57 @@
...
+        ICON_RADIUS = 16
+        DELTA_X = (ICON_RADIUS * 2) + 2
+
+        def preferred_render_locations
+          if layout == :pointy
+            if @num_cities > 1
+              [{
+                 region_weights: { BOTTOM_RIGHT_CORNER => 1 },
+                 x: (DELTA_X / 2) * (@icons.size - 1),
+                 y: 79.5,
```
tobymao, line 18: the only difference here is the y, can we make this more dry?

assets/app/view/game/part/icons.rb
```diff
@@ -0,0 +1,57 @@
...
+        end
+
+        def load_from_tile
+          @icons = @tile.icons
+          @num_cities = @tile.cities.size
+        end
+
+        def render_part
+          children = @icons.map.with_index do |icon, index|
+            h(:image,
```
tobymao, line 43: can be a one liner

assets/app/view/game/tile.rb
```diff
@@ -76,6 +76,8 @@ def render
         @tile.reservations.each { |x| children << render_tile_part(Part::Reservation, reservation: x) }
         children << h(Part::Borders, tile: @tile) if @tile.borders.any?
 
+        children << render_tile_part(Part::Icons) if @tile.icons.any?
```
tobymao, line 79: let's move icons before borders, borders are always the top layer

lib/engine/game/base.rb
```diff
@@ -191,6 +191,10 @@ def self.load_from_json(json)
         data['hexes'].transform_keys!(&:to_sym)
         data['hexes'].transform_values!(&:invert)
 
+        hex_ids = data['hexes'].values.map(&:keys).flatten
+        dup_hexes = hex_ids.each_with_object(Hash.new(0)) { |o, h| h[o] += 1 }.select { |_k, v| v > 1 }.keys
```
tobymao, line 195: hex_ids.group_by(&:itself).select { |_, v| v.size > 1 } .keys

lib/engine/part/icon.rb
```diff
@@ -0,0 +1,23 @@
...
+require_relative 'base'
+
+module Engine
+  module Part
+    class Icon < Base
+      attr_accessor :preprinted
+      attr_reader :image, :name, :sticky
+
+      def initialize(image, name = nil, sticky = true, preprinted = true)
+        @image = "/icons/#{image}.svg"
```
tobymao, line 12: you need to update nginx

tobymao: would be nice if you could also DRY up my nginx config, i couldn't figure out how to OR multiple image paths

lib/engine/tile.rb
```diff
@@ -96,6 +96,12 @@ def self.part(type, params, cache)
         junction = Part::Junction.new
         cache << junction
         junction
+      when 'icon'
+        Part::Icon.new(params['image'],
+                       params.fetch('name', params['image']),
```
tobymao, line 101: let's not do anything fancy here, do it in the initializer, so just raw params['name'], params['sticky']

then in intiailizer

@name = name || image etc

also sticky should be a boolean, so it should be real truthy value

tobymao:
i don't really like layable by, i think perhaps it should be unlayable: true and then special entities can lay it like companies with their tile laying abilities,

lib/engine/part/icon.rb
michaeljb: hmm, guessing you tried separating the paths with `|` like the line `location ~ \.(xml|png|ico|webmanifest) {`?

Can't get my local prod stack working, so I can't test out nginx changes right now. I've got these values set in my environment:

```
export POSTGRES_DB=18xx_production
export POSTGRES_PASSWORD=secret
export POSTGRES_USER=18xx
```
Getting this error:

```
queue_1  | Sequel failed to connect: PG::ConnectionBad: FATAL:  password authentication failed for user "18xx"
```

Issue #755, opened 2020-06-18
1846 Teleport tokens for the B&O to Cincy not working properly. #Bug


mhrabak:
It seems that you cannot teleport the token to Cincy for the B&O and pay the higher associated cost for not having a connected track.

KillerOrca:
Game ID 2517 also affected by this.

Issue #748, opened 2020-06-17
[1846] IC's $40 token in Centralia not implemented


pence:
IC has a discount for placing a token in Centralia, but it still appears to cost $80 here

michaeljb:
Looks like this was fixed by  #837

Issue #57, opened 2020-04-16
session report


sharunasbresson:
dear Toby, i write you about my first long game session with your app.
First of all, thanks fro your great work! I've tried before playing online 18xx but it's the first time I'm looking forward doing it again and again after one session!

Here my feedback for the current state of the art (apologies for my poor english, I'm an italian after all :-) ) :
The game I'm writing about is the id74
There are minor presentation issues and I know you're dealing quickly with them (yesterday when i opened the game the new corporation look was a very pleasant surprise).
I would like to get some more information on the game maps (train phases, train limits).
Especially as soon as more 18xx game will be implemented with your system the need of this kind of minutiae will get more important.
The bank pool with the already sold stock would be also useful.
Maybe different color for the different player name (but I'm nitpicking now).

The bug or major issues I found are the following:
when a corporation bought the sumitomo company and tried to lay a mountain track without river
the power to ignore the building cost was not active, meaning the 80 $ was regularly charged on the company treasure (maybe we didn't find a way to make this work, in this better scenario my “complaint” would regard the usability, but I suspect the this rules was just not working).

Same thing with the train limit. The agenda showed us that as expected we were entering the phase 5 and the train limits of 2 but it was only when we entered phase six that we were forced to have just two trains in the corporation. In phase 5 a corporation was also able to buy a third train.
But even if we were careful to fair play, not being able to discard the trains in excess was a gamebreaker.

The other and possibly bigger issue we had was the ability to buy trains form other corporations at a proce different from 1$. We used the price tab and tried by clicking but also by writing the amountof money we wished to pay and it always decided we were going to pay 1$.
Obviously after that we just continued the game for test, not for “enjoying the game” (wich really was until then really enjoyabe, and we are very very grateful).

Probably a minor bug (but I think I wouldn't be an accurate tester if I wouldn't mention it) occured just once, when the Iyo railway was able to run the same route twice. The director of the company saw what was happening and used the undo button to avoid the mistake, but the app was allowing him to do it, while in an other occurrenc eof the same mistake from another corporation the app showed us the mistake.

Finally, I saw on the forum that someone perceived as useful the possibility to control form one computer the other online player's moves.
We thought the opposite, because sometimes inadvertently someone thinking to act for himself made  adecision for another player (we manage to undo, but we were uncomfortable about it).

I hope that with this letter I'm “making my job” by offering you something in exchange of the wonderful gift you are providing us all.

[...]

PR #813, opened 2020-06-21
Fixes 1846 token teleport and discount powers.

adamranderson:
Changes 1846 to use the already existing teleport ability for tokens. Adds a price to the ability that 46 can set for each company and other games (Ches, 36jr) can set to 0 to make the teleport token free. This leads to a lot of cleanup in the 1846 operating round subclass.

Adds a discounted_token ability to cover to cover the 1846 token discounts for being connected to special hexes.

#748 #755

### DIFF
--- a/lib/engine/config/game/g_1836_jr30.rb
+++ b/lib/engine/config/game/g_1836_jr30.rb
@@ -321,7 +321,8 @@ module Engine
                 ],
                "hexes":[
                   "J8"
-               ]
+               ],
+               "token_price": 0
             }
          ]
       },
--- a/lib/engine/config/game/g_1846.rb
+++ b/lib/engine/config/game/g_1846.rb
@@ -285,7 +285,8 @@ module Engine
           "hexes": [
             "E11"
           ],
-          "token_price": 60
+          "token_price": 60,
+          "track_optional": true
         },
         {
           "type": "discounted_token",
@@ -330,7 +331,8 @@ module Engine
           "hexes": [
             "H12"
           ],
-          "token_price": 100
+          "token_price": 100,
+          "track_optional": true
         },
         {
           "type": "discounted_token",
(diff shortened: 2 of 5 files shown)


tobymao:
so this is amazing if it works, but i went down the path of trying to mix this into teleporting and i ran into a lot of issues which is why i make a new ability.

for example, a teleport requires you to lay a tile in order to trigger. but in 1846, you can token first and then lay a tile.

from the code it looks like this won't work


` if ability[:hexes].include?(hex_id) && ability[:tiles].include?(action.tile.name)
          ability[:tiles].include?(action.tile.name)	            @teleported = true
            @teleport_token_price = ability[:token_price] || 0
          end```

adamranderson:
Hmmm yes, the teleport code does expect the track lay to happen before the token which is incorrect here. I think it would be possible to change teleport to allow the token without track. It looks like the rules of 18Chesapeake need it too.

"P4 Chesapeake and Ohio Canal. Blocks hex D2 while
owned by a player. The owning company **may** place a #57
(straight city) tile in hex D2. The company does not need to
have a route to this hex. The tile placed counts as the
company’s tile lay action and the company must pay the
terrain cost. The company may then immediately place a
station token in hex D2 free of charge. The token used must
be the next available cheapest station."

The rules for 1836jr Charbonnages du Hainaut are ambiguous.
"Owning Corporation may place a tile and
station token in the CdH hex for only the F60
cost of the mountain."

scottredracecar:
36Jr follows 1830 D&H, same intent with Chesapeake.

tobymao:
is it possible for you to clean up your PR, it's hard to tell what's changed now that you've done a big merge.

i'm also doing a large refactor on abilities so i was thinking about just cherry picking your changes, it might be easier to chat on the 18xx slack if you're available, i've pinged you

PR #821, opened 2020-06-21
1836Jr30: Can only by one of each rank from the depot on a round

jenf:
Implements
A corporation may only buy one of each train type per operating turn

### DIFF
--- a/lib/engine/game/g_1836_jr30.rb
+++ b/lib/engine/game/g_1836_jr30.rb
@@ -15,6 +15,10 @@ module Engine
 
       SELL_BUY_ORDER = :sell_buy_sell
       TRACK_RESTRICTION = :permissive
+
+      def operating_round(round_num)
+        Round::G1836Jr30::Operating.new(@corporations, game: self, round_num: round_num)
+      end
     end
   end
 end
--- a/lib/engine/round/g_1836_jr_30/operating.rb
+++ b/lib/engine/round/g_1836_jr_30/operating.rb
@@ -17,7 +17,7 @@ module Engine
         end
 
         def buyable_trains
-          super.reject { |x| x.from_depot? && @depot_trains_bought.include?(x.distance) }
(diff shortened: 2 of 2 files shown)


lib/engine/game/g_1836_jr30.rb
```diff
@@ -15,6 +15,10 @@ class G1836Jr30 < Base
 
       SELL_BUY_ORDER = :sell_buy_sell
       TRACK_RESTRICTION = :permissive
+
+      def operating_round(round_num)
```
tobymao, line 19: why do you need this?

lib/engine/round/g_1836_jr_30/operating.rb
```diff
@@ -0,0 +1,36 @@
...
+          @depot_trains_bought = []
+        end
+
+        def start_operating
+          @depot_trains_bought = []
+          super
+        end
+
+        def buyable_trains
+          super.reject { |x| x.from_depot? && @depot_trains_bought.include?(x.distance) }
```
tobymao, line 20: better to track by the sym, train.sym

lib/engine/round/g_1836_jr_30/operating.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        def can_buy_train?
+          super && (@depot_trains_bought.empty? || buyable_trains.any?)
+        end
+
+        def process_buy_train(action)
+          # Since the train won't be in the depot after being bougth store the state now.
+          add_to_list = action.train.from_depot?
+          super
```
tobymao, line 30: can't we do

@depoat_trains_bought << action.train.sym if action.train.from_depot?
super

jenf: nope, since process_buy_trains calls buyable_trains and means no train can be bought

PR #826, opened 2020-06-22
1882: Implement NWR event and fix a few minor issues

jenf:
Place the "NWR Card" under the first [3/4/5/6] train (select 3/4/5/6 randomly) at game setup
It's probably best implementing this as an event. However it should also be included in the 'Info' page as well as put into the log.
When the NWR card is revealed, remove all yellow tiles in hexes marked NWR from the board (back to the tile stock).
Improve the info page for events
P2 stays open until Phase 6 unlike the others that close on Phase 5.

Not on 1882 list:
Fix phase changes
Fix company purchasing
Fix storage of neutral token

### DIFF
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -81,21 +81,34 @@ module View
               Lib::Color.convert_hex_to_rgba(Part::MultiRevenue::COLOR[phase_color], 0.4)
           end
 
-          buy_text = if phase[:buy_companies]
-                       'Can Buy'
-                     elsif phase[:events]&.include?(:close_companies)
-                       'Close'
-                     else
-                       ''
-                     end
+          buy_text = []
+
+          buy_text << 'Can Buy Companies' if phase[:buy_companies]
+          phase[:events]&.each do |name, _value|
+            buy_text << (@game.class::EVENTS_TEXT[name] ? "#{@game.class::EVENTS_TEXT[name][0]}*" : name)
+          end
 
           h(:tr, [
             h(:td, td_props, phase[:name] + (current_phase == phase ? ' (Current) ' : '')),
             h(:td, td_props, phase[:operating_rounds]),
             h(:td, td_props, phase[:train_limit]),
             h(:td, phase_props, phase_color.capitalize),
-            h(:td, td_props, buy_text),
+            h(:td, td_props, buy_text.join(',')),
           ])
+        end
+
+        phase_text = @game.class::EVENTS_TEXT.map do |_sym, desc|
+          h(:tr, [h(:td, td_props, desc[0]), h(:td, td_props, desc[1])])
+        end
+
+        if phase_text.any?
+          phase_text = [h(:table, [
+            h(:tr, [
+              h(:th, td_props, 'Event'),
+              h(:th, td_props, 'Description'),
+              ]),
+            *phase_text,
+          ])]
         end
 
         props = {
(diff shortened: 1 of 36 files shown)


lib/engine/game/base.rb
```diff
@@ -766,6 +769,14 @@ def operating_round(round_num)
         Round::Operating.new(@corporations, game: self, round_num: round_num)
       end
 
+      def event_close_companies!
+        @log << '-- Event: Private companies close --'
+
+        @companies.each do |company|
+          company.close! unless company.abilities(:closes) || company.abilities(:never_closes)
```
tobymao, line 776: don't use unless with multiple conditions

lib/engine/game/g_1882.rb
```diff
@@ -48,6 +64,21 @@ def init_corporations(stock_market)
         end
         corporations
       end
+
+      def event_nwr!
+        @log << '-- Event: North West Rebellion! --'
+        name = '1882/NWR'
+        @hexes.each do |hex|
+          next unless hex.tile.icons.any? { |x| x.name == name }
```
tobymao, line 72: icon.name

lib/engine/game/g_1882.rb
```diff
@@ -26,6 +26,22 @@ class G1882 < Base
...
+        'nwr' => ['North West Rebellion',
+                  'Remove all yellow tiles from NWR-marked hexes. Station markers remain']
+      ).freeze
+
+      def init_phase
+        phases = self.class::PHASES
+        nwr_phases = %w[3 4 5 6]
+        nwr_phase = nwr_phases[rand % nwr_phases.size]
+        @log << "NWR Rebellion occurs at start of phase #{nwr_phase}"
+        phases.each do |x|
```
tobymao, line 39: phase

assets/app/view/game/game_info.rb
```diff
@@ -81,23 +81,35 @@ def phases
...
 
-          buy_text = if phase[:buy_companies]
-                       'Can Buy'
-                     elsif phase[:events]&.include?(:close_companies)
-                       'Close'
-                     else
-                       ''
-                     end
+          event_text = []
+          event_text << 'Can Buy Companies' if phase[:buy_companies]
```
tobymao, line 85: why don't we make this an event too

tobymao: or maybe not since it's oanly on etime

lib/engine/game/g_1882.rb
```diff
@@ -48,6 +64,21 @@ def init_corporations(stock_market)
...
         corporations
       end
+
+      def event_nwr!
+        @log << '-- Event: North West Rebellion! --'
+        name = '1882/NWR'
+        @hexes.each do |hex|
+          next unless hex.tile.icons.any? { |x| x.name == name }
+
+          next unless hex.tile.color == :yellow && hex.tile != hex.original_tile
```
tobymao, line 74: unless with multiple args

lib/engine/game/g_1882.rb
jenf: ?

lib/engine/game/base.rb
```diff
@@ -772,6 +775,18 @@ def operating_round(round_num)
         Round::Operating.new(@corporations, game: self, round_num: round_num)
       end
 
+      def event_close_companies!
+        @log << '-- Event: Private companies close --'
+
+        @companies.each do |company|
+          ability = company.abilities(:close)
```
tobymao, line 782: let's use block style 

company.abilities(:close) do |ability|
  next if ability.when == 'never' # note this is a string and not a symbol like you have it here

i think there's a bug with the when stuff that we need to fix i guess when should always be a string?  since phase names are strings

PR #831, opened 2020-06-22
Game page map inline

yzemaze:
- map inline
- map-size calculation corrected to prevent pagebreaks due to map width set wider than necessary
- buy_companies wrapped & display: flex to avoid setting max-width (which could be a problem later on)
- min. 3 companies will be rendered inline (with rem <= 16px), max. depends on width of map
- route_selection slimmed

### DIFF
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -8,6 +8,7 @@ module View
       needs :selected_company, default: nil, store: true
       needs :game, store: true
       needs :tile_selector, default: nil, store: true
+      needs :inline, default: true
 
       def selected?
         @company == @selected_company
@@ -65,14 +66,13 @@ module View
 
         props = {
           style: {
-            display: 'inline-block',
             cursor: 'pointer',
             border: 'solid 1px gainsboro',
             'border-radius': '10px',
             overflow: 'hidden',
             padding: '0.5rem',
             margin: '0.5rem 0.5rem 0 0',
-            width: '300px',
+            width: '19rem',
             'text-align': 'center',
             'font-weight': 'bold',
             'vertical-align': 'top',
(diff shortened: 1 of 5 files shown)


assets/app/view/game/company.rb
```diff
@@ -65,14 +66,13 @@ def render
...
           style: {
-            display: 'inline-block',
             cursor: 'pointer',
             border: 'solid 1px gainsboro',
             'border-radius': '10px',
             overflow: 'hidden',
             padding: '0.5rem',
             margin: '0.5rem 0.5rem 0 0',
-            width: '300px',
+            width: '19rem',
```
tobymao, line 75: is this going to work on all browsers?

assets/app/view/game/map.rb
```diff
@@ -92,9 +92,16 @@ def map_y
       end
 
       def render_map
-        w_size, h_size = @layout == :flat ? [85, 50] : [50, 85]
-        width = (@cols.size * w_size) + GAP
-        height = (@rows.size * h_size) + GAP
+        flat2flat = 85
+        edge_length = 50
```
tobymao, line 96: make these constants

i'm not sure what flat2flat means maybe choose a better name and use snake case

assets/app/view/game/company.rb
yzemaze: I can’t test ’em all (FF, Chrome + Safari/iOS worked fine), but 19rem + 1rem padding = 20rem = width of corps and player cards. So if nobody did complain about them until now I wouldn’t expect any problems.
If font-size (user-setting in browser!) is set to 16px all these cards are 320px wide (+1 px border each side)

assets/app/view/game/map.rb
yzemaze: flat2flat = flat side to flat side = [apothem](https://mathworld.wolfram.com/Apothem.html) * 2

PR #836, opened 2020-06-22
Allow tiles to have a zero 'movement' cost to run for free by a train

jenf:
There is a fishing hex that counts for free onto a route.

### DIFF
--- a/lib/engine/config/game/g_1882.rb
+++ b/lib/engine/config/game/g_1882.rb
@@ -637,7 +637,7 @@ module Engine
       ]
     },
     "blue": {
-      "offboard=revenue:yellow_20|brown_30,movement_cost:0;path=a:0,b:_0;path=a:1,b:_0": [
+      "offboard=revenue:yellow_20|brown_30,visit_cost:0;path=a:0,b:_0;path=a:1,b:_0": [
         "B6"
       ]
     }
--- a/lib/engine/config/game/g_1882.rb
+++ b/lib/engine/config/game/g_1882.rb
@@ -637,7 +637,7 @@ module Engine
       ]
     },
     "blue": {
-      "offboard=revenue:yellow_20|brown_30;path=a:0,b:_0;path=a:1,b:_0": [
+      "offboard=revenue:yellow_20|brown_30,movement_cost:0;path=a:0,b:_0;path=a:1,b:_0": [
         "B6"
       ]
     }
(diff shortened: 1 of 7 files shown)


lib/engine/config/game/g_1882.rb
```diff
@@ -637,7 +637,7 @@ module G1882
       ]
     },
     "blue": {
-      "offboard=revenue:yellow_20|brown_30;path=a:0,b:_0;path=a:1,b:_0": [
+      "offboard=revenue:yellow_20|brown_30,movement_cost:0;path=a:0,b:_0;path=a:1,b:_0": [
```
tobymao, line 640: maybe we should rename it distance since it maps to train distance

lib/engine/tile.rb
```diff
@@ -73,15 +73,25 @@ def self.part(type, params, cache)
 
         Part::Path.new(params['a'], params['b'])
       when 'city'
-        city = Part::City.new(params['revenue'], params.fetch('slots', 1), params['groups'], params['hide'])
+        city = Part::City.new(params['revenue'],
+                              params.fetch('slots', 1),
+                              params['groups'],
+                              params['hide'],
+                              params.fetch('movement_cost', 1))
```
tobymao, line 80: there's no reason to fetch because you have a default already

lib/engine/part/revenue_center.rb
```diff
@@ -5,14 +5,19 @@
...
 
-      def initialize(revenue, groups = nil, hide = false)
+      def initialize(revenue, groups = nil, hide = false, movement_cost = 1)
         @revenue = parse_revenue(revenue)
         @groups = (groups || '').split('|')
         @hide = hide
+        @movement_cost = movement_cost.to_i
+      end
+
+      def free_stop?
```
tobymao, line 19: vestigial?

lib/engine/config/game/g_1882.rb
scottredracecar: I don't know if this makes sense, but the train terminology is visits so if it instead were to be visits=-1, to me that would mean that it grants a minus 1 visits to counterbalance the visit to reach it. Not sure if that is sensible or not.

lib/engine/tile.rb
jenf: unfortunately you do need the fetch, I've tried it without and it just sets all the visit costs to zero.
Presumably because you have an argument it'll take it even if it's nil.

lib/engine/config/game/g_1882.rb
jenf: visit_cost is what I went with

scottredracecar: even better

PR #842, opened 2020-06-23
Render non-home token reservations

michaeljb:
Also:
- add Erie's token ability
- fix a bug with the static 1846 map page where Erie's home token is rendered

Screenshot from 2020-06-22 21-29-02

Screenshot from 2020-06-22 21-24-46

### DIFF
--- a/assets/app/view/map_page.rb
+++ b/assets/app/view/map_page.rb
@@ -14,7 +14,7 @@ module View
 
       return h(:p, "Bad game title: #{game_title}") unless game_class
 
-      names = %w[p1 p2 p3 p4]
+      names = %w[p1 p2 p3 p4 p5]
       h(Game::Map, game: game_class.new(names))
     end
   end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -551,13 +551,18 @@ module Engine
 
       def init_hexes(companies, corporations)
         blockers = {}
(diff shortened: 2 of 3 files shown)


assets/app/view/map_page.rb
```diff
@@ -14,7 +14,7 @@ def render
 
       return h(:p, "Bad game title: #{game_title}") unless game_class
 
-      names = %w[p1 p2 p3 p4]
+      names = %w[p1 p2 p3 p4 p5]
```
tobymao, line 17: this breaks games that don't support 5 players like 1836jr.

in lib/engine.rb i have a function called player_range

lib/engine/game/base.rb
```diff
@@ -560,6 +559,12 @@ def init_hexes(companies, corporations)
           end
         end
 
+        reservations = Hash.new { |k, v| k[v] = [] }
+        corporations.each do |corp|
+          reservations[corp.coordinates].unshift(corp)
```
tobymao, line 564: why are we unshifting here?

michaeljb: It ensures home tokens are always first in the reservation list for cities like Erie in 1846 with reservations for multiple corps. Not essential though.

michaeljb:
Going to re-do this as a new reservation ability.

PR #852, opened 2020-06-23
Implement 1836jr ports and Paris

jenf:
A train that includes one of these three locations at one end of its run increases the total of its run by F20 per station
token (not City) on the route in the green phase, and F30 per station token (not City) in the brown phase. As with
Off-Map Locations, only one Paris/Port Location may be scored (either at the beginning or the end of the run). Such
a run must include at least two other scoring locations (e.g., a run from Lille to Paris is not legal) and thus requires at
least a 3 train. A train may include both an Off-Map Location and a Paris/Port Location in its run, one at each end.

### DIFF
--- a/lib/engine/config/game/g_1836_jr30.rb
+++ b/lib/engine/config/game/g_1836_jr30.rb
@@ -628,13 +628,13 @@ module Engine
          ]
       },
       "blue":{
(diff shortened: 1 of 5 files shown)


tobymao:
https://github.com/tobymao/18xx/commit/bb418cd9e72b441b32ea49c08faba8f822905f46#diff-b0e5a3507cd857378e055b4609e2a46eR123

PR #868, opened 2020-06-25
Add reservation ability for token reservations.

michaeljb:
1846

    - Give relevant corporations and C&WI the reservation ability.
    - Add Erie's token ability.
    - Close private companies that are removed from the game before deleti
      them. This keeps O&I and MC from blocking tiles if they're not in play.

Abilities

- Add teardown method to token abilities; called when the ability is being
  removed.
- Add `remove` key to abilities to indicate the phase in which that ability
  is removed; check for these at every phase change.
- When a company is closed, remove all of its abilities.

Reservations

- Reservations are the actual corp/company now rather than their short name.
- Reservation ability has tile and city properties that can be updated in order
  to handle tile upgrades.
- If a slot is reserved by a company, it also counts as being reserved by the
  owning corp.
- Reservations are more efficiently initialized, following pattern of blockers:
  build hash of hexes => reservations rather than looping through all the
  corp/company abilities on every hex.

Notes on things that could use refactoring (maybe after 1846 is done):

- `Engine::Corporation` now has `attr_reader :sym`; we want the sym/short name
  for both corporations *and* companies when rendering a reservation, so
  corporations/companies need a common interface for that (called in
  `View::Game::Part::CitySlot#reservation`; currently, corp.name returns the
  short name, company.name returns the long name)
- `corporation#abilities` returns an Array of abilities, but `company#abilities`
  returns `nil` or the ability; a consistent interface there would be
  nice. `lib/engine/game/base.rb` and `lib/engine/hex.rb` use the same logic to
  get to a single ability, or return if there isn't one.

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -39,7 +39,7 @@ module View
           h(
             :text,
             { attrs: { fill: 'black', transform: 'translate(0 9) scale(1.75)' } },
-            @reservation,
+            @reservation.sym,
           )
         end
 
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -39,7 +39,7 @@ module View
           h(
             :text,
             { attrs: { fill: 'black', transform: 'translate(0 9) scale(1.75)' } },
-            @reservation.sym,
+            @reservation.id,
           )
         end
 
--- a/lib/engine/ability/reservation.rb
+++ b/lib/engine/ability/reservation.rb
@@ -8,10 +8,10 @@ module Engine
       attr_accessor :city, :tile
       attr_reader :hex, :slot
 
-      def setup(hex:, city: 0, slot: 0)
+      def setup(hex:, city: nil, slot: nil)
         @hex = hex
-        @city = city
-        @slot = slot
+        @city = city || 0
+        @slot = slot || 0
         @tile = nil
       end
 
--- a/lib/engine/ability/reservation.rb
+++ b/lib/engine/ability/reservation.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class Reservation < Base
+      attr_accessor :city, :tile
+      attr_reader :hex, :slot
+
+      def setup(hex:, city: 0, slot: 0)
+        @hex = hex
+        @city = city
+        @slot = slot
+        @tile = nil
+      end
+
+      def teardown
+        tile.cities[@city].reservations.delete(owner) if tile
+      end
+    end
+  end
+end
--- a/lib/engine/company.rb
+++ b/lib/engine/company.rb
@@ -35,6 +35,8 @@ module Engine
 
     def close!
       @closed = true
+
+      @abilities.each { |t, _| remove_ability(t) }
       return unless owner
 
       owner.companies.delete(self)
--- a/lib/engine/company.rb
+++ b/lib/engine/company.rb
@@ -36,7 +36,7 @@ module Engine
     def close!
       @closed = true
 
-      @abilities.each { |t, _| remove_ability(t) }
+      @abilities.keys.each { |a| remove_ability(a) }
       return unless owner
 
       owner.companies.delete(self)
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -21,11 +21,12 @@ module Engine
     include Spender
 
     attr_accessor :ipoed, :share_price
-    attr_reader :capitalization, :companies, :min_price, :name, :full_name
+    attr_reader :capitalization, :companies, :min_price, :name, :full_name, :sym
     attr_writer :par_price
 
     def initialize(sym:, name:, **opts)
       @name = sym
+      @sym = sym
       @full_name = name
       [
         Share.new(self, president: true, percent: 20),
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -21,12 +21,11 @@ module Engine
     include Spender
 
     attr_accessor :ipoed, :share_price
-    attr_reader :capitalization, :companies, :min_price, :name, :full_name, :sym
+    attr_reader :capitalization, :companies, :min_price, :name, :full_name
     attr_writer :par_price
 
     def initialize(sym:, name:, **opts)
       @name = sym
-      @sym = sym
       @full_name = name
       [
         Share.new(self, president: true, percent: 20),
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -143,7 +143,7 @@ module Engine
     def abilities(type, time = nil)
       abilities = []
 
-      if (ability = super)
+      if (ability = super(type, time, &nil))
         abilities << ability
         yield ability, self if block_given?
       end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -559,13 +559,29 @@ module Engine
 
       def init_hexes(companies, corporations)
         blockers = {}
-
         companies.each do |company|
           company.abilities(:blocks_hexes) do |ability|
             ability.hexes.each do |hex|
               blockers[hex] = company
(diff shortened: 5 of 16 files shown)


lib/engine/ability/reservation.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class Reservation < Base
+      attr_accessor :city, :tile
+      attr_reader :hex, :slot
+
+      def setup(hex:, city: 0, slot: 0)
```
tobymao, line 11: it's better if city and slot are default nil and then city slot are filled on lines 13 14 with ||

lib/engine/ability/reservation.rb
```diff
@@ -0,0 +1,23 @@
...
+        @hex = hex
+        @city = city
+        @slot = slot
+        @tile = nil
+      end
+
+      def teardown
+        tile.cities[@city].reservations.delete(owner) if tile
```
tobymao, line 19: so reservations is an entity?

lib/engine/company.rb
```diff
@@ -35,6 +35,8 @@ def min_bid
 
     def close!
       @closed = true
+
+      @abilities.each { |t, _| remove_ability(t) }
```
tobymao, line 39: this may break meat packing and steam boat because they get closed but their abilities don't go away will have to think

lib/engine/corporation.rb
```diff
@@ -19,11 +19,12 @@ class Corporation
     include Spender
 
     attr_accessor :ipoed, :share_price
-    attr_reader :capitalization, :companies, :min_price, :name, :full_name
+    attr_reader :capitalization, :companies, :min_price, :name, :full_name, :sym
```
tobymao, line 22: you don't need sym, you can just use name

lib/engine/game/base.rb
```diff
@@ -564,6 +563,23 @@ def init_hexes(companies, corporations)
           end
         end
 
+        reservations = Hash.new { |k, v| k[v] = [] }
+        corporations.each do |c|
+          reservations[c.coordinates] << { entity: c,
+                                           city: c.city }
+        end
+        (corporations + companies).each do |c|
+          abilities = c.abilities(:reservation)
```
tobymao, line 572: c.abilities(:reservation) do |ability, _|
  may be cleanear

assets/app/view/game/part/city_slot.rb
```diff
@@ -39,7 +39,7 @@ def reservation
           h(
             :text,
             { attrs: { fill: 'black', transform: 'translate(0 9) scale(1.75)' } },
-            @reservation,
+            @reservation.sym,
```
tobymao, line 42: can just do reservation.name

lib/engine/company.rb
michaeljb: Maybe abilities for companies could default to `"remove": "close"`, but meat packing and steam boat could have `"remove": "4"`, and this loop here could be

```
@abilities.each { |t, _| remove_ability(t) if t.remove == :close }
```

lib/engine/ability/reservation.rb
michaeljb: Yeah. Previously it was just the name. With the entity it's easier to tell if the reservation belongs to a company that is owned by the corporation (e.g., for `reserved_by?`)

assets/app/view/game/part/city_slot.rb
michaeljb: A company's `#name` is the full name:

Screenshot from 2020-06-25 21-33-34

I made an attempt at adding `Company#full_name` and moving `Company#sym` to `Company#name` to be consistent with `Corporation`, but kept running into the out of sync "It's not Player 1's turn" issue. https://github.com/michaeljb/18xx.games/commit/f3f318df2123847d2ef846a91931321f49592cd4 - [sample game JSON](https://gist.github.com/michaeljb/f6548901afa080dbacc67e5e29474123) (this game works on 3bbdbd1a, but doesn't work after merging in the latest master code, still looking into that)

I'm guessing I missed a spot where `#name` should be `#full_name`. Or could there be an issue with the MS/BIG4 minors having the same name as the companies? I'm not really sure what's going on there

lib/engine/ability/reservation.rb
michaeljb: 7abca3c1

assets/app/view/game/part/city_slot.rb
tobymao: ah my bad i thought it was corporation.name not company. company it should be sym

tobymao: i thought @reservation was a corporation and not a company

tobymao: if it's both, you can do id

lib/engine/game/base.rb
michaeljb: I did try that, but `Corporation`'s override of `#abilities` is doing something I don't understand, it seems to be duplicating the abilities.

See where the `puts` block is passed, PRR's reservation gets printed twice while C&WI's just once:

```
> bundle exec rake irb
irb -I lib/ -I assets/app/
2.7.0 :001 > require 'engine'
 => true 
2.7.0 :002 > g = Engine::GAMES_BY_TITLE['1846'].new(['p1', 'p2', 'p3', 'p4', 'p5'])
2.7.0 :003 > prr = g.corporations[0]
2.7.0 :004 > cwi = g.companies[2]
2.7.0 :005 > prr.abilities(:reservation)
 => [#<Engine::Ability::Reservation:0x000055a756561000 @type=:reservation, @owner_type=nil, @when=nil, @count=nil, @used=false, @remove="4", @hex="E11", @city=0, @slot=0, @tile=<Engine::Tile: E11, hex: E11>, @owner=<Engine::Corporation: PRR>>] 
2.7.0 :006 > prr.abilities(:reservation) { |a| puts a }
#<Engine::Ability::Reservation:0x000055a756561000>
#<Engine::Ability::Reservation:0x000055a756561000>
 => [#<Engine::Ability::Reservation:0x000055a756561000 @type=:reservation, @owner_type=nil, @when=nil, @count=nil, @used=false, @remove="4", @hex="E11", @city=0, @slot=0, @tile=<Engine::Tile: E11, hex: E11>, @owner=<Engine::Corporation: PRR>>] 
2.7.0 :007 > 
2.7.0 :008 > 
2.7.0 :009 > cwi.abilities(:reservation)
 => #<Engine::Ability::Reservation:0x000055a756974908 @type=:reservation, @owner_type=nil, @when="sold", @count=nil, @used=false, @remove=nil, @hex="D6", @city=3, @slot=0, @tile=<Engine::Tile: D6, hex: D6>, @owner=<Engine::Company: C&WI>> 
2.7.0 :010 > 
2.7.0 :011 > cwi.abilities(:reservation) { |a| puts a }
#<Engine::Ability::Reservation:0x000055a756974908>
 => #<Engine::Ability::Reservation:0x000055a756974908 @type=:reservation, @owner_type=nil, @when="sold", @count=nil, @used=false, @remove=nil, @hex="D6", @city=3, @slot=0, @tile=<Engine::Tile: D6, hex: D6>, @owner=<Engine::Company: C&WI>> 

```

This output is what I would expect, so I think the problem is in `Corporation#abilities`:

```
2.7.0 :020 > prr.instance_variable_get(:@abilities)[:reservation]
 => #<Engine::Ability::Reservation:0x000055a756561000 @type=:reservation, @owner_type=nil, @when=nil, @count=nil, @used=false, @remove="4", @hex="E11", @city=0, @slot=0, @tile=<Engine::Tile: E11, hex: E11>, @owner=<Engine::Corporation: PRR>> 
```

tobymao: if the corporation owns the ability, then it will print out it's child's abilities

assets/app/view/game/part/city_slot.rb
michaeljb: >  (this game works on 3bbdbd1, but doesn't work after merging in the latest master code, still looking into that)

This fixed the issue I was seeing after merging latest master into this branch - 1c49aa07

Issue #878, opened 2020-06-26
1846 - B&O can't play $40 token if it has less than $80 on charter.


adamranderson:
https://gist.github.com/adamranderson/1210b78e7fa9629cbc6ce9b44020a8ff

PR #879, opened 2020-06-26
Chicago & Western Indianna token teleport power.

adamranderson:
1. Refactored token_price_ability to handle CW&I being an extra token. 
2. This also fixed a bug where city.tokenable? checked token.price (was $80) instead of price returned by token_price_ability (e.g. $40 for B&O). #878


lib/engine/ability/token.rb
```diff
@@ -5,12 +5,13 @@
 module Engine
   module Ability
     class Token < Base
-      attr_reader :hexes, :price, :teleport_price
+      attr_reader :hexes, :price, :teleport_price, :extra_token
 
-      def setup(hexes:, price:, teleport_price: nil)
+      def setup(hexes:, price:, teleport_price: nil, extra_token: false)
```
tobymao, line 10: rename to extra: nil

lib/engine/ability/token.rb
```diff
@@ -5,12 +5,13 @@
...
         @hexes = hexes
         @price = price
         @teleport_price = teleport_price
+        @extra_token = extra_token
```
tobymao, line 14: @extra = extra || false

lib/engine/round/operating.rb
```diff
@@ -657,16 +657,21 @@ def min_token_price(token)
       end
 
       def token_price_ability(token, hex)
-        return [0, :teleport] if @teleported
+        if @teleported
```
tobymao, line 660: this is now a dangerous method 

rename to

def adjust_token_price_ability! 

maybe?

PR #882, opened 2020-06-26
Fix token selector on right hand side of the map

### LINKED ISSUE #849: Tile Selector on Right Hidden
Try laying on L4

https://gist.githubusercontent.com/tobymao/c6c54e662e719ff8953ca7e385ec1878/raw/1e3024b025801bca1493fcdfacdb59730740e506/gistfile1.txt

jenf:
fixes #849

### DIFF
--- a/assets/app/lib/radial_selector.rb
+++ b/assets/app/lib/radial_selector.rb
@@ -25,5 +25,7 @@ module Lib
         'pointer-events' => 'auto',
       }
     end
+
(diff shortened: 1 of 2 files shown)


yzemaze:
The problem isn’t restricted to the right hand side. At least on 1836jr the bottom side is problematic, on other (future) maps left and top will be as well.

jenf:
Left Is already fixed, I can double check how 36jr looks

PR #887, opened 2020-06-26
Companies on cards

yzemaze:
- companies included on corp/player cards
- player owned companies with abilities during OR (e. g. Mitsubishi Ferry) are still rendered below active corp (alternative: render player card)
- toggle description by click/tap on company name
- click/tap to toggle selection (selected = light blue bg)
- prevent selection of not usable or not purchasable companies (should prevent using other corp’s companies from companies tab etc.)
- cursor: pointer only when selectable
- active corp/player highlighted with border and name underline instead of bg color
- (needs: display in part for later)

- still two selects needed when ability allows laying two tiles

I’m not convinced `@table` is the best method to choose which render method to use for corp/player. (… and results in annoying white-space changes. render as switch plus render_tab/page might be better.)
Logic in company ll. 18 – 34 does (should?!) work, but might not be flexible enough for other cases.

### DIFF
--- a/assets/app/view/game/companies.rb
+++ b/assets/app/view/game/companies.rb
@@ -7,22 +7,56 @@ module View
     class Companies < Snabberb::Component
       needs :game
       needs :user, default: nil
+      needs :owner, default: nil
+      needs :table, default: false
 
       def render
-        children = @game
-          .companies
-          .select(&:owner)
-          .group_by(&:owner)
-          .sort_by { |owner, _| owner.name == @user&.dig(:name) ? '' : owner.name }
-          .map { |owner, companies| render_companies(owner, companies) }
+        if @table
+          render_companies_table(@owner)
+        else
+          children = @game
+            .companies
+            .select(&:owner)
+            .group_by(&:owner)
+            .sort_by { |owner, _| owner.name == @user&.dig(:name) ? '' : owner.name }
+            .map { |owner, companies| render_companies(owner, companies) }
 
-        h(:div, children)
+          h(:div, children)
+        end
       end
 
       def render_companies(owner, companies)
         h(:div, { style: { 'margin-bottom': '0.5rem' } }, [
           h(:div, { style: { 'border-bottom': '1px solid gainsboro' } }, owner.name),
           *companies.map { |c| h(Company, company: c) },
+        ])
+      end
+
+      def render_companies_table(owner)
+        companies = owner.companies.map do |c|
+          h(Company, company: c, table: true)
+        end
+
+        table_props = {
+          style: {
+            padding: '0 0.5rem',
+            justifySelf: 'stretch',
+          },
+        }
+        row_props = {
+          style: {
+            grid: owner.player? ? '1fr / 4fr 1fr 1fr' : '1fr / 5fr 1fr',
+            justifySelf: 'stretch',
+          },
+        }
+
+        h('div#company_table', table_props, [
+          h('div.bold', row_props, [
+            h(:div, 'Company'),
+            owner.player? ? h('div.right', 'Value') : '',
+            h('div.right', 'Income'),
+          ]),
+          h(:div, [*companies]),
         ])
       end
     end
--- a/assets/app/view/game/companies.rb
+++ b/assets/app/view/game/companies.rb
@@ -56,7 +56,7 @@ module View
             owner.player? ? h('div.right', 'Value') : '',
             h('div.right', 'Income'),
           ]),
-          h(:div, [*companies]),
+          h(:div, companies),
         ])
       end
     end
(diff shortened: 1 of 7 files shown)


tobymao:
Screen Shot 2020-06-26 at 8 49 41 AM

Is there a border here?

assets/app/view/game/companies.rb
```diff
@@ -25,6 +31,34 @@ def render_companies(owner, companies)
...
+          },
+        }
+
+        h('div#company_table', table_props, [
+          h('div.bold', row_props, [
+            h(:div, 'Company'),
+            owner.player? ? h('div.right', 'Value') : '',
+            h('div.right', 'Income'),
+          ]),
+          h(:div, [*companies]),
```
tobymao, line 59: h(:div, companies)

assets/app/view/game/company.rb
```diff
@@ -8,12 +8,30 @@ class Company < Snabberb::Component
...
       def selected?
         @company == @selected_company
       end
 
+      def purchasable?
+        !@company.owner || @game.round.operating? && @company.owner.player?
+      end
+
+      def ability_usable?
+        return if (@company.all_abilities.map(&:type) & View::Game::Round::Operating::ABILITIES).empty?
```
tobymao, line 23: you can just do Round::Operating::Abilities

assets/app/view/game/company.rb
```diff
@@ -8,12 +8,30 @@ class Company < Snabberb::Component
...
+
+        @game.round.can_act?(@company.owner)
```
tobymao, line 25: i don't think that's right because players can use abilities at any time

assets/app/view/game/company.rb
```diff
@@ -8,12 +8,30 @@ class Company < Snabberb::Component
...
+      end
+
+      def select_company(event)
+        event.JS.stopPropagation
+        selected_company = (ability_usable? || purchasable?) && !selected? ? @company : nil
+        store(:tile_selector, nil)
```
tobymao, line 31: you're doing a double render here, skip: true

assets/app/view/game/round/operating.rb
```diff
@@ -42,10 +42,10 @@ def render
           left = [action]
           corporation = round.current_entity
           left << h(Corporation, corporation: corporation)
-          (corporation.companies + corporation.owner.companies).each do |c|
+          corporation.owner.companies.each do |c|
             next if (c.all_abilities.map(&:type) & ABILITIES).empty?
 
-            left << h(Company, inline: false, company: c)
+            left << h(Company, display: 'block', company: c, game: @game)
```
tobymao, line 48: why block?

assets/app/view/game/companies.rb
yzemaze: > Is there a border here?
Looks like 4px black, but it should be in user-selected font-color (corporation 23) like so …
image
But there’s no need for explicit color anyway, inheritance should be sufficient.

assets/app/view/game/company.rb
yzemaze: upper case. So no need to add full path if parts are shared – of course …

assets/app/view/game/company.rb
yzemaze: ok, I’ll limit it to corps then.

assets/app/view/game/companies.rb
tobymao: do you need to fix something?

assets/app/view/game/round/operating.rb
yzemaze: because inline-block would render it right next to the corp which would move the map below them …

assets/app/view/game/companies.rb
yzemaze: color just removed, please try

tobymao:
i cannot click on mitsubishi ferry to buy it

can you please do some more testing?

{
  "status": "active",
  "actions": [
    {
      "type": "bid",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 1,
      "company": "TR",
      "price": 20
    }
  ],
  "id": "hs_ffhcvhji_1593223946",
  "players": [
    {
      "name": "Player 1"
    },
    {
      "name": "Player 2"
    },
    {
      "name": "Player 3"
    }
  ],
  "title": "1889",
  "description": "",
  "max_players": "3",
  "mode": "hotseat",
  "user": {
    "id": 0,
    "name": "You"
  },
  "created_at": "2020-06-26",
  "loaded": true,
  "result": {}
}

yzemaze:
> i cannot click on mitsubishi ferry to buy it

Order of || is relevant (else I’d need to use more & to prevent running into nils).

PR #889, opened 2020-06-26
Fix paging of hotseat games

### LINKED ISSUE #866: [Home: Hotseat] Next has no effect / no paging
Even with 12+ games there’s no way to go to the next page. Next changed URL param (to hotseat=1), but nothing happens.
Opening https://www.18xx.games/?hotseat=0 (or =1) leads to hotseat games not rendered at all. (Maybe related to #762.)

jenf:
fixes #866

### DIFF
--- a/assets/app/app.rb
+++ b/assets/app/app.rb
@@ -52,7 +52,7 @@ class App < Snabberb::Component
       case @app_route
       when /new_game/
         h(View::CreateGame)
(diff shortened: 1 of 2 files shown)


tobymao:
looks like the tests broke

jenf:
yeah, screwed up the regex.

PR #901, opened 2020-06-27
Allow skip history by round

### LINKED ISSUE #684: Jump by round in history as well as by turn
At the moment, we have two options:

- |< , >| : jump to beginning/end
- < , > : jump by action

Would it be possible to add an intermediate level (<< , >> maybe?), that allows jumping by round? This would allow reviewing a game quickly, but still fine-grained enough to be able to track how the game changed.

jenf:
fixes #684


jenf:
needs a bit more work, will resubmit when ready

PR #902, opened 2020-06-27
Believe Postgres is Requirement - Failing without

rheretic:
Step 1/6 : FROM postgres:12.2
ERROR: Service 'db' failed to build: Get "https://registry-1.docker.io/v2/": dial tcp: lookup registry-1.docker.io on [::1]:53: read udp [::1]:41924->[::1]:53: read: connection refused
make: *** [Makefile:24: dev_up] Error 1

I assume this means that my box needs to have Postgres loaded.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -10,6 +10,7 @@ If configuring the droplet from scratch, these are the requirements:
 * `docker`
 * `docker-compose`
 * `make`
+* `postgres`
 * this repo (via `git clone`)
 
 ### Docker


michaeljb:
No, shouldn't need postgres on your host machine. That error message looks like a failure to download the postgres image layer from the Docker website.

What output do you get from running `docker pull postgres:12.2`? If you get the same error, you might need to setup a Docker account - https://docs.docker.com/docker-id/
If the `docker pull` works, then it was probably just an intermittent network issue.

rheretic:
THANKS michaeljb!  From the cmd line your docker pull command is doing something!

And make is getting further!  

Weird!  (Closing)

PR #908, opened 2020-06-28
slight refactor

benjaminxscott:
avoids executing the removal logic or printing empty 'removal' log lines w/ 5 players

### DIFF
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -135,6 +135,9 @@ module Engine
 
       def remove_from_group!(group, entities)
         removals = group.sort_by { rand }.take([5 - @players.size, 2].min)
+        # This looks verbose, but it works around the fact that we can't modify code which includes rand() w/o breaking existing games
+        return unless removals
+
         @log << "Removing #{removals.join(', ')}"
         entities.reject! do |entity|
           if removals.include?(entity.name)
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -129,14 +129,16 @@ module Engine
       end
 
       def remove_from_group!(group, entities)
-        remove = group.sort_by { rand }.take([5 - @players.size, 2].min)
-        @log << "Removing #{remove.join(', ')}"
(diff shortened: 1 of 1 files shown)


tobymao:
this will probably break every 5 player game because now the random seeds have changed

benjaminxscott:
oh does the digest of the engine file get used as part of the prng seed?

benjaminxscott:
Oh would it regenerate current games based on the new `rand` call?

tobymao:
> Oh would it regenerate current games based on the new `rand` call?

yes, every game is purely made up by it's actions. if the number of rands change, that will change the subsequent random decisions in the game, so i'm inclined to not merge this change although I do appreciate your effort in the PR.

benjaminxscott:
Ah, so if we instead check for `remove.nil?` on the print statement, would that be a non-breaking prng seed change? 

(since it's still calling rand() the same number of times, it's just not printing)

tobymao:
yes, if the goal is to simply clean up the log in a 5 player game.

also you can run lint locally

docker-compose exec rack rubocop

pass -e if you want it to auto correct

benjaminxscott:
Ah yep my commits probably look funny, haven't setup a ruby env on my chromebook 
(also i don't have much xp with ruby)

tobymao:
you just need docker, not sure if that can work on a chromebook

lib/engine/game/g_1846.rb
```diff
@@ -134,10 +134,13 @@ def setup
       end
 
       def remove_from_group!(group, entities)
-        remove = group.sort_by { rand }.take([5 - @players.size, 2].min)
-        @log << "Removing #{remove.join(', ')}"
+        removals = group.sort_by { rand }.take([5 - @players.size, 2].min)
```
tobymao, line 137: this doesn't do what you think it does, sort_by.take returns an array

an empty array is truthy in ruby unlike in python

ruby only false and nil are falsey

you need to do return if removals.empty?

benjaminxscott: Ah, thanks !

Issue #688, opened 2020-06-12
After token is layed show the coordinates of the token


jenf:
After a token is layed it gets greyed out, it'd be ideal if the code would show it's location on the map as it does with the initial token.

PR #913, opened 2020-06-28
Show position for layed token (#688)

perwestling:
When token is placed all tokens will show the
hex name instead of showing token cost for
non-home tokens.

This is the request as given in issue #688 
Pros:
- Easy to see where a token is placed
Cons:
- What did I pay for it? This you have to check the log for.
Might also cause problems in 18xx where a token can be returned, but I believe this problem is similar if showing cost.

This is my first contribution to this open source project so I am interested to know if I have missed anything?
Also, was not able to locate any unit test for this.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -189,7 +189,12 @@ module View
           }
           img_props[:style][:filter] = 'contrast(50%) grayscale(100%)' if token.used
 
-          token_text = i.zero? ? @corporation.coordinates : token.price
+          token_text =
+            if i.zero?
+              @corporation.coordinates
+            else
+              token.city.nil? ? token.price : token.city.hex.name
+            end
 
           h(:div, token_column_props, [
             h(:img, img_props),


assets/app/view/game/corporation.rb
```diff
@@ -189,7 +189,7 @@ def render_tokens
           }
           img_props[:style][:filter] = 'contrast(50%) grayscale(100%)' if token.used
 
-          token_text = i.zero? ? @corporation.coordinates : token.price
+          token_text = i.zero? ? @corporation.coordinates : token.hex_name
```
tobymao, line 192: you can just do token.city.hex.name instead

lib/engine/round/operating.rb
```diff
@@ -580,13 +580,14 @@ def place_token(action)
           price_log = " for #{@game.format_currency(token.price)}"
         end
 
+        token.hex_name = action.city.hex.name
```
tobymao, line 583: delete this line

lib/engine/token.rb
```diff
@@ -3,7 +3,7 @@
 module Engine
   class Token
     attr_reader :corporation, :logo, :type
-    attr_accessor :city, :price, :used
+    attr_accessor :city, :price, :used, :hex_name
```
tobymao, line 6: remove this accessor

tobymao:
thanks for your contribution. there are not individual tests for this particular case, but we test all the games with full scale run throughs. they are located in spec/lib/engine/game/game_spec

they run through every game and check that the money at the end matches up, so usually if you break something, it will be caught in those

assets/app/view/game/corporation.rb
tobymao: if you'd prefer, you could add a 

def hex to the token and call token.hex.name, but up to you

perwestling:
I suppose test this would require some kind of UI test, as presentation details is not something that would be caught by those "end-to-end" tests.

But the main question is still if this is a recommended change?

Anyway, I will fix your comments and push up a new version.

tobymao:
I think it’s fine. Do you feel like it’s an improvement?

It is possible to make a ui test for this. You’d need to get game data that outs the ui in a state that had just placed a token, then check the html contains this text. Look at assets spec. This is non trivial effort though.

perwestling:
It does feel like overkill in this case.
I suppose shoulda or similar could be used to do some simple unit test.
Anyway, I just made the trivial change in this commit, and have now pushed up a new version of it with your comments handled.

PR #915, opened 2020-06-28
Added bank cash to more pages - resolves #262

### LINKED ISSUE #262: Make the bank cash visible on more pages
Some comments/suggestions:
1. I think it would be helpful to include the bank on the spreadsheet view
2. Is it possible to include the phase of the game with current train limit, number of ORs on each screen?

andypymont:


assets/app/view/game/round/stock.rb
```diff
@@ -16,6 +16,7 @@ class Stock < Snabberb::Component
 
         needs :selected_corporation, default: nil, store: true
         needs :last_player, default: nil, store: true
+        needs :show_bank, default: false
```
tobymao, line 19: what's the point of show_bank?

maybe we should make bank a component?

andypymont: Pretty much copied that attribute from the existing stock market tab, but I think you're right that it doesn't look great repeated three times. I'm getting my head around the project structure but can absolutely try and pull it out into its own component which will then be used in those three places.

PR #932, opened 2020-06-30
[1846] implement tunnel blasting

### LINKED ISSUE #894: [1846] Tunnel Blasting Company discount not applied to track lays/upgrades
The B&O bought in the TBC at the end of OR1.1. In OR1.2, it upgrades Wheeling with #14, but is charged the normal rate of $40 ($20 + $20 for the tunnel to Pittsburgh) instead of just $20 (discounted for owning the TBC). Likewise, laying #9 on G17 should cost $40, not $60.

> B&O spends $40 and lays tile #14 with rotation 1 on G19
> B&O spends $60 and lays tile #9 with rotation 1 on G17

[1846-TBC-discount-not-applied.txt](https://github.com/tobymao/18xx/files/4838904/1846-TBC-discount-not-applied.txt)

benjaminxscott:
fixes #894 

Refactors `ignore_terrain` into a generic `tile_discount` ability

I'm not 100% sure if this properly implements a discount for "mountain edges" in 1846 tiles, since I don't have a good grasp of [the code that determines that](https://github.com/tobymao/18xx/blob/master/lib/engine/round/base.rb#L274)

### DIFF
--- a/lib/engine/ability/tile_discount.rb
+++ b/lib/engine/ability/tile_discount.rb
@@ -4,11 +4,11 @@ require_relative 'base'
 
 module Engine
   module Ability
-    class IgnoreTerrain < Base
-      attr_reader :terrain
-
-      def setup(terrain:)
+    class TileDiscount < Base
+      attr_reader :terrain, :discount
+      def setup(terrain:, discount:)
         @terrain = terrain.to_sym
+        @discount = discount
       end
     end
   end
--- a/lib/engine/config/game/g_1846.rb
+++ b/lib/engine/config/game/g_1846.rb
@@ -209,7 +209,15 @@ module Engine
       "value": 60,
       "revenue": 20,
       "desc": "Reduces, for the owning corporation, the cost of laying all mountain tiles and tunnel/pass hexsides by $20.",
-      "sym": "TBC"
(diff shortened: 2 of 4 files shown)


tobymao:
other than that looks good to me

tobymao:
hm i guess it won't work for border cost, you're going to need to apply the discount there as well

benjaminxscott:
Hmm, would you be able to implement that? I honestly have no idea how the border code works

tobymao:
sure

tobymao:
can you fix the issue with the definition?

Apreche:
I think this might need a protection to ensure that total_cost doesn't go below 0.

benjaminxscott:
> can you fix the issue with the definition?
Oh I wasn't sure what this meant? @tobymao

benjaminxscott:
> I think this might need a protection to ensure that total_cost doesn't go below 0.

Makes sense to add something like `max(total_discount, 0)`, as a "negative cost" could happen if there are varying terrain costs and also a private that makes things "cheaper by X until free"

e.g. "40 discount on mountains, some mountains cost 20 to lay"

(For the current implemented games, this patch should work)
@Apreche

PR #935, opened 2020-06-30
Allow skip back and forwards between rounds

### LINKED ISSUE #684: Jump by round in history as well as by turn

jenf:
fixes #684

### DIFF
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -9,6 +9,7 @@ module View
       def self.included(base)
         base.needs :game_data, default: {}, store: true
         base.needs :game, store: true
+        base.needs :fullgame, default: nil, store: true
         base.needs :flash_opts, default: {}, store: true
         base.needs :connection, store: true, default: nil
         base.needs :user, store: true, default: nil
@@ -68,6 +69,7 @@ module View
 
         store(:tile_selector, nil, skip: true)
         store(:game, game)
+        store(:fullgame, game)
       rescue StandardError => e
         store(:game, @game.clone(@game.actions), skip: true)
         store(:flash_opts, e.message)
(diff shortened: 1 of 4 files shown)


assets/app/view/game/actionable.rb
```diff
@@ -68,6 +69,7 @@ def process_action(action)
 
         store(:tile_selector, nil, skip: true)
         store(:game, game)
+        store(:fullgame, game)
```
tobymao, line 72: double render

tobymao:
i'm not a fan of this approach. i think we should instead store outside like round_history, storing 2 game objects seems very heavy weight. either that or in the game_data we could store pointers to round history

something like round -> action .id maybe.

assets/app/view/game/actionable.rb
jenf: ahh, i'd assumed it would swallow the second render

jenf:
Yeah, we can probably just store the round history list, but we'd still have to process the full game if your in history mode.

tobymao:
don't we always process the full game

jenf:
Not if you come in via an URL argument (e.g. http://localhost:9292/hotseat/hs_qnhsajwz_2284?action=449)
But I can cover that case.

tobymao:
why does that change?

jenf:
The game object only loads the first n actions, so you cannot go forward again

tobymao:
that's fine

jenf:
so, just for clarity, store the round history list in a needs?

PR #946, opened 2020-06-30
Move 1846 draft to step mechanism

jenf:
Also fixes anyone can act in the initial auction


lib/engine/round/auction.rb
```diff
@@ -59,8 +59,10 @@ def pass_description
 
       def process_action(action)
         clear_cache!
+
         step = @steps.find { |s| s.actions(action.entity).include?(action.type) }
         raise GameError, "No step found for this action: #{action.to_h}" unless step
+        raise GameError, "It is not #{action.entity.name}'s turn" unless step.active_entities.include?(action.entity)
```
tobymao, line 65: how is this possible? there should be no step available if it is not that user's turn

lib/engine/step/g_1846/draft_distribution.rb
```diff
@@ -0,0 +1,139 @@
...
+
+        def description
+          'Draft Companies'
+        end
+
+        def finished?
+          all_blank? || @companies.empty?
+        end
+
+        def actions(_entity)
```
tobymao, line 56: you should only return actions if it's the user's turn, that way we don't have to check active entities anymore

lib/engine/round/auction.rb
jenf: yep, your quite right, i hadn't understood about the entity parameter

lib/engine/round/g_1846/draft.rb
```diff
@@ -1,125 +1,22 @@
...
 
 module Engine
   module Round
     module G1846
-      class Draft < Base
-        attr_reader :companies, :index
-
-        def initialize(entities, game:, **_kwargs)
+      class Draft < Auction
+        def initialize(game, steps)
```
tobymao, line 9: can you remove the initialize call and just override def select to reorder the players in reverse order?

any changes to state can be done in setup or round_state

lib/engine/round/g_1846/draft.rb
```diff
@@ -1,125 +1,22 @@
...
-            when michigan_southern.full_name
-              michigan_southern.owner = player
-              michigan_southern
-            when big4.full_name
-              big4.owner = player
-              big4
-            end
-
-          @game.bank.spend(company.value, minor) if minor
+        def next_index!
```
tobymao, line 18: maybe woe don't need this

lib/engine/round/g_1846/draft.rb
jenf: ironically that's how i originally coded it, then there was problems of player order after exiting the draft. I'll have another stab at it.

Issue #891, opened 2020-06-26
[1846] Test Case: Use C&WI to teleport to Chicago


wcraigtrader:
> Chicago & Western Indiana
> Reserves a token slot in Chicago (D6), in which the owning corporation may place an extra token at no cost.

The NYC (owned by Barb) should be able to buy in the C&WI (owned by Barb) and then immediately use the free token on the C&WI spot in Chicago, but cannot.

[1846-NYC-should-be-able-to-teleport-to-Chicago-with-CWI.txt](https://github.com/tobymao/18xx/files/4838729/1846-NYC-should-be-able-to-teleport-to-Chicago-with-CWI.txt)

michaeljb:
Fixed by #947

PR #947, opened 2020-07-01
Fix 1846 C&WI edge cases

### LINKED ISSUE #895: [1846] bug - able to token any slot in CHI using C&WI private
[repro, B&O owns CWI and is about to teleport to CHI](https://github.com/tobymao/18xx/files/4839804/cwi_bo.json.txt)

B&O is able to token any open slot in CHI, rather than being limited to the southeast slot

Related: I wasn't able to reproduce #891 , since I was able to teleport to CHI by clicking the city slot

### LINKED ISSUE #897: [1846] C&WI private should provide an additional token action
[repro, B&O owns CWI and is about to teleport to CHI](https://github.com/tobymao/18xx/files/4839804/cwi_bo.json.txt)

Since CW&I provides an additional token + token laying action, B&O should be able to lay a token in Detroit as well

Technically the token is laid when the private is bought in, so a strict rules interpretation would only allow a token to be laid at purchase-time, which would obliviate the need to create an edge case in the existing "token laying step" logic

```
Rules: Upon being purchased, place an additional token of its purchasing corporation at no cost
```

adamranderson:
- C&WI token ability is now used by clicking the company and then city. It uses the special round instead of the operating round.
- Since it uses the special round it count as an extra token lay action. Fixes #897 
- C&WI checks that the city is reserved for it. Fixes #895 
- C&WI ability now only shows up when owned by a corporation
- C&WI ability must be used right when purchased. This matches the reservation and comments made by Tom Lehman on bgg forums.


adamranderson:
It seems I broke a spec test by requiring the player to click C&WI first.

michaeljb:
**tl;dr 2087 is broken at action 293; this branch seems to fix the bug somehow, decreasing one player's available cash**

In game 2087, the IC bought in C&WI but didn't use it, PRR tokened in the C&WI's spot the next turn.

Eventually, the IC did lay a token in Chicago and it was counted as an additional token, so without this code change, it still had one token left, but after this code change, that token IC eventually placed in Chicago came out of IC's normal supply of tokens.

So, on master, IC's pass action 371 is passing on laying a token since it has one left; with this code change, IC is out of tokens and has been autopassed to running routes, so the pass is interpreted as skipping running routes.

After deleting action 371, the spec now fails on action 395--the player hitoaji doesn't have quite enough cash to buy a Grand Trunk share.

At the beginning of Stock Round 4 (action 307), hitoaji buys a share of IC. On master, IC costs $150, but on this branch, IC costs $137. That $13 difference means that hitoaji later cannot afford the share of Grand Trunk.

The price difference was caused at action 293: IC runs, pays out half, and increases to $150. On master, there are additional lines showing it did not run, and drops back to $137.

master:
master screenshot

this branch:
branch screenshot



I haven't figured out how the code changes on this branch could be impacting the dividends payout to fix that issue, but I am confident now that 2087 is broken. There are a bunch of undo/redo actions around there, maybe that has something to do with it? 
 But I still don't see anything on this branch that would relate to that either :thinking:

tobymao:
when something like this happens, usually the game is out of sync. meaning master is out of sync

michaeljb:
@tobymao so what needs to be done to get this merged? I tried adding a few more completed games of 1846 to replace the broken 2087, but I couldn't get any of them to pass. Some had the error about re-using a train that has operated (like in #967) and some referred to an action being an invalid step. Would it be ok to just delete 2087?

michaeljb:
This PR fixes #891

tobymao:
well, we should try to fix the build first

michaeljb:
Well yeah, that's what [my first comment](https://github.com/tobymao/18xx/pull/947#issuecomment-652776304) was about.

2807 is broken beyond all repair, the extra "IC does not run" and price drop happening at https://www.18xx.games/game/2807?action=293 enabled hitoaji to buy an IC share for $13 less than it should have cost in SR4 (https://www.18xx.games/game/2807?action=307), then in SR5 hitoaji buys a GT share and has less than $13 remaining (https://www.18xx.games/game/2807?action=395), ergo with that extra "IC does not run" removed, the game breaks when it reaches the action where hitoaji buys the GT share.

I don't know what "master is out of sync" means or how to fix it, and I also don't understand how this PR gets rid of the extra "IC does not run" unless there was some weird trickle down effect from IC buying in C&WI at action 120 and PRR tokening the slot previously reserved for C&WI at 128.

tobymao:
ok, then let's just delete that spec?

adamranderson:
Thanks Michael for fixing the test. I have merged in the changes and this is ready for another look.

lib/engine/action/place_token.rb
```diff
@@ -11,7 +11,7 @@ def initialize(entity, city:, slot:, token_type: nil)
         @entity = entity
         @city = city
         @slot = slot
-        @token = @entity.find_token_by_type(token_type)
+        @token = @entity.find_token_by_type(token_type) if defined? @entity.find_token_by_type
```
tobymao, line 14: i don't see how this code can work. if @token is nil, you're going to hit a null pointer exception. 

i suggest you have company implement find_token_by_type, and given the token type it should return it's corresponding token

lib/engine/company.rb
```diff
@@ -59,6 +59,10 @@ def corporation?
       false
     end
 
+    def find_token_by_type(token_type)
+      return Token.new(@owner)
```
tobymao, line 63: remove the return

lib/engine/company.rb
```diff
@@ -59,6 +59,10 @@ def corporation?
       false
     end
 
+    def find_token_by_type(token_type)
```
tobymao, line 62: _token_type

tobymao, line 62: should theis instead check for the company ability for token? you don't want every company creating tokens right?

lib/engine/action/place_token.rb
adamranderson: It worked since the token in the in this action wasn't being used but created inline. This was more hacked together tho so I took your suggestion.

lib/engine/company.rb
adamranderson: This was why I went with the first approach to be creating the token in 1 spot for the ability code. I'll add another check here and raise a game error if not.

Issue #650, opened 2020-06-07
Eliminate unneeded train buying wording


ChrisTheCat:
From a game log on OR1.

> LV must buy a train either from The Depot, The Discard or other corporations.

There are no trains in "The Discard" and buying across isn't allowed, as a 3 train hasn't been purchased. I suggest that in this case, the display should only be "LV must buy a train from The Depot." In other scenarios, it might be appropriate to say "from The Depot or other corporations" for example.

scottredracecar:
You can buy across anytime in 18Chesapeake.

ChrisTheCat:
That's interesting. Suggestion still stands - don't mention locations that aren't possible to buy a train

perwestling:
@ChrisTheCat I have made a commit that change so that the text would be
  LV must buy an available train

What do you think about that?

ChrisTheCat:
I like it. Good solution.

PR #952, opened 2020-07-01
Improve must-buy-train wording (#650)

perwestling:
Vary text depending if there is available trains in the
discard and/or allowed to buy from other corporations.

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -66,11 +66,7 @@ module View
         children << h(:div, [h(UndoAndPass, pass: !must_buy_train)])
 
         if must_buy_train
-
-          children << h('div.margined',
-                        "#{@corporation.name} must buy a train either from The Depot, The Discard"\
-                        "#{other_corp_trains.any? ? ' or other corporations' : ''}")
-
+          children << h('div.margined', "#{@corporation.name} must buy an available train")
           children += render_president_contributions if @corporation.cash < @depot.min_depot_price
         end
 


assets/app/view/game/buy_trains.rb
```diff
@@ -68,8 +68,11 @@ def render
         if must_buy_train
 
           children << h('div.margined',
-                        "#{@corporation.name} must buy a train either from The Depot, The Discard"\
-                        "#{other_corp_trains.any? ? ' or other corporations' : ''}")
+                        "#{@corporation.name} must buy a train"\
+                        "#{round.can_buy_train? || !@depot.discarded.empty? ? ' either' : ''}" \
```
tobymao, line 72: should it be round.can_buy_train? && @depot.discarded.any?

tobymao: ah i see, because you can always buy from the depot, you're checking if you can buy trains from others or the discard to choose either, 

let's make it round.can_buy_train? || @depot.discarrded.any?

perwestling: @tobymao: yes, the idea was to get the test "XXX must buy a train from The Depot" if it is the only place to buy a train, and to get "XX must buy a train either from The Depot...." if there is more places.
Now I am not native English speaking so I do not know if it would work to remove "either" completly?
Also, would it be possible to have a situation there there a no trains available in The Depot? I know there are variants where the trains can run out, so it might not be 100%.

Another solution to this might be to change the text to 
  **XXX must buy an available train.**

tobymao: yes, in 1836, there was a game when all the diesels were bought.

the either is not necessary,

X must by a train from the depot, the discord, or other corporations is valid

XXX must buy an available train is a good solution and the simplest.

PR #954, opened 2020-07-01
Hide JSON by default, remove pretty print

### LINKED ISSUE #888: Collapse the JSON representation for the game data in Tools tab
When a user goes to the Tools tab, the game data is display in JSON form. For a new game, this is a fairly short list, but later in the game, the rendered JSON can be quite long. Also, with some games, some of the game data should be private information, at least initially (yes, I'm thinking of the 1846 private draft).

I think it would be better when a user goes to the Tools tab, the JSON game data is not displayed until the user clicks a button. This would speed up page rendering and avoid inadvertently revealing private game information.

jenf:
It's less confusing to non-engineers if we don't show a bunch of JSON data by default.

fixes #888

### DIFF
--- a/assets/app/view/game/game_data.rb
+++ b/assets/app/view/game/game_data.rb
@@ -8,11 +8,12 @@ module View
       include GameManager
 
       needs :allow_clone, default: true
+      needs :show_json, default: false, store: true
       needs :actions
 
       def render
         @game_data = @game_data.merge(actions: @actions)
-        @json = `JSON.stringify(#{@game_data.to_n}, null, 2)`
+        @json = `JSON.stringify(#{@game_data.to_n}, null)`
 
         props = {
           style: {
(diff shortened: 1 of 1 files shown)


assets/app/view/game/game_data.rb
```diff
@@ -8,22 +8,22 @@ class GameData < Snabberb::Component
       include GameManager
 
       needs :allow_clone, default: true
+      needs :show_json, default: false, store: true
       needs :actions
 
       def render
         @game_data = @game_data.merge(actions: @actions)
-        @json = `JSON.stringify(#{@game_data.to_n}, null, 2)`
+        @json = `JSON.stringify(#{@game_data.to_n}, null)`
```
tobymao, line 16: i  think we should keep the pretty print

PR #965, opened 2020-07-02
change the $0 space in 1846 to black, update legend

### LINKED ISSUE #945: [1846] Market page erroneously claims that you can own more than 60% of companies @ $0 price
The 1846 market legend has an amber color that claims `Company shares can be held above 60%` -- that's not a feature of 1846. The market uses amber for the `$0` level, but any company whose market price drops to $0 is automatically closed.

I would recommend changing the legend to `Company closes` for the amber ($0) levels.

michaeljb:
* consistent wording in stock market legend: "Company" -> "Corporation"

[Fixes #945]

Screenshot from 2020-07-01 23-42-29

Screenshot from 2020-07-01 23-44-19

### DIFF
--- a/lib/engine/config/game/g_1846.rb
+++ b/lib/engine/config/game/g_1846.rb
@@ -116,7 +116,7 @@ module Engine
   },
   "market": [
     [
-      "Closed",
+      "Closedblk",
       "10",
       "20",
       "30",
--- a/lib/engine/config/game/g_1846.rb
+++ b/lib/engine/config/game/g_1846.rb
@@ -116,7 +116,7 @@ module Engine
   },
   "market": [
     [
-      "Closedblk",
+      "0blk",
       "10",
       "20",
       "30",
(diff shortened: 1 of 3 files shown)


lib/engine/config/game/g_1846.rb
```diff
@@ -116,7 +116,7 @@ module G1846
   },
   "market": [
     [
-      "Closed",
+      "Closedblk",
```
tobymao, line 119: hm what the heck is this? i guess i never looked at it, but it should be "0blk" or we can do "0l" or since we're naming everything now "0black" and everything else should be "0brown" etc

michaeljb: Changed to 0 in 9dc0c943 -- that makes more sense anyway since "0" is what's displayed in the actual market cell. Used "blk" for "black" since "b" was taken for "brown". Kinda made me laugh when I realized the "o" in "Closed" was why "orange" was used.

PR #969, opened 2020-07-02
Fix (game) cards’ margin on mobile

yzemaze:
found another culprit, maybe the last one …
Please check if 96vw works on iPhone now.


tobymao:
home page still wiggles

Issue #209, opened 2020-04-29
Allow programmed moved for stock rounds/merger rounds


jenf:
rr18xx has a nice stock round automation which would be useful to improve upon especially when playing async games. It'd be nice to be able to fill in your request at any time during the stock round.

I think a lot of the different conditions can be merged into a 'Stop' condition whereby it'll stop if it thinks that the player might want to consider their move more. This would be better than getting users to choose their stop conditions as it may make them regret their decision.

- Buy to Float
- Buy to Float with Stop
- Buy n from Bank with Stop
- Buy n from IPO with Stop
- Pass once
- Pass until OR with Stop

AutoStop for both pass and buy stops when
- Someone sells something in a company i'm invested in
- I can now buy a share, I couldn't previously
- Someone is close to buying a company out from under me.

Buy would also stop when:
- Any company has changed price (I may value company B over company A)
- A new company has floated
- Someone has bought shares in the company I am buying
- 
Implementation plan for Programmed moves
### 'Auto' actions
- [x] Underlying support for adding extra actions

### Underlying support
- [x] Add actions to add programmed move to be added/removed/modified
- [x] Move logging to individual player view (why program deactivated)
- [x] Determine how best to de-arm the programmed moves when moving to next round

### Future
- [x] UI work
- [x] Buy to float (only on first round)
- [x] Merger/float auto-pass on convert, merge, for sale for 1817/1867/1861 etc
- [x] Disable Merger option in stock round
- [x] Add asset tests
- [x] Rename to 'Auto' rather than Async
- [x] Add fixture tests for auto actions
- [x] Make it clearer that Programmed actions are active for a player
- [x] Buy till float unless shenanigan
- [x] Pass unless shenanigan
- [x] SR for 1822
- [x] SR for 1817/1861/1867
- [ ] Buy till custom

jcoveney:
So I've thought about this a bit. I agree that there are reasonable defaults... I feel like the dream would be a little sublanguage that lets you query basic game state and build a "should I stop or not" based off of that. Then we can do whatever we want. That said, not sure if toby is interested in maintaining something like that :) And it may be overkill when there are probably some sane defaults that cover most use cases (eg what is in the OP!)

benjaminxscott:
FWIW rr18xx [has something like this](http://www.rr18xx.com/cgi-bin/cmd_auto_order.cgi?gname=JJJS_1) as reference for the options

[...]

PR #971, opened 2020-07-02
Add a notepad

### LINKED ISSUE #297: Add private notepad
private notepad/text field if possible, that is not seen by other players in game. Users could record own notes/short term strategy might help remember what we intend to do when game comes back to our turn

jenf:
fixes #297

While this could have been done as local storage, I often switch devices and so losing my notes would be annoying.

(This is partially to implement user storage for #209)


assets/app/view/game/notepad.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        h(:div, [
+          notepad,
+          render_button("Save Notepad#{saved ? '' : ' (Unsaved)'}") { submit },
+          ])
+      end
+
+      def local_save
+        # non-persistently save the notepad in case the user switches tabs
+        store(:user_notes, params['notepad'])
```
tobymao, line 46: hm this will cause a full rerender every time you type a key stroke?

routes/game.rb
```diff
@@ -36,6 +40,14 @@ class Api
             game.to_h
           end
 
+          # POST '/api/game/<game_id>/user_settings'
+          r.is 'user_settings' do
+            DB.with_advisory_lock(:action_lock, game.id) do
```
tobymao, line 45: i don't think this lock is necessary

routes/game.rb
```diff
@@ -11,7 +11,11 @@ class Api
 
         # '/api/game/<game_id>/'
         r.is do
-          game.to_h(include_actions: true)
+          game_data = game.to_h(include_actions: true)
```
tobymao, line 14: instead could we make this an option on the to_h, so to_h takes in a param who it's for

assets/app/view/game/notepad.rb
jenf: On Firefox it's only when you exit the textbox, I'll test with chrome

tobymao:
oh ok cool then

routes/game.rb
jenf: If two users try to save the notepad at the same time, won't one user lose their content?

tobymao:
@jenf still waiting on the changes to to_h

jenf:
yeah, i'll pick this up in a few days. Trying to progress 1882 more

assets/app/view/game/notepad.rb
jenf: Same on Edge and Chrome.

routes/game.rb
jenf: @tobymao do you want me to remove the lock? I think it might have race conditions if we remove it.

tobymao:
can keep it

PR #974, opened 2020-07-03
entities view: consolidation players/corps/companies

yzemaze:
- players and corporations merged into entities view, sorted by player order [logged in player first, bank last], corps below their pres sorted by Sym A – Z
- minors (1846: Big4, MS) will turn up on players’ cards as well as separate as corps
- number (1 – n) and width (17 – 20rem) of columns (with player/corp cards) depend on window width
- finetuning of player/corp cards
- game nav/menu refactored (semantics!)
- menu entries for corps/players/comps removed (route_anchor cases still in place)
- companies’ `layout: 'list'` isn’t used in this setup. Removal _if_ we definitely don’t implement dedicated pages with sorting options.


assets/app/view/game/company.rb
```diff
@@ -9,7 +9,7 @@ class Company < Snabberb::Component
       needs :game, store: true
       needs :tile_selector, default: nil, store: true
       needs :display, default: 'inline-block'
-      needs :table, default: false
+      needs :layout, default: ''
```
tobymao, line 12: default nil is fine

assets/app/view/game/company.rb
```diff
@@ -45,7 +45,7 @@ def render_bidders
       end
 
       def render
-        if @table
+        if @layout == 'table'
```
tobymao, line 48: in general let's use symbols even though it doesn't matetr

assets/app/view/game/companies.rb
```diff
@@ -8,11 +8,14 @@ class Companies < Snabberb::Component
       needs :game
       needs :user, default: nil
       needs :owner, default: nil
-      needs :table, default: false
+      needs :layout, default: 'list'
 
       def render
-        if @table
+        case @layout
+        when 'table'
```
tobymao, line 15: change to symbols

assets/app/view/game/entities.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        players = @game.players
+        if (i = players.map(&:name).rindex(@user&.dig(:name)))
+          players = players.rotate(i)
+        end
+
+        player_owned, bank_owned = (@game.corporations + @game.minors).sort_by(&:name).partition(&:owner)
+
+        children = players.map do |p|
+          corps = player_owned.select { |c| c.owner == p }.map { |c| h(Corporation, corporation: c) }
```
tobymao, line 28: can we avoid the double for loop here, do a player_owned = player_owned.group_by(&:owner)

assets/app/view/game_page.rb
```diff
@@ -81,6 +81,8 @@ def render
           h(Game::Companies, game: @game, user: @user)
         when 'corporations'
```
tobymao, line 82: get rid of the other tabs

yzemaze: Tabs are long gone, these are just the route anchors. I’d leave them in place as I expect the need for them later when people clamour for sorting options … but we could of course revert then.
If you want the code to be pruned, just say so. There are more unnecessary hunks due to entities. (Maybe transitional, maybe for good – depends on sorting yes/no.)

tobymao:
delete all unused code, can always bring it back, but most likely it will
just sit there unused if left

assets/app/view/game/entities.rb
yzemaze: I’ll need to traverse players (not everyone "owns" corps!) and then their corps for print. group_by(&:owner) would get rid of the select, but I’d still need map.

tobymao: well yes, but you don't need to do players.map |...| player_owned.select

yzemaze: Well, I don’t know if …
`        player_owned = player_owned.group_by(&:owner)

        children = players.map do |p|
          corps = player_owned[p]&.map { |c| h(Corporation, corporation: c) }`
performs any better than select, but if that’s what you’d like. Is group_by essentially a DB-op?

tobymao:
no, there's no db interaction in the client

group by takes an iterable and turns into into a hash

group_by: List[X] => Map(Y, X)

spec/assets_spec.rb
```diff
@@ -120,13 +120,13 @@ def render(**needs)
       }
 
       expect(render(app_route: '/game/1', **needs)).to include('Takamatsu E-Railroad')
-      expect(render(app_route: '/game/1#players', **needs)).to include('Player 1')
-      expect(render(app_route: '/game/1#corporations', **needs)).to include('Awa Railroad')
+      expect(render(app_route: '/game/1#entities', **needs)).to include('Entities')
+      expect(render(app_route: '/game/1#entities', **needs)).to include('Player 1')
```
tobymao, line 124: there's no need to call this 3x

tobymao:
this wiggles on chrome and firefox. to recreate just shrink the browser

Screen Shot 2020-07-04 at 2 07 01 PM

yzemaze:
> this wiggles on chrome and firefox. to recreate just shrink the browser
I’m using responsive design mode(s) and my phone. I can’t reproduce wiggling with FF or Chrome (same on browserstack for Safari/Chrome). Additionally there are no margins on the right side on any of these cards. (… which were the reason for wiggling in all cases up to now.)
btw: FF + Chrome on iOS do use the WebKit (Safari) engine: https://en.wikipedia.org/wiki/Firefox_for_iOS

PR #985, opened 2020-07-04
[WIP] Autorouting

NickHeiner:
This is very preliminary and there's a ton wrong with it.

### DIFF
--- a/assets/app/view/game/autoroute.rb
+++ b/assets/app/view/game/autoroute.rb
@@ -3,7 +3,11 @@ module View
     class Autoroute
       def self.calculate(game, corporationId)
         corp = game.corporation_by_id(corporationId)
-        corp.trains
+        trains = corp.trains
+
+        if trains.empty?
+          return []
+        end
       end
     end
   end
--- a/assets/app/view/game/autoroute.rb
+++ b/assets/app/view/game/autoroute.rb
@@ -1,0 +1,9 @@
+module View
+  module Game
+    class Autoroute
+      def self.calculate(game, company)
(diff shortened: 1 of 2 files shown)


assets/app/view/game/autoroute.rb
```diff
@@ -0,0 +1,14 @@
+module View
```
tobymao, line 1: the calculation should be in the lib/engine, but only called in the views

assets/app/view/game/autoroute.rb
```diff
@@ -0,0 +1,14 @@
...
+  module Game
+    class Autoroute
+      def self.calculate(game, corporationId)
```
tobymao, line 4: always use snake_case

assets/app/view/game/autoroute.rb
```diff
@@ -0,0 +1,14 @@
...
+        corp = game.corporation_by_id(corporationId)
+        trains = corp.trains
+
+        if trains.empty?
+          return []
```
tobymao, line 9: in ruby, the last line is always return, so you don't need to add a return

assets/app/view/game/autoroute.rb
NickHeiner: Heh! Of course. I haven't run rubocop on this but obviously I will before submitting this for real.

assets/app/view/game/autoroute.rb
NickHeiner: I added `return` here because I'm going to add more after it.

PR #990, opened 2020-07-05
Limit LSL's ability to Phase 2

### LINKED ISSUE #906: [1846] LSL should only be usable in Phase 2
From page 12 of GMT rules:

> **Lake Shore Line:** This company’s ability may not be used until green tiles are available.

Sample game showing the ability being used in phase 1 - https://gist.github.com/michaeljb/965a96690fe66c59a3bc2267a04a9568

michaeljb:
Prevent laying tiles of colors that are not allowed by the current phase.

[Fixes #906]

### DIFF
--- a/lib/engine/abilities.rb
+++ b/lib/engine/abilities.rb
@@ -35,6 +35,9 @@ module Engine
           true
         end
 
+      phase = @game&.phase&.current && @game.phase.current['name']
+      return nil if phase && ability.phases && !ability.phases.include?(phase)
+
       return nil if time && ability.when != time.to_s
       return nil unless correct_owner_type
 
--- a/lib/engine/abilities.rb
+++ b/lib/engine/abilities.rb
@@ -35,9 +35,6 @@ module Engine
           true
         end
 
-      phase = @game&.phase&.current && @game.phase.current['name']
-      return nil if phase && ability.phases && !ability.phases.include?(phase)
-
       return nil if time && ability.when != time.to_s
       return nil unless correct_owner_type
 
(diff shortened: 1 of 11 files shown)


tobymao:
hm i don't know about this. maybe we should just remove lsl's ability in phase 3

tobymao:
i don't like the ability of giving a game to everything as well

michaeljb:
> hm i don't know about this. maybe we should just remove lsl's ability in phase 3

The issue isn't the LSL doing anything in phase 3, it's that the LSL can be used in phase 1.

tobymao:
can we just check that the tile upgrade needs to respect the upgrade rules?

michaeljb:
> i don't like the ability of giving a game to everything as well

hmmm maybe abilities could have an ability that lists which phase it should become active, then `Engine::Phase#trigger_events!` could enable those abilities?

michaeljb:
> can we just check that the tile upgrade needs to respect the upgrade rules?

It already does. This sample game shows NYC laying a yellow on Toledo with its normal tile lay, and then using LSL to do the normal valid upgrade to green. The only invalid part is that it's Phase 1 but the LSL's ability shouldn't be available until Phase 2 - https://gist.github.com/michaeljb/965a96690fe66c59a3bc2267a04a9568

tobymao:
but laying a green is not legal in phase 1, if we deny laying greens there that should fix it

michaeljb:
> but laying a green is not legal in phase 1, if we deny laying greens there that should fix it

https://github.com/tobymao/18xx/pull/990/commits/dc5a5364d078afc82dce9a7044e16eb2ad5e08e3 now raises an error banner if the wrong color is laid. It does break a couple `connected_hexes` specs

tobymao:
cool much simpler

PR #991, opened 2020-07-05
Assign steamboats before the private company is bought in

### LINKED ISSUE #851: [1846] [rules] bonus of the Steamboats private can be assigned before it's bought in
Edit: fixed an inaccuracy in my writeup

The player owner of Steamboats may choose to assign / shift its power to any company , during the private income step

(This means the assigned company may receive the bonus before Steamboats is bought in)

In practice this means the owner of BO can get the steamboats bonus without having to buy in the private, since BO is quite expensive to start

This is likely a one-off edge case for privates, as a counterexample note that Meat Packing must be bought in for a company to receive its bonus

Perhaps have a button + dropdown for "Assign Power" after clicking on the private company (similar to the 'buy' dialogue)

adamranderson:
How to use: Click Steamboat Company then click the corporation. 

- Removes previous assignment when clicked again
- Removes previous assignment when purchased

image

Fixes #851

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -9,22 +9,24 @@ module View
       include Actionable
       include Lib::Color
       needs :corporation
-      needs :round, default: nil, store: true
+      needs :selected_company, default: nil, store: true
       needs :selected_corporation, default: nil, store: true
       needs :game, store: true
       needs :display, default: 'inline-block'
 
       def render
         select_corporation = lambda do
-          if @round&.can_assign_corporation?
-            puts "Assigning corporation"
-            process_action(Engine::Action::Assign.new(@round.current_entity, target: @corporation))
-            #store(:selected_corporation, nil)
-            puts "Done assigning corporation"
-          else
-            selected_corporation = selected? ? nil : @corporation
-            store(:selected_corporation, selected_corporation)
-          end
+          selected_corporation = selected? ? nil : @corporation
+          store(:selected_corporation, selected_corporation)
+
+          puts "#{@round.inspect}"
+
+
+          if selected_corporation && @selected_company && @game.special.can_assign_corporation?
+            process_action(Engine::Action::Assign.new(@selected_company, target: @corporation))
+            store(:selected_corporation, nil)
+            store(:selected_company, nil)
+          end       
         end
 
         card_style = {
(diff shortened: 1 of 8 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 
+require 'view/game/actionable'
 require 'view/game/companies'
 
 module View
   module Game
     class Corporation < Snabberb::Component
+      include Actionable
```
tobymao, line 9: by including actionable you no longer need the need for game

lib/engine/round/operating.rb
```diff
@@ -540,19 +540,29 @@ def buy_company(company, price)
           raise GameError, "Price must be between #{@game.format_currency(min)} and #{@game.format_currency(max)}"
         end
 
+        log_later = []
         company.owner = @current_entity
         entity.companies.delete(company)
 
         company.abilities(:assign_corporation) do |ability|
+          Assignable.remove_from_all!(@game.corporations, company.sym) do |unassigned|
```
tobymao, line 548: let's do company.id

lib/engine/assignable.rb
```diff
@@ -17,5 +17,14 @@ def assign!(key, value = true)
     def remove_assignment!(key)
       assignments.delete(key)
     end
+
+    def self.remove_from_all!(assignables, key)
```
tobymao, line 21: should we just remove from the previous as opposed to doing everything?

lib/engine/round/special.rb
```diff
@@ -81,10 +90,21 @@ def _process_action(action)
...
-          company.abilities(:assign_hexes, &:use!)
-          @game.log << "#{company.name} activates #{hex.name}"
+          target = action.target
+          raise GameError, "#{company.name} is already assigned to #{target.name}" if target.assigned?(company.id)
+
+          if target.is_a?(Hex) && company.abilities(:assign_hexes)
+            target.assign!(company.id)
+            company.abilities(:assign_hexes, &:use!)
+            @game.log << "#{company.name} is assigned to #{target.name}"
+          end
```
tobymao, line 100: elsif

lib/engine/assignable.rb
adamranderson: Looking at previous could be problematic. Since order could be:

1. Assign corp
2. Assign hex
3. Re-assign corp.

The previous in #3 would be the hex but we want to leave that and remove the corp.

PR #993, opened 2020-07-05
change font to Lato; improve legibility of numbers on map

yzemaze:
Except for the logo margins (around .) and bold on game menu hover Lato worked out of the box for text, too.
Logo: Keep old one / use Lato, too?

### DIFF
--- a/public/assets/main.css
+++ b/public/assets/main.css
@@ -1,11 +1,17 @@
-* { font-family: 'Inconsolata', monospace; }
+* { font-family: Lato, sans; }
 
 a {
   color: currentColor;
   text-underline-offset: 0.2em;
 }
+nav a {
+  text-decoration: none;
+}
+nav a.active {
+  text-decoration: underline;
+}
 nav a:focus, nav a:hover {
-  font-weight: bold;
+  text-decoration: underline dotted;
   outline: 0;
 }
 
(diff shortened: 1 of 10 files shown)


public/assets/main.css
```diff
@@ -1,11 +1,20 @@
...
+}
 
 a {
   color: currentColor;
   text-underline-offset: 0.2em;
 }
+nav a {
+  text-decoration: none;
+}
+nav a.active {
```
tobymao, line 13: does this need to be a class? if it's not does the focus not work?

tobymao:
everything lato

tobymao:
Screen Shot 2020-07-05 at 8 16 55 AM

3 digit revenue too big

public/assets/main.css
yzemaze: No, it was just the/my usual way. `route_anchor == anchor[1..-1] ? a_props[:style][:textDecoration] = 'underline' : ''` would work, too.

tobymao: ok let’s revert it

yzemaze:
> everything lato

?
Is it not?

> 3 digit revenue too big

I know. Do you
a) want it to fit the small circle
b) the circle to expand
or c) neglect this edge case?

tobymao:
ideally fit in the circle

PR #988, opened 2020-07-05
game_info: provide obsoletes and variant names and prices

jaysachs:
Addresses #798

Issue #798, opened 2020-06-20
[1846] Information improvement in "info"


Spobster:
It would be nice to see the N/M trains also in the info tab, including their prices. Also an extra column depicting "phasing out" phase in addition to the "closing" phase would be useful.

PR #997, opened 2020-07-05
Provide variants info on remaining trains when buying trains

jaysachs:
Similar to #988 / #798

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -228,11 +228,9 @@ module View
         }
 
         rows = @depot.upcoming.group_by(&:name).flat_map do |_, trains|
-          train = trains.first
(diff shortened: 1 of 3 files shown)


assets/app/view/game/buy_trains.rb
```diff
@@ -227,9 +227,12 @@ def remaining_trains
           },
         }
 
-        rows = @depot.upcoming.group_by(&:name).flat_map do |name, trains|
-          [h(:div, name),
-           h(:div, @game.format_currency(trains.first.price)),
+        rows = @depot.upcoming.group_by(&:name).flat_map do |_, trains|
+          train = trains.first
+          names =  train.variants.values.map { |var| var[:name] }
```
tobymao, line 232: can you create a helper function in train

def names_and_prices
[names, prices]
end

and refactor these two places where you do this pattern?

PR #1008, opened 2020-07-06
info text only when not logged in

yzemaze:
no mode check to prevent content jumping


assets/app/view/form.rb
```diff
@@ -26,7 +26,7 @@ def params
       end.to_h
     end
 
-    def render_form(name, inputs)
+    def render_form(name, inputs, msg = nil)
```
tobymao, line 29: msg is not really a good name for what this is

Issue #961, opened 2020-07-02
Special rules for Mark Derrick Games, 18AL


scottredracecar:
(this is a work in progress)

**All**

- [x] Private buy in is 1/2 to 1-1/2x face
- [x] All except 18MS, one train per turn until Phase 4

**18AL**

- [x] Towns don’t count against visits
- [x] Implement private companies

**18AL Tiles**

- [x] yellow 441 elyton/birmingham 10 lolipop
- [x] yellow 445 20 lumber terminal gentle curve town
- [x] yellow 451 atlanta triple lolipop 30
- [x] green 442 birmingham bird foot 2 slots 30
- [x] green 443 montgomery 5-way 2 slots 40
- [x] brown 444 birmingham montgomery 5-way, 2 slots 50
- [x] gray 446 6-way 3city 70 birmingham

See the list on https://github.com/perwestling/18xx/wiki/18AL

scottredracecar:
If someone wants to dive into one of these games, please have at it! Otherwise, I plan to start fleshing out some of these special rules mid-July and will resume work.

Some of these games have map issues that require different tile logic to display right.

Most of these games have special city-specific tile upgrades which aren't defined yet. I listed the ones for GA and AL.

perwestling:
I will to start look at 18AL. That, together with 18MEX, are the only Mark Derrick games I have played.
It will be my first dive into any game at 18xx.games.
After 18AL I intend to start on 18SJ which is a Swedish prototype.

perwestling:
@scottredracecar: I looked through the tiles for 18AL in your list above, and checked what was present in the 18xx.game right now.

**Does exists in the game and look OK**
- 441 - as 441a
- 442 - as 442a
- 443 - as 443a
- 444 - as 444a
- 446

**In 18xx.games but incorrect**
- 445 - missing name and wrong revenue

**Missing in 18AL**
- 451 - not in my game. Atlanta is an off-board hex. And the one on the map matches my game.

So is only correction of 445 needed? The rest done?

I have also two other querries:
1. Coal markers, should they appear on the map?
2. Destinations, should they appear on the map? And/or the company charters?

Otherwise map looks good. Nashville and Mobile hexes can be slightly improved.
Does the game support non-pass-through off-board hexes like Nashville and Mobile?

If I do make any commit for 18AL, should I just put it up as a pull request on master branch, and notify you about it?

perwestling:
[...]

PR #1010, opened 2020-07-06
Improvement for 18AL, 18GA, 18AL (#961)

perwestling:
- Implement "max 1 train from depot before phase 4" for all three
- Correct operation round count for 18AL and 18TN

### DIFF
--- a/lib/engine/config/game/g_18_al.rb
+++ b/lib/engine/config/game/g_18_al.rb
@@ -467,67 +467,81 @@ module Engine
    "phases":[
       {
          "name":"2",
+         "on":"2",
          "train_limit":4,
          "tiles":[
             "yellow"
-         ]
+         ],
+         "operating_rounds": 1
       },
       {
          "name":"3",
+         "on":"3",
          "train_limit":4,
          "tiles":[
             "yellow",
             "green"
          ],
+         "operating_rounds": 2,
          "buy_companies":true
       },
       {
          "name":"4",
+         "on":"4",
          "train_limit":3,
          "tiles":[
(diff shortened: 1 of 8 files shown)


perwestling:
The "max 1 train from depot before phase 4" contains exactly the same implementation. So question is if this should be generalized so that all three uses one and same implementation?

tobymao:
this is using the old style of operating round and mostly copy and paste, for now, please remove all of those changes and only keep the config changes

perwestling:
@tobymao: Without "operating_rounds" lines in 18AL I encountered a bug in OR1: After the last company has bought its train, and then Skip buying more, the game got "stuck". It did not switch to SR2.
I describes the problem in #961. 

If I add the "operating_rounds" that problem disappeared.

But I will try to update my commit.

A question: What do you think about the operation - should I generalize it to make one ruby file for handling the operation round feature of just allowing one buy before phase 4, or is it OK with 3 files that are essentially the same?

tobymao:
@perwestling i'm currently rewriting the operating rounds to make it easier to do things like you want

3 files the same is not ok, and working on this now in the "old way of doing things" will make it more difficult for me to complete my rewrite

perwestling:
@tobymao: OK, I can wait for the rewrite before trying to implement this feature. I will close this pull-request and make a new one further on.

tobymao:
Are you in the 18xx slack? It will be more efficient to communicate there.

https://join.slack.com/t/18xxgames/shared_invite/zt-8ksy028m-CSZC~G5QtiFv60_jdqqulQ

PR #1013, opened 2020-07-06
1882: Allow SC to convert into SC and choose its home location

jenf:
Another 2 from the 1882 list:
the president discards P2 from the game, selects a par for SC, pays for one share, receives the presidency of the SCR
Until P2 is exchanged SC(R) cannot be opened


lib/engine/config/game/g_1882.rb
```diff
@@ -263,6 +263,12 @@ module G1882
           "hexes": [
             "H4"
           ]
+        },
+        {
+          "type": "exchange",
```
tobymao, line 268: this is a very specific ability, i don't know if we should put it in the generic framework

jenf: i'm struggling to find anything similar.

Is there a suggestion on how to do it which won't look terrible :)

jenf:
1824 is similar in that you exchange a mining company for the 20% share, but the par price is preset.

Maybe, it's best to list this as a president share exchange which is a generic ability. But then custom code specific to 1882 to add the cost of one share.

tobymao:
sounds good to me

PR #1027, opened 2020-07-07
prefers_color_scheme: dark => different fallback colors

### LINKED ISSUE #1016: 1846 Dark Mode "hides" the closed box on the stock market
The closed box is black and cannot be distinguished from the background in the market and in the tooltip below. Perhaps it should have a thin border in a contrasting color?

Screenshot 2020-07-06 17 19 13

yzemaze:
might fix #1016 – if not color_for(:bg|font) will at least work with pcs: dark when not logged in, too.

### DIFF
--- a/assets/app/lib/color.rb
+++ b/assets/app/lib/color.rb
@@ -4,10 +4,11 @@ require 'lib/hex'
 
 module Lib
   module Color
+    dark = `window.matchMedia('(prefers-color-scheme: dark)').matches`
     COLORS = {
-      bg: '#ffffff',
+      bg: dark ? '#000000' : '#ffffff',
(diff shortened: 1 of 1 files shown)


assets/app/lib/color.rb
```diff
@@ -4,10 +4,11 @@
 
 module Lib
   module Color
+    dark = `window.matchMedia('(prefers-color-scheme: dark)').matches`
```
tobymao, line 7: should be a constant DARK

tobymao: tested and does work on safari

yzemaze:
.freeze was missing. Please add it.

PR #1036, opened 2020-07-08
when closing corporations, only "spend" money if more than zero

### LINKED ISSUE #1029: 1846: Corporations cannot close during a SR
If a company has a share price of 10, either the president selling a share or the corporation having shares on the market should cause it to close. However, both selling a share and simply passing while there are shares on the market causes a "0 is not valid to spend" error.

This occurred in SR3 here: https://www.18xx.games/game/3559?action=562

michaeljb:
[Fixes #1029]

screenshots from a clone of game 3559 where I had the GT president sell shares to close it:

Screenshot from 2020-07-07 21-19-23

Screenshot from 2020-07-07 21-19-32

### DIFF
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -238,7 +238,7 @@ module Engine
           end
         end
 
-        corporation.spend(corporation.cash, @bank)
+        corporation.spend(corporation.cash, @bank) if corporation.cash.positive?
         corporation.trains.each { |t| t.buyable = false }
         if corporation.companies.any?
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1846.rb
```diff
@@ -238,7 +238,7 @@ def close_corporation(corporation)
           end
         end
 
-        corporation.spend(corporation.cash, @bank)
+        corporation.spend(corporation.cash, @bank) unless corporation.cash.zero?
```
tobymao, line 241: better to do if corporation.cash.positive?

tobymao: feel free to merge after that change

PR #1038, opened 2020-07-08
Apply company discount to border terrain, log terrain discounts

### LINKED ISSUE #877: 1889 - Sumimoto Mines Railway discount should appear in log
When a company lays a tile in 1889 using the Sumimoto Mines Railway company power, it would be nice if there was an indication in the log. For example, instead of "TR lays tile #7 with rotation 1 on H13" it could say "TR lays tile #7 with rotation 1 on H13 for free using Sumimoto Mines Railway"

### LINKED ISSUE #1001: [1846] Tunnel Blasting Company discount not applied when upgrading Detroit or Wheeling
GT has already bought in the TBC, tries to upgrade Detroit to tile 294. This would normally cost $100, but with TBC it should only cost $80, but GT gets charged full price.

[Game data right before upgrading tile](https://github.com/tobymao/18xx/files/4875436/1846-tbc-bonus-not-used-for-detroit.txt)

michaeljb:
[Fixes #1001]
[Closes #877]

Example mentioned in #877:
Screenshot from 2020-07-07 23-23-34

From the game data provided in #1001:
Screenshot from 2020-07-07 23-22-58
Screenshot from 2020-07-07 23-23-14

color mountains in middle of hex to match color of mountains on borders:
Screenshot from 2020-07-07 23-33-34

### DIFF
--- a/lib/engine/round/base.rb
+++ b/lib/engine/round/base.rb
@@ -269,9 +269,9 @@ module Engine
           if free
             0
           else
-            border, border_types = border_cost(tile)
+            border, border_types = border_cost(tile, entity)
             terrain += border_types if border.positive?
-            tile_cost(old_tile, entity.all_abilities) + border
+            tile_cost(old_tile, entity, @game) + border
           end
 
         entity.spend(cost, @game.bank) if cost.positive?
@@ -294,7 +294,7 @@ module Engine
         end
       end
 
-      def border_cost(tile)
+      def border_cost(tile, entity)
         hex = tile.hex
         types = []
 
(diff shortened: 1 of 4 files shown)


lib/engine/round/base.rb
```diff
@@ -309,13 +309,24 @@ def border_cost(tile)
...
+          end
+
+          cost - discount
         end
         [total_cost, types]
       end
 
-      def tile_cost(tile, abilities)
-        tile.upgrade_cost(abilities)
+      def tile_cost(tile, entity, game)
```
tobymao, line 328: theres no reason to pass in game here, you already have it

lib/engine/tile.rb
```diff
@@ -190,11 +190,18 @@ def lawson?
         ((cities.empty? && towns.one?) && edges.size > 2)
     end
 
-    def upgrade_cost(abilities)
-      ability = abilities.find { |a| a.type == :tile_discount }
+    def upgrade_cost(entity, game)
```
tobymao, line 193: so this is a smell, upgrade_cost should be moved out of tile.rb and into the round next to border_cost

tobymao:
bed time for me will merge tomorrow, or you can now if you're confident of the canges

benjaminxscott:
@michaeljb - Thanks for implementing the borders costs which I totally just handwaved in implementing tunnel blasting for yellow tiles in a prior commit 

I think your approach to calculating / handling in `border_cost()` is way cleaner than what I did in `tile_cost()`, btw
https://github.com/tobymao/18xx/blob/96c86d7d167b35e773d794c610861dc08a152868/lib/engine/round/base.rb#L328

PR #1040, opened 2020-07-08
Add the posibility to configure destination for a corporation

perwestling:
This will make it possible to implement things like
- Destination in 1822, 1844, 1856
- Historical objective in 18AL

There are more 18xx that have similar concepts.

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -43,6 +43,7 @@ module Engine
       @min_price = opts[:min_price]
       @always_market_price = opts[:always_market_price] || false
       @needs_token_to_par = opts[:needs_token_to_par] || false
+      @destination = opts[:destination] || nil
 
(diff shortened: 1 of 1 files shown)


perwestling:
An example on how this can be used in 18AL:

In lib/engine/config/game/g_18_al.rb:
- Add a historical objective (ie destination) for each corporation
  So for ABC that has Birmingham as historical objective:
  "destination":"G4"
- Subclass engine/round/operating.rb for 18AL and add a method similar to this:
   def place_token(action)
      super
      return unless action.entity.destination?(action.city.hex)

      action.entity.cash += 100
   end

By adding this a company in 18AL that puts a token in the city that is its historical objective, will get 100 added to its treasury.

tobymao:
this should be modeled as an ability as opposed to an attribute

perwestling:
I will try to do it as ability modelling instead.

PR #1042, opened 2020-07-08
Revision game_info & spreadsheet

yzemaze:
details s. commit comments
tr highlight on hover/tap limited to game_info & spreadsheet for now (only makes sense on larger tables, definitely not on .card)

### DIFF
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -43,7 +43,7 @@ module View
           h('div.title.nowrap', title_props, [h(:em, 'The Bank')]),
           h(:div, body_props, [
             h(:div, @game.format_currency(@game.bank.cash)),
-            *h(GameInfo, game: @game, layout: 'discarded_trains'),
+            h(GameInfo, game: @game, layout: 'discarded_trains'),
           ]),
         ])
       end
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -43,7 +43,7 @@ module View
           h('div.title.nowrap', title_props, [h(:em, 'The Bank')]),
           h(:div, body_props, [
             h(:div, @game.format_currency(@game.bank.cash)),
-            h(GameInfo, game: @game, layout: 'discarded_trains'),
+            *h(GameInfo, game: @game, layout: 'discarded_trains'),
           ]),
         ])
       end
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -49,8 +49,9 @@ module View
           phase_color = Array(phase[:tiles]).last
           phase_props = { style: {} }
           if Part::MultiRevenue::COLOR.include?(phase_color)
-            phase_props[:style][:backgroundColor] = Part::MultiRevenue::COLOR[phase_color]
-            phase_props[:style][:color] = contrast_on(Part::MultiRevenue::COLOR[phase_color])
+            bg_color = color_for(phase_color)
+            phase_props[:style][:backgroundColor] = bg_color
+            phase_props[:style][:color] = contrast_on(bg_color)
           end
 
           event_text = []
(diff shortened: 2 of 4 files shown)


assets/app/view/game/bank.rb
```diff
@@ -43,7 +43,7 @@ def render_card
           h('div.title.nowrap', title_props, [h(:em, 'The Bank')]),
           h(:div, body_props, [
             h(:div, @game.format_currency(@game.bank.cash)),
-            h(GameInfo, game: @game, layout: 'discarded_trains'),
+            *h(GameInfo, game: @game, layout: 'discarded_trains'),
```
tobymao, line 46: render should always return one item, so remove the * and make sure GameInfo returns one div

assets/app/view/game/game_info.rb
```diff
@@ -14,71 +16,41 @@ def render
...
-            discarded_trains,
-          ])
-        end
-
-        children << phases
-        children << game_info
-
-        h(:div, children)
+        children = upcoming_trains
+        children.concat discarded_trains unless @depot.discarded.empty?
```
tobymao, line 25: always use parenthesis

tobymao: prefer positive conditionals

if @depot.discarded.any?

assets/app/view/game/game_info.rb
```diff
@@ -88,54 +60,51 @@ def phases
           end
 
           h(:tr, [
-            h(:td, td_props, phase[:name] + (current_phase == phase ? ' (Current) ' : '')),
-            h(:td, td_props, phase[:operating_rounds]),
-            h(:td, td_props, phase[:train_limit]),
+            h(:td, (current_phase == phase ? '⮕ ' : '') + phase[:name]),
```
tobymao, line 63: unicode looks inconsistent in browsers, what about -> or >

assets/app/view/game/game_info.rb
```diff
@@ -147,62 +116,66 @@ def upcoming_trains
         rows = @depot.upcoming.group_by(&:name).map do |name, trains|
           train = trains.first
           discounts = train.discount&.group_by { |_k, v| v }&.map do |price, price_discounts|
-            price_discounts.map(&:first).join(',') + ' => ' + @game.format_currency(price)
+            price_discounts.map(&:first).join(', ') + ' ⮕ ' + @game.format_currency(price)
```
tobymao, line 119: same here

assets/app/view/game/game_info.rb
dfannius: It would be a crime to represent an arrow by the ASCII string `->` in the year 2020.

tobymao: lol ok maybe the arrow is ok or possible you can make an svg

tobymao:
can you change "Operating Order" to be "Order"

assets/app/view/game/bank.rb
yzemaze: It already returns `''` if there are no discarded trains. I was avoiding the div-wrapper around header + table on purpose as these are exactly the ones I tried to eradicate. I’ll find another solution.

assets/app/view/game/game_info.rb
yzemaze: *notetoself* don’t copy anyone’s code blindly

assets/app/view/game/game_info.rb
yzemaze: Arrows for consistency with spreadsheet (up/down arrow already indicates sort order there).
Even if it gets rendered slightly differently on different browsers it should still be clearly identifiable as an arrow ;) svg would just be more load.

assets/app/view/game/game_info.rb
yzemaze: > can you change "Operating Order" to be "Order"

Sure. I’ll forward any hate-mail from spreadsheets grognards to you ;)

assets/app/view/game/game_info.rb
yzemaze: Real question: Is the arrow sufficient to indicate current round? Additionally highlight row with background color maybe?

tobymao:
i think an arrow is sufficient

PR #1046, opened 2020-07-08
Correct tiles in 18AL (#961)

perwestling:
- Split 444a into two tiles with different labels (444b for Birmingham
  with label B, and 444m for Montgomery with label M)
- Block tile laying of tile 445 as can only be placed via company BLC
- Put icon of tile 445
  - to be able to use attribute for tile block
  - to separate it form normal gentle tile with town
- Add sticky coal icon for Warrior Coal token locations (5 hexes)
- When tile is placed in Montgomery ensure label is set to M to
  allow for a yellow tile to be put there and further tiles to be
  put there needing the label M.

### DIFF
--- a/lib/engine/tile.rb
+++ b/lib/engine/tile.rb
@@ -328,6 +328,11 @@ module Engine
       @revenue_to_render ||= stops.map(&:revenue_to_render)
     end
 
+    # Used to set label for a recently placed tile
+    def label=(label_name)
+      @label = Part::Label.new(label_name)
+    end
+
     private
 
     def separate_parts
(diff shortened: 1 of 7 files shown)


perwestling:
Although specific to 18AL I have created a label! method in Engine::Tile as the chosen tile need to be modified to allow for general hex becoming specialized. In case of 18AL it is Montgomery that has specific tiles from green phase, but is a general yellow city.
A similar situation can be found for Guadalajara in 18MEX which has specific tiles in later phases.

lib/engine/tile.rb
```diff
@@ -328,6 +328,11 @@ def revenue_to_render
       @revenue_to_render ||= stops.map(&:revenue_to_render)
     end
 
+    # Used to set label for a recently placed tile
+    def label!(label_name)
```
tobymao, line 332: instead this should be a setter function

def label=(name)
  @label = ...
end

perwestling: Have now updated according to this. Also did some other corrections in the commit.

PR #1049, opened 2020-07-09
render routes after (on top of) black track

yzemaze:
image

=>

image

Only TrackCurvilinearPath, because the few other occassions don’t really matter (and would hurt performance more than they’d do good). [as of now & afaik]

### DIFF
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -39,7 +39,7 @@ module View
           else
             @tile.paths.select { |path| path.edges.size == 2 }
             .map { |path| [path, color_for(path)] }
-            .sort_by { |_, color| color == TRACK_COLOR ? 0 : 1 }
+            .sort_by { |_, color| color == TRACK_COLOR ? 0 : ROUTE_COLORS.index(color) + 1 }
             .map do |path, color|
               h(TrackCurvilinearPath, region_use: @region_use, path: path, color: color)
             end
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -36,7 +36,7 @@ module View
           elsif @tile.cities.any?
             render_track_for_curvilinear_city
           else
-            @tile.paths.select { |path| path.edges.size == 2 }.map do |path|
+            @tile.paths.select { |path| path.edges.size == 2 }.sort_by { |path| color_for(path) }.map do |path|
               h(TrackCurvilinearPath, region_use: @region_use, path: path, color: color_for(path))
             end
           end
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -10,6 +10,7 @@ module View
     module Part
       class Track < Snabberb::Component
         # http://mkweb.bcgsc.ca/colorblind/ 13 color palette
+        TRACK_COLOR = '#000000'
         ROUTE_COLORS = %i[#AA0A3C #0A9B4B #005AC8 #8214A0].freeze
 
         needs :tile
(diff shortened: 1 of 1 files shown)


assets/app/view/game/part/track.rb
```diff
@@ -36,7 +36,7 @@ def render
           elsif @tile.cities.any?
             render_track_for_curvilinear_city
           else
-            @tile.paths.select { |path| path.edges.size == 2 }.map do |path|
+            @tile.paths.select { |path| path.edges.size == 2 }.sort_by { |path| color_for(path) }.map do |path|
```
tobymao, line 39: this is obscure, how can you sort by a color? yes it's a string but seems like a hack

if you want to do this for real let's do it for real

@tiles.path
.select...
.map color
.sort by something reasonnable

yzemaze: There’s just one reasonable attribute to sort paths by here: is it part of a route or not? Luckily someone already wrote color_for(path) that checks exactly that (if I got L79+ right). It even has the added benefit of returning a value (instead of 0/1) to sort route-containing paths by so we don’t render color1 over color2 on some tiles and vice versa on others. Yes one could use the same algorithm again and just return the index, but why write code twice? As long as the color for standard track is #000000 (and you pretty decidedly said it shouldn’t be configurable) this works fine.

tobymao: as i eluded to in my previous comment

you can sort by whether or not the color is the default color which is much more explicit than relying on #00000 being alphabetically behind colors

yzemaze: a rather elusive allusion … especially the reverted map/sort_by in the stub doesn’t help at all (to say the least). If you just wanted index (or another numerical value) instead of color as sorting criterion you should’ve just said so.

Same algorithm inline would be:
`.sort_by { |path| @routes_paths.find_index { |route_paths| route_paths.any? { |p| path == p } } || -1 }`
… or rather more "verbose"? (=> copycat method of color_for)

btw: I haven’t seen an explicit definition of a default track color anywhere – afaik it’s just the string in color_for.

tobymao: no it should be

`
@tiles
.path
.select { ... }
.map { |path| [path, color_for(path)] }
.sort_by { |_, color| color == :black 1 : 0 } 
`

yzemaze: … or sth. along these lines …
```
@tile.paths.select { |path| path.edges.size == 2 }
            .map { |path| [path, color_for(path)] }
            .sort_by { |_, color| color == '#000000' ? 0 : 1 }
            .map do |path, color|
              h(TrackCurvilinearPath, region_use: @region_use, path: path, color: color)
```
But as I said, this can lead to renderings like this:
image
It’s illegal, so as long as there are no games implemented with track-sharing (1822, 18Ardennes, other?) this doesn’t matter. Replacing 1 by an index/hash should work to cover that use case later on.

tobymao: how does you other one work? because all the colors are relative?

yzemaze: Which other one exactly?

yzemaze: Actually the principle is the same: Just sort all colors except standard (black) the same way on every hex. Whether you sort by hex-code lexicographically/alphabetically, by their index, a hash (of hex-code) or sth. else doesn’t matter – as long as the resulting sorting order is the same on all hexes to prevent renderings like in the screenshot.
The result [e. g. red on green on … for all hexes] isn’t relevant for now as it doesn’t matter which route color is rendered on top on crossings and trains don’t share track (yet).

With track sharing games order of routes will become relevant, but in combination with route width, dash-array or outlines this shouldn’t be a big problem. (Rendering routes next to each other might look nicer but would likely be way more trouble.)

assets/app/view/game/part/track.rb
```diff
@@ -36,8 +37,11 @@ def render
           elsif @tile.cities.any?
             render_track_for_curvilinear_city
           else
-            @tile.paths.select { |path| path.edges.size == 2 }.map do |path|
-              h(TrackCurvilinearPath, region_use: @region_use, path: path, color: color_for(path))
+            @tile.paths.select { |path| path.edges.size == 2 }
+            .map { |path| [path, color_for(path)] }
+            .sort_by { |_, color| color == TRACK_COLOR ? 0 : 1 }
```
tobymao, line 42: can you change this to work with indicies?

PR #1050, opened 2020-07-09
18AL: Add support for historical objective (#961)

perwestling:
Introduce an ability Destination that can be used to specify
a destination hex. This ability can be used in e.g. 1822,
1844, 1856, 18AL.

Configure historical objective hex for all corporations in 18AL
using the destination ability. When a corporation places its
token in its destination hex, it receives $100 to treasury as
it has reached its historical objective.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -48,6 +48,9 @@ module View
           children << h(Companies, owner: @corporation, game: @game) if @corporation.companies.any?
         end
 
+        abilities = @corporation.all_abilities.select { |ability| ability.owner.corporation? }
+        children << render_abilities(abilities) if abilities&.any?
+
         if @corporation.owner
           subchildren = (@corporation.operating_history.empty? ? [] : [render_revenue_history])
           children << h('table.center', [h(:tr, subchildren)])
@@ -298,6 +301,29 @@ module View
       def render_revenue_history
         last_run = @corporation.operating_history[@corporation.operating_history.keys.max].revenue
         h('td.bold', "Last Run: #{@game.format_currency(last_run)}")
+      end
+
+      def render_abilities(abilities)
+        attribute_lines = []
+
+        abilities.each do |ability|
+          attribute_lines << h('div.name.nowrap', ability.name)
+          attribute_lines << h('div.right', ability.display_value || '')
+        end
+
+        table_props = {
+          style: {
+            padding: '0 0.5rem',
+            grid: 'auto / 3fr 2fr',
+            gap: '0 0.2rem',
+          },
+        }
+
+        h('div#company_table', table_props, [
+          h('div.bold', 'Attribute'),
+          h('div.bold.right', 'Value'),
+          *attribute_lines,
+        ])
       end
 
       def selected?
--- a/lib/engine/ability/base.rb
+++ b/lib/engine/ability/base.rb
@@ -9,14 +9,17 @@ module Engine
       include Helper::Type
       include Ownable
 
+      attr_accessor :name, :display_value
       attr_reader :type, :owner_type, :remove, :when, :count
 
-      def initialize(type:, owner_type: nil, count: nil, remove: nil, **opts)
+      def initialize(type:, name: nil, owner_type: nil, count: nil, remove: nil, display_value: nil, **opts)
         @type = type&.to_sym
+        @name = name&.to_s || @type
         @owner_type = owner_type&.to_sym
         @when = opts.delete(:when)&.to_s
         @count = count
         @used = false
+        @display_value = display_value&.to_s
         @remove = remove&.to_s
 
         setup(**opts)
--- a/lib/engine/config/game/g_18_al.rb
+++ b/lib/engine/config/game/g_18_al.rb
@@ -238,7 +238,17 @@ module Engine
             100
          ],
          "coordinates":"A4",
-         "color":"blue"
+         "color":"blue",
+         "abilities": [
+            {
+              "type": "assign_hexes",
+              "name": "Historical objective",
+              "hexes": [
+                "G4"
+              ],
+              "count": 1
+            }
+         ]
       },
       {
          "sym":"M&O",
@@ -251,7 +261,17 @@ module Engine
             100
          ],
          "coordinates":"Q2",
-         "color":"orange"
+         "color":"orange",
+         "abilities": [
+            {
+               "type": "assign_hexes",
+               "name": "Historical objective",
+               "hexes": [
+                 "K2"
+               ],
+               "count": 1
+            }
+         ]
       },
       {
          "sym":"WRA",
@@ -264,7 +284,17 @@ module Engine
             100
          ],
          "coordinates":"L5",
-         "color":"red"
+         "color":"red",
+         "abilities": [
+            {
+               "type": "assign_hexes",
+               "name": "Historical objective",
+               "hexes": [
+                 "J7"
+               ],
+               "count": 1
+            }
+         ]
       },
       {
          "sym":"ATN",
@@ -276,7 +306,17 @@ module Engine
             100
          ],
          "coordinates":"F1",
-         "color":"black"
+         "color":"black",
+         "abilities": [
+            {
+               "type": "assign_hexes",
+               "name": "Historical objective",
+               "hexes": [
+                 "L1"
+               ],
+               "count": 1
+            }
+         ]
       },
       {
          "sym":"ABC",
@@ -287,7 +327,17 @@ module Engine
             40
          ],
          "coordinates":"G6",
-         "color":"green"
+         "color":"green",
+         "abilities": [
+            {
+               "type": "assign_hexes",
+               "name": "Historical objective",
+               "hexes": [
+                 "G4"
+               ],
+               "count": 1
+            }
+         ]
       },
       {
          "sym":"TAG",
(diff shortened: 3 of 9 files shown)


lib/engine/round/g_18_al/operating.rb
```diff
@@ -0,0 +1,23 @@
...
+require_relative '../operating'
+
+module Engine
+  module Round
+    module G18AL
+      class Operating < Operating
+        def place_token(action)
+          super
+
+          ability = action.entity.all_abilities.find do |a|
```
tobymao, line 12: is cleaner

action.entity.abilities(:destination) do |ability|
 ...
end

tobymao: also make sure to call ability.remove so they can't use it twice

lib/engine/config/game/g_18_al.rb
```diff
@@ -229,7 +229,7 @@ module G18AL
    "corporations":[
       {
          "sym":"L&N",
-         "name":"Louisville & Nashville Railroad",
+         "name":"Louisville & Nashville Railroad (HO: Birmingham)",
```
tobymao, line 232: the historical objective should not be in the name, it can be referenced with the abilitiy

perwestling: @tobymao: Where can the users find information about which is the historical objective for each corporation? In the physical game this is shown on the corporation charters as this is an ability that the corporation has from the beginning of the game and it does not change.

perwestling: Ah, you mean that it should be appended to the corporation name during setup, by using the ability? Or?

perwestling: I will use "Decorator pattern" to add historical objective to the full name.

perwestling: Did not get decorator to work, so instead I added an attribute in corporation (and minor) that can be used to set some extra info. 
Any extra info is now shown as " (...extra info)" after the corporation name.

An improvement could be to add an extra line on the corporation charter to show the extra information. This line could be at the end, maybe.
But that can be done when the next 18xx that need to show corporation information in a similar way.

assets/app/view/game/corporation.rb
```diff
@@ -81,9 +81,11 @@ def render_title
           },
         }
 
+        full_name = @corporation.full_name
+        full_name += " (#{@corporation.extra_info})" unless @corporation.extra_info.nil?
```
tobymao, line 85: we don't need the extra info attribute, delete that, we should simply render abilities for corporations, if that's not clear let's chat about it on slack

perwestling: @tobymao: This is not done today, I suppose. Is the idea to show the name of all abilities that is not part of a company?

So it would be
_ABC         4   $70
            Trains Cash
...
Company                              Income
Memphis & Charle....                    $0

Ability
Historical Objective:     Birmingham_

perwestling: This is how the Entity page looks like with this feature.
Screenshot from 2020-07-13 19-02-18

perwestling:
ability/base.rb should now be the same in both this one and #1089.

assets/app/view/game/corporation.rb
```diff
@@ -44,6 +44,9 @@ def render
           children << h(Companies, owner: @corporation, game: @game) if @corporation.companies.any?
         end
 
+        abilities = @corporation.all_abilities.select { |ability| ability.owner.corporation? }
+        children << render_abilities(abilities) unless abilities.nil? || abilities.empty?
```
tobymao, line 48: don't use unless with multiple statements

children << render... if abilities&.any?

lib/engine/ability/base.rb
```diff
@@ -9,14 +9,17 @@ class Base
       include Helper::Type
       include Ownable
 
+      attr_accessor :name, :value
       attr_reader :type, :owner_type, :remove, :when, :count
 
-      def initialize(type:, owner_type: nil, count: nil, remove: nil, **opts)
+      def initialize(type:, name: nil, owner_type: nil, count: nil, remove: nil, value: nil, **opts)
         @type = type&.to_sym
+        @name = name.nil? ? @type : name&.to_s
```
tobymao, line 17: @name = name&.to_s || @type

lib/engine/game/g_18_al.rb
```diff
@@ -18,6 +18,12 @@ class G18AL < Base
 
       def setup
         setup_company_price_50_to_150_percent
+
+        @corporations.each do |corporation|
+          corporation.abilities(:destination) do |ability|
+            ability.value = @hexes.detect { |h| h.name == ability.hex }.location_name
```
tobymao, line 24: use find

lib/engine/ability/base.rb
```diff
@@ -9,14 +9,17 @@ class Base
...
         @owner_type = owner_type&.to_sym
         @when = opts.delete(:when)&.to_s
         @count = count
         @used = false
+        @value = value.nil? ? '' : value&.to_s
```
tobymao, line 22: what's the point of value? i don't really see it being used

lib/engine/ability/base.rb
tobymao: do we really need name? can't we just use the type, you've copied and pasted historical object multiple times, destination may be fine

lib/engine/round/g_18_al/operating.rb
perwestling: Moved to g_18_al/token.rb. Modified to use :destination, and then remove_ability.

lib/engine/config/game/g_18_al.rb
perwestling: Think this conversation is resolved.

assets/app/view/game/corporation.rb
perwestling: @tobymao Do you think this view is OK? 

In case of a corporation that owns both companies and have attributes the view is
Corporation_with_ability_and_company

assets/app/view/game/corporation.rb
perwestling: Fixed next push.

lib/engine/ability/base.rb
perwestling: Applied 1st suggestion in next push.
Name field is used when presenting the ability on the corporation charter.

lib/engine/game/g_18_al.rb
perwestling: As far as I can tell detect and find are equivalent, and it seems find does call detect internally. But sure, I will switch in the next push.

lib/engine/ability/base.rb
perwestling: It is used in render_abilities in asserts/app/view/game.corporation.
So e.g. is is used to show the hex name where Coal Field token is assigned, it is used to save the city name for the historical objective.

perwestling:
@tobymao I have now updated based on your review comments (see above).
Did also include refactoring of revenue_of part to make it possible to use the same in 18MEX with minimal duplication.
This is base for the other PRs of 18AL, except the end game check on that is separate, so I will look into that one next.

assets/app/view/game/corporation.rb
```diff
@@ -48,6 +48,9 @@ def render
           children << h(Companies, owner: @corporation, game: @game) if @corporation.companies.any?
         end
 
+        abilities = @corporation.all_abilities.select { |ability| ability.owner.corporation? }
+        children << render_abilities(abilities) if abilities&.any?
```
tobymao, line 52: you don't need & here, abilities will always be an array

lib/engine/ability/base.rb
```diff
@@ -9,14 +9,17 @@ class Base
...
 
-      def initialize(type:, owner_type: nil, count: nil, remove: nil, **opts)
+      def initialize(type:, name: nil, owner_type: nil, count: nil, remove: nil, value: nil, **opts)
         @type = type&.to_sym
+        @name = name&.to_s || @type
         @owner_type = owner_type&.to_sym
         @when = opts.delete(:when)&.to_s
         @count = count
         @used = false
+        @value = value.nil? ? '' : value&.to_s
```
tobymao, line 22: ok a better name for value i think would be description because it's used for displaying text, value makes it seem like it has actual functional meaning

lib/engine/ability/destination.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
```
tobymao, line 5: i wonder if we don't even need the destination ability because we already have the assign_hexes ability

what do you think?

lib/engine/ability/base.rb
perwestling: Description could possibly be shown when clicking on the ability in the view.
I can rename it to display_value instead?

lib/engine/game/g_18_al.rb
```diff
@@ -15,9 +17,17 @@ class G18AL < Base
...
       include CompanyPrice50To150Percent
+      include Revenue4D
+      include TerminusCheck
 
       def setup
         setup_company_price_50_to_150_percent
+
+        @corporations.each do |corporation|
+          corporation.abilities(:destination) do |ability|
+            ability.value = @hexes.find { |h| h.name == ability.hex }.location_name
```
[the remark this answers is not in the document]
perwestling: I use display of value for several abilities, not just destination. So it is a general thing.
But I will try to see if I can change this to using the assigns_hex ability instead.

PR #1056, opened 2020-07-10
1846: hotseat draft displays drafted companies and adjusted cash/value

### LINKED ISSUE #727: [1846] Private Draft doesn't show a player the companies they've drafted
During the private draft (in a hotseat game), nothing shows the current player how much money they have already allocated for the privates they've already chosen, or what privates they've already chosen. I realize that's hidden information until the end of the draft, but a player should be able to see that information when they click "show".

jaysachs:
Fixes #727.

### DIFF
--- a/assets/app/view/game/round/draft.rb
+++ b/assets/app/view/game/round/draft.rb
@@ -90,14 +90,14 @@ module View
           drafted = draft_dist.choices[@player]
 
           player = @player.clone
-          dummy = Engine::Player.new("_dummy_")
+          dummy = Engine::Player.new('_dummy_')
           drafted.reject { |company| draft_dist.blank?(company) }
             .each do |company|
-            unless draft_dist.blank?(company)
-              company.owner = player
-              player.spend(company.min_bid, dummy)
-              player.companies << company
-            end
+            next if draft_dist.blank?(company)
+
+            company.owner = player
+            player.spend(company.min_bid, dummy)
+            player.companies << company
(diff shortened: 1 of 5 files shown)


assets/app/view/game_page.rb
```diff
@@ -231,7 +231,7 @@ def render_action
       when Engine::Round::Operating
         h(Game::Round::Operating, game: @game)
       when Engine::Round::G1846::Draft
-        h(Game::Round::Draft, game: @game)
+        h(Game::Round::Draft, game: @game, round: @round)
```
tobymao, line 234: this is unnecessary, you can get @game.round

assets/app/view/game/round/draft.rb
```diff
@@ -85,9 +86,31 @@ def render_input
         def render_player
           return nil if @hidden
 
-          h(:div, [
-            h(Player, player: @player, game: @game),
-          ])
+          draft_dist = @round.steps[-1]
+          drafted = draft_dist.choices[@player]
+
+          player = @player.clone
```
tobymao, line 92: i'm not a big fan of this approach, it seems a bit hacky, can we instead do a simpler view?

tobymao: just listing the companies name that have been drafted and the remaining money so we don't have to clone and mock objects?

jenf:
solution based on suggest approach done as #1307

PR #1064, opened 2020-07-10
18AL/18GA/18MEX/18TN: Limit depot train buy to one before phase 4 (#961)

perwestling:
This applies to the following 18xx:
  - 18AL
  - 18GA
  - 18MEX
  - 18TN

When buying train from depot in phase 2 and phase 3
only one train can be bought.

### DIFF
--- a/lib/engine/game/g_18_al.rb
+++ b/lib/engine/game/g_18_al.rb
@@ -19,6 +19,21 @@ module Engine
       def setup
         setup_company_price_50_to_150_percent
       end
+
+      def operating_round(round_num)
+        Round::Operating.new(self, [
+          Step::Bankrupt,
+          Step::DiscardTrain,
+          Step::BuyCompany,
+          Step::HomeToken,
+          Step::Track,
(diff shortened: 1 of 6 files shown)


jenf:
In general the approach looks fine, but I think it'll need to wait until after the steps rewrite gets merged. It'll be a single class then

perwestling:
@jenf and @tobymao: Now this is rewritten to match the new steps so it should be possible to merge.

lib/engine/step/g_18_al/train.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
jenf, line 1: Looks good, only real comment is the use of 'g_18_al' when it's common to nearly all of Mark Derrick's games.
How about g_18_markderrick?

or train_buyone_until_phase4.rb?

perwestling: Renamed it to a descriptive name, and put the file in lib/engine/step.

PR #1073, opened 2020-07-11
18AL: Implement company Brown & Sons Lumber Co

perwestling:
Add the posibility to configure skip of feature check when
laying tile.

Add ability for Brown & Sons Lumber Co to lay tile 445
as a one of ability, for free.

Tile 445 does now ignore town count in the hex it is placed.
This tile will only be placed via the Brown & Sons Lumber
Co special ability.

A similar ability exists in 18MEX, but that will be implemented
later.

### DIFF
--- a/lib/engine/ability/tile_lay.rb
+++ b/lib/engine/ability/tile_lay.rb
@@ -5,12 +5,13 @@ require_relative 'base'
 module Engine
   module Ability
     class TileLay < Base
-      attr_reader :hexes, :tiles, :free
+      attr_reader :hexes, :tiles, :free, :discount
 
-      def setup(hexes:, tiles:, free: false)
+      def setup(hexes:, tiles:, free: false, discount: nil)
         @hexes = hexes
         @tiles = tiles
         @free = free
+        @discount = discount || 0
       end
     end
   end
--- a/lib/engine/config/game/g_18_al.rb
+++ b/lib/engine/config/game/g_18_al.rb
@@ -209,7 +209,26 @@ module Engine
          "name":"Brown & Sons Lumber Co.",
          "value":70,
          "revenue":15,
-         "desc":"Owning corporation may lay the Lumber Terminal track tile (number 445) in an empty swamp hex. It need not be connected to an existing station of the corporation.  The corporation need not pay the $20 cost of the swamp hex.  And it does not count as the corporation's one tile lay per turn.  (But it still must be laid during the tile-laying step of the corporation's turn). Laying the tile does not close the B&SLC. The Lumber Terminal tile is permanent, and cannot be upgraded."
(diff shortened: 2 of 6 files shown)


lib/engine/config/tile.rb
```diff
@@ -52,7 +52,7 @@ module Tile
         # '403' => 'city=revenue:30;town=revenue:30;path=a:0,b:_0;label=B;upgrade=cost:40',
         '437' => 'town=revenue:30;path=a:0,b:_0;path=a:_0,b:2;icon=image:port,blocks_lay:1',
         '438' => 'city=revenue:40;path=a:0,b:_0;path=a:2,b:_0;label=H;upgrade=cost:80',
-        '445' => 'town=revenue:20;path=a:0,b:_0;path=a:_0,b:2;icon=image:18_al/tree,blocks_lay:1',
+        '445' => 'town=revenue:20;path=a:0,b:_0;path=a:_0,b:2;icon=image:18_al/tree,blocks_lay:1;skip_check=town_size',
```
tobymao, line 55: i don't think we should have this property on the tile, it should be a part of the ability, i also don't know if it's necessary to have this check

lib/engine/config/game/g_18_al.rb
```diff
@@ -209,7 +209,26 @@ module G18AL
...
+         "desc":"Owning corporation may lay the Lumber Terminal track tile (number 445) in an empty swamp hex. It need not be connected to an existing station of the corporation.  The corporation need not pay the $20 cost of the swamp hex.  And it does not count as the corporation's one tile lay per turn.  (But it still must be laid during the tile-laying step of the corporation's turn). Laying the tile does not close the B&SLC. The Lumber Terminal tile is permanent, and cannot be upgraded.",
+         "abilities": [
+             {
+               "type": "tile_lay",
+               "free":true,
+               "owner_type": "corporation",
+               "tiles": [
+                  "445"
+               ],
+               "hexes": [
```
tobymao, line 221: this may be a lot easier if when using special ability tile lays, we only check to see if the color is correct as opposed to the full validation

perwestling: I can give it a try. As it is right now (in master) there are a couple of checks that stops 445 from appearing as a tile to select.

lib/engine/round/base.rb
```diff
@@ -294,6 +294,10 @@ def lay_tile(action)
         end
       end
 
+      def game_error_not_upgradeable(old_tile, tile)
```
tobymao, line 297: this change is unnecessary, if you did this due to linters, just do multiline unless statement

lib/engine/config/tile.rb
perwestling: The property was removed in later versions of this PR.

lib/engine/round/base.rb
perwestling: Fixed in later versions.

lib/engine/ability/tile_lay.rb
```diff
@@ -5,12 +5,13 @@
...
     class TileLay < Base
-      attr_reader :hexes, :tiles, :free
+      attr_reader :hexes, :tiles, :free, :discount
 
-      def setup(hexes:, tiles:, free: false)
+      def setup(hexes:, tiles:, free: false, discount: 0)
         @hexes = hexes
         @tiles = tiles
         @free = free
+        @discount = discount
```
tobymao, line 14: it will be safer if we do discount: nil
@discount ||= 0

perwestling: Fixed

PR #1074, opened 2020-07-11
18AL/18MS: Implement company that gives discount for some trains

perwestling:
Configurable attribute 'train_discount' added. When this is set
for a corporation, the trains that has been configured in the
attribute, will appear as extra lines in the BuyTrains view,
shown with the discounted price.

When train bought for a discounted price the company should close.

This attribute has been configured for:
- 18AL: New Decatur Yards (50% discount for 3, 4, 5 trains)
- 18MS: Mobile & Ohio Railway (100$ discount for 3+, 4+ trains)

This commit is for peer review. The close of companies when
used does not seem to work as it should.

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -54,6 +54,8 @@ module View
       def render
         step = @game.round.active_step
         @corporation = step.current_entity
+        @ability = @selected_company&.abilities(:train_discount, 'train') if @selected_company&.owner == @corporation
+
         @depot = @game.depot
 
         available = step.buyable_trains.group_by(&:owner)
@@ -135,10 +137,11 @@ module View
             .sort_by { |_, v| v[:price] }
             .flat_map do |name, variant|
             price = variant[:price]
+            price = @ability&.discounted_price(train, price) || price
 
             buy_train = lambda do
               process_action(Engine::Action::BuyTrain.new(
-                @corporation,
+                @ability ? @selected_company : @corporation,
                 train: train,
                 price: price,
                 variant: name,
--- a/lib/engine/ability/train_discount.rb
+++ b/lib/engine/ability/train_discount.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class TrainDiscount < Base
+      attr_reader :discount, :trains
+      def setup(discount:, trains:)
+        @discount = discount
+        @trains = trains
+      end
+
+      def discounted_price(train, price)
+        return price unless @trains.include?(train.name)
+
+        price - (discount > 1 ? discount : (price * discount))
+      end
+    end
+  end
+end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1030,7 +1030,7 @@ module Engine
           Step::Token,
           Step::Route,
           Step::Dividend,
-          Step::Train,
+          Step::BuyTrain,
           [Step::BuyCompany, blocks: true],
         ], round_num: round_num)
       end
--- a/lib/engine/step/single_depot_train_buy_before_phase4.rb
+++ b/lib/engine/step/single_depot_train_buy_before_phase4.rb
@@ -6,24 +6,24 @@ module Engine
   module Step
     class SingleDepotTrainBuyBeforePhase4 < Train
       def buyable_trains
-        super.reject { |x| x.from_depot? && @depot_trains_bought.any? && !@game.phase.available?('4') }
-      end
-
-      def setup
-        super
-        @depot_trains_bought = []
+        super.reject { |x| x.from_depot? && !@game.phase.available?('4') && already_bought_from_depot? }
       end
 
       def process_buy_train(action)
-        # Since the train won't be in the depot after being bought store the state now.
         from_depot = action.train.from_depot?
         super
-
         return unless from_depot
 
-        @depot_trains_bought << action.train.sym
-
+        @round.trains_bought << {
+          entity: action.entity,
+        }
         pass! unless buyable_trains.any?
+      end
+
+      private
+
+      def already_bought_from_depot?
+        @round.trains_bought.find { |e| e[:entity].name == current_entity.name }
       end
     end
   end
--- a/lib/engine/step/single_depot_train_buy_before_phase4.rb
+++ b/lib/engine/step/single_depot_train_buy_before_phase4.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-require_relative '../train'
+require_relative 'buy_train'
 
 module Engine
   module Step
-    class SingleDepotTrainBuyBeforePhase4 < Train
+    class SingleDepotTrainBuyBeforePhase4 < BuyTrain
       def buyable_trains
         super.reject { |x| x.from_depot? && !@game.phase.available?('4') && already_bought_from_depot? }
       end
--- a/lib/engine/step/special_buy_train.rb
+++ b/lib/engine/step/special_buy_train.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+require_relative 'train'
+
+module Engine
+  module Step
+    class SpecialBuyTrain < Base
+      include Train
+
+      ACTIONS = %w[buy_train].freeze
+
+      def actions(entity)
+        return [] unless ability(entity)
+
+        ACTIONS
+      end
+
+      def blocks?
+        false
+      end
+
+      def process_buy_train(action)
+        actual_buyer = action.entity.owner
+
+        buy_train_action(action, actual_buyer)
+
+        @round.trains_bought << {
+          entity: actual_buyer,
+        } if action.train.from_depot?
+
+        ability = ability(action.entity)
+        ability.use! if action.price < action.train.price &&
+          ability.discounted_price(action.train, action.train.price) == action.price
+        action.entity.close! if ability.count.zero?
+
+        pass! unless can_buy_train?(actual_buyer)
+      end
+
+      def ability(entity)
+        return unless entity.company?
+
+        entity.abilities(:train_discount, 'train')
+      end
+    end
+  end
+end
--- a/lib/engine/step/train.rb
+++ b/lib/engine/step/train.rb
@@ -91,6 +91,7 @@ module Engine
 
         @log << "#{entity.name} #{verb} a #{train.name} train for "\
           "#{@game.format_currency(price)} from #{source}"
+
         entity.buy_train(train, price)
         pass! unless can_buy_train?(entity)
       end
@@ -166,6 +167,12 @@ module Engine
 
       def issuable_shares
         []
+      end
+
+      def round_state
+        {
+          trains_bought: [],
+        }
       end
     end
   end
--- a/lib/engine/step/train.rb
+++ b/lib/engine/step/train.rb
@@ -4,39 +4,7 @@ require_relative 'base'
 
 module Engine
   module Step
-    class Train < Base
-      def actions(entity)
-        # 1846 and a few others minors can't buy trains
-        return [] if entity.minor?
-
-        # TODO: This needs to check it actually needs to sell shares.
-        return ['sell_shares'] if entity == current_entity.owner
-
-        return [] if entity != current_entity
-        # TODO: Not sure this is right
-        return %w[sell_shares buy_train] if must_buy_train?(entity)
-        return %w[buy_train pass] if can_buy_train?(entity)
-
-        []
-      end
-
-      def description
-        'Buy Trains'
-      end
-
-      def pass_description
-        @acted ? 'Done (Trains)' : 'Skip (Trains)'
-      end
-
-      def sequential?
-        true
-      end
-
-      def pass!
-        @last_share_sold_price = nil
-        super
-      end
-
+    module Train
       def can_buy_train?(entity = nil)
         entity ||= current_entity
         can_buy_normal = room?(entity) &&
@@ -55,8 +23,8 @@ module Engine
         @game.must_buy_train?(entity)
       end
 
-      def process_buy_train(action)
-        entity = action.entity
+      def buy_train_action(action, entity = nil)
+        entity ||= action.entity
         train = action.train
         train.variant = action.variant
         price = action.price
--- a/lib/engine/config/game/g_18_al.rb
+++ b/lib/engine/config/game/g_18_al.rb
@@ -256,7 +256,21 @@ module Engine
          "name":"New Decatur Yards",
          "value":120,
          "revenue":20,
-         "desc":"A corporation which owns the New Decatur Yards may purchase any one new train from the bank for half the regular price. This action closes the New Decatur Yards."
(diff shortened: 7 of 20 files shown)


perwestling:
This is more for peer review. Note problem mentioned in commit.

assets/app/view/game/buy_trains.rb
```diff
@@ -86,10 +86,12 @@ def render
 
         if (round.can_buy_train? && round.corp_has_room?) || round.must_buy_train?
           children << h(:h3, h3_props, 'Available Trains')
-          children << h(:div, div_props, [
-            *from_depot(depot_trains),
-            *other_corp_trains.any? ? other_trains(other_corp_trains) : '',
-          ])
+          lines = [*from_depot(depot_trains)]
```
tobymao, line 89: this is redundant, just do lines = from_depot

assets/app/view/game/buy_trains.rb
```diff
@@ -86,10 +86,12 @@ def render
...
+          @corporation.abilities(:train_discount) do |ability|
+            lines += [*from_depot_discounted(depot_trains, ability)]
```
tobymao, line 91: redundant

and better to do lines.concat(from_depot)

assets/app/view/game/buy_trains.rb
```diff
@@ -154,6 +156,38 @@ def from_depot(depot_trains)
...
+            price = variant[:price]
+            price -= if ability.absolute
+                       ability.discount
+                     else
+                       price * (ability.discount / 100.0)
+                     end
+
+            buy_train = lambda do
+              # FIXME: Does not seem to remove company from corporation
+              ability.use!
```
tobymao, line 173: you cannot make any changes in the view and you should not

this should all be done in the engine, this is why it does not work

perwestling: OK. Have now pushed up a new commit which does the update in the engine. Now it works.

assets/app/view/game/buy_trains.rb
perwestling: I refactored this so that the "gritty stuff" is done in from_depot.

assets/app/view/game/buy_trains.rb
```diff
@@ -129,20 +129,37 @@ def render
...
         depot_trains.flat_map do |train|
+          next unless ability.nil? || (ability.trains.include? train.name)
+
           train.variants.sort_by { |_, v| v[:price] }.flat_map do |name, variant|
             price = variant[:price]
+            price -= if ability.absolute
+                       ability.discount
+                     else
+                       price * (ability.discount / 100.0)
+                     end unless ability.nil?
```
tobymao, line 149: if ability

assets/app/view/game/buy_trains.rb
```diff
@@ -129,20 +129,37 @@ def render
...
                 @corporation,
                 train: train,
                 price: price,
                 variant: name,
+                ability: ability&.type
               ))
             end
 
             source = @depot.discarded.include?(train) ? 'The Discard' : 'The Depot'
+            source += ' discounted' unless ability.nil?
```
tobymao, line 162: if ability

lib/engine/action/buy_train.rb
```diff
@@ -5,14 +5,15 @@
 module Engine
   module Action
     class BuyTrain < Base
-      attr_reader :train, :price, :exchange
+      attr_reader :train, :price, :exchange, :ability
 
-      def initialize(entity, train:, price:, variant: nil, exchange: nil)
+      def initialize(entity, train:, price:, variant: nil, exchange: nil, ability: nil)
```
tobymao, line 10: we should not do this (add ability here)

instead, just change the price when you buy the train

lib/engine/step/g_18_al/train.rb
```diff
@@ -0,0 +1,26 @@
...
+module Engine
+  module Step
+    module G18AL
+      class Train < SingleDepotTrainBuyBeforePhase4
+        def process_buy_train(action)
+          ability = action.ability
+
+          super
+
+          return if ability.nil?
```
tobymao, line 14: super

action.entity.ability(:train) do |ability|
 ...

assets/app/view/game/buy_trains.rb
perwestling: In the latest version I moved the discount calculation to the train_discount ability class, and rewrote it.

assets/app/view/game/buy_trains.rb
perwestling: Fixed

lib/engine/action/buy_train.rb
perwestling: Fixed

lib/engine/step/g_18_al/train.rb
perwestling: Fixed, and this part was rewritten as well. Now the use! is done in the g_18_al/train.rb code.

assets/app/view/game/buy_trains.rb
```diff
@@ -129,10 +129,21 @@ def render
       end
 
       def from_depot(depot_trains)
+        trains_from_depot = buyable_trains_in_depot(depot_trains)
+        @corporation.abilities(:train_discount) do |ability|
```
tobymao, line 133: why do we need to do this separately, why can't you just call buyable_trains_in_depot one time?

assets/app/view/game/buy_trains.rb
```diff
@@ -129,10 +129,21 @@ def render
...
+          trains_from_depot.concat(buyable_trains_in_depot(depot_trains, ability))
+        end
+        trains_from_depot
+      end
+
+      def buyable_trains_in_depot(depot_trains, ability = nil)
         depot_trains.flat_map do |train|
+          trains_on_interest = ability&.trains || [train.name]
```
tobymao, line 141: trains_of_interest

tobymao: this code doesn't make much sense to me, i think if we removed this next statement we could just call the whole function once

lib/engine/ability/train_discount.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class TrainDiscount < Base
+      attr_reader :discount, :absolute, :trains
```
tobymao, line 8: new line

tobymao: ok i was confused, i thought discount absolute meant that the new price is X

maybe we should just make discount an integer subtraction and a float for percentage?

so you can do discount 50 for minus 50$ or 0.5 for 50% off

lib/engine/step/g_18_al/train.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module Engine
+  module Step
+    module G18AL
+      class Train < SingleDepotTrainBuyBeforePhase4
+        def process_buy_train(action)
+          super
+
+          action.entity.abilities(:train_discount) do |ability|
+            if ability.owner&.sym == 'NDY' && action.train.price > action.price
```
tobymao, line 13: let's call id

lib/engine/ability/train_discount.rb
perwestling: If discount > 1 the new price will be old_price - discount
otherwise the new price is old_price - discount*old_price, ie old_price*(1-discount)

Fixed

lib/engine/step/g_18_al/train.rb
perwestling: Fixed

assets/app/view/game/buy_trains.rb
perwestling: This is to get different prices for the same train.
See included picture
Screenshot from 2020-07-23 13-41-46

assets/app/view/game/buy_trains.rb
perwestling: See picture above.
So is it possible to get two lines for the available 3T with different prices, with just one call to the method?

perwestling:
I will push up a new branch where some of the comments have been fixed. 
How to solve the UI problem of buying a train discounted (intentionally choosing the ability) is still an open question.

As the idea is to use Specials for this buy, and that part is undergoing redesign, I think this PR can wait until that is ready.

lib/engine/step/special_train.rb
```diff
@@ -0,0 +1,44 @@
...
+        ACTIONS
+      end
+
+      def blocks?
+        false
+      end
+
+      def process_buy_train(action)
+        buy_train_action(action, action.entity.owner)
+        ability = ability(action.entity)
```
jenf, line 25: This has a problem in EMR that you can sell more shares than needed to use the discount, probably can be done in a follow up PR

perwestling: @jenf Thinking about it I am sceptical that company abilities could be used by the president to affect his/her personal resources. So I do not think the train discount ability would be applicable during an EMF.
Anyway, I have asked a question about it on BGG.

I suggest that we do not support that. I will do a test to see that this special train are not triggered by an emergency buy.

assets/app/view/game/buy_trains.rb
```diff
@@ -54,6 +54,8 @@ def render_president_contributions
       def render
         step = @game.round.active_step
         @corporation = step.current_entity
+        @entity = @selected_company&.owner == @corporation ? @selected_company : @corporation
```
tobymao, line 57: you should only set the entity if the company has train buying abilities, wouldn't the current code allow any company to buy a train?

lib/engine/ability/train_discount.rb
```diff
@@ -0,0 +1,21 @@
...
+      attr_reader :discount, :trains
+      def setup(discount:, trains:)
+        @discount = discount
+        @trains = trains
+      end
+
+      def discounted_price(train, price)
+        return price unless @trains.include?(train.name)
+
+        price - (discount > 1 ? discount : (price * discount))
```
tobymao, line 17: might be safer to do discount.is_a?(Integer), not sure test it

lib/engine/game/base.rb
```diff
@@ -1022,6 +1022,7 @@ def operating_round(round_num)
           Step::Token,
           Step::Route,
           Step::Dividend,
+          Step::SpecialTrain,
```
tobymao, line 1025: this should not be added to the base class

lib/engine/step/train.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 require_relative 'base'
+require_relative 'trainer'
 
 module Engine
   module Step
     class Train < Base
+      include Trainer
```
tobymao, line 9: lol trainer is not a great name, it is like saying this step likes to work out at the gym

maybe TrainBuyer

tobymao: i don't see why this was split out, isn't the only other use casse in the special track subclass?

assets/app/view/game/buy_trains.rb
perwestling: No longer relevant as code changed.

assets/app/view/game/buy_trains.rb
perwestling: In latest versions the only thing that changes in the view is the price shown. So this picture is not relevant any longer.

lib/engine/step/special_train.rb
perwestling: I think we need to support using NDY during EMF, but as you say, it can be done in a future PR.

lib/engine/ability/train_discount.rb
perwestling: Done

assets/app/view/game/buy_trains.rb
perwestling: Have rewritten this part. Now only set @ability here and do an explicit use of entity where it is needed.

lib/engine/game/base.rb
perwestling: Fixed

lib/engine/ability/train_discount.rb
perwestling: But it did not work...  If I check which type/class discount has Ruby says Number.

Anyway, I changed back to using the > operator.

lib/engine/step/train.rb
perwestling: I thought token -> tokener
track -> tracker
train -> trainer....

But in the new commit I'll try step buy_train that uses a module train.

Now I felt I needed having the step + module pattern, which in part was due to the track / token patern, and in part that I did not get it to work subclassing Step::Train.

But if you still think it is a bad idea I can give it a new try.

In any case, isn't Step::BuyTrain a better name than Step::Train?

lib/engine/step/base.rb
```diff
@@ -67,7 +67,9 @@ def active_entities
       end
 
       def round_state
-        {}
+        {
+          trains_bought: [],
```
tobymao, line 71: this should not go in the base step, only in the train buying step

lib/engine/step/train.rb
```diff
@@ -90,8 +58,14 @@ def process_buy_train(action)
 
         @log << "#{entity.name} #{verb} a #{train.name} train for "\
           "#{@game.format_currency(price)} from #{source}"
+        @round.trains_bought << {
+          entity: entity,
+          train: train,
+          from_depot?: train.from_depot?,
```
tobymao, line 64: do you need all these params? it doesn't seem like you use them all

tobymao: why don't we just put this where it's necessary, the single buy, i don't think you need to track all this stuff, just whether or not you bought a train right?

assets/app/view/game/buy_trains.rb
perwestling: Different solution in later version, so not relevant any longer.

lib/engine/step/base.rb
perwestling: Fixed

lib/engine/step/train.rb
perwestling: Have now changed so that trains_bought is only updated by special_train_buy and single_depot_train_buy_before_phase4.

The information is now just entity, which seems to be the only one needed. (Was not sure when the @round state was reset, therefor the extra values.) 
If adding train.sym it would be possible to refactor 1836Jr30 to use this state as well, but in that case one could consider to add methods in train.rb to handle this in a general way.

lib/engine/step/single_depot_train_buy_before_phase4.rb
```diff
@@ -1,30 +1,30 @@
...
-        # Since the train won't be in the depot after being bought store the state now.
         from_depot = action.train.from_depot?
         super
-
         return unless from_depot
 
-        @depot_trains_bought << action.train.sym
-
+        @round.trains_bought << {
+          entity: action.entity,
```
tobymao, line 18: why store a hash, this should be called bought_trains and just be the entity

lib/engine/step/special_buy_train.rb
```diff
@@ -0,0 +1,47 @@
...
+          ability.discounted_price(action.train, action.train.price) == action.price
+        action.entity.close! if ability.count.zero?
+
+        pass! unless can_buy_train?(actual_buyer)
+      end
+
+      def ability(entity)
+        return unless entity.company?
+
+        entity.abilities(:train_discount, 'train')
```
tobymao, line 43: how do you know that this is the train buying phase?

PR #1077, opened 2020-07-12
render non-home reservations as grey instead of black

### LINKED ISSUE #1071: [1846] Differentiate the home station of the Erie and the reserved station in the City of Erie
Right now the two circles look identical on the map and it's not unreasonable for an unfamiliar player to assume that Erie starts in Erie.

michaeljb:
* on /tiles/ pages, use real hex coordinates instead of A1 where applicable;
  without this, all reservations are grey except when the corporation's home
  happens to be A1

[Fixes #1071]

Screenshot from 2020-07-11 18-48-59

Screenshot from 2020-07-11 18-51-23

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -48,8 +48,12 @@ module View
         def reservation
           text = @reservation.id
 
+          non_home = @reservation.corporation? && (@reservation.coordinates != @city.hex.coordinates)
+          color = non_home ? '#808080' : 'black'
+
           attrs = {
-            fill: 'black',
+            fill: color,
+            stroke: color,
             'font-size': "#{RESERVATION_FONT_SIZE[text.size]}px",
             'dominant-baseline': 'central',
           }
--- a/assets/app/view/tiles.rb
+++ b/assets/app/view/tiles.rb
@@ -12,7 +12,8 @@ module View
       location_name: nil,
       scale: 1.0,
       opacity: 1.0,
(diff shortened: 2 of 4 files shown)


JEBrosius:
This looks great, Michael.  Thanks!  I know that the reservations already disappeared in the gray phase, but you might want to check that this continues to be true.  (It's an odd thing -- the discounts for the previously-reserved IC, Erie, B&O, and PRR spaces remain, but the reservations go away in Phase IV.  I don't know what behavior we actually want about that.  But your version is a clear improvement to me.

michaeljb:
Thanks for the feedback @JEBrosius!

The current implementation has the reservation disappearing from the map in Phase IV, but the discount does remain. Unfortunately in the current UI, there's nothing to remind players of the discount. Personally I'd prefer to keep the map's current behavior and have the discount info added to the corporation charter.

assets/app/view/game/part/city_slot.rb
```diff
@@ -48,8 +48,12 @@ def render_part
         def reservation
           text = @reservation.id
 
+          non_home = @reservation.corporation? && (@reservation.coordinates != @city.hex.coordinates)
+          color = non_home ? '#808080' : 'black'
+
           attrs = {
```
tobymao, line 54: can we instead set this only if it's non home? that way the html doesn't get as bloated

michaeljb: 1aa98365

PR #1084, opened 2020-07-12
Allow revenues to be formatted for 1836jr30 ports

### LINKED ISSUE #957: 1836jr30 offboard should have +
Posted this on the list of special rules but was asked to make a bug. Text here refers to Scott’s [post](https://github.com/tobymao/18xx/issues/692) :

Issue on the post’s list is checked as complete, but I’m not seeing/reading that fix on the board:

“ Fix display for port/paris off-maps: it's not at all clear that they have funny rules.”

Maybe implement a +20/+30 as on the printed map?

jenf:
fixes #957


lib/engine/part/revenue_center.rb
```diff
@@ -19,14 +19,25 @@ def initialize(revenue, **opts)
       end
 
       # number, or something like "yellow_30|green_40|brown_50|gray_70|diesel_90"
-      def parse_revenue(revenue)
+      def parse_revenue(revenue, format)
```
tobymao, line 22: format = nil so the tests don't need to pass nil in

jenf: done, should be ready for merge.

Issue #1053, opened 2020-07-09
[1846] 3/5 train with EW + meatpacking incorrect payout


adamranderson:
https://gist.github.com/adamranderson/7dc3e12622e97d21c73079ca38347238

In this game IC is running a 3/5 train from Chicago connections to Windsor. The meat packing token is assigned to the IC and is assigned to Chicago. The game calculates the best run as: Chicago connections (7) - Detroit (5) - Windsor (7) for 19 total.  With meatpacking the route should replace Detroit with Chicago (7) for 21 total.

Spobster:
Please check whether the 3/5 run from IC in this one is fixed as well after fixing the bug. It does not involve the meatpacking company.

https://gist.github.com/Spobster/6bbecab69d3d2fc3d069e4f0debc0daa

(was posted as #1127)

vantdog:
have another issue with a 3/5 and meat packing - involving PRR. 

Game 6005. First time I ran this train - it ran correctly. Paid out Windsor, Chicago (with Meat packing) and Chicago Connections for 210 (in Green phase - so Chicago upgraded to Green). 

Next OR - I tokened in Detroit - and ran the same route with the 3/5. Only paying out 190 that time. I think it paid Detroit instead of Chicago?

-Tim

adamranderson:
Hit this again in a similar way to vantdog. Game data https://gist.github.com/adamranderson/150fcf54a0dbe5cbc6ed92fa2e372379

I suspect that the 3/5 train is picking Detroit ($50) over Chicago ($40) at face value and not seeing the +$30 meat packing token.

adamranderson:
This is working as intended. The 3/5 will only pick up Chicago and the EW run if the company has a token in Chicago. One of the 3 stops must have a token.

From the rules: 
"N/M trains, such as a 3/5 (“three out of five”) train. These may
count any number of cities and off-board areas totaling up to
N towards its revenue (with at least one of them containing a
token belonging to the corporation), while visiting up to M
cities and off-board areas. An N/M train cannot visit more than
M locations."

adamranderson:
Looks like the bug still exists if a token is Chicago. It was only working as intended if the company only has a token in Detroit.

https://gist.github.com/adamranderson/150fcf54a0dbe5cbc6ed92fa2e372379

dfannius:
Fixed with https://github.com/tobymao/18xx/commit/d5437b0f1bef304dc58ecf44e850ce873a9ecdd7.

PR #1085, opened 2020-07-12
Add detail about bonuses and clarify 3/5 trains

### LINKED ISSUE #976: Explain route payouts.
1) When running a 3/5 or similar train in 1846, it isn't obvious which revenue centers are actually being counted.
2) When running a route with bonuses due to privates and/or tokens, there is no announcement as to the amount.

Suggest adding the amount of revenue from each hex next to the coordinates of the hex, and adding an extra item for route-level bonuses.

### LINKED ISSUE #1063: Indicate mail contract bonus when running train and in log
The mail contract should be indicated when setting the route and then later when payout or withhold in the log.

This run looks like it is calculated incorrectly, as a 3/5 should be 40 + 60 + 70 = 170. However, the mail contract bonus is 40.

Here, it should show as part of the on-screen text while selecting cities.

Screenshot 2020-07-10 06 10 05

Here, it should appear in the log.

NYC spends $20 and lays tile #23 with rotation 4 on E15
NYC spends $20 and lays tile #8 with rotation 2 on E13
NYC passes
NYC runs a 3/5 train for $210 (D14, E17, G19, G21)
NYC pays out $210 = $21 x 10 shares
ChrisShaffer receives $105 = $21 x 5 shares
Halekio receives $21 = $21 x 1 shares
Akado receives $21 = $21 x 1 shares
NYC's share price changes from $80 to $100

jenf:
fixes #1063
fixes #976

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -53,6 +53,7 @@ module View
           store(:selected_route, route, skip: true)
         end
 
+        have_bonuses = false
         trains = trains.flat_map do |train|
           onclick = lambda do
             unless (route = @routes.find { |t| t.train == train })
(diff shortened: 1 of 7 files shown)


tobymao:
this pr adds a ton of complexity :/

jenf:
yeah, this makes #1053 harder to fix as well

jenf:
This is probably best left until at least all the known 1846 routing bugs are fixed, then we can see the best way of doing this.

PR #1089, opened 2020-07-13
18AL: Implement M&C RR and S&NA companies

perwestling:
Added route_bonus ability that can be used to give revenue
bonus if a route include the specified hexes.

Although generic, the use of ability implemented in 18AL only.

Configure ability for company Memphis & Charleston RR.

### DIFF
--- a/lib/engine/ability/hex_bonus.rb
+++ b/lib/engine/ability/hex_bonus.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class HexBonus < Base
+      attr_reader :hexes, :amount
+
+      def setup(hexes:, amount:)
+        @hexes = hexes
+        @amount = amount
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_18_al.rb
+++ b/lib/engine/game/g_18_al.rb
@@ -17,6 +17,8 @@ module Engine
       GAME_DESIGNER = 'Mark Derrick'
       GAME_END_CHECK = { bankrupt: :immediate, stock_market: :current_or, bank: :current_or }.freeze
 
+      EVENTS_TEXT = Base::EVENTS_TEXT.merge('remove_tokens' => ['Remove Tokens', 'Coal Field token removed']).freeze
+
       include CompanyPrice50To150Percent
       include Revenue4D
       include TerminusCheck
@@ -35,6 +37,7 @@ module Engine
         Round::Operating.new(self, [
           Step::Bankrupt,
           Step::DiscardTrain,
+          Step::G18AL::Assign,
           Step::BuyCompany,
           Step::HomeToken,
           Step::SpecialTrack,
@@ -51,7 +54,22 @@ module Engine
         # Mobile and Nashville should not be possible to pass through
         ensure_termini_not_passed_through(route, %w[A4 Q2])
 
-        adjust_revenue_for_4d_train(route, super)
+        revenue = adjust_revenue_for_4d_train(route, super)
+
+        route.corporation.abilities(:hexes_bonus) do |ability|
+          revenue += route.stops.sum { |stop| ability.hexes.include?(stop.hex.id) ? ability.amount : 0 }
+        end
+
+        revenue
+      end
+
+      def event_remove_tokens!
+        @corporations.each do |corporation|
+          corporation.abilities(:hexes_bonus) do |a|
+            @log << "#{corporation.name} removes: #{a.description}"
+            corporation.remove_ability(a)
+          end
+        end
       end
 
       def get_location_name(hex_name)
--- a/lib/engine/game/g_18_al.rb
+++ b/lib/engine/game/g_18_al.rb
@@ -19,9 +19,15 @@ module Engine
 
       EVENTS_TEXT = Base::EVENTS_TEXT.merge('remove_tokens' => ['Remove Tokens', 'Coal Field token removed']).freeze
 
+      ROUTE_BONUSES = %i[atlanta_birmingham mobile_nashville].freeze
+
       include CompanyPrice50To150Percent
       include Revenue4D
       include TerminusCheck
+
+      def route_bonuses
+        ROUTE_BONUSES
+      end
 
       def setup
         setup_company_price_50_to_150_percent
@@ -38,7 +44,7 @@ module Engine
           Step::Bankrupt,
           Step::DiscardTrain,
           Step::G18AL::Assign,
-          Step::BuyCompany,
+          Step::G18AL::BuyCompany,
           Step::HomeToken,
           Step::SpecialTrack,
           Step::G18AL::Track,
@@ -48,6 +54,13 @@ module Engine
           Step::SingleDepotTrainBuyBeforePhase4,
           [Step::BuyCompany, blocks: true],
         ], round_num: round_num)
+      end
+
+      def stock_round
+        Round::Stock.new(self, [
+          Step::DiscardTrain,
+          Step::G18AL::BuySellParShares,
+        ])
       end
 
       def revenue_for(route)
@@ -60,7 +73,23 @@ module Engine
           revenue += route.stops.sum { |stop| ability.hexes.include?(stop.hex.id) ? ability.amount : 0 }
         end
 
+        route_bonuses.each do |type|
+          revenue += route_bonus(route, type)
+        end
+
         revenue
+      end
+
+      def routes_revenue(routes)
+        # Ensure we only get each route_bonus at most one time
+        total_revenue = super
+        route_bonuses.each do |type|
+          bonus_amount = routes.first.corporation.abilities(&:amount)
+          times_received = routes.count { |r| route_bonus(r, type).positive? }
+
+          total_revenue -= bonus_amount * (times_received - 1) if times_received > 1
+        end if routes.any?
+        total_revenue
       end
 
       def event_remove_tokens!
@@ -74,6 +103,14 @@ module Engine
 
       def get_location_name(hex_name)
         @hexes.find { |h| h.name == hex_name }.location_name
+      end
+
+      private
+
+      def route_bonus(route, type)
+        route.corporation.abilities(type).sum do |ability|
+          ability.hexes == (ability.hexes & route.hexes.map(&:name)) ? ability.amount : 0
+        end
       end
     end
   end
--- a/lib/engine/game/g_18_al.rb
+++ b/lib/engine/game/g_18_al.rb
@@ -26,8 +26,7 @@ module Engine
 
         @corporations.each do |corporation|
           corporation.abilities(:assign_hexes) do |ability|
-            historical_objective_city_name = @hexes.find { |h| h.name == ability.hexes.first }.location_name
-            ability.description = "Historical objective: #{historical_objective_city_name}"
+            ability.description = "Historical objective: #{get_location_name(ability.hexes.first)}"
           end
         end
       end
@@ -53,6 +52,10 @@ module Engine
         ensure_termini_not_passed_through(route, %w[A4 Q2])
 
         adjust_revenue_for_4d_train(route, super)
+      end
+
+      def get_location_name(hex_name)
+        @hexes.find { |h| h.name == hex_name }.location_name
       end
     end
   end
--- a/lib/engine/step/g_18_al/assign.rb
+++ b/lib/engine/step/g_18_al/assign.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require_relative '../assign'
+
+module Engine
+  module Step
+    module G18AL
+      class Assign < Assign
+        def process_assign(action)
+          company = action.entity
+          @game.game_error("#{company.owner.name} owns no trains") if company.owner.trains.empty?
+
+          target = action.target
+          hexes = company.abilities(:assign_hexes)&.hexes
+          if !@game.loading && @game.graph.reachable_hexes(company.owner).find { |h, _| h.id == target.id }.nil?
+            @game.game_error("#{@game.get_location_name(target.id)} is not reachable")
+          end
+
+          super
+
+          # Add a revenue bonus for the corporation
+          # that will be removed in phase 6
+          ability = Engine::Ability::HexBonus.new(
+            type: :hexes_bonus,
+            description: "Coal Field token: #{@game.get_location_name(target.id)}",
+            hexes: [target.id],
+            amount: 10
+          )
+          company.owner.add_ability(ability)
+          @game
+            .hexes
+            .select { |hex| hexes.include?(hex.name) }
+            .each { |hex| hex.tile.icons = [] }
+        end
+      end
+    end
+  end
+end
--- a/lib/engine/step/g_18_al/buy_company.rb
+++ b/lib/engine/step/g_18_al/buy_company.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require_relative '../buy_company'
+
+module Engine
+  module Step
+    module G18AL
+      class BuyCompany < BuyCompany
+        def process_buy_company(action)
+          super
+
+          return unless action.company.sym == 'M&C'
+
+          rob_bonus, pan_bonus = @game.route_bonuses
+          action.entity.add_ability(create_hexes_bonus_ability(rob_bonus, 'Robert E. Lee', %w[G8 G4], 20))
+          action.entity.add_ability(create_hexes_bonus_ability(pan_bonus, 'Pan American', %w[Q2 A4], 40))
+        end
+
+        private
+
+        def create_hexes_bonus_ability(type, name, hex_names, amount)
+          terminus1, terminus2 = hex_names.map { |n| @game.get_location_name(n) }
+          description = "#{name}: #{terminus1}-#{terminus2} (#{@game.format_currency(amount)})"
+          Engine::Ability::HexBonus.new(type: type, description: description, hexes: hex_names, amount: amount)
+        end
+      end
+    end
+  end
+end
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -192,7 +192,7 @@ module View
         }
 
         revenue = begin
-                    @game.format_currency(active_routes.sum(&:revenue))
+                    @game.format_currency(@game.routes_revenue(active_routes))
                   rescue Engine::GameError
                     '(Invalid Route)'
                   end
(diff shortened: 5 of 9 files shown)


perwestling:
Two things remains for this feature:

1. Ability/Company does not close upon change of presidency
2. Text probably need to be updated to reflect how the implementation is. No assign of chits for trains. Instead this is done automatically. Effect should be the same. But I think text should be updated to reflect this. (This means the text will differ from the actual text in the game.)

perwestling:
Have now fixed so that:
1. M&C RR does close when change of presidency (of owning corporation)
2. Updated M&C RR description to better match how it works in the online

Also, added implementation and configuration for S&NA company.

perwestling:
Think this commit needs a bit of a redesign. And will have to wait for https://github.com/tobymao/18xx/pull/1050 to be accepted.

perwestling:
I have now improved this commit, using abilities in corporations for the revenue bonus things. Am using the almost the same changes to ability/base.rb as the historical objective commit so they should be independent of each other.

lib/engine/ability/hex_bonus.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class HexBonus < Base
+      attr_reader :bonus_name, :hex, :bonus
```
tobymao, line 8: can just call it name

tobymao: maybe bonus would be better renamed to amount?

lib/engine/g_18_al/phase.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative '../phase'
+
+module Engine
```
tobymao, line 5: this is going to be rewritten

lib/engine/game/g_18_al.rb
```diff
@@ -34,6 +43,40 @@ def operating_round(round_num)
...
+          company.owner.add_ability(ability)
+        end
+      end
+
+      def revenue_for(route)
+        revenue = super
+
+        route.corporation.abilities(:hex_bonus) do |ability|
+          bonus = route.stops.sum { |stop| ability.hex == stop.hex.id ? ability.bonus : 0 }
+          revenue += bonus if bonus.positive?
```
tobymao, line 75: why bother check if bonus is positive?

lib/engine/round/g_18_al/stock.rb
```diff
@@ -0,0 +1,45 @@
...
+
+require_relative '../stock'
+
+module Engine
+  module Round
+    module G18AL
+      class Stock < Stock
+        def initialize(entities, game:, **opts)
+          super
+          @bonus_controller = player_controlling_route_bonus_ability
```
tobymao, line 11: this could just be route_bonus_ability&.owner&.owner

lib/engine/round/g_18_al/stock.rb
```diff
@@ -0,0 +1,45 @@
...
+          @game = game
+        end
+
+        def log_pass(entity)
+          begin
+            ability = route_bonus_ability
+            ability.owner.remove_ability(ability.type)
+            @bonus_controller = nil
+            @game.log << "Removed from #{ability.owner.name} due to presidency change: #{ability.name}"
+          end unless @bonus_controller.nil? || @bonus_controller == player_controlling_route_bonus_ability
```
tobymao, line 21: never use unless with multiple statements

if !@bonus_controller && ...

lib/engine/round/operating.rb
```diff
@@ -71,6 +71,10 @@ def recalculate_order
       def teleported?(entity)
         entity.abilities(:teleport)&.find(&:used?)
       end
+
+      def calculate_route_bonus
```
tobymao, line 75: i don't think we need a route bonus

lib/engine/round/route_bonus.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
```
tobymao, line 1: not sure we should have this class

perwestling:
@tobymao : About the route bonus feature. 

In 18AL the M&C RR company gives the ability to add two chits to trains, giving a route bonus if the train travels Nashville to Mobile or Atlanta [$40 bonus] to Birmingham [$20 bonus]. (It is exclusive or as three of those cities are terminus.) The chit can be moved freely and is kept in company if train sold/rusted/discarded to be reassigned. This ability is kept in the owning corporation in the whole game and lost if presidency changes.

I choose to implement this as an ability (route_bonus) which do include the two routes. The ability is transfered to the corporation at purchase of the company, and closure checks is done every stock turn.

Now a complication here is that the route bonus can only occur once for all routes of an OR. It is definitly possible for a corporation to run two trains från Atlanta to Birmingham and then the dividend addition should be a simple $20 not a double. This means I could not do the route bonus addition in the revenue_for. So I added this in the common code to check for the ability and add the bonus to the dividend sum. This is so I can add the bonus if one or more routes comply with the requirement.
This looks a bit weird, e.g. if the corporation run one train only and it gets $170 plus $20 route bonus it will show the route as $170 and the  total dividend as $190.

**So if we should not have the route_bonus ability class, what should we have instead?**

I would also preferred to have 2 instances of the route_bonus, so one could have one ability for route_bonus Atlanta-Birmingham [$20] and one ability for route_bonus Nashville-Mobile. Have an issue for allowing multiple abilities of the same type.

This commit also contains a 2nd company, but in this case it was more straightforward. Here it is just one hex that gives bonus, and that bonus can be applied to all routes that visit that hex. So that bonus is a straight forward hex_bonus. This was easy to add in the 18AL module revenue_for method.

tobymao:
can we instead use the assign action?

to remove the ability, you can use the event system, check out 1846 remove_tokens event

lib/engine/ability/hex_bonus.rb
perwestling: Done

lib/engine/game/g_18_al.rb
perwestling: Done

lib/engine/round/g_18_al/stock.rb
perwestling: Fixed

lib/engine/round/g_18_al/stock.rb
perwestling: Fixed

lib/engine/round/operating.rb
perwestling: I will try to use hex_bonus (which I have renamed to hexes_bonus) for both this cases.

lib/engine/round/route_bonus.rb
perwestling: I will try to use hex_bonus (which I have renamed to hexes_bonus) for both this cases.

lib/engine/g_18_al/phase.rb
perwestling: Have now used the new step structure.

lib/engine/ability/hexes_bonus.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class HexesBonus < Base
```
tobymao, line 7: HexesBonus kind of reads funny, slight preference to just calling this HexBonus with hexes reader

lib/engine/game/g_18_al.rb
```diff
@@ -46,11 +54,86 @@ def operating_round(round_num)
...
+        Round::Stock.new(self, [
+          Step::DiscardTrain,
+          Step::G18AL::BuySellParShares,
+        ])
+      end
+
+      def action_processed(action)
+        super
+        case action
+        when Action::Assign
```
tobymao, line 67: let's hold off on merging this pr until i finish the special action

lib/engine/game/g_18_al.rb
```diff
@@ -46,11 +54,86 @@ def operating_round(round_num)
...
+          company = action.entity
+          target = action.target
+
+          # Add a revenue bonus for the corporation
+          # that will be removed in phase 6
+          ability = Engine::Ability::HexesBonus.new(
+            type: :hexes_bonus,
+            name: 'Coal Field token',
+            display_value: target.id,
+            **{
```
tobymao, line 77: there's no point in doing this splat

you can just do hexes: [target.id], amount: 10 inline

lib/engine/game/g_18_al.rb
```diff
@@ -46,11 +54,86 @@ def operating_round(round_num)
...
+              'hexes' => [target.id],
+              'amount' => 10,
+            }
+          )
+          company.owner.add_ability(ability)
+          @hexes.select { |hex| %w[H3 G4 H5 G6 E6].include?(hex.name) }
```
tobymao, line 83: you're creating this array multiple times, you should make it a constant

lib/engine/game/g_18_al.rb
```diff
@@ -46,11 +54,86 @@ def operating_round(round_num)
...
+            .each { |hex| hex.tile&.icons = [] }
```
tobymao, line 84: are there hexes without tiles?

lib/engine/step/g_18_al/bye_sell_par_shares.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require_relative '../buy_sell_par_shares'
+
+module Engine
+  module Step
+    module G18AL
+      class BuySellParShares < BuySellParShares
+        def initialize
```
tobymao, line 9: you should never ovverride initialize, you can use def setup instead

lib/engine/ability/hexes_bonus.rb
perwestling: Done

lib/engine/game/g_18_al.rb
perwestling: I had some problem when I tried, but I'll give it a new try.

lib/engine/game/g_18_al.rb
perwestling: Done

lib/engine/game/g_18_al.rb
perwestling: I will correct and investigate if it fails.

lib/engine/step/g_18_al/bye_sell_par_shares.rb
perwestling: Done

perwestling:
Have pushed a new version, with the review comments handled.
This PR will have to wait for a rewrite of Specials (as it uses this for assigning the Coal Field token).

lib/engine/step/g_18_al/assign.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Step
+    module G18AL
+      class Assign < Assign
+        def process_assign(action)
+          company = action.entity
+          raise GameError, "#{company.owner.name} owns no trains" if company.owner.trains.empty?
+
+          target = action.target
+          hexes = company.abilities(:assign_hexes)&.hexes
+          reachable = @game.graph.reachable_hexes(company.owner).find { |h, _| h.id == target.id }
```
tobymao, line 15: only run this unless @game.is_loading because this is very expensive

check for other examples

lib/engine/game/g_18_al.rb
```diff
@@ -17,17 +17,24 @@ class G18AL < Base
       GAME_DESIGNER = 'Mark Derrick'
       GAME_END_CHECK = { bankrupt: :immediate, stock_market: :current_or, bank: :current_or }.freeze
 
+      EVENTS_TEXT = Base::EVENTS_TEXT.merge('remove_tokens' => ['Remove Tokens', 'Coal Field token removed']).freeze
+
+      ROUTE_BONUSES = %i[route_bonus_Atlanta_Birmingham route_bonus_Mobile_Nashville].freeze
```
tobymao, line 22: why the funny names? why not just atlanta_birmingham and mobile_nashville

lib/engine/game/g_18_al.rb
```diff
@@ -48,11 +56,66 @@ def operating_round(round_num)
...
+        route_bonuses.each do |type|
+          bonus_received_for_this_type = routes.sum { |r| route_bonus(r, type) }
+          next if bonus_received_for_this_type.zero?
+
+          bonus_amount = 0
+          routes.first.corporation.abilities(type).each { |a| bonus_amount = a.amount }
+          next if bonus_received_for_this_type == bonus_amount
+
+          total_revenue -= bonus_received_for_this_type
+          total_revenue += bonus_amount
```
tobymao, line 95: i'm having trouble understanding this piece of code, could it be cleaned up a bit?

why are you subtracting the bonus received for this type and then adding the bonus amount

lib/engine/step/g_18_al/assign.rb
```diff
@@ -0,0 +1,35 @@
...
+          raise GameError, "#{@game.get_location_name(target.id)} is not reachable" unless reachable
```
tobymao, line 16: if you do a merge with master there's a new method now

@game.raise_error

lib/engine/step/g_18_al/assign.rb
```diff
@@ -0,0 +1,35 @@
...
+          # Add a revenue bonus for the corporation
+          # that will be removed in phase 6
+          ability = Engine::Ability::HexBonus.new(
+            type: :hexes_bonus,
+            description: "Coal Field token: #{@game.get_location_name(target.id)}",
+            hexes: [target.id],
+            amount: 10
+          )
+          company.owner.add_ability(ability)
+          @game.hexes.select { |hex| hexes.include?(hex.name) }
```
tobymao, line 29: i prefer multiline syntax

@game
.hexes
.select...

lib/engine/step/g_18_al/assign.rb
```diff
@@ -0,0 +1,35 @@
...
+            .each { |hex| hex.tile&.icons = [] }
```
tobymao, line 30: doesn't a hex always have a tile? have you seen this break?

lib/engine/step/g_18_al/buy_company.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Step
+    module G18AL
+      class BuyCompany < BuyCompany
+        def process_buy_company(action)
+          super
+
+          return unless action.company.sym == 'M&C'
+
+          # Add two permanent route bonus abilities
+          # that is removed if president changes
```
tobymao, line 15: could also do

rob_bonus, pan_bonus = @game.route_bonuses

lib/engine/step/g_18_al/bye_sell_par_shares.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require_relative '../buy_sell_par_shares'
```
tobymao, line 3: you spelled the file wrong, it's buy

tobymao: and not bye

lib/engine/step/g_18_al/bye_sell_par_shares.rb
```diff
@@ -0,0 +1,50 @@
...
+
+          @log << "#{route_bonus_ability.owner.name} removes route bonuses as presidency changed"
+          president_changed(route_bonus_ability.owner)
+          @controller_at_start = nil
+        end
+
+        private
+
+        def current_controller
+          route_bonus_ability&.owner&.owner
```
tobymao, line 27: i know i told you to do this but there's a better method, route_bonus_ability&.player

lib/engine/step/g_18_al/bye_sell_par_shares.rb
```diff
@@ -0,0 +1,50 @@
...
+          @controller_at_start = current_controller
+        end
+
+        def log_pass(entity)
+          super
+
+          return if @controller_at_start.nil? || @controller_at_start == current_controller
+
+          @log << "#{route_bonus_ability.owner.name} removes route bonuses as presidency changed"
+          president_changed(route_bonus_ability.owner)
```
tobymao, line 20: i'm not sure this is the appropriate method for this, overriding process_pass makes more sense

lib/engine/game/g_18_al.rb
perwestling: Have updated commit to use special action.

lib/engine/step/g_18_al/assign.rb
perwestling: Ok, now this is done very seldom, just after trying to assign one hex connected to a one-time company ability, but I have now added game.is_loading.

lib/engine/game/g_18_al.rb
perwestling: Probably because they existed as an attribute route_bonus before....
By updated now to the shorter names.

lib/engine/game/g_18_al.rb
perwestling: The idea here is that if you have two or three trains that run from Birmingham to Atlanta, you should only receive the Birmingham to Atlanta once.

So I go through all routes and add the bonuses received (which have been added to each route specifically) and if this amount is larger than the bonus one time, I need to deduct the extra amount received.

Now this is a rare occurance, but it might happen. 

Anyway, I have rewritten the code to make it clearer (hopefully)

lib/engine/step/g_18_al/assign.rb
perwestling: Done

lib/engine/step/g_18_al/assign.rb
perwestling: Done

lib/engine/step/g_18_al/assign.rb
perwestling: I have now removed the ampersand. Will change if it breaks.

lib/engine/step/g_18_al/buy_company.rb
perwestling: Done

lib/engine/step/g_18_al/bye_sell_par_shares.rb
perwestling: Done

lib/engine/step/g_18_al/bye_sell_par_shares.rb
perwestling: Done

lib/engine/step/g_18_al/bye_sell_par_shares.rb
perwestling: Done

perwestling: Done

PR #1093, opened 2020-07-14
18AL: Fix revenue calculation for trains

perwestling:
4D now double off-board and cities.

6T and 7T now change phases.

All trains count towns for revenue, but
does not count them for distance.

### DIFF
--- a/lib/engine/config/game/g_18_al.rb
+++ b/lib/engine/config/game/g_18_al.rb
@@ -14,7 +14,7 @@ module Engine
    "currencyFormatStr":"$%d",
    "bankCash":8000,
    "certLimit":{
-      "3":25,
+      "3":15,
       "4":12,
       "5":10
    },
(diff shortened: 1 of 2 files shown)


perwestling:
I forgot to add the files.... New try.

jenf:
Given that 1828 and 18C2C also have the concept of Doubling trains, would recommend a generalized solution for this. Depending on if @tobymao  agrees of course :)

perwestling:
I suppose it would be possible to generalize trains but it is a bit of a mess....

But I guess a simple variant could be to add a "double" field in the distance.
Default value false. If used the revenue for that type of revenue centers would be doubled.

So 4D definition in 18AL would look like:
```
      {
         "name":"4D",
         "distance":4,
         "distance":[
            {
               "nodes":[
                  "city",
                  "offboard"
               ],
               "pay":4,
               "visit":4.
               "double":true
            },
            {
               "nodes":[
                  "town"
               ],
               "pay":99,
               "visit":99
            }
         ],
         "price":800,
         "num":5
      }
```

Issue #1096, opened 2020-07-15
Improvement: Support non-passable citites in off-board hexes


perwestling:
In some 18xx there are off-board hexes that contains cities that are one-way; they can only be terminus of routes.

Example of such:
- Nashville and Mobile in 18AL
- Sarajevo in 1824
- London in 1862

It would be nice if hexes could be defined to allow for this. Preferably the tracks into the hex should look like they do for off board hexes.

(For 18AL I will do a workaround, letting the game check for invalid route, but it will not be obvious on the map that pass-through rotes are not allowed.)

perwestling:
Terminal attribute added to hex configuration.
Supported in 18xx.games, and used in 18AL and 18MEX.

PR #1098, opened 2020-07-15
18AL: Nashville/Mobile must be terminus if in route

perwestling:
Workaround for 18AL to the problem described in #1096

### DIFF
--- a/lib/engine/game/g_18_al.rb
+++ b/lib/engine/game/g_18_al.rb
@@ -34,6 +34,22 @@ module Engine
           [Step::BuyCompany, blocks: true],
         ], round_num: round_num)
       end
+
+      def revenue_for(route)
+        ensure_terminus_first_or_last(route, %w[A4 Q2])
+
+        super
+      end
+
+      private
+
+      def ensure_terminus_first_or_last(route, termini)
+        route.hexes[1...-1].each do |hex|
+          next unless termini.include? hex.name
+
+          raise GameError, "#{hex.location_name} must be first or last in route"
+        end
+      end
     end
   end
 end


jenf:
Looks good to me, waiting for Toby

perwestling:
I think this one and #1093 can be merged (if they are approved). It is trivial to merge them, just put the terminus check of this PR first in revenue_for, then super, and last the handling of 4D.

lib/engine/game/g_18_al.rb
```diff
@@ -19,6 +19,22 @@ class G18AL < Base
...
+        ensure_terminus_first_or_last(route, %w[A4 Q2])
+
+        super
+      end
+
+      private
+
+      def ensure_terminus_first_or_last(route, termini)
+        termini.each do |terminus|
+          index = route.hexes.index { |hex| hex.name == terminus }
```
jenf, line 33: Cleaner doing
route.hexes[1...-1].map(&:name).include?(terminus)
or even
(route.hexes[1...-1].map(&:name) || termini).empty?
The above removes the need for the loop

(Note the three dots)

perwestling: Now rewritten using & operation for arrays.
Yes, it looks better.

lib/engine/game/g_18_al.rb
```diff
@@ -34,6 +34,22 @@ def operating_round(round_num)
...
+        ensure_terminus_first_or_last(route, %w[A4 Q2])
+
+        super
+      end
+
+      private
+
+      def ensure_terminus_first_or_last(route, termini)
+        route.hexes[1...-1].each do |hex|
+          next unless termini.include? hex.name
```
tobymao, line 48: @jenf @perwestling parens around this

lib/engine/game/g_18_al.rb
```diff
@@ -34,6 +34,22 @@ def operating_round(round_num)
           [Step::BuyCompany, blocks: true],
         ], round_num: round_num)
       end
+
+      def revenue_for(route)
+        ensure_terminus_first_or_last(route, %w[A4 Q2])
```
tobymao, line 39: this doesn't need to be a new method, just put it all in line

lib/engine/game/g_18_al.rb
perwestling: Where do you want parentheses?

lib/engine/game/g_18_al.rb
perwestling: I can refactor this in the revenue commit.

lib/engine/game/g_18_al.rb
tobymao: include?(hex.name)

PR #1101, opened 2020-07-15
Improve game end: current OR and/or stock market reach max

perwestling:
Added GAME_END which can be set to:
 :full_round    - Complete full OR set when end game triggered
 :current_round - Complete current OR where end game triggered
Default value is :full_round

Added GAME_END_CHECK_STOCKMARKET which is a boolean, default false.

Implemented a check for stock market max price, and use this
for checking end game condition (if GAME_END_CHECK_STOCKMARKET
is true).

Change so that if GAME_CHECK is :current_round a game end
triggered game will end in the current OR.

Configure 18AL to use stock market end game trigger, and
to end in current OR if bank runs out or stock market maxed.

Improved IPO price order so that it is always ascending.

TODO: There are more Mark Derrick games that could use this
configuration.
TODO: No test has been implemented for this.


lib/engine/game/base.rb
```diff
@@ -42,6 +42,8 @@ class Base
       GAME_RULES_URL = nil
       GAME_DESIGNER = nil
       GAME_PUBLISHER = nil
+      GAME_END_CHECK_STOCKMARKET = false
```
jenf, line 45: Could we generalize this as a hash
{stockmarket: :full_round, bank: :full_round} and apply it to all games

perwestling: Good idea. I did that and also rewrote the code handling this a bit. I think it now will be easier to extend it with more variants.

Have also added light blue squares to the stock market (a la 1849) for game end triggering squares. It is now possible to have multiple such squares with different values.

perwestling: Hm... the light blue of 1849 was a different meaning....  So I might need another color for game end trigger in stock market.

lib/engine/game/base.rb
```diff
@@ -43,6 +43,13 @@ class Base
       GAME_DESIGNER = nil
       GAME_PUBLISHER = nil
 
+      # Allowed game end check values, if triggered:
+      #  none - selected trigger does not cause game to end
+      #  immediate - ends in current turn
+      #  current_or - ends at the next end of an OR
+      #  full_or - ends at the next end of a complete OR set
+      GAME_END_CHECK = { bankrupt: :immediate, stockmarket: :none, bank: :full_or }.freeze
```
tobymao, line 51: stock_market

lib/engine/game/base.rb
```diff
@@ -818,11 +825,7 @@ def total_rounds
       end
 
       def next_round!
-        if (_reason, after = game_end_reason)
-          if after != :full_round || (@round.is_a?(Round::Operating) && @round.round_num == @operating_rounds)
-            return end_game!
-          end
-        end
+        return end_game! unless game_end_check.first.nil?
```
tobymao, line 828: if game_end_check.first

lib/engine/game/base.rb
```diff
@@ -849,19 +852,43 @@ def next_round!
         @round_history << @actions.size
       end
 
+      def game_end_check
+        when_points = [:immediate]
```
tobymao, line 856: you can just call this when

lib/engine/game/base.rb
```diff
@@ -849,19 +852,43 @@ def next_round!
...
+        when_points << :current_or if @round.is_a?(Round::Operating)
+        when_points << :full_or if @round.is_a?(Round::Operating) && @round.round_num == @operating_rounds
+
+        triggers = {
+          bankrupt: -> { bankruptcy_limit_reached? },
+          bank: -> { @bank.broken? },
+          stockmarket: -> { @stock_market.max_reached? },
```
tobymao, line 863: there's no need to make these lambdas, just leave them booleans

lib/engine/game/base.rb
```diff
@@ -849,19 +852,43 @@ def next_round!
...
+        }
```
tobymao, line 864: triggers = {
...
}.select(&:second)

lib/engine/game/base.rb
```diff
@@ -849,19 +852,43 @@ def next_round!
...
+
+        when_points.each do |when_point|
+          triggers.keys.each do |what|
+            next if self.class::GAME_END_CHECK[what] != when_point
+
+            return [what, when_point] if triggers[what].call
```
tobymao, line 870: don't need the call

lib/engine/game/base.rb
```diff
@@ -849,19 +852,43 @@ def next_round!
...
+          end
+        end
+
+        [nil, nil]
+      end
+
       def game_ending_description
-        reason, after = game_end_reason
-        return unless after
+        game_end_check_result = game_end_check
```
tobymao, line 878: reason, when  = game_end_check

lib/engine/game/base.rb
```diff
@@ -849,19 +852,43 @@ def next_round!
...
+        reason = game_end_check_result.first
+        after = game_end_check_result.last
+        return if reason.nil?
```
tobymao, line 881: return unless reason

lib/engine/game/base.rb
perwestling: Fixed next push

lib/engine/game/base.rb
perwestling: Fixed next push

lib/engine/game/base.rb
perwestling: Fixed next push

lib/engine/game/base.rb
perwestling: Looks better now. Fixed next push

lib/engine/game/base.rb
perwestling: Fixed next push

lib/engine/game/base.rb
perwestling: Fixed next push

lib/engine/game/base.rb
```diff
@@ -746,19 +749,43 @@ def next_round!
...
+          end
+        end
+
+        [nil, nil]
+      end
+
       def game_ending_description
-        reason, after = game_end_reason
-        return unless after
+        game_end_check_result = game_end_check
```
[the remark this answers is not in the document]
perwestling: Fixed next push

lib/engine/game/base.rb
```diff
@@ -746,19 +749,43 @@ def next_round!
...
+
+        when_points.each do |when_point|
+          triggers.keys.each do |what|
+            next if self.class::GAME_END_CHECK[what] != when_point
+
+            return [what, when_point] if triggers[what].call
+          end
+        end
+
+        [nil, nil]
```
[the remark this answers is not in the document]
perwestling: Fixed next push

lib/engine/game/base.rb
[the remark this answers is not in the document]
perwestling: Fixed next push

lib/engine/game/base.rb
tobymao: yea after is good

lib/engine/stock_market.rb
```diff
@@ -19,6 +19,7 @@ def initialize(market, unlimited_colors, multiple_buy_colors: [])
           price
         end
       end
+      @par_prices.sort_by!(&:price).reverse!
```
tobymao, line 22: why is this being sorted? makes sense but, it's not enough to sort by price, you also need to sort by the indicies since two par prices can have the same value

perwestling: If I choose 18AL from master, when I buy the presidency share from market it will look like this:

Par Price:
$60   $75   $70    $105    $90

After sort it instead looks like:
$60   $70   $75   $90   $105

But I have now added so it is sorted by price, column, row - reversed.

PR #1108, opened 2020-07-18
Multi revenue with user defined bg colors

yzemaze:


assets/app/view/game/game_info.rb
```diff
@@ -47,12 +47,13 @@ def phases
         current_phase = @game.phase.current
         rows = @game.phase.phases.map do |phase|
           phase_color = Array(phase[:tiles]).last
-          phase_props = { style: {} }
-          if Part::MultiRevenue::COLOR.include?(phase_color)
```
yzemaze, line 51: The if seems obsolete to me, because we don’t have any game where phase_color could be nil. And if there’d be one in the future a rewrite of this part would be necessary anyway.

tobymao:
@yzemaze does this work with yoru latest commit?

yzemaze:
> @yzemaze does this work with yoru latest commit?

Now it does.

PR #1110, opened 2020-07-18
separate user settings from lib/color

yzemaze:
some overdue cleanup, some groundwork for user defined route colors

### DIFF
--- a/assets/app/view/user.rb
+++ b/assets/app/view/user.rb
@@ -2,17 +2,21 @@
 
 require 'game_manager'
 require 'user_manager'
+require 'lib/settings'
 require 'view/game_row'
 require 'view/logo'
 require 'view/form'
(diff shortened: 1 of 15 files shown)


assets/app/view/user.rb
```diff
@@ -2,18 +2,22 @@
...
 module View
   class User < Form
-    include Lib::Color
+    include Lib::Settings
     include GameManager
     include UserManager
 
     needs :type
 
+    DARK = `window.matchMedia('(prefers-color-scheme: dark)').matches`.freeze
```
tobymao, line 18: why is this copied from settings?

yzemaze: It’s the other way round, but obsolete anyway – just like a bit more.

PR #1126, opened 2020-07-20
ease number input on mobile

yzemaze:
On mobile backspace didn’t function as expected. (Workaround by selecting number and then changing is a pita.)
Allowing the fields to be temporary empty prevents this, but necessitates "validation" on mode switch.
Empty max_players gets catched for multiplayer games (according to my tests) and doesn’t matter for hotseat, because player inputs don’t get wiped. `required` had no effect apart from highlighting the field if empty on Firefox.

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -70,8 +70,10 @@ module View
 
       enforce_range = lambda do
         elm = Native(@inputs[:max_players]).elm
-        elm.value = elm.max.to_i unless (elm.min.to_i..elm.max.to_i).include?(elm.value.to_i)
-        store(:num_players, elm.value.to_i) if @mode == :hotseat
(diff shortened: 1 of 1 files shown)


assets/app/view/create_game.rb
```diff
@@ -70,8 +70,10 @@ def render_inputs
 
       enforce_range = lambda do
         elm = Native(@inputs[:max_players]).elm
-        elm.value = elm.max.to_i unless (elm.min.to_i..elm.max.to_i).include?(elm.value.to_i)
-        store(:num_players, elm.value.to_i) if @mode == :hotseat
+        unless elm.value.empty?
```
tobymao, line 73: if elm.value.any?

yzemaze: thx, I can’t remember this … (tried present? …)

PR #1128, opened 2020-07-20
settings: route colors, width & dash

yzemaze:
I’m no fan of dasharray, but if we’re serious about making the site friendlier to people with visual impairments color and width very likely won’t suffice.

width.to_i because user-settings from inputs are saved as string afaik. I don’t know where/how exactly I could change this for specific inputs.
`4.times` in models/user.rb is ugly – couldn’t figure out how to access Lib::Settings::ROUTE_COLORS there.
no hash with route settings because of problems with settings-array in models/user.rb and accessing values with color_for/settings_for. I wouldn’t mind rewriting these parts, but would likely need a lot of help.

### DIFF
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -5,6 +5,15 @@ require 'lib/hex'
 module Lib
   module Settings
     DARK = `window.matchMedia('(prefers-color-scheme: dark)').matches`.freeze
+    # http://mkweb.bcgsc.ca/colorblind/ 12 color palette
+    ROUTE_COLORS = %i[#A40122 #008DF9 #00FCCF #FF5AAF].freeze
+
+    routes = {}
+    ROUTE_COLORS.each_with_index do |color, n|
+      routes["r#{n}_color"] = color
+      routes["r#{n}_dash"] = '0'
+      routes["r#{n}_width"] = 8
+    end
 
     SETTINGS = {
       notifications: true,
@@ -14,6 +23,7 @@ module Lib
       font: DARK ? '#ffffff' : '#000000',
       font2: '#000000',
       **Lib::Hex::COLOR,
+      **routes,
     }.freeze
 
     def self.included(base)
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -8,8 +8,8 @@ module Lib
     # http://mkweb.bcgsc.ca/colorblind/ 12 color palette
     ROUTE_COLORS = %i[#A40122 #008DF9 #00FCCF #FF5AAF].freeze
 
-    ROUTES = ROUTE_COLORS.flat_map.with_index do |color, n|
-      [["r#{n}_color", color], ["r#{n}_dash", '0'], ["r#{n}_width", 8]]
+    ROUTES = ROUTE_COLORS.flat_map.with_index do |color, index|
+      [["r#{index}_color", color], ["r#{index}_dash", '0'], ["r#{index}_width", 8]]
     end.to_h
 
     SETTINGS = {
@@ -37,8 +37,12 @@ module Lib
 
     alias color_for setting_for
 
-    def route_prop(n, prop)
-      setting_for("r#{n}_#{prop}")
+    def route_prop(index, prop)
+      setting_for(route_prop_string(index, prop))
+    end
+
+    def route_prop_string(index, prop)
+      "r#{index}_#{prop}"
     end
   end
 end
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -8,12 +8,9 @@ module Lib
     # http://mkweb.bcgsc.ca/colorblind/ 12 color palette
     ROUTE_COLORS = %i[#A40122 #008DF9 #00FCCF #FF5AAF].freeze
 
-    routes = {}
-    ROUTE_COLORS.each_with_index do |color, n|
-      routes["r#{n}_color"] = color
-      routes["r#{n}_dash"] = '0'
-      routes["r#{n}_width"] = 8
-    end
+    ROUTES = ROUTE_COLORS.flat_map.with_index do |color, n|
+      [["r#{n}_color", color], ["r#{n}_dash", '0'], ["r#{n}_width", 8]]
+    end.to_h
 
     SETTINGS = {
       notifications: true,
@@ -23,7 +20,7 @@ module Lib
       font: DARK ? '#ffffff' : '#000000',
       font2: '#000000',
       **Lib::Hex::COLOR,
-      **routes,
+      **ROUTES,
     }.freeze
 
     def self.included(base)
@@ -39,5 +36,9 @@ module Lib
     end
 
     alias color_for setting_for
+
+    def route_prop(n, prop)
+      setting_for("r#{n}_#{prop}")
+    end
   end
 end
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -32,14 +32,14 @@ module View
           if @tile.offboards.any?
             @tile.paths.select(&:offboard).map do |path|
               h(TrackOffboard, offboard: path.offboard, path: path, region_use: @region_use,
-                               color: prop_for(path, :color), width: prop_for(path, :width),
-                               dash: prop_for(path, :dash),)
+                               color: value_for(path, :color), width: value_for(path, :width),
+                               dash: value_for(path, :dash),)
             end
           elsif @tile.lawson?
             @tile.paths.select { |path| path.edges.one? }.map do |path|
               h(TrackLawsonPath, path: path, region_use: @region_use,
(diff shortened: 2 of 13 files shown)


assets/app/view/game/part/track.rb
```diff
@@ -86,12 +88,12 @@ def index_for(path)
...
 
-        def color_for_index(index)
-          index ? self.class::ROUTE_COLORS[index] : TRACK_COLOR
+        def prop_for_index(index, prop)
+          index ? setting_for("r#{index}_#{prop}") : TRACK[prop]
         end
 
-        def color_for(path)
-          color_for_index(index_for(path))
+        def prop_for(path, prop)
```
tobymao, line 95: change all of these to props

assets/app/view/user.rb
```diff
@@ -89,9 +91,17 @@ def reset_settings
       input_elm(:bg2).value = default_for(:bg2)
       input_elm(:font2).value = default_for(:font2)
       input_elm(:red_logo).checked = false
+
       TILE_COLORS.each do |color, hex_color|
         input_elm(color).value = hex_color
       end
+
+      ROUTE_COLORS.each_with_index do |hex_color, index|
```
tobymao, line 99: this looks copy pasted

models/user.rb
```diff
@@ -10,9 +10,13 @@ class User < Base
 
   RESET_WINDOW = 60 * 15 # 15 minutes
 
-  SETTINGS = %w[
-    notifications bg font bg2 font2 red_logo white yellow green brown gray red blue
-  ].freeze
+  opts = %w[notifications red_logo bg font bg2 font2 white yellow green brown gray red blue]
```
tobymao, line 13: just do SETTINGS = [].map

assets/app/view/game/part/track.rb
yzemaze: ?

assets/app/view/user.rb
yzemaze: ?

models/user.rb
yzemaze: ?

assets/app/view/game/part/track.rb
tobymao: %s/prop/props/g

assets/app/view/user.rb
tobymao: how is this different from the top part that looks the same? in lib/settings

models/user.rb
tobymao: rewrite this to do it all inline

using the form SETTINGS = %w[...].map

assets/app/view/user.rb
yzemaze: this resets the inputs to defaults

models/user.rb
yzemaze: hm, but I don’t want to map these settings, they are ok as is

tobymao: what? you don't use opts anywhere else, it's used to populate settings, instead of creating opts and mutating it, create SETTINGS in a functional style

assets/app/view/game/part/track.rb
yzemaze: on second thought: `value_for(path, prop)` might be more adequate
yep still prop, because it asks for _a_ property, not multiple properties

tobymao: sounds good

assets/app/lib/settings.rb
```diff
@@ -5,6 +5,15 @@
 module Lib
   module Settings
     DARK = `window.matchMedia('(prefers-color-scheme: dark)').matches`.freeze
+    # http://mkweb.bcgsc.ca/colorblind/ 12 color palette
+    ROUTE_COLORS = %i[#A40122 #008DF9 #00FCCF #FF5AAF].freeze
+
+    routes = {}
```
tobymao, line 11: can we again not use lower case for constants, and similarly do

ROUTES = ROUTE_COLERS.flat_map.with_index do |color, n|
 [["r#{n}_color, color]...]
end.to_h

something like that

assets/app/lib/settings.rb
```diff
@@ -5,6 +5,15 @@
...
+    ROUTE_COLORS.each_with_index do |color, n|
+      routes["r#{n}_color"] = color
```
tobymao, line 13: since you use these keys later on, i think we should make a helper function to get them so if we decide to change it it's not too error prone

models/user.rb
```diff
@@ -10,9 +10,12 @@ class User < Base
 
   RESET_WINDOW = 60 * 15 # 15 minutes
 
-  SETTINGS = %w[
-    notifications bg font bg2 font2 red_logo white yellow green brown gray red blue
-  ].freeze
+  SETTINGS = Array.new(4) do |n|
+    %w[color dash width].map do |prop|
```
tobymao, line 14: again here we make a similar string

models/user.rb
```diff
@@ -10,9 +10,12 @@ class User < Base
 
   RESET_WINDOW = 60 * 15 # 15 minutes
 
-  SETTINGS = %w[
-    notifications bg font bg2 font2 red_logo white yellow green brown gray red blue
-  ].freeze
+  SETTINGS = Array.new(4) do |n|
```
tobymao, line 13: why are you make a size 4 array? 

maybe you wanted to do 4.time.flat_map ?

models/user.rb
```diff
@@ -10,9 +10,12 @@ class User < Base
...
+    %w[color dash width].map do |prop|
+      "r#{n}_#{prop}"
+    end
+  end
+  SETTINGS.concat(%w[notifications red_logo bg font bg2 font2 white yellow green brown gray red blue]).flatten!.freeze
```
tobymao, line 18: you shouldn't mutate constants

models/user.rb
yzemaze: I had `4.times.flat_map` there, but stumbled upon a benchmark somewhere stating Array.new(n) is significantly more performant than n.times.map. But flat_map doesn’t need flatten!, so here we go …

yzemaze: Is there a way to access Lib::Settings::ROUTE_COLORS here? (… to get rid of hardcoded 4)

assets/app/view/user.rb
```diff
@@ -89,9 +91,17 @@ def reset_settings
...
+        input_elm("r#{index}_color").value = hex_color
```
tobymao, line 100: use your function that you added here?

models/user.rb
```diff
@@ -10,9 +10,12 @@ class User < Base
...
 
-  SETTINGS = %w[
-    notifications bg font bg2 font2 red_logo white yellow green brown gray red blue
-  ].freeze
+  SETTINGS = 4.times.flat_map do |n|
+    %w[color dash width].map do |prop|
+      "r#{n}_#{prop}"
+    end
+  end
+  SETTINGS.concat(%w[notifications red_logo bg font bg2 font2 white yellow green brown gray red blue]).freeze
```
tobymao, line 18: you're mutating a constant here SETTINGS.concat mutates

assets/app/view/user.rb
yzemaze: I’m not reading sth. there. This is just resetting the input-value.

tobymao: i don't want "r#{index_x}" everywhere in the code can we clean that up

models/user.rb
yzemaze: so
SET1 = 4.times …
SET2 = %w[…]
SETTINGS = SET1 + SET2

… or sth else?

tobymao: Settiings = 4.times... + %w[...]

models/user.rb
```diff
@@ -10,9 +10,12 @@ class User < Base
 
   RESET_WINDOW = 60 * 15 # 15 minutes
 
-  SETTINGS = %w[
-    notifications bg font bg2 font2 red_logo white yellow green brown gray red blue
-  ].freeze
+  SETTINGS = 4.times.flat_map do |n|
+    %w[color dash width].map do |prop|
+      "r#{n}_#{prop}"
```
tobymao, line 15: i don't want "r#{n}_#{props}" in the code if possible

assets/app/view/user.rb
yzemaze: The inputs need to have a unique id. This is the easiest solution to generate one known to me and easy to understand, too.

tobymao: can you create a helper function that generates this key?

models/user.rb
yzemaze: Well, the settings do need a name. The other option would be hashes, but as I already wrote in my PR comment: 

> no hash with route settings because of problems with settings-array in models/user.rb and accessing values with color_for/settings_for. I wouldn’t mind rewriting these parts, but would likely need a lot of help.

PR #1131, opened 2020-07-20
Rewrite SR to use steps

jenf:
I haven't yet run this over the database, but the UI seems to work. I'll run the database tomorrow.

I've tested Stock rounds, EMR

The one big quirk I found was in spec/lib/engine/games/game_spec.rb 

This is due to the fact that the player had an option on which shares to sell
https://www.18xx.games/game/247
The code creates entries in the hash map shares_by_corporation which causes the each in bankrupcy to be done in a different order causing a different run.


tobymao:
lgtm, i'll let you merge after you've double checked the games

jenf:
11 games broke. Will need to examine them all to see why

jenf:
Fixed two bugs.

We are likely to need another migration, as the old code forced users to pass when they had no valid actions
(e.g. https://www.18xx.games/game/603?action=90)

PR #1142, opened 2020-07-21
pruning + token popup

### LINKED ISSUE #1122: Feature request - Allow expanding market or include operating order information on charters
Sometimes enough company tokens can end up on a space on the market to obscure them. Operating order can be important to know during the stock round. If the space on the market could be expanded or the current operating order included on the charters would clear things up.

image

yzemaze:
fixes #1122 (partially?)
on hover/tap:
image => image

Additionally some pruning. I could do more, but don’t know about your intentions regarding the load of CONSTs. If these values won’t be changed in the future we could eliminate at least some of them. (… and maybe even switch to .box and css to en passant eliminate the style mergings.)

What’s the purpose of empty boxes without borders? https://github.com/tobymao/18xx/blob/master/assets/app/view/game/stock_market.rb#L85

PS: How about reducing the width of empty boxes on pretty wide 1D markets like 1846?

### DIFF
--- a/public/assets/main.css
+++ b/public/assets/main.css
@@ -351,3 +351,26 @@ div.game.card em {
   font-style: normal;
   text-decoration: underline;
 }
+
+#stock_market .splayed {
+  display: flex;
+  flex-direction: row-reverse;
+  position: absolute;
+  left: 2px;
+  z-index: 1000;
+  padding: 2px;
+  height: 33px;
+  box-sizing: border-box;
+  border: 2px solid;
+  border-color: inherit;
+  border-radius: 5px;
+  box-shadow: 2px 2px #555555;
+  background-color: inherit;
+}
+#stock_market object {
+  position: absolute;
+}
+#stock_market .splayed object {
+  position: static;
+  padding-right: 1px;
+}
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'lib/settings'
+require 'view/game/bank'
 require 'view/game/token'
 
 module View
@@ -11,6 +12,7 @@ module View
       needs :game
       needs :show_bank, default: false
       needs :explain_colors, default: false
+      needs :splayed, default: nil, store: true
 
       COLOR_MAP = {
         red: '#ffaaaa',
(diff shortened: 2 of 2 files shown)


dfannius:
I put in all the CONSTs when writing the original code for token spacing because I find that explicitly naming and computing all the intermediate values makes tricky code a lot more maintainable for me (otherwise I basically have to rederive the formulas every time I visit the code). I won't be offended if you get rid of some of them.

tobymao:
how does clicking work?

yzemaze:
> how does clicking work?

hover/tap

tobymao:
i don't see any javascript that would allow tapping to work

yzemaze:
> i don't see any javascript that would allow tapping to work

There’s no need for JS. hover => tapping on mobile, the diff (to desktop) is just that you need to tap again (anywhere) to let it disappear.

tobymao:
since when do iphones support hover?

yzemaze:
> I put in all the CONSTs when writing the original code for token spacing because I find that explicitly naming and computing all the intermediate values makes tricky code a lot more maintainable for me (otherwise I basically have to rederive the formulas every time I visit the code). I won't be offended if you get rid of some of them.

Yeah, I got that – and these would be helpful if needed. Question is: Will we need this in the future?

yzemaze:
> since when do iphones support hover?

oh well, iOS, Safari especially … according to browserstack it works on Chrome on iOS13 and on iOS12 on Safari, too.
Given that this is a minute detail (niche³) I’m reluctant to waste more time on it, but I’ll look for a dedicated Safari iOS13 solution later on.

tobymao:
just make it tap only, and no hover

public/assets/main.css
```diff
@@ -339,3 +339,26 @@ div.game.card em {
   font-style: normal;
   text-decoration: underline;
 }
+
+#stock_market .splayed {
```
tobymao, line 343: i don't think this css is necessary here is opposed to a prop

yzemaze: The class splayed is toggled. I’d have to toggle prop, don’t know if that works as easily.

tobymao:
why not? how is it different from anything else?

tobymao:
sorry, after revisiting, i don't think this is the best option, i don't think it's obvious you need to click on it, the order is already displayed. for 1D stock markets where this will happen more often, I think we should have a taller market, i'm going to close this for now

yzemaze:
> sorry, after revisiting, i don't think this is the best option, i don't think it's obvious you need to click on it, the order is already displayed. for 1D stock markets where this will happen more often, I think we should have a taller market, i'm going to close this for now

Yep, it was more or less just nice to have and might not be the best option (depends on your use case). But it was "good enough" and your arguments can’t convince me. It’s not obvious, but so are using abilites or rotating tiles. Order isn’t displayed on market tab. The latter might be a good way though to help with crowded boxes for all markets. There’s enough space next to Bank cash …

PR #1146, opened 2020-07-22
Add 1882 Initial auction rules, fix 18Chesapeake train export during initial auction

### LINKED ISSUE #1092: If OR1 happens before all private companies are sold, then export a train
I am not proud to submit this bug, but in the interest of having a perfect implementation of 18Chesapeake rules, here it is.

If all the players pass during the initial stock round after a private company has been purchased, we run an OR. After this OR, a train should be exported. So this screenshot is correct, but also there should be an export which doesn't happen currently.
image

jenf:
Fixes #1092

Also implements 1882
The first available company reduces by $5, not just the first.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -605,6 +605,8 @@ module Engine
         end
       end
 
+      def or_set_finished; end
+
       def place_home_token(corporation)
         return unless corporation.next_token # 1882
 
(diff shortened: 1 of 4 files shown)


jenf:
Looks like 50 games hit the chesapeake case, so will need a pinning.

tobymao:
will let you handle merge and pin

PR #1150, opened 2020-07-23
1836jr oo tile rotations

### LINKED ISSUE #962: Incorrect tile lay for Brown OO in E5 Rotterdam
E5 Rotterdam Brown OO tile allowed place a tile that moved one of the cities and should not be allowed. 
See https://www.18xx.games/game/3234

jenf:
fixes #962


lib/engine/tile.rb
```diff
@@ -274,6 +274,19 @@ def token_blocked_by_reservation?(corporation)
       end
     end
 
+    def city_town_edges
+      # Returns a list of each edge a city/town goes to
```
tobymao, line 278: if the intention was to return a list, maybe you should do values at the end here intsead of calling it on 206 and 210 of the other file

PR #1158, opened 2020-07-24
Fix/issue 1157

perwestling:
Presentation has now been improved:
Screenshot from 2020-07-24 13-02-21

Made some changes in the ability presentation:
- Do not show abilities without name
- Make presentation layout more dynamic
The reason for having abilities without name is to hide them from presentation. In some places an ability can be represented as multiple attributes internally.

Also, it will also mean that adding abilities to appear will be an intentional action.

The dynamic part of presentation was created by @yzemaze

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -48,8 +48,10 @@ module View
           children << h(Companies, owner: @corporation, game: @game) if @corporation.companies.any?
         end
 
-        abilities = @corporation.all_abilities.select { |ability| ability.owner.corporation? }
-        children << render_abilities(abilities) if abilities&.any?
+        abilities_to_display = @corporation.all_abilities.select do |ability|
+          ability.owner.corporation? && ability.description
+        end
+        children << render_abilities(abilities_to_display) if abilities_to_display.any?
 
         if @corporation.owner
           subchildren = (@corporation.operating_history.empty? ? [] : [render_revenue_history])
@@ -308,23 +310,21 @@ module View
 
       def render_abilities(abilities)
         attribute_lines = []
+        props = { style: { display: 'inline-block' } }
 
         abilities.each do |ability|
-          attribute_lines << h('div.name.nowrap', ability.name)
-          attribute_lines << h('div.right', ability.display_value || '')
+          attribute_lines << h('div.nowrap', props, ability.description)
         end
 
         table_props = {
           style: {
-            padding: '0 0.5rem',
-            grid: 'auto / 3fr 2fr',
-            gap: '0 0.2rem',
+            padding: '0.5rem',
+            justifyContent: 'center',
           },
         }
 
-        h('div#company_table', table_props, [
-          h('div.bold', 'Attribute'),
-          h('div.bold.right', 'Value'),
+        h('div#attribute_table', table_props, [
+          h('div.bold', 'Ability'),
           *attribute_lines,
         ])
       end
--- a/lib/engine/config/game/g_1846.rb
+++ b/lib/engine/config/game/g_1846.rb
@@ -399,6 +399,7 @@ module Engine
       "abilities": [
         {
           "type": "token",
+          "description": "Reserved $40/$60 Ft. Wayne token",
           "hexes": [
             "E11"
           ],
@@ -444,6 +445,7 @@ module Engine
       "abilities": [
         {
           "type": "token",
+          "description": "Reserved $40/$100 Cincinnati token",
           "hexes": [
             "H12"
           ],
@@ -491,6 +493,7 @@ module Engine
       "abilities": [
         {
           "type": "token",
+          "description": "Reserved $40 Erie token",
           "hexes": [
             "D20"
           ],
@@ -538,6 +541,7 @@ module Engine
          {
             "type":"tile_lay",
             "free":true,
+            "description": "Free tile lay: E5, F6, G5, H6, J4",
             "hexes":[
                "E5",
                "F6",
@@ -553,6 +557,7 @@ module Engine
         },
         {
           "type": "token",
+          "description": "Reserved $40 Centralia token",
           "hexes": [
             "I5"
           ],
--- a/lib/engine/step/g_18_al/token.rb
+++ b/lib/engine/step/g_18_al/token.rb
@@ -10,10 +10,10 @@ module Engine
           super
 
           entity.abilities(:assign_hexes) do |ability|
-            next unless ability.name == 'Historical objective'
+            next unless ability.hexes.length == 1
 
             if city.hex.name == ability.hexes.first
-              @log << "#{entity.name} receives $100 - #{ability.name} reached in #{ability.display_value}"
+              @log << "#{entity.name} receives $100 - #{ability.description}"
               entity.cash += 100
               entity.remove_ability(ability)
             end
--- a/lib/engine/ability/base.rb
+++ b/lib/engine/ability/base.rb
@@ -9,17 +9,16 @@ module Engine
       include Helper::Type
       include Ownable
 
-      attr_accessor :name, :display_value
+      attr_accessor :description
       attr_reader :type, :owner_type, :remove, :when, :count
 
-      def initialize(type:, name: nil, owner_type: nil, count: nil, remove: nil, display_value: nil, **opts)
+      def initialize(type:, description: nil, owner_type: nil, count: nil, remove: nil, **opts)
         @type = type&.to_sym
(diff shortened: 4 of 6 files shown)


lib/engine/config/game/g_1846.rb
```diff
@@ -491,6 +495,8 @@ module G1846
       "abilities": [
         {
           "type": "token",
+          "name": "Reserved Eerie token",
```
jenf, line 498: Erie?

lib/engine/config/game/g_1846.rb
```diff
@@ -444,6 +446,8 @@ module G1846
       "abilities": [
         {
           "type": "token",
+          "name": "Reserved Cincinatti token",
```
jenf, line 449: Cincinnati

jenf:
Apart from the typos looks good for me, will need Toby to review.

dfannius:
I would write "Ability" instead of "Attribute", and "Cost" instead of "Value".

perwestling:
> I would write "Ability" instead of "Attribute", and "Cost" instead of "Value".

@dfannius It is not always Cost. E.g. in 18AL uses this for its Historical objective, and then the "display_value" is the name of the city that is the destination. Also in 18AL when Coal Field token is assigned to one of the 5 cities, the hex assigned is shown in Value. 
The third one in 18AL is the route bonuses, where the name of the route is shown in Attribute and the bonus ($20 or $40) is shown in Value.

So I saw it as a value to the attribute....

For 1846 one could consider to instead use e.g. PRR
Attribute                                Value
Reserved $40 token          Cincinatti

The ones I was least happy with is the "Free tile lay       $0", I think it would be better to show the 5 hexes in the Value.

Ability/Attribute - yes, I think that can be a good one.

Cost/Value - do we have something more neutral?

perwestling:
About 18AL historical objective vs Coal Field token: Maybe I should be consistent and use hex name in both cases?  So G4 instead of Birmingham. (Alternatively to use city name also for Coal Field token.)

dfannius:
I see, your "value" means the value of a variable, not a monetary value (which is what people would interpret the word as).

If the cells in the second column can mean very different things from row to row, I would not even make a header for it. Ideally the cells could say things like "costs $40" but I know there is not much space. Parenthetical expressions like "($40)" are used in accounting to denote negative numbers and maybe could get the idea of a cost across.

perwestling:
It is the associated value of the ability/attribute, and if it is a cost, a bonus, an income, a hex, a city depends on the context of the ability....  
Removing the column header for that part is an idea, but I think it can be confusing if the Company table above Ability/Attribute shows Value and Income on the right part.

Another variant is to just use "name" and then each game can put whatever they like in that name, and not have anything to the right.
So in 18AL it would be
`Historical objective Birmingham`
And in 1846 is would be
`Reserved $40/$60 Ft. Wayne token`

perwestling:
18AL version:
Screenshot from 2020-07-24 16-05-21

1846 updated version:
Screenshot from 2020-07-24 16-19-31

assets/app/view/game/corporation.rb
```diff
@@ -48,8 +48,10 @@ def render
           children << h(Companies, owner: @corporation, game: @game) if @corporation.companies.any?
         end
 
-        abilities = @corporation.all_abilities.select { |ability| ability.owner.corporation? }
-        children << render_abilities(abilities) if abilities&.any?
+        abilities_to_display = @corporation.all_abilities.select do |ability|
+          ability.owner.corporation? && !ability.name.nil?
```
tobymao, line 52: ability.name

assets/app/view/game/corporation.rb
```diff
@@ -48,8 +48,10 @@ def render
...
+        end
+        children << render_abilities(abilities_to_display) if abilities_to_display&.any?
```
tobymao, line 54: abilities_to_display will never be nil no need for &

assets/app/view/game/corporation.rb
```diff
@@ -308,23 +310,22 @@ def render_revenue_history
 
       def render_abilities(abilities)
         attribute_lines = []
+        props = { style: { display: 'inline-block' } }
```
tobymao, line 313: i think we need a bit of margin / padding for the abilities

lib/engine/config/game/g_1846.rb
```diff
@@ -399,6 +399,7 @@ module G1846
       "abilities": [
         {
           "type": "token",
+          "name": "Reserved $40/$60 Ft. Wayne token",
```
tobymao, line 402: i think we shouldn't use name, it should be description

can you change the ability.name to be ability.description?

lib/engine/step/g_18_al/token.rb
```diff
@@ -10,7 +10,7 @@ def place_token(entity, city, token, teleport: false)
           super
 
           entity.abilities(:assign_hexes) do |ability|
-            next unless ability.name == 'Historical objective'
+            next unless ability.name.start_with == 'Historical objective'
```
tobymao, line 13: this is not a method?

it should be name.start_with?('Historical objective')

yzemaze:
> 1846 updated version:

Left-aligned looks really wonky, even out of place (quite literally ;) ). If most abilities will be on companies’/minors’ cards center-aligned will fit way better. After toby is done with the specials rewrite I’ll look into headers of company tables to fit the centered style better.

assets/app/view/game/corporation.rb
```diff
@@ -308,23 +310,22 @@ def render_revenue_history
...
+          attribute_lines << h('div.nowrap', props, ability.name)
         end
 
         table_props = {
           style: {
-            padding: '0 0.5rem',
-            grid: 'auto / 3fr 2fr',
-            gap: '0 0.2rem',
+            padding: '0.5rem',
+            grid: 'auto',
```
yzemaze, line 322: grid & gap are obsolete if you’re just using lines

lib/engine/config/game/g_1846.rb
perwestling: Fixed

lib/engine/config/game/g_1846.rb
perwestling: Fixed

assets/app/view/game/corporation.rb
perwestling: Fixed

assets/app/view/game/corporation.rb
perwestling: Fixed

lib/engine/config/game/g_1846.rb
perwestling: ok, changed

lib/engine/step/g_18_al/token.rb
perwestling: I removed the name/description from the configuration and check for hexes size instead. The other asssign_hexes ability in 18AL has more than 1 hexes to choose from.

assets/app/view/game/corporation.rb
perwestling: Fixed

assets/app/view/game/corporation.rb
perwestling: Fixed

assets/app/view/game/corporation.rb
perwestling: Fixed

lib/engine/config/game/g_1846.rb
perwestling: Ok, fixed

lib/engine/step/g_18_al/token.rb
perwestling: Yeah, was a bug. But I changed this check to checking for hexes.length instead, so I could remove name/description from the config file.

assets/app/view/game/corporation.rb
perwestling: Removed now. And put in centralized as you had in your commit.

assets/app/view/game/corporation.rb
perwestling: I tried to add padding here but it just moved the ability lines a bit to the right, and kept Ability header as is, which looked fishy.

Any suggestion what should be done?

perwestling:
After the latest push 1846 looks like this:
Screenshot from 2020-07-25 13-03-23

And when ability and company appear both on the corporation charter it looks like this:
Screenshot from 2020-07-25 13-27-22

PR #1160, opened 2020-07-24
NWR should happen after the original first n-train, if SC is setup

jenf:
SC adding a train causes NWR to happen a train later.

Show this both on the interface and implement it correctly.

Create the SC reserve trains at the start to ensure the ids work.


assets/app/view/game/game_info.rb
```diff
@@ -62,6 +62,20 @@ def phases
           phase[:events]&.each do |name, _value|
             event_text << (@game.class::EVENTS_TEXT[name] ? "#{@game.class::EVENTS_TEXT[name][0]}*" : name)
           end
+          trains = @depot.upcoming.select { |train| train.name == phase[:on] }
+          trains.each.with_index do |train, index0|
```
tobymao, line 66: don't like the name index0

assets/app/view/game/game_info.rb
```diff
@@ -62,6 +62,20 @@ def phases
...
+            train.events.each do |name|
+              index = index0 + 1
+              ordinal =
```
tobymao, line 69: prefer not to do this, if needed make a library or change the verbiage of the text to not depend on this

event_text when purchasing train.name train #index

lib/engine/train.rb
```diff
@@ -7,7 +7,7 @@ module Engine
   class Train
     include Ownable
 
-    attr_accessor :obsolete, :operated
+    attr_accessor :obsolete, :operated, :events
```
tobymao, line 10: hmm do we really need to do this? there's no other way to track it?

jenf: Maybe I could add a field to the events to delay it by a number of trains bought

jenf: Then, it'd make the UI simpler as well

PR #1165, opened 2020-07-25
operating order on corp cards

yzemaze:
second part of fix for #1122

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -60,7 +60,7 @@ module View
             },
           }
 
-          subchildren = (n = @game.round.entities.find_index(@corporation)) ? [render_operating_order(n)] : []
+          subchildren = render_operating_order
           subchildren << render_revenue_history if @corporation.operating_history.any?
           children << h(:div, props, subchildren)
         end
@@ -315,10 +315,14 @@ module View
         h('div.bold', "Last Run: #{@game.format_currency(last_run)}")
       end
 
-      def render_operating_order(n)
+      def render_operating_order
         round = @game.round
-        div_class = '.bold' if n >= round.entities.find_index(round.current_entity)
-        h("div#{div_class}", "Order: #{n + 1}")
+        if (n = @game.round.entities.find_index(@corporation))
+          div_class = '.bold' if n >= round.entities.find_index(round.current_entity)
+          [h("div#{div_class}", "Order: #{n + 1}")]
+        else
+          []
+        end
       end
 
       def render_abilities(abilities)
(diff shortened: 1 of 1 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -52,8 +52,17 @@ def render
...
-          children << h('table.center', [h(:tr, subchildren)])
+          props = {
+            style: {
+              grid: '1fr / repeat(2, max-content)',
+              gap: '2rem',
+              justifyContent: 'center',
+            },
+          }
+
+          subchildren = (n = @game.round.entities.find_index(@corporation)) ? [render_operating_order(n)] : []
```
tobymao, line 63: %s/find_index/index/g

tobymao: it'd be cleaner if it was just 

sub_children = render_operating_order(@game.round.entities.index(@corporation))

assets/app/view/game/corporation.rb
```diff
@@ -303,7 +312,13 @@ def render_shares
 
       def render_revenue_history
         last_run = @corporation.operating_history[@corporation.operating_history.keys.max].revenue
-        h('td.bold', "Last Run: #{@game.format_currency(last_run)}")
+        h('div.bold', "Last Run: #{@game.format_currency(last_run)}")
+      end
+
+      def render_operating_order(n)
+        round = @game.round
```
tobymao, line 319: don't need to store this var you only use it once

assets/app/view/game/corporation.rb
yzemaze: yep, but won’t work for one test case:

```
Assets #html 1889 renders float
     Failure/Error: context.eval(script, filename: @file)
     
     MiniRacer::RuntimeError:
       >=: undefined method `>=' for nil
```

hence: assignment ? …

assets/app/view/game/corporation.rb
yzemaze: twice (was three times, hence it’s there). Where’s your cutoff? ;)

tobymao: ah didn't see you used it twice that's fine

assets/app/view/game/corporation.rb
tobymao: i mean obviously if you copy what i did literally, but you can write code to make it work, like using a compact or returning an empty array if nil in your function

yzemaze: sure, had that, but this looked cleaner to me …
I’ll change it again.

PR #1168, opened 2020-07-25
Improve payout ui

### LINKED ISSUE #886: [1846] Payout / Withhold / Half Pay UI enhancement
While selecting routes, the **submit** button name changes to reflect the total amount of the run(s). Once you submit the runs, the individual runs and their values are logged but not the total value. After the submit, the screen changes to offer **Payout** / **Withhold** / **Half Pay** buttons, but at this point there is no indication of the value that would go to stock holders or the company, or what that would do to the stock price.

It would be nice if the UI displayed more information.

Assume that a company is going to pay out $430, and its current share price is $150: 
* If it pays in full, that's $43/share and a double jump to $180. 
* If it half-pays, that's $210 to the company, and $22/share, and a single jump to $165. 
* If it withholds, thats $430 to the company and it falls back to $137.

### Idea 1: Change the button text to reflect the effects

**[ Full pay, $43/share ]** **[ Half pay, $210 + $22/share ]** **[ Withhold $430 ]**

### Idea 2: Create a table that describes the effects, with a plain button for each section

| [ Full Pay ] | [ Half Pay ] |[  Withhold ] |
| --- | --- | --- |
| Pay $43/share | Pay $23/share |  |
| | Keep $210 | Keep $430 |
| Right 2 ($180) | Right 1 ($165) | Left 1 ($137) |

jenf:
Make the outcome of dividend, withhold, half pay available to the user
Fixes #886


assets/app/view/game/dividend.rb
```diff
@@ -27,11 +30,37 @@ def render
...
+          direction = "#{option[:share_times]} #{option[:share_direction]}"
 
-          h('button.button.margined', { on: { click: click } }, text)
+          props = { style: { paddingRight: '1rem' } }
+          h(:tr, [
+            button,
+            h('td.right', props, [@game.format_currency(option[:company])]),
+            h('td.right', props, [@game.format_currency(option[:per_share])]),
+            h(:td, [direction]),
+            ])
```
tobymao, line 42: indentation

assets/app/view/game/dividend.rb
```diff
@@ -27,11 +30,37 @@ def render
...
+        }
+        share_props = { style: { width: '2.7rem' } }
+
+        h(:table, table_props, [
+          h(:thead, [
+            h(:tr, [
+              h(:th, 'Dividend'),
+              h(:th, 'Treasury'),
+              h(:th, share_props, 'Per Share'),
+              h(:th, 'Stock moves'),
```
tobymao, line 59: Stock Moves

lib/engine/step/dividend.rb
```diff
@@ -40,26 +50,52 @@ def process_dividend(action)
         )
 
         entity.trains.each { |train| train.operated = true }
+
         @round.routes = []
-        send(action.kind, entity, revenue)
+
+        unless Dividend::DIVIDEND_TYPES.include? action.kind
```
tobymao, line 56: parens

lib/engine/step/dividend.rb
```diff
@@ -28,9 +28,19 @@ def skip!
...
+        dividend_types.map do |type|
+          payout = send(type, entity, revenue)
+          [type, payout.merge(share_price_change(entity, revenue - payout[:company]))]
+        end.to_h
+      end
+
       def process_dividend(action)
         entity = action.entity
         revenue = routes.sum(&:revenue)
+        payout = dividend_options(entity)[action.kind.to_sym]
```
tobymao, line 42: should we instead change it to string? i see on line 56 you don't use to_sym, could be danagerous?

jenf: When coming from a load it's a string, when coming from the UI it's a sym. I've converted it to sym uniformly.

jenf:
Think I've addressed everything

PR #1174, opened 2020-07-26
NWR Private bonus, and close Canadian Pacific properly

jenf:
Implements remaining 1882 privates

### DIFF
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -18,16 +18,16 @@ module View
         def render
           round = @game.round
           @step = round.active_step
-          corporation = round.current_entity
-          @current_actions = round.actions_for(corporation)
-          corporation = corporation.owner if corporation.company?
+          entity = round.current_entity
+          @current_actions = round.actions_for(entity)
+          entity = entity.owner if entity.company?
 
           left = [h(UndoAndPass, pass: @current_actions.include?('pass'))]
           left << h(RouteSelector) if @current_actions.include?('run_routes')
           left << h(Dividend) if @current_actions.include?('dividend')
           left << h(BuyTrains) if @current_actions.include?('buy_train')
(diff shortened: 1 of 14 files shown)


assets/app/view/game/round/operating.rb
```diff
@@ -20,6 +20,7 @@ def render
           @step = round.active_step
           corporation = round.current_entity
           @current_actions = round.actions_for(corporation)
+          corporation = corporation.owner if corporation.company?
```
tobymao, line 23: maybe change corporation to entity

lib/engine/action/remove_token.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Action
+    class RemoveToken < Base
+      attr_reader :city, :slot, :token
```
tobymao, line 8: where is token set?

lib/engine/step/g_1882/nwr_track_bonus.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Engine
+  module Step
+    module G1882
+      module NwrTrackBonus
+        def tile_nwr?(tile)
+          nwr_name = 'NWR'
```
tobymao, line 8: just do this inline icon.name == 'NWR'

lib/engine/step/g_1882/special_nwr.rb
```diff
@@ -0,0 +1,162 @@
...
+        def pass_description
+          'NWR: Skip (Track)'
+        end
+
+        def active_entities
+          [@entity]
+        end
+
+        def blocks?
+          @state != nil
```
tobymao, line 57: just do @state

lib/engine/action/remove_token.rb
jenf: it's not :)

PR #1203, opened 2020-07-27
1882 Don't show tiles that don't have free token spots

### LINKED ISSUE #1178: 1882: tiles that cannot be tokened show up on the token lay step or NWR

jenf:
Fixes #1178

### DIFF
--- a/lib/engine/step/g_1882/special_nwr.rb
+++ b/lib/engine/step/g_1882/special_nwr.rb
@@ -138,7 +138,8 @@ module Engine
             end
 
           when :place_token
-            ability(entity).hexes.include?(hex.id)
+            ability(entity).hexes.include?(hex.id) &&
+            hex.tile.cities.any? { |c| c.tokenable?(entity.owner, free: true) }
           when :lay_tile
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_1882/special_nwr.rb
```diff
@@ -138,7 +138,8 @@ def available_hex(entity, hex)
             end
 
           when :place_token
-            ability(entity).hexes.include?(hex.id)
+            ability(entity).hexes.include?(hex.id) &&
+            hex.tile.cities.any? { |c| c.tokenable?(entity.owner, free: true) }
```
tobymao, line 142: so one thing to note is that avialble_hex should return an array of edges, what you have now is a boolean which is hacky and sort of works, but is not consistent with how this method is used elsewhere

jenf: yeah, i think i ran into that previously, it seems to be subtle different between tile laying and tokening.
Maybe we should rename the tile version to available_edges which gets called by available_hex by default

Issue #1130, opened 2020-07-20
Lake Shore Line can upgrade to green tile before Phase 2


JacCCMF:
{
  "id": 4428,
  "description": "Live 46: in 2 hours or less Monty/Akado/Jac one open spot? ",
  "user": {
    "id": 24,
    "name": "Jac"
  },
  "players": [
    {
      "id": 523,
      "name": "MontyBrewster71"
    },
    {
      "id": 24,
      "name": "Jac"
    },
    {
      "id": 529,
      "name": "Akado"
    },
    {
      "id": 648,
      "name": "starchitect"
    }
  ],
  "max_players": 5,
  "title": "1846",
  "settings": {
    "seed": 1.620502458385741e+38
  },
  "status": "active",
  "turn": 1,
  "round": "Operating Round",
  "acting": [
    529
  ],
  "result": {},
  "actions": [
    {
      "type": "bid",
      "entity": "starchitect",
      "entity_type": "player",
      "id": 1,
      "company": "MS",
      "price": 60
    },
    {
      "type": "message",
      "entity": "starchitect",
      "entity_type": "player",
      "id": 2,
      "message": "glhf"
    },
    {
      "type": "message",
      "entity": "MontyBrewster71",
      "entity_type": "player",
      "id": 3,
      "message": "ditto"
    },
    {
      "type": "message",
      "entity": "Jac",
      "entity_type": "player",
      "id": 4,
      "message": "hi everyone! we have a great group here "
    },
    {
      "type": "message",
      "entity": "Akado",
      "entity_type": "player",
      "id": 5,
      "message": "good luck, have fun!"
    },
    {
      "type": "bid",
      "entity": "Akado",
      "entity_type": "player",
      "id": 6,
      "company": "Pass (1)",
      "price": 0
    },
    {
      "type": "bid",
      "entity": "Jac",
      "entity_type": "player",
      "id": 7,
      "company": "MPC",
      "price": 60
    },
    {
      "type": "bid",
      "entity": "MontyBrewster71",
      "entity_type": "player",
      "id": 8,
      "company": "LSL",
      "price": 40
    },
    {
      "type": "bid",
      "entity": "starchitect",
      "entity_type": "player",
      "id": 9,
      "company": "C&WI",
      "price": 60
    },
    {
      "type": "bid",
      "entity": "Akado",
      "entity_type": "player",
      "id": 10,
      "company": "Pass (3)",
      "price": 0
    },
    {
      "type": "bid",
[...]

PR #1208, opened 2020-07-27
Ensure special tile lay follows phase colors (#1130)

perwestling:
For special tile tile, restrict available tiles to
the ones that are allowed to lay during current phase.

Add special lay to 18AL and 18MEX so that they can lay
Copper Canyon and Lumber Terminal again.

Put check for honoring existing track before short cut
for special lay, so existing track is also honored by
special lay.
(This was a potential bug with the special lay short cut.
 It should have no change of existing functionality for
 non special lays as this was an existing check that
 has just been moved. The remaining checks are still
 skipped by special lay.)

### DIFF
--- a/lib/engine/step/special_track.rb
+++ b/lib/engine/step/special_track.rb
@@ -32,10 +32,11 @@ module Engine
       end
 
       def potential_tiles(entity, hex)
+        colors = @game.phase.tiles
         (ability(entity)&.tiles || [])
           .map { |name| @game.tiles.find { |t| t.name == name } }
           .compact
-          .select { |t| hex.tile.upgrades_to?(t, true) }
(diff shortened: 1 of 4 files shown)


lib/engine/step/special_track.rb
```diff
@@ -32,8 +32,10 @@ def available_hex(entity, hex)
       end
 
       def potential_tiles(entity, hex)
+        colors = @game.phase.tiles
         (ability(entity)&.tiles || [])
           .map { |name| @game.tiles.find { |t| t.name == name } }
+          .select { |t| colors.include?(t.color) }
```
tobymao, line 38: you can add this inline to line 40

perwestling: Fixed

tobymao: i don't see the change

Issue #1061, opened 2020-07-10
[spreadsheet view] Change way certicate counts are displayed


zurfarosa:
Currently in spreadsheet view the Certs row counts only the certificates that count towards certificate limits (i.e. it ignores certificates from yellow or orange companies). But I find that towards the end of the game, knowing a player's *total* certificate count allows me to more easily gauge their earning potential. 

Can I suggest instead that we display the certificate information as follows:
1. **Certs** (this would be a total count of all certificates)
2. **of which count towards limit**  (can't think of anything less wordy at the moment!)

EDIT: 
I've updated my suggestion after reading Dan's reply. 

If you had 7 shares in just one company, which happened to be a yellow company, it would say: 
1. **Shares**: 7
2. **Certs**: 0/18

Having the "/18" makes it a little clearer that "certs" only includes those certificates which count towards the certificate limit. It's also a reminder of the certificate limit for that particular game or player count.

dfannius:
I think the word for #2 is "shares". For example, I imagine you want the president's certificate to count as 2 for purposes of this stat.

(That said, I don't have a problem just summing the column in my head, but I can appreciate other people's desire for this number.)

zurfarosa:
> I think the word for #2 is "shares". For example, I imagine you want the president's certificate to count as 2 for purposes of this stat.
> 
> (That said, I don't have a problem just summing the column in my head, but I can appreciate other people's desire for this number.)

Yes, that makes sense. So if you had 7 shares in just one company, which happened to be a yellow company, it would say: 
1. **Shares**: 7
2. **Certs**: 0/18

Having the "/18" makes it obvious that "certs" only includes those certificates which count towards the certificate limit. It's also a reminder of the certificate limit for that particular game or player count.

Issue #1155, opened 2020-07-24
Add priority to spreadsheet


ChrisTheCat:
It would be nice if the spreadsheet indicated who has priority deal.

Issue #872, opened 2020-06-25
[spreadsheet view] ability to sort companies by a user's number of shares


zurfarosa:
You should be able to click on user's name and sort the table according to how many shares in the company that user has.

Issue #473, opened 2020-05-23
Add liquid cash to player info/spreadsheet


jenf:
When I was trying to setup a company is forgotten to take into account the privates.

It'd be useful to show how much a player has if they sell all the shares they can sell.

hast18xx:
one more nth is to see players available cash in the names row

![image](https://user-images.githubusercontent.com/64490299/82729841-30f04b00-9cfb-11ea-8014-a47e9bd3537f.png)

ChrisTheCat:
Now that this is on the player cards, could it also be added to the spreadsheet?

yzemaze:
@hast18xx 
Players’ cards with cash is rendered right below the corporations. If that’s not enough please open a new issue.

PR #1209, opened 2020-07-27
Spreadsheet: some infos added, styling/language consolidated

yzemaze:
I left "Privates" as header, because it’s a tad shorter and a bit easier to grasp instantly than "Companies".

Should resolve: #1061, #872, #473, #1155


tobymao:
it looks good but i think we should switch privates to companies since it's more consistent.

Issue #1105, opened 2020-07-17
Db container problem on CentOS


perwestling:
Did try to get 18xx.games developer site started on a CentOS installation.

When trying to do "make dev_up_b" the 18xxgames_db_1 container fail to start.
(Gets a chown error for /var/lib/postgres/db)

@jenf trouble shooted this one and found out that there is a different default user ID on CentOS.

So by changing line 12 in db/Dockerfile to use the actual user id of the user instead of the default 1000, removing the containers and then retry it worked.

So it would be good to use the actual userid of the user starting the container instead of a hardcoded one.

benjaminxscott:
Is this still an issue?

tobymao:
let’s close it won’t get fixed

perwestling:
Yes it is still an issue. I have a local commit that I apply to the branches I use on CentOS, and I will continue to use.

My reason for having this issue was two fold:
- Maybe it would be fixed someday?
- Anyone that encounter this problem could find the issue, and then also the solution

PR #1229, opened 2020-07-28
Setup db user in an entrypoint script

yoyoson:
Trying to address #1105. I have tested this on Centos7, but not elsewhere. 

After this, should not have to edit db/Dockerfile to replace '1000' with your actual UID.

Overall I was trying to make the process of running "sudo make dev_up_b" from a fresh checkout as smooth as possible.  To that end, a new dev using Centos might still need to first disable selinux.

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -1,13 +1,11 @@
 FROM postgres:12.2
 
-# use uid 1000 to (hopefully) match host user (ie, you)
-RUN useradd --uid 1000 --user-group --create-home --home-dir /home/db db
-
-RUN chown db:db /var/lib/postgresql/data && \
-    chown db:db /var/run/postgresql
-
-USER db:db
+USER root:root
 
 COPY ./postgresql.conf /etc/postgresql/postgresql.conf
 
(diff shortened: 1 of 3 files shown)


db/Dockerfile
```diff
@@ -1,13 +1,11 @@
 FROM postgres:12.2
 
-# use uid 1000 to (hopefully) match host user (ie, you)
-RUN useradd --uid 1000 --user-group --create-home --home-dir /home/db db
-
-RUN chown db:db /var/lib/postgresql/data && \
-    chown db:db /var/run/postgresql
-
-USER db:db
+USER root:root
```
tobymao, line 3: we don't use sudo or root

yoyoson: I will rethink this.

yoyoson: Since /etc/passwd is owned and only writable by root, I don't see any way to do the `useradd` after the container starts and w/o also running as root.

I have another approach that goes back to baking the 'db' UID into the image, thought it seems that hasn't been an issue.  Unfortunately it involves some more conscious upfront effort.

PR #1246, opened 2020-07-29
button styling consolidated

yzemaze:


public/assets/main.css
```diff
@@ -11,11 +11,24 @@ a {
   text-underline-offset: 0.2em;
 }
 
-button {
-  margin: 0.5rem 0.5rem 0 0;
+button, .button-link {
```
tobymao, line 14: let's rename button_link to underscores to be consistent ongoing

public/assets/main.css
```diff
@@ -11,11 +11,24 @@ a {
...
+  cursor: pointer;
+  text-decoration: none;
 }
-button + button {
-  margin-left: 0.5rem;
+button.small {
+  padding: 0.2rem 0.5rem;
+  margin-right: 0.5rem;
+}
+button:last-child, button.par_price:nth-of-type(6n), button.sell_share:nth-of-type(3n) {
```
tobymao, line 30: what does 6n and 3n mean

yzemaze: x*n-th element (6th, 12th etc.)
Used here to remove margin-right _only_ from the last element per row of (possibly) multiple rows [1846!] so the full width can be occupied.

PR #1256, opened 2020-07-30
18Chesapeake 2p rules

### LINKED ISSUE #477: Implement 2-player 18Chesapeake rules
Per section 14 of the rulebook.

- President's certificates (except for Vanderbilt which is still 20%) are 30%
- When a company floats, all remaining IPO shares are placed in the market
- When a non-president buys a share from the market, one share is exported (before the buy proceeds)
- Stocks float up if market is empty (don't need 100% in player hands)

michaeljb:
* 30% presidents share, except for Cornelius
* on corporation float, put remaining IPO shares into market
* when non-president buys shares, first remove a share from the market
* at the end of the stock round, the share price increases when no shares are in
  the market, even though 100% cannot be owned by players

Co-authored-by: @scottredracecar

[Fixes #477]


lib/engine/game/g_18_chesapeake.rb
```diff
@@ -90,6 +124,16 @@ def cornelius
       def or_set_finished
         depot.export! if %w[2 3 4].include?(@depot.upcoming.first.name)
       end
+
+      def float_corporation(corporation)
+        super
+
+        return unless players.size == 2
```
tobymao, line 123: return super unless players

michaeljb: I want to call super no matter what. For the order of how logs will come out, I want super called before the special 2p stuff is handled.

michaeljb:
A couple things from Slack:
- make the Bank the owner of "removed" shares
- don't subclass the corporation to change share percentages

lib/engine/game/g_18_chesapeake.rb
michaeljb: Would this be better as an `after_float` method?

tobymao: nah, tthis is fine

michaeljb:
> * make the Bank the owner of "removed" shares
> * don't subclass the corporation to change share percentages

Addressed both of these, but something went wrong and now neither player can buy shares once the corp is floated. Will come back to this later tonight.

tobymao:
ping me when this is good to go

michaeljb:
@tobymao it should be all set now

PR #1258, opened 2020-07-30
acting player? => favicon / browser tab color change

yzemaze:
Tested across sys (Manjaro/Android/Win) and browsers (FF/Chrome/Edge). Worked smoothly. Please check MacOS/iOS/Safari.

Standard theme color was/is set to white (although it might not have worked due to missing name). If you’d like a different one change content in L28+.


assets/app/lib/settings.rb
```diff
@@ -50,5 +50,19 @@ def route_prop(index, prop)
...
     end
+
+    def change_favicon(change)
+      `document.getElementById('favicon_svg').href = '/images/icon' + #{change ? '_red' : ''} + '.svg'`
+      `document.getElementById('favicon_16').href = '/images/favicon-16x16' + #{change ? '_red' : ''} + '.png'`
+      `document.getElementById('favicon_32').href = '/images/favicon-32x32' + #{change ? '_red' : ''} + '.png'`
+      `document.getElementById('favicon_apple').href = '/apple-touch-icon' + #{change ? '_red' : ''} + '.png'`
+    end
+
+    def change_tab_color(change)
```
tobymao, line 61: change is not a good name, it should be active

assets/app/lib/settings.rb
```diff
@@ -50,5 +50,19 @@ def route_prop(index, prop)
     def route_prop_string(index, prop)
       "r#{index}_#{prop}"
     end
+
+    def change_favicon(change)
```
tobymao, line 54: same here

PR #1267, opened 2020-07-30
end date on finished games

### LINKED ISSUE #1264: Feature Request:  "Date Ended" timestamp for Completed Games
The "Completed Games" section in your profile is currently sorted by Game ID, with no indication of when a game ended.  Some async games may end when a user is sleeping or not available, and it would disappear from their "Your Games" section on the main page and appear somewhere in the "Completed Games" section with no great way to search for it or identify it as having ended recently.

Being able to sort by Game ID and Last Move or Date Updated or Date Ended would be awesome, but even just putting the Date Ended there would be great.

yzemaze:
fixes #1264

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -132,8 +132,8 @@ module View
       )
     end
 
-    def render_time_or_date(timestamp)
-      ts = Time.at(@gdata[timestamp].to_i)
+    def render_time_or_date(ts_key)
+      ts = Time.at(@gdata[ts_key]&.to_i || 0)
       time_or_date = ts > Time.now - 82_800 ? ts.strftime('%T') : ts.strftime('%F')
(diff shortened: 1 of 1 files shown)


assets/app/view/game_card.rb
```diff
@@ -132,8 +132,10 @@ def render_link(href, click, text)
       )
     end
 
-    def time_or_date(ts)
-      ts > Time.now - 82_800 ? ts.strftime('%T') : ts.strftime('%F')
+    def render_time_or_date(timestamp)
```
tobymao, line 135: this is not a timestamp, this is a key

assets/app/view/game_card.rb
```diff
@@ -132,8 +132,10 @@ def render_link(href, click, text)
...
+      ts = Time.at(@gdata[timestamp].to_i)
```
tobymao, line 136: this will crash if there's no data should be safer here and do

@gdata[ts_key]&.to_i || 0

PR #1274, opened 2020-07-30
history links titles

yzemaze:
Does "latest" work? "most recent" or sth. else better?
(Splitting into latest / game end seems not worth it.)


tobymao:
How about 

Start, Previous Round, Previous Action, Next Action, Next Round, End

yzemaze:
> How about
> 
> Start, Previous Round, Previous Action, Next Action, Next Round, End

End doesn’t really work for still running games, but I wouldn’t mind.

tobymao:
Current

PR #1280, opened 2020-07-31
sort games by date DESC

### LINKED ISSUE #1113: Option to sort completed games by finish date in profile
This would make it much easier to find the games that "just finished and disappeared from my home page."

yzemaze:
fixes #1113

On a related note: Is the sorting in https://github.com/tobymao/18xx/blob/30b2d4aab3ee1850a040a56dbd152d1fb16c53dc/assets/app/game_manager.rb#L155-L160 necessary? Performance reasons?

### DIFF
--- a/assets/app/view/home.rb
+++ b/assets/app/view/home.rb
@@ -32,10 +32,10 @@ module View
       # Ready, then active, then unstarted, then completed
       your_games.sort_by! do |game|
         [
-          user_is_acting?(@user, game) ? -game['id'] : 0,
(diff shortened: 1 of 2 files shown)


tobymao:
does this break paging? can you check

tobymao:
> fixes #1113
> 
> On a related note: Is the sorting in
> 
> https://github.com/tobymao/18xx/blob/30b2d4aab3ee1850a040a56dbd152d1fb16c53dc/assets/app/game_manager.rb#L155-L160
> necessary? Performance reasons?

i don't remember

PR #1283, opened 2020-07-31
Merge auction and draft in preparation for 1817

jenf:
This merges 1846's draft and other auctions.

This should now be flexible enough to cope with any kind of auction or draft of the private companies ready for implementing 1817.


assets/app/view/game/round/auction.rb
```diff
@@ -13,23 +13,40 @@ class Auction < Snabberb::Component
         include Actionable
 
         needs :selected_company, default: nil, store: true
+        needs :hidden, default: true, store: true
+        needs :user, store: true, default: nil
```
tobymao, line 17: let's pass this in intsead of store

Issue #1288, opened 2020-07-31
18AL Memphis & Charleston Railroad


ChrisTheCat:
The text on the Memphis & Charleston Railroad company needs to be updated to reflect the fact that each bonus may only be scored by one train (both bonuses can be scored by a single train, but neither bonus may be scored by more than one train).

tobymao:
you can fix this one yourself 

https://github.com/tobymao/18xx/blob/master/lib/engine/config/game/g_18_al.rb#L252

PR #1289, opened 2020-07-31
Enhanced company descriptions

ChrisTheCat:


ChrisTheCat:
This closes issue #1288 but I'm not sure I did the right thing to make that happen.

Issue #1204, opened 2020-07-27
NWR marker covers income on some tiles


jenf:
![NWR Problem](https://user-images.githubusercontent.com/71923/88554600-9b38ab00-d01e-11ea-9517-4aca4202c519.png)

18xx.games/game/4978

jenf:
fixed by #1306

PR #1306, opened 2020-08-02
Improve 1882 and 18AL icons location

### LINKED ISSUE #1300: 18AL: Rendering of icon in Birmingham tile collides with revenue
In 18AL a mining icon is used to show where the Warrior Coal Field token can be assigned, or has been assigned. This is rendered on a bad place on the Birgmingham tile, at 3 o'clock, which means it collides with the revenue for that tile. 

I think one possibility would be to make it possible to specify where the icon should be placed. In the case of Birmingham, putting the icon at 1 o'clock (North East) should be good enough.

Here is how Birmingham would look in green with mining icon still there:
Birmingham_mining
From the beginning Birmingham also has a mountain icon:
Empty_Birmingham

Birmingham has the B tiles, special in each color. Note! The mining icon will be removed no later than the 6T is bought, so it will never appear on the gray tile.
Tiles

jenf:
Fixes #1300 and #1204
Screenshot from 2020-08-02 12-48-57
Screenshot from 2020-08-02 12-49-28

### DIFF
--- a/assets/app/view/game/part/icons.rb
+++ b/assets/app/view/game/part/icons.rb
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 require 'view/game/part/base'
+require 'view/game/part/small_item'
 
 module View
   module Game
     module Part
       class Icons < Base
+        include SmallItem
         ICON_RADIUS = 16
         DELTA_X = (ICON_RADIUS * 2) + 2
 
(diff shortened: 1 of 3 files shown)


perwestling:
Nice! The only small issue is that the icon and the name label of the city does compete for the same space a bit. In this case the upper left corner would be the optional place (as upper right corner in an upgraded tile contains a mountain icon) but I do think the current solution is good enough.

Issue #1251, opened 2020-07-29
Usability: Add the possibility to show implementation details for the game on the info page


perwestling:
On the info page today is shown designer, a link to the rules, and some more.

But I would like to see a section where one could write implementation details. E.g. that optional rules not supported, differences from the rule book based on differences in the digital implementation, and mayne other things. It could be enough with a simple text.

ChrisTheCat:
As discussed on Slack, we could add a link to the wiki on the game Info tab, and then put implementation details / rules clarifications in the wiki where it's easier for non-programmers to edit/test

jenf:
Wiki is the appropriate place, closing

PR #1309, opened 2020-08-02
Show wiki page as game info link if such exists (#1251)

perwestling:
If the GAME_INFO_URL is set, do show that as a link under
Game Info part of the Info page. Do not show any other Game
Info parts.


tobymao:
why remove the other info? why not show both

perwestling:
I thought it was bad to duplicate the information. But I will push up a new version that just adds the external link if present.

Issue #1265, opened 2020-07-30
Usability: Add possibility to color log texts


perwestling:
I think color could be used as an enhancement if it is used restrictively and as a complement. (The information presented should be noticeable even if colors not used or color difference noticed.)

An example on how this could be used:
Red text - for warnings, that needs player action. 
Green text - for new rounds text, for end game notice

Usecases for red text:
- Warning that End Game condition triggered
- If a company ability check is not easy to implement it could be possible to notify players that we need to check that it was valid (ex 18AL, SNA RR - is token allowed to assign?). Having it colored would make it a more feasible to use in this way.
- Bankrupcy, Emergency Funding and similar might require red text

If introducing this it might be good to document design rules on Wikipedia when colors could/should be used to avoid overuse of it.

perwestling:
The bold style you now get if you put "--" at the start of the log message might be good enough?
Have added that to an 18AL message where players are expected to check a private, and to Game End.

So good enough I think.

PR #1315, opened 2020-08-03
Added styling to important log messages

AnthonyPAlicea:
Similar to #1265 this is to very simply, for now, make it harder to miss the important automated log messages and make the overall log a bit easier to parse when looking through history. Admittedly a brittle solution, but simple enough for now and will work across games in history.

### DIFF
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -49,12 +49,12 @@ module View
       end
 
       lines = @log.map do |line|
+        line_props = { style: {} }
         if line.is_a?(String)
           if line.start_with?('--')
-            h('div.chatlog-important', line)
-          else
-            h(:div, line)
+            line_props[:style][:fontWeight] = 'bold'
           end
+          h(:div, line_props, line)
         elsif line.is_a?(Engine::Action::Message)
           h(:div, { style: { fontWeight: 'bold' } }, "#{line.entity.name}: #{line.message}")
         end
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -50,7 +50,11 @@ module View
 
       lines = @log.map do |line|
         if line.is_a?(String)
-          h(:div, line)
+          if line.start_with?('--')
+            h(:div, { style: { fontWeight: 'bold', color: 'green' } }, line)
+          else
+            h(:div, line)
+          end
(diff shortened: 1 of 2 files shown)


assets/app/view/log.rb
```diff
@@ -50,7 +50,11 @@ def render
 
       lines = @log.map do |line|
         if line.is_a?(String)
-          h(:div, line)
+          if line.start_with?('--')
```
tobymao, line 53: i think the green might be too much

also i think this may be a bit cleaner if you only do the style programatically


so
line_props = { style: {}}
line_props[:style][:fontWeight] = 'bold' if line.start_with?('--')
h(:div, line_props, line)

perwestling: Green doesn't look important. Red might...
I agree with Toby that bold might be enough to start with.

AnthonyPAlicea:
Tried something different then, rather than start cluttering up the code with future styling. Just added a CSS class, which makes it possible to deal more cleanly deal with stacked important messages, and potentially in the future can add adjustment classes for particular kinds of messages. Still would like to differentiate between important messages and user-generated messages (my groups tend to be chatty). 

It looks like this:

18xxGifs

assets/app/view/log.rb
AnthonyPAlicea: Changed to a CSS class.

tobymao: sorry to be picky, but could we do this inline and not use css classes. 

i try to avoid css classes as much as possible.

by changing this to a CSS class, the notifications emails will not have any styling since we can't send CSS to mail providers. inline styles are more universal and easier to maintain in this project

AnthonyPAlicea: Oh I understand. No problem at all, thanks for explaining.

tobymao:
thanks!

assets/app/view/log.rb
AnthonyPAlicea: Ok, changed back to inline styles and cleaned it up a bit. Didn't touch the user-generated messages render.

Thanks for the patience! Really appreciating the site and wanted to start trying to contribute if possible.

tobymao:
a lot of discussion happens in the slack so feel free to drop in and say
hello!

PR #1324, opened 2020-08-03
1817 initial auction

jenf:
First pass of the initial auction for 1817

### DIFF
--- a/lib/engine/step/g_1817/selection_auction.rb
+++ b/lib/engine/step/g_1817/selection_auction.rb
@@ -1,0 +1,168 @@
+# frozen_string_literal: true
+
+require_relative '../base'
+
+module Engine
+  module Step
+    module G1817
+      class SelectionAuction < Base
+        include Auctioner
+        ACTIONS = %w[bid pass].freeze
+
+        attr_reader :bids, :companies
+
+        def description
+          'Bid on Companies'
+        end
+
+        def available
+          @companies
+        end
+
+        def process_pass(action)
+          entity = action.entity
+
+          if auctioning_company
+            entity.pass!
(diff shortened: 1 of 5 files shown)


lib/engine/step/g_1817/selection_auction.rb
```diff
@@ -0,0 +1,168 @@
...
+          @companies
+        end
+
+        def process_pass(action)
+          entity = action.entity
+
+          if auctioning_company
+            entity.pass!
+            pass_auction(action)
+
```
tobymao, line 28: new line

lib/engine/step/g_1817/selection_auction.rb
```diff
@@ -0,0 +1,168 @@
...
+        end
+
+        def available
+          @companies
+        end
+
+        def process_pass(action)
+          entity = action.entity
+
+          if auctioning_company
```
tobymao, line 25: maybe the pass should go in the line above since you call it twice?

so an idea, not sure if it makes it cleaner but it's what i'm doing for 1817

instead of mixing pass usage for bidding on an active company and placing bids, when a bid starts, i add all eligible entities into bidders. when they pass, instead of calling .pass! i remove them from the bidders, that way pass! is only used one way

tobymao:
i'm curious if this could be reworked into the 1817 corporation token auction as well, i may have a go at it after this gets merged in

lib/engine/step/g_1817/selection_auction.rb
jenf: yeah, i don't really like the entity.pass, i'd rather the step keep track of the pass than the entity itself

PR #1329, opened 2020-08-04
Add two bold texts, and improve phase status

perwestling:
Applied bold text to Game over message, and one Note! in 18AL.

Made it possible to show phase status, like e.g. Limited Train Buy in 18AL. See picture.

Screenshot from 2020-08-04 13-34-46

### DIFF
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -53,7 +53,15 @@ module View
 
       def phases
         current_phase = @game.phase.current
+        phases_events = []
         rows = @game.phase.phases.map do |phase|
+          row_events = []
+
+          phase[:status]&.each do |status|
+            row_events << @game.class::STATUS_TEXT[status] if @game.class::STATUS_TEXT[status]
+          end
+          phases_events.concat(row_events)
+
           phase_color = Array(phase[:tiles]).last
           bg_color = color_for(phase_color)
           phase_props = {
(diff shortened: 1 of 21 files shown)


perwestling:
A possible improvement of this commit can be to refactor Can Buy Companies so it is just a status among others. 

Still another could be in the SignleDepotBuyBeforePhase4 feature, to check for limited_train_buy status in current phase instead of check explicitly that phase < 4.

assets/app/view/game/game_info.rb
```diff
@@ -63,18 +72,32 @@ def phases
...
+            h(:td, row_events.map(&:first).join(', ')),
           ])
         end
 
+        status_text = phases_events.uniq.map do |short, long|
+          h(:tr, [h(:td, short), h(:td, long)])
+        end
+
+        if status_text.any?
+          status_text = [h(:table, [
```
tobymao, line 90: there's no point to making this an array because you flatten it anyways on line 118

tobymao: ah i see it's because you're using status_text previously and relying on this being empty, that's a bit tricky

lib/engine/game/base.rb
```diff
@@ -563,7 +566,7 @@ def end_game!
 
         @finished = true
         scores = result.map { |name, value| "#{name} (#{format_currency(value)})" }
-        @log << "Game over: #{scores.join(', ')}"
+        @log << "-- Game over: #{scores.join(', ')}"
```
tobymao, line 569: this should also have trailing --

lib/engine/config/game/g_18_al.rb
```diff
@@ -697,7 +697,10 @@ module G18AL
          "tiles":[
             "yellow"
          ],
-         "operating_rounds": 1
+         "operating_rounds": 1,
+         "status":[
```
tobymao, line 701: these statuses seem like they are just descriptions, should we do the refactor you're talking about?

tobymao:
> A possible improvement of this commit can be to refactor Can Buy Companies so it is just a status among others.
> 
> Still another could be in the SignleDepotBuyBeforePhase4 feature, to check for limited_train_buy status in current phase instead of check explicitly that phase < 4.

i think we should do this, i don't like that we have text only statuses without any function

assets/app/view/game/game_info.rb
perwestling: I must confess that this is more or less copy-paste from the Events presentation in another place in this file.

lib/engine/game/base.rb
perwestling: Done

lib/engine/config/game/g_18_al.rb
perwestling: Have now done the refactoring.

Issue #1135, opened 2020-07-21
Allow users to change their user name


jenf:
At the moment one user has their email address as their user name. They might want to be changing it, but the software won't allow it.

tobymao:
this is difficult because user names are used as keys in actions

jenf:
I think when a user changes names we add an alias into every game.
Bob changes to Bob2, then we update each of the games with Bob2 in so that Bob2 aliases to Bob.
This could also be used to replace players in a game, as I think a few people are finding a lot of dropouts

jenf:
{"seed": 5, "aliases": {"Bob2": ["Bob"]}}

game would take the list of aliases so any reference to Bob it treats as an action by Bob2

jenf:
After we've migrated players to use ids, we'll need to add functionality to allow players to change names.

I'd was thinking of adding a setting stating the last time they changed their name, and disallow changes for 60 days (or unless they bug an admin). I'd rather discourage changing names too often as it might cause confusion to other users.

PR #1330, opened 2020-08-04
Modify engine to use playerid->username hash for future modifications to allow players to rename themselves.

jenf:
Step 1 for #1135

Engine now takes a hash of players id's to names. This is currently ignored but lays the foundation for future fixes.
This can go in now, and more changes can come later.

Next steps
1. Clean up parts of the UI that use id rather than name. (I've seen a few)
2. Write migration script
3. Modify user settings to use id rather than name
4. Add UI to change player names

One thing I've spotted is that player names are used in the game results, my solution to this is to rerun unpinned games with that player in and store the new result.

### DIFF
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -52,8 +52,10 @@ module View
       return if game_id == @game&.id &&
         ((!cursor && @game.actions.size == @num_actions) || (cursor == @game.actions.size))
 
+      # Hotseat doesn't have player ids, use names instead.
+      players = @game_data['players'].map { |p| [p['id'] || p['name'], p['name']] }.to_h
       @game = Engine::GAMES_BY_TITLE[@game_data['title']].new(
-        @game_data['players'].map { |p| p['name'] },
+        players,
         id: game_id,
         actions: cursor ? actions.take(cursor) : actions,
         pin: @pin,
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -313,8 +313,15 @@ module Engine
         @finished = false
         @log = []
         @actions = []
-        @names = names.freeze
-        @players = @names.map { |name| Player.new(name) }
+        @names = if names.is_a?(Hash)
+                   names.freeze
+                 else
+                   names.map { |n| [n, n] }.to_h
+                 end
+
+        # This intentionally ignores player id for now until the database is migrated.
+        @players = @names.map { |_playerid, name| Player.new(name, name) }
+
         @optional_rules = init_optional_rules(optional_rules)
 
         @seed = @id.to_s.scan(/\d+/).first.to_i % RAND_M
--- a/routes/game.rb
+++ b/routes/game.rb
@@ -80,9 +80,11 @@ class Api
 
                 game.save
               else
+                players = users.map { |u| [u.id, u.name] }.to_h
                 engine = Engine::GAMES_BY_TITLE[game.title].new(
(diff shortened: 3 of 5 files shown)


tobymao:
i don't think we should do this, instead we should switch player ids to be the user id, and pass in the player name

so game can be initialize with [ { id: 1, name: 'jen'}] or ['jen']

this would also allow us to figure out who actually did the action in things like master mode and game end, because now we have access to the user_id => name mapping.

jenf:
would that leave the entity in the action as 'jen' or as '1'?

tobymao:
it would be 1

tobymao:
so player.id = 1, player.name = 'jen'

jenf:
that means rewriting all the actions in the database :(

tobymao:
yes

tobymao:
is this ready for a final look?

jenf:
Yep, but it's just part one but can be merged without migration

assets/app/view/game_page.rb
```diff
@@ -51,8 +51,10 @@ def load_game
       return if game_id == @game&.id &&
         ((!cursor && @game.actions.size == @num_actions) || (cursor == @game.actions.size))
 
+      # Hotseat doesn't have player ids, use names instead.
+      players = @game_data['players'].map { |p| [p['id'] || p['name'], p['name']] }.to_h
```
tobymao, line 55: you can just make the initializer take in an either an array of names or the hash. if an array of names, then you can do this logic

lib/engine/game/base.rb
```diff
@@ -173,6 +173,15 @@ class Base
 
       def setup; end
 
+      def self.player_range
+        self::CERT_LIMIT.keys.minmax
+      end
+
+      def self.create_with_max_players
+        players = player_range.max.times.map { |p| [p.to_s, "Player #{p + 1}"] }.to_h
```
tobymao, line 181: can be simplified with the initialize fix

routes/game.rb
```diff
@@ -80,9 +80,11 @@ class Api
 
                 game.save
               else
+                players = users.map { |u| [u.id, u.name] }.to_h
```
tobymao, line 83: i wonder will this break compatability with all pins?

jenf: No, since this is in the non-pin case

tobymao:
but what happens when we start pinning games where the engine has hash?

jenf:
I think as long as it's consistent it should be fine.

Any place where it's shown to the user we will need to use the hash otherwise you'll see "23 spends x"

yeknom02:
I was referred to this thread since I  was asking how to change my username on the site. I think it might be beneficial since I know a few contacts from different spaces

jenf:
It's being worked on, i'll pick it up again when 1817 hits alpha.

yeknom02:
@jenf if 1817 gets an alpha release, then I am sure my priorities will experience a temporary, yet severe, shift while this gets addressed. ;-)  Thanks for the reply.

assets/app/view/game_page.rb
jenf: done

jenf:
Finally picked this up again, should be ready for review. but this is only part 1 of the changes

tobymao (approved):

tobymao:
lgtm i'll let you handle merge and migrations

jenf:
This one doesn't need migrations, will start to write the script then check if anything falls apart

PR #1339, opened 2020-08-05
allow complete set of upgradeable checks for special tile lay abilities

### LINKED ISSUE #1285: Lake Shore Line Upgrades Cleveland to wrong type of Tile
When using the Lake Shore Line's ability, it upgrades Cleveland to a non Z tile, which means it is undervalued.

cheesinglee:
By default, special tile lay abilities skip the check for matching labels and matching number of cities/towns. One consequence of this is that the LSL private ability in 1846 allows upgrading Cleveland to a non-Z city or Toledo to a Z-city. This PR introduces a parameter to the ability configuration which will restore the full set of checks. Setting this parameter for LSL prevents the illegal upgrade.

Fixes #1285

### DIFF
--- a/lib/engine/ability/tile_lay.rb
+++ b/lib/engine/ability/tile_lay.rb
@@ -5,13 +5,14 @@ require_relative 'base'
 module Engine
   module Ability
     class TileLay < Base
-      attr_reader :hexes, :tiles, :free, :discount
+      attr_reader :hexes, :tiles, :free, :discount, :special_lay
 
-      def setup(hexes:, tiles:, free: false, discount: nil)
+      def setup(hexes:, tiles:, free: false, discount: nil, special_lay: true)
         @hexes = hexes
         @tiles = tiles
         @free = free
         @discount = discount || 0
+        @special_lay = special_lay
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/engine/ability/tile_lay.rb
```diff
@@ -5,13 +5,14 @@
 module Engine
   module Ability
     class TileLay < Base
-      attr_reader :hexes, :tiles, :free, :discount
+      attr_reader :hexes, :tiles, :free, :discount, :special_lay
 
-      def setup(hexes:, tiles:, free: false, discount: nil)
+      def setup(hexes:, tiles:, free: false, discount: nil, special_lay: true)
```
tobymao, line 10: i'd prefer special_lay: nil

and then @special_lay ||= true

lib/engine/step/special_track.rb
```diff
@@ -32,11 +32,12 @@ def available_hex(entity, hex)
       end
 
       def potential_tiles(entity, hex)
+        special_lay = ability(entity)&.special_lay.nil? || ability(entity)&.special_lay
```
tobymao, line 35: this calls ability 3x, you should call it once and then save it

lib/engine/ability/tile_lay.rb
tobymao: ah yes i see in your code you check both nil and special lay, this is error prone

tobymao:
https://github.com/tobymao/18xx/commit/c101409dd47943d2359a385821a37479f120d8cf 

cleanup

PR #1348, opened 2020-08-05
Spreadsheet: fix alignment, +letter-spacing

### LINKED ISSUE #1333: Player finance columns do not line up with player columns above
**Describe the bug**
A clear and concise description of what the bug is. Include steps to reproduce the behavior. If applicable, add screenshots to help explain your problem.

**Please provide details about your system**
 - OS [e.g. iOS13, Win10] & Device [e.g. iPhone XS]
 - Browser & Version [e.g. Chrome 83]

Chrome browser.  Win10.  Player finance columns in 1882 do not line up with the player columns above; they are evenly spaced while the above columns have different spacing, resulting in offset.

yzemaze:
fixes #1333 

ante
image

post
image

### DIFF
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -34,10 +34,7 @@ module View
           h(:thead, render_title),
           h(:tbody, render_corporations),
           h(:thead, [
-            h(:tr, [
-              h(:th, ''),
-              h(:th, { attrs: { colspan: @game.players.size } }, 'Player Finances'),
-            ]),
+            h(:tr, { style: { height: '1rem' } }, ''),
           ]),
           h(:tbody, [
             render_player_cash,
@@ -91,9 +88,11 @@ module View
       end
 
       def render_title
+        spacing = ->(cols) { { attrs: { colspan: cols }, style: { letterSpacing: '0.4rem' } } }
+
         or_history_titles = render_history_titles(@game.corporations)
 
-        p_props = {
+        pd_props = {
           style: {
             background: 'salmon',
             color: 'black',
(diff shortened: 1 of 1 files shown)


dfannius:
The letter spacing looks pretty ugly to me, to be honest (see https://practicaltypography.com/letterspacing.html for an actual expert agreeing with me). Alternatives that would look okay to my eye:

- A horizontal line going across a whole group, under the group title
- Put a subtle background behind each group title (spanning the group)
- Put a subtle background behind each group (the whole table)

(I also actually prefer the original alignment, also, but don't feel strongly about it.)

yzemaze:
> I also actually prefer the original alignment, also, but don't feel strongly about it.

Me too, but there are definitely cases were left aligned share numbers and right aligned finances look horrible:
image
image

=> The issue is legit and this way it’s more consistent.

subtle bg-color (on headers only to prevent conflicts!) instead of letter-spacing sounds fine to me. In the end it’s toby’s choice.

assets/app/view/game/spreadsheet.rb
```diff
@@ -104,18 +103,18 @@ def render_title
         [
           h(:tr, [
             h(:th, ''),
-            h(:th, { attrs: { colspan: @game.players.size } }, 'Players'),
-            h(:th, { attrs: { colspan: 2 } }, 'Bank'),
-            h(:th, { attrs: { colspan: 2 } }, 'Prices'),
-            h(:th, { attrs: { colspan: 4 } }, 'Corporation'),
+            h(:th, spacing[@game.players.size], 'Players'),
```
tobymao, line 106: shouldn't you be calling the lambda with parens?

yzemaze: You tell me. I don’t mind changing these to `spacing.call(x)`.

tobymao:
hm ok it's correct, didn't know we can keep it

PR #1350, opened 2020-08-05
Allow users to see tile upgrade path on the tile page

### LINKED ISSUE #605: Some sort of visual indicator when there is no valid track for a hex
Right now, as far as I know, if you click on a hex with no valid updates, there is no visual feedback that there is no valid track. The hex could be greyed or reddened or something.

I realize that "lack of tiles" probably implies no tiles, but a) visual feedback is good and b) lately, the tile laying has been a little buggy for me. I'll see if it's a known issue, but regardless, sometimes there are valid tiles, but they appear out of view...in this case, when I click on a tile, I don't know how to differentiate immediately between "you can upgrade the tile you just have to go looking for where they are" (usually far on the bottom left), and "you cannot upgrade this tile".

### LINKED ISSUE #865: See full upgrade path on the tiles view
Hi, I think it would be nice to see which tile is an upgrade for which other one. For example in the "tile-tab", if you would click on green tile #28, that you would shade out all the brown tiles that don't fit on this green tile and keep for example #46 highlighted. Or alternatively, clicking on a tile in the "tile-tab" would make it turn around.
Furthermore, it would also be nice to be able to play around in the map a bit more with tiles, even though you may not have a legal connection, it is sometimes important to sort out which routes are possible in the long run, or which tiles may be used by your opponents, so you can anticipate better. I haven't really sorted out how this could be nicely implemented, but I would definitely use a feature like that.
Thanks for all the great work!

### LINKED ISSUE #1349: Tile selector should be removed when clicking generally on other tiles
**Describe the bug**
If you have the tile selector open and click on another tile it doesn't dismiss it, it makes it feel harder to dismiss than it should be.

jenf:
fixes #865
fixes #1349
fixes #605

As discussed removed "Illegal".

Screenshot from 2020-08-07 17-31-39


jenf:
Screenshot from 2020-08-05 17-19-23

assets/app/view/game/tile_manifest.rb
```diff
@@ -5,20 +5,70 @@
...
     class TileManifest < Tiles
-      needs :tiles
-      needs :all_tiles
-      needs :layout
+      needs :game
+      needs :tile_selector, default: nil, store: true
+
+      def potential_upgrades(tile)
+        colors = Array(@game.phase.phases.last[:tiles])
+        @game.init_tiles
```
tobymao, line 13: do we need to call init_tiles twice?

assets/app/view/game/tile_manifest.rb
```diff
@@ -5,20 +5,70 @@
...
+        remaining = @game.tiles.group_by(&:name)
 
-        children = @all_tiles.sort.group_by(&:name).flat_map do |name, tiles|
+        children = @game.init_tiles.sort.group_by(&:name).flat_map do |name, tiles|
           num = remaining[name]&.size || 0
           opacity = num.positive? ? 1.0 : 0.5
-          render_tile_blocks(name, tile: tiles.first, num: num, opacity: opacity, layout: @layout)
+          tile = tiles.first
+          extra_children = []
+          if @tile_selector&.hex&.tile&.name == tile.name
```
tobymao, line 28: let's move this into it's own function that returns an empty array or this

assets/app/view/game/hex.rb
```diff
@@ -77,7 +80,12 @@ def transform
       end
 
       def on_hex_click
-        return if @actions.empty?
+        return if @actions.empty? && @role != :tile_page
+
+        if !@clickable || @hex == @tile_selector&.hex
```
tobymao, line 85: could be a one liner return store(:tile_selector, nil) if

assets/app/view/game/history_controls.rb
```diff
@@ -46,6 +47,7 @@ def link(text, title, action_id = nil)
         route = Lib::Params.add(@app_route, 'action', action_id)
 
         click = lambda do
+          store(:tile_selector, nil)
```
tobymao, line 50: hm this is not the best, i wonder if there's any other way around it

jenf: Another solution would be to store the action id in the selector and ignoring if it doesn't match.

lib/engine/game/base.rb
```diff
@@ -1074,6 +1075,15 @@ def cache_objects
       def bankruptcy_limit_reached?
         @players.any?(&:bankrupt)
       end
+
+      def all_potential_upgrades(tile)
```
tobymao, line 1079: is this code in tracker?

jenf: It's very similar, but this doesn't limit on phase or if the tile is available

assets/app/view/game/history_controls.rb
jenf: Alternatively move the tile selector and selected_company into a 'clear_ui_state' method in actionable

lib/engine/game/base.rb
jenf: unsure what needs changing if any on this.

tobymao: i'm going to move potential tiles out of step

tobymao: eh maybe later

PR #1356, opened 2020-08-05
18AL: Add support for buying companies between players

perwestling:
This feature makes it possible for a player to buy a private company from another player during the stock round.

This should maybe be a draft instead?

There are two things remaining:
- The layout did not look good if a player buys the first company of two. So need help to get that correct.
- Need to fit this buy together with other so a player cannot buy more than one thing during an SR

### DIFF
--- a/assets/app/view/game/buy_company_from_other_player.rb
+++ b/assets/app/view/game/buy_company_from_other_player.rb
@@ -25,9 +25,9 @@ module View
           size: max,
         })
         children = [input]
-        @game.round.active_step.purchasable_companies_from_others(@game.current_entity).each do |company|
+        @game.round.active_step.purchasable_companies(@game.current_entity).each do |company|
           children << render_company_buy(input, company)
-        end if @game.round.active_step.can_buy_any_company?(@game.current_entity)
+        end
         h(:div, props, children)
       end
 
--- a/assets/app/view/game/buy_company_from_other_player.rb
+++ b/assets/app/view/game/buy_company_from_other_player.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+require_relative 'actionable'
+
+module View
+  module Game
+    class BuyCompanyFromOtherPlayer < Snabberb::Component
+      include Actionable
+
+      def render
+        props = {
+          style: {
+            grid: '1fr / 5rem 4rem',
+            alignItems: 'center',
+            margin: '1rem 0 1.5rem 0',
+          },
+        }
+
+        max = @game.current_entity.cash.to_s.size
+        input = h(:input, style: { marginBottom: '0.5rem', marginRight: '1rem' }, props: {
+          value: 1,
+          type: 'number',
+          min: 1,
+          max: max,
+          size: max,
+        })
+        children = [input]
+        @game.round.active_step.purchasable_companies_from_others(@game.current_entity).each do |company|
+          children << render_company_buy(input, company)
+        end if @game.round.active_step.can_buy_any_company?(@game.current_entity)
+        h(:div, props, children)
+      end
+
+      private
+
+      def render_company_buy(input, company)
+        buy_company = lambda do
+          price = Native(input).elm.value.to_i
+          process_action(Engine::Action::BuyCompany.new(
+            @game.current_entity,
+            company: company,
+            price: price,
+          ))
+        end
+        h(:button, { on: { click: buy_company } }, "Buy #{company.id} from #{company.owner.name}")
+      end
+    end
+  end
+end
--- a/assets/app/view/game/round/stock.rb
+++ b/assets/app/view/game/round/stock.rb
@@ -46,14 +46,10 @@ module View
 
           children.concat(render_corporations)
           children << h(Players, game: @game)
-          children << h(BuyCompanyFromOtherPlayer, game: @game) if buy_any_company?
+          children << h(BuyCompanyFromOtherPlayer, game: @game) if @step.purchasable_companies(@current_entity).any?
           children << h(StockMarket, game: @game)
 
           h(:div, children)
-        end
-
-        def buy_any_company?
-          @step.can_buy_any_company?(@current_entity) && @step.purchasable_companies_from_others(@current_entity).any?
         end
 
         def render_corporations
--- a/assets/app/view/game/round/stock.rb
+++ b/assets/app/view/game/round/stock.rb
@@ -46,9 +46,14 @@ module View
 
           children.concat(render_corporations)
           children << h(Players, game: @game)
+          children << h(BuyCompanyFromOtherPlayer, game: @game) if buy_any_company?
           children << h(StockMarket, game: @game)
 
           h(:div, children)
+        end
+
+        def buy_any_company?
+          @step.can_buy_any_company?(@current_entity) && @step.purchasable_companies_from_others(@current_entity).any?
         end
 
         def render_corporations
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -10,7 +10,7 @@ module Engine
     class BuySellParShares < Base
       include ShareBuying
 
-      PURCHASE_ACTIONS = [Action::BuyShares, Action::Par].freeze
+      PURCHASE_ACTIONS = [Action::BuyCompany, Action::BuyShares, Action::Par].freeze
 
       def actions(entity)
         return [] unless entity == current_entity
@@ -19,6 +19,7 @@ module Engine
         actions = []
         actions << 'buy_shares' if can_buy_any?(entity)
         actions << 'par' if can_ipo_any?(entity)
+        actions << 'buy_company' if can_buy_any_company?(entity)
         actions << 'sell_shares' if can_sell_any?(entity)
 
         actions << 'pass' if actions.any?
@@ -188,6 +189,17 @@ module Engine
         !bought? && @game.corporations.any? { |c| c.can_par?(entity) && can_buy?(entity, c.shares.first&.to_bundle) }
       end
 
+      def purchasable_companies_from_others(entity)
+        @game.purchasable_companies.reject { |c| c.owner == entity }
+      end
+
+      def can_buy_any_company?(entity)
+        !bought? &&
+        entity.cash.positive? &&
+        @game.phase.status.include?('can_buy_companies_from_other_players') &&
+        purchasable_companies_from_others(entity).any?
+      end
+
       def get_par_prices(entity, _corp)
         @game
           .stock_market
@@ -203,7 +215,24 @@ module Engine
       end
 
       def bought?
-        @current_actions.any? { |x| PURCHASE_ACTIONS.include?(x.class) }
+        @current_actions.any? { |x| self.class::PURCHASE_ACTIONS.include?(x.class) }
+      end
+
+      def process_buy_company(action)
+        entity = action.entity
+        company = action.company
+        price = action.price
+        owner = company.owner
+
+        @game.game_error("Cannot buy #{company.name} from #{owner.name}") unless owner.player?
+
+        company.owner = entity
+        owner.companies.delete(company)
+
+        entity.companies << company
+        entity.spend(price, owner)
+        @current_actions << action
+        @log << "-- #{entity.name} buys #{company.name} from #{owner.name} for #{@game.format_currency(price)}"
       end
     end
   end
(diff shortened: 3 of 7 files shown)


perwestling:
This is how it looks before and after you click on the top company.
So, no good.

Screenshot from 2020-08-06 00-53-11
Screenshot from 2020-08-06 00-53-33

tobymao:
Does this even need a separate step? Can't we just modify the original buy companies step to support this?

tobymao:
i don't think the button should be here, i think it will be easier for the ui if when you click on a company, it creates selected_company

and then you just put the button below the company to click buy

perwestling:
Improved the commit:
- Now the restriction of just one buy per SR does include the companies, so at max a player can buy one share or one company, not both
- The entry field and the button have now been moved to between player cards and the bank saldo / stockmarket

Remaining problem:
- Selection of companies in the player cards does not propagate to the BuyCompanyFromOtherPlayer component. Attempt was made to introduce a :selected_companies, which is updated when a company is selected to show description, but this does not seem to propagate from the Company component

perwestling:
Did not get the selection of company to propagate. And it was possibly a bit messy as you could select multiple companies, and also select companies owned by the corporation.

So instead I choose to present a line with one button for each purchasable company. But the line is only shown if the player has not yet bought anything this stock turn.

The button row is below the player cards, so it is similar to when happens when you click on a corporation.

Also, it is a bit down as it is a very rare action, and we much more often want to access the player information.

So I think this can be a Good Enough (TM) solution that has very little impact outside of 18AL. If more 18xx have this functionality, we can consider to make it more general.

Screenshot from 2020-08-06 14-56-07

perwestling:
> Does this even need a separate step? Can't we just modify the original buy companies step to support this?

It is of course possible to modify the original step. The main differences is in process_buy_company. 
But keeping it in a step under g_18_al means it does not complicate the main code with special cases.
When a similar functionality is needed for another 18XX we can think about refactoring it to incorporate this into the common code.

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module View
+  module Game
+    class BuyCompanyFromOtherPlayer < Snabberb::Component
+      include Actionable
+
+      needs :game, store: true
```
tobymao, line 8: do not list this as store since you're passing it in

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
```
tobymao, line 2: need to have require actionable to be safe here

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
...
+module View
+  module Game
+    class BuyCompanyFromOtherPlayer < Snabberb::Component
+      include Actionable
+
+      needs :game, store: true
+      needs :selected_companies, default: [], store: true
```
tobymao, line 9: why do you need selected_companies?

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
...
+      needs :selected_company, default: nil, store: true
+
+      def render
+        @step = @game.round.active_step
+        @current_actions = @step.current_actions
+        return h(:div, {}, '') if !@step.can_buy_any?(@game.current_entity) || !@current_actions.include?('buy_company')
```
tobymao, line 15: can just do h(:div) i think? or maybe h(:div, '')

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
...
+        input = h(:input, style: { marginBottom: '0.5rem', marginRight: '1rem' }, props: {
+          value: 1,
+          type: 'number',
+          min: 1,
+          max: @game.current_entity.cash.to_s.size,
+          size: @game.current_entity.cash.to_s.size,
+        })
+        children = [input]
+        buyable_companies
+          .select { |c| purchasable?(c) }
```
tobymao, line 37: why are you doing select purchasable twice?

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
...
+            alignItems: 'center',
+            margin: '1rem 0 1.5rem 0',
+          },
+        }
+
+        input = h(:input, style: { marginBottom: '0.5rem', marginRight: '1rem' }, props: {
+          value: 1,
+          type: 'number',
+          min: 1,
+          max: @game.current_entity.cash.to_s.size,
```
tobymao, line 32: maybe store this as a var

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
...
+          .select { |c| purchasable?(c) }
+          .each { |c| children << render_company_buy(input, c) }
+        h(:div, props, children)
+      end
+
+      private
+
+      def purchasable?(company)
+        @game.class::COMPANY_BUY_BETWEEN_PLAYERS &&
+        company&.owner&.player? &&
```
tobymao, line 46: why would company be nil?

assets/app/view/game/buy_company_from_other_player.rb
```diff
@@ -0,0 +1,64 @@
...
+        company&.owner != @game.current_entity
+      end
+
+      def render_company_buy(input, company)
+        buy_company = lambda do
+          price = input.JS['elm'].JS['value'].to_i
```
tobymao, line 52: you can do Native(input).elm.value.to_i

lib/engine/step/g_18_al/buy_sell_par_shares.rb
```diff
@@ -6,11 +6,19 @@ module Engine
...
+        PURCHASE_ACTIONS = [Action::BuyCompany, Action::BuyShares, Action::Par].freeze
+
         def setup
           super
           @controller_at_start = current_controller
         end
 
+        def actions(entity)
+          actions = super
+          actions << 'buy_company' if actions.include?('buy_shares')
```
tobymao, line 18: hm, shouldn't you only add buy_company action if there are companies to buy? then you wouldn't need to do all those checks in the view that you do

assets/app/view/game/round/stock.rb
```diff
@@ -46,6 +46,7 @@ def render
 
           children.concat(render_corporations)
           children << h(Players, game: @game)
+          children << h(BuyCompanyFromOtherPlayer, game: @game) if @game.class::COMPANY_BUY_BETWEEN_PLAYERS
```
tobymao, line 49: you can get rid of all those return h(:div, '') if the actions returns buy company, it will be cleaner

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Fixed

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Fixed

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Removed in latest version

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Need for this removed in later revision

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Ops. Fixed

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Fixed

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Hm, fixed

assets/app/view/game/buy_company_from_other_player.rb
perwestling: Fixed

assets/app/view/game/round/stock.rb
perwestling: Yes, it became much cleaner after applying your suggestion

lib/engine/step/g_18_al/buy_sell_par_shares.rb
perwestling: Rewritten this quite a bit, but now the buy_company will only be available if there is a legal buy.

perwestling:
Have now incorporated review comments by @tobymao 

Did a rebase and force push as I rewrote this a bit. Now use the event status feature.
Also have moved the handling of buy_company into the general buy_sell_par_shares class.
Was not able to get adding buy_company to the stock round.

In this case I think buy_company belongs to the buy_sell_par_shares as otherwise we need to introduce a state that would be shared between buy_sell_par_shares and buy_company (@current_actions)

Even if doing that, using the normal buy_company did introduce some special things. Think I needed to check if OR or SR at some places.

Anyway I did not get it to work, but when I tried I think the code was not fully correct yet. So should I give it a new try to make buy_company work for both OR and SR? Introducing a shared round state between buy_company and buy_sell_par_shares? (The shared state is to keep track if anything has been bought during the SR.)

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -188,6 +189,17 @@ def can_ipo_any?(entity)
         !bought? && @game.corporations.any? { |c| c.can_par?(entity) && can_buy?(entity, c.shares.first&.to_bundle) }
       end
 
+      def purchasable_companies_from_others(entity)
+        @game.purchasable_companies.reject { |c| c.owner == entity }
+      end
+
+      def can_buy_any_company?(entity)
```
tobymao, line 196: i think this is not necessary you can just put this all in purchasebl_companies

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -188,6 +189,17 @@ def can_ipo_any?(entity)
         !bought? && @game.corporations.any? { |c| c.can_par?(entity) && can_buy?(entity, c.shares.first&.to_bundle) }
       end
 
+      def purchasable_companies_from_others(entity)
```
tobymao, line 192: return [] if bought? || !entity.cash.positive?...

@game.purchasable_companies.reject...

this is how you can consolidate can_buy_any_company?

tobymao:
> Have now incorporated review comments by @tobymao
> 
> Did a rebase and force push as I rewrote this a bit. Now use the event status feature.
> Also have moved the handling of buy_company into the general buy_sell_par_shares class.
> Was not able to get adding buy_company to the stock round.
> 
> In this case I think buy_company belongs to the buy_sell_par_shares as otherwise we need to introduce a state that would be shared between buy_sell_par_shares and buy_company (@current_actions)
> 
> Even if doing that, using the normal buy_company did introduce some special things. Think I needed to check if OR or SR at some places.
> 
> Anyway I did not get it to work, but when I tried I think the code was not fully correct yet. So should I give it a new try to make buy_company work for both OR and SR? Introducing a shared round state between buy_company and buy_sell_par_shares? (The shared state is to keep track if anything has been bought during the SR.)

so is this pr working or missing something? it seems like you're saying it doesn't work in the OR?

lib/engine/step/buy_sell_par_shares.rb
perwestling: Done

lib/engine/step/buy_sell_par_shares.rb
perwestling: Have made an attempt now. Using name purchasable_companies in the step.

perwestling:
> so is this pr working or missing something? it seems like you're saying it doesn't work in the OR?

It does work and passes the rake test for 18AL (which has added two purchases in SR1).

What does not (or rather did not) work is (was) to use buy_company as a step in StockRound.

So the question is if I should make another try? It might complicate the buy_company code.

An alternative could be to extract the process_buy_company code to a shared module, which then buy_company and buy_sell_par_shares could share.

perwestling:
It should be noted that this PR will break existing alpha 18AL games, as players will be needing to pass during SR. If a player have $1 up to less than cheapest share, in SR1, they before were passed automatically but now have to manually pass, as there are purchasable privates. 

But this is the last needed feature to make 18AL feature complete, and maybe ready for Beta.

PR #1368, opened 2020-08-06
Adjust multi_revenue placement; Add support for terminal paths for off-board cities fixes #1290

### LINKED ISSUE #1290: 18AL Mobile city rendering
The name of the city of Mobile (Q2) overlaps the values.

roseundy:
Fixes #1290 
Adjusted multi_revenue placement when in tow or bottom row.
Added support for "terminal:1" parameter to path in tile definition. Will create tapered track to city.

### DIFF
--- a/assets/app/view/game/part/revenue.rb
+++ b/assets/app/view/game/part/revenue.rb
@@ -31,12 +31,12 @@ module View
               {
                 region_weights: TOP_ROW,
                 x: 0,
-                y: -48,
+                y: -72,
               },
               {
                 region_weights: BOTTOM_ROW,
                 x: 0,
-                y: 48,
+                y: 69,
               },
             ]
           else
--- a/assets/app/view/game/part/revenue.rb
+++ b/assets/app/view/game/part/revenue.rb
@@ -14,30 +14,30 @@ module View
           if multi_revenue?
             [
               {
-                region_weights: CENTER,
+                region_weights: { CENTER => 1.5 },
                 x: 0,
                 y: 0,
               },
               {
-                region_weights: TOP_MIDDLE_ROW,
+                region_weights: { TOP_MIDDLE_ROW => 1.5 },
                 x: 0,
-                y: -24,
+                y: -48,
               },
               {
-                region_weights: BOTTOM_MIDDLE_ROW,
+                region_weights: { BOTTOM_MIDDLE_ROW => 1.5 },
                 x: 0,
-                y: 24,
+                y: 45,
               },
-              {
-                region_weights: TOP_ROW,
-                x: 0,
-                y: -72,
-              },
-              {
-                region_weights: BOTTOM_ROW,
-                x: 0,
-                y: 69,
-              },
+              # {
+              #   region_weights: { TOP_ROW => 1.5 },
+              #   x: 0,
+              #   y: -72,
+              # },
+              # {
+              #   region_weights: { BOTTOM_ROW => 1.5 },
+              #   x: 0,
+              #   y: 69,
+              # },
             ]
           else
             SMALL_ITEM_LOCATIONS
--- a/assets/app/view/game/part/track_lawson_path.rb
+++ b/assets/app/view/game/part/track_lawson_path.rb
@@ -38,31 +38,25 @@ module View
         def render_part
           rotation = 60 * @edge_num
 
-          props =
-            if @terminal == 1
-              {
-                attrs: {
-                  transform: "rotate(#{rotation})",
-                  d: 'M6 60 L 6 85 L -6 85 L -6 60 L 0 25 Z',
-                  fill: @color,
-                  stroke: 'none',
-                  'stroke-linecap': 'butt',
-                  'stroke-linejoin': 'miter',
-                  'stroke-width': @width.to_i * 0.75,
-                  'stroke-dasharray': @dash,
-                 },
-              }
-            else
-              {
-                attrs: {
-                  transform: "rotate(#{rotation})",
-                  d: 'M 0 87 L 0 0',
-                  stroke: @color,
-                  'stroke-width': @width,
-                  'stroke-dasharray': @dash,
-                },
-              }
-            end
+          props = {
+            attrs: {
+              transform: "rotate(#{rotation})",
+              d: 'M 0 87 L 0 0',
+              stroke: @color,
+              'stroke-width': @width,
+              'stroke-dasharray': @dash,
+            },
+          }
+
+          props[:attrs].merge!(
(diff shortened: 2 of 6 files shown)


tobymao:
can you attach before and after pics of what this changed?

assets/app/view/game/part/revenue.rb
```diff
@@ -31,12 +31,12 @@ def preferred_render_locations
               {
                 region_weights: TOP_ROW,
                 x: 0,
-                y: -48,
+                y: -72,
```
tobymao, line 34: does this change the region_weights?

assets/app/view/game/part/track_lawson_path.rb
```diff
@@ -37,15 +38,31 @@ def preferred_render_locations
...
-            attrs: {
-              transform: "rotate(#{rotation})",
-              d: 'M 0 87 L 0 0',
-              stroke: @color,
-              'stroke-width': @width,
-              'stroke-dasharray': @dash,
-            },
-          }
+          props =
+            if @terminal == 1
```
tobymao, line 42: i think terminal should be a boolean value

you could also just do

props = { ... }

props[attrs].merge(d: ...) if @terminal

lib/engine/part/path.rb
```diff
@@ -5,12 +5,13 @@
 module Engine
   module Part
     class Path < Base
-      attr_reader :a, :b, :branch, :city, :edges, :junction, :node, :offboard, :stop, :town
+      attr_reader :a, :b, :branch, :city, :edges, :junction, :node, :offboard, :stop, :town, :terminal
 
-      def initialize(a, b)
+      def initialize(a, b, terminal = 0)
```
tobymao, line 10: let's change to be terminal: nil

lib/engine/part/path.rb
```diff
@@ -5,12 +5,13 @@
...
         @a = a
         @b = b
         @edges = []
+        @terminal = terminal
```
tobymao, line 14: @terminal = !!terminal

lib/engine/tile.rb
```diff
@@ -63,15 +63,20 @@ def self.part(type, params, cache)
       case type
       when 'path'
         params = params.map do |k, v|
-          case v[0]
-          when '_'
-            [k, cache[v[1..-1].to_i]]
+          case k
+          when 'terminal'
+            [k, v.to_i]
```
tobymao, line 68: no need to do a conversion here since any value w ill be truthy

lib/engine/tile.rb
```diff
@@ -63,15 +63,20 @@ def self.part(type, params, cache)
...
+            when '_'
+              [k, cache[v[1..-1].to_i]]
+            else
+              [k, Part::Edge.new(v)]
+            end
           end
         end.to_h
 
-        Part::Path.new(params['a'], params['b'])
+        Part::Path.new(params['a'], params['b'], params['terminal'])
```
tobymao, line 79: will need to add terminal: ...

assets/app/view/game/part/revenue.rb
roseundy: No. I originally thought it was a problem with the weights, but the actual placement needed to be lined up with what the weights were saying.

roseundy:
Before:
old Nashville
Mobile

After:
new Nashville
new Mobile

lib/engine/tile.rb
roseundy: I'm afraid I don't understand, why won't params['terminal'] work?

lib/engine/tile.rb
roseundy: OK

roseundy:
I realized that I should add an update to TILES.md to this as well.

assets/app/view/game/part/track_lawson_path.rb
roseundy: OK, I'll change to a boolean. It's a pretty drastic change to attrs: so I don't know if the merge would be as pretty.

roseundy: But I will defer to your judgement.

lib/engine/tile.rb
roseundy: Are you saying we don't directly pass in the value for terminal, but rather the hash {terminal: value}?

assets/app/view/game/part/revenue.rb
roseundy: No. Just the position.

lib/engine/part/path.rb
roseundy: OK

lib/engine/part/path.rb
roseundy: OK

lib/engine/part/path.rb
roseundy: Wait, you mean terminal = nil, right?

roseundy: Gotcha: keyword argument.

lib/engine/tile.rb
roseundy: Nevermind. Keyword argument.

roseundy:
After looking at this more, I wasn't happy with how multi_revenue appears when it is on the very top or bottom of a tile. I'm removing those as an option from preferred_location and increasing their weight slightly so they will bias the location name away from them more reliably.

Meridian with originally proposed changes:
newMeridian


With latest changes to multi_revenue preferred location:
newnewNashville
newnewMobile
newnewMeridian

assets/app/view/game/part/revenue.rb
roseundy: But my latest commit does change the region_weights: reduced number of locations (to avoid having revenue appear at the very top or bottom of a tile) and a greater than unity weight (to bias location text from appearing on top of the revenue).

tobymao:
this still lets you run through the city though right? i don't see anywhere in the code preventing you from running through

assets/app/view/game/part/revenue.rb
```diff
@@ -14,30 +14,30 @@ def preferred_render_locations
...
                 y: -48,
               },
               {
-                region_weights: BOTTOM_ROW,
+                region_weights: { BOTTOM_MIDDLE_ROW => 1.5 },
                 x: 0,
-                y: 48,
+                y: 45,
               },
+              # {
```
tobymao, line 31: you shouldn't leave commented out code, i've removed it for you

roseundy:
This change has no effect other than rendering. Currently, 18AL has special code to prevent running through the two offboard cities. Now that paths can be marked as "terminal", this could be generalized.

PR #1372, opened 2020-08-06
Special tile lay fixes

### LINKED ISSUE #1173: 1846: MC and O&I allowing illegal tile lays
When purchased by a corporation, the MC and O&I allow you to build two free bonus tiles at any time in marked hexes. According to the game rules, if you lay both tiles using this power, they must connect to one another. 

In game 4811, the MC owner was able to lay two disconnected tiles in the MC hexes (two sharp curves, not connected to each other). We had them undo it, but this is against the game rules. You may lay one or both tiles in the hexes, but if you lay both, they must connect (also, if you choose to only lay one, I believe you don't get to choose to lay the other one for free on a subsequent operating round; it must connect and you must pay at that point). 

I tested this in a hotseat game and both MC and O&I were allowing disconnected tile-lays after the private was bought in. I believe in all instances I observed this behavior, it happened after the Buy Companies step (which is legal; you can have a corporation buy in a private and then exercise its power as the last part of an operating turn in 1846).

### LINKED ISSUE #1276: 1846: O&I Can be used during an SR to lay tracks
Yep, some dude decided to do this instead of buying a share.  Good tester!

```
    {
      "type": "buy_company",
      "entity": "PRR",
      "entity_type": "corporation",
      "id": 298,
      "company": "O&I",
      "price": 40
    },
...
    {
      "type": "pass",
      "entity": "twotwotrainer",
      "entity_type": "player",
      "id": 327
    },
    {
      "type": "lay_tile",
      "entity": "O&I",
      "entity_type": "company",
      "id": 330,
      "hex": "F16",
      "tile": "8-6",
      "rotation": 1
    },
    {
      "type": "lay_tile",
      "entity": "O&I",
      "entity_type": "company",
      "id": 331,
      "hex": "F14",
      "tile": "9-5",
      "rotation": 1
    },
    {
      "type": "pass",
      "entity": "bstover",
      "entity_type": "player",
      "id": 332
    }
  ],
  "loaded": true,
  "created_at": 1596163045,
  "updated_at": 1596168821
}
```

cheesinglee:
Fixes #1276 
Fixes #1173 

It seems the only thing left in `Step::G18Chesapeake::SpecialTrack` after this change is the modified blocking logic, so maybe that could be factored out into another flag and the subclass eliminated.


lib/engine/step/g_18_chesapeake/special_track.rb
```diff
@@ -14,38 +14,13 @@ def actions(entity)
           ACTIONS
         end
 
-        def description
```
tobymao, line 17: why was this removed?

lib/engine/step/g_18_chesapeake/special_track.rb
```diff
@@ -14,38 +14,13 @@ def actions(entity)
...
-          "Lay Track for #{@company.name}"
-        end
-
         def active_entities
           @company ? [@company] : super
         end
 
         def blocks?
```
tobymao, line 21: this can be easily factored out and should be with this PR

lib/engine/step/special_track.rb
```diff
@@ -48,6 +53,27 @@ def ability(entity)
           entity == @round.just_sold_company
         ability || entity.abilities(:tile_lay, 'track')
       end
+
+      def check_connect(action)
+        company = action.entity
+        a = ability(company)
+        hex_ids = a.hexes
+        return unless a&.connect && hex_ids.size > 1
```
tobymao, line 61: never use unless with 2 booleans

lib/engine/step/special_track.rb
```diff
@@ -48,6 +53,27 @@ def ability(entity)
           entity == @round.just_sold_company
         ability || entity.abilities(:tile_lay, 'track')
       end
+
+      def check_connect(action)
+        company = action.entity
+        a = ability(company)
```
tobymao, line 59: call this track_ability or something

lib/engine/step/g_18_chesapeake/special_track.rb
tobymao: i see because you put it in special

lib/engine/step/special_track.rb
```diff
@@ -48,6 +53,27 @@ def ability(entity)
...
+        hex_ids = a.hexes
+        return unless a&.connect && hex_ids.size > 1
+
+        if company == @company
```
tobymao, line 63: doesn't this fail if the company doesn't lay the two tracks together? what if two companies are doing this separately

cheesinglee:
I've factored out the blocking logic and eliminated the 18Chesapeake subclass for the special track step. I also implemented logic to forfeit the second tile lay for MC/O&I if tile lays are interrupted by a different action. Finally, since this PR adds a couple new attributes to the tile lay ability, I've started a file that documents all the different abilities and the attributes they take.

lib/engine/step/special_track.rb
cheesinglee: Of the privates we have currently implemented that lay more than one track, all of them either block until all tiles are laid, or lose the ability if the track lays are interrupted, so I don't think we'll run into the issues you raised. I admit that's a weak defense though. Can you propose a different approach?

michaeljb:
I see this PR adds documentation about abilities; I haven't had time to review yet, but I wouldn't be surprised if #1396 wants to add/change something.

lib/engine/step/special_track.rb
tobymao: i don't think your code does this properly

Issue #1353, opened 2020-08-05
[1889] Train could run illegal route


augustoribeiro:
SR company could run a train on an apparently illegal path where the track section on the D9-E8 hex-side was used twice
![Screenshot 2020-08-05 at 12 52 25](https://user-images.githubusercontent.com/2705049/89450921-aa74d280-d75b-11ea-96c5-307b19e35423.png)

![Screenshot 2020-08-05 at 10 28 01](https://user-images.githubusercontent.com/2705049/89450822-79949d80-d75b-11ea-9f9a-cdb0269905aa.png)

https://www.18xx.games/game/5033?action=713

Please request if you need any extra info. Best regards.

jenf:
I don't think this will be specifically 1889.

cmaggard:
I've been taking a look into this- definitely not specific to 1889.

timbovelander:
We ran into the same bug, but with multiple trains in 18chesapeake.

![IMG-20200913-WA0001](https://user-images.githubusercontent.com/2109937/93016404-37b01000-f5c1-11ea-9e0e-45f1ec2c7144.jpg)

PR #1377, opened 2020-08-06
Track Reuse Bugfix [#1353]

cmaggard:
Two separate commits:

* A fixture loader to clean up some boilerplate that I had copied from the initial game spec when loading JSON game data (optional)
* An edge reuse detector that runs inline with the other route error checks.

Feedback is greatly appreciated as this is my first open-source contribution in a while and my skills are rusty.

### DIFF
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -34,6 +34,10 @@ module Engine
         on.keys.select { |p| on[p] == 1 }
       end
 
+      def edge_set
+        exits.map { |e| [hex.name, e].join('-') }
+      end
+
       def walk(skip: nil, visited: nil, on: nil)
         return if visited&.[](self)
 
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -34,10 +34,6 @@ module Engine
         on.keys.select { |p| on[p] == 1 }
       end
 
-      def edge_set
-        exits.map { |e| [hex.name, e].join('-') }
-      end
-
       def walk(skip: nil, visited: nil, on: nil)
         return if visited&.[](self)
 
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -206,11 +206,14 @@ module Engine
     end
 
     def check_edge_reuse!
-      dupes = collected_used_edges.group_by(&:itself).select { |_k, v| v.length > 1 }
+      used_edges = paths.flat_map do |path|
+        path.exits.map { |edge| [path.hex, edge] }
+      end
+      dupes = used_edges.group_by(&:itself).select { |_k, v| v.length > 1 }
       return if dupes.empty?
 
-      hexes = dupes.map(&:first).join(',')
-      @game.game_error("Routes cannot reuse the same sections of track: #{hexes.inspect}")
+      hexes = dupes.keys.map(&:first).map(&:name).sort.join(',')
+      @game.game_error("Routes cannot reuse the same sections of track: #{hexes}")
     end
 
     def check_connected!(token)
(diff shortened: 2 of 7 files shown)


lib/engine/part/path.rb
```diff
@@ -34,6 +34,10 @@ def select(paths)
         on.keys.select { |p| on[p] == 1 }
       end
 
+      def edge_set
```
tobymao, line 37: i don't think we should do this, it's enough to just have the edge object which is unique, this just creates a more inefficient representation of this

so if you just do path.edges, that returns all you need to know

spec/lib/engine/games/game_spec.rb
```diff
@@ -147,16 +147,23 @@ module Engine
...
         context game_id do
           it 'matches result exactly' do
-            game_path = game.title.gsub(/(.)([A-Z])/, '\1_\2').downcase
-            data = JSON.parse(File.read("spec/fixtures/#{game_path}/#{game_id}.json"))
-            players = data['players'].map { |p| p['name'] }
-            expect(game.new(players, id: game_id, actions: data['actions']).result).to eq(result)
-            rungame = game.new(players, id: game_id, actions: data['actions'], strict: true)
-            expect(rungame.result).to eq(result)
-            expect(rungame.finished).to eq(true)
+            loaded_game = load_game_fixture(game.title, game_id)
```
tobymao, line 150: i don't think we need to do this whole thing, would be cleaner for this one of situation to jmake a pure unit test

lib/engine/part/path.rb
tobymao: ah maybe you just want the exit number, in that case, i still wouldn't do it like this,

just check it manually in the route

hex_edges = paths.flat_map do |path|
path.exits.map { |edge| [hex, edge] }
}

tobymao:
i'm not sure how we would do it because we don't support it now, but how would you support 

https://www.boardgamegeek.com/image/4188536/1822ca

spec/lib/engine/games/game_spec.rb
cmaggard: That's fair, I was just thinking down the road of loading other 'erroneous' game states as checks. I can remove this if desired.

lib/engine/part/path.rb
cmaggard: Good call- I was trying to look for something like this. Will fix up. :+1:

cmaggard:
I'm still pretty new to the codebase so I'm not 100% sure on a suggestion for double track exits from a single edge. I did see some suggestions in the slack on additional edge numbering but as the comment pointed out, I don't know if there's any modulus stuff going on in the codebase to limit the range to 1-6.

spec/lib/engine/games/game_spec.rb
cmaggard: If it helps your thoughts on it some, I added some bits to make it lazy-evaluating, so we're not actually evaluating a JSON game state unless explicitly requested.

lib/engine/route.rb
```diff
@@ -205,6 +205,17 @@ def check_overlap!
       end
     end
 
+    def check_edge_reuse!
+      used_edges = paths.flat_map do |path|
+        path.exits.map { |edge| [path.hex, edge] }
+      end
+      dupes = used_edges.group_by(&:itself).select { |_k, v| v.length > 1 }
```
tobymao, line 212: always use size, v.size

tobymao: this can be simplified as

hexes = paths
  .flat_map { |path| path.exits.map ...}
  .group_by(&:itself)
  .select { |_, v| v.size > 1 }
  .map { |k, v| k.first.name }
  .sort
  .join(',')
  
@game.game_error('') if hexes.any?

spec/fixtures/1889/5033.json
```diff
@@ -0,0 +1,7766 @@
+{
```
tobymao, line 1: can we get rid of all the spec changes you made?

PR #1381, opened 2020-08-07
Operating round - don't move to next entity on messages

### LINKED ISSUE #1352: 1846 - Bankruptcy
Game 5715 - 1846, 4 player

player brootal went bankrupt - was president of 2 companies at the time it happened (NYC and PRR). Both went into receivership as per the rules - as in both cases there was only 1 stock held by another player. Both were operated by "Sharepool" according to the game history. PRR followed the rules - ran its train and withheld revenue. NYC didn't do that though - it doesn't appear that it ran its train, nor did it withhold. 

Not likely to happen a whole lot - but thought I'd let you know. 
-Tim

michaeljb:
[Fixes #1352]

Also fix order of log messages for corp in receivership to say

> Sharepool operates NYC
> NYC skips issue or redeem shares
> NYC skips place a token or lay track

instead of

> NYC skips issue or redeem shares
> NYC skips place a token or lay track
> Sharepool operates NYC

### DIFF
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -58,8 +58,8 @@ module Engine
         end
         entity.trains.each { |train| train.operated = false }
         @game.place_home_token(entity) if @home_token_timing == :operate
-        skip_steps
         @game.log << "#{entity.owner.name} operates #{entity.name}" unless finished?
+        skip_steps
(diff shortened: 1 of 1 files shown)


michaeljb:
This will break games like 5715 from #1352 -- I doubt it will be many.

PR #1383, opened 2020-08-07
Make 18TN playable

perwestling:
Feature complete implementation of 18TN.

About the commits, the only commit that changes the general code is the change needed to handle the Memphis upgrade. Introduces a method in base.rb (used in two tile laying cases) that can be implemented in the games to handle cases where label should be handled in a special way.


lib/engine/g_18_tn/share_pool.rb
```diff
@@ -0,0 +1,21 @@
...
+
+require_relative '../share_pool'
+
+module Engine
+  module G18TN
+    class SharePool < SharePool
+      def buy_shares(entity, shares, exchange: nil, exchange_price: nil)
+        super
+
+        return if @game.turn != 1
```
tobymao, line 11: return if !@game.turn.one? || (lnr != @game.company(..)
return unless lnr.owner

lnr.close!

lib/engine/game/g_18_tn.rb
```diff
@@ -13,27 +14,89 @@ class G18TN < Base
...
+
+      EVENTS_TEXT = Base::EVENTS_TEXT.merge(
+        'civil_war' => ['Civil War', 'Companies with trains loose revenue of one train its next OR']
+      ).freeze
 
       include CompanyPrice50To150Percent
 
       def setup
-        setup_company_price_50_to_150_percent
+        @corporations.each do |corporation|
```
tobymao, line 29: return unless (ic = @corporation.find { |c| c.id == 'IC' })

lib/engine/step/g_18_tn/buy_company.rb
```diff
@@ -0,0 +1,79 @@
...
+
+module Engine
+  module Step
+    module G18TN
+      class BuyCompany < BuyCompany
+        def can_buy_company?(entity)
+          companies = @game.purchasable_companies
+
+          entity == current_entity &&
+            (@game.turn == 1 || (@game.phase. available?('3') && @game.phase.buy_companies)) &&
```
tobymao, line 13: so you can buy companies if turn == 1 or available 3 and buy companies? i'm not exactly sure what's going on here

lib/engine/step/g_18_tn/buy_company.rb
```diff
@@ -0,0 +1,79 @@
...
+            companies.any? &&
+            companies.map(&:min_price).min <= entity.cash
+        end
+
+        def blocks?
+          @opts[:blocks]
+        end
+
+        def description
```
tobymao, line 22: this is not needed

lib/engine/step/g_18_tn/buy_company.rb
```diff
@@ -0,0 +1,79 @@
...
+          'Buy Companies'
+        end
+
+        def pass_description
```
tobymao, line 26: it looks like you just copy and pasted step/buy_company

can you clean this up?

lib/engine/tile.rb
```diff
@@ -206,7 +206,7 @@ def terrain
       @upgrades.flat_map(&:terrains).uniq
     end
 
-    def upgrades_to?(other, special = false)
+    def upgrades_to?(game, other, special = false)
```
tobymao, line 209: i don't think we should make this change or ghave label_check_override, 

didn't the other games just add the label or remove it specially?

lib/engine/g_18_tn/share_pool.rb
perwestling: Fixed

lib/engine/game/g_18_tn.rb
perwestling: Fixed (slightly rewritten)

lib/engine/step/g_18_tn/buy_company.rb
perwestling: By using event status, and introducing an event status for OR 1 buy company, I think the code became clearer.

lib/engine/step/g_18_tn/buy_company.rb
perwestling: Copy paste error.

lib/engine/step/g_18_tn/buy_company.rb
perwestling: Now it has been cleaned up.

lib/engine/tile.rb
perwestling: I have not come up with a better solution. Further comment below.

perwestling:
@tobymao I have now fixed the review comments, all but one.
Had to rebase vs upstream master, to get access to the phase status feature, so therefor did a force push.

But there is one feature in this commit that was not accepted. And it is the possibility of overriding the label check during tile lay.

Now adding labels (as is done here, and in 18AL) or changing labels (as is done here) handles most cases in 18TN.

But there is one hex that has a special one: Memphis.

It starts as a white city hex, revenue 20.
It can be upgraded to standard yellow. And later to standard green.
But for brown it can now use two different tile - a standard brown or a tile with a label. 
The label is shared with 2 other cities, and there are only 2 tiles with that brown label.

I have not found any solution to this except to change the standard code to let the game handle the label check.

But maybe it can be changed? One possibility is that the call back is asked if it will be reponsible for the label check. If it will, it will return true/false if label check OK. If it wont, the normal label check can be done.
So 18TN could then just do the label check for brown lay in hex H3 (Memphis) and for all other situation in can let the main code do the checking.

Do you have a better suggestion how to solve this use case?

lib/engine/g_18_tn/share_pool.rb
tobymao: this doesn't look fixed

lib/engine/tile.rb
tobymao: ok i've thought about it a bit. move upgrades_to? out of tile and into game, this will require shifting around some specs but i think it makes sense it will give the game full control over upgrade rules

perwestling: Ok, I will implement that change. I think it can be applied to 18AL as well - the special handling for Montgomery can then be done in its upgrades_to?

Using that (in 18AL) I suppose it does not need to add any labels. Instead the upgrades_to? will make the green M tile available when upgrading to Green, and exclude the non M green tiles.

And the Wiki page should be updated for 18AL (and 18TN) to reflect this.

I will probably do this implementation later today. I intend to create a spec file first, so that it will manage regression test for 18TN when doing this change. Should be possible to add these commits. Do not think there is any conflicts in master.
(Assuming there is not problem with other spec files - we will see.)

lib/engine/g_18_tn/share_pool.rb
perwestling: Ah, another try has been made.

perwestling:
@tobymao Have now added 4 commits that fix review comments, refactor upgrades_to?, corrects some minor bugs and adds a spec file for 18TN.

PR #1388, opened 2020-08-08
lib/truncate

yzemaze:
Default truncation at 15 chars (=> rendering: 12 + ...) seems useful in our case to render usernames the same in multiple places with less hassle.
It’s only necessary for elements where it’s not possible to rely on CSS (cf. `.nowrap`), because they aren’t contained in an element with a max-width.


tobymao:
fix conflict

PR #1389, opened 2020-08-08
entity order: divs => ul/li, pruning

yzemaze:
ul not just because of semantics, it will come in handy when a vertical (space-saving) order is needed.
btw: I didn’t miss the borders on dark bg until now. Do we need them?


tobymao:
dunno

PR #1393, opened 2020-08-08
fix dumping presidency when presidency is 30% instead of 20%

### LINKED ISSUE #1376: Support other presidency percentage than 20
20% is very common for presidency shares. But there are a bunch that has other numbers.

One of these are 18TN which has a 30% presidency share.

It would be good if it was possible to configure per game.

michaeljb:
The new code outside of Engine::Corporation#initialize should work for any size
of share corporations (e.g., 5 share corporations in 1817, 4 share corporations
in 18CZ), regardless of how many shares a president certificate is.

[Fixes #1376]

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -29,10 +29,13 @@ module Engine
     def initialize(sym:, name:, **opts)
       @name = sym
       @full_name = name
-      [
+      shares = [
         Share.new(self, president: true, percent: 20),
         *8.times.map { |index| Share.new(self, percent: 10, index: index + 1) },
-      ].each { |share| shares_by_corporation[self] << share }
+      ]
+      shares.each { |share| shares_by_corporation[self] << share }
+      @presidents_share = shares.first
+      @second_share = shares[1]
 
       @share_price = nil
       @par_price = nil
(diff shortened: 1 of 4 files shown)


michaeljb:
@perwestling would you mind doing a quick check of this branch with 18TN?

michaeljb:
I think it would be best for 2p 18Chesapeake, and smaller corporations in 1817, 18CZ, etc., to define how many shares a corp has and how many shares are on the president certificate at the time of initialization (i.e., as args to `Corporation#initialize`).

For 2p Chesapeake, that would require selecting which corp Cornelius is assigned to before the Corporation object is initialized, which seems doable but maybe messy? And I figured doing anything with other sized corporations for other games was out of scope for this.

michaeljb:
I can set up some 2p Chesapeake unit tests for this later, need to do some house work for now...

michaeljb:
The promised unit test - https://github.com/tobymao/18xx/commit/5241a9f73b228d81a3ded64d783e7e3c19fb5206

perwestling:
@michaeljb Seems to work with 18TN as well.

michaeljb:
@perwestling that's great! Thanks for taking a look.

PR #1395, opened 2020-08-09
Fix revenue position on pointy maps

roseundy:
1. Swap transformation on revenue for pointy maps to allow revenue to be properly horizontally centered on tiles
2. Add correct multi_revenue preferred_render_location for pointy maps
3. Update TILES.md for border options and minor formatting issue with last edit

With old revenue position (on 18TN):
Birmingham_old

With this change:
Birmingham_new


roseundy:
I would have added issue #1395, but it had been already closed.

PR #1396, opened 2020-08-09
Allow multiple abilities of same type; add per-OR abilities

### LINKED ISSUE #1316: [1846] Steamboat company can be assigned multiple times in an OR while owned by a player
Per the GMT rules:

> **Meat Packing Company and the Steamboat Company**: After being purchased, the corporation owning each
of these companies may place—or shift with the Steamboat Company—a marker among the listed locations, adding its value to the corporation’s routes that count this location. Use the second marker for the corporation after removing the company in phase III (before these markers are removed in phase IV).
>
> Before the Steamboat Company is purchased by a corporation, its owner can assign its marker **during the private income step** to a corporation or Independent Railroad, using its second marker to indicate this. Remove both markers once it is purchased.

As currently implemented, each operating company may use the SBC ability to re-assign the SBC token to itself during an OR, as long as it is owned by a player.

**Steps to reproduce:**

1. Load Gist: https://gist.github.com/wcraigtrader/17bada28517e0038eb046934d67c4329
  The MS has already assigned the SBC to itself, at Toledo.
2. Currently NYC is operating, have NYC assign SBC to itself.

### LINKED ISSUE #1344: [1846] Steamboat company can be assigned to a different corporation while owned by a corporation
**Describe the bug**

Once the SBC is owned by a corp, it should not be re-assignable.

**Steps to reproduce:**

1. Load Hotseat Gist: https://gist.github.com/michaeljb/5fcbcb7ad595da2a99ae33ccdf5182f8
    B&O already owns the SBC.
2. Click the SBC, move to Entities tab, click on another corporation
3. View the log to see the assignment, set up a route for B&O to see it no longer gets the bonus

### LINKED ISSUE #1359: [1846] SBC token can be moved to a new location too many times
As of #1358, the Steamboat company may be re-assigned too many times. While owned by a player, it should be assignable once per OR (before any corporations operate to be _strictly_ accurate); while owned by a corporation, it should be assignable once.

michaeljb:
For 1846's Steamboat company, split the assign abilities into separate
abilities when player/corporation-owned, and add the proper restriction on how
many times it can be re-assigned.

[Fixes #1316]
[Fixes #1344]
[Fixes #1359]

This probably breaks some games.


michaeljb:
This still does not fully resolve the Steamboat Company, see #1397

michaeljb:
This probably breaks some games.

lib/engine/abilities.rb
```diff
@@ -10,46 +10,53 @@
...
-
         ability.owner = self
-        @abilities[ability.type] = ability
+        @abilities << ability
       end
     end
 
-    def abilities(type, time = nil)
-      return nil unless (ability = @abilities[type])
+    def ability_by_type(type)
```
tobymao, line 23: i don't like how we can't handle duplicate abilities i wonder if there's a way around  this?

michaeljb: How about this? f2844fdd

I think code that calls `abilities` will need to be changed when we have some corp/company that has duplicate abilities, but `abilities` itself should do the right thing.

michaeljb: These two lines are included for backward compatibility without having to change other code, ideally we'd just return `active_abilities`; that could be fixed when we need to actually implement duplicate abilities - https://github.com/tobymao/18xx/pull/1396/files#diff-9abc48bc27003a6922f597d541dc4f3eR48-R49

michaeljb:
I will update the ability README added in  #1372

PR #1401, opened 2020-08-09
[1882] link directly to rules pdf

benjaminxscott:


jenf:
@scottredracecar will this be the final link? Previously you didn't want the doc linked directly

scottredracecar:
That revised link will break shortly (probably within one week)

PR #1402, opened 2020-08-09
Correct 1817 map

etowle:
* Add location name to New York
* Change Chicago and Atlanta revenue to 30/50/60/80
* Add impassable border northeast of Cleveland
* Fix track on yellow New York tile


jenf:
Changes match hard copy.

PR #1411, opened 2020-08-10
1817 stock market

jenf:


lib/engine/share_price.rb
```diff
@@ -2,51 +2,56 @@
...
                    unlimited_colors: [],
                    multiple_buy_colors: [])
       @coordinates = coordinates
       @price = price
       @color = color
-      @can_par = can_par
-      @end_game_trigger = end_game_trigger
+      @type = type
+      @can_par = type == :par
+      @end_game_trigger = type == :endgame
```
tobymao, line 52: do we need vars for all of these? can we just have @type?

tobymao: or perhaps a stock market space could have multiple types, not sure if that's needed now

jenf: I can mod the existing ones to not need the variables.

lib/engine/share_price.rb
```diff
@@ -2,51 +2,51 @@
...
+
+      color, type =
         case
-        when can_par
-          :red
-        when end_game_trigger
-          :blue
-        when code.include?('blk')
-          :black
+        when code.include?('p')
```
tobymao, line 14: cleaner as a regex

case code
when /p/
when /c/
...

PR #1414, opened 2020-08-11
A new attempt to make 18MS, first part 

perwestling:
Remaining are at least:

- [x] Implement 2+, 3+, 4+ trains
- [x] Rusting after OR 2.2, 3.2, 4.2
- [x] Game end after 10 ORs (actually after 5.2, as we use 1.1, 1.2, 2.1, 2.2, ... 5.1, 5.2 END
- [x] Fix map for New Orleans and Atlanta
- [ ] Private Alabama Great Southern Railroad
- [ ] Private Birmingham Southern Railroad
- [ ] Private Meridian & Memphis Railway
- [ ] Private Mississippi Central Railway
- [ ] President can (maybe) help when buying permanent trains
- [ ] Bonus for Atlanta tokenize
- [ ] Bonus for Chattanooga connection
- [ ] Restrict Atlanta revenue to once per run
- [ ] Loan during emergency buy
- [ ] Illegal to buy train that has run during OR, if it will rust at the end of the OR (idea: set them as obsolete and then rust obsoleted trains)
- [x] Corporations can lay 2 yellow tiles during its first OR
- [x] P1/P2 can be bought at face value OR 1
- [x] Purchase of private companies can only be done from those that the president own
- [ ] D phase should be triggered by purchase of any D train, not just 2D
- [x] 70% is max ownership in company
- [x] Update wiki page and add link to it
- [ ] Optional: Timeline presentation, for events and status

### DIFF
--- a/lib/engine/game/g_18_ms.rb
+++ b/lib/engine/game/g_18_ms.rb
@@ -12,7 +12,9 @@ module Engine
       GAME_LOCATION = 'Mississippi, USA'
       GAME_DESIGNER = 'Mark Derrick'
       GAME_PUBLISHER = Publisher::INFO[:all_aboard_games]
-      GAME_END_CHECK = { final_or_set: 10 }.freeze
+
+      # Game ends after 5 * 2 ORs
+      GAME_END_CHECK = { final_or_set: 5 }.freeze
 
       HOME_TOKEN_TIMING = :operating_round
 
@@ -24,6 +26,12 @@ module Engine
 
       def setup
         setup_company_price_50_to_150_percent
+      end
+
+      def new_operating_round(round_num = 1)
+        # Switch to phase 3 if OR1.2 is to start
+        @phase.next! if @turn == 1 && round_num == 2
+        super
       end
 
       def operating_round(round_num)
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_ms.rb
```diff
@@ -42,6 +74,32 @@ def operating_round(round_num)
...
+        when 4 then rust('3+', 30)
+        when 5 then rust('4+', 60)
+        end
+      end
+
+      private
+
+      def rust(train, salvage_value)
+        rusted_trains = []
+        trains.each do |t|
```
jenf, line 90: Would be tempted to do this as a rusted_trains = trains.select

Then do the rust later as a rusted_trains.each

perwestling: Fixed

lib/engine/game/g_18_ms.rb
```diff
@@ -12,18 +12,61 @@ class G18MS < Base
...
       def setup
-        setup_company_price_50_to_150_percent
+        @previously_floated_corporations = []
+      end
+
+      def purchasable_companies(entity = nil)
+        entity ||= current_entity
+        # Only companies owned by the president may be bought
+        companies = super.reject do |c|
+          entity.corporation? && entity.player == c.player
```
tobymao, line 43: do you want this, or do you want owned_by?(entity)

perwestling: Yes, but it was reversed check. Have corrected that now, and using owned_by?

lib/engine/game/g_18_ms.rb
```diff
@@ -42,6 +85,46 @@ def operating_round(round_num)
...
+        case @turn
+        when 3 then rust('2+', 20)
+        when 4 then rust('3+', 30)
+        when 5 then rust('4+', 60)
+        end
+      end
+
+      def sr_finished
+        # Check for any newly floated corporations
+        @newly_floated_corporations = []
```
tobymao, line 99: this should be initialized elsewhere

lib/engine/game/g_18_ms.rb
```diff
@@ -42,6 +85,46 @@ def operating_round(round_num)
...
+        @corporations.each do |c|
+          next if !c.floated? || @previously_floated_corporations.include?(c)
+
+          @newly_floated_corporations << c
+          @previously_floated_corporations << c
+        end
+      end
+
+      def tile_lays(entity)
+        return super unless @newly_floated_corporations&.include?(entity)
```
tobymao, line 109: then this won't be nil / undefined

perwestling:
Am doing the small step variant instead, with small pull requests. And the check point is done in a separate issue.

PR #1425, opened 2020-08-12
Fix exception when parsing tile with a town connected to just one edge

### LINKED ISSUE #1385: 18MS: Crashes when attempting to configure off board hexes
There are two errors in the 18MS map. Included maps below is from the physical game.

1. Atlanta on the east should be a group of D12, E13, F12, with track out on the board. When I try to add a hex E13, with a path, it crashes with: undefined method 'edge?' for nil. 
Atlanta
2. I3 on the map is a workaround to solve the city + town hex of I1. If I try to add a track on I3 to connect to I1 I get uncaught exception: Type Error, NilClass can't be coerced into Numeric. Stack trace for this include reservation.rb:3, town_rect.rb:64, base.rb etc
NewOrleans

roseundy:
Fixes #1385 - specifically part 2.

### DIFF
--- a/lib/engine/tile.rb
+++ b/lib/engine/tile.rb
@@ -199,7 +199,8 @@ module Engine
       @lawson ||=
         !!@junction ||
         (@cities.one? && @towns.empty?) ||
-        ((cities.empty? && towns.one?) && edges.size > 2)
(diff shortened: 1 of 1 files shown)


michaeljb:
Thanks!

PR #1443, opened 2020-08-14
Step-blocking privates (Ehime in 1889, Steamboat in 1846)

### LINKED ISSUE #1397: [1846] - Steamboat Company can be assigned any time in OR instead of during private income step, allowing multiple Corporations to benefit in a single OR
**Describe the bug**

When the Steamboat Company is owned by a player, it should only be assignable during the "private income step", or before any corporations operate.

Without this implemented, it will be possible for the Steamboat to be assigned to one corp, get it the bonus, and then be assigned to another corp so both get the bonus in one OR.

**Steps to reproduce:**

1. Start an 1846 game with Steamboat Company not removed from the game
2. Have the owner of the Steamboat Company assign it when it is their turn to operate a corporation

sample game JSON: https://gist.github.com/michaeljb/d170e9aa2a8820593f1064c9b8bf7813

Thanks @wcraigtrader for originally pointing out this timing issue in https://github.com/tobymao/18xx/issues/1316; for #1396 it was easier to solve the "multiple times in an OR" part of the issue, and put off the "private income step" part for later.

michaeljb:
Interrupt normal game flow when the Ehime is bought in (1889) so that the selling player must lay a tile or pass, and when the Steamboat Company must be assigned (1846) at the start of every OR that a player owns it. Fixes #1397 and gets rid of one "this is crap" code comment :wink: 

### TODO:

- [x] fix bug where can't assign Steamboat a corporation after undoing--`@selected_company` seems to get undefined in the view; weirdly enough, a hex can still be assigned
- [x] fix tests
- [x] add assets tests for views at Ehime and Steamboat steps
- [x] add engine tests for those steps
- [x] write migration for existing 1889 games; skipping Ehime's tile lay now requires an explicit pass (1846 migration would be much more involved, and it's beta so less of a big deal to just pin everything instead)
- [x] multiplayer game test of Ehime power (test corner case where the buying corp's president and the selling player are different)
- [x] when a corporation buys Steamboat, ~it should be unassigned its currently assigned hex~ keep the assignment and log it 
- [x] while owned by a corporation, the Steamboat can be re-assigned once per OR (https://boardgamegeek.com/thread/1770162/article/26036090#26036090)
- [x] test `migrate_game.rb` / `process_pass` error in 1846

### Screenshots:

Interrupt for Ehime's tile lay:

Screenshot from 2020-08-13 19-00-34

Steamboat is assigned to a minor and a hex; the currently assigned minor and hex are not clickable options to assign it now (preventing the user from seeing a GameError for assigning to the already-assigned value):

Screenshot from 2020-08-13 18-57-39

Steamboat's first "turn" at the beginning of OR 1.1 (note that D14/Toledo is clickable):

Screenshot from 2020-08-13 19-02-29


assets/app/view/game/help.rb
```diff
@@ -0,0 +1,17 @@
...
+  module Game
+    class Help < Snabberb::Component
+      needs :game
+
+      def render
+        help_text = @game.round.active_step.help
+        return '' if help_text.size.zero?
+
+        help_text_lines = Array(help_text)
+        h('div', help_text_lines.map { |l| h('div', l) })
```
tobymao, line 13: one liner

assets/app/view/game/round/operating.rb
```diff
@@ -30,7 +31,27 @@ def render
             left << h(BuyTrains)
           end
           left << h(IssueShares) if @current_actions.include?('buy_shares')
-          left << h(Corporation, corporation: entity)
+          if (entity.minor? || entity.corporation?) && @game.active_players.include?(entity.owner)
```
tobymao, line 34: if entity.operator?

lib/engine/game/g_1846.rb
```diff
@@ -326,6 +326,7 @@ def operating_round(round_num)
           Step::SpecialToken,
           Step::SpecialTrack,
           Step::G1846::BuyCompany,
+          Step::G1846::AssignSteamboat,
```
tobymao, line 329: lol wat

lib/engine/game/g_1889.rb
```diff
@@ -27,6 +28,26 @@ class G1889 < Base
...
+          Step::Token,
+          Step::Route,
+          Step::Dividend,
+          Step::BuyTrain,
+          [Step::BuyCompany, blocks: true],
+        ], round_num: round_num)
+      end
+
+      def active_players
+        current_entity == company_by_id('ER') ? [@round.company_seller] : super
```
tobymao, line 49: it's not the company player?

tobymao: why does this need an override

lib/engine/step/assign.rb
```diff
@@ -48,13 +48,14 @@ def process_assign(action)
         end
       end
 
-      def assignable_corporations
-        @game.corporations
+      def assignable_corporations(company = nil)
+        @game.corporations.select(&:floated?).reject { |c| c.assigned?(company&.id) }
```
tobymao, line 52: select { |c| c.floated && !c.assigned?(company&.id))

lib/engine/step/g_1846/assign_steamboat.rb
```diff
@@ -0,0 +1,110 @@
...
+          blocking_for_steamboat? || super
+        end
+
+        def blocks?
+          true
+        end
+
+        def pass!
+          super
+          @round.start_operating
```
tobymao, line 62: hm why ?

lib/engine/step/g_1846/buy_company.rb
```diff
@@ -10,6 +10,10 @@ module G1846
       class BuyCompany < BuyCompany
         include ReceivershipSkip
 
+        def assignable_corporations
```
tobymao, line 13: all methods assignable_corporations should have the entity arg

lib/engine/step/g_1889/special_track.rb
```diff
@@ -0,0 +1,39 @@
...
+
+        def process_pass(action)
+          ability = @company.abilities(:tile_lay, 'sold')
+          @company.remove_ability(ability)
+          @log << "#{action.entity.name} passes lay track"
+          pass!
+        end
+
+        def blocking_for_sold_company?
+          if (just_sold_company = @round.respond_to?(:just_sold_company) && @round.just_sold_company)
```
tobymao, line 27: combine the ifs

lib/engine/step/g_1889/special_track.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require_relative '../special_track'
+
+module Engine
+  module Step
+    module G1889
+      class SpecialTrack < SpecialTrack
```
tobymao, line 8: i wonder if this should just be generic in special track? does this need to be 1889 specific?

lib/engine/game/g_1889.rb
michaeljb: Not the company player, but the previous owner. Screenshot from 2020-08-18 21-42-51

Needs an override because normally if a company is active, the player who can take actions, and is alerted about their turn, is the owner of the company.

lib/engine/step/g_1889/special_track.rb
michaeljb: I think 1889 is the only game we have (at least so far) where a company's special track lay can forcibly interrupt the whole flow.

I guess with the C-P in Chesapeake, once one tile is laid, it forces you to lay the second, but after buying it in, you can use it at any point. 1889's Ehime is right when sold, no other time. I couldn't think of any other companies in any other games with that when-sold timing off the top of my head.

lib/engine/game/g_1889.rb
tobymao: like 99% of the time it's the same owner, why would you sell to someone else. sigh

lib/engine/step/g_1846/assign_steamboat.rb
michaeljb: I did some weird stuff to get the steamboat assignment to happen at the beginning of the OR, including changing `Round::Operating#setup` to not call `start_operating`, so I figured `AssignSteamboat` would need to call it. Don't remember if I actually tested it out without this line though.

tobymao: we definitely should not do it like this

michaeljb:
This most likely breaks most, if not all, 1846 games where Steamboat Company was not removed, and all 1889 games where Ehime Railroad does not lay a tile (an explicit pass is required for that now).

If we don't want to just pin the broken 1889 games, a migration should just need to add the pass action after Ehime is bought in, but I guess undo actions make that slightly more complicated.

tobymao:
maybe you can try to do a migration? @jenf made a script that could help you a lot

tobymao:
migrate_game.rb

assets/app/view/game/help.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module View
+  module Game
+    class Help < Snabberb::Component
+      needs :game
+
+      def render
+        help_text = @game.round.active_step&.help || ''
+        return '' if help_text.size.zero?
```
tobymao, line 10: help_text.empty?

michaeljb:
Was able to try out the migrate script with these steps:
1) `docker-compose exec rack bash`
2) `bundle exec rake irb`
3) `require_relative "migrate_game"`
4) `migrate_all()`

Found some things to tweak, and now it correctly migrates the broken 1889 games. Well, it correctly migrated the one broken test case I made up on my local machine, I can try something with a copy of the prod db or something later. Also still need to make sure 1846 games get pinned appropriately.

PR #1445, opened 2020-08-15
18TN: Bug fixes

perwestling:
Correct Nashville. Now starts as white hex. [This breaks existing games. Correcting Louisville will not break existing games.]

Correct show of tile upgrades in tile manifest for green tiles related Chattanooga and Nashville.
Brown P is also shown as an option for normal green city tiles.

Now corporations can only buy 1 company during OR 1.

Corrected handling of Civil War when only owning one train. (Withhold only option + Stock market price unchanged.)

Replaced spec file with a new 5 player game.


assets/app/view/game/tile_manifest.rb
```diff
@@ -12,8 +12,10 @@ def render_tile_selector(remaining, tile)
         return [] unless @tile_selector
         return [] if @tile_selector.role != :tile_page || @tile_selector.hex&.tile&.name != tile.name
 
-        upgrade_tiles = @game.all_potential_upgrades(@tile_selector.hex.tile).map do |t|
-          [t, remaining[t.name]&.any? ? nil : 'None Left']
+        upgrade_tiles = @game.all_potential_upgrades(@tile_selector.hex.tile).map do |t_info|
```
tobymao, line 15: i don't think we should have this

perwestling: I will make a simplified solution, which would remove the changes in the general code.

perwestling:
Still some bug in the implementation, so I will push another version soon.

perwestling:
Screenshot from 2020-08-16 12-37-55
Screenshot from 2020-08-16 12-38-33

Screenshot from 2020-08-16 12-40-33
Screenshot from 2020-08-16 12-40-55
Screenshot from 2020-08-16 15-01-03
Screenshot from 2020-08-16 15-01-37

assets/app/view/game/map.rb
```diff
@@ -76,7 +76,7 @@ def render
               end
 
               tiles = step.upgradeable_tiles(current_entity, @tile_selector.hex)
-              all_upgrades = @game.all_potential_upgrades(@tile_selector.hex.tile)
+              all_upgrades = @game.all_potential_upgrades(@tile_selector.hex.tile, @tile_selector.hex)
```
tobymao, line 79: is this really necessary? doesnn't the tile already have access to it's own hex?

perwestling: <<Face palm>> You are right. Pushed a commit that removed that parameter.

PR #1451, opened 2020-08-15
Handle empty depot smothly (fixes #1446)

### LINKED ISSUE #1446: Got “undefined method `price' for nil” when trying to buy the last 4D train on 18AL pinned to version d43f9f0c
This happened when I was trying to buy last 4D on 18AL pinned to version d43f9f0c https://www.18xx.games/game/5459?action=549

When I try to buy the last train for WRA, I get the error “ undefined method `price' for nil” I tried to search for a similar problem that was fixed, but could not find anything.

perwestling:
If the last train is bought the game does not try to use
the cheapest train in the depot.

And if no trains are in the depot then a corporation is not
forced to buy trains.

Included a test case for an 18AL game where all 9 permanent
trains are bought, and the trains distributed among the 6
corporations so that one corporation is left without. Game
runs to completion with L&N trainless.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -557,7 +557,7 @@ module Engine
       end
 
       def must_buy_train?(entity)
-        !entity.rusted_self && entity.trains.empty? &&
+        !entity.rusted_self && entity.trains.empty? && !depot.depot_trains.empty? &&
           (self.class::MUST_BUY_TRAIN == :always ||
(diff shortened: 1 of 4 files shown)


lib/engine/game/base.rb
```diff
@@ -557,7 +557,7 @@ def can_run_route?(entity)
       end
 
       def must_buy_train?(entity)
-        !entity.rusted_self && entity.trains.empty? &&
+        !entity.rusted_self && entity.trains.empty? && !depot.depot_trains.empty? &&
```
tobymao, line 560: depot_trains.any?

PR #1452, opened 2020-08-15
Show all money corp receives in preview, incl. divs (fixes #1367)

### LINKED ISSUE #1367: 'Payout' preview should show money paid to bank shares
The payout preview is great, but I think that the 'Treasury' value for 'payout' should show the treasury gain for dividends paid to bank shares.

dfannius:
Also I finally changed "Payout" to "Pay Out"; "payout" is a noun and the button names should be verbs.

### DIFF
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -32,6 +32,7 @@ module Engine
         revenue = @game.routes_revenue(routes)
         dividend_types.map do |type|
           payout = send(type, entity, revenue)
+          payout[:divs_to_company] = company_dividends(entity, payout[:per_share])
           [type, payout.merge(share_price_change(entity, revenue - payout[:company]))]
         end.to_h
       end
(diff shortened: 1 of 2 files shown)


dfannius:
Forgot to run rake, I'm rusty!

dfannius:
By the way, I was originally going to show the money that a corporation gets from dividends in a separate column, but we're already really low on horizontal space, so I just lumped it in together for the Treasury column. The information is published separately, though, so we could break it out in the display if we wanted.

lib/engine/step/dividend.rb
```diff
@@ -88,13 +89,31 @@ def payout(entity, revenue)
         { company: 0, per_share: payout_per_share(entity, revenue)[0] }
       end
 
+      def dividends_for_entity(entity, holder, per_share)
+        return 0 if (percent = holder.percent_of(entity)).zero?
```
tobymao, line 93: can you just do num_shares_of

dfannius: Sure, although looking at the code I see that `num_shares_of` just calls `percent_of` and divides by 10.

dfannius: Actually, when I changed `percent_of` to `num_shares_of` a ton of unit tests failed. I can look into why tomorrow.

dfannius: Never mind, I see what your point was.

lib/engine/step/dividend.rb
```diff
@@ -88,13 +89,29 @@ def payout(entity, revenue)
         { company: 0, per_share: payout_per_share(entity, revenue)[0] }
       end
 
+      def dividends_for_entity(entity, holder, per_share)
+        return 0 if (shares = holder.num_shares_of(entity)).zero?
```
tobymao, line 93: can't this just be holder.num_shares_of(entity) * per_share, why do you need to check for zero

dfannius: Sure, I was just changing the existing code as little as possible, but it used to have more of a reason to early-out.

Issue #1424, opened 2020-08-11
18TN -Louisville Hex


Bearcat89:
There should be a track segment in the Louisville hex that allows a route to bypass the L&N station there (both for purposes of train runs and track placement).  This should be similar to how the Altoona hex (PRR home hex) in 1830 is handled.

perwestling:
![image](https://user-images.githubusercontent.com/2631151/89955521-802b8500-dc33-11ea-9859-c2046ba93825.png)

perwestling:
Has been corrected and deployed.

![Louisville](https://user-images.githubusercontent.com/2631151/90681931-7b368900-e264-11ea-8521-d007de4ee9e3.PNG)

PR #1464, opened 2020-08-17
Minor rewrite of track rendering to be path-specific

roseundy:
Also rewrite of code to handle track connected to cities. Impetus was to fix Louisville in 18TN. This required changing the track rendering code to be path-specific rather than essentially categorizing a tile as a "city" tile, for example.

Fixed some broken tile specifications in config/tile.rb along the way of testing my code.

Still to do:
- Adding "loc:<edge>" specifiers to cities/towns in the tile definitions and acting on them
- Supporting city-city, town-town and town-city paths in the tile definition


roseundy:
Louisville new renders like this:
Louisville_start

Old multi-city tile rendering:
457-459old

New rendering:
457-459new

assets/app/view/game/part/track.rb
```diff
@@ -29,58 +29,37 @@ def render
           # Array<Array<Path>>
           @routes_paths = @routes.map { |route| route.paths_for(@tile.paths) }
 
-          if @tile.offboards.any?
-            @tile.paths.select(&:offboard).map do |path|
+          tracks = []
+          @tile.paths.select { |path| path.edges.size == 2 }
+          .map { |path| [path, index_for(path)] }
+          .sort_by { |_, index| index || -1 }
+          .map do |path, index|
```
tobymao, line 36: map returns an array, you shouldn't be doing tracks << x

tracks = x.map...

assets/app/view/game/part/track.rb
```diff
@@ -29,58 +29,37 @@ def render
...
+            tracks << h(TrackCurvilinearPath, region_use: @region_use, path: path,
+                                              color: value_for_index(index, :color),
+                                              width: value_for(path, :width),
+                                              dash: value_for(path, :dash),)
+          end
+
+          tracks << @tile.paths.map do |path|
```
tobymao, line 43: this doesn't make sense you're adding an array of paths to tracks?

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,148 @@
...
+        def control_location
+          # calculate the position of the quadratic control point for a bezier curve to
+          # be drawn between two points on a tile. If a point is on an edge, the control
+          # point will be chosen to make the curve perpendicular to the tile edge.
+          #
+          loc0 = begin_location
+          loc1 = end_location
+
+          # Currently this only handles the case where the start poing is on an edge
+          edge_perp_angle = [90, 30, -30, -90, -150, 150][@edge]
```
tobymao, line 58: make this first array a constant

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,148 @@
...
+          }
+        end
+
+        def need_bezier
+          # probably not general enough
+          (end_location != { x: 0, y: 0 }) && ((@ct_edge != @edge) && ((@ct_edge - @edge).abs != 3))
+        end
+
+        def preferred_render_locations
+          regions = {
```
tobymao, line 92: constantize

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,148 @@
...
+          internal_angle = edge_perp_angle / 180 * Math::PI - angle
+          offset = (distance / 2 * Math.tan(internal_angle)).abs
+
+          {
+            x: mid[:x] + Math.cos(normal_angle) * offset,
+            y: mid[:y] - Math.sin(normal_angle) * offset,
+          }
+        end
+
+        def need_bezier
```
tobymao, line 86: def bezier? or need_bezier?

assets/app/view/game/part/track.rb
roseundy: Gotcha.

assets/app/view/game/part/track_node_path.rb
roseundy: OK

assets/app/view/game/part/track_node_path.rb
roseundy: OK

assets/app/view/game/part/track_node_path.rb
roseundy: I'll switch to need_bezier?

assets/app/view/game/part/track.rb
roseundy: It's an array of h() objects.

roseundy:
Fixes #1424

assets/app/view/game/part/track.rb
michaeljb: `tracks` is also an array of `h()` objects, probably better to `.concat` them, or `.flatten` afterward (though it does all get flattened later, at https://github.com/tobymao/18xx/blob/master/assets/app/view/game/tile.rb#L69)

assets/app/view/game/part/track.rb
```diff
@@ -29,58 +29,36 @@ def render
...
-
-        private
-
-        def render_track_for_curvilinear_city
-          @tile.cities.flat_map do |city|
-            exits = city.exits
-
-            city.paths.map do |path|
-              h(TrackCurvilinearHalfPath, exits: exits, path: path, region_use: @region_use,
+            elsif path.junction || path.city || (path.town && path.town.exits.size != 2)
```
michaeljb, line 47: Why exclude towns with 2 exits here? Seems like it would be better for those to use `TrackNodePath` and get rid of `TrackCurvilinearHalfPath`. I think that would fix #264

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,150 @@
...
+            full_distance = 50
+            full_distance -= 15 if @tile.borders.any? { |border| border.edge == @ct_edge }
+          end
+          {
+            x: -Math.sin((@ct_edge * 60) / 180 * Math::PI) * full_distance,
+            y: Math.cos((@ct_edge * 60) / 180 * Math::PI) * full_distance,
+          }
+        end
+
+        def control_location
```
michaeljb, line 60: I'm wondering if it would be possible to memoize `start_location`, `end_location`, and `control_location` somewhere to save some recalculations. I added a few quick debug lines, and `control_location` had a good number of repeated calls where `begin_location`, `end_location`, and `@edge` all had the same values.

Haven't thought about it too much, they'd probably need to be pulled out of `View::Game::Part` and put higher up in the module hierarchy.

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,150 @@
...
+          ]
+        end
+
+        def render_part
+          rotation = 60 * @edge
+
+          props = {
+            attrs: {
+              d: "M #{begin_location[:x].round(2)} #{begin_location[:y].round(2)} "\
+                 "L #{end_location[:x].round(2)} #{end_location[:y].round(2)}",
```
michaeljb, line 118: Would there be any disadvantage to making `begin_location` and `end_location` round the result before they return it? That would make this part (and below) a little cleaner. That would cost `control_location` a little precision, but I doubt there would be much of a difference unless you're zoomed way in--am I wrong about that?

assets/app/view/game/part/track_node_path.rb
michaeljb: @tobymao you have any thoughts on that?

assets/app/view/game/part/track.rb
roseundy: I may eventually do that, but it would require changing town rendering to make sure they land on top of the bezier curves.

assets/app/view/game/part/track_node_path.rb
roseundy: I wasn't sure how to do this. Perhaps storing in the path object?

assets/app/view/game/part/track_node_path.rb
roseundy: No, good point.

assets/app/view/game/part/track.rb
tobymao: yea, it should be concat

assets/app/view/game/part/track_node_path.rb
tobymao: begin_location could definitely be a constant of edge => map

end_location looks more dynamic, you can memoize doing 

def end_location
  @end_location ||= begin 
...
end

you can grep for other examples

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,150 @@
...
+          {
+            x: -Math.sin((@edge * 60) / 180 * Math::PI) * 87,
+            y: Math.cos((@edge * 60) / 180 * Math::PI) * 87,
+          }
+        end
+
+        def end_location
+          return { x: 0, y: 0 } if @junction
+
+          if (@tile.towns.size + @tile.cities.size) < 2
```
tobymao, line 47: could this be @tile.nodes.size < 2 ? nodes includes offboards, cities, junctions, and towns

assets/app/view/game/part/track_node_path.rb
tobymao: it would be easier to set all these up in the setup method as well

tobymao: load_from_tile

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,150 @@
...
+
+        def control_location
+          # calculate the position of the quadratic control point for a bezier curve to
+          # be drawn between two points on a tile. If a point is on an edge, the control
+          # point will be chosen to make the curve perpendicular to the tile edge.
+          #
+          loc0 = begin_location
+          loc1 = end_location
+
+          # Currently this only handles the case where the start poing is on an edge
```
tobymao, line 68: start poing?

assets/app/view/game/part/track_node_path.rb
roseundy: Yes, won't be called with offboards and junctions are handled above.
Might be better as @tile.stops < 2

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,150 @@
...
+          edge_perp_angle = EDGE_PERP_ANGLES[@edge]
+
+          distance = Math.sqrt((loc0[:x] - loc1[:x])**2 + (loc0[:y] - loc1[:y])**2)
+          mid = { x: (loc0[:x] + loc1[:x]) / 2, y: (loc1[:y] + loc0[:y]) / 2 }
+          angle = Math.atan2(loc0[:y] - loc1[:y], loc1[:x] - loc0[:x])
+
+          normal_angle = Math.atan2(loc1[:x] - loc0[:x], loc1[:y] - loc0[:y])
+
+          # determine what side of curve control point should be on
+          # -> want to always rc into the center of tile
```
tobymao, line 78: rc?

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,150 @@
...
+          center_angle = Math.atan2(loc0[:y], -loc0[:x])
+          if (angle >= 0) && (center_angle < angle) && ((angle - Math::PI) < center_angle)
```
tobymao, line 80: these are all the same result?

it could be

normal_angle -= Math::PI if () 
  || (...) 
  || (...)

assets/app/view/game/part/track_node_path.rb
tobymao: cool, stops makes sense

assets/app/view/game/part/track_node_path.rb
roseundy: So my longer term plan is for begin_location to resemble end_location more since eventually it will have to handle paths that don't go to edges (city/town to city/town paths on a hex). I'll change to have these cache in addition to control_location.

assets/app/view/game/part/track_node_path.rb
roseundy: start point. I'll fix.

assets/app/view/game/part/track_node_path.rb
roseundy: arc. I'll fix.

assets/app/view/game/part/track_node_path.rb
roseundy: OK.

assets/app/view/game/part/track.rb
roseundy: OK

assets/app/view/game/part/track.rb
roseundy: Is it OK if I hold off for a bit on this specific change?

assets/app/view/game/part/track_node_path.rb
roseundy: And now I know why it was written this way. @tile.stops and @tile.nodes doesn't find cities/towns with no paths connected to them :-/

assets/app/view/game/part/track.rb
michaeljb: No problem!

assets/app/view/game/part/track_node_path.rb
```diff
@@ -0,0 +1,154 @@
...
+        EDGE_REGIONS = {
+          0 => TRACK_TO_EDGE_0,
+          1 => TRACK_TO_EDGE_1,
+          2 => TRACK_TO_EDGE_2,
+          3 => TRACK_TO_EDGE_3,
+          4 => TRACK_TO_EDGE_4,
+          5 => TRACK_TO_EDGE_5,
+        }.freeze
+
+        def self.lookup_pos_x(edge, distance)
```
tobymao, line 26: these don't need to be class methods since you're using @@

it should just be lookup_pos, then the next calls won't need to call TrackNodePath

PR #1472, opened 2020-08-18
18GA: Add some missing features

perwestling:
Has some problems so do not merge this yet.

### DIFF
--- a/lib/engine/game/g_18_ga.rb
+++ b/lib/engine/game/g_18_ga.rb
@@ -18,6 +18,8 @@ module Engine
       GAME_END_CHECK = { bankrupt: :immediate, stock_market: :current_or, bank: :current_or }.freeze
 
       STATUS_TEXT = Base::STATUS_TEXT.merge(
+        'can_buy_companies_from_other_players' => ['Interplayer Company Buy', 'Companies can be bought between players']
+      ).merge(
         Step::SingleDepotTrainBuyBeforePhase4::STATUS_TEXT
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_ga.rb
```diff
@@ -27,17 +29,17 @@ def setup
         setup_company_price_50_to_150_percent
 
         # Place neutral tokens in the off board cities
-        neutral = Corporation.new(
+        @neutral = Corporation.new(
```
tobymao, line 32: why make these instance variables? you never use them

lib/engine/train.rb
```diff
@@ -10,7 +10,7 @@ class Train
     attr_accessor :obsolete, :operated, :events
     attr_reader :available_on, :name, :distance, :discount, :obsolete_on,
                 :rusts_on, :rusted, :sym, :variant, :variants
-    attr_writer :buyable
+    attr_writer :buyable, :index
```
tobymao, line 13: what's the point of index?

perwestling: When adding the new 2 train I do that by cloning another 2 train, and then I set index so it wont collide with another 2 train.

lib/engine/game/g_18_ga.rb
perwestling: Removed

perwestling:
Have split this into smaller PRs.

PR #1482, opened 2020-08-19
Improve handling of special label cases in 18TN and 18AL

perwestling:
18TN has now Memphis corrected - it should only allow upgrades to P tile in brown.

Removed adding of labels in 18TN as this would cause same problem as was fixed in 18AL.

Refactored upgrades_to? and all_potential_upgrades in 18TN to make the special cases (Memphis, Chattanooga., Nashville when upgrading to brown, and Nashville when upgrading to green) clearer (I hope).

Also used similar refactoring in 18AL to do the same pattern there.

If this is approved a similar refactoring will be done in 18GA.


lib/engine/game/base.rb
```diff
@@ -1118,6 +1118,11 @@ def all_potential_upgrades(tile)
           .select { |t| upgrades_to?(tile, t) }
           .reject(&:blocks_lay)
       end
+
+      # In all_potential_upgrades context
+      def from_tile_manifest?(tile)
+        !tile.hex || tile.hex.name == 'A1'
```
tobymao, line 1124: this is not good

perwestling: Have made a new attempt

tobymao: what i didn't like was the hex.name == 'A1' why is that needed is there another way?

lib/engine/game/g_18_tn.rb
```diff
@@ -107,24 +112,33 @@ def lnr
...
+        super
       end
 
       def all_potential_upgrades(tile)
-        upgrades = super
+        suggested_upgrades = super
+
+        # Upgrade to green in Nashville (F11) is only allowed with N tile (#TN2)
+        # Tile manifest for yellow standard cities should show N tile (TN1) as an option
+        special_upgrade, updated_upgrades =
```
tobymao, line 131: this pattern is strange, why are you returning two arrays here

lib/engine/game/label_breaking_upgrade.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module LabelBreakingUpgrade
+  # Check if a special upgrade should be done
+  def special_upgrade?(suggested_upgrades, tile_from, from_tiles_of_interest, hexes_that_are_special, special_tile)
```
tobymao, line 5: maybe lets chat about this design sometime

lib/engine/game/g_18_tn.rb
perwestling: That's true - I can just return nil / updated_upgrades

PR #1484, opened 2020-08-19
18GA: Implement Ocilla Southern RR company

perwestling:
Buying corporation receives a free non-sellable 2 train if it has
room for it and the 2 trains hasn't rusted.

### DIFF
--- a/lib/engine/step/g_18_ga/buy_company.rb
+++ b/lib/engine/step/g_18_ga/buy_company.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require_relative '../buy_company'
+
+module Engine
+  module Step
+    module G18GA
+      class BuyCompany < BuyCompany
+        def process_buy_company(action)
+          super
+
+          return unless action.company.sym == 'OSR'
+
+          owner = action.company.owner
+          return if owner.player? || owner.trains.size == @game.phase.train_limit || @game.phase.available?('4')
+
+          free_two_train = @game.trains.select { |t| t.name == '2' }.last.dup
+          free_two_train.buyable = false
+          free_two_train.index = 5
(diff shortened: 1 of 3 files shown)


lib/engine/step/g_18_ga/buy_company.rb
```diff
@@ -0,0 +1,26 @@
...
+      class BuyCompany < BuyCompany
+        def process_buy_company(action)
+          super
+
+          return unless action.company.sym == 'OSR'
+
+          owner = action.company.owner
+          return if owner.player? || owner.trains.size == @game.phase.train_limit || @game.phase.available?('4')
+
+          free_two_train = @game.trains.select { |t| t.name == '2' }.last.dup
```
tobymao, line 17: this should be trains.find  { |t| t.name == '2' }.dup, but what happens if this train is owned by someone else already?

tobymao: why don't you just make a new train?

lib/engine/train.rb
```diff
@@ -10,7 +10,7 @@ class Train
     attr_accessor :obsolete, :operated, :events
     attr_reader :available_on, :name, :distance, :discount, :obsolete_on,
                 :rusts_on, :rusted, :sym, :variant, :variants
-    attr_writer :buyable
+    attr_writer :buyable, :index
```
tobymao, line 13: yea i think you should get rid of this and just create the train in buy_company

lib/engine/step/g_18_ga/buy_company.rb
perwestling: Ok, making a new train

lib/engine/train.rb
perwestling: ok

PR #1488, opened 2020-08-21
Add location to tile defintions

roseundy:
Add "loc:<edge>" specifier to tile definitions.
Add support for non-integer edges (0.5, 1,5, etc).
Updated a handful of tile definitions in config/tile.rb
Updated Louisville tile definition to get this:
Louisville_newnew


roseundy:
Note that I haven't fully implemented this yet for towns. I'm holding off on delving into town rendering until I work on city-to-town paths.

PR #1503, opened 2020-08-23
18MS: Allow extra yellow tile lay first OR

perwestling:
Every corporation can lay 2 yellow (or one upgrade) in its first OR.
Remaining ORs it is 1 yellow or one upgrade.


lib/engine/game/base.rb
```diff
@@ -943,6 +945,7 @@ def next_round!
           when Round::Stock
             @operating_rounds = @phase.operating_rounds
             reorder_players
+            sr_finished
```
jenf, line 948: This doesn't seem to be used, otherwise it's fine.

perwestling: Ah, solution evolved over time....

Have removed that method in the new version. Had to rebase it so it was a push force.

PR #1506, opened 2020-08-23
18MS: Initial draft

perwestling:
In the initial draft each player (in reverse order) buy one company
each for face value. The remaining companies, if any, is discarded.


lib/engine/step/g_18_ms/draft_distribution.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
```
jenf, line 1: Can any of the 1846 code be shared?

perwestling: I do not know if that would be good? 
The 18MS draft is a straightforward one, where each player buy one private - they cannot pass. It is open. Variants of this exists in 1893 and 1876.
Possibly 1848 and 18CZ could reuse this type of draft.

So in the same way as we have several different auction maybe we need to have several different drafts?

assets/app/view/game_page.rb
```diff
@@ -251,6 +251,8 @@ def render_action
         end
       when Engine::Round::Operating
         h(Game::Round::Operating, game: @game)
+      when Engine::Round::G18MS::Draft
```
tobymao, line 254: can we at least do Round::Draft share?

tobymao: you should be able to just when Draft

perwestling: Changed to Engine::Round::Draft

lib/engine/step/g_18_ms/draft_distribution.rb
perwestling: Draft from 1846 is now shared, and moved up one level.

perwestling:
@tobymao Now it has been updated and rebased.

PR #1508, opened 2020-08-23
18MS: Update private companies handling

perwestling:
Make it possible for companies to buy P1/P2 during OR1.

AGS (P1) and BS (P2) can be bought by companies during OR1, but
only at face value. Otherwise buy of companies is as usual, can
be done for 50%-150% of face value during green phase. In 18MS
brown phase are triggered when 1st 6 train is bought.

Only private companies owned by the president may be bought


tobymao:
can you fix the merge conflict

PR #1509, opened 2020-08-23
18MS: Max ownership in a corporation is 70%

perwestling:


jenf (commented):
Looks good, but Toby should probably glance at it.

PR #1534, opened 2020-08-27
Add tokens for assign_hex and assign_corp and add tokens for 1846 and 18AL

### LINKED ISSUE #1323: [1846] Indicate on the map where the SBC and MPC tokens are placed
When playing the physical board game, the Steamboat Company (SBC) and Meat Packing Company (MPC) have tokens that are placed on the map to indicate which cities get their respective bonuses. There is no analogous icon on the map for those companies.

My recommendation would be to put a halo around the appropriate icon for the assigned city on the map; i would recommend the halo be blue, since that would have an adequate contrast to yellow, green, brown, and red.

jenf:
fixes #1323 

Screenshot from 2020-08-27 20-43-53

### DIFF
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -52,6 +52,11 @@ module Engine
       TILE_COST = 20
       EVENTS_TEXT = Base::EVENTS_TEXT.merge('remove_tokens' => ['Remove Tokens', 'Remove private company tokens']).freeze
 
+      ASSIGNMENT_TOKENS = {
+        "MPC": '/icons/1846/mpc_token.svg',
+        "SC": '/icons/1846/sc_token.svg',
+      }.freeze
+
       # Two tiles can be laid, only one upgrade
       TILE_LAYS = [{ lay: true, upgrade: true }, { lay: true, upgrade: :not_if_upgraded }].freeze
 
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -53,8 +53,8 @@ module Engine
       EVENTS_TEXT = Base::EVENTS_TEXT.merge('remove_tokens' => ['Remove Tokens', 'Remove private company tokens']).freeze
 
       ASSIGNMENT_TOKENS = {
-        "MPC": '/icons/1846/mpc_token.svg',
-        "SC": '/icons/1846/sc_token.svg',
+        'MPC' => '/icons/1846/mpc_token.svg',
+        'SC' => '/icons/1846/sc_token.svg',
       }.freeze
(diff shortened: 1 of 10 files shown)


perwestling:
I tried it for 18AL and it worked fine. No token (like MPC) did appear on the charter but I do not know if it need to?

Anyway, one improvement could be done in 18AL. Before this change the Mining symbol remained in the selected hex. As the token is placed it is no longer needed.
This can be fixed by changing line 141 in g_18_al.rb: Remove exclude != hex.name.
Also the exclude: parameter can be removed in this method and in the call in assign.rb.

Then there is a bug. In 18AL all assigned tokens are removed when 6 train is bought.
This is handled in event_remove_tokens! in g_18_al.rb. So I think token can be removed in this method.
The remove_mining_incons call from here is no longer needed, in case the mining icon has already been removed.

lib/engine/game/g_1846.rb
```diff
@@ -52,6 +52,11 @@ class G1846 < Base
       TILE_COST = 20
       EVENTS_TEXT = Base::EVENTS_TEXT.merge('remove_tokens' => ['Remove Tokens', 'Remove private company tokens']).freeze
 
+      ASSIGNMENT_TOKENS = {
+        "MPC": '/icons/1846/mpc_token.svg',
```
tobymao, line 56: are these strings or symbols? if symbols do MPC: ... if strings "MPC" => 'x'

jenf:
> I tried it for 18AL and it worked fine. No token (like MPC) did appear on the charter but I do not know if it need to?
> 
> Anyway, one improvement could be done in 18AL. Before this change the Mining symbol remained in the selected hex. As the token is placed it is no longer needed.
> This can be fixed by changing line 141 in g_18_al.rb: Remove exclude != hex.name.
> Also the exclude: parameter can be removed in this method and in the call in assign.rb.
> 
> Then there is a bug. In 18AL all assigned tokens are removed when 6 train is bought.
> This is handled in event_remove_tokens! in g_18_al.rb. So I think token can be removed in this method.
> The remove_mining_incons call from here is no longer needed, in case the mining icon has already been removed.

I'm afraid I'm not quite clear what needs fixing, so it's probably best if I let you do it on the branch.

lib/engine/game/g_18_al.rb
```diff
@@ -35,6 +35,9 @@ class G18AL < Base
 
       STANDARD_YELLOW_CITY_TILES = %w[5 6 57].freeze
 
+      ASSIGNMENT_TOKENS = {
```
jenf, line 38: This needs modifications to 18AL to remove the assignment

PR #1535, opened 2020-08-27
Support city and town on same tile, support intra-tile paths

### LINKED ISSUE #853: Show a city and town on the same hex
This is going to come up a fair amount for upcoming games including 18MEX and 18NewEngland, others I'm sure, 1822, 1822CA if we ever get there. Currently the app can't parse the json if a hex includes both a city and a town like this.
image

It would be nice if the system also handled this case, but probably not a big deal and could be hacked around with an icon.
image

cc @michaeljb

roseundy:
Several rendering changes:

1. Add support for city and town on same tile
2. Add support for intra-tile paths (i.e. path between city/town and city/town)
3. Single class now supports pure track, track to junctions, track to cities, track to towns and track between cities and towns
4. Small improvements for label positions
5. TownRect now used for towns with a single exit
6. All track rendering now uses pure arcs instead of bezier curves or a combination of arcs and line segments

NOTE: corresponding routing support for intra-tile paths have not yet been implemented.

Technically this fixes #853


roseundy:
Examples of what is now supported:
I3_test
603_new
603a

roseundy:
With latest commit, can now do this:
18MEX_470
MexicoCity
Tampico
Mazatlan

roseundy:
Note the above 18MEX map/tile changes haven't been checked in, only the code that supports them.

assets/app/view/game/part/city.rb
```diff
@@ -127,6 +112,86 @@ class City < Base
...
         }.freeze
 
+        ANGLE_UPPER_RIGHT = -60
+        ANGLE_LOWER_RIGHT = 10
+        ANGLE_LOWER_LEFT = 170
+        ANGLE_UPPER_LEFT = -120
+
+        REVENUE_LOCATIONS_BY_EDGE = {
+          0 => [
+                { regions: [19], angle: ANGLE_LOWER_LEFT },
```
tobymao, line 122: your indenting is not great here, do you have an auto indenter?

assets/app/view/game/part/city.rb
```diff
@@ -228,6 +302,14 @@ def render_part
           h(:g, props, children)
         end
 
+        # look for a location for revenue based on edge and region_use
+        def find_revenue_location(edge)
+          revenue_location = REVENUE_LOCATIONS_BY_EDGE[edge].min_by do |cand|
```
tobymao, line 307: what's cand?

tobymao: i think you can just do this inline on line 335

revenue_location = REVENUE_LOCATIONS_BY_EDGE[edge].min_by { |loc| combined_cost(loc[:regions]) }
regions = revenue_location[...]
rotation = revenue_location[...]

assets/app/view/game/part/city.rb
```diff
@@ -241,20 +323,33 @@ def render_revenue
...
-              regions, negative_displacement = OO_REVENUE_REGIONS[@edge]
-              displacement *= -1 if negative_displacement
+          elsif @edge && @city.slots == 1
+            regions, rotation = find_revenue_location(@edge)
+          elsif @edge
+            regions, negative_displacement = OO_REVENUE_REGIONS[@edge]
+            displacement *= -1 if negative_displacement
+          else
+            # pick an edge where there isn't another stop
+            other_stops = (@tile.cities + @tile.towns).reject { |s| s == @city }
```
tobymao, line 341: this whole else statement seems inefficient

edges = SOME_CONSTANT - (@tiles.cities + @tile.towns).flat_map do |stop|
    next [] if stop == @city || !(edge = @tile.preferred_city_town_edges[stop])
    [edge, (edge - 1) % 6]
end

revenue_edge = edges[0] || 0
rotation = 60 * revenue_edge + 120
regions = CENTER_REVENUE_REGIONS[revenue_edge]

assets/app/view/game/part/town_rect.rb
```diff
@@ -35,63 +35,113 @@ class TownRect < Base
...
-              edge_b = (@town.exits - [edge_a]).first
-              edges = [edge_a, edge_b]
-              edges[edges.index(edges.min)] += 6 if (edge_a - edge_b).abs > 3
-              edges
-            end
-        end
-
-        def min_edge
-          @min_edge ||= normalized_edges.min
+        def self.normalized_edges(edge_a, exits)
```
tobymao, line 116: you shouldn't remove the caching here. can we make this module and have class level caching? it seems like this would be a candidate for that let me know if you need help with that

assets/app/view/game/part/town_rect.rb
```diff
@@ -35,63 +35,113 @@ class TownRect < Base
...
+          if edge_a && exits.size == 2
+            edge_b = (exits - [edge_a]).first
```
tobymao, line 118: edge_b = exits.first { |e| e != edge_a } maybe since this doesn't do 2 allocations

assets/app/view/game/part/town_rect.rb
```diff
@@ -35,63 +35,113 @@ class TownRect < Base
...
+          else
+            [edge_a, nil, nil]
+          end
         end
 
         # Returns a symbol of :straight, :gentle or :sharp based on the absolute
         # difference of the normalized edges
-        def track_type
-          @track_type ||=
+        def self.track_type(edges)
```
tobymao, line 129: as with normalize edges maybe thes method belongs in a separate module

assets/app/view/game/part/town_rect.rb
```diff
@@ -35,63 +35,113 @@ class TownRect < Base
...
+          @@track_type ||= {}
+          @@track_type[edges] ||=
             begin
-              edge_a, edge_b = normalized_edges
+              edge_a, edge_b, = edges
```
tobymao, line 133: why is there an extra comma here

tobymao: why is there a nil in the array?

tobymao: could this also be [:sharp, :gentle, :straight[... - 1] ? also maybe make that into a constant

assets/app/view/game/part/town_rect.rb
```diff
@@ -35,63 +35,113 @@ class TownRect < Base
...
-        def min_edge
-          @min_edge ||= normalized_edges.min
+        def self.normalized_edges(edge_a, exits)
+          if edge_a && exits.size == 2
+            edge_b = (exits - [edge_a]).first
+            edges = [edge_a, edge_b]
+            edges[edges.index(edges.min)] += 6 if (edge_a - edge_b).abs > 3
+            edges.append(edges.min)
+          else
+            [edge_a, nil, nil]
```
tobymao, line 123: ok, this is kind of confusing coding, storing [edge_a, edge_b, min_edge] and then doing edges.last is not very clear, i don't think we should do this pattern

tobymao: can you explain to me what the point of normalize edges is, i'm not actually sure what it does

assets/app/view/game/part/town_rect.rb
```diff
@@ -144,27 +196,43 @@ def rotation_angles
                     end
 
             [(edge_a * 60) + delta]
+          elsif edge_a
+            [edge_a * 60]
+          else
+            # This town is in the center. Find the orientation of the town
+            # by looking at the first path connected to it
+            path = town.paths.first
+            if !path.edges.empty?
```
tobymao, line 205: if path.edges.any?

lib/engine/part/base.rb
```diff
@@ -93,6 +93,10 @@ def blocks_lay?
         false
       end
 
+      def rect?
```
tobymao, line 96: does this need to be in the base? should it only be on nodes?

assets/app/view/game/part/track_node_path.rb
```diff
@@ -160,96 +162,187 @@ def edge_y_pos(edge, distance)
...
+
+            @begin_edge = @path.edges.first.num
+            @begin_x = edge_x_pos(@begin_edge, 87)
+            @begin_y = edge_y_pos(@begin_edge, 87)
+
+            @end_edge = @path.edges.last.num
+            @end_x = edge_x_pos(@end_edge, 87)
+            @end_y = edge_y_pos(@end_edge, 87)
+          elsif @num_exits == 1
+            @begin_edge = @path.edges.first.num
```
tobymao, line 254: @path.exits[0]

assets/app/view/game/part/track_node_path.rb
```diff
@@ -160,96 +162,187 @@ def edge_y_pos(edge, distance)
...
+          # these are the only possibilities this Class will handle:
+          # 1. exit - exit
+          # 2. exit - junction
+          # 3. exit - city/town
+          # 4. city/town - city/town
+
+          if @num_exits > 1
+            # exit - exit
+
+            @begin_edge = @path.edges.first.num
```
tobymao, line 246: path.exits[0]

assets/app/view/game/tile.rb
```diff
@@ -25,13 +25,16 @@ def render_tile_part(part_class, **kwargs)
...
+        return true if revenue.any? { |r| !r.is_a?(Numeric) }
+
         return false if revenue.empty?
 
         return false if revenue.first.is_a?(Numeric) && (@tile.cities + @tile.towns).one?
 
         return false if revenue.uniq.size > 1
 
-        return false if @tile.cities.sum(&:slots) < 3 && @tile.stops.size == 2
+        return false if @tile.cities.sum(&:slots) < 3 && (@tile.cities + @tile.towns).size == 2
```
tobymao, line 37: we call cities + towns . size a lot, maybe make a helper method on tiles

lib/engine/part/path.rb
```diff
@@ -5,12 +5,16 @@
 module Engine
   module Part
     class Path < Base
-      attr_reader :a, :b, :branch, :city, :edges, :junction, :node, :offboard, :stop, :town, :terminal
+      attr_reader :a, :b, :branch, :city, :edges, :junction, :node, :offboard,
```
tobymao, line 8: we should not do this, if paths have multiple X, there's no reason to store just one

you should replace all old things with plurals, stop and stops is wrong

so let's make everything

branches, cities, junctions nodes etc

lib/engine/tile.rb
```diff
@@ -336,22 +337,25 @@ def compute_city_town_edges
...
+      exitless_cts.select do |xct|
+        ct_edges[xct] = compute_loc(xct.loc) if xct.loc
       end
 
       ct_edges
     end
 
     def revenue_to_render
-      @revenue_to_render ||= stops.map(&:revenue_to_render)
+      @revenue_to_render ||= (@cities + @towns + @offboards).map(&:revenue_to_render)
```
tobymao, line 358: why not stops?

assets/app/view/game/part/city.rb
roseundy: Yeah, this was originally much more complicated than it needed to be. I simplified it down and should have just skipped the added method.

assets/app/view/game/part/city.rb
roseundy: No. I'm embarrassed to admit I just use vim. I'll see if there is better Ruby support I can add.

assets/app/view/game/part/city.rb
roseundy: OK, I'll recode.

assets/app/view/game/part/town_rect.rb
roseundy: I pulled out caching because of the re-use. Instead I just save the result where it is needed rather than calling normalized_edges over and over again.

tobymao: it will still get called for every town_rect and everything that calls these methods which will be thousands of  times

assets/app/view/game/part/city.rb
tobymao: vim is all i use

assets/app/view/game/part/town_rect.rb
roseundy: I didn't actually change this part of normalized_exits. I can change if you like.

assets/app/view/game/part/town_rect.rb
roseundy: I would need help (or at least more explanation for what you are looking for).

assets/app/view/game/part/town_rect.rb
roseundy: To your first point, this was to avoid having to calculate and cache min_edge over and over again in the original code.

normalized_edges was a pre-existing method to generate pairs of town exits that obeyed constraints to make calculating track_type, track_direction and ration_angles easier. See the comment directly above it.

I was trying to avoid ripping up the TownRect class even more than I did. I just wanted access to the location calculations. Since normalized_edges was called throughout that code I had to parameterize it.

assets/app/view/game/part/town_rect.rb
roseundy: Sure. Interesting that rubocop missed that.

lib/engine/part/base.rb
roseundy: Funny. I originally put it in Node and then decided to put it into base to match the other derived class overrides.

assets/app/view/game/part/track_node_path.rb
roseundy: OK

assets/app/view/game/part/town_rect.rb
tobymao: sure

assets/app/view/game/part/track_node_path.rb
roseundy: OK

assets/app/view/game/part/town_rect.rb
tobymao: ah i see makes sense, but the thing is instead of doing edges.min, you're now doing edges.last which is much more confusing, i think we can add caching back in

assets/app/view/game/tile.rb
roseundy: Yeah. I originally thought @tile.stops would work, but it only contains cities and town that are connected.

assets/app/view/game/part/town_rect.rb
tobymao: sure, happy to discuss in slack

lib/engine/part/path.rb
roseundy: Sorry, I was trying to be conservative and avoid touching more files. I was going to clean this up when I visited routing. I'll fix it up now.

assets/app/view/game/tile.rb
tobymao: well that should be fixed then, that sounds like a bug

lib/engine/part/path.rb
tobymao: i'd prefer more aggressive clean up

lib/engine/tile.rb
roseundy: Only to handle the special case of pathless cities/towns. One could make a pretty decent argument that pathless revenue_centers shouldn't display revenue. Maybe that would be the better fix.

tobymao: i think we should fix the stops not working for pathless, that sounds wrong

assets/app/view/game/part/label.rb
```diff
@@ -83,13 +83,13 @@ class Label < Base
...
         def preferred_render_locations
-          if (@tile.cities + @tile.towns).one?
+          if @tile.city_towns.one?
             if @tile.cities.one? && (@tile.cities.first.slots > 1)
               [P_LEFT_CORNER[layout]]
             else
               SINGLE_CITY_ONE_SLOT
             end
-          elsif @tile.cities.size > 1
+          elsif (@tile.cities.size + @tile.towns.size) > 1
```
tobymao, line 92: @tile.city_towns.size > 1

assets/app/view/game/part/town_location.rb
```diff
@@ -0,0 +1,250 @@
...
+
+        # Returns the two edges plus min edge so that a is the edge to render close to, and
+        # (a - b).abs is 1, 2 or 3.
+        def normalized_edges(edge_a, exits)
+          @@normalized_edges ||= {}
+          @@normalized_edges[[edge_a, exits]] ||= if edge_a && exits.size == 2
+                                                    edge_b = exits.find { |e| e != edge_a }
+                                                    edges = [edge_a, edge_b]
+                                                    edges[edges.index(edges.min)] += 6 if (edge_a - edge_b).abs > 3
+                                                    edges.append(edges.min)
```
tobymao, line 94: delete this

assets/app/view/game/part/town_location.rb
```diff
@@ -0,0 +1,250 @@
...
+                    end
+
+            [(edge_a * 60) + delta]
+          elsif edge_a
+            [edge_a * 60]
+          else
+            # This town is in the center. Find the orientation of the town
+            # by looking at the first path connected to it
+            path = town.paths.first
+            if path.edges.any?
```
tobymao, line 183: if (edge = path.exits&.first)
[edge * 60]

assets/app/view/game/part/town_location.rb
```diff
@@ -0,0 +1,250 @@
...
+                else
+                  CENTER
+                end
+          [loc]
+        end
+
+        # Returns an array of rotation options for the town rectangle that
+        # corresponds to the positions given from track_location
+        def town_rotation_angles(tile, town, edges)
+          edge_a, = edges
```
tobymao, line 153: should this be cacheD?

assets/app/view/game/part/town_location.rb
```diff
@@ -0,0 +1,250 @@
...
+              else
+                [0]
+              end
+            end
+          end
+        end
+
+        # Returns an array of weights, location and rotations
+        def town_position(tile, town, edges)
+          edge_a, = edges
```
tobymao, line 199: should this cache? on some good key though since tile and town are objects

assets/app/view/game/part/town_rect.rb
```diff
@@ -209,7 +53,7 @@ def preferred_render_locations
         def load_from_tile
           @tile = @town.tile
           @edge = @tile.preferred_city_town_edges[@town]
-          @num_towns = @tile.towns.size
+          @num_cts = @tile.cities.size + @tile.towns.size
```
tobymao, line 56: @tile.cite_towns.size

lib/engine/part/path.rb
```diff
@@ -5,12 +5,16 @@
 module Engine
   module Part
     class Path < Base
-      attr_reader :a, :b, :branch, :city, :edges, :junction, :node, :offboard, :stop, :town, :terminal
+      attr_reader :a, :b, :city, :edges, :junction, :offboard,
```
tobymao, line 8: put these back in alphabetical order

lib/engine/tile.rb
```diff
@@ -305,7 +309,7 @@ def compute_city_town_edges
 
       # populate ct_edges and edge_count as described in above comments
       paths.each do |path|
-        next unless (ct = path.city || path.town)
+        next unless path.exits.any? && (ct = path.city || path.town)
```
tobymao, line 312: i don't think this is necessary. if there are no exits the iterator won't do anything

lib/engine/tile.rb
```diff
@@ -415,9 +423,9 @@ def separate_parts
         end
       end
 
-      @nodes = @paths.map(&:node).compact.uniq
-      @branches = @paths.map(&:branch).compact.uniq
-      @stops = @paths.map(&:stop).compact.uniq
+      @nodes = @paths.flat_map(&:nodes).compact.uniq
```
tobymao, line 426: i believe none of these should need compact, can you check?

assets/app/view/game/part/town_location.rb
roseundy: Hmmm.  It will only be called once with a given town.

assets/app/view/game/part/town_location.rb
roseundy: Again, it will only be called once for a given town.

assets/app/view/game/part/town_rect.rb
roseundy: Yep, missed it.

lib/engine/part/path.rb
roseundy: Roger.

lib/engine/tile.rb
roseundy: Yep, I'll check.

assets/app/view/game/part/label.rb
roseundy: Yep, missed it.

assets/app/view/game/part/town_location.rb
roseundy: Rats. Didn't mean to do that.

assets/app/view/game/part/town_location.rb
roseundy: Got it

lib/engine/tile.rb
roseundy: Roger

lib/engine/tile.rb
roseundy: #compact does appear to be superfluous here.

lib/engine/tile.rb
roseundy: Well, technically is a stop without a path really a stop (from a routing standpoint)?

Issue #1516, opened 2020-08-24
18GA: Corporation owning W&S company does not pay for normal tokening of Waycross


perwestling:
Multi player game. Fails after latest update with W&S company.

Game fails at [(action 284)](https://18xx.games/game/7315?action=284) 
Game expects Place token but game log is Run route
Closer inspection shows that corporation has $113 and available token, and city, so Place token is correct.

The reason for this can be found at [(action 116)](https://18xx.games/game/7315?action=116) where ACL is placing a token and do not pay anything despite it should cost $40 as it is not using the free teleport - does not have to as it has normal connection.

So the theory is that before the update ACL spent $40 for the token, and when action 284 come around ACL now had $73 and not enough to pay for a token so that the tokening was silently skipped.

perwestling:
Code investigation confirms this. If putting token in hex (that is specified in the ability) then the ability is used automatically. Reverting the W&S commit solves the problem, but it need to be fixed and then this should be tested again.

perwestling:
After the fix corporation ACL now pays $40 in action 116 which removes the bug. So this do no longer break the game.

PR #1544, opened 2020-08-29
18GA: Implement Waycross & Southern RR (fixes #1496, #1516)

### LINKED ISSUE #1496: 18GA: Private Waycross & Southern RR does not support teleport of token
Private is correct when it comes to price, income and not-blocking.

But a corporation who have bought Waycross & Southern RR does not get any teleport token ability for the Waycross hex.

perwestling:
Now it should be possible to use Waycross & Southern RR to teleport
the token for free. And when making a normal, connected, tokening
in Waycross the token is paid for.

### DIFF
--- a/lib/engine/ability/token.rb
+++ b/lib/engine/ability/token.rb
@@ -7,7 +7,7 @@ module Engine
     class Token < Base
       attr_reader :hexes, :price, :teleport_price, :extra, :from_owner
 
-      def setup(hexes:, price:, teleport_price: nil, extra: nil, from_owner: false)
+      def setup(hexes:, price:, teleport_price: nil, extra: nil, from_owner: nil)
         @hexes = hexes
         @price = price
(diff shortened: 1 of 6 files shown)


lib/engine/ability/token.rb
```diff
@@ -5,13 +5,14 @@
 module Engine
   module Ability
     class Token < Base
-      attr_reader :hexes, :price, :teleport_price, :extra
+      attr_reader :hexes, :price, :teleport_price, :extra, :from_owner
 
-      def setup(hexes:, price:, teleport_price: nil, extra: nil)
+      def setup(hexes:, price:, teleport_price: nil, extra: nil, from_owner: false)
```
tobymao, line 10: this can just be nil because you already do or false

perwestling: Fixed

PR #1546, opened 2020-08-29
18MS: Handle special rules for Atlanta, Chattanooga and New Orleans.

perwestling:
First to token in New Orleans gets a one time corporation bonus
of $100. And also receives a route bonus of $10 for any route
include New Orleans. The route bonus are removed when the first
6 train is bought.

A coins icon has been added on the New Orleans hex. When bonus is
triggered the icon is removed.

The first corporation connecting to Chattanooga receives a one
time bonus to treasury of $50.

A corporation can only use Atlanta once per run.


perwestling:
Notice how the coin icon looks like in New Orleans and Chattanooga

image

lib/engine/game/g_18_ms.rb
```diff
@@ -109,11 +117,20 @@ def purchasable_companies(entity = nil)
       end
 
       def revenue_for(route)
-        # Diesels double to normal revenue
-        route.train.name.end_with?('D') ? 2 * super : super
+        # Diesels double normal revenue
```
tobymao, line 120: revenue = super
revenue *= 2 if ...

lib/engine/step/g_18_ms/track.rb
```diff
@@ -0,0 +1,31 @@
...
+        ACTION_HEX_OF_INTEREST = 'B10'
+        BONUS_HEX = 'B12'
+
+        def lay_tile(action, extra_cost: 0, entity: nil)
+          entity ||= action.entity
+          super
+
+          return if action.tile.hex.name != ACTION_HEX_OF_INTEREST ||
+            @game.chattanooga_reached ||
+            !@game.loading ||
```
tobymao, line 18: this seems wrong. this means there will be different behavior

lib/engine/game/g_18_ms.rb
perwestling: Fixed

lib/engine/step/g_18_ms/track.rb
perwestling: I have made a fix. Better now?

lib/engine/step/g_18_ms/track.rb
```diff
@@ -0,0 +1,30 @@
...
+        ACTION_HEX_OF_INTEREST = 'B10'
+        BONUS_HEX = 'B12'
+
+        def lay_tile(action, extra_cost: 0, entity: nil)
+          entity ||= action.entity
+          super
+
+          return if action.tile.hex.name != ACTION_HEX_OF_INTEREST ||
+            @game.chattanooga_reached ||
+            (!@game.loading && @game.graph.reachable_hexes(entity).find { |h, _| h.name == 'B12' }.nil?)
```
tobymao, line 18: still needs to be fixed

perwestling: I might not have understood. Have made yet another attempt.

What I do want to accomplish is to check if the tile placed in B10 creates a connection for the entity towards B12 (Chattanooga). So is there a better way to do this?

tobymao: the reason why you see game.loading checks is because any graph call (reachable hexes etc) are very expensive to compute. but anytime you see a game.loading check, it's purely to determine if something was legal or not, it doesn't have any side effects. however in your case, based on the graph call, you make a side effect.

that means that you cannot have different behavior for when game.loading and when game is not loading. thus, you should get rid of game.loading and pay the cost. or come up with something more clever.

perwestling: Ok, got it. In this case I think the cost is OK, as it just checked when placing tiles in one hex on the board (B10) that can trigger it, and it is very likely that they do connect to B12 when they do tile B10. 
So I have removed the game.load, and put the flag check first for readability.

perwestling:
Force-push? No, I did a normal push? A bug in github?

PR #1548, opened 2020-08-29
18MS: Implement Meridian & Memphis Railway private

perwestling:
This private has a one time ability (for a corporation) to place
a token at half price. A similar way of specifying this as for
train_discount is used. If the value is between 0 and 1, it is
how big part is a discount of the tokening price. So 0.5 means
half price.

Use similar method as was done in 18GA Waycross & Southern to
just use ability if ability is actived explicitly.

Have made a change to hex selection. If hexes is empty this
means all (connected) hexes is allowed.

Use of ability does not close the private.


perwestling:
The method used to have hexes as [] in the token ability, and use that as any-hex might not be the correct solution. Compare tile_lay for NWR in 1882.

perwestling:
I suspect the implementation might be wrong, allowing for teleportation of token. So I need to try this further.

perwestling:
If one do attempt to token a non-connected city, you get the game error "Cannot place token on XX because it is not connected".
I think this can be good enough.

lib/engine/config/game/g_18_ms.rb
```diff
@@ -179,7 +179,17 @@ module G18MS
...
          "revenue":15,
          "desc":"The owning Major Company may lay their cheapest available token for half price. This is not an extra token placement. This ability can only be used once, and using it does not close the private company.",
-         "sym":"M&M"
+         "sym":"M&M",
+         "abilities": [
+            {
+              "type": "token",
+              "owner_type":"corporation",
+              "hexes": [],
+              "price": 0.5,
```
tobymao, line 188: would it be better to call it a discount? what do the other things do?

tobymao:
still has unaddressed comments and merge conflict

lib/engine/config/game/g_18_ms.rb
perwestling: I have now added a new attribute 'discount' that is used for this purpose.

perwestling:
Have now updated the commit, and rebased towards latest master.

lib/engine/ability/token.rb
```diff
@@ -5,14 +5,15 @@
 module Engine
   module Ability
     class Token < Base
-      attr_reader :hexes, :price, :teleport_price, :extra, :from_owner
+      attr_reader :hexes, :price, :teleport_price, :extra, :from_owner, :discount
 
-      def setup(hexes:, price:, teleport_price: nil, extra: nil, from_owner: nil)
+      def setup(hexes:, price: nil, teleport_price: nil, extra: nil, from_owner: nil, discount: nil)
```
tobymao, line 10: why did you change price?

lib/engine/step/tokener.rb
```diff
@@ -72,17 +72,25 @@ def adjust_token_price_ability!(entity, token, hex)
...
           return [token, ability]
         end
 
         [token, nil]
       end
+
+      private
+
+      def new_token_price(token, ability)
+        return ability.price if ability.price
```
tobymao, line 90: shouldn't this just be abstracted in the ability class?

in class ability, you can have the price reflect the discount

perwestling: Have now replaced price attribute with price method in ability.

lib/engine/ability/token.rb
perwestling: In case if discount used there is no meaning of price, so there is no reasonable default value.
Should I add a guard/game_error in case both price and discount are nil?

PR #1550, opened 2020-08-30
18MS: Implement private P1 and P2 + general tile lay anywhere

perwestling:
This PR implements private P1 and P2 in 18MS, which are privates that can lay extra yellow tiles during tile lay.
This require the possibility to specify empty hexes and tiles for the ability, so supporting this in the general code has been added.

### DIFF
--- a/lib/engine/step/special_track.rb
+++ b/lib/engine/step/special_track.rb
@@ -29,13 +29,13 @@ module Engine
       end
 
       def process_lay_tile(action)
-        lay_tile(action)
+        lay_tile(action, spender: action.entity.owner)
         check_connect(action)
         ability(action.entity).use!
       end
 
       def available_hex(entity, hex)
-        return unless ability(entity).hexes.include?(hex.id)
+        return if ability(entity).hexes.any? && !ability(entity).hexes.include?(hex.id)
 
         @game.hex_by_id(hex.id).neighbors.keys
       end
@@ -43,9 +43,10 @@ module Engine
       def potential_tiles(entity, hex)
         return [] unless (tile_ability = ability(entity))
 
-        tile_ability
-          .tiles
-          .map { |name| @game.tiles.find { |t| t.name == name } }
+        tiles = tile_ability.tiles.map { |name| @game.tiles.find { |t| t.name == name } }
+        tiles = @game.tiles.uniq(&:name) if tile_ability.tiles.empty?
+
+        tiles
           .compact
           .select { |t| @game.phase.tiles.include?(t.color) && @game.upgrades_to?(hex.tile, t, tile_ability.special) }
       end
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -72,7 +72,15 @@ module Engine
         discount = 0
 
         entity.abilities(:tile_lay) do |ability|
-          next if !ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name)
+          next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
+
+          if ability.reachable &&
+            !@game.loading &&
+            !@game.graph.reachable_hexes(spender)[hex] &&
+            hex.name != spender.coordinates
+
+            @game.game_error("Track laid must be connected to one of #{spender.id}'s stations")
+          end
 
           free = ability.free
           discount = ability.discount
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -74,13 +74,10 @@ module Engine
         entity.abilities(:tile_lay) do |ability|
           next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
 
-          if ability.reachable &&
+          @game.game_error("Track laid must be connected to one of #{spender.id}'s stations") if ability.reachable &&
+            hex.name != spender.coordinates &&
             !@game.loading &&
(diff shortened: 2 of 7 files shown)


lib/engine/step/special_track.rb
```diff
@@ -35,17 +35,18 @@ def process_lay_tile(action)
...
       end
 
       def potential_tiles(entity, hex)
         return [] unless (tile_ability = ability(entity))
 
-        tile_ability
-          .tiles
-          .map { |name| @game.tiles.find { |t| t.name == name } }
+        tiles = tile_ability.tiles.map { |name| @game.tiles.find { |t| t.name == name } }
+        tiles = @game.tiles.uniq(&:name) if tiles.empty?
```
tobymao, line 47: this seems like it could be bug prone. what happens if you've run out of tiles, then you allow all?

instead shouldn't you check if tile_abilities doesn't have any tiles?

lib/engine/step/tracker.rb
```diff
@@ -71,14 +71,22 @@ def lay_tile(action, extra_cost: 0, entity: nil)
         discount = 0
 
         entity.abilities(:tile_lay) do |ability|
-          next if !ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name)
+          next unless ability.hexes.empty?
+
+          corporation = entity.owner
+          @game.game_error("Track laid in must be connected to one of #{corporation.id}'s stations") if
+            !@game.graph.reachable_hexes(corporation)[hex] && hex.name != corporation.coordinates
```
tobymao, line 78: will this break 1889? is this rule actually generic, also you shouldn't do reachable checks outside of a game.loading for performance reasons

lib/engine/step/tracker.rb
```diff
@@ -71,14 +71,22 @@ def lay_tile(action, extra_cost: 0, entity: nil)
...
+        entity.abilities(:tile_lay) do |ability|
+          next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
 
           free = ability.free
           discount = ability.discount
         end
 
         entity.abilities(:teleport) do |ability, _|
-          ability.use! if ability.hexes.include?(hex.id) && ability.tiles.include?(tile.name)
+          ability.use! if ability.hexes.any? && ability.hexes.include?(hex.id) && ability.tiles.include?(tile.name)
```
tobymao, line 89: don't you want to use the ability regardless? if abilitiy.hexes is an empty array, you'll never use the teleport ability

lib/engine/step/tracker.rb
```diff
@@ -71,14 +71,22 @@ def lay_tile(action, extra_cost: 0, entity: nil)
...
-          next if !ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name)
+          next unless ability.hexes.empty?
+
+          corporation = entity.owner
+          @game.game_error("Track laid in must be connected to one of #{corporation.id}'s stations") if
+            !@game.graph.reachable_hexes(corporation)[hex] && hex.name != corporation.coordinates
+        end if entity.company?
+
+        entity.abilities(:tile_lay) do |ability|
+          next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
```
tobymao, line 82: same here

lib/engine/step/special_track.rb
perwestling: Check tile_ability.tiles instead.

lib/engine/step/tracker.rb
perwestling: The only tile_lay ability outside 18MS with no hexes specified is NWR in 1882. The 1889 one have hexes as it specifies the coast towns.
Now I suppose 1882 should work anyway, but I have added 
  ability.when == 'track' &&
which excludes that ability.

But does not 1882 NWR require the tile lay to be connected?

An alternative is to add an attribute that specifies if tile lay need to be connected and use that to decide if this check should be done or not.

lib/engine/step/tracker.rb
perwestling: Changed backed.

lib/engine/step/tracker.rb
perwestling: I think it is correct, after the change of the code. We do want to use free and discount ability attributes if hexes is empty. 
In case of the tile_lay in 1882 it will mean that the following code is exercised, which is wasn't before but the NWR does not specify free or discount, so it will use default which is free=false and discount=0, so it wont change anything.

So this should be OK.

perwestling:
@tobymao Have now updated based on your comments, and rebased the commits.
The problem of doing tile lay outside of track lay still remains but that is a known problem.

lib/engine/step/tracker.rb
```diff
@@ -72,7 +72,15 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
         discount = 0
 
         entity.abilities(:tile_lay) do |ability|
-          next if !ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name)
+          next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
+
+          if ability.reachable &&
+            !@game.loading &&
+            !@game.graph.reachable_hexes(spender)[hex] &&
+            hex.name != spender.coordinates
```
tobymao, line 80: move the hex.name check first to improve performance, also can we make it inline

game_error if x &&
y &&
z

perwestling: Correction pushed.

PR #1551, opened 2020-08-30
18MS: Assisted permanent train buy

perwestling:
If a corporation owns trains but cannot afford to buy an available
permanent train (but has at least 50% of its price) the president
can assist in the purchase buy spending the difference (and an
extra $50 "fee").

Have added a president_assisted_buy in the base class that can
be overridden if the game support this feature.

### DIFF
--- a/lib/engine/step/g_18_ms/buy_train.rb
+++ b/lib/engine/step/g_18_ms/buy_train.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+require_relative '../buy_train'
+
+module Engine
+  module Step
+    module G18MS
+      class BuyTrain < BuyTrain
+        def process_buy_train(action)
+          entity ||= action.entity
+          price = action.price
+          train = action.train
+          player = entity.player
+
+          president_assist, fee_amount = @game.president_assisted_buy(entity, train, price)
+
+          if president_assist.positive?
+            player.spend(president_assist + fee_amount, @game.bank)
+            @game.bank.spend(president_assist, entity)
+            assist = @game.format_currency(president_assist).to_s
+            fee = @game.format_currency(fee_amount).to_s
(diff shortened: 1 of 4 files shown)


tobymao:
i wonder if president assist needs to be in the action. it can be inferred by simply doing it, for example with a lower price.

if you create a train buy action with less money, then the 18MS specific code could make up the difference automatically

perwestling:
> i wonder if president assist needs to be in the action. it can be inferred by simply doing it, for example with a lower price.
> 
> if you create a train buy action with less money, then the 18MS specific code could make up the difference automatically

I have now redesigned it. So the changes to the general code is introduction of the president_assisted_buy method in the base class (which 18MS overrides) and the change of button text from "Buy" to "Assisted buy" in buy_trains.rb in case president_assisted_buy method returns any assist amount. 

Now I do not know any other 18xx that has such an assisted buy. Do not know if it possible to make more general?

lib/engine/step/g_18_ms/buy_train.rb
```diff
@@ -0,0 +1,29 @@
...
+          price = action.price
+          train = action.train
+          player = entity.player
+
+          president_assist, fee = @game.president_assisted_buy(entity, train, price)
+
+          if president_assist.positive?
+            player.spend(president_assist + fee, @game.bank)
+            @game.bank.spend(president_assist, entity)
+            @log << "#{player.name} pay #{@game.format_currency(president_assist)} to assist buy of #{train.name} train"
```
tobymao, line 20: Player pays $100 and an additional $5 fee to assist buying a 5 train.

perwestling: image

perwestling: 6 trains normally cost $550. The log text are added before super method is called so it will be before the phase change message, and not connected to the log text about train buy. The emergency buy assist does appear together with the buy train log.
I think this might be good enough.

Issue #1553, opened 2020-08-30
Cannot emergency buy trains as it is not buyable


perwestling:
Testing 18MS on master and came in a situation where a corporation needs to buy a train. 

The player has $869 so it should not be a problem. But for some reason a "Train not buyable" error appear.

To me the code in train.rb, buy_train_action, looks incorrect as the following:
  @game.game_error('Not a buyable train') unless buyable_train_variants(train, entity).include?(train.variant)
is before the 
  if remaining.positive? && must_buy_train?(entity)
that adds player money.
![image](https://user-images.githubusercontent.com/2631151/91663015-6c629880-eae6-11ea-8a06-efedbca870a6.png)

{
  "status": "active",
  "actions": [
    {
      "type": "bid",
      "entity": "Player 4",
      "entity_type": "player",
      "id": 1,
      "company": "AGS",
      "price": 30
    },
    {
      "type": "bid",
      "entity": "Player 3",
      "entity_type": "player",
      "id": 2,
      "company": "BS",
      "price": 40
    },
    {
      "type": "bid",
      "entity": "Player 2",
      "entity_type": "player",
      "id": 3,
      "company": "M&M",
      "price": 50
    },
    {
      "type": "bid",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 4,
      "company": "MC",
      "price": 60
    },
    {
      "type": "par",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 5,
      "corporation": "GMO",
      "share_price": "70,1,2"
    },
    {
      "type": "par",
      "entity": "Player 2",
      "entity_type": "player",
      "id": 6,
      "corporation": "IC",
      "share_price": "70,1,2"
    },
    {
      "type": "par",
      "entity": "Player 3",
      "entity_type": "player",
      "id": 7,
      "corporation": "L&N",
      "share_price": "70,1,2"
    },
    {
      "type": "par",
      "entity": "Player 4",
      "entity_type": "player",
      "id": 8,
      "corporation": "Fr",
      "share_price": "70,1,2"
    },
    {
      "type": "buy_shares",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 9,
      "shares": [
        "GMO_1"
      ]
    },
    {
      "type": "buy_shares",
      "entity": "Player 2",
      "entity_type": "player",
      "id": 10,
      "shares": [
        "IC_1"
      ]
    },
    {
      "type": "buy_shares",
      "entity": "Player 3",
[...]

PR #1559, opened 2020-08-31
18MS: Handle emergency buy

perwestling:
Fix bug #1553.

If player is missing resources during emergency buy, the player must borrow the needed money.

### DIFF
--- a/assets/app/view/game/emergency_money.rb
+++ b/assets/app/view/game/emergency_money.rb
@@ -24,7 +24,7 @@ module View
           children << h(:div, props, corp)
         end
 
-        children << render_bankruptcy
+        children << render_bankruptcy if @game.round.steps.find { |s| s.respond_to?(:process_bankrupt) }
         children
       end
 
--- a/assets/app/view/game/emergency_money.rb
+++ b/assets/app/view/game/emergency_money.rb
@@ -24,13 +24,13 @@ module View
           children << h(:div, props, corp)
         end
 
-        children << render_bankruptcy if @game.round.steps.find { |s| s.respond_to?(:process_bankrupt) }
+        children << render_bankruptcy if @game.round.actions_for(entity).include?('bankrupt')
         children
       end
 
       def render_bankruptcy
         resign = lambda do
-          process_action(Engine::Action::Bankrupt.new(@game.round.active_step.current_entity))
+          process_action(Engine::Action::Bankrupt.new(entity))
         end
 
         props = {
@@ -41,6 +41,12 @@ module View
         }
 
         h(:button, props, 'Declare Bankruptcy')
+      end
+
+      private
+
+      def entity
+        @game.round.active_step.current_entity
       end
     end
   end
--- a/lib/engine/game/g_18_ms.rb
+++ b/lib/engine/game/g_18_ms.rb
@@ -70,12 +70,21 @@ module Engine
           @phase.next! if @turn == 1 && round_num == 2
         end
 
+        if round_num == 1
+          @players.each do |p|
+            next unless p.cash.negative?
+
+            debt = -p.cash
+            interest = (debt / 2.0 + 0.5).floor
+            p.spend(interest, @bank, check_cash: false)
+            @log << "#{p.name} has to borrow another #{format_currency(interest)} as being in debt at end of SR"
+          end
+        end
         super
       end
 
       def operating_round(round_num)
         Round::Operating.new(self, [
-          Step::Bankrupt,
           Step::Exchange,
           Step::DiscardTrain,
           Step::SpecialTrack,
--- a/lib/engine/game/g_18_ms.rb
+++ b/lib/engine/game/g_18_ms.rb
@@ -75,7 +75,7 @@ module Engine
             next unless p.cash.negative?
 
             debt = -p.cash
-            interest = (debt / 2.0 + 0.5).floor
+            interest = (debt / 2.0).ceil
             p.spend(interest, @bank, check_cash: false)
             @log << "#{p.name} has to borrow another #{format_currency(interest)} as being in debt at end of SR"
           end
--- a/lib/engine/step/train.rb
+++ b/lib/engine/step/train.rb
@@ -33,12 +33,12 @@ module Engine
         train.variant = action.variant
         price = action.price
         exchange = action.exchange
-        @game.phase.buying_train!(entity, train)
 
         # Check if the train is actually buyable in the current situation
         @game.game_error('Not a buyable train') unless buyable_train_variants(train, entity).include?(train.variant)
 
         remaining = price - entity.cash
+        delayed_log = nil
         if remaining.positive? && must_buy_train?(entity)
           cheapest = @depot.min_depot_train
           if train != cheapest && (!@game.class::EBUY_OTHER_VALUE || train.from_depot?)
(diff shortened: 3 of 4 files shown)


lib/engine/game/g_18_ms.rb
```diff
@@ -62,12 +62,22 @@ def new_operating_round(round_num = 1)
           @phase.next! if @turn == 1 && round_num == 2
         end
 
+        if round_num == 1
+          @players.each do |p|
+            next unless p.cash.negative?
+
+            debt = -p.cash
+            interest = (debt / 2 + 0.99).floor
```
tobymao, line 70: what's the point of the 0.99?

int / int = int

lib/engine/game/g_18_ms.rb
```diff
@@ -62,12 +62,22 @@ def new_operating_round(round_num = 1)
...
+            @log << "#{p.name} has to borrow another #{format_currency(interest)} as being in debt at end of SR"
+          end
+        end
         super
       end
 
       def operating_round(round_num)
         Round::Operating.new(self, [
-          Step::Bankrupt,
+          Step::G18MS::Bankrupt,
```
tobymao, line 80: why do we need step::bankrupt at all?can't you just not have anything

lib/engine/step/g_18_ms/bankrupt.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
tobymao, line 1: this file seems like it's not needed

lib/engine/step/train.rb
```diff
@@ -61,6 +61,9 @@ def buy_train_action(action, entity = nil)
 
         source = @depot.discarded.include?(train) ? 'The Discard' : train.owner.name
 
+        @game.phase.buying_train!(entity, train)
```
tobymao, line 64: i wonder how if this will break games

tobymao: i'm pretty sure this change is going to break hundreds of games

lib/engine/game/g_18_ms.rb
perwestling: When doing emergency buy, before you have sold all shares, you can press the "Declare Bankrupt" button in the view.

Removing the file would mean that you get some game exception as bankrupt action is not handled.

So to avoid this problem I had two choices:
- Remove the button (based on a [class] constant) - e.g show only if DECLARE_BANKRUPCY is true [by default this is true, 18MS it is false]
- If button pressed make a game_error raise explaining that you cannot declare bankrupcy but have to buy the train

With the option to remove the step from 18MS, which one do you prefer?

lib/engine/step/g_18_ms/bankrupt.rb
perwestling: See above - would result in an unhandled action without changes of the view code.

lib/engine/step/train.rb
perwestling: I do not think it will break any games but I have only tested with rake tests, and it passes them.

The implementation before did like this:
- Oh, I would like to buy this train from the depot
- Ok, before doing that let's change phase if that train triggers a phase change
- And now do give us the cheapest train if we add new trains for the next phase

For all other game titles implemented the new trains are not cheaper than the train triggering phase change. But for 18MS 6 train cost $550, and the newly available 2D cost $500. This means that old implementation stops you from buying the 6 train....

But is it possible to try this change with all games to see if it does break any game at all?

lib/engine/game/g_18_ms.rb
tobymao: can't you just check the available actions?

lib/engine/step/train.rb
tobymao: i think the issue will be that the buying_train triggers train rusting allowing people to buy the next train even though it becomes traight tight

tobymao: yes, i can try this change before we deploy it

lib/engine/game/g_18_ms.rb
```diff
@@ -62,12 +62,22 @@ def new_operating_round(round_num = 1)
           @phase.next! if @turn == 1 && round_num == 2
         end
 
+        if round_num == 1
+          @players.each do |p|
+            next unless p.cash.negative?
+
+            debt = -p.cash
+            interest = (debt / 2.0 + 0.99).floor
```
tobymao, line 70: i still don't quite understand the + 0.99 can you explain to me what your intention is?

lib/engine/game/g_18_ms.rb
perwestling: My idea was that
- if debt is 50, interest should be 25
- if debt is 51, interest should be 26

So (51 / 2.0 + 0.99) = 26.49 which floor gives 26
And (50 / 2.0 + 0.99) = 25.99 which floor gives 25

This is a VERY minor thing. 

But in case of debt is 1, 1 / 2 = 0.
Which would mean
  Player 1 has to borrow another $0 as being in debt at the end of SR

lib/engine/game/g_18_ms.rb
perwestling: See comment above.
But the idea was to make interest for debt of $1 to be positive. That is odd debt are rounded up.

lib/engine/game/g_18_ms.rb
perwestling: Ah, yes it is a better solution. Do know not render the "Declare Bankruptcy" button if there is not step that handle process_bankrupt action.

lib/engine/game/g_18_ms.rb
perwestling: But I have changed to + 0.5; which might make it clearer that it is just to get correct rounding.

The 0.99 was just to ensure that any problem with float handling would not cause any problem.

lib/engine/step/train.rb
perwestling: I do not see that this should cause any problems. The check if you are allowed to buy any trains is done with the phase that are active before you press the button. Besides it is a bug in the current implementation that I am trying to fix.

But trying it out on the existing games should clarify if there is a problem. At least the spec test games does not report any problem.

tobymao:
why don’t you just do ceiling?

perwestling:
> why don’t you just do ceiling?
Why indeed.... Must have forgotten about ceiling.
Have now added a commit where I change to 
  interest = (debt / 2.0).ceil

assets/app/view/game/emergency_money.rb
```diff
@@ -24,7 +24,7 @@ def render_emergency_money_raising(player)
           children << h(:div, props, corp)
         end
 
-        children << render_bankruptcy
+        children << render_bankruptcy if @game.round.steps.find { |s| s.respond_to?(:process_bankrupt) }
```
tobymao, line 27: this doesn't look good

i think maybe your intention here is to 

@game.round.actions_for(entity).include?('bankrupt')

lib/engine/game/g_18_ms.rb
```diff
@@ -70,12 +70,21 @@ def new_operating_round(round_num = 1)
           @phase.next! if @turn == 1 && round_num == 2
         end
 
+        if round_num == 1
```
tobymao, line 73: round_num.one?

perwestling: round_num is a number. Is does not have the one? method, does it?

assets/app/view/game/emergency_money.rb
perwestling: Ok, modified the code.

lib/engine/game/g_18_ms.rb
tobymao: you're right

tobymao (approved):

PR #1574, opened 2020-09-02
(fixes #1450) Privates in auction now show discount

### LINKED ISSUE #1450: Indicate company discount on card
In 1882, when everyone passes the price of the cheapest company decreases by 5. This is indicated in the log but the card for the company still shows just the full value. It would be nice if, when a company was available for a discount, the true selling price were indicated on the card.

Presumably this also shows up in 1846 (when people pass on the very last company in the draft) and in 1830-ish games where no one buys a private at all (although that very rarely happens).

captianobi:
I have fixed the issue #1450 and the company cards when auctioned now indicate the proper discount instead of only showing the initial value. 
image
as seen the card now matches the new value that is shown in chat.

### DIFF
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -100,7 +100,7 @@ module View
             h(:div, { style: header_style }, 'PRIVATE COMPANY'),
             h(:div, @company.name),
             h(:div, { style: description_style }, @company.desc),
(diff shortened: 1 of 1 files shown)


captianobi:
@dfannius recomended some changes regrading 1846 and original value

dfannius:
The discount should be displayed separately from the value, since the original value still matters (for buying in).

Also, I think the 1846 privates with debt need to be handled specially, since their debt is currently implemented as a negative discount.

PR #1576, opened 2020-09-02
(fixes #1450) Fixes the issue where the price on the cards would not go down if there was an discount. 

### LINKED ISSUE #1450: Indicate company discount on card

captianobi:
(fixes #1450) The issue has been fixed. A new price: value has been added to show the current price of the company. I would like to request assistance for fixing the issue for 1846 as the discount includes the debt of companies like the big four.
image

### DIFF
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -100,7 +100,7 @@ module View
             h(:div, { style: header_style }, 'PRIVATE COMPANY'),
             h(:div, @company.name),
             h(:div, { style: description_style }, @company.desc),
-            h(:div, { style: value_style }, "Value: #{@game.format_currency(@company.value-@company.discount)}"),
+            h(:div, { style: value_style }, "Value: #{@game.format_currency(@company.value - @company.discount)}"),
             h(:div, { style: revenue_style }, "Revenue: #{@game.format_currency(@company.revenue)}"),
           ]
 
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -72,6 +72,10 @@ module View
             float: 'right',
           }
 
+          current_value_style = {
+            float: 'center'
+          }
+
           bidders_style = {
             marginTop: '0.5rem',
             display: 'inline-block',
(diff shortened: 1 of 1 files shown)


assets/app/view/game/company.rb
```diff
@@ -109,6 +113,14 @@ def render
             children << render_bidders
           end
 
+          if @company.discount != 0
+            children << h(
+            :div,
+            { style: current_value_style },
```
tobymao, line 119: move line 75 inline here

captianobi:
I would like to request help with 1846 and it's debt, I'd recommend creating a new value for the display such as realDiscount and then using that to actually show the discount instead of using just @company.discount

assets/app/view/game/company.rb
```diff
@@ -109,6 +113,14 @@ def render
             children << render_bidders
           end
 
+          if @company.discount != 0
```
tobymao, line 116: if company.discount.positive? or unless company.discount.zero?

captianobi: Unless company.discount.zero. making it positive won't help fix the 1846 issue.

dfannius: The 1846 minors can be discounted too. I'm not sure how it's implemented on this site and it may not even be correct if the implementation overloads discount and debt, since I believe the discount is supposed to apply to the buying price rather than the debt.

dfannius: (Confirmed: https://boardgamegeek.com/thread/1880571/article/27393557#27393557)

captianobi:
So the issue with the 1846 minors is that the discount starts at -60 to implement the 60$ debt that the purchaser of the companies has to pay. So off the bat the code will display price: 100. As the value is 40 plus the 60$ debt.

tobymao (approved):

PR #1578, opened 2020-09-02
(fixes 7582) stops Big 4 and MS from going less than 0

captianobi:
(fixes 7582)


lib/engine/step/g_1846/draft_distribution.rb
```diff
@@ -90,7 +90,15 @@ def process_pass(action)
             "to #{@game.format_currency(new_value)}"
 
           @round.next_entity_index!
-          return if new_value.positive?
+          if company.name == 'Big 4'
```
tobymao, line 93: case company.id
when 'BIG4'
return if new_value >= 60
when 'MS' 
return if new_value >= 80
else
return if new_value.positive?

jrimmer:
Nice work, @captianobi !

lib/engine/step/g_1846/draft_distribution.rb
```diff
@@ -90,14 +90,13 @@ def process_pass(action)
...
-          elsif new_value.positive?
-            return true
+          case company.id
+          when 'Big 4'
+            return if new_value >= 60
+          when 'MS'
+            return if new_value >= 80
           else
-            return false
+            return new_value.positive?
```
tobymao, line 99: this is not correct, it should be return if new_value.positive?

captianobi: All fixed

tobymao (approved):

PR #1584, opened 2020-09-04
Cache stop calculation for routes

dfannius:


tobymao:
let's change this be caching style

def stops
@stops ||= begin do
...
end
end

and then when you want to clear the cache, just do @stops = nil

dfannius:
Do you really mean Stops with a capital s? That doesn't look like the usual style. (And if not, then is there a conflict between the @stops variable and the stops method?

tobymao:
no, just @stops, and there's no conflict, stops is the method, @stops is the instance variable

dfannius:
I'm changing it to that style on the branch that implements "smart stops", so I'm going to submit it as part of that PR rather than doing this separately here and then rebasing.

PR #1587, opened 2020-09-05
Smart stops (fixes #1002, #1053)

### LINKED ISSUE #1002: 1846 - "Always include the ends of a route" has no workaround if company has the mail contract
https://github.com/tobymao/18xx/blob/bb418cd9e72b441b32ea49c08faba8f822905f46/lib/engine/route.rb#L148-L149

dfannius:
New stop-choosing code looks at all possibilities with the maximum usable number of stops and chooses the optimal one according to the game's rules, rather than picking them generically and then applying game-specific bonuses.


lib/engine/route.rb
```diff
@@ -15,6 +15,7 @@ def initialize(game, phase, train, connection_hexes: [], routes: [], override: n
       @routes = routes
       @override = override
       @game = game
+      @stops = []
```
tobymao, line 18: this should be nil

line 45 is not needed

make restore_connections private

lib/engine/route.rb
```diff
@@ -141,37 +145,61 @@ def visited_stops
     end
 
     def stops
+      @stops ||= compute_stops
+    end
+
+    def compute_stops
```
tobymao, line 151: lets make this private as well for now

tobymao: or put it inline in stops

tobymao:
diff --git a/lib/engine/route.rb b/lib/engine/route.rb
index b0ee78bc..f451d70c 100644
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -152,7 +152,6 @@ module Engine
       visits = visited_stops
       distance = @train.distance
       return visits if distance.is_a?(Numeric)
-
       return [] if visits.empty?
 
       # types_pay lists how many locations of each type can be hit. A
@@ -163,44 +162,40 @@ module Engine
       # Stops use the first available slot, so for each stop in this case
       # we'll try to put it in a town slot if possible and then
       # in a city/town/offboard slot.
-      types_pay = distance.map { |h| [h['nodes'], h['pay']] }.sort_by { |t, _| t.size }
+      distance = distance.sort_by { |types, _| types.size }
 
-      best_stops = []
-      best_revenue = 0
       max_num_stops = [distance.sum { |h| h['pay'] }, visits.size].min
+
       max_num_stops.downto(1) do |num_stops|
-        visits.combination(num_stops.to_i).each do |stops| # to_i to work around Opal bug
+        # to_i to work around Opal bug
+        stops, revenue = visits.combination(num_stops.to_i).map do |stops|
           # 1) At least one stop must have a token
           next if stops.none? { |stop| stop.tokened_by?(corporation) }
 
           # 2) We can't ask for more revenue centers of a type than are allowed
-          ok = true
-          types_used = Array.new(types_pay.size, 0) # how many slots of each row are filled
-          stops.each do |stop|
-            row = types_pay.each_index.find(-> { ok = false }) do |i|
-              types_pay[i][0].include?(stop.type) && types_used[i] < types_pay[i][1]
+          types_used = Array.new(distance.size, 0) # how many slots of each row are filled
+
+          next unless stops.all? do |stop|
+            row = distance.index.with_index do |h, i|
+              h['nodes'].include?(stop.type) && types_used[i] < h['pay']
             end
-            break unless ok
 
             types_used[row] += 1
+            row
           end
 
-          next unless ok
-
-          revenue = @game.revenue_for(self, stops)
-          if revenue > best_revenue
-            best_stops = stops
-            best_revenue = revenue
-          end
-        end
+          [stops, @game.revenue_for(self, stops)]
+        end.compact.max_by(&:last)
 
         # We assume that no stop collection with m < n stops could be
         # better than a stop collection with n stops, so if we found
         # anything usable with this number of stops we return it
         # immediately.
-        return best_stops if best_revenue.positive?
+        return stops if revenue.positive?
       end
+
+      []
     end
 
     def hexes

tobymao:
i refactored it a bit 

https://gist.githubusercontent.com/tobymao/64f62082e1f6f2ffeab8ba9d078a98c7/raw/efa6f60cbe89601e28de214cfc6040aad71e16c5/diff

lib/engine/route.rb
dfannius: Making `compute_stops` inline is awkward because it has a few early `return`s, so I'll just make it private.

dfannius:
In your refactor, `types_used[row] += 1` has no effect if `row` is `nil`?

tobymao:
you're right that's wrong i kept thinking types_used was a hash, it should be types_used[row] += 1 if row

dfannius:
Ready for review again.

tobymao (approved):

PR #1588, opened 2020-09-05
Intra-tile path routing support

roseundy:
Now supports routing for tiles with a single intra-tile path (i.e. 18TN, 18MEX).

A fully robust fix will require more substantial changes (and will likely break existing games without taking special action).

### DIFF
--- a/lib/engine/connection.rb
+++ b/lib/engine/connection.rb
@@ -7,10 +7,10 @@ module Engine
     def self.connect!(hex)
       connections = {}
 
-      node_paths, paths = hex.tile.paths.partition { |p| p.nodes.any? }
+      node_paths, paths = hex.tile.paths.partition(&:node?)
 
       paths.each do |path|
-        path.walk { |p| node_paths << p if p.nodes.any? }
+        path.walk { |p| node_paths << p if p.node? }
       end
 
       node_paths.uniq.each do |node_path|
@@ -19,14 +19,13 @@ module Engine
           hex = path.hex
           connection = connection.branch!(path)
           next if connection.paths.include?(path)
-          next if connection.nodes.include?(path.nodes[0])
-          next if connection.nodes.include?(path.nodes[1])
+          next if (connection.nodes & path.nodes).any?
           next if connection.paths.any? { |p| p.hex == hex && (p.exits & path.exits).any? }
 
           connections[connection] = true
           connection.add_path(path)
 
-          if path.exits.none?
+          if path.exits.empty?
             hex_connections = hex.connections[:internal]
             hex_connections << connection unless hex_connections.include?(connection)
           else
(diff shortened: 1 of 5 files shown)


lib/engine/connection.rb
```diff
@@ -21,16 +19,21 @@ def self.connect!(hex)
           hex = path.hex
           connection = connection.branch!(path)
           next if connection.paths.include?(path)
-          # FIXME: for intra-node paths
           next if connection.nodes.include?(path.nodes[0])
+          next if connection.nodes.include?(path.nodes[1])
```
tobymao, line 23: you can replace these two lines with 

(connection.nodes & path.nodes).any?

lib/engine/connection.rb
```diff
@@ -21,16 +19,21 @@ def self.connect!(hex)
...
           next if connection.paths.any? { |p| p.hex == hex && (p.exits & path.exits).any? }
 
           connections[connection] = true
           connection.add_path(path)
 
-          path.exits.each do |edge|
-            hex_connections = hex.connections[edge]
+          if path.exits.none?
```
tobymao, line 29: empty?, none? would be if you expect nils inside

lib/engine/connection.rb
```diff
@@ -94,8 +96,7 @@ def clear_cache
     end
 
     def nodes
-      # FIXME: for intra-tile paths
-      @nodes ||= @paths.map { |p| p.nodes[0] }.compact
+      @nodes ||= @paths.map(&:nodes).flatten
```
tobymao, line 99: @paths.flat_map(&:nodes)

lib/engine/connection.rb
```diff
@@ -7,12 +7,10 @@ class Connection
     def self.connect!(hex)
       connections = {}
 
-      # FIXME: for intra-node paths
-      node_paths, paths = hex.tile.paths.partition { |p| p.nodes[0] }
+      node_paths, paths = hex.tile.paths.partition { |p| p.nodes.any? }
```
tobymao, line 10: p.nodes.any? is called everywhere,

let's add a helper function in path 

def node?
  @_node ||= @nodes.any?

lib/engine/connection.rb
roseundy: OK. Still learning the array/hash methods.

lib/engine/connection.rb
roseundy: OK

lib/engine/connection.rb
roseundy: OK

roseundy:
Fixed the rookie mistakes.

lib/engine/connection.rb
roseundy: Done.

tobymao (approved):

PR #1591, opened 2020-09-05
Don't allow obsolete trains in pool (fixes #1438)

### LINKED ISSUE #1438: 1846 discarded green trains are not removed from back pool on grey train
B&O discarded a 3/5 train due to train limit decreasing. In grey phase this train is still in the bank pool preventing anyone from emergency buying a 6T.  Since the rules enforce buying the cheapest available train (the 3/5). The game does let you buy the obsolete 3/5.

https://gist.github.com/adamranderson/cd6b79fd6cbda4fc8ad57904004d6b68

dfannius:
Since this fix just adds another condition for rusting, it does generate a log message of the form `-- Event: 3/5 trains rust --`, even though the only 3/5 trains that have rusted are the ones in the pool. If this is too confusing I can break it out into separate check that doesn't trigger that event notice (or triggers a different one).

### DIFF
--- a/lib/engine/phase.rb
+++ b/lib/engine/phase.rb
@@ -10,6 +10,7 @@ module Engine
       @index = 0
       @phases = phases
       @game = game
+      @depot = @game.depot
       @log = @game.log
       setup_phase!
     end
@@ -100,7 +101,8 @@ module Engine
       end
 
       @game.trains.each do |t|
-        next if t.rusted || t.rusts_on != train.sym
+        next if t.rusted
+        next unless t.rusts_on == train.sym || (t.obsolete_on == train.sym && @depot.discarded.include?(t))
 
         rusted_trains << t.name
         entity.rusted_self = true if entity && entity == t.owner
--- a/lib/engine/phase.rb
+++ b/lib/engine/phase.rb
@@ -102,7 +102,9 @@ module Engine
 
       @game.trains.each do |t|
         next if t.rusted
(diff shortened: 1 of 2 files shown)


lib/engine/phase.rb
```diff
@@ -100,7 +101,8 @@ def rust_trains!(train, entity)
       end
 
       @game.trains.each do |t|
-        next if t.rusted || t.rusts_on != train.sym
+        next if t.rusted
+        next unless t.rusts_on == train.sym || (t.obsolete_on == train.sym && @depot.discarded.include?(t))
```
tobymao, line 105: don't use unless with multiple statements, change this to a next if

dfannius: If I change it to
```
next if !(t.rusts_on == train.sym || (t.obsolete_on == train.sym && @depot.discarded.include?(t)))
```
then rubocop complains.

I could DeMorgan it into
```
 next if t.rusts_on != train.sym && !(t.obsolete_on == train.sym && @depot.discarded.include?(t))
```
but I think that's pretty ugly and actually obfuscates the meaning, which is "skip the following code unless the train should be rusted".

I would be fine with
```
 should_rust = t.rusts_on == train.sym || (t.obsolete_on == train.sym && @depot.discarded.include?(t))
 next unless should_rust
```

But it's your call.

tobymao (approved):

PR #1597, opened 2020-09-06
18MEX: Add more tiles - Mexico City tiles remains.

perwestling:


perwestling:
Current tile manifest

image

tobymao (approved):

PR #1605, opened 2020-09-06
18MEX: Corrections for tiles and map

perwestling:
Add missing tiles. All tiles are now present.

Correct orientation and count.

Add labels for all tiles and hexes where this should be.

The brown Mexico City tiles are still a problem as there is no
support for having separate tracks in the same location.


perwestling:
image

Note, 485P and 486P are the same tile, but all P tiles are used together with a corresponding MC tile.

PR #1612, opened 2020-09-06
Don't barf on nil stock market cells

dfannius:


dfannius:
Was causing a nil reference error in the 1836jr market.

PR #1613, opened 2020-09-07
Add config option for only allowing the cheapest depot train on EMR

### LINKED ISSUE #1497: 1846 forced train purchase didn't allow buy of a 7/8
During OR4.1 of game 7051 BO had to force buy a train, but after selling the minimum shares required for the 6T (800$) it wouldn't allow the sale of more shares to buy the 7/8 (900$), namely the NYC share that should have been possible, of even up to 2*BOs since that wouldn't change the presidency

michaeljb:
Once the President is contributing cash in 1846, any train can be bought

[Fixes #1497]


michaeljb:
Just realized my fix probably only works if the president has the cash straight up, but the president is allowed to sell more shares to raise the needed money for a 7/8 ($900) even if they have or can reach the $800 needed for a 6.

PR #1632, opened 2020-09-08
Welcome message grammar

dfannius:


perwestling (approved):
Ah, the pit falls of not being native English speaking....

Issue #1469, opened 2020-08-18
1846 - buying President's share from market disappears a share from the game


michaeljb:
https://www.18xx.games/game/6822?action=658 (copied to a [gist here](https://gist.github.com/michaeljb/7ac19622f73dfc9fb76af817151360ef))

DrAwesome buys a second share from the market, becoming the president of the IC, which had been in receivership. There should be one share left in the market, but that share has apparently disappeared; the game reports 4 in the treasury, 2 held by DrAwesome, and 1 each held by 18xxperson, Mark894, and adam, for a total of 9 shares.

On the IC's turn, they were presented the option to redeem a share, but could not actually redeem it, and could not skip that step without hitting a GameError

crericha:
Okay, so the game lets DrAwesome buy the president share outright, paying 2x the market price; however, it only shows him owning 20% of the company, when it should be 30%. His cert count of 8, confirms the game recognizes him as the owner of the IC. So my guess is the game tried to exchange his 10% share with the 20% in the market, but instead the 10% share got lost someplace.

I think I see two issues here:
1. The 20% share should not be purchasable outright from the market. DrAwesome should only be paying for a single share, not a double share.
2. The logic to exchange the the single share for the double share has a bug where the 10% share doesn't end up in the market.

crericha:
The share class knows which share is the president share, so my thinking is there should be logic around dealing with purchasing the president share from the market.

1. If you own no shares of the corporation, then you should not be able to buy the presidents share.
2. If you own 1 share less than the amount it takes to own the corporation (assuming logic to change presidency is correct), then you should pay 1x market price and exchange you additional shares for the presidents share. I'm being intentionally generic here as games like 1880 have 30 or 40 percent presidents shares.

PR #1637, opened 2020-09-10
Fix buying president share from market

crericha:
Fix for #1469. First verify the player has a share and is thus able to exchange for the president's share. Then update the president's share bundle to a 10% partial when being bought out of the market, so the exchange logic works correctly. This fix also removes the old logic that didn't work correctly.


lib/engine/share_pool.rb
```diff
@@ -160,6 +160,10 @@ def transfer_shares(bundle, to_entity, spender: nil, receiver: nil, price: nil)
       num_shares.times { move_share(shares_of(corporation).first, owner) }
     end
 
+    def companies
```
tobymao, line 163: what is this for?

lib/engine/step/g_1846/buy_sell_par_shares.rb
```diff
@@ -7,15 +7,31 @@ module Step
...
+          bundle = bundle.to_bundle
+          corporation = bundle.corporation
+
+          if corporation.receivership? && bundle.presidents_share
+            return false if entity.num_shares_of(corporation).zero?
+
+            bundle = ShareBundle.new(bundle.shares, 10)
+          end
+
+          super(entity, bundle)
```
tobymao, line 21: hmm interesting since you mutate bundle here, 

i'm not sure, but if you just call super, (without the args) will it do the same thing?

lib/engine/step/g_1846/buy_sell_par_shares.rb
```diff
@@ -7,15 +7,31 @@ module Step
...
+        end
+
+        def process_buy_shares(action)
+          bundle = action.bundle
```
tobymao, line 25: return unless (bundle = action.bundle)

rather than all the safe operators

lib/engine/step/g_1846/buy_sell_par_shares.rb
```diff
@@ -7,15 +7,31 @@ module Step
...
+
+          if bundle&.corporation&.receivership? && bundle.presidents_share
+            action = Action::BuyShares.new(action.entity,
+                                           shares: bundle.shares,
+                                           share_price: bundle.share_price,
+                                           percent: 10)
           end
 
-          super
+          super(action)
```
tobymao, line 34: again i wonder if can just call super here

lib/engine/share_pool.rb
crericha: I believe it is an artifact from Michael's start of a fix for the bug. Its not used anywhere in the fix, so will remove.

lib/engine/step/g_1846/buy_sell_par_shares.rb
crericha: In stock Ruby, super with no arguments works great; however, I confirmed I need to explicitly pass the parameters here. Perhaps a limitation/bug with Opal?

tobymao:
sounds good we can keep it then

lib/engine/step/g_1846/buy_sell_par_shares.rb
crericha: That's what I initially had; however, it prevents the parent from being called, which adds the game error to be displayed to the player. Two Options come to mind.

1) Handle the error message here. This duplicates functionality handled later.
@game.game_error("Cannot buy a share of #{shares&.corporation&.name}") unless (bundle = action.bundle)

2) Call super early and let the parent handle the error.
return super unless (bundle = action.bundle)

tobymao: ah you're right. let's do number 2

crericha: Option 2 implemented.

lib/engine/share_pool.rb
michaeljb: Yeah I had added it in lib/engine/g_1846/share_pool.rb, I feel like something was breaking in ORs if a corporation was in receivership or something but I don't remember exactly. From a little bit of local testing taking out the function doesn't seem to break anything

tobymao (approved):

PR #1650, opened 2020-09-11
1846 - fix buying trains w/ emergency issuing of shares

### LINKED ISSUE #1623: [1846] Train buying restrictions after issuing shares
As of #1621, I believe 18xx.games now allows all legal train-buying behavior in 1846, but it also allows some illegal behavior. These rules are not currently enforced:

* after issuing shares, a corporation may not buy a train from another corporation - [sample.json](https://gist.github.com/michaeljb/403326d577495ef5891b9362cb52531c)
* if the desired train can be afforded by issuing 1 share, the corporation may not issue 2 shares to buy it (the current UI should only allow issuing more shares than necessary when doing so enables buying the more expensive train variant, in which case that is the only variant that should be allowed for purchase) - [sample.json](https://gist.github.com/michaeljb/403326d577495ef5891b9362cb52531c)
* the president cannot contribute cash before shares are issued - [sample.json](https://gist.github.com/michaeljb/4dab2cf831545f865ceab6b4dedb0fec)

~I expect~ This will require pinning when fixed.

michaeljb:
Also 1846 and 1889 selling in blocks during EMR

- prevent president from selling shares if the corporation can and must issue
  shares first

- prevent corporation from buying a train from another corporation after
  emergency issuing

- prevent corporation from buying a train after emergency issuing more shares
  than was necessary to do so

- if the corporation could have emergency issued 1 or 2 shares, and issued 2,
  @last_share_issued_price is the difference of the bundles' totals; in other
  words, how much *more* money was raised by issuing N shares rather than issuing
  N-1 shares; because of 1846's weird emergency issuing rules, that amount does
  not equal the actual price of one share

- min_depot_price uses safe operators because when 18AL gets to the 4Ds,
  min_depot_price returns nil (I don't understand what's going on here)

- remove spec game 4949; president contributed cash to a train purchase before
  the corporation issued shares

[Fixes #1623]

I expect this to break many 1846 games--players buying trains w/ president cash when they should have issued shares first--and some 1889 games--`MUST_SELL_IN_BLOCKS` was not enforced during EMR


lib/engine/depot.rb
```diff
@@ -47,7 +47,7 @@ def min_depot_train
     end
 
     def min_depot_price
-      min_depot_train.variants.map { |_, v| v[:price] }.min
+      min_depot_train&.variants&.map { |_, v| v[:price] }&.min
```
tobymao, line 50: maybe just do return unless min_depot_tarin

lib/engine/step/base.rb
```diff
@@ -90,6 +90,10 @@ def help
         ''
       end
 
+      def last_share_issued_price
```
tobymao, line 93: is this really relevant for other steps?

lib/engine/step/train.rb
```diff
@@ -86,39 +93,51 @@ def available_cash(player)
...
-            # 18Chesapeake and most others, it's legal to buy trains from other corps until
-            # if the player has just sold a share they can buy a train between cash-price_last_share_sold and cash
-            # e.g. If you had $40 cash, and if the train costs $100 and you've sold a share for $80,
-            # you now have $120 cash the $100 train should still be available to buy
-            min_cash_to_spend = (entity.cash + entity.owner.cash) - @last_share_sold_price
-            return depot_trains + (other_trains.reject { |t| t.price < min_cash_to_spend })
+            other_trains =
+              if @game.class::EBUY_OTHER_VALUE
+                must_spend = (entity.cash + entity.owner.cash) - @last_share_sold_price + 1
+                other_trains.reject { |t| t.price < must_spend }
```
tobymao, line 104: instead of the if else and other_equals you could just do other_trains.reject!

lib/engine/step/train.rb
```diff
@@ -86,39 +93,51 @@ def available_cash(player)
...
+              else
+                []
+              end
           end
         end
-        depot_trains + other_trains
+
+        depot_trains.concat(other_trains)
```
tobymao, line 111: hm ya since you return it here, i'd probably keep it as a +, if you were doing more stuff after this with depot_trains, yea concat it

lib/engine/step/train.rb
michaeljb: How about this? https://github.com/tobymao/18xx/pull/1650/commits/4c8a9dcc9d65eeb9422b66acb476d605a007bc28#diff-97e9f18fd60923583d013e4cd042366c

PR #1652, opened 2020-09-12
18MS: Modify Atlanta off-board hexes

perwestling:
Use the track restriction instead of specific code, to stop
multiple visits to Atlanta in a run.

Before:

image

After:

image

And here is how parts of it looks like on the physical map:
image


tobymao:
none of that is necessary, you can just use groups if you want

if you assign the red offboard hex to a group, you can't start and end there, do you want to do that instead?

scottredracecar:
No, the players don't have a way to understand that these are part of the group. This design more closely matches the physical game/rules

tobymao (approved):

PR #1653, opened 2020-09-12
1846 - prevent incorrect C&WI tokening, remove reservation when sold

### LINKED ISSUE #1565: 1846 C&WI token placement
In game 7893, a game of 1846, B&O purchased the private company C&WI, then placed the token in Chicago. However, they placed it in a spot other than the reserved C&WI slot. As far as I can tell, the token MUST be placed in the reserved spot. The C&WI card reads ""Reserves a token slot in Chicago (D6), in which the owning corporation may place an extra token at no cost." 

I do not see anywhere in the rules that this allows for placement of the token anywhere other than this reserved spot. Am I wrong, or is this just a mistake/oversight? 

https://www.18xx.games/game/7893?action=54

Annotation 2020-09-01 102458

michaeljb:
- change `Abilities#remove_ability_when` to use `Ability#remove` instead of
  `Ability#when`; use this for slightly DRYer code in
  `Engine::Phase#trigger_events!`

- add `city` property to the `Ability::Token` so that C&WI can be restricted to
  a particular city on the Chi hex

- add `"remove": "sold"` to C&WI's reservation, as other corporations are
  allowed to token there once it has been bought by a corporation

- change tile config for green, brown, gray Chi tiles so that the city indexes
  are consistent across all of them, matching the yellow preprinted tile; this
  breaks some of the specs, and I expect will break a lot of existing games

- `spec/fixtures/1846/3099.json` is broken by the above change but I've kept it
  because it has some UI tests in `spec/assets_spec.rb` that are still valid
  before it reaches the broken state

- add some unit tests for all of the C&WI reservation/token placing behavior

[Fixes #1565]


tobymao:
does this break every 1846 game?

tobymao (approved):

michaeljb:
Just about...don't think it'll break any games where Chicago was only tokened in yellow :P

tobymao:
can you fix the merge conflict?

PR #1656, opened 2020-09-13
1846 - better hiding of stuff during draft

### LINKED ISSUE #1474: 1846 Initial Auction: [Show Player] button for final drafted card shows all players
GameID: 6949
In the initial draft of 1846, the final card is flipped face up and players can click [Show Player] to see what they've already drafted.  This allows them to determine if they want to buy the Mail Contract.

However, once a player has clicked [Show Player] on their turn, then all players after them are shown with their respective holdings.  If the player picks first, they can then see all the holdings of others until the private is picked.

While the advantage gained from this is minor at best, as very few mail contracts would be passed completely around and player holdings are shown once it is bought anyway, it's still something.  Maybe a "nice to have" feature?

### LINKED ISSUE #1585: [1846] During initial draft other players hands and choices can be seen
In the initial draft of 1846 - when the “Show player” or “Show companies” button is used during my turn - the player info and company offer is correctly shown. But if the history arrows are used with my info shown, other players info or company offers can be seen - so its possible to see the whole draft history.

michaeljb:
- Add new method `hidden?` to return true if `@block_show` is true; this fixes
  hiding breaking when using the history feature [Fixes #1585]

- Add attr `pre_process_pass` to steps so that active step can have some
  behavior set; other buttons created within the Auction view update
  `store(:hidden)` before they carry on with any action processing, but the Pass
  button is created outside of that, so the Pass button cannot update anything
  in the Auction view; the Auction view sets a lambda that updates the store and
  then cleans itself up. This fixes the issue where in 1846, if the draft is
  down to one company, a player can click "Show Player" and then pass, and then
  they are shown the info about the next player. [Fixes #1474]

### DIFF
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -41,7 +41,7 @@ module View
         end
 
         def render_players
-          return nil if !@step.players_visible? && @hidden
+          return nil if !@step.players_visible? && hidden?
 
           if @step.players_visible?
             h(Players, game: @game)
@@ -68,6 +68,13 @@ module View
             if @block_show
               store(:flash_opts, 'Enter master mode to reveal other hand. Use this feature fairly.')
             else
+              if @current_actions.include?('pass')
+                @step.before_process_pass = lambda do
+                  store(:hidden, true, skip: true)
+                  @step.before_process_pass = -> {}
+                end
+              end
+
               store(:hidden, !@hidden)
             end
           end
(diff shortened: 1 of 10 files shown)


assets/app/view/game/round/auction.rb
```diff
@@ -68,6 +68,13 @@ def render_show_button
             if @block_show
               store(:flash_opts, 'Enter master mode to reveal other hand. Use this feature fairly.')
             else
+              if @current_actions.include?('pass')
+                @step.before_process_pass = lambda do
```
tobymao, line 72: hm i'm not sure about this. is there another way? it seems weird for the client to mutate the engine like this

michaeljb: Yeah, this is definitely super weird. Typing up a comment got me thinking more about this, and I had an idea for keeping this `before_process_pass` stuff only under `assets/app/view/` instead of in the engine -- https://github.com/tobymao/18xx/pull/1656/commits/7ee232ade161c56a9bb6083155afe33a6c8e33dc

tobymao (approved):

PR #1659, opened 2020-09-13
Improvements for trains buy

perwestling:
Introduce a possibility to sell more shares than needed during emergency buy.
Only title that supports this at the moment are 18MS.

Refactor emergency buy so that check for sellable bundles can be
used by sub classes of train buy.

Bug fix: Do not enforce purchase of cheapest train if game title
(1846, 18MS) allows buy of more expensive train during emergency buy.

Bug fix 18MS: Do not allow player loans to purchase train if there
still are shares to sell. This stops the bugs there game made it possible to
take loans to buy a non-emergency train via president assist.

Sort of bug fix: Avoid doing emergency buy in case there is an affordable cheaper train.

### DIFF
--- a/lib/engine/step/g_18_ms/buy_train.rb
+++ b/lib/engine/step/g_18_ms/buy_train.rb
@@ -23,12 +23,17 @@ module Engine
           end
 
           emergency_buy_with_loan = false
+          available_funds = entity.cash + player.cash
 
-          if price > entity.cash + player.cash && must_buy_train?(entity)
+          if price > available_funds && must_buy_train?(entity)
 
             if can_sell_anything?(player, entity)
               @game.game_error("#{player.name} may not loan money as long as #{entity.name} has shares to sell")
             end
+
+            cheapest = @game.depot.min_depot_train
+            @game.game_error("#{player.name} may not loan money when affordable trains exist") if
+              cheapest.price <= available_funds
 
             # Prepare to take a loan
             emergency_buy_with_loan = true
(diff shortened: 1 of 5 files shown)


lib/engine/step/g_18_ms/buy_train.rb
```diff
@@ -57,6 +58,18 @@ def process_buy_train(action)
           @log << "An extra #{@game.format_currency(interest)} is added to the debt"
           pass!
         end
+
+        private
+
+        def can_sell_anything?(player)
+          bundles = []
```
tobymao, line 65: this is unnneded

lib/engine/step/g_18_ms/buy_train.rb
```diff
@@ -57,6 +58,18 @@ def process_buy_train(action)
...
+          @game.corporations.each do |c|
+            bundles = player.bundles_for_corporation(c)
```
tobymao, line 67: player.bundles_for_corporation(c).any ? { bundle| sellable_bundle?(bundle) }

lib/engine/step/g_18_ms/buy_train.rb
perwestling: Ok

lib/engine/step/g_18_ms/buy_train.rb
perwestling: Ok

perwestling:
@tobymao Have fixed review comments.
Also found a bug, where corporation + player had more money than 6 train and 2D, but less than 4D/5D. It was then possible to buy 4D/5D and take loan. 
After correction an error message appear.

tobymao (approved):

PR #1666, opened 2020-09-14
Add parallel path support (lanes) to rendering and routing

roseundy:
These changes are needed for the Mexico City tile in 18MEX

### DIFF
--- a/assets/app/view/game/part/track_node_path.rb
+++ b/assets/app/view/game/part/track_node_path.rb
@@ -335,7 +335,7 @@ module View
             @end_y
           ) if @need_arc
 
-          return unless @path.parallel? && @begin_edge
+          return if !@path.parallel? || !@begin_edge
 
           shift = (@path.lane_index * 2 - @path.lanes + 1) * (@width + PARALLEL_SPACING[@path.lanes - 2]) / 2.0
           delta_x = (shift * Math.cos(@begin_edge * 60.0 * Math::PI / 180.0)).round(2)
--- a/assets/app/view/game/part/track_node_path.rb
+++ b/assets/app/view/game/part/track_node_path.rb
@@ -15,6 +15,8 @@ module View
         needs :width, default: 8
         needs :dash, default: '0'
 
+        PARALLEL_SPACING = [8, 6, 4].freeze
+
         EDGE_PERP_ANGLES = [90, 30, -30, -90, -150, 150].freeze
 
         EDGE_REGIONS = {
@@ -332,6 +334,17 @@ module View
             @end_x,
             @end_y
           ) if @need_arc
+
+          return unless @path.parallel? && @begin_edge
+
+          shift = (@path.lane_index * 2 - @path.lanes + 1) * (@width + PARALLEL_SPACING[@path.lanes - 2]) / 2.0
+          delta_x = (shift * Math.cos(@begin_edge * 60.0 * Math::PI / 180.0)).round(2)
+          delta_y = (shift * Math.sin(@begin_edge * 60.0 * Math::PI / 180.0)).round(2)
+
+          @begin_x += delta_x
+          @begin_y += delta_y
+          @end_x += delta_x
+          @end_y += delta_y
         end
 
         def preferred_render_locations
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -5,9 +5,9 @@ require_relative 'base'
 module Engine
   module Part
     class Path < Base
-      attr_reader :a, :b, :branches, :city, :edges, :junction, :nodes,
-                  :offboard, :stops, :terminal, :town
-      def initialize(a, b, terminal: nil)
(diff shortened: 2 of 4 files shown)


assets/app/view/game/part/track_node_path.rb
```diff
@@ -332,6 +334,17 @@ def load_from_tile
             @end_x,
             @end_y
           ) if @need_arc
+
+          return unless @path.parallel? && @begin_edge
```
tobymao, line 338: don't use unless with multiple

return if !parallel || !begin_edge

lib/engine/route.rb
```diff
@@ -235,11 +235,13 @@ def connection_hexes
...
 
-    def find_connections(connections_a, connections_b)
+    def find_connections(connections_a, connections_b, other_paths)
       connections_a.each do |a|
+        next if (a.paths & other_paths).any?
+
         connections_b.each do |b|
           middle = (a.nodes & b.nodes)
-          next if middle.empty?
+          next if middle.empty? || (b.paths & other_paths).any? || (b.paths & a.paths).any?
```
tobymao, line 244: maybe we should move these & other_paths to a select+ that way it's N + M as opposed to N X M?

lib/engine/tile.rb
```diff
@@ -63,10 +63,14 @@ def self.from_code(name, color, code, **opts)
     def self.part(type, params, cache)
       case type
       when 'path'
+        num_lanes = 1
```
tobymao, line 66: i don't think this is necessary, you could just check if it's nil, because default is 1

lib/engine/part/path.rb
```diff
@@ -5,16 +5,18 @@
...
+      def initialize(a, b, terminal: nil, lanes: nil, lane_index: 0)
         @a = a
         @b = b
         @edges = []
         @branches = []
         @stops = []
         @nodes = []
         @terminal = !!terminal
+        @lanes = lanes || 1
+        @lane_index = lane_index
```
tobymao, line 19: i wonder if we can make lane_index inferred, it sucks that you have to pass @lanes and @lane_index

perhaps if parallel? you can just return the index of the particular path as a helper method

lib/engine/route.rb
```diff
@@ -44,7 +44,7 @@ def next_connection(node, connection, other)
     end
 
     def select(node, other)
-      other_paths = @routes.reject { |r| r == self }.flat_map(&:paths)
+      other_paths = @routes.flat_map(&:paths)
```
tobymao, line 47: so why is this one changed?

assets/app/view/game/part/track_node_path.rb
roseundy: Funny - I originally had this as a if block (w/o return) and rubocop wanted me to change it to be a "return unless", but "return if" works.

lib/engine/route.rb
roseundy: Can you clarify? I don't understand.

lib/engine/tile.rb
roseundy: I needed an integer to iterate over later.

lib/engine/part/path.rb
roseundy: I don't understand your solution. We need two attributes for a path: how many lanes in parallel and what is the position within those lanes, so both @lanes and @lane_index need to be stored with a path.

lib/engine/route.rb
roseundy: Because otherwise clicking on a node can re-use a path on the current route. This was the first issue I encountered when I prototyped parallel paths.

lib/engine/route.rb
roseundy: I could build a new version of other_paths inside the outer loop that is other_paths + a.paths. Is that what you mean?

tobymao: well since you do a next if b.paths & other_paths.any? in this inner loop, you could instead do connections_b.select(...) outside and use that instead of connections_b

lib/engine/part/path.rb
tobymao: can't you infer the lane_index based on the position in the array?

lib/engine/tile.rb
tobymao: you could just do

(num_lanes || 1).map do |...|

lib/engine/route.rb
roseundy: Gotcha

lib/engine/tile.rb
roseundy: OK. My Ruby newbie is showing again.

lib/engine/part/path.rb
roseundy: Which array? There are potentially many paths on a tile, not all of them part of parallel paths.

tobymao (approved):

PR #1673, opened 2020-09-16
18MS: Set to alpha

perwestling:
Updated welcome page to include 18MS.
Sorted the mentioned 18xx in alphabetical order.


perwestling:
@scottredracecar - should we press the button?

perwestling:
It looks like we have a go!
So you can merge and deploy this whenever it is appropriate, @tobymao

PR #1690, opened 2020-09-18
Change lane attributes to be per exit rather than per path

roseundy:
Update 18MEX tileset.
Refactored code in Path and track_node_path to use per exit (actually "endpoint" - a or b) lane width and index attributes.
Updated Path#walk to only follow paths across tiles if they have compatible lane attributes.

New MC/P tiles for 18MEX:
18MEX_MC_P_Split_Track

### DIFF
--- a/assets/app/view/game/part/track_node_path.rb
+++ b/assets/app/view/game/part/track_node_path.rb
@@ -255,7 +255,7 @@ module View
             @end_edge = @path.edges.last.num
             @end_x = edge_x_pos(@end_edge, 87)
             @end_y = edge_y_pos(@end_edge, 87)
-            lanes = @path.ab_lanes
+            lanes = @path.lanes
           elsif @num_exits == 1
             @begin_edge = @path.exits[0]
             @begin_x = edge_x_pos(@begin_edge, 87)
@@ -279,11 +279,8 @@ module View
                                  ]
                                end
             end
-            lanes = if @path.a.edge?
-                      @path.ab_lanes
-                    else
-                      [@path.ab_lanes.last, @path.ab_lanes.first]
-                    end
+            lanes = @path.lanes
+            lanes.reverse! if @path.b.edge?
           else
             # city/town - city/town
             @ct_edge0 = @tile.preferred_city_town_edges[@stop0] if @stop0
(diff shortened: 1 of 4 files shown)


assets/app/view/game/part/track_node_path.rb
```diff
@@ -278,6 +279,11 @@ def load_from_tile
                                  ]
                                end
             end
+            lanes = if @path.a.edge?
+                      @path.ab_lanes
+                    else
+                      [@path.ab_lanes.last, @path.ab_lanes.first]
```
tobymao, line 285: could just be 

lanes = @path.ab_lanes
lanes.reverse! if @path.a.edge?

tobymao: same with all the other ones

assets/app/view/game/part/track_node_path.rb
```diff
@@ -335,16 +351,30 @@ def load_from_tile
...
-          delta_x = (shift * Math.cos(@begin_edge * 60.0 * Math::PI / 180.0)).round(2)
-          delta_y = (shift * Math.sin(@begin_edge * 60.0 * Math::PI / 180.0)).round(2)
+          begin_shift_edge = @begin_edge || @end_edge
+          end_shift_edge = @end_edge || @begin_edge
 
-          @begin_x += delta_x
-          @begin_y += delta_y
-          @end_x += delta_x
-          @end_y += delta_y
+          if lanes.first[0] > 1
```
tobymao, line 359: make lanes.first a var or something

first_lane = lanes[0]

tobymao: or since you use both

first, last = lanes

lib/engine/part/path.rb
```diff
@@ -5,18 +5,43 @@
 module Engine
   module Part
     class Path < Base
-      attr_reader :a, :b, :branches, :city, :edges, :junction, :lanes, :lane_index,
+      attr_reader :a, :ab_lanes, :b, :branches, :city, :edges, :exit_lanes, :junction,
```
tobymao, line 8: why don't we just call it lanes? 

why is an ab_lane and what's a lane?

assets/app/view/game/part/track_node_path.rb
roseundy: Nice. Will do.

assets/app/view/game/part/track_node_path.rb
roseundy: OK

lib/engine/part/path.rb
roseundy: OK, I can rename.

roseundy:
Here is an image showing how the lanes match up between hexes:

tobymao (approved):

Issue #1685, opened 2020-09-17
18MS - Display the Phase and Rusting Schedule


Bearcat89:
It would be nice if the Info tab would show the phase change and rusting schedule for trains.

perwestling:
This is similar to #1675 
There will be improvements done in this area shortly.

Issue #1675, opened 2020-09-16
[18MS] Display progress chart on info tab and update OR names


KillerOrca:
Game https://www.18xx.games/game/9795

https://all-aboardgames.com/products/international-18ms-the-railroads-come-to-mississippi

As found on the third image the round of play tracker should be added to the info tab.

The OR description also went from OR 1.1 to OR 1.2 instead of OR 1 to OR 2.

perwestling:
There will always be differenes between the looks of physical games and how it is presented online.
Now the game is run for 10 ORs, 5 OR sets, and I took the design decision to use the existing presentation of OR1.1, OR1.2, OR2.1, OR2.2, that is first OR of first OR set, second OR of first OR set etc.
This was also described in the wiki page: https://github.com/tobymao/18xx/wiki/18MS#rules-highlights

It is of course possible to change this to using OR 1, OR 2, OR 3, ... and I might give it a new try. But I thought it most important to test it in alpha to see the reactions.

When it comes to play tracker this is something that would require a lot of specialized implementation. And as far as I know it would only be used in 18MS and 18CZ. Question is if is worth the effort; how useful would it be?

Mark Derrick had another suggestion, and that was to show the last OR more clearly in the log.
That might be a good improvement which do not require much implementation.

I did have an idea how to add a time line, but that was mostly to show the events that occur during the ORs. Now this is far from how it appear in the physical game, and I do not know if it will be clear enough for the players?  Here is a link to my suggestion: https://github.com/tobymao/18xx/issues/1507

Any more thoughts on this subject, @KillerOrca ?

TIE53:
I think it needs some sort of timeline/round marker.  As far as I can tell, there is nothing in the current implementation that displays the train rusting schedule, the length of the game, when phase 3 begins, etc. This is not a problem for experienced players, but when I was teaching a new player yesterday, I had to write all of this information in the chat, which was mildly burdensome. 

I don't think it necessarily needs a moving round marker, but an image/display analagous to the round marker on the physical board would be very helpful.

KillerOrca:
@perwestling My thinking was a static image of the play tracker then players could match the current round to see how for along they are to the game end. This might not be everyone's preferred version but I think as a quick reference it would work well.

Speaking of 18CZ since it also has the static rounds I think renaming to match the OR structure, in this instance for 18MS, will make it clearer for players. When it goes from OR1.1 to OR1.2 new players may think they have more time left in the game than they really do.

[...]

PR #1693, opened 2020-09-19
18MS: Improve display of OR description, and add special info

perwestling:
The game now shows OR 1 (of 10) etc instead of OR 1.1 (of 2).
The OR numbering is local to 18MS, so the change of OR presentation
is only done in 18MS, but it is possible to override the
round_name method in game base class if a title want to
modify this.

18MS also add information in the log, and game page, about the next
upcoming time based event. So e.g.
In the log:
  -- Operating Round 6 (of 10) - 3+ trains rust after OR 6 --
In the game page:
  18MS: Operating Round 6 (of 10) - 3+ trains rust after OR 6 - Lay Track
Or
  18MS: Operating Round 1 (of 10) - Change to Phase 3 after OR 1 -
  2+ trains rust after OR 4 - Buy Trains
This is done by extracting the description to the round_description
method in game base class, and overriding it in 18MS.

Stock round description has not been changed.

Draft round now says the following in the game page for 18MS:
  18MS: Draft - Draft One Company Each
It was "Draft round 1" but as there is only one round this was
confusing.

The description for all other titles should be unchanged.

Have added a "Timeline Info" to the Info tab that only display if the
list TIMELINE_INFO contains anything. If it do the Info page will just
append the elements in the list as paragraph with text, one paragraph
per element. This has only been added to 18MS in this commit but it
can easily be added other titles as needed, e.g. 18Chesapeake could add
a paragraph on how a non-parament train is exported at end of each
OR set.

Internally the game end check for specific OR has been removed. This
was only used in 18MS and it is better to implement it just in that
title, as the OR numbering is local to 18MS. It is also a game
end that is very few titles, only 18CZ as far as I know.

This commit is an attempt to handle the tickets #1675 and #1685.

### DIFF
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -27,7 +27,21 @@ module View
       def render_body
         children = upcoming_trains
         children.concat(discarded_trains) if @depot.discarded.any?
-        children.concat(phases, game_info)
+        children.concat(phases)
+        children.concat(timeline) if timeline
+        children.concat(game_info)
+      end
+
+      def timeline
+        return nil if @game.class::TIMELINE.empty?
+
+        children = [h(:h3, 'Timeline')]
+
+        @game.class::TIMELINE.each do |line|
+          children << h(:p, line)
+        end
+
+        children
       end
 
       def game_info
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -56,8 +56,6 @@ module Engine
       #  current_round - ends at the end of the current round
       #  current_or - ends at the next end of an OR
       #  full_or - ends at the next end of a complete OR set
-      # Also, you can use final_or_set: <number> to trigger game
-      # end (:full_or) when that OR is reached.
       GAME_END_CHECK = { bankrupt: :immediate, bank: :full_or }.freeze
 
       BANKRUPTCY_ALLOWED = true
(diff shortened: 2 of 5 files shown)


perwestling:
Here is examples on how it looks like in the log:

image

image
image

image

And here is part of the Info page:
image

assets/app/view/game/game_info.rb
```diff
@@ -27,7 +27,21 @@ def render
       def render_body
         children = upcoming_trains
         children.concat(discarded_trains) if @depot.discarded.any?
-        children.concat(phases, game_info)
+        children.concat(phases)
+        children.concat(timeline_info) if timeline_info
```
tobymao, line 31: we could probably just call this timeline

lib/engine/game/base.rb
```diff
@@ -1042,8 +1043,7 @@ def game_end_check
           end
         end
 
-        return :final_or_set, :full_or if @round.is_a?(Round::Operating) &&
-          self.class::GAME_END_CHECK[:final_or_set]&.to_i == turn
+        nil
```
tobymao, line 1046: why do you need to return nil here

assets/app/view/game/game_info.rb
perwestling: Ok, I will change that.

lib/engine/game/base.rb
perwestling: The reason is that the construction that ends on line 1043 is not nil. When line 1045 was removed game_end_check breaks as any non-nil value will trigger end game. With 1045 left in it either returned something or resulted in nil.

perwestling:
@tobymao Do you have any further comments?  What do you think of the
"Operating Round 3 (of 10)"?

tobymao:
ok please push latest and merge, i think the text is fine

PR #1695, opened 2020-09-20
Streamline issue redeem shares [Fixes #1433]

### LINKED ISSUE #1433: 1846 Issue/redeem Streamlining
Currently it’s a pain point having people use multiple undos to get back to the start. I know it’s not technically allowed but the rules suggest allowing due to the pain of
Undoing multiple times for a company because you're short or have spare cash at the end of the OR.

Either have a return to start of company turn button or have a button to allow issue/redeem Through out - obviously modifying payouts and cash auto.

crericha:


crericha:
Note: This PR will break current 1846 games.

Migrating the rspec JSON to remove the "pass" for the issue/redeem step was anything but trivial. I probably put in as many hours on that as I did the fix. It is not worth pursuing for existing games.

lib/engine/round/g_1846/operating.rb
```diff
@@ -43,6 +43,12 @@ def after_process(action)
             end
           end
 
+          dividend_step = steps.find { |step| step.class == Step::G1846::Dividend }
```
tobymao, line 46: maybe this should be

unless current_step.is_a?(Steps::1846::Dividend)
steps.find { ... is_a? }.pass!

assets/app/view/game/issue_shares.rb
```diff
@@ -10,18 +10,18 @@ class IssueShares < Snabberb::Component
...
         @step = @game.round.active_step
         @entity = @game.current_entity
-
+        @current_actions = @game.round.actions_for(@entity)
         children = []
 
-        if @step.current_actions.include?('sell_shares')
-          children << render_shares('Issue', @step.issuable_shares(@entity), Engine::Action::SellShares)
+        if @current_actions.include?('sell_shares')
+          children << render_shares('Issue', @game.issuable_shares(@entity), Engine::Action::SellShares)
```
tobymao, line 17: why does this need to be in game, can't you just do this directly here?

round.step_for(entity, 'sell_shares').issuable_shares

lib/engine/round/g_1846/operating.rb
crericha: The dividend step doesn't happen if no trains run. Will after_process() still be called in that case? When does current_step change to the next step?

assets/app/view/game/issue_shares.rb
crericha: When we discussed in slack, you mentioned the client code shouldn't have to know much detail about the server code. In this case, should the client have to know that issuable_shares is step dependent? I leaned towards no and abstracted via the game class; however, both approaches are fine by me. Let me know which you prefer.

lib/engine/round/g_1846/operating.rb
crericha: Confirmed the dividend step can be non-blocking, so it doesn't always become the current_step.

assets/app/view/game/issue_shares.rb
tobymao: sounds fine we'll keep

lib/engine/round/g_1846/operating.rb
tobymao: ok then let's just do

unless steps.find { |step| step.is_a?(Step::) }.active?

tobymao (approved):

lib/engine/round/g_1846/operating.rb
crericha: Fixed.

PR #1697, opened 2020-09-20
18GA: Avoid abilities used out of phase and without owning (fixes #1681)

### LINKED ISSUE #1681: 18GA - Private abilities usable by non-owning companies
https://www.18xx.games/game/9231?action=187

W&A company was able to use the Midland and Waycross private abilities during SAL's turn.  Likely because they are both owned by the same player, but the private abilities state:
```
 (But it still must be laid during the tile-laying step of the corporation's turn...
```
```
However, it can only be done during the token-placement step of the corporation's turn...
```

perwestling:
Only show abilities during corporation turn if the corporation
owns the ability.

Only allow special token/tile during tokening and tile lay phase
respectively.

In case special tile lay ability has not yet been used, require
owning entity to pass tile lay. For tokening this is not required
as the special tokening is just to put out standard token for free.

This breaks several games, where special tile lay was used despite
corporation not owning company. One of these broken games was
used as spec test - therefor this has been swapped.


lib/engine/step/g_18_ga/track.rb
```diff
@@ -0,0 +1,35 @@
...
+          pass! unless remaining_tile_lay?(action.entity)
+        end
+
+        private
+
+        def remaining_tile_lay?(entity)
+          return true if can_lay_tile?(entity)
+          return false unless @game.p2_company.owner == entity
+
+          ability = @game.p2_company.abilities(:tile_lay)
```
tobymao, line 29: this can all be done inline

Issue #686, opened 2020-06-12
1817 Special Rules Summary


jenf:
### Map
 - [x] Some tiles are declared as water terrain when they are not water terrain causing problems with Bridge companies

### Initial auction

- [x] Choose and bid, with seed capital
- [x] PD should be player to the left of the last player to initiate a bid

### Market
- [x] Needs a liquidation and three acquisition spots

### IPO
- [x] When IPOing a company first place station location players choice
- [x] Then a bidding for that IPO
- [x] The winning bidder chooses if the company is 2, 5 or 10 shares depending on the phase
- [x] Currently a bug, once you've passed on the IPO bid you can rejoin
- [x] Currently a bug, if all other players pass, you still need to up the bid
- [x] Currently a bug, if you try to PAR a company without any privates it explodes.

### SR
- [x] Players can act as the corporations they own to take loans
- [x] Players can act as the corporations to purchase their own shares from the open market
- [x] Corps must buy their tokens before end of stock round.
- [x] Players can buy shorts on 5/10 share companies
- [x] New Companies need to buy their tokens
- [x] 2 Stock companies should not change stockprice when sold out.
- [x] Stocks in the market move share price one to the left

### Short selling
- [x] Players that holding shorts that declares a dividend must paying that amount to the bank.
- [x] Cash-Crisis if cannot, effectively EMR with the no-change-of-president variant.
- [x] Player bankruptcies if cannot.

### Interest/Loans

 - [x] All Actions should see if you have buying power to cover them (after taking out loans)
- [x] OR

- [x] Acquisitions

### OR
- [x] Lay one or two track tiles (one may be an upgrade)
- [x] A station marker may not be placed in Toronto.
- [x] Pay / Half Pay / Withhold Dividend - (see rules)
- [x] Public Companies are never required to purchase a train
- [x] Therefore, no EMR for trains, but see Cash-Crisis
- [x] Pay Loan Interest
- [x] Repay Loans
- [x] Liquidate companies without trains -> move to liquidation spot
- [x] End of every Operating Round - export a train
- [x] Companies cannot buy trains from companies in acquisition

## Conversion/Merge
- [x] Convert or merge companies in white section of the
stock market, in operating order
- [x] Companies converted must buy extra station markers
- [x] Must reduce down to 8 station markers as liquidate rules

## Liquidate, acquire, sell
- [x] Liquidate companies in the red liquidation area, in
reverse order they were placed in liquidation.
[...]

PR #1700, opened 2020-09-21
Initial work on 1817 acquisition/liquidation/sales

jenf:
This isn't the full acquisition implementation but about 80% of the way there.

I've split the remaining items on https://github.com/tobymao/18xx/issues/686, these are better done separately


assets/app/view/game/round/bidder.rb
```diff
@@ -0,0 +1,47 @@
...
+require 'view/game/players'
+require 'view/game/sell_shares'
+require 'view/game/stock_market'
+require 'view/game/undo_and_pass'
+
+module View
+  module Game
+    module Round
+      module Bidder
+        def render_bid(entity, corporation)
```
tobymao, line 18: i think this would be better as a widget with needs

lib/engine/round/g_1817/acquisition.rb
```diff
@@ -5,34 +5,22 @@
 module Engine
   module Round
     module G1817
-      class Acquisition < Merger
+      class Acquisition < Round::Merger
```
tobymao, line 8: should this be a merger or an auction?

jenf: Merger fits more, since you'll have to select the company to merge with.

assets/app/view/game/round/bidder.rb
jenf: done

tobymao (approved):

Issue #1709, opened 2020-09-23
[1846] Issuing should be a blocking step when other pre-route steps would be skipped due to lack of cash


michaeljb:
If a corporation has no cash for buying privates or for laying track, its turn will skip to running routes; if it has no train, it will skip to buying a train. If a corporation has no cash but has issuable shares, it should be able to issue shares to raise funds for laying track, etc., but the site does not currently allow this.

* https://www.18xx.games/game/10481
* [10481.json](https://gist.github.com/michaeljb/a535e655ddb448e1724aeb4634c277b5)

In this game, the B&O in OR 1.2 should be able to issue shares and then place a token and/or lay track, but is unable to do so.

PatrickoftheIsles:
We abandoned a game today after experiencing this bug because it happened at a crucial moment.

https://18xx.games/game/10486

PR #1710, opened 2020-09-23
Implement 1846 2p Variant

michaeljb:
1846 2p Variant Changes:

- add "1846 2p Variant" as a separate title (per Tom's request), inheriting from
  1846; keep all logic in 1846 class so it will be easily inheritable
  later (e.g., for 18 Los Angeles)
    - name class `G1846TwoPlayerVariant` since that is more readable than
      `G18462pVariant`; I also considered `G1846Variant2p` but scrapped that so
      that the class name could be more similar to the `.title`

- list as alpha, mention it on the welcome page

- sort visible companies in view during draft

- 2p rules; see https://boardgamegeek.com/thread/1616729/draft-2-player-1846-rules-game-designer

`Game` Changes:

- add method `two_player?`, use it in 18Chesapeake code as well

- add instance var `@removals` to track removed companies and corporations

- add method `corporation_opts` so that max ownership of all corporations
  can vary by player count (1846 2p uses 70%)

- wrap constant `GAME_END_CHECK` in method `game_end_check_values` so that game
  end can vary by player count

- add support for game end "reason" `final_train`, and game end "after"
  `one_more_full_or_set`
    - add method `Depot#empty?` for checking on `final_train`
    - add `Game` instance var `@final_turn` to help trigger the game after the
      next full OR set

- add some DRAFT configuration constants--this should be re-worked so that 1846
  2p just uses a different draft class than
  `Engine::Step::G1846::DraftDistribution`

- allow multiple rules links by defining a hash

Other Changes:

- implement `Engine::Company#<==>`, sorting first by min_bid then by name

- 1846
    - fix bug with Big 4 / Michigan Southern discounts; don't discount beyond
      their debt amount
    - remove unnecessary blue hex (C19) from 1846

TODO before "beta":

- draft should be different class
    - make it easily reusable for titles such as 18CZ and 18 Los Angeles
    - selected company should be paid for immediately

- indicate that removed corporations will place a second token in the relevant
  city when upgraded to green; log message? icon? "reservation" like QLL in
  1882? need to make sure it doesn't prevent other corporations tokening in
  yellow


michaeljb:
I discovered #1709 while putting together one of the game spec fixtures added here, so it'll probably break when #1709 is done. I can fix/replace it after #1709 is done and before merging this one.

lib/engine/game/base.rb
```diff
@@ -119,6 +121,11 @@ class Base
 
       COMPANIES = [].freeze
 
+      DRAFT_HAND_SIZE = nil
```
tobymao, line 124: this is weird to be in base class

tobymao (approved):

lib/engine/game/base.rb
michaeljb: 69877369

tobymao:
you can merge whenever ready

PR #1717, opened 2020-09-23
More verbose route descriptions in log (fixes #1063)

### LINKED ISSUE #1063: Indicate mail contract bonus when running train and in log

dfannius:
It's not awesome that there is now a `revenue_str` method that parallels `revenue_for`. However, refactoring the code so that the logic all happens in one place (building up the descriptive string while calculating the revenue) is not trivial and would also lead to some wasted computation (since `revenue_for` is called many times when the correct stops have to be chosen).

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -625,6 +625,10 @@ module Engine
         stops.sum { |stop| stop.route_revenue(route.phase, route.train) }
       end
 
+      def revenue_str(route)
+        route.hexes.map(&:name).join('-')
+      end
+
       def routes_revenue(routes)
         routes.sum(&:revenue)
       end
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -210,6 +210,34 @@ module Engine
         revenue
       end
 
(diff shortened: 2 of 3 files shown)


dfannius:
I should also note that the bonuses are explained only for 1846, and other titles would have to implement descriptive text separately.

tobymao (approved):

PR #1719, opened 2020-09-24
Do not skips steps that can be executed by issuing a share. [Fixes #1709]

### LINKED ISSUE #1709: [1846] Issuing should be a blocking step when other pre-route steps would be skipped due to lack of cash

crericha:


lib/engine/game/g_1846.rb
```diff
@@ -391,6 +391,31 @@ def sellable_bundles(player, corporation)
         super
       end
 
+      def issuable_shares(entity)
+        return [] unless entity.corporation?
+        return [] unless round.steps.find { |step| step.class == Step::G1846::Dividend }.active?
+
```
crericha, line 397: I don't like inferring that the player can issue and redeem as long as the Dividend step is still active in multiple places. I think it would be clearer if it were instead to check if the IssueShares step is active; however, right now the code never skips that step and instead returns no actions if the Dividend step isn't active. Is it safe in IssueShares::actions() to call skip!? That would allow me to check for the IssueShares step being active here.

tobymao: probably better to call step.pass!

crericha:
Also, this is game breaking again as it will add new possible steps that now must be passed.

lib/engine/game/g_1846.rb
crericha: That's what I meant. :-)

Issue #1558, opened 2020-08-31
[1889] port tile illegal placement


SirYaro:
**Please provide**
* Multiplayer? Yes
https://www.18xx.games/game/7696?action=91

**Describe the bug**
TR (owner KrzysP) was able to place port tile in G10 while Mitsubishi Ferry was hold by another player (Sir_Yaro).

tobymao:
@michaeljb is this fixed?

michaeljb:
No the Mitsubishi Ferry is still on `any` time. To follow the rules completely, we'll need to add some more times that can be checked, like "between OR turns".

PR #1729, opened 2020-09-26
Fixes for using abilities out of turn

### LINKED ISSUE #1725: 1889: Cannot use Mitsubishi Ferry ability if not president of any corporation

The rules have this wording:

perwestling:
1889: Correct how Ferry private is used. (Fixes #1725 and #1558) 

General: Make it possible for owner to use ability out-of-turn. And stop non-user for doing this.

The general fix works for both Duogo Railway and Mitsubishi Ferry privates.

The general fix need to be verified versus existing games that it does not break anything, or at least that it works as intended in the cases it do break some game. Have verified that it work as intended in a multi-player 1889.

### DIFF
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -35,8 +35,18 @@ module View
             !action.free? &&
             participant &&
             !@game.active_players.map(&:name).include?(@user['name']) &&
-            !Lib::Storage[@game.id]&.dig('master_mode')
+            !Lib::Storage[@game.id]&.dig('master_mode') &&
+            !out_of_turn_allowed?(action)
           return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
+        end
+
+        if !hotseat &&
+          !action.free? &&
+          participant &&
+          @game.active_players.map(&:name).include?(@user['name']) &&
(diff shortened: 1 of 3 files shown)


assets/app/view/game/actionable.rb
```diff
@@ -35,10 +35,20 @@ def process_action(action)
             !action.free? &&
             participant &&
             !@game.active_players.map(&:name).include?(@user['name']) &&
-            !Lib::Storage[@game.id]&.dig('master_mode')
+            !Lib::Storage[@game.id]&.dig('master_mode') &&
+            !out_of_turn_allowed?(action)
           return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
         end
 
+        if !hotseat &&
```
tobymao, line 43: you should store this in a new var so it's not repeated

assets/app/view/game/actionable.rb
```diff
@@ -86,6 +96,24 @@ def clear_ui_state
         store(:selected_company, nil, skip: true)
         store(:tile_selector, nil, skip: true)
       end
+
+      private
+
+      def out_of_turn_allowed?(action)
+        entity = action.entity
+        entity.company? && entity.owner.name == @user['name']
```
tobymao, line 104: is this always true?

PR #1730, opened 2020-09-26
18MEX: Implement needed stuff for phase 3½

perwestling:
Introduce 5% shares for NdM.

Reserve shares used for mergers.

Implement phase 3½ merge of minors.

General: Add buyable attribute for shares and bundles. For a bundle, it is buyable if all shares within it is buyable.


perwestling:
Remaining: Sell of 5% shares should not lower market price.

tobymao (approved):

PR #1731, opened 2020-09-26
Disallow infinitesmal track-sharing (fixes #1353)

### LINKED ISSUE #1353: [1889] Train could run illegal route

dfannius:
Tested on the reported example in #1353.

Someone who understands lanes should update `Path::track_resources_used` to take them into account. It should be very straightforward.

Profiling before and after this change, `Route::check_overlap!` went from 0.6% to 2.0% of total time. It could be sped up if necessary (the current code is optimized for clarity, not for speed), but so could plenty of other things.


roseundy (commented):
Make this change to path.rb line 135: `TrackResource.new(hex, p.num, @lanes[p == @a ? 0 : 1][1])`

Change "lane" in track_resource.rb to "lane_index" everywhere, and it should work.

I'm wondering though if adding a structure to all paths is the way to go. Would it be more efficient to create a list of [hex, edge, lane_index]  in route.rb and check to see if you have duplicates?

dfannius:
> I'm wondering though if adding a structure to all paths is the way to go. Would it be more efficient to create a list of [hex, edge, lane_index] in route.rb and check to see if you have duplicates?

Indeed, that was how I first implemented it, but it felt a little bare-metal and non-abstract, and I wasn't certain we'd never have to generalize it further (in future titles). I'm fine with whatever the consensus is.

lib/engine/route.rb
```diff
@@ -152,8 +152,8 @@ def check_cycles!
     end
 
     def check_overlap!
-      @routes.flat_map(&:paths).group_by(&:itself).each do |k, v|
-        @game.game_error("Route cannot use same path twice #{k.inspect}") if v.size > 1
+      @routes.flat_map { |r| r&.paths&.flat_map(&:track_resources_used) }.group_by(&:itself).each do |k, v|
```
tobymao, line 155: how can route be nil? i think we should switch it to 'bare metal', it's simple enough

@routes.flat_map do |route|
route.paths.flat_map ...

dfannius:
Bare metal it is. `Route::check_overlap!` time went down from 2.0% to 1.7%.

roseundy:
We need to check to make sure works correctly with 18MEX.

tobymao (approved):

dfannius:
P.S. After Toby's further performance improvements (29552831b1f98dc68ad17c19ed29117ef7aff069), time spent in `Route::check_overlap!` went down from 1.7% to 1.6%.

Issue #1434, opened 2020-08-12
1882: Request abbreviation changes


TheExcalabur:
In 1882 the assigned abbreviations are confusing as they stand.  

With both the Canadian National and the Canadian Northern in the game, that's an understandable problem.  However, the chosen on-site abbreviations don't match the usage in other games, some historical sources, or the spreadsheets/some paper copies.  

I would recommend:
CNR (or CN) -- Canadian National Railway
CNoR (or CNo) -- Canadian Northern Railway.  

Minor fix, but I think it'd help.

dfannius:
In (the latest revision of) the AAG rules for 1882, the Canadian National Railway is CN and the Canadian Northern Railway is CNoR.

tobymao:
matches closely to aag, closing

TheExcalabur:
That "o" is _really_ important.  Please?  You effectively currently have the 'CN' and the 'CN Railway' in the game.

PR #1735, opened 2020-09-27
[1882] CNR -> CNoR

benjaminxscott:
The [AAG rules](https://boardgamegeek.com/filepage/206629/1882-rules) indicate that the Canadian Northern Railway's mark is `CNoR` and Canadian National Railway's mark is `CN`

Historically it looks like the Canadian Northern Railway never used the mark `CNR` (Canadian National was known as `CNR` or `CN`)

Related to #1434


tobymao:
i'm not really inclined to fix this, you need to rename the svg and create a migration to change all existing actions to account for this ID change

benjaminxscott:
ah yeah that would be a lot more work, makes sense since current games would need to be migrated

PR #1736, opened 2020-09-27
allow "empty" hexes in game conf definition, fix axes for 18MS column 14

michaeljb:
18MS has no hexes in column 14, now we can add an "empty" hex like B14 to fix axes rendering. Seemed like this would be simpler than trying to identify and fill gaps in letters or numbers.

Before:

Screenshot from 2020-09-27 11-47-34

After:

Screenshot from 2020-09-27 11-46-15


michaeljb:
wrong branch name, will re-open

PR #1737, opened 2020-09-27
allow "empty" hexes in game conf definition, fix axes for 18MS column 14

michaeljb:
18MS has no hexes in column 14, now we can add an "empty" hex like B14 to fix axes rendering. Seemed like this would be simpler than trying to identify and fill gaps in letters or numbers.

Before:

Screenshot from 2020-09-27 11-47-34

After:

Screenshot from 2020-09-27 11-46-15

### DIFF
--- a/lib/engine/hex.rb
+++ b/lib/engine/hex.rb
@@ -8,7 +8,7 @@ module Engine
     include Assignable
 
     attr_accessor :x, :y
-    attr_reader :connections, :coordinates, :layout, :neighbors, :tile, :location_name, :original_tile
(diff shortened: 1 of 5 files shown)


lib/engine/hex.rb
```diff
@@ -75,6 +76,7 @@ def initialize(coordinates, layout: nil, axes: nil, tile: Tile.for('blank'), loc
       @original_tile = @tile = tile
       @tile.hex = self
       @activations = []
+      @empty_placeholder = empty_placeholder
```
tobymao, line 79: can we just call it empty, don't need the helper method on line 224

tobymao (approved):

PR #1739, opened 2020-09-27
Add crossover rendering

### LINKED ISSUE #513: Redraw tiles to make it obvious that you cannot change track at a crossover
These Vassal tiles (for a different game) should give you an idea of what I mean:

image

It's visually obvious that you can't change track at the crossover as one track lies on top of the other.  

Probably not necessary for experienced players, but as a noob I've been caught out by misreading the tiles. 

EDIT: The Carthaginian redesign came up with an elegant way of doing the crossovers: 

image

roseundy:
Fixes #513 

Also fixed boolean calls in path.rb and one random tile.

New rendering:
crossover
crossover_2
56_crossover
crossover_route

### DIFF
--- a/assets/app/view/game/part/track_node_path.rb
+++ b/assets/app/view/game/part/track_node_path.rb
@@ -18,6 +18,8 @@ module View
         needs :dash, default: '0'
 
         CROSSOVER_GAP = 3
+        STRAIGHT_CROSSOVER = '1 55 63 56'
+        GENTLE_CROSSOVER = '1 55 47 56'
 
         PARALLEL_SPACING = [8, 6, 4].freeze
 
@@ -262,9 +264,9 @@ module View
             lanes = @path.lanes
 
             if @tile.crossover? && @path.straight?
-              @crossover_dash = '1 55 63 56'
+              @crossover_dash = STRAIGHT_CROSSOVER
             elsif @tile.crossover? && @path.gentle_curve?
-              @crossover_dash = '1 55 47 56'
+              @crossover_dash = GENTLE_CROSSOVER
             end
 
           elsif @num_exits == 1
@@ -293,11 +295,12 @@ module View
             lanes = @path.lanes
             lanes.reverse! if @path.b.edge?
 
-            if @tile.crossover? && @path.straight?(@ct_edge0)
-              @crossover_dash = '1 55 63 56'
-            elsif @tile.crossover? && @path.gentle_curve?(@ct_edge0)
-              @crossover_dash = '1 55 47 56'
+            if @tile.crossover? && @path.straight?
+              @crossover_dash = STRAIGHT_CROSSOVER
+            elsif @tile.crossover? && @path.gentle_curve?
+              @crossover_dash = GENTLE_CROSSOVER
             end
+
           else
             # city/town - city/town
             @ct_edge0 = @tile.preferred_city_town_edges[@stop0] if @stop0
@@ -428,17 +431,15 @@ module View
             'stroke-dasharray': @dash,
           ) if @terminal
 
-          children = []
-          children.append(h(:path, props))
+          children = [h(:path, props)]
           if @crossover_dash
-            intersect_props = props.dup
-            intersect_props[:attrs].merge!(
+            props[:attrs].merge!(
               stroke: setting_for(@tile.color),
               'stroke-width': @width + CROSSOVER_GAP * 2,
               'stroke-dasharray': @crossover_dash,
               'stroke-dashoffset': 1,
             )
-            children.prepend(h(:path, intersect_props))
+            children.prepend(h(:path, props))
           end
           children
         end
--- a/assets/app/view/game/part/track_node_path.rb
+++ b/assets/app/view/game/part/track_node_path.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'lib/settings'
 require 'view/game/part/base'
 require 'view/game/part/town_location'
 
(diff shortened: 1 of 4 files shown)


assets/app/view/game/part/track_node_path.rb
```diff
@@ -256,6 +260,13 @@ def load_from_tile
             @end_x = edge_x_pos(@end_edge, 87)
             @end_y = edge_y_pos(@end_edge, 87)
             lanes = @path.lanes
+
+            if @tile.crossover? && @path.straight?
+              @crossover_dash = '1 55 63 56'
```
tobymao, line 265: these strings should be constants

assets/app/view/game/part/track_node_path.rb
```diff
@@ -411,7 +428,19 @@ def render_part
             'stroke-dasharray': @dash,
           ) if @terminal
 
-          h(:path, props)
+          children = []
```
tobymao, line 431: children = [h(:path, props)]

assets/app/view/game/part/track_node_path.rb
```diff
@@ -411,7 +428,19 @@ def render_part
...
+          children.append(h(:path, props))
+          if @crossover_dash
+            intersect_props = props.dup
```
tobymao, line 434: does this dup actually do what you want? dup is usually just a shallow copy, so nested dictionaries would still be mutated?

lib/engine/part/path.rb
```diff
@@ -98,21 +98,55 @@ def path?
...
       def node?
-        @_node ||= @nodes.any?
+        return @_node if defined?(@_node)
+
+        @_node = @nodes.any?
       end
 
       def terminal?
-        @_terminal ||= !!@terminal
+        return @_terminal if defined?(@_terminal)
```
tobymao, line 107: there's no reason to cache this, it's not expensive at all just do

def terminal?
!!@terminal

lib/engine/tile.rb
```diff
@@ -360,6 +360,43 @@ def compute_city_town_edges
...
+    def crossover?
+      return @_crossover if defined?(@_crossover)
+
+      @_crossover = compute_crossover
+    end
+
+    def compute_crossover
+      edge_paths = Hash.new { |h, k| h[k] = [] }
+      @paths.each do |p|
+        edge_paths[p.a.num].append(p) if p.a.edge?
```
tobymao, line 373: use << in favor of append

assets/app/view/game/part/track_node_path.rb
roseundy: It seems to work, but I can change to a a deep copy.

lib/engine/part/path.rb
roseundy: OK.

lib/engine/tile.rb
roseundy: OK.

assets/app/view/game/part/track_node_path.rb
roseundy: OK

assets/app/view/game/part/track_node_path.rb
roseundy: OK.

lib/engine/tile.rb
```diff
@@ -360,6 +360,43 @@ def compute_city_town_edges
...
+        edge_paths[p.b.num].append(p) if p.b.edge?
+        next unless p.nodes.size == 1
```
tobymao, line 375: net unless p.nodes.one?

lib/engine/tile.rb
```diff
@@ -360,6 +360,43 @@ def compute_city_town_edges
...
+
+        ct_edge = preferred_city_town_edges[p.nodes.first]
```
tobymao, line 377: what's the point of these 3 lines? they don't do anything

lib/engine/tile.rb
```diff
@@ -360,6 +360,43 @@ def compute_city_town_edges
...
+        p.straight?(ct_edge)
+        p.gentle_curve?(ct_edge)
+      end
+
+      @paths.each do |p|
+        next if p.nodes.size > 2
```
tobymao, line 383: how can a path have more than two nodes?

lib/engine/tile.rb
roseundy: OK.

lib/engine/tile.rb
roseundy: It calls those methods in path with the proper city/town thus causing them to cache. I don't want to call preferred_city_town_edges in Path.

lib/engine/tile.rb
roseundy: Damn. Should be 1.

assets/app/view/game/part/track_node_path.rb
tobymao: maybe separate out the hashes

lib/engine/tile.rb
tobymao: what's the point of caching them here? why don't you just do it lazily?

tobymao: ohhhhh, wait what, you rely on the order for caching? that's a really dangerous pattern. the cache should take into consideration arguments

assets/app/view/game/part/track_node_path.rb
roseundy: Actually I don't need to do either. It seems each call of h(:path, props) is essentially a call by value of the properties, so I'm free to change props between calls.

lib/engine/tile.rb
roseundy: Revamped this to not use arguments at all and to avoid this entire problem.

tobymao (approved):

PR #1740, opened 2020-09-27
Make it possible to introduce a different train limit for a corporation

perwestling:
Add a train_limit ability, with attribute increase, that can be added
to a corporation to set a permanent increase in the train limit.

### DIFF
--- a/lib/engine/phase.rb
+++ b/lib/engine/phase.rb
@@ -126,8 +126,8 @@ module Engine
     private
 
     def train_limit_increase(entity)
-      ability = entity.abilities(:train_limit)
-      ability.any? ? ability.first.increase : 0
+      entity.abilities(:train_limit) { |ability| return ability.increase }
+      0
     end
   end
 end
--- a/lib/engine/phase.rb
+++ b/lib/engine/phase.rb
@@ -4,7 +4,7 @@ require_relative 'action/buy_train'
 
 module Engine
   class Phase
-    attr_reader :name, :operating_rounds, :train_limit, :tiles, :phases, :status
+    attr_reader :name, :operating_rounds, :tiles, :phases, :status
 
     def initialize(phases, game)
       @index = 0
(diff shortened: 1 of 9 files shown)


lib/engine/phase.rb
```diff
@@ -31,6 +31,10 @@ def current
       @phases[@index]
     end
 
+    def train_limit(entity)
```
tobymao, line 34: entity = nil

tobymao: and handle nil entity

lib/engine/phase.rb
```diff
@@ -118,5 +122,12 @@ def next!
       @index += 1
       setup_phase!
     end
+
+    private
+
+    def train_limit_increase(entity)
+      ability = entity.abilities(:train_limit)
```
tobymao, line 129: entity.abilities(:train_limit) { |ability| return ability.increase } 
0

lib/engine/phase.rb
tobymao: i see i guess you enforce it to exist, i guess it's ok

lib/engine/phase.rb
perwestling: Done

lib/engine/phase.rb
perwestling: I think it is fine to not have default value here, and instead enforce it to be specified. Did update all calls, and they all had access to a corporation.

perwestling:
Compare #1742 

Is this a good way of adding special abilities for a corporation in a title? 

An alternative is to add a description ability without any functionality (so it shows up on the charter) and let the title implement the ability.

I do think it is good to, if possible, show special rules for an entity, in the game.

@tobymao Have pushed a fix for review comment.

tobymao:
what do you mean add special abilities for a corporation? doesn't 1846 do this?

tobymao (approved):

PR #1741, opened 2020-09-27
18MEX: Add mail feature

perwestling:
If a major corporation has trains at start of its OR, add revenue
for the home city to the treasury.

### DIFF
--- a/lib/engine/round/g_18_mex/operating.rb
+++ b/lib/engine/round/g_18_mex/operating.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require_relative '../operating'
+
+module Engine
+  module Round
+    module G18MEX
+      class Operating < Operating
+        def skip_steps
+          entity = @entities[@entity_index]
+          return super if entity.minor?
+
+          entity.trains.empty? ? handle_no_mail(entity) : handle_mail(entity)
+          super
+        end
+
+        private
+
+        def handle_no_mail(entity)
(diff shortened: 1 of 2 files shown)


perwestling:
I considered adding a mail ability, to all corporations, but thought this mail might be too special.

Anyway, the house keeping method addition in the operating is something that I think can be used in more games. E.g. L trains bought by minors in 1822. The term "house keeping" is from 1822.
It is things done at the very start of the OR, before the first tile lay. So in 18MEX it is mail.

lib/engine/round/g_18_mex/operating.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require_relative '../operating'
+
+module Engine
+  module Round
+    module G18MEX
+      class Operating < Operating
+        def house_keeping
```
tobymao, line 9: can't you just override start operating? i don't know if house keeping adds that much

perwestling: That I tried at first. If I do "house_keeping" + super, this means that the "XXX operating YYY" message comes after the mail message.
So I could solve that by overriding skip_steps to do mail + super.
Then I got the idea about 1822, and its housekeeping.

But sure, I can go for the override of skip_steps.

perwestling:
As only 18MEX code was changed after acting upon the review comment, I rebased it as well.

tobymao (approved):

PR #1748, opened 2020-09-28
18MEX: Support NdM merge

perwestling:
When first 5 train is bought the NdM merge event is triggered
if NdM has floated.

The merge process if that player in order, left of the player
that has bought the 5 train, is asked if their corporation wants
to merge. (Excluding "American" corporations, and corporations
controlled by NdM president.) If there is no floated corporation
that accepts, NdM will merge with any non floated, if there is
any eligible one.
The president of the merged corporation receives a 10% reserved
NdM share in exchange for their presidency. Remaining shares
in merged corporation is force sold for half amrket price.
Treasury and trains are transfered to NdM.
NdM also exchanges up to two tokens, one should at least be
the home token if any.

There are a new step: merge, which is located in module 18MEX.
This is used for the merge question. A connected view has
been added for the merge step.

18MEX also uses an assign step which is used in case NdM need
to choose between several tokens on the map.

---------

Have added a buyable attribute to share. This is set to true by
default, and is set to false if share reserved. In 18MEX this
is done for exchange shares that will be used during minor
closure and NdM merge.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -50,6 +50,7 @@ module View
 
         unless @corporation.minor?
           children << render_shares
+          children << render_reserved if @corporation.reserved_shares.any?
           children << h(Companies, owner: @corporation, game: @game) if @corporation.companies.any?
         end
 
@@ -335,6 +336,18 @@ module View
         ])
       end
 
+      def render_reserved
+        bold = { style: { fontWeight: 'bold' } }
+        h('table.center', [
+          h(:tbody, [
+            h('tr.reserved', [
+              h('td.left', bold, "#{@game.class::IPO_RESERVED_NAME}:"),
+              h('td.right', @corporation.reserved_shares.map { |s| "#{s.percent}%" }.join(', ')),
+            ]),
+          ]),
+        ])
+      end
+
       def render_revenue_history
         last_run = @corporation.operating_history[@corporation.operating_history.keys.max].revenue
         h(:div, { style: { display: 'inline' } }, [
@@ -347,8 +360,9 @@ module View
         return [] unless @game.round.current_entity&.operator?
 
         round = @game.round
-        if (n = @game.round.entities.find_index(@corporation))
-          span_class = '.bold' if n >= round.entities.find_index(round.current_entity)
+        if (n = @game.round.entities.index(@corporation))
+          m = round.entities.index(round.current_entity)
+          span_class = '.bold' if n && m && n >= m
           [h(:div, { style: { display: 'inline' } }, [
             'Order: ',
             h("span#{span_class}", n + 1),
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -251,7 +251,7 @@ module View
           h(Game::Round::Stock, game: @game)
         end
       when Engine::Round::Operating
-        if (%w[merge] & current_actions).any?
+        if current_actions.include?('merge')
           h(Game::Round::Merger, game: @game)
         else
           h(Game::Round::Operating, game: @game)
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -251,7 +251,11 @@ module View
           h(Game::Round::Stock, game: @game)
         end
       when Engine::Round::Operating
-        h(Game::Round::Operating, game: @game)
+        if (%w[merge] & current_actions).any?
+          h(Game::Round::Merger, game: @game)
+        else
+          h(Game::Round::Operating, game: @game)
+        end
       when Engine::Round::Draft
         h(Game::Round::Auction, game: @game, user: @user, before_process_pass: @before_process_pass)
       when Engine::Round::Auction
--- a/lib/engine/game/g_18_mex.rb
+++ b/lib/engine/game/g_18_mex.rb
@@ -16,6 +16,8 @@ module Engine
       GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18MEX'
       GAME_END_CHECK = { bankrupt: :immediate, stock_market: :current_or, bank: :current_or }.freeze
 
+      IPO_RESERVED_NAME = 'Trade-in shares'
+
       STANDARD_GREEN_CITY_TILES = %w[14 15 619].freeze
       CURVED_YELLOW_CITY = %w[5 6].freeze
 
@@ -48,8 +50,8 @@ module Engine
         @minor_b_reserved_share ||= ndm.shares[8]
       end
 
-      def ndm_merge_reserved_share
-        @ndm_merge_reserved_share ||= ndm.shares.last
+      def ndm_merge_share
+        @ndm_merge_share ||= ndm.shares.last
       end
 
       def udy
@@ -72,6 +74,15 @@ module Engine
         @minor_c ||= minor_by_id('C')
       end
 
+      # Set to 1 if no-one accepts NdM merge
+      def cert_limit_adjust
+        @cert_limit_adjust ||= 0
+      end
+
+      def cert_limit
+        super + cert_limit_adjust
+      end
+
       include CompanyPrice50To150Percent
       include Revenue4D
 
@@ -87,6 +98,15 @@ module Engine
           hex.tile.cities[0].place_token(minor, minor.next_token)
           minor.float!
         end
+
+        # TODO: Can neutral be removed? Move shares to market instead
+        # before deleting them.
+        @neutral = Corporation.new(
+          sym: 'N',
+          name: 'Neutral',
+          tokens: [],
+        )
+        @neutral.owner = @bank
 
         @brown_g_tile ||= @tiles.find { |t| t.name == '480' }
         @gray_tile ||= @tiles.find { |t| t.name == '455' }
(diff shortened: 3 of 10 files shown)


assets/app/view/game/merge.rb
```diff
@@ -0,0 +1,35 @@
...
+        children << render_description(corporation)
+        children << h(:div, [h(:button, { on: { click: merge } }, 'Merge into NdM')])
+        children << h(Corporation, corporation: corporation)
+        children << h(Corporation, corporation: @game.ndm)
+        h(:div, children)
+      end
+
+      def render_description(corporation)
+        player = corporation.player
+        player ||= @game.ndm.player
```
jenf, line 30: Bit confused on this, should probably be a needs into the view to say the player name.

assets/app/view/game/merge.rb
```diff
@@ -0,0 +1,35 @@
...
+        merge = lambda do
+          process_action(Engine::Action::Merge.new(
+            corporation.player ? @game.current_entity : @game.ndm,
+            corporation: corporation,
+          ))
+        end
+
+        children = [h(:div, "Merge candidates: #{@game.mergable_candidates.map(&:name)}")]
+        children << render_description(corporation)
+        children << h(:div, [h(:button, { on: { click: merge } }, 'Merge into NdM')])
```
jenf, line 22: Should be generic and have NdM as a const.

assets/app/view/game/merge.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+require 'view/game/undo_and_pass'
+
+module View
+  module Game
+    class Merge < Snabberb::Component
```
jenf, line 8: Is there any way we can get 1817 to use some of this code?

perwestling:
Here is how the Merge view looks like when a player is asked if their corporation should be merged into NdM:
image

And this is how it looks when a non-floated corporation should be merged into NdM:
image

And here is how it looks like when CHI accepts the merge, and NdM has to select which token (besdies CHI home) that should be swapped (the non-chosen is removed):
image

perwestling:
That dialog that let's NdM confirm that SPM should be merged could instead be automatic - if the last candidate is a non-floated one it should be selected. The only case for NdM to have to do a choice should be if there are several non-floated candidates that have player owned 20% shares; which is very unlikely to happen.

BTW, this is how it looks like if NdM player clicks on Mazatlán (removing CHI token in Torreón):
image

Note that CHI has disappeared from the order. And in the Entities view all CHI has been removed:
image

assets/app/view/game/merge.rb
perwestling: See pictures below. The idea here is the merge candidate is shown, and the president of it makes the decision if it should merge. In case it has no president, it is a non-started one, and then NdM should make the decision. (Note my remark below that this is very likely automatic.)

I know realize that this is probably an error - only if the corporation is floated the player can decide. Non-floated is automatically merged if one if players decline. If non-floated and non-sold presidency, that is chosen if players decline. And if there is two non-floated eligible (both non-sold, or sold) - then NdM gets to choose.

perwestling: As can be seen from the screen shots, the merge candidate is shown underlined in the Order, as well as in text above the button. Only time it is not shown underlined is if it is a non-floated corporation that is up for merge.

assets/app/view/game/merge.rb
perwestling: That's an idea. I need to make @game.ndm generic as well. Probably add a method to base that g_18_mex.rb can override?  What do you call the one that you merge into? Mergee for the one that is merged into the... Merger?

Or should this view be only used by 18MEX? Can it be put in a Module in that case?

assets/app/view/game/merge.rb
perwestling: I have not studied 1817 in detail but it would be good if it was. I don't recall such a situation in 1817, where there is a list of candidate corporations to be merged into a target, but maybe there is? 
I suppose it could be used for a Merge/Decline question of one mergee into one merger.

jenf: In acquisitions once a player has won the bid they then choose the corporation to merge with it.

perwestling: Is there such a dialog implemented today? If not it sounds like this can be generalized to cover for that as well.  
Will try to generalize it a bit.

assets/app/view/game/corporation.rb
```diff
@@ -348,7 +348,8 @@ def render_operating_order
 
         round = @game.round
         if (n = @game.round.entities.find_index(@corporation))
-          span_class = '.bold' if n >= round.entities.find_index(round.current_entity)
+          m = round.entities.find_index(round.current_entity)
```
tobymao, line 351: index i believ is the same

assets/app/view/game/merge.rb
tobymao: the views shouldn't have any game specific logic like this

lib/engine/game/g_18_mex.rb
```diff
@@ -136,8 +158,140 @@ def event_minors_closed!
...
+      # as well.
+      def merge_major(major = nil)
+        @ndm_merge_completed = true
+        @mergable_candidates = []
+        exchange_tokens = [ndm.tokens[-2], ndm.tokens.last]
+
+        # Make reserved share available
+        ndm_merge_share.buyable = true
+
+        if major.nil?
```
tobymao, line 183: unless major

perwestling: Done

assets/app/view/game/merge.rb
perwestling: Ok, I have now generalized the view so it can be used by other titles.

assets/app/view/game/merge.rb
perwestling: Improved the code a bit to make it clearer.

assets/app/view/game/corporation.rb
perwestling: Yes, index is just an alias to find_index. I changed to use index in both lines now as it is shorter.

perwestling:
@jenf and @tobymao : I have made some improvements, and handled review comments.

One thing I added was to display the reserved (in 18MEX: trade-in) IPO shares in the corporation charter. These cannot be bought. (The title should ensure these are the last ones in the share array so that all other shares can be bought.)
Here is a screen shot on how this looks from the start in 18MEX:
image

And here how it looks just before the NdM merge (when the last trade-in share is "up for grabs")
image

Regardless if merge take place or not that last share is changed to buyable, so the "Trade-in shares" section disappear from NdM. As can be seen CHI does not have any trade-in shares at all. In 18MEX NdM has 3 and UdY has 1 from the start.

assets/app/view/game/merge.rb
perwestling: Have generalized it a bit. But still not sure if it is possible to use it in 1817. Maybe in 1824?

assets/app/view/game/merge.rb
```diff
@@ -0,0 +1,29 @@
...
+        corporation = @game.mergable_candidates.first
+        merge = lambda do
+          process_action(Engine::Action::Merge.new(
+            @game.merge_target,
+            corporation: corporation,
+          ))
+        end
+
+        children = [h(:div, "Merge candidates: #{@game.mergable_candidates.map(&:name)}")]
+        children << h(:div, "To make the decision: #{@game.merge_decider.player.name}")
```
jenf, line 21: This is a bit of a different flow than normal, but we are hitting issues where the 'active' player is not one that's normally listed on the round including in 1817.

Maybe we should have a standard way of showing that the normal turn sequence flow has been interrupted and who the interrupting entity is. But still showing the normal turn sequence as before.

@tobymao this would be nice to include on 1817 as the acquisiton is player based, but when your gaining/paying off loans it's corporation based. Would also help on the cash crisis as well, as it will get confusing whose turn it is.

jenf: The current entity should be the 'merge_decider' anyway. So that the email gets sent to the right user.

perwestling: In g_18_mex/merge.rb, active_entities is [@game.merge_decider].
Normally, the merge_decider in 18MEX would be the president of the corporation to be merged into NdM. But there is one exception - if there is two(+) eligible corporations, that has not floated, then NdM will be the merge decider. It is probably a rare case.
If there is just one non-floated to choose from NdM is not asked but instead it is auto merged.

But NdM might have to select among multiple tokens placed, and that will also interrupt normal flow, before returning to Buy Trains step.

jenf: Merging up to the latest should add the current entity to the list, removing the need for the "To make the decision" part

perwestling:
Have removed the "To make the decision" part, moved the "Merge candidates" to the right of the Merge button. Rebased versus master.

Ready to be reviewed again.

assets/app/view/game/merge.rb
tobymao: has this been attempted yet? that's the only thing i'd look into for this commit

perwestling: I have not done any 1817 related work. So maybe @tobymao or @jenf could try to use it for acquisitions?

assets/app/view/game/merge.rb
perwestling: Have removed the "To make the decision" part, moved the "Merge candidates" to the right of the Merge button. Rebased versus master.

Have not uploaded new screen shots as the changes are very minor.

perwestling:
I will resolve the conflicts, but I am waiting with that until it has been enough reviewed.

perwestling:
@tobymao and @jenf I have now updated this commit. Have rebased it versus latest master.
In the commit it now uses the already existing merge dialog. That had to be updated a bit to make it work for 18MEX.
I also introduced preselected corporation in case only one candidate exists (which is the case for 18MEX where one candidate at a time is presented as a choice).

perwestling:
I have now fixed the exchange tokens (so that they appear on the charter only as part of the NdM merge.

So with this PR I think 18MEX is ready, functionally, to be put to alpha. I am doing some pre-alpha testing on the test site. It uses the same code, except the possibility to inter-buy corporations (as it breaks the started game if added). 

So it would be good to have it reviewed.

assets/app/view/game_page.rb
```diff
@@ -251,7 +251,11 @@ def render_action
           h(Game::Round::Stock, game: @game)
         end
       when Engine::Round::Operating
-        h(Game::Round::Operating, game: @game)
+        if (%w[merge] & current_actions).any?
```
tobymao, line 254: current_actions.include?('merge')

tobymao (approved):

assets/app/view/game_page.rb
perwestling: Fixed it

perwestling:
@tobymao I haved fixed your comment. Time to merge it?

PR #1763, opened 2020-10-01
Add support for Lawson tiles: upgrades and routes

### LINKED ISSUE #1756: Cannot upgrade plain yellow track to lawson style green track
When using Lawson style track (ie tiles 80-83 for plain green track) they won't show up as possible upgrades for yellow tiles 7 - 9 (or 77-79)

roseundy:
Fixes #1756 

1. Recoded Path "<=" method to comprehend junctions (via new jwalk method). This was needed to allow upgrades to Lawson tiles.
2. Junction class no longer is a subclass of Node, but of Base directly
3. Modified Path "walk" method to comprehend junctions
4. Added Path "match?" method that is identical to old "<=" method
5. Modified Connection "branch!" method to deal with junctions
6. Removed branches array from Path and Tile

### DIFF
--- a/lib/engine/connection.rb
+++ b/lib/engine/connection.rb
@@ -114,25 +114,20 @@ module Engine
     #    see paths for ealier connections
     # 2. @paths are in order (i.e. the head of one connects to the tail of the next)
     def branch!(path)
-      ends = Hash.new(0)
-      @paths.each do |p|
-        ends[p.a_id] += 1 if p.a.junction? || p.a.edge?
-        ends[p.b_id] += 1 if p.b.junction? || p.b.edge?
-      end
-
-      return self unless ends[path.a_id] > 1 || ends[path.b_id] > 1 ||
-                         (ends[path.a_id].positive? && path.a.edge?) ||
-                         (ends[path.b_id].positive? && path.b.edge?)
-
-      # only keep paths leading to edge/junction
       branched_paths = []
       @paths.each do |p|
+        # we've seen this edge before
         break if path.a.edge? && (path.a_id == p.a_id || path.a_id == p.b_id)
         break if path.b.edge? && (path.b_id == p.a_id || path.b_id == p.b_id)
 
         branched_paths << p
-        break if ([path.a_id, path.b_id] & [p.a_id, p.b_id]).any?
+
+        # we've seen this junction before
+        break if path.a.junction? && (path.a == p.a || path.a == p.b)
+        break if path.b.junction? && (path.b == p.a || path.b == p.b)
       end
+
+      return self if branched_paths.size == @paths.size
 
       branch = self.class.new(branched_paths)
 
--- a/lib/engine/connection.rb
+++ b/lib/engine/connection.rb
@@ -20,7 +20,6 @@ module Engine
           connection = connection.branch!(path)
           next if connection.paths.include?(path)
           next if (connection.nodes & path.nodes).any?
-          next if connection.paths.any? { |p| p.hex == hex && (p.exits & path.exits).any? }
 
           connections[connection] = true
           connection.add_path(path)
@@ -115,23 +114,24 @@ module Engine
     #    see paths for ealier connections
     # 2. @paths are in order (i.e. the head of one connects to the tail of the next)
     def branch!(path)
-      branched_paths = path.select(@paths)
-
       ends = Hash.new(0)
       @paths.each do |p|
-        ends[p.a] += 1 if p.a.junction?
-        ends[p.b] += 1 if p.b.junction?
+        ends[p.a_id] += 1 if p.a.junction? || p.a.edge?
+        ends[p.b_id] += 1 if p.b.junction? || p.b.edge?
       end
 
-      return self if ends[path.a] < 2 && ends[path.b] < 2 && @paths.size == branched_paths.size
+      return self unless ends[path.a_id] > 1 || ends[path.b_id] > 1 ||
+                         (ends[path.a_id].positive? && path.a.edge?) ||
+                         (ends[path.b_id].positive? && path.b.edge?)
 
-      # if junction, only keep prior paths
-      if ends[path.a] == 2 || ends[path.b] == 2
-        branched_paths = []
-        @paths.each do |p|
-          branched_paths << p
-          break if path.a == p.a || path.a == p.b || path.b == p.a || path.b == p.b
-        end
+      # only keep paths leading to edge/junction
+      branched_paths = []
+      @paths.each do |p|
+        break if path.a.edge? && (path.a_id == p.a_id || path.a_id == p.b_id)
+        break if path.b.edge? && (path.b_id == p.a_id || path.b_id == p.b_id)
+
+        branched_paths << p
+        break if ([path.a_id, path.b_id] & [p.a_id, p.b_id]).any?
       end
 
       branch = self.class.new(branched_paths)
--- a/lib/engine/connection.rb
+++ b/lib/engine/connection.rb
@@ -110,9 +110,29 @@ module Engine
       hexes.include?(hex)
     end
 
+    # this code relies on two things:
+    # 1. Path::walk is depth-first - whenever a new connection is started due to branching, we will never
+    #    see paths for ealier connections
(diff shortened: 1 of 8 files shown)


lib/engine/part/path.rb
```diff
@@ -51,8 +50,32 @@ def <=>(other)
...
+        other.jwalk { |t| other_list << t }
+        this_list = []
+        jwalk { |t| this_list << t }
+        this_list.all? { |tt| other_list.any? { |ot| tt <= ot } }
+      end
+
+      def match?(other)
         (@a <= other.a && @b <= other.b) ||
-          (@a <= other.b && @b <= other.a)
+        (@a <= other.b && @b <= other.a)
```
tobymao, line 62: leave the indentation the same

lib/engine/connection.rb
```diff
@@ -20,7 +20,6 @@ def self.connect!(hex)
           connection = connection.branch!(path)
           next if connection.paths.include?(path)
           next if (connection.nodes & path.nodes).any?
-          next if connection.paths.any? { |p| p.hex == hex && (p.exits & path.exits).any? }
```
tobymao, line 23: was this removed due to lanes? will this cause issues?

roseundy: Two reasons:
1. It's redundant given the new way branch! has been coded
2. It doesn't handle lanes

I initially changed it to handle lanes and realized it looked exactly like what is now in branch!

lib/engine/part/path.rb
```diff
@@ -51,10 +50,36 @@ def <=>(other)
...
+      def match?(other)
         (@a <= other.a && @b <= other.b) ||
           (@a <= other.b && @b <= other.a)
       end
 
+      def ends
+        @ends ||= calculate_ends
+      end
+
+      def calculate_ends
```
tobymao, line 66: this could be

[@a, @b].flat_map do |j|
  next [j] unless j.junction?

  j.paths.map do |p|
    next if p == self
    [p.a, p.b].select(&:junction?)
  end

lib/engine/part/path.rb
```diff
@@ -65,13 +90,24 @@ def select(paths)
...
 
-      def walk(skip: nil, visited: nil, on: nil)
+      def walk(skip: nil, jskip: nil, visited: nil, on: nil)
         return if visited&.[](self)
 
         visited = visited&.dup || {}
         visited[self] = true
         yield self, visited
 
+        [@a, @b].each do |j|
```
tobymao, line 100: i thin you could replace this with just grabbing @junction because it is an instance variable of path

lib/engine/part/path.rb
roseundy: Ok

lib/engine/part/path.rb
roseundy: Yep

lib/engine/part/path.rb
```diff
@@ -152,6 +188,22 @@ def rotate(ticks)
         path
       end
 
+      def a_id
```
tobymao, line 191: i don't like the a_id methods and b_id methods, is it possible to call edge.id?

maybe change edge id so it works this?

roseundy: edge doesn't have access to lanes, which is why I made it part of Path.

roseundy: The problem is that lanes apply to things other than edges.

lib/engine/part/path.rb
```diff
@@ -51,10 +50,36 @@ def <=>(other)
       end
 
       def <=(other)
+        other_ends = other.ends
+        ends.all? { |t| other_ends.any? { |o| t <= o } }
+      end
+
+      def match?(other)
```
tobymao, line 57: do we need the match method? should the new <= method just replace it?

roseundy: We still need it to make sure we haven't added unreachable track.

lib/engine/part/path.rb
roseundy: I'm adding lanes to base and setting them in path.separate_parts

lib/engine/part/path.rb
roseundy: It's only a problem when upgrading Lawson tiles.

lib/engine/part/base.rb
```diff
@@ -7,7 +7,7 @@ module Part
     class Base
       include Helper::Type
 
-      attr_accessor :index, :tile
+      attr_accessor :index, :tile, :lanes
```
tobymao, line 10: i don't think lanes should be in base unless every single part has a lane

lib/engine/part/edge.rb
```diff
@@ -7,6 +7,10 @@ module Part
     class Edge < Base
       attr_accessor :num
 
+      def id
+        @id ||= "#{hex.id}-#{@num}-#{@lanes[1]}"
```
tobymao, line 11: in the previous code didn't you have lanes[0] and lanes[1]

tobymao: let's do underscores here as well

lib/engine/part/path.rb
tobymao: why doesn't <= work when upgrading lawson?

lib/engine/part/path.rb
```diff
@@ -51,10 +50,31 @@ def <=>(other)
...
 
+      def ends
+        @ends ||= calculate_ends
+      end
+
+      def calculate_ends
+        [@a, @b].flat_map do |j|
+          next [j] unless j.junction?
+
+          j.paths.map do |p|
```
tobymao, line 70: this looks like it's gonna break, i wonder why it didn't?

it should be paths.flat_map

next [] if p == self

lib/engine/part/edge.rb
tobymao: i see you did it in line 215 lanes[a or b]

lib/engine/tile.rb
```diff
@@ -467,9 +467,10 @@ def separate_parts
       end
 
       @nodes = @paths.flat_map(&:nodes).uniq
-      @branches = @paths.flat_map(&:branches).uniq
       @stops = @paths.flat_map(&:stops).uniq
       @edges = @paths.flat_map(&:edges).uniq
+
+      @edges.each { |e| e.tile = self }
```
tobymao, line 473: did this not work previously?

tobymao: why is this needed?

lib/engine/part/base.rb
roseundy: Any part that can be an endpoint of a path should probably have one. I can move it to node, edge and possibly junction.

lib/engine/part/edge.rb
roseundy: Lanes in Path is a two-dimensional array (array of arrays)

lib/engine/part/path.rb
roseundy: Because it should only affect Lawson tiles, and I hadn't gotten around to retesting them after these changes. My bad.

lib/engine/tile.rb
roseundy: No. I thought tile.separate_parts would have done this too in lines 462-467, but it wasn't working and truthfully I wasn't sure what that code was supposed to do.

lib/engine/part/edge.rb
roseundy: The first dimension is "a" vs. "b", the second dimension is lane size and lane index

lib/engine/part/path.rb
roseundy: Scratch that. It doesn't break. a and b don't have paths unless they are junctions.

roseundy: Still need to test with lawson tiles

roseundy: The select needs to be a reject

lib/engine/part/path.rb
roseundy: Fixed it so that it does now.

tobymao (approved):

PR #1773, opened 2020-10-03
Modify share ability to support multiple certificates

crericha:


crericha:
Tested with all the games that were affected.

assets/app/view/game/round/auction.rb
```diff
@@ -56,12 +56,16 @@ def render_players
         end
 
         def render_company_pending_par
-          corporation = @step.company_pending_par.abilities(:share).share.corporation
+          render = []
```
tobymao, line 59: call this children because render is a method name in snabberb::component

tobymao (approved):

assets/app/view/game/round/auction.rb
crericha: Change applied.

PR #1776, opened 2020-10-03
18MEX: Special handling of NdM 5% shares

perwestling:
Do not count 5% shares for cert limit

5% are sold one at a time, in separate bundles.

When selling a 5% share it does not affect share price.

TODO: What remains to implement is two things:
- If owning 2 5% it looks as if it is 1 share
- The rules have a special case where you can sell 5%
  if you own 10% or more, and there is a 5% in the open market.
  Maybe this should be an exchange action? As we are
  not allowed to sell 10% and then buy 5%.

### DIFF
--- a/lib/engine/player.rb
+++ b/lib/engine/player.rb
@@ -15,11 +15,10 @@ module Engine
     attr_accessor :bankrupt
     attr_reader :name, :companies
 
-    def initialize(name, count_companies: true)
+    def initialize(name)
       @name = name
       @cash = 0
       @companies = []
-      @count_companies = count_companies
     end
 
     def value
@@ -44,11 +43,6 @@ module Engine
 
     def player?
       true
-    end
-
-    def num_certs
-      num_companies = @count_companies ? companies.size : 0
-      num_companies + shares.count { |s| s.corporation.counts_for_limit }
     end
 
     def to_s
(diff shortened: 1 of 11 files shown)


lib/engine/player.rb
```diff
@@ -15,8 +15,9 @@ class Player
     attr_accessor :bankrupt
     attr_reader :name, :companies
 
-    def initialize(name, count_companies: true)
+    def initialize(name, game, count_companies: true)
```
tobymao, line 18: we should avoid this, models shouldn't take games

can you instead put a property on the share?

perwestling: Sure. Have pushed a refactored version.

tobymao:
sorry, that's not what i meant, but after looking at what you did and the current code,

i think it might be better to be a property of game

game.num_certs(player)

that way you don't have to pass the game in and we can get rid of the property on player "count_companies" 

do you mind making this refactor?

perwestling:
@tobymao I must have failed to notice your attribute comment; sorry about that.
Yes, that is probably a better way of doing it, so I rebased and refactored, making a new version of this commit. 

About "count_companies" I do find any title that uses that. I know where are titles where you do not count the privates for cert limit but none has so far been implemented.
So I did incorporate that change in the latest commit. It will make it easy to change the code to add game configuration to change the count.

tobymao (approved):

PR #1784, opened 2020-10-04
1846 2p - place Erie's second token during setup

### LINKED ISSUE #1781: 1846 2p - ERIE's second token should be placed in Erie during setup if ERIE is removed from the game
When ERIE is removed, its second token will be placed in Erie, so its reservation needs to stick around so that two other corporations may not token in Erie. When two corporations have incorrectly tokened there, when a green tile is placed, an error is thrown as there is no slot for ERIE to put its second token down.

https://www.18xx.games/game/11557?action=187

michaeljb:
[Fixes #1781]

### DIFF
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -159,8 +159,14 @@ module Engine
             end
 
             if two_player?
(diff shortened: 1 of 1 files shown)


michaeljb:
Breaks 11557 and any other games where ERIE was removed and a non-NYC corp placed a token in Erie.

Issue #1611, opened 2020-09-06
Seeing other company's routes


slapphappyjazz:
A general request from our group would be to update the game map for all players with the routes a company has selected.  Currently, only the acting director/president gets to see the routes selected.  Other players would like to see this too, once the director has submitted the routes.

dfannius:
I think this is covered by #563 and #1250.

slapphappyjazz:
#563 and #1250 seem a bit more general: they seem to be asking for the abilty to view a company's runs at any time.  I'm just suggesting that they runs are shown to other players during the active company's turn, from the point where the runs are submitted until the end of the turn.
The more general feature would be useful as well, but would be for when the player goes looking for that information between turns.  Showing the routes during the turn would be equivalent to the f-t-f explanation of the route being run during the turn.   I also assume it would be easier to implement, as it would only be a matter of showing other players the same view of the map as the active player.
This suggestion only makes sense for live play.  During asynchronous play it wouldn't make any differenc.e.

dfannius:
OK, let's keep them all open for now.

jenf:
Eilert Fleetjer  11:53
Hi, I don't have a github account but I would like to encourage a new feature:
- I would like to see visually which routes the trains of the other companies took the last run
- This would be really helpful in async games
- You could show an icon of each company in the map tab and upon selection it would show the lines where the trains went in the last run

jenf:
The new map enhancements give me a way to approach this.

Issue #1250, opened 2020-07-29
Allow viewing of corporations routes and payout on the map


PedroRCSequeira:
It's nice to know how much a company paid last time  when buying shares and so on, but maybe the company has a new train, or 2 of it's trains rusted.
If it showed what we had on the last run it would help figure this stuff out (without having to look at the log
Maybe when you click you see it?
Or it is just displayed after the run value $$

Another useful thing may be to see the runs. Maybe we could combine both

You click on a company, and then on the map, and you see the same info as when you are running it. Which trains ran, and where

It would tell you everything I talked about before, *and* it would make it easier to know where to block someone *insert devil smile*

Issue #563, opened 2020-05-29
Leave companies previous run on the map during the operating round


jenf:
It'd be useful when reviewing other players turns and when on your own to see how they ran their companies on the map.

(This needs the last route improvements i'm currently working on)

jenf:
I'm going to expand this, outside of the corps turn you should be able to use the map to see their last route.

PedroRCSequeira:
as a "quick look" version, I believe it could be useful to see, in the entities tab, which trans ran last time, so it would have something like: Last Run: $200 (2, 2, 3/5)

perwestling:
A suggestion that came up in a Slack channel (dfan, I believe) was that you should be able to click on the Last run on the corporation, and this would display that run.

PR #1785, opened 2020-10-05
Show route during dividend step

daniel-sousa-me:
A frequently requested feature is the ability to see the route of a company on the map. There are numerous suggestions of ways of doing this, but I think the easiest one to implement is to show the route during the dividend step.

This should (at least partially) address issues: #1611, #1250 ant #563

### DIFF
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -8,10 +8,14 @@ module View
     class Dividend < Snabberb::Component
       include Actionable
 
+      needs :routes, store: true, default: []
+
       def render
         @step = @game.active_step
 
         options = @step.dividend_options(@step.current_entity)
+
+        store(:routes, @step.routes, skip: true)
 
         payout_options = @step.dividend_types.map do |type|
           option = options[type]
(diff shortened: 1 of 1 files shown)


assets/app/view/game/dividend.rb
```diff
@@ -52,6 +57,9 @@ def render
             margin: '0.5rem 0 0 0',
             textAlign: 'left',
           },
+          hook: {
```
tobymao, line 60: in order to have a hook you need a key, look at view/chat.rb as an example

daniel-sousa-me: Is the hook needed?
I don't know what it does. I just copy-pasted code from route_selector.rb

tobymao: i don't know, what happens when you remove it?

daniel-sousa-me: lol

It works without the hook on the cases I tried. But I assume it was there because maybe some combination of clicks may remove the view, but not advance the action in the "normal way". Then the function `cleanup` would not be called and the storage of the `:routes` thingy might conflict with something else somewhere else.

daniel-sousa-me: I went back and found out that you added the hook in this commit: https://github.com/tobymao/18xx/commit/cfe18c198fe0f014356bea8c54f080ec439ff4fd#diff-c0fbc2481cab263a628acff8e00c30eb

So my guess is that it would cause problems with the Undo.
Can I just put whatever string I want as key? If so, I already edited the PR.

tobymao (approved):

PR #1801, opened 2020-10-08
Allows routes to carry over properly when upgrading a simple tile to …

### LINKED ISSUE #1797: Previous routes lost when upgrading to lawson track
When upgrading from a non-lawson to a lawson track (most notably any yellow track to a lawson green track) any existing train routes through that tile are lost. I understand this is due to adding the node in the middle to allow for the lawson pathing but felt it should at least be noted in case someone finds a way to preserve the route.

roseundy:
…a Lawson tile

Fixes #1797 

Since I was mucking around with Connection::id, I simplified it to take advantage of the fact that route paths are always in order and start and end at nodes.

### DIFF
--- a/lib/engine/connection.rb
+++ b/lib/engine/connection.rb
@@ -89,16 +89,13 @@ module Engine
       @id ||=
         begin
           sorted = []
-          path_map = {}
-          node_path = nil
+          junction_map = {}
 
+          # skip over paths that have a junction we've already seen
           @paths.each do |path|
-            node_path = path if path.node?
-            path_map[path] = true
-          end
-
-          node_path.walk(on: path_map) do |path|
-            sorted << path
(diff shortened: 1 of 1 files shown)


lib/engine/connection.rb
```diff
@@ -89,16 +89,13 @@ def id
...
 
+          # skip over paths that have a junction we've already seen
           @paths.each do |path|
-            node_path = path if path.node?
-            path_map[path] = true
-          end
-
-          node_path.walk(on: path_map) do |path|
-            sorted << path
+            sorted << path unless junction_map[path.a] || junction_map[path.b]
```
tobymao, line 96: don't use unless with multiple statements, switch it an if statement

roseundy: Fixed.

tobymao (approved):

PR #1803, opened 2020-10-08
18CO Base Files

ryandriskel:
I was advised to submit a PR before I got too deep into implementation. This code adds all the assets for 18CO: map, market, corporations, trains, privates (sans some abilities) and tiles. It also adds some basic game functionality to allow it to be playable ( though with incorrect rules ).

There are a few files within the step directory that are not yet referenced by the module. These will be utilized in future PRs where their implementation actually occurs. I didn't feel it was worth the time to remove them from this PR. I will comment on those files below.

Files run through Rubocop and all style problems have been fixed.

If submitting a PR via a fork is not the preferred method, please let me know. It was requested on the last project where I contributed.

Note: The doubling trains seem ripe for a refactor. I plan to look into that later.


lib/engine/step/g_18_co/track.rb
```diff
@@ -0,0 +1,20 @@
...
+require_relative '../track'
+
+module Engine
+  module Step
+    module G18CO
+      class Track < Track
+        def process_lay_tile(action)
+          lay_tile_action(action)
+
+          # TODO: Implement Mine Collection
```
ryandriskel, line 12: I expect to need this file later as when laying a yellow tile on a hex containing a mine, the corporation receives the mine token. Each mine token adds 10 to its treasury after every run. This collection is not yet implemented, so this is a placeholder.

lib/engine/step/g_18_co/mine_income.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
```
ryandriskel, line 1: I am unsure if adding a new step for adding the mine income to the treasury ( post running trains ) is the proper solution as there is no real user interaction. The alternative would seem to be to add that logic to the dividend step.

lib/engine/step/g_18_co/dividend.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
ryandriskel, line 1: This file is not yet used. It will contain necessary changes to allow for multiple jumps and potentially the payout to the corporations owning shares.

lib/engine/step/g_18_co/company_pending_par.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
ryandriskel, line 1: This file is not yet used. It is intended to contain the implementation of the "par groups" as certain corporations can only be parred in certain sections of the par choices.

lib/engine/config/game/g_18_co-full_b_group.json
```diff
@@ -0,0 +1,48 @@
+{
```
ryandriskel, line 1: This file could be removed. It contains the full details of the actual 18CO private companies. Only one company from "group B" is added to the game. I was able to accomplish this with a single existing private / ability combo. We lose out on the flavor though.

tobymao: please delete this file

lib/engine/config/game/g_18_co.rb
```diff
@@ -0,0 +1,1225 @@
+# frozen_string_literal: true
+
+# rubocop:disable Lint/RedundantCopDisableDirective, Layout/LineLength, Layout/HeredocIndentation
+
+module Engine
+  module Config
+    module Game
+      module G18CO
+        JSON = <<-'DATA'
+{
```
tobymao, line 10: can you pas sthis through a json formatter?

http://jsonviewer.stack.hu/

lib/engine/game/g_18_co.rb
```diff
@@ -0,0 +1,133 @@
...
+require_relative '../config/game/g_18_co'
+require_relative 'base'
+require_relative 'company_price_50_to_150_percent'
+require_relative 'revenue_4d'
+require_relative 'revenue_5d'
+
+module Engine
+  module Game
+    class G18CO < Base
+      register_colors(green: '#237333',
```
tobymao, line 12: can you run an auto indent on this file?

lib/engine/game/g_18_co.rb
```diff
@@ -0,0 +1,133 @@
...
+        revenue
+      end
+
+      def routes_revenue(routes)
+        total_revenue = super
+        # TODO: East/SLC Bonus
+        total_revenue
+      end
+
+      def get_location_name(hex_name)
```
tobymao, line 98: this is copied move it to the base class

lib/engine/game/g_18_co.rb
```diff
@@ -0,0 +1,133 @@
...
+        @hexes.find { |h| h.name == hex_name }.location_name
+      end
+
+      def upgrades_to?(from, to, special = false)
```
tobymao, line 102: delete this

lib/engine/game/revenue_5d.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+# This module can be included and then called from
+# the revenue_for method in a subclass to the Engine::Game class.
+#
+# It will double the revenue for all cities and off-board
+# hexes in the route.
+#
+# This is used in any games that uses 5D trains, e.g. 18CO.
+#
+module Revenue5D
```
tobymao, line 12: it's unnecessary to make this a module

lib/engine/step/g_18_co/mine_income.rb
tobymao: correct, you don't need a step, just do it in the dividend

lib/engine/game/g_18_co.rb
ryandriskel: I'll just remove it. It was copy pasta from getting started and I don't know if I need it.

lib/engine/game/g_18_co.rb
ryandriskel: Hmm, I'm not getting any differences.

lib/engine/game/g_18_co.rb
```diff
@@ -0,0 +1,130 @@
...
+                      white: '#FFFFFF')
+      load_from_json(Config::Game::G18CO::JSON)
+      AXES = { x: :number, y: :letter }.freeze
+
+      # DEV_STAGE = :beta
+
+      GAME_LOCATION = 'Colorado, USA'
+      GAME_RULES_URL = 'https://drive.google.com/open?id=0B3lRHMrbLMG_eEp4elBZZ0toYnM'
+      GAME_DESIGNER = 'R. Ryan Driskel'
+      GAME_INFO_URL = 'https://boardgamegeek.com/thread/1526930/18co-rock-stock-playtesting-files-available'
```
perwestling, line 30: I think you can make the GAME_INFO_URL as the other games, that is to the wiki page
(18MEX: https://github.com/tobymao/18xx/wiki/18MEX)
and from the wiki page you can possibly have the playtesting link.

lib/engine/game/g_18_co.rb
```diff
@@ -0,0 +1,130 @@
...
+      end
+
+      def stock_round
+        Round::Stock.new(self, [
+        Step::DiscardTrain,
+        Step::BuySellParShares,
+        ])
+      end
+
+      def adjust_revenue_for_5d_train(route, stops, revenue)
```
perwestling, line 82: I think it could be possible to refactor the 4D revenue thing to cover for both 4D and 5D. There are a couple of tiles that use these 4D (18AL, 18MEX, 18GA?) and then 18MS has another type of 4D.

lib/engine/config/game/g_18_co.rb
```diff
@@ -0,0 +1,1260 @@
...
+		"rows": "numbers",
+		"columns": "letters"
+	},
+	"mustSellInBlocks": false,
+	"locationNames": {
+		"A11": "Laramie, WY",
+		"A17": "Cheyenne, WY",
+		"B10": "Walden",
+		"B22": "Sterling",
+		"B26": "Lincoln, NE (SLC +100)",
```
perwestling, line 39: Wonder if this makes the map messy? An alternative could be to add an ability to SLC that gives this bonus.
(Hm, I do not know what SLC is....)

lib/engine/config/game/g_18_co.rb
```diff
@@ -0,0 +1,1260 @@
...
+			"abilities": [
+
+			]
+		},
+		{
+			"sym": "GJGR",
+			"name": "Grand Junction and Grand River Valley Railway",
+			"value": 40,
+			"revenue": 10,
+			"desc": "An owning Corporation may upgrade a yellow town to a green city in additional to its normal tile lay. Action closes the company or closes on purchase of “5” train.",
```
perwestling, line 337: Check if there is a way to express via an ability? As it is now, the use will just remove the ability. But doing it in code is OK.

lib/engine/config/game/g_18_co.rb
```diff
@@ -0,0 +1,1260 @@
...
+			"sym": "Toll",
+			"name": "Saguache & San Juan Toll Road Company",
+			"value": 60,
+			"revenue": 10,
+			"desc": "An owning Corporation receives a $20 discount on the cost of tile lays. Closes on purchase of “5” train.",
+			"abilities": [
+				{
+					"type": "tile_discount",
+					"discount": 20,
+					"terrain": "mountain"
```
perwestling, line 397: Is it only mountain costs? If so maybe the description needs to be updated.

perwestling (commented):
Just some minor comments.

lib/engine/game/g_18_co.rb
crericha: There is a 3+D in 1828 that is similar too. The only difference is it can also run infinite towns before, in the middle, and after the 3 cities.

perwestling: The 4D trains of 18AL/18MEX is like that. So naming them with + is a good thing. (Ie 18MEX and 18AL should probably have been called 4+D, and the normal trains in 18MEX/18AL should have been called 2+, 3+, ...)

But the naming of trains is a bit complex to make general as the "+" can have different meaning in different games.

Anyway I think it could be possible to refactor this to have reusable code for all games that use these types of Diesels, alternatively we could enhance the configuration to make it possible to specify what kind of stops are counted double? In the latter case we would not have to do any special code for that. I will submit an enhancement ticket with this suggestion.

lib/engine/config/game/g_18_co.rb
ryandriskel: Salt Lake City. I was following the example set by 1882; ( HB +100 ) on the two hexes signifying that a connection to Hudson Bay gives a +100 bonus.

lib/engine/game/g_18_co.rb
ryandriskel: I fully support a refactor to allow specifying a multiplier on stop type as indicated in my original PR text. Do we wish to pursue that ahead of this PR, or loop back on it later?

lib/engine/config/game/g_18_co.rb
ryandriskel: I see your point. I suppose I wrote it that way because the only kind of tile cost in the game is mountainous terrain, but it should be removed.

lib/engine/config/game/g_18_co.rb
ryandriskel: This is one of those pieces I was in the middle of working when Jen suggested I get a PR up before I get too deep. I haven't actually had a chance to test this bit yet. Should I just remove the ability completely and leave it for a later PR?

Side question: Why would it just remove the ability? I read it as "Lay a tile of 14 or 15 in one of the following hexes" ... I know that isn't quite accurate to the ability, but it seemed it should have done something close to what I wanted.

lib/engine/config/game/g_18_co.rb
ryandriskel: @perwestling Terrain type is a requirement of this ability, and mountain is the only type of terrain in 18CO.

lib/engine/game/g_18_co.rb
ryandriskel: I have done this.

lib/engine/config/game/g_18_co.rb
ryandriskel: Here is what it looks like on the map.

Screen Shot 2020-10-12 at 5 54 02 PM

tobymao (approved):

PR #1812, opened 2020-10-11
Initial commit for 18xx.game's 1828 implementation.

crericha:


tobymao:
you need to add an override for title or change the class names

crericha:
I'm interested in your thoughts here. My current thoughts o the topic is use the wiki for the full name, reference to JC and 1828, and have a rules section listing the visual and naming differences that matter when reading the rulebook and a link to the 1828 rulebook. Then on the site for the game dropdown list and source files, we'd use the shortened name of 1828.. The rationale is I don't think we'd list 1830 as 1830: Railways & Robber Barons, so its okay to use the short name there.

tobymao:
i don't want  there to be any confusion, as we do this for the 1846 2 player variant

crericha:
Changed name to 1828.Games and added implementer note to address.

crericha:
Heads up that I already had a couple of other commits ready to go, so they are included here too.

tobymao (approved):

PR #1819, opened 2020-10-11
Filter out illegal paths during run_routes

### LINKED ISSUE #1125: Don't suggest invalid routes when valid ones exist
Game 1306, OR7.1 (it's an old game but I don't think it is pinned)

As SRR, I started my route by clicking on Ohio, then Green Spring, then Washington DC, then Strasburg. There are two paths from DC to Strasburg, but one of them would reuse existing track and one wouldn't. The site gave me the former illegal path by default. Of course I could click again to get the legal one, but it would be nice for illegal ones to be filtered out if some other ones are legal.

roseundy:
Fixes #1125 

Modify behavior of Route.touch_node to filter out connections that use paths already in the partially completed route being built. Unlike the last time I tried this - I specifically do _not_ filter out the paths in the most recently added connection so that clicking on the last node in the connection list will correctly switch between alternatives (but now, only legal alternatives).

For example, when selecting this route, the existing code initially creates an illegal route:
Before:
OldRouting

After:
NewRouting

### DIFF
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -38,13 +38,15 @@ module Engine
     end
 
     def next_connection(node, connection, other)
-      connections = select(node, other)
+      connections = select(node, other, connection)
       index = connections.find_index(connection) || connections.size
       connections[index + 1]
     end
 
-    def select(node, other)
+    def select(node, other, keep = nil)
       other_paths = @routes.reject { |r| r == self }.flat_map(&:paths)
+      other_paths.concat(@connections.reject { |c| c[:connection] == keep }
+                          .flat_map { |c| c[:connection].paths })
       nodes = [node, other]
 
       node.hex.all_connections.select do |c|
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -50,9 +50,7 @@ module Engine
       nodes = [node, other]
 
       node.hex.all_connections.select do |c|
-        (c.nodes & nodes).size == 2 &&
-          !@connections.include?(c) &&
-          (c.paths & other_paths).empty?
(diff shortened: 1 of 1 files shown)


lib/engine/route.rb
```diff
@@ -38,13 +38,15 @@ def connections
...
+      connections = select(node, other, connection)
       index = connections.find_index(connection) || connections.size
       connections[index + 1]
     end
 
-    def select(node, other)
+    def select(node, other, keep = nil)
       other_paths = @routes.reject { |r| r == self }.flat_map(&:paths)
+      other_paths.concat(@connections.reject { |c| c[:connection] == keep }
+                          .flat_map { |c| c[:connection].paths })
```
tobymao, line 49: i don't understand this code

c[:connection] is a connection object, how can that every equal an array of paths?

also line 52 !@connection.include?(c) looks like it's a bug, can you fix that?

roseundy: I'm filtering out the connection that matches keep first, then calling flat_map on all of the paths in the rest of the connections in @connections.

roseundy: Pretty sure if line 52 worked as intended, it would break being able to cycle paths. The intent, though is to do exactly what my PR is trying to do.  I need to think about whether is is a better way.

roseundy: Scratch that. I don't know exactly what the intent was, but if I fix it to actually work it breaks cycling paths. It's preventing from selecting a connection already in the connection list, which is OK, except when you'd like to cycle paths via "next_connection". My PR will do what the intent was (except cycling works). So I'm just tossing the old code.

tobymao (approved):

PR #1820, opened 2020-10-12
18 Los Angeles - allow Freight and Passenger trains to use the same track

michaeljb:
only trains in the same overlap group are not allowed to reuse track; this
allows for 18 Los Angeles's Freight and Passenger trains to work correctly

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -661,6 +661,28 @@ module Engine
         routes.sum(&:revenue)
       end
 
+      def compute_other_paths(routes, route)
+        routes.reject { |r| r == route }.flat_map(&:paths)
+      end
+
+      def check_overlap(routes)
+        tracks = []
+
+        routes.each do |route|
+          route.paths.each do |path|
+            a = path.a
+            b = path.b
+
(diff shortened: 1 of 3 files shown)


tobymao:
@roseundy is touching this bit right now so this conflicts. i wonder if it's worth adding this concept just for 18LA. does this exist anywhere else?

if it's doesn't exist perhaps, we could use the technique where game computes the function so it exists as just one override

michaeljb:
I don't know of any others that do it this way, I can probably move it into the game to keep it to one override

michaeljb:
@tobymao I've pushed a fix that moves some logic from the route class into the game base class. I'm happy to wait for #1819 to be merged first and then I could fix the conflict before merging this PR.

tobymao:
sounds good

roseundy:
1862 also allows different trains to reuse the same track

roseundy:
FYI, I think that you can leave the code I added to route.rb (i.e. the line that concatenates onto other_routes) in route.rb. The line I added just makes sure the route being built doesn't reuse any paths within itself. It should affect the ability to re-use paths _between_ routes (trains).

PR #1824, opened 2020-10-12
18MS: Add line to P1/P2 description about OR 1 buy-in.

perwestling:
It seems some players do forget to use this feature so adding it
should hopefully work as a reminder.

### DIFF
--- a/lib/engine/config/game/g_18_mex.rb
+++ b/lib/engine/config/game/g_18_mex.rb
@@ -252,7 +252,7 @@ module Engine
          "name":"Kansas City, Mexico, & Orient Railroad",
          "value":40,
          "revenue":10,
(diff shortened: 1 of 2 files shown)


perwestling:
@scottredracecar The added lines are not in the physical game. I think it might be a good idea as players seem to miss out on that possibility. Also I am wondering about the "Major Company". The 18xx.games uses corporation for "major company" but the descriptions of the privates do use the same text in the physical game. So I used "Major Company" in the added text as well.

So do you think this is OK?

With this PR handled, the rules link updated, then I think we can add a PR to make 18MS production state.

perwestling:
I sent a question about it to Mark Derrick as well.

tobymao:
you shouldn't use the word major company, call it a corporation

perwestling:
@tobymao Got thumbs up from designer so I have now made it consistent according to 18xx.games standard.
Also updated 18MEX in the same style.  18AL/18GA/18TN were all ok.

tobymao (approved):

PR #1846, opened 2020-10-13
Add ability for games to have unlimited tiles

### LINKED ISSUE #1843: 1817
There are not enough tiles. Non-lettered tiles are considered unlimited.

jenf:
fixes #1843


dfannius:
Once this is in, it's probably worth making a pass through all existing games looking for tiles that are unlimited in the rules but currently represented with a large number, typically the amount printed in the physical game (e.g., the plain yellow tiles in 1846).

assets/app/view/tiles.rb
```diff
@@ -38,7 +38,11 @@ def render_tile_blocks(
         text = tile.preprinted ? '' : '#'
         text += name
         text += "-#{rotation}" unless rotations == [0]
-        text += " × #{num}" if num
+        if tile.unlimited
+          text += ' x ∞'
```
perwestling, line 42: It looks like the multiplication sign is different sizes? Should they not be the same?

lib/engine/game/base.rb
```diff
@@ -856,6 +856,18 @@ def two_player?
         @two_player ||= @players.size == 2
       end
 
+      def add_extra_tile(tile)
+        game_error('Add extra tile only works if unlimited') unless tile.unlimited
+
+        # Find the highest tile that exists of this type in the tile list and duplicate it.
```
perwestling, line 862: An alternative would be to make the find-max code part of the dup method. When you would guarantee that dup isn't called with a non max tile.

lib/engine/game/base.rb
```diff
@@ -993,17 +1005,18 @@ def init_hexes(companies, corporations)
 
       def init_tiles
         self.class::TILES.flat_map do |name, val|
-          if val.is_a?(Integer)
-            count = val
+          if val.is_a?(Integer) || val == 'unlimited'
+            count = val == 'unlimited' ? 1 : val
```
perwestling, line 1009: Thinking on how readable this construct is, and on line 1019. Would 
 count = (val == 'unlimited' ? 1 : val) 
improve things?

perwestling: Another possibility would be to add a private
  def unlimited?(val)
     val == 'unlimited'
  end
and then use this on line 1009, 1019 and 1029.

perwestling (commented):
Looks good, have just some minor comments.

tobymao (approved):

tobymao:
maybe change the name dup because it's not a dup, you're incrementing the index

PR #1851, opened 2020-10-13
Multiplier Attribute for Train Nodes ( Issue 1814 )

ryandriskel:
Add a new attribute "multiplier" that allows a user to specify an amount the revenue is multiplied by. This allows for simple "double" trains and works for any revenue type, allowing trains to double all type or only specific types. See: 18AL / 18Mex / 18CO


ryandriskel:
If accepted, we could update older titles to use this functionality, but it is not required to get the functionality in the base and I didn't want to mess with titles already in production.

lib/engine/game/base.rb
```diff
@@ -650,7 +650,19 @@ def end_game!
       end
 
       def revenue_for(route, stops)
-        stops.sum { |stop| stop.route_revenue(route.phase, route.train) }
+        stops.sum { |stop| revenue_for_stop(route, stop) }
+      end
+
+      def revenue_for_stop(route, stop)
+        distance = route.train.distance
+        multiplier = 1
```
ryandriskel, line 658: There might be a cleaner way to write this in Ruby. I come from the C/JS/PHP world and am still learning Ruby syntax.

ryandriskel:
Apparently I need to get better at running the test suite. I have to add some additional checks to handle potential missing attributes. Converting to draft.

ryandriskel:
Not sure why I am not doing this in stop.route_revenue -- will make that refactor later today.

ryandriskel:
The new changes allow for a train like

```
{
  "name":"2D",
  "multiplier":2
  "price":200,
  "rusts_on":"4",
  "num":5
},
```

as well as the multiplier within the distance node. This makes simple "double everything" trains possible.

lib/engine/part/revenue_center.rb
```diff
@@ -50,9 +50,25 @@ def max_revenue
...
+
+      def route_base_revenue(phase, train)
+        return (@revenue[:diesel]) if train.name.upcase == 'D' && @revenue[:diesel]
+
+        phase.tiles.reverse.each { |color| return (@revenue[color]) if @revenue[color] }
+      end
+
+      def revenue_multiplier(train)
+        distance = train.distance
+        base_multiplier = train.multiplier.nil? ? 1 : train.multiplier
```
tobymao, line 64: train.multiplier || 1

lib/engine/part/revenue_center.rb
```diff
@@ -50,9 +50,25 @@ def max_revenue
...
+
+        return base_multiplier if distance.is_a?(Numeric)
 
-        phase.tiles.reverse.each { |color| return @revenue[color] if @revenue[color] }
+        row = distance.index do |h|
+          h['nodes'].include?(type)
+        end
+        distance[row].fetch('multiplier', base_multiplier)
```
tobymao, line 71: why is there both a train multiplier and a distance multiplier?

tobymao:
looks good, please refactor everything to use this technique

lib/engine/part/revenue_center.rb
ryandriskel: For a simple train that doubles everything:
```
{
  "name":"2D",
  "multiplier":2
  "price":200,
  "rusts_on":"4",
  "num":5
},
```

For a more complex implementation where only cities/offboards are doubled:
```
{
			"name": "4D",
			"distance": [
				{
					"nodes": [
						"city",
						"offboard"
					],
					"multiplier": 2,
					"pay": 4,
					"visit": 4
				},
				{
					"nodes": [
						"town"
					],
					"pay": 99,
					"visit": 99
				}
			],
			"available_on": "5",
			"price": 650,
			"num": 3
		},
```

The final line ensures that the train multiplier is respected if the multiplier is not explicitly included in the distance section.

ryandriskel:
Hmm ... something very wrong looking has happened to file changes after rebasing to latest.

ryandriskel:
Im going to kill this PR and try recreating to get the proper changed files.

PR #1852, opened 2020-10-13
Fix not adding the new cities to list of cities

jenf:


tobymao (approved):

tobymao:
do you need to do towns and other nodes as well?

PR #1865, opened 2020-10-14
Implement 1828 auction and other minor updates

crericha:


lib/engine/step/g_1828/waterfall_auction.rb
```diff
@@ -0,0 +1,78 @@
...
+    module G1828
+      class WaterfallAuction < WaterfallAuction
+        def may_purchase?(company)
+          return false unless @companies.first.value == company.value && @bids[company].empty?
+
+          rest_of_row_has_bids?(company)
+        end
+
+        def can_auction?(company)
+          return true if @process_round_end_auction && @bids[company].count > 1
```
tobymao, line 16: prefer to use size instead of size

tobymao: count

lib/engine/step/g_1828/waterfall_auction.rb
```diff
@@ -0,0 +1,78 @@
...
+            end
+
+            round_end_auction_complete if all_bids_processed?
+          else
+            super
+          end
+        end
+
+        def resolve_bids_for_company(company)
+          return false unless @process_round_end_auction || rest_of_row_has_bids?(company)
```
tobymao, line 37: don't use unless for multiple statements

lib/engine/step/g_1828/waterfall_auction.rb
```diff
@@ -0,0 +1,78 @@
...
+
+          entities.each(&:unpass!)
+        end
+
+        def all_bids_processed?
+          @bids.values.flatten.empty?
+        end
+
+        def rest_of_row_has_bids?(company)
+          @companies.find_all { |c| c.value == company.value }.each do |c|
```
tobymao, line 69: what is find_all? is this the same as select? if so use select

lib/engine/step/waterfall_auction.rb
```diff
@@ -66,14 +68,14 @@ def actions(entity)
 
       def setup
         setup_auction
-        @companies = @game.companies.sort
+        @companies = @game.companies.sort_by(&:value)
```
tobymao, line 71: is this correct for all games? seems like company line 31 sorts by min_bid, name

lib/engine/step/waterfall_auction.rb
```diff
@@ -98,7 +100,39 @@ def max_bid(player, company)
...
+          @log << "#{@auctioning.name} goes up for auction"
+        end
+
+        resolved
+      end
+
+      def active_auction
+        company = @auctioning
+        bids = @bids[company]
+        yield company, bids if bids.count > 1
```
tobymao, line 130: size

lib/engine/step/waterfall_auction.rb
```diff
@@ -98,7 +100,39 @@ def max_bid(player, company)
...
+      end
+
+      def can_auction?(company)
+        company == @companies.first && @bids[company].count > 1
```
tobymao, line 134: size

lib/engine/step/waterfall_auction.rb
crericha: Micheal recently added the <=> operator to company for 1846 2P, so no impact to the waterfall auction games. In my opinion, the company class shouldn't dictate sorting for auctions. Each auction type is different with different needs.

Looking at his commit. I did remove a sort that he added in the view. I will re-add that.

crericha: I added the sorts to the draft auction and verified they work correctly.

Issue #1814, opened 2020-10-11
Make the doubling feature of diesels a configurable thing


perwestling:
Right now there are diesels in some title that double revenue from stops but with slightly smaller difference

- 4D of 18MEX/18AL/18CO is a 4+ train that double cities and off-boards, but can run an unlimited number of towns
- 5D of 18CO is a similar 5+ train
- 2D/4D/5D of 18MS is a standard 2/4/5 train that double any stops

The "4+D" functionality used in 18MEX/18AL(/18CO) is a module you can include in the game code. It could be refactored to work for any number, in which case it would cover 5D of 18CO as well.

The 2D/4D/5D functionality of 18MS is handled in the 18MS game code.

My suggestion is to enhance the game configuration for trains to support doubling of certain stops.

Here is how the configuration on an "4+D" train is done in 18CO. (Similar in 18AL/18MEX)
```
		{
			"name": "4D",
			"distance": [
				{
					"nodes": [
						"city",
						"offboard"
					],
					"pay": 4,
					"visit": 4
				},
				{
					"nodes": [
						"town"
					],
					"pay": 0,
					"visit": 99
				}
			],
			"available_on": "5",
			"price": 650,
			"num": 3
		},
```
My suggestion is to add a "doubling" which would specify which stops that should be doubled.
It could look like this for the "4+D" train type:
```
		{
			"name": "4D",
			"doubling": [
				{
					"nodes": [
						"city",
						"offboard"
					]
				},
```
Or add as a field in the distance structure:
```
			"distance": [
				{
					"nodes": [
						"city",
						"offboard"
					]
                                        "double":true,
					"pay": 4,
					"visit": 4
				},
				{
					"nodes": [
						"town"
					],
					"pay": 0,
					"visit": 99
				}
			],
```
In all cases if not double specified, it should pay normal.

A question is this should be generalized further? E.g. if the double attribute in the last case should be some general attribute that take ["double"] as value? E.g. "modifier":["double"],

That last comment is this could be used for handling E trains. In 1822 the E trains double all stations with the corporation token in (if I remember correctly) so how should that be expressed?

ryandriskel:
Note: The 4D/5D in 18CO _do not pay towns at all_, unlike 18AL.

I have a slightly different suggestion, a multiplier attribute, assumed 1.

`"distance": [
				{
					"nodes": [
						"city",
						"offboard"
					]
                                        "multiplier":2,
					"pay": 4,
					"visit": 4
				},
				{
[...]

PR #1874, opened 2020-10-14
Issue 1814 Generic train revenue multiplier

ryandriskel:
Add a new attribute "multiplier" that allows a user to specify an amount the revenue is multiplied by. This allows for simple "double" trains and works for any revenue type, allowing trains to double all type or only specific types. See: 18AL / 18Mex / 18CO

Proposed solution for #1814 

For a simple train that doubles everything:
```
{
  "name":"2D",
  "multiplier":2
  "price":200,
  "rusts_on":"4",
  "num":5
},
```
For a more complex implementation where only cities/offboards are doubled:
```
{
  "name": "4D",
  "distance": [
    {
      "nodes": [
        "city",
        "offboard"
      ],
      "multiplier": 2,
      "pay": 4,
      "visit": 4
    },
    {
      "nodes": [
        "town"
      ],
      "pay": 99,
      "visit": 99
    }
  ],
  "available_on": "5",
  "price": 650,
  "num": 3
},
```


perwestling:
Looks a bit strange, commit wise, as there are duplicated commits. I suspect that you are not that used to git....

perwestling:
I would probably have made two commits; one with the general code change, and one with changes for the titles (or maybe one per type of change of the titles). But that's me.

Anyway, I think 18MS could be changed as well. It uses a simple multiplier of 2 for all D trains. And it implements this doubling in g_18_ms.rb. It has 2D, 4D and 5D.

perwestling (commented):
The changes looked fine to me.

ryandriskel:
> I would probably have made two commits; one with the general code change, and one with changes for the titles (or maybe one per type of change of the titles). But that's me.

I had originally submitted the PR with just the base code as its is independent. I then had some trouble with a rebase ( cause for duplicated commits ) that borked the pr https://github.com/tobymao/18xx/pull/1851. Toby asked me to update with the changes to the actual titles there.

perwestling:
@ryandriskel If you want to clean it up you could do the following 

1. git -mv 1814_Generic_Doubling_Trains some_name
   some_name just an example
2. git checkout master
   preparing to get latest master from upstream
3. git fetch upstream
    assuming the remote you forked from is called upstream
4. git reset --hard upstream/master
    now we have same local base as the upstream base
5. git checkout -b 1814_Generic_Doubling_Trains 
    the 1814_Generic_Doubling_Trains will now have the latest upstream master as it base
6. git cherry-pick some_name
   would pick the top commit from your old branch and add it to your new branch
   You can cherry-pick with SHA1 or some_name^ or .... if you want to pick code from other commit(s)
7. git push -f --set-upstream origin 1814_Generic_Doubling_Trains 
   This will overwrite this branch on your fork with the new version, and it will update the PR.

That is one way of cleaning up the commit.

perwestling:
And I would recommend "git rebase -i" if one want to clean up commit history.
E.g.
   git rebase -i upstream/master
To possible do things with all commits since the base of this branch

ryandriskel:
18MS updated and we are now in an alternate timeline.

lib/engine/part/revenue_center.rb
```diff
@@ -50,9 +50,25 @@ def max_revenue
...
+        distance = train.distance
+        base_multiplier = train.multiplier || 1
+
+        return base_multiplier if distance.is_a?(Numeric)
 
-        phase.tiles.reverse.each { |color| return @revenue[color] if @revenue[color] }
+        row = distance.index do |h|
+          h['nodes'].include?(type)
+        end
+        distance[row].fetch('multiplier', base_multiplier)
```
tobymao, line 71: why does the train and the distance both have multiplier?

ryandriskel: The train multiplier allows a basic "multiply all stop types" implementation. See 18MS trains.

The distance multiplier allows for multiplying specific nodes. 18AL/18Mex only double cities.

perwestling:
I think it looks good now.

Issue #1836, opened 2020-10-12
18CO Special Rules Summary


ryandriskel:
This is a summary of the rules for 18CO: Rock & Stock that are not (suspected) currently available on 18xx.games. I haven't played all the games in the works so some of these may be possible. Please comment if so!

### Auction

- [x] A bid on a Private Company may be moved to a different Private Company as long as the bid value is increased by 5, in increments of 5.
- [x] A player may have multiple bids on the same Private Company.
- [x] The auction ends when all players pass consecutively.
- [x] Privates that received no bids are placed in the Market.

### Private Companies

- [x] New Ability: An owning Corporation may return a station token to its charter to gain the token cost. Corporation must always have at least one token on the board.
- [x] New Ability: Money gained from mine tokens is doubled for the owning Corporation.
- [x] Durango​ ​and​ ​Silverton​ ​Narrow​ ​Gauge should start with the money from the Presidency being awarded through the Private Company. It currently floats as expected, but with $0 instead of twice the par value.

### Stock Rounds

**Launching Corporations**
- [x] Par values are grouped into three zones. Each zone has a different minimum number of shares required to float ( 4/5/6).
- [x] Corporations have different classifications that determine the minimum par value zone where they are allowed to launch.
- [ ] A​ ​Corporation​ ​cannot​ ​be​ ​launched​ ​if​ ​its​ ​reserved​ ​station​ ​location​ ​is​ ​not​ ​available,​ ​or​ ​no​ ​tile​ ​upgrade​ ​containing​ ​a​ ​station location​ ​on​ ​that​ ​tile​ ​is​ ​available.

**Share Transactions**
- [x] A player may purchase a third party corporation share from the treasury of any corporation where they are President.
- [x] A single share sold by a non-President does not cause a stock to drop in value.
- [x] The DSNG has share breakdown as follows: 2P,2,2,1,1,1,1

**End of Stock Round Movement**
- [x] Fully owned: If​ ​the​ ​Stock​ ​Token​ ​cannot​ ​move​ ​up,​ ​move​ ​the​ ​Stock​ ​Token​ ​one​ ​space​ ​to​ ​the​ ​right.

**President’s​ ​Choice​ ​Stock​ ​Round**
[...]

PR #1886, opened 2020-10-15
Generic support for Multi-directional Dividend Movement ( 18CO rule )

ryandriskel:

### Running Trains
**Dividends**: Adjust​ ​the​ ​Stock​ ​Token​ ​one​ ​space​ ​right.​ ​If​ ​the​ ​revenue​ ​was​ ​at​ ​least​ ​twice​ ​the​ ​original​ ​Stock​ ​Market​ ​Value​ ​this round,​​ ​move​ ​the​ ​token​ ​one​ ​space​ ​up.

-----

The current code only allows for one direction of movement, as well as only right and left.

My solution is to allow an array to be specified for both share_direction and share_times. This is supported by the changes to **lib/engine/step/dividend.rb** . I have chosen to make this change to the base dividend.rb (rather than 18CO specific) so any future game could take advantage of the capability. I chose to add support for _down_ for completeness. Current integer format still works the same way. Here is a working example of the newly supported format:

`{ share_direction: %i[right up], share_times: [1, 1] }`

The changes to **assets/app/view/game/dividend.rb** allow for displaying the array based information, as seen in this example from my testing:

Screen Shot 2020-10-15 at 3 09 48 PM

### DIFF
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -143,10 +143,8 @@ module Engine
 
         prev = entity.share_price.price
 
-        Array(payout[:share_times]).each_with_index do |share_times, share_times_index|
+        Array(payout[:share_times]).zip(Array(payout[:share_direction])).each do |share_times, direction|
           share_times.times do
-            direction = Array(payout[:share_direction])[share_times_index]
-
             case direction
             when :left
               @game.stock_market.move_left(entity)
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -142,12 +142,23 @@ module Engine
         return unless payout[:share_direction]
 
         prev = entity.share_price.price
-        payout[:share_times].times do
-          case payout[:share_direction]
-          when :left
-            @game.stock_market.move_left(entity)
-          when :right
-            @game.stock_market.move_right(entity)
+
+        share_times = Array(payout[:share_times])
+
+        share_times.each_with_index do |share_time, share_time_index|
+          share_time.times do
+            direction = Array(payout[:share_direction])[share_time_index]
+
+            case direction
+            when :left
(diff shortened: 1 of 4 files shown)


lib/engine/step/dividend.rb
```diff
@@ -142,12 +142,21 @@ def change_share_price(entity, payout)
...
 
         prev = entity.share_price.price
-        payout[:share_times].times do
-          case payout[:share_direction]
-          when :left
-            @game.stock_market.move_left(entity)
-          when :right
-            @game.stock_market.move_right(entity)
+
+        Array(payout[:share_times]).each_with_index do |share_times, share_times_index|
```
tobymao, line 146: what about Array(times).zip(Array(direction)).each do |times, direction)

tobymao:
can you refactor the other games to use this new method?

lib/engine/step/dividend.rb
ryandriskel: I'll check out the zip method.

ryandriskel:
> can you refactor the other games to use this new method?

@tobymao Do any other games use multi-directional stock movement? The code for games like 1846 (left-only multi-jump) is still fully supported. This new code is backwards compatible.

lib/engine/step/dividend.rb
ryandriskel: Yep, that's a bit cleaner, thanks!

tobymao:
i don't think so, but don't other games have double tripple jumps that could be refactored to use this style?

roseundy:
1860 has multi-jumps, but it is on a "zig-zag" (essentially 1D with different rules for price movements between ORs and SRs) market. Since 1862 also has a zig-zag market, I was thinking I would make the general StockMarket handle it. I'm not close to doing that yet.

PR #1889, opened 2020-10-16
Render optional rules when creating a game, implement 18 Los Angeles's optional LA Title company

### LINKED ISSUE #1082: Feature suggestion: It should be possible to select variants of 18xx when creating a game
Some 18xx has variants/optional rules that is of interest to try. 
Example:
- 18AL: Use "hard-rust" instead of "obsoleteness" when buying 7T
- 18MEX: Allow corporations to lay 3 yellow tiles instead of 2 during its first OR

It would be nice if when creating the game it would be possible to select which optional rules (if any) that should be used for the game. The selection could be shown in the description for the game.
It should be possible to select multiple optional rules (each game should check if a combination is invalid).

I see no use case for changing optional rules.

michaeljb:
Taken from @perwestling's branch for #1082. I left off the optional 18AL rules added there to keep this branch smaller. I also added a minor fix so that the optional rules render correctly when swapping between multiplayer and hotseat, as well as changing player count in hotseat mode.

Seems to work just fine with LA Title, and there's an added bonus to making it optional right away--I was worried that games created before LA Title was implemented would break since they would be expecting a different number of companies in the draft, but since it's not included by default, those games will be fine.

I guess we could say this Fixes #1082 and use different issues to track specific optional rules that people want added?

Screenshot from 2020-10-15 23-07-58

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -72,19 +72,17 @@ module View
         val = range.value.to_i
         range.value = (min..max).include?(val) ? val : max
         store(:num_players, range.value.to_i, skip: true)
-
-        if Native(@inputs[:optional_rules_selected])
-          Native(@inputs[:optional_rules_selected]).elm.value = ''
-          @optional_rules_selected = nil
-        end
-        store(:optional_rules_available, @optional_rules_per_title[title])
+        update_optional_rules_store
       end
 
       enforce_range = lambda do
         elm = Native(@inputs[:max_players]).elm
         if elm.value.to_i.positive?
           elm.value = elm.max.to_i unless (elm.min.to_i..elm.max.to_i).include?(elm.value.to_i)
-          store(:num_players, elm.value.to_i) if @mode == :hotseat
+          if @mode == :hotseat
+            store(:num_players, elm.value.to_i)
+            update_optional_rules_store
+          end
         end
       end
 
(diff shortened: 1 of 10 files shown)


perwestling:
Have you tried to having multiple optional rules and selecting, say, 2 out of 3? In the branch I had the problem that even if I selected several only one took effect.

@michaeljb I think it is fine to have the base functionality, and e.g. 18 Los Angeles as a "pilot" title to try optional rules out. I have implemented optional rules for 18AL on a branch but I can add them as a separate PR once this one is on master. Think there was some bug in the "hard rust" 4 train optional rule so I need to test those more before doing the PR.

Also, I have not checked the commit yet, but are the optional rule(s) chosen displayed in the game presentation on the main page? I think this is essential knowledge when you want to select if you should join a game. 
Another thing I thought about is if this should be possible to use as a filter criteria the day when we implement that? Although that might be something we can consider in the future, when filtering is in place...

perwestling:
About the screen shot - maybe None should be added and the default option? Or is it clear enough for the users that deselecting all will mean that no optional rules are used? 
When doing this I were considering if vertical "radio buttons" were better, clearer, and that I might have done if my GUI skills was up to it. (I did work a couple of years with GUIs but that was not in web context but in stand alone GUI applications, which used their own systems, like ESL, QT, and some "home made" ones.)

michaeljb:
I haven’t tested out multiple options yet. 18 Los Angeles does have another optional privates company, I could add a stub for that other company and try it out.

I was imagining a list of checkboxes for the interface, that would make it more clear when something isn’t being used. I’ll put some more work in that direction and then reopen the PR.

tobymao:
let's hold off on variants for now as we're growing new games quickly. when it is done though i'd like to make it a really clean integation

Issue #1888, opened 2020-10-16
[1817] Incorrect dividend when half-paying as a 2-share company.


mtsay:
When a 2-share company generates a revenue of $130 and chooses to half-pay, the dividend should be $65 but I'm seeing $66 (based on the log and game data).

- [complete game log](https://gist.github.com/mtsay/86f4f4d5f394e435df4fecb0e3cc9dc5)
- [highlight](https://gist.github.com/mtsay/86f4f4d5f394e435df4fecb0e3cc9dc5#file-game_log-txt-L146) from the same log.
- [complete game data](https://gist.github.com/mtsay/3a5423df9462a224f2c011f8773a12f0)

mtsay:
I also ran a quick-check on half-paying as 5-share companies, and they appear to be correct.

mtsay:
I'm no Ruby programmer nor am I familiar with the codebase, so bear with me on the following investigation.

The issue appears to be from the generic half-pay logic at [this line](https://github.com/tobymao/18xx/blob/69a1fbe72d94f3f1d20b593f9f4c9d94542afc4b/lib/engine/step/half_pay.rb#L7).

The withhold amount is correct for 5-share companies and 10-share companies in 1817, but should simply be `revenue / 2` for 2-share companies.

Searching through the code base, I've found the following:

- the rest of dividend logic looks fine as long as the withhold amount is correct.
  - we go through the `payout_per_share` [logic](https://github.com/tobymao/18xx/blob/69a1fbe72d94f3f1d20b593f9f4c9d94542afc4b/lib/engine/step/dividend.rb#L116). Here I'm assuming the variable `per_share` can hold a value like `32.5` with no issues.
- this half-pay logic is used for 1817, 1846, and 18EU.
  - 1846 doesn't use `total_shares` nor `share_percent`.
  - I'm not familiar with 18EU, but I'm assuming the current logic is correct.

My suggestion is have a 1817-specific logic for half-paying, say extending from `HalfPay`. The 1817-specific logic simply sets `withheld` to `revenue / 2` when `entity.total_shares` is 2.

That being said, I don't have a dev environment nor tests, so I'm going to need some dev's help on this.

mtsay:
[PR-1890](https://github.com/tobymao/18xx/pull/1890) is an attempt to address this issue.

mtsay:
[PR-1890](https://github.com/tobymao/18xx/pull/1890) is merged and I can confirm this issue is no longer occurring on the site.

PR #1890, opened 2020-10-16
Issue 1888 Half-pay dividend for 2-share companies.

mtsay:
Half-paying as 2-share companies in 1817 is off as reported in [issue 1888](https://github.com/tobymao/18xx/issues/1888). This PR updates dividend logic to support this case.

I can think of 2 sensible approaches:
- extend 1817 dividend logic to accommodate half-paying as 2-share companies.
- update half-paying logic to work for 2-share companies in 1817.

While private companies and minor companies are common in 18xx games, I cannot think of another title, 18USA excluded, with 2-share companies. The key questions for choosing between the above options are:
- how likely are we to add another title with variable number of shares?
- when that day comes, are we happier refactoring 1817 logic for the new title (downside of first approach) or branching out 2 different logic for 2 different titles (downside of second approach).

For the time being, this PR implements the second approach. And I will happily revise it to take the first approach if we think the first approach is better.

### DIFF
--- a/lib/engine/step/g_1817/dividend.rb
+++ b/lib/engine/step/g_1817/dividend.rb
@@ -10,6 +10,10 @@ module Engine
         DIVIDEND_TYPES = %i[payout half withhold].freeze
         include HalfPay
 
+        def half_pay_withhold_amount(entity, revenue)
+          entity.total_shares == 2 ? revenue / 2 : super
+        end
+
         def share_price_change(entity, revenue = 0)
           price = entity.share_price.price
           return { share_direction: :left, share_times: 1 } unless revenue.positive?
--- a/lib/engine/step/half_pay.rb
+++ b/lib/engine/step/half_pay.rb
@@ -4,8 +4,12 @@ module Engine
   module Step
     module HalfPay
       def half(entity, revenue)
-        withheld = (revenue / 2 / entity.total_shares).to_i * entity.total_shares
+        withheld = half_pay_withhold_amount(entity, revenue)
         { corporation: withheld, per_share: payout_per_share(entity, revenue - withheld) }
+      end
+
+      def half_pay_withhold_amount(entity, revenue)
+        (revenue / 2 / entity.total_shares).to_i * entity.total_shares
       end
     end
   end


mtsay:
I've successfully built a local dev environment, manually tested the simple case, and verified that when a 2-share company generates $70 in revenue and half pays, $35 dividend is paid to the president.

- [game log](https://gist.github.com/mtsay/129019bdf851c2f2a9974b30667dcc36#file-game_log-txt-L88)
- [game data](https://gist.github.com/mtsay/f290e69e0da672c8a93736783b196e39)
- there is a syntax error when I ran RuboCop:
Screen Shot 2020-10-16 at 12 24 17 AM
it looks like RuboCop succeeded anyway and the error isn't related to my changes.
- I'm currently working on making `stackprof.dump` available.

lib/engine/step/half_pay.rb
```diff
@@ -4,9 +4,13 @@ module Engine
   module Step
     module HalfPay
       def half(entity, revenue)
-        withheld = (revenue / 2 / entity.total_shares).to_i * entity.total_shares
+        withheld = half_pay_withhold_amount(entity, revenue)
         { corporation: withheld, per_share: payout_per_share(entity, revenue - withheld) }
       end
+
+      def half_pay_withhold_amount(entity, revenue)
```
mtsay, line 11: I'm not sure if adding a non-publicly-exposed method in a module in Ruby is kosher.

lib/engine/step/g_1817/dividend.rb
```diff
@@ -10,6 +10,10 @@ class Dividend < Dividend
         DIVIDEND_TYPES = %i[payout half withhold].freeze
         include HalfPay
 
+        def half_pay_withhold_amount(entity, revenue)
+          entity.total_shares == 2 ? revenue / 2 : super
```
mtsay, line 14: override the half-pay logic for 1817 only.

mtsay (commented):
Please review.

tobymao (approved):

PR #1894, opened 2020-10-16
Add master mode to log ( Issue 509 )

### LINKED ISSUE #509: Identify on log that some action was made by master mode and who did it
Identify on the log that some action was made by master mode and who did it, currently looks like the own person made.

ryandriskel:
Resolves #509 

This is a normal action
```
{
  "type": "bid",
  "entity": "Vanish",
  "entity_type": "player",
  "id": 2,
  "company": "O&I",
  "price": 40
},
```

This an action when taken with master mode and it isn't your turn
```
{
  "type": "bid",
  "entity": "Garramon",
  "entity_type": "player",
  "id": 3,
  "master_user": "Vanish",
  "company": "C&WI",
  "price": 60
},
```
_Note: Currently, the value of master_user is the user['name'] following the pattern set by entity. It could be the user['id'], with the name fetched at display time. That might be a better implementation for the future. Thoughts?_

Sample of log output
Screen Shot 2020-10-16 at 1 05 38 PM

**lib/engine/action/base.rb** - I don't love the code here. I feel like there is probably a cleaner way to implement within the other methods of the action class. However, subclasses directly implement _self.h_to_args_  without using super and thus any changes to those functions would require refactoring many classes. Maybe that refactor is still the way to go?

### DIFF
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -34,15 +34,12 @@ module View
         if !hotseat &&
             !action.free? &&
             participant &&
-            !@game.active_players.map(&:name).include?(@user['name']) &&
-            !Lib::Storage[@game.id]&.dig('master_mode')
-          return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
-        end
-
-        if !hotseat &&
-          Lib::Storage[@game.id]&.dig('master_mode') &&
-          !@game.active_players.map(&:name).include?(@user['name'])
-          action.master_user = @user['name']
+            !@game.active_players.map(&:name).include?(@user['name'])
+            if !Lib::Storage[@game.id]&.dig('master_mode')
+              return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
+            else
+              action.master_user = @user['name']
+            end
         end
 
         game = @game.process_action(action)
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -39,6 +39,12 @@ module View
           return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
         end
 
+        if !hotseat &&
+          Lib::Storage[@game.id]&.dig('master_mode') &&
+          !@game.active_players.map(&:name).include?(@user['name'])
+          action.master_user = @user['name']
+        end
+
         game = @game.process_action(action)
         @game_data[:actions] << action.to_h
         store(:game_data, @game_data, skip: true)
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -40,7 +40,7 @@ module View
             return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
           end
 
-          action.master_user = @user['name']
+          action.user = @user['name']
         end
 
         game = @game.process_action(action)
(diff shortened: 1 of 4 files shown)


assets/app/view/game/actionable.rb
```diff
@@ -32,11 +32,15 @@ def process_action(action)
...
-          return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
+          !action.free? &&
+          participant &&
+          !@game.active_players.map(&:name).include?(@user['name'])
+
+          unless Lib::Storage[@game.id]&.dig('master_mode')
+            return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
+          end
+
+          action.master_user = @user['name']
```
tobymao, line 43: i don't think this is needed, we already know which user made every action in the database

lib/engine/action/base.rb
```diff
@@ -8,11 +8,13 @@ class Base
       include Helper::Type
 
       attr_reader :entity
-      attr_accessor :id
+      attr_accessor :id, :master_user
```
tobymao, line 11: just call it :user

lib/engine/action/base.rb
```diff
@@ -8,11 +8,13 @@ class Base
...
 
       def self.from_h(h, game)
         entity = game.get(h['entity_type'], h['entity'])
-        new(entity, **h_to_args(h, game))
+        obj = new(entity, **h_to_args(h, game))
+        obj.master_user = h['master_user'] unless h['master_user'].nil?
```
tobymao, line 16: if h['user']

lib/engine/action/base.rb
```diff
@@ -37,6 +39,7 @@ def to_h
           'entity' => entity.id,
           'entity_type' => type_s(entity),
           'id' => @id,
+          'master_user' => @master_user,
```
tobymao, line 42: user

basically, we need to check that the user is not the same as the player

lib/engine/game/base.rb
```diff
@@ -449,6 +449,8 @@ def process_action(action)
           return clone(@actions)
         end
 
+        @log << "• Action(#{action.type}) via Master Mode by #{action.master_user}:" unless action.master_user.nil?
```
tobymao, line 452: if action.user

lib/engine/action/base.rb
```diff
@@ -8,11 +8,13 @@ class Base
...
 
       attr_reader :entity
-      attr_accessor :id
+      attr_accessor :id, :user
 
       def self.from_h(h, game)
         entity = game.get(h['entity_type'], h['entity'])
-        new(entity, **h_to_args(h, game))
+        obj = new(entity, **h_to_args(h, game))
+        obj.user = h['user'] unless h['user'] == entity.owner.name
```
tobymao, line 16: it needs to be entity.player 

entity.owner could be a corporation

assets/app/view/game/actionable.rb
ryandriskel: Without this code, the user that made the master mode call will not see the master mode log until new data is returned from the server (either refresh or other player action), as _@game.process_action_ (line 42 of actionable) is run on the client, where there is no user data yet.

It is one line to add the user data server side, and even with this code I would suggest adding it as it is a more "reliable" method for determining if it is a master mode action, but I don't see how we can do without the above code unless we just don't care about the acting user see the log.

I've added that line to the PR, but I am interested if there is something obvious I am missing.

tobymao: action.master_user needs to change to user

ryandriskel: Did I convince you we should keep this then? :D

tobymao: yes it's fine, thanks for the explanation

PR #1902, opened 2020-10-17
Added invite link to new games

daniel-sousa-me:
When adding ?invite=73 to the url, if game 73 is new, it will appear above all others on the home page with a message inviting the user to join. Furthermore now the owner sees an "Invite" button on their new games.

Up next: unlisted games

### DIFF
--- a/api.rb
+++ b/api.rb
@@ -67,7 +67,7 @@ class Api < Roda
   use Rack::Deflater unless PRODUCTION
 
   STANDARD_ROUTES = %w[
-    / about hotseat login map new_game profile signup tiles tutorial forgot reset
+    / about hotseat login map new_game invite profile signup tiles tutorial forgot reset
   ].freeze
 
   Dir['./routes/*'].sort.each { |file| require file }
@@ -113,7 +113,8 @@ class Api < Roda
   end
 
   def render_with_games
-    render(pin: request.params['pin'], games: Game.home_games(user, **request.params).map(&:to_h))
+    invite = request.params['invite'].to_i if request.params['invite']
+    render(pin: request.params['pin'], games: Game.home_games(user, **request.params).map(&:to_h), invite: invite)
   end
 
   def render(**needs)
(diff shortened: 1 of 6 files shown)


api.rb
```diff
@@ -113,7 +113,8 @@ class Api < Roda
   end
 
   def render_with_games
-    render(pin: request.params['pin'], games: Game.home_games(user, **request.params).map(&:to_h))
+    invite = request.params['invite'].to_i if request.params['invite']
+    render(pin: request.params['pin'], games: Game.home_games(user, **request.params).map(&:to_h), invite: invite)
```
tobymao, line 117: let's just add all params  so remove invite and pin and do render(**request.params)

tobymao: or instead, can we just create an actual invite route and not use params? 

so it should be /invite/1

models/game.rb
```diff
@@ -55,10 +55,23 @@ class Game < Base
     ) personal_games
   SQL
 
+  INVITE_QUERY = <<~SQL
```
tobymao, line 58: i don't like how we do a full query here

maybe instead, you should make an entirely new page dedicated to viewing unstarted games like in rolling stock.net

daniel-sousa-me: Well, that was my initial idea, but I was having trouble figuring out how to grab the game data in that page, so I fell back to this one.

Meanwhile I have learned a bit more about how everything is rendered, so I'll try that way again

daniel-sousa-me: In any case, the query is looking for an id=-1, which should abort pretty fast

daniel-sousa-me: I tried doing another page, but it's complicated. If I replace the `Game.home_games` buy something else, it just gets rewritten because there is a `user_refresh` on `assets/app/app.rb`. I tried completely replacing the `assets/app/app.rb` and not use `GameManager`, but then I lose all the auto updating logic, so we'd need to rewrite at least `GameCard` and some other stuff to handle this in a different way.

I don't know what's a better solution.

tobymao: you can get game data easily from the games api we can chat on slack about details

PR #1904, opened 2020-10-17
fix initial turn in 2p 18 Los Angeles

### LINKED ISSUE #1893: [18LA] lists titles of SRs and ORs wrong
https://www.18xx.games/game/12602

In this game, the first SR was listed as **Stock Round**, and the 2 subsequent ORs were called **Operating Round 1 (of 2)** and **Operating Round 2 (of 2)**. The subsequent stock round was called *Stock Round 1**, followed by **Operating Round 1.1 (of 2)** and **Operating Round 1.2 (of 2)** and then **Stock Round 2** etc.

It really helps us having a sample game to reproduce the issue.

It seems that they're all off by one round.

michaeljb:
[Fixes #1893]

In 2p games of 18 Los Angeles, operating order is currently broken for the first two sets of ORs by this; the first set should be low-high then high-low, but is currently high-low for both; the second is the reverse (since during the first one the turn is set to 0, and then for the second set the turn is set to 1)


michaeljb:
There are a few games this will break that I can pin

PR #1908, opened 2020-10-18
[1846, 18 Los Angeles] fix C&WI (and PMC) giving free token in connected other city

### LINKED ISSUE #1897: [18LA] Token in Long Beach is free with PMC
* Game Gist - https://gist.github.com/aplassard/e375d0274cc6b5b5c606787a9e347609

The LA was able to lay a token in the left most slot of Long Beach for free. It should only be able to lay a free token in the third slot in LB using the PMC (I'm assuming this bug is tied to the PMC and not any other behavior).

michaeljb:
Should only be able to give the free token in the originally reserved city, not another city even if the owning corporation is connected.

Reported as an 18 Los Angeles problem, but unsurprisingly also applies to 1846

[Fixes #1897]

### DIFF
--- a/lib/engine/step/g_18_ga/token.rb
+++ b/lib/engine/step/g_18_ga/token.rb
@@ -17,7 +17,7 @@ module Engine
           []
         end
 
-        def adjust_token_price_ability!(entity, token, hex)
+        def adjust_token_price_ability!(entity, token, hex, city)
           return [token, nil] if @game.active_step.current_entity.corporation?
 
           super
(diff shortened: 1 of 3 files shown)


michaeljb:
This could break some 1846 games

michaeljb:
Looks like some other specs are broke too, closing for now

michaeljb:
My local copy of the db has through game 11493, only three 1846 games up to that point are broken by this change

tobymao:
go ahead and merge and pin

tobymao (approved):

PR #1914, opened 2020-10-18
18MEX: Improve presentation of fractional/reserved shares (fixes #1848)

### LINKED ISSUE #1848: 18MEX: Improve share presentation for NdM (and UdY)
NdM (and UdY) charter might need improvement:
image (2)

1. Consider adding a Trade-in line between IPO and Market, and do not show trade-in shares in the IPO count
2. When having non-integer shares improve justification to make the number appear adjusted. Consider if the NdM view should use 4.0, 3.0, 2.5, 0.5 to make it consistent.

The first one applies to UdY as well, while NdM is the only one with half-shares in 18MEX.

perwestling:
If there are reserved shares, do deduct them from IPO nunber shown
and show them on a line of their own (without price).

In case the corporation uses fractional shares (e.g. NdM uses
5% shares that appear as 0.5) then show one digit fraction even
if the share number to show has fraction .0. This will make for
better reading of the share number.

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -23,7 +23,7 @@ module Engine
     include Spender
 
     attr_accessor :ipoed, :share_price, :par_via_exchange
-    attr_reader :capitalization, :companies, :min_price, :name, :full_name
+    attr_reader :capitalization, :companies, :min_price, :name, :full_name, :fraction_shares
     attr_writer :par_price
 
     SHARES = ([20] + Array.new(8, 10)).freeze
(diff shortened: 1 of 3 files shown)


lib/engine/corporation.rb
```diff
@@ -37,6 +37,7 @@ def initialize(sym:, name:, **opts)
       end
 
       shares.each { |share| shares_by_corporation[self] << share }
+      @fraction_shares = shares.find { |s| (s.percent % 10).positive? }
```
tobymao, line 40: @fractional_shares = shares.any? { |s| (s.percent & 10).positive?

tobymao (approved):

lib/engine/corporation.rb
perwestling: Yeah, fractional is a better name so I can change that.
But isn't '%' (modulo) the correct operation here?

tobymao: yes that was a typo

PR #1930, opened 2020-10-18
Interface for Optional Rules, add 18 Los Angeles optional company LA Title

### LINKED ISSUE #1082: Feature suggestion: It should be possible to select variants of 18xx when creating a game

michaeljb:
Locally, I tested multiple variants with a company that was a copy of LA Title, and selecting one or the other or both all worked correctly.

Multiplayer:
Screenshot from 2020-10-18 17-28-25

Hotseat:
Screenshot from 2020-10-18 17-28-33

Game cards, one created with LA Title and one created without:
Screenshot from 2020-10-18 17-29-14

This PR replaces/is a follow-up to #1889. Fixes #1082. /cc @perwestling


michaeljb:
@tobymao I just now saw your comment on #1889 about holding off on variants, should I close this PR down?

tobymao:
let me look at the pr, again, if it's good i guess we can continue

perwestling:
I think this interface looks better than the first attempt. I would like to see how it acts with 2 or 3 choices as well, where all is chosen. 
But I think I can try this PR on 18AL to see how it behaves for this.

assets/app/view/create_game.rb
```diff
@@ -11,6 +11,7 @@ class CreateGame < Form
     needs :num_players, default: 3, store: true
     needs :flash_opts, default: {}, store: true
     needs :user, default: nil, store: true
+    needs :optional_rules_available, default: nil, store: true
```
tobymao, line 14: should we just call this optional_rules?

assets/app/view/create_game.rb
```diff
@@ -113,6 +157,7 @@ def mode_input(mode, text)
         elm = Native(@inputs[:max_players]).elm
         elm.value = [elm.value.to_i, elm.min.to_i].max
         store(:num_players, elm.value.to_i)
+        update_optional_rules_store
```
tobymao, line 160: should we have one method that sets all the stuff, like num players and optional rules?

models/game.rb
```diff
@@ -103,7 +103,13 @@ def ordered_players
 
   def to_h(include_actions: false, player: nil)
     actions_h = include_actions ? actions.map(&:to_h) : []
-    settings_h = include_actions ? settings.to_h : {}
+
+    settings_h =
```
tobymao, line 107: hm i wonder why we don't just always include the settings

perwestling:
Have now tried to  apply the 4 optional rules of 18AL to the above branch (locally for me) and it seems to work fine. Selected all 4 and they work fine. 

I have added a printout method in 18AL that prints out the short name (description) of each optional rule used in the game. Not sure this should be used in general.

models/game.rb
michaeljb: I'll probably have time to address comments tomorrow, want me to just make it always include settings?

michaeljb:
> I have added a printout method in 18AL that prints out the short name (description) of each optional rule used in the game. Not sure this should be used in general.

@perwestling I think that sounds like it'd be worth adding for general use. If you want to push it up I could integrate it with this branch, or we could wait till this one is merged in :slightly_smiling_face:

models/game.rb
tobymao: maybe

perwestling:
I have pushed a branch to my fork: https://github.com/perwestling/18xx/pull/new/feature/optional_rules/add_18AL

The print code can be found in g_18_al.rb

assets/app/view/create_game.rb
michaeljb: `update_inputs` does this now

models/game.rb
michaeljb: I did this, settings is now always included, but `seed` is only included if the actions are included

michaeljb:
Link to all changes I've added since @perwestling's c00c92bf - https://github.com/tobymao/18xx/pull/1930/files/c00c92bfb3e3be9214a4338dbaf907880ef48393..eeecda97136d5cdb1fff3d8ddec51ce1ee1565e8

perwestling (commented):
Is this part of the commit?

perwestling:
It is a bit messy when it is so many commits....

I am not sure it was a good idea to add 18AL to this commit. Some of the general code changes that was needed (mainly accessors) was due to 18AL optional rules needs, and that might be clearer if having that in a separate commit. 
Oh well, this PR should be a good pilot for optional rules.

I will wait with looking into optional rules for 18MEX until this is approved.  They look pretty straightforward to implement so these we can add later.

michaeljb:
With the [squash and merge](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) option you can always make a single commit on master when merging a PR, no matter how messy the branch :slightly_smiling_face: 

I've pushed up some commits reverting all of the 18AL pieces, so now this PR isn't adding any of 18AL's optional rules.

perwestling:
Ok. I will apply the fixes to the 18AL optional rules commit when it is time to make a PR for that. At least we have tested that it works fine with multiple optional rules.

models/game.rb
```diff
@@ -103,7 +103,9 @@ def ordered_players
 
   def to_h(include_actions: false, player: nil)
     actions_h = include_actions ? actions.map(&:to_h) : []
-    settings_h = include_actions ? settings.to_h : {}
+    settings_h = settings.to_h
+
+    settings_h.delete('seed') unless include_actions
```
tobymao, line 108: why do we delete seed? just keep it in. once you make this change go ahead and merge

tobymao (approved):

models/game.rb
michaeljb: That was the one thing I saw that was really being excluded by the old logic of only including settings when actions were also included. Will make this change when I get back to a computer...

PR #1936, opened 2020-10-19
Add desktop notification toggle

pablonen:
Adds a setting to toggle whether desktop notifications should be triggered while the game is in focus.

If this is ok, I can finish this up with the notification sending code and focus checking as discussed in #386 .

### DIFF
--- a/assets/app/view/user.rb
+++ b/assets/app/view/user.rb
@@ -3,6 +3,7 @@
 require 'game_manager'
 require 'user_manager'
 require 'lib/settings'
+require 'lib/notification'
 require 'view/game_row'
 require 'view/logo'
 require 'view/form'
@@ -39,6 +40,7 @@ module View
         when :profile
           ['Profile Settings', [
             render_notifications(setting_for(:notifications)),
+            render_desktop_notifications(setting_for(:desktop_notifications)),
             h('div#settings__colors', [
               render_logo_color(setting_for(:red_logo)),
               h(:div, [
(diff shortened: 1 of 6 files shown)


tobymao:
yes, i'm fine with it, just wanted it to be cleaned up as said in the previous PR

ryandriskel:
Testing on chrome

Confirmed new setting is added to page as expected
Screen Shot 2020-10-22 at 12 40 57 PM

Confirmed after enabling, saving, leaving the page and returning the option is selected as expected.

Screen Shot 2020-10-22 at 12 41 22 PM

I did not receive a permissions request. Further diagnosis showed my browser already had Notification.permission set to 'denied', so the code worked properly in not asking permission. However, there was no indication to me, the user, that my browser had denied the request. I found this confusing. I would suggest letting the user know and perhaps disabling the option if it was denied.

I will run a second test where my browser is not defaulting to 'denied'

assets/app/view/user.rb
```diff
@@ -123,6 +125,18 @@ def render_notifications(checked = true)
...
     end
 
+    def render_desktop_notifications(checked = false)
+      h('div#settings__desktop_notifications', [
+        render_input(
+          'Allow Desktop Turn and Message Notifications while in game',
+          id: :desktop_notifications,
+          type: :checkbox,
+          attrs: { checked: checked },
+          on: { change: -> { Lib::Notification.ask_permission } },
```
ryandriskel, line 135: I could be wrong, but this appears to ask permission whether the user is checking or unchecking the option. Shouldn't it only ask if the box is checked?

ryandriskel:
Ah yes, I cannot easily test where permission != denied locally because I am not running SSL locally.

tobymao:
closing due to lack of progress, feel free to open again when ready

PR #1958, opened 2020-10-20
[18CO] Denver and Colorado Springs fixed tile rotations

ryandriskel:

Denver and Colorado Springs can only be laid with orientation pointing north.

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -37,6 +37,9 @@ module Engine
 
       IPO_NAME = 'Treasury'
 
+      # First 3 are Denver, Second 3 are CO Springs
+      TILES_FIXED_ROTATION = %w[co1 co2 co3 co5 co6 co7].freeze
+
       EVENTS_TEXT = Base::EVENTS_TEXT.merge(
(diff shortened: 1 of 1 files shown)


perwestling (approved):

perwestling:
Even cleaner would have been to squashed all three commits to one before pushing. 
Can be done with "git rebase -i".

ryandriskel:
> Even cleaner would have been to squashed all three commits to one before pushing.
> Can be done with "git rebase -i".

If that's how we roll around here, will do in the future.

lib/engine/game/g_18_co.rb
```diff
@@ -116,6 +119,10 @@ def sell_shares_and_change_price(bundle)
         log_share_price(corporation, price) if self.class::SELL_MOVEMENT != :none
       end
 
+      def legal_tile_rotation?(_entity, _hex, tile)
+        super unless TILES_FIXED_ROTATION.include?(tile.name) && tile.rotation != 0
```
tobymao, line 123: don't use unless with multiple statements

ryandriskel: I've switched back to the previous version using if rather than unless

PR #1962, opened 2020-10-21
1828: Implement first to pass player ordering

crericha:


lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -159,8 +159,10 @@ def process_par(action)
       def pass!
         super
         if @current_actions.any?
+          @round.pass_order.delete(current_entity)
           current_entity.unpass!
         else
+          @round.pass_order << current_entity unless @round.pass_order.include?(current_entity)
```
tobymao, line 165: @round.pass_order |= [current_entity]

lib/engine/game/base.rb
```diff
@@ -1261,11 +1265,20 @@ def priority_deal_player
...
 
-        @players.rotate!(@players.index(player))
-        @log << "#{player.name} has priority deal"
+      def reorder_players
+        case self.class::NEXT_SR_PLAYER_ORDER
+        when :after_last_to_act
+          player = @players.reject(&:bankrupt)[@round.entity_index]
+          @players.rotate!(@players.index(player))
+        when :first_to_pass
+          @players = @round.pass_order if @round.pass_order.any?
```
tobymao, line 1279: what happens if there is no pass order? it's ok not to set players?

lib/engine/game/base.rb
```diff
@@ -1261,11 +1265,20 @@ def priority_deal_player
...
+        player_order.reject(&:bankrupt).index(entity)
+      end
 
-        @players.rotate!(@players.index(player))
-        @log << "#{player.name} has priority deal"
+      def reorder_players
+        case self.class::NEXT_SR_PLAYER_ORDER
+        when :after_last_to_act
+          player = @players.reject(&:bankrupt)[@round.entity_index]
+          @players.rotate!(@players.index(player))
```
tobymao, line 1277: should everything use pass order instead of last to pass?

lib/engine/step/buy_sell_par_shares.rb
crericha: |= appends to the existing array?

lib/engine/game/base.rb
crericha: If there is no pass order, then there wasn't a round for players to pass, so the order would stay the same.

lib/engine/game/base.rb
crericha: Pass order is a bit different. Its tracking when players first started passing, not just the last round of passes. For example, if the SR actions were like:

P1 buy, pass
P2 pass, pass
P3 buy, pass

The pass order would be P2, P1, P3; not P1, P2, P3.

lib/engine/step/buy_sell_par_shares.rb
crericha: Basically, append current_entity to the array if it isn't already present in the array.

tobymao: yes, it's a set union operator

crericha: Nice. Learn something new every day.

tobymao (approved):

PR #1971, opened 2020-10-22
[18CO] Par groups and float percent adjustment

ryandriskel:

Alright, starting to get into the slightly more complicated stuffs.

### Background

Par values are grouped into three zones. Each zone has a different number of shares required to float ( 4 / 5 / 6 ).
Corporations have different classifications that determine the par value zones where they are allowed to launch ( X / A / B / C ).

X is always 75 and 20% to float. This is the DSNG, which floats immediately upon the President's share being awarded.

**Par Groups**
C group corps can be parred in C / B / A
B group corps can be parred in B / A
A group corps can only be parred in A

**Float amounts**
Regardless of corporation group:
C par zone requires 40% to float
B par zone requires 50% to float
A par zone requires 60% to float

### View Samples

Showing X group at par time
Screen Shot 2020-10-22 at 11 04 56 AM

Showing B group at par time
Screen Shot 2020-10-22 at 11 04 48 AM

Showing A group at par time
Screen Shot 2020-10-22 at 11 04 44 AM

Sample of the log when a corporation changes float percentage
Screen Shot 2020-10-22 at 11 04 07 AM

### Notes

I didn't want to change the base corporation.rb but I couldn't determine any way to adjust the float_percent without adding the accessor. I suppose I could add a G18CO specific child class that adds the accessor but it seemed worthwhile to start here.

I opted not to attempt any visual changes to the Stock Market with this PR as I am more concerned with the rules implementation at this time.

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -22,7 +22,7 @@ module Engine
     include ShareHolder
     include Spender
 
-    attr_accessor :ipoed, :share_price, :par_via_exchange, :max_ownership_percent
+    attr_accessor :ipoed, :share_price, :par_via_exchange, :max_ownership_percent, :float_percent
     attr_reader :capitalization, :companies, :min_price, :name, :full_name, :fraction_shares
     attr_writer :par_price
 
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -41,6 +41,27 @@ module Engine
       # First 3 are Denver, Second 3 are CO Springs
       TILES_FIXED_ROTATION = %w[co1 co2 co3 co5 co6 co7].freeze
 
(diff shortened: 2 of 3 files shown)


lib/engine/corporation.rb
```diff
@@ -22,7 +22,7 @@ class Corporation
     include ShareHolder
     include Spender
 
-    attr_accessor :ipoed, :share_price, :par_via_exchange, :max_ownership_percent
+    attr_accessor :ipoed, :share_price, :par_via_exchange, :max_ownership_percent, :float_percent
```
tobymao, line 25: this needs to be dynamically set as opposed to being passed in as an option?

tobymao: i see it does

tobymao (approved):

PR #1975, opened 2020-10-22
Add 1860 Turn Auction

roseundy:
Implemented "turn" auction for 1860.
Created logos for the eight companies


assets/app/view/game/round/auction.rb
```diff
@@ -32,12 +33,14 @@ def render
 
           store(:before_process_pass, -> { hide! }, skip: true) if @current_actions.include?('pass')
 
-          if @current_actions.include?('par')
+          if @current_actions.include?('par') && @step.companies_pending_par
```
tobymao, line 36: this change looks too invasive i wonder if it needs another dedicated view or a different approach

assets/app/view/game/round/auction.rb
```diff
@@ -32,12 +33,14 @@ def render
...
             h(:div, render_company_pending_par)
           else
             h(:div, [
+              render_simple_bid,
               render_show_button,
               *render_companies,
+              *render_corporations,
```
tobymao, line 43: i don't think auctions need to show corporations

assets/app/view/game/round/auction.rb
```diff
@@ -94,7 +97,11 @@ def render_show_button
 
         def render_companies
           return [] if hidden? && !@step.visible?
-          return [] unless @current_actions.include?('bid')
+          unless @current_actions.include?('bid') ||
```
tobymao, line 100: don't use unless with multiples

lib/engine/action/simple_bid.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require_relative 'base'
```
tobymao, line 3: let's just use bid and not require a company or corporation

lib/engine/g_1860/corporation.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative '../corporation'
+
+module Engine
+  module G1860
```
tobymao, line 6: i'd like to avoid having a custom corporation for 1860, 

we already have the concept of receivership i think

lib/engine/step/g_1860/buy_cert.rb
```diff
@@ -0,0 +1,281 @@
...
+require_relative '../../action/buy_shares'
+require_relative '../../action/par'
+
+module Engine
+  module Step
+    module G1860
+      class BuyCert < Base
+        attr_reader :companies, :in_auction
+
+        AUCTION_ACTIONS = %w[simple_bid pass].freeze
```
tobymao, line 14: should this be a subclass of auction?

assets/app/view/game/round/auction.rb
roseundy: I was hesitant to add a new view just for a single game. It would also require a new round type, since that is what the game_page view uses to select the round view. None of that is hard, but I was trying to avoid game-specific view classes. I have tried to keep the changes to the shared parts auction view to a minimum.

I'll let you make the call as to whether a view used by just one game is the way to go.

assets/app/view/game/round/auction.rb
roseundy: In the initial auction in 1860 you are bidding on the right to choose what private company to buy or corporation to start. It is important to see those choices, including the corporations, when bidding.

assets/app/view/game/round/auction.rb
roseundy: Ok

lib/engine/action/simple_bid.rb
roseundy: Yeah, I see now I was being stupid. I conflated the values returned by actions with the actual Action classes.

lib/engine/g_1860/corporation.rb
roseundy: 1860 is going to be an outlier here.
1860 corporations have corporation-specific allowed par prices, layers (order of availability to start) and at least four stage flags: Receivership, Insolvency, Bankrupt, and Closed (different from Bankrupt). The first two flags are independent of each other.

roseundy: Can you help me understand why a custom corporation for 1860 is bad?

lib/engine/step/g_1860/buy_cert.rb
roseundy: Do you mean auctioner? There is no Auction step class. Auctioneer would have been a force-fit as it assumes bids are for a specific company.

lib/engine/action/simple_bid.rb
roseundy: And now I see I wasn't being stupid. I need to distinguish between "bid" and "simple_bid" so that bids are placed in the correct location in the auction view. I guess this is an excellent reason to add a new type of auction view (that isn't an "Auction").

assets/app/view/game/round/auction.rb
roseundy: See below. I think I now agree we need a new view. The problem is that it is still an auction, but one that violates a lot of assumptions that the current auction view was written with (e.g. bids are always associated with a private company).

roseundy: Updated step to use auction view with only a few changes

lib/engine/action/simple_bid.rb
roseundy: Now using "auctioning" to distinguish between a bid on a company and a bid for turn order

assets/app/view/game/round/auction.rb
```diff
@@ -13,6 +13,7 @@ class Auction < Snabberb::Component
         include Actionable
 
         needs :selected_company, default: nil, store: true
+        needs :selected_corporation, default: nil, store: true
```
tobymao, line 16: i don't think you need this

assets/app/view/game/round/auction.rb
```diff
@@ -168,6 +171,58 @@ def render_input(company)
...
+          })
+
+          create_bid = lambda do
+            hide!
+            price = input.JS['elm'].JS['value'].to_i
+            process_action(Engine::Action::Bid.new(
+              @current_entity,
+              price: price,
+            ))
+            store(:selected_company, nil, skip: true)
```
tobymao, line 193: do you actually need this?

lib/engine/game/g_1860.rb
```diff
@@ -43,11 +52,79 @@ def init_stock_market
...
+          # tie-breaker: lowest total face value in private companies
+          player = @players.select { |p| p.companies.any? }.min_by { |p| p.companies.sum(&:value) }
+          reason = 'least value of private companies'
+        end
+
+        @players.rotate!(@players.index(player))
+        @log << "#{player.name} has priority deal (#{reason})"
+      end
+
+      def next_round!
```
tobymao, line 90: what's the summary of changes here?

public/logos/1860/SC.svg
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
tobymao, line 1: can you remove the white spaces in these svgs?

lib/engine/step/g_1860/buy_cert.rb
```diff
@@ -0,0 +1,268 @@
...
+        def available
+          @companies.select { |c| can_afford?(current_entity, c) }
+        end
+
+        def may_purchase?(_company)
+          true
+        end
+
+        def auctioning
+          :turn if @in_auction
```
tobymao, line 33: why do you make this :turn instead of just true or false?

lib/engine/step/g_1860/buy_cert.rb
```diff
@@ -0,0 +1,268 @@
...
+        def setup_auction
+          @bids.clear
+          @first_player = current_entity
+          start_idx = entity_index
+          size = entities.size
+          # initialize bids to preserve player order starting with current player
+          entities.each_index do |idx|
+            @bids[entities[idx]] = -size + (idx - start_idx) % size
+          end
+          @in_auction = true
```
tobymao, line 196: shouldn't you just have one method that is like auctioning?

assets/app/view/game/round/auction.rb
roseundy: I need "selected_corporation", otherwise I can't respond to a Corporation UI element being clicked.

assets/app/view/game/round/auction.rb
roseundy: I think so - to control which company/corporation gets to respond to a click.

lib/engine/game/g_1860.rb
roseundy: I needed a new method of calculating the priority deal between the auction round and the first stock round.

public/logos/1860/SC.svg
roseundy: Sure. I created these with inkscape, but I'll look for a tool to remove the whitespace.

lib/engine/step/g_1860/buy_cert.rb
roseundy: No particular reason, other than being more explicit about the reason for the check in the auction view.

assets/app/view/game/round/auction.rb
tobymao: when do you respond to a corporation being clicked though? isn't it auto populated through step.auctioning?

assets/app/view/game/round/auction.rb
tobymao: but when you're auctioning do you ever choose tho company / corp to auction? i thought you're bidding for rights

lib/engine/game/g_1860.rb
tobymao: instead of doing all of that could you just override a different method like round_finished?

lib/engine/step/g_1860/buy_cert.rb
roseundy: I don't understand. Are you suggesting I don't need setup_auction ? Or that I shouldn't be checking @in_action directly?

lib/engine/step/g_1860/buy_cert.rb
tobymao: let's do 

def auctioning?
...
end

and then in all the code instead of referencing in_auction, just calling auctioning?

lib/engine/game/g_1860.rb
roseundy: I could have put it at the end of the auction round in the step logic, but that seemed out of place.

lib/engine/step/g_1860/buy_cert.rb
roseundy: In other games' auctions, it is either nil or a company.

roseundy: I'll still need to def auctioning, since the auctioning view calls @step.auctioning and it expects either nil or a company.

tobymao: ah i see what you're doing, ok let's keep it then

lib/engine/step/g_1860/buy_cert.rb
tobymao: is there a way where you can determine in_auction without having to set it?

tobymao: if not, just call it @auctioning and make it attr_reader

lib/engine/game/g_1860.rb
roseundy: Crap! I didn't realize there was an "init_round_finished" method.

roseundy: I'll just need to make sure the standard "reorder_players" doesn't re-reorder, or override that too.

assets/app/view/game/round/auction.rb
roseundy: I fixed my reply - it's a mistake.

assets/app/view/game/round/auction.rb
roseundy: It's populated, but I don't show the "Par" buttons until the corp has been clicked. No different that the private companies not showing the "Buy" button until the company has been clicked.

tobymao: but you don't need to actually click it, you know which one is gonna be, so no need for another user action unless i'm misunderstanding

lib/engine/step/g_1860/buy_cert.rb
roseundy: Ok

assets/app/view/game/round/auction.rb
roseundy: You have to click a company or corporation first in order to bid/buy/par/etc it. Otherwise you don't get the "input" controls. At the beginning of the round, there are six choices. It isn't until it is clicked that the view "knows" which company/corp you are buying/parring and attaches the input.

roseundy: And in fact, during the turn auction, I don't want the companies/corps to be clicked, since they aren't being bid on.

roseundy:
Updated code with suggested changes. Still have 'selected_corporation' and still have 1860-specific corporation

roseundy:
OK, removed game-specific corporation.

Issue #1982, opened 2020-10-23
[1817] Players skipped when president cannot afford buy treasury share after merger


SmithInMN:
Player 4 just merged SR and UR.  New price is $110.  Player 4 buys one share from the treasury.  Player 1 is skipped as it has only $89.  Players 2 and 3 have enough money to buy a share, but were skipped.  Rule states that every president should have the opportunity to buy one share of a newly converted/merged company.

{
  "status": "active",
  "actions": [
    {
      "type": "bid",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 1,
      "company": "MAJM",
      "price": 100
    },
    {
      "type": "pass",
      "entity": "Player 2",
      "entity_type": "player",
      "id": 2
    },
    {
      "type": "pass",
      "entity": "Player 3",
      "entity_type": "player",
      "id": 3
    },
    {
      "type": "pass",
      "entity": "Player 4",
      "entity_type": "player",
      "id": 4
    },
    {
      "type": "bid",
      "entity": "Player 2",
      "entity_type": "player",
      "id": 5,
      "company": "MAIL",
      "price": 70
    },
    {
      "type": "pass",
      "entity": "Player 3",
      "entity_type": "player",
      "id": 6
    },
    {
      "type": "pass",
      "entity": "Player 4",
      "entity_type": "player",
      "id": 7
    },
    {
      "type": "pass",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 8
    },
    {
      "type": "bid",
      "entity": "Player 3",
      "entity_type": "player",
      "id": 9,
      "company": "MINM",
      "price": 40
    },
    {
      "type": "bid",
      "entity": "Player 4",
      "entity_type": "player",
      "id": 10,
      "company": "MINM",
      "price": 45
    },
    {
      "type": "bid",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 11,
      "company": "MINM",
      "price": 50
    },
    {
      "type": "undo",
      "entity": "Player 2",
      "entity_type": "player",
      "id": 12
    },
    {
      "type": "undo",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 13
    },
    {
      "type": "pass",
      "entity": "Player 4",
      "entity_type": "player",
      "id": 14
    },
    {
      "type": "pass",
      "entity": "Player 1",
      "entity_type": "player",
      "id": 15
    },
    {
      "type": "pass",
      "entity": "Player 2",
[...]

PR #1987, opened 2020-10-23
1817 Purchase of treasury share would end when one player could not afford.

jenf:
fixes (#1982)


tobymao (approved):

tobymao:
go ahead and deploy and delet as necessary

PR #1994, opened 2020-10-23
Sort corporations in stock rounds.

### LINKED ISSUE #1838: 1817 Corporation ordering improvement during SRs
The amount of corporations at play makes seeing what's going on and which corporation belong to visually overwhelming.

Would be nice:

- The corporations that belong to you were outlined
- There were two columns: one column of started companies and one for unstarted companies.
         -my thinking here is you'd be able to buy stock on the left and start a new company on the right.
         -Also, maybe during SR1 the companies could be displayed in reverse. Unstarted at top/left, started on bottom/righ

jenf:
For 1817 the amount of corporations gets confusing

Sort those that have been IPOed by their running order
Then others leave in symbol order

Done in base class to game so that other games (18CO and 1849 etc) can override.

This does have the slight impact to the UI that the position can change dynamically,
but this makes it clearer the change in stock price.

fixes #1838

### DIFF
--- a/assets/app/view/game/round/stock.rb
+++ b/assets/app/view/game/round/stock.rb
@@ -57,7 +57,7 @@ module View
             },
           }
 
-          @game.corporations.map do |corporation|
+          @game.sorted_corporations.map do |corporation|
             next if @auctioning_corporation && @auctioning_corporation != corporation
 
             children = []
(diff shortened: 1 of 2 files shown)


tobymao (approved):

giovandc:
I'd argue that this suggestion could be useful for the private auction too.
The private which is being auctioned should be displayed as first, even better if separated from the others. This is even more needed in mobile view.

roseundy:
I'd argue that the order of things to be displayed during the auction should be controlled by the step logic, not the view logic.

PR #1995, opened 2020-10-24
better 18 Los Angeles and 18MEX maps

michaeljb:
- render location names even if there is no node there (e.g., Beverly Hills and
  Copper Canyon)
- render BHC blocker in middle of hex

## Before

Screenshot from 2020-10-23 18-16-08
Screenshot from 2020-10-23 18-15-54

## After
Screenshot from 2020-10-23 18-15-23
Screenshot from 2020-10-23 18-15-32


lib/engine/company.rb
```diff
@@ -59,6 +59,14 @@ def company?
       true
     end
 
+    def border?
```
tobymao, line 62: why are companies parts?

michaeljb: There is no `Engine::Part::Blocker`, and this is the `Tile` method that gets called when adding a company as a blocker:

https://github.com/tobymao/18xx/blob/master/lib/engine/tile.rb#L250-L253

michaeljb: In `View::Game::Part::Blocker`, `@blocker` is just a company everywhere

tobymao (approved):

PR #1998, opened 2020-10-24
Update all_abilities method to return an array instead of a hash

crericha:
Returning a hash from the all_abilities method can be an intermix of ability objects and arrays of ability objects, which is more complicated to process. Since all consumers use the output of the method as an array anyhow, simplify by returning as an array.

Don't delete elements from an array while iterating over it. Bad things can happen.

### DIFF
--- a/lib/engine/ability/tile_lay.rb
+++ b/lib/engine/ability/tile_lay.rb
@@ -7,7 +7,7 @@ module Engine
     class TileLay < Base
       attr_reader :hexes, :tiles, :free, :discount, :special, :connect, :blocks, :reachable
 
-      def setup(hexes:, tiles:, free: false, discount: nil, special: nil,
+      def setup(tiles:, hexes: nil, free: false, discount: nil, special: nil,
                 connect: nil, blocks: nil, reachable: nil)
         @hexes = hexes
         @tiles = tiles
--- a/lib/engine/company.rb
+++ b/lib/engine/company.rb
@@ -44,7 +44,9 @@ module Engine
     def close!
       @closed = true
 
-      all_abilities.each { |a| remove_ability(a) }
+      while (a = all_abilities.first)
+        remove_ability(a)
+      end
       return unless owner
 
       owner.companies.delete(self)
--- a/lib/engine/company.rb
+++ b/lib/engine/company.rb
@@ -44,9 +44,7 @@ module Engine
     def close!
       @closed = true
 
-      while (a = all_abilities.first)
-        remove_ability(a)
-      end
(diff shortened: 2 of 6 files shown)


lib/engine/company.rb
```diff
@@ -44,7 +44,9 @@ def min_bid
     def close!
       @closed = true
 
-      all_abilities.each { |a| remove_ability(a) }
+      while (a = all_abilities.first)
```
tobymao, line 47: change to all_abilities.dup.each

lib/engine/corporation.rb
```diff
@@ -157,11 +157,7 @@ def holding_ok?(share_holder, extra_percent = 0)
     end
 
     def all_abilities
-      all = @companies.flat_map(&:all_abilities)
-      @abilities.each do |ability|
-        abilities(ability.type) { |a| all << a }
-      end
-      all
+      @companies.map(&:all_abilities).flatten
```
tobymao, line 160: why was this changed? you've removed corporation abilities here?

crericha: That's a mistake. Will fix. Surprised it passed all the tests!

lib/engine/company.rb
crericha: Okay. I went with the while loop because I thought it would be more efficient than calling the constructor for each entry.

lib/engine/corporation.rb
```diff
@@ -157,11 +157,7 @@ def holding_ok?(share_holder, extra_percent = 0)
     end
 
     def all_abilities
-      all = @companies.flat_map(&:all_abilities)
-      @abilities.each do |ability|
-        abilities(ability.type) { |a| all << a }
-      end
-      all
+      @companies.map(&:all_abilities).append(@abilities).flatten
```
tobymao, line 160: @companies.flat_map(&:all_abilities) + @abilities

lib/engine/ability/tile_lay.rb
```diff
@@ -7,7 +7,7 @@ module Ability
     class TileLay < Base
       attr_reader :hexes, :tiles, :free, :discount, :special, :connect, :blocks, :reachable
 
-      def setup(hexes:, tiles:, free: false, discount: nil, special: nil,
+      def setup(tiles:, hexes: nil, free: false, discount: nil, special: nil,
```
ryandriskel, line 10: How is this change tied to the PR?

crericha: Its not. I tried to prune these changes from all the 1828 company ability changes I'm making. I saw this made it in, but didn't remove because it is an innocuous change and will be built upon in my next PR.

PR #2005, opened 2020-10-24
Improvements for handling merged corporation

### LINKED ISSUE #1913: [18Mex][1846][18LA][1817][Others?] Spreadsheet and Entities tab don't show closed / merged companies
The spreadsheet tab and entities tab doesn't show closed/merged companies. If a company operated for a few ORs and then closed in 1846 it would not show that company in the SS.

### LINKED ISSUE #2003: 18MEX: Merge fails in case the current entity tries to merge into NdM
Consider game 13258, action 308. 
If CHI accepts the game crashes.

I presume this is cause that CHI is deleted from the list of entities, which result in nil.
The handling of merged corporation needs to be improved. Consider using closed, and avoid doing steps for closed corporations, but they can still be shown in entity list, possibly with (CLOSED) after.

perwestling:
Merged corporation is now kept as entity for operating round,
with (CLOSED) after it. If current entity is merged, any remaining
actions is skipped silently.

The merged (closed) corporation are now skipped on entities
and game (stock) page, but remains on the spread sheet tab.

Correct removal of shares from merged corporation in 18MEX.

This fixes the following titles:
- Fixes #2003, the crash was caused by removing of a corporation
  from round entities - caused index error. Now the corporation
  remains but it will not act in the current operation round.
- Fixes #1913 for 18MEX - merged corporation does now remain on
  spread sheet, with all 10 shares in IPO, no share or par price,
  but do show operating history.
  For 1846, 18LA, 1817, Others? it depends on how close of
  corporation is handled.

Added close! and closed? methods to minor and corporation. These
only set closed attribute, sets owner to bank, and set floated to
false. Any clean up needed have to be done in the title that calls
the close! method.

### DIFF
--- a/assets/app/view/game/entities.rb
+++ b/assets/app/view/game/entities.rb
@@ -22,8 +22,10 @@ module View
           players = players.rotate(i)
         end
 
-        entity_candidates = (@game.corporations + @game.minors).reject(&:closed?)
-        player_owned, bank_owned = entity_candidates.sort_by(&:name).partition(&:owner)
+        player_owned, bank_owned = (@game.corporations + @game.minors)
+          .reject(&:closed?)
+          .sort_by(&:name)
+          .partition(&:owner)
         player_owned = player_owned.group_by(&:owner)
 
         children = players.map do |p|
--- a/assets/app/view/game/entities.rb
+++ b/assets/app/view/game/entities.rb
@@ -17,13 +17,13 @@ module View
             gap: '3rem 1.2rem',
           },
         }
-
         players = @game.players
         if (i = players.map(&:name).rindex(@user&.dig(:name)))
           players = players.rotate(i)
         end
 
-        player_owned, bank_owned = (@game.corporations + @game.minors).sort_by(&:name).partition(&:owner)
+        entity_candidates = (@game.corporations + @game.minors).reject(&:closed?)
+        player_owned, bank_owned = entity_candidates.sort_by(&:name).partition(&:owner)
         player_owned = player_owned.group_by(&:owner)
 
         children = players.map do |p|
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -256,9 +256,9 @@ module View
           h('td.padded_number', { style: { borderLeft: border_style } }, corporation.num_shares_of(corporation)),
           h('td.padded_number', { style: { borderRight: border_style } },
             "#{corporation.receivership? ? '*' : ''}#{@game.share_pool.num_shares_of(corporation)}"),
-          h('td.padded_number', corporation.par_price ? @game.format_currency(corporation.par_price.price) : ''),
+          h('td.padded_number', par_price?(corporation) ? @game.format_currency(corporation.par_price.price) : ''),
           h('td.padded_number', market_props,
-            corporation.share_price ? @game.format_currency(corporation.share_price.price) : ''),
+            share_price?(corporation) ? @game.format_currency(corporation.share_price.price) : ''),
           h('td.padded_number', @game.format_currency(corporation.cash)),
           h('td.left', order_props, operating_order_text),
           h(:td, corporation.trains.map(&:name).join(', ')),
@@ -268,6 +268,14 @@ module View
           h(:th, name_props, corporation.name),
           *render_history(corporation),
         ])
+      end
+
+      def par_price?(corporation)
+        corporation.par_price && !corporation.closed?
+      end
+
+      def share_price?(corporation)
+        corporation.share_price && !corporation.closed?
       end
 
       def render_companies(entity)
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -256,9 +256,9 @@ module View
           h('td.padded_number', { style: { borderLeft: border_style } }, corporation.num_shares_of(corporation)),
           h('td.padded_number', { style: { borderRight: border_style } },
             "#{corporation.receivership? ? '*' : ''}#{@game.share_pool.num_shares_of(corporation)}"),
-          h('td.padded_number', par_price?(corporation) ? @game.format_currency(corporation.par_price.price) : ''),
+          h('td.padded_number', corporation.par_price ? @game.format_currency(corporation.par_price.price) : ''),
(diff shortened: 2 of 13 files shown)


assets/app/view/game/entities.rb
```diff
@@ -17,13 +17,13 @@ def render
...
         }
-
         players = @game.players
         if (i = players.map(&:name).rindex(@user&.dig(:name)))
           players = players.rotate(i)
         end
 
-        player_owned, bank_owned = (@game.corporations + @game.minors).sort_by(&:name).partition(&:owner)
+        entity_candidates = (@game.corporations + @game.minors).reject(&:closed?)
+        player_owned, bank_owned = entity_candidates.sort_by(&:name).partition(&:owner)
```
tobymao, line 26: you can just do this all in one line

x, y = (...)
.rject(&:closed?)
.sort_by
.partition

assets/app/view/game/spreadsheet.rb
```diff
@@ -270,6 +270,14 @@ def render_corporation(corporation, operating_order, current_round, index)
         ])
       end
 
+      def par_price?(corporation)
```
tobymao, line 273: should you just put the closed in the corporation getter intsead?

assets/app/view/game/stock_market.rb
```diff
@@ -145,7 +145,7 @@ def grid_2d
         @game.stock_market.market.flat_map do |row_prices|
           row = row_prices.map do |price|
             if price
-              corporations = price.corporations
+              corporations = price.corporations.reject(&:closed?)
```
tobymao, line 148: if a corporation is closed they shouldn't be in the price anyways, so i don't know if we should do this

lib/engine/corporation.rb
```diff
@@ -213,5 +214,15 @@ def transfer(ownable_type, to)
       ownables.clear
       transferred
     end
+
+    def closed?
+      @closed
+    end
+
+    def close!(bank)
```
tobymao, line 222: maybe here we sohuld remove the corporation from the share price

lib/engine/game/g_18_mex.rb
```diff
@@ -502,6 +486,11 @@ def replace_token(major, major_token, exchange_tokens)
         city.place_token(ndm, ndm_replacement, check_tokenable: false)
         exchange_tokens.delete(ndm_replacement)
       end
+
+      def close_major(major)
+        major.ipoed = false
```
tobymao, line 491: shoudln't this be done in close! ?

lib/engine/minor.rb
```diff
@@ -3,10 +3,10 @@
 require_relative 'assignable'
 require_relative 'abilities'
 require_relative 'entity'
-require_relative 'ownable'
 require_relative 'passer'
 require_relative 'spender'
 require_relative 'operator'
+require_relative 'ownable'
```
tobymao, line 9: why is this moved?

lib/engine/corporation.rb
tobymao: why pass bank here? just set owner to be nil

lib/engine/minor.rb
```diff
@@ -57,5 +58,15 @@ def float!
     def inspect
       "<#{self.class.name}: #{id}>"
     end
+
+    def closed?
+      @closed
+    end
+
+    def close!(bank)
```
tobymao, line 66: just set owner nil

assets/app/view/game/entities.rb
perwestling: Fixed

assets/app/view/game/spreadsheet.rb
perwestling: Ok, have now put some logic in the par_price and share_price getters.
Had to change accessor method for share_price

assets/app/view/game/stock_market.rb
perwestling: I added a close(corporation) method in stock_market that removes it from price. That works.

lib/engine/corporation.rb
perwestling: Have added a deletion of share price from stock market. Now I pass game instead. And set owner to nil.

lib/engine/game/g_18_mex.rb
perwestling: Ok, moved.

lib/engine/minor.rb
perwestling: I think I moved it to have it after operator. Have now moved them so they are alphabetical.

perwestling:
@tobymao Now I have pushed an updated version of this, with your review comments handled.

The only thing that I wonder about is if the cleanup code should be moved to general code?
It could be:

1. Remove tokens on map
2. Delete all trains in entity
3. Move money in entity to bank
4. Move shares from players to IPO

Then it is up to the game title to transfer stuff, give rebate, swap tokens etc if it needs to do things.


-

lib/engine/stock_market.rb
```diff
@@ -117,6 +117,10 @@ def move(corporation, row, column, force: false)
       share_price.corporations << corporation
     end
 
+    def close(corporation)
```
tobymao, line 120: delete this

lib/engine/corporation.rb
```diff
@@ -213,5 +222,17 @@ def transfer(ownable_type, to)
       ownables.clear
       transferred
     end
+
+    def closed?
+      @closed
+    end
+
+    def close!(game)
+      game.stock_market.close(self)
```
tobymao, line 231: you don't need to pass the game,

just do share_price.corporations.delete(self)

lib/engine/minor.rb
perwestling: I corrected that, but forgot to comment. So it has been fixed.

lib/engine/stock_market.rb
perwestling: ok

lib/engine/corporation.rb
perwestling: ok

Added & as the share price might not have been set (in case of merging a corporation that has never been bought shares from),

perwestling:
@tobymao Another PR that has been updated.

tobymao (approved):

PR #2007, opened 2020-10-24
18 Los Angeles - better handling of passenger/freight route overlap

michaeljb:
- add help text in route step to tell user the routes can overlap
- render both routes, making the bottom one much thicker so it is visible

## Before

Screenshot from 2020-10-24 12-49-54

## After

Screenshot from 2020-10-24 12-50-26

### DIFF
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -32,7 +32,7 @@ module View
           path_indexes = paths_and_stubs.map { |p| [p, indexes_for(p)] }.to_h
 
           sorted = paths_and_stubs
-            .map { |path| path_indexes[path].map { |i| [path, i] } }.flatten(1)
+            .flat_map { |path| path_indexes[path].map { |i| [path, i] } }
             .sort_by { |_, index| index || -1 }
 
           sorted.map do |path, index|
@@ -75,7 +75,7 @@ module View
               [1, 3 * path_indexes[path].reverse.index(index)].max
             end
 
-          value_for_index(index, :width) * multiplier
+          value_for_index(index, :width).to_f * multiplier.to_i
         end
       end
     end
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -28,14 +28,17 @@ module View
           # Array<Array<Path>>
           @routes_paths = @routes.map { |route| route.paths_for(@tile.paths) }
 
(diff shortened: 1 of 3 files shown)


assets/app/view/game/part/track.rb
```diff
@@ -28,14 +28,17 @@ def render
           # Array<Array<Path>>
           @routes_paths = @routes.map { |route| route.paths_for(@tile.paths) }
 
-          sorted = (@tile.paths + @tile.stubs)
-            .map { |path| [path, index_for(path)] }
+          paths_and_stubs = @tile.paths + @tile.stubs
+          path_indexes = paths_and_stubs.map { |p| [p, indexes_for(p)] }.to_h
+
+          sorted = paths_and_stubs
+            .map { |path| path_indexes[path].map { |i| [path, i] } }.flatten(1)
```
tobymao, line 35: why do you only need to flatten one level?  why not just do a flat_map?

michaeljb: The previous line was this:

```
.map { |path| [path, index_for(path)] }
```

That produced an Array with elements that were a 2-length array containing a path and the route (index) that path was used by, like this:

```
[[pathA, 0], [pathB, nil], [pathC, 0], [pathD, 1]]
```

Now that a path can belong to more than one route, I replaced the `index_for` call (which returned an integer) at that point with a reference to the `path_indexes` hash, giving an array of integers. After my `map` call, the data structure looks like this:

```
[[[pathA, 0], [pathA, 1]], [[pathB, nil]], [[pathC, 0], [pathC, 2]], [[pathD, 1]]]
```

Flattening that once produces a structure like the initial one, where we get a list of path+index pairs:

```
[[pathA, 0], [pathA, 1], [pathB, nil], [pathC, 0], [pathC, 2], [pathD, 1]]
```

Flattening all the way would break up the pairs:

```
[pathA, 0, pathA, 1, pathB, nil, pathC, 0, pathC, 2] pathD, 1]
```

tobymao: ok, can you just do path_and_stubs.flat_map {|path| path.map... }

michaeljb: OK, I see, I mistakenly thought flat_map flattened all the way

2b1cf270 fixes that, and another error, I had this happen:

Screenshot from 2020-10-25 16-53-30

michaeljb:
This one is ready to go

PR #2012, opened 2020-10-25
Check for terminal path in middle of route

### LINKED ISSUE #1967: [18MEX] UdY can run a train through Mérida (off-board hex Q14)
in 18MEX, UdY was allowed to run a 3-train route through its home station in Mérida (Q14), despite it being an off-board location.  (P13-Q14-S12)

The move has been undone, but here is the preceding action:
https://www.18xx.games/game/12968?action=206

roseundy:
Current code doesn't check to see if a route passes through a city that has "terminal" paths.
This fixes that by checking to see if a route has a terminal flag on any path that is not at the very beginning or end of the route.
Only games known to be impacted: 18AL and 18MEX.

Fixes #1967

### DIFF
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -166,9 +166,19 @@ module Engine
     end
 
     def check_terminals!
-      if paths.size > 2
-        @game.game_error('Route cannot pass through terminal') if path[1..-2].any? { |p| p.terminal? }
+      return unless paths.size > 2
+
+      ordered_paths = []
+      @connections.each do |c|
+        cpaths = c[:connection].paths
+        if cpaths.first.nodes.any?(c[:left])
+          ordered_paths.concat(cpaths)
+        else
+          ordered_paths.concat(cpaths.reverse)
+        end
       end
+
+      @game.game_error('Route cannot pass through terminal') if ordered_paths[1..-2].any?(&:terminal?)
     end
 
     def distance
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -166,16 +166,11 @@ module Engine
     end
 
     def check_terminals!
-      return unless paths.size > 2
+      return if paths.size < 3
 
-      ordered_paths = []
-      @connections.each do |c|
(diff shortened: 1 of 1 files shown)


roseundy:
There is a non-zero probability that a number of existing games will fail this new test (i.e. they run routes that break the rules).

perwestling:
Looks good.

lib/engine/route.rb
```diff
@@ -165,6 +165,22 @@ def check_connected!(token)
       # rubocop:enable Style/GuardClause, Style/IfUnlessModifier
     end
 
+    def check_terminals!
+      return unless paths.size > 2
```
tobymao, line 169: had to read this over a couple times, can we do

return if paths.size < 3

lib/engine/route.rb
```diff
@@ -165,6 +165,22 @@ def check_connected!(token)
...
+
+      ordered_paths = []
+      @connections.each do |c|
```
tobymao, line 172: ordered_paths = @connections.map do |c|
  paths = c[:connections].paths
  paths[0].nodes.any? { ... } // i'm not sure what your code is doing

lib/engine/route.rb
roseundy: Of course.

lib/engine/route.rb
roseundy: Yep. I think it needs to be flat_map, though.

I'm putting the paths into the order implied by the order of connections. If a connection goes between node A and node B, the paths associated with that connection can be A...B or B...A. I need to have a list of paths in exactly the order encountered going from node A to node B. This is so I can check that that none of the paths - other than the start and end paths - have the "terminal" attribute.

roseundy: Ok, pushed again. I left the variable name "cpaths" to avoid confusion with the method named paths.

michaeljb:
> There is a non-zero probability that a number of existing games will fail this new test (i.e. they run routes that break the rules).

Found a few broken games and have pinned them

PR #2010, opened 2020-10-25
Added invite link to games

daniel-sousa-me:
When the url of a new game, instead of showing an error message now we see the game card allowing the user to join the game. Furthermore now the owner sees an "Invite" button on their new games.

Up next: unlisted games

PR #2015, opened 2020-10-25
Added option to create unlisted/private game

daniel-sousa-me:
The unlisted games are only hidden from the new games queue. Should they be also hidden from active games and finished games? Maybe we could have that as a separate option?

The unlisted games are hidden on the client side, after being fetched from the db. This means that each page will have less games. I don't think that's a big problem. It's not trivial to workaround, because the db is "not aware" of this field.
An alternative would be to always show all the cards, but hide the Join (and maybe Enter) buttons.

Do not pull this before #2010

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -19,7 +19,9 @@ module View
         render_inputs,
       ]
 
-      if @mode == :hotseat
+      if @mode == :multi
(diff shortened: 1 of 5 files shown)


daniel-sousa-me:
I don't know what happened, I had fixed these errors. Let's try again...

PR #2018, opened 2020-10-25
Added option to create unlisted/private game

daniel-sousa-me:
The unlisted games are only hidden from the new games queue. Should they be also hidden from active games and finished games? Maybe we could have that as a separate option?

The unlisted games are hidden on the client side, after being fetched from the db. This means that each page will have less games. I don't think that's a big problem. It's not trivial to workaround, because the db is "not aware" of this field.
An alternative would be to always show all the cards, but hide the Join (and maybe Enter) buttons.

Do not pull this before #2010


daniel-sousa-me:
I removed condition from `models/game.rb` that I don't know why is there. Please check if my change doesn't have the potential to break something.

tobymao (approved):

daniel-sousa-me (commented):
Nice. I was precisely thinking how nice it would be to have this. Do we have any issues tracking variants we'd like to implement?

PR #2020, opened 2020-10-25
"cheater" tokens for 18LA DC&H, 1846 2p Variant setup

### LINKED ISSUE #1680: [1846] GT token put on Chicago city 2 (east side of tile) rather than city 1 (south-east side of tile) when buying in the C & WI
multiplayer game ID: 9175

In OR2.1, Grand Trunk buys in the C&WI private. The free GT token was placed in the wrong location initially or more likely, it has moved during a code update as none of the players noticed the error at the time and it's current location could only have been the result of an illegal move. The game was started on 9/9/2020. 

2020-09-17 00_25_31-1846 - hs_opumancl_9175 - 18xx Games
prior to the buy in
2020-09-17 00_27_52-1846 - hs_opumancl_9175 - 18xx Games
immediately following the buy in
2020-09-17 00_28_20-1846 - hs_opumancl_9175 - 18xx Games
immediately after placing the token.

A link to the moves in question:
https://www.18xx.games/game/9175?action=234

I doubt that someone would not have noticed this during live play, most especially the GT player. Is there any way to move the token over to the correct spot?

Thanks in advance,
Mike Ward

### LINKED ISSUE #1806: 18 Los Angeles Special Rules Summary
[18 Los Angeles Rules](https://drive.google.com/file/d/1G_fLbak96VWQ0Vfvg7-Qh2gXgv7r0BQi/view?usp=sharing)
[18 Los Angeles Map](https://drive.google.com/file/d/1MUSljm9Hoe39pEXp9R_UjV5mVeVaNXGN/view)
[1846 Rules](https://s3-us-west-2.amazonaws.com/gmtwebsiteassets/1846/1846-RULES-GMT.pdf)

Listing differences from 1846.

# Setup
- [x] different company groups than those inherited from 1846
- [x] removed corporations place two tokens

# Private Company Draft
- [x] uses 1846 2p Variant style draft for all player counts
- [x] if all players pass, the draft is over and SR1 begins, no ORs; unpicked companies are simply removed
- [x] snake turn order
- [x] if one company is left and all players pass, the draft is over instead of discount happening

# Priority Deal
- [x] determined at beginning of each stock round
- [x] turn order is lowest cash to highest (tie keeps same order as before)

# Trains/Routes
- [x] N-trains are "Passenger Trains"
- [x] N/M-trains are "Freight Trains"
- [x] a Passenger Train and Freight Train can each use the same track; multiple trains of the same type may not reuse track
- [x] a route can only include two offboards if they N-S or E-W; some hexes count as multiple directions; it is allowed to run from a N/E (San Bernardino) to a S/E (Irvine), because they are N-S even though both are also E; initial groups config may cover this already, but it needs testing
- [x] bonuses should apply for running E-W or N-S

# Minors / Independent Railroads
- [x] may only lay one tile per OR
- [x] initial capital is different from their initial "value" (Gardena starts with $60, not $140; Orange County starts with $40, not $100)

# Map
- [x] render and enforce preprinted track "stubs"; they must be honored by upgrades (i.e., when Cypress is upgraded, one of the new paths must point to Long Beach)
- [x] there is a hex in the row above Row A in column 9 so that Pasadena (A8) can upgrade to brown; what to call this row?
- [x] hex G14 should be added (similar to the above, but less problematic); it and the above hex could start as grey hexes with a $0 offboard with no location name for their paths to "nowhere" (these paths exist to allow adjacent hexes on the map to have more freedom when being upgraded)
- [x] all the hexes have names, even if there are no cities/offboards in the hex (not needed for alpha)

# Corporations
- [x] token teleports/discounts are not implemented
    - SP: $40/$100 token in Los Angeles
    - LA: $40 token in Alhambra
    - ELA: $40/$60 token in Culver City
    - SF: $40 token in Montebello

# Private Companies
- [x] does Pacific Maritime work exactly like Chicago & Western Indiana? Specifically, does the reservation disappear once it is bought in?
- [x] optional rules for the two optional private companies
- [x] LA Title
- [x] Dewey, Cheatham, and Howe
- [x] does US Mail Contract behave like 1846's Mail Contract, or does it close in Phase III?
- [x] new icons for Los Angeles Citrus (replace `public/icons/18_los_angeles/meat.svg` and `public/icons/18_los_angeles/lac_token.svg`) (not needed for alpha)
- [x] new icon for South Bay Line (replace `public/icons/18_los_angeles/sbl.svg`) (not needed for alpha)
- [x] update references in logs to revenue bonuses for LA Citrus and LA Steamship (not needed for alpha)

# 2 player
- [x] just one group of corporations for removals
- [x] removed corporations' second tokens are placed during setup (as in 3p-5p 18 Los Angeles) instead of on green tile lay

michaeljb:
- 18 Los Angeles's private company "Dewey, Cheatham, and Howe": place a token in
  city with no available slots (and provide the name for this token argument)
- 1846 2p Variant setup: removed corporations place a second token in slot 1
  instead of slot 0; this is functionally equivalent to current behavior (token
  is placed when green tile is laid, has no real effect before then) but way
  more obvious what is happening, and takes advantage of the digital platform

[Closes #1806]

## 1846 2p Before

1846 2p before log 
1846 2p before map

## 1846 2p After

1846 2p cheaters log
1846 2p cheaters
Screenshot from 2020-10-25 12-18-38
Screenshot from 2020-10-25 12-18-49
Screenshot from 2020-10-25 12-18-58

## 18 Los Angeles - DC&H

DCH before
DCH after

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -82,7 +82,8 @@ module View
           entity = @selected_company || step.current_entity
           actions = step.actions(entity)
           return if (%w[remove_token place_token] & actions).empty?
-          return if @token && !step.can_replace_token?(entity, @token)
+          return if @token && !step.can_replace_token?(entity, @token) &&
+                    !(cheater = entity.abilities(:token)&.cheater)
 
           event.JS.stopPropagation
 
@@ -102,7 +103,7 @@ module View
               action = Engine::Action::PlaceToken.new(
                 @selected_company || @game.current_entity,
                 city: @city,
-                slot: @slot_index,
+                slot: cheater || @slot_index,
                 token_type: next_tokens[0].type
               )
               store(:selected_company, nil, skip: true)
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -15,9 +15,9 @@ module View
 
     def render_notification
       message = <<~MESSAGE
+        <p>18 Los Angeles is now in beta!</p>
         <p>1817 is now more stable and in alpha! Feel free to play multiplayer games.</p>
         <p>18MEX is now in alpha!</p>
-        <p>18 Los Angeles by Tony Fryer is now in alpha! It is based on 1846, and plays 2-5 players. Learn more on the <a href='https://traxx-denver.com/games/'>TraXX</a> website.</p>
         <p>Please file <a href='https://github.com/tobymao/18xx/issues'>issues and ideas</a> on
         <a href='https://github.com/tobymao/18xx/issues'>GitHub</a>.<br>
         If you have any questions, check out the <a href="https://github.com/tobymao/18xx/wiki/FAQ">FAQ</a> and other
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -15,7 +15,6 @@ module View
 
     def render_notification
       message = <<~MESSAGE
-        <p>18 Los Angeles is now in beta!</p>
         <p>1817 is now more stable and in alpha! Feel free to play multiplayer games.</p>
         <p>18MEX is now in alpha!</p>
         <p>Please file <a href='https://github.com/tobymao/18xx/issues'>issues and ideas</a> on
--- a/lib/engine/ability/token.rb
+++ b/lib/engine/ability/token.rb
@@ -6,10 +6,11 @@ module Engine
   module Ability
     class Token < Base
       attr_reader :hexes, :teleport_price, :extra, :from_owner, :discount, :city,
-                  :neutral
+                  :neutral, :cheater
 
       def setup(hexes:, price: nil, teleport_price: nil, extra: nil,
-                from_owner: nil, discount: nil, city: nil, neutral: nil)
+                from_owner: nil, discount: nil, city: nil, neutral: nil,
+                cheater: nil)
         @hexes = hexes
         @price = price
         @teleport_price = teleport_price
@@ -18,6 +19,7 @@ module Engine
         @discount = discount
         @city = city
         @neutral = neutral || false
+        @cheater = cheater || false
       end
 
       def price(token = nil)
--- a/lib/engine/config/game/g_18_los_angeles.rb
+++ b/lib/engine/config/game/g_18_los_angeles.rb
@@ -324,6 +324,29 @@ module Engine
       ]
     },
     {
+      "name": "Dewey, Cheatham, and Howe",
+      "value": 40,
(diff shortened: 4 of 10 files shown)


michaeljb:
After this is merged, I'd like to give it a couple days for anyone to see if this breaks 1846 2p, if it doesn't we can call that production.

Once this and #2007 are in, 18 Los Angeles will be ready for beta. This PR does have a small conflict with #2007 that I'll fix as soon as one of them is merged.

assets/app/view/game/part/city_slot.rb
```diff
@@ -82,7 +82,10 @@ def on_click(event)
           entity = @selected_company || step.current_entity
           actions = step.actions(entity)
           return if (%w[remove_token place_token] & actions).empty?
-          return if @token && !step.can_replace_token?(entity, @token)
+          return if @token && !step.can_replace_token?(entity, @token) &&
+                    !(cheater = entity.abilities(:token)&.cheater)
+
+          puts "cheater || @slot_index = #{cheater || @slot_index}"
```
tobymao, line 88: puts

assets/app/view/game/part/city_slot.rb
```diff
@@ -82,7 +82,10 @@ def on_click(event)
           entity = @selected_company || step.current_entity
           actions = step.actions(entity)
           return if (%w[remove_token place_token] & actions).empty?
-          return if @token && !step.can_replace_token?(entity, @token)
+          return if @token && !step.can_replace_token?(entity, @token) &&
```
tobymao, line 85: can we use block syntex 

entity.abilities(:token) do |token|
 ...

for the future of multiple abitiies

assets/app/view/game/part/city_slot.rb
```diff
@@ -82,7 +82,13 @@ def on_click(event)
           entity = @selected_company || step.current_entity
           actions = step.actions(entity)
           return if (%w[remove_token place_token] & actions).empty?
-          return if @token && !step.can_replace_token?(entity, @token)
+
+          cheaters = []
+          entity.abilities(:token) do |ability|
```
tobymao, line 87: ok, i change my mind since you require the first ability, switch it back sorry

michaeljb: np, it's not that I need the first so much as I don't know how we'd deal with the situation where one entity has multiples of this ability type; with or without the block syntax this code would need changes once we have a game with that situation anyway

michaeljb:
@tobymao this is ready again

tobymao:
loooks good feel free to merge

tobymao (approved):

PR #2022, opened 2020-10-25
[18CO] Specific Mine Implementation

ryandriskel:
_This change only impacts 18CO files._

### Background

- Placing a tile on a location with a mine places the mine token onto the corporation's charter.
- Mines award $10 per mine directly to the treasury after the dividend step as long as a train was run
- Idarado Mining Company: Money gained from mines is doubled for the owning Corporation. Closes phase 5 if owned by a player. Closes phase 6 if owned by a corporation.
- Mines are removed from the board in phase 6

### Results

Mines show up as a corporation ability
Screen Shot 2020-10-25 at 10 13 18 AM

Log is written to when a corporation picks up a mine as well as when mine money is awarded
Screen Shot 2020-10-25 at 10 18 04 AM

### Implementation Notes

Rather than define new Ability classes I am working with the Base class. I don't know any other games that use tokens like this. One downside is the need to build a New object when adding a mine as the attributes are read only.

Piggybacking on **change_share_price** isn't the most ideal place to do this, but the timing is right and the data needed to determine if mines should pay is available.

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -66,6 +66,8 @@ module Engine
 
       EAST_HEXES = %w[A26 J26 E27 G27].freeze
 
+      BASE_MINE_VALUE = 10
+
       EVENTS_TEXT = Base::EVENTS_TEXT.merge(
           'remove_mines' => ['Mines Close', 'Mine tokens removed from board and corporations']
         ).freeze
@@ -76,9 +78,26 @@ module Engine
         @dsng ||= corporation_by_id('DSNG')
       end
 
+      def imc
+        @imc ||= company_by_id('IMC')
+      end
+
       def setup
         setup_company_price_50_to_150_percent
+        setup_companies
+        setup_corporations
+      end
 
+      def setup_companies
+        imc.add_ability(Engine::Ability::Base.new(
+            type: :mine_multiplier,
+            description: 'Mine values are doubled',
+            count_per_or: 2
+          ))
+      end
+
+      def setup_corporations
+        # The DSNG comes with a 2P train
         train = @depot.upcoming[0]
         train.buyable = false
         dsng.buy_train(train, :free)
@@ -90,6 +109,35 @@ module Engine
           self.class::CERT_LIMIT_COLORS,
           multiple_buy_colors: self.class::MULTIPLE_BUY_COLORS
         )
+      end
+
+      def mines_count(entity)
+        entity.abilities(:mine_income).sum(&:count_per_or)
+      end
+
+      def mine_multiplier(entity)
+        entity.abilities(:mine_multiplier).map(&:count_per_or).reject(&:zero?).reduce(1, :*)
+      end
+
+      def mines_total(entity)
+        BASE_MINE_VALUE * mines_count(entity) * mine_multiplier(entity)
+      end
+
+      def mines_remove(entity)
+        entity.abilities(:mine_income) do |ability|
+          entity.remove_ability(ability)
+        end
+      end
+
+      def mine_add(entity)
+        mine_count = mines_count(entity) + 1
+        mines_remove(entity)
+        entity.add_ability(Engine::Ability::Base.new(
+              type: :mine_income,
+              description: "Corporation owns #{mine_count} mines",
+              count_per_or: mine_count,
+              remove: '6'
+            ))
       end
 
       def operating_round(round_num)
@@ -181,10 +229,12 @@ module Engine
       def event_remove_mines!
         @log << '-- Event: Mines close --'
 
-        @log << 'Mines removed from board (TODO)'
+        hexes.each do |hex|
+          hex.tile.icons.reject! { |icon| icon.name == 'mine' }
+        end
 
-        @companies.each do |company|
-          @log << "Mines removed from  #{company.name} (TODO)"
+        @corporations.each do |corporation|
+          mines_remove(corporation)
         end
       end
 
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -84,16 +84,7 @@ module Engine
 
       def setup
         setup_company_price_50_to_150_percent
-        setup_companies
         setup_corporations
-      end
-
-      def setup_companies
-        imc.add_ability(Engine::Ability::Base.new(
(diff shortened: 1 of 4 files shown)


tobymao:
i think normally this was done with the assign ability. could that be used instead for consistency?

lib/engine/game/g_18_co.rb
```diff
@@ -90,6 +109,35 @@ def init_stock_market
         )
       end
 
+      def mines_count(entity)
+        entity.abilities(:mine_income).sum(&:count_per_or)
+      end
+
+      def mine_multiplier(entity)
+        entity.abilities(:mine_multiplier).collect(&:count_per_or).reject(&:zero?).inject(1, :*)
```
michaeljb, line 117: Throughout the codebase, `map` is used instead of `collect`, and `reduce` is used instead of `inject`.

lib/engine/game/g_18_co.rb
```diff
@@ -74,9 +76,26 @@ def dsng
...
+      end
+
       def setup
         setup_company_price_50_to_150_percent
+        setup_companies
+        setup_corporations
+      end
 
+      def setup_companies
+        imc.add_ability(Engine::Ability::Base.new(
```
michaeljb, line 90: Should this be defined in the JSON in `lib/engine/config/game/g_18_co.rb`?

michaeljb: Oh right, it's using `Ability::Base` instead of its own ability class.

ryandriskel:
> i think normally this was done with the assign ability. could that be used instead for consistency?

I've looked through all the games using assign and I'm having a difficult time visualizing how to use the Assign ability here. I guess I should first try to clarify whether you meant the assign_hexes Ability or whether you meant Assignable ?

Where I see assign_hexes Ability being used is when one corporation _places_ tokens on hexes to gain a bonus. This is the exact opposite - all corporations are removing tokens from the map to gain their respective bonuses. Additionally, once the mine is collected, the hex it came from is no longer relevant.

If Mine was its own object, using Assignable on said class would make sense, but I haven't seen a pattern of games defining their own object types.

tobymao:
it's not an ability, it's an action. an entity can assign a target

lib/engine/step/g_18_co/track.rb
```diff
@@ -9,8 +9,18 @@ class Track < Track
         def process_lay_tile(action)
           lay_tile_action(action)
 
-          # TODO: Implement Mine Collection
-          @log << 'TODO: implement mine token collection'
+          # Mine Token Collection
+          unless action.hex.tile.icons.select { |icon| icon.name == 'mine' }.empty?
```
michaeljb, line 13:
```suggestion
          if action.hex.tile.icons.map(&:name).include?('mine')
```

michaeljb:
@tobymao @ryandriskel I took a crack at doing this via assignment; it does use the assignment hash a little differently than we have elsewhere in that the value really matters and can be something other than `true`. Ditching the abilities also has the benefit of not using `count_per_or: 2` for something that is really just used once in the OR. In my commit I also check for positive revenue in an override of `process_dividend`, it seemed more natural to do that and take the revenue from the history.

I did a bit of local testing and this works for at least a couple mines and with the Idarado Mining Company (I didn't keep going up till phase 6 to test closing of the mines).

https://github.com/michaeljb/18xx.games/commit/cab6938870562a59429136b2c34a4281240eed0c

ryandriskel:
Thanks @michaeljb - That does look a lot cleaner. I really appreciate you giving it a shot. I'll load it up on my computer tomorrow and run through everything.

Moving this to Draft for now.

ryandriskel:
I've got the basic code working following Michael's lead. It is cleaner, but I feel this trades misusing Ability count_per_or for misusing Assignment active.

The one thing lost by not using the Abilities section is the number of mines owned by the corp.

Also, the corporation never _places_ a mine on the board, so it might be misleading having the mine show up in the token list.

Screen Shot 2020-10-28 at 8 45 55 AM

One possibility I see is to extend the assignment data to allow an object rather than just an active flag. I'm going to play around with that and see if there's a clean solution.

ryandriskel:
Argh, I didn't mean to actually push the assignment version of this as I am not sure that's the way to go based on feedback in slack.

lib/engine/game/g_18_co.rb
```diff
@@ -167,6 +172,15 @@ def new_auction_round
         ])
       end
 
+      def action_processed(action)
+        super
+
+        case action
+        when Action::BuyCompany
+          mine_update_text(action.entity) if action.company == imc && action.entity.corporation?
```
ryandriskel, line 180: When the IMC is bought in, the text on the Ability needs to be updated to reflect the change in mine value and total.

lib/engine/step/g_18_co/dividend.rb
```diff
@@ -14,17 +14,15 @@ def share_price_change(entity, revenue = 0)
           { share_direction: %i[right up], share_times: [1, 1] }
         end
 
-        def change_share_price(entity, payout)
-          tfm = @game.mines_total(entity)
-
-          # Only pay mines if a train produce revenue regardless of withhold or pay
-          if tfm.positive? && (payout[:corporation].positive? || payout[:per_share].positive?)
-            @game.bank.spend(tfm, entity)
+        def process_dividend(action)
```
ryandriskel, line 17: Pulled this from @michaeljb 's suggestions as it is a much more appropriate place to execute the action.

lib/engine/step/g_18_co/track.rb
```diff
@@ -17,9 +17,7 @@ def process_lay_tile(action)
             # Add mine to corporation data
             @game.mine_add(action.entity)
 
-            @log << "#{action.entity.name} collects a mine token from
-              #{action.hex.name} for a total of
-              #{@game.format_currency(@game.mines_total(action.entity))} from mines"
+            @log << "#{action.entity.name} collects a mine token from #{action.hex.name}"
```
ryandriskel, line 20: I shortened the log message to avoid redundancy since the payout is shown on the corporation now

tobymao (approved):

tobymao:
conflict

ryandriskel:
> conflict

Rebased

PR #2023, opened 2020-10-25
18MEX: Implement support for optional rule 8a and 8d

perwestling:
8a: Allow corporations (not minors) to lay a 3rd yellow tile (or 1
upgrade) during its first OR. For its second OR and later it is 2
yellow or 1 upgrade.

8d: Use normal (hard rust) 4 trains

Optional rule 8b and 8c are not yet implemented.


ryandriskel (approved):
Looks good here!

PR #2050, opened 2020-10-26
1817: Add the underscores for safe par spots and 

### LINKED ISSUE #1831: 1817 cosmetic improvement to stock market
Values 50, 70, and 120 should have an underscore, or some way to indicate that they are the values needed for a 2,5,10 share company to float, take maximum loans, and not enter acquisition.

Values 50-200 should have an indication of valid par values.

jenf:
Make it clearer that the acquisition values are not 'real' by putting them in brackets

fixes #1831

Moves the text to the game class, so they can be overwritten and extended if needed


assets/app/view/game/stock_market.rb
```diff
@@ -208,18 +218,7 @@ def render
...
-            [:unlimited, 'Corporation shares can be held above 60%'],
-            [:multiple_buy, 'Can buy more than one share in the corporation per turn'],
-            [:close, 'Corporation closes'],
-            [:endgame, 'End game trigger'],
-            [:liquidation, 'Liquidation'],
-            [:acquisition, 'Acquisition'],
-            [:repar, 'Par value after bankruptcy'],
-            [:ignore_one_sale, 'Ignore first share sold when moving price'],
-          ]
+          type_text = @game.class::MARKET_TEXT
```
ryandriskel, line 221: Hooray! I support this refactor.

jenf: I was tempted to remove all notion of colors from the stock market class, one day maybe.

PR #2063, opened 2020-10-27
Option flag to allow town to upgrade to city

ryandriskel:
### Background

Add a new option that allows a town tile to upgrade to a city tile. Used in 18CO, 18NEB, 1825, 1829, 1856, etc ( Section 5.4 http://www.fwtwr.com/18xx/rules_difference_list/single_list.htm#5 )

### Implementation Notes

The non-18CO specific changes are adding the `to_city` option and changing the definition of `<=` for Town. It would be reasonable to debate the name of the attribute `to_city`, its just what was in my head at the time.

In `<=` I choose to make it explicit that the only time we don't use the `super` definition is when both `to_city && other.city?` are true. There could be an argument made to use the shorthand `return to_city if other.city?` but that would change the logic of how `<=` works for a town tile that doesn't use the new feature and I wanted to avoid that.

Game creators still need to define the upgrade path for town->city tiles within upgrades_to? as the automatic upgrade detection will not determine cities are upgrades of towns. I think this is desirable as most games specify the exact tile(s) and not just any city tile for town upgrades.

Within `all_potential_upgrades`, I originally used the pattern set by other games of pre-determining a tile and saving in memory. This ended up presenting issues when that tile was placed on the board. The manifest would show the tile with tokens placed on it. Other games only have one specific tile, so the manifest would be empty once that tile was placed. Below is a screenshot of that in effect before my change.

Screen Shot 2020-10-26 at 3 37 17 PM

### Game Usage

Greeley about to be upgraded
Screen Shot 2020-10-26 at 3 43 32 PM

Greeley upgraded shows the upgrade icon removed
Screen Shot 2020-10-26 at 6 14 42 PM

Tile Manifest
Screen Shot 2020-10-26 at 6 15 01 PM

Limon doesn't have the upgrade icon, and selecting the tile properly shows no tiles.
Screen Shot 2020-10-26 at 6 15 58 PM

### Additional Notes

This change involved some 18CO tiles that were defined slightly incorrectly, so I fixed those along the way. The config changes didn't seem worth a separate PR.
1.) Denver green tile was rotated one step in the config
2.) The co8, co9 and co10 tile counts did not match the manifest
3.) Acquisition is the applicable section to show on the Stock Market.

### DIFF
--- a/lib/engine/config/game/g_18_co.rb
+++ b/lib/engine/config/game/g_18_co.rb
@@ -111,24 +111,24 @@ module Engine
 		"28": 2,
 		"29": 2,
 		"co8": {
-			"count": 1,
+			"count": 5,
 			"color": "green",
-			"code": "town=revenue:20;junction;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0"
+			"code": "town=revenue:20,to_city:1;junction;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0"
 		},
 		"co9": {
-			"count": 1,
+			"count": 5,
 			"color": "green",
-			"code": "town=revenue:20;junction;path=a:0,b:_0;path=a:5,b:_0;path=a:3,b:_0"
+			"code": "town=revenue:20,to_city:1;junction;path=a:0,b:_0;path=a:5,b:_0;path=a:3,b:_0"
 		},
 		"co10": {
-			"count": 1,
+			"count": 2,
 			"color": "green",
-			"code": "town=revenue:20;junction;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0"
+			"code": "town=revenue:20,to_city:1;junction;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0"
 		},
 		"co2": {
 			"count": 1,
 			"color": "green",
-			"code": "city=revenue:50,slots:3;city=revenue:50,hide:1;path=a:0,b:_0;path=a:1,b:_0;path=a:4,b:_0;path=a:5,b:_0;path=a:2,b:_1;path=a:3,b:_1;label=D;"
+			"code": "city=revenue:50,slots:3;city=revenue:50,hide:1;path=a:0,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:5,b:_0;path=a:1,b:_1;path=a:2,b:_1;label=D;"
 		},
 		"co6": {
 			"count": 1,
@@ -297,25 +297,25 @@ module Engine
 			"40"
 		],
 		[
-			"10b",
+			"10a",
 			"15",
 			"20",
 			"25",
 			"30"
 		],
 		[
-			"10b",
-			"10b",
+			"10a",
+			"10a",
 			"15",
 			"20",
 			"20"
 		],
 		[
-			"10b",
-			"10b",
-			"10b",
-			"15b",
-			"15b"
+			"10a",
+			"10a",
+			"10a",
+			"15a",
+			"15a"
 		]
 	],
 	"companies": [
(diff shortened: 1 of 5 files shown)


lib/engine/config/game/g_18_co.rb
```diff
@@ -111,24 +111,24 @@ module G18CO
 		"28": 2,
 		"29": 2,
 		"co8": {
-			"count": 1,
+			"count": 5,
 			"color": "green",
-			"code": "town=revenue:20;junction;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0"
+			"code": "town=revenue:20,to_city:1;junction;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0"
```
jenf, line 116: I'm not sure the 'to_city' should be part of 'town' since a tile can have multiple towns on it.
Maybe this should be part of 'upgrade' instead

lib/engine/game/g_18_co.rb
```diff
@@ -150,12 +152,28 @@ def revenue_str(route)
         str
       end
 
+      def upgrades_to?(from, to, special = false)
+        # Green towns can't be upgraded to brown cities unless the hex has the upgrade icon
+        if GREEN_TOWN_TILES.include?(from.hex.tile.name)
+          return BROWN_CITY_TILES.include?(to.name) if from.hex.tile.icons.select { |icon| icon.name == 'upgrade' }.any?
```
jenf, line 158: How is the upgrade icon being set?

jenf (commented):
Generally okay, but I think it needs to be part of 'upgrade' rather than 'town'

lib/engine/config/game/g_18_co.rb
ryandriskel: Being a flag on town would allow a potential for some weird tile where one town upgrades to a city yet a different town stays a town. I don't know of such a case but I don't see why multiple towns precludes the option from being attached to the town.

lib/engine/game/g_18_co.rb
ryandriskel: The upgrade icon is specific to 18CO and set in the map config.

lib/engine/config/game/g_18_co.rb
jenf: I suppose... I'll leave it to Toby to decide.

ryandriskel: Phrased another way, its indicating "This town is eligible to convert to a city" rather than "This tile is eligible to convert to a city".

tobymao: yes, it's more flexible in that case, but a bit more verbose

lib/engine/step/g_18_co/track.rb
```diff
@@ -9,6 +9,12 @@ class Track < Track
         def process_lay_tile(action)
           lay_tile_action(action)
 
+          # Remove the upgrade icon when the town is converted to a city
+          if action.hex.tile.cities.any? &&
+             action.hex.tile.icons.select { |icon| icon.name == 'upgrade' }.any?
```
tobymao, line 14: tile.icons.any? { ...}

lib/engine/game/g_18_co.rb
tobymao: icons.any? { ... }

lib/engine/config/game/g_18_co.rb
ryandriskel: Toby, I'm unsure if you're encouraging we move the flag to upgrade. I see how to write the code, however the more I think about it, the more I prefer having it on town because it is more specific. For example, if there were no towns on the tile I'm not sure what the expected usage of the flag would be.

tobymao: i'll leave that decision to you but it seems like town is fine

lib/engine/game/g_18_co.rb
ryandriskel: Done

lib/engine/step/g_18_co/track.rb
ryandriskel: Done

tobymao (approved):

PR #2072, opened 2020-10-27
1817: Corps entering acquisition in merger round should skip acqusition round.

### LINKED ISSUE #2061: 1817: Companies falling into acquisition during M&A should not be acquirable
https://www.18xx.games/game/13593 is the game and EJE is the company that went into acquisition.

It fell into acquisition during the M&A round so it shoudl not be offered for acquisition.  I imagine liquidations might also be affected by this.

jenf:
Keep a track of corps share price before the merger round and filter out those
which just entered acquisition (fixes #2061)

### DIFF
--- a/lib/engine/round/g_1817/acquisition.rb
+++ b/lib/engine/round/g_1817/acquisition.rb
@@ -21,7 +21,9 @@ module Engine
           # Things that are offered up for acquisition, sale etc
           @offering =  @game
                       .corporations
-                      .select(&:floated?)
+                      .select do |corp|
+                        corp.floated? && !corp.share_price.acquisition? || @game.stock_prices_start_merger[corp].acquisition?
+                      end
                       .sort.reverse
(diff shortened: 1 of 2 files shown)


lib/engine/round/g_1817/acquisition.rb
```diff
@@ -23,6 +23,9 @@ def select_entities
                       .corporations
                       .select(&:floated?)
```
tysen, line 24:
```suggestion
                      .select do |corp|
                        corp.floated? && !corp.share_price.acquisition? || @game.stock_prices_start_merger[corp].acquisition?
                      end
```

michaeljb:
I would merge this but it will (presumably) break some 1817 games, so I'll just "approve" it and leave the pinning or deletion to Toby or Jen.

michaeljb (approved):

tobymao (approved):

PR #2077, opened 2020-10-28
Stock drop per 10% sold

ventusignis:
- Add SELL_MOVEMENT where the market price falls once for every 10% sold (for example drops twice when 25% sold)

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -654,7 +654,7 @@ module Engine
         when :down_share
           bundle.num_shares.times { @stock_market.move_down(corporation) }
         when :down_per_10
-          (10 * bundle.num_shares / entity.total_shares).to_i.times { @stock_market.move_down(corporation) }
+          (bundle.percent / 10).to_i.times { @stock_market.move_down(corporation) }
         when :left_block_pres
           stock_market.move_left(corporation) if was_president
         when :none
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -654,18 +654,7 @@ module Engine
         when :down_share
           bundle.num_shares.times { @stock_market.move_down(corporation) }
         when :down_per_10
-          case entity.total_shares
-          when 5
-            (2 * bundle.num_shares).times { @stock_market.move_down(corporation) }
-          when 10
-            bundle.num_shares.times { @stock_market.move_down(corporation) }
-          when 20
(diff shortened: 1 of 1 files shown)


michaeljb (approved):

michaeljb:
Based on http://www.fwtwr.com/18xx/rules_difference_list/single_list.htm#2 I'm guessing this is for 1856? An issue with a list like https://github.com/tobymao/18xx/issues/686 might be useful

lib/engine/game/base.rb
```diff
@@ -652,6 +653,8 @@ def sell_shares_and_change_price(bundle)
         case self.class::SELL_MOVEMENT
         when :down_share
           bundle.num_shares.times { @stock_market.move_down(corporation) }
+        when :down_per_10
+          (10 * bundle.num_shares / entity.total_shares).to_i.times { @stock_market.move_down(corporation) }
```
tobymao, line 657: can you just do (bundle.percent / 10).to_i

perwestling: For selling 25% does bundle.num_shares return 2.5? Or 2? In the former case one could add a 'ceil: true' parameter to num_shares. Have done similar changes to handle the 5% shares of NdM in 18MEX.

Isn't the down_per_10 the normal behavior in 18xx? I don't remember 1856 have any special rules for that.

perwestling: Ah, now I remember? The "big one" have 5% shares that do not affect the price if sold one, but you need to sell two to drop it. Right.
In case of 18MEX you wont drop price if selling 2 5%, but 18MEX has special handling in the game title code.

ventusignis: > For selling 25% does bundle.num_shares return 2.5? Or 2? In the former case one could add a 'ceil: true' parameter to num_shares. Have done similar changes to handle the 5% shares of NdM in 18MEX.
> 
> Isn't the down_per_10 the normal behavior in 18xx? I don't remember 1856 have any special rules for that.

I think it's the normal for 20 share companies in most 18xx games, but this is distinctly different from dropping once per share (in a 20 share company)

The intention is that it gets rounded down, so selling 25% would result in 2 drops

ventusignis:
> Based on http://www.fwtwr.com/18xx/rules_difference_list/single_list.htm#2 I'm guessing this is for 1856? An issue with a list of special rules like #686 might be useful

I created https://github.com/tobymao/18xx/issues/2085
It's a bit longer than I thought it would be

PR #2078, opened 2020-10-28
When selecting par value, display number of purchasable shares.

dlpierce:
Displays the number of shares the player could buy with cash in each par price button. Adjusts the button padding so the number is on its own line.

This is a feature that the old DOS 1830 game has that makes it easy to see at which par prices you can float a company on your own.

This does not take into consideration the per corporation share limit, so games with low par prices (e.g. 1846) will show an unobtainable amount of shares.

### DIFF
--- a/assets/app/view/game/par.rb
+++ b/assets/app/view/game/par.rb
@@ -29,11 +29,13 @@ module View
           props = {
             style: {
               width: 'calc(17.5rem/6)',
-              padding: '0.2rem 0',
+              padding: '0.2rem',
             },
             on: { click: par },
           }
-          h('button.small.par_price', props, @game.format_currency(share_price.price))
+          purchasable_shares = entity.cash.div(share_price.price)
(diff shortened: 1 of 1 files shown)


michaeljb:
Couple of example screenshots:

Screenshot from 2020-10-27 21-02-20
Screenshot from 2020-10-27 21-02-06

michaeljb (approved):

assets/app/view/game/par.rb
```diff
@@ -29,11 +29,13 @@ def render
           props = {
             style: {
               width: 'calc(17.5rem/6)',
-              padding: '0.2rem 0',
+              padding: '0.2rem',
             },
             on: { click: par },
           }
-          h('button.small.par_price', props, @game.format_currency(share_price.price))
+          purchasable_shares = entity.cash.div(share_price.price)
```
tobymao, line 36: is div divide? if so just do (entity.cash / share_price.price).to_i

PR #2094, opened 2020-10-28
18MEX: Correct bug with hard rust 4 train (fixes #2049)

### LINKED ISSUE #2049: 18MEX: 'Hard rust' does not hard rust
https://18xx.games/game/13900?action=561

Game uses optional rule 8d: Hard rust
But game still soft rust 4 trains when 2nd 6 train bought

Checking Info page for newly created games with this optional rule it looks OK there.

perwestling:
Need to update variant as well when setting obsolete_on, rusts_on.

Update description text for optional rules.

Also remove unnecessary parantheses in the general presentation
of optional rules.

Before:

-- Phase 2 (Operating Rounds: 1, Train Limit: 3, Available Tiles: Yellow ) --
Optional rules used in this game:
* Hard rust: (8d: Hard rust for 4 trains)

After change:

-- Phase 2 (Operating Rounds: 1, Train Limit: 3, Available Tiles: Yellow ) --
Optional rules used in this game:
* Hard rust: 4 trains rust when 6' train is bought

### DIFF
--- a/lib/engine/train.rb
+++ b/lib/engine/train.rb
@@ -7,9 +7,9 @@ module Engine
   class Train
     include Ownable
 
-    attr_accessor :obsolete, :operated, :events, :obsolete_on, :rusts_on
+    attr_accessor :obsolete, :operated, :events
     attr_reader :available_on, :name, :distance, :discount, :multiplier, :rusted, :sym,
-                :variant, :variants
+                :variant, :variants, :obsolete_on, :rusts_on
     attr_writer :buyable
 
     def initialize(name:, distance:, price:, index: 0, **opts)
@@ -53,6 +53,14 @@ module Engine
 
       @variant = @variants[new_variant]
       @variant.each { |k, v| instance_variable_set("@#{k}", v) }
+    end
+
+    def update_end_of_life(rusts_on, obsolete_on)
+      @rusts_on = rusts_on
+      @obsolete_on = obsolete_on
+      @variants = @variants.map do |name, variant|
+        [name, variant.merge(rusts_on: rusts_on, obsolete_on: obsolete_on)]
+      end.to_h
     end
 
     def names_to_prices
--- a/lib/engine/train.rb
+++ b/lib/engine/train.rb
@@ -58,9 +58,7 @@ module Engine
     def update_end_of_life(rusts_on, obsolete_on)
       @rusts_on = rusts_on
       @obsolete_on = obsolete_on
-      @variants = @variants.map do |name, variant|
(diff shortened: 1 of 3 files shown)


lib/engine/train.rb
```diff
@@ -55,6 +55,14 @@ def variant=(new_variant)
       @variant.each { |k, v| instance_variable_set("@#{k}", v) }
     end
 
+    def update_end_of_life(rusts_on, obsolete_on)
+      @rusts_on = rusts_on
+      @obsolete_on = obsolete_on
+      @variants = @variants.map do |name, variant|
+        [name, variant.merge(rusts_on: rusts_on, obsolete_on: obsolete_on)]
+      end.to_h
```
michaeljb, line 63:
```suggestion
      @variants.each { |_, v| v.merge!(rusts_on: rusts_on, obsolete_on: obsolete_on) }
```

lib/engine/train.rb
```diff
@@ -55,6 +55,12 @@ def variant=(new_variant)
       @variant.each { |k, v| instance_variable_set("@#{k}", v) }
     end
 
+    def update_end_of_life(rusts_on, obsolete_on)
```
tobymao, line 58: why is this only called in 18mex? should this just be done inline in 18mex? will this be used for other games?

perwestling: @tobymao This is to change the configuration (when a train is obsoleted, rusted) so I do not know if that many games will use it.
But it is at least used by optional rules in other Mark Derrick games, at least 18AL. I have 18AL optional rules ready since before - just have to update them with using this.

If it were to be done in the game, don't I have to change the accessors (rusts_on, obsolete_on, variants)? The rusts_on and obsolete_on was changed to read-write access because of the 18MEX optional rule need so this is going back to not exposing it.  Also if doing it outside does expose details on how @variants is represented, which might be unnecessary.

But there is another potential here - it is WHY is the @variants created for a game that does not have any variant? The bug was caused by the train class creating variant duplication so maybe I should look into that instead? If I fix it so variant is not created then I can do as I did before, that is making rusts_on and obsolete_on accessible and update them unline in 18MEX. That is probably a better fix.

perwestling: Thinking about it it is probably just to avoid initiating @variants in case :variants is nil.

tobymao: sound sgood

perwestling:
@tobymao I did not get it to work with changing how variants are handled, so I tried your first suggestion instead - doing the work in g_18_mex.rb

So that meant I had to expose variants.

tobymao (approved):

PR #2099, opened 2020-10-29
Improved chat style

dlpierce:
Improves the styling of the homepage chat.

- Bolded username
- Smaller timestamp
- Timestamp only shows date for messages older than one day
- Slight indentation of word wrapped messages (also applied to the in-game log)
- Slightly increased spacing between messages

Screenshot_20201028_224216

### DIFF
--- a/assets/app/view/log.rb
+++ b/assets/app/view/log.rb
@@ -56,6 +56,17 @@ module View
             line_props[:style][:marginTop] = '0.5em' if index.positive?
           end
           h(:div, line_props, line)
+
+        elsif line.is_a?(Hash) # Homepage chat
+          timestamp = Time.at(line[:created_at])
+          time = timestamp.strftime(timestamp + 86_400 < Time.now ? '%F %T' : '%T')
+
+          h('div.chatline', line_props, [
+            h('span.username', line[:user][:name]),
+            h('span.timestamp', time),
+            h('span.message', line[:message]),
+          ])
+
         elsif line.is_a?(Engine::Action::Message)
           h(:div, { style: { fontWeight: 'bold' } }, "#{line.entity.name}: #{line.message}")
         end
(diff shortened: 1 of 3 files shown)


assets/app/view/log.rb
```diff
@@ -56,6 +56,17 @@ def render
             line_props[:style][:marginTop] = '0.5em' if index.positive?
           end
           h(:div, line_props, line)
+
```
tobymao, line 59: remove extra line

assets/app/view/log.rb
```diff
@@ -56,6 +56,17 @@ def render
...
+        elsif line.is_a?(Hash) # Homepage chat
+          timestamp = Time.at(line[:created_at])
+          time = timestamp.strftime(timestamp + 86_400 < Time.now ? '%F %T' : '%T')
+
+          h('div.chatline', line_props, [
```
tobymao, line 64: can we use inline styles here?

ryandriskel:
I'm curious how others feel about the order. I like the direction you're headed, but I preferred the timestamp before the name.

dfannius:
I also strongly prefer timestamp first, as it's a fixed width (except when the date has to be added), so it keeps everything aligned as much as possible (if the timestamp comes first, the names will be aligned, but if the name comes first, the timestamps won't be aligned).

dlpierce:
> I'm curious how others feel about the order. I like the direction you're headed, but I preferred the timestamp before the name.

It was suggested to switch it this way, but I'm impartial. Putting the time first with a fixed width font would ensure the usernames all line up.

dfannius:
Most fonts, hopefully including this one (it looks like it), use a fixed width for numerals, even if they're variable-width otherwise.

dlpierce:
Timestamp comes first now.

tobymao (approved):

PR #2102, opened 2020-10-29
1860 Stock Round

roseundy:
Changes to Stock Round for 1860:
- Implement "layers" for limiting what corps can be ipoed
- Implement selling private company to bank
- Implement buying private company from bank
- Reserve shares associated with a private company
- Restrict private company exchange to ipoed corps
- Selling private company becomes a stock action
- Half-price sales of trainless corp stock
- Max ownership: 100%

This PR makes a few changes to non-1860 files:
- New action
- Changes to UI stock round view to display companies in bank, companies to sell and to prevent selecting unavailable corps
- Change to game base to allow above selecting (always true unless overridden by game) and companies in bank
- Change in buy_sell_shares to allow mode that allows selling shares in SR1

**Selling a private company to the bank:**
sell_company

**Buying a private company from the bank:**
buy_company


lib/engine/game/g_1860.rb
```diff
@@ -136,6 +145,70 @@ def par_prices(corp)
...
+        end.values
+        if layers.empty?
+          1
+        else
+          layers.max + 1
+        end
+      end
+
+      def sorted_corporations
+        # Corporations sorted by some potential game rules
```
jenf, line 166: Remove comment

lib/engine/game/g_1860.rb
```diff
@@ -136,6 +145,70 @@ def par_prices(corp)
...
+          bundles
+        end
+
+        bundles
+      end
+
+      def sell_shares_and_change_price(bundle)
+        corporation = bundle.corporation
+        price = corporation.share_price.price
+        # bundle.share_price = (bundle.price_per_share / 2).to_i if corporation.trains.empty?
```
jenf, line 205: remove commented code

lib/engine/step/g_1860/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,74 @@
...
+        end
+
+        def sell_shares(entity, shares)
+          @game.game_error("Cannot sell shares of #{shares.corporation.name}") unless can_sell?(entity, shares)
+
+          @players_sold[shares.owner][shares.corporation] = :now
+          @game.sell_shares_and_change_price(shares)
+        end
+
+        def bought?
```
jenf, line 52: Isn't this in the base class?

jenf (commented):
Few minor changes, but generally looks good. Would need a toby approval however.

lib/engine/game/g_1860.rb
roseundy: OK

lib/engine/game/g_1860.rb
roseundy: Shoot. missed that one.

lib/engine/step/g_1860/buy_sell_par_shares.rb
roseundy: Yep, nice catch.

roseundy:
Ok, cleaned up.

assets/app/view/game/round/stock.rb
```diff
@@ -120,6 +122,41 @@ def render_buy_tokens
...
+            },
+          }
+
+          @game.companies_in_bank.map do |company|
+            children = []
+            children << h(Company, company: company)
+            children << h('div.margined_bottom', { style: { width: '20rem' } },
+                          render_buy_input(company)) if @selected_company == company
+            h(:div, props, children)
+          end.compact
```
tobymao, line 141: why do you need a compact?

lib/engine/ability/sell_to_bank.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class SellToBank < Base
```
tobymao, line 7: only certain companies can be sold to the bank?

lib/engine/action/sell_company.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Action
+    class SellCompany < Base
+      attr_reader :entity, :price
```
tobymao, line 8: you don't need company?

assets/app/view/game/round/stock.rb
roseundy: Because render_buy_input can return [].

lib/engine/ability/sell_to_bank.rb
roseundy: No, all private companies can. I need an ability so it shows up during the Stock round. Is there another way of doing that?

lib/engine/action/sell_company.rb
roseundy: No, entity is the company. entity.owner is the player who is selling.

tobymao: i don't think this is how actions are generally done.

the entity is the one doing the action ( the entity.owner) and then you need a target, which is the company

lib/engine/ability/sell_to_bank.rb
tobymao: yes, i don't think you need entities, just have them show up in the stock round through a method in the step, sellable_companies or whatever and remove all these ability references

roseundy: But it needs to show up with the other abilities in the UI, doesn't it? How else would it appear?

lib/engine/action/sell_company.rb
roseundy: I was just copying how the buy_shares action is used during an exchange ability on a private company. entity is the company and entity.owner is the player in that case too.

tobymao: i think in buy_shares entity is the person buying the share, and shares refers to the shares you are buying?

lib/engine/game/base.rb
```diff
@@ -1436,6 +1436,14 @@ def round_description(name, round_number = nil)
         description.strip
       end
 
+      def can_select?(_entity)
+        true
+      end
+
+      def companies_in_bank
```
tobymao, line 1443: this isn't necessary, you can just do game.companies.select { ... }

lib/engine/game/base.rb
```diff
@@ -1436,6 +1436,14 @@ def round_description(name, round_number = nil)
         description.strip
       end
 
+      def can_select?(_entity)
```
tobymao, line 1439: why do you need can_select?

lib/engine/game/base.rb
roseundy: Ok.

lib/engine/game/base.rb
roseundy: I needed a way to prevent parring a corporation that is in an unavailable layer (1860 specific thing). If can_select? returns false, then the corp isn't selectable in the list. I thought this was a cool way of doing it.  The alternative was to add a can_ipo? or similar qualifier to this line:
`children << render_input if @selected_corporation == corporation`

Since I didn't want to modify all the game-specific steps that use the UI:Round, I had to made it part of game/base.

I can switch to do the latter, or give the method a different name.

roseundy: Ok, switching to the approach where the corp is selectable, but it won't have any par buttons rendered unless @game.corporation_available? returns true (which it will in the game/base definition).

tobymao (approved):

PR #2104, opened 2020-10-29
1817 Mines restrictions

### LINKED ISSUE #2082: [1817] Coal tile placement restrictions
When you use a coal private, the games allow you to place any tile in any hex in the list. It should only allow placing a tile you would normally be able to place. Furthermore it needs to be directly adjacent to a city

jenf:
Mines can only be placed next to existing track
Mines must be connected to a revenue location
or could potentially be upgraded to be connected

fixes #2082

### DIFF
--- a/lib/engine/step/g_1817/special_track.rb
+++ b/lib/engine/step/g_1817/special_track.rb
@@ -29,7 +29,8 @@ module Engine
         def available_hex(entity, hex)
           return super if entity.company? && entity.id == 'PSM'
 
-          return if ability(entity)&.hexes&.any? && !ability(entity)&.hexes&.include?(hex.id)
+          hexes = ability(entity)&.hexes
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_1817/special_track.rb
```diff
@@ -25,6 +25,42 @@ def process_lay_tile(action)
           end
           step.laid_track = step.laid_track + 1
         end
+
+        def available_hex(entity, hex)
+          return super if entity.company? && entity.id == 'PSM'
+
+          return if ability(entity)&.hexes&.any? && !ability(entity)&.hexes&.include?(hex.id)
```
tobymao, line 32: cleaner if you do two lines
_ability = ability(entity)
return if...

tobymao (approved):

PR #2112, opened 2020-10-30
1828: Incremental development of abilities for private companies and corporations

crericha:
**Summary of changes**
M&H: Add support to exchange for any share. Includes a change to how its displayed (see below). Allow players to continue to hold more than 60% of a corporation if done via exchange or when the company's stock market position used to allow and now doesn't.
C&P: Treat like a minor. 
E&K: Add support for two tile-laying abilities in one private company. Allow tile-laying abilities to cost money.
B&M, NYH: Display extra tile lay as an ability, but implement via game class's tile_lays method

**Exchange view sample**

Current:
image

New:
image

New with multiple shares:
image

### DIFF
--- a/lib/engine/ability/exchange.rb
+++ b/lib/engine/ability/exchange.rb
@@ -8,7 +8,7 @@ module Engine
       attr_reader :corporation, :from
 
       def setup(corporation:, from:)
-        @corporation = corporation == 'any' ? corporation.to_sym : corporation
+        @corporation = corporation
         @from = Array(from).map(&:to_sym)
       end
     end
--- a/lib/engine/ability/exchange.rb
+++ b/lib/engine/ability/exchange.rb
@@ -8,7 +8,7 @@ module Engine
       attr_reader :corporation, :from
 
       def setup(corporation:, from:)
-        @corporation = corporation
+        @corporation = corporation == 'any' ? corporation.to_sym : corporation
         @from = Array(from).map(&:to_sym)
       end
     end
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -52,7 +52,7 @@ module Engine
       @float_percent = opts[:float_percent] || 60
       @floated = false
       @max_ownership_percent = opts[:max_ownership_percent] || 60
-      @hold_above_max_ownership_percent_ok = opts[:hold_above_max_ownership_percent_ok] || false
+      @can_hold_above_max = opts[:can_hold_above_max] || false
       @min_price = opts[:min_price]
       @always_market_price = opts[:always_market_price] || false
       @needs_token_to_par = opts[:needs_token_to_par] || false
@@ -164,7 +164,7 @@ module Engine
 
     # Is it legal to hold percent shares in this corporation?
     def holding_ok?(share_holder, extra_percent = 0)
-      return true if @hold_above_max_ownership_percent_ok
+      return true if @can_hold_above_max
 
       percent = share_holder.percent_of(self) + extra_percent
       %i[orange brown].include?(@share_price&.color) || percent <= @max_ownership_percent
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -52,6 +52,7 @@ module Engine
       @float_percent = opts[:float_percent] || 60
       @floated = false
       @max_ownership_percent = opts[:max_ownership_percent] || 60
+      @hold_above_max_ownership_percent_ok = opts[:hold_above_max_ownership_percent_ok] || false
       @min_price = opts[:min_price]
       @always_market_price = opts[:always_market_price] || false
       @needs_token_to_par = opts[:needs_token_to_par] || false
@@ -163,6 +164,8 @@ module Engine
 
     # Is it legal to hold percent shares in this corporation?
     def holding_ok?(share_holder, extra_percent = 0)
+      return true if @hold_above_max_ownership_percent_ok
+
       percent = share_holder.percent_of(self) + extra_percent
       %i[orange brown].include?(@share_price&.color) || percent <= @max_ownership_percent
     end
--- a/lib/engine/game/g_1828.rb
+++ b/lib/engine/game/g_1828.rb
@@ -146,7 +146,7 @@ module Engine
       def event_close_companies!
         super
 
-        @minors.dup.each { |minor| remove_minor(minor) }
+        @minors.dup.each { |minor| remove_minor!(minor) }
       end
 
       def event_remove_corporations!
@@ -163,7 +163,7 @@ module Engine
         @phase.current[:name] == 'Purple'
       end
 
-      def remove_minor(minor)
+      def remove_minor!(minor)
         @round.force_next_entity! if @round.current_entity == minor
         minor.spend(minor.cash, @bank) if minor.cash.positive?
         @minors.delete(minor)
--- a/lib/engine/game/g_1828.rb
+++ b/lib/engine/game/g_1828.rb
@@ -72,6 +72,8 @@ module Engine
       def stock_round
         Round::G1828::Stock.new(self, [
           Step::DiscardTrain,
+          Step::Exchange,
+          Step::SpecialTrack,
           Step::G1828::BuySellParShares,
         ])
       end
@@ -79,6 +81,7 @@ module Engine
       def operating_round(round_num)
         Round::Operating.new(self, [
           Step::Bankrupt,
+          Step::Exchange,
           Step::DiscardTrain,
           Step::SpecialTrack,
           Step::BuyCompany,
@@ -104,6 +107,10 @@ module Engine
         sm.enable_par_price(79)
 
         sm
+      end
+
+      def corporation_opts
+        { hold_above_max_ownership_percent_ok: true }
       end
 
       def init_round_finished
--- a/lib/engine/game/g_1828.rb
+++ b/lib/engine/game/g_1828.rb
@@ -120,7 +120,7 @@ module Engine
       end
 
       def corporation_opts
-        { hold_above_max_ownership_percent_ok: true }
+        { can_hold_above_max: true }
       end
 
       def init_round_finished
--- a/lib/engine/game/g_1828.rb
+++ b/lib/engine/game/g_1828.rb
@@ -84,17 +84,18 @@ module Engine
           Step::Exchange,
           Step::DiscardTrain,
           Step::SpecialTrack,
-          Step::BuyCompany,
+          Step::G1828::BuyCompany,
           Step::Track,
           Step::Token,
           Step::Route,
-          Step::Dividend,
+          Step::G1828::Dividend,
           Step::BuyTrain,
           [Step::BuyCompany, blocks: true],
         ], round_num: round_num)
       end
 
       def setup
+        setup_minors
         remove_extra_private_companies
         remove_extra_trains
       end
@@ -107,6 +108,15 @@ module Engine
         sm.enable_par_price(79)
 
         sm
+      end
+
+      EXTRA_TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false, cost: 40 }].freeze
+      EXTRA_TILE_LAY_CORPS = %w[B&M NYH].freeze
+
+      def tile_lays(entity)
+        return self.class::EXTRA_TILE_LAYS if EXTRA_TILE_LAY_CORPS.find(entity.id)
+
+        super
       end
 
       def corporation_opts
@@ -133,6 +143,12 @@ module Engine
         stock_market.enable_par_price(120)
       end
 
+      def event_close_companies!
+        super
+
+        @minors.dup.each { |minor| remove_minor(minor) }
+      end
+
       def event_remove_corporations!
         @log << "-- Event: #{EVENTS_TEXT['remove_corporations'][1]}. --"
         @corporations.reject(&:ipoed).each do |corporation|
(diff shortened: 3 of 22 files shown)


lib/engine/round/operating.rb
```diff
@@ -14,7 +14,7 @@ def name
       end
 
       def select_entities
-        @game.minors + @game.corporations.select(&:floated?).sort
+        @game.minors.select(&:floated?) + @game.corporations.select(&:floated?).sort
```
crericha, line 17: This is a bug. An unfloated minor shouldn't operate and if it doesn't have an owner, it'll crash the game.

lib/engine/step/special_track.rb
```diff
@@ -30,13 +30,14 @@ def blocks?
...
         tile_ability = ability(action.entity)
-        lay_tile(action, spender: action.entity.owner)
+        lay_tile(action, spender: action.entity.owner, tile_ability: tile_ability)
         check_connect(action, tile_ability)
         tile_ability.use!
       end
 
       def available_hex(entity, hex)
         return if ability(entity)&.hexes&.any? && !ability(entity)&.hexes&.include?(hex.id)
+        return if ability(entity)&.reachable && !@game.graph.connected_hexes(entity.owner)[hex]
```
crericha, line 40: Improvement: If the tile laying ability requires the hex to be reachable, only show the reachable hexes. Was originally handled in tracker via a game error (code still there in case), but, in my opinion, better to not allow in the first place.

lib/engine/step/tracker.rb
```diff
@@ -73,7 +73,7 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
         free = false
         discount = 0
 
-        entity.abilities(:tile_lay) do |ability|
+        entity.abilities(:tile_lay, tile_ability ? tile_ability.when : nil) do |ability|
```
crericha, line 76: I'll admit, this is kind of a hack. Let me explain how I got there and maybe someone else has a better idea.

The reason I need to change the code is because the special_track class supports the same private company having both a "when sold" track lay and an anytime track lay, which the E&K needs. However, the tracker module isn't aware of this and will try to use all the tile_laying abilities, which breaks the E&K. In summary, I need a way to leverage the special_track behavior without breaking the normal track step that also uses the tracker module.

The abilities method has its own hack, where if a single ability is found, it returns that and if multiple abilities are found, it returns an array. At first, I wrote code that supported those possibilities. In my opinion, it was quite messy and hard to read.

What makes the abilities return easier to handle is that it also yields, so you can simply add a block that handles both the single and multiple case...simple. I tried to leverage by writing a method in tracker called tile_lay_abilities, which took in the block and passed to the abilities method. I then overrode the method in the special_track class to provide the correct behavior for private company track lay abilities. It worked, but I worried about the tight coupling between this code, the new method, and the abilities method from the abilities module, so I abandoned in favor of what is in the code now.

lib/engine/ability/exchange.rb
```diff
@@ -8,7 +8,7 @@ class Exchange < Base
       attr_reader :corporation, :from
 
       def setup(corporation:, from:)
-        @corporation = corporation
+        @corporation = corporation == 'any' ? corporation.to_sym : corporation
```
tobymao, line 11: keep the types the same, so always do sym or always expect str

lib/engine/corporation.rb
```diff
@@ -52,6 +52,7 @@ def initialize(sym:, name:, **opts)
       @float_percent = opts[:float_percent] || 60
       @floated = false
       @max_ownership_percent = opts[:max_ownership_percent] || 60
+      @hold_above_max_ownership_percent_ok = opts[:hold_above_max_ownership_percent_ok] || false
```
tobymao, line 55: name kinda long

lib/engine/game/g_1828.rb
```diff
@@ -140,8 +163,24 @@ def custom_end_game_reached?
         @phase.current[:name] == 'Purple'
       end
 
+      def remove_minor(minor)
```
tobymao, line 166: remove_minor!

lib/engine/step/special_track.rb
tobymao: is this only called from the UI? if so, that's ok, there's a performance implication to calling the graph

if it's run in a hot loop of actions, we need to put this in restrictions

lib/engine/step/g_1817/track.rb
```diff
@@ -15,7 +15,7 @@ def setup
           @hex = nil
         end
 
-        def lay_tile(action, extra_cost: 0, entity: nil)
+        def lay_tile(action, extra_cost: 0, entity: nil, tile_ability: nil)
```
tobymao, line 18: does tile_ability always come from action_entity? if it does, you can infer it here instead of passing it in

lib/engine/step/tracker.rb
tobymao: hm i think i favor using the yield as that was designed to handle multiple abilities of the same type, could i see what that looks like?

lib/engine/corporation.rb
crericha: I agree. I went for clarity over length. Any suggestions? can_hold_above_max maybe?

lib/engine/step/g_1817/track.rb
crericha: What I'm really passing in is the "when" argument from ability. I could change to only be that. Note: this will be removed completely if I pass in the block instead.

lib/engine/step/special_track.rb
crericha: It gets called by the UI in a loop for each hex when a private company is selected and has a tile laying ability that is reachable. Is there a way to cache the result until an action is taken?

lib/engine/step/tracker.rb
crericha: Here is the diff.

diff --git a/lib/engine/step/g_1818/track.rb b/lib/engine/step/g_1817/track.rb
index 19b7520a..6d868a25 100644
--- a/lib/engine/step/g_1817/track.rb
+++ b/lib/engine/step/g_1817/track.rb
@@ -15,7 +15,7 @@ module Engine
           @hex = nil
         end
 
-        def lay_tile(action, extra_cost: 0, entity: nil, tile_ability: nil)
+        def lay_tile(action, extra_cost: 0, entity: nil)
           @game.game_error('Cannot lay and upgrade the same tile in the same turn') if action.hex == @hex
           @game.game_error('Cannot upgrade mines') if action.hex.assigned?('mine')
           super
diff --git a/lib/engine/step/g_1882/track.rb b/lib/engine/step/g_1882/track.rb
index cf9579a2..ccffdc74 100644
--- a/lib/engine/step/g_1882/track.rb
+++ b/lib/engine/step/g_1882/track.rb
@@ -10,7 +10,7 @@ module Engine
       class Track < Track
         include NwrTrackBonus
 
-        def lay_tile(action, extra_cost: 0, entity: nil, tile_ability: nil)
+        def lay_tile(action, extra_cost: 0, entity: nil)
           entity ||= action.entity
           super
           gain_nwr_bonus(action.tile, entity)
diff --git a/lib/engine/step/g_1889/special_track.rb b/lib/engine/step/g_1889/special_track.rb
index 78cd62c6..6d597cef 100644
--- a/lib/engine/step/g_1889/special_track.rb
+++ b/lib/engine/step/g_1889/special_track.rb
@@ -20,7 +20,7 @@ module Engine
           return super unless action.entity == @company
 
           lay_tile(action, spender: @round.company_seller)
-          ability(action.entity).use!
+          tile_lay_abilities(action.entity).use!
         end
 
         def process_pass(action)
diff --git a/lib/engine/step/g_18_ms/lay_tile_with_chattanooga_check.rb b/lib/engine/step/g_18_ms/lay_tile_with_chattanooga_check.rb
index 98f14a1f..2bde56f1 100644
--- a/lib/engine/step/g_18_ms/lay_tile_with_chattanooga_check.rb
+++ b/lib/engine/step/g_18_ms/lay_tile_with_chattanooga_check.rb
@@ -6,7 +6,7 @@ module Engine
       module LayTileWithChattanoogaCheck
         ACTION_HEX_OF_INTEREST = 'B10'
 
-        def lay_tile(action, extra_cost: 0, entity: nil, spender: nil, tile_ability: nil)
+        def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
           super
 
           entity ||= action.entity
diff --git a/lib/engine/step/special_track.rb b/lib/engine/step/special_track.rb
index bc6dc610..7f9cca8e 100644
--- a/lib/engine/step/special_track.rb
+++ b/lib/engine/step/special_track.rb
@@ -11,7 +11,7 @@ module Engine
       ACTIONS = %w[lay_tile].freeze
 
       def actions(entity)
-        return [] unless ability(entity)
+        return [] unless tile_lay_abilities(entity)
 
         ACTIONS
       end
@@ -25,25 +25,26 @@ module Engine
       end
 
       def blocks?
-        ability(@company)&.blocks
+        tile_lay_abilities(@company)&.blocks
       end
 
       def process_lay_tile(action)
-        tile_ability = ability(action.entity)
-        lay_tile(action, spender: action.entity.owner, tile_ability: tile_ability)
-        check_connect(action, tile_ability)
-        tile_ability.use!
+        ability = tile_lay_abilities(action.entity)
+        lay_tile(action, spender: action.entity.owner)
+        check_connect(action, ability)
+        ability.use!
       end
 
-        return if ability(entity)&.hexes&.any? && !ability(entity)&.hexes&.include?(hex.id)
-        return if ability(entity)&.reachable && !@game.graph.connected_hexes(entity.owner)[hex]
+        ability = tile_lay_abilities(entity)
+        return if ability&.hexes&.any? && !ability&.hexes&.include?(hex.id)
+        return if ability&.reachable && !@game.graph.connected_hexes(entity.owner)[hex]
 
         @game.hex_by_id(hex.id).neighbors.keys
       end
 
-        return [] unless (tile_ability = ability(entity))
+        return [] unless (tile_ability = tile_lay_abilities(entity))
 
         tiles = tile_ability.tiles.map { |name| @game.tiles.find { |t| t.name == name } }
         tiles = @game.tiles.uniq(&:name) if tile_ability.tiles.empty?
@@ -53,12 +54,12 @@ module Engine
           .select { |t| @game.phase.tiles.include?(t.color) && @game.upgrades_to?(hex.tile, t, tile_ability.special) }
       end
 
-      def ability(entity)
+      def tile_lay_abilities(entity, &block)
         return unless entity&.company?
 
-        ability = entity.abilities(:tile_lay, 'sold') if @round.respond_to?(:just_sold_company) &&
+        ability = entity.abilities(:tile_lay, 'sold', &block) if @round.respond_to?(:just_sold_company) &&
           entity == @round.just_sold_company
-        ability || entity.abilities(:tile_lay, 'track')
+        ability || entity.abilities(:tile_lay, 'track', &block)
       end
 
       def check_connect(action, tile_ability)
diff --git a/lib/engine/step/tracker.rb b/lib/engine/step/tracker.rb
index 353ea824..19589d66 100644
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -38,7 +38,11 @@ module Engine
         @laid_track += 1
       end
 
-      def lay_tile(action, extra_cost: 0, entity: nil, spender: nil, tile_ability: nil)
+      def tile_lay_abilities(entity, &block)
+        entity.abilities(:tile_lay, &block)
+      end
+
+      def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
         entity ||= action.entity
         spender ||= entity
         tile = action.tile
@@ -73,7 +77,7 @@ module Engine
         free = false
         discount = 0
 
-        entity.abilities(:tile_lay, tile_ability ? tile_ability.when : nil) do |ability|
+        tile_lay_abilities(entity) do |ability|
           next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
 
           @game.game_error("Track laid must be connected to one of #{spender.id}'s stations") if ability.reachable &&

lib/engine/ability/exchange.rb
crericha: This is where the json string is converted to a symbol. It's used as a symbol throughout. Was there a different place you were thinking for the conversion?

lib/engine/step/g_1817/track.rb
crericha: Removed.

lib/engine/step/tracker.rb
crericha: Refactored to the block option.

lib/engine/step/special_track.rb
crericha: Actually, it is already cached.

    def reachable_hexes(corporation)
      compute(corporation) unless @reachable_hexes[corporation]
      @reachable_hexes[corporation]
    end

lib/engine/ability/exchange.rb
tobymao: i'd just leave at as @corporation = corporation.to_sym then

lib/engine/corporation.rb
tobymao: yea
sounds goods

lib/engine/step/special_track.rb
```diff
@@ -25,24 +25,26 @@ def active_entities
...
         lay_tile(action, spender: action.entity.owner)
-        check_connect(action, tile_ability)
-        tile_ability.use!
+        check_connect(action, ability)
+        ability.use!
       end
 
       def available_hex(entity, hex)
-        return if ability(entity)&.hexes&.any? && !ability(entity)&.hexes&.include?(hex.id)
+        ability = tile_lay_abilities(entity)
```
tobymao, line 39: return unless (ability = ...)
then you don't need all the safe operators

lib/engine/ability/exchange.rb
crericha: Using a symbol will break individual certificates. I went ahead and used a string for any, so the type is the same regardless of value.

PR #2125, opened 2020-10-30
1817: Disallow bids that aren't valid combinations of companies

### LINKED ISSUE #2117: [1817] - Player bidding on new companies an amount s/he cannot pay
It happened in this game https://www.18xx.games/game/13922

The player Merlo Piccante has 0 cash and two privates, one with face value 120, one with 30, and he was able to bid 135, winning the auction, but now he's stuck, because his bid power is either 120 or 150, no inbetween.
He souldn't have been allowed to bid anything else but 120 and 150.

jenf:
A player should only be able to bid the values of the companies combined.
(fixes #2117)

### DIFF
--- a/lib/engine/step/g_1817/buy_sell_par_shares.rb
+++ b/lib/engine/step/g_1817/buy_sell_par_shares.rb
@@ -175,10 +175,25 @@ module Engine
           end
         end
 
+        def available_company_options(entity)
+          value = entity.companies.map(&:value)
+          # Rather naive way of doing this, but the number of potential values is small
+          options = [0]
+          value.each do |v|
+            options += options.map { |o| o + v }
+          end
+          options.uniq
+        end
+
         def add_bid(action)
           entity = action.entity
           corporation = action.corporation
           price = action.price
+
+          options = available_company_options(entity)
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_1817/buy_sell_par_shares.rb
```diff
@@ -175,11 +175,26 @@ def process_bid(action)
           end
         end
 
+        def available_company_options(entity)
+          value = entity.companies.map(&:value)
+          # Rather naive way of doing this, but the number of potential values is small
+          options = [0]
+          value.each do |v|
```
tobymao, line 182: are you trying to do value.combination ?

tobymao: values = entity.companies.map(&:values)
(1..values.size).flat_map { |size| values.combination(size).to_a }

jenf: Yeah, I'm away until Monday so won't be able to pick it up. I'm tempted to rework a bit more of this so it's automatic if players have no choice. Would then do a migration for existing games

tobymao (approved):

PR #2128, opened 2020-10-31
In case different percentages is in market, separate buy buttons

### LINKED ISSUE #2097: [18MEX] Cannot buy 10% NdM with the both 5% in market
Game link: https://www.18xx.games/game/14188?action=611

Both 5% shares of NdM are in the market, along with a 10% share. I (auzzy) want to buy the 10% share. I only have 12 certs out of the 15 cert limit, so I have space. I also have plenty of money, so I can afford it. However, when I click on NdM, I am only presented with a single button labeled "Buy Market Share". When I buy the share, it purchases a 5% share (as noted in the log at the link above).

perwestling:
This is currently only for 18MEX but it is general code that will
be used as soon as there are different percentage shares in the
share pool.

In case there is just one type of shares the behavior will not be changed.

In case there are more the one type of percentage in the share pool
the percentage of the market share will be shown on the Buy button.
This is also presented even if only one of the types are buyable.

So, using 18MEX as an example:

* If 5% and 10% are in market pool, and player can buy both
  Show: "Buy 5% Market Share" "Buy 10% Market Share"

* If 5% and 10% are in market pool, but player can only buy 5%
  Show: "Buy 5% Market Share"

* If just 5% or 10% shares are in market pool
  Show "Buy Market Share"

* And this means - if not 18MEX
  Show "Buy Market Share"

This fixes #2097

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -16,7 +16,6 @@ module View
         current_entity = step.current_entity
 
         ipo_share = @corporation.shares[0]
-        pool_share = @game.share_pool.shares_by_corporation[@corporation][0]
 
         children = []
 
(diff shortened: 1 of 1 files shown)


perwestling:
If you want to know how this change will look in the stock market, check out #2097 which include screen shots for the different cases.
But for all current titles implemented in 18xx.games, there should be no presentation change.

PR #2145, opened 2020-10-31
Fix city assignment on upgrade and set 18CO Token Cities

ryandriskel:
Without the _any?_ check, the first city will always match regardless of which city index is specified in the config. We should instead assume the city has the same index. In the case where that city's index is no longer available on the tile, use index 0, which is the same as the original code would have returned.

Ref: https://github.com/tobymao/18xx/issues/1836
DSL, DRG, DPAC and CM have fixed home token locations on tiles that have multiple stations.

---

DSL is assigned the top left position correctly.
Screen Shot 2020-10-31 at 4 06 31 PM

**Before the fix**, the token would jump to the bottom right because no exits is a subset of exits 0+5, even though the city index is set to 1 in the config file.
Screen Shot 2020-10-31 at 4 06 40 PM

**After the fix**, the token is in the correct position.
Screen Shot 2020-10-31 at 4 07 05 PM

Green Tile upgrade is also grouped correctly ( Different game shown here, so different tokens in play )
Screen Shot 2020-10-31 at 5 07 04 PM

Brown Tile also merges correctly
Screen Shot 2020-10-31 at 5 08 36 PM

### DIFF
--- a/lib/engine/config/game/g_18_co.rb
+++ b/lib/engine/config/game/g_18_co.rb
@@ -525,6 +525,7 @@ module Engine
 				0,
 				40
 			],
+			"city": 2,
 			"coordinates": "E15",
 			"color": "purple"
 		},
@@ -539,6 +540,7 @@ module Engine
 				0,
 				40
 			],
+			"city": 1,
 			"coordinates": "E15",
 			"color": "green"
 		},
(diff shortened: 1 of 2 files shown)


ryandriskel:
I should note that I have a slight concern that there are edge cases out there (looking at you OO tiles), but I have to identify any.

ryandriskel:
Converted to draft as specs were passing but no longer are. I wonder if this is the edge case I am looking for?!

tobymao (approved):

PR #2154, opened 2020-11-01
In case different percentages is in market, separate buttons (#2128)

### LINKED ISSUE #2097: [18MEX] Cannot buy 10% NdM with the both 5% in market

perwestling:
This is currently only for 18MEX but it is general code that will
be used as soon as there are different percentage shares in the
share pool.

In case there is just one type of shares the behavior will not be changed.

In case there are more the one type of percentage in the share pool
the percentage of the market share will be shown on the Buy/Exchange button.
This is also presented even if only one of the types are buyable.

So, using 18MEX, and Buy, as an example:

* If 5% and 10% are in market pool, and player can buy both
  Show: "Buy 5% Market Share" "Buy 10% Market Share"

* If 5% and 10% are in market pool, but player can only buy 5%
  Show: "Buy 5% Market Share"

* If just 5% or 10% shares are in market pool
  Show "Buy Market Share"

* And this means - if not 18MEX, but still Buy
  Show "Buy Market Share"

* For Exchange a similar behavior

This fixes #2097

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -16,7 +16,7 @@ module View
         current_entity = step.current_entity
 
         ipo_share = @corporation.shares[0]
-        pool_share = @game.share_pool.shares_by_corporation[@corporation][0]
+        pool_shares = @game.share_pool.shares_by_corporation[@corporation].group_by(&:percent).values.collect(&:first)
 
         children = []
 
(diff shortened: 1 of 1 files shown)


perwestling:
Have verified this fix by buying/exchanging B shares in https://18xx.games/game/12466

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -16,7 +16,7 @@ def render
         current_entity = step.current_entity
 
         ipo_share = @corporation.shares[0]
-        pool_share = @game.share_pool.shares_by_corporation[@corporation][0]
+        pool_shares = @game.share_pool.shares_by_corporation[@corporation].group_by(&:percent).values.collect(&:first)
```
ryandriskel, line 19: I suggest using map instead of collect as that seems to be the pattern in the codebase.

tobymao: agreed, i can make this change locally

tobymao (approved):

PR #2160, opened 2020-11-01
Adding several lib-engine-config-game files

philcampeau:
Using the 18xx-maker export-rb function, I created a /lib/engine/config/game file for:

1817NA
1832
1835
1836jr (1856 variant)
1841
1849
1856
1870


tobymao:
i don't want this in until they are ready, closing for now because it's just dead weight

PR #2161, opened 2020-11-01
Update

philcampeau:
corrected the trailing new line missing from 1817NA and 1856 in pull #2160


philcampeau:
Not sure if I handled this correctly. I was initially missing a trailing newline break in #2160 , so i created a new pull request with the corrections done.

perwestling:
Probably not. The title of this PR should be about the main thing, as is the description.

PR #2164, opened 2020-11-01
1860 Implement Fishbourne Ferry Company

roseundy:
Fishbourne Ferry Company now closes all other companies
It become available on the first 6 train purchase

Also implement logging of when new corporation layer becomes available.

Non-1860 code change: modify company.rb to allow closing when not owned by player or corp (i.e. the bank)


roseundy:
I wonder if this code would be better in Company::close!

     ` owner.companies.delete(self) if owner.respond_to?(:companies)
`

lib/engine/company.rb
```diff
@@ -47,7 +47,7 @@ def close!
       all_abilities.dup.each { |a| remove_ability(a) }
       return unless owner
 
-      owner.companies.delete(self)
+      owner.companies.delete(self) if owner.player? || owner.corporation?
```
tobymao, line 50: why was this added? will thins break other games?

lib/engine/game/g_1860.rb
```diff
@@ -204,6 +215,15 @@ def sell_shares_and_change_price(bundle)
         num_shares.times { @stock_market.move_left(corporation) }
         log_share_price(corporation, price)
       end
+
+      def close_other_companies!(company)
+        any = @companies.reject { |c| c == company }.reject(&:closed?)
+        return unless any
```
tobymao, line 221: return unless @companies.any? { |c| c == company && ... }

lib/engine/company.rb
roseundy: Because I need to close a company even if it's owned by the bank. The bank doesn't have a companies method, so the original code will blow up.  This shouldn't break anything since the only things that have companies are players and corporations.

I could write it like this:
`owner.companies.delete(self) if owner.respond_to?(:companies)`
Would that be preferable?

lib/engine/game/g_1860.rb
roseundy: Ok. Left over from a previous approach.

PR #2166, opened 2020-11-02
Add temporary announcement of the 1846 tournament

daniel-sousa-me:
I asked Toby for permission.

The announcement will disappear automatically and then I'll do another PR to remove that code.

### DIFF
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -16,17 +16,9 @@ module View
     end
 
     def render_notification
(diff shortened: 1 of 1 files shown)


perwestling (commented):
After 2020-11-05 is the tournament_announcement set? What will happen on line 27 if not set?

daniel-sousa-me:
It will just include the empty string in the announcements, but I'll change it so that the variable is always defined

tobymao:
don't put the time, just put the string, and make a new PR when you want to remove it

PR #2173, opened 2020-11-02
1817: Add par values to stock market (fixes #2155)

### LINKED ISSUE #2155: 1817 Stock market should indicate par prices (ie max 200)
1817 Stock market - par prices indicated (ie max 200).

jenf:
This also refactors and removes the notion of color from the stock market itself.

The game can now override the color (Par is generally red, but on 1817 this is used for liquidation)

fixes #2155


roseundy:
This is great. I've always wondered about the wisdom of hard-coded colors in the stock market.

tobymao (approved):

PR #2188, opened 2020-11-04
Initial 1870 configuration

daniel-sousa-me:


assets/app/view/game/part/upgrade.rb
```diff
@@ -73,6 +73,7 @@ def render_part
               swamp: svg(delta_x: delta_x, delta_y: delta_y, icon: 'swamp'),
               desert: svg(delta_x: delta_x, delta_y: delta_y, icon: 'cactus'),
               lake: svg(delta_x: delta_x, delta_y: delta_y, icon: 'lake'),
+              mississippi: svg(delta_x: delta_x, delta_y: delta_y, icon: 'mississippi'),
```
tobymao, line 76: this doesn't seem like it should be here

daniel-sousa-me: Where can I put it? It's just a modified "water" icon, so that we can distinguish the Mississippi river from the other rivers. Maybe I can give it a more generic name like "river"?

tobymao: yea, i think 1817 may do something similar?

daniel-sousa-me: 1817 added the "lake" icon, which I'm also using. There's too much water on 1870 :P

tobymao (approved):

PR #2203, opened 2020-11-05
Don't imply shares can be over limit in par step

dlpierce:
Enhances the par share count button helper text by limiting the amount of shares shown to be purchasable to the corporation ownership limit. Appends an "L" flag in this case.

Screenshot_20201105_011651

### DIFF
--- a/assets/app/view/game/par.rb
+++ b/assets/app/view/game/par.rb
@@ -33,7 +33,8 @@ module View
             },
             on: { click: par },
           }
-          purchasable_shares = [(entity.cash / share_price.price).to_i, @corporation.max_ownership_percent / @corporation.total_shares].min
+          purchasable_shares = [(entity.cash / share_price.price).to_i,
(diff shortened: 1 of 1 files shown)


assets/app/view/game/par.rb
```diff
@@ -33,8 +33,10 @@ def render
             },
             on: { click: par },
           }
-          purchasable_shares = (entity.cash / share_price.price).to_i
-          text = "#{@game.format_currency(share_price.price)} (#{purchasable_shares})"
+          purchasable_shares = [(entity.cash / share_price.price).to_i, @corporation.max_ownership_percent / 10].min
```
tobymao, line 36: shares aren't always 10, i think you want total_shares

dlpierce: Thanks, that makes sense. I poked around in 1817 to see if that case might come up, but didn't see anywhere that it did. I'll get it changed.

tobymao (approved):

PR #2211, opened 2020-11-05
1817 Acqusition and liquidation fixes

### LINKED ISSUE #1828: 1817: Log does not get eminated when changing share price from $40 normal to $40 acquisition
**Describe the bug**
There is a lack of logging as the share prices are identical when changing share price from $40 to $40 moving into the acquisition area.

This area should probably be revisited and pass the share object in rather than the price

### LINKED ISSUE #2179: 1817 entering liquidation during M&A round
The rules are clear that corporations that move into the grey acquisition area during an M&A round are not auctioned during that same round. They are silent about what happens to corporations that move into the red liquidation area during an M&A round (which can happen if a liquidation drives someone bankrupt).

Craig Bartell, the designer, has stated several times that his intent was that the M&A round auction order is set at the beginning of the M&A round, and that companies that enter liquidation during an M&A round are not auctioned until the next M&A round (i.e. they sit in limbo for the current M&A round).

As I noted, the rules do not address this topic, so I think it is fine for 18xx.games developers to agree on which interpretation they would like to use. However, we should document that choice in the implementation notes in the wiki. And it should be an actual choice by the developers, and not something that happened because we copied the rr18xx model, or simply just did it that way.

According to reports (which I have not yet verified) 18xx.games currently implements the "auction them this round" version.

### LINKED ISSUE #2194: [1817] Create a separation between Liquidation and Acquisition
With the stock market as is, it gives the impression to new players that it is possible to fall from acquisition into liquidation just by moving a step down.

It would be nice to have some space between than or a thicker line.

Another option could be putting the liquidation spot on the right of 600. I believe no one will think that you can go up from 600 into liquidation xD

### LINKED ISSUE #2195: [1817] Companies in acquisition are worth $0
Companies in acquisition count as $40 for jumping purposes, but their value is $0 (they can't be bought sold during an SR, but for the M&A and the end game they count as $0)

### LINKED ISSUE #2210: [1817] Round number for M&A
On the tile on top, during an OR we have "Operating Round 2.1 (of 2)", but on M&A it only shows "Merger Round 2", which always confuses me. We should add the rest of the info there.

jenf:
Add round number for M&A rounds (fixes #2210)
Acquisition stocks are worth $0, but paying out >$40 moves to right (fixes #2195)
Log share price move into acquisition (fixes #1828)
Add separation between liquidation and acquisition on stock market (fixes #2194)
Don't do acquisitions for things that entered liquidation/acquisition during the M&A round (fixes #2179)
Highlight liquidation and acquisition on corporate shares (improves 2186)


daniel-sousa-me:
Wow

Thank you very much for being so fast implementing my random little recommendations! You've been doing an amazing job on this monster of a game :)

tobymao (approved):

PR #2212, opened 2020-11-05
1860 tile placement and upgrades

roseundy:
- Home track tile lay when corp is floated if no tile on home hex
- Yellow tile placement rules (2x, unless one is a city)
- Upgrade rules (must be able to reach with a single train corporation owns)
- Constrained rotation for (some) N, M and R tiles
- Fixed Halt to allow upgrade to Town.

### DIFF
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -138,24 +138,6 @@ module Engine
         check_new_layer
       end
 
-      def place_home_track(corporation)
-        hex = hex_by_id(corporation.coordinates)
-        tile = hex&.tile
-
-        # skip if a tile is already in home location
-        return unless tile.color == :white
-
-        @log << "#{corporation.name} must choose tile for home location"
-
-        @round.pending_tracks << {
-          entity: corporation,
-          hexes: [hex],
-        }
-
-        @round.clear_cache!
-        nil
-      end
-
       def event_fishbourne_to_bank!
         ffc = @companies.find { |c| c.sym == 'FFC' }
         ffc.owner = @bank
@@ -273,14 +255,14 @@ module Engine
       end
 
       def node_distance_walk(node, distance, node_distances: {}, corporation: nil, path_distances: {})
-        return if node_distances&.[](node) && node_distances[node] <= distance
+        return if (node_distances[node] || 999) <= distance
 
         node_distances[node] = distance
-        this_distance = node.city? ? distance + 1 : distance
+        distance += 1 if node.city?
 
         node.paths.each do |node_path|
-          path_distance_walk(node_path, this_distance, path_distances: path_distances) do |path|
-            yield path, this_distance
+          path_distance_walk(node_path, distance, path_distances: path_distances) do |path|
+            yield path, distance
             path.nodes.each do |next_node|
               next if next_node == node
               next if corporation && next_node.blocks?(corporation)
@@ -288,7 +270,7 @@ module Engine
 
               node_distance_walk(
                 next_node,
-                this_distance,
+                distance,
                 node_distances: node_distances,
                 corporation: corporation,
                 path_distances: path_distances,
@@ -303,7 +285,7 @@ module Engine
       end
 
       def path_distance_walk(path, distance, skip: nil, jskip: nil, path_distances: {})
-        return if path_distances&.[](path) && path_distances[path] <= distance
+        return if (path_distances[path] || 999) <= distance
 
         path_distances[path] = distance
 
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -257,7 +257,7 @@ module Engine
       end
 
       def biggest_train(corporation)
-        corporation.trains.max_by(&:distance) || 0
+        corporation.trains.map(&:distance).max || 0
       end
 
       def get_token_cities(corporation)
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -24,13 +24,15 @@ module Engine
 
       EBUY_PRES_SWAP = false # allow presidential swaps of other corps when ebuying
       EBUY_OTHER_VALUE = false # allow ebuying other corp trains for up to face
-      HOME_TOKEN_TIMING = :operating_round
+      HOME_TOKEN_TIMING = :float
       SELL_AFTER = :any_time
       SELL_BUY_ORDER = :sell_buy
 
       EVENTS_TEXT = Base::EVENTS_TEXT.merge(
         'fishbourne_to_bank' => ['Fishbourne', 'Fishbourne Ferry Company available for purchase']
       ).freeze
+
+      TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded_or_city, upgrade: false }].freeze
 
       PAR_RANGE = {
         1 => [74, 100],
@@ -77,6 +79,7 @@ module Engine
       def stock_round
         Round::Stock.new(self, [
           Step::DiscardTrain,
+          Step::G1860::HomeTrack,
           Step::G1860::Exchange,
           Step::G1860::BuySellParShares,
         ])
@@ -86,7 +89,7 @@ module Engine
         Round::Operating.new(self, [
           Step::Bankrupt,
           Step::DiscardTrain,
-          Step::Track,
+          Step::G1860::Track,
           Step::Token,
           Step::Route,
           Step::Dividend,
@@ -122,6 +125,24 @@ module Engine
 
       def or_set_finished
         check_new_layer
+      end
+
+      def place_home_track(corporation)
+        hex = hex_by_id(corporation.coordinates)
+        tile = hex&.tile
+
+        # skip if a tile is already in home location
+        return unless tile.color == :white
+
+        @log << "#{corporation.name} must choose tile for home location"
+
+        @round.pending_tracks << {
+          entity: corporation,
+          hexes: [hex],
+        }
+
+        @round.clear_cache!
+        return
       end
 
       def event_fishbourne_to_bank!
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -59,6 +59,14 @@ module Engine
         'VYSC' => 4,
       }.freeze
 
+      NO_ROTATION_TILES = %w[
+        758
+        761
+        763
+        773
+        775
+      ].freeze
+
       def setup
         @bankrupt_corps = []
         @receivership_corps = []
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_1860.rb
```diff
@@ -223,6 +255,125 @@ def close_other_companies!(company)
         @companies.reject { |c| c == company }.each(&:close!)
         @log << '-- Event: starting private companies close --'
       end
+
+      def biggest_train(corporation)
+        corporation.trains.map(&:distance).max || 0
```
tobymao, line 260: trains.max(&:distance)

roseundy: Ok

roseundy: biggest_train needs to return an integer, not a train.

lib/engine/game/g_1860.rb
```diff
@@ -124,6 +138,24 @@ def or_set_finished
         check_new_layer
       end
 
+      def place_home_track(corporation)
+        hex = hex_by_id(corporation.coordinates)
+        tile = hex&.tile
```
tobymao, line 143: why do you do &.tile here, line 146 indicates that tile is never nil

lib/engine/game/g_1860.rb
```diff
@@ -124,6 +138,24 @@ def or_set_finished
         check_new_layer
       end
 
+      def place_home_track(corporation)
```
tobymao, line 141: why is this method in game? why not just do it in the step, seems like the only place where it is called

lib/engine/game/g_1860.rb
tobymao: this can go in the step, it's not used anywhere else it looks like

lib/engine/game/g_1860.rb
```diff
@@ -223,6 +255,125 @@ def close_other_companies!(company)
...
+            next unless city.tokened_by?(corporation)
+
+            tokens << city
+          end
+        end
+        tokens
+      end
+
+      def node_distance_walk(node, distance, node_distances: {}, corporation: nil, path_distances: {})
+        return if node_distances&.[](node) && node_distances[node] <= distance
```
tobymao, line 276: node_distancecs be nil?

if it's not you can do (node_distances[node] || 0) <= distance

lib/engine/game/g_1860.rb
```diff
@@ -223,6 +255,125 @@ def close_other_companies!(company)
...
+
+        node_distances[node] = distance
+        this_distance = node.city? ? distance + 1 : distance
```
tobymao, line 279: just call it

distance = node.city? ...

lib/engine/game/g_1860.rb
```diff
@@ -223,6 +255,125 @@ def close_other_companies!(company)
...
+          end
+        end
+      end
+
+      def lane_match?(lanes0, lanes1)
+        lanes0 && lanes1 && lanes1[0] == lanes0[0] && lanes1[1] == (lanes0[0] - lanes0[1] - 1)
+      end
+
+      def path_distance_walk(path, distance, skip: nil, jskip: nil, path_distances: {})
+        return if path_distances&.[](path) && path_distances[path] <= distance
```
tobymao, line 306: will path_distances be nil?

lib/engine/step/g_1860/track.rb
```diff
@@ -0,0 +1,123 @@
...
+
+          if node.city?
+            node_distances[node] < max_distance
+          else
+            node_distances[node] <= max_distance
+          end
+        end
+
+        def reachable_hex?(entity, hex)
+          max_distance = @game.biggest_train(entity)
```
tobymao, line 55: this pattern is repeated multiple times can probably be dryed up

lib/engine/step/g_1860/home_track.rb
```diff
@@ -0,0 +1,68 @@
...
+        def reachable_hex?(_entity, _hex)
+          true
+        end
+
+        def available_hex(_entity, hex)
+          pending_track[:hexes].include?(hex)
+        end
+
+        def hex_neighbors(entity, hex)
+          @game.graph.connected_hexes(entity)[hex]
```
tobymao, line 63: does this get hit on loading?

lib/engine/step/g_1860/track.rb
```diff
@@ -0,0 +1,123 @@
...
+              pd = path_distances[tile_path] if path_distances[tile_path]
+              return true if pd && pd <= max_distance
+            end
+          end
+
+          false
+        end
+
+        def available_hex(entity, hex)
+          return false unless (neighbors = @game.graph.connected_hexes(entity)[hex])
```
tobymao, line 86: does this get hit on loading?

lib/engine/step/g_1860/track.rb
```diff
@@ -0,0 +1,123 @@
...
+
+            return true if hex.tile.cities.any? && path_distances[npath] < max_distance
+            return true if hex.tile.cities.empty? && path_distances[npath] <= max_distance
+          end
+
+          false
+        end
+
+        def hex_neighbors(entity, hex)
+          @game.graph.connected_hexes(entity)[hex]
```
tobymao, line 118: double check this graph

lib/engine/step/g_1860/tracker.rb
```diff
@@ -0,0 +1,279 @@
...
+          @upgraded = false
+          @laid_track = 0
+          @laid_city = false
+        end
+
+        def can_lay_tile?(entity)
+          action = get_tile_lay(entity)
+          return false unless action
+
+          entity.tokens.any? && (@game.buying_power(entity) >= action[:cost]) && (action[:lay] || action[:upgrade])
```
tobymao, line 19: i don't think actions use symbols, i think they are all strings, double check this

lib/engine/step/g_1860/tracker.rb
```diff
@@ -0,0 +1,279 @@
...
+          when :restrictive
+            @game.game_error('Must use new track') unless used_new_track
+          when :semi_restrictive
+            @game.game_error('Must use new track or change city value') if !used_new_track && !changed_city
+          else
+            raise
+          end
+        end
+
+        def potential_tiles(_entity, hex)
```
tobymao, line 230: isn't this just a copy of step/tracker.rb? did you just copy and paste it?

lib/engine/step/g_1860/tracker.rb
```diff
@@ -0,0 +1,279 @@
...
+            used_new_track = true unless op
+
+            next unless old_tile.color == :white
+
+            old_revenues = op&.nodes && op.nodes.map(&:max_revenue).sort
+            new_revenues = np&.nodes && np.nodes.map(&:max_revenue).sort
+            changed_city = true unless old_revenues == new_revenues
+          end
+
+          case @game.class::TRACK_RESTRICTION
```
tobymao, line 216: is this all copy and pasted?

lib/engine/game/g_1860.rb
roseundy: I was just mirroring place_home_token. I can move it.

lib/engine/game/g_1860.rb
roseundy: Yeah, that's inconsistent. Is there ever a case where a hex doesn't have a tile?

lib/engine/game/g_1860.rb
roseundy: OK

lib/engine/game/g_1860.rb
roseundy: OK

lib/engine/game/g_1860.rb
roseundy: No, just empty.

lib/engine/step/g_1860/track.rb
roseundy: Yeah, I wasn't happy about that. Not sure what you mean by "dryed up", though.

lib/engine/step/g_1860/home_track.rb
roseundy: No. hex_neighbors is not called then. It's used by legal_tile_rotation? only.

lib/engine/step/g_1860/track.rb
roseundy: No. Same answer as hex_neighbors. This is no different from the base Tracker module.

lib/engine/step/g_1860/track.rb
roseundy: Why? This is exactly how it works in every game now.

lib/engine/step/g_1860/tracker.rb
roseundy: I didn't change this.

roseundy: This is a different "action" than the Action class. I didn't name it.

lib/engine/step/g_1860/tracker.rb
roseundy: Many of the methods are identical, but several have 1860-specific changes (distances, for example). I would have sub-classed Tracker, but it isn't a class. Can you override methods outside of a class?

lib/engine/step/g_1860/tracker.rb
roseundy: No. See above.

lib/engine/step/g_1860/tracker.rb
roseundy: Hmmm. Tracker is a module included by the step/Track class. I needed my own Track because I needed to modify some of the methods in Tracker (i.e. I need an 1860-specific Tracker).  If I sub-class my 1860 Track from the generic one, any method redefined in my Tracker will be used by the other non-modified methods in the generic Track/Tracker, right?

I should be able to simply my code then.

lib/engine/game/g_1860.rb
roseundy: (node_distances[node] || 999) <= distance
Since I don't want to return if it isn't defined yet.

lib/engine/step/g_1860/track.rb
roseundy: OK, refactored so that max_distances is only calculated once across most of these methods.

lib/engine/step/g_1860/tracker.rb
roseundy: OK, just figured out that you *can* override methods outside of a class. The 1860 Tracker will now include the base Tracker

lib/engine/step/g_1860/tracker.rb
tobymao: above where?

roseundy: The code snippet above this one. Tracker was copied over because I had to change some, but not all, of it's methods. I've since changed it so that the 1860-specific Track includes the generic one and overloads the changed methods.

roseundy: Specifically, I had to modify check Tracker::check_track_restrictions! to compare reachable revenues before and after the tile lay. Unfortunately this meant having to make a very small modification to lay_tile_action.

PR #2213, opened 2020-11-06
Allow the owner of a game to act even if they aren't playing.

daniel-sousa-me:
While doing this I took the chance to review the permission check that
got a little less strict on the other commit related to this (57860854)

I thought the previous commit dealt with all actions, but it turns out I had just peeled the first layer.

### DIFF
--- a/assets/app/app.rb
+++ b/assets/app/app.rb
@@ -58,15 +58,15 @@ class App < Snabberb::Component
       when /[^?](game|hotseat|tutorial)/
         render_game
       when /signup/
-        h(View::User, user: @user, type: :signup)
+        h(View::User, type: :signup)
       when /login/
-        h(View::User, user: @user, type: :login)
+        h(View::User, type: :login)
+      when /forgot/
+        h(View::Forgot)
+      when /reset/
+        h(View::Reset)
       when /profile/
         h(View::User, user: @user, type: :profile)
-      when /forgot/
-        h(View::Forgot, user: @user)
-      when /reset/
-        h(View::Reset, user: @user)
       when /about/
         h(View::About)
       when /tiles/
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -23,9 +23,14 @@ module View
         @game_data['user_settings'].merge!(settings)
       end
 
+      def participant?
+        return @participant if defined?(@participant)
+
+        @participant = (@game.players.map(&:id) + [@game_data['user']['id']]).include?(@user&.dig('id'))
+      end
+
       def process_action(action)
         hotseat = @game_data[:mode] == :hotseat
-        participant = @game.players.map(&:name).include?(@user&.dig('name'))
 
         if Lib::Params['action']
           return store(:flash_opts, 'You cannot make changes in history mode. Press >| to go current')
@@ -33,8 +38,8 @@ module View
 
         if !hotseat &&
           !action.free? &&
-          participant &&
-          !@game.active_players.map(&:name).include?(@user['name'])
+          participant? &&
+          !@game.active_players.map(&:id).include?(@user['id'])
 
           unless Lib::Storage[@game.id]&.dig('master_mode')
             return store(:flash_opts, 'Not your turn. Turn on master mode in the tools tab to act for others.')
@@ -57,13 +62,13 @@ module View
 
         if hotseat
           Lib::Storage[@game_data[:id]] = @game_data
-        elsif participant
+        elsif participant?
           json = action.to_h
           if @game_data&.dig('settings', 'pin')
             meta = {
               'game_result': @game_data[:result],
               'game_status': @game_data[:status],
-              'active_players': game.active_players.map(&:name),
+              'active_players': game.active_players.map(&:id),
               'turn': game.turn,
               'round': game.round.name,
             }
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -25,12 +25,13 @@ module View
             message = event['target']['value']
             if message.strip != ''
               event['target']['value'] = ''
-              process_action(Engine::Action::Message.new(@player, message: message))
+              sender = @player || Engine::Player.new(@game_data['user']['id'], @game_data['user']['name'])
+              process_action(Engine::Action::Message.new(sender, message: message))
             end
           end
         end
 
-        if @player
+        if participant?
           children << h(:div, { style: {
             margin: '1vmin 0',
             display: 'flex',
--- a/assets/app/view/user.rb
+++ b/assets/app/view/user.rb
@@ -19,77 +19,96 @@ module View
     ROUTE_COLORS = Lib::Settings::ROUTE_COLORS.freeze
 
     def render_content
-      title, inputs =
+      children =
         case @type
-        when :signup
-          ['Signup', [
-            render_input('User Name', id: :name),
-            render_input('Email', id: :email, type: :email, attrs: { autocomplete: 'email' }),
-            render_input('Password', id: :password, type: :password, attrs: { autocomplete: 'new-password' }),
-            render_notifications,
-            h(:div, [render_button('Create Account') { submit }]),
-          ]]
-        when :login
-          ['Login', [
-            render_input('Email', id: :email, type: :email, attrs: { autocomplete: 'email' }),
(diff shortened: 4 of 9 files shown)


assets/app/app.rb
```diff
@@ -58,15 +58,23 @@ def render_content
       when /[^?](game|hotseat|tutorial)/
         render_game
       when /signup/
-        h(View::User, user: @user, type: :signup)
+        return h('h3', 'You are already logged in') if @user
+
+        h(View::User, type: :signup)
```
tobymao, line 63: why don't you just do this in the user view?

assets/app/app.rb
```diff
@@ -58,15 +58,23 @@ def render_content
...
       when /forgot/
-        h(View::Forgot, user: @user)
+        return h('h3', 'You are already logged in') if @user
+
+        h(View::Forgot)
       when /reset/
-        h(View::Reset, user: @user)
+        h(View::Reset)
+      when /profile/
+        return h('h3', 'You are not logged in') unless @user
```
tobymao, line 75: same here

assets/app/view/game/actionable.rb
```diff
@@ -26,14 +26,15 @@ def save_user_settings(settings)
       def process_action(action)
         hotseat = @game_data[:mode] == :hotseat
         participant = @game.players.map(&:name).include?(@user&.dig('name'))
+        owner = @game.owner == @user&.dig('name')
```
tobymao, line 29: maybe just set particiant = this ? is participant gonna be different from owner?

lib/engine/.#game.rb
```diff
@@ -0,0 +1 @@
+daniel@daniel-pc.3418657:1604050323
```
tobymao, line 1: what is this file

lib/engine/game/base.rb
```diff
@@ -330,8 +330,10 @@ def self.load_from_json(*jsons)
         const_set(:LAYOUT, data['layout'].to_sym)
       end
 
-      def initialize(names, id: 0, actions: [], pin: nil, strict: false, optional_rules: [])
+      def initialize(names, id: 0, owner: nil, actions: [], pin: nil, strict: false, optional_rules: [])
```
tobymao, line 333: i don't like this

models/game.rb
```diff
@@ -49,8 +49,10 @@ class Game < Base
     FROM (
       SELECT g.*
       FROM games g
-      JOIN user_games ug
-        ON g.id = ug.id
+      WHERE EXISTS (
+          SELECT * from user_games ug
+          WHERE g.id = ug.id)
```
tobymao, line 54: no, this should just be changed to a left join

assets/app/app.rb
daniel-sousa-me: Surprisingly, there is a reason. In the view there is a similar `case` distinction, but it goes like this:
```
title, inputs =
  case @type
  when :signup
    ...      
```

And that doesn't allow to `return`

As I see it, there were 3 options:
1. The one I did, moving it "outside"
2. Repeating the lhs of the assignment in the 3 options
3. Add a third `case` in the beginning just to deal with this

If you see a 4th or prefer one of the other 2, I'll change it.

lib/engine/.#game.rb
daniel-sousa-me: Emacs stuff :/
Sorry

assets/app/view/game/actionable.rb
daniel-sousa-me: Right. I did it like this in the beginning, because there's a semantic distinction that I thought was going to be relevant, but looking back, it isn't.

lib/engine/game/base.rb
daniel-sousa-me: Where would you put it? The game needs to know who is its owner somehow, so it can tell the other components

assets/app/app.rb
tobymao: what do you mean it doesn't allow return?

daniel-sousa-me: I mean
`MiniRacer::SnapshotError: TypeError: value.$respond_to? is not a function`

I can't figure out why this is happening, but it only happens when I `return` somewhere inside the `title, inputs = ...`

tobymao: are you returning an h element?

daniel-sousa-me: Yes

tobymao: hit me up on slack sometime and i can help you debug

assets/app/view/game/actionable.rb
```diff
@@ -26,14 +26,15 @@ def save_user_settings(settings)
...
+        participant ||= @game.owner == @user&.dig('name')
 
         if Lib::Params['action']
           return store(:flash_opts, 'You cannot make changes in history mode. Press >| to go current')
         end
 
         if !hotseat &&
           !action.free? &&
-          participant &&
+          (participant || owner) &&
```
tobymao, line 37: i think you forgot to revert this line

assets/app/view/game/actionable.rb
```diff
@@ -57,7 +58,7 @@ def process_action(action)
 
         if hotseat
           Lib::Storage[@game_data[:id]] = @game_data
-        elsif participant
+        elsif participant || owner
```
tobymao, line 61: and here

lib/engine/game/base.rb
tobymao: i think it's a bit strange, if we're going to do something like this i'd prefer something less invasive. 

can we remove the concept of owner as a fake player, and instead have that be inferred? 

somehow allowing you to make actions but not having to have owner/owner_player within the engine

daniel-sousa-me: The owner is not something new I'm adding, it was already there but was only being used for new games. I'm just making the game aware of it and allowing that user to act.

This is only being used in actionable and game_log (which includes actionable). In actionable we have @game and @game_data. A workaround would be to not pass the owner to @game and just use the information from @game_data.

I think any other solution would be more invasive, because we'd always need to have the game aware of it somehow.

assets/app/view/game/actionable.rb
```diff
@@ -23,24 +23,27 @@ def save_user_settings(settings)
         @game_data['user_settings'].merge!(settings)
       end
 
+      def participant?
+        @game.players.map(&:id).append(@game_data['user']['id']).include?(@user&.dig('id'))
```
tobymao, line 27: return @participant if defined(@participant)

(@game.players.map(&:id) + [@game_data['user'][']id}.include

assets/app/view/game/game_log.rb
```diff
@@ -16,6 +16,7 @@ def render
         ]
 
         @player = @game.player_by_id(@user['id']) if @user
+        sender = @player || Engine::Player.new(@game_data['user']['id'], @game_data['user']['name'])
```
tobymao, line 19: why is this created outside of the lambda?

tobymao: why does this use a name instead of an id?

assets/app/view/user.rb
```diff
@@ -19,71 +19,81 @@ class User < Form
...
-          ['Profile Settings', [
-            render_username,
-            render_notifications(setting_for(:notifications)),
-            h('div#settings__colors', [
-              render_logo_color(setting_for(:red_logo)),
-              h(:div, [
-                render_color('Main Background', :bg, color_for(:bg)),
-                render_color('Main Font Color', :font, color_for(:font)),
+      children =
+        if @user
```
tobymao, line 23: this function is too nested and big can you clean it up a bit

lib/engine/action/base.rb
```diff
@@ -11,7 +11,7 @@ class Base
       attr_accessor :id, :user
 
       def self.from_h(h, game)
-        entity = game.get(h['entity_type'], h['entity'])
+        entity = game.get(h['entity_type'], h['entity']) || Player.new(h['entity'], h['entity'])
```
tobymao, line 14: what if we make ID 0 or nil?

assets/app/view/game/game_log.rb
daniel-sousa-me: It is creating the object with both

lib/engine/action/base.rb
daniel-sousa-me: Yeah, that should happen. I wrote this before the change to id's and it went unnoticed when I rebased

PR #2232, opened 2020-11-08
refactor: drop "selected" from `optional_rules_selected`

michaeljb:
Migration script I used locally to fix game settings for this change:

```
require_relative 'validate'

where_args = {
  Sequel.pg_jsonb_op(:settings).has_key?('optional_rules_selected') => true,
}

Game.where(**where_args).each do |game|

  game.settings['optional_rules'] =
    if game.settings.has_key?('pin')
      # keep the old key on pinned games so they work correctly, but still add
      # the new key so their info is displayed on game cards on the home page
      # correctly as well
      game.settings['optional_rules_selected'] || []
    else
      game.settings.delete('optional_rules_selected') || []
    end

  game.save
end
```

### DIFF
--- a/assets/app/mail/turn.rb
+++ b/assets/app/mail/turn.rb
@@ -14,7 +14,7 @@ class Turn < Snabberb::Component
       @game_data['players'].map { |p| p['name'] },
       id: @game_data['id'],
       actions: @game_data['actions'],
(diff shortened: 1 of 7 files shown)


tobymao (approved):

tobymao:
go ahead and deploy

michaeljb:
Deployed

PR #2234, opened 2020-11-08
Fixes for NdM 5% in market pool

perwestling:
1. (18MEX specific) If at cert limit, 5% NdM share should still be possible to buy

2. (General) If buying non-10% for Market is an option, percentage is always shown on Buy button.


jenf (approved):

perwestling:
Disregard the push of 6b62faf - was another issue.

Issue #2171, opened 2020-11-02
18MEX merger round should show players, not corporations


ChrisTheCat:
In the screenshot below, the NdM just purchased the 5 train, triggering the merger. Priority order is thedaber (NdM + PAC) -> hobie (MC + UdY) -> Chris (MEX + TM) -> winn1098 (Chi). The merger order is hobie (MC or UdY) -> Chris (MEX) -> winn1098 (Chi). The game is asking in the right order, but the row (under history, above undo/decline buttons) is corporations in operating order. That row really should show the players, as I've described them here, and not the corporations.

<img width="1333" alt="Screen Shot 2020-11-02 at 11 18 21 AM" src="https://user-images.githubusercontent.com/8732435/97909511-336cbd00-1cfd-11eb-8db7-7115fbfad829.png">

perwestling:
Player order is thedaber, Hobie (controlling UdY & MC), ChrisShaffer (MEX), win1098(CHI), and SPM not floated
Merge should there for be UdY or MC, followed by MEX, followed by CHI, followed by SPM
Game log shows:
  Merge candidates: UdY,MC,MEX,CHI,SPM
In turn it shows UdY, MC, MEX, CHI as the active entity (underlined) and ask the controlling player if they want to merge or not.
If all decline SPM is automatically chosen.

The operation order is displayd as we are in operating round.

If we change the order, how would you tell what the operating order is? Is this not important for your decision?

perwestling:
Would it be better if the logged line would be:

Merge decision: Hobie (UdY,MC),ChrisShaffer (MEX), win1098 (CHI), auto/thedaber (SPM)

ChrisTheCat:
Operating order is not really relevant at all.

ChrisTheCat:
yes, that would be great.

ChrisTheCat:
Except that SPM isn't eligible, because thedaber owns NdM

perwestling:
If showing players, which player should be shown for SPM? (None I suppose as it is merged automatically)

If so the NdM player would only be shown if there is a choice among non-floated.

Would the players understand that SPM will be automatically merged?

ChrisTheCat:
Perhaps this would be best:

Merge decision: Hobie (UdY,MC), ChrisShaffer (MEX), win1098 (CHI), automatic (SPM)

ChrisTheCat:
If there were two corporations unfloated (hard to imagine that happening, but it's possible), then:

[...]

PR #2238, opened 2020-11-08
Improve log message Fix/18 mex/merge candidates log

perwestling:
Does improve the log message for showing merge candidates.
This addresses issue #2171 although it does not solve it.

This is a slight improvement that address part of issue #2171 

Here is how it looks in the logs after this change:
image

Note: If a player controls several corporations these will not be grouped.
So it is
  CHI (Player A), MEX (Player A), SPM (Player B)

'automatic' is shown on the last corporation, if it is non-floated. This is as this will be automatically merged, if all corporations before that decline.

The president of NdM appear on the non-floated corporations, except the last one. This is because the NdM president does choose if there are multiple non-floated corporations that are merge candidates. (In case there are non-IPOed corporations, any IPOed but non-floated cannot be merge candidates.)

### DIFF
--- a/lib/engine/game/g_18_mex.rb
+++ b/lib/engine/game/g_18_mex.rb
@@ -282,14 +282,14 @@ module Engine
           return merge_major
         end
 
-        @mergable_candidates = mergable_corporations
-        @log << "Merge candidates: #{present_mergable_candidates(@mergable_candidates)}" if @mergable_candidates.any?
+        @mergeable_candidates = mergeable_corporations
+        @log << "Merge candidates: #{present_mergeable_candidates(@mergeable_candidates)}" if @mergeable_candidates.any?
         possible_auto_merge
       end
 
       def decline_merge(major)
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_mex.rb
```diff
@@ -283,7 +283,7 @@ def event_ndm_merger!
         end
 
         @mergable_candidates = mergable_corporations
-        @log << "Merge candidates: #{@mergable_candidates.map(&:name)}" if @mergable_candidates.any?
+        @log << "Merge candidates: #{present_mergable_candidates(@mergable_candidates)}" if @mergable_candidates.any?
```
jenf, line 286: Needs the 'e' Mergeable

Otherwise fine

perwestling: I have pushed a commit that change all 21 occurrences of mergable to mergeable.

jenf (approved):

PR #2239, opened 2020-11-08
18MEX: Rename PAC to FCP

perwestling:
In the physical game FCP is shown as the logo. FCP is the acronym for the
spannish name (Ferrocarril del Pacífico) while PAC is acronum for the
english name (Pacific Railroad).

Now Ferrocarril is the spannish word for 'railway' and ferrocarril
appear in most of the spannish names... but the remaining corporation
will not be renamed as it would cause confusion.

The reason for this change is that most players is familliar with
the FCP logo in the physical game.

This breaks all 18MEX games. It would be possible to convert games
but I do not think it is worth the effort in case there are other
game breaking changes.


perwestling:

perwestling:
Need migration, but it should be straightforward.

PR #2240, opened 2020-11-08
18MEX: Skip minor track lay if it cannot afford to lay in home hex

### LINKED ISSUE #1944: 18MEX tile laying step should be skipped if no money and only tile has cost
This might be a more global request. In 1846, tile laying is skipped if the corporation has no money. It would be nice if the same thing happened in 18MEX when the only tile a corporation can lay has a cost the corporation cannot afford. An easy example of this is the Minor Corporations in the first operating round.

perwestling:
If minor corporation does not have enough cash to lay a yellow in
its home hex, the track lay is skipped.

In case home hex has a yellow/green tile then normal track laying
check is done. This might mean that minor gets to manually pass
as there are no allowed track lays it can afford.

But that last is an exceptional case.

Fixes #1944

### DIFF
--- a/lib/engine/step/g_18_mex/track.rb
+++ b/lib/engine/step/g_18_mex/track.rb
@@ -33,6 +33,15 @@ module Engine
           super
         end
 
+        def can_lay_tile?(entity)
+          return super unless entity.minor?
+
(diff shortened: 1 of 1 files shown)


perwestling:

PR #2241, opened 2020-11-08
1817: Improve display of loan info, add buying power

### LINKED ISSUE #2186: 1817 Liquidation & debt visuals
Currently it’s not very clear if a company is in liquidation (apart from in the market tab).

Could it be possible to highlight a company in liquidation? May change the colour graphics to red or something on the game tab? Especially if a player is about to pay out when they could withhold and stop the liquidation. Saves big rewinds. I know it’s an unforgiving game but obvious blunders like that can be avoided. 

Also I’d suggest making the interest red if the company doesn’t have the funds so it’s more apparent.

On rr18xx it also includes a value called Debt margin, which shows how much the company would have if you took max loans less interest for all loans. Which gives you an idea of how close to the line you’re driving.

debt margin = company cash + 100*(loans remaining) - (size * loan fee)

### LINKED ISSUE #2192: 1817 Change place of loan interest info, remove bank size info
While it is appropriate to have the total number of loans issued, the actual and the future interest rate displayed in the bank card in the entities tabs, it also makes it much difficult for players playing from mobile to find them, since they are forced to scroll through many other cards.
I think that it would make it easier to move or also display these informations in the info tab, which is much less hoarded and clearer, and would it make it much easier to parse these information.

The bank size info should be removed or just hidden, since the game has a infinite bank, so it's actually incorrect to show a very big bank size.

jenf:
Change stock market tab to use the card form (fixes #2192)
Add detail showing how loans go up and down (also #2192)
Add colour for acquisition and liquidation (fixes #2186)
Highlight red when corp cannot pay its loans (also #2186)
Add buying power for games where loans are present (also #2186)

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -60,8 +60,8 @@ module View
         children << render_abilities(abilities_to_display) if abilities_to_display.any?
 
         extras = []
-        extras += render_loans if @corporation.loans.any?
-        extras += render_buying_power if @game.total_loans.positive?
+        extras << render_loans if @corporation.loans.any?
+        extras << render_buying_power if @game.total_loans.positive?
         if extras.any?
           props = { style: { borderCollapse: 'collapse' } }
           children << h('table.center', props, [h(:tbody, extras)])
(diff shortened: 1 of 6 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -58,7 +58,14 @@ def render
           ability.owner.corporation? && ability.description
         end
         children << render_abilities(abilities_to_display) if abilities_to_display.any?
-        children << render_loans if @corporation.loans.any?
+
+        extras = []
+        extras += render_loans if @corporation.loans.any?
```
tobymao, line 63: extras <<

PR #2246, opened 2020-11-08
1828: Implement Virginia Coalfields

crericha:
VA Coalfields specific rules
- Can buy coal marker if connected to VA coalfields
- Acquire a coal marker when laying VA tunnel (straight town only legal lay)
- Cannot run to VA coalfields without a coal marker
- Cannot see past VA coalfields without a coal marker (acts like its tokened out)
- Cannot token VA coalfields

Coal marker implemented as an ability


lib/engine/part/city.rb
```diff
@@ -6,14 +6,15 @@
 module Engine
   module Part
     class City < RevenueCenter
-      attr_accessor :reservations
+      attr_accessor :reservations, :block_if
```
tobymao, line 9: what is this for?

assets/app/view/game/buy_special.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+
+module View
```
tobymao, line 5: so what is this used for?

lib/engine/part/city.rb
crericha: Allow the game to add any conditions, beyond the token, that make a city blocked. For 1828, the VA coalfields tile is blocked unless the corporation has bought a coal marker.

assets/app/view/game/buy_special.rb
crericha: To purchase a coal marker. I made it general, so that other games that have game specific purchases, can reuse it. 1856 and the tunnel/bridge access, OE with port access, pullmans, and more.

lib/engine/part/city.rb
tobymao: ok, let's not do this. since it's a one off thing for for 1828.

instead, just overwrite the whole function on g_1828#244

lib/engine/step/g_1828/buy_special.rb
```diff
@@ -0,0 +1,27 @@
...
+        def can_buy_special?(entity)
+          @game.can_buy_coal_marker?(entity)
+        end
+
+        def short_description
+          'Coal Marker'
+        end
+
+        def items
+          [{ description: 'Coal Marker', cost: 120 }]
```
tobymao, line 18: should we make a real item class?

tobymao: or should we instead use assign?

lib/engine/part/city.rb
crericha: What technique do you suggest? Overwrite the method dynamically in the instance? I ask because before I did the block_if approach, I went down the inheritance path. I had to overwrite city.rb, tile.rb, and init_tiles in the game base class. Seemed like a lot and the approach above felt like a more elegant way to inject behavior from the game.

tobymao: I think you should be able to dynamically reassign the block? method of this one city just like you assign block_if

tobymao: `irb(main):001:1* class X
irb(main):002:2*   def y
irb(main):003:2*     puts 'hi'
irb(main):004:1*   end
irb(main):005:0> end
=> :y
irb(main):006:0> x = X.new
irb(main):007:0> x.y
hi
=> nil
irb(main):008:1* def x.y
irb(main):009:1*   puts 'z'
irb(main):010:0> end
=> :y
irb(main):011:0> x.y
z

`

lib/engine/step/g_1828/buy_special.rb
crericha: Sure, I can make it a class.

I looked at assign. If I understand it correctly, the player clicks on the object to assign and then on what to assign it to. The coal marker doesn't really have the first half of that, so in my opinion, a button makes sense.

lib/engine/part/city.rb
crericha: Okay, sounds good. I'm familiar with that approach and surprised Opal can handle. Cool.

PR #2278, opened 2020-11-12
1856 initial commit

ventusignis:
This has a (nearly?) complete 1856 config JSON, barebones game class, fixes for game tiles, and placeholder corporation token SVGs.

1856 has a weird behavior where the rules for how a company starts / is capitalized depends not on the current phase, but on the *facing train*
To get around this I made "prime phases" / "second phases" for the last train of each set of trains. This will make coding up the rules for the game easy enough but it significantly clutters up the UI for users - this is almost certainly a solution where the end-user loses.

The game is not in a state where it can use it yet but if someone has an implementation of "getFacingTrain()" I can remove the "second phases" and use that instead

This also uses placeholders for tokens for all the companies, these need to be replaced at some point. If there is a way to get SVGs usable for 18xx.games out of 18xx-maker I can update the logos.


tobymao (approved):

daniel-sousa-me:
@ventusignis I'm doing 1870. We probably should coordinate

daniel-sousa-me:
Well, I just found the rules summary for 1856 and it's waaaayy more different then I expected. I think the only new thing they have in common are the destination routes (and even those seem to work a little differently). So let's say I do that, since you already have a ridiculous amount of things to implement?

ventusignis:
> Well, I just found the rules summary for 1856 and it's waaaayy more different then I expected. I think the only new thing they have in common are the destination routes (and even those seem to work a little differently). So let's say I do that, since you already have a ridiculous amount of things to implement?

If permissions to implement 18C2C are given, that's a lot closer to 1870, even if enormous 

If you implement a function that given two hexes (or cities later?) and a boolean flag to determine if stations are blocking or not that will return if they are connected by track that can be used by both 1870 and 1856

ventusignis:
> > Well, I just found the rules summary for 1856 and it's waaaayy more different then I expected. I think the only new thing they have in common are the destination routes (and even those seem to work a little differently). So let's say I do that, since you already have a ridiculous amount of things to implement?
> 
> If permissions to implement 18C2C are given, that's a lot closer to 1870, even if enormous
> 
> If you implement a function that given two hexes (or cities later?) and a boolean flag to determine if stations are blocking or not that will return if they are connected by track that can be used by both 1870 and 1856

It needs to be city; the THB in 1856 starts in a OO and since it has two cities, the the city with the THB home token is the one that matters

PR #2306, opened 2020-11-14
1860 trains, routes and route selection

roseundy:
This is going to be one of the biggest set of changes for 1860.
This implements:
- 1860 m+n trains
- All of the various and sundry (and non-standard) route rules
- Allowing one train to run through a tokened-out city
- Automatically finding best towns to run to
- UI change to allow selecting between halt subsidies and town revenues
- Support for save/restore of connections to tiles with multiple intra-tile paths

One of the biggest problems I faced was not only allowing the user to choose between halts and towns for each train, but preserving that choice for the next OR. This required modifying the run_routes action as well as operating_info.

New UI for selecting number of halts in route_selector:
route_selector

The following common files had to be changed:
- Moved game-specific logic from Engine::Route to Game::Base
- Engine::Connection::id to support "ambiguous" connections by appending node ids to hex name
- Engine::Tile and Engine::Part::Node to support linear naming of node parts (node_id)
- Engine::Route, Action::RunRoutes, View::RouteSelector => connection_hexes remain strings until use
- Engine::Route, Action::RunRoutes support for halts
- View::RouteSelector suport for halts
- Engine::OperatingInfo support for halts


roseundy:
This is what the new JSON for run_routes can look like for a route with halts and multiple intra-tile paths:
json

assets/app/view/game/route_selector.rb
```diff
@@ -149,16 +158,29 @@ def render
             ]),
             h(:tbody, trains),
           ]),
-          actions,
+          actions(render_halts),
         ])
       end
 
+      def halt_actions(route, revenue, subsidy)
```
tobymao, line 165: would it be easier to just pay out max halts instead of having this action?

tobymao:
```One of the biggest problems I faced was not only allowing the user to choose between halts and towns for each train, but preserving that choice for the next OR. This required modifying the run_routes action as well as operating_info.```

maybe we should just simplify it by not alloiwg it

File: lib/engine/connection.rb [hunk 2]
```diff
@@ -81,17 +81,23 @@ def matches?(connection_hexes)
...
-          junction_map = {}
-
-          # skip over paths that have a junction we've already seen
-          @paths.each do |path|
-            uniq_paths << path if !junction_map[path.a] && !junction_map[path.b]
-            junction_map[path.a] = true if path.a.junction?
-            junction_map[path.b] = true if path.b.junction?
-          end
+          # deal with ambiguous intra-tile path
+          if @paths&.one? && @paths[0].tile.ambiguous_connection?
```
tobymao, line 85: why would paths be nil?

lib/engine/connection.rb
```diff
@@ -81,17 +81,23 @@ def matches?(connection_hexes)
...
+            node_ids = @paths[0].nodes.map(&:node_id).sort
```
tobymao, line 86: youc ould do 

first, last @paths[0].nodes.map9&:nodeid).sort

lib/engine/game/base.rb
```diff
@@ -775,6 +779,105 @@ def check_overlap(routes)
         end
       end
 
+      def check_connected(route, token)
```
tobymao, line 782: did you change anything here that i need to look at or is it just movement?

lib/engine/tile.rb
```diff
@@ -239,6 +241,12 @@ def terrain
       @upgrades.flat_map(&:terrains).uniq
     end
 
+    # if tile has more than one intra-tile paths, those paths
+    # cannot be identified with just a hex name
+    def ambiguous_connection?
+      @ambiguous_connection ||= @paths.select { |p| p.nodes.size > 1 }.size > 1
```
tobymao, line 247: could you do @paths.count { } ?

assets/app/view/game/route_selector.rb
roseundy: I feel it's a pretty important tactical decision about running for revenue or to help the company. The rules specifically call this out as a choice the director makes.

lib/engine/connection.rb
roseundy: In initialize: @path = nil

lib/engine/connection.rb
roseundy: OK

lib/engine/game/base.rb
roseundy: Just moving game logic out of route

lib/engine/tile.rb
roseundy: Sure. Always learning something new...

lib/engine/connection.rb
tobymao: then how come the line below doesn't fail?

roseundy: Good point.

assets/app/view/game/route_selector.rb
tobymao: hm i thought the choice was whether or not to use the halt or not. is there another choice? i don't think halts pay out, they only ever help the company. the choice would be to use the halt or not, you'd only not want to use it in very rare situations

roseundy: In a real life game with a M+N train, you decide how the +N stops are split between towns and halts. Since you can choose between upgrading to another halt or to a town, presumably you'd want the choice to get the revenue from the town if they were available.

roseundy: How strongly do you dislike how I have it now?

tobymao: i'll look at it some more this weekend

assets/app/view/game/route_selector.rb
```diff
@@ -56,6 +55,7 @@ def render
           store(:selected_route, route, skip: true)
         end
 
+        render_halts = false
```
tobymao, line 58: render_halts = route.num_halts

assets/app/view/game/route_selector.rb
```diff
@@ -95,7 +95,13 @@ def render
             td_props = { style: { paddingRight: '0.8rem' } }
 
             children << h('td.right', td_props, route.distance)
-            children << h('td.right', td_props, revenue)
+            if route.num_halts
+              render_halts = true
```
tobymao, line 99: remove this

assets/app/view/game/route_selector.rb
```diff
@@ -149,16 +158,29 @@ def render
...
 
+      def halt_actions(route, revenue, subsidy)
+        change_halts = lambda do
+          route.cycle_halts
+          store(:selected_route, route)
+        end
+
+        [
+          revenue,
+          h(:br),
```
tobymao, line 173: i don't think we need a br, just use a style

lib/engine/config/game/g_1860.rb
```diff
@@ -576,35 +576,35 @@ module G1860
   "trains": [
     {
       "name": "2+1",
-      "distance": 2,
+      "distance": [2, 1],
```
tobymao, line 579: why is distance an array?

look at the format of distance in g_1846

tobymao: how can your code work with distance in this format?

lib/engine/game/g_1860.rb
```diff
@@ -239,7 +241,7 @@ def close_other_companies!(company)
       end
 
       def biggest_train(corporation)
-        corporation.trains.map(&:distance).max || 0
+        corporation.trains.map { |t| t.distance.first }.max || 0
```
tobymao, line 244: there is no such thing as distance first, the format doesn't look correct

lib/engine/route.rb
```diff
@@ -4,9 +4,10 @@
 
 module Engine
   class Route
+    attr_accessor :num_halts
```
tobymao, line 7: why is this an accesor?

lib/engine/route.rb
```diff
@@ -4,9 +4,10 @@
...
     attr_reader :last_node, :phase, :train, :routes
 
-    def initialize(game, phase, train, connection_hexes: [], routes: [], override: nil)
+    def initialize(game, phase, train, connection_hexes: [], routes: [], override: nil, num_halts: nil)
```
tobymao, line 10: should we just call this halts?

lib/engine/game/g_1860.rb
```diff
@@ -356,6 +359,175 @@ def legal_tile_rotation?(_entity, _hex, tile)
...
+        n_towns = route.stops.select { |n| n.town? && !n.halt? }.size
+        "#{n_cities}+#{n_towns}"
+      end
+
+      def revenue_for(route, stops)
+        stops.sum { |stop| stop.route_base_revenue(route.phase, route.train) }
+      end
+
+      def subsidy_for(_route, stops)
+        stops.select(&:halt?).size * HALT_SUBSIDY
```
tobymao, line 521: stops.count(&:halt?)

tobymao: but how does this work with num_halts in route?

lib/engine/route.rb
```diff
@@ -270,8 +255,8 @@ def find_connections(connections_a, connections_b, other_paths)
     private
 
     def restore_connections(connection_hexes)
-      possibilities = connection_hexes.map do |hexes|
-        hex_ids = hexes.map(&:id)
+      possibilities = connection_hexes.map do |hex_ids|
+        hexes = hex_ids.map { |hex_str| @game.hex_by_id(hex_str.split.first) }
```
tobymao, line 259: why did you change this to get the hexes here every time instead of on init?

lib/engine/tile.rb
```diff
@@ -90,7 +90,8 @@ def self.part(type, params, cache)
                               visit_cost: params['visit_cost'],
                               route: params['route'],
                               format: params['format'],
-                              loc: params['loc'])
+                              loc: params['loc'],
+                              node_id: cache.size)
```
tobymao, line 94: why do you need to add a node_id?

assets/app/view/game/route_selector.rb
roseundy: Ok

roseundy: Ok

assets/app/view/game/route_selector.rb
roseundy: Ok

assets/app/view/game/route_selector.rb
roseundy: Ok, I'll have to figure out how I can use a style to place the button under the revenue.

lib/engine/config/game/g_1860.rb
roseundy: M+N train => distance = [M, N]
Since I was rewriting the code dealing with distances, I chose to use a simple array over the more complicated hash used by 1846 and the others. In addition, the existing code and the hash structure couldn't handle how 1860 train distances work - specifically how an unused city/offboard allowance can be used for towns/halts (i.e. a run than doesn't use all M cities).

lib/engine/game/g_1860.rb
roseundy: distance is an array here.

lib/engine/route.rb
roseundy: Because on a brand new route that contains halts, num_halts is calculated and set by @game.compute_stops. Effectively the default num_halts for a new route is set to a number that maximizes revenue first, subsidies second.

lib/engine/route.rb
roseundy: Yeah, probably.

lib/engine/game/g_1860.rb
roseundy: route.num_halts is used by @game.compute_halts to determine how many stops are at halts.

lib/engine/route.rb
roseundy: restore_connection is only called on init, so I'm not sure what you mean by "every time"
I moved the @game.hex_by_id lookup here from Action::run_routes. It shouldn't be called any more often.

I moved it here because I needed to leave connections_hexes in string format so that hex_ids was the same string value stored in the JSON so that the Connection::id generated for "ambiguous" intra-tile paths would match correctly.

lib/engine/tile.rb
roseundy: Because I needed an id that was unique across all forms of nodes for use in disambiguating nodes in intra-tile paths/connections. Tile generates indexes per class, but those indexes aren't unique across the node classes (i.e. there can be a City with index 0 as well as a Town with index 0).

lib/engine/route.rb
roseundy: This is how num_halts ever becomes anything other than nil for a new route.

lib/engine/tile.rb
roseundy: I also needed node_id to be completely deterministic and based on the tile definition so that upgraded tiles will work correctly with existing routes/connections.

assets/app/view/game/route_selector.rb
roseundy: Wait, render_halts is calculated across *all* routes. If any of the routes has halts, then render_halts is set true.

lib/engine/route.rb
roseundy: I renamed num_halts to halt

assets/app/view/game/route_selector.rb
roseundy: Changed to use div, which avoids the need for br

lib/engine/config/game/g_1860.rb
tobymao: i don't think that's true. the current hash used in 1846 should allow for city and offboard overflowing to towns and halts.

lib/engine/game/g_1860.rb
tobymao: doesn't that mean that this method should check num_halts and not select halts? don't you need to know which ones the user chose?

lib/engine/tile.rb
tobymao: let's get rid of node_id and just change the part.index = index to not group by class, would that work?

lib/engine/config/game/g_1860.rb
roseundy: OK, I'll look at the code again. I wasn't seeing how that could work.

lib/engine/game/g_1860.rb
roseundy: All halts are the same, I just need to know how many halts are being stopped at. I could use either Route::halts or count the stops that are halts. They should always be the same.

lib/engine/tile.rb
roseundy: Will the be guaranteed to result in the same index for a node after an upgrade? Since node_id is a function of the tile definition, as long as the node appears in the same order in both the original tile definition and it's upgraded tile, it will work correctly.

roseundy: It will work as long as all city and town definitions occur before any paths are defined. I can live with that.

roseundy: Done.

lib/engine/config/game/g_1860.rb
roseundy: Done.

lib/engine/game/g_1860.rb
roseundy: Moved to standard train definitions.

lib/engine/connection.rb
```diff
@@ -81,17 +81,23 @@ def matches?(connection_hexes)
...
-
-          # skip over paths that have a junction we've already seen
-          @paths.each do |path|
-            uniq_paths << path if !junction_map[path.a] && !junction_map[path.b]
-            junction_map[path.a] = true if path.a.junction?
-            junction_map[path.b] = true if path.b.junction?
-          end
+          # deal with ambiguous intra-tile path
+          if @paths.one? && @paths[0].tile.ambiguous_connection?
+            node0, node1 = @paths[0].nodes.map(&:index).sort
```
tobymao, line 86: should this be id and not index?

roseundy: No, because id includes the tile number which would break the JSON stored route on upgrades.

tobymao (approved):

PR #2307, opened 2020-11-14
Allow using abilities with "when: owning_corp_or_turn" during BuyCompany step

### LINKED ISSUE #2032: [1846] buying in a private shouldn't force end turn
1846
When buying in the meat packing plant at the end of my company's turn, it instantly ended that company's turn.  So I sat there clicking and trying to get the token placed for that private, but didn't realize it had moved on.  The brown phase was started shortly thereafter and I was left with the option to beg for my company's turn back to place the token.

The company turns should never auto-end, especially on something as innocuous as buying a private.  There are a number of other things I could still have done, all taken away with not a whit of notification.

michaeljb:
- change 1846's BuyCompany description to "Buy/Use Companies"
- add new ability "when" option; value of "owning_corp_or_turn" means the
  ability can be used any time during the turn; if "track" or
  "owning_corp_or_turn" are passed as the time to an ability with
  "owning_corp_or_turn", then the ability is usable
- add logic to block in BuyCompany if any abilities with "owning_corp_or_turn"
  are active and available
- refactor `Abilities#abilities`, including extracting a bunch of private methods
    - allow a call with no type passed in, to collect all abilities that match
      the other given parameters
    - change how checking the time works
        - add new parameter `require_when`; if false (which is default), then
          this works as before, i.e., if no time is specified, then the ability
          is active even if it has a `when` attr; if true, then only abilities
          with a `when` are active
    - add new parameter `owner_type`; filter abilities by owner type so that
      abilities on player-owned companies don't cause the BuyCompany step to
      block
- add broken 18GA game as fixture
- "migrate" all broken fixtures

[Fixes #2032] [Relates to ##1704]


michaeljb:
The commits on this branch are pretty well organized right now, the interesting logic changes are all in https://github.com/tobymao/18xx/pull/2307/commits/280876103e18edf850f830c371253416e3ff3111

michaeljb:
All of the broken 1846-derived games should work by just adding "pass" actions. The one broken 18GA game actually needs a pass action removed.

michaeljb:
Broken games:

```
18 Los Angeles - 161/298
1846 - 515/1175
1846 2p Variant - 107/181
18GA - 1/75

```

this script seems to have worked for all but 55 of those broken games - https://github.com/michaeljb/18xx.games/commit/4862388df27cd67a1af4b0346643e4866d5e8434

michaeljb:
I think I've got everything looking good now, going to run another full migration+validation locally to be sure. Probably won't have more time today to work on this after I kick that off.

I've done the migration locally with a modified `migrate_game.rb` (see 
79d316f  and https://github.com/michaeljb/18xx.games/compare/fix-2032...michaeljb:migrate-2032), and it takes over half an hour on my machine to repair all the games but that script doesn't take advantage of multiple cores or anything.

michaeljb:
> I think I've got everything looking good now, going to run another full migration+validation locally to be sure. Probably won't have more time today to work on this after I kick that off.

@tobymao the migration worked and pinned 0 games, but the validation hasn't quite finished (I'm expecting 0 issues since the migration was applied to all games that threw exceptions on the prior validation). I should be able to do the release for this tomorrow, or possibly very late tonight if review gets done before then, but I'm happy to wait.

michaeljb:
The validation I kicked off finished as expected with no errors, converting from "draft" to "ready for review".

lib/engine/abilities.rb
```diff
@@ -90,5 +76,64 @@ def ability_uses
       end
       count
     end
+
+    private
+
+    def right_type?(ability, type)
+      type ? ability.type == type : true
```
tobymao, line 83: !type || ability.type == type

tobymao (approved):

PR #2308, opened 2020-11-14
18MEX: Don't try to spend $0 when the home hex has not tile (fixes $2304)

daniel-sousa-me:


perwestling:
@daniel-sousa-me I did work on this as well, and have pushed a fix.

I think my solution was slightly better as I gives different log messages. In your case it will no log any message about the missing mail. Also, my commit explains WHY this happened.

daniel-sousa-me:
And on top of everything, your commit doesn't fail the checks xD

PR #2313, opened 2020-11-14
1870: Ledge, share protection, close corp and half dividends

daniel-sousa-me:


lib/engine/action/protect.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Action
+    class Protect < Base
```
jenf, line 7: Would prefer this to be move explicit, what are you protecting? 'PriceProtect'

daniel-sousa-me: Good point. It was also a bit inconsistent, because I did that in some places

assets/app/view/game/round/stock.rb
```diff
@@ -27,6 +27,9 @@ def render
           @auctioning_corporation = @step.auctioning_corporation if @step.respond_to?(:auctioning_corporation)
           @selected_corporation ||= @auctioning_corporation
 
+          @price_protection = @step.protecting_bundle if @step.respond_to?('protecting_bundle')
```
tobymao, line 30: use a symbol for respond_to

lib/engine/action/price_protect.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
```
tobymao, line 5: why does this need a new action? can't it just be a buy share?

lib/engine/config/game/g_1846.rb
```diff
@@ -20,9 +20,21 @@ module G1846
     "5": 9000
   },
   "certLimit": {
-    "3": 14,
-    "4": 12,
-    "5": 11
+    "3": {
```
tobymao, line 23: are these changes part of a different commit?

daniel-sousa-me: Yeah, it's part of a commit that was already merged. I'll rebase and I think this should disappear.

lib/engine/action/price_protect.rb
daniel-sousa-me: Indeed it doesn't. It didn't occur to me I could use the "same action" on a different step

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -21,6 +21,8 @@ def render
         children = []
 
         if step.current_actions.include?('buy_shares')
+          price_protection = step.protecting_bundle if step.respond_to?(:protecting_bundle)
```
tobymao, line 24: should the method be called price_protection as well?

lib/engine/game/g_1846.rb
```diff
@@ -311,9 +311,7 @@ def action_processed(action)
 
         check_special_tile_lay(action)
 
-        @corporations.dup.each do |corporation|
-          close_corporation(corporation) if corporation.share_price&.price&.zero?
-        end
+        super
```
tobymao, line 314: why is this calling super

tobymao: shouldn't this call @corporations.dup close corporation?

tobymao: i see, you changed action_processed to close corporations... that should not be done, it's going to happen on every single action

lib/engine/step/g_1870/price_protection.rb
```diff
@@ -0,0 +1,110 @@
...
+        def round_state
+          { sell_queue: [] }
+        end
+
+        def purchasable_companies(_entity = nil)
+          []
+        end
+
+        def protecting_bundle
+          @round.sell_queue.dig(0)
```
tobymao, line 32: why call dig?

lib/engine/step/g_1870/price_protection.rb
```diff
@@ -0,0 +1,110 @@
...
+        end
+
+        def can_sell?
+          false
+        end
+
+        def can_buy?(entity, bundle)
+          return unless bundle
```
tobymao, line 40: return unless bundle&.buyable

lib/engine/game/g_1846.rb
daniel-sousa-me: Why? Doesn't it get called exactly in the cases? The point was to generalise this for all games that have a close corp space on the stock market.
Maybe I should put a condition surrounding the `each`?

lib/engine/step/g_1870/price_protection.rb
daniel-sousa-me: This was copy-pasted from `buy_sell_par_shares.rb`. I'm sneaking in a change in there as well

assets/app/view/game/buy_sell_shares.rb
daniel-sousa-me: Is there a more concise way of writing this in the spirit of `step&.protecting_bundle`?

lib/engine/game/g_1846.rb
tobymao: because after every single action, you're going to duplication corporations and iterate over them

lib/engine/step/g_1870/price_protection.rb
tobymao: you can still clean it up

tobymao: what did you change?

lib/engine/game/g_1846.rb
daniel-sousa-me: But that's not something new I did. I just moved the code from 1846 to the main method.

What do you propose I do?

lib/engine/step/g_1870/price_protection.rb
daniel-sousa-me: The code you suggested I edited was copy-pasted from another step. I changed it in my new step and in the original step.

lib/engine/game/g_1846.rb
tobymao: so are you saying this code doesn't run in 1889?

daniel-sousa-me: I was not understanding if you wanted to change how this is done altogether or just check if we're in a game where a corp can close.

Now it doesn't run in 1889.

assets/app/view/game/part/divisions.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+require 'lib/color'
```
tobymao, line 3: what is this? is this new code? i don't think i've reviewed this yet. 

if it's new, i can't really review this if it's a moving target

daniel-sousa-me: I'm so sorry. This is part of the code I'm working on now and for some reason it escaped `git stash`

daniel-sousa-me: I'll stop using "amend" and force push. I don't seem to find a satisfactory way to send the revisions.

PR #2318, opened 2020-11-15
1817: Make corp logos bigger on 5 & 10 share corps

### LINKED ISSUE #2088: 2/5/10 share graphic presentation improvement suggestions
Screen Shot 2020-10-28 at 8 26 53 AM

Would it be possible to indicate on the market which corporations are 2, 5, and 10 share? This could be by making the icons different sizes, or putting colored borders around the icons.

Also, during an OR or M&A round, it would be nice to indicate on the market which corporations have acted, or are ineligible (e.g. if a company moves into the acquisition zone during the round, it won't be eligible for acquisition that same round).

Screen Shot 2020-10-28 at 8 30 11 AM

Similarly, it would be nice to know which corporations are 2, 5, and 10 share size during the M&A round (shown above).

jenf:
Also add a field to say how big the corporation is
fixes #2088

### DIFF
--- a/lib/engine/game/g_1817.rb
+++ b/lib/engine/game/g_1817.rb
@@ -459,6 +459,15 @@ module Engine
         @operating_rounds if ['Operating', 'Merger', 'Merger and Conversion', 'Acquisition'].include?(name)
       end
 
+      def corporation_size(entity)
+        # For display purposes is a corporation small, medium or large
+        { 2 => :small, 5 => :medium, 10 => :large }[entity.total_shares]
+      end
+
+      def show_corporation_size?(_entity)
+        true
+      end
+
       private
 
       def new_auction_round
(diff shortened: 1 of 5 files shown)


michaeljb (approved):

lib/engine/game/g_1817.rb
```diff
@@ -459,6 +459,15 @@ def total_rounds(name)
         @operating_rounds if ['Operating', 'Merger', 'Merger and Conversion', 'Acquisition'].include?(name)
       end
 
+      def corporation_size(entity)
+        # For display purposes is a corporation small, medium or large
+        { 2 => :small, 5 => :medium, 10 => :large }[entity.total_shares]
```
michaeljb, line 464: Should this hash be a constant on the class so it only gets allocated once per game?

PR #2320, opened 2020-11-15
Do not allow president share to be bought from the market (fixes #2316)

### LINKED ISSUE #2316: After recent GUI changes in 18MEX: Can buy 20% share from market
https://www.18xx.games/game/16134?action=374

When a player bought a share of B&O in the next SR, two buttons were offered for the next share purchase:  [Buy 10% share from market] and [Buy 20% share from market].  I'm not entirely sure why there are two buttons, as they did the same thing and you can't buy a pres share directly from the market.

image

perwestling:


perwestling:
It seems this is correct. There are titles where you are allowed to buy the president's share in case it is the last one. Altough it is not a buy as such but a swap.

PR #2323, opened 2020-11-15
18TN: Civil War should affect triggering company (fixes #1873)

### LINKED ISSUE #1873: 18TN - Civil War did not impact triggering company
A company (GMO) with no train bought the 3' train as it's first.  It had a valid route (and train, via the purchase) when the Civil War was triggered, but it was able to run for revenue in the next OR.

Civil War Trigger captured here:
https://www.18xx.games/game/11452?action=185

perwestling:
There was a bug if the triggering company's only train was the
one triggering the Civil War. In which case trains was still
empty when checking for Civil War token.

Have now included the trigger in the event even if it has no
other trains.

Also added a log line per corporation describing if it gets
Civil War token or not, to make it more visible.

### DIFF
--- a/lib/engine/game/g_18_tn.rb
+++ b/lib/engine/game/g_18_tn.rb
@@ -103,15 +103,26 @@ module Engine
 
       def event_civil_war!
         @log << '-- Event: Civil War! --'
-        @corporations.each do |c|
-          # No effect if corporation has no trains
-          next if c.trains.empty?
+        @corporations.sort_by(&:name).each do |c|
+          unless c.floated?
+            @log << "#{c.name} does not revceive any Civil War token as it has not floated yet"
+            next
+          end
+
+          if c.trains.empty? && current_entity != c
+            # No effect if corporation has no trains, current entity does not yet have
+            # any trains as it is in the middle of a train purchase (which triggered the event)
+            # but as it will have a train after the buy is completed it gets the token anyway.
+            @log << "#{c.name} does not revceive any Civil War token as it owns no trains" if c.floated?
+            next
+          end
 
           c.add_ability(Engine::Ability::Base.new(
             type: :civil_war,
             description: 'Civil War! (One time effect)',
             count: 1,
           ))
+          @log << "#{c.name} receives a Civil War token which affects its next OR"
         end
       end
 
--- a/lib/engine/game/g_18_tn.rb
+++ b/lib/engine/game/g_18_tn.rb
@@ -103,27 +103,25 @@ module Engine
 
       def event_civil_war!
         @log << '-- Event: Civil War! --'
-        @corporations.sort_by(&:name).each do |c|
-          unless c.floated?
-            @log << "#{c.name} does not receive any Civil War token as it has not floated yet"
-            next
-          end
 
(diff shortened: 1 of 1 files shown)


perwestling:
Here is how the logs look like after this change:

image

lib/engine/game/g_18_tn.rb
```diff
@@ -103,15 +103,26 @@ def init_share_pool
 
       def event_civil_war!
         @log << '-- Event: Civil War! --'
-        @corporations.each do |c|
-          # No effect if corporation has no trains
-          next if c.trains.empty?
+        @corporations.sort_by(&:name).each do |c|
+          unless c.floated?
+            @log << "#{c.name} does not revceive any Civil War token as it has not floated yet"
```
jenf, line 108: typo receive

lib/engine/game/g_18_tn.rb
```diff
@@ -103,15 +103,26 @@ def init_share_pool
...
+            next
+          end
+
+          if c.trains.empty? && current_entity != c
+            # No effect if corporation has no trains, current entity does not yet have
+            # any trains as it is in the middle of a train purchase (which triggered the event)
+            # but as it will have a train after the buy is completed it gets the token anyway.
+            @log << "#{c.name} does not revceive any Civil War token as it owns no trains" if c.floated?
```
jenf, line 116: as above

lib/engine/game/g_18_tn.rb
perwestling: Fixed

lib/engine/game/g_18_tn.rb
perwestling: Fixed

michaeljb (commented):
I think logging all of the corporations that _don't_ receive a token makes the log cluttered.

Here's a version that gets the log down to one line:

```
      def event_civil_war!
        @log << '-- Event: Civil War! --'
        corps = @corporations.select do |c|
          (c == current_entity) || (c.floated? && c.trains.any?)
        end

        corps.each do |corp|
          corp.add_ability(Engine::Ability::Base.new(
            type: :civil_war,
            description: 'Civil War! (One time effect)',
            count: 1,
          ))
        end

        @log << "#{corps.map(&:name).sort.join(', ')} each receive a Civil "\
                'War token which affects their next OR'
      end
```

Screenshot from 2020-11-15 12-53-06

perwestling:
@michaeljb I have now incorporated your suggestion and pushed up a new version.
Added a bit of comments.

michaeljb:
I found that this breaks some games, which makes sense since the Civil War impacts revenue. I think that also means they'll need to be pinned and can't be migrated - https://gist.github.com/michaeljb/bff40b79330034abab37ffe7e28f1daf

michaeljb (approved):

perwestling:
Yes, it is likely that the situation is probably pretty common that a corporation buys its first train, and it triggers the Civil War event. The bug was that they were now not affected, which meant that the corporation got an extra train income next OR. And this should have some effects on the rest of the game. 

So pin is fine.

tobymao (approved):

PR #2330, opened 2020-11-15
Activate players using id

jenf:
This activates changing players to use id rather than name as their actions, and provides the rewrite script.

This enables username changes #1135

There's a hook to detect if the hotseat game is from before the change which should avoid us having to delete every hotseat game.

DO NOT MERGE UNTIL WE ARE READY TO MIGRATE ALL GAMES!


michaeljb (approved):

migrate_names.rb
```diff
@@ -0,0 +1,75 @@
...
+# rubocop:disable all
+
+require_relative 'models'
+
+Dir['./models/**/*.rb'].sort.each { |file| require file }
+Sequel.extension :pg_json_ops
+require_relative 'lib/engine'
+$failed = []
+
+def migrate_db_actions(data)
```
tobymao, line 11: if you want this to be faster, you can do a bulk update or transaction

tobymao (approved):

lib/engine/game/base.rb
```diff
@@ -345,7 +345,7 @@ def initialize(names, id: 0, actions: [], pin: nil, strict: false, optional_rule
                  end
 
         # This intentionally ignores player id for now until the database is migrated.
-        @players = @names.map { |_playerid, name| Player.new(name, name) }
+        @players = @names.map { |playerid, name| Player.new(playerid, name) }
```
tobymao, line 348: underscore

jenf:
I need to fix a bug with the chat bar.

Issue #2325, opened 2020-11-15
1817: Corporation share sell buttons inconsistent (UI)


crericha:
The first sell button says "Sell a 20% ($60)" and the second "Sell 2 ($120)". In my opinion, the buttons should be consistent. Either "Sell 1" and "Sell 2" or "Sell 1 20%" and "Sell 2 20%".

![image](https://user-images.githubusercontent.com/67359577/99193143-04b30580-2745-11eb-8dbf-ea80631ebb1a.png)

[sell share inconsistency.data.txt](https://github.com/tobymao/18xx/files/5542877/sell.share.inconsistency.data.txt)

jenf:
@perwestling ?

jenf:
resolved I think

PR #2332, opened 2020-11-15
Fix problems with Buy and Sell button changes

### LINKED ISSUE #2316: After recent GUI changes in 18MEX: Can buy 20% share from market

perwestling:
Previous commit introduced possibility to show non-standard
percentage in 18MEX, but it introduced problems in other
titles. So this commit improves the solution and solves
a couple of issues.

Disregard president shares for pool shares if there are other
shares that can be bought. This means it will not be possible
to buy president's share directly from the market unless it
is the last one. (And the title need to handle that case.)

Do not let Treasury or Short share presence affect GUI
presentation. Before this fix, the "Buy a 10% Market share"
appeared when Treasury share was present, instead of just
"Buy Market Share".

Also, do not use 10% as the standard share size. Instead use
the share_percent of the corporation in question. This solves
the problem of 20% appearing for 5 share corporations in 1817.

Similar fixes for Sell view. Problem reported in issue #2325.

Fixes #2316 and #2325

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -30,15 +30,17 @@ module View
           end
 
           # Put up one buy button for each buyable percentage share type in market.
-          # In case there are more than one type of percentages in market or if shares are not 10%
-          # (e.g. 18MEX), show percentage type on button.
-          pool_shares
+          # In case there are more than one type of percentages in market or if shares are not the
+          # standard percent (e.g. 5% in 18MEX), show percentage type on button.
+          # Do skip president's share in case there are other shares available.
+          buyables = pool_shares
             .select { |share| step.can_buy?(current_entity, share) }
-            .each do |share|
-              text = pool_shares.size > 1 || share.percent != 10 ? "#{share.percent}% " : ''
-              children << h(:button, { on: { click: -> { buy_share(current_entity, share) } } },
-                            "Buy #{text}Market Share")
-            end
+            .reject { |share| share.president && pool_shares.size > 1 }
+          buyables.each do |share|
+            text = buyables.size > 1 || share.percent != @corporation.share_percent ? "#{share.percent}% " : ''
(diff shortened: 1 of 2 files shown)


perwestling:
After this fix the situation in #2325 looks like before the original change:
image

perwestling:
And after the fix the problem reported in #2316 now looks good:

image

perwestling:
And in 18MEX the 10% shares are handled as normal shares:
image

If the player has 5% shares they get a "Sell a 5% Share" at a time. That is because selling 5% shares does not change the stock price. And it is a game specific bundler in 18MEX that achieves that only one is bundled at a time.
image

And finally, if there is a choice in market what can be bought, then both percentages is shown. This is general code but it might only affect 18MEX currently.
image

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -30,15 +30,17 @@ def render
           end
 
           # Put up one buy button for each buyable percentage share type in market.
-          # In case there are more than one type of percentages in market or if shares are not 10%
-          # (e.g. 18MEX), show percentage type on button.
-          pool_shares
+          # In case there are more than one type of percentages in market or if shares are not the
+          # standard percent (e.g. 5% in 18MEX), show percentage type on button.
+          # Do skip president's share in case there are other shares available.
+          buyables = pool_shares
```
tobymao, line 36: one statement buyable = pool_shares.select do ...

perwestling: I think it needs to be done in two steps, as I use the size of pool_shares (line 38) and buyable (line 40).

tobymao (approved):

PR #2344, opened 2020-11-16
Allow users to change their name

### LINKED ISSUE #1135: Allow users to change their user name

jenf:
Modify user settings to use id (needed converting to string)
Update migrate script to modify user settings

fixes #1135

### DIFF
--- a/assets/app/view/user.rb
+++ b/assets/app/view/user.rb
@@ -38,6 +38,7 @@ module View
           ]]
         when :profile
           ['Profile Settings', [
+            render_username,
             render_notifications(setting_for(:notifications)),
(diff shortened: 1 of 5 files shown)


tobymao:
so just to confirm, if a user changes their name, there will be no migrations necessary? since everything is stored as ID now? name changes are just done automatically?

jenf:
yep, we'll need one migration just to sort the user settings out on all the games, but it took about 5 minutes to run.

tobymao (approved):

PR #2354, opened 2020-11-18
Don't show loans in spreadsheet view for games without loan mechanics

benjaminxscott:
Fixes a logic issue (Ruby apparently treats numeric zero as "True")

This caused a "Loans" column to show up, even if games didn't have a loan mechanic

Tested locally, seems to work as expected for '46 and '17


tobymao:
correct, in ruby only nil and false are truthy (as should any sane language!)

PR #2377, opened 2020-11-20
1828: Get the waterfall auction step explicitly. 

crericha:
It is not active in multiplayer games until the game starts.

### DIFF
--- a/lib/engine/game/g_1828.rb
+++ b/lib/engine/game/g_1828.rb
@@ -268,7 +268,7 @@ module Engine
                              .take(7 - @players.size)
         to_remove.each do |company|
           company.close!
-          @round.steps.find {|step| step.is_a?(Step::G1828::WaterfallAuction) }.companies.delete(company)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1828.rb
```diff
@@ -268,7 +268,7 @@ def remove_extra_private_companies
                              .take(7 - @players.size)
         to_remove.each do |company|
           company.close!
-          @round.active_step.companies.delete(company)
+          @round.steps.find { |step| step.is_a?(Step::G1828::WaterfallAuction) }.companies.delete(company)
```
tobymao, line 271: should this be out of the hot loop?

crericha: The loop is executed a maximum of four times, over two steps and only called during game setup, so I didn't optimize; however, if you want me to. I can cache the step. Let me know.

tobymao:
if it only happens during the setup thats ok

lib/engine/game/g_1828.rb
crericha: Yeah, I try to avoid optimizations that won't have an impact on performance.

PR #2383, opened 2020-11-20
better help text for routes: N/M trains and LA's overlapping routes

michaeljb:
Addresses the FAQ of N/M trains being "wrong" when really they're just counting a tokened spot.

# 1846 N/M

1846 NM

# 18 Los Angeles N/M

LA NM

# 18 Los Angeles N and N/M

LA N and NM

# 18 Los Angeles N and N/M old behavior

Replacing the passenger/freight text with "N" and "N/M" for simplicity and consistency with 1846

Screenshot from 2020-11-20 10-15-39


michaeljb:
The `train_help` method should be easy to override for other unusual trains such as 18MEX's 4D train

tobymao (approved):

Issue #2113, opened 2020-10-30
18MEX overlapping icons and names


ChrisTheCat:
The yellow Veracruz tile has the word "Veracruz" overlapping the symbol "V." Similarly Mexico City overlaps MC, Mazatlan overlaps M, etc. Some of the revenue values overlap the city names as well, as shown here:

<img width="194" alt="Screen Shot 2020-10-29 at 9 10 16 PM" src="https://user-images.githubusercontent.com/8732435/97659065-36308f00-1a2b-11eb-97f6-2dd4d934ac75.png">

perwestling:
@ChrisTheCat - is the improvement enough? If so I think this can be closed.

ChrisTheCat:
Works for now.

PR #2384, opened 2020-11-20
Add Show/Hide buttons for location names, hex coordinates, and "starting" map

### LINKED ISSUE #322: Toggle tile coordinates
A minor feature request that would help when reviewing others' routes or when describing routes verbally during synch games: overlay tile coordinates with an on/off toggle. Similar to the Map Ops-> Toggle coordinates in rr18xx:

Screenshot from 2020-05-07 15-06-02

michaeljb:
* use localStorage to persist the last show/hide click
* workaround location names overlapping with important info by making it easy to
  hide (partially addresses #2113)
* default location names to the localStorage value iff there is a value there;
  this preserves the current behavior for users (showing the names) until they
  click on something to hide them, and keeps the tests passing which expect city
  names to be there
* show the "starting" map by re-using `init_hexes`; it's not exactly the
  starting map depending on how some starting tokens/ability icons are handled,
  but should be close enough for those that want it (I don't see a GitHub issue
  for this, but it's a request that's come up a few times in chat)

[Fixes #322]

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require '../lib/storage'
 require 'view/game/axis'
 require 'view/game/hex'
 require 'view/game/tile_confirmation'
@@ -14,6 +15,9 @@ module View
       needs :selected_route, default: nil, store: true
       needs :selected_company, default: nil, store: true
       needs :opacity, default: nil
+      needs :show_coords, default: Lib::Storage['show_coords'], store: true
+      needs :show_location_names, default: (s = Lib::Storage['show_location_names']).nil? ? true : s, store: true
+      needs :show_starting_map, default: false, store: true
 
       EDGE_LENGTH = 50
       SIDE_TO_SIDE = 87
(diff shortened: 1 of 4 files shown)


michaeljb:
# Screenshots

The button now says "Show Starting Map" but when I took most of these screenshots it was still "Show Blank Map"

## Default View

Screenshot from 2020-11-20 11-05-13

## locations hidden

Screenshot from 2020-11-20 10-59-38

## locations hidden, hex coords shown

Screenshot from 2020-11-20 10-59-44

## hex coords shown

Screenshot from 2020-11-20 10-59-50


## All options flipped - location hidden, hexes visible, "starting" map shown

Screenshot from 2020-11-20 10-59-27

jenf:
I assume flat maps look okay as well?

michaeljb:
Thanks for thinking of that @jenf! I must have checked flat because I know I tweaked the position of the hex coords a little, but when I saw your comment I was worried I'd forgotten. Anyway, I think flat looks just fine: Screenshot from 2020-11-20 13-16-19

assets/app/view/game/map.rb
```diff
@@ -14,6 +15,9 @@ class Map < Snabberb::Component
       needs :selected_route, default: nil, store: true
       needs :selected_company, default: nil, store: true
       needs :opacity, default: nil
+      needs :show_coords, default: Lib::Storage['show_coords'], store: true
```
jenf, line 18: I'm not entirely sure about this.

This will work differently than other player preferences which are stored on the server.

Maybe there should be an option on the profile, and the local option is just a temporary option

michaeljb: I thought someone might want to hide the names and leave them hidden, but yeah `localStorage` just applies to the current device.

I feel like it would be fine to have the buttons by the map use (and update) the server player preferences instead of `localStorage`, seems simpler that way.

daniel-sousa-me:
Another useful button would be zoom in/out

tobymao (approved):

PR #2386, opened 2020-11-20
render upgradable preprinted tiles with color borders

### LINKED ISSUE #2087: Show printed tiles in a slightly different color
The other day I wanted to lay a tile that was no longer available, so I tried upgrading a random spot to have the tile. Then I realised that I was upgrading a tile that was printed on the map and so it wouldn't make an extra tile available.

Maybe it would be nice to have some way to distinguish printed tiles from laid tiles

michaeljb:
* this prevents them from being mistaken for tiles from the tile manifest
* sort of uses a precedent from AH's 1830, where OO hexes on the map are
  rendered with a thin yellow border

[Fixes #2087]

Screenshots with `View::Game::Hex::BORDER_COLOR_STROKE_WIDTH` set to 25 are available at https://github.com/tobymao/18xx/issues/2087#issuecomment-731370389. 25 was arbitrarily chosen but I think it looks fine, and it's easy to change if anyone wants it to.

### DIFF
--- a/lib/engine/game/g_1882.rb
+++ b/lib/engine/game/g_1882.rb
@@ -39,6 +39,8 @@ module Engine
       # Two lays or one upgrade, second tile costs 20
       TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false, cost: 20 }].freeze
 
+      PREPRINT_COLOR_ON_BORDER = %w[F12].freeze
+
       def stock_round
         Round::Stock.new(self, [
           Step::DiscardTrain,
--- a/lib/engine/game/g_1882.rb
+++ b/lib/engine/game/g_1882.rb
@@ -39,8 +39,6 @@ module Engine
       # Two lays or one upgrade, second tile costs 20
       TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false, cost: 20 }].freeze
 
-      PREPRINT_COLOR_ON_BORDER = %w[F12].freeze
-
       def stock_round
         Round::Stock.new(self, [
           Step::DiscardTrain,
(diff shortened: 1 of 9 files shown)


lib/engine/game/g_1882.rb
```diff
@@ -39,6 +39,8 @@ class G1882 < Base
       # Two lays or one upgrade, second tile costs 20
       TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false, cost: 20 }].freeze
 
+      PREPRINT_COLOR_ON_BORDER = %w[F12].freeze
```
jenf, line 42: Could this auto-calculate the preprinted tiles as part of the base class?

michaeljb: Maybe, but i was feeling lazy.

I think that ends up being more expensive, every hex with track on it would need to compare with every rotation of every tiles until it finds a match--in 1882's case the match should bearly since F12 is like a tile 8--18 Los Angeles has some yellow cities including a higher-numbered 57. I guess you'd first sort the tiles by color instead of just number... :thinking:

But then all the hexes with track that _don't_ match a tile _do_ have to check against all the tiles in the manifest.

Maybe I could finally learn how to use stackprof to investigate this :slightly_smiling_face:

jenf: Would it not be easier to make all 'printed' upgradable tiles have a border that matches the base color?

michaeljb: This is the new behavior at 077560ae, here are some screenshots:

Screenshot from 2020-11-20 22-37-52
Screenshot from 2020-11-20 22-37-37
Screenshot from 2020-11-20 22-37-29

michaeljb:
Experimenting with changing `BORDER_COLOR_STROKE_WIDTH` a little bit, I might like it better at 15:

Screenshot from 2020-11-20 22-42-37

jenf:
I also prefer 15

jenf (approved):

tobymao (approved):

PR #2393, opened 2020-11-21
18MEX: Fix bugs for valuation of share holdings (fixes #2355)

### LINKED ISSUE #2355: 18MEX bankruptcy didn't sell all shares
https://18xx.games/game/16377

```
MEX does not run
MEX's share price changes from $65 to $60
-- ChrisShaffer goes bankrupt and sells remaining shares --
ChrisShaffer sells 3 shares NdM and receives $240
NdM's share price changes from $80 to $65
-- Game over: claudio ($961), 18xxperson ($943), Hanigan ($546), ChrisShaffer ($463) --
```

ChrisShaffer should have also sold 3 shares of MEX and 0.5 shares of NdM automatically as part of the bankruptcy.

perwestling:
Valuation of NdM holdings was incorrect - if holding both 5% share(s)
and normal shares, the value of it was just the 5% share(s).

Have change the general code for bankruptcy sell - now it continues
to sell in case there are sellable bundles after the largest
package has been sold. And it will sell the biggest bundles first.
This means that the 5% shares of NdM in 18MEX will eventually be
sold.

Set 18MEX to decrease stock price per 10% sold - this means that
bankruptcy sell of one 5% at a time wont affect stock price.
Also ment that the overriden method that handle sell and change
stock price could be removed from 18MEX base class.

Refactored base class a bit to make it possible for 18MEX to
override parts of the base class to handle the 5% shares.
Also introduced all_bundles_for_corporation as it need to
be called from 18MEX. The bundles_for_corporation is overridden
by e.g. 18MEX but 18MEX also need to call the base method
directly when it needs to get bundles that include any and all
5% shares. (During valuation and dump.)


lib/engine/step/bankrupt.rb
```diff
@@ -39,10 +39,14 @@ def process_bankrupt(action)
...
-          next unless (bundle = @game.sellable_bundles(player, corporation).max_by(&:price))
 
-          @game.sell_shares_and_change_price(bundle)
-          @round.recalculate_order
+          # Do a potential repeated sell of bundles. This is important for NdM in 18MEX
+          # which might have 5% bundle(s) besides the 10%+ bundles.
+          # Most other titles this will just sell one, the largest, bundle.
+          while (bundle = @game.sellable_bundles(player, corporation).max_by(&:price))
+            @game.sell_shares_and_change_price(bundle)
+            @round.recalculate_order
```
jenf, line 48: Recalculate order only needs doing once

perwestling: Correct.
Before it was once per corporation that could be sold. 
Have now moved it out of that loop so it is just called once in all - should work.

PR #2396, opened 2020-11-21
18MEX: Introduce swap functionality (fixes #1921)

### LINKED ISSUE #1921: [18MEX] Player Should Be Allowed to "upgrade" 5% share of NdM for a 10% Share
https://gist.github.com/TIE53/7a0b026e26209670a4c66a1d95908d63

A player owning a 5% share of the NdM should be able to purchase a 10% share of the NdM from the open market by "trading in" their 5% share for a 10% share and paying the difference in cost (rule 3.2.C.5).  This does not appear to be implemented.

perwestling:
In 18MEX it is possible to swap a 5% share as part of a sell/buy.

If selling shares and there is a 5% share in the market, player
can choose to get the 5% as part of the deal, reducing the money
received for the sell. Note: Selling 30%, with a 5% swap, just
decreases stock market price as if 20% had been sold.

If buying shares and the player owns one 5% share, the player
can choose to use the 5% as a partial payment, reducing the
money needed by the value of the 5% share. This is counted
as a buy for all purposes.

Have introduced swap_sell and swap_share methods, which if
implemented will enable the swap buttons to appear in the GUI.
So far only 18MEX has these implemented.

Note! Swap button is Sell/Buy buttons, which do show info
about 5% share as part of deal.

A bunch of smaller bug fixes for 18MEX 5% share handling.
 - Now handles the 60% roof correctly for NdM.
   Note! If holding 55%, buying 10% is allowed if you swap
   your 5%. Similar if 25% in market, you can sell 30% if
   you swap in 5%.
 - Use SELL_MOVEMENT :down_by_10 instead of override of
   sell_shares_and_change_price
 - Override num_certs to handle the special case for NdM 5%.
   Have rewritten can_gain? to use similar guard as
   can_buy? and can_sell? (stop NdM actions before phase
   minors have merged)

[Fixes #1921]

### DIFF
--- a/lib/engine/action/buy_shares.rb
+++ b/lib/engine/action/buy_shares.rb
@@ -5,12 +5,13 @@ require_relative 'base'
 module Engine
   module Action
     class BuyShares < Base
-      attr_reader :entity, :bundle
+      attr_reader :entity, :bundle, :swap
 
-      def initialize(entity, shares:, share_price: nil, percent: nil)
+      def initialize(entity, shares:, share_price: nil, percent: nil, swap: nil)
         @entity = entity
         @bundle = ShareBundle.new(Array(shares), percent)
         @bundle.share_price = share_price
+        @swap = swap
       end
 
       def self.h_to_args(h, game)
@@ -18,6 +19,7 @@ module Engine
           shares: h['shares'].map { |id| game.share_by_id(id) },
           share_price: h['share_price'],
           percent: h['percent'],
+          swap: game.share_by_id(h['swap']),
         }
       end
 
@@ -26,6 +28,7 @@ module Engine
           'shares' => @bundle.shares.map(&:id),
           'percent' => @bundle.percent,
           'share_price' => @bundle.share_price,
+          'swap' => @swap&.id,
         }
       end
     end
--- a/lib/engine/action/sell_shares.rb
+++ b/lib/engine/action/sell_shares.rb
@@ -6,12 +6,13 @@ require_relative '../share_bundle'
 module Engine
   module Action
     class SellShares < Base
-      attr_reader :entity, :bundle
+      attr_reader :entity, :bundle, :swap
 
-      def initialize(entity, shares:, share_price: nil, percent: nil)
+      def initialize(entity, shares:, share_price: nil, percent: nil, swap: nil)
         @entity = entity
         @bundle = ShareBundle.new(shares, percent)
         @bundle.share_price = share_price
+        @swap = swap
       end
 
       def self.h_to_args(h, game)
@@ -19,6 +20,7 @@ module Engine
           shares: h['shares'].map { |id| game.share_by_id(id) },
           share_price: h['share_price'],
           percent: h['percent'],
+          swap: game.share_by_id(h['swap']),
         }
       end
 
@@ -27,6 +29,7 @@ module Engine
           'shares' => @bundle.shares.map(&:id),
           'percent' => @bundle.percent,
           'share_price' => @bundle.share_price,
+          'swap' => @swap&.id,
         }
       end
     end
--- a/lib/engine/step/buy_train.rb
+++ b/lib/engine/step/buy_train.rb
@@ -53,6 +53,8 @@ module Engine
         buy_train_action(action)
         pass! unless can_buy_train?(action.entity)
       end
+
+      def swap_sell(_player, _corporation, _bundle, _pool_share); end
     end
   end
 end
--- a/lib/engine/step/g_18_mex/buy_sell_par_shares.rb
+++ b/lib/engine/step/g_18_mex/buy_sell_par_shares.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../buy_sell_par_shares'
+require_relative 'swap_buy_sell'
 
 module Engine
   module Step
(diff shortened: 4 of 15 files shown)


perwestling:
Maybe not the prettiest gui change.... but here is how the swap buttons look like when selling a bundle:
image

And here is a buy situation:
image

lib/engine/action/buy_shares.rb
```diff
@@ -26,6 +28,7 @@ def args_to_h
           'shares' => @bundle.shares.map(&:id),
           'percent' => @bundle.percent,
           'share_price' => @bundle.share_price,
+          'swap' => @swap ? @swap.id : nil,
```
tobymao, line 31: @swap&.id

lib/engine/action/buy_shares.rb
```diff
@@ -5,19 +5,21 @@
...
         @bundle.share_price = share_price
+        @swap = swap
       end
 
       def self.h_to_args(h, game)
         {
           shares: h['shares'].map { |id| game.share_by_id(id) },
           share_price: h['share_price'],
           percent: h['percent'],
+          swap: h['swap'] ? game.share_by_id(h['swap']) : nil,
```
tobymao, line 22: game.share_by_id(h['swap'])

lib/engine/action/sell_shares.rb
```diff
@@ -6,19 +6,21 @@
...
         @bundle.share_price = share_price
+        @swap = swap
       end
 
       def self.h_to_args(h, game)
         {
           shares: h['shares'].map { |id| game.share_by_id(id) },
           share_price: h['share_price'],
           percent: h['percent'],
+          swap: h['swap'] ? game.share_by_id(h['swap']) : nil,
```
tobymao, line 23: same

lib/engine/step/g_18_mex/buy_sell_par_shares.rb
```diff
@@ -15,16 +15,40 @@ def can_sell?(entity, bundle)
...
+        # Check if it is possible to sell an NdM bundle if player swap 5% share from pool
+        def swap_sell(player, corporation, bundle, pool_share)
+          return if @game.ndm != corporation || pool_share.percent != 5 || bundle.percent == 5
 
-          # NdM 5% shares does not affect cert limit
-          bundle.corporation.holding_ok?(entity, bundle.percent)
+          # If we were allowed to buy another 5% then swap is OK. Test this by
+          # creating a cloned bundle where one of the shares has its percentage reduced
+          # by 5. This way we can test if the swap will not exceed market limit of 50%.
+          updated_shares = bundle.shares.map(&:dup)
```
tobymao, line 43: why do you need to dup these shares?

lib/engine/step/buy_train.rb
```diff
@@ -53,6 +53,8 @@ def process_buy_train(action)
         buy_train_action(action)
         pass! unless can_buy_train?(action.entity)
       end
+
+      def swap_sell(_player, _corporation, _bundle, _pool_share); end
```
tobymao, line 57: why do you need swap_sell in buy_train?

lib/engine/action/buy_shares.rb
perwestling: Fixed

lib/engine/action/buy_shares.rb
perwestling: Fixed

lib/engine/action/sell_shares.rb
perwestling: Fixed

lib/engine/step/g_18_mex/buy_sell_par_shares.rb
perwestling: The bundle sent in is the one that are to be sold, so I do not want to affect it, but create a clone. And for that clone I reduce the percentage with 5 (which is the size of NdM 5% shares to be swapped with).ö
Anyway, I refactored this part of the code to use a helper method. Used in both the can_buy? and can_sell? tests.

lib/engine/step/buy_train.rb
perwestling: If I don't have it then I get an undefined method 'swap_sell' from Engine::Step::BuyTrain.
From MiniRacer.

I think that emergency funding requires the Sell shares part, and that uses swap functionality.
Probably 18MEX should implement swap sell (and share the code) as there might be a situation there you want to reduce your holdings in NdM to finance a train purchase. 
I guess I should subclass buy_train.rb for 18MEX and share the implementation for swap_sell. 

What do you think about that?

tobymao: maybe it's ok

perwestling: I think I will break out the swap_buy and swap_sell (and the helper function) to an 18Mex module, and import it in both 18Mex version of buy_sell_par_shares.rb and buy_train.rb. That way it will be available during emergency buy of trains in 18Mex. (The swap_buy wont be used there but it is good to have it in the same module.)

I'll add it as another commit to this PR.

perwestling:
@tobymao I have now rebased this to latest master, and forced push it.
Changes:
- Extract the swap methods to a module
- Include this in two classes in 18Mex
- Checked that the map(&:dup) was needed. Otherwise the percentage was actually affected (which ment the cost for a market share was halfed).

I will make remove draft status as this is ready.

PR #2409, opened 2020-11-23
Add zoom buttons to map

daniel-sousa-me:


assets/app/view/game/map.rb
```diff
@@ -17,7 +17,9 @@ class Map < Snabberb::Component
       needs :opacity, default: nil
       needs :show_coords, default: Lib::Storage['show_coords'], store: true
       needs :show_location_names, default: (s = Lib::Storage['show_location_names']).nil? ? true : s, store: true
+      needs :map_zoom, default: (s = Lib::Storage['map_zoom']).nil? ? 1 : s, store: true
```
tobymao, line 20: refactor this it's bad, i didn't catch it in the previous review but the defaults should not be accessed like this

you have have them in the render method and set default values

ryandriskel:
Confirming the new buttons work as expected. Going to be very useful on larger maps!

Screen Shot 2020-11-24 at 12 03 36 PM

assets/app/view/game/map.rb
michaeljb: https://github.com/tobymao/18xx/pull/2440

michaeljb: https://github.com/tobymao/18xx/pull/2440 fixes the defaults for the properties that I added; when it's merged, @daniel-sousa-me, you could update this PR to match the pattern that's used there. I suggest waiting till it's merged in case it changes between now and then :wink:

daniel-sousa-me: Yeah, I'll wait. I just realise that rebasing now to your branch would probably fail, because merging a PR changes the commits.

tobymao:
needs updating

assets/app/view/game/map_controls.rb
```diff
@@ -7,14 +7,15 @@ module Game
...
+      needs :map_zoom, default: 1, store: true
       needs :show_starting_map, default: false, store: true
 
       def render
         children = [
           location_names_controls,
           hex_coord_controls,
           starting_map_controls,
-        ]
+        ] + map_zoom_controls
```
tobymao, line 18: put it inline with *map_szoom_conto

daniel-sousa-me:
Updated and corrected

assets/app/view/game/map.rb
```diff
@@ -188,6 +191,13 @@ def show_location_names
 
         @show_location_names.nil? ? default : @show_location_names
       end
+
+      def map_zoom
+        stored = Lib::Storage['map_zoom']
+        default = stored.nil? ? 0 : stored
```
tobymao, line 197: this is not necessary, you can just do

@map_zoom || Lib::Storage['map_zoom'] || 0

daniel-sousa-me: Yeah lol
I only did it like this because it's how the others were. I'm changing them all.

tobymao (approved):

PR #2411, opened 2020-11-23
1860 dividends, separate bank for corp transactions

roseundy:
This implements the 1860 dividend step
It tweaks the colors of the 1860 stock market
In implements an 1860-specific bank whose cash is 10000 - sum of player cash

This makes a few changes to common files:
- New colors in the stock market UI
- Game/Base to implement a bank_cash method for use by UI:Bank
- UI:Bank to call @game.bank_cash method instead of @game.bank.cash directly


lib/engine/depot.rb
```diff
@@ -6,13 +6,13 @@ module Engine
...
 
-    def initialize(trains, game)
+    def initialize(trains, game, bank)
       @game = game
       @trains = trains
       @trains.each { |train| train.owner = self }
       @upcoming = @trains.dup
       @discarded = []
-      @bank = @game.bank
+      @bank = bank
```
tobymao, line 15: init(bank = nil)
@bank = bank ||@game.bank

lib/engine/game/g_1860.rb
```diff
@@ -6,6 +6,8 @@
 module Engine
   module Game
     class G1860 < Base
+      attr_reader :cobank
```
tobymao, line 9: what's a cobank

roseundy: Company bank.

roseundy:
My last commit didn't get all of my changes :-/

I've refactored to put the second bank into common code, with @cobank set to @bank for games that don't need split banks.

lib/engine/game/g_1860.rb
roseundy: I'm open to suggestions for a better name. The idea is to have two banks: the player bank (that breaks), and a company bank used for funds into and out of corporations. Most of the time (except for 1860 now) the company bank is just an alias for the player bank.

lib/engine/depot.rb
roseundy: Refactored.

lib/engine/game/g_1860.rb
tobymao: is it just an infinite pool of money?

roseundy: Yes

tobymao: ok rename it something like that instead of cobank

roseundy: But for most games the corporation is transacting with a finite bank, so naming it "infinite_pool" wouldn't make sense either. What would you suggest? Maybe "corporate_bank"?

lib/engine/depot.rb
```diff
@@ -12,7 +12,7 @@ def initialize(trains, game)
       @trains.each { |train| train.owner = self }
       @upcoming = @trains.dup
       @discarded = []
-      @bank = @game.bank
+      @bank = game.cobank
```
tobymao, line 15: this is wrong, it should be the game.bank or a random bank you pass in right?

lib/engine/game/g_1860.rb
tobymao: this is an 1860 only construct, so corporate_bank or infinite_bank is fine.

lib/engine/depot.rb
roseundy: Ok, so the idea I'm not explaining very well is that corporations would (in general) interact with one bank and that players would interact with another. In many games (all except 1860 at this point), the two would be set to the same bank in Game::Base, but in 1860 they would be separate. So I need two different bank names depending on whether it's a corporation doing the spending/receiving or not.

lib/engine/game/g_1860.rb
roseundy: But I'm trying to refactor common code to use either the player bank or the corporate bank. Otherwise, I wind up doing ugly things like creating duplicates of a Steps just to change the bank in the spend method.

roseundy:
Lastest push has re-refactored code: no split banks, instead using an 1860-specific bank that calculates bank cash and bank breakage based on 10000 - sum of player cash

tobymao (approved):

Issue #2350, opened 2020-11-17
1817 - Company logos in different size in the order section of the main view


giovandc:
Recently it's been implemented a feature the displays different company token sizes in the market, depending on the size of the company (2/5/10). This effect should be limited to the market, and not to the operation order section in the main view too
![image](https://user-images.githubusercontent.com/72523908/99410621-fc003200-28f2-11eb-99c6-027bf99a344b.png)

jenf:
Why? Another user requested it

daniel-sousa-me:
I didn't even realise it was related to company size. I think it looks weird in both places

jenf:
Yeah, this is a case of 'can't please everyone'

I can be persuaded to remove it from the bar, but the market i think probably has value.

PedroRCSequeira:
I really like it in both places, just extra info is nice!!

giovandc:
I honestly thought that this was a bug since they look so weird skewed and out of place in the operation round order bar.
I don't really care about the market, even though I personally find it superfluous, but sure as hell isn't needed in the operation order bar, which was, before this change, neat and clean, as it should be.

jenf:
Vote up this comment if you want it removed, vote down this comment if you want it retained.

daniel-sousa-me:
I suggest that while the icons get smaller, their container stays the same size. That should improve the look

daniel-sousa-me:
I figured out what is my problem with the logos. You made the big corps bigger instead of making the small corps smaller. I think that if you reduce the size of all of them it will look a lot better.

This applies doubly to the size in the bar. If the biggest logos are of the size of the text (and the others are centred vertically), I think it will look good. Even if the logos are not legible, you still get some info with the color.

I may try to play around with this later and send some screenshots.

daniel-sousa-me:
I've done the new version, you can find the source at: https://github.com/daniel-sousa-me/18xx/tree/logo_sizes
I'll do a PR if you all think this is an improvement.

Here are the new versions:

![image](https://user-images.githubusercontent.com/3337865/99918490-285ee880-2d0f-11eb-996d-c043e55114be.png)

![image](https://user-images.githubusercontent.com/3337865/99918493-344aaa80-2d0f-11eb-80ae-122582dd0cf4.png)

daniel-sousa-me:
The thing I still don't like is the bold and especially the underline, but I don't have an alternative that's consistent with the rest of the site

giovandc:
[...]

PR #2412, opened 2020-11-23
UI work

daniel-sousa-me:
Lots of minor improvements on the spreadsheet and other places (see commits)

### DIFF
--- a/assets/app/view/game/entity_list.rb
+++ b/assets/app/view/game/entity_list.rb
@@ -54,7 +54,7 @@ module View
           children = []
           if entity.corporation?
             size = TOKEN_SIZES[@game.corporation_size(entity)]
-            vpadding = (TOKEN_SIZES[:large] - size)/2
+            vpadding = (TOKEN_SIZES[:large] - size) / 2
             logo_props = {
               attrs: { src: entity.logo },
               style: {
@@ -68,7 +68,7 @@ module View
 
           text_props = { style: { margin: 'auto' } }
           small_props = { style: { fontSize: 'smaller' } }
-          owner = "#{entity.owner.name.truncate}" if !entity.player? && entity.owner
+          owner = entity.owner.name.truncate if !entity.player? && entity.owner
           owner = 'CLOSED' if entity.closed?
 
           text = [entity.name]
--- a/assets/app/view/game/entity_list.rb
+++ b/assets/app/view/game/entity_list.rb
@@ -22,6 +22,7 @@ module View
               float: 'left',
               listStyle: 'none',
               paddingRight: '1rem',
+              display: 'flex',
             },
           }
 
@@ -53,19 +54,26 @@ module View
           children = []
           if entity.corporation?
             size = TOKEN_SIZES[@game.corporation_size(entity)]
+            vpadding = (TOKEN_SIZES[:large] - size)/2
             logo_props = {
               attrs: { src: entity.logo },
               style: {
-                padding: "#{TOKEN_SIZES[:large] - size}rem 0.4rem 0 0",
+                padding: "#{vpadding}rem 0.4rem #{vpadding}rem 0",
                 height: "#{size}rem",
+                margin: 'auto',
               },
             }
             children << h(:img, logo_props)
           end
 
-          owner = " (#{entity.owner.name.truncate})" if !entity.player? && entity.owner
-          owner = ' (CLOSED)' if entity.closed?
-          children << h(:span, "#{entity.name}#{owner}")
+          text_props = { style: { margin: 'auto' } }
+          small_props = { style: { fontSize: 'smaller' } }
+          owner = "#{entity.owner.name.truncate}" if !entity.player? && entity.owner
+          owner = 'CLOSED' if entity.closed?
+
+          text = [entity.name]
+          text += [h(:br), h(:span, small_props, owner)] if owner
+          children << h(:span, text_props, text)
 
           h(:li, entity_props, children)
         end
--- a/assets/app/view/game/entity_list.rb
+++ b/assets/app/view/game/entity_list.rb
@@ -71,8 +71,8 @@ module View
           owner = entity.owner.name.truncate if !entity.player? && entity.owner
           owner = 'CLOSED' if entity.closed?
 
-          text = [entity.name]
-          text += [h(:br), h(:span, small_props, owner)] if owner
+          text = [h(:div, entity.name)]
+          text << h(:div, small_props, owner) if owner
           children << h(:span, text_props, text)
 
           h(:li, entity_props, children)
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -108,7 +108,10 @@ module View
             },
           }
 
-          @step.available.select(&:company?).map do |company|
+          companies = @step.available.select(&:company?)
+          companies = companies.partition { |c| c == @step.auctioning }.flatten(1)
+
+          companies.map do |company|
             children = [h(Company, company: company, bids: @step.bids[company])]
             children << render_input(company) if @selected_company == company
             h(:div, props, children)
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -147,10 +147,12 @@ module View
             end,
             h(:th, @game.ipo_name),
             h(:th, 'Market'),
-            *(@hide_ipo ?
-                [h(:th, render_sort_link('Price', :share_price))] :
+            *(if @hide_ipo
+                [h(:th, render_sort_link('Price', :share_price))]
+              else
                 [h(:th, render_sort_link(@game.ipo_name, :par_price)),
-                  h(:th, render_sort_link('Market', :share_price))]),
+                 h(:th, render_sort_link('Market', :share_price))]
+              end),
             h(:th, render_sort_link('Cash', :cash)),
             h(:th, render_sort_link('Order', :order)),
             h(:th, 'Trains'),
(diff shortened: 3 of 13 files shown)


assets/app/view/game/spreadsheet.rb
```diff
@@ -253,26 +277,31 @@ def render_corporation(corporation, operating_order, current_round, index)
...
+            share_holding = num_shares_of(p, corporation).to_s unless corporation.minor?
+            share_holding = '*' if corporation.minor? && corporation.president?(p)
+            h('td.padded_number', sold_props, share_holding || '')
           end,
           h('td.padded_number', { style: { borderLeft: border_style } }, num_shares_of(corporation, corporation).to_s),
           h('td.padded_number', { style: { borderRight: border_style } },
             "#{corporation.receivership? ? '*' : ''}#{num_shares_of(@game.share_pool, corporation)}"),
-          h('td.padded_number', corporation.par_price ? @game.format_currency(corporation.par_price.price) : ''),
+          (h('td.padded_number',
+             corporation.par_price ? @game.format_currency(corporation.par_price.price) : '') unless @hide_ipo),
```
perwestling, line 293: If @hide_ipo, then it will be nil. Does it show something good?
Alternative would be 
 (h('td.padded_number',
             corporation.par_price && !@hide_ipo ? @game.format_currency(corporation.par_price.price) : '')),

daniel-sousa-me: I put a `.compact` in the end of the Array to remove the `nil`s usually I'd be appending things to a stores Array conditionally, but since this was in the middle of a very big block I thought that adding a lot of `<<` or concatenating two big Arrays would be more clumsy.

I don't like that alternative very much, because you end up with an empty column in the table.

ryandriskel:
I would suggest adding screenshots of the impacts of the changes as without a corresponding issue I am not sure how to verify the changes have the intended result.

daniel-sousa-me:
https://github.com/tobymao/18xx/issues/2350#issuecomment-731855079

daniel-sousa-me:
image

daniel-sousa-me:
In the current version, for comparison:

image

daniel-sousa-me:
They are all very minor things x)

For the others I don't think a screen shot would be more useful than the commit message

daniel-sousa-me:
There's one more thing!

Original:

image

New version:

image

michaeljb:
@daniel-sousa-me in case you missed it, there are now conflicts on this branch

daniel-sousa-me:
Thanks for the warning. I'll rebase in a moment.

Is there a way to get an email when this happens?

daniel-sousa-me:
The conflict was with my own PR x)

assets/app/view/game/spreadsheet.rb
```diff
@@ -20,6 +20,8 @@ def render
         @spreadsheet_sort_order = Lib::Storage['spreadsheet_sort_order']
 
         @players = @game.players.reject(&:bankrupt)
+        @hide_ipo = @game.all_corporations.reject(&:minor?).all?(&:always_market_price)
+        @show_corporation_size = @game.all_corporations.any?(&@game.method(:show_corporation_size?))
```
tobymao, line 24: what is this?

assets/app/view/game/spreadsheet.rb
```diff
@@ -133,8 +137,12 @@ def render_title
             end,
             h(:th, @game.class::IPO_NAME),
             h(:th, 'Market'),
-            h(:th, render_sort_link(@game.class::IPO_NAME, :par_price)),
-            h(:th, render_sort_link('Market', :share_price)),
+            *(if @hide_ipo
```
tobymao, line 140: don't use this syntax, put it outside

assets/app/view/game/spreadsheet.rb
tobymao: don't use this syntax

assets/app/view/game/stock_market.rb
```diff
@@ -91,16 +91,18 @@ def grid_1d_price(price)
 
       def grid_1d
         token_height = @game.stock_market.market.first.map do |p|
-          p.corporations.sum { |c| TOKEN_SIZES[@game.corporation_size(c)] + VERTICAL_TOKEN_PAD }
-        end.push(MIN_NUM_TOKENS * (TOKEN_SIZE + VERTICAL_TOKEN_PAD)).max
+          (p.corporations.length + 1) * TOKEN_SIZE + VERTICAL_TOKEN_PAD
```
tobymao, line 94: don't use length, use size

assets/app/view/game/entity_list.rb
```diff
@@ -53,19 +54,26 @@ def render
...
-          owner = " (#{entity.owner.name.truncate})" if !entity.player? && entity.owner
-          owner = ' (CLOSED)' if entity.closed?
-          children << h(:span, "#{entity.name}#{owner}")
+          text_props = { style: { margin: 'auto' } }
+          small_props = { style: { fontSize: 'smaller' } }
+          owner = entity.owner.name.truncate if !entity.player? && entity.owner
+          owner = 'CLOSED' if entity.closed?
+
+          text = [entity.name]
+          text += [h(:br), h(:span, small_props, owner)] if owner
```
tobymao, line 75: use concat, why use a br instead of a style

assets/app/view/game/round/auction.rb
```diff
@@ -108,7 +108,10 @@ def render_companies
             },
           }
 
-          @step.available.select(&:company?).map do |company|
+          companies = @step.available.select(&:company?)
+          companies = companies.partition { |c| c == @step.auctioning }.flatten(1)
```
tobymao, line 112: what is this code doing

assets/app/view/game/spreadsheet.rb
```diff
@@ -183,40 +191,51 @@ def render_corporations
...
-            corporation.par_price&.price || 0
-          when :share_price
-            corporation.share_price&.price || 0
-          else
-            @game.player_by_id(@spreadsheet_sort_by)&.num_shares_of(corporation)
-          end
+        result = result.map { |c, _, order| [c, order, order] } if result.to_enum.with_object(1).map(&:[]).all?(0)
+
+        result.sort_by! do |corporation, operating_order, order|
+          [case @spreadsheet_sort_by
```
tobymao, line 210: don't use this syntax, split it out

assets/app/view/game/spreadsheet.rb
```diff
@@ -183,40 +191,51 @@ def render_corporations
...
+           when :short
+             @game.available_shorts(corporation)
+           when :size
+             corporation.total_shares
+           else
+             @game.player_by_id(@spreadsheet_sort_by)&.num_shares_of(corporation)
+           end, order]
         end
 
+        result.each(&:pop)
```
tobymao, line 232: what are you trying to do here

assets/app/view/game/spreadsheet.rb
```diff
@@ -183,40 +191,51 @@ def render_corporations
...
-          when :order
-            operating_order
-          when :par_price
-            corporation.par_price&.price || 0
-          when :share_price
-            corporation.share_price&.price || 0
-          else
-            @game.player_by_id(@spreadsheet_sort_by)&.num_shares_of(corporation)
-          end
+        result = result.map { |c, _, order| [c, order, order] } if result.to_enum.with_object(1).map(&:[]).all?(0)
```
tobymao, line 207: what is this?

assets/app/view/game/spreadsheet.rb
daniel-sousa-me: Which one?

assets/app/view/game/spreadsheet.rb
daniel-sousa-me: It's checking if there is some corp where those columns will be used. If not, this variable is checked later to hide the column

assets/app/view/game/spreadsheet.rb
daniel-sousa-me: Never mind. Got it from the other comment

assets/app/view/game/round/auction.rb
daniel-sousa-me: During the bidding of privates in 1817 all the 11 privates are shown, but we are only bidding on one of them. This is putting the one we're currently bidding on the beginning.

assets/app/view/game/spreadsheet.rb
daniel-sousa-me: If corporations are equal in the attribute we are sorting by, it reverts to sorting by market order. During the OR we have an array explicitly storing the order, so we use it. When we don't have that array, we use the current order in the market.

assets/app/view/game/spreadsheet.rb
daniel-sousa-me: We are storing the corp order in 2 ways, but after sorting we've figured out which one we have so we're removing the other.

assets/app/view/game/round/auction.rb
tobymao: too much movement, i don't think wee need to do that revert this change

assets/app/view/game/spreadsheet.rb
tobymao: rewrite it any? { |c| c.repsondsto}

assets/app/view/game/spreadsheet.rb
tobymao: i feel like you could do this easier, make it more clean

assets/app/view/game/spreadsheet.rb
tobymao: too convoluted

assets/app/view/game/spreadsheet.rb
daniel-sousa-me: The problem is that order can mean 2 different things and sometimes we care about both of those at the same time.

I changed it the little that I could. I hope it's good enough

assets/app/view/game/round/auction.rb
daniel-sousa-me: I think that we really need to do something. Everyone is always complaining that it's hard to find where is the private that it's up for auction. There is barely any visual indication. This just puts one corp to the top when it's auctioned. In all other comparable auctions you don't even see the other entities.

daniel-sousa-me:
Since this doesn't seem to be going anywhere, I'm marking this as a draft and will resubmit in parts when I get time (which probably won't happen in the next 2 weeks)

tobymao:
closing for now

PR #2415, opened 2020-11-23
Hide bankruptcies from the spreadsheet and put them in the end in the entities tab

daniel-sousa-me:


assets/app/view/game/entities.rb
```diff
@@ -43,6 +45,10 @@ def render
           *bank_owned.map { |c| h(Corporation, corporation: c) },
         ].compact)
 
+        children += bankrupt_players.map do |p|
```
tobymao, line 48: concat

lib/engine/step/g_1817/bankrupt.rb
```diff
@@ -45,6 +45,8 @@ def process_bankrupt(action)
             corporation.owner = @game.share_pool
           end
 
+          @round.recalculate_order
```
jenf, line 48: looks like this needs a merge from master

assets/app/view/game/entities.rb
```diff
@@ -22,6 +22,8 @@ def render
           players = players.rotate(i)
         end
 
+        bankrupt_players, players = players.partition(&:bankrupt)
```
ryandriskel, line 25: TIL the partition function 👍

daniel-sousa-me: Since I don't know Ruby very well, I always have to search for solutions to most things. So follow my commits if you want to keep learning obscure random methods xD

PR #2417, opened 2020-11-23
recalculate_order was not being called in 1817 some times

daniel-sousa-me:
It now also removes corporations in liquidation

### DIFF
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -74,6 +74,8 @@ module Engine
         # to change order. Re-sort only them.
         index = @entity_index + 1
         @entities[index..-1] = @entities[index..-1].sort if index < @entities.size - 1
+
+        @entities.pop while @entities.last.corporation? && @entities.last.share_price.liquidation?
       end
 
       def teleported?(entity)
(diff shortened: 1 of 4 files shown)


daniel-sousa-me:
This is high priority, because it's breaking games

lib/engine/step/train.rb
```diff
@@ -90,7 +90,6 @@ def process_sell_shares(action)
         @last_share_sold_price = action.bundle.price_per_share
         super
         @corporations_sold << action.bundle.corporation
-        @round.recalculate_order
```
jenf, line 93: This needs to be left

lib/engine/round/operating.rb
```diff
@@ -74,6 +74,8 @@ def recalculate_order
         # to change order. Re-sort only them.
         index = @entity_index + 1
         @entities[index..-1] = @entities[index..-1].sort if index < @entities.size - 1
+
+        @entities.pop while @entities.last.corporation? && @entities.last.share_price.liquidation?
```
jenf, line 78: would this break if entities is empty?

lib/engine/step/train.rb
daniel-sousa-me: Doesn't `super` call the one in `emergency_money.rb`?

lib/engine/round/operating.rb
daniel-sousa-me: It does... I don't know how I missed that :disappointed:

jenf (approved):

PR #2426, opened 2020-11-24
Add links to dividend action on the OR history table on the spreadsheet

daniel-sousa-me:


assets/app/view/game/actionable.rb
```diff
@@ -95,6 +97,30 @@ def clear_ui_state
         store(:selected_company, nil, skip: true)
         store(:tile_selector, nil, skip: true)
       end
+
+      def link(text, title, action_id = nil, style_extra = {})
```
jenf, line 101: probably better this being called history_link

assets/app/lib/params.rb
```diff
@@ -9,6 +9,7 @@ def self.[](key)
     end
 
     def self.add(route, k, v)
+      route = route.split('#')[0]
```
ryandriskel, line 12: What is happening here?

ryandriskel: I didn't do a good job asking my question.

Perhaps a single line of `route.split(/[\?\#]/)[0]`

daniel-sousa-me: That should also work. I don't know which is preferred

tobymao: yea make it one line

tobymao:
what does this look like?

assets/app/view/game/actionable.rb
```diff
@@ -95,6 +97,30 @@ def clear_ui_state
         store(:selected_company, nil, skip: true)
         store(:tile_selector, nil, skip: true)
       end
+
+      def history_link(text, title, action_id = nil, style_extra = {})
```
tobymao, line 101: why do you need sytle extra seems unnecessary

daniel-sousa-me: Because the links in the history control (where this is being generalised from) have a margin in the anchor tag that shouldn't apply to links else. I can wrap them in a span with the margin instead, if you prefer.

daniel-sousa-me:
> what does this look like?

It doesn't

When you click on the amount of the dividend on the table it automatically takes you to the action, but there is no visual indication "by default", because it is already too cluttered. When you hover, you get then hand cursor and the color slightly changes as with other links, but that's it.

tobymao (approved):

PR #2431, opened 2020-11-24
[18CO] Corporate Buy Shares

ryandriskel:

### Background

18CO allows corporations to buy shares from two possible places:
- The market
- The corporation's president

Once a purchase of a corporation is made, only purchases of that specific corporation can be made in that round, though from either source.

### Implementation

I chose to implement this as a generic ( non-18CO ) specific class as I believe there will be other games that allow cross-corp share purchasing.

A new action had to be created to allow the correct actions / display to work, though it implements no specific code of its own.

The displays have a decent amount of repeated code which I am not a fan of, but the code was also just different enough to seem difficult to refactor.

Flag: **CORPORATE_BUY_SHARE_SINGLE_CORP_ONLY** - Without this flag, a corporation can purchase shares of multiple other corporations during the CBS step.
Flag: **CORPORATE_BUY_SHARE_ALLOW_BUY_FROM_PRESIDENT** - Without this flag, a corporation can only purchase shares from the market.

**_There currently is nothing to handle what happens when one corporation should become president of another corporation. This triggers a takeover in 18CO, but I want to handle that in a separate PR. Right now the player remains President._**

### Display

Corporation's holdings of other corporation shares show up in a list similar to the player's shares, with the corporation logo indicating it is a share in a different corporation. Here, CM owns 30% of DSNG and 10% of CS.

Screen Shot 2020-11-24 at 10 08 13 AM

Buttons are available only for the specific shares that can be bought beneath potential sources.

Screen Shot 2020-11-24 at 9 20 40 AM

This shows the changes to the share pool, allowing the Corporation to buy from the President and the corresponding log working as expected.

Screen Shot 2020-11-24 at 9 21 29 AM

**Styling suggestions appreciated. 😃**

### DIFF
--- a/assets/app/view/game/corporate_buy_shares.rb
+++ b/assets/app/view/game/corporate_buy_shares.rb
@@ -1,13 +1,12 @@
 # frozen_string_literal: true
 
 require 'view/game/actionable'
+require 'view/game/button/buy_share'
 
 module View
   module Game
     class CorporateBuyShares < Snabberb::Component
       include Actionable
-
-      needs :game, store: true
 
       def render
         @step = @game.round.active_step
@@ -29,11 +28,13 @@ module View
         }
 
         @step.source_list(@entity).map do |source|
+          next if source.corporation? && !@game.corporation_available?(source)
+
           children = []
           if source.player?
             children << h(Player, player: source, game: @game)
             children << render_player_input(source)
-          elsif source.corporation? && @game.corporation_available?(source)
+          elsif source.corporation?
             children << h(Corporation, corporation: source)
             children << render_corporation_input(source)
           end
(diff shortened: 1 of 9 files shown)


ryandriskel:
I have converted this to draft not because I am still working on it, but because I will be traveling over the coming week and thus don't want it to get merged in while I am not available to troubleshoot, should the unexpected need arise. I would still appreciate feedback in the meantime.

assets/app/view/game/corporate_buy_shares.rb
```diff
@@ -0,0 +1,96 @@
...
+          children = []
+          if source.player?
+            children << h(Player, player: source, game: @game)
+            children << render_player_input(source)
+          elsif source.corporation? && @game.corporation_available?(source)
+            children << h(Corporation, corporation: source)
+            children << render_corporation_input(source)
+          end
+          h(:div, props, children)
+        end.compact
```
tobymao, line 41: why do you call compact here

assets/app/view/game/corporate_buy_shares.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      def render_player_input(player)
+        return unless @step.current_actions.include?('corporate_buy_shares')
+
+        input = []
+        player.shares.group_by(&:corporation).values.each do |corp_shares|
```
tobymao, line 48: input = palyer.shares.map...

assets/app/view/game/corporate_buy_shares.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        h('div.margined_bottom', { style: { width: '20rem' } }, input.compact)
+      end
+
+      def render_player_buttons(shares)
+        children = []
+
+        shares
+          .select { |share| @step.can_buy?(@entity, share.to_bundle) }
+          .each do |share|
```
tobymao, line 60: i prefer to just an each next inteasd of a select each

shares.each do |share|
next if...

tobymao: rather do select map without the chidlren <<

assets/app/view/game/corporate_buy_shares.rb
```diff
@@ -0,0 +1,96 @@
...
+        pool_shares = @game.share_pool.shares_by_corporation[corporation].group_by(&:percent).values.map(&:first)
+        input = [render_corporation_buttons(pool_shares)]
+
+        h('div.margined_bottom', { style: { width: '20rem' } }, input.compact)
+      end
+
+      def render_corporation_buttons(shares)
+        children = []
+
+        shares
```
tobymao, line 81: same herechange to select map

assets/app/view/game/corporate_buy_shares.rb
ryandriskel: I started with block 73-85 from view/game/round/stock.rb but I see .compact is used there because the map can produce nils. However, this lead me to finding a tiny bug with my code where an empty div was produced if the corporation wasn't available. I will move that check into a "next" line and now the .compact should be applicable.

If it is still not applicable, then I'm not sure why we do it in stock.rb either?

assets/app/view/game/corporate_buy_shares.rb
ryandriskel: I have taken this pattern from lines 107-112 in view/game/buy_sell_shares.rb

Shall I update there, too?

tobymao: yes

assets/app/view/game/corporate_buy_shares.rb
tobymao: it's applicable in view/lround/stock because it uses next which causes nils

assets/app/view/game/corporate_buy_shares.rb
ryandriskel: I'll submit changes to view/game/buy_sell_shares.rb in a separate PR. There's a few there and I don't want to dilute the purpose of this PR.

ryandriskel:
Looks like the logic for share buy buttons of shares of different percentages has changed since I created this. I'm going to need to re-examine what's happening in lines 38-46 of buy_sell_shares.rb to see what should be changed.

Considering the shared logic, I am considering if I should create a view class for the button.

ryandriskel:
I have updated this to use the new Buy Share Button and verified it works as intended.

assets/app/view/game/corporate_buy_shares.rb
```diff
@@ -0,0 +1,83 @@
...
+
+require 'view/game/actionable'
+require 'view/game/button/buy_share'
+
+module View
+  module Game
+    class CorporateBuyShares < Snabberb::Component
+      include Actionable
+
+      needs :game, store: true
```
tobymao, line 11: this is not necessary because it's a part of actionable

ryandriskel: Removed and verified

ryandriskel:
I believe this one is up to date.

tobymao (approved):

PR #2432, opened 2020-11-24
[18CO] Fix missing market text and color definitions

ryandriskel:
**_This change only affects 18CO_**

Recent changes removing Acquisition from the base game definition caused it to be missing from the Market. Additionally, the colors were backwards.

Additionally, add some helpful text to the market zone definitions, as implementing separate par zones as currently not supported.

Screen Shot 2020-11-24 at 10 55 51 AM

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -43,6 +43,16 @@ module Engine
       GREEN_TOWN_TILES = %w[co8 co9 co10].freeze
       BROWN_CITY_TILES = %w[co4 63].freeze
 
+      STOCKMARKET_COLORS = {
+        par: :yellow,
+        acquisition: :red,
+      }.freeze
+
+      MARKET_TEXT = {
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_co.rb
```diff
@@ -43,6 +43,16 @@ class G18CO < Base
       GREEN_TOWN_TILES = %w[co8 co9 co10].freeze
       BROWN_CITY_TILES = %w[co4 63].freeze
 
+      STOCKMARKET_COLORS = {
+        par: :yellow,
+        acquisition: :red,
+      }.freeze
+
+      MARKET_TEXT = {
+        par: 'Par: C [40, 50, 60, 75] - 40%, B/C [80, 90, 100, 110] - 50%, A/B/C: [120, 135, 145, 160] - 60%',
```
jenf, line 52: Looks good, i'm putting a change in for 1867 that might make the par group a bit cleaner

jenf (approved):

lib/engine/game/g_18_co.rb
jenf: but that improvement can be done later.

PR #2436, opened 2020-11-25
First part of the improvements on the log

daniel-sousa-me:
Rewriting how the log is created to store more structured data, trying make it more readable and useful. Built upon work started by dlpierce

* Added logger class

* Added types of logs for actions, messages and undo (consequences and
  events will be added later)

* Show time of the action/message in the log

* Links to a point in history from the log

* Show small icon on log when some undos happened

### DIFF
--- a/assets/app/view/chat.rb
+++ b/assets/app/view/chat.rb
@@ -14,7 +14,6 @@ module View
 
     def render
       @connection.subscribe('/chat', 0) do |data|
-        data[:created_at] = Time.at(data[:created_at])
         add_line(data)
       end unless @subscribed
 
@@ -35,7 +34,7 @@ module View
         if code && code == 13
           message = event['target']['value']
           if message.strip != ''
-            add_line(user: @user, created_at: Time.now, message: message)
+            add_line(user: @user, created_at: Time.now.to_i, message: message)
             event['target']['value'] = ''
             @connection.post('/chat', message: message)
           end
--- a/assets/app/view/chat.rb
+++ b/assets/app/view/chat.rb
@@ -74,6 +74,7 @@ module View
     end
 
     def add_line(data)
+      data[:created_at] = Time.at(data[:created_at])
       store(:log, @log << data)
     end
   end
--- a/assets/app/view/chat.rb
+++ b/assets/app/view/chat.rb
@@ -14,6 +14,7 @@ module View
 
     def render
       @connection.subscribe('/chat', 0) do |data|
+        data[:created_at] = Time.at(data[:created_at])
         add_line(data)
       end unless @subscribed
 
@@ -34,7 +35,7 @@ module View
         if code && code == 13
           message = event['target']['value']
           if message.strip != ''
-            add_line(user: @user, created_at: Time.now.to_i, message: message)
+            add_line(user: @user, created_at: Time.now, message: message)
             event['target']['value'] = ''
             @connection.post('/chat', message: message)
           end
@@ -74,7 +75,6 @@ module View
     end
 
     def add_line(data)
-      data[:created_at] = Time.at(data[:created_at])
       store(:log, @log << data)
     end
   end
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -99,7 +99,7 @@ module View
         store(:tile_selector, nil, skip: true)
       end
 
-      def history_link(text, title, action_id = nil, style_extra = {})
+      def history_link(text, title, action_id = nil)
         route = Lib::Params.add(@app_route, 'action', action_id)
 
         click = lambda do
@@ -118,7 +118,6 @@ module View
           style: {
             color: 'currentColor',
             textDecoration: 'none',
-            **style_extra,
           },
         )
       end
--- a/assets/app/view/game/actionable.rb
+++ b/assets/app/view/game/actionable.rb
@@ -122,29 +122,5 @@ module View
         )
       end
     end
-
-    def history_link(text, title, action_id = nil, style_extra = {})
-      route = Lib::Params.add(@app_route, 'action', action_id)
-
-      click = lambda do
-        store(:round_history, @game.round_history, skip: true) unless @round_history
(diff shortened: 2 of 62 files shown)


daniel-sousa-me:
This PR includes #2426. So it might make sense to merge it first

daniel-sousa-me:
image

image

daniel-sousa-me:
It would be really nice if this could be merged within a week or so. The next season of the 1846 league will start on December 10th. One of the consequences of this PR is that it exposes the time the players acted. I'd like to use that data to try to make the groups as timezone-uniform as possible.

assets/app/lib/params.rb
```diff
@@ -9,6 +9,7 @@ def self.[](key)
     end
 
     def self.add(route, k, v)
+      route = route.split('#')[0]
```
tobymao, line 12: this looks similar to before

assets/app/view/chat.rb
```diff
@@ -74,6 +74,7 @@ def render
     end
 
     def add_line(data)
+      data[:created_at] = Time.at(data[:created_at])
```
tobymao, line 77: why is this done here?

tobymao:
this PR has changes from your history link commit, please fix both

assets/app/lib/params.rb
daniel-sousa-me: This includes the commits from the other PR. I've fixed it there and rebased this one.

assets/app/view/chat.rb
daniel-sousa-me: Because this code was written in patches and it got confusing keeping the data structures consistent on the chat and on the actual log. But looking back it makes no sense to be here, because on sending a message it's converted to timestamp and back to object.

It's moved.

assets/app/lib/params.rb
daniel-sousa-me: I just realised that this strategy will probably fail because merging PRs changes the commits, so I'll probably have to rebase again when the other is merged.

If you want to merge them at the same time, you probably could just merge this one. (I'm sorry, I'm experienced with git but not with github and especially with PRs)

tobymao:
can you clean up this commit?

lib/engine/logger.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
```
tobymao, line 1: not sure we need this logger class

assets/app/view/chat.rb
```diff
@@ -14,6 +14,7 @@ class Chat < Snabberb::Component
 
     def render
       @connection.subscribe('/chat', 0) do |data|
+        data[:created_at] = Time.at(data[:created_at])
```
tobymao, line 17: why is this done?

assets/app/view/game/actionable.rb
```diff
@@ -95,6 +97,54 @@ def clear_ui_state
         store(:selected_company, nil, skip: true)
         store(:tile_selector, nil, skip: true)
       end
+
+      def history_link(text, title, action_id = nil, style_extra = {})
```
tobymao, line 101: fix this

assets/app/view/log.rb
```diff
@@ -48,37 +48,70 @@ def render
...
-          h(:div, { style: { fontWeight: 'bold' } }, "#{sender}: #{line.message}")
+
+          children << h(:div, line_props, line)
+        elsif line.is_a?(Hash)
+          require 'view/log_line'
+          children << if line[:type]
+                        h('div.logline', line_props, [h(LogLine, line: line)])
+                      else
+                        h('div.chatline', line_props, [
+                                      h('span.time', time_props, time_str),
```
tobymao, line 106: formatting is off

lib/engine/action/assign.rb
```diff
@@ -8,7 +8,7 @@ class Assign < Base
       attr_reader :target
 
       def initialize(entity, target:)
-        @entity = entity
+        super(entity)
```
tobymao, line 11: why are all of these changes? i think it's probably too much change and we should rethink it

lib/engine/action/base.rb
```diff
@@ -40,6 +43,7 @@ def to_h
           'entity_type' => type_s(entity),
           'id' => @id,
           'user' => @user,
+          'created_at' => @created_at&.to_i,
```
tobymao, line 46: this seems unnecessary, you already have created_at in the db

tobymao:
this pr probably needs a lot of work, it looks too complicated right now

lib/engine/logger.rb
daniel-sousa-me: Everything could be included in the Game class, but it getting too big and this way the code will be more organised. This class will still grow a bit, because some lines of the logs will probably have relationships with other lines.

assets/app/view/chat.rb
daniel-sousa-me: The time is stored on the server as a timestamp, but then it's handy to have the variable as a Time object that we can directly call its methods

lib/engine/action/assign.rb
daniel-sousa-me: That part was done by @dlpierce. I'm not sure why he did that

lib/engine/action/base.rb
daniel-sousa-me: This was also done by @dlpierce. I think you may need it because of hotseat games

daniel-sousa-me:
@tobymao Can you take a look at this again? I have addressed every comment you made

lib/engine/action/assign.rb
dlpierce: This is so every action will have @created_at from Action::Base.

lib/engine/action/base.rb
dlpierce: I didn't think the action timestamp was exposed in the json. If it's available some other way then this is probably not necessary.

tobymao:
closing for now

PR #2438, opened 2020-11-25
[18CO] Towns on route ends are counted against city limit.

ryandriskel:

In 18CO, unlimited towns are paid but towns on route ends are counted against city limit. It is legal for a 2 train to run "C t t t t" for example, but not "t C t t t". Similarly, a 4 train can run "t C C t" or "C t C C t" but not "t C C C t".

Screen Shot 2020-11-24 at 7 12 58 PM

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -185,6 +185,23 @@ module Engine
         end
       end
 
+      def check_distance(route, visits)
+        super
+
+        distance = route.train.distance
+
+        return if distance.is_a?(Numeric)
+
+        cities_allowed = distance.find { |d| d['nodes'].include?('city') }['pay']
+        cities_visited = visits.count { |v| v.city? || v.offboard? }
+        start_at_town = visits.first.town? ? 1 : 0
+        end_at_town = visits.last.town? ? 1 : 0
+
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_co.rb
```diff
@@ -185,6 +185,23 @@ def action_processed(action)
...
       end
 
+      def check_distance(route, visits)
+        super
+
+        distance = route.train.distance
+
+        return if distance.is_a?(Numeric)
+
+        cities_allowed = distance.select { |d| d['nodes'].include?('city') }.first['pay']
```
tobymao, line 195: distance.find { ...}[0]['pay']

ryandriskel: Updated

michaeljb: `distance.select { ... }` returns an array that is filtered down from `distance`, while `distance.find { ...}` returns just the first element from `distance` that matches the filter.

I think instead of `distance.find { ...}[0]['pay']`, you just want `distance.find { ...}['pay']`

ryandriskel: Thanks @michaeljb - yeah, I have a bad habit of that. I updated to your fix and actually tested this time.

PR #2439, opened 2020-11-25
1856 tile upgrades

ventusignis:
Tile upgrades:
L cities 
Hamilton
Dits to Plain Track (Phase 6+)
Dits to Yellow Cities (Phase 6+)
Double Dits to Green Cities (Phase 6+)
Barrie/London


lib/engine/game/g_1856.rb
```diff
@@ -47,6 +48,96 @@ class G1856 < Base
...
+
+        @straight_track ||= @tiles.find { |t| t.name == '9' }
+        @sharp_track ||= @tiles.find { |t| t.name == '7' }
+        @gentle_track ||= @tiles.find { |t| t.name == '8' }
+
+        @x_city ||= @tiles.find { |t| t.name == '14' }
+        @k_city ||= @tiles.find { |t| (t.name == '15') }
+      end
+
+      def gray_phase?
```
tobymao, line 64: why do you need this? can't you just do @phase.name == :gray or something?

lib/engine/game/g_1856.rb
```diff
@@ -47,6 +48,96 @@ class G1856 < Base
...
+
+      #
+      # Get the currently possible upgrades for a tile
+      # from: Tile - Tile to upgrade from
+      # to: Tile - Tile to upgrade to
+      # special - ???
+      def upgrades_to?(from, to, special = false)
+        return false if from.name == '470'
+        # double dits upgrade to Green cities in gray
+        return @phase.tiles.include? 'gray' if to.name == '14' && %w[55 1].include?(from.name)
```
tobymao, line 87: clean this up to just check @phase.name helper

lib/engine/game/g_1856.rb
```diff
@@ -47,6 +48,96 @@ class G1856 < Base
...
+      # Get all possible upgrades for a tile
+      # tile: The tile to be upgraded
+      # tile_manifest: true/false Is this being called from the tile manifest screen
+      #
+      def all_potential_upgrades(tile, tile_manifest: false)
+        upgrades = super
+        return upgrades unless tile_manifest
+
+        # In phase 6+ single dits may be turned into plain yellow track or yellow cities
+        if @phase.tiles.include?('gray')
```
tobymao, line 126: use your helper

lib/engine/game/g_1856.rb
```diff
@@ -47,6 +48,96 @@ class G1856 < Base
...
+          upgrades |= [@straight_city, @straight_track] if tile.name == '4'
+          upgrades |= [@gentle_city, @gentle_track] if tile.name == '58'
+          upgrades |= [@sharp_city, @sharp_track] if tile.name == '3'
+          # furthermore, double dits may be upgraded to green cities, if track can be preserved
+          upgrades |= [@x_city] if tile.name == '55'
+          upgrades |= [@x_city] if tile.name == '1'
+          upgrades |= [@k_city] if tile.name == '56'
+          upgrades |= [@k_city] if tile.name == '2'
+        end
+        upgrades |= @tiles.find { |t| t.name == '127' } if tile.name == '125'
```
tobymao, line 136: a 125 can upgrade to a 127??

lib/engine/step/g_1856/track.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
```
tobymao, line 1: what's the point of this file? it looks copy and pasted

lib/engine/game/g_1856.rb
ventusignis: Can lib/engine/step/g_1856 access @phase ? If so this method is not needed

lib/engine/step/g_1856/track.rb
ventusignis: It is copy and pasted; the difference here is that the 1856 tracker is used instead of the base one. Is there a cleaner way of doing this? I am still trying to figure out how to effectively use the framework

lib/engine/game/g_1856.rb
ventusignis: No; I went and fixed the tile manifest (it also shows OO -> Hamilton now)

lib/engine/game/g_1856.rb
ventusignis: If you are asking why have the helper, this keeps the logic inside here, so that if it needs to be optimized later it only gets optimized in one place

If you are asking why `@phase.tiles.include?('gray')`, as an alternative `%w[6 8 D].include?(@phase.name)` can be used but between the two `@phase.tiles.include?('gray')` seems to be more straightforward of the two

lib/engine/step/g_1856/track.rb
ryandriskel: Instead of duplicating the whole file, you should extend the default Track class, and then override the specifics with what you need. It is possible that the "include Tracker" in the base needs to be moved until a place you can override.

ventusignis: The contents of the class have been removed and it extends from the base Track step

lib/engine/step/g_1856/tracker.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require_relative '../tracker'
+
+module Engine
+  module Step
+    module G1856
+      module Tracker
+        include Step::Tracker
+        def available_hex(entity, hex)
```
tobymao, line 10: why do you need this? it's the default behavior of step track

lib/engine/step/g_1856/tracker.rb
```diff
@@ -0,0 +1,56 @@
...
+        def old_paths_are_preserved(old_paths, new_paths)
+          # if gray phase, towns can be upgraded or downgraded
+          # and there are no tiles mixed with towns and other things
+          # so if it is gray phase, and the tile has towns, then we only need
+          # to test that exits are preserved
+
+          return old_paths.all? do |path|
+            path.exits.all? do |exit|
+              new_paths.any? do |new_path|
+                new_path.exits.any? do |new_exit|
```
tobymao, line 41: this is a really heavy nested loop is it really necessary?

for example new_payths.exits.include?(exit)

or (path.exists & new_path.exist).any?

lib/engine/step/g_1856/tracker.rb
```diff
@@ -0,0 +1,56 @@
...
+          old_paths = hex.tile.paths
+          old_ctedges = hex.tile.city_town_edges
+
+          new_paths = tile.paths
+          new_exits = tile.exits
+          new_ctedges = tile.city_town_edges
+          extra_cities = [0, new_ctedges.size - old_ctedges.size].max
+
+          new_exits.all? { |edge| hex.neighbors[edge] } &&
+          (new_exits & available_hex(entity, hex)).any? &&
```
tobymao, line 24: add indentation here

lib/engine/step/g_1856/tracker.rb
ventusignis: I removed the quad nesting by using hashes, is this good?
```ruby
        def old_paths_are_preserved(old_paths, new_paths)
          # if gray phase, towns can be upgraded or downgraded
          # and there are no tiles mixed with towns and other things
          # so if it is gray phase, and the tile has towns, then we only need
          # to test that exits are preserved
          old_exits = {}
          new_exits = {}
          old_paths.each { |path| path.exits.each { |exit| old_exits[exit] = true } }
          new_paths.each { |path| path.exits.each { |exit| new_exits[exit] = true } }
          return old_exits.all? { |exit, exits| !exits or new_exits[exit] } if
            @game.gray_phase? && old_paths.any? { |old_path| path_has_town(old_path) }
          old_paths.all? { |path| new_paths.any? { |p| path <= p } }
        end
```

tobymao: why don't you just use arrays set functions?

ventusignis: If I try to use `to_set` on the arrays, it looks like I need to `require 'set'` but if I do `require 'set'` I get this error on the page `error | "LoadError: cannot load such file -- set"`

tobymao: you don't need to use a set, just use an array

lib/engine/step/g_1856/track.rb
```diff
@@ -0,0 +1,15 @@
...
+
+require_relative '../base'
```
tobymao, line 3: what's the point of this file? should we just delete it?

ventusignis: The point of this file is to inject the 1856::Tracker into the default Track logic

ventusignis: It looks like I can just put the contents of G1856::Tracker here and have G1856::Track work just the same so I'll do that once I've been able to clean up G1856::Tracker

tobymao:
can you change from hash to using array set logic

ventusignis:
Yeah, I should be able to get to it this afternoon. Working on this has been interrupted by "need to do this now" things over the past few days

lib/engine/step/g_1856/tracker.rb
ventusignis: I've updated the branch; the logic is now in `lib/engine/step/g_1856/track.rb`

ventusignis:
I just broke the branch badly by trying to rebase it to current master; I'm looking into this

ventusignis:
I fixed it but there are some merge master commits; I'm not sure if that's a big deal for the repo or not. I'd be fine with making a new branch once the code here is good and submitting that as a clean PR

lib/engine/step/g_1856/track.rb
```diff
@@ -0,0 +1,46 @@
...
+          new_ctedges = tile.city_town_edges
+          extra_cities = [0, new_ctedges.size - old_ctedges.size].max
+
+          new_exits.all? { |edge| hex.neighbors[edge] } &&
+            (new_exits & available_hex(entity, hex)).any? &&
+            old_paths_are_preserved(old_paths, new_paths) &&
+            # Count how many cities on the new tile that aren't included by any of the old tile.
+            # Make sure this isn't more than the number of new cities added.
+            # 1836jr30 D6 -> 54 adds more cities
+            extra_cities >= new_ctedges.count { |newct| old_ctedges.all? { |oldct| (newct & oldct).none? } }
```
tobymao, line 24: use empty? instead of none?

lib/engine/step/g_1856/track.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+
+        def old_paths_are_preserved(old_paths, new_paths)
+          # if gray phase, towns can be upgraded or downgraded
+          # and there are no tiles mixed with towns and other things
+          # so if it is gray phase, and the tile has towns, then we only need
+          # to test that exits are preserved
+          old_exits = old_paths.map(&:exits).flatten.uniq
```
tobymao, line 32: flat_map(&:exist)

lib/engine/step/g_1856/track.rb
```diff
@@ -0,0 +1,46 @@
...
+          new_exits = new_paths.map(&:exits).flatten.uniq
+          return old_exits.all? { |exit| new_exits.include?(exit) } if
```
tobymao, line 34: (old_exits - new_exits).empty?

lib/engine/step/g_1856/track.rb
```diff
@@ -0,0 +1,46 @@
...
+            @game.gray_phase? && old_paths.any? { |old_path| path_has_town(old_path) }
```
tobymao, line 35: don't need this helper method, can just do old_paths.any?(&:town)

PR #2440, opened 2020-11-25
move the localStorage lookup out of the needs declaration

michaeljb:
addresses https://github.com/tobymao/18xx/pull/2409#discussion_r528455233

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -15,8 +15,8 @@ module View
       needs :selected_route, default: nil, store: true
       needs :selected_company, default: nil, store: true
       needs :opacity, default: nil
-      needs :show_coords, default: Lib::Storage['show_coords'], store: true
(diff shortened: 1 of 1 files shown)


assets/app/view/game/map.rb
```diff
@@ -174,6 +174,22 @@ def render_map
           ]),
         ])
       end
+
+      def show_coords
+        default = false
+        stored = Lib::Storage['show_coords']
+        default = stored.nil? ? default : stored
```
tobymao, line 181: default = stored.nil? ? false : Lib::Storage['show_coords']

ryandriskel: toby, Why access Lib twice? ( the first time defines stored )

ryandriskel: `def show_coords
        stored = Lib::Storage['show_coords']
        default = stored.nil? ? false : stored`

Seems like maybe what you meant?

tobymao: yes thanks

PR #2441, opened 2020-11-25
Refactor logic used to verify connection for the lay_tile ability

crericha:
The logic to verify connection for the lay_tile ability made a lot of assumptions. Refactored the logic to be more robust. Moved logic specific to 1889 to the 1889's special_track class. This paves the way for 1828 to reuse the logic in a future commit.


lib/engine/step/special_track.rb
```diff
@@ -16,16 +16,8 @@ def actions(entity)
...
-        "Lay Track for #{@company.name}"
-      end
-
-      def active_entities
-        @company ? [@company] : super
-      end
-
       def blocks?
-        tile_lay_abilities(@company)&.blocks
+        false
```
tobymao, line 20: is this gonna break stuff?

lib/engine/ability/base.rb
```diff
@@ -10,7 +10,7 @@ class Base
       include Ownable
 
       attr_accessor :count_this_or, :description
-      attr_reader :type, :owner_type, :remove, :when, :count, :count_per_or, :start_count
+      attr_reader :type, :owner_type, :remove, :when, :count, :count_per_or, :start_count, :show_count
```
tobymao, line 13: why do we need to add show_count? what was wrong with the original

crericha: I want to use start_count in other places and it was dependent on show_count being true. This allows me to break that dependency without affecting the show_count logic.

crericha: Discussed further in slack. I think we are on the same page.

lib/engine/step/special_track.rb
crericha: I don't think that line actually worked lol. Traced the line back to a commit to a fix 1846 special tile lays. I tested the 1846 special tile lay privates by hand and it all worked great. Cases that should fail, failed. Cases that should work, worked.

michaeljb: Just found that this did break stuff for 18Chesapeake. C-P and B&S are supposed to block after their first tile is laid, and now they don't. The abilities don't set `blocks` explicitly, but it defaults to `true` in `engine/ability/tile_lay.rb`

PR #2443, opened 2020-11-25
1856 companies may close

ventusignis:
Based off of [1856 tile upgrades](https://github.com/tobymao/18xx/pull/2439)

I saw some behavior when testing tile upgrades that corporations didn't close but I was able to get corporations to close in a new game by selling shares and by withholding. 

This implements the dynamic certificate limit, with stubs related to nationalization which still needs to be implemented

### DIFF
--- a/lib/engine/config/game/g_1856.rb
+++ b/lib/engine/config/game/g_1856.rb
@@ -523,7 +523,7 @@ module Engine
       "num": 2,
       "events":[
       {
-         "type":"Nationalization"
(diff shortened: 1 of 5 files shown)


tobymao:
why is this pr based off the other one? which one do you want me to review? if only this one is valid close the other one, but i've already reviewed that one...

ventusignis:
I put this one up so that after the other one gets merged in that this one is a small PR to review

PR #2446, opened 2020-11-25
Refactor BuySellShares view and add [18CO] DSNG Share Distribution

ryandriskel:

### Background

This started as a simple task to add the unusual "shares" distribution for DSNG. That was a one liner, but in the process I discovered that the view file did not respect different sizes of shares in the treasury, only the market. I found doing so following the current pattern would create a lot of duplicated code for creating the buttons as well determining percentages and text. This also gave me the chance to write a lot of the code using fail first methodology as well as map functions.

The changes will also reduce the complexity needed for creating the buy buttons in https://github.com/tobymao/18xx/pull/2431 as the same Button view file can be reused, passing a different action.

### Implementation

A change I've made is to always show shares in descending percentage. Currently, they were always showing in the order defined in the JSON. This meant a 10% might be first, then a 20%. Once the 10% was purchased, the 20% would show first. Now, they are always in the same order.

Additionally, I've standardized the calculations of share percentage between buy and exchange. They no longer use similar but different code, which I believe was an oversight. All buttons will show the percentage if there are multiple types available or if the share has a different percentage than the "normal" percentage for the corporation, as defined by the second share (not new) in the configuration.

### Status

I have verified the normal use cases of Buy from IPO/Treasury and Market. I still need to verify things don't explode on:

- [x] Price Protection
- [x] Share Exchange
- [x] Reduce Price Shares
- [x] Shorts

Screen Shot 2020-11-25 at 2 13 46 PM

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -82,11 +82,9 @@ module View
 
         return [] unless price_protection
 
-        protect = lambda do
-          process_action(Engine::Action::BuyShares.new(@current_entity, shares: price_protection.shares))
-        end
+        protect = -> { process_action(Engine::Action::BuyShares.new(@current_entity, shares: price_protection.shares)) }
 
-        [h(:button, { on: { click: -> { protect } } }, 'Protect Shares')]
+        [h(:button, { on: { click: protect } }, 'Protect Shares')]
       end
 
       def render_reduced_price_shares(shares, source: 'Market')
@@ -106,9 +104,7 @@ module View
         return [] unless @step.current_actions.include?('short')
         return [] unless @step.can_short?(@current_entity, @corporation)
 
-        short = lambda do
-          process_action(Engine::Action::Short.new(@current_entity, corporation: @corporation))
-        end
+        short = -> { process_action(Engine::Action::Short.new(@current_entity, corporation: @corporation)) }
 
         [h(:button, { on: { click: short } }, 'Short Share')]
       end
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -3,6 +3,7 @@
 require 'view/game/actionable'
 require 'view/game/corporation'
 require 'view/game/sell_shares'
+require 'view/game/button/buy_share'
 
 module View
   module Game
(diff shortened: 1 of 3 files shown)


ryandriskel:
Reduced Price Shares OK

Screen Shot 2020-11-25 at 2 31 45 PM

ryandriskel:
Turns out Share Exchange for 'any' wasn't working.

Before Refactor

Screen Shot 2020-11-25 at 2 54 30 PM

After Refactor

Screen Shot 2020-11-25 at 2 53 29 PM

ryandriskel:
Shorts look OK

Screen Shot 2020-11-25 at 3 23 08 PM

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -12,114 +13,138 @@ class BuySellShares < Snabberb::Component
...
       def render
-        step = @game.round.active_step
-        current_entity = step.current_entity
+        @step = @game.round.active_step
+        @current_entity = @step.current_entity
 
-        ipo_share = @corporation.shares[0]
-        pool_shares = @game.share_pool.shares_by_corporation[@corporation].group_by(&:percent).values.map(&:first)
+        @ipo_shares = @corporation.shares.group_by(&:percent).values
+          .map(&:first).sort { |a, b| b.percent <=> a.percent }
```
tobymao, line 20: are you just trying to do a reverse or something?

sort.reverse

or sort_by { |x| -x.percent }

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -12,114 +13,138 @@ class BuySellShares < Snabberb::Component
...
+
+        @pool_shares = @game.share_pool.shares_by_corporation[@corporation].group_by(&:percent).values
+          .map(&:first).sort { |a, b| b.percent <=> a.percent }
```
tobymao, line 23: same

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -12,114 +13,138 @@ class BuySellShares < Snabberb::Component
...
 
-            puts("pool_share = #{pool_share.class} with price #{pool_share.price}")
-            reduced_price = @game.format_currency(pool_share.price - swap_share.price)
-            children << h(
-              :button,
-              { on: { click: -> { buy_share(current_entity, pool_share, swap: swap_share) } } },
-              "Buy #{pool_share.percent}% Market Share (#{reduced_price} + #{swap_share.percent}% Share)",
-            )
-          end
+        children.compact
```
tobymao, line 47: do you need to call compact if these just all return arrays?

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -12,114 +13,138 @@ class BuySellShares < Snabberb::Component
...
-            { on: { click: short } },
-            'Short Share',
-          )
+          h(Button::BuyShare,
+            share: share,
+            swap_share: swap_share,
+            entity: @current_entity,
+            percentages_available: shares.size,
+            source: source)
         end
```
tobymao, line 102: this on eis missing a compact

assets/app/view/game/buy_sell_shares.rb
ryandriskel: Just reverse. Will use the cleaner code.

assets/app/view/game/buy_sell_shares.rb
ryandriskel: Good point, might as well just do a single compact at the end.

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -12,114 +13,138 @@ class BuySellShares < Snabberb::Component
...
-                            "#{prefix} an #{@game.class::IPO_NAME} share")
-            end
+        return [] unless price_protection
+
+        protect = lambda do
+          process_action(Engine::Action::BuyShares.new(@current_entity, shares: price_protection.shares))
+        end
 
-            next unless ability.from.include?(:market)
+        [h(:button, { on: { click: -> { protect } } }, 'Protect Shares')]
```
daniel-sousa-me, line 89: I checked the price_protection as you asked and it's broken, but only because you have a lambda inside another lambda on the click. To keep the style similar to the rest, I suggest you put:

`protect = -> { process_action(Engine::Action::BuyShares.new(@current_entity, shares: price_protection.shares)) }

[h(:button, { on: { click: protect } }, 'Protect Shares')]`

ryandriskel: Oh wow, I didn't notice the -> at all. I thought I was following what was done for shorts but a couple extra characters makes a difference!

ryandriskel: Alright, both lambdas / buttons use the same syntax now. I verified Short Sale still works.

tobymao (approved):

PR #2448, opened 2020-11-26
Add 1817NA by inheriting from 1817

ventusignis:
1817NA has 4 loans per row instead of 1817's 5 loans per row. I removed 5 corporations arbitrarily to match 1817NA's rules.

I slightly altered Track in 1817 so that it gets the private name & pittsburg hex from the hex instead of being hardcoded

1817NA is pre-alpha in this version of the pull request.


ventusignis:
See #2450

PR #2450, opened 2020-11-26
Add 1817na

ventusignis:
I fixed the weird rebase issues in this branch

Adds 1817 by inheriting from 1817
Changes 1817 Track to use class constant to determine Pittsburgh private name & location
Overrides loan logic to implement 1817NA's 4 loans per row

Game is marked as pre-alpha


lib/engine/game/g_1817_na.rb
```diff
@@ -0,0 +1,57 @@
...
+      }.freeze
+      GAME_DESIGNER = 'Mark Voyer'
+
+      def self.title
+        '1817NA'
+      end
+
+      # Not genericifying 1817's loan logic just so it can be kept simpler, at least for now
+      def init_loans
+        @loan_value = 100
```
tobymao, line 30: why do you need this to be an instance variable?

ventusignis: The implementation of 1817 has it as an instance variable and to maximize reuse either g_1817.rb needed to be refactored or the behavior copied

For the case of loans, I chose to copy the behavior because the logic was complicated enough that I didn't want to meddle in it if I could avoid it

In the case of the Pittsburgh private, the logic was simple enough I was comfortable refactoring it to use constants which `G1817NA` overrides

tobymao (approved):

tobymao:
when will you release the alpha?

ventusignis:
> when will you release the alpha?

What is the normal schedule for releasing games on the site? 

This is rules complete. The only thing left I would like to do with this is find a way to have all 20 1817 corporations available to start, but limit the game to only being able to have 15 corporations in play at at time. Currently 5 arbitrarily chosen corporations are removed out of the 20 from 1817.

tobymao:
You mean you want the user to choose which 5 are removed but currently 5
random ones are removed? I don't see why it matters which ones are removed.

PR #2460, opened 2020-11-27
18NA: Make it alpha, add all 20 corporations, make it so that only 15…

ventusignis:
… out of 20 corporations can be IPOed

### DIFF
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -68,6 +68,12 @@ module View
             h('td.right', @game.format_currency(@game.round.active_step.seed_money)),
           ])
         end
+        if (max_corps_ipoed = @game.max_corps_ipoed)
+          trs << h(:tr, [
+            h(:td, 'IPOed Corporations'),
+            h('td.right', "#{@game.num_corps_ipoed}/#{max_corps_ipoed}"),
+          ])
+        end
 
         return unless trs.any?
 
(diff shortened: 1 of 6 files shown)


daniel-sousa-me:
Having 20 corps on the game page is a bit annoying. I understand why it's "needed" in 1817, but here I'd suggest you just pick 5 to take out. Or if you don't want to decide that yourself, have 5 randomly removed during setup (1846 does this, so the code should be trivial)

ventusignis:
> 
> 
> Having 20 corps on the game page is a bit annoying. I understand why it's "needed" in 1817, but here I'd suggest you just pick 5 to take out. Or if you don't want to decide that yourself, have 5 randomly removed during setup (1846 does this, so the code should be trivial)

Originally this was implemented by removing the first 5 corporations of 1817 alphabetically.

The reason I put the effort to add the other corporations back in was that I believe that the benefits of allowing the players to actively choose which corporations are used outweigh the downsides of having 20 corporations on the game page.

daniel-sousa-me:
Well, it's your call.

Maybe the page could show just one placeholder corp and the winner of the auction could pick the corp they want instead of it being picked by who starts the auction (just an idea for the future)

tobymao:
just keep it as is was, keep only 15 corps in the config and remove the other code so it's simpler

ventusignis:
That's basically all this patch did so I'll close it and submit a new one to just make the game alpha

PR #2463, opened 2020-11-27
1828: Allow the game class's cache to be cleared

crericha:
- Add function to clear cache of the specified type
- Use to clear share_prices cache when new par prices become unlocked.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1641,6 +1641,8 @@ module Engine
       end
 
       def clear_cache(type)
+        return unless CACHABLE.any? {|t,_n| t == type }
+
         ivar = "@_#{type}"
         instance_variable_set(ivar, send(type).map { |x| [x.id, x] }.to_h)
       end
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -1640,6 +1640,13 @@ def cache_objects
         end
       end
 
+      def clear_cache(type)
+        return unless CACHABLE.any? { |t, _n| t == type }
```
tobymao, line 1644: i'm a bit wary of this function? what's the intention?

crericha: Short answer: to clear the cached par prices. 
Longer answer: In 1828, more par prices become available throughout the game and if the cache isn't updated with the new par prices, then share_price_by_id won't be able to get at them. The game crashes on the par action if the screen is refreshed or undo used.

Maybe update_cache is a better name that clear_cache?

crericha: Essentially, the cache implies all of those objects are immutable. That's not true for all games.

tobymao: ok change to update

crericha: Change made.

tobymao (approved):

PR #2469, opened 2020-11-28
Fix abilities with multiple track lays that need to be laid at the same time

crericha:
New lay_tile param called "must_lay_together". The special_track step will block, after the first tile lay, if must_lay_together is true.

*** May need current games to be pinned. Recommend testing against the production db to verify ***

### DIFF
--- a/lib/engine/ability/tile_lay.rb
+++ b/lib/engine/ability/tile_lay.rb
@@ -5,10 +5,10 @@ require_relative 'base'
 module Engine
   module Ability
     class TileLay < Base
-      attr_reader :hexes, :tiles, :free, :discount, :special, :connect, :blocks, :reachable, :cost
+      attr_reader :hexes, :tiles, :free, :discount, :special, :connect, :blocks, :reachable, :must_lay_together, :cost
 
       def setup(tiles:, hexes: nil, free: false, discount: nil, special: nil,
-                connect: nil, blocks: nil, reachable: nil, cost: 0)
+                connect: nil, blocks: nil, reachable: nil, must_lay_together: false, cost: 0)
         @hexes = hexes
         @tiles = tiles
         @free = free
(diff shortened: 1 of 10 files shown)


crericha:
@michaeljb - here is the PR for the 18Chesapeake tile lay issue

spec/fixtures/1846/12666.json
```diff
@@ -1201,8 +1201,8 @@
     },
     {
       "type": "lay_tile",
-      "entity": "GT",
-      "entity_type": "corporation",
+      "entity": "MC",
+      "entity_type": "company",
```
crericha, line 1205: This is how current game will fail. Pass done in the track step by the corporation, rather than special_track.

lib/engine/spender.rb
```diff
@@ -7,6 +7,7 @@ module Spender
     attr_accessor :cash
 
     def check_cash(amount)
+      amount.bad_func if (@cash - amount).negative?
```
tobymao, line 10: bad_func ????

lib/engine/ability/tile_lay.rb
```diff
@@ -5,10 +5,10 @@
 module Engine
   module Ability
     class TileLay < Base
-      attr_reader :hexes, :tiles, :free, :discount, :special, :connect, :blocks, :reachable, :cost
+      attr_reader :hexes, :tiles, :free, :discount, :special, :connect, :blocks, :reachable, :must_lay_together, :cost
 
       def setup(tiles:, hexes: nil, free: false, discount: nil, special: nil,
-                connect: nil, blocks: nil, reachable: nil, cost: 0)
+                connect: nil, blocks: nil, reachable: nil, must_lay_together: false, cost: 0)
```
tobymao, line 11: should this default to true? what are the inverses?

lib/engine/spender.rb
crericha: Debug code because the game error wouldn't tell me which action was causing the issue for the above json. Needs to be removed!

lib/engine/ability/tile_lay.rb
crericha: My preference is to default everything to off, so no surprises as others add new games. What do you think?

PR #2471, opened 2020-11-28
1889: Set @company to nil after the track has been laid. Fixes bad interactions after using Ehime.

crericha:


michaeljb:
What sort of bad interactions?

tobymao (approved):

PR #2481, opened 2020-11-28
1828: Fix hex and tile labels for NYC, Boston, and Baltimore

crericha:


lib/engine/game/g_1828.rb
```diff
@@ -122,6 +122,17 @@ def init_stock_market
         sm
       end
 
+      def init_tiles
+        tiles = super
+
+        tiles.find { |tile| tile.name == '53' }.label = 'Ba'
```
tobymao, line 128: is there only one tile of this type? if not you could do

crericha: Yes, only one of each.

PR #2482, opened 2020-11-29
1828: Implement D&H teleport ability

crericha:


lib/engine/ability/README.md
```diff
@@ -130,6 +130,8 @@ Lay a tile and place a station token without connectivity
   teleport destination.
 - `tiles`: An array of tile numbers which may be placed at the
   teleport destination.
+- `cost`: Cost to use the telport ability.
```
michaeljb, line 133: typo

lib/engine/step/tracker.rb
```diff
@@ -91,9 +91,20 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
...
+          next if !ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name)
+
+          teleport = true
+          free = true if ability.free_tile_lay
+          if ability.cost
+            spender.spend(ability.cost, @game.bank) if ability.cost.positive?
+            @log << "#{spender.name} spends #{@game.format_currency(ability.cost)} and teleports to #{hex.name}"
+          end
+
+          ability.use!
```
michaeljb, line 103: `process_lay_tile` in `special_track.rb` already calls `ability.use!`. Would it be possible for `lay_tile` to take teleport and cost as parameters, and get rid of the `.abilities(:teleport)` call here?

https://github.com/crericha/18xx/blob/fbc83f1fa66e894c5c883bc643ced201dbaf0e49/lib/engine/step/special_track.rb#L23-L28

lib/engine/ability/README.md
crericha: Fixed.

lib/engine/step/tracker.rb
crericha: I didn't add ability.use!. It was already there. (line 94)

There are other games that allow teleport without a tile lay (1849 comes to mind), so I think the teleport ability is still needed?

michaeljb: > I didn't add ability.use!. It was already there (line 94). Is it something that should be removed?

Oh sorry, I missed that! Yeah looks like we need to leave it

lib/engine/ability/README.md
```diff
@@ -130,6 +130,8 @@ Lay a tile and place a station token without connectivity
   teleport destination.
 - `tiles`: An array of tile numbers which may be placed at the
   teleport destination.
+- `cost`: Cost to use the teleport ability.
+- `fee_tile_lay`: If true, the tile is laid with 0 cost. Default false.
```
tobymao, line 134: free?

lib/engine/ability/teleport.rb
```diff
@@ -5,11 +5,13 @@
...
     class Teleport < Base
-      attr_reader :hexes, :tiles
+      attr_reader :hexes, :tiles, :cost, :free_tile_lay
 
-      def setup(hexes:, tiles:)
+      def setup(hexes:, tiles:, cost: nil, free_tile_lay: false)
         @hexes = hexes
         @tiles = tiles
+        @cost = cost
+        @free_tile_lay = free_tile_lay
```
tobymao, line 14: safer to do free_tile_lay: nil

free_tile_lay = free_tile_lay || false

tobymao (approved):

PR #2485, opened 2020-11-29
Make it possible to trigger event last during train buy

perwestling:
Some events require the trains to be rusted before they can trigger,
so by setting 'when' to 'post' for the event, the will be done last,
post. If not set, or set to another value, it will be done first,
before the rusting (which is the current behavior).


perwestling:
I tried with 'when' but it did not work.

perwestling:
This caused problems (as phase change occurs during train buy, which means the train buyer might not yet have bought any train) so I will use a different solution.

PR #2486, opened 2020-11-29
Add 1817WO; Nieuw Zeeland rule is not implemented

ventusignis:
Marc Voyer sent me the 18xx-maker JSON for 1817WO so I put together the base for it as a pre-alpha title. 
The rule for Nieuw Zeeland is not implemented here and would need to be implemented before it can be taken alpha.

### DIFF
--- a/lib/engine/config/game/g_1817_wo.rb
+++ b/lib/engine/config/game/g_1817_wo.rb
@@ -1,0 +1,679 @@
+# frozen_string_literal: true
+
+# File original exported from 18xx-maker/export-rb
+# https://github.com/18xx-maker/export-rb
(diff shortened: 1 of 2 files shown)


daniel-sousa-me:
I think that the interaction between near and the coal mine also needs to be implemented

ventusignis:
@daniel-sousa-me What do you mean by "near"?

daniel-sousa-me:
if NYC has been upgraded to the south, you can't place a coal in E4. I thought the code was written in a less generic way, but I jus checked and it should be accounting for that. So disregard my comment

ventusignis:
I did encounter an issue with the generic coal mine logic in 1817NA but I submitted a PR for that which is already merged; the coal near NYC won't be an issue

tobymao (approved):

PR #2490, opened 2020-11-29
1860 Receivership

roseundy:
- Added ability to sell President's certificate to market and to sell a partial share with a swap
- Added ability to buy partial share of President's certificate in market with a swap
- No track laying or tokening for Receivership companies
- Automatic withholding for Receivership companies
- No train purchase for Receivership companies if they already have trains

I had to modify one common file to allow the purchase of a partial President's certificate from the market. This is because share purchases are not bundled and I need to modify the Pres share to be a partial share. The only place this was a pain was in the UI because it wants to show the full 20% as a purchase option:

- UI::Button::BuyShare - suppress displaying percentage on 'buy' button for president's share

### DIFF
--- a/lib/engine/g_1860/share_pool.rb
+++ b/lib/engine/g_1860/share_pool.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+require_relative '../share_pool'
+
+module Engine
+  module G1860
+    class SharePool < SharePool
+      def buy_shares(entity, shares, exchange: nil, exchange_price: nil, swap: nil)
+        bundle = shares.is_a?(ShareBundle) ? shares : ShareBundle.new(shares)
+        if bundle.presidents_share && bundle.owner == self
+          puts 'Detected purchase of presidents share from market'
+          bundle = ShareBundle.new(bundle.shares, bundle.corporation.share_percent)
+        end
+
+        super(entity, bundle, exchange: exchange, exchange_price: exchange_price, swap: swap)
+      end
+
+      def transfer_shares(bundle, to_entity,
+                          spender: nil,
+                          receiver: nil,
+                          price: nil,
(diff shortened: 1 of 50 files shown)


lib/engine/g_1860/share_pool.rb
```diff
@@ -0,0 +1,111 @@
...
+      def buy_shares(entity, shares, exchange: nil, exchange_price: nil, swap: nil)
+        bundle = shares.is_a?(ShareBundle) ? shares : ShareBundle.new(shares)
+        if bundle.presidents_share && bundle.owner == self
+          bundle = ShareBundle.new(bundle.shares, bundle.corporation.share_percent)
+        end
+
+        super(entity, bundle, exchange: exchange, exchange_price: exchange_price, swap: swap)
+      end
+
+      def transfer_shares(bundle, to_entity,
```
tobymao, line 17: how much of this has changed? what did you need to do

lib/engine/game/base.rb
```diff
@@ -659,6 +659,10 @@ def redeemable_shares(_entity)
         []
       end
 
+      def allow_percentage_on_buy(_share)
```
tobymao, line 662: what does this mean?

lib/engine/g_1860/share_pool.rb
roseundy: The original transfer_shares doesn't deal with selling or buying president's shares to/from the pool.

lib/engine/game/base.rb
roseundy: 1860 needs a way to suppress printing out "20%" on the buy from market button for the president's share. No other game needs that. This was the least intrusive way I could think of doing that.

tobymao (approved):

PR #2497, opened 2020-11-30
Use city as centre for TokenSelector, instead of the centre of the tile

daniel-sousa-me:
I think this only affects Pile o' Bones in 1882, but looks nicer

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -111,8 +111,8 @@ module View
               process_action(action)
             else
               coords = Hex.coordinates(@tile.hex)
-              coords[0] += @city_render_location[:x]
-              coords[1] += @city_render_location[:y]
+              coords[0] += @city_render_location[:x] if @city_render_location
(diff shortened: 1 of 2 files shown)


tobymao:
show a picture

jenf:
also does this have any issue with tiles on the edges of the board?

daniel-sousa-me:
It is not checking for the edges, but this commit doesn't make that any better or any worse. I thought about also addressing that, but in the end it's very hard to imagine a case where that would be relevant, since the tokens never appear very far from the hex

daniel-sousa-me:
Old:

image

New:

image

assets/app/view/game/part/city_slot.rb
```diff
@@ -22,6 +22,7 @@ class CitySlot < Base
         needs :tile_selector, default: nil, store: true
         needs :reservation, default: nil
         needs :game, default: nil, store: true
+        needs :city_render_location, default: nil
```
tobymao, line 25: how can the default be nil if on line 114 you expect it to always exist

tobymao (approved):

PR #2500, opened 2020-11-30
Fix unable to lay coal in hex B8 from a Brown Dawson City

ventusignis:


jenf:
Please provide a sample game so I can validate.

ventusignis:
Here is a clone of the game I had yesterday where we hit it
https://gist.github.com/ventusignis/92a87978cddd74af42595bafae90bf0f

ventusignis:
`.edges()` returns the edges of the tile, as it is in the tile manifest while `.exits()` returns the rotated exits; This never came up in 1817 because all of the K cities near mountains face up and the other cities are spiders or Boston (which lucks out?)

jenf:
code looks just running to check no regressions on existing 1817 games now

jenf (approved):

PR #2510, opened 2020-12-01
Stock Market View Page

ryandriskel:
I found it useful to have a view directly into the stock market while building 18CO. This allows a user to view the market for a specific game without bring in the game. (eg: /market/18AL )

http://0.0.0.0:9292/market/1846

Screen Shot 2020-12-01 at 7 17 56 AM

http://0.0.0.0:9292/market/18CO

Screen Shot 2020-12-01 at 7 18 10 AM

### DIFF
--- a/assets/app/view/market_page.rb
+++ b/assets/app/view/market_page.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'view/game/stock_market'
+
+module View
+  class MarketPage < Snabberb::Component
+    needs :route
+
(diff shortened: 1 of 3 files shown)


assets/app/view/market_page.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require_tree 'engine'
```
tobymao, line 3: i don't think it's necessary to require

ryandriskel: Removed.

tobymao (approved):

tobymao:
please add a unit test for this later in spec/assets

PR #2514, opened 2020-12-01
1856 Dynamic Certificate Limit

ventusignis:
Add 1856's dynamic certificate limit; the certificate limit is static until nationalization but after nationalization it changes whenever corporations close.

I put in some stubs to implement later once Nationalization is implemented

### DIFF
--- a/lib/engine/game/g_1856.rb
+++ b/lib/engine/game/g_1856.rb
@@ -82,6 +82,34 @@ module Engine
         # trade all shares
       end
 
+      def post_nationalization
+        # TODO: Update this with something more correct once nationalization is implemented
+        true
+      end
+
+      def num_corporations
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1856.rb
```diff
@@ -82,6 +82,34 @@ def event_nationalization!
...
+        true
+      end
+
+      def num_corporations
+        # TODO: Update this with something more correct once nationalization is implemented
+        @corporations.size - 1
+      end
+
+      def cert_limit
+        return { 3 => 20, 4 => 16, 5 => 13, 6 => 11 }[@players.size] unless post_nationalization
```
tobymao, line 96: make these constantns

tobymao (approved):

PR #2518, opened 2020-12-01
Add ability to see last run on map at nearly any time*

### LINKED ISSUE #1611: Seeing other company's routes

jenf:
* not during run routes/dividends

fixes #1611 & #563

### DIFF
--- a/assets/app/view/game/map_controls.rb
+++ b/assets/app/view/game/map_controls.rb
@@ -107,11 +107,11 @@ module View
           end
         end
 
-        @route_input = render_select('Show Route', id: :route, on: { input: route_change }, children: operators)
-        ['Last Route:', @route_input]
+        @route_input = render_select(id: :route, on: { input: route_change }, children: operators)
+        ['Show Last Route For:', @route_input]
       end
 
-      def render_select(_label, id:, on: {}, children: [])
+      def render_select(id:, on: {}, children: [])
         input_props = {
           attrs: {
             id: id,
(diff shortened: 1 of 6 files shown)


benjaminxscott:
Would `Show Last Routes For:` be a good descriptive label? 

Then it's more clear the operation that button refers to 
- "Show", similar to the other toggles
- "Last Route(s)", to make it clear this shows all routes
- "For", to make clear that this selector is per-company (and being a bit more clear in the "None" case)

benjaminxscott:
Thanks for addressing out my suggestion - to clarify, I meant it as a pure suggestion rather than a 'review', fwiw

assets/app/view/game/map_controls.rb
```diff
@@ -55,6 +58,69 @@ def starting_map_controls
...
 
+      def generate_last_route(entity)
+        trains = entity.trains
+        operating = entity.operating_history
+        last_run = operating[operating.keys.max]&.routes
+        return [] unless last_run
+
+        halts = operating[operating.keys.max]&.halts
+        routes = []
+        last_run.map do |train, connections|
```
tobymao, line 69: why is this a map compact if you don't use the output?

roseundy:
I suspect this might not play nice with the 1860 leased trains - not that you should worry about that.

assets/app/view/game/map_controls.rb
```diff
@@ -55,6 +58,69 @@ def starting_map_controls
...
+          next unless trains.include?(train)
```
roseundy, line 70: Do you really want this? Doesn't this mean that a route won't be shown if the corporation no longer owns the train with which the route was run?

PR #2519, opened 2020-12-01
Add player value over rounds to the spreadsheet

### LINKED ISSUE #2201: Provide table of player and company history
Currently there is a table of revenue for each public company by round in the "OR History" on the "Spreadsheet" tab. This is a great feature for tracking the progress of the game (e.g. for making plots or performing an analysis), but it would be even greater if a similar "history" was provided for other player and company information:

- Player:
    - Cash
    - Value
    - Liquidity
    - Shares
    - Certs
- Company:
    - Market Price
    - Number of each kind of train
   - Player held shares

Currently the only way to get at this information is to run the game and copy out this information after each OR (or whatever), but it would be great to have this as part of the "Spreadsheet", suitable for simply copying and pasting into a Goole Sheet or Excel, for use in analysis or creating plots.

jenf:
fixes #2201

### DIFF
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -52,6 +52,10 @@ module View
             render_player_companies,
             render_player_certs,
           ]),
+          h(:thead, [
+            h(:tr, { style: { height: '1rem' } }, ''),
+          ]),
+          *render_player_history,
         ])
         # TODO: consider adding OR information (could do both corporation OR revenue and player change in value)
         # TODO: consider adding train availability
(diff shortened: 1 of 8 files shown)


assets/app/view/game/spreadsheet.rb
```diff
@@ -65,6 +69,25 @@ def render_history_titles(corporations)
         or_history(corporations).map { |turn, round| h(:th, @game.or_description_short(turn, round)) }
       end
 
+      def render_player_history
+        # OR history should exist in all
+        zebra_row = true
+        last_values = nil
+        @game.players.first.history.map do |h|
```
tobymao, line 76: cleaner to do map.with_index 

index.even? 
then manual zebra rowing

jenf: it's not exactly the same, as the values aren't shown if they are identical

PR #2528, opened 2020-12-03
Enhancement: log company sym when its ability is used for tokens/tiles

michaeljb:
e.g., "ELA (DC&H) places a token" instead of "ELA places a token" to indicate
that ELA used the ability from the company Dewey, Cheatham, and Howe to place
that token

### DIFF
--- a/lib/engine/step/tokener.rb
+++ b/lib/engine/step/tokener.rb
@@ -37,7 +37,12 @@ module Engine
         @game.game_error('Token is already used') if token.used
 
         token, ability = adjust_token_price_ability!(entity, token, hex, city)
-        entity.remove_ability(ability) if ability
+        tokener = entity.name
+        if ability
+          tokener = "#{entity.name} (#{ability.owner.sym})" if ability.owner != entity
+          entity.remove_ability(ability)
+        end
+
         free = !token.price.positive?
         city.place_token(entity, token, free: free, cheater: special_ability&.cheater)
         unless free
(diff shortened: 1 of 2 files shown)


lib/engine/step/tokener.rb
```diff
@@ -49,9 +55,9 @@ def place_token(entity, city, token, teleport: false, special_ability: nil)
         when :neutral
           entity.tokens.delete(token)
           token.corporation.tokens << token
-          @log << "#{entity.name} places a neutral token on #{hex.name}#{price_log}"
+          @log << "#{tokener} places a neutral token on #{hex.name}#{price_log}"
```
tobymao, line 58: won't tokener be nil?

michaeljb: tokener gets set in either branch of the if statement at line 40. I could set it to “entity.name” before that block if that would be more clear.

tobymao: it doesn't every time

 tokener = "#{entity.name} (#{ability.owner.sym})" if ability.owner != entity

lib/engine/step/tokener.rb
```diff
@@ -37,7 +37,13 @@ def place_token(entity, city, token, teleport: false, special_ability: nil)
         @game.game_error('Token is already used') if token.used
 
         token, ability = adjust_token_price_ability!(entity, token, hex, city)
-        entity.remove_ability(ability) if ability
+        if ability
```
ryandriskel, line 40: How about instead:

```
tokener = entity.name
if ability
  tokener = "#{entity.name} (#{ability.owner.sym})" if ability.owner != entity
  entity.remove_ability(ability)
end
```

lib/engine/step/tokener.rb
ryandriskel: How about instead:
```
tokener = entity.name
if ability
  tokener = "#{entity.name} (#{ability.owner.sym})" if ability.owner != entity
  entity.remove_ability(ability)
end
```

michaeljb: Oh, right of course. Thanks!

PR #2534, opened 2020-12-03
Use buttons instead of links for the activated abilities

daniel-sousa-me:
Most new users get confused with the links. This should help

image

### DIFF
--- a/assets/app/view/game/abilities.rb
+++ b/assets/app/view/game/abilities.rb
@@ -37,7 +37,7 @@ module View
             on: { click: toggle_show },
           }
           children << h(:button, props, "#{@show_other_abilities ? 'Hide' : 'Show'} Others")
-          children.concat(render_companies(others)) if @show_other_abilities
+          children << h(:div, render_companies(others)) if @show_other_abilities
         end
 
         if companies.include?(@selected_company)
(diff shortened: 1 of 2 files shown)


jenf:
Code changes themselves looks good, but it's probably good to get a second opinion on the change to the UI.

michaeljb:
I am a fan of this change, however in my quick test, clicking on one company showed its text but did not make the button look different the way @daniel-sousa-me's screenshot does :thinking: 

Screenshot from 2020-12-04 10-22-55

daniel-sousa-me:
Which browser+OS?

daniel-sousa-me:
This is using the CSS. Maybe you need to do Ctrl+Alt+R?

michaeljb:
Ubuntu 18, Firefox 83

I do see the button element does have `class="active"`...

I also see in the CSS, the `button.active` definition is under `@media(hover: hover) and (pointer: fine) {`. I don't know exactly what `@media` does, good it be activating the class only when the element is actually being hovered?

michaeljb:
Rebuilding the container did the trick :slightly_smiling_face:

michaeljb (approved):

daniel-sousa-me:
I didn't notice the @media there. It is meant to prevent the :hover effects in devices where they don't make sense (like smartphones).

I'll update the CSS

PR #2535, opened 2020-12-04
pre-alpha impl of 1848

benjaminxscott:
Initial features re: #2548

Pulled most of this code from @eheiden's branch

Goal is to have a working baseline for people who want to implement this game's chrome/mechanics

Logos courtesy of @kelsin [at 18xx-maker](https://github.com/18xx-maker/18xx-maker/tree/master/src/data/logos/1848)

Open to a rebase / squash


assets/app/view/game/round/auction.rb
```diff
@@ -170,6 +170,14 @@ def render_input(company)
               ]
             end
 
+          if @step.may_reduce?(company)
+            company_actions << h(:button, { on: { click: lambda {
+                                                           process_action(
```
benjaminxscott, line 175: blame rubocop for this formatting

lib/engine/step/g_1848/dutch_auction.rb
```diff
@@ -0,0 +1,148 @@
...
+          company.discount += 5
+          price = company.min_bid
+          @log << "#{current_entity.name} reduces #{company.name} by £5 to #{@game.format_currency(price)}"
+          @round.next_entity_index!
+        end
+
+        def may_reduce?(company)
+          # Each private can be discounted a maximum of 6 times
+
+          if company.sym == 'P1' && company.min_bid.positive?
```
benjaminxscott, line 70: Probably a more generic way to handle this than hardcoding (e.g. "if discount > 30")
However, discount is already overloaded to handle the "buy-in price" differential, so I couldn't be bothered

lib/engine/step/g_1848/dutch_auction.rb
```diff
@@ -0,0 +1,148 @@
...
+          # P1 can be reduced to free, so we disable `check_positive`
+          player.spend(price, @game.bank, check_positive: false)
+          @companies.delete(company)
+          @log << "#{player.name} buys #{company.name} for #{@game.format_currency(price)}"
+
+          company.abilities(:share) do |ability|
+            share = ability.share
+
+            if share.president
+              # TODO: - I think this isn't necessary since we must par CAR to 100
```
benjaminxscott, line 138: I started implementing the "hand the player a free share" similar to 18Ches, but ran into some errors

perwestling:
I think we might need a check list for all the functionality of 1848, with those parts that have been implemented checked.
Create a new issue for that.

lib/engine/step/g_1848/dutch_auction.rb
jenf: or a {'P2'=>40} hash and just look it up rather than multiple lines

benjaminxscott: Good idea!

benjaminxscott:
@perwestling agree

lib/engine/step/g_1848/dutch_auction.rb
benjaminxscott: 410d102

lib/engine/action/reduce.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Engine
+  module Action
+    class Reduce < Base
+      attr_reader :company
+
+      def initialize(entity, company: nil)
+        @entity = entity
+        @company = company
+        raisenp 'Company cannot be nil' unless @company
```
tobymao, line 13: raisenp? just raise a GameError

lib/engine/action/reduce.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
tobymao, line 1: what is a reduce action

lib/engine/step/g_1848/dutch_auction.rb
```diff
@@ -0,0 +1,142 @@
...
+        # On your turn, must buy any private or reduce price of any private
+        # Once you own a private, you may pass
+        # if everything is reduced to minimum price, and nobody has bought a private, you must buy one
+        # if any privates are left and everyone passes in a row, owned privates pay and auction continues
+        # if someone has bought a private and everything is minimum, you can pass
+
+        ACTIONS = %w[bid reduce].freeze
+        ACTIONS_WITH_PASS = %w[bid reduce pass].freeze
+
+        MIN_PRICES = {
```
tobymao, line 23: why is this here? it's not the value of the company?

lib/engine/action/reduce.rb
tobymao: i feel like you can just use the assign action for this

lib/engine/step/g_1848/dutch_auction.rb
```diff
@@ -0,0 +1,142 @@
...
+          company.min_bid > MIN_PRICES[company.sym]
+        end
+
+        def actions(entity)
+          return [] if @companies.empty?
+          return [] unless entity.player?
+
+          actions = entity.player.companies.empty? ? ACTIONS : ACTIONS_WITH_PASS
+
+          entity == current_entity ? actions : []
```
tobymao, line 87: why don't you do this current_entity check on line 83

lib/engine/action/reduce.rb
benjaminxscott: It "reduces the cost of a private"

assets/app/view/game/round/auction.rb
tobymao: put the hash on a new line

h(:button, {
  on: ...
})

lib/engine/action/reduce.rb
tobymao: yea, let's use that

lib/engine/step/base.rb
```diff
@@ -36,6 +36,10 @@ def did_sell?(_corporation, _entity)
         false
       end
 
+      def may_reduce?(_company)
```
tobymao, line 39: don't put this here, since no game uses it cept 48, just have it called in the view the action exists

lib/engine/step/g_1848/dutch_auction.rb
```diff
@@ -0,0 +1,142 @@
...
+
+        def process_pass(action)
+          entity = action.entity
+          @log << "#{entity.name} passes"
+          entity.pass!
+          all_passed! if entities.all?(&:passed?)
+          @round.next_entity_index!
+        end
+
+        def process_reduce(action)
```
tobymao, line 67: you'll need to change this to process_assign

lib/engine/step/g_1848/dutch_auction.rb
```diff
@@ -0,0 +1,142 @@
...
+          @game.payout_companies
+          @game.or_set_finished
+
+          entities.each(&:unpass!)
+        end
+
+        def buy_company(player, company, price)
+          company.owner = player
+          player.companies << company
+          # P1 can be reduced to free, so we disable `check_positive`
```
tobymao, line 123: why don't you only do it when it's p1 then

lib/engine/step/g_1848/dutch_auction.rb
benjaminxscott: Privates in this game are a bit weird, for instance

P1 has a "value" of 40 (can be bought in by a corp for up to 40)

P1 has a "price" of 30 (can be purchased by a player for 30)

P1 has a "minimum price" of zero (can be discounted by players taking "reduce" actions for up to 30$ total)

Other privates are similar, and each one can only be discounted by 30$ total

The `discount` data member is overloaded to also handle the difference between "price" and "value", so we can't just check whether discount >= 30 and so we have this hardcoded hash

lib/engine/step/g_1848/dutch_auction.rb
benjaminxscott: I think this was copy/pasted from [the base `waterfall_auction`](https://github.com/tobymao/18xx/blob/master/lib/engine/step/waterfall_auction.rb#L64)

See 9b7e7b30

lib/engine/action/reduce.rb
benjaminxscott: 9b7e7b30

lib/engine/step/base.rb
benjaminxscott: 69c49307

lib/engine/step/g_1848/dutch_auction.rb
benjaminxscott: You mean company.sym == 'P1' ... spend(..., check_positive=false)
else: spend(...)
?

assets/app/view/game/round/auction.rb
benjaminxscott: 57ed230f

lib/engine/step/g_1848/dutch_auction.rb
tobymao: or spend(price, @game,bank, check_positive: company.sym != 'P1')

benjaminxscott: 780735bd

lib/engine/action/reduce.rb
benjaminxscott: bca96f79017673bb5f853eb2899a76f486a7d7b8

lib/engine/step/g_1848/dutch_auction.rb
benjaminxscott: bca96f79017673bb5f853eb2899a76f486a7d7b8

benjaminxscott:
I would def squashmerge this, since some of the early commits e.g. 411a056 were not relevant, and my later commit messages are not particularly meaningful

tobymao (approved):

PR #2545, opened 2020-12-04
Refactor: move abilities() method from Abilities module to Game::Base

michaeljb:
Changes:
* find an entity's abilities via `@game.abilities(entity, ...)` instead of `entity.abilities(...)`
  * on current master, there are only a few instances where the `abilities()` caller does not have access to a `@game`, and in those cases it happens to work fine to use `all_abilities`; d2fa7d2fa6c425da68436034ce1ec5cddd9fb1e0
* `corporation.abilities(...)` always returned an Array of abilities, but the new `@game.abilities(entity, ...)` matches the interface for companies; that is, if there are no abilities, `nil` is returned; if there is exactly one, that one ability is returned; if there are more than one, then the Array is returned (code that expected the Array has been updated in ab6492fcc6b36dac6338fdb4795c130adf6702cf)
* `Corporation` and `Minor` no longer need their own overrides for `abilities()`
* the private methods from the `Abilities` module that `abilities()` called have moved to private methods in `Game::Base`, with `ability_` prepended to their names

Putting `abilities()` in the `Game` class will allow for more precise control of the timing of when private companies can and cannot be used, as the code in `abilities()` will have access to the whole game state. This will be very helpful in addressing issues like the ones I've labeled with [ability timing](https://github.com/tobymao/18xx/issues?q=is%3Aopen+is%3Aissue+label%3A%22ability+timing%22), but I didn't want to pile any of those fixes on top of this change.

The tests are all passing (and a few test failures along the way did help me catch bugs in my refactor), and last night I validated against an entire DB backup and there were no issues.

-----
Probably easier to look at the diff one commit at a time instead of all at once; the earliest commits are the most interesting, the last one touches many files but has only very simple changes in it--basically swapping the order of arguments for `abilities()`.

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -129,7 +129,7 @@ module View
         children = []
 
         @game.companies.each do |company|
-          company.abilities(:exchange) do |ability|
+          @game.abilities(company, :exchange) do |ability|
             next if ability.corporation != @corporation.name && ability.corporation != 'any'
             next unless company.owner == @current_entity
 
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -65,7 +65,7 @@ module View
         step = @game.round.active_step
         @corporation = step.current_entity
         if @selected_company&.owner == @corporation
-          @ability = @selected_company&.abilities(:train_discount, time: 'train')
+          @ability = @game.abilities(@selected_company, :train_discount, time: 'train')
         end
 
         @depot = @game.depot
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -541,7 +541,7 @@ module View
       end
 
       def can_assign_corporation?
-        @selected_corporation && @selected_company&.abilities(:assign_corporation)
+        @selected_corporation && @game.abilities(@selected_company, :assign_corporation)
       end
     end
   end
--- a/assets/app/view/game/exchange.rb
+++ b/assets/app/view/game/exchange.rb
@@ -23,7 +23,7 @@ module View
       end
 
       def render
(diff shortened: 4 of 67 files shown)


ryandriskel:
Standardizing returns is good, but moving from `object.function` to `game.function(object...` has my OOP spidey senses tingling.

michaeljb:
@ryandriskel understandable. Standardizing the returns across corporations/companies wasn't the main goal, but doing so was simpler than continuing to have corporations always expect an array (or changing companies to always expect an array).

The bigger picture goal of this refactor is to enable expanding the `Abilities#right_time?` (moved in this PR to `Game::Base#ability_right_time?`) method to more precisely tell what "time" is it in the game, rather than depending on a string the caller passes in. I want to use that to be able to strictly allow abilities to be used when the rules of their games allow, and forbid it at other times. Having access to the full game state will be crucial for this, and `Company` and `Corporation` objects do not have access to that.

For instance, right now 18xx.games allows the C-P's tile lay ability in 18Chesapeake to be used during the turn of another corporation or during any step of the owning corporation's turn, but by the rules that ability should only be usable during the owning corporation's track step.

tobymao:
have you run this against the full db?

michaeljb:
I see that this will cause conflicts/API issues with several other outstanding PRs: #2568, #2574, and #2576. I'm happy to wait for those to be merged and then resolve conflicts, or if this gets merged first I can push fixes directly to those branches.

michaeljb:
Updated for the changes merged from #2574. Added TODO list to OP.

tobymao:
@michaeljb go ahead and merge this whenever you feel ready

michaeljb:
I've rebased with the latest master, running another validate with the latest db backup and planning to release later tonight.

michaeljb:
Real life got in the way, will try again tomorrow...

michaeljb:
Found 3 games that failed validation, but they are games which have already been removed from the live DB.

Issue #1649, opened 2020-09-11
Handle borders meant to hide an edge differently from others


roseundy:
All borders are drawn after all other tile elements are drawn. This should be changed so that borders that trying to hide the edge of the tile (type == nil) are drawn first. This would allow off-board with track to be rendered properly.

Without this change, if we try to hide edges we get:
![BorderIssue](https://user-images.githubusercontent.com/8494213/92953218-938e6380-f41e-11ea-8010-0992ec36921b.png)

So, we are forced to not use "border" at all, and we get the hex edges:
![Offboard_lines](https://user-images.githubusercontent.com/8494213/92953268-aef96e80-f41e-11ea-8a39-e230044d8c5c.png)

roseundy:
A bonus would be to have edge hiding border respect the transparency property when a map was drawn.

roseundy:
Looked into the transparency (opacity) problem. It's caused by the fact that the border extends into the neighboring hex. This has implications for the primary bug fix since drawing wide borders will take out neighboring track.

michaeljb:
Closed via #2551 (but I'm happy to re-open if you think that's warranted @roseundy)

PR #2551, opened 2020-12-05
Enhancement: "Nil" border rendering

michaeljb:
When a hex/tile has a border with no defined type, don't render a border element on top of it; instead, remove the "stroke" from the hex `<polygon>` and draw a partial outline with `<path>` elements. Drastically improves rendering of 18MEX's Mexico City double-tile, and somewhat improves rendering of paired offboards when opacity is low.

It's not perfect, there seems to still be either a slight seam or overlap that is visible, depending on the orientation of the two tiles being rendered. For instance, see the two different offboard pairs in the 1846 "after" screenshot; the lower one has a visible line that isn't present on the higher one

## Before

MC yellow before
MC green before
MC brown before
1846 low opacity before

## After

MC yellow after
MC green after
MC brown after
1846 low opacity after

### DIFF
--- a/assets/app/view/game/hex.rb
+++ b/assets/app/view/game/hex.rb
@@ -45,7 +45,7 @@ module View
           else
             @hex.tile
           end
-        children = [h(:polygon, attrs: { points: Lib::Hex::POINTS })]
+        children = hex_outline
         if @tile
           children << h(
             Tile,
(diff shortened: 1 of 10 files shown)


roseundy:
Um, see issue #1649.

michaeljb:
Whoops, I didn't think to look for an actual issue on this! Here are some screenshots with the borders added for 18MS:

Screenshot from 2020-12-05 10-58-08
Screenshot from 2020-12-05 10-58-13

michaeljb:
@roseundy I'll see if rendering the "nil" border in a different order/width than other borders (the solution you describe in #1649) can fix this

michaeljb:
I'm still getting a small gap in the track that should be connected across the two tiles, and the rendering is different based on whether I'm zoomed in or out--it seems the browser still wants the line from the `<polygon>` to be visible even with the `<path>` or `<line>` from the border drawn on top of it. These are with the border's stroke-width set to 2, only difference is whether I'm zoomed in or not:

stroke width 2 zoomed out
stroke width 2 zoomed in

assets/app/view/game/hex.rb
```diff
@@ -75,6 +75,24 @@ def render
...
 
+      def hex_outline
+        polygon_props = { attrs: { points: Lib::Hex::POINTS } }
+
+        invisible_edges = @tile.borders.select { |b| b.type.nil? }.map(&:edge) if @tile
+        if invisible_edges&.any?
+          polygon_props[:attrs][:stroke] = 'none'
+          shapes = [h(:polygon, polygon_props)]
+
+          (Engine::Tile::ALL_EDGES - invisible_edges).each do |edge|
```
tobymao, line 86: this can be a map

PR #2554, opened 2020-12-05
18CO Add Discarded Train Discount

ryandriskel:

Trains in the Market ( from discards ) are purchased at 50% face value.

Normal price of the 3 was 180. With discount its 90.
Screen Shot 2020-12-05 at 11 47 58 AM

### Implementation Notes

Since the final price is calculated in the view itself, I couldn't figure out a good way to make this 18CO specific. The calculation seemed straightforward enough to roll into the base, though.

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -162,6 +162,7 @@ module View
             price = variant[:price]
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             price = @ability&.discounted_price(train, price) || price
+            price = @game.apply_discard_discount(train, price)
 
             buy_train = lambda do
               process_action(Engine::Action::BuyTrain.new(
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -162,7 +162,7 @@ module View
             price = variant[:price]
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             price = @ability&.discounted_price(train, price) || price
-            price = @game.apply_discard_discount(train, price)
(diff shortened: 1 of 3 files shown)


assets/app/view/game/buy_trains.rb
```diff
@@ -162,6 +162,7 @@ def from_depot(depot_trains)
             price = variant[:price]
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             price = @ability&.discounted_price(train, price) || price
+            price = @game.apply_discard_discount(train, price)
```
ryandriskel, line 165: I could use `respond_to` for this instead of putting it in game base, but that felt even ickier

lib/engine/game/base.rb
```diff
@@ -763,6 +764,13 @@ def must_buy_train?(entity)
            (self.class::MUST_BUY_TRAIN == :route && @graph.route_info(entity)&.dig(:route_train_purchase)))
       end
 
+      def apply_discard_discount(train, price)
+        return price unless DISCARDED_TRAIN_DISCOUNT
+        return price unless @depot.discarded.include?(train)
+
+        (price * (100 - DISCARDED_TRAIN_DISCOUNT) / 100).ceil
```
tobymao, line 771: make it explicitly a float / 100.0

lib/engine/game/base.rb
```diff
@@ -763,6 +764,13 @@ def must_buy_train?(entity)
            (self.class::MUST_BUY_TRAIN == :route && @graph.route_info(entity)&.dig(:route_train_purchase)))
       end
 
+      def apply_discard_discount(train, price)
```
tobymao, line 767: call it discard_discount

Issue #2122, opened 2020-10-30
1817 abilities shown during M&A round


ChrisTheCat:
<img width="567" alt="Screen Shot 2020-10-30 at 1 57 42 PM" src="https://user-images.githubusercontent.com/8732435/97756488-2f9c2900-1ab8-11eb-8d7f-8c92c50c1d49.png">

These abilities should not be shown during the M&A round.

PedroRCSequeira:
They should also not be shown outside that company turn, right? for instance coals from another of your companies

ChrisTheCat:
Correct

PR #2556, opened 2020-12-05
18CO Moving Bid Auction

ryandriskel:
### Background

18CO does not use a waterfall style auction. All privates are available for bids starting at face value.

A bid on a Private Company may be moved to a different Private Company as long as the bid value is increased by 5, in increments of 5. A player may have multiple bids on the same Private Company (via moving a bid between privates), but when "Place Bid" happens on a company with multiple bids already, it always adds to the player's highest bid. All bids are considered committed cash.

The auction ends when all players pass consecutively.

### Implementation Notes

The `&& company.owner` addition fixes the problem of private abilities showing during the SR and will probably fix https://github.com/tobymao/18xx/issues/2122 as you shouldn't be able to activate an ability that's not owned by anyone.

In the view, the new code won't run unless the actions include "move_bid", which no other games implement, and thus the view cannot be broken for current games.

The handling for companies that are not purchased during the auction is done in a follow up PR.

### Impacts

When the first company is selected, Place Bid is shown instead of Purchase. This is NOT a generic change and is done through the already existing "Purchasable Companies" functionality.
Screen Shot 2020-12-05 at 10 19 21 AM

Once a player has a bid on a company, selecting a different company to place a bid on will give the option of moving that bid. The bid will be updated to the amount in the text box.
Screen Shot 2020-12-05 at 10 19 42 AM

Showing multiple bids on one company from the same player
Screen Shot 2020-12-05 at 10 54 21 AM

In this case, the user tried to move a higher bid from DNP ($50) to a $45 bid on GJGR, which is not allowed. To move a bid you must increase it by at least $5.
Screen Shot 2020-12-05 at 10 21 17 AM

Log samples from moving bids
Screen Shot 2020-12-05 at 10 42 12 AM

Once all players pass consecutively, all the auctions are immediately resolved with the companies going to the highest bidders.
Screen Shot 2020-12-05 at 10 42 48 AM

### DIFF
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -187,25 +187,25 @@ module View
           moveable_bids = @step.moveable_bids(@current_entity, company)
           return [] if moveable_bids.empty?
 
-          moveable_bids.map do |old_company, old_bids|
-            old_bids.map do |old_bid|
+          moveable_bids.flat_map do |from_company, from_bids|
+            from_bids.map do |from_bid|
               move_bid = lambda do
                 hide!
                 price = input.JS['elm'].JS['value'].to_i
                 process_action(Engine::Action::MoveBid.new(
                   @current_entity,
                   company: company,
-                  old_company: old_company,
+                  from_company: from_company,
                   price: price,
-                  old_price: old_bid.price,
+                  from_price: from_bid.price,
                 ))
                 store(:selected_company, nil, skip: true)
               end
 
               h(:button, { on: { click: move_bid } },
-                "Move #{old_company.sym} #{@game.format_currency(old_bid.price)} Bid")
+                "Move #{from_company.sym} #{@game.format_currency(from_bid.price)} Bid")
             end
-          end.flatten
+          end
         end
 
         def render_turn_bid
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -164,10 +164,12 @@ module View
                 ))
                 store(:selected_company, nil, skip: true)
               end
+
               [
                 input,
                 h(:button, { on: { click: create_bid } }, 'Place Bid'),
-              ]
+                *render_move_bid_inputs(company, input),
+              ].compact
             end
 
           if @step.may_reduce?(company)
@@ -177,6 +179,33 @@ module View
           end
 
           h(:div, { style: { textAlign: 'center', margin: '1rem' } }, company_actions)
+        end
+
+        def render_move_bid_inputs(company, input)
+          return [] unless @current_actions.include?('move_bid')
+
+          moveable_bids = @step.moveable_bids(@current_entity, company)
+          return [] if moveable_bids.empty?
+
+          moveable_bids.map do |old_company, old_bids|
+            old_bids.map do |old_bid|
+              move_bid = lambda do
+                hide!
+                price = input.JS['elm'].JS['value'].to_i
+                process_action(Engine::Action::MoveBid.new(
+                  @current_entity,
+                  company: company,
+                  old_company: old_company,
+                  price: price,
+                  old_price: old_bid.price,
+                ))
+                store(:selected_company, nil, skip: true)
+              end
+
+              h(:button, { on: { click: move_bid } },
+                "Move #{old_company.sym} #{@game.format_currency(old_bid.price)} Bid")
+            end
+          end.flatten
         end
 
         def render_turn_bid
--- a/lib/engine/step/g_18_co/moving_bid_auction.rb
+++ b/lib/engine/step/g_18_co/moving_bid_auction.rb
@@ -66,7 +66,8 @@ module Engine
           return unless company
 
           high_bid = highest_bid(company)
-          (high_bid ? high_bid.price + min_increment : company.min_bid)
+
+          high_bid ? high_bid.price + min_increment : company.min_bid
         end
 
         # can never purchase directly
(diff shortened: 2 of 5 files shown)


assets/app/view/game/round/auction.rb
```diff
@@ -179,6 +181,33 @@ def render_input(company)
           h(:div, { style: { textAlign: 'center', margin: '1rem' } }, company_actions)
         end
 
+        def render_move_bid_inputs(company, input)
+          return [] unless @current_actions.include?('move_bid')
+
+          moveable_bids = @step.moveable_bids(@current_entity, company)
+          return [] if moveable_bids.empty?
+
+          moveable_bids.map do |old_company, old_bids|
```
tobymao, line 190: flat_map

assets/app/view/game/round/auction.rb
```diff
@@ -179,6 +181,33 @@ def render_input(company)
...
+            old_bids.map do |old_bid|
+              move_bid = lambda do
+                hide!
+                price = input.JS['elm'].JS['value'].to_i
+                process_action(Engine::Action::MoveBid.new(
+                  @current_entity,
+                  company: company,
+                  old_company: old_company,
+                  price: price,
+                  old_price: old_bid.price,
```
tobymao, line 200: is it necessary to store the old price? isn't that inferred from the old_company?

what do you think about changing the name to be from_company, to_company, price

lib/engine/step/g_18_co/moving_bid_auction.rb
```diff
@@ -0,0 +1,215 @@
...
+        def auctioning
+          nil
+        end
+
+        # min bid is face value or $5 higher than previous bid
+        def min_bid(company)
+          return unless company
+
+          high_bid = highest_bid(company)
+          (high_bid ? high_bid.price + min_increment : company.min_bid)
```
tobymao, line 69: parens are unnecessary

assets/app/view/game/round/auction.rb
ryandriskel: It is required. A Player can have multiple bids on a single Company, passing the price associated is the key to which of those bids are being moved.

ryandriskel: Changing it to "from_company" and "to_company" seems good.

lib/engine/step/g_18_co/moving_bid_auction.rb
```diff
@@ -0,0 +1,215 @@
...
+
+        def committed_cash(player, _show_hidden = false)
+          bids = bids_for_player(player)
+          return 0 if bids.empty?
+
+          bids.sum(&:price)
+        end
+
+        def current_bid_amount(player, company)
+          @bids[company]&.select { |b| b.entity == player }&.sort(&:price)&.last&.price || 0
```
tobymao, line 85: maybe do a safe return first 

return unless (bids = @bids[company])

bids.select...min(&:price)  || 0

lib/engine/step/g_18_co/moving_bid_auction.rb
```diff
@@ -0,0 +1,215 @@
...
+          if @game.class::MUST_BID_INCREMENT_MULTIPLE && ((price - min) % min_increment).nonzero?
+            @game.game_error("Must increase bid by a multiple of #{@game.format_currency(min_increment)}")
+          end
+          if price > max_bid(entity, company)
+            @game.game_error("Cannot afford bid. Maximum possible bid is
+              #{@game.format_currency(max_bid(entity, company))}")
+          end
+
+          bids = @bids[company]
+          highest_player_bid = bids.sort(&:price).reverse.find { |b| b.entity == entity }
```
tobymao, line 170: can you bid twice? why do you need to sort?

lib/engine/step/g_18_co/moving_bid_auction.rb
```diff
@@ -0,0 +1,215 @@
...
+
+          @game.game_error("Minimum bid is #{@game.format_currency(min)} for #{company.name}") if price < min
+          if @game.class::MUST_BID_INCREMENT_MULTIPLE && ((price - min) % min_increment).nonzero?
+            @game.game_error("Must increase bid by a multiple of #{@game.format_currency(min_increment)}")
+          end
+          if price > max_bid(entity, company)
+            @game.game_error("Cannot afford bid. Maximum possible bid is
+              #{@game.format_currency(max_bid(entity, company))}")
+          end
+          if price < (old_price + min_increment)
```
tobymao, line 193: parens are unnecessary

lib/engine/step/g_18_co/moving_bid_auction.rb
```diff
@@ -0,0 +1,215 @@
...
+
+          bids = @bids[company]
+          bids << bid
+
+          @log << "#{entity.name} moves #{@game.format_currency(old_price)} bid from
+            #{old_company.name} to bid #{@game.format_currency(price)} for #{bid.company.name}"
+        end
+
+        def bids_for_player(player)
+          @bids.values.map do |bids|
```
tobymao, line 208: flat_map, compact seems unnecessary

lib/engine/step/g_18_co/moving_bid_auction.rb
ryandriskel: Yes, you can have multiple bids on a single company. For example, you could have a low bid on Company B, get outbid and then move a bid from Company A to Company B.

lib/engine/step/g_18_co/moving_bid_auction.rb
ryandriskel: I think we want max, not min ... otherwise yeah that's cleaner.

lib/engine/step/g_18_co/moving_bid_auction.rb
ryandriskel: I'm going to standardize the method for finding the highest player bid using max as we identified earlier

assets/app/view/game/round/auction.rb
ryandriskel: I made the change from `old_...` to `from_...` but I did not make the change to `to_...` as this allows the bid company and amount to be the same as the normal bid action and thus it greatly simplifies the logic on comparing bid values.

Issue #2559, opened 2020-12-05
[1846] [18LA]


mkkohls:
1846 18LA private drafts cannot select private.
Open a game of 46 or LA and try to draft a private. No options show up for selecting the private after clicking on it. Undo and redo work as well as passing in 18LA.

Game with issue:17895
{
  "id": 17895,
  "description": "Banal Game Title",
  "user": {
    "id": 1972,
    "name": "mkkohls"
  },
  "players": [
    {
      "id": 263,
      "name": "mrouser"
    },
    {
      "id": 146,
      "name": "Skanadron"
    },
    {
      "id": 1972,
      "name": "mkkohls"
    },
    {
      "id": 2857,
      "name": "mikeXIV"
    }
  ],
  "max_players": 5,
  "title": "1846",
  "settings": {
    "seed": 1148775833,
    "unlisted": true,
    "optional_rules": []
  },
  "user_settings": null,
  "status": "active",
  "turn": 1,
  "round": "Draft Round",
  "acting": [
    146
  ],
  "result": {},
  "actions": [
    {
      "type": "bid",
      "entity": 2857,
      "entity_type": "player",
      "id": 1,
      "company": "Pass (4)",
      "price": 0
    },
    {
      "type": "bid",
      "entity": 1972,
      "entity_type": "player",
      "id": 2,
      "company": "C&WI",
      "price": 60
    },
    {
      "id": 3,
      "type": "undo",
      "entity": 146,
      "entity_type": "player",
      "user": 1972
    },
    {
      "id": 4,
      "type": "redo",
      "entity": 1972,
      "entity_type": "player",
      "user": 1972
    },
    {
      "id": 5,
      "type": "undo",
      "entity": 146,
      "entity_type": "player",
      "user": 1972
    },
    {
      "id": 6,
      "type": "redo",
      "entity": 1972,
      "entity_type": "player",
      "user": 1972
    }
  ],
  "loaded": true,
  "created_at": 1607178799,
  "updated_at": 1607192185
}

michaeljb:
Fixed by #2562 and f952bb59cb2407b1fdebd73cda50b64d21d6cdff

PR #2561, opened 2020-12-05
Fix 1846/18MS Draft due to 1848 May_Reduce? Addition

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/2559


benjaminxscott:
Perhaps using `respond_to()` would handle this more generically than requiring changes to auction defs to other games (and possibly future games)

https://github.com/tobymao/18xx/pull/2562

michaeljb:
#2562 was merged instead of this one

PR #2562, opened 2020-12-05
revert de25ac2df with more generic fix

benjaminxscott:
Fixes auction issues in non-1848 games caused by a bug introduced in [my prior patch](https://github.com/tobymao/18xx/pull/2535/files/472e68df8ba3bb36fea95e5ec73413a59ae3c24e#r536492916)

### DIFF
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -170,7 +170,7 @@ module View
               ]
             end
 
-          if @step.may_reduce?(company)
+          if @step.respond_to?(:may_reduce) && @step.may_reduce?(company)
             company_actions << h(:button, {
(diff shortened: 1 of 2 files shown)


assets/app/view/game/round/auction.rb
```diff
@@ -170,7 +170,7 @@ def render_input(company)
               ]
             end
 
-          if @step.may_reduce?(company)
+          if @step.respond_to?(:may_reduce) and @step.may_reduce?(company)
```
michaeljb, line 173: This appears to work in the compiled JS, but if this was executing in Ruby it wouldn't work; the method is `may_reduce?` so the symbol would need to be `:may_reduce?` rather than `may_reduce`

```suggestion
          if @step.respond_to?(:may_reduce?) and @step.may_reduce?(company)
```

tobymao (commented):
tthere are problems but i will fix it

PR #2565, opened 2020-12-05
Exposes the field created_at for action in the json api

daniel-sousa-me:


models/action.rb
```diff
@@ -12,6 +12,6 @@ def before_create
   end
 
   def to_h
-    action.to_h.merge('user' => user_id)
+    action.to_h.merge('user' => user_id).merge(created_at: created_at_ts)
```
tobymao, line 15: change to be

merge('user' => user_id, 'created_at' => created_at)

tobymao (approved):

PR #2567, opened 2020-12-05
fix starting tile definitions for 1867's Toronto and Montreal hexes

michaeljb:
The third Montreal city gets placed in the center, will take some more effort to get that placed next to edge 5, but the track is connecting the correct cities at least.

I haven't read the 1867 rules, is it green on the board because it cannot be tokened before a green phase? Nobody can get connected to it until a green M tile is placed, so I was wondering if we could just leave that city off for now, but if there's some teleport-style token ability that would enable a token placement there that option is a no-go.

/cc @jenf 

## AAG board

Screenshot from 2020-12-05 14-56-25
Screenshot from 2020-12-05 14-56-29

## [Before](https://18xx.games/tiles/1867/F16+L12)

Screenshot from 2020-12-05 14-57-58

## After

Screenshot from 2020-12-05 14-58-24

### DIFF
--- a/lib/engine/config/game/g_1867.rb
+++ b/lib/engine/config/game/g_1867.rb
@@ -718,10 +718,10 @@ module Engine
       ]
     },
     "yellow": {
-      "city=revenue:40;city=revenue:40;city=revenue:40;path=a:1,b:_0;path=a:3,b:_0;label=M;upgrade=cost:20,terrain:water": [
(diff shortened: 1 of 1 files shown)


tobymao (approved):

jenf:
Yeah the greens can't be used until phase 3

roseundy:
Use "loc:5" to move the third (empty) city near edge 5.

michaeljb:
Fixed in fc61f08f, thanks @roseundy

PR #2568, opened 2020-12-05
1870: Implement privates

daniel-sousa-me:


lib/engine/ability/blocks_division.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class BlocksDivision < Base
```
jenf, line 7: This isn't intuitive to what 'Division' means

jenf: Add an overview to lib/engine/ability/README.md

jenf: Would suggest 'Partition' as this tends to relate separation of geographical things.

And BlocksCrossingPartitions would make it clear what the ability does.

jenf: Otherwise good from my pov.

daniel-sousa-me: All done :)

jenf:
Initial review looks good, but changes are extensive and would probably deserve a glance by @tobymao

daniel-sousa-me:
I'm sorry, when making the rename I realised there were some files that were not ported from the other branch, and a couple of edge cases I forgot to implement.

assets/app/view/game/buy_companies.rb
```diff
@@ -28,7 +28,7 @@ def render_companies
           },
         }
 
-        companies = @game.purchasable_companies.sort_by do |company|
+        companies = @game.purchasable_companies(@game.current_entity).sort_by do |company|
```
tobymao, line 31: why is this changed? current_entity should be the default

lib/engine/ability/blocks_crossing_partition.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class BlocksCrossingPartition < Base
```
tobymao, line 7: can just call this BlocksPartition, the crossing is inferred because it's a partition

lib/engine/game/g_1870.rb
```diff
@@ -69,15 +78,121 @@ def init_stock_market
...
+      def event_companies_buyable!
+        river_company.max_price = 2 * river_company.value
+      end
+
+      def event_remove_tokens!
+        removals = Hash.new { |h, k| h[k] = {} }
+
+        @corporations.each do |corp|
+          corp.assignments.dup.each do |company, _|
+            if ASSIGNMENT_TOKENS.keys.any?(company)
```
tobymao, line 94: ASSIGNMENT_TOKENS[company]

lib/engine/game/g_1870.rb
```diff
@@ -69,15 +78,121 @@ def init_stock_market
...
+              removals[company][:corporation] = corp.name
+              corp.remove_assignment!(company)
+            end
+          end
+        end
+
+        @hexes.each do |hex|
+          hex.assignments.dup.each do |company, _|
+            if ASSIGNMENT_TOKENS.keys.any?(company)
```
tobymao, line 103: ASSIGNMENT_TOKENS[company[

lib/engine/game/g_1870.rb
```diff
@@ -69,15 +78,121 @@ def init_stock_market
...
+        @ssw_corporation ||= corporation_by_id('SSW')
+      end
+
+      def river_corporations
+        [ssw_corporation, mp_corporation]
+      end
+
+      def purchasable_companies(entity = nil)
+        return super unless @phase.name == '1'
+        return [river_company] if [mp_corporation, ssw_corporation].any?(entity)
```
tobymao, line 139: include?

lib/engine/game/g_1870.rb
```diff
@@ -69,15 +78,121 @@ def init_stock_market
...
+        (tile.exits & hex.tile.borders.select { |b| b.type == :water }.map(&:edge)).empty? &&
+          hex.tile.partitions.all? do |partition|
+            tile.paths.all? do |path|
+              (path.exits - partition.inner).empty? || (path.exits - partition.outer).empty?
+            end
+          end
+      end
+
+      def upgrades_to?(from, to, _special = false)
+        return to.name == '170' if from.color == :green && P_HEXES.any?(from.hex.name)
```
tobymao, line 185: include?

lib/engine/game/g_1870.rb
```diff
@@ -69,15 +78,121 @@ def init_stock_market
...
+
+        return to.name == '171K' if from.color == :brown && from.hex.name == 'B11'
+        return to.name == '172L' if from.color == :brown && from.hex.name == 'C18'
+
+        super(from, to, false, to.cities.size.positive? && %i[yellow green].any?(to.color))
```
tobymao, line 190: to.cities.any? && [].include?(to.color)

lib/engine/part/partition.rb
```diff
@@ -0,0 +1,48 @@
...
+module Engine
+  module Part
+    class Partition < Base
+      attr_reader :a, :a_sign, :b, :b_sign, :type, :restrict, :magnet, :blockers, :inner, :outer
+
+      SIGN = {
+        '-' => -1,
+        nil => 0,
+        '+' => 1,
+      }.freeze
```
tobymao, line 14: why do you need sign?

lib/engine/part/partition.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def initialize(a, b, type, restrict, magnet)
```
tobymao, line 16: what is magnet?

lib/engine/part/partition.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        @type = type
+        # If restrict==inner, only allow paths between a and b. If outer, only between b and a
+        @restrict = restrict
+        # Pulls the river to this vertex. Also only cosmetic
+        @magnet = magnet&.to_i
+        @blockers = []
+
+        @inner = (@a..(@b - 1)).to_a
+        @outer = (0..5).to_a - inner
```
tobymao, line 34: shouldn't you only create these if it's the appropriate type

lib/engine/step/g_1870/buy_company.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module Engine
+  module Step
+    module G1870
+      class BuyCompany < BuyCompany
+        def can_buy_company?(entity)
+          return super unless @game.phase.name == '1'
+
+          entity == current_entity &&
+            @game.purchasable_companies(entity).any? &&
```
tobymao, line 13: include

lib/engine/tile.rb
```diff
@@ -444,11 +447,6 @@ def revenue_to_render
       @revenue_to_render ||= @stops.map(&:revenue_to_render)
     end
```
tobymao, line 449: why did you change this

lib/engine/step/g_1870/special_track.rb
```diff
@@ -0,0 +1,35 @@
...
+
+          extra_cost = tile_lay[:cost] - 40
+          extra_cost -= 20 if owner == @game.ssw_corporation && action.hex.name == 'H17'
+
+          lay_tile(action, extra_cost: extra_cost, entity: action.entity, spender: owner)
+
+          ability = tile_lay_abilities(action.entity)
+          ability.use!
+
+          if @game.river_corporations.any?(owner) && !owner.operated?
```
tobymao, line 26: include?

assets/app/view/game/buy_companies.rb
daniel-sousa-me: My bad, then. But everywhere else where it's called, it is passing then entity

lib/engine/game/g_1870.rb
daniel-sousa-me: I love how I forgot that `any?` existed when it actually was what I wanted to do :man_facepalming:

lib/engine/part/partition.rb
daniel-sousa-me: It's only a rendering thing. It draws the line a little bit to one edge or the other. It's important because the hex needs some information about what the river is going to avoid doing sharp corners (and I thought the sign was a simpler solution then having the code actually look at the bordering hexes)

lib/engine/part/partition.rb
daniel-sousa-me: It's just a rendering thing. I needed the river to go around this city:

image

and is also used here:

image

lib/engine/tile.rb
daniel-sousa-me: This was messing up code I was doing to generalise being able to put unlabeled tiles on labeled hexes, but I ended up throwing it away because my approach was getting too messy. The removal got lost on the revert. I've put it back

assets/app/view/game/buy_companies.rb
tobymao: then that should be refactored as it's redundant

daniel-sousa-me: I'll take a look to see which ones are redundant and open a separate PR removing them if possible. (this one is already re-removed)

tobymao:
how difficult will it be to remove the visual hints? if we don't i'd like to change the name of magnet

daniel-sousa-me:
The first version without the signs around the vertices looked like this:

image

I don't think this looks remotely ok, so I made the effort to find ways to smooth the curves. I truly don't believe there's a better approach to this, unless you want to overlay an svg over the entire map.

The "magnets" are mostly because of St. Louis. In this case I'm not convinced this is the best solution, but here it is really necessary to do something, because the city would appear on the wrong side of the river.

I triied an alternative moving the city instead:
image

But this also required a lot of fiddling and in the end the tiles looked a little bit weird. Also the rendering of the parts on that tile got messed up.

I have zero attachment to the name 'magnets'. I can change it to whichever term you prefer.

tobymao:
If you used the first version are there any playability concerns or is it just styling? It seems like the first version will be functionally better and not need to have magnets.

daniel-sousa-me:
When the river is going through the side of the hex, it is giving wrong information, because the river is just in one of the hexes, so you can lay track on the other hex.

Writing this made me realise I don't actually need to pass the magnet as a parameter. It can be drawn every time there's an inner/outer restriction.

tobymao (approved):

PR #2569, opened 2020-12-05
Add a shadow to cheater tokens to make them distinguishable from normal tokens

daniel-sousa-me:
Before:

image

After:

image

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -44,13 +44,30 @@ module View
 
         def render_part
           children = []
-          children << h(:circle, attrs: { r: @radius, fill: 'white' })
+          circle_attrs = { r: @radius, fill: 'white' }
+
+          props = { on: { click: ->(event) { on_click(event) } },
+                    attrs: { transform: '' } }
+
+          props[:attrs][:transform] = rotation_for_layout if @edge
+
+          if @city && @slot_index >= @city.normal_slots
+            children << h(:defs, [
+              h(:filter, { attrs: { id: 'shadow', x: '-50%', y: '-50%', width: '200%', height: '200%' } }, [
+                h(:feOffset, attrs: { result: 'offOut', in: 'SourceAlpha', dx: 2, dy: 2 }),
+                h(:feGaussianBlur, attrs: { result: 'blurOut', in: 'offOut', stdDeviation: '5' }),
+                h(:feBlend, attrs: { in: 'SourceGraphic', in2: 'blurOut', mode: 'normal' }),
+              ]),
+            ])
+            circle_attrs[:filter] = 'url(#shadow)'
+
+            # The shadow makes the token look larger, this offsets the effect a little
+            props[:attrs][:transform] += ' scale(0.95)'
+          end
+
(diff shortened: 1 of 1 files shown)


daniel-sousa-me:
I did this because of 1870, but I think it's a good improvement for all cheater tokens

ventusignis:
How does this work with 1870 if a company destinates on a yellow hex; but it upgrades to green later?

michaeljb:
> How does this work with 1870 if a company destinates on a yellow hex; but it upgrades to green later?

I just tested with 1846 2p Variant, and after the green tile was laid, the cheater token was then rendered as a regular one

tobymao:
this doesn't seem necessary

daniel-sousa-me:
In 1870 the token is always outside the slot. For 1870 I'm not using exactly cheater tokens, but the rendering is done in the same way.

In 1870 it's more important to have a distinction, because the value of the city is doubled if the route ends there. Also it looked confusing that you were upgrading a tile with 2 spaces and 2 tokens and when you upgrade you get to one that's supposed to have 2, but then an extra empty space appears.
It may be less relevant for the cheater token, but I thought it might be useful regardless, since they can share the same code.

>on a yellow hex; but it upgrades to green later
Yes, that is pretty common, since usually when you reach the destination, you weren't already there. It can also happen from brown to gray.

daniel-sousa-me:
> I just tested with 1846 2p Variant, and after the green tile was laid, the cheater token was then rendered as a regular one

That was the intended behaviour, because at that point the cheater token is exactly like a normal token for all purposes, right?

michaeljb:
> That was the intended behaviour, because at that point the cheater token is exactly like a normal token for all purposes, right?

Yep, was just commenting to answer ventusignis's question.

> In 1870 it's more important to have a distinction, because the value of the city is doubled if the route ends there. Also it looked confusing that you were upgrading a tile with 2 spaces and 2 tokens and when you upgrade you get to one that's supposed to have 2, but then an extra empty space appears.
It may be less relevant for the cheater token, but I thought it might be useful regardless, since they can share the same code.

I know some other games have destination tokens (1822 is the only one off the top of my head), and it may be useful to make all of _those_ different in some way. Maybe instead of shadow, a white or black circle around the edge of the token? That would be trickier than a shadow right now since the token is just a saved .svg file instead of code that we generate.

ryandriskel:
I can understand why we'd want to render them differently, but the shadow isn't distinctive enough for my eyes.

daniel-sousa-me:
Putting a border was the first thing I tried, but couldn't get it to look good. We can just have different svg's for the destination tokens. I'm doing that for the icons that show up before reaching the destination.

I know it's not easy to distinguish, but I didn't want to make them draw too much attention, because in the end they're just tokens.

It's pretty trivial to just apply the style to destination-like tokens than to cheater tokens, because they're in separate arrays. I just thought that this might be useful for cheater tokens. But now that I think more about it, I see it doesn't make sense to have rendering differently, because after being place they are indeed normal tokens (they behave exactly the same way as the other tokens there).

I'm closing this PR and will put the style only on the destination (and bundle it in that PR). But I'd like to continue the discussion of how they look, because I'm not too happy with the result either.

PR #2573, opened 2020-12-06
18CO Special Tile Lays

ryandriskel:
### Background 

**GJGR**: An owning Corporation may upgrade a yellow town to a green city in additional to its normal tile lay. Action closes the company or closes on purchase of “5” train.

**LNPW**: An owning Corporation may lay an extra tile at no cost in addition to its normal tile lay. Action closes the company or closes on purchase of “5” train.

### Implementation Notes

I moved some shared code between the Normal Track Step and Special Track Step into a shared Tracker module.

I don't love having to modify the tiles in the json. However, I can't see a way to fake it at runtime. The explicit code in upgrades_to? allows it to work, though. No need to detect if the hex has a town since the ability definition restricts the upgrade to town hexes.

### Impacts 

GJGR only activates on towns, and only shows green cities on yellow towns
Screen Shot 2020-12-05 at 8 59 00 PM

LNPW can lay a yellow tile
Screen Shot 2020-12-05 at 9 02 26 PM

LNPW doesn't allow upgrades
Screen Shot 2020-12-05 at 9 02 20 PM

### DIFF
--- a/lib/engine/step/g_18_co/special_track.rb
+++ b/lib/engine/step/g_18_co/special_track.rb
@@ -10,7 +10,7 @@ module Engine
         include Tracker
 
         def process_lay_tile(action)
-          super(action)
+          super
 
           clear_upgrade_icon(action.hex.tile)
           collect_mines(action.entity.owner, action.hex)
--- a/lib/engine/step/g_18_co/special_track.rb
+++ b/lib/engine/step/g_18_co/special_track.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require_relative '../special_track'
+require_relative 'tracker'
+
+module Engine
+  module Step
+    module G18CO
+      class SpecialTrack < SpecialTrack
+        include Tracker
+
+        def process_lay_tile(action)
+          super(action)
+
+          clear_upgrade_icon(action.hex.tile)
+          collect_mines(action.entity.owner, action.hex)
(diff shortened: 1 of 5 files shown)


lib/engine/step/g_18_co/special_track.rb
```diff
@@ -0,0 +1,22 @@
...
+require_relative 'tracker'
+
+module Engine
+  module Step
+    module G18CO
+      class SpecialTrack < SpecialTrack
+        include Tracker
+
+        def process_lay_tile(action)
+          super(action)
```
tobymao, line 13: super, the action is not necessary

lib/engine/step/g_18_co/tracker.rb
```diff
@@ -0,0 +1,30 @@
...
+        def clear_upgrade_icon(tile)
+          return if tile.cities.empty?
+          return if tile.icons.empty? { |icon| icon.name == 'upgrade' }
+
+          tile.icons.reject! { |icon| icon.name == 'upgrade' }
+        end
+
+        def collect_mines(corporation, hex)
+          # Mine Token Collection
+          return unless hex.tile.icons.map(&:name).include?('mine')
```
tobymao, line 17: hex.tile.icons.any? { ... }

ryandriskel: oh sure pick on old code 😄

PR #2574, opened 2020-12-06
1828: Implement C&P company

crericha:
Implemented the following abilities for C&P:

- Pays $15 in revenue until the end of the auction
- Train doesn't rust
- Must run the tile it laid
- Run does not need to include its station
- Does not run if no tile laid
- Buying company has the option to replace C&P home token with one of its tokens for free, provided it has a token and has not laid a token this round.
- If the buying company cannot or does not replace the C&P token, a blocking token is placed instead

Cleaned-up the blocking token logic.

### DIFF
--- a/lib/engine/game/g_1828.rb
+++ b/lib/engine/game/g_1828.rb
@@ -261,6 +261,12 @@ module Engine
         end
       end
 
+      def can_run_route?(entity)
+        return false if entity.id == 'C&P' && !@round.last_tile_lay
+
+        super
+      end
+
       private
 
       def setup_minors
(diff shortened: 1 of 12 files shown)


lib/engine/game/g_1828.rb
```diff
@@ -264,12 +282,25 @@ def coalfields.blocks?(corporation)
...
       end
 
+      def can_run_route?(entity)
+        return false if entity.id == 'C&P' && !@round.last_tile_lay
+
+        super
+      end
+
+      def city_tokened_by?(city, entity)
+        return @graph.connected_nodes(entity).keys.include?(city) if entity.id == 'C&P'
```
tobymao, line 292: why would you do keys.include?

can't you just do connected_nodes(entity)[city]

crericha: Good catch. Let's pretend that never happened lol.

PR #2576, opened 2020-12-06
18CO Corporations can buy Un-auctioned Companies

ryandriskel:

A​ ​Private​ ​Company​ ​in​ ​the​ ​Market​ ​is​ ​purchased​ ​at​ ​50%​ ​of​ ​the​ ​Private​ ​Company’s​ ​face​ ​value.

Private companies wind up in the market if they are not purchased during auction. They can only be bought by corporations, not players.

### Implementation Notes

Changes to the generic **buy_company.rb**
- Don't run the companies.delete call if the owner is nil, because it will throw an error
- Pay to bank if owner is nil ( the market )
- Add text to show "the market" if owner is nil

Changes to the generic **buy_company.rb** were not strictly necessary for 18CO as a player cannot buy from market but rather done for consistency and to prevent errors if a game allowed it in the future.
- Add text to show "the market" if owner is nil

Corporations can buy from market
Screen Shot 2020-12-06 at 10 27 26 AM

Players correctly cannot buy from market
Screen Shot 2020-12-06 at 9 42 34 AM

### DIFF
--- a/lib/engine/step/buy_company.rb
+++ b/lib/engine/step/buy_company.rb
@@ -58,7 +58,7 @@ module Engine
 
         log_later = []
         company.owner = entity
-        owner.companies.delete(company)
+        owner&.companies&.delete(company)
 
         company.abilities(:assign_corporation) do |ability|
           Assignable.remove_from_all!(assignable_corporations, company.id) do |unassigned|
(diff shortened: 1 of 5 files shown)


lib/engine/step/buy_company.rb
```diff
@@ -58,7 +58,7 @@ def process_buy_company(action)
 
         log_later = []
         company.owner = entity
-        owner.companies.delete(company)
+        owner&.companies&.delete(company)
```
ryandriskel, line 61: Might need to delete from @game.companies here when owner is nil?

ryandriskel:
Converting to draft because I am about to go out of town and don't want the remotest chance of something breaking while I am away.

jenf (approved):

PR #2577, opened 2020-12-06
Variants should also be considered for discount (fixes #2492)

### LINKED ISSUE #2492: Discount does not work for variants
The trains to choose from does not include the variants. Even if discount is specified for the variants.

The reason is that discountable_trains_for in depot.rb ignores variants. Suggestion is to return variants that are discountable as well.

Requires changes in depot.rb and buy_trains.rb

perwestling:
Now variants is added to discountable trains, if the base version
has a discount for the a train. The discount will be the same for
the variants as for the base version.

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -109,20 +109,21 @@ module View
         if discountable_trains.any?
           children << h(:h3, h3_props, 'Exchange Trains')
 
-          discountable_trains.each do |train, discount_train, price|
+          discountable_trains.each do |train, discount_train, discount_name, price|
             exchange_train = lambda do
               process_action(
                 Engine::Action::BuyTrain.new(
                   @corporation,
                   train: discount_train,
                   price: price,
+                  variant: discount_name,
                   exchange: train,
                 )
               )
             end
 
             children << h(:div, [
-              "#{train.name} -> #{discount_train.name} #{@game.format_currency(price)} ",
+              "#{train.name} -> #{discount_name} #{@game.format_currency(price)} ",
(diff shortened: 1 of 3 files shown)


assets/app/view/game/buy_trains.rb
```diff
@@ -109,20 +109,21 @@ def render
         if discountable_trains.any?
           children << h(:h3, h3_props, 'Exchange Trains')
 
-          discountable_trains.each do |train, discount_train, price|
+          discountable_trains.each do |train, discount_train, discount_name, price|
```
tobymao, line 112: should the variable be called variant?

lib/engine/depot.rb
```diff
@@ -89,9 +89,20 @@ def discountable_trains_for(corporation)
...
+            next if v[:name] == name
+
+            price = v[:price] - (discount_train.price - discounted_price)
+            discount_info << [train, discount_train, v[:name], price]
+          end
+
+          discount_info
         end.compact
-      end
+      end.flatten(1)
```
tobymao, line 105: shouldn't flat_map just work?

assets/app/view/game/buy_trains.rb
perwestling: Good idea

lib/engine/depot.rb
perwestling: This is me being unused to Ruby... you are correct, flat_map gives the same result.

Consider line 93-100: is there a better way to inject extra elements during a map operation?
Normally the map [t1, t2, t3] would result in [[t1, _, _], [t2, _, _], [t3, _, _]] but I wanted to insert a [t1-variant, _, _] extra. The code I wrote do this but it felt a bit clunky...

PR #2579, opened 2020-12-06
Bug fix, 1846: add potential cash from minor to entity's buying power

### LINKED ISSUE #2566: 1846 - track step is skipped if corp doesn't have enough cash, even though they could buy a minor and get enough cash to lay a tile
https://18xx.games/game/17894?action=120 ([and in gist here](https://gist.github.com/michaeljb/63cc81e5b3bc2b5addae83ae4f5e19ca))

In 1846, track lays cost $20; the track step gets skipped automatically since NYC doesn't have enough cash, but NYC could buy MS from the owner and gain the MS's $40 treasury, and then afford a track lay.

michaeljb:
This prevents auto-skipping track/train steps when the corporation doesn't currently have enough cash for the action, but could have enough cash if they buy in one of the minors.

* because the minors come with a train, the entity's current trains must be
  taken into account
* when considering train buying power for an entity, if they have 0
  trains (i.e., if they must buy a train), the minor's potential cash should
  *not* be considered, as this prevents the president from contributing cash for
  EMR

[Fixes #2566]

-----

The migration is very straightforward, just inserting pass actions when necessary. I ran it locally for all "active" games and then all "finished" games; active took around 4 minutes while finished took around 40, so when released I'll do the same.

### DIFF
--- a/lib/engine/game/g_1846.rb
+++ b/lib/engine/game/g_1846.rb
@@ -568,6 +568,24 @@ module Engine
 
         help
       end
+
+      def potential_minor_cash(entity, allowed_trains: (0..3))
+        if entity.corporation? && entity.cash.positive? && allowed_trains.include?(entity.trains.size)
+          @minors.reduce(0) do |memo, minor|
+            minor.owned_by_player? && minor.cash.positive? ? memo + minor.cash - 1 : memo
+          end
+        else
+          0
+        end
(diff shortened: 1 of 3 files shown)


michaeljb:
Whoops, forgot to migrate specs, will fix and re-open...

PR #2582, opened 2020-12-06
Attempts to optimize connections

jenf:
Skip reconnection walking if all new paths do not connect to anything,
simplify it to just migrate the references to the new hex

### DIFF
--- a/lib/engine/hex.rb
+++ b/lib/engine/hex.rb
@@ -98,11 +98,20 @@ module Engine
       new_tile.hex = self
     end
 
-    def lay(tile)
+    def lay(tile, downgrade = false)
+      
       # key: city on @tile (AKA old_city)
       # values: city on tile (AKA new_city)
       #
       # map old cities to new based on edges they are connected to
+      #isolated = false
+      isolated = isolated_lay?(tile) && !downgrade
+      #if id == 'E20'
+        
+      #  puts isolated, $actionid
+      #  puts "before", @connections
+      #end
+
       city_map =
         # if @tile is blank, map cities by index
         if @tile.cities.flat_map(&:exits).empty? && (@tile.cities.size == tile.cities.size)
@@ -167,6 +176,7 @@ module Engine
       tile.borders.concat(@tile.borders)
       @tile.borders.clear
 
+      old_tile = @tile
       @tile.hex = nil
       tile.hex = self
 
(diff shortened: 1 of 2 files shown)


jenf:
@tobymao here's an optimisaiton, a few connections related tests are failing.

The code can probably be done better, but I won't be able to look at it for a bit

jenf:
It can probably be generalized to only connect new exits if unambigious, and migrate the existing ones.
Original:
`     TOTAL    (pct)     SAMPLES    (pct)     FRAME
     16474  (46.1%)        4866  (13.6%)     Engine::Part::Path#walk
      5803  (16.2%)        2404   (6.7%)     Engine::Connection.valid_connection?
      2311   (6.5%)        2311   (6.5%)     Engine::Tile#id
      2125   (5.9%)        2125   (5.9%)     (sweeping)
      3739  (10.5%)        1433   (4.0%)     Engine::Part::Base#id
      1332   (3.7%)        1332   (3.7%)     (marking)`

New:
`     10968  (36.0%)        3634  (11.9%)     Engine::Part::Path#walk
      1531   (5.0%)        1531   (5.0%)     (sweeping)
      3187  (10.5%)        1300   (4.3%)     Engine::Connection.valid_connection?
      1259   (4.1%)        1259   (4.1%)     Engine::Tile#id
       915   (3.0%)         915   (3.0%)     (marking)
`

lib/engine/hex.rb
```diff
@@ -205,6 +218,29 @@ def connect!
...
+      #wanted_hex = 'D6'
+      #if id == wanted_hex
+      #  puts "old", @tile.paths 
+      #  puts "new", tile.paths
+      #  puts tile.paths.map(&:exits)
+      #  puts tile.paths.map(&:exits).uniq
+      #end
+
+      # The migration code cannot handle Chicago like upgrades
+      return false if tile.paths.map(&:exits) != tile.paths.map(&:exits).uniq
```
tobymao, line 232: this doesn't seem quite right and neither does just tile.exits

tobymao:
it seems like we could actually save some time by just caching ids?

tobymao:
just thinking aloud, would it be possible to instead of connecting every lay, to only connect stuff when we need it for running routes?

tobymao:
Let's chat more tomorrow, but it may be possible to do away with connections as we know it. It's going to be quite a large piece of work but will probably shave off a huge amount of time.

It seems like connections are only used in routing, which means that we really only need have them when a user is acutally creating a route. When rehydrating them from the DB, we can store the literal representation so in a sense they are only calculated when needed, and only when the user is actually triggering them. That means no connections are ever created when loading games.

jenf:
closing this as we need a better solution

PR #2583, opened 2020-12-06
18CO Takeovers

ryandriskel:

If,​ ​at​ ​any​ ​time,​ ​a​ ​Corporation​ ​owns​ ​the​ ​President’s​ ​Certificate​ ​of​ ​another​ ​Corporation,​ ​the​ ​owning​ ​Corporation​ ​must​ ​take​ ​over​ ​the owned​ ​Corporation​ ​as​ ​follows:

- [x] Third-party shares on the owned Corporation charter are sold to the Market with no impact to share price. This money is placed in the owned Corporation Treasury.
- [x] Remove​ ​any​ ​owned​ ​Station​ ​Tokens​ ​in​ ​hexes​ ​containing​ ​the​ ​owning​ ​Corporation’s​ ​Station​ ​tokens.
- [x] Up​ ​to​ ​two​ ​owned​ ​Station​ ​Tokens​ ​on​ ​the​ ​board​ ​may​ ​be​ ​replaced​ ​with​ ​the​ ​owning​ ​Corporation’s​ ​Station​ ​Tokens​ ​at​ ​no​ ​cost. Use​ ​tokens​ ​from​ ​the​ ​right​ ​side​ ​of​ ​the​ ​Corporation​ ​Charter​ ​for​ ​this​ ​purpose.
- [x] The​ ​owning​ ​Corporation​ ​receives​ ​the​ ​Trains,​ ​Mine​ ​Tokens​ ​and​ ​Private​ ​Companies.
- [x] The​ ​owned​ ​Corporation’s​ ​Treasury,​ ​rounded​ ​down​ ​to​ ​the​ ​nearest​ ​$10,​ ​is​ ​paid​ ​out​ ​to​ ​the​ ​owned​ ​Corporation’s​ ​shareholders as​ ​if​ ​it​ ​were​ ​a​ ​dividend.​ ​Treasury​ ​money​ ​paid​ ​to​ ​shares​ ​in​ ​the​ ​Market​ ​is​ ​returned​ ​to​ ​the​ ​Bank.​ ​Any​ ​leftover​ ​treasury​ ​cash​ ​is transferred​ ​to​ ​the​ ​owning​ ​Corporation’s​ ​treasury.
- [x] The​ ​owning​ ​Corporation’s​ ​Train​ ​limit​ ​is​ ​increased​ ​by​ ​one.
- [x] Any​ ​excess​ ​trains​ ​are​ ​discarded​ ​to​ ​the​ ​Market​ ​down​ ​to​ ​the​ ​new​ ​limit.
- [x]  The​ ​owned​ ​Corporation​ ​is​ ​closed.​ ​Remove​ ​its​ ​Shares,​ ​Stock​ ​Token,​ ​any​ ​remaining​ ​Station​ ​Tokens​ ​and​ ​any​ ​previously​ ​taken​ ​over Charters​ ​underneath​ ​the​ ​owned​ ​Corporation’s​ ​charter​ ​from​ ​the​ ​game.

### Implementation Notes

TBD

### In Action

Most of the Takeover is automated. The only player choice is which of up to 2 tokens to swap to the new corporation.
Screen Shot 2020-12-06 at 3 18 07 PM

Showing the DPAC Denver redundant token was removed, and the remaining DPAC token hex is highlighted.
Screen Shot 2020-12-06 at 3 18 33 PM

The DPAC token was replaced with a DSNG token. Since no further DPAC tokens remained on the board, the Takeover step ended.
Screen Shot 2020-12-06 at 3 19 45 PM

DSNG has 5 trains, legal with the +1 ability from the takeover. The ability is also shown on the card.
Screen Shot 2020-12-06 at 3 16 41 PM

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -76,15 +76,23 @@ module View
           h(:text, { attrs: attrs }, text)
         end
 
+        def cheater_token(entity)
+          token_abilities = entity.abilities(:token)
+          return nil if token_abilities.empty?
+
+          token_abilities&.cheater
+        end
+
         def on_click(event)
           return if @tile_selector&.is_a?(Lib::TileSelector)
 
           step = @game.round.active_step(@selected_company)
           entity = @selected_company || step.current_entity
           actions = step.actions(entity)
+
           return if (%w[remove_token place_token] & actions).empty?
-          return if @token && !step.can_replace_token?(entity, @token) &&
-                    !(cheater = entity.abilities(:token)&.cheater)
+
+          return if @token && !step.can_replace_token?(entity, @token) && !(cheater = cheater_token(entity))
 
           event.JS.stopPropagation
 
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -76,23 +76,15 @@ module View
           h(:text, { attrs: attrs }, text)
         end
 
-        def cheater_token(entity)
-          token_abilities = entity.abilities(:token)
-          return nil if token_abilities.empty?
-
-          token_abilities&.cheater
-        end
-
         def on_click(event)
           return if @tile_selector&.is_a?(Lib::TileSelector)
 
           step = @game.round.active_step(@selected_company)
           entity = @selected_company || step.current_entity
           actions = step.actions(entity)
-
           return if (%w[remove_token place_token] & actions).empty?
-
-          return if @token && !step.can_replace_token?(entity, @token) && !(cheater = cheater_token(entity))
+          return if @token && !step.can_replace_token?(entity, @token) &&
+                    !(cheater = entity.abilities(:token)&.cheater)
 
           event.JS.stopPropagation
 
--- a/lib/engine/step/g_18_co/takeover.rb
+++ b/lib/engine/step/g_18_co/takeover.rb
@@ -23,7 +23,7 @@ module Engine
         end
 
         def active?
-          takeover_in_progress?
+          takeover_in_progress
         end
 
         def current_entity
@@ -84,29 +84,28 @@ module Engine
 
           return if place_count.positive? &&
             taken_entity.placed_tokens.any? &&
-            current_entity.unplaced_tokens.empty?
+            current_entity.unplaced_tokens.any?
 
           close_corporation(taken_entity)
         end
 
-        def takeover_in_progress?
+        def takeover_in_progress
           return true unless current_entity.nil?
 
-          @game.corporations.find do |source|
-            next if source.nil?
-            next unless source.floated?
-
-            takeover_corporation?(source)
+          @game.corporations.dup.find do |source|
+            source == takeover_corporation(source)
           end
         end
 
-        def takeover_corporation?(source)
+        def takeover_corporation(source)
+          return unless source.floated?
+
           president_share_count = source.owner.num_shares_of(source)
           source.corporate_share_holders.find do |c_s_h|
             corporation, corporate_share_percent = c_s_h
             next unless president_share_count < (corporate_share_percent / source.share_percent)
 
-            execute(source, corporation)
+            source == execute(source, corporation)
           end
         end
 
@@ -139,8 +138,7 @@ module Engine
             .map { |key, vals| [key, vals.size] }
 
           cash = {}
-          # must use while as shares array is modified by transfer
-          while (share = source.corporate_shares.first)
+          source.corporate_shares.dup.each do |share|
             @game.bank.spend(share.price, source)
             cash[share.corporation] =
               cash[share.corporation].nil? ? share.price : cash[share.corporation] + share.price
@@ -201,7 +199,7 @@ module Engine
         # Treasury money paid to shares in the Market is returned to the Bank.
         # Any leftover treasury cash is transferred to the owning Corporation's treasury.
         def distribute_treasury(source, destination)
-          payout = (source.cash.to_f / 10).floor.to_i
+          payout = (source.cash.to_f / 10).floor
 
           payout_shareholders(source, payout)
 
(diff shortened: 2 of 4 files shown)


assets/app/view/game/part/city_slot.rb
```diff
@@ -76,15 +76,23 @@ def reservation
           h(:text, { attrs: attrs }, text)
         end
 
+        def cheater_token(entity)
+          token_abilities = entity.abilities(:token)
+          return nil if token_abilities.empty?
+
+          token_abilities&.cheater
```
tobymao, line 83: if token_abilities can be nil, line 81 will crash, if it can't be nil, why do you use &?

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,267 @@
...
+            current_entity.unplaced_tokens.empty?
+
+          close_corporation(taken_entity)
+        end
+
+        def takeover_in_progress?
+          return true unless current_entity.nil?
+
+          @game.corporations.find do |source|
+            next if source.nil?
```
tobymao, line 96: next unless corporation&.floated? should be all you need

or even better just put return unless source&floated? in takeover_corporation?

tobymao: using find here and in takeover isn't super idiomatic, since you're using ? i think the intent is to return a bool, so many any? makes more sense

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,267 @@
...
+            next unless source.floated?
+
+            takeover_corporation?(source)
+          end
+        end
+
+        def takeover_corporation?(source)
```
tobymao, line 103: a ? method shouldn't have side effects, execute looks like a side effect

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,267 @@
...
+        end
+
+        def return_corporate_shares_to_market(source)
+          return unless source.corporate_shares.any?
+
+          returned_certs = source.corporate_shares.group_by(&:corporation)
+            .map { |key, vals| [key, vals.size] }
+
+          cash = {}
+          # must use while as shares array is modified by transfer
```
tobymao, line 142: if this is the issue, just do dup.each, comments like this are a red flag

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,267 @@
...
+          @game.mines_add(destination, mine_count)
+          @game.mines_remove(source)
+        end
+
+        # The owned Corporation's Treasury, rounded down to the nearest $10
+        # is paid out to the owned Corporation's shareholders as if it were a dividend.
+        # Treasury money paid to shares in the Market is returned to the Bank.
+        # Any leftover treasury cash is transferred to the owning Corporation's treasury.
+        def distribute_treasury(source, destination)
+          payout = (source.cash.to_f / 10).floor.to_i
```
tobymao, line 204: floor.to_i seems redundant

lib/engine/step/g_18_co/takeover.rb
ryandriskel: I think this was another instance of the array being modified caused problems. I'll use .dup here, too.

assets/app/view/game/part/city_slot.rb
ryandriskel: I rebased to the latest master and the original cheater code now works. I will return this file's code to the previous implementation.

lib/engine/step/g_18_co/takeover.rb
tobymao: this comment is unecessary now, please delete

tobymao (approved):

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,265 @@
...
+            taken_entity.placed_tokens.any? &&
+            current_entity.unplaced_tokens.any?
+
+          close_corporation(taken_entity)
+        end
+
+        def takeover_in_progress
+          return true unless current_entity.nil?
+
+          @game.corporations.dup.find do |source|
```
jenf, line 95: I don't generally expect find to cause side effects and confused me when reading this.

maybe each with a break?

lib/engine/step/g_18_co/takeover.rb
jenf: I'd say this should have a !

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,265 @@
...
+            source == takeover_corporation(source)
+          end
+        end
+
+        def takeover_corporation(source)
```
jenf, line 100: change to takeover_corporation!

jenf (changes requested):

lib/engine/step/g_18_co/takeover.rb
tobymao: takeover_in_progress shouldn't have any side effects or you should rename this

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,265 @@
...
+
+          return if place_count.positive? &&
+            taken_entity.placed_tokens.any? &&
+            current_entity.unplaced_tokens.any?
+
+          close_corporation(taken_entity)
+        end
+
+        def takeover_in_progress
+          return true unless current_entity.nil?
```
tobymao, line 93: return true if current_entity

lib/engine/step/g_18_co/takeover.rb
```diff
@@ -0,0 +1,265 @@
...
+          @game.corporations.dup.find do |source|
+            source == takeover_corporation(source)
+          end
+        end
+
+        def takeover_corporation(source)
+          return unless source.floated?
+
+          president_share_count = source.owner.num_shares_of(source)
+          source.corporate_share_holders.find do |c_s_h|
```
tobymao, line 104: i think in general you shouldn't be executing stuff in a find either, can you instead find the corporation and then execute it?

PR #2602, opened 2020-12-09
1860 Corporate Bankruptcy

roseundy:
Last of the big corporate status implementations: Corporate Bankruptcy

Unlike closing a corporation, a bankrupt corporation keeps it's assets and can be re-parred with a wider range of prices than before it was bankrupt. Any outstanding stock shares are moved from players/pool back to IPO. A new priority deal is assigned and if the bankruptcy occurs in the middle of a stock round, the stock round is restarted. It also implements the concept of "flipped" tokens for bankrupted corps that block track laying and tokening, but not running trains:

Flipped_token

Common file changes:

- UI: Corporation: handle showing assets for bankrupt corps, shrink font for really long train strings (1860 is bad in this regard)
- UI: Token: render flipped tokens
- Engine::Token: new status instance var
- Engine::Corporation: new unfloat! method
- Round::Stock: game configuration to prevent share price change when stock is sold out
- Game::Base: add config variable for above

bankrupte corp card

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -87,19 +87,22 @@ module View
           children << h(:div, props, subchildren)
         end
 
-        if @game.status_str(@corporation)
-          props = {
-            style: {
-              grid: '1fr / repeat(2, max-content)',
-              gap: '2rem',
-              justifyContent: 'center',
-              backgroundColor: color_for(:bg2),
-              color: color_for(:font2),
-            },
-          }
+        status_props = {
+          style: {
+            grid: '1fr / repeat(2, max-content)',
+            gap: '2rem',
+            justifyContent: 'center',
+            backgroundColor: color_for(:bg2),
+            color: color_for(:font2),
+          },
+        }
 
-          children << h(:div, props, render_status)
+        if @corporation.trains.any? && !@corporation.floated?
+          if @corporation.trains.any? && !@corporation.floated?
+            children << h(:div, status_props, @game.float_str(@corporation))
+          end
         end
+        children << h(:div, status_props, render_status) if @game.status_str(@corporation)
 
         h('div.corp.card', { style: card_style, on: { click: select_corporation } }, children)
       end
@@ -166,7 +169,7 @@ module View
         }
 
         holdings =
-          if !@corporation.corporation? || @corporation.floated?
+          if !@corporation.corporation? || @corporation.floated? || @corporation.trains.any?
             h(:div, holdings_props, [render_trains, render_cash])
           elsif @corporation.cash.positive?
             h(:div, holdings_props, [render_to_float, render_cash])
@@ -203,12 +206,17 @@ module View
             grid: '25px auto / 1fr',
           },
         }
+
         value_props = {
           style: {
             maxWidth: '7.5rem',
             fontWeight: 'bold',
           },
         }
+
+        value_props[:style].merge!(fontSize: 'small') if value.size > 6
+        value_props[:style].merge!(fontSize: 'x-small') if value.size > 10
+
         key_props = {
           style: {
             alignSelf: 'end',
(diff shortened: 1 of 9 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -87,19 +87,22 @@ def render
...
+            gap: '2rem',
+            justifyContent: 'center',
+            backgroundColor: color_for(:bg2),
+            color: color_for(:font2),
+          },
+        }
 
-          children << h(:div, props, render_status)
+        if @corporation.trains.any? && !@corporation.floated?
+          if @corporation.trains.any? && !@corporation.floated?
```
tobymao, line 101: typo?

assets/app/view/game/token.rb
```diff
@@ -17,6 +27,39 @@ def render
           },
         )
       end
+
+      def render_stroke
+        redw = 7
```
tobymao, line 32: constants up top

lib/engine/game/g_1860.rb
```diff
@@ -252,11 +320,16 @@ def current_layer
...
 
       def float_corporation(corporation)
-        # Designer says that bankrupt corps keep insolvency flag
-        clear_bankrupt(corporation)
+        clear_bankrupt!(corporation)
         super
       end
 
+      def action_processed(_action)
+        @corporations.each do |corporation|
```
tobymao, line 328: you need to do this on every single action? that's gonna make the action * N

lib/engine/game/g_1860.rb
```diff
@@ -450,6 +528,56 @@ def compute_distance_graph(corporation)
...
 
+      # needed for custom_node_walk
+      def custom_node_select(node, paths, corporation: nil)
+        on = paths.map { |p| [p, 0] }.to_h
+
+        custom_node_walk(node, on: on, corporation: corporation) do |path|
+          on[path] = 1 if on[path]
+        end
+
+        val = on.keys.select { |p| on[p] == 1 }
```
tobymao, line 539: you don't need the val, just do the selcet statement, you also don't need to do 0, just do nil, then you can do select { on[p] }

lib/engine/game/g_1860.rb
```diff
@@ -450,6 +528,56 @@ def compute_distance_graph(corporation)
...
+        val
+      end
+
+      # needed for custom_blocks?
+      def custom_node_walk(node, visited: nil, on: nil, corporation: nil, visited_paths: {})
```
tobymao, line 544: what's this do?

assets/app/view/game/corporation.rb
roseundy: Yep. Reorganized this section and biffed the cleanup.

assets/app/view/game/token.rb
roseundy: OK.

lib/engine/game/g_1860.rb
roseundy: This is already in the base class, and if anything, mine is simpler:

```
      def action_processed(_action)
        @corporations.dup.each do |corporation|
          close_corporation(corporation) if corporation.share_price&.type == :close
        end if stock_market.has_close_cell
      end
```

lib/engine/game/g_1860.rb
roseundy: Yep, code left over from when I was debugging.

lib/engine/game/g_1860.rb
roseundy: Fundamentally I need a version of Node::select that uses a custom City::blocks? method that understands about flipped tokens for bankrupt corporation.

roseundy: I thought about temporarily replacing flipped tokens with neutral tokens during route selection and dividend steps, but that seemed too inelegant and prone to bugs.

tobymao (approved):

PR #2607, opened 2020-12-10
1867: Add support for multiple types on stock price at same time

jenf:
Add support for conversion prices for minors
Add maximum stock price on minors
Add unlimited 7,8 9 track
Tidy up A17 & D2 tiles

### DIFF
--- a/lib/engine/share_price.rb
+++ b/lib/engine/share_price.rb
@@ -4,6 +4,24 @@ module Engine
   class SharePrice
     attr_reader :coordinates, :price, :corporations, :can_par, :type, :types
 
+    TYPE_MAP = {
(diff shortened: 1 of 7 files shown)


lib/engine/share_price.rb
```diff
@@ -2,58 +2,53 @@
...
-          :par_1
-        when /z/
-          :par_2
-        end
+      m = code.match(/(\d*)([a-zA-Z]*)/)
+      price = m[1].to_i
+
+      types = []
+      m[2].chars.each do |char|
+        type_map = {
```
tobymao, line 15: this should be taken out of the hot loop

tobymao (approved):

Issue #2592, opened 2020-12-08
1817WO Rules Differences


ventusignis:
- [x] A corporation with a token in Nieuw Zeeland gets $20 at the end of the OR if it has no loans
- [x] A corporation with a token in Nieuw Zeeland gets to ignore one loan's worth of interest if it has loans
- [x] Nieuew Zeeland cannot be chosen for IPO before Phase 3
- [x] Nieuew Zeeland cannot be started as a 2 share corporation
- [x] A player (may also choose not to) place another token for the Nieuew Zeeland corporation in another city slot in the SR it starts when it buys tokens

PR #2612, opened 2020-12-11
1817wo nieuwzeeland

ventusignis:
This resolves 4 of the rules in https://github.com/tobymao/18xx/issues/2592

Namely: 
* A corporation with a token in Nieuw Zeeland gets $20 at the end of the OR if it has no loans
* A corporation with a token in Nieuw Zeeland gets to ignore one loan's worth of interest if it has loans
* Nieuw Zeeland cannot be chosen for IPO before Phase 3
* Nieuw Zeeland cannot be started as a 2 share corporation

### DIFF
--- a/lib/engine/round/g_1817_wo/operating.rb
+++ b/lib/engine/round/g_1817_wo/operating.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require_relative '../g_1817/operating'
+
+module Engine
+  module Round
+    module G1817WO
+      class Operating < G1817::Operating
+        def pay_interest!(entity)
+          @cash_crisis_due_to_interest = nil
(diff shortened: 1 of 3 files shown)


lib/engine/round/g_1817_wo/operating.rb
```diff
@@ -0,0 +1,40 @@
...
+
+module Engine
+  module Round
+    module G1817WO
+      class Operating < G1817::Operating
+        def pay_interest!(entity)
+          @cash_crisis_due_to_interest = nil
+          return if @paid_loans[entity]
+
+          return unless @steps.any? { |step| step.passed? && step.is_a?(Step::BuyTrain) }
```
tobymao, line 13: might be better to use round state here

tobymao (approved):

PR #2613, opened 2020-12-11
18CO Corporate Share Sales

ryandriskel:

Shares (of other corporations)​ ​owned​ ​by​ ​a​ ​Corporation​ ​can​ ​be​ ​sold​ ​if​ ​the​ ​operating​ ​Corporation​ ​did​ ​not​ ​purchase​ ​any​ ​shares​ ​in​ ​the​ ​target Corporation​ ​earlier in​ ​the​ ​same​ ​operating round.

### Implementation Notes

Changes to _operating.rb_ were required to add the new view
Changes to _sell_shares.rb_ allow for the corporate sell action which is necessary to allow for the corporate view.
Changes to _share_pool.rb_ correct the assumption that if a corporation is selling shares, its an issue. That's only true if the corporation is the same as the bundle.
Changes to _corporate_buy_shares.rb_ allow for distinguishing between which corporation bought which other corporations

### Impacts

Here we see DPAC owns both CM and KPAC. DPAC can sell CM. DPAC cannot sell KPAC because, as the log shows, they just bought KPAC this round.

Screen Shot 2020-12-11 at 10 03 10 AM

### DIFF
--- a/assets/app/view/game/corporate_sell_shares.rb
+++ b/assets/app/view/game/corporate_sell_shares.rb
@@ -10,16 +10,15 @@ module View
 
       def render
         @step = @game.round.active_step
-        current_actions = @step.current_actions
-        @entity ||= @game.current_entity
-        children = []
+        @current_actions = @step.current_actions
+        @entity = @game.current_entity
 
-        children.concat(render_sources) if current_actions.include?('corporate_sell_shares')
-
-        h('div.margined', children.compact)
+        h('div.margined', render_sources)
       end
 
       def render_sources
+        return [] unless @current_actions.include?('corporate_sell_shares')
+
         @step.source_list(@entity).flat_map do |source|
           [
             h(Corporation, corporation: source),
--- a/assets/app/view/game/corporate_sell_shares.rb
+++ b/assets/app/view/game/corporate_sell_shares.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+require 'view/game/sell_shares'
+
+module View
+  module Game
+    class CorporateSellShares < Snabberb::Component
+      include Actionable
+
+      def render
+        @step = @game.round.active_step
+        current_actions = @step.current_actions
+        @entity ||= @game.current_entity
+        children = []
+
+        children.concat(render_sources) if current_actions.include?('corporate_sell_shares')
+
+        h('div.margined', children.compact)
+      end
+
+      def render_sources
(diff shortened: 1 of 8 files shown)


lib/engine/action/corporate_sell_shares.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require_relative 'sell_shares'
+
+module Engine
```
jenf, line 5: Is a separate action actually needed? The entity type says it's a corporation anyway

ryandriskel: Similar to Without a separate action class, the IssuesShares view is invoked ( see line 40 of view/game/round/operating.rb )

jenf:
In general looks okay, but will need a look by toby

assets/app/view/game/corporate_sell_shares.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module View
+  module Game
+    class CorporateSellShares < Snabberb::Component
+      include Actionable
+
+      def render
+        @step = @game.round.active_step
+        current_actions = @step.current_actions
+        @entity ||= @game.current_entity
```
tobymao, line 14: why are you doing @entity ||= @entity is not defined anywhere else right?

assets/app/view/game/corporate_sell_shares.rb
```diff
@@ -0,0 +1,35 @@
...
+        children = []
+
+        children.concat(render_sources) if current_actions.include?('corporate_sell_shares')
+
+        h('div.margined', children.compact)
```
tobymao, line 19: there's no reason to compact, you will never have nils, maybe you can try

h('div.margined', render_sources)

def render_sources
  return [] unless current_actions.include?

lib/engine/action/corporate_sell_shares.rb
tobymao: couldn't we remove CorporateBuyShares action by just checking a type specific to the round.

it seems strange to create a new action type just for one game when we already have actions for these things.

how clean or difficult is it to just use the existing buy_shares / sell_shares and just have a special flag for 18co

ryandriskel: buy_shares / sell_shares actions are already used in the Operating Round for the issue shares and redeem shares views. Therefore, we can't base the decision "by checking a type specific to the round" as far as I interpret your suggestion.

tobymao (approved):

PR #2618, opened 2020-12-12
[DRAFT] 1817WO: Add action to place second token when buying tokens

ventusignis:
Let me know how this looks and if there is an easier way than doing it this way

This is the last of the 1817WO rules needed to be implemented

### DIFF
--- a/assets/app/view/game/round/stock.rb
+++ b/assets/app/view/game/round/stock.rb
@@ -90,6 +90,7 @@ module View
             @selected_corporation.ipoed ? h(BuySellShares, corporation: @selected_corporation) : render_pre_ipo,
             render_loan,
             render_buy_tokens,
+            render_buy_and_place_tokens,
           ]
           if @step.actions(@selected_corporation).include?('buy_shares')
             inputs << h(IssueShares, entity: @selected_corporation)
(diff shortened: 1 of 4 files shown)


tobymao:
why does this need to be a new action and what is the exact rule?

ventusignis:
Nieuw Zeeland 

Players may select Nieuw Zeeland as an IPO starting in phase 3. Must choose 5 share or higher IPO. A company winning the bid places a token in Nieuw Zeeland. _After concluding paying for additional token(s) in the IPO, a 2nd token from the company is immediately placed anywhere on the map that has a free city, if there are any_. A company owning a token in Nieuw Zeeland earns +$20 from the bank per OR if it has no loans, or one interest payment for one loan is ignored per OR.

The goal of this PR is to to allow placing the second token. 

I originally tried to implement this by using a modification of the HomeToken step but I had trouble with getting it to be able to pass, so I went in this direction to let the user choose to, or not to, place a second home when buying tokens.

However after putting the rules here, I see a very important word.
After concluding paying for additional token(s) in the IPO, a 2nd token from the company **is** immediately placed anywhere on the map that has a free city, if there are any

I thought that said 'may be' when I was implementing this. I think I need to give this a second go; since this is compulsory, there is no need to allow passing

tobymao:
ok, yea lets get rid of it and just use the existing home token stuff or similar

ventusignis:
Closing this because it's easier to make a new branch

PR #2619, opened 2020-12-12
18CO President's Choice Round

ryandriskel:

**President's Choice Round**

This​ ​special​ ​stock​ ​round​ ​will​ ​occur​ ​once​ ​before​ ​the​ ​Stock​ ​Round​ ​following​ ​the​ ​first​ ​purchase​ ​of​ ​a​ ​5D​ ​or​ ​E​ ​train. Starting​ ​with​ ​the​ ​Priority​ ​Deal​ ​and​ ​moving​ ​clockwise,​ ​each​ ​player​ ​may​ ​purchase​ ​one​ ​Certificate​ ​from​ ​the​ ​Treasury​ ​of​ ​a Corporation​ ​for​ ​which​ ​they​ ​are​ ​President. Once​ ​a​ ​player​ ​passes,​ ​they​ ​may​ ​no​ ​longer​ ​purchase​ ​Certificates​ ​during​ ​this​ ​special​ ​stock​ ​round. Once​ ​all​ ​players​ ​pass,​ ​proceed​ ​to​ ​the​ ​normal​ ​Stock​ ​Round.​ ​All​ ​Certificates​ ​in​ ​the​ ​Treasuries​ ​of​ ​Corporations​ ​are​ ​now available​ ​for​ ​all​ ​players​ ​to​ ​purchase​ ​for​ ​the​ ​remainder​ ​of​ ​the​ ​game.

This PR also implements another feature from that list almost incidentally:

For 18CO, A player may purchase a third party corporation share from the treasury of any corporation where they are President.

### Implementation Notes

**buy_sell_shares.rb** - 1.) The change to using ipo_shares instead of shares is actually a bug fix. Corporate shares could be bought once the treasury was empty, but they were displayed as ipo shares on the button. 2.) Displaying the corporate shares will only impact games that utilize them.

**g_18_co/buy_sell_par_shares.rb** - This change actually _prevents_ a player from buying a corporate share unless they are president, as once I fixed the share/ipo_share display bug in _buy_sell_shares.rb_ I found that any players could buy any corporate share at any point in the game.

### In Action

Player 5 bought a share and thus stays in the round.
Player 1 was at cert limit and thus was forced to leave the round. This is automatic when there is no share a player can buy ( ie Skip ).
Player 2 is considering whether to buy that CM share from KPAC ( where he is President ).
Screen Shot 2020-12-12 at 7 09 18 AM

The PCR has finished and Player 5 resumes normal SR ( Priority does not change due to PCR ). Here we can see DPAC's CS share can now be bought ( _All​ ​Certificates​ ​in​ ​the​ ​Treasuries​ ​of​ ​Corporations​ ​are​ ​now available​ ​for​ ​all​ ​players​ ​to​ ​purchase​ ​for​ ​the​ ​remainder​ ​of​ ​the​ ​game._ )

Screen Shot 2020-12-12 at 7 15 08 AM

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -16,7 +16,7 @@ module View
         @step = @game.round.active_step
         @current_entity = @step.current_entity
 
-        @ipo_shares = @corporation.shares.group_by(&:percent).values
+        @ipo_shares = @corporation.ipo_shares.group_by(&:percent).values
           .map(&:first).sort_by(&:percent).reverse
 
         @pool_shares = @game.share_pool.shares_by_corporation[@corporation].group_by(&:percent).values
@@ -40,6 +40,7 @@ module View
 
         children.concat(render_ipo_shares)
         children.concat(render_market_shares)
+        children.concat(render_corporate_shares)
         children.concat(render_price_protection)
         children.concat(render_reduced_price_shares(@ipo_shares, source: @game.ipo_name(@corporation)))
         children.concat(render_reduced_price_shares(@pool_shares))
@@ -72,6 +73,20 @@ module View
             entity: @current_entity,
             percentages_available: @ipo_shares.size,
             source: @game.ipo_name(share.corporation))
+        end
+      end
+
+      def render_corporate_shares
+        @corporation.corporate_shares.group_by(&:corporation).values.flat_map do |corp_shares|
+          corp_shares.group_by(&:percent).values.map(&:first).sort_by(&:percent).reverse.map do |share|
+            next unless @step.can_buy?(@current_entity, share)
+
+            h(Button::BuyShare,
+              share: share,
+              entity: @current_entity,
+              percentages_available: @ipo_shares.size,
+              source: share.corporation.name)
+          end
         end
       end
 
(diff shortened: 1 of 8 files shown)


assets/app/view/game/buy_sell_shares.rb
```diff
@@ -75,6 +76,20 @@ def render_ipo_shares
         end
       end
 
+      def render_corporate_shares
+        @corporation.corporate_shares.group_by(&:corporation).values.flat_map do |corp_shares|
```
tobymao, line 80: won't this have nils in it?

ryandriskel: I don't think so as @corporation.corporate_shares will return an empty array if there are no shares held by corporations. The group_by and values calls then cannot return nils.

Could you be more specific about what "this" is that might have nils?

tobymao: next unless @step.can_buy?(@current_entity, share)

ryandriskel: So, the short answer is yes. However, so do the other share render (render_ipo_shares, render_market_shares) functions currently, as the results all get compacted in the render function.

PR #2620, opened 2020-12-12
Corporate Share Buying: Fix Undefined Buttons and Useless Displays

ryandriskel:
It was possible for a corporation or player to be displayed and have no possible inputs ( due to not being able to afford any shares, for example ). This change ensures that if there are no actions possible, that entity is not displayed.

Before Change:
DSL only had $134, and thus could not afford DPAC or KPAC shares
Screen Shot 2020-12-12 at 7 22 30 AM

After Change
Screen Shot 2020-12-12 at 7 51 48 AM


ryandriskel:
I was working on a branch of master so I don't understand why I ran into a conflict.

tobymao (approved):

Issue #2609, opened 2020-12-11
18CO Gameplay Issues


ryandriskel:
I'll be using this issue to document the items I find not working as I'm completing the features listed in https://github.com/tobymao/18xx/issues/1836

- [x] Corporate shares should pay dividends
- [x] 5D / E should be available once the first 6 is purchased ( This only happens when the corp just triggered the change. This might be global, ie the "available_on" doesn't quite work properly for any game )
- [x] Track lay changes on E train to single tile

Private Companies
- [x] GJGR should close when the ability is used ( town upgraded )
- [x] DNPR should close when the ability is used ( token returned )
- [x] LNPW should close when the ability is used ( tile laid )
- [x] DSNG should be removed if DRGR is not purchased in the auction

Emergency Money Raising:
- [x] Should buy cheapest train only, 
- [x] Should issue shares
- [x] Compute the train cost correctly

PR #2622, opened 2020-12-12
Fix Corporate Shares not paying dividends

ryandriskel:

Seems fairly self-explanatory. Corporate shareholders should be paid dividends. Before this change, the code only looks at players.

Here we see CM was paid $3 after the change.
Screen Shot 2020-12-12 at 8 39 51 AM

### DIFF
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -117,8 +117,10 @@ module Engine
         per_share = payout_per_share(entity, revenue)
 
         payouts = {}
-        @game.players.each do |player|
-          payout_entity(entity, player, per_share, payouts)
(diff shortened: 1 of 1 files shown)


lib/engine/step/dividend.rb
```diff
@@ -117,8 +117,10 @@ def payout_shares(entity, revenue)
         per_share = payout_per_share(entity, revenue)
 
         payouts = {}
-        @game.players.each do |player|
-          payout_entity(entity, player, per_share, payouts)
+        [@game.players, @game.corporations].flatten.each do |payee|
```
tobymao, line 120: (@game.players + @game.corporations).each

PR #2630, opened 2020-12-13
18CO Company Closings ( with bonus base corporation_close fix! )

ryandriskel:

 GJGR should close when the ability is used ( town upgraded )
 DNPR should close when the ability is used ( token returned )
 LNPW should close when the ability is used ( tile laid )
 DSNG should be removed if DRGR is not purchased in the auction

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -105,10 +105,6 @@ module Engine
         @imc ||= company_by_id('IMC')
       end
 
-      def dnp
-        @dnp ||= company_by_id('DNP')
-      end
-
       def setup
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_co.rb
```diff
@@ -213,6 +221,11 @@ def action_processed(action)
         case action
         when Action::BuyCompany
           mine_update_text(action.entity) if action.company == imc && action.entity.corporation?
+        when Action::RemoveToken
```
jenf, line 224: This should be done as part of a specialization for remove token.

jenf (approved):

PR #2633, opened 2020-12-13
fix typo when buying privates

benjaminxscott:
There was a missing space in the output logline


ryandriskel:
I think I broke this. Thanks for the fix!

ryandriskel (approved):

PR #2634, opened 2020-12-13
Fix base Emergency Buy + Issue Shares, add 18CO Implementation

ryandriskel:

Each commit is specific to a change.

**Refactor Emergency Selling Minimum Shares**
This is a readability refactor

**EMR Refactor Presidency Swap Check …**
Change player to seller, as if it is a share issue, the bundle owner will be the corporation, not a player. No need to check if Presidency changes since issuing shares to the market cannot trigger a presidency change.

**Fix train available_cash double counting …**
If the corp is issuing shares, this prevents double counting the corporation cash, which would prevent the corporation from ever being able to buy a train if there are issuable shares.

**Fix train can_sell? preventing issuing …**
An issue share is a share sale by the current entity (the corporation) and should not be prevented by the must_issue_before_ebuy

**Fix train_buy issue shares modifying president sales variables …**
last_share_sold_price and corporations_sold are only applicable to what the President has done, not the corporation.

**Refactor EMR issuable_shares …**
The default should be to return the game's emergency_issuable_bundles for the corporation. All games that don't allow EMR issuing aren't going to implement that anyway and thus it will still return []

**18CO EMR Don't show issue options if can already afford a train**
Before this change the Issue Share buttons would show, but not work, since the player did not need to raise money anymore.

**Only show information about issuing if there are shares to issue**

The text "DPAC can issue shares to raise..." was showing regardless of whether there were shares to issue.
Screen Shot 2020-12-13 at 11 05 37 AM

After the change the paragraph no longer renders since it is irrelevant
Screen Shot 2020-12-13 at 11 05 25 AM

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -133,7 +133,9 @@ module View
         children << remaining_trains
 
         children << h(:div, "#{@corporation.name} has #{@game.format_currency(@corporation.cash)}.")
-        if (issuable_cash = @game.emergency_issuable_cash(@corporation)).positive?
+
+        if step.issuable_shares(@corporation).any? &&
+           (issuable_cash = @game.emergency_issuable_cash(@corporation)).positive?
           children << h(:div, "#{@corporation.name} can issue shares to raise up to "\
                               "#{@game.format_currency(issuable_cash)} (the corporation "\
                               'must issue shares before the president may contribute).')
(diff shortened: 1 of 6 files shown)


ryandriskel:
Verified that after all the changes, the corporation can issue all shares and then if still not enough cash, have the normal president contribution.

Screen Shot 2020-12-13 at 12 24 59 PM

Screen Shot 2020-12-13 at 12 25 08 PM

tobymao (approved):

tobymao:
@ryandriskel can you check out these broken games?

15713, 18MS, finished, Cannot sell shares of IC
17230, 18MS, finished, Cannot sell shares of WRA
17492, 18MS, finished, Cannot sell shares of L&N
17510, 18MS, finished, Cannot sell shares of WRA
15628, 18MS, finished, Cannot sell shares of L&N

Issue #2477, opened 2020-11-28
Better use of the word "pass" in the log


ChrisTheCat:
In the stock round, it would be clearer if the word "pass" only referred to an actual pass that changes priority. In this example, after I bought the NdM share, the button said "Done (share)" but the log says this:

> ChrisShaffer buys a 10% share of NdM from the market for $75
> ChrisShaffer passes selling shares

I suggest it would be better if it said something like this:

> ChrisShaffer buys a 10% share of NdM from the market for $75
> ChrisShaffer finished stock action

Or perhaps:

> ChrisShaffer buys a 10% share of NdM from the market for $75
> ChrisShaffer declined to sell shares

PR #2635, opened 2020-12-13
Issue 2477 Clarify Passing your action vs Ending your action

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/2477

Using the word decline instead of pass more clearly indicates that an action was still taken that will affect priority deal rather than passing your action.

### DIFF
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -30,12 +30,8 @@ module Engine
       def log_pass(entity)
         return @log << "#{entity.name} passes" if @current_actions.empty?
 
-        action = if bought?
-                   'selling'
-                 else
-                   'buying'
-                 end
-        @log << "#{entity.name} passes #{action} shares"
+        action = bought? ? 'to sell' : 'to buy'
+        @log << "#{entity.name} declines #{action} shares"
       end
 
       def log_skip(entity)


lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -30,12 +30,8 @@ def actions(entity)
...
         return @log << "#{entity.name} passes" if @current_actions.empty?
 
-        action = if bought?
-                   'selling'
-                 else
-                   'buying'
-                 end
-        @log << "#{entity.name} passes #{action} shares"
+        action = bought? ? 'selling' : 'buying'
+        @log << "#{entity.name} declines #{action} shares"
```
jenf, line 34: declines to buy/sell shares

jenf (changes requested):

lib/engine/step/buy_sell_par_shares.rb
ryandriskel: I didn't write the logic here, just changed the word, but I'm not sure it should be changed.

If you bought before hitting Done ( log would indicate "Entity declines selling shares" ), you no longer had the option to buy, and thus you declined selling ( not buy/sell )

If you sold before hitting Done ( log would indicate "Entity declines buying shares" ), then saying you declined selling (via "declines buy/sell shares") is not entirely accurate.

jenf: Sorry, what I mean is "Entity declines selling shares" is bad English. "Entity declines to sell shares"

ryandriskel: Well then, that I can do!

jenf (approved):

Issue #2597, opened 2020-12-09
Master mode showing player took actions when they didn't


michaeljb:
![Screenshot from 2020-12-08 17-57-44](https://user-images.githubusercontent.com/1045173/101559714-ed032180-397e-11eb-9d40-7383a14ad05d.png)

In https://18xx.games/game/14927#tools, the browser find tells me there are 633 instances of `"user"` on in the game JSON, with 632 of them being `"user": 136`; the on other being `"user": {` at the top of the JSON, so the system reports that 632 of the 904 actions were taken by user 136 (Stephen_Yu).

michaeljb:
May be caused by migrations - https://github.com/tobymao/18xx/blob/master/migrate_game.rb#L314

michaeljb:
I think that needs to be changed to `action.user` or something, and we could do a sweep for games where the owner is marked as the user on the action an unusually high amount of times and remove those so it doesn't look like they master moded through the whole game.

tobymao:
this is not a bug in the engine, this is a bug in the migration script

tobymao:
is this fixed?

michaeljb:
Yeah, fixed in https://github.com/tobymao/18xx/commit/160d2ec478944576e865711e32f1ce1a1aa8967e#diff-ec939cb63abf76c70afca5bcb759020a2e3dfe6149f1f8b2e3a113801c2966cdR331 and https://github.com/tobymao/18xx/commit/07219f62af13d407ceb0124faa16309075e6ca6c#diff-ec939cb63abf76c70afca5bcb759020a2e3dfe6149f1f8b2e3a113801c2966cd

Note that any future migrations that create new actions also need to appropriately set `user` on that action

PR #2636, opened 2020-12-13
1817 Bugs:

### LINKED ISSUE #2605: 1817: Bankrupt player expected to bid or pass on company acquisitions
https://18xx.games/game/14464?action=1889

1) It's funny because the game is 1817, the action is 1889, but in my region, we call bringing back a player from the dead to do your bidding "Necromancy."

2) A player went bankrupt through no obviously-horrible decisionmaking of their own, truly just an unfortunate soul making only the best decisions yet the slings and arrows of outrageous fortune still conspired to bring them down.  It happens and we're not here to point fingers.  However, the player (with $0) was still prompted to bid on companies in the Acquisition Zone.

This isn't the Twilight Zone, and players with only companies that have $0 buying power or 0 companies total, should be automatically skipped in the Acquisition phase, just like players with 0 companies get skipped in the Merger phase.

It's not a gamebreaking issue, but it's a nice "Quality of Life" or "Peace in Death" thing, and it would make 1817 go a bit less-slow in the M&A rounds.  Thanks!

### LINKED ISSUE #2606: [1817] Company 100+% sold only lowers in value based on market shares (does not first increase value for 100% in player hands)
**Describe the bug**

e.g. "GT's share price changes from $120 to $80" yet there is 120% of the shares in players hands so it fails to increase the value first and then go back four spaces to 80$.

https://18xx.games/hotseat/hs_ueysilmu_17862

jenf:
Sold out should disregard shorts (fixes #2606, but mass pinnning needed)
Bankrupt player shouldn't take part in acquisitions (fixes #2605, migrateable)

### DIFF
--- a/lib/engine/round/g_1817/stock.rb
+++ b/lib/engine/round/g_1817/stock.rb
@@ -25,7 +25,7 @@ module Engine
         end
 
         def sold_out?(corporation)
-          corporation.total_shares > 2 && corporation.player_share_holders.values.sum >= 100
+          corporation.total_shares > 2 && corporation.player_share_holders.values.select(&:positive?).sum >= 100
         end
       end
     end
--- a/lib/engine/step/g_1817/acquire.rb
+++ b/lib/engine/step/g_1817/acquire.rb
@@ -79,7 +79,13 @@ module Engine
                 end
 
               players = @game.players.rotate((@game.players.index(owner) + 1) % @game.players.size)
-              [players.find { |p| @active_bidders.include?(p) }]
+              player = players.find { |p| @active_bidders.include?(p) }
+              if player.bankrupt
+                pass_auction(player)
+                active_entities
+              else
+                [player]
+              end
             end
           elsif @buyer
             [@buyer]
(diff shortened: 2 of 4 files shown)


michaeljb:
With 1817 in production mode, we definitely want #2597 sorted out before applying the migrations for this. I'll see if I can figure out anything for that this evening.

tobymao (approved):

michaeljb:
With the changes on this branch, I'm seeing the pass actions from bankrupt player still get processed (which means no migration at least):

{EE08CECD-5C75-4654-BCAA-E9278C2610DE}

https://gist.github.com/michaeljb/a9b7c97eb5eb8e8493f8e0f075f73e31

Adding `return [] if entity.player? && entity.bankrupt` to the top of `Step::G1817::Acquire#actions` prevents the bankrupt player from being allowed to take actions. With that line added however, the migration fails (at least for game 13678)

jenf:
I'll rework it as suggested by @michaeljb and fix the migration

jenf:
Actually avoiding the need for migration is probably worth it.

PR #2649, opened 2020-12-15
1860 endgame - feature complete for 1860

### LINKED ISSUE #2451: 1860: bank break check not correct
Because of the way 1860's bank breaks, G1860::Bank.check_cash doesn't work correctly. It has a problem:
- It always checks, even if it isn't a player doing the spending, so that it will incorrectly detect a break when a corp is spending

The solution is to neuter the check_cash method and instead have the dividend and stock steps check the bank breaking explicitly. These are the only place it needs to be checked since those are the only places players receive cash from the bank.

roseundy:
Last set of changes to make 1860 feature complete:

Game now ends on
- Bank breaking
- Corp reaching rightmost stock price
- Nationalization

Implements 8+4 event
Implements 9+5 event - triggering Southern formation
Implements Nationalization

Moves player value calculation into Game::base

Common file changes:
- UI::Player - now calls @game.player_value(player)
- UI::Spreadsheet - now calls @game.player_value(player)
- Game::Base - implements player_value(player), modified call to PlayerInfo
- Engine::PlayerInfo - takes player value directly from calling parameters

Fixes #2451

### DIFF
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -100,7 +100,7 @@ module View
         trs.concat([
           h(:tr, [
             h(:td, 'Value'),
-            h('td.right', @game.format_currency(@player.value)),
+            h('td.right', @game.format_currency(@game.player_value(@player))),
           ]),
           h(:tr, [
             h(:td, 'Liquidity'),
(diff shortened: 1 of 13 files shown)


tobymao (approved):
ready for alpha?

PR #2656, opened 2020-12-15
Add support for hills and rough terrain

tysen:
1849 distinguishes between these three


perwestling:
I was curious how this would look but could not find any rough/hill/mountain icons in the commit.
So I tried to create a hotseat 1849 using the production code. Did not work, got an error Undefined metod 'description' for nil.
render_game -> render_round -> description

tysen:
@perwestling This just re-uses the mountain icon for hills/rough for now. Thanks for reporting the error. I'll see if I can duplicate it but the 1849 code is nascent and changing a lot right now, especially on my machine.

tysen:
Not sure this is necessary right now.

PR #2661, opened 2020-12-15
1867: First pass on merger

jenf:
Mostly working, needs token reduction and share purchasing

### DIFF
--- a/lib/engine/step/g_1867/merge.rb
+++ b/lib/engine/step/g_1867/merge.rb
@@ -7,14 +7,18 @@ module Engine
   module Step
     module G1867
       class Merge < Base
+        LIMIT_OWNED_BY_ONE_ENTITY = 6
+        LIMIT_MERGE = 10
+
         def actions(entity)
           return [] if !entity.corporation? || entity != current_entity
           return [] if @round.converted
 
           actions = []
 
-          return ['merge'] if @converting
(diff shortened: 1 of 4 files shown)


lib/engine/step/g_1867/merge.rb
```diff
@@ -92,14 +101,100 @@ def move_assets(from, to)
...
+            owner = corporation.owner
+            owners << owner
+            @game.close_corporation(corporation)
+
+            share = target.shares.last
+
+            if target.shares.first.president && owner.percent_of(target) == 10
+              # give the 10% back
+              presidency = target.shares.first
+              @game.share_pool.move_share(owner.shares_of(target).first, presidency.owner)
```
ryandriskel, line 147: Seems there are a number of ways to do this. In 18CO, I used `share.transfer(...)`. When should we use `@game.share_pool` instead?

tobymao: not sure, can both accomplish the same thing?

jenf: i can rework this later, i'll need to do the later on the case that people are buying shares when they're at 10% anyway

jenf: I remember, I added it for 1817 to avoid making that public :)

Issue #2665, opened 2020-12-16
1817 cannot cross-player merge 5share corporations with bare presidencies


ventusignis:
It looks like in the effort to prevent presidentless corporations being created from mergers (IE: shorts) the implementation of 1817 prohibits mergers where the new president will have only the bare presidency.

It looks like the `>` should be `>=` here in `lib/engine/step/g_1817/conversion.rb`
`            (target.total_shares != 5 || merged_max_share_holder(corporation, target) > 40) &&`

jenf:
Resolved by #2666

PR #2666, opened 2020-12-16
Allow mergers in 1817 that end in the president holding 20%

ventusignis:
Implementation of the change suggested in https://github.com/tobymao/18xx/issues/2665


jenf:
Will cause some companies to now be mergable, so will need a pin/migration

tobymao (approved):

PR #2667, opened 2020-12-16
[1817, 1817WO] Ohio Bridge Company should only be usable once

### LINKED ISSUE #2664: [1817] Ohio Bridge Company allows to place two bridge tokens
[https://18xx.games/game/14443?action=1017](url)

I was able to place a second bridge token. The first one was placed at the 1st OR at Charleston. If you see the action on the link, I have just placed the second token at Cincinnati. I should not be able to place more than one.

I don't know if I can use the Ohio Bridge Company while operating a different company than the owner. This could be a second issue with the rules, but I am not sure of this one.

benjaminxscott:
fixes #2664

### DIFF
--- a/lib/engine/config/game/g_1817.rb
+++ b/lib/engine/config/game/g_1817.rb
@@ -183,7 +183,7 @@ module Engine
       "name" : "Ohio Bridge Company",
       "value" : 40,
       "revenue" : 0,
-      "desc" : "Comes with one $10 bridge token that may be placed by the owning corp in Louisville, Cincinnati, or Charleston, max one token per city, regardless of connectivity..  Allows owning corp to skip $10 river fee when placing yellow tiles.",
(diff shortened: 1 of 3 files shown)


ventusignis:
You beat me by 5 minutes to it!

lib/engine/config/game/g_1817.rb
```diff
@@ -224,6 +226,7 @@ module G1817
             "H9"
           ],
           "count": 2,
+          "when": "owning_corp_or_turn",
```
jenf, line 229: Shouldn't this be done in 1817NA?

ventusignis: Yes. I'll submit a separate PR so this one doesn't get held up

jenf: it'd be easier to do one just in case something needs pinning

benjaminxscott: pushed

benjaminxscott:
@jenf - it seems like at least the [game from the issue](https://18xx.games/game/14443?action=1017) would be pinned, since its current state is invalid due to the double Ohio bridge usage, possibly other games as well

jenf (approved):

PR #2677, opened 2020-12-16
Fix interest discount in 1817WO being applied to every corp

ventusignis:
This small fix got lost in code cleanup yesterday


ventusignis:
Without this all corporations get the new zealand bonus

jenf (approved):

PR #2686, opened 2020-12-17
18CO Remove City Reservations, Close Corp w/o Homes, Re-reserve Newly Parred Corps

ryandriskel:

From phase 5 onward, cities are no longer reserved.
If a corporation can never place its home token, it is removed from the game.
Once a corporation is parred, its reservation is returned to the hex.
A hex might be a single city, but the corporation may upgrade it to place its home token.

**Implementation Notes**

Tile label for Colorado Springs changed from **CS** to **C** to avoid confusion over the Corporation CS if it needs re-reservation.

I feel like once again I have a side effect inside a ? function, similar to the takeover. I don't know how to avoid that and still get the nice automation this provides. Its the same kind of reason: its the Home Token placement step, and if it can find a suitable automatic placement (in this case when a tile placement opens up a slot), it executes it automatically, otherwise it prompts the player to make the choice.

### In Action

CS was not parred. CM placed a token on CS home hex once "closable corporations" was in play. Later, CS was parred, placing a reservation on the yellow hex (rereserve_home_station), as it can be upgraded to green, which has two slots.
Screen Shot 2020-12-16 at 2 52 43 PM

In the next OR, CM considered upgrading the tile to green. The reservation moves from the tile to the city ( migrate_reservations function ).
Screen Shot 2020-12-16 at 10 20 42 PM

CM undid its tile lay. When it became CS's turn to operate, HomeToken was not active at first because there was no available slot, so CS proceeded to Track. Its only option was to lay the green tile. This triggered the migrate_reservations, which assigned a slot, and thus HomeToken's multiple_home_slot ran and placed the token in the newly added slot automatically.
Screen Shot 2020-12-16 at 11 09 20 PM

In this SR, CS and DPAC were parred. DPAC's station was still open, so it doesn't have to worry about placing track on its first OR, while CS must upgrade to place its token.
Screen Shot 2020-12-16 at 11 36 19 PM

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -50,6 +50,7 @@ module Engine
       GREEN_TOWN_TILES = %w[co8 co9 co10].freeze
       GREEN_CITY_TILES = %w[14 15].freeze
       BROWN_CITY_TILES = %w[co4 63].freeze
+      MAX_CITY_TILES = %w[14 15 co1 co2 co3 co4 co7 63].freeze
 
       STOCKMARKET_COLORS = {
         par: :yellow,
@@ -91,11 +92,20 @@ module Engine
           'presidents_choice' => [
             'President\'s Choice Triggered',
             'President\'s choice round will occur at the beginning of the next Stock Round',
+          ],
+          'unreserve_home_stations' => [
+            'Remove Reservations',
+            'Home stations are no longer reserved for unparred corporations.',
           ]
         ).freeze
 
       STATUS_TEXT = Base::STATUS_TEXT.merge(
-        'reduced_tile_lay' => ['Reduced Tile Lay', 'Corporations place only one tile per OR.']
+        'reduced_tile_lay' => ['Reduced Tile Lay', 'Corporations place only one tile per OR.'],
+        'closable_corporations' => [
+          'Closable Corporations',
+          'Unparred corporations are removed if there is no station available to place their home token. '\
+          'Parring a corporation restores its home token reservation.',
+        ]
       ).freeze
 
       include CompanyPrice50To150Percent
@@ -191,7 +201,7 @@ module Engine
         Step::Bankrupt,
         Step::G18CO::Takeover,
         Step::DiscardTrain,
-        Step::HomeToken,
+        Step::G18CO::HomeToken,
         Step::G18CO::ReturnToken,
         Step::BuyCompany,
         Step::G18CO::RedeemShares,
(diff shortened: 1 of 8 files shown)


lib/engine/game/g_18_co.rb
```diff
@@ -266,9 +276,44 @@ def action_processed(action)
...
+          rereserve_home_station(action.corporation) if @phase.status.include?('closable_corporations')
         end
       end
 
+      def remove_corporations_if_no_home(city)
+        tile = city.tile
+
+        return unless tile_has_max_cities(tile)
+
+        @corporations.dup.reject(&:ipoed).each do |corp|
```
tobymao, line 291: there's no reason to call dup, reject already does that

tobymao: or if you're already using the each next pattern, then do

dup.each do 

next if ipoed

lib/engine/game/g_18_co.rb
```diff
@@ -357,6 +402,12 @@ def event_remove_mines!
         end
       end
 
+      def event_unreserve_home_stations!
+        @log << '-- Event: Home station reservations removed --'
+
+        hexes.map { |h| h.tile.cities.map(&:remove_all_reservations!) }
```
tobymao, line 408: why are you calling map? should this be an each?

lib/engine/step/g_18_co/tracker.rb
```diff
@@ -77,6 +79,25 @@ def all_new_paths_have_new_exits?(brand_new_exit_junctions)
           brand_new_exit_junctions.size >=
             brand_new_exit_junctions.sum { |_bnp, new_exit_junctions| new_exit_junctions.size }
         end
+
+        def pending_token(entity)
+          @round.pending_tokens.find { |pt| pt[:entity] == entity }
```
tobymao, line 84: doesn't look like you use pending_token anywhere, you should switch to any? instead of find

lib/engine/tile.rb
```diff
@@ -281,7 +281,7 @@ def reserved_by?(corporation)
 
     def add_reservation!(entity, city, slot = 0)
       # Single city, assume the first
-      city = 0 if @cities.one?
+      city = 0 if @cities.one? && available_slot?
```
ryandriskel, line 284: I have an itch on this change. The point is that if the city is full, you should place the reservation on the tile.

Without this change, its possible to have a reservation on the same slot as a token and actually replace that token.

I'm wondering if instead it would be better to make a slightly more verbose function.

```
def add_reservation!(entity, city, slot = nil)
  city = 0 if @cities.one?
  slot = @cities[city].get_slot(entity) if slot.nil?

  if city && slot
    @cities[city].add_reservation!(entity, slot)
  else
    @reservations << entity
  end
end

File: lib/engine/step/g_18_co/tracker.rb (hunk 3 above)
Comment (ryandriskel) [reply lib/engine/step/g_18_co/tracker.rb:84]: Tracker is an include and is used in those files.

File: lib/engine/tile.rb (hunk 4 above)
Comment (tobymao) [reply lib/engine/tile.rb:284]: doesn't that change behavior since you're no longer calling available_slot?

prefer unless slot over if slot.nil?

Comment (ryandriskel) [reply lib/engine/tile.rb:284]: Aside: So far this change works fine and none of the tests are breaking. It just has a bad smell to me.

Comment (tobymao) [reply lib/engine/tile.rb:284]: why not do slot = available_slot unless slot

Comment (ryandriskel) [reply lib/engine/tile.rb:284]: > doesn't that change behavior since you're no longer calling available_slot?

The later if statement `if city && slot` handles whether there is a slot just as the PR does, but it _seems_ improved in a multiple city case. It may be irrelevant as this may never come up.

File: lib/engine/part/city.rb [hunk 5]
```diff
@@ -61,6 +61,10 @@ def remove_reservation!(entity)
         @reservations.delete(entity)
       end
 
+      def remove_all_reservations!
+        @reservations = []
```
ryandriskel, line 65: Would it be better to use `.clear` here?

lib/engine/game/g_18_co.rb
```diff
@@ -357,6 +403,12 @@ def event_remove_mines!
         end
       end
 
+      def event_unreserve_home_stations!
+        @log << '-- Event: Home station reservations removed --'
+
+        hexes.each { |h| h.tile.cities.map(&:remove_all_reservations!) }
```
tobymao, line 409: you did'nt resolve this this shouldn't be an each

ryandriskel: Argh, I changed the one but not the other.

ryandriskel: They should _both_ be each, shouldn't they?

tobymao (approved):

PR #2698, opened 2020-12-17
1856 variable cap

ventusignis:
1856 corporations are full cap, incremental cap, or escrow based on the facing train at presidency purchase

### DIFF
--- a/lib/engine/g_1856/corporation.rb
+++ b/lib/engine/g_1856/corporation.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require_relative '../corporation'
+
+module Engine
+  module G1856
+    class Corporation < Corporation
+      def initialize(game, sym:, name:, **opts)
+        @game = game
+        @started = false
+        super(sym: sym, name: name, **opts)
+      end
+
+      # ~Ab~RE-using floated? to represent whether or not a corporation has operated
+      def floated?
+        @started
+      end
+
+      def floatable?
+        percent_of(self) <= 100 - percent_to_float
+      end
+
+      def float!
+        @started = true
+      end
+
+      def par!
+        @capitalization = capitalization_type
+      end
+
+      def capitalization_type
+        # TODO: escrow
+        return :incremental if @game.phase.status.include? :escrow
+        return :incremental if @game.phase.status.include? :incremental
+        return :full if @game.phase.status.include? :fullcap
+
+        # This shouldn't happen
+        raise NotImplementedError
+      end
+
+      def percent_to_float
+        return 20 if @game.phase.status.include? :facing_2
+        return 30 if @game.phase.status.include? :facing_3
+        return 40 if @game.phase.status.include? :facing_4
+        return 50 if @game.phase.status.include? :facing_5
+        return 60 if @game.phase.status.include? :facing_6
+
+        # This shouldn't happen
+        raise NotImplementedError
+      end
+    end
+  end
+end
--- a/lib/engine/g_1856/corporation.rb
+++ b/lib/engine/g_1856/corporation.rb
@@ -28,6 +28,19 @@ module Engine
         @capitalization = capitalization_type
       end
 
+      def capitalization_type_desc
+        case @capitalization
+        when :full
+          'Full'
+        when :incremental
+          'Incremental'
+        when :escrow
+          'Escrow'
+        else
+          raise NotImplementedError
+        end
+      end
+
       def capitalization_type
         # TODO: escrow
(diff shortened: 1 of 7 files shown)


ventusignis:
image
image

lib/engine/g_1856/corporation.rb
```diff
@@ -0,0 +1,88 @@
...
+        return :incremental if @game.phase.status.include? :incremental
+        return :full if @game.phase.status.include? :fullcap
+
+        # This shouldn't happen
+        raise NotImplementedError
+      end
+
+      # As long as this is only used in core code for display we can re-use it
+      def percent_to_float
+        return 20 if @game.phase.status.include? :facing_2
```
tobymao, line 77: you parens

tobymao: use

lib/engine/g_1856/corporation.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require_relative '../corporation'
```
tobymao, line 3: is it really necessary to have a new corp type? can these methods just be on game?

lib/engine/g_1856/share_pool.rb
```diff
@@ -0,0 +1,149 @@
+# frozen_string_literal: true
+
+require_relative '../share_pool'
+
+module Engine
+  module G1856
+    class SharePool < SharePool
+      def buy_shares(entity, shares, exchange: nil, exchange_price: nil, swap: nil)
```
tobymao, line 8: this looks copy and pasted what did you change?

lib/engine/step/g_1856/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,27 @@
...
+  module Step
+    module G1856
+      class BuySellParShares < BuySellParShares
+        def process_par(action)
+          share_price = action.share_price
+          corporation = action.corporation
+          entity = action.entity
+          @game.game_error("#{corporation} cannot be parred") unless corporation.can_par?(entity)
+
+          corporation.par!
```
tobymao, line 15: what's the point of overriding this? it looks 99% copy pasted

lib/engine/g_1856/share_pool.rb
ventusignis: Lines 62-63 and 104 are changed
`buy_shares` treats `:escrow` corporations similarly to `:incremental` capitalization corporations when invoking `transfer_shares`

`transfer_shares` invokes `escrow_share_buy!` on the corporation if appropriate, this relates to the response to your comment on `corporation.rb` I have yet to finish

lib/engine/step/g_1856/buy_sell_par_shares.rb
ventusignis: Overriding it lets me call `par!` on the corporation in the middle of the method, I'll refer to this in my response to your comment on corporation.rb

lib/engine/g_1856/corporation.rb
ventusignis: I'm not super familiar with ruby so it's completely possible there are things that Ruby supports that I don't realize yet. It seems like the only way to inject the logic into the corporation would be to try to pass methods instead of values as some of the options or to monkey patch the Corporation class

Passing methods instead of values (such as `float_percent`) aren't flexible enough to work without making significant modifications to Corporation and putting defaults in the Base game and at that point I believe it would make it more complex overall.

I'm not comfortable with casually using monkey patching in general and in this case I don't know how that would affect the server side or client side code; I'm not sure if it could cause side effects with other games. 

That's why I've kept it as a game-specific subclass and some of the other subclasses and overrides that you've commented on elsewhere are done to support use of this game-specific class

There is an argument that the escrow functionality can be moved into the base class and I agree that it can be done, but I don't think it's time yet and that it'd be better to wait until we have another game or two that also uses "escrow" accounts for corporations to implement that in a general way. 
I know that 18SY (Original rules) keeps *shares* in escrow which get redeemed at market value at time of escrow release, so a static escrow balance wouldn't work there

lib/engine/g_1856/corporation.rb
ventusignis: Fixed that here and below

lib/engine/step/g_1856/buy_sell_par_shares.rb
ventusignis: The class will be needed for implementing vesting later as well

tobymao (approved):

PR #2700, opened 2020-12-17
Generalize 1817 token reduction code and add to 1867

jenf:
Also fix repaying of loans for 1867


lib/engine/step/token_merger.rb
```diff
@@ -0,0 +1,72 @@
...
+        others.flat_map(&:tokens)
+      end
+
+      def remove_duplicate_tokens(surviving, others)
+        # If there are 2 station markers on the same city the
+        # surviving company must remove one and place it on its charter.
+        # In the case of NY tiles this is ambigious and must be solved by the user
+
+        others = others_tokens(others).map(&:city).compact
+        surviving.tokens.each do |token|
```
tobymao, line 28: do you need to do a dup here? dose remove! mutate tokens?

tobymao (approved):

lib/engine/step/token_merger.rb
jenf: huh?

jenf: what dup? Sorry unsure what you mean

tobymao: surviving.tokens.dup.each do 

because you do a remove right?

jenf: oh, i see what you mean. no the token.remove! just un-places the token. (This is exactly the same as the original 1817 code)

PR #2704, opened 2020-12-18
Hex trains for 1849

tysen:


lib/engine/game/g_1849.rb
```diff
@@ -39,6 +39,41 @@ class G1849 < Base
...
+        when types.include?(:broad)
+          :broad
+        else
+          :dual
+        end
+      end
+
+      def hex_edge_cost(conn, train)
+        track = track_type(conn.paths)
+        edges = conn.paths.each_cons(2).sum do |pair|
```
tobymao, line 57: you could do |a, b|

lib/engine/config/game/g_1849.rb
```diff
@@ -473,19 +473,120 @@ module G1849
...
   "corporations": [
   ],
   "trains": [
+    {
+      "name": "4H",
+      "num": 4,
+      "distance": 99,
+      "price": 100,
+      "rusts_on": "8H",
+      "hex_edge_limit": 4
```
tobymao, line 543: can we change this to

"distance": { "hex": 4 }

tobymao:
merge conflicts

lib/engine/config/game/g_1849.rb
jenf: +1, I think we should also add the R6H type here somehow

lib/engine/game/g_1849.rb
jenf: I'd also suggest putting this into a common hex_train.rb class as 18Ireland will use it as well as other hex based games in the future.

lib/engine/config/game/g_1849.rb
jenf: "distance": {"hex": 4, "terrain_cost": {"broad": 2}}
or something similar.

tysen: I considered this but you get a lot of mileage out of having `distance` be a numeric, and a lot more code would have to be changed to support hex-typed `distance` than how I've done it. Also, I'm not sure I see the benefit of allowing players to select their routes hex by hex - ultimately revenue centers are what matter for a train run.

lib/engine/game/g_1849.rb
tysen: I originally had a new `Train` class but trains don't do any of this computation - games and routes do.

lib/engine/config/game/g_1849.rb
tobymao: which part needs to change? distance is already not numeric in some games

i don't think we're suggesting allowing players to select routes hex by hex

tysen: Sorry, I should have been more clear. My primary concern is with `compute_stops` which expects that `distance` be mostly about revenue centers, and uses `route.visited_stops` (again, concerned with revenue centers) in its computation. wrt selecting routes hex by hex - of course, no, but that's the ad absurdum result of redefining e.g. `Connection` (used in `visited_stops`) to include hexes.

tobymao: if you change distance to be { 'hex': 4 }  you could  just check distance.is_a?(Numeric) || distance['hex']

if you don't like this perhaps then change 'hex_limit': true and keep distance a real number

tysen: I took your second suggestion. I don't see a lot of value in specifying this further in config given that the few games that have hex trains and narrow gauge all use them quite differently:

- 18Ireland: Trains can't run on narrow gauge.
- 1858: Has 4 different kinds of train only one of which counts hexes and only one of which uses narrow gauge (they're not the same)
- 1849: As implemented here.

jenf: yeah probably the case.

lib/engine/train.rb
```diff
@@ -23,6 +23,7 @@ def initialize(name:, distance:, price:, index: 0, **opts)
       @available_on = opts[:available_on]
       @discount = opts[:discount]
       @multiplier = opts[:multiplier]
+      @hex_edge_limit = opts[:hex_edge_limit]
```
tobymao, line 26: these are not eneded anymore rgiht, yo ucan cleanup this file and we can merge

tysen: Oops, silly. Done.

PR #2705, opened 2020-12-18
Terrain costs for 1849

tysen:
Transfer upgrade cost to newly placed tiles, charge discounted rates for
narrow and broad-placed-while-owning-P2


lib/engine/game/g_1849.rb
```diff
@@ -49,6 +49,57 @@ def event_earthquake!
...
+                               Step::BuyTrain,
+                               [Step::BuyCompany, blocks: true],
+                             ], round_num: round_num)
+      end
+
+      def new_track(old_tile, new_tile)
+        # Assume path retention checked elsewhere
+        old_track = old_tile.paths.map(&:track)
+        added_track = new_tile.paths.map(&:track)
+        old_track.each { |t| added_track.slice!(added_track.index(t) || added_track.length) }
```
tobymao, line 73: always use size as opposed to length

lib/engine/game/g_1849.rb
```diff
@@ -49,6 +49,57 @@ def event_earthquake!
...
+          added_track.include?(:broad) ? :broad : :narrow
+        end
+      end
+
+      def tile_cost(tile, _hex, entity, new_tile: nil)
+        return 0 if tile.upgrades.empty?
+
+        upgrade = tile.upgrades[0]
+        track = new_track(tile, new_tile)
+        case track
```
tobymao, line 86: simply to 

case new_track(...)

lib/engine/game/g_1849.rb
```diff
@@ -49,6 +49,57 @@ def event_earthquake!
...
+        when :dual
+          upgrade.cost
+        when :narrow
+          @log << "#{entity.name} pays 1/4 cost for narrow gauge track"
+          upgrade.cost / 4
```
tobymao, line 91: is there rounding? care with division as js and ruby have different interpretations

lib/engine/part/path.rb
```diff
@@ -61,7 +61,18 @@ def <=>(other)
 
       def <=(other)
         other_ends = other.ends
-        ends.all? { |t| other_ends.any? { |o| t <= o } }
+        ends.all? { |t| other_ends.any? { |o| t <= o } } && tracks_match(other.track)
+      end
+
+      def tracks_match(other)
```
tobymao, line 67: use constants, i'm afraid this is a hot loop and will cerate too many objects

lib/engine/game/base.rb
```diff
@@ -1034,7 +1034,7 @@ def place_home_token(corporation)
         city.place_token(corporation, token)
       end
 
-      def tile_cost(tile, hex, entity)
+      def tile_cost(tile, hex, entity, new_tile: nil) # rubocop:disable Lint/UnusedMethodArgument
```
tobymao, line 1037: was new_tile really necessary? can't you just do hex.tile?

lib/engine/game/g_1849.rb
tysen: No rounding - base costs are 160, 80, and 40

lib/engine/game/base.rb
tysen: Good point. Didn't realize `hex.lay` had already been called by now.

tobymao (approved):

PR #2708, opened 2020-12-18
Force repayment of excess loans in 1856

ventusignis:


lib/engine/round/g_1856/operating.rb
```diff
@@ -25,9 +26,45 @@ def after_process(_action)
...
 
+        # nil if the entity doesn't need to pay off loans, # otherwise
+        def must_payoff?(entity)
+          loans_to_repay = entity.loans.size - @game.maximum_loans(entity)
+          loans_to_repay.positive? ? loans_to_repay : nil
+        end
+
+        def force_repay_loans!(entity)
+          @cash_crisis_due_to_forced_repay = nil
+          return unless @steps.any? { |step| step.passed? && step.is_a?(Step::BuyTrain) } && must_payoff?(entity)
```
ryandriskel, line 41: Calling `must_payoff?(entity)` several times even though you set `loans_to_payoff = must_payoff?(entity)`

Maybe call it once before anything else?

ventusignis: It's only used in `force_repay_loans!` so I'll just remove it

tobymao (approved):

PR #2720, opened 2020-12-18
18CO - Fix when the DSNG closes

ryandriskel:
The DSNG closes when the company runs its first train, not buys, as it comes with a train, but starts on a yellow hex with no track.

Sorry @michaeljb another usage of company.abilities https://github.com/tobymao/18xx/pull/2545


michaeljb:
@ryandriskel no worries! I just merged #2545 and then added 7751a80 to this branch.

tobymao (approved):

PR #2722, opened 2020-12-18
Add delta value mode for player history

dlpierce:
Adds a button to the bottom of the spreadsheet that toggles a difference mode of the player value history table.

relative_player_history

### DIFF
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -20,10 +20,12 @@ module View
       def render
         @spreadsheet_sort_by = Lib::Storage['spreadsheet_sort_by']
         @spreadsheet_sort_order = Lib::Storage['spreadsheet_sort_order']
+        @delta_value = Lib::Storage['delta_value']
 
         children = []
         children << h(Bank, game: @game)
         children << render_table
+        children << render_spreadsheet_controls
 
         h('div#spreadsheet', { style: {
           overflow: 'auto',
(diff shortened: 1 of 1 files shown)


assets/app/view/game/spreadsheet.rb
```diff
@@ -77,11 +79,17 @@ def render_player_history
           end
           next if values == last_values
 
+          rel_values = (last_values || Array.new(values.length, 0)).map(&:-@).zip(values).map(&:sum) if @rel_player_hist
```
tobymao, line 82: use size instead of length

i had no idea you could do &:-@, pretty crazy

assets/app/view/game/spreadsheet.rb
```diff
@@ -77,11 +79,17 @@ def render_player_history
...
           last_values = values
           zebra_row = !zebra_row
           h(:tr, zebra_props(zebra_row), [
             h('th.left', h.round),
-            *values.map { |v| h('td.padded_number', @game.format_currency(v)) },
+            *values.map.with_index do |v, i|
```
tobymao, line 87: i'm not a fan of embedding this inside the array, can you move this out and just call it a var 

xxx = values.map...

assets/app/view/game/spreadsheet.rb
```diff
@@ -210,6 +218,23 @@ def toggle_sort_order
...
 
+      def toggle_relative_player_history
+        Lib::Storage['relative_player_history'] = !@rel_player_hist
+        update
+      end
+
+      def render_spreadsheet_controls
+        props = {
+          on: {
+            click: lambda do
```
tobymao, line 229: make this a one liner -> {... }

perwestling:
I am not sure I understand this view? Relative to what?

dfannius:
> I am not sure I understand this view? Relative to what?

That player's net worth in the previous round.

dfannius:
> I am not sure I understand this view? Relative to what?

That player's net worth in the previous round.

dlpierce:
> > I am not sure I understand this view? Relative to what?
> 
> That player's net worth in the previous round.

Correct. If anyone has a better name for this mode, please suggest it.

dlpierce:
Now referred to as delta, with code renamed appropriately. Button now says "Show Total Value" or "Show Delta Value".

tobymao (approved):

PR #2728, opened 2020-12-19
Token fees, extra shares, at float

tysen:


lib/engine/step/g_1849/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,23 @@
...
+      class BuySellParShares < BuySellParShares
+        def can_buy_multiple?(entity, corp)
+          super || (corp.owner == entity && just_parred(corp) && num_shares_bought(corp) < 2)
+        end
+
+        def just_parred(corporation)
+          @current_actions.any? { |x| x.is_a?(Action::Par) && x.corporation == corporation }
+        end
+
+        def num_shares_bought(corporation)
```
ryandriskel, line 17: Should the percentage of shares bought matter here?

ryandriskel: Ok, I think I am following a little better. Are we trying to determine if only the president's share has been bought this round?

tysen: The only time you can buy multiple shares at once in 1849 is on the same turn that you par a corp, on which you can buy up to an additional 20% of that corp. Parring (and thus receiving the president's certificate) is not a `Action::BuyShares` though, so that's why we have to check for the `Action::Par`. You can maximally get two shares from `Par` and two additional shares from `BuyShares`

tobymao (approved):

PR #2731, opened 2020-12-19
Fix all_bundles_for_corporation sorting when same % share exists

ryandriskel:
Context: The DNSG has 20% presidency, two 20% shares and four 10% shares.

The assumption was made that the President's certificate would always have a higher percentage and thus price than any other share. However, in a game where there are shares with the same percentage, the President's cert could wind up anywhere in the shares list. This could cause any or all bundles to contain the President's certificate.

The sorting at the end handles out of order partial bundles, because the partialing occurs as the last bundle added but could be a lower percentage.

---

Before

Screen Shot 2020-12-19 at 8 15 56 AM

After

Screen Shot 2020-12-19 at 10 01 08 AM

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -711,24 +711,26 @@ module Engine
       def all_bundles_for_corporation(share_holder, corporation, shares: nil)
         return [] unless corporation.ipoed
 
-        shares = (shares || share_holder.shares_of(corporation)).sort_by(&:price)
+        shares = (shares || share_holder.shares_of(corporation)).sort_by { |h| [h.president ? 1 : 0, h.price] }
 
         bundles = shares.flat_map.with_index do |share, index|
           bundle = shares.take(index + 1)
           percent = bundle.sum(&:percent)
           bundles = [Engine::ShareBundle.new(bundle, percent)]
-          if share.president
-            normal_percent = corporation.share_percent
-            difference = corporation.presidents_percent - normal_percent
-            num_partial_bundles = difference / normal_percent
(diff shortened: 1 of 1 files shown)


lib/engine/game/base.rb
```diff
@@ -711,24 +711,26 @@ def bundles_for_corporation(share_holder, corporation, shares: nil)
...
           bundles = [Engine::ShareBundle.new(bundle, percent)]
-          if share.president
-            normal_percent = corporation.share_percent
-            difference = corporation.presidents_percent - normal_percent
-            num_partial_bundles = difference / normal_percent
-            (1..num_partial_bundles).each do |n|
-              bundles.insert(0, Engine::ShareBundle.new(bundle, percent - (normal_percent * n)))
-            end
-          end
+          bundles.concat(partial_bundles_for_presidents_share(corporation, bundle, percent)) if share.president
```
ryandriskel, line 720: This is just a refactor for readability.

lib/engine/game/base.rb
```diff
@@ -711,24 +711,26 @@ def bundles_for_corporation(share_holder, corporation, shares: nil)
       def all_bundles_for_corporation(share_holder, corporation, shares: nil)
         return [] unless corporation.ipoed
 
-        shares = (shares || share_holder.shares_of(corporation)).sort_by(&:price)
+        shares = (shares || share_holder.shares_of(corporation)).sort_by { |h| [h.president && 1 || 0, h.price] }
```
tobymao, line 714: h.president ? 1 : 0 is more clear

perwestling (approved):

PR #2732, opened 2020-12-19
18CO - Fix detection of city having max stations

ryandriskel:
Corp closing worked fine on red tiles but not in places like Denver as the check is based on the tile being laid, not the hex it is laid in. Not sure how this made it through final checks before.

Tweaking the variable/function names to be more accurate.

---

After

Screen Shot 2020-12-19 at 10 35 02 AM

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -50,7 +50,7 @@ module Engine
       GREEN_TOWN_TILES = %w[co8 co9 co10].freeze
       GREEN_CITY_TILES = %w[14 15].freeze
       BROWN_CITY_TILES = %w[co4 63].freeze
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_co.rb
```diff
@@ -320,8 +320,8 @@ def remove_corporations_if_no_home(city)
         end
       end
 
-      def tile_has_max_cities(tile)
-        tile.color == :red || MAX_CITY_TILES.include?(tile.hex.name)
+      def tile_has_max_stations(tile)
+        tile.color == :red || MAX_STATION_TILES.include?(tile.name)
```
ryandriskel, line 324: `tile.hex.name` > `tile.name` is the core change

Issue #1302, opened 2020-08-01
18AL: Bonus not aparent


perwestling:
In #1293 @SmithInMN wrote:

"Likewise, it would be helpful if the named train bonus was named in the routes which use them. They are being added in correctly, but some notation to show which routes triggered the bonuses would be helpful."

In 18AL there are two types of bonuses:

1. Warrior Coal Field token, giving $10 each time a route contains a specific hex
2. Robert E Lee (och Pan American) giving a $20 (and/or $40) bonus for a run if any route contains two specific hexes

The Warrior Coal Field token bonus is added to the route revenue, but it is not apparent that a bonus has been received.

The Robert E Lee (and Pan American) is added to the revenue as a whole. 
Example:
- Route 1 is from Birmingham to Atlanta and gives $70
- Route 2 is from Birmingham to Atlanta (another path) and gives $80
In this case the total revenue is shown as $170, not $150, as it includes a one time bonus of Robert E Lee.

It would be nice if it somehow were more obvious that a bonus has been triggered.

perwestling:
Similar request can be found for other 18xx in #976 and #1063 

An attempt that might solve this has been done via #1085 but it is still not clear if it would solve the run bonus; possibly a separate line with extra bonus(es) might need to be added.

perwestling:
This might also apply to the Civil War event in 18TN, where when running the next runs, one of the trains are excluded from giving any revenue (which is implemented as the minimum one are just removed from total revenue). So it can be seen as a negative bonus. It has the same problem of not being apparent.

ryandriskel:
18CO and 1882 have examples of showing a bonus in the run log. Its not terribly pretty though.

perwestling:
18MS also has the Atlanta bonus that could use this, see issue 2000.

One lack I noticed when I used this in another game was that the bonus information did not appear until the log (besides the increased value). But anyway it is an improvement.
In case of Mobile-Nashville: Wonder how to best present it? M/N? 
Also, in case of two trains were managed to hit these bonus-routes - does it show on just one (if I recall it correctly you cannot get this bonus for more than one run).

ryandriskel:
Also, in case of two trains were managed to hit these bonus-routes - does it show on just one (if I recall it correctly you cannot get this bonus for more than one run).

Yessir!

[...]

PR #2743, opened 2020-12-20
18AL Bonus Revenue Display and Logging

ryandriskel:
Fixes: https://github.com/tobymao/18xx/issues/1302

---

There are a series of titles that award bonuses for stops at certain hexes or hitting specific routes. We may be able to make some generic functionality for this. In the meantime, I am tackling the titles missing this data in the route selector and logs.

---

Screen Shot 2020-12-20 at 2 45 31 PM

Screen Shot 2020-12-20 at 2 45 48 PM

Screen Shot 2020-12-20 at 2 46 26 PM

Screen Shot 2020-12-20 at 2 46 39 PM

### DIFF
--- a/lib/engine/game/g_18_al.rb
+++ b/lib/engine/game/g_18_al.rb
@@ -29,7 +29,10 @@ module Engine
         Step::SingleDepotTrainBuy::STATUS_TEXT
       ).freeze
 
-      ROUTE_BONUSES = %i[atlanta_birmingham mobile_nashville].freeze
+      ROUTE_BONUSES = {
+        atlanta_birmingham: "ATL>BRM",
+        mobile_nashville: "MOB>NSH",
+      }.freeze
 
       STANDARD_YELLOW_CITY_TILES = %w[5 6 57].freeze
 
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_18_al.rb
```diff
@@ -108,24 +112,52 @@ def stock_round
...
-          total_revenue += routes.map { |r| route_bonus(r, type) }.max
-        end if routes.any?
-        total_revenue
+      def routes_revenue_bonuses(routes)
+        return [] if routes.empty?
+
+        route_bonuses.map do |type, _description|
+          next unless abilities(routes.first.corporation, type)
+
+          possible_bonuses = routes.map { |r| route_bonus(r, type) }.compact
```
perwestling, line 142: What is the idea with the best bonus?

roseundy (commented):
You don't need to pass routes as a new parameter to revenue_for since you are already passing self.

lib/engine/route.rb
```diff
@@ -220,7 +220,7 @@ def revenue
         @game.game_error("Cannot use group #{key} more than once") unless group.one?
       end
 
-      @game.revenue_for(self, stops)
+      @game.revenue_for(self, stops, routes: @routes)
```
roseundy, line 223: You don't need to pass @routes, since you are already passing self. the revenue_for method can already access routes via that reference.

roseundy, line 223: You don't need to pass @routes, since you are already passing self. the revenue_for method can already access routes via that reference.

ryandriskel: Thanks Steve! Yeah, I didn't realize route had a reference to routes originally. That's going to simplify this greatly!

lib/engine/game/g_18_al.rb
```diff
@@ -206,9 +238,10 @@ def tile_lays(entity)
       private
 
       def route_bonus(route, type)
```
ryandriskel, line 240: I think I may have made a mistake here when I rebased. This will probably bork if there are multiple abilities.

lib/engine/game/g_18_al.rb
ryandriskel: This is almost the same code that was there before, but rather than applying the additional revenue immediately, route_bonus pairs the total bonus with the route that matched it. There could be multiple matches ( like ATL > BIR with two trains ), so the max call reduces it to applying to one route, the one where the bonus gives the best overall revenue.

perwestling: I wrote the original code ;-)

The problem here is that we have
- Route hex bonus
- Route bonus
- Routes bonus

And they can be unique or multiple.

18AL does have multiple route hex bonus, and two different unique routes bonuses.
Does the code allow for the case where two different (ATL>BRM,  MOB>NSH) routes bonuses are both triggered?

ryandriskel: I know you wrote it, that's why I requested your review. ;-)

Yeah, the naming is getting a little unwieldy. I'm in support of giving them more specific names. I'll do that.

"Does the code allow for the case where two different (ATL>BRM, MOB>NSH) routes bonuses are both triggered?"

Yeah, the variable `best_bonus` is really the best bonus for that bonus type ( atlanta-birmingham and mobile-nashville ). Note it is a map function, so we return results for both bonus types.

ryandriskel: I've got a local game where one company can just about run both routes, so I'll give it validation.

ryandriskel: Screen Shot 2020-12-20 at 6 38 24 PM
Screen Shot 2020-12-20 at 6 38 39 PM

lib/engine/game/g_18_al.rb
```diff
@@ -112,20 +115,25 @@ def revenue_for(route, stops)
...
 
-      def routes_revenue(routes)
-        total_revenue = super
-        route_bonuses.each do |type|
-          return total_revenue unless abilities(routes.first.corporation, type)
+      def revenue_str(route)
+        str = super
+
+        r_bonuses = []
+        r_bonuses.concat(bonuses_for_routes(route.routes).select { |b| b[:route] == route })
```
tobymao, line 130: no need to init and concat, just do r_bonuses =

perwestling:
Besides what toby commented on it looks fine.

PR #2744, opened 2020-12-20
[1860] add direct link to rules pdf

benjaminxscott:
Previously the rules link was to BGG, which requires login to view the PDF

This is on my personal dropbox as a shared file, can change to something else if we want

### DIFF
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -21,7 +21,7 @@ module Engine
       load_from_json(Config::Game::G1860::JSON)
 
       GAME_LOCATION = 'Isle of Wight'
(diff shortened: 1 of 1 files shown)


roseundy (commented):
Can we move this to a location on github? I don't think pointing to a personal Dropbox location is a good idea.

Under 18xx/public (like 18xx/public/rules) would work.

benjaminxscott:
Added a link to the eventual 'raw hosting' link on master

roseundy:
Sorry about that, but Toby would prefer your original solution. He doesn't want copyrighted material on github. So your original commit was fine :-/

benjaminxscott:
blew away the history just in case

PR #2749, opened 2020-12-21
1856 cannot par cgr

ventusignis:
I added the 'national' method so that 1836jr56 can share functionality with 1856 in the future


lib/engine/g_1856/corporation.rb
```diff
@@ -36,6 +36,11 @@ def can_buy?
         false
       end
 
+      def can_par?
+        return super unless self == @game.national
+        false
```
tobymao, line 41: this is probably a linter error

tobymao: self == @game.natinoal ? false : super

ventusignis:
I'm pushing up a new branch that has a cleaner history than this one

PR #2757, opened 2020-12-21
1860 add first two options, fix map UI to allow missing row/column

roseundy:
Add two options to 1860: 2-3 player map, and original (first edition) insolvency rules

When testing 2-3 player map, rendering of map was broken due to the lack of hexes in column "A". There was some code in UI:Map that tries to deal with axes not starting with "A" or "0", but it wasn't complete. I modified UI::Map, Hex and Axis to allow shifting the display of hexes.

New 2-3 player map and rendering:
optional_map

### DIFF
--- a/assets/app/view/game/axis.rb
+++ b/assets/app/view/game/axis.rb
@@ -15,6 +15,7 @@ module View
       needs :gap
       needs :map_x, default: 0
       needs :map_y, default: 0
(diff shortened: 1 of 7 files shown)


jenf (commented):
Initial glance looks okay

tobymao (approved):

PR #2764, opened 2020-12-22
Fix for route_selector when switching entities, but not steps

### LINKED ISSUE #2762: route_selector needs to refresh when moving between companies
As currently implemented, the route_selector UI view won't properly fill in with the previous run's route if moving directly from one corporation's routes step to another using "undo", "redo" or the history controls or, I suspect, when playing multi-player.

Unfortunately, this comes up often in the 1860 end-game.

roseundy:
Fixes #2762 

Add logic to route_selector to clear stored route when switching between corps/rounds with the same run_routes action.

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -12,6 +12,8 @@ module View
       include Lib::Color
       include Lib::Settings
 
+      needs :last_entity, store: true, default: nil
+      needs :last_round, store: true, default: nil
       needs :routes, store: true, default: []
       needs :selected_route, store: true, default: nil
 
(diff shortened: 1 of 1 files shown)


assets/app/view/game/route_selector.rb
```diff
@@ -39,7 +41,15 @@ def generate_last_routes!
       end
 
       def render
-        trains = @game.route_trains(@game.round.current_entity)
+        current_entity = @game.round.current_entity
+        # this is needed for the rare case when moving directly between run_routes steps
+        if @last_entity != current_entity || @last_round != @game.round
+          cleanup
+          store(:last_entity, current_entity)
```
tobymao, line 48: skip: true

roseundy: OK.

jenf (approved):

tobymao (approved):

Issue #2723, opened 2020-12-18
18Chesapeake - Company can use extra tile laying ability from private owned by company with same president.


ccxvii:
18Chesapeake lets both the CP and BS privates lay their extra tiles during a non-owning public company's turn if the same player is president of both the private-owning and currently operating public companies.

Example:
* Player A is president of both B&O and LV.
* LV owns CP and BS. 
* During the B&O operating round, B&O can use LV's CP and BS privates to lay extra tiles. 

Here's the game log:

```
-- Operating Round 4.1 (of 2) --
LV collects $10 from Baltimore and Susquehanna Railroad
LV collects $10 from Columbia - Philadelphia Railroad
C&O collects $5 from Delaware and Raritan Canal
C&O collects $15 from Chesapeake and Ohio Canal
Player A operates B&O
LV (C-P) lays tile #9 with rotation 2 on H2
LV (C-P) lays tile #8 with rotation 2 on I3
LV (B&S) lays tile #9 with rotation 2 on F4
LV (B&S) lays tile #9 with rotation 2 on G5
B&O lays tile #8 with rotation 3 on F12
```

PR #2766, opened 2020-12-22
Allow for increased control over who can use a company

ryandriskel:
I should have slept on the previous PR because I literally woke up inspired this morning with this addition!

---

This update allows for filtering not only companies owned by other players, but also those owned by your other corporations.

Some games allow using the ability of a company another of your corporations own. 18CO only allows you to utilize your own owned companies.

Both those options could probably wind up within the ability config and implemented in the game/base function here.


ryandriskel:
Relates to https://github.com/tobymao/18xx/issues/2723

tobymao (approved):

PR #2769, opened 2020-12-22
1849 share price behavior

tysen:
- Par values become available with phases
- Blue zone on market blocked until 16H
- Share price moves right only if dividend >= price
- Also, fewer trains in 3p game

### DIFF
--- a/lib/engine/g_1849/stock_market.rb
+++ b/lib/engine/g_1849/stock_market.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative 'stock_market'
+
+module Engine
+  module G1849
+    class StockMarket < StockMarket
+      attr_writer :game
+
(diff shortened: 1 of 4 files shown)


lib/engine/g_1849/stock_market.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative 'stock_market'
+
+module Engine
+  module G1849
+    class StockMarket < StockMarket
+      attr_writer :game
```
tobymao, line 8: i don't think this is necessary, was this just added for line 30?

tysen: The market needs access to `game` to write to the log and also to get the current phase information. Setting it in a writer was simpler than changing the signature of `initialize` for this subclass.

PR #2772, opened 2020-12-23
Last/double cert funkiness

tysen:
This cert must be bought last from wherever it is (corporate treasury or
market), issued last, and redeemed last. It can be sold by a player into
the market as they please. The current location of the double share is
marked with a 'd' on the corporatation charter.

Not included in this commit is the optional swap for the last cert by
the outgoing president in exchange for the presidency.

### DIFF
--- a/lib/engine/game/g_1849.rb
+++ b/lib/engine/game/g_1849.rb
@@ -73,7 +73,10 @@ module Engine
       def setup
         @corporations.sort_by! { rand }
         remove_corp if @players.size == 3
-        @corporations.each { |c| c.next_to_par = false }
+        @corporations.each do |c|
+          c.next_to_par = false
+          c.shares.last.last_cert = true
+        end
         @corporations[0].next_to_par = true
       end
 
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_1849.rb
```diff
@@ -179,14 +183,50 @@ def issuable_shares(entity)
...
+      end
+
+      def bundles_for_corporation(share_holder, corporation, shares: nil)
+        return [] unless corporation.ipoed
+
+        shares = (shares || share_holder.shares_of(corporation))
+
+        bundles = []
+
+        (1..shares.size).each do |n|
```
tobymao, line 212: bundle = (1..).flat_map

perwestling:
18Mex does have some funkiness, with its 5% shares, and exchange.

lib/engine/game/g_1849.rb
tysen: 'course. fixed.

tobymao (approved):

PR #2773, opened 2020-12-23
Add player choose to stock view

perwestling:
Make it possible to control from a step if choose should appear
in the view. It is useful in case there are several choose.

The main usage for this is in 18SJ where it is possible for a player
to trigger priority deal "stealing" due to an ability.

This commit contains the general code update needed to accomplish this.
Should not affect 1867 or 1817 which are the titles using choose so far.

### DIFF
--- a/assets/app/view/game/choose.rb
+++ b/assets/app/view/game/choose.rb
@@ -28,7 +28,7 @@ module View
 
         div_class = choice_buttons.size < 5 ? '.inline' : ''
         h(:div, [
-          h("div#{div_class}", { style: { marginTop: '0.5rem' } }, "#{@game.round.active_step.choice_name}:"),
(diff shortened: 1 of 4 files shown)


tobymao:
why would choice be an action but choice_available be false? i don't understand the use case

perwestling:
@tobymao I am trying to add a private ability where a player can trigger it to become priority dealer. So I add a choice (Yes/No more or less) that the player with the private has to answer.

When I add this to stock.rb it means that choose appear as an available action for the player in the stock round. Which meant that the choices appeared under EVERY corporation in the view..... (The choose that is on line 83 in stock.rb [86 in the new version])

So by adding this method it is possible to have different behavior (choice / no choice) depending on if the choice is for player or for corporation.

tobymao (approved):

PR #2774, opened 2020-12-23
18MEX: Add help text for token removal (fixes #2706)

### LINKED ISSUE #2706: [18Mex] - Prompt for NdM to remove an extra token.
Link: https://18xx.games/game/19294?action=342

CHI has merged into NdM, and NdM needs to remove a token, but nothing prompts the user to do so. A message in the game log would be enough.

perwestling:
See comment on the issue for how it looks like.

### DIFF
--- a/lib/engine/step/g_18_mex/assign.rb
+++ b/lib/engine/step/g_18_mex/assign.rb
@@ -41,9 +41,8 @@ module Engine
         end
 
         def help
-          help_text = ["Select a #{@game.merged_major.name} token to replace with an #{@game.ndm.name} exchange token."]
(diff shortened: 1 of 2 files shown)


lib/engine/step/g_18_mex/assign.rb
```diff
@@ -40,6 +40,12 @@ def available_hex(entity, hex)
           @game.merged_cities_to_select.find { |t| t.city.hex == hex }
         end
 
+        def help
```
tobymao, line 43: just make this one array?

['x', 'y'[

perwestling: Fixed

PR #2775, opened 2020-12-23
Various UI tweaks:

### LINKED ISSUE #1792: Don't show the "upgrade discount" column when the game has no discounts
On the info page, the upcoming trains table has an "upgrade discount" column that is always displayed, even if the game does not have any train discounts. The column shouldn't appear for these games.

### LINKED ISSUE #2541: Add Train roster to Entities and Spreadsheet tabs
The site UI can be improved. To do that, you should take in consideration two user types. The desktop user and the mobile user.
Every change in UI, unless they are separated (they aren't right now, and I guess it would be a pain to do so), should take in consideration both environments.
One of the shortcomings of the actual UI is that the informations are too scattered around and should be regrouped and organized on a functional criterion.
The "Game", "Entities", "Market" and "Tools" tabs are fine as they are, right now.
The "Map" and "Tiles" tabs can be merged together to avoid jumping between the two to check the tile roster and then decide where to expand your routes. Which is literally the way rr18xx handles this. The tile roster can be placed horizontally under the map, or vertically at the left of it. Right now I guess the tiles tab isn't used much since the UI actually suggests you what tile can be placed if you click on a hex, but for games with complex tile roster you should always be able to look at the map and the tile roster simultaneously. And we all know that having two separate tabs opened in a browser, one set on the map tab and one on the tiles map can be really taxing on the resources of the device you are using.

The info and spreadsheet tabs literally have, taken together, all the information needed to decide your action during a stock round. When you act, you want to know in which phase you are right now, how many trains are left, how many trains the company has etc. Qualititatively, they are all part of a single block of information needed to act tipically during a stock round., so it's only natural to group them in a single tab, similarly to this concept.

image

As a matter of fact, the inexperienced player will always go look at the info tab to keep track of the progress of the game and of the trains; the expert player will too, but way less, probably only to check the trains.
Any change should take in consideration every kind of user, both newcomers and veterans, right now having the info and spreadsheet information separated is more of an issue for the noobies than for the experienced players. In 1817 it might take me just a glance at the info tab , after that I can focus only on the spreadsheet, since I've played it many times and memorized the train roster and phases.
A newcomer might forget it really easily. Then, when analyzing the situation on the spreadsheet, he'll have to switch between spreadsheet and info tabs way more than an experienced player. And that is annoying, because it's fixable extremely easily, given that the spreadsheet and info tabs are mostly empty.
As shown in the concept I posted previously, rearranging the elements on the info tab on the "top" of it while leaving space for the spreadsheet under that will make the info tab actually complete, with all the needed info in one single place, without any need to annoyingly switch between tabs.
All the arguments against this proposal don't sound convincing.

- It will make the experience better not only for the experienced player, but (expecially) for the newcomers too, so it caters to everyone.
- The scrolling argument is not convincing. 1920x1080 fits the concept perfectly, 1366x768, the most common resolution after that, fits it with only a little bit of scrolling needed to see the cut off 10% portion of the spreadsheet. But, at that resolution, every other tab will need scrolling, so why is it an issue only for the info/spreadsheet one? Same for mobile, this change wouldn't change practically the amount of scrolling needed to use the site. It won't require a higher resolution from desktop too. It basically doesn't require anything, resolution wise. Here's the difference between height of the concept and the actual info tab
image
- It's a _tested_ and _working_ concept, since it's basically a lighter version of rr18xx main view.
- One could also argue that a player might not be interested in using the spreadsheet. That's not a problem! You can simply not look at it. All the "old" info is regrouped on the higher half of the tab, with a nice natural division made by the first line of the spreadsheet. Hence, visually, it isn't a hindrance for those who don't use the spreadsheet.

jenf:
Screenshot from 2020-12-23 14-34-29

Add the simplified train roster to entity and spreadsheet (fixes #2541)
Hide non-started corporations in 1817/1867 since they don't really matter (also #2541)
Always show loan amounts on 17 & 67 (also #2541)
Remove corporation size (also #2541)
Hide discount and available fields if they are unused (fixes #1792)

### DIFF
--- a/assets/app/view/game/entities.rb
+++ b/assets/app/view/game/entities.rb
@@ -39,10 +39,15 @@ module View
           ])
         end
 
+        extra_bank = []
+        unless @game.respond_to?(:unstarted_corporation_summary)
+          extra_bank += bank_owned.map { |c| h(Corporation, corporation: c) }
+        end
         children << h(:div, [
           h(Bank, game: @game),
+          h(GameInfo, game: @game, layout: 'upcoming_trains'),
           *@game.corporations.select(&:receivership?).map { |c| h(Corporation, corporation: c) },
-          *bank_owned.map { |c| h(Corporation, corporation: c) },
+          *extra_bank,
         ].compact)
 
         children = children.concat(bankrupt_players.map { |p| h(:div, [h(Player, player: p, game: @game)]) })
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -68,6 +68,12 @@ module View
             h('td.right', @game.format_currency(@game.round.active_step.seed_money)),
           ])
         end
+        if @game.respond_to?(:unstarted_corporation_summary)
+          trs << h(:tr, [
+            h(:td, 'Unstarted corporations'),
+            h('td.right', @game.unstarted_corporation_summary),
+          ])
+        end
 
         return unless trs.any?
 
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -61,7 +61,7 @@ module View
         children << render_abilities(abilities_to_display) if abilities_to_display.any?
 
         extras = []
-        extras.concat(render_loans) if @corporation.loans.any?
+        extras.concat(render_loans) if @game.total_loans&.nonzero?
         if @corporation.corporation? && @corporation.floated? &&
               @game.total_loans.positive? && @corporation.can_buy?
           extras << render_buying_power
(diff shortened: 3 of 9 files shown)


assets/app/view/game/entities.rb
```diff
@@ -39,10 +39,15 @@ def render
           ])
         end
 
+        extra_bank = []
+        unless @game.respond_to?(:unstarted_corporation_summary)
+          extra_bank += bank_owned.map { |c| h(Corporation, corporation: c) }
```
tobymao, line 44: concat

tobymao (approved):

PR #2776, opened 2020-12-23
Stop inability to create 1867 games

jenf:
Fix halt caused by home tokening twice


lib/engine/step/g_1867/single_item_auction.rb
```diff
@@ -136,6 +136,9 @@ def win_bid(winner, company)
         end
 
         def post_win_bid(winner, company)
+          # Avoid startup infinite recursion
+          return if @game.players.none?
```
tobymao, line 140: better to do players.empty?

tobymao (approved):

Issue #2781, opened 2020-12-23
18CO - Logging of negative discount


ryandriskel:
https://18xx.games/game/19885

Log incorrectly stated -$20. CS did not gain money ( correct )

<img width="612" alt="Screen Shot 2020-12-23 at 11 22 02 AM" src="https://user-images.githubusercontent.com/15675400/103026414-1c826400-4511-11eb-8b29-f6d62a4b7d8d.png">

PR #2792, opened 2020-12-23
Fix Issue 2781 - Negative Discount Logged

ryandriskel:
FIxes https://github.com/tobymao/18xx/issues/2781

https://18xx.games/game/19885

Log incorrectly stated -$20. CS did not gain money ( correct )

Screen Shot 2020-12-23 at 11 22 02 AM


ryandriskel:
Borders costs are in a different part of the code. This would only apply to the border cost if there was also a tile cost, which is not the intention.

PR #2795, opened 2020-12-24
Fix Issue 2781 - Negative Discount Logged

ryandriskel:
FIxes https://github.com/tobymao/18xx/issues/2781

https://18xx.games/game/19885

Log incorrectly stated -$20. CS did not gain money ( correct )

Screen Shot 2020-12-23 at 11 22 02 AM

---

Showing correct log on an upgrade and a no-cost hex.

Screen Shot 2020-12-23 at 6 27 30 PM

And correctly logging with one upgrade and one border.

Screen Shot 2020-12-23 at 6 33 16 PM

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1054,7 +1054,7 @@ module Engine
             (!a.hexes || a.hexes.include?(hex.name))
         end
 
-        cost = tile.upgrades.sum do |upgrade|
+        tile.upgrades.sum do |upgrade|
           discount = ability && upgrade.terrains.uniq == [ability.terrain] ? ability.discount : 0
 
           log_cost_discount(entity, ability, discount)
@@ -1062,8 +1062,6 @@ module Engine
           total_cost = upgrade.cost - discount
           total_cost
         end
-
-        cost
       end
 
       def tile_cost_with_discount(_tile, hex, entity, cost)
@@ -1078,8 +1076,7 @@ module Engine
         discount = [cost, ability.discount].min
         log_cost_discount(entity, ability, discount)
 
-        end_cost = cost - discount
-        end_cost
+        cost - discount
       end
 
       def log_cost_discount(entity, ability, discount)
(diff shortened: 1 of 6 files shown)


lib/engine/game/base.rb
```diff
@@ -1051,31 +1051,41 @@ def place_home_token(corporation)
...
-          log_tile_cost_discount(entity, ability, discount)
+          log_cost_discount(entity, ability, discount)
 
           total_cost = upgrade.cost - discount
           total_cost
         end
 
-        if ability && ability.terrain.nil?
-          log_tile_cost_discount(entity, ability, ability.discount)
+        cost
```
tobymao, line 1069: just remove the assigment on line 1060 and return that

lib/engine/game/base.rb
```diff
@@ -1051,31 +1051,41 @@ def place_home_token(corporation)
         city.place_token(corporation, token)
       end
 
-      def tile_cost(tile, hex, entity)
+      def upgrade_cost(tile, hex, entity)
```
tobymao, line 1054: why is this renamed?

lib/engine/game/base.rb
```diff
@@ -1051,31 +1051,41 @@ def place_home_token(corporation)
...
+            (!a.hexes || a.hexes.include?(hex.name))
         end
 
-        cost
+        return cost unless ability
+
+        discount = [cost, ability.discount].min
+        log_cost_discount(entity, ability, discount)
+
+        end_cost = cost - discount
```
tobymao, line 1084: you don't use end_cost, just return the expression

lib/engine/game/base.rb
ryandriskel: I found the function name misleading. The actual tile cost in the step is a much longer calculation involving tile lay abilities, borders, extra costs, etc. All this function does is calculate and log the discounts for _upgrades_ for the newly placed tile, hence the name change.

tobymao (approved):

Issue #2671, opened 2020-12-16
18CO Quality of Life Improvements


ryandriskel:
This is a list of improvements that can be made to make the game play smoother or more intuitive, but do not block the game from heading to production:

- [x] Update to use the new par groups from 1867 to make them more apparent on the Stock Market
- [x] Make the button text on Corporate Share Buying more context sensitive (ie "Pass Buying Shares" instead of "Pass")
- [x] Show the corporation class on the corp card
- [ ] Show the bidding corporation in the Acquisition Round
- [x] Make corporate shares more distinguishable on the card
- [x] Try to make the Denver Tile Clearer
- [x] Do not show emergency money (like how much you can raise) info if you aren't in EBuy (Global)
- [x] End train buying if there are no trains to buy from anywhere
- [x] Max company buy amount could be set to corp cash if its less than 150%
- [x] Allow issuing more than one share at a time ( complex due to price drops )
- [x] Hide abilities for other corporations as they aren't usable
- [x] In log, mark ignored town stops for doubling trains (see 1860)
- [x] In revenue selector, show E/W bonus
- [ ] Once President's Choice is Triggered, remove the event from the Info tab, as it is one time only.
- [x] Illustrate the DSNG 20% shares
- [x] Show all Corporate Shares available in Status after President's Choice 6b, 7
- [x] Dsitributes text should be skipped when there is no distribution

ryandriskel:
The last two items are so insignificant or rare that it isn't worth the time to do them.

PR #2789, opened 2020-12-23
1867 pre-alpha bug fixes

jenf:
Fix auto-pass when player didn't have enough money to pay 'normal' par price
Fix merge if it contains a town
Tidy up button text on merge
Remove the 'N' from the green token
Add a spec

PR #2796, opened 2020-12-24
18CO - Par Groups on the Market

ryandriskel:
This totally works, BUT I am interested to know if the change to `can_par?` impacts 1867.

Screen Shot 2020-12-23 at 7 17 15 PM

For: https://github.com/tobymao/18xx/issues/2671


jenf:
Possibly, you should also look at https://github.com/tobymao/18xx/pull/2789
As the minimum price needs setting for corps correctly.

Im not entirely sure if I've actually done higher level pars for majors yet in 1867

tobymao (approved):

PR #2802, opened 2020-12-24
add game state specs for fixtures

michaeljb:


spec/fixture_game_state_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+def full_desc(action_id, desc)
+  "at action #{action_id} #{desc}"
+end
+
+def at_action(action_id, desc)
+  it(full_desc(action_id, desc)) do
+    # share this across all individual tests within a describe block; doing this
+    # inside of an `it` prevents it from being executed if the `-e` flag is used
+    # on the commandline to run specific tests, and this test would be excluded
+    unless self.class.class_variable_defined?(:@@data)
```
michaeljb, line 19: Using `@@data` directly instead of via `class_variable_*` makes rspec print out a bunch of warnings

tobymao:
what's the value of these fixtures? would these specs catch anything that the full game specs wouldn't?

also i feel like it's a bit over engineered

michaeljb:
> what's the value of these fixtures? would these specs catch anything that the full game specs wouldn't?

You can verify that at a certain point in the game, some specific action was *not* allowed; the full game spec only verifies that all actions in the game were allowed.

Tests for restrcictions can be added, e.g., in 1846 the cert limit drops when a corporation closes. With the full game spec you'd just have a game where players happened to start holding fewer shares, but with this you can inspect the state more precisely.

> also i feel like it's a bit over engineered

I do like the interface of `at_action` down in the actual tests, but I guess it would also work to have a simpler helper function that just needs to be called in each `it` block.

tobymao:
what's up with storing the state as well? can we simplify these specs?

michaeljb:
> what's up with storing the state as well? can we simplify these specs?

I wanted to accomplish two things with that:

1) don't load the state if not running the specs that use it
2) load the state from the filesystem just once for all the specs that use it

In c34f745f I changed it to use `let` which easily solves (1), but not (2). `before(:all)` could probably be used to solve (2) but I don't that to have to be repeated in each `describe GAME_ID` block.

Probably fine to stop worrying about that before it's clearly a problem...

tobymao (approved):

Issue #2784, opened 2020-12-23
18CO- Current Implentation Allows Companies to issue shares in excess of the 50% Bank Pool Limit


Maxwell-Frank:
https://18xx.games/game/19885?action=498

The DSL here is able to issue a 6th share into the bank pool despite the 50% pool limit.

PR #2805, opened 2020-12-24
Fix 2784 - 18CO Issue Shares should respect 50% market limit

ryandriskel:
Fixes https://github.com/tobymao/18xx/issues/2784

Will break http://18xx.games/game/19885?action=498 - but that game is finished

Screen Shot 2020-12-24 at 9 49 20 AM

Also allows the DSNG to issue its 20% shares, as num_shares was 2, even though it was a single certificate.


tobymao (approved):

ryandriskel:
I think the fix itself is correct, but my current spec file is broken because of it. Will review. I thought I checked this before pushing but I must have had a different branch loaded.

Issue #2817, opened 2020-12-26
18CO Tiebreaker between multiple corporations having more shares than President not implemented.


ryandriskel:
```Rule 4.5.2 If multiple entities have more shares than the current President, the closest player seated clockwise from the current President becomes the new President, followed by Corporations in Operating order.```

Issue #2812, opened 2020-12-26
18CO unable to dump DPAC on UP


ChrisTheCat:
Foxrik is unable to sell 5 shares of DPAC, which should result in a takeover by UP.

https://18xx.games/game/20036?action=569

Issue #2780, opened 2020-12-23
18CO - DSNG Shares displayed incorrectly


ryandriskel:
After buying 10% share, then next turn buying a 20% share, Vanish took the presidency correctly, but the share count was wrong.

<img width="427" alt="Screen Shot 2020-12-23 at 10 50 49 AM" src="https://user-images.githubusercontent.com/15675400/103026323-e1803080-4510-11eb-8016-3d62b0ee96b9.png">

PR #2820, opened 2020-12-27
18CO Corporation can actually take President's Share

ryandriskel:
Fixes https://github.com/tobymao/18xx/issues/2812 
Fixes https://github.com/tobymao/18xx/issues/2780
Fixes https://github.com/tobymao/18xx/issues/2817 

### Implementation Notes

Looking for a better name for the `corporate_player_share_holders` function if ya got one. It is the set of the corporate share holders and player share holders. For all other current games, it results in the same set as `player_share_holders`

Could also use a better name for `totaling_num`. This function creates an array of shares that have a total percentage matching the number of shares needed to do the swap. This is needed to allow for 20% shares being owned. The default assumes all shares have the same percentage, and thus doesn't even need to consider the share's percentage. The 18CO override finds a bundle totaling the number of shares needed, but swapping 10% shares if the entity owns them.

These changes allow the Corporation to actually obtain the President's certificate, rendering the initial implementation of `find_takeover_corporation` unnecessary. Instead, we can simply check if the corporation is owned by another corporation.

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -140,12 +140,10 @@ module Engine
       @share_holders ||= Hash.new(0)
     end
 
-    def corporate_player_share_holders
-      share_holders.select { |s_h, _| s_h.player? || (s_h.corporation? && s_h != self) }
-    end
-
-    def player_share_holders
-      share_holders.select { |s_h, _| s_h.player? }
+    def player_share_holders(corporate: false)
+      share_holders.select do |s_h, _|
+        s_h.player? || (corporate && s_h.corporation? && s_h != self)
+      end
     end
 
     def corporate_share_holders
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -140,6 +140,10 @@ module Engine
       @share_holders ||= Hash.new(0)
     end
 
+    def corporate_player_share_holders
+      share_holders.select { |s_h, _| s_h.player? || (s_h.corporation? && s_h != self) }
+    end
+
     def player_share_holders
       share_holders.select { |s_h, _| s_h.player? }
     end
--- a/lib/engine/g_18_co/share_pool.rb
+++ b/lib/engine/g_18_co/share_pool.rb
@@ -18,6 +18,13 @@ module Engine
         # we want the bundle with the most shares, as higher percent in fewer shares in more valuable
         matching_bundles.max_by(&:size)
       end
+
+      def distance(entity_a, entity_b)
+        return 0 if !entity_a || !entity_b
+        return @game.players.size + 485 - entity_b.share_price.price if entity_b.corporation?
+
+        super
+      end
     end
   end
 end
--- a/lib/engine/g_18_co/share_pool.rb
+++ b/lib/engine/g_18_co/share_pool.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require_relative '../share_pool'
+
+module Engine
+  module G18CO
+    class SharePool < SharePool
+      def totaling_num(shares, num_shares)
+        return [] if shares.empty?
+        return [] unless num_shares
+        return shares if shares.one?
+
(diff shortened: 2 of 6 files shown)


lib/engine/g_18_co/share_pool.rb
```diff
@@ -0,0 +1,30 @@
...
+          shares.combination(n).to_a.select { |b| b.sum(&:percent) == percent }
+        end
+
+        # we want the bundle with the most shares, as higher percent in fewer shares in more valuable
+        matching_bundles.max_by(&:size)
+      end
+
+      def distance(entity_a, entity_b)
+        return 0 if !entity_a || !entity_b
+        return @game.players.size + 485 - entity_b.share_price.price if entity_b.corporation?
```
tobymao, line 24: what's 485

ryandriskel: Max share price in 18CO

ryandriskel: Lower overall number is "closer", so this inverts the share price, making higher share value corps have preference.

lib/engine/corporation.rb
```diff
@@ -140,6 +140,10 @@ def share_holders
       @share_holders ||= Hash.new(0)
     end
 
+    def corporate_player_share_holders
```
tobymao, line 143: remove this method and pass a flag to line 147 player_share_holders(corporations: true) or whatever

lib/engine/share_pool.rb
```diff
@@ -207,10 +207,16 @@ def change_president(presidents_share, swap_to, president)
...
       num_shares = presidents_share.percent / corporation.share_percent
 
-      possible_reorder(president.shares_of(corporation)).take(num_shares).each { |s| move_share(s, swap_to) }
+      totaling_num(possible_reorder(president.shares_of(corporation)), num_shares).each do |s|
+        move_share(s, swap_to)
+      end
       move_share(presidents_share, president)
     end
 
+    def totaling_num(shares, num_shares)
```
tobymao, line 216: totaling_num?

ryandriskel: It returns the set of shares that total the number of shares needed for the presidency swap (which is not equal if there are non-10% shares). It is not a boolean.

tobymao: can you rename it ? i still don't really get it

lib/engine/g_18_co/share_pool.rb
ryandriskel: I made it an 18CO constant so its more clear.

lib/engine/corporation.rb
ryandriskel: Done

lib/engine/share_pool.rb
ryandriskel: Had to sleep on it to come up with the relevant name.

lib/engine/g_18_co/share_pool.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative '../share_pool'
+
+module Engine
+  module G18CO
+    class SharePool < SharePool
+      def presidency_check_shares(corporation)
+        corporation.player_share_holders(corporate: true)
```
tobymao, line 9: this is already default true, why do you use it?

lib/engine/share_bundle.rb
```diff
@@ -50,7 +50,8 @@ def price
     def can_dump?(entity)
       return true unless presidents_share
 
-      (corporation.player_share_holders.reject { |k, _| k == entity }.values.max || 0) >= presidents_share.percent
+      sh = corporation.player_share_holders(corporate: true)
```
tobymao, line 53: this is already default true

ryandriskel: Good catch. Its should not be the default, that's why.

tobymao (approved):

PR #2833, opened 2020-12-27
skip implicit use of tile-laying abilities

### LINKED ISSUE #2818: 1846 and others: can use tile-laying privates implicity for free without using up the ability
Ideally there would be an interface for selecting whether the corporation's regular tile lay or the company's special ability should be used (and I have some work on a branch for that), but since it currently applies the discount, we should treat it as fully using the ability.

Sample game where NYC uses LSL twice - https://gist.github.com/michaeljb/4bc47b0ea4af22b411337617e690ddee

michaeljb:
[Fixes #2818]

Implicitly using tile lay abilities from companies allows the benefit of the ability to be applied without `use!`-ing the ability. With this change, in order for a company's tile laying ability to be used during a corporation's OR turn, the ability button must be pressed.

### DIFF
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -78,6 +78,7 @@ module Engine
         teleport = false
 
         tile_lay_abilities(entity) do |ability|
+          next if ability.owner != entity
           next if ability.hexes.any? && (!ability.hexes.include?(hex.id) || !ability.tiles.include?(tile.name))
 
           @game.game_error("Track laid must be connected to one of #{spender.id}'s stations") if ability.reachable &&


ryandriskel:
Does this fix https://github.com/tobymao/18xx/issues/2723 ... I couldn't tell from the report what the user was saying is the expected behavior.

michaeljb:
Nope not yet. The fix for #2723 and others will be a bigger change, and I discovered I could separate this fix out.

ryandriskel:
I guess I am confused why the functions being called return an ability where the corporation isn't the owner.

tobymao (approved):

michaeljb:
`ability.owner` can be a private company that is owned by the corporation, so the corporation does have access to that ability; in that case, `ability.owner.owner` would be the corporation.

In the scenario where the corporation is `entity` and `ability.owner` is a private company owned by the corporation, the ability hasn't been explicitly selected in the UI.

This is an approximate recreation of the sample hotseat game I posted on #2818 - https://18xx.games/game/20234?action=58
NYC upgraded Toledo (D14) for free by just clicking on it; it should cost them $20, but the Lake Shore Line's special ability was used implicitly, and did not get `use!`d, so then the ability is still available to explicitly use (by clicking on the ability button) to upgrade Cleveland (E17) for free.

PR #2837, opened 2020-12-27
Feat(Tile): Added frame option

Zwergenpunk:
Adds an Option to the tile code to add an colored frame to a tile, which is needed for 18CZ.

You can add the code `frame=color:purple` to get a ourple frame.

It looks like that:
image

### DIFF
--- a/assets/app/view/game/hex.rb
+++ b/assets/app/view/game/hex.rb
@@ -62,9 +62,9 @@ module View
         children << h(TriangularGrid) if Lib::Params['grid']
         children << h(TileUnavailable, unavailable: @unavailable, layout: @hex.layout) if @unavailable
 
-        if @tile&.frame && @tile&.frame.color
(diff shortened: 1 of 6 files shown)


assets/app/view/game/hex.rb
```diff
@@ -59,6 +62,15 @@ def render
         children << h(TriangularGrid) if Lib::Params['grid']
         children << h(TileUnavailable, unavailable: @unavailable, layout: @hex.layout) if @unavailable
 
+        if @tile&.frame && @tile&.frame.color
```
tobymao, line 65: @tile&.farme&.color

Zwergenpunk: thanks ;) I fixed it

tobymao:
you need to fix the linter

tobymao (approved):

PR #2839, opened 2020-12-27
Minor tile_lay ability tweaks

michaeljb:
* fix docs and default for `blocks` attr
* add attr `must_lay_all` for abilities that must lay all of their tiles
  together and cannot be interrupted by other actions (i.e., prevent C-P from
  illegally interrupting B&S at https://18xx.games/game/1273?action=165)


michaeljb:
This just has a couple of games to pin, was thinking I could release and pin alongside #2833

tobymao (approved):

Issue #2533, opened 2020-12-03
Very laggy interface on mobile devices


yraelrocks:
Turns get out of sync easily, refresh very slow & over multiple users. Particularly 1817 game: 14388

yrael - iOS 14.2 iPhone 11 Safari

yraelrocks:
Sunnd: Chrome on MacBook (work)plus windows (home) and iOS 14 (mobile) are the 3 platforms I use.

yraelrocks:
Phil: Windows on an Acer laptop and whatever OS samsung currently has on an S7

jenf:
yeah, 1817 seems to really stress the code out, we probably need to see if any performance improvements can be made especially with tile laying (which seem to take the majority of cpu time)

jenf:
One potential improvement. Skip rebuilding the graph if a tile has no revenue centers and doesn't connect to another tile.

jenf:
To fix this the connections code needs a rewrite. We should not generate connections while processing previous moves (similar to how the graph is mostly done). Then only calculate when needed.

tobymao:
this has been fixed. 26 seconds -> 4 seconds for game 18300

PR #2850, opened 2020-12-28
Connections performance improvement

jenf:
Change connections calculation to on-demand rather than when connected.

When a tile is layed, it walks the paths invalidating the cache of connected tiles
When route requests the connections at that point it calculates them if the cache is not already calculated.

This also simplifies the downgrade case as it invalidates it anyway

Stackprof goes from 138.76 seconds for 100 runs to 103.32 on 15528 (25% reduction)
The connect! goes from 33.92% of CPU to 16% (3.117% disconnect + 12.848% connect)
Loading unoptimized goes from 16.47 seconds to 14.23 seconds (15% reduction)

improves #2533, I'm sure we can do extra improvements

### DIFF
--- a/lib/engine/connection.rb
+++ b/lib/engine/connection.rb
@@ -5,39 +5,25 @@ module Engine
     attr_reader :paths
 
     def self.connect!(hex)
-      connections = {}
-      node_paths = []
-      hex_edges = {}
-
       hex.tile.paths.each do |path|
-        path.walk { |p| node_paths << p if p.node? }
-      end
-
-      node_paths.uniq.each do |node_path|
-        node_path.walk(chain: []) do |chain|
+        path.walk(chain: []) do |chain|
           next unless valid_connection?(chain)
 
-          connection = Connection.new(chain)
-          connections[connection] = true
+          path = chain[0]
 
-          [chain[0], chain[-1]].each do |path|
-            hex = path.hex
-            if path.exits.empty?
-              hex_edges[[hex, :internal]] = true
-              hex.connections[:internal] << connection
-            else
-              path.exits.each do |edge|
-                hex_edges[[hex, edge]] = true
-                hex.connections[edge] << connection
-              end
+          connection = Connection.new(chain)
+
+          if path.exits.empty?
+            hex.connections_cached[:internal] << connection
+          else
+            path.exits.each do |edge|
+              hex.connections_cached[edge] << connection
             end
           end
         end
       end
 
-      hex_edges.keys.each do |hex_, edge|
-        connections = hex_.connections[edge]
-        connections.select!(&:valid?)
+      hex.connections_cached.values.each do |connections|
         connections.uniq!(&:hash)
       end
     end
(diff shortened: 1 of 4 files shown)


lib/engine/hex.rb
```diff
@@ -181,32 +182,40 @@ def lay(tile)
...
 
-      tile.restore_borders
+    def connections
+      connect unless @connected
+      @connections
     end
 
-    def connect!
-      Connection.connect!(self)
+    def connections_cached
```
tobymao, line 211: do we need this if we already have line 206 connections?

lib/engine/connection.rb
```diff
@@ -5,39 +5,25 @@ class Connection
...
-      node_paths = []
-      hex_edges = {}
-
       hex.tile.paths.each do |path|
-        path.walk { |p| node_paths << p if p.node? }
-      end
-
-      node_paths.uniq.each do |node_path|
-        node_path.walk(chain: []) do |chain|
+        path.walk(chain: []) do |chain|
```
tobymao, line 9: why aren't we walking node paths anymore? why do we walk all paths?

lib/engine/hex.rb
tobymao: i see, you're using it as a setter, i think that pattern is a bit weird

lib/engine/connection.rb
```diff
@@ -5,39 +5,25 @@ class Connection
...
           next unless valid_connection?(chain)
 
+          path = chain[0]
```
tobymao, line 12: this still preserves the old code? it looks difference because you're no longer using chains?

lib/engine/connection.rb
```diff
@@ -5,39 +5,25 @@ class Connection
...
+              hex.connections_cached[edge] << connection
             end
           end
         end
       end
 
-      hex_edges.keys.each do |hex_, edge|
-        connections = hex_.connections[edge]
-        connections.select!(&:valid?)
+      hex.connections_cached.values.each do |connections|
```
tobymao, line 26: you don't need select! valid? anymore because there's never an invalid connection?

lib/engine/hex.rb
tobymao: perhaps it makes sense to move the connect code in Hex so you don't have to expose the private connections_cached, or just have the connection code set it directly

lib/engine/connection.rb
jenf: The old code did both hexes in a chain, this just does the one it works on

lib/engine/connection.rb
jenf: That was really for when an tile without nodes was laid. It'd then update all the hexes connected. The walk paths still happens but only to invalidate the cache

lib/engine/connection.rb
jenf: True will update

lib/engine/hex.rb
jenf: It can call connections directly, will update

PR #2866, opened 2020-12-29
Nationalization payoff

ventusignis:
I've been working on this for a little while and I think it's time to share what I have right now to see if the big picture solution to this is acceptable while I work on the little details


lib/engine/config/game/g_1856.rb
```diff
@@ -517,18 +517,24 @@ module G1856
...
+      "abilities": [
+         {
+            "type": "train_buy",
+            "description": "Inter train buy/sell at face value",
+            "face_value": true
+         },
+         {
+            "type": "train_limit",
+            "description": "+1 train limit",
+            "increase": 100
```
ventusignis, line 531: By making the increase unreasonably big the goal is to prevent the forced train discards until after nationalization has completed

lib/engine/g_1856/corporation.rb
```diff
@@ -15,7 +16,9 @@ def initialize(game, sym:, name:, **opts)
         @game = game
         @started = false
         @escrow = nil
+        @log = @game.log
         super(sym: sym, name: name, **opts)
+        @all_shares = shares_by_corporation[self].dup
```
ventusignis, line 21: Is there a better way?

lib/engine/config/game/g_1856.rb
perwestling: I think you can modify description to something better. An alternative is to introduce a new attribute to cater for this case, but I suppose the solution is OK.

lib/engine/config/game/g_1856.rb
```diff
@@ -517,18 +517,24 @@ module G1856
...
+         },
+         {
+            "type": "no_buy",
+            "description": "Unavailable until nationalization"
```
perwestling, line 535: I think we might beed to general work on this ability. Maybe it should be part of the general code to check for this no_buy (in can_par, can_buy, can_exchange) and I think it is not today.
Also we could consider having some timing attribute, when it should stop?

Is used for NdM in 18MEX and I am using it in the upcoming 1824.

lib/engine/g_1856/corporation.rb
```diff
@@ -37,7 +40,7 @@ def can_buy?
       end
 
       def can_par?
-        self == @game.national ? false : super
+        self == @game.national ? !@ipoed : super
```
perwestling, line 43: As my comment about no_buy above maybe this can be handled in general code?

lib/engine/g_1856/corporation.rb
```diff
@@ -56,6 +59,27 @@ def release_escrow!
...
+
+          @all_shares.each_with_index do |share, index|
+            share.percent = index.zero? ? 10 : 5
+            share.cert_size = index.zero? ? 1 : 0.5
+          end
+        end
+
+        @log << "#{@name} issues 10 more shares"
+        num_shares = @all_shares.count
+        # Yes, this can create 50 5% shares in one corp. 1891 does this. It is weird
```
perwestling, line 75: 50 5% shares? Isn't 20 max in 1856?

lib/engine/config/game/g_1856.rb
ventusignis: Right now this ability is only there as a visual aid, and it is controlled by the 1856::Corporation class. I think we could make it into an actual ability where having the ability makes it unparrable until the ability is removed

lib/engine/config/game/g_1856.rb
ventusignis: After the CGR forms during nationalization it does actually have a +1 train limit. Are you suggesting I change it to say "3 train limit" instead of "+1 train limit"?

lib/engine/g_1856/corporation.rb
ventusignis: Yeah 20 is max in 1856. 1891 (a 1856 variant) lets the national keep issuing shares. it handles paying dividends by taking the run into the treasury then paying out if the treasury has enough to pay dividends, otherwise it keeps the money

ventusignis: It gets a stock bump if it pays out and if it doesn't have enough to pay out it spins in place

lib/engine/g_1856/corporation.rb
ventusignis: I made an Unparrable ability that prevents a corporation from being parred until the ability is removed. This doesn't help 18MEX because the NdM is parred in the private auction but it may be useful in other games

lib/engine/corporation.rb
```diff
@@ -92,6 +92,7 @@ def buy_multiple?
     def can_par?(entity)
       return false if @par_via_exchange && @par_via_exchange.owner != entity
       return false if @needs_token_to_par && @tokens.empty?
+      return false if all_abilities.find { |a| a.type == :unparrable }
```
tobymao, line 95: it might be easier to move can_par? into the game class

lib/engine/g_1856/corporation.rb
```diff
@@ -15,7 +16,9 @@ def initialize(game, sym:, name:, **opts)
         @game = game
         @started = false
         @escrow = nil
+        @log = @game.log
```
tobymao, line 19: we shouldn't do this

lib/engine/g_1856/corporation.rb
tobymao: probably an indication that the corporation subclass is doing too much

lib/engine/g_1856/corporation.rb
```diff
@@ -56,6 +55,28 @@ def release_escrow!
         @capitalization = :incremental
       end
 
+      # Issue more shares
+      def issue_shares!
```
tobymao, line 59: should be in the game class and not in corporation

tobymao:
seems like you're jamming too much stuff into g1856 corporation that really should be in the g1856 game class.

lib/engine/g_1856/corporation.rb
ventusignis: This is now in `g_1856.rb` as `national_issue_shares!`

lib/engine/g_1856/corporation.rb
ventusignis: I changed `corporation.rb` to use `@game.log` instead in the remaining log statements

lib/engine/corporation.rb
ventusignis: `can_par?` in `Corporation` now calls `corporation_parrable?` in `Game` so that `SharePool`'s `can_par?` can be differentiated later as `parrable_at?` or something like that if needed in the future

lib/engine/g_1856/corporation.rb
ventusignis: I've pulled out the logic from here. I still hope there's a better way to get access to the shares in the corporation so that they can be modified (change percentage & cert_size) for share issuing and the super weird handling of the case of 1-share CGR presidency

lib/engine/corporation.rb
ventusignis: This broke pretty thoroughly; `Corporation` doesn't have a reference to `@game`. 
I think there are three main solutions to this:
1: remove `can_par?` from corporation and just use a method in `game`
 - This is doable without much effort, I'd want to give heads up to the 1824 and 1849 developers that this will slightly affect their games
 2: add `@game` to `Corporation`
  - This seems like it would be contrary to the goal of avoiding packing logic in `Corporation`
 3: revert to how it was in the PR originally
 
 I'm fine with any direction on this, @tobymao do you have any preferences?

tobymao: i don't know what version 3 is but i was suggesting 1.

entities shouldn't require game objects, if they do, it's a smell because it's game -> entity -> game, by moving it to game, yes game gets bigger, but it's just much easier to override etc.

entities should be mostly simple data container objects without business logic. because if there's logic, then a game can and will override it

lib/engine/g_1856/corporation.rb
tobymao: what i meant by this was Corporation shouldn't have access to the game or the log

ventusignis:
Thanks for the feedback @tobymao and @perwestling !
I'm closing this for now to split some of it apart as a separate PR then resubmit this later

PR #2867, opened 2020-12-29
Make it possible to sell unsold privates during SR.

perwestling:
This is a preparation for 1824 which does not have an initial auction,
but instead the privates are sold together with shares during SR 1.

### DIFF
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -212,7 +212,7 @@ module Engine
       end
 
       def purchasable_unsold_companies
-        @game.companies.select { |c| c.owner.nil? }
+        @game.companies.reject(&:owner)
       end
 
       def get_par_prices(entity, _corp)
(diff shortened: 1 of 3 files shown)


lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -211,6 +211,10 @@ def purchasable_companies(entity)
         @game.purchasable_companies(entity)
       end
 
+      def purchasable_unsold_companies
+        @game.companies.select { |c| c.owner.nil? }
```
tobymao, line 215: companies.reject{ |c| c.owner }

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -239,15 +243,18 @@ def process_buy_company(action)
...
         company.owner = entity
-        owner.companies.delete(company)
+        owner&.companies&.delete(company)
 
         entity.companies << company
-        entity.spend(price, owner)
+        entity.spend(price, owner.nil? ? @game.bank : owner)
         @current_actions << action
-        @log << "-- #{entity.name} buys #{company.name} from #{owner.name} for #{@game.format_currency(price)}"
+        @log << "#{owner.nil? ? '' : '-- '}"\
```
tobymao, line 254: owner ? '--' ..

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -239,15 +243,18 @@ def process_buy_company(action)
...
+                "#{entity.name} buys #{company.name} from "\
+                "#{owner.nil? ? 'the market' : owner.name} for "\
```
tobymao, line 256: same here inverse so you don't have to call explicit nil

lib/engine/step/buy_sell_par_shares.rb
perwestling: Fixed

lib/engine/step/buy_sell_par_shares.rb
perwestling: Fixed

lib/engine/step/buy_sell_par_shares.rb
perwestling: Fixed

PR #2843, opened 2020-12-28
Properly remove reservations when a token is placed, fix cert limit in 1846 games

michaeljb:
Game 19962 is broken by the cert limit change, so for the fixture, instead of
allowing actions that violate the cert limit, add an `end_game` action and
remove the rest of the game's actions

[Fixes #2799]
[Fixes #2800]

This will require a small handful of games to be pinned; like the one reported in issues #2799 and #2800, they are 1846 games involving a bankruptcy and then corporation closure resulting in incorrect cert limit.

PR #2870, opened 2020-12-29
only allow actions from the relevant company during a blocking SpecialTrack step

michaeljb:
This is a necessary change for the broader ability timing rework.

It breaks 3 games locally, but they can all be easily migrated by adding a pass action.

The `current_step` check is a sort of workaround the circular fact that `game.active_step` needs to check what actions are available, and some actions are only available if the ability is, but the abilities also need to know what the game step is, so if `game.active_step` is called inside of `abilities()`, we'll hit an infinite loop.

In `Round::Base#skip_steps`, `DiscardTrain` gets skipped if it has no crowded corporations, but it does not get marked as passed, so when current_step is defined in `ability_right_time?`, `DiscardTrain` will always seem to be the active step, unless we move it later in the ordered list of steps in the OR; therefore, I put it as late as possible.


lib/engine/game/g_18_co.rb
```diff
@@ -210,7 +210,6 @@ def operating_round(round_num)
         Round::Operating.new(self, [
         Step::Bankrupt,
         Step::G18CO::Takeover,
```
ryandriskel, line 212: For 18CO, a corporation can gain trains during Takeover, which could cause a DiscardTrain to be required. I have a concern that putting it after Route will allow the corporation to run an extra train for a round.

michaeljb: Maybe I could change up the logic here, if a step gets passed over in this loop because `active` returned false, it could be marked as inactive? Then in my case, DiscardTrain would be inactive, so I could reject steps that were either `inactive` or have had `passed` called?

https://github.com/tobymao/18xx/blob/891ed4f699ab5f4bea0d950775c1ec9b29e4412c/lib/engine/round/base.rb#L156

ryandriskel: As long as the `inactive` flags are cleared with every action, that seems like it would work.

ryandriskel: My brain is exploding a bit from that 18CO playthrough so I don't even know if that made sense ... but yes, essentially cache the result of active? until the gamestate changes.

michaeljb: My first attempt at this only breaks 8 out of 10,512 games, not too bad.

michaeljb: Turns out those games are all the ones that I pinned for #2843, I just forgot to pin them in my local copy of the DB 😆

lib/engine/round/base.rb
```diff
@@ -153,7 +153,8 @@ def operating?
 
       def skip_steps
         @steps.each do |step|
-          next if !step.active? || !step.blocks? || @entities[@entity_index]&.closed?
+          step.inactive = !step.active?
```
michaeljb, line 156: `skip_steps` is called every time an action is processed, so an `@inactive` value won't overstay its welcome

https://github.com/tobymao/18xx/blob/891ed4f699ab5f4bea0d950775c1ec9b29e4412c/lib/engine/round/base.rb#L97

lib/engine/game/g_18_co.rb
michaeljb: Did another local verification, this is all looking good. Thanks @ryandriskel!

lib/engine/game/base.rb
```diff
@@ -1875,6 +1875,11 @@ def ability_right_time?(ability, time, strict_time)
         return false if strict_time && !ability.when
         return true unless time
 
+        current_step = @round.steps.find { |s| s.blocks? && !s.passed? && !s.inactive }
```
tobymao, line 1878: can you use active_step? and if not maybe add it to the round base

lib/engine/step/base.rb
```diff
@@ -16,6 +17,14 @@ def initialize(game, round, **opts)
         @round = round
         @opts = opts
         @acted = false
+        @name = name
+      end
+
+      def name
```
tobymao, line 23: why do you need this? can't you just check that it's a SpecialTrack?

lib/engine/step/base.rb
```diff
@@ -6,7 +6,8 @@ module Engine
   module Step
     class Base
       include Passer
-      attr_accessor :acted
+      attr_accessor :acted, :inactive
```
tobymao, line 9: why do you need inactive?

lib/engine/game/base.rb
michaeljb: `active_step` leads to an infinite loop:

* `Game#active_step` just calls `Round#active_step` https://github.com/tobymao/18xx/blob/d025f57c5679b011a9312136fd899032b3970f91/lib/engine/game/base.rb#L477-L479
* `Round#active_step` with no entity passed calls `step.blocking?` https://github.com/tobymao/18xx/blob/d025f57c5679b011a9312136fd899032b3970f91/lib/engine/round/base.rb#L122-L126
* `Step#blocking?` calls `current_actions` https://github.com/tobymao/18xx/blob/d025f57c5679b011a9312136fd899032b3970f91/lib/engine/step/base.rb#L75-L77
* `Step#current_actions` calls actions` https://github.com/tobymao/18xx/blob/d025f57c5679b011a9312136fd899032b3970f91/lib/engine/step/base.rb#L57-L61
* In some cases, a step's `actions` calls `Game#abilities`, completing the loop, e.g., `SpecialTrack` calls `tile_lay_abilities` which in turn calls `Game#abilities` https://github.com/tobymao/18xx/blob/d025f57c5679b011a9312136fd899032b3970f91/lib/engine/step/special_track.rb#L14-L19 https://github.com/tobymao/18xx/blob/d025f57c5679b011a9312136fd899032b3970f91/lib/engine/step/tracker.rb#L41-L43


I'll add it to the round base

lib/engine/step/base.rb
michaeljb: I don't want `DiscardTrain` showing up as the `current_step` when we're really on the `Track` step. If no train needs to be discarded, `Round#skip_steps` will just call `next` when looking at the `DiscardTrain` step, as `DiscardTrain#active?` returns `false`; it doesn't get `pass!` called on it.

Up where `abilities` wants to know what the current step is, without `inactive`, it can't filter out `DiscardTrain` from the list.

My first approach was to just move the `DiscardTrain` step to right before `BuyTrain`, but this is incompatible with 18CO (see comment thread at https://github.com/tobymao/18xx/pull/2870#discussion_r549552231).

tobymao: i'd rather we handle this another way instead of making the system more complex just for 18co. can you have 18co do an explicit pass or something?

lib/engine/step/base.rb
michaeljb: On my branch with the more extensive ability timing changes, it is assumed that when an ability's `when` refers to `track`, that really means the `Track` step, so I added `#name` here to keep the ability `when` consistent. Switched to using `.is_a?(SpecialTrack)` in 54d5d9e6, and will give some more thought about how to handle this on my branch (which has a few other things that need cleaning up before I open a PR for it anyway)

lib/engine/step/base.rb
michaeljb: I'll see if I can find a sample 18CO game where a takeover forced a train discard and experiment with it a bit.

michaeljb: Doesn't look like any such games exist yet:

```
irb(main):008:0> Game.where(title: '18CO').all.select { |game| game.actions.any? { |a| a.action['type'] == 'discard_train' } }
=> [] 
```

I'll try and make it happen in a game now, just have to do a bit of reading up on 18CO first 😄

michaeljb: Got a sample hotseat game, now to experiment with code changes - https://gist.github.com/michaeljb/9eb00cb70ad0efaf2659d9a190e95f0d

michaeljb: As @ryandriskel suspected, moving DiscardTrain to right before BuyTrain breaks with the above game:

```
 GameError: Step Lay Track cannot process action {"type"=>"discard_train", "entity"=>"DSNG", "entity_type"=>"corporation", "id"=>81, "train"=>"2-3"}
```

I could have `potential_blocking_step` reject the `DiscardTrain` step since it's generally unlikely to be a step where special abilities are happening. Or maybe move `potential_blocking_step` from `Round` to `Game`, then 18CO could override it to reject `DiscardTrain`, and any other game can do something similar if needed?

lib/engine/round/base.rb
```diff
@@ -125,6 +125,10 @@ def active_step(entity = nil)
         @active_step ||= @steps.find { |step| step.active? && step.blocking? }
       end
 
+      def potential_blocking_step
```
michaeljb, line 128: I'm not super happy with this name, I didn't want it to be `blocking_step` since it doesn't actually call `step.blocking?`. Maybe if I move it into Game (see https://github.com/tobymao/18xx/pull/2870#discussion_r549816469) I could rename it to `ability_blocking_step` or `ability_current_step` or something? That groups it with the other `ability_` methods and makes it a little more clear this is really only for the special case of looking at abilities.

lib/engine/step/base.rb
michaeljb: Got rid of `inactive`, moved `Round#potential_blocking_step` to `Game#ability_blocking_step`, and gave 18CO a special override to ignore `DiscardTrain` in ffefe35d, running local validation now...

lib/engine/step/base.rb
```diff
@@ -7,6 +7,7 @@ module Step
     class Base
       include Passer
       attr_accessor :acted
+      attr_reader :company
```
tobymao, line 10: does this need to be in base?

michaeljb: Nope, sorry 824dcab8

tobymao (approved):

Issue #2672, opened 2020-12-16
Stock round indicator on corporations with no action


ryandriskel:
During the stock round, all corporations are shown. This is good to have a general sense of relative positioning. However, many of the corporations will have no actions a player can take once they are selected. It would help reduce mental processing if there was a quick indication of whether there are actions available for that corporation.

A few potential (but not exhaustive) implementation ideas:
- reduced transparency
- colored border
- icon

benjaminxscott:
I like the idea of lighter transparency for corp cards for which the active player has no valid actions

roseundy:
Perhaps a dashed line for their borders?

ryandriskel:
This as at 60% opacity. I'm not sure it is distinguishable enough, but I want to still be able to read the corp card.

<img width="1351" alt="Screen Shot 2020-12-21 at 5 56 49 PM" src="https://user-images.githubusercontent.com/15675400/102897991-16fc1f80-4426-11eb-9d36-ad86251474b4.png">

<img width="1363" alt="Screen Shot 2020-12-21 at 5 57 21 PM" src="https://user-images.githubusercontent.com/15675400/102897995-182d4c80-4426-11eb-9cb7-bd57a7c1f57e.png">

benjaminxscott:
imo it's not very distinguishable, perhaps some sort of [striped gradiant](https://css-tricks.com/stripes-css/) for the background of cards?

ryandriskel:
Here with are with a very light grey background on corps with no actions. Thoughts?

<img width="1357" alt="Screen Shot 2020-12-29 at 11 29 40 AM" src="https://user-images.githubusercontent.com/15675400/103305791-4a602080-49c9-11eb-852a-b2a9acb10305.png">

benjaminxscott:
Hmm it seems not that distinguishable, but I'm not the only opinion fwiw (just I get email notifs for this thread)

Perhaps  the [disabled](https://www.w3schools.com/tags/att_button_disabled.asp) attribute could be combined to make it obvious that it's unclickable when hovered?

ryandriskel:
The stripes work and are noticable, but I feel like they make the text difficult to read

<img width="1321" alt="Screen Shot 2020-12-29 at 12 45 28 PM" src="https://user-images.githubusercontent.com/15675400/103309981-c495a280-49d3-11eb-90d2-8fe7d1d569ea.png">

crericha:
[...]

PR #2892, opened 2020-12-29
2672 - Stock Round - Denote corps with no actions

ryandriskel:
Closes https://github.com/tobymao/18xx/issues/2672

The active player has no buy/sell/choose actions for the corps with a light gray background.

Screen Shot 2020-12-29 at 11 29 40 AM

We could do more with this like disable selectable and turn the cursor hand to pointer, but I didn't want to make assumptions about why someone might need to select a corp..

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -30,7 +30,8 @@ module View
 
         children << h(SellShares, player: @current_entity, corporation: @corporation)
 
-        h(:div, children.compact)
+        children = children.compact
+        h(:div, children) if children.any?
       end
 
       def render_buy_shares
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -31,7 +31,7 @@ module View
         children << h(SellShares, player: @current_entity, corporation: @corporation)
 
         children = children.compact
-        h(:div, children) if children.any?
+        return h(:div, children) if children.any?
 
         nil
       end
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -32,6 +32,8 @@ module View
 
         children = children.compact
         h(:div, children) if children.any?
+
+        nil
       end
 
       def render_buy_shares
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -39,10 +39,16 @@ module View
         card_style = {
           cursor: 'pointer',
         }
-        card_style[:border] = '4px solid' if @game.round.can_act?(@corporation)
+
         card_style[:display] = @display
 
-        card_style[:backgroundColor] = '#ededed' unless interactive?
+        unless interactive?
+          factor = Native(`window.matchMedia('(prefers-color-scheme: dark)').matches`) ? 0.8 : 0.5
+          card_style[:backgroundColor] = convert_hex_to_rgba(color_for(:bg2), factor)
+          card_style[:border] = '1px dashed'
+        end
+
+        card_style[:border] = '4px solid' if @game.round.can_act?(@corporation)
 
         if selected?
           card_style[:backgroundColor] = 'lightblue'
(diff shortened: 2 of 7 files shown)


ryandriskel:
Style fixes incoming :/

michaeljb:
Looks good in normal mode, but very difficult to read in dark mode (unfortunately I don't know where to look to style things differently in dark mode)

Screenshot 2020-12-29 170444

ryandriskel:
PR for the win. I forgot dark mode is even a thing! Thanks for catching that @michaeljb !

daniel-sousa-me:
It's worse than "dark mode being a thing". The users can change the color to whatever they want.

roseundy:
Can you outline the corp card with dashed lines? Too subtle?

ryandriskel:
> Can you outline the corp card with dashed lines? Too subtle?

I can try it. You suggested that earlier and I glossed over it.

ryandriskel:
I quite like the combo of dashed border and graying

Screen Shot 2020-12-31 at 8 19 12 AM

ryandriskel:
Dark mode

Screen Shot 2020-12-31 at 8 21 50 AM

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -30,7 +30,8 @@ def render
 
         children << h(SellShares, player: @current_entity, corporation: @corporation)
 
-        h(:div, children.compact)
+        children = children.compact
+        h(:div, children) if children.any?
```
tobymao, line 34: this is not really well defined behavior as it implicitly returns nil i believe, you shouldn't dot his pattern,  always return a div

tobymao: or if you do need to return the nil explicitly and fix all references of it to handle it

assets/app/view/game/corporation.rb
```diff
@@ -38,9 +39,17 @@ def render
         card_style = {
           cursor: 'pointer',
         }
-        card_style[:border] = '4px solid' if @game.round.can_act?(@corporation)
+
         card_style[:display] = @display
 
+        unless interactive?
+          factor = Native(`window.matchMedia('(prefers-color-scheme: dark)').matches`) ? 0.8 : 0.5
```
daniel-sousa-me, line 46: I think this should be avoided. I prefer dark mode and have it enabled on my computer, but on my phone I don't have that setting. In both cases I use the same color scheme (defined on the profile settings).

I'd suggest you just pick an intermediate value like 0.6

ryandriskel: This is used elsewhere in the system, however I could calculate it based on whether the background color is light or dark rather than using a window function.

ryandriskel: I've changed this to use math based on bg2 instead of the dark mode detection.

Screen Shot 2021-01-01 at 8 24 30 AM
Screen Shot 2021-01-01 at 8 24 53 AM

daniel-sousa-me: Afaict it's only used twice. One of them is to set the default. The other is in the spreadsheet. These 2 screenshots are taken with the same account, but on different browsers:

image

image

I'll try to open a PR changing the spreadsheet to work like what you did.

assets/app/view/game/corporation.rb
```diff
@@ -568,6 +577,10 @@ def selected?
         @corporation == @selected_corporation
       end
 
+      def interactive?
```
tobymao, line 580: this method is unecessary

assets/app/view/game/sell_shares.rb
```diff
@@ -46,7 +46,10 @@ def render
           end
         end
 
-        h(:div, buttons.compact)
+        buttons = buttons.compact
+        h(:div, buttons) if buttons.any?
```
tobymao, line 50: are you missing a return here? 

did you test this code?

ryandriskel: My savior!

ryandriskel: I fixed this in both places where I added the explicit nil return. However, if you'd prefer me to hold on to this one until we've got more UI tests, I understand. I have verified buy/sell/par buttons but not the "choose"

tobymao (approved):

PR #2895, opened 2020-12-29
add new Game method to load game from JSON or DB

michaeljb:
cc @jenf

### DIFF
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -53,20 +53,7 @@ module View
       return if game_id == @game&.id &&
         ((!cursor && @game.actions.size == @num_actions) || (cursor == @game.actions.size))
 
-      # Some Hotseat games don't have player ids, use names instead.
-      players = @game_data['players'].map { |p| [p['id'] || p['name'], p['name']] }.to_h
-      # Back compatibility, make hotseat games continue to work after the change to play names
-      if actions&.first&.dig(:entity_type) == 'player' && actions.first[:entity].is_a?(String)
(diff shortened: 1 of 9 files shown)


assets/app/view/game_page.rb
```diff
@@ -53,20 +53,7 @@ def load_game
       return if game_id == @game&.id &&
         ((!cursor && @game.actions.size == @num_actions) || (cursor == @game.actions.size))
 
-      # Some Hotseat games don't have player ids, use names instead.
-      players = @game_data['players'].map { |p| [p['id'] || p['name'], p['name']] }.to_h
-      # Back compatibility, make hotseat games continue to work after the change to play names
-      if actions&.first&.dig(:entity_type) == 'player' && actions.first[:entity].is_a?(String)
-        players = @game_data['players'].map { |p| [p['name'], p['name']] }.to_h
```
michaeljb, line 60: Is it important to keep this part for backward compatibility?

tobymao: na

lib/engine/game/base.rb
```diff
@@ -30,6 +30,46 @@
 
 module Engine
   module Game
+    def self.load(source, at_action: nil, **kwargs)
+      hash_game = nil
+      db_game = nil
+      case source
```
tobymao, line 36: maybe game_data = case source
when Hash, String

{ }...
else
{] ...
end

michaeljb: Done in 9d89a026f6e0afd0eed322b20ea3cdf3dcfc5531

tobymao (approved):

PR #2909, opened 2020-12-30
Ability train_discount improvement

perwestling:
Add missing documentation.

Add new attribute that will control if the private with this ability
is closed or not when used up.

### DIFF
--- a/lib/engine/ability/README.md
+++ b/lib/engine/ability/README.md
@@ -200,6 +200,15 @@ Modify train buy in some way.
 - `face_value`: If true, any inter corporation train buy must be at
   face value. Default false.
 
+## train_discount
+
+Discount the train buy cost. The `count` attribute specify how many times the discount can be used.
+
+- `discount`: Discount amount. If > 1 this is an absolute amount. If 0 < amount < 1 it is the fraction, e.g. 0.75 is a 75% discount.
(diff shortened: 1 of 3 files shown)


jenf:
Changes look good, just question if this should be generic to all, but that's for the next user :)

jenf (approved):

perwestling:
@jenf The train discount is used by 18AL, 18MS and 18SJ. 18AL and 18SJ use a 50% discount for any train buy while 18MS use $100 discount for 3 and 4 trains.

Now this is part of special_train_buy.rb, and only appear if the train_discount ability is present, so it should not affect any other title. The effect in game is that the train discount ability appear as a button, as usual, and if clicked on the purchase price of the train buy is modified in the GUI. That was implemented for 18AL.

Was a bit surprised that documentation was not there - it might have been lost in some commit?
But the main reason for this commit was that 18SJ had a train discount in which the private did not close when the train discount was used. That is different from 18AL and 18MS.

perwestling:
So should the closed_when_used_up be a general attribute for train_discount? 

Well, as the train_discount code do close the private in its code, and this seems to be non-general way of treating it, I think it is a good idea to make the configurable. Maybe there should be a general mechanism for that instead but I chose this way.

PR #2916, opened 2020-12-30
simplify game_spec.rb, just test everything in fixtures/ dir

michaeljb:
* expected results must now be set in the fixture's JSON file (some fixtures
  needed updating; any that needed updating now have the result that was
  originally in game_spec.rb)
* rewrite game_spec.rb to load all JSON fixtures; went from running 34 tests to
  36 tests; fixtures 16852 (1817) and 17510 (18MS) didn't have tests defined and
  so were missed, but we won't have to worry about that happening again


tobymao (approved):

tobymao:
nice

PR #2918, opened 2020-12-31
allow games to catch and store exceptions when processing actions

michaeljb:
Use this to show more when rendering a broken game:

* always show the broken action
* provide full link to the broken point in the game
* show the game as it was at the last valid action

Validate/migrate scripts still expect the exception to be raised, but now they
could be modified to get more info on broken games for faster debugging (e.g.,
right now the produced validate.json just shows the exception message, not the
broken action or a link for the last valid action or anything like that).

-----

I originally added `game_error` so that more error messages could link to the last valid action without needing to pass it in everywhere a GameError was raised. With this change, the game page can show the error message, the broken action, and the last valid game state for *all* errors, not just those occurring in a scope with access to a Game instance.

~I do have another commit that actually gets rid of all `game_error` calls, but left it off this PR for now because it's both large and not very interesting - https://github.com/tobymao/18xx/commit/93b77954732bc4aeb7826c29df3b9de6bb5ee733 (though there is one interesting bit, rubocop caught an unreachable line: at https://github.com/tobymao/18xx/commit/93b77954732bc4aeb7826c29df3b9de6bb5ee733#diff-4240657053470220def50bb227cd5b94d88ab4d805eec718b0f3327c85816406L78)~

I went ahead and added that commit to this PR, doesn't seem worth it to leave `game_error()` around if it's not doing anything any more. I kept the commits separate so that the first one with the actual interesting changes can still be examined by itself.

### Old

Screenshot (1)

### New

Screenshot (2)

### New with `?debug=1`

Screenshot (3)

PR #2922, opened 2020-12-31
Initial commit for 1824 Cisleithania

perwestling:
1824 is a 2 or 3 player variant for 1824, that uses an alternative
map, and removes some privates, corporations, changes cert limits,
bank, starting cash etc for 3 players compared to 1824 base game.

This commit does not contain the reduction stuff but is at least
loadable and the Cisleithania is handled as an optional rule.
Note 1! If player count is invalid (e.g. Cislethania optional rule)
a Game Error is thrown but this will not be handled gracefully.
Note 2! Optional rules "Goods Time" is not yet implemented, but
preparation is done.

Also corrects purchasable_unsold_companies which did crash at
start of 1824 or 1824 Cislethania. Moved it to base class (default
returning empty list) so that 1824 can override it.


tobymao:
do you still need the custom corporation code?

michaeljb (changes requested):
All `game_error(msg)` calls need to be replaced with `raise GameError, msg` now (see #2918)

michaeljb:
> All game_error(msg) calls need to be replaced with raise GameError, msg now (see #2918)

I went ahead and fixed this in 4e79dccc 

But looks like there's merge conflicts, I guess take care to not use `game_error` when resolving them

perwestling:
New attempt. Think all GameError is correct this time.


> do you still need the custom corporation code?

Maybe in the future. How special handling of shares in Regional Corporations to solve, but not at the moment.

lib/engine/game/g_1824.rb
```diff
@@ -33,6 +31,144 @@ class G1824 < Base
...
+        players.each do |player|
+          bank.spend(CASH_CISLEITHANIA[@players.size], player)
+        end
+      end
+
+      def init_corporations(stock_market)
+        return super unless option_cislethania
+
+        # Remove Coal Railway C4 (SPB), Regional Railway BH and SB
+        CORPORATIONS.reject! { |c| %w[SPB SB BH].include?(c['sym']) }
```
tobymao, line 88: you should not mutate the constants! fix all of these instances

lib/engine/game/g_1824.rb
```diff
@@ -89,14 +211,211 @@ def setup
...
+      def validate_optional_rules(optional_rules)
+        if optional_rules&.include?(:cisleithania)
+          raise GameError 'Cisleithania optional rule is for 2-3 players' if @players.size > 3
+          raise GameError 'Cannot use Cisleithania and Goods Time at the same time' if option_goods_time
+        elsif @players.size < 3
+          raise GameError '2 player count requires to use Cisleithania optional rule'
+        end
+      end
+
+      def mountain_railway_definition(index)
```
tobymao, line 239: put all of these definitions in constants not in methods like this

spec/lib/engine/games/config_spec.rb
```diff
@@ -9,8 +9,13 @@ module Engine
   players = ('a'..'z')
 
   Engine::GAMES.each do |title|
+    min_p = players.take(Engine.player_range(title).min)
+
+    # 1824 2 player is only supported with Optional Rule Cisleithania activated
+    min_p = players.take(3) if title.to_s == 'Engine::Game::G1824'
```
tobymao, line 15: this doesn't seem to be the right approach

michaeljb (dismissed):

perwestling:
@michaeljb It's ok. Just wanted you to check that I did not introduce any new GameError error.

spec/lib/engine/games/config_spec.rb
perwestling: @tobymao I suppose I should maybe assume that Cisleithania is always used if 2 players? So the optional rules is always active for 2.
The optional rule can dictate which flavor of 1824 you get for 3 players. 
And for 4-6 players I can ignore the optional rule.

This would mean the above code can be removed.

Also the 'Goods Times' optional rule, which is only relevant for 3-6 players, and not when using Cisleithania I guess can just be ignored.  
Can just do appropriate logging and remove the error checks for this.

tobymao: that's a good idea

lib/engine/game/g_1824.rb
perwestling: fixed

Had to create location_name method to be able to handle LOCATION_NAMES changes.

lib/engine/game/g_1824.rb
perwestling: Have attempted to fix this.

spec/lib/engine/games/config_spec.rb
perwestling: Ok, attempt done.

lib/engine/game/g_1824.rb
```diff
@@ -33,6 +29,173 @@ class G1824 < Base
...
+      def bank_cash
+        return super unless option_cisleithania
+
+        BANK_CASH_CISLEITHANIA[@players.size] - @players.sum(&:cash)
+      end
+
+      def init_bank
+        return super unless option_cisleithania
+
+        Engine::Bank.new(BANK_CASH_CISLEITHANIA[@players.size], log: @log)
```
tobymao, line 77: better to change base to have overridable bank_cash

tobymao (approved):

PR #2938, opened 2020-12-31
make Engine::Game.load gentler on the db and more selective about kwargs

michaeljb:
* if a param is given (like actions), no need to hit the db
* don't allow users to pass in arbitrary title, names, or id, it doesn't make
  any sense to load a game while changing those params

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -31,38 +31,44 @@ require_relative '../player_info'
 
 module Engine
   module Game
-    def self.load(data, at_action: nil, **kwargs)
-      game_data =
-        case data
-        when String
-          return load(JSON.parse(File.exist?(data) ? File.read(data) : data), at_action: at_action, **kwargs)
-        when Hash
-          {
-            title: data['title'],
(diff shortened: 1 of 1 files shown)


lib/engine/game/base.rb
```diff
@@ -31,38 +31,44 @@
...
-            names: data.ordered_players.map { |u| [u.id, u.name] }.to_h,
-            id: data.id,
-            actions: data.actions.map(&:to_h),
-            pin: data.settings['pin'],
-            optional_rules: data.settings['optional_rules'] || [],
-          }
-        end
-      title = game_data.delete(:title)
-      names = game_data.delete(:names)
-      game_data.merge!(kwargs)
```
michaeljb, line 62: I ended up deciding I didn't like dealing with the merge and deleting certain items to avoid them merging with kwargs, so I got rid of the hash and just explicitly set all of the relevant variables. `strict` and `disable_user_errors` can still be passed as kwargs to `load` and they will be forwarded on to `initialize`

michaeljb:
Full local validation is complete, and rubocop fix is pushed.

PR #2943, opened 2020-12-31
clean up more game_error stragglers missed in #2918

michaeljb:


michaeljb:
since github doesn't turn PR/Issue numbers into links when they're in the title, just want to make a comment linking this to #2918

Issue #2855, opened 2020-12-28
18MEX - track step blocks even though there's no valid track action


michaeljb:
https://18xx.games/game/12969?action=152

CHI owns KCMO, and KCMO hasn't used its ability, so this step blocks; however, KCMO cannot use its ability as Copper Canyon (F5) has already had a tile placed.

`Step::18Mex::Track#remaining_tile_lay?` should probably check for `copper_canyon_hex_empty?`, like `Step::18Mex::SpecialTrack#tile_lay_abilities` does, or even better, if we could genericize a tile-laying ability being removed when its corresponding hex is laid, then this could be handled in the config with an extra field on KCMO's ability.

(I am considering making a new issue for tile-laying abilities and a separate one for teleport abilities and throwing checklists in them, but didn't want to interrupt what I'm doing now to take the time to do that and didn't want to forget about this particular case, so just making another issue for now)

Issue #2105, opened 2020-10-29
18MS: Tile laying private ability skips token lay.


JLamkin:
In 18MS with the two privates that give you either 1 or 2 extra tile lays. I found a bug with order of operations.

If in the first OR you lay both of your tiles, than buy in the private and lay the extra 1 or 2 tiles it skips the tokening step.
However, if you buy in the private than lay both of your tiles, than lay the extra 1 or 2 tiles it gives you the option to lay the token.

Not a huge bug but thought I would point it out.

crericha:
May be fixed by the special_track/tracker rework done for teleport. Needs to be tested.

michaeljb:
I think that for this to work, we need the track step to block not only for abilities the active corporation currently has, but also if the active corporation could possibly buy a company that has a tile lay ability with `when: 'track'`.

Issue #1793, opened 2020-10-06
Bug 1836Jr30 - System allows private company power for tile placement to be exercised out of operational order when two corporations are owned by one player.


Liturgista:
Bug. 
In OR4.1 in game 11084 (1836Jr30), system allowed placement of additional free tile by GCL for ESF power out of turn  - in between B operation and NBDS operation. This caused a problem with tile laying for GCL during its operation (after NBDS). Players had to backtrack and redo operations for B, NBDS and then GCL to fix and continue as intended.

PR #2947, opened 2021-01-01
Ability timing refactor

michaeljb:
Goal
* special abilities should work only at the times allowed by the rules
* those times should be succinctly listed in the ability config/declaration

Current Status

* the abilities on this branch are currently designed to not break any games, so                                                                                                                                                                                                                                                  in many cases they have times explicitly defined that are actually against the                                                                                                                                                                                                                                                  rules but match current 18xx.games implementation, e.g., 18Chesapeak's B&S and                                                                                                                                                                                                                                                  C-P privates have `"when": ["SpecialTrack", "other_or"]`; to change it to                                                                                                                                                                                                                                                       match rules would be `"when": "Track"`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        

Changes
* added `on_phase`; used for abilities triggered by a certain/train phase (such
  as companies that close on the 6T instead of 5T)
* `when`/`time` can now use the last part of the name of the step class to
  indicate that game step, eg "TrackAndTroken", in some cases this allows for
`abilities()` to match the ability without passing a `time`; they can still
use an arbitrary string to only match when called
* `"when": "Track"` ensures that the ability is only available during the
Track phase, and also prevents the Track step from passing if there is a
Track ability available
* with `"when": "Track"`, 1662 out of 2426 18Chesapeake games are
broken (69%)
* with `migrate_abilities.rb`, 1356 of those 1662 are successfully migrated,
leaving 306 pinned
* `ability_blocking_step`: only include steps that include the module "Tracker";
if/when more Steps are added, this check can be expanded to a list of
modules/classes to check for; the main thing is to start with a list of
allowed steps and filter down, instead of looking at the full list of possible
steps
* set prealpha company configs with timings that should work (need to verify
with game devs)
* change "train" time to "buying_train" and "bought_train" to be more
clear (TODO: can use "Train" to match the step?)
* some cleanup for 18GA, 18MS 18TN custom steps; some of the logic is now
handled by using `"when": "Track"` for the relevant company configs
* `allow_passive` is now `passive_ok`


lib/engine/ability/README.md
```diff
@@ -17,11 +17,27 @@ These attributes may be set for all ability types
   order for the ability to be active. Either "player" or
   "corporation".
 - `remove`: Game phase when this ability is removed
-- `when`: The game step or phase when this ability is active
 - `count`: The number of times the ability may be used
 - `count_per_or`: The number of times the ability may be used in each OR; the
   property `count_this_or` is reset to 0 at the start of each OR and increments
   each time the ability is used
+- `when`: The game steps, phases, or special times when this ability is
```
michaeljb, line 24: The doc updates here are WIP and not yet complete

lib/engine/game/base.rb
```diff
@@ -1924,30 +1924,62 @@ def ability_right_owner?(entity, ability, owner_type)
...
       end
 
       def ability_usable_this_or?(ability)
         !ability.count_per_or || (ability.count_this_or < ability.count_per_or)
       end
 
-      def ability_right_time?(ability, time, strict_time)
-        return false if strict_time && !ability.when
-        return true unless time
+      def ability_right_time?(ability, time, allow_passive)
```
michaeljb, line 1935: The changes in this function are the main point of this PR.

lib/engine/game/base.rb
```diff
@@ -1924,30 +1924,62 @@ def ability_right_owner?(entity, ability, owner_type)
...
+
+        return false if @round.operating? &&
+                        ability.owner.owned_by_corporation? &&
+                        (@round.current_operator != ability.corporation) &&
+                        !ability.when?('other_or')
+
+        return false if time && !ability.when?(time)
+
+        ability.when.any? do |ability_when|
+          case ability_when
```
michaeljb, line 1965: This `case` block was kind of the main idea I had for this PR, with an ability listing out strings that describe when it can be used, and then bits of code to tell when that string matches the current game state.

lib/engine/player.rb
```diff
@@ -35,6 +35,10 @@ def player
       self
     end
 
+    def corporation
```
michaeljb, line 38: Added so that an ownable can call `.corporation`

lib/engine/step/buy_company.rb
```diff
@@ -14,12 +14,9 @@ def actions(entity)
         return [] if entity.minor?
         return blocks? ? ACTIONS : ACTIONS_NO_PASS if can_buy_company?(entity)
 
-        return PASS if blocks? &&
-                       entity.corporation? &&
-                       @game.abilities(entity,
-                                       time: 'owning_corp_or_turn',
-                                       owner_type: 'corporation',
-                                       strict_time: true)
+        if blocks? && entity.corporation? && @game.abilities(entity, time: 'owning_corp_or_turn', allow_passive: false)
```
michaeljb, line 17: This line is one of two places where `allow_passive: false` is used. This is ensuring that a company's ability can be used if the company is bought in right before the end of the turn. This doesn't apply for always-on / passive abilities like terrain-discount companies such as Sumitomo Mines Railway in 1889, but could apply for many types of abilities, which is why no ability type is passed in.

lib/engine/step/g_18_ga/track.rb
```diff
@@ -25,7 +25,8 @@ def process_lay_tile(action)
 
         def remaining_tile_lay?(entity)
           can_lay_tile?(entity) ||
-          (@game.p2_company.owner == entity && @game.abilities(@game.p2_company, :tile_lay)&.count&.positive?)
+            (@game.p2_company.owner == entity &&
+             @game.abilities(@game.p2_company, :tile_lay, time: 'Track')&.count&.positive?)
```
michaeljb, line 29: Most likely this whole `remaining_tile_lay?` function can be removed once the relevant company has been changed to have `"when": "Track"` instead of `"when": ["SpecialTrack", "Track"]`

lib/engine/step/special_track.rb
```diff
@@ -76,12 +76,17 @@ def potential_tiles(entity, hex)
...
-        ability = @game.abilities(entity, :tile_lay, time: 'sold', &block) if @round.respond_to?(:just_sold_company) &&
-          entity == @round.just_sold_company
-        ability || @game.abilities(entity, :tile_lay, time: 'track', &block)
+        if @round.respond_to?(:just_sold_company) && entity == @round.just_sold_company
+          ability = @game.abilities(entity, :tile_lay, time: 'sold', **kwargs, &block)
+        end
+
+        ability ||
+          @game.abilities(
+            entity, :tile_lay, time: %w[SpecialTrack %current_step% owning_corp_or_turn], **kwargs, &block
```
michaeljb, line 88: After `ability_right_time?`, this function is probably the next most important logic change in the PR

lib/engine/step/tracker.rb
```diff
@@ -11,6 +11,9 @@ def setup
       end
 
       def can_lay_tile?(entity)
+        step_name = self.class.name.split('::').last
+        return true if tile_lay_abilities(entity, time: step_name, allow_passive: false)
```
michaeljb, line 15: This allows for abilities to generally block a Track-laying step from moving on (see https://github.com/tobymao/18xx/pull/2947/files#r550727703)

lib/engine/step/tracker.rb
```diff
@@ -78,7 +83,7 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
         discount = 0
         teleport = false
 
-        tile_lay_abilities(entity) do |ability|
+        tile_lay_abilities(entity, allow_passive: true) do |ability|
```
michaeljb, line 86: `allow_passive` defaults to `true`, I should get rid of this

migrate_abilities.rb
```diff
@@ -0,0 +1,290 @@
...
+    actions.delete(broken_action)
+    return
+
+  elsif game.active_step.is_a?(Engine::Step::Route) &&
+     broken_action['type'] == 'place_token' &&
+     prev_action['type'] == 'pass'
+    actions.delete(prev_action)
+    return
+
+  elsif (company = game.company_by_id(broken_action['entity'])) &&
```
michaeljb, line 68: this `elsif` branch does a lot of the work for fixing up 18Chesapeake games where people bought/used C-P and B&S at the wrong time of the turn by shifting actions around where possible (i.e., if in the actual game they bought the company in after running routes and used it immediately, but they had enough money at the track step, this will move those actions back to the track step)

michaeljb: Toby told me not to bother with this approach on Slack but I was already in too deep and wanted to solve it and make it work, even if we end up not using it 😝 

On one hand, 1662 games is a lot of games to pin, but on the other hand 1356 is definitely a lot of games to migrate on the prod server. 🤷 

Just using the other branches in this `repair()` does handle some cases where it's just a matter of deleting or inserting a pass action

michaeljb: Also eventually I want to get some of these changes (like the forking support) into the  main migrate script but I've been kicking that can down the road

michaeljb:
I'm going to tryadding a new `on_phase` parameter, hopefully be able to clean up the logic with the abilities that have something like `"when": ["6", "other_or"]` right now.

Making the `passive` stuff more explicit in the configs might also help, will try that next if `on_phase` seems to work well.

lib/engine/game/g_18_ga.rb
```diff
@@ -83,8 +83,8 @@ def operating_round(round_num)
           Step::G18GA::SpecialToken,
           Step::G18GA::BuyCompany,
           Step::HomeToken,
-          Step::G18GA::SpecialTrack,
-          Step::G18GA::Track,
```
michaeljb, line 87: The Track and SpecialTrack classes for 18GA were just implementing the correct timing that can now be achieved with `"when": "Track"`. 18MEX, 18MS, and 18TN also enforce the timing and so their custom Step classes need to be looked at as well.

migrate_abilities.rb
michaeljb: > On one hand, 1662 games is a lot of games to pin, but on the other hand 1356 is definitely a lot of games to migrate on the prod server. 🤷

Solution: we'll be able to proxy a connection to the DB from my machine to run the migration logic locally

lib/engine/game/g_18_chesapeake.rb
```diff
@@ -34,11 +34,20 @@ def init_share_pool
...
+          end
+        end
+      end
+
       def action_processed(action)
         case action
         when Action::LayTile
-          check_special_tile_lay(action, columbia)
-          check_special_tile_lay(action, baltimore)
+          flush_log!('special_tile_lay')
```
michaeljb, line 50: With the timing changes, having this check after the action processed meant the action was already used up and removed by the time `action_processed` is called, but it looked weird with the log lines about losing the ability to lay happening before the tile lay.

This initially was behaving strangely with the first place `queue_log!` was used (the phase change in `train.rb`), so I changed queued_log to be a hash that could accept different string descriptions and use them as keys for the log messages to display later.

michaeljb:
I updated the ability readme, might take this out of "draft" soon.

lib/engine/game/base.rb
```diff
@@ -1983,35 +1985,60 @@ def ability_right_owner?(entity, ability, owner_type)
...
       end
 
       def ability_usable_this_or?(ability)
         !ability.count_per_or || (ability.count_this_or < ability.count_per_or)
       end
 
-      def ability_right_time?(ability, time, strict_time)
-        return false if strict_time && !ability.when
-        return true unless time
+      def ability_right_time?(ability, time, on_phase, passive_ok)
```
michaeljb, line 1996: The changes in this function are the main point of this PR

michaeljb:
Rebased, merge conflicts resolved

assets/app/view/game/buy_trains.rb
```diff
@@ -70,12 +70,12 @@ def render
...
         @corporation = step.current_entity
         if @selected_company&.owner == @corporation
-          @ability = @game.abilities(@selected_company, :train_discount, time: 'train')
+          @ability = @game.abilities(@selected_company, :train_discount, time: 'buying_train')
         end
 
         @depot = @game.depot
 
-        available = step.buyable_trains(@corporation).group_by(&:owner)
+        available = step.buyable_trains(@corporation).group_by(&:owner) || []
```
tobymao, line 78: this can never be nil why did you change this

lib/engine/game/base.rb
```diff
@@ -1375,16 +1376,17 @@ def ipo_reserved_name(_entity = nil)
         'IPO Reserved'
       end
 
-      def abilities(entity, type = nil, time: nil, owner_type: nil, strict_time: false)
+      def abilities(entity, type = nil, time: nil, on_phase: nil, passive_ok: nil)
```
tobymao, line 1379: we don't need to make the change now, but i wanted to make note that the call to all_abilities is very expensive and is called quite often for garbage collection since it makes a ton of of new arrays

lib/engine/game/base.rb
tobymao: should time any be default?

lib/engine/game/base.rb
```diff
@@ -1991,35 +1993,60 @@ def ability_right_owner?(entity, ability, owner_type)
...
       def ability_blocking_step
-        @round.steps.find { |step| step.blocks? && !step.passed? }
+        @round.steps.find do |step|
+          # currently, abilities only care about Tracker, the is_a? check could
+          # be expanded to a list of possible classes/modules when needed
+          step.blocks? && !step.passed? && step.is_a?(Step::Tracker)
+        end
+      end
+
+      def ability_blocking_step_and_name
```
tobymao, line 2046: why is this needed?

lib/engine/game/base.rb
```diff
@@ -1991,35 +1993,60 @@ def ability_right_owner?(entity, ability, owner_type)
...
+          end
         end
       end
 
       def ability_blocking_step
-        @round.steps.find { |step| step.blocks? && !step.passed? }
+        @round.steps.find do |step|
+          # currently, abilities only care about Tracker, the is_a? check could
+          # be expanded to a list of possible classes/modules when needed
+          step.blocks? && !step.passed? && step.is_a?(Step::Tracker)
```
tobymao, line 2042: why?

lib/engine/game/base.rb
```diff
@@ -1353,15 +1354,15 @@ def train_help(_runnable_trains)
         []
       end
 
-      def queue_log!
+      def queue_log!(log_name)
```
tobymao, line 1357: make this default nil so that you can queue_log without specify name and change the trivial references to use that

lib/engine/game/base.rb
tobymao: do you even need ability_blocking_step?

lib/engine/game/base.rb
tobymao: step.type

lib/engine/ownable.rb
```diff
@@ -14,6 +14,10 @@ def player
       owner.is_a?(Player) ? owner : owner&.player
     end
 
+    def corporation
+      is_a?(Corporation) ? self : owner&.corporation
```
tobymao, line 18: corporation?

lib/engine/ownable.rb
```diff
@@ -14,6 +14,10 @@ def player
...
+    end
+
```
tobymao, line 20: change the bottom here too to be owner.corporation?

lib/engine/step/tracker.rb
```diff
@@ -11,6 +11,9 @@ def setup
       end
 
       def can_lay_tile?(entity)
+        step_name = self.class.name.split('::').last
```
tobymao, line 14: is this needed?

tobymao: and if so you just want to call type

lib/engine/step/tracker.rb
```diff
@@ -30,16 +33,18 @@ def get_tile_lay(entity)
...
+        raise GameError, 'Cannot lay a yellow now' if tile.color == :yellow && !(tile_lay && tile_lay[:lay])
 
         lay_tile(action, extra_cost: tile_lay[:cost])
         @upgraded = true if action.tile.color != :yellow
         @laid_track += 1
       end
 
-      def tile_lay_abilities(entity, &block)
-        @game.abilities(entity, :tile_lay, &block)
+      def tile_lay_abilities(entity, **kwargs, &block)
```
tobymao, line 44: same here

lib/engine/game/base.rb
michaeljb: Could we make a cache that's cleared after every action is processed? Any other ideas for ways to improve that?

assets/app/view/game/buy_trains.rb
michaeljb: don't remember, overlooked it when I was reviewing this myself last night, will delete

lib/engine/game/base.rb
tobymao: yea but it's difficult because abilities can change when companies are closed etc

lib/engine/ownable.rb
michaeljb: Should the `is_a?(Player)` calls here be replaced with `.player?`? This has the odd side effect of ownable things needing to implement basically everything on `Entity`. Should `Ownable` have `include Entity`?

Anyway with `corporation?` or with doing all the changes I mentioned, I'm getting a bunch of errors, will come back to this one later.

ryandriskel:
If I am reading this correctly, in 18CO,
```
		{
			"sym": "DNP",
			"name": "Denver, Northwestern and Pacific Railroad",
			"value": 50,
			"revenue": 10,
			"desc": "An owning Corporation may return a station token to its charter to gain the token cost. Corporation must always have at least one token on the board. Action closes the company or closes on purchase of “5” train.",
			"abilities": [
				{
					"type": "return_token",
					"owner_type": "corporation",
					"count": 1,
					"reimburse": true
				}
			]
		},
```

This ability cannot be triggered until `when: ['owning_corp_or_turn']` is added to the config as it is not a passive ability.

michaeljb:
@ryandriskel yes you are correct

lib/engine/ownable.rb
michaeljb: Got this working, can't be 100% sure but the first time around I think I was missing some `&`s when `owner` might be `nil`

lib/engine/game/g_18_chesapeake.rb
michaeljb: idk what happened but I found I was actually able to eliminate the string arg to `queue_log!`

lib/engine/game/base.rb
michaeljb: I think no, using any as the default opens up more possibilities for abilities to be used out-of-turn before the check for `other_or` is done

lib/engine/game/base.rb
michaeljb: Yeah, it's used to sort-of figure what the active step is without calling `active_step`, since that triggers a loop of `active_step` -> what actions are available? -> what abilities are active? -> `ability_right_time?` -> `active_step`.

In 18GA, Per avoided that with these track step definitions:

* https://github.com/tobymao/18xx/pull/2947/files#diff-aa552812d2ce056cec0285385023d04fa1e6fff31b3ae39a458d2ea126bebe7e
* https://github.com/tobymao/18xx/pull/2947/files#diff-9630d124a4a448e25c7be52e26d00d79c9e1b87901e654a80c2f4effab7fb08a

Maybe something more like this could work, but company abilities that can be used any time during the corporation’s OR turn make it more challenging.

michaeljb:
@ryandriskel I tried adding `when: ['owning_corp_or_turn']` for you on this branch, but it breaks the 18CO fixtures; I didn't dig in more on that yet, but I'd guess that has to do with BuyCompany blocking for `owning_corp_or_turn` abilities:

https://github.com/tobymao/18xx/blob/bce27c91bb4be48a790b11ca4897c0966bcb583e/lib/engine/step/buy_company.rb#L17-L22

michaeljb:
@tobymao ready for one more look tomorrow, I resolved most of the your comments with some code changes, and posted some response comments on the last two. Full validation against 10,977 games is passing.

tobymao (approved):

michaeljb:
Now that this is merged, I'm unassigning from several ability timing issues. I still plan to fix them soon-ish, but don't want to block anyone else if they want to take them.

Will mention them here so they see the link to this PR: #1558, #1704, #1725, #1793, #2855, #2105,

PR #2955, opened 2021-01-01
1856 nationalization core

ventusignis:
This isn't everything for nationalization but this does do a good chunk of it and the goal is that I'll be able to add on extra functionality later with small PRs

What this does do is:
- Nationalization starts with the first 6T
- Player order starts with the player who bought the 6T
- Companies are forced to repay loans
- Presidents of companies, in player order, choose to cover the difference to payoff loans or let the company merge in
- Companies when they merge transfer leftover cash from paying off loans and trains
- Tokens of merged corporations are given to the national
- Trains of merged corporations are given to the national
- Shares of merged corporations are traded 2-1 for national shares
- Market price of national is the average* of the merged corporations prices


lib/engine/g_1856/corporation.rb
```diff
@@ -16,6 +17,7 @@ def initialize(game, sym:, name:, **opts)
         @started = false
         @escrow = nil
         super(sym: sym, name: name, **opts)
+        @all_shares = shares_by_corporation[self].dup
         @capitalization = nil
       end
```
ventusignis, line 23: The floating behavior in 1856 is going to keep the 1856::Corporation class around although escrow logic can probably be pulled into the base classes at some point which could reduce the amount of custom 1856 code

lib/engine/ability/unparrable.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
```
tobymao, line 5: why is this needed, can't you just define can_par? in game like others?

lib/engine/g_1856/corporation.rb
tobymao: what about it?

lib/engine/game/g_1856.rb
```diff
@@ -355,6 +385,346 @@ def stock_round
...
+          @loans << corp.loans.pop(loans_repaid)
+          @log << "#{corp.name} repays #{format_currency(amount_repaid)} to redeem #{loans_repaid} loans"
+        end
+      end
+
+      def merge_major(major)
+        @national_formed = true
+        @log << "-- #{major.name} merges into #{national.name} --"
+        # Trains are transferred
+        major.trains.clone.each do |t|
```
tobymao, line 432: dup

lib/engine/game/g_1856.rb
```diff
@@ -355,6 +385,346 @@ def stock_round
...
+        home_bases = @nationalized_corps.map do |c|
+          nationalize_home_token(c, create_national_token)
+        end
+        # So the national will get 11 tokens if and only if all 11 majors merge in
+        remaining_tokens = [national_token_limit - home_bases.size, 0].max
+
+        # Other tokens second, ignoring duplicates from the home token set
+        @nationalized_corps.each do |corp|
+          corp.tokens.each do |token|
+            next if token.city.nil? || home_bases.include?(token.city.hex)
```
tobymao, line 620: !token.city

lib/engine/game/g_1856.rb
```diff
@@ -355,6 +385,346 @@ def stock_round
...
+        @national.tokens.each do |token|
+          city = token.city
+          token.remove! if cities.include?(city)
+        end
+      end
+
+      # Convert the home token of the corporation to one of the national's
+      # Return the nationalized corps home hex
+      def nationalize_home_token(corp, token)
+        if token.nil?
```
tobymao, line 683: unless token

lib/engine/game/g_1856.rb
ventusignis: Is `dup` generally preferred to `clone`, or just in this case?

tobymao: dup is just a soft copy, clone does all internals which is not necessary

lib/engine/game/g_1856.rb
ventusignis: Changed

lib/engine/game/g_1856.rb
ventusignis: OK thanks, I've changed this

lib/engine/game/g_1856.rb
ventusignis: Changed

lib/engine/ability/unparrable.rb
ventusignis: I did some changes to remove this ability; I also defined `float_str` which will communicate the same information to the player
image

lib/engine/g_1856/corporation.rb
ventusignis: You've mentioned a desire to put the custom logic in `G1856::Coporation` into the 1856 game file before. I have some ideas on how to incrementally remove some of the custom classes for 1856 but I think that the floating behavior is going to keep the custom 1856 corporation around for that purpose

lib/engine/g_1856/corporation.rb
```diff
@@ -16,6 +17,7 @@ def initialize(game, sym:, name:, **opts)
         @started = false
         @escrow = nil
         super(sym: sym, name: name, **opts)
+        @all_shares = shares_by_corporation[self].dup
```
tobymao, line 20: is this really necesasry? i see it's for total_shares, but can't your 1856 game class count this without having to dup the shares?

lib/engine/g_1856/corporation.rb
```diff
@@ -45,11 +47,8 @@ def capitalization_type_desc
         CAPITALIZATION_STRS[@capitalization || _capitalization_type]
       end
 
-      def release_escrow!
-        @log << "Releasing #{@game.format_currency(@escrow)} from escrow for ${@name}"
-        @cash += @escrow
-        @escrow = nil
-        @capitalization = :incremental
+      def total_shares
```
tobymao, line 50: this doesn't seem like it needs to be in corporation class

lib/engine/g_1856/corporation.rb
ventusignis: It's also used in `Engine::Game::G1856::national_issue_shares!` for converting the existing shares to 5% shares and for accessing the shares in `Engine::Game::G1856::national_share_swap` for distributing shares among players. I also have to solve how to give a player a 1-share presidents cert in one of the nationalization edge cases, but that's not solved yet

lib/engine/g_1856/corporation.rb
ventusignis: This is overriding `total_shares` from the base Corporation class

tobymao: what was wrong with the initial implementation?

ventusignis: It helps with some of the weird edge cases in 1856 but since I cut off this PR before I started to tackle them I'm going to remove that from here

tobymao (approved):

PR #2970, opened 2021-01-01
fix broken action message; don't just use the last action as the broken one

michaeljb:


lib/engine/game/base.rb
```diff
@@ -618,6 +618,7 @@ def process_action(action)
       rescue Engine::GameError => e
         @exception = e
         @actions |= [action]
```
michaeljb, line 620: I'm not sure if this line should be deleted or not--if the action is broken, should we prevent it from being added to `@actions`?

tobymao (approved):

lib/engine/game/base.rb
tobymao: deleted

PR #2980, opened 2021-01-02
Make Nieuw Zeeland green in 1817WO until it is parrable

ventusignis:
The thought behind this is that since 1867 has city slots that are not IPOable until green that the same visual style should be used in 1817WO for Nieuw Zeeland because that hex is unIPOable until green (phase 3)
image

### DIFF
--- a/lib/engine/game/g_1817_wo.rb
+++ b/lib/engine/game/g_1817_wo.rb
@@ -22,6 +22,7 @@ module Engine
         '1817 Rules' => 'https://drive.google.com/file/d/0B1SWz2pNe2eAbnI4NVhpQXV4V0k/view',
       }.freeze
       GAME_DESIGNER = 'Mark Voyer & Brennan Sheremeto'
+      EVENTS_TEXT = G1817::EVENTS_TEXT.merge('nieuw_zeeland_available' => ['Nieuw Zealand opens for new IPOs'])
 
       def self.title
         '1817WO'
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_1817_wo.rb
```diff
@@ -30,6 +31,15 @@ def self.title
...
         @new_zealand_city = hexes.find { |hex| hex.location_name == 'Nieuw Zeeland' }.tile.cities[0]
+        # Put an 1867 style green token in the New Zealand hex
+        @green_token = Token.new(nil, price: 0, logo: '/logos/1817/nz.svg', type: :neutral)
+        @new_zealand_city.exchange_token(@green_token)
+      end
+
+      def event_nieuw_zeeland_available!
+        # Remove the 1867-style green token from the New Zealand hex
+        @log << 'Corporations can now be IPOed in Nieuw Zeeland'
+        @green_token.amp(&:remove!)
```
tobymao, line 42: what is amp? you mean map?

did you test this code?

tobymao: shouldn't it just be @green_token.remove!

ventusignis: I tested then I cleaned up, which introduced the typo. I'm going to add a spec for 1817WO and change it to @green_token.remove!`

tobymao (approved):

PR #2982, opened 2021-01-02
Upgrade postgresql to 13

michaeljb:


michaeljb:
I've tested this with a local stack running in dev mode, the script took about 3min 15s to run. I want to test with a local stack running in prod mode, should be able to try that out a little later tonight

tobymao (approved):

PR #2983, opened 2021-01-02
Fix to make it possible to exchange for variant

perwestling:
In case there is a variant of a purchasable train we need to look
among the variants to allow it.

Before the change, when attempting to exchange for the variant train, it was not among the buyable_train_variants as the base variant was checked against.

### DIFF
--- a/lib/engine/step/train.rb
+++ b/lib/engine/step/train.rb
@@ -135,7 +135,7 @@ module Engine
       end
 
       def buyable_train_variants(train, entity)
-        return [] unless train_among_buyable?(entity, train)
+        return [] unless buyable_trains(entity).any? { |bt| bt.variants[bt.name] }
 
         variants = train.variants.values
         return variants if train.owned_by_corporation?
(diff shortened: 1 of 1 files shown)


perwestling:
There is still a bug left here.

Situation is D train cost 1100, with a discount of 300 for a 4 train.
Variant is E train cost 1300 (so 1000 after discount).

If corporation has 1100 and a 4 train, the exchange of an E train can be done, which cost 1000.
If corporation has 1000 and a 4 train, the exchange of an E train CANNOT be done. 

Probably because D train cannot be bought without.

So still needs some further fix.

perwestling:
It is needed to be able to exchange the variant train, so I will need it to make the fixture work for 18SJ.

The remaining problem I will look into further.

lib/engine/step/train.rb
```diff
@@ -135,10 +135,9 @@ def buyable_trains(entity)
       end
 
       def buyable_train_variants(train, entity)
-        return [] unless buyable_trains(entity).include?(train)
+        return [] unless train_among_buyable?(entity, train)
```
tobymao, line 138: this can be simplified

buyable_trains(entity).any? { |train| train.variants[train.name] }

perwestling: Fixed

tobymao (approved):

PR #2985, opened 2021-01-02
[18CZ] Feat(Company): Value is changeing from round to round

Zwergenpunk:
In 18CZ the values of companies is changing from round to round, so I have to change it during the game. 

Thus, it needs to be writable. 

Usage can be seen here:
https://github.com/Zwergenpunk/18xx/blob/c42b57cbc42c3cbd436d1637f3e69915cda67bf7/lib/engine/game/g_18_cz.rb#L91

### DIFF
--- a/lib/engine/company.rb
+++ b/lib/engine/company.rb
@@ -10,8 +10,8 @@ module Engine
     include Entity
     include Ownable
 
-    attr_accessor :desc, :max_price, :min_price, :revenue, :discount
-    attr_reader :name, :sym, :value, :min_auction_price, :treasury
+    attr_accessor :desc, :max_price, :min_price, :revenue, :discount, :value
+    attr_reader :name, :sym, :min_auction_price, :treasury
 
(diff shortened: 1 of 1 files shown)


tobymao:
it's not a discount but the actual value of the company gets more? so at the end of the game it will count more to the score?

Zwergenpunk:
Yeah. If you hold a company at the end of the Game it has a value of 120. and you can sell it up to the current value

ryandriskel:
An alternative is to override 18CZ `player_value(player)` function. 

For example in 18CO there is one company that has _no_ player value, even though it has a price of 120 for the auction:

```
def player_value(player)
        value = player.value
        return value unless drgr&.owner == player

        value - drgr.value
      end
```

Zwergenpunk:
But the companies do not really have a fixed value. The starting of a specific company is somewhat random. Of course I could calculate thr value of a player differently but wouldn't it be easier to adjust the company.

tobymao (approved):

PR #2987, opened 2021-01-02
Fix Spreadsheet view to display correct operating order text even when company has no operating history. Fixes #2912.

### LINKED ISSUE #2912: [18GA] Minor issue - Company order "number" not showing in Spreadsheet
* Multiplayer? Link to the game
 https://18xx.games/game/19189?action=347

* Hotseat? 
 https://gist.github.com/yoyoson/74b07d1d6dbd7ea2e510d3c2f237d784

It really helps us having a sample game to reproduce the issue.

**Describe the bug**
Looking under "Corporation" section of Spreadsheet tab, under "Order"  column. The companies are acting in the correct order so this is only a cosmetic issue. But the GA corp doesn't show the "3" as it should since it's acting third.

timpelican:


assets/app/view/game/spreadsheet.rb
```diff
@@ -303,8 +303,12 @@ def render_corporation(corporation, operating_order, current_round, index)
         order_props = { style: { paddingLeft: '1.2em' } }
         operating_order_text = ''
         if operating_order.positive?
-          corporation.operating_history.each do |history|
-            operating_order_text = "#{operating_order}#{history[0] == current_round ? '*' : ''}"
+          if corporation.operated?
```
tobymao, line 306: let's change it to 

operating_order_text = 
if corporation.operated?
  "#{operating_order}{corporation.operating_history[-1][0] == current_round ? "*" : ""}"
else
  operating_order.to_s
end

timpelican: `corporation.operating_history` is a hash, so I don't think `corporation.operating_history[-1]` works (at least, it doesn't for me).  

`corporation.operating_history.keys[-1] == current_round`

does do the right thing for me, and `keys[-1]` is defined but nil for empty hashes / non-operated companies, so it's possible to replace the whole conditional with one line:

`operating_order_text = "#{operating_order}#{corporation.operating_history.keys[-1] == current_round ? '*' : ''}"`

For non-operated companies, `nil != current_round`, and we do the right thing (get the operating_order number, but don't append the '*').

Is that ok?

tobymao:
yep

tobymao (approved):

PR #2998, opened 2021-01-02
[18CZ] Initial Map and some basic features

Zwergenpunk:
This PR does only affect 18CZ game intern code and  includes 
- the map
- train config
- companies
   - config
   - buy phase
   - value adjustment
- corporation 
   - config
   - tokens
   - par values
   - different entry points into game
- Stock Market config
- Phase Config
- Tiles + purple tiles


lib/engine/config/game/g_18_cz.rb
```diff
@@ -0,0 +1,1537 @@
...
+                        "J14",
+                        "I17"
+                    ],
+                    "color": "orange",
+                    "type": "large"
+                },
+                {
+                    "float_percent": 50,
+                    "sym": "Ug",
+                    "name": "ungar. Staatsbahn",
```
benjaminxscott, line 500: seems like it's called `Ungarische Staatsbahn` in the rules, why is it different here?

lib/engine/config/game/g_18_cz.rb
```diff
@@ -0,0 +1,1537 @@
...
+            "currencyFormatStr": "K%d",
+            "bankCash": 99999,
+            "certLimit": {
+                "3": 14,
+                "4": 12,
+                "5": 10,
+                "6": 9
+            },
+            "startingCash": {
+                "3": 350,
```
benjaminxscott, line 24: s.b. 380 for 3p

lib/engine/step/g_18_cz/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require_relative '../buy_sell_par_shares.rb'
+
+module Engine
```
tobymao, line 5: what is this for

Zwergenpunk: I copied too much. I removed the require statement

lib/engine/game/g_18_cz.rb
```diff
@@ -0,0 +1,160 @@
+# frozen_string_literal: true
+
+require_relative '../config/game/g_18_cz'
+require_relative 'base'
+
+module Engine
+  module Game
+    class G18CZ < Base
+      register_colors(brightGreen: '#c2ce33', beige: '#e5d19e', lightBlue: '#1EA2D6', mintGreen: '#B1CEC7',
```
perwestling, line 9: might be easier to read if format parameters in one column

lib/engine/game/g_18_cz.rb
```diff
@@ -0,0 +1,160 @@
...
+                      yellow: '#ffe600', lightRed: '#F3B1B3')
+
+      load_from_json(Config::Game::G18CZ::JSON)
+
+      DEV_STAGE = :prealpha
```
perwestling, line 14: :prealpha is default value so not needed.

lib/engine/game/g_18_cz.rb
```diff
@@ -0,0 +1,160 @@
...
+      GAME_LOCATION = 'Czech Republic'
+      GAME_RULES_URL = 'https://www.lonny.at/app/download/9940504884/rules_English.pdf'
+      GAME_DESIGNER = 'Leonhard Orgler'
+      GAME_PUBLISHER = nil
```
perwestling, line 18: Can put :lonny_games here

lib/engine/game/g_18_cz.rb
```diff
@@ -0,0 +1,160 @@
...
+      SELL_BUY_ORDER = :sell_buy
+      SELL_MOVEMENT = :down_block
+
+      HOME_TOKEN_TIMING = :operate
+
+      PAR_RANGE = {
+        1 => [50, 55, 60, 65, 70],
+        2 => [60, 70, 880, 90, 100],
+        3 => [90, 100, 110, 120],
+
```
perwestling, line 30: Extra line

lib/engine/game/g_18_cz.rb
```diff
@@ -0,0 +1,160 @@
...
+        'BTE' => 2,
+        'KFN' => 2,
+        'EKJ' => 1,
+        'OFE' => 1,
+        'BCB' => 1,
+        'MW' => 1,
+        'VBW' => 1,
+      }.freeze
+
+      EVENTS_TEXT = Base::EVENTS_TEXT.merge(
```
perwestling, line 53: Could possibly be done as STATUS instead. EVENT is more .... events. While STATUS is for status. But both work.

benjaminxscott: I think these should be events since they have triggers

https://github.com/tobymao/18xx/blob/f7222bd1611a80c8d0921197f2bf0bda42da96ab/lib/engine/game/g_18_cz.rb#L140

lib/engine/game/g_18_cz.rb
Zwergenpunk: I mean it will be added once it is in alpha, so why should I remove it?

lib/engine/step/g_18_cz/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
```
tobymao, line 2: this needs to require BuySellParShares

Zwergenpunk: Then I did not understand the last commment correctly. 
This is needed because every coroporations has different par values.

tobymao (approved):

PR #3003, opened 2021-01-03
[1882] make it more clear that P2 closes on phase 6

benjaminxscott:
P2 wasn't being closed since `close_companies()` was only called once, at phase 5

This patch creates a new event on phase 6 that calls `close_companies()` again to close P2

Also added more detail in the event copy to clarify those differences

Screenshot 2021-01-02 at 11 04 29 PM

### DIFF
--- a/lib/engine/config/game/g_1882.rb
+++ b/lib/engine/config/game/g_1882.rb
@@ -467,7 +467,10 @@ module Engine
       "name": "6",
       "distance": 6,
       "price": 630,
-      "num": 3
+      "num": 3,
+      "events": [
+        {"type": "close_remaining_companies"}
+      ]
     },
     {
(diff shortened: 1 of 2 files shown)


tobymao:
@jenf

michaeljb:
Looks like it closed just fine here? https://18xx.games/game/10526?action=271

Do you have an example game where it didn't close?

benjaminxscott:
Ah you're right and I was mistaken, looks like phase change logic handles privates with an explicit `close` ability , and otherwise `event_close_companies()` handles them

Would it make sense to include the changes in this branch to create a new event to make it more clear that P2 closes at a different phase than the rest?

tobymao:
i think i'll close for now

Issue #2996, opened 2021-01-02
[18CO] Denver & Rio Grande Railway Silverton Branch purchasable by corporation after it should have closed


PJBarns:
After the Denver & Rio Grande Railway Silverton Branch private closes and leaves the players hand, it seems to be purchasable from the market by clicking "Show companies from other players." 

Game state right before I attempted it:
https://18xx.games/game/20978?action=449

Hot-seat game right after I had the DSNG purchase it:
https://18xx.games/hotseat/hs_quzvabnr_20978
Edit: The hot-seat action link:
https://18xx.games/hotseat/hs_quzvabnr_20978?action=458 

The description says that the private is purchased from the market.

PR #2936, opened 2020-12-31
18CO Force Pass Player out of Cash

ryandriskel:
Closes https://github.com/tobymao/18xx/issues/2864

If the player has all their cash committed, they cannot participate in the auction any more, so skip them.

This will probably break any active 18CO game where a player committed all their cash. Not sure how likely that is.

Screen Shot 2020-12-31 at 12 09 22 PM

PR #3007, opened 2021-01-03
Fix 18CO Closed companies being available for purchase

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/2996

This will break games because passes turn into skips. Suggest merge at the same time as https://github.com/tobymao/18xx/pull/2936


ryandriskel:
It appears this may have occurred within my fixture. I'll need to get a new one, sigh.

jenf:
should be easy to change migrate_game to cope with this.

ryandriskel:
Moving to Draft while I learn migrate_game

ryandriskel:
> should be easy to change migrate_game to cope with this.

The updated fixture was generated by running the migration against the fixture. :)

tobymao (approved):

PR #3015, opened 2021-01-03
Don't tell the player the company value in 1817 variants will move beyond end of market

### LINKED ISSUE #3013: Don't say stock will move right/left if it's at the end/start of the market
A company will never exceed $600 in value; do not say a company at $600 can move right
https://18xx.games/game/18114?action=1420

ventusignis:
fixes #3013

Before this if a company is at the highest values in the market with accordingly high dividends the game will falsely tell the player that the stock price will move beyond the end of the market

### DIFF
--- a/lib/engine/step/g_1817/dividend.rb
+++ b/lib/engine/step/g_1817/dividend.rb
@@ -8,6 +8,8 @@ module Engine
     module G1817
       class Dividend < Dividend
         DIVIDEND_TYPES = %i[payout half withhold].freeze
+        PENULTIMATE_PRICE = 540
+        FINAL_PRICE = 600
         include HalfPay
 
         def half_pay_withhold_amount(entity, revenue)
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_1817/dividend.rb
```diff
@@ -8,6 +8,8 @@ module Step
     module G1817
       class Dividend < Dividend
         DIVIDEND_TYPES = %i[payout half withhold].freeze
+        PENTULTIMATE_PRICE = 540
```
jenf, line 11: penultimate

ventusignis: Thanks!

jenf:
A similar sort of issue will exist in other games. It's probably best done in a generic way up in the UI rather than down in game specific code.I think it finds the final share price afterwards already

ventusignis:
I'll try to reproduce this in 1889

ventusignis:
This happens in 1889 too. The reverse happens when at the bottom corner of the market too; it'll say withholding makes the "move left" but it doesn't move. It also says "move right" or "move left" if the share price would move vertically in the stock market instead of horizontally

I'm going to close this PR for now as I think this is a bigger question than I originally thought

PR #3022, opened 2021-01-03
Last cert exchange

tysen:
When the presidency changes and the incoming president has both the
double cert and two single certs, the outgoing president has the option
of exchanging the presidency cert for either.

### DIFF
--- a/lib/engine/g_1849/share_pool.rb
+++ b/lib/engine/g_1849/share_pool.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require_relative 'share_pool'
+
+module Engine
+  module G1849
+    class SharePool < SharePool
+      def change_president(presidents_share, swap_to, president)
+        corporation = presidents_share.corporation
+
+        incoming_pres_shares = president.shares_of(corporation)
+
+        double = incoming_pres_shares.find(&:last_cert)
+        shares =
+          if double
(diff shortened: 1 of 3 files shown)


lib/engine/g_1849/share_pool.rb
```diff
@@ -0,0 +1,42 @@
...
+  module G1849
+    class SharePool < SharePool
+      def change_president(presidents_share, swap_to, president)
+        corporation = presidents_share.corporation
+
+        incoming_pres_shares = president.shares_of(corporation)
+
+        shares = []
+        double = incoming_pres_shares.find(&:last_cert)
+        if double
```
tobymao, line 15: shares = 
  if double
...
else 
...

lib/engine/step/g_1849/swap_choice.rb
```diff
@@ -0,0 +1,68 @@
...
+        def choice_name
+          'Swap for Presidency'
+        end
+
+        def process_choose(action)
+          choice = action.choice
+          entity = action.entity
+
+          if choice == 'Two 10% certs'
+            @game.log << "#{entity.name} chooses two 10% certificates"
```
tobymao, line 56: you haev access to log here you can just do @log <<

tobymao (approved):

PR #3050, opened 2021-01-04
18Mag: starting draft

roseundy:
Add starting draft for 18Mag.
Start corps parred with all stock in pool (all in receivership).

Common file changes:
- action::bid: add ability to bid on minors
- round::draft: add "rotating" draft type
- UI::round::auction: add ability to bid on minors and on parred corp stocks

### DIFF
--- a/lib/engine/action/bid.rb
+++ b/lib/engine/action/bid.rb
@@ -5,12 +5,13 @@ require_relative 'base'
 module Engine
   module Action
     class Bid < Base
-      attr_reader :company, :corporation, :price
+      attr_reader :company, :corporation, :minor, :price
 
-      def initialize(entity, price:, company: nil, corporation: nil)
+      def initialize(entity, price:, company: nil, corporation: nil, minor: nil)
         @entity = entity
         @company = company
         @corporation = corporation
+        @minor = minor
         @price = price
       end
 
@@ -18,6 +19,7 @@ module Engine
         {
           company: game.company_by_id(h['company']),
           corporation: game.corporation_by_id(h['corporation']),
+          minor: game.minor_by_id(h['minor']),
           price: h['price'],
         }
       end
@@ -26,6 +28,7 @@ module Engine
         {
           'company' => @company&.id,
           'corporation' => @corporation&.id,
+          'minor' => @minor&.id,
           'price' => @price,
         }
       end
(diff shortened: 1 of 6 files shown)


lib/engine/action/bid.rb
```diff
@@ -5,19 +5,21 @@
...
         @corporation = corporation
+        @minor = minor
         @price = price
       end
 
       def self.h_to_args(h, game)
         {
           company: game.company_by_id(h['company']),
           corporation: game.corporation_by_id(h['corporation']),
+          minor: game.minor_by_id(h['minor']),
```
tobymao, line 22: in 1846 we use the company card as a minor, does that not work here as well?

lib/engine/game/g_18_mag.rb
```diff
@@ -40,6 +42,38 @@ def setup
           hex.tile.cities[0].place_token(@mine, @mine.next_token)
         end
         @mine.tokens.each { |t| t.type = :neutral }
+
+        # IPO and float all corporations with semi-randomly chosen prices
+        # They will start off in receivership with all shares in market
+        prices = START_PRICES.dup
+        rand_prices = @corporations.size.times.map do |_|
+          prices.rotate!(rand % prices.size)
+          prices.pop
```
tobymao, line 51: what is this trying to do? could you just do START_PRICES.sort_by { rand }

lib/engine/game/g_18_mag.rb
```diff
@@ -40,6 +42,38 @@ def setup
...
+        end
+        @corporations.each do |corp|
+          share_price = @stock_market.par_prices.find { |p| p.price == rand_prices[0] }
+          rand_prices.shift
+          @stock_market.set_par(corp, share_price)
+          corp.ipoed = true
+
+          corp.ipo_shares.each do |share|
+            bundle = ShareBundle.new([share])
```
tobymao, line 60: just call share.to_bundle

lib/engine/action/bid.rb
roseundy: 1846 doesn't process bids for minors directly. Instead it uses a private with the same name as the minor. I'd rather not do that for 18Mag since it has 13 minors and no privates. I can do that if you feel strongly about not adding to action::bid

lib/engine/game/g_18_mag.rb
roseundy: Yeah, I figured there was an elegant way of during this, but I just brute-forced it (and not very efficiently either).

lib/engine/game/g_18_mag.rb
roseundy: Ok

tobymao (approved):

PR #3068, opened 2021-01-05
Messina earthquake

tysen:
Also a couple bug fixes for track laying and corp removal.

### DIFF
--- a/lib/engine/game/g_1849.rb
+++ b/lib/engine/game/g_1849.rb
@@ -130,6 +130,10 @@ module Engine
       def remove_corp
         removed = @corporations.pop
         @log << "Removed #{removed.name}"
+        return if removed.name == 'AFG'
+
+        hex_by_id(removed.coordinates).tile.city_towns.first.remove_reservation!(removed)
+        @log << "Removed token reservation at #{removed.coordinates}"
       end
 
       def num_trains(train)
@@ -202,6 +206,7 @@ module Engine
 
       def new_stock_round
         @corporations.each { |c| c.closed_recently = false }
+        @messina_upgradeable = true
         super
       end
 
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1849.rb
```diff
@@ -396,13 +409,29 @@ def event_brown_par!
...
+        if (garibaldi = @corporations.find { |c| c.name == 'AFG' })
+          if city.tokened_by?(garibaldi) && garibaldi.placed_tokens.one?
+            @log << '-- AFG loses only token, closing. --'
+            close_corporation(garibaldi)
+          end
+        end
+
+        # Remove from game tokens on Messina
+        @log << '-- Removing tokens from game. --'
+        city.tokens.map(&:destroy!) unless city.tokens.empty?
```
perwestling, line 426: empty? not needed as empty list means no destroy anyway

perwestling, line 426: I suppose each would have been clearer but both work.

tysen: `empty?` returns `true` if the array has elements but all are `false` or `nil`, which is the effect I'm looking for here, since the `tokens` array is initialized with `slots`-many `nil`s. Anyway, I changed it to look better.

tobymao: that's not correct, empty only returns true if the element is actually empty, you're looking for tokens.any?

tobymao: city.tokens.each { |token| token&.destroy! }

tysen: You're right, I had them mixed up. The updated version already includes your suggestion.

tobymao (approved):

PR #3071, opened 2021-01-05
Corporation ordering

tysen:
Sort unopened corps on the entites tab in the order in which they'll be
opened. Also handle when AFG can't open.

When there are no free token slots in any of the hexes in which AFG can start, it is removed from the ordering and cannot be opened until a spot becomes free, which can occur either when a different corporation closes or when a relevant city is upgraded.

### DIFF
--- a/assets/app/view/game/entities.rb
+++ b/assets/app/view/game/entities.rb
@@ -26,9 +26,9 @@ module View
 
         player_owned, bank_owned = (@game.corporations + @game.minors)
           .reject(&:closed?)
-          .sort_by(&:name)
           .partition(&:owner)
-        player_owned = player_owned.group_by(&:owner)
+        player_owned = player_owned.sort_by(&:name).group_by(&:owner)
+        bank_owned = @game.bank_sort(bank_owned)
 
         children = players.map do |p|
           corps = player_owned[p]&.map { |c| h(Corporation, corporation: c) }
--- a/lib/engine/game/g_1849.rb
+++ b/lib/engine/game/g_1849.rb
@@ -121,6 +121,7 @@ module Engine
         @corporations.sort_by! { rand }
         remove_corp if @players.size == 3
         @corporations.each do |c|
+          c.slot_open = true
           c.next_to_par = false
           c.shares.last.last_cert = true
         end
@@ -156,7 +157,7 @@ module Engine
         corporation.next_to_par = false
         index = @corporations.index(corporation)
 
-        @corporations[index + 1].next_to_par = true unless index == @corporations.length - 1
+        @corporations[index + 1].next_to_par = true unless @corporations.last == corporation
       end
 
       def home_token_locations(corporation)
@@ -195,13 +196,28 @@ module Engine
         Engine::G1849::SharePool.new(self)
       end
 
+      def update_garibaldi
+        afg = @corporations.find { |c| c.name == 'AFG' }
+        return unless afg && !afg.slot_open && !home_token_locations(afg).empty?
+
+        afg.slot_open = true
+        afg.closed_recently = true
+        @log << 'AFG now has a token spot available and can be opened in the next stock round.'
+      end
+
       def close_corporation(corporation, quiet: false)
         super
         corporation = reset_corporation(corporation)
         corporation.shares.last.last_cert = true
-        @corporations.push(corporation)
+        @corporations << corporation
         corporation.closed_recently = true
-        corporation.next_to_par = true if @corporations[@corporations.length - 2].floated?
+        index = @corporations.index(corporation)
+        unless @corporations[index - 1].slot_open
+          @corporations[index - 1].next_to_par = false
+          @corporations[index - 1], @corporations[index] = @corporations[index], @corporations[index - 1]
+        end
+        corporation.next_to_par = true if @corporations[index - 1].floated?
+        update_garibaldi
       end
 
       def float_str(entity)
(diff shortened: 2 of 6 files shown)


assets/app/view/game/entities.rb
```diff
@@ -26,9 +26,9 @@ def render
 
         player_owned, bank_owned = (@game.corporations + @game.minors)
           .reject(&:closed?)
-          .sort_by(&:name)
           .partition(&:owner)
-        player_owned = player_owned.group_by(&:owner)
+        player_owned = player_owned.sort_by(&:name).group_by(&:owner)
+        bank_owned.sort_by!(&@game.bank_sort)
```
tobymao, line 31: i don't like &@game.bank_sort

lib/engine/game/g_1849.rb
```diff
@@ -191,13 +192,28 @@ def init_share_pool
         Engine::G1849::SharePool.new(self)
       end
 
+      def update_garibaldi
+        afg = @corporations.find { |c| c.name == 'AFG' }
+        return unless afg && !afg.slot_open && home_token_locations(afg).any?
```
tobymao, line 197: !empty?

lib/engine/step/g_1849/token.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Step
+    module G1849
+      class Token < Token
+        def process_place_token(action)
+          entity = action.entity
+
+          place_token(entity, action.city, action.token)
+
+          index = @game.corporations.index { |c| c.name == 'AFG' }
+          afg = index ? @game.corporations[index] : nil
```
tobymao, line 15: why don't you just do @game.corporations.find ...

lib/engine/step/g_1849/token.rb
```diff
@@ -0,0 +1,32 @@
...
+          if afg && !afg.floated? && @game.home_token_locations(afg).empty?
+            if afg.next_to_par && afg != @game.corporations.last
+              @game.corporations[index + 1].next_to_par = true
+              afg.next_to_par = false
+            end
+            afg.slot_open = false
+            @game.corporations.delete(afg)
+            @game.corporations.push(afg)
```
tobymao, line 23: corporatios << afg

lib/engine/step/g_1849/token.rb
tysen: I need the index on line 18.

tobymao: then afg = @game.corporations[index] is the simplification

assets/app/view/game/entities.rb
tysen: Have you a better idea?

tobymao: no haha

tobymao: not having this custom sort is one...

lib/engine/step/g_1849/token.rb
tysen: isn't `array[nil]` a `TypeError`?

assets/app/view/game/entities.rb
tysen: The order in which the corporations will be opened is otherwise inaccessible during the private auction. I could log the order but that's pretty lame. Also, supporting these custom sorts can also be useful in games with e.g. minors (say, 1846, in which at the start of the game the minors are mixed in with the majors alphabetically. I might argue it'd be better if they were grouped together instead.)

lib/engine/step/g_1849/token.rb
tysen: To clarify, AFG may not be in the game.

assets/app/view/game/entities.rb
daniel-sousa-me: Why not change the definition of '<=>' for unfloated corps?

tobymao: if that solution is generic that'd be great

lib/engine/step/g_1849/token.rb
tobymao: correct, i'm mistaken

assets/app/view/game/entities.rb
tysen: I wish that could work but I don't think it will. Corps have no knowledge of where they are in the opening order (that information exists only in the ordering of the `@corporations` array in `Game`). For that to work I think I'd need to add `@game` as a field in `Corporation`, and `<=>` would have to walk through the `@game.corporations` array to find its index and `other`'s index. Is that better than this?

tysen: Another way would be to store the corp's index in `Game`'s `@corporations` array as a separate field, and update it every time the order changes, but that's a lot more code in many more places than what I have here.

tobymao: ok can we instead change it to be

game.nice_method_name_to_sort_for_bank(corporations)

tysen: done

daniel-sousa-me: But given a corporation A and a corporation B, they can't figure out by themselves which one starts before?

tysen: Not without doing one of the things I listed above.

daniel-sousa-me: I don't know 1849. I assumed it was easier.

tobymao (approved):

PR #3079, opened 2021-01-05
Require permission checks to buy trains

### LINKED ISSUE #1619: Call out in UI when actions require another player's permission
By design, nothing prevents players from taking actions that require another player's permission (such as buying one of their privates, or one of their corporation's trains). And actions accidentally (or intentionally) taken can always be undone.

But it would be nice to call out those actions before they are taken, so that players are aware beforehand of which actions require permission, especially if they're new.

The lightest-weight solution (at least from the user's side) is probably to have a confirmation button somewhere for these actions. I'm not too concerned that it will add much of a speedbump to the UI, since there are very few actions that require other players' permission, and for the ones that do, you presumably already spent some time obtaining their permission, so what's one more button-press?

jenf:
This covers trains, privates and mergers

fixes #1619

Flash_opts had a bug which meant that errors wouldn't flash when buying privates, so i've had to effectively clone the flash code
Screenshot from 2021-01-08 22-35-20

### DIFF
--- a/assets/app/app.rb
+++ b/assets/app/app.rb
@@ -9,6 +9,7 @@ require 'view/about'
 require 'view/create_game'
 require 'view/invite_game'
 require 'view/home'
+require 'view/confirm'
 require 'view/flash'
 require 'view/game_page'
 require 'view/map_page'
@@ -39,6 +40,7 @@ class App < Snabberb::Component
     h(:div, props, [
       h(View::Navigation),
       h(View::Flash),
+      h(View::Confirm),
       render_content,
     ])
   end
--- a/assets/app/view/confirm.rb
+++ b/assets/app/view/confirm.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module View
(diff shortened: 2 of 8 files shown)


tobymao:
Yes, but make it a bit more dry, you defined the process_action twice, and perhaps the verbiage could be a little more concise?

'This action requires consent from x' -> Confirm ??

tobymao:
Maybe you can make this Confirm View that takes in a message and a lambda

jenf:
Done, one problem is if your on a mobile and buying the company you won't be focused in the right spot.

One possibility is making it a fixed position element to make it be a true dialog box

jenf:
@yzemaze any opinions on how this should be styled?

I've made it an overlay dialog as it might be that the top bar isn't visible.

I think it'd need to be CSS based

jenf:
Actually i'll just reuse flash with the addition of a button option.

tobymao (approved):

Issue #3073, opened 2021-01-05
[18CO] - Illegal action Blocking step Buy Shares cannot process action 873


wheresvic:
The 6 train and the E train were bought in the same set of ORs

In the subsequent SR we had the President's choice round followed by the Stock round. In the SR we could not buy any shares due to the above error.

Game json: https://gist.github.com/wheresvic/2b427b182e6e6cfe98bd1f27bfbc3679

wheresvic:
Could be related to #3066

ryandriskel:
From https://18xx.games/game/21298

ryandriskel:
I don't have a solution on this yet but found the problem. The browser thought it was Viz's turn while the server thinks its wheresvic's turn.

wheresvic:
Looks like @ryandriskel is right - we got it again https://18xx.games/game/21415?action=854 and here you can see that on the home page it shows Aavar's turn but when you jump into the game it is showing DrAwesome to make a move.

PR #3089, opened 2021-01-06
18CO - Fix detection of the end of the President's Choice round on Ruby

ryandriskel:
Fix: https://github.com/tobymao/18xx/issues/3073
Fix: https://github.com/tobymao/18xx/issues/3094

Can this really be this simple? I feel like I saw :pass used elsewhere, but running the original sample game as a spec ( which has gone poof on 18xx.games ), the President's choice round never ended as the `@entities.delete(action.entity)` was never triggered because nothing matched :pass. Switching to a string worked in both ruby and browser.

### DIFF
--- a/lib/engine/round/g_18_co/presidents_choice.rb
+++ b/lib/engine/round/g_18_co/presidents_choice.rb
@@ -23,9 +23,9 @@ module Engine
         end
 
         def after_process(action)
-          return if action.type == :message
+          return if action.free?
 
-          if action.type == :pass
+          if action.pass?
             @entities.delete(action.entity)
             return finish_round if finished?
 
@@ -58,7 +58,7 @@ module Engine
         end
 
         def finished?
-          @game.finished || @entities.empty?
+          @game.finished || @game.presidents_choice == :done || @entities.empty?
         end
 
         private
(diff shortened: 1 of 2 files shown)


lib/engine/round/g_18_co/presidents_choice.rb
```diff
@@ -23,9 +23,9 @@ def select_entities
         end
 
         def after_process(action)
-          return if action.type == :message
+          return if action.type == 'message'
 
-          if action.type == :pass
+          if action.type == 'pass'
```
tobymao, line 28: action.pass?

lib/engine/round/g_18_co/presidents_choice.rb
```diff
@@ -23,9 +23,9 @@ def select_entities
         end
 
         def after_process(action)
-          return if action.type == :message
+          return if action.type == 'message'
```
tobymao, line 26: action.free?

tobymao:
correct, opal doesn't have :symbol, ruby does. type is always a string, so when you do type == :x it is always false. but i think there are other ways you can handle your checks

ryandriskel:
I've incorporated Toby's suggestions and have the correct spec included this time. Somehow I had reverted to one from before the migrations and that's why the build was failing.

tobymao (approved):

PR #3096, opened 2021-01-06
update 18sj icons

scottredracecar:


perwestling:
Here is how the icons look like in my file handler:

image

perwestling:
And here is how the looked like before the update:

image

perwestling:
And here is the looks for a completed game.

image

Nisse75:
The new tokens looks great! For printed prototypes I will be using the colours (and cmyk #) below. So if possible and not to much work, the online game should be the same for both tokens and charters.

BJ Brown 7C3A2D
TGOJ Orange FC4C02
UGJ Bright green 4CAE04
SNJ Blue 0067B9
KFJ White FFFFFF (if white is a problem use a light shade of grey?)
ÖSJ Red D50032
MYJ Yellow FFD100
STJ Black 000000
MÖJ Gold C97A00
SWB Green 00843D
ÖKJ Violet 7248BD
KHJ Lavender A08BCB

perwestling:
This is how it looks when KFJ has placed its token, and no other corporation has. It is pretty hard to tell the difference in online play. For the physical game white tokens do work.  So my suggestion is to use some color to separate KFJ from white. 
Using light gray is possible - wont be a problem with gray tiles as KFJ has been long time merged when gray tiles appear....

image

perwestling:
Unless there is a "real" logo to use for KHJ?

Nisse75:
I have researched logos a bit for swedish rail and found almost nothing. My
guess is that logos was not ”a thing” here at the time.

I really don’t think the colours are super important so not using white for
the online implementation is ok with me.

Örjan

ons 6 jan. 2021 kl. 17:41 skrev Per Westling <notifications@github.com>:

Nisse75:
ons 6 jan. 2021 kl. 18:13 skrev Örjan Wennman <orjanwennman@gmail.com>:

PR #3101, opened 2021-01-06
1856 rights nationalization

ventusignis:
The tunnel & bridge rights get transferred to the national during nationalization; duplicates get popped to the bank pool where other corporations may purchase them

### DIFF
--- a/lib/engine/step/g_18_sj/special_buy.rb
+++ b/lib/engine/step/g_18_sj/special_buy.rb
@@ -27,7 +27,7 @@ module Engine
 
         def setup
           super
-          @gkb_bonus_item = Item.new(description: 'GKB bonus', cost: 0)
+          @gkb_bonus_item ||= Item.new(description: 'GKB bonus', cost: 0)
           @gkb_bonus_bought = false
         end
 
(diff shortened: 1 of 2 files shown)


lib/engine/step/g_18_sj/special_buy.rb
```diff
@@ -27,7 +27,7 @@ def process_special_buy(action)
 
         def setup
           super
-          @gkb_bonus_item = Item.new(description: 'GKB bonus', cost: 0)
+          @gkb_bonus_item ||= Item.new(description: 'GKB bonus', cost: 0)
```
tobymao, line 30: why is this changed?

ventusignis: This was supposed to be in the prior PR but the push from my working copy failed due to a typo

tobymao: is this needed?

ventusignis: It will work as-is. This change will prevent a new instance of the Item being created every operating round

tobymao (approved):

PR #3109, opened 2021-01-06
18SJ: Improve GKB bonus

### LINKED ISSUE #3102: 18SJ - Göta kanalbolag private ability not working
https://18xx.games/game/21455
When Raymong bought the private into MOJ company, Ability didnt show up until the next OR started.
then clicking the Buy GKR Bonus ability button gave error and reset the game back position as if button wasn't pressed.
thre was a different error but it went by too fast to catch

image

perwestling:
Show multiple bonuses in revenue string.

Improve description.

Correct acquire bug.

Removed the "special buy" button. Instead you activate the ability by clicking on Abilities: Göta Kanalbolaget.
This will clear out the route shown in the view, and when route is specified it will apply the bonuses.
If ability is activated, and a run is done, this will decrease the count for the ability (regardless if GKB bonus are used or not).
If ability is deselected the latest run is restored (if any) and the ability count is unaffected.

This should fix #3102


lib/engine/game/g_18_sj.rb
```diff
@@ -680,11 +680,11 @@ def gkb_bonus(stops)
...
 
-        revenue = 0
-        stops.select { |s| GKB_HEXES.include?(s.hex.name) }.each { |_s| revenue += @gkb_bonus }
-        if revenue.positive?
-          bonus[:revenue] = revenue
+        bonuses = stops.count { |s| GKB_HEXES.include?(s.hex.name) }
+        if bonuses.positive?
+          bonus[:revenue] = @gkb_bonus * bonuses
           bonus[:description] = 'GKB'
+          bonus[:description] += "x#{bonuses}" if bonuses > 1
```
tobymao, line 687: in here do route.clear_cache or something

tobymao: actually no, how exactly does the ability work and where is it applied?

perwestling: @tobymao As it is now the ability makes a button appear when you have the run route view up, and there are attempts left. If you click on the button a choose action is triggered which change the route revenue and string calculation change in the 18SJ game file. (A state is set in 18_sj/game.rb which is used when calculating route revenue.)
This does work if you in any way updates the route(s) in the GUI, but not for an existing route.

perwestling: @tobymao Have now changed so that ability is used instead. Have tried this with the game in question and it works fine, decrease the amount for each try, and remove the ability when used up.

lib/engine/action/run_routes.rb
```diff
@@ -24,6 +24,7 @@ def self.h_to_args(h, game)
             revenue_str: route['revenue_str'],
             subsidy: route['subsidy'],
             halts: route['halts'],
+            ability: route['ability'],
```
tobymao, line 27: should this be multiple abilities?

lib/engine/game/g_18_sj.rb
```diff
@@ -299,7 +296,7 @@ def revenue_for(route, stops)
          bergslagen_bonus(icons),
          orefields_bonus(icons),
          sveabolaget_bonus(route, stops),
-         gkb_bonus(stops)].map { |b| b[:revenue] }.each { |r| revenue += r }
+         gkb_bonus(route, stops)].map { |b| b[:revenue] }.each { |r| revenue += r }
```
tobymao, line 299: route already has stops, so you don't need to pass it in separately right?

lib/engine/action/run_routes.rb
perwestling: All routes will have same value for ability, but that was a simplification that I think might be ok.

I think it might be possible to have multiple route related abilities active, so maybe the code should support that? But right now there is only one so I did not implement support for that.

lib/engine/game/g_18_sj.rb
perwestling: Right. So I should fix that, and the sveabolaget_bonus which do the same thing.

lib/engine/action/run_routes.rb
perwestling: Ok, I made it into a list instead, making it possible to support multiple abilities in the future.

lib/engine/step/route.rb
```diff
@@ -41,6 +41,10 @@ def process_run_routes(action)
           @log << "#{entity.name} runs a #{train.name} train for #{revenue}: #{route.revenue_str}"
         end
         pass!
+
+        return unless (route = @round.routes.first)
```
tobymao, line 45: this doesn't look correct? why do you only use one route?

tobymao: why not just put it in the each loop ?

lib/engine/route.rb
```diff
@@ -19,6 +19,7 @@ def initialize(game, phase, train, **opts)
       @revenue_str = opts[:revenue_str]
       @subsidy = opts[:subsidy]
       @halts = opts[:halts]
+      @abilities = opts[:abilities] || []
```
tobymao, line 22: by setting [] here you will always store an empty array in the database because array is truthy maybe this should be nil or in the Action::Route code you should remove empty abilities

perwestling: Have refactored it a bit now.

lib/engine/step/route.rb
perwestling: Changed this code a bit.

tobymao (approved):

PR #3111, opened 2021-01-06
Share value ordering

tysen:
Corporations that end up on the same stock market space after a series
of sales by the same player should retain the relative operating order
they had before the sales.

### DIFF
--- a/lib/engine/share_price.rb
+++ b/lib/engine/share_price.rb
@@ -59,6 +59,10 @@ module Engine
       @limited = !unlimited_types.include?(type)
     end
 
+    def ==(other)
+      @coordinates == other.coordinates
+    end
+
     def id
       "#{@price},#{@coordinates.join(',')}"
     end
--- a/lib/engine/step/g_1849/buy_sell_par_shares.rb
+++ b/lib/engine/step/g_1849/buy_sell_par_shares.rb
@@ -6,6 +6,23 @@ module Engine
   module Step
     module G1849
       class BuySellParShares < BuySellParShares
+        def setup
+          super
+          @game.old_operating_order = @game.corporations.sort
+        end
+
+        def pass!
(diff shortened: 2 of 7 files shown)


lib/engine/share_price.rb
```diff
@@ -2,7 +2,8 @@
 
 module Engine
   class SharePrice
-    attr_reader :coordinates, :price, :corporations, :can_par, :type, :types
+    attr_reader :coordinates, :price, :can_par, :type, :types
```
tobymao, line 5: i don't this change is necessary, you can just mutate corporations

corporations.clear
corporations.concat(reordered)

lib/engine/step/g_1849/buy_sell_par_shares.rb
```diff
@@ -6,6 +6,23 @@ module Engine
...
     module G1849
       class BuySellParShares < BuySellParShares
+        def setup
+          super
+          @game.old_operating_order = @game.corporations.sort
+        end
+
+        def pass!
+          @passed = true
+          if @current_actions.any?
```
tobymao, line 16: unless empty

lib/engine/share_price.rb
```diff
@@ -59,6 +60,10 @@ def initialize(coordinates,
       @limited = !unlimited_types.include?(type)
     end
 
+    def ==(other)
```
tobymao, line 63: why did you add this?

lib/engine/share_price.rb
tysen: good point

lib/engine/share_price.rb
tysen: For the `group_by` in `reorder_corps`

tobymao (approved):

PR #3115, opened 2021-01-07
18Mag: double-sided tiles, K-tiles

roseundy:
- Double-sided tiles, which is basically pairing up certain tiles. When one is used the other is put on a separate list.
- Support for laying plain yellow track on K-hexes.

Common file changes:
UI::tile_manifest, UI::tiles - support for double-sided manifest
Game::base - DOUBLE_SIDED_TILE flag
Step::tracker - broke tile list updates out to allow overriding
Engine::tile - added "opposite" method and instance var

18Mag_Manifest

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -74,7 +74,7 @@ module Engine
     class Base
       attr_reader :actions, :bank, :cert_limit, :cities, :companies, :corporations,
                   :depot, :finished, :graph, :hexes, :id, :loading, :loans, :log, :minors,
-                  :phase, :players, :operating_rounds, :round, :share_pool, :stock_market,
+                  :phase, :players, :operating_rounds, :round, :share_pool, :stock_market, :tile_groups,
                   :tiles, :turn, :total_loans, :undo_possible, :redo_possible, :round_history, :all_tiles,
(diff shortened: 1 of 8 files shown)


lib/engine/game/base.rb
```diff
@@ -226,6 +226,8 @@ class Base
 
       MUST_BUY_TRAIN = :route # When must the company buy a train if it doesn't have one (route, never, always)
 
+      DOUBLE_SIDED_TILES = false # if true, requires @tile_groups to be set
```
tobymao, line 229: just set tile groups in all and check the emptyness of that instead of this constant

lib/engine/tile.rb
```diff
@@ -243,6 +244,10 @@ def terrain
       @upgrades.flat_map(&:terrains).uniq
     end
 
+    def opposite!(tile)
```
tobymao, line 247: uhhh why? just make opposite assecor and do tile.opposite = x

lib/engine/game/base.rb
roseundy: Ok

lib/engine/tile.rb
roseundy: Sure.

tobymao (approved):

PR #3122, opened 2021-01-07
1824: Emergency buy

perwestling:
Some special things for 1824 emergency buy
- A player cannot go bankrupt - if needed a player can take a loan
  with 50% direct interest, and 50% of current debt after each SR
- A player can take the loan during emergency buy even though
  still owning sellable shares
- If emergency buy, if buying train from other player's corporation
  the price cannot exceed face value of train bought
- Presidency change not allowed during emergency buy

This code has not been tested in action. That will have to
wait until more functionality has been implemented - so therefor
it is a draft commit.

### DIFF
--- a/assets/app/view/game/loans.rb
+++ b/assets/app/view/game/loans.rb
@@ -17,7 +17,7 @@ module View
           children << h(
             :button,
             { on: { click: lambda do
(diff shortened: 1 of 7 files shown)


perwestling:
Thinking of another solution. Using a choice button to buy a train with a loan.

PR #3148, opened 2021-01-08
[18CZ] Added placing purple tiles ability of privates

Zwergenpunk:
I changed the block_lay property to writable because the purple tiles have do not need an icon.

There is also the posibility to remove the block_lay option from icon and place it as an own property.


lib/engine/config/game/g_18_cz.rb
```diff
@@ -303,126 +303,576 @@ module G18CZ
...
+                               "216p",
+                               "611p",
+                               "889p",
+                               "8894p",
+                               "8895p",
+                               "8896p",
+                               "595p",
+                               "8857p"
+                            ],
+                            "when": "track",
```
benjaminxscott, line 839: I think `when` s.b . "owning_corp_or_turn" since it can be laid at any point in the corp's turn

lib/engine/config/game/g_18_cz.rb
```diff
@@ -303,126 +303,576 @@ module G18CZ
...
                     "value": 75,
                     "revenue": 20,
                     "sym": "NM",
-                    "abilities": []
+                    "abilities": [
+                        {
+                            "type":"tile_lay",
+                            "count": 1,
+                            "owner_type":"corporation",
+                            "free": false,
```
benjaminxscott, line 823: `free` is false by default so can omit

lib/engine/config/game/g_18_cz.rb
Zwergenpunk: yeah, you're right. Fixed it

tobymao (approved):

PR #3150, opened 2021-01-09
Private companies for 1849

tysen:
Now including the right token thanks to @scottredracecar


lib/engine/step/g_1849/track.rb
```diff
@@ -14,11 +14,24 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
...
         end
+
+        def available_hex(entity, hex)
+          return super unless entity.sms_hexes
+
+          return [0, 1, 2, 3, 4, 5] if entity.sms_hexes.find { |h| h == hex.id }
+        end
+
+        def check_track_restrictions!(entity, old_tile, new_tile)
+          super
```
tobymao, line 31: why not just do super unless entity.sms_hexes

Issue #1570, opened 2020-09-01
Allow inspection of a title without playing it


dfannius:
When I am learning the rules of a new game that is implemented on 18xx.games, I find a game in progress, rewind to the beginning, and look at the privates, market, phase info, etc. It would be nice if there were a way to view all of that information in a read-only manner without finding an actual live (or dead) game to snoop on.

dfannius:
Thanks, that's more convenient than rewinding the history. It would still be nice not to have to first find an active game of the title you're interested in.

ChrisTheCat:
You can also just create a new hotseat (local) game yourself and play with
it.

jenf:
Maybe "preview/1889" should give you that option

PR #3162, opened 2021-01-09
[HOLD] MVP Title Page - Showing info for a specific Game title

ryandriskel:
Title page, eg https://18xx.games/title/1889

Place where a user can see basic information ( stuff from Game Info, link to rules ) about a particular title ( 1846, 1889, etc), games open / in progress / complete / etc for that title, stats on that title ( number of completed games, popular player counts ? ), link to create a game for ( with the title already specified ), link to map / market, same chat as homepage.

This PR does not incorporate all of those features, but gives a starting point to build on top. A bonus feature is that game developers can easily find alpha/beta games in progress/completed for reference.

**I could use someone with a large local db of games to verify everything works as expected with a larger set of games.**

Some examples of title pages ( without many games listed because I don't have much locally beyond hotseat )

Screen Shot 2021-01-09 at 10 57 34 AM
Screen Shot 2021-01-09 at 11 09 32 AM

Screen Shot 2021-01-09 at 11 14 06 AM

I also added links to the title pages for each game.

Screen Shot 2021-01-09 at 11 17 11 AM

I verified the in-game info is still working as expected.

Screen Shot 2021-01-09 at 11 11 21 AM

### DIFF
--- a/api.rb
+++ b/api.rb
@@ -100,6 +100,11 @@ class Api < Roda
       render_with_games
     end
 
+    r.on 'title', String do |title|
+      request.params['title'] = title
+      render(pin: request.params['pin'], games: Game.home_games(user, **request.params).map(&:to_h))
+    end
+
     r.on 'game', Integer do |id|
       halt(404, 'Game not found') unless (game = Game[id])
 
--- a/assets/app/view/game/game_meta.rb
+++ b/assets/app/view/game/game_meta.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module View
+  class GameMeta < Snabberb::Component
+    needs :game
+
+    def render
+      children = [h(:h3, 'Game Info')]
+
+      if (publisher = @game.class::GAME_PUBLISHER)
+        children << h(:p, [
+            'Published by ',
+            *Lib::Publisher.link_list(component: self, publishers: Array(publisher)),
+          ])
+      end
+      children << h(:p, "Designed by #{@game.class::GAME_DESIGNER}") if @game.class::GAME_DESIGNER
+      children << h(:p, "Implemented by #{@game.class::GAME_IMPLEMENTER}") if @game.class::GAME_IMPLEMENTER
+      if @game.class::GAME_RULES_URL.is_a?(Hash)
+        @game.class::GAME_RULES_URL.each do |desc, url|
+          children << h(:p, [h(:a, { attrs: { href: url, target: '_blank' } }, desc)])
+        end
+      else
+        children << h(:p, [h(:a, { attrs: { href: @game.class::GAME_RULES_URL, target: '_blank' } }, 'Rules')])
+      end
+      if @game.optional_rules.any?
+        children << h(:h3, 'Optional Rules Used')
+        @game.class::OPTIONAL_RULES.each do |o_r|
+          next unless @game.optional_rules.include?(o_r[:sym])
+
+          children << h(:p, " * #{o_r[:short_name]}: #{o_r[:desc]}")
+        end
+      end
+
+      if @game.class::GAME_INFO_URL
+        children << h(:p, [h(:a, { attrs: { href: @game.class::GAME_INFO_URL, target: '_blank' } }, 'More info')])
+      end
+
+      h(:div, children)
+    end
+  end
+end
--- a/assets/app/view/title_page.rb
+++ b/assets/app/view/title_page.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+require 'game_manager'
+require 'lib/settings'
+require 'lib/storage'
+require 'view/chat'
+require 'view/game_row'
+require 'view/game/game_meta'
+
+module View
+  class TitlePage < Snabberb::Component
+    include GameManager
+    include Lib::Settings
+
+    needs :user
+    needs :route
+    needs :refreshing, default: nil, store: true
+
+    ROUTE_FORMAT = %r{/title/([^/?]*)/?}.freeze
+
+    def render
+      @game_title = @route.match(ROUTE_FORMAT)[1].gsub('%20', ' ')
+
+      your_games, other_games = @games
+        .partition { |game| user_in_game?(@user, game) || user_owns_game?(@user, game) }
+
+      children = [
+        render_game_info,
+        h(Chat, user: @user, connection: @connection),
+      ]
+
+      # these will show up in the profile page
+      your_games.reject! { |game| game['status'] == 'finished' }
+
+      grouped = other_games.group_by { |game| game['status'] }
+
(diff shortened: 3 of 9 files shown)


ryandriskel:
Have a db backup and will be testing this today.

assets/app/view/game/game_meta.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module View
+  class GameMeta < Snabberb::Component
```
michaeljb, line 4: Since this is shown on the site as "Info", could we call the class and file "Info" as well? Easier to find the relevant class that way 🙂

ryandriskel: This component is a shared sub-component of the "Game Info" component that already exists.

ryandriskel:
Neat side effect of the latest changes, http://0.0.0.0:9292/?title=1889 allows one to filter the homepage to that title (1889 here). It will be very simple to add a select to filter the homepage.

ryandriskel:
Screen Shot 2021-01-23 at 9 59 16 PM

yzemaze:
> Neat side effect of the latest changes, http://0.0.0.0:9292/?title=1889 allows one to filter the homepage to that title (1889 here). It will be very simple to add a select to filter the homepage.

I had added a filter for titles (and would have added more), but Toby didn’t like it that way and preferred fulltextsearch on more columns. When implementing this feature I ran into trouble due to user_refresh quirks. These are gone since the cookie based auth was implemented this month. Stay tuned :)

api.rb
```diff
@@ -100,6 +100,11 @@ class Api < Roda
       render_with_games
     end
 
+    r.on 'title', String do |title|
+      request.params['title'] = title
```
tobymao, line 104: why do you do this?

api.rb
```diff
@@ -100,6 +100,11 @@ class Api < Roda
...
+      render(pin: request.params['pin'], games: Game.home_games(user, **request.params).map(&:to_h))
```
tobymao, line 105: why do you need pin?

assets/app/view/game/game_meta.rb
tobymao: aren't you missing a module Game ?

assets/app/view/game/game_meta.rb
```diff
@@ -0,0 +1,41 @@
...
+      children << h(:p, "Designed by #{@game.class::GAME_DESIGNER}") if @game.class::GAME_DESIGNER
+      children << h(:p, "Implemented by #{@game.class::GAME_IMPLEMENTER}") if @game.class::GAME_IMPLEMENTER
+      if @game.class::GAME_RULES_URL.is_a?(Hash)
+        @game.class::GAME_RULES_URL.each do |desc, url|
+          children << h(:p, [h(:a, { attrs: { href: url, target: '_blank' } }, desc)])
+        end
+      else
+        children << h(:p, [h(:a, { attrs: { href: @game.class::GAME_RULES_URL, target: '_blank' } }, 'Rules')])
+      end
+      if @game.optional_rules.any?
```
tobymao, line 25: empty

assets/app/view/game/game_meta.rb
```diff
@@ -0,0 +1,41 @@
...
+        children << h(:p, [
+            'Published by ',
+            *Lib::Publisher.link_list(component: self, publishers: Array(publisher)),
+          ])
+      end
+      children << h(:p, "Designed by #{@game.class::GAME_DESIGNER}") if @game.class::GAME_DESIGNER
+      children << h(:p, "Implemented by #{@game.class::GAME_IMPLEMENTER}") if @game.class::GAME_IMPLEMENTER
+      if @game.class::GAME_RULES_URL.is_a?(Hash)
+        @game.class::GAME_RULES_URL.each do |desc, url|
+          children << h(:p, [h(:a, { attrs: { href: url, target: '_blank' } }, desc)])
```
tobymao, line 20: i don't think we do target blank anywhere else

assets/app/view/title_page.rb
```diff
@@ -0,0 +1,116 @@
...
+      your_games, other_games = @games
+        .partition { |game| user_in_game?(@user, game) || user_owns_game?(@user, game) }
+
+      children = [
+        render_game_info,
+        h(Chat, user: @user, connection: @connection),
+      ]
+
+      # these will show up in the profile page
+      your_games.reject! { |game| game['status'] == 'finished' }
```
tobymao, line 33: you're missing archived

assets/app/view/title_page.rb
```diff
@@ -0,0 +1,116 @@
...
+        key: 'title_page',
+        hook: {
+          destroy: destroy,
+        },
+      }
+
+      h('div#titlepage', props, children)
+    end
+
+    def game_refresh
```
tobymao, line 74: is this stuff all copy and pasted from the game page?

models/game.rb
```diff
@@ -16,6 +16,7 @@ class Game < Base
       SELECT *
       FROM games
       WHERE status = '%<status>s'
+        AND title ~ :title
```
tobymao, line 19: why are you using a regex

assets/app/view/title_page.rb
ryandriskel: No ... mostly copy and pasted from the home page.

models/game.rb
ryandriskel: This method for creating SQL queries is a bit new to me. I couldn't figure out a way to _not_ use title in the query if none was specifed. Ideally the title parameter wouldn't be there at all when none was specified, then '=' when it is.

assets/app/view/game/game_meta.rb
ryandriskel: I didn't add it. I moved this code from game_info.rb > game_info into its own module.

api.rb
ryandriskel: To force the initial set of games shown on the title page to be that title. Otherwise the initial set is all games.

assets/app/view/game/game_meta.rb
ryandriskel: This was pre-existing code, but I can change it.

dfannius:
This would implement #1570.

ChrisTheCat:
A bunch of the features/information that are requested here are already in the wiki. https://github.com/tobymao/18xx/wiki Based on comments about this PR, I'm planning to add mini-maps for each game to the wiki. I'm also talking to a user about rewording the wiki link on the 18xx.games home page to make it more obvious what it contains.

ChrisTheCat:
Links to the map, tiles, and market for each game have been added to all the game entries on the wiki

ryandriskel:
Closing this PR for now. Will reopen when I pick it back up.

PR #3202, opened 2021-01-10
publish commit hash

### LINKED ISSUE #1609: Display deployed version of the site
This has been discussed but I don't think there's an issue for it.

It would be nice to display (maybe in the About page) the commit ID of the version that has been deployed, so people who have had PRs merged don't have to wonder if those changes have gone live yet.

yzemaze:
Low level and definitely not as nice as showing deployed version in /about, but it works fine locally. You have to decide if this works with the current deploy process though.
.gitignore addition to prevent annoying mishaps.
This would fix #1609 – sort of. Triggered by https://18xxgames.slack.com/archives/C012K0CNY5C/p1610272399481100

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -113,6 +113,7 @@ task :precompile do
   pin_dir = Assets::OUTPUT_BASE + Assets::PIN_DIR
   FileUtils.mkdir_p(pin_dir)
   FileUtils.cp("#{bundle}.gz", "#{pin_dir}/#{git_rev}.js.gz")
+  sh("echo #{git_rev} > #{Assets::OUTPUT_BASE}/commit")
 end
 
 desc 'Profile loading data'
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 /vendor/
 /public/assets/*
 /public/pinned/*
+/public/commit
 
 .tags
 .DS_STORE


michaeljb:
I think [`scripts/deploy_prod.sh`](https://github.com/tobymao/18xx/blob/master/scripts/deploy_prod.sh) would need this line added, perhaps after the `ssh` command so that the commit is only updated if the rest of the deploy was successful:

```
scp public/commit deploy@18xx:~/18xx/public/
```

michaeljb:
Also embedding the SHA in the JS bundle somewhere and using this to auto-trigger a refresh if the client is out of date would be neat, obviously that's a lot more work than this 😄

daniel-sousa-me:
> Also embedding the SHA in the JS bundle somewhere and using this to auto-trigger a refresh if the client is out of date would be neat, obviously that's a lot more work than this

Probably we can use the message-bus to do this easily. Just have the clients subscribe to a "/version" channel and have Rake (or whatever makes sense, I never understood how this pipeline works) publish a message when there is a file change/restart/whatever

Rakefile
```diff
@@ -113,6 +113,7 @@ task :precompile do
   pin_dir = Assets::OUTPUT_BASE + Assets::PIN_DIR
   FileUtils.mkdir_p(pin_dir)
   FileUtils.cp("#{bundle}.gz", "#{pin_dir}/#{git_rev}.js.gz")
+  sh("echo #{git_rev} > #{Assets::OUTPUT_BASE}/commit")
```
daniel-sousa-me, line 116: Shouldn't this be:

```File.write("#{Assets::OUTPUT_BASE}/commit", git_rev)```

tobymao:
sha of the bundle is done easily by changing the name of the file in Assets so that HTML references the new javascript.

tobymao:
#725277

PR #3218, opened 2021-01-11
Auto action prototype

ventusignis:
Gist here for playing around with this
https://gist.github.com/ventusignis/80a7f6bee3e476bce3c5add1015846a7


ventusignis:
Forgot to put `unless @game.loading` when executing the auto action and it doesn't work in this case so closing to look into that

Issue #2777, opened 2020-12-23
18CO Laramie, North Park and Western Railroad usable outside of Track Lay Step


ryandriskel:
Laramie, North Park and Western Railroad allows the corporation to lay a free additional yellow tile during its track lay. Currently, this can be done at any point during its turn, as opposed to only while laying tile.

PR #3220, opened 2021-01-11
18CO Ability Timing

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/2777

Correct the timing of LNPW to be during Track only.
Correct GJGR as it cannot be used during other's ORs.
Use a modern fixture that used these abilities correctly.

Note: I was concerned about `Denver, Northwestern and Pacific Railroad` as it has no `when` but I tried using it and it worked correctly.


michaeljb (approved):

tobymao (approved):

jenf:
i'll do the 1817 migration and 18CO pins at the same time in about 2 hours

PR #3223, opened 2021-01-11
Auto action prototype

ventusignis:
This is a prototype based on the discussion on auto actions.
I tried calling them "derived actions" to distinguish them from actions that are automatically passed.

I added a throwaway Action `Bonus` that just `@Logs` a silly message to demonstrate that the derived actions work, even with undos & redos
*edit* The prototype only affects 1856 but since 1856 still needs so much work to be 'done' I don't really mind adding this log message for now
Here is a gist to a hotseat game you can play around with this.
https://gist.github.com/ventusignis/4afbbe32c196d3cea887f73df361e822

### DIFF
--- a/lib/engine/action/base.rb
+++ b/lib/engine/action/base.rb
@@ -33,7 +33,7 @@ module Engine
 
       def initialize(entity)
         @entity = entity
-        @dervied = false
+        @derived = false
         @round_override = false
         @created_at = Time.now
         @derived_children = []
--- a/lib/engine/action/base.rb
+++ b/lib/engine/action/base.rb
@@ -8,13 +8,14 @@ module Engine
       include Helper::Type
 
       attr_reader :entity
-      attr_accessor :id, :user, :created_at
+      attr_accessor :id, :user, :created_at, :derived
 
       def self.from_h(h, game)
         entity = game.get(h['entity_type'], h['entity']) || Player.new(nil, h['entity'])
         obj = new(entity, **h_to_args(h, game))
         obj.user = h['user'] if entity.player && h['user'] != entity.player&.id
         obj.created_at = h['created_at'] || Time.now
+        obj.derived = h['derived'] || false
         obj
       end
 
@@ -28,6 +29,7 @@ module Engine
 
       def initialize(entity)
         @entity = entity
+        @dervied = false
         @created_at = Time.now
       end
 
@@ -43,6 +45,7 @@ module Engine
           'id' => @id,
           'user' => @user,
           'created_at' => @created_at.to_i,
+          'derived' => @derived,
           **args_to_h,
         }.reject { |_, v| v.nil? }
       end
--- a/lib/engine/action/base.rb
+++ b/lib/engine/action/base.rb
@@ -8,7 +8,7 @@ module Engine
       include Helper::Type
 
       attr_reader :entity
-      attr_accessor :id, :user, :created_at, :derived
+      attr_accessor :id, :user, :created_at, :derived, :round_override
 
       def self.from_h(h, game)
         entity = game.get(h['entity_type'], h['entity']) || Player.new(nil, h['entity'])
@@ -16,6 +16,7 @@ module Engine
         obj.user = h['user'] if entity.player && h['user'] != entity.player&.id
         obj.created_at = h['created_at'] || Time.now
         obj.derived = h['derived'] || false
+        obj.round_override = h['round_override'] || false
         obj
       end
 
@@ -30,6 +31,7 @@ module Engine
       def initialize(entity)
         @entity = entity
         @dervied = false
+        @round_override = false
         @created_at = Time.now
       end
 
@@ -46,6 +48,7 @@ module Engine
           'user' => @user,
           'created_at' => @created_at.to_i,
           'derived' => @derived,
+          'round_override' => @round_override,
           **args_to_h,
         }.reject { |_, v| v.nil? }
       end
--- a/lib/engine/action/base.rb
+++ b/lib/engine/action/base.rb
@@ -8,7 +8,7 @@ module Engine
       include Helper::Type
 
       attr_reader :entity
-      attr_accessor :id, :user, :created_at, :derived, :round_override
+      attr_accessor :id, :user, :created_at, :derived, :round_override, :derived_children
 
       def self.from_h(h, game)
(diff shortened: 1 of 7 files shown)


lib/engine/action/base.rb
```diff
@@ -28,7 +33,10 @@ def self.split(klass)
 
       def initialize(entity)
         @entity = entity
+        @dervied = false
```
jenf, line 36: Typo

jenf, line 36: Typo

lib/engine/step/g_1856/dividend.rb
```diff
@@ -24,7 +24,7 @@ def process_dividend(action)
           revenue = @game.routes_revenue(routes) - penalty
           kind = action.kind.to_sym
           payout = dividend_options(entity)[kind]
-
+          @game.process_derived_action(Engine::Action::Bonus.new(entity)) unless @game.loading
```
jenf, line 27: So I have a different way of thinking about this.

For instance in the 1867 token case.
The step should occur as normal, and the method should return a

def automatic_action(entity)
  Engine::Action::Pass.new(entity) if rather_expensive_check_returns_that_only_option_is_pass
end


This would return either nil for no action, or an action to be added.

process_action when it's finished it's normal actions would check the current step to see if it's got a derived action and if so process it until it runs out of derived actions. These derived actions would be passed upwards to the callee so that they can be added to the json/database.

ventusignis: OK so you are suggesting is that instead of calling `@game.process_derived_action(<new action>)` inside `process_#{action}` to have `process_#{action}` return an array of actions, or to have `automatic_action(entity)` be a separate method that gets invoked by the code that called `process_#{action}`?

tobymao:
I think this is a bit too complicated. 

Can we instead just have #auto_action which will attach a child action to the regular action

so Action:;Base just needs one attr_reader -- auto_actions which is the array of actions that were created

jenf:
My problem with that approach is it assumes that the tile lay knows about the next action, unless i'm misparsing

In the 1867 case the tile lay shouldn't need to know about if the token should be passed.

lib/engine/game/base.rb
```diff
@@ -592,8 +599,32 @@ def initialize_actions(actions)
...
+      def process_derived_action(action)
+        action = action_from_h(action) if action.is_a?(Hash)
+        action.derived = true
+
+        @action_queue.prepend(action)
+      end
+
       def process_action(action)
         action = action_from_h(action) if action.is_a?(Hash)
+        raise GameError, "Derived actions may not use process_action, especially #{action}" if action.derived
```
daniel-sousa-me, line 611: This is probably (and hopefully) my most useless comment ever on this repo: instead of "especially" it should be "specially" or "specifically".

lib/engine/step/g_1856/dividend.rb
jenf: At the end of process_action (or an alternative say generate_derived_actions which can get called when trying to write actions) it'll keep trying to see if there's an automatic action available to it.

This could be limited to the UI and submitted as a bunch of actions.

tobymao:
> My problem with that approach is it assumes that the tile lay knows about the next action, unless i'm misparsing
> 
> In the 1867 case the tile lay shouldn't need to know about if the token should be passed.

round/base.rb could trigger this when it's choosing to skip things.

jenf:
Could the prototype show how 1867 token would work

ventusignis:
> 
> 
> Could the prototype show how 1867 token would work

I'm not familiar with 1867, could you explain a little more how the 1867 token works? I should be able to put a POC of that up tomorrow evening

jenf:
The token cost is based on distance as crow flies from the nearest existing token.

This is quite CPU intensive, so it was skipped as part of this PR:
https://github.com/tobymao/18xx/pull/3077/files

lib/engine/game/base.rb
```diff
@@ -592,8 +599,32 @@ def initialize_actions(actions)
...
+
+        # Process the action we came here to do first
+        result = process_single_action(action)
+
+        # Process any other actions put on the queue since our action
+        until @action_queue.empty?
+          derived_action = @action_queue.pop
```
daniel-sousa-me, line 618: If it's a queue, it should be `shift`. `pop` is for stacks (FIFO vs FILO)

ventusignis:
@tobymao @jenf 
I tried to get this to work for the 1867 can_token? scenario and it's not simple to use. 

The way this prototype is designed allows for doing checks *after* the main action has been completed - the aforementioned scenario needs the check done *before*

In this scenario, changing the Token action to have a `token_check` field or doing the check after finishing the Track action would be the way to implement this scenario for 1867 with the prototype. Changing the Token action looks to be the simpler of the two cases

lib/engine/game/base.rb
ventusignis: If this PR goes anywhere I need to fix this, and test it, before it can get merged

jenf:
> The way this prototype is designed allows for doing checks _after_ the main action has been completed - the aforementioned scenario needs the check done _before_

> 
> In this scenario, changing the Token action to have a `token_check` field or doing the check after finishing the Track action would be the way to implement this scenario for 1867 with the prototype. Changing the Token action looks to be the simpler of the two cases

Yeah this was the cause of my concern, we need a way of doing this check after as I'd expect 99% of them to be done after as making the track aware of the token actions would make it unmaintanable.

The 'add_auction_action' approach mentioned by toby would seem to work doing it after the main step has been processed

tobymao:
can you redo it with my suggested approach? can make a new PR

tobymao:
or perhaps jen may take a stab at it

ventusignis:
I don't think the other ideas proposed would solve this for 1867; I believe that in all cases the main Token action (whether it's a Pass or PlaceToken) would get executed before the followup autoaction - if I understand both proposals correctly Jen's proposal would have it executed at the end of `process_action` and Toby's proposal would behave like my POC but the action would be stored as a child of Action N instead of being stored as action N+1

This could possibly be doable by having `pre-auto-actions` which get executed before the "main" action. To implement the 1867 case the `pre-auto-action` would set round state which could be checked in the Token step instead of manually calculating the `can_token?` method

tobymao:
closing for now

PR #3252, opened 2021-01-12
18CZ Added stubs to Praha

Zwergenpunk:


jenf:
note the stubs by themselves don't restrict tile laying, see 1867 on how to do this.

Zwergenpunk:
> note the stubs by themselves don't restrict tile laying, see 1867 on how to do this.

Ah thanks. good to know ;)

PR #3275, opened 2021-01-13
18CZ Added Restrictions for stubs

Zwergenpunk:
@jenf Is that all thats needed? It was all I found in 1867

### DIFF
--- a/lib/engine/game/g_1867.rb
+++ b/lib/engine/game/g_1867.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 require_relative '../config/game/g_1867'
-require_relative '../loan.rb'
+require_relative '../loan'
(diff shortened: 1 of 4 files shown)


jenf:
Yeah, can we do it as an module include since we've got it three times now

Zwergenpunk:
like the `CompanyPriceUpToFace` ? 

Yeah that sounds reasonable

Zwergenpunk:
I'm not sure about the naming of the module. If someone has a better name. I'm fine with that ;)

Should I add it to 1867 myself? I don't want to break anything I don't know :/

jenf:
I would, same with 18 la

Zwergenpunk:
I changed 18LA and 1867.

tobymao (approved):

PR #3286, opened 2021-01-13
Improve Round info on game card (fixes #3200)

### LINKED ISSUE #3200: 18MS: Incorrect round information in front page
Consider this game:

image

If I click on that game it turns out to be OR 7...
https://18xx.games/game/21238?action=187

So round information need to be corrected for 18MS.

perwestling:
This commit makes it possible for titles to control how the Round
information is presented on the game cards. To make this possible
we now starts to store round number as well as turn and round class.

The only title that uses this new method is 18MS.
It now shows the 18MS OR based on available turn and round num.
If round num has not yet been stored, it presents the possible
ORs as a range.
For Stock Round, no change.

Example:
  operating round, turn 4, round_num 2 => Operating 8
  stock round, turn 4 => Stock 4
  operating round, turn 4, round_num nil => Operating 7-8


assets/app/view/game/actionable.rb
```diff
@@ -86,6 +86,7 @@ def process_action(action)
         if hotseat
           @game_data[:turn] = game.turn
           @game_data[:round] = game.round.name
+          @game_data[:round_num] = game.round.round_num
```
tobymao, line 89: what about for non hot seat??

assets/app/view/game_card.rb
```diff
@@ -299,5 +299,14 @@ def render_broken
         ]),
       ])
     end
+
+    private
+
+    def game_card_round_info
+      game = Kernel.const_get(Engine::GAMES_BY_TITLE[@gdata['title']].to_s)
```
tobymao, line 306: isn't there a method in engine for this? why do you need to const_get

lib/engine/game/g_18_ms.rb
```diff
@@ -200,10 +200,31 @@ def or_set_finished
...
       def or_description_short(turn, round)
+        G18MS.or_description_short(turn, round)
+      end
+
+      def self.or_description_short(turn, round)
         ((turn - 1) * 2 + round).to_s
       end
 
-      # Game will end directly after the end of OR 10
+      def self.game_card_round_info(gdata)
```
tobymao, line 210: game engines shouldn't take in gdata

assets/app/view/game/actionable.rb
perwestling: Doesn't line 102 handle that? I looked where turn and round was stored, and only found these lines and line 101+103.

assets/app/view/game_card.rb
perwestling: It creates a symbol, which I can use to call the class method with.

Example:
class Test
  def self.fum; end
end
a = "Test"
b = Kernel.const_get(a)

a.fum throws exception
b.fum works

Well, I did not find any prettier solution, but maybe there is?

lib/engine/game/g_18_ms.rb
perwestling: Ok. I will pass the values on line 211-213 as parameters instead.

tobymao:
i don't think we should do this approach at all. i think we should do a db migration to change what is stored in the game and have games set this string  directly from the engine

perwestling:
Yes, that is probably better. The solution was kind of ugly.... So doing the migration sounds better.

perwestling:
I put this as Draft but it can probably be closed.

jenf:
Note pinned games won't update it properly

perwestling:
New commit message:

**Improve Round info on game card (fixes #3200)**

Game keeps the current game info, and saves this the game is
stored.

The saved game info is used on game card. If game does not have
any it will show blank instead.

The round info is round name (minus Round) followed by the
or_description_short.
This means:
- 18MS show Operating 10 etc
- Standard 18xx show Operating 3.1 etc

The 18MS base class also contains a small refactoring which is
not connected to the fix.

lib/engine/game/base.rb
```diff
@@ -75,7 +75,7 @@ def self.load(data, at_action: nil, actions: nil, pin: nil, optional_rules: nil,
     class Base
       attr_reader :raw_actions, :actions, :bank, :cert_limit, :cities, :companies, :corporations,
                   :depot, :finished, :graph, :hexes, :id, :loading, :loans, :log, :minors,
-                  :phase, :players, :operating_rounds, :round, :share_pool, :stock_market, :tile_groups,
+                  :phase, :players, :operating_rounds, :round, :round_info, :share_pool, :stock_market, :tile_groups,
```
tobymao, line 78: why make this an attr? seems better as just a method that is computed dynamically

def round_info

tobymao:
this doesn't migrate yet? also are you going to remove round and round_num since they may be unused?

perwestling:
Here is how it looks like in game card, one game that I have done action in, and one that I have not:
image

perwestling:
> this doesn't migrate yet? also are you going to remove round and round_num since they may be unused?

I run migration and it did not update round_info. Have not checked the details but is it so that game is not written down to database if unchanged? 
I will try to have round as a method instead and see if that makes a difference.

tobymao:
the round should never be blank

perwestling:
This is a new attempt to get this improved way of showing Round info.

> the round should never be blank

Don't we need to do this for broken games, or hotseat games? What should these display?

perwestling:
Sorry, now I got it, leave out "Round: " as well if empty

perwestling:
This is how it looks after latest change:

image

perwestling:
Updated is left out if round_info is missing. 
So should I show Updated: always if not finished, or new?

perwestling:
Pause this for now. Maybe pick it up at a later time.

PR #3288, opened 2021-01-13
18cz Added buying of only specific trains

Zwergenpunk:
Which train you can buy depends on the corporation type


Zwergenpunk:
this is not correct for the rules -.-

PR #3300, opened 2021-01-14
use cookies instead of local storage for auth so we can avoid refresh…

tobymao:
…ing to login

### DIFF
--- a/api.rb
+++ b/api.rb
@@ -46,8 +46,6 @@ class Api < Roda
   plugin :error_handler
 
   error do |e|
-    puts e.backtrace.reverse
-    puts "#{e.class}: #{e.message}"
     LOGGER.error e.backtrace
     { error: e.message }
   end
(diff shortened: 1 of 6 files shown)


yzemaze:
Performance is good. Home loads fine without flashing background on user/refresh when bg is different than default according to prefers-color-scheme 
Login and Logout work as designed and redirect to Home.

Problems:
1. Signup: After giving consent to pp home is loaded, but user is not logged in. Refresh: User logged in.
2. Profile: `Save Changes` and `Reset to Defaults` both lead to [You are not logged in](https://github.com/tobymao/18xx/blob/9894f31fe079bb7e3b3f0b4c3d5928754e302be2/assets/app/view/user.rb#L40). Refresh: User logged in.
3. Account deletion (either in signup process or later) leads to error msg: `Attempt to delete object did not result in a single row modification (Rows Deleted: 0, SQL: DELETE FROM "sessions" WHERE ("id" = 24))`. User is deleted nonetheless, session did exist and was deleted, too.

PR #3320, opened 2021-01-15
Publish turn notification when game is started

daniel-sousa-me:


daniel-sousa-me:
Even though this is not run in js, I changed the "any?" to "empty?" for consistency. Can remove that if preferred.

routes/game.rb
```diff
@@ -145,7 +145,19 @@ class Api
               halt(400, 'Player count not supported')
             end
 
-            set_game_state(game, engine, users)
+            acting = set_game_state(game, engine, users)
+
+            unless acting.empty?
```
tobymao, line 150: why would this ever be empty?

routes/game.rb
```diff
@@ -145,7 +145,19 @@ class Api
...
+              MessageBus.publish(
```
tobymao, line 151: game_url looks changed here maybe you should turn this into a function

routes/game.rb
daniel-sousa-me: I don't know. Just copied what was above. But indeed it seems highly unlikely that nobody is acting when the games is starting.

routes/game.rb
daniel-sousa-me: Indeed. I don't know how that could have happened (there was no change on this line..)

tobymao (approved):

PR #3322, opened 2021-01-15
18Mag: add RABA, G&C and SNW power purchase

roseundy:
Use SpecialBuy to allow minor to pay for RABA, G&C and SNW "railcar" powers

G&C, RABA, and SNW powers are available via button above route selector:
18Mag_buys

Common file change:
- UI::route_selector - generalize case of showing subsidies (since mines generate those)

G&C power - convert to N+N train:
18Mag_Plus

RABA power - offboard bonus:
18Mag_RABA

SNW power - allow use of mines:
18Mag_Mine

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -257,7 +257,8 @@ module View
                     '(Invalid Route)'
                   end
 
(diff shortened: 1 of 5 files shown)


jenf:
Quick review looks ok, but I've probably missed something

tobymao (approved):

PR #3326, opened 2021-01-15
Rubocop 1.7.0, c_lexer removed

yzemaze:
c_lexer removed
rubocop 0.78.0 => 1.7.0 [1.8.1 clashes with opal’s dependencies]
offenses fixed, cf. commits

I activated new cops if I deemed them useful and was certain their auto-corrections didn’t break anything. Nonetheless please have at least a look at each commit.

I disabled many new/pending cops, which fulfilled at least one of these criteria:
- unsafe
- unsafe auto-correction (false-positives)
- corrections made sense, but made code harder to grasp
- I wasn’t sure if the corrections were truely safe or might/could break something
- I wasn’t sure which style to enforce (cf. EnforcedStyle)
Some cops are marked as TODO. These should be examined by a more experienced dev.

I don’t know if/when `TargetRubyVersion` should be set to 3.0.

Minor issues/warnings:
Rubocop complains:
```
warning: parser/current is loading parser/ruby30, which recognizes
warning: 3.0.0-dev-compliant syntax, but you are running 3.0.0.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

make dev_up_b:
```
queue_1  | Failed to load WithCLexer, using pure Ruby lexer
rack_1   | Failed to load WithCLexer, using pure Ruby lexer
```


tobymao:
we can set TargetRubyVersion to 3

assets/app/view/game/bid.rb
```diff
@@ -17,13 +17,13 @@ def render
...
           max_bid = step.max_bid(@entity, @corporation)
           price_input = h(:input, style: { marginRight: '1rem' }, props: {
-            value: min_bid,
-            step: min_increment,
-            min: min_bid + min_increment,
-            max: max_bid,
-            type: 'number',
-            size: [@entity.cash.to_s.size, max_bid.to_s.size].max,
-          })
+                            value: min_bid,
```
tobymao, line 20: i prefer the previous style, is that easy to fix?

assets/app/view/game/dividend.rb
```diff
@@ -135,13 +135,18 @@ def cleanup
...
         max = (@step.variable_max(entity) / entity.total_shares).to_i
 
-        input = h(:input, style: { margin: '1rem 0px', marginRight: '1rem' }, props: {
-          value: max,
-          min: 0,
-          max: max,
-          type: 'number',
-          size: max.to_s.size,
-        })
+        input = h(:input,
```
tobymao, line 138: all these should be new lines
h(
  :input,
..

assets/app/view/game/bid.rb
yzemaze: Yes, that’s due to `Layout/FirstHashElementIndentation`, but disabling would allow crap like:
```
          price_input = h(:input, style: { marginRight: '1rem' }, props: {
  value: min_bid,
  step: min_increment,
  min: min_bid + min_increment,
  max: max_bid,
  type: 'number',
  size: [@entity.cash.to_s.size, max_bid.to_s.size].max,
})
          place_bid = lambda do
```
So choose wisely.

assets/app/view/game/dividend.rb
yzemaze: Multiline could be enforced by https://docs.rubocop.org/rubocop/1.7/cops_layout.html#layoutmultilinemethodargumentlinebreaks, but that would also affect hundreds of other cases were this wouldn’t be desirable.
I haven’t found a cop to enforce a newline on multiline method class, especially not limited to only h().

PR #3331, opened 2021-01-15
generically capitalize based on number of shares

benjaminxscott:
This covers cases where companies have 4 or 5 shares (etc) without them needing additional logic

Ran the test fixtures, passed

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1278,7 +1278,7 @@ module Engine
 
         return if corporation.capitalization == :incremental
 
-        @bank.spend(corporation.par_price.price * 10, corporation)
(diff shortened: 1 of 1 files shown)


tobymao:
i'm curious why this doesn't break any games? should you refactor other games to leverage this change?

benjaminxscott:
> i'm curious why this doesn't break any games?
I think other games either used incremental cap which doesn't hit this codepath, or the default "10 shares worth of capital" was something that worked for them

> should you refactor other games to leverage this change?
It doesn't look like there's any game engines redefine `float_corporation()` in a way that would break on this change

For context, this is a usecase in 18CZ 
(and presumably others with variable float amounts)

tobymao (approved):

Issue #3297, opened 2021-01-14
[18CO] Able to upgrade normal track when it should be unavailable


Talbatross:
https://18xx.games/game/22553?action=451

CS is able to upgrade J18 even though it can't run any of the track that exists on that tile.

PR #3332, opened 2021-01-15
Fix 18CO Prior Connected Paths Timing

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/3297

In lay_tile, before check_track_restrictions! is called, the new tile is laid and the graph updated. This meant the original timing for calling prior_connected_paths was after the graph was updated, and thus we were examining newly connected paths as well. If the corporation could trace a loop back to the old segment of track, it would be returned as connected.

The new logic sets the prior_connected_paths before the graph is updated, so we are working with the true set of prior_connected_paths.

**Example**

Here CS should NOT have been able to upgrade the tile to connect between Pueblo and Trinidad (as happened in https://18xx.games/game/22553?action=451 ). The old logic failed because of the loop between Colorado Springs and Pueblo.

Screen Shot 2021-01-14 at 5 54 31 PM

### DIFF
--- a/lib/engine/step/g_18_co/tracker.rb
+++ b/lib/engine/step/g_18_co/tracker.rb
@@ -25,6 +25,15 @@ module Engine
           @log << "#{corporation.name} collects a mine token from #{hex.name}"
         end
 
+        def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
+          entity ||= action.entity
+          @previous_connections = prior_connected_paths(entity, action.hex.tile)
+
+          super
+
+          @previous_connections = []
+        end
+
         def check_track_restrictions!(entity, old_tile, new_tile)
           return true if pending_token(entity)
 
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_18_co/tracker.rb
```diff
@@ -25,6 +25,15 @@ def collect_mines(corporation, hex)
           @log << "#{corporation.name} collects a mine token from #{hex.name}"
         end
 
+        def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
+          entity ||= action.entity
+          @previous_connections = prior_connected_paths(entity, action.hex.tile)
```
tobymao, line 30: looks like this calls the graph outside of game load, this will kill performance

ryandriskel: > looks like this calls the graph outside of game load, this will kill performance

I don't see how I can avoid checking the graph ( the previous implementation also used the graph ) but previously it only checked on a very limited set of tiles. In order to get back to that state, I would have to to duplicate the code in both the base check_track_restrictions! as well as 18COs check_track_restrictions!.

Do you have a different suggestion for how I can find out the reachable segments of track on a hex for an entity before a tile is laid?

tobymao: can you only check this when !@game.loading

ryandriskel: Why would this kill performance anyway? Its not regenerating the graph. Base `check_track_restrictions!` calls the same `@game.graph.connected_paths(entity)` repeatedly.

tobymao: only when !game.loading

tobymao (approved):

PR #3333, opened 2021-01-16
Remove closed AFG from round after quake

### LINKED ISSUE #3328: 1849, Issue with earthquake
We had the Messina earthquake happen, and then AFG was supposed to close. It does show as possible to start next stock round, but it is still forced to buy a train during the OR. When we try to bypass it, it says that the server did not accept the request. 

Game id is:

https://18xx.games/game/23001

It's 3p, the issue occured in OR5.2 when I bought the 12H train.

tysen:
Fixes #3328

### DIFF
--- a/lib/engine/game/g_1849.rb
+++ b/lib/engine/game/g_1849.rb
@@ -180,7 +180,7 @@ module Engine
       def remove_corp
         removed = @corporations.pop
         @log << "Removed #{removed.name}"
-        return if removed.name == 'AFG'
+        return if removed == afg
 
         hex_by_id(removed.coordinates).tile.city_towns.first.remove_reservation!(removed)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1849.rb
```diff
@@ -180,7 +180,7 @@ def setup_companies
       def remove_corp
         removed = @corporations.pop
         @log << "Removed #{removed.name}"
-        return if removed.name == 'AFG'
+        return if removed.name == afg
```
tobymao, line 183: shouldn't this just be removed == afg or removed.name == afg.name?

afg is a corp object right

tysen: oops. fixed.

tobymao (approved):

PR #3347, opened 2021-01-17
1824: Implement more, part 1

perwestling:
General codes changes needed for 1824:
- Added read/write attribute 'type' to minor (no default) which is
  needed for train limit handling when minor buy trains. The current
  implementation has Pre-Staatsbahn as minor corporation that can
  buy trains.
- Move purchasable_unsold_companies to game base, and change default
  to empty list. So far only 1824 uses this and then 1824 override
  it in its game class.
- Added read/write attribute 'floatable' to corporation (default true).
  If false, will not allow corporation to be floated regardless of
  amount sold.
  This is used in 1824 for corporations where president
  share is reserved, unsold, and 10% shares can still be bought.
  Shown in corporation charter that it cannot float.
- Implemented SELL_MOVEMENT :down_block which moves share prize one
  step down regardless of number of shares sold. This is needed by
  1824, 18CZ and more.
- Set can_buy_any_from_ipo? to true in case any share is buyable.
  Previously it just checked first share. This is needed for
  Staatsbahn which can have 10% shares sold even though presidency
  share is reserved.

But the main part of this commit is features for 1824, as well
as lots of fixes and improvements. Here are a few of the implementations:
- Implement removal (from game) of unsold Mountain, Coal and Pre-
  Staatshbahn railways.
- When Coal Railway sold (buying the 100% certificate) this pars
  the aassociated Staatsbahn.
- Set up reserved shares for Coal Railways.
- Set up train limits per type of corporation/minor.
- 50/50 dividend for Pre-Staatsbahn, and Coal railways.
- g-Trains must visit exactly one mine. Normal trains may not
  visit mines.
- Anyone can buy/use g-trains, but Coal railways can only own g-trains.
- g-trains cannot be bought until first 2 train bought.
- Trains can only be bought from other corporation from phase 3.
- Mine visit pays to treasury.
- Mine now show revenue for all colors.
- Operating is:
  1. Coal railways in pre-set order
  2. Pre-staatsbahn in pre-set order
  3. Regionals and Staatsbahn in usual (stock market prize) order

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -22,8 +22,7 @@ module Engine
     include ShareHolder
     include Spender
 
-    attr_accessor :ipoed, :par_via_exchange, :max_ownership_percent, :float_percent, :capitalization, :max_share_price,
-                  :floatable
+    attr_accessor :ipoed, :par_via_exchange, :max_ownership_percent, :float_percent, :capitalization, :max_share_price
     attr_reader :companies, :min_price, :name, :full_name, :fraction_shares, :type, :id, :needs_token_to_par,
                 :presidents_share
     attr_writer :par_price, :share_price
@@ -54,7 +53,6 @@ module Engine
       @closed = false
       @float_percent = opts[:float_percent] || 60
       @floated = false
-      @floatable = true
       @max_ownership_percent = opts[:max_ownership_percent] || 60
       @min_price = opts[:min_price]
       @always_market_price = opts[:always_market_price] || false
@@ -159,11 +157,7 @@ module Engine
     end
 
     def floated?
-      @floatable && (@floated ||= percent_of(self) <= 100 - @float_percent)
-    end
-
-    def floatable?
-      @floatable
+      @floated ||= percent_of(self) <= 100 - @float_percent
     end
 
     def percent_to_float
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -22,7 +22,8 @@ module Engine
     include ShareHolder
     include Spender
 
-    attr_accessor :ipoed, :par_via_exchange, :max_ownership_percent, :float_percent, :capitalization, :max_share_price
+    attr_accessor :ipoed, :par_via_exchange, :max_ownership_percent, :float_percent, :capitalization, :max_share_price,
+                  :floatable
     attr_reader :companies, :min_price, :name, :full_name, :fraction_shares, :type, :id, :needs_token_to_par,
                 :presidents_share
     attr_writer :par_price, :share_price
@@ -53,6 +54,7 @@ module Engine
       @closed = false
       @float_percent = opts[:float_percent] || 60
       @floated = false
+      @floatable = true
       @max_ownership_percent = opts[:max_ownership_percent] || 60
       @min_price = opts[:min_price]
       @always_market_price = opts[:always_market_price] || false
@@ -157,7 +159,11 @@ module Engine
     end
 
     def floated?
-      @floated ||= percent_of(self) <= 100 - @float_percent
+      @floatable && (@floated ||= percent_of(self) <= 100 - @float_percent)
+    end
+
+    def floatable?
+      @floatable
     end
 
     def percent_to_float
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -762,10 +762,6 @@ module Engine
         end
       end
 
-      def purchasable_unsold_companies
-        []
-      end
-
       def player_value(player)
         player.value
       end
@@ -950,9 +946,7 @@ module Engine
       end
 
       def float_str(entity)
-        return unless entity.corporation?
-
-        entity.floatable? ? "#{entity.percent_to_float}% to float" : 'Not floatable'
+        "#{entity.percent_to_float}% to float" if entity.corporation?
       end
 
       def route_distance(route)
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -892,6 +892,8 @@ module Engine
           (percent / 10).to_i.times { @stock_market.move_down(corporation) }
         when :down_block
           @stock_market.move_down(corporation)
+        when :left_block
+          @stock_market.move_left(corporation)
         when :left_block_pres
           stock_market.move_left(corporation) if was_president
         when :none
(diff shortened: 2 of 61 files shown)


lib/engine/corporation.rb
```diff
@@ -157,7 +159,11 @@ def president?(player)
     end
 
     def floated?
-      @floated ||= percent_of(self) <= 100 - @float_percent
+      @floatable && (@floated ||= percent_of(self) <= 100 - @float_percent)
```
tobymao, line 162: why does this need to check floatable also if it does it should probably check floatable?

tobymao: since it's attr accessor, i would remove the method definition floatable? and just use floatable

lib/engine/game/base.rb
```diff
@@ -762,6 +762,10 @@ def purchasable_companies(entity = nil)
         end
       end
 
+      def purchasable_unsold_companies
```
tobymao, line 765: who uses this method? didn't this change from the definition in step?

lib/engine/game/g_1824.rb
```diff
@@ -228,41 +237,126 @@ def init_round
...
       def ipo_name(entity)
-        return 'Treasury' if coal_railways.include?(entity)
+        return 'Treasury' if coal_railway?(entity)
 
         'IPO'
       end
 
       def can_par?(corporation, parrer)
-        super && !corporation.all_abilities.find { |a| a.type == :no_buy }
+        super && !corporation.all_abilities.find { |a| a.type == :no_buy } &&
```
tobymao, line 295: all_ability.any? or none instead of find beacuse you don't use the return type

lib/engine/game/g_1824.rb
```diff
@@ -228,41 +237,126 @@ def init_round
...
+          state_sd
+        when 'U1', 'U2'
+          state_ug
+        when 'K1', 'K2'
+          state_kk
+        end
+      end
+
+      def operating_order
+        @corporations.select(&:floated?).select { |c| coal_railway?(c) } +
```
tobymao, line 357: combined the selects so you don't do two passes

lib/engine/round/g_1824/stock.rb
```diff
@@ -37,6 +37,58 @@ def next_entity_index!
...
+            remove_reservation(pre_state)
+
+            # Remove home token
+            pre_state.tokens.first.remove!
+
+            pre_state.close!
+            p.close!
+          end
+
+          @game.corporations.select { |c| @game.coal_railway?(c) }.reject(&:floated?).each do |c|
```
tobymao, line 66: just do a next if, you're doing 3 passes over the data here

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -177,7 +177,7 @@ def can_buy_any_from_market?(entity)
       end
 
       def can_buy_any_from_ipo?(entity)
-        @game.corporations.any? { |c| c.ipoed && can_buy?(entity, c.shares.first&.to_bundle) }
+        @game.corporations.any? { |c| c.ipoed && c.shares.any? { |s| can_buy?(entity, s&.to_bundle) } }
```
tobymao, line 180: why would s be nil?

lib/engine/corporation.rb
perwestling: Have now replaced floatable? with floatable accessor.

The idea of this accessor is to make it possible to sell 10% shares, while still having 20% reserved. The sell of 10% might still exceed floated percent but as long as the 20% certificate is not sold/exchanged the corporation cannot float.

So when the exchange/merge happens 1824 will change floatable to true. (As it is set to false for these Staatsbahn corporation which have presidency certificate reserved for exchange with Coal corporation.)

perwestling: An alternative is to subclass corporation (as 1856 does) but I think it might be confusing for the corporation charter if 50% of the shares is sold and it would show "0% to float" or even "-10% to float"? That could be handled by introducing a game specific method for the presentation.
Have not tried that case though, could do that. Maybe I could handle that case in a graceful way in the presentation?
(e.g. if floated? is false, and percent to float not positive, put in a suitable text)

lib/engine/game/base.rb
perwestling: In general code it is use in stock.rb (view) as one of the criterias if BuyCompaniesAtFaceValue should be rendered.
The BuyCompaniesAtFaceValue is in general code as well.

Default implementation is in lib/engine/game/base.rb and return empty list, which means the BuyCompaniesAtFaceValue is not normally rendered in the Stock round. 
1860 had the situation that it could have unsold purchasable companies, and therefor this was rendered by mistake in that game, causing it to break.

The reason it broke was that the default was to show any unsold purchasable companies, and then override in 1860.

But in 1824 I now thought I needed this method in round/g_1824/stock.rb as the active step is no longer present when stock round is finished.

So I refactored this to put the onus on 1824 to implement the method, and remove the implementation in 1860 (that did override it to return empty list).

If you want I can change it back a bit, putting it in step base instead?  I still think default can be empty list.
1824 can handle it by having it in its game base, and call it from the 1824 step and 1824 stock round, to avoid duplication of implementation.

lib/engine/game/g_1824.rb
perwestling: I changed it to none? 
Also realized there was a bug with the regional so I refactored that statement, and a bit of the code as well.

lib/engine/game/g_1824.rb
perwestling: Changed to use partition instead

lib/engine/round/g_1824/stock.rb
perwestling: Have refactored a bit so it is one loop over companies and one loop over corporations.

lib/engine/game/base.rb
perwestling: When coming down to your comment about stock I realized that I do get a double loop over companies. So avoiding the duplication is probably worthless....

So I will change it back a bit to move method to step instead. But I still think changing default to [] is good to avoid problems in other titles.

lib/engine/corporation.rb
perwestling: In my new commit I subclassed corporation (and minor) to remove this from general code. Note sure if any other 18xx title uses this so it might be just as well.

lib/engine/step/buy_sell_par_shares.rb
perwestling: Have now removed &.

perwestling:
@tobymao I put some info about the second commit in that commit message.

Maybe I should squash these into one commit with one commit message?

PR #3352, opened 2021-01-17
18Mag: functionality additions

roseundy:
Most minor changes:
- Fixed orientation for BudaPest
- Move left by one on block sales
- Set must_buy_train false
- Permissive tile lays
- Corp token placement and bonus

Common file changes:
- UI::route_select - clear route caches when refreshing
- Game::base - support for :down_block and :left_block
- Step::tracker - add log message when home token needs to be placed on upgrade
- Engine::route - calculate connection_data before clearing cache

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -233,6 +233,7 @@ module View
 
         clear = lambda do
           @selected_route&.reset!
+          @game.reset_route!(@selected_route)
           store(:selected_route, @selected_route)
         end
 
@@ -240,12 +241,14 @@ module View
           @selected_route = nil
           store(:selected_route, @selected_route)
           @routes.clear
-          store(:routes, @routes)
           @game.round.routes = @routes
+          @game.reset_all_routes!
+          store(:routes, @routes)
         end
 
         clear_all = lambda do
           @routes.each(&:reset!)
+          @game.reset_all_routes!
           store(:routes, @routes)
         end
 
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -86,7 +86,6 @@ module View
           @selected_route = @routes.first
           store(:routes, @routes, skip: true)
           store(:selected_route, @selected_route, skip: true)
-          @game.round.routes = @routes
         end
 
         if !@selected_route && (first_train = trains[0])
@@ -94,8 +93,9 @@ module View
           @routes << route
           store(:routes, @routes, skip: true)
           store(:selected_route, route, skip: true)
-          @game.round.routes = @routes
         end
+
+        @routes.each(&:clear_cache!)
 
         render_halts = false
         trains = trains.flat_map do |train|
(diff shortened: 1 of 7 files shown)


assets/app/view/game/route_selector.rb
```diff
@@ -86,13 +86,15 @@ def render
...
           store(:selected_route, @selected_route, skip: true)
+          @game.round.routes = @routes
         end
 
         if !@selected_route && (first_train = trains[0])
           route = Engine::Route.new(@game, @game.phase, first_train, abilities: @abilities, routes: @routes)
           @routes << route
           store(:routes, @routes, skip: true)
           store(:selected_route, route, skip: true)
+          @game.round.routes = @routes
```
tobymao, line 97: why does this need to be done?

lib/engine/game/g_18_mag.rb
```diff
@@ -427,6 +454,18 @@ def routes_subsidy(routes)
         routes.sum(&:subsidy)
       end
 
+      def reset_route!(route)
+        @round.snw_train = nil if route.train == @round.snw_train
```
tobymao, line 458: why dous the route have any affect on the round like this?

assets/app/view/game/route_selector.rb
roseundy: I need a way to clear the route cache when the user buys route "powers" using SpecialBuy actions through Step::Route. The process_special_buy method needs to clear the cache because the powers bought can/will affect revenue calculations. This change to the UI::route_selector is needed so that process_special_buy can reference the routes being manipulated by the route_selector before the final submit action. I could have added a new method to do this, but round.routes is already accessible.

lib/engine/game/g_18_mag.rb
roseundy: I use round_state to communicate between the SpecialBuy actions that are used to purchase "powers" that affect the route (bonus revenues, plus-train upgrades, ability use mines) and the revenue calculations that are done in Route via game-specific methods.

roseundy: I needed to add a reset_route! method to force the powers to be recalculated when the user clears or resets the routes via the UI.

assets/app/view/game/route_selector.rb
roseundy: Deleted. UI now clears route cache

lib/engine/game/g_18_mag.rb
roseundy: Deleted. The "railcar" calculations are now stateless.

lib/engine/route.rb
```diff
@@ -27,6 +27,7 @@ def initialize(game, phase, train, **opts)
     end
 
     def clear_cache!
+      connection_data
```
tobymao, line 30: maybe only do this if @connection_data already?

roseundy: Actually, it needs to happen if !@connection_data. That's the case where blowing away @connection_hexes prevents reconstruction of @connection_data

roseundy: Changed, per our discussion.

tobymao (approved):

PR #3358, opened 2021-01-17
Fixed purchasable share count

Zwergenpunk:
The previous soultion did not work as expected

With 10 Shares and 60% max_ownership:
60/10 = 6 - > works
with 4 shares and 75% max ownership:
75/4 => 18,75 -> meeh

the new soultion should work for all values
With 10 Shares and 60% max_ownership:
(60/100) * 10 = 6
with 4 shares and 75% max ownership:
(75/100) * 4 = 3

### DIFF
--- a/assets/app/view/game/par.rb
+++ b/assets/app/view/game/par.rb
@@ -33,8 +33,9 @@ module View
             },
             on: { click: par },
           }
+
           purchasable_shares = [(entity.cash / share_price.price).to_i,
-                                @corporation.max_ownership_percent / @corporation.total_shares].min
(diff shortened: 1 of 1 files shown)


Zwergenpunk:
I just doublechecked and if you set both equations equal it holds only if share value = 10

benjaminxscott:
seems like it works for 10-share companies
Screenshot 2021-01-17 at 3 30 07 PM

seems like it works for 4-share companies (max ownership is 60% so player can buy up to 3 shares)
Screenshot 2021-01-17 at 3 28 25 PM

PR #3382, opened 2021-01-18
First commit for 1835

thepackrat:
I'm working through the abandoned 1835 data file which didn't actually work out of the box and which was padded with a bunch of placeholder data. 

I've stripped it right back to something which works well enough to display a trivial map/market and am committing this  ahead of slowly working through checking all the data against my copy and I'll submit them as a series of small merges which will be easier to review.

### DIFF
--- a/lib/engine/config/game/g_1835.rb
+++ b/lib/engine/config/game/g_1835.rb
@@ -1,0 +1,586 @@
+# coding: utf-8
+# frozen_string_literal: true
+
+# TODO list (packrat)
+# trains are correct
+# cert limits are correct
+# cash is correct
+
+# phases aren't checked yet.
+
+# tiles are not correct.
+# market is not correct
+# map hexes are not correct
+# companies are not correct.
+# minors aren't there
+# privates aren't correct.
+
+# File original exported from 18xx-maker/export-rb
+# https://github.com/18xx-maker/export-rb
+# rubocop:disable Lint/RedundantCopDisableDirective, Layout/LineLength, Layout/HeredocIndentation
+
+module Engine
+  module Config
+    module Game
+      module G1835
+        JSON = <<-'DATA'
+{
+  "filename": "1835",
+  "modulename": "1835",
+  "currencyFormatStr": "%dM",
+  "bankCash": 12000,
+  "certLimit": {
+    "3": 19,
+    "4": 15,
+    "5": 12,
+    "6": 11,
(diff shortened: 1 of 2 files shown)


lib/engine/config/game/g_1835.rb
```diff
@@ -0,0 +1,583 @@
...
+      "75",
+      "80",
+      "90p",
+      "100"
+    ]
+  ],
+  "companies": [
+    {
+      "name": "Private with an icon",
+      "sym": "P1",
```
perwestling, line 76: Think you can have different syms. Not sure if this cause any problems though.

lib/engine/config/game/g_1835.rb
```diff
@@ -0,0 +1,583 @@
...
+      "sym": "MS",
+      "name": "Black Railroad",
+      "logo": "1835/MS",
+      "tokens": [
+        0,
+        40,
+        100
+      ],
+      "coordinates": "C11",
+      "color": "red"
```
perwestling, line 129: Shouldn't it be black? ;-)

lib/engine/config/game/g_1835.rb
```diff
@@ -0,0 +1,583 @@
...
+    {
+      "name": "1.1",
+      "train_limit": 4,
+      "tiles": [
+        "yellow"
+      ],
+      "operating_rounds": 1
+    },
+    {
+      "name": "1.1",
```
perwestling, line 437: Use unique names for the phases.
I haven't played 35 in over 25 years so I do not remember any details.
If phase is triggered by train buy you might need 'on'.

perwestling (approved):
Just some minor comments.

lib/engine/config/game/g_1835.rb
thepackrat: All the phase stuff and companies are placeholders that I carved out of the original template that are left because they weren't causing problems. THere' sa list at the top, including these, of what still needs to be done. They're placeholders becuase I need them to work on the map.

lib/engine/config/game/g_1835.rb
thepackrat: placeholder only. THey're all getting cleaned up in the next commit, ~none of the company stuff was correct from where I started, but at least this renders a map and keeps each commit phase relatively small.

tobymao (approved):

PR #3383, opened 2021-01-18
login with username

### LINKED ISSUE #3319: Allow login via username
A few people are logging in using username rather than email.
Allowing username would avoid this confusion.

yzemaze:
Caveats as discussed earlier.

fixes #3319

### DIFF
--- a/routes/user.rb
+++ b/routes/user.rb
@@ -8,8 +8,7 @@ class Api
       r.post do
         # POST '/api/user/login'
         r.is 'login' do
-          halt(400,
-               'Could not find user') unless (user = User.by_email(r['email']) || user = User.first(name: r['email']))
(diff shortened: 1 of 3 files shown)


routes/user.rb
```diff
@@ -8,7 +8,8 @@ class Api
       r.post do
         # POST '/api/user/login'
         r.is 'login' do
-          halt(400, 'Could not find user') unless (user = User.by_email(r['email']))
+          halt(400,
+               'Could not find user') unless (user = User.by_email(r['email']) || user = User.first(name: r['email']))
```
tobymao, line 12: you don't need to the user =

tobymao: instead change by_email to also search for the username, 

so in models/user.rb adjust that method

it's important to use the lower case functions

tobymao (approved):

PR #3393, opened 2021-01-18
1824: Implement stock round exchange for mountain railway

perwestling:
Make it possible to exchange Mountain Railway private for
share in any regional railway, during phase 3.

Also implement that exchanged trains are removed from game.

To support this, two changes was needed in the general code:

1. Make it possible to specify an array of corporations names
   in ability exchange. Have updated this handling in all
   places this was used. Some titles do only use one value
   and in that cases only the first corporation is used.

2. Made tracker of actions for player round state. This
   round state is reset when a player has completed its
   turn (via unpass!).
   This is needed to make it possible for exchange step
   to handle an exchange as a buy. (This is so in 1824
   while in 1889 you can exchange + buy [but not buy +
   exchange in that order].)
   The effect is that an exchange in 1824 is like a buy,
   except that you get the message that player pass as
   player does not have any valid actions. But at least
   it will not trigger end of stock round.

### DIFF
--- a/lib/engine/ability/README.md
+++ b/lib/engine/ability/README.md
@@ -107,7 +107,9 @@ Provide a description for an ability that is implemented outside of the ability 
 
 Exchange this company for a share of a corporation.
 
-- `corporation`: The corporation whose share may be exchanged. Use `"any"` to allow for all corporations.
+- `corporation`: A single corporation, an array of corporations, or `"any"`.
+  Specifies which corporation whose share may be exchanged.
+  Use `"any"` to allow for all corporations.
 - `from`: Where the share may be take from, either `"ipo"`,
   `"market"`, or an array containing both.
 
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -28,7 +28,7 @@ module Engine
       end
 
       def log_pass(entity)
-        return @log << "#{entity.name} passes" if @current_actions.empty?
+        return @log << "#{entity.name} passes" if @round.player_actions.empty?
         return if bought? && sold?
 
         action = bought? ? 'to sell' : 'to buy'
@@ -51,7 +51,7 @@ module Engine
       end
 
       def pass_description
-        if @current_actions.empty?
+        if @round.player_actions.empty?
           'Pass (Share)'
         else
           'Done (Share)'
@@ -59,7 +59,13 @@ module Engine
       end
 
       def round_state
-        { players_sold: Hash.new { |h, k| h[k] = {} } }
+        { players_sold: Hash.new { |h, k| h[k] = {} }, player_actions: [] }
+      end
+
+      def unpass!
+        super
+
+        @round.player_actions = []
       end
 
       def setup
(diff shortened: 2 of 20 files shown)


lib/engine/ability/README.md
```diff
@@ -107,7 +107,9 @@ Provide a description for an ability that is implemented outside of the ability
 
 Exchange this company for a share of a corporation.
 
-- `corporation`: The corporation whose share may be exchanged. Use `"any"` to allow for all corporations.
+- `corporation`: A single corporation, an array of corporations, or `"any"`.
```
tobymao, line 110: no, this is bad, change it to be corporations so it's not ambiguous what corporation is

tobymao: there can be corporation which is used when you expect one,

or corporations, when you expect multiple

tobymao: or you should change everything to use corporations

choose one approach

tobymao: choices

corporation and corporations attributes -- this seemse asier

only corporations attribute

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -70,7 +76,7 @@ def setup
           corps.each { |corp, _k| corps[corp] = :prev }
         end
 
-        @current_actions = []
+        @round.player_actions = []
```
tobymao, line 79: is this needed? shouldn't the initializer do that?

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -145,12 +151,12 @@ def process_par(action)
...
         @game.after_par(corporation)
         @round.last_to_act = entity
-        @current_actions << action
+        @round.player_actions << action
       end
 
       def pass!
         super
-        if @current_actions.any?
+        if @round.player_actions.any?
```
tobymao, line 159: unless empty

lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -51,15 +51,21 @@ def description
...
+        if @round.player_actions.empty?
           'Pass (Share)'
         else
           'Done (Share)'
         end
       end
 
       def round_state
-        { players_sold: Hash.new { |h, k| h[k] = {} } }
+        { players_sold: Hash.new { |h, k| h[k] = {} }, player_actions: [] }
```
tobymao, line 62: why don't you make the exchange a part of your custom buy_sell_par_shares, that way you don't have to make this global

tobymao:
this change is too invasive, i think you should try to make it more custom to 1824

perwestling:
I think I will split this into two commits.
1. One with the general stuff about the exchange ability, and I think I will add a 'corporations' attribute that will take an array. Both 'corporation' and 'corporations' attribute will exist, but probably a check that not both used at the same time.
2. One 1824 specific commit, that adds exchange functionaly to buy_sell_par (which will probably be renamed to buy_sell_par_exchange ;-) ) in 1824.

So will probably make a new PR for the 1 above, and change this PR to be 2 above. Might also continue to work on the 2 part as there still need some functionality for Mountain Railways exchange, as in 1824 there is an interrupt step of exchange in case there is any Mountain Railway privates left when the first 4 train is bought. (And Mountain Railway is lost without compensation if there is no available shares to exchange.) 

In other words, the general code change PR I should be able to put up later today, while the 1824 specific exchange code I will do later. Have put this PR in draft in the mean time.

perwestling:
I close this one as I almost have a working new version.

PR #3401, opened 2021-01-19
1867, 1817*: Add user setting to use plain logos for major's tokens

ventusignis:
The optional logos are simply the corporation's abbreviation on a colored token
image
image
image

Tried extending it to 1817
image

User setting page
image


tobymao:
this is interesting, and not very much code, but i'm a bit opposed to this. we should replace the current tokens with these  or not have the fancy ones.

compress the svgs please

tobymao:
@jenf thoughts?

jenf:
I don't like it being a game option because that makes everyone play with those logos.

I have no problem it being a personal preference

tobymao:
What about making all the logos the plain ones?

ventusignis:
I can look into how to make this controlled by player config instead of a game option; while I'm at it I might loop in 1817* for this as I've seen complaints of certain pairs looking too similar (STR/BA, H/PW come to mind)

ventusignis:
The current logos look nice and mimic the GTG print of 1867 so I think there will be a large portion of players that would prefer to use the existing set of tokens

ventusignis:
Considering the alt-logos are just 400 bytes each (circle, fill, a text box) the alt-logos shouldn't add a significant size to the deployable bundle

jenf:
This probably will be a can't please everyone situation. So I'd probably suggest having it as an option.

ryandriskel:
I've had a similar request for 18CO. The tokens currently use the same art as the corporation's real logo and the game's components, but some would prefer the ticker. I think the player preference option to use text-based tokens when available would please most people, with the only downsides being additional work for creators and possible confusion when two players compare their games?

tobymao:
My preference is to only have text tokens.

jenf:
Maybe have basic tokens be the default and the high-graphics being a user option.

Question is why should these tokens be text and not say 1846 or 1817

tobymao:
My preference would be for it to all be text, but creativity to the
implementor or designer of the game.

ventusignis:
I changed this PR a bit, I'm curious what everyone thinks of this.
It's now a user setting that controls this; the code changes to support this is a little more intrusive as several view files had to be altered. Games can support this by adding a `simple_logo` field to a corporation in their config and that logo will be used when a user has the "simple logos" setting enabled. I also added simple logos for the 1817 games as they have many corporations whose tokens can be a bit similar

The SVGs are about 240 bytes each after I simplified them as much as reasonably possible. Since they are so small, new logos in the same style can be created for other games by simply copying and manually editing the SVGs. All that needed to be altered for the 1817 corporations were:
* background color
* text color
* corporation name

perwestling:
Interesting. Maybe I should try this for 1824? Here we added the actual SVGs from the physical game instead of the trivial ones. But using simplified logos might be worth trying.

Also, should we make a markdown documentation file for these attributes? Or is there one?
Anyway, that could be a separate enhancement.

jenf:
Generally looks good to me :)

tobymao (approved):

jenf:
We could generate the simplified logos in code

yzemaze:
We could, but why if we can use browser caching?

jenf:
yeah, browser caching and a script is best

PR #3402, opened 2021-01-19
1835: Small corrections to private powers in regard to share ownership

neutronc:
Private Bavarian Companies come with 10 percent  of the Bavarian Corporation


lib/engine/config/game/g_1835.rb
```diff
@@ -197,15 +197,24 @@ module G1835
           ],
           "free": true,
           "count": 1
-        }
+        },
```
perwestling, line 200: Formatting seems a bit off, here and below.

perwestling (approved):
This seems to be on the smaller side for PRs.... There is some overhead with review, merge, deploy so I think it is a good idea to not have too many PRs. At the same to it is good to break down into managable PRs.

lib/engine/config/game/g_1835.rb
jenf: Please fix the probably tabs/space issue, then it's good to go.

neutronc:
Cant get the merge out of this one, sorry. Next time i'll remember to rebase my fork :(

tobymao (approved):

PR #3404, opened 2021-01-19
game cards: fix placement/alignment of buttons

yzemaze:
Same width for all game related buttons (join, leave, start, delete, confirm, invite and review). Placement rtl ttb. If we’d ever need more buttons they’d follow suit. If longer captions were needed (minor) changes would become necessary.
Same height on all headers would be possible but would waste space on short titles (basically all but 18Chessie:OtR) with <= 2 buttons when the games are running.

ante:
image

post:
image

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -18,6 +18,13 @@ module View
     needs :confirm_kick, store: true, default: nil
     needs :flash_opts, default: {}, store: true
 
+    BUTTON_STYLE = {
+      margin: '0',
+      padding: '0.2rem 0',
+      textAlign: 'center',
+      width: '3.5rem',
+    }.freeze
+
     def render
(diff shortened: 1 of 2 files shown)


assets/app/view/game_card.rb
```diff
@@ -158,6 +169,20 @@ def render_optional_rules
       h(:div, [h(:strong, 'Optional Rules: '), rendered_rules])
     end
 
+    def render_invite_link
+      return unless new? && owner?
```
tobymao, line 173: change to if statement

tobymao (approved):

assets/app/view/game_card.rb
yzemaze: That would trigger Rubocop Style/GuardClause.

tobymao (approved):

PR #3413, opened 2021-01-19
18CZ Added laying Home Track + Token when operated

Zwergenpunk:


lib/engine/game/g_18_cz.rb
```diff
@@ -252,6 +269,58 @@ def train_of_size?(item, size)
...
+
+      def place_home_token(corporation)
+        return unless corporation.next_token # 1882
+        # If a corp has laid it's first token assume it's their home token
+        return if corporation.tokens.first&.used
+
+        hex = hex_by_id(corporation.coordinates)
+
+        tile = hex&.tile
+        if !tile || (tile.reserved_by?(corporation) && tile.paths.any?)
```
tobymao, line 286: !tile.paths.empty?

lib/engine/step/g_18_cz/home_track.rb
```diff
@@ -0,0 +1,109 @@
...
+
+        def reachable_hex?(_entity, _hex)
+          true
+        end
+
+        def available_hex(_entity, hex)
+          pending_track[:hexes].include?(hex)
+        end
+
+        def hex_neighbors(entity, hex)
```
tobymao, line 87: what is this, it's not called anywhere? did you just copy this stuff ?

Zwergenpunk: Yes. You're right. I should remove all unneeded code. I just didn't look if I need it.

tobymao: please double check everything you've done in this commit to see if it's necessary and to understand what it's doing

lib/engine/game/g_18_cz.rb
```diff
@@ -87,6 +87,22 @@ class G18CZ < Base
...
 
+      OFFBOARD_TILES_FOR_CORPORATION = {
+        'Ug' => '8894',
+        'kk' => '8895',
+        'SX' => '8896',
+        'PR' => '8897',
+        'BY' => '8898',
+      }.freeze
+
+      COORDINATES_FOR_LARGE_CORPORATION = {
```
benjaminxscott, line 98: just to note that the [config already has the coords](https://github.com/tobymao/18xx/blob/master/lib/engine/config/game/g_18_cz.rb#L1136)

lib/engine/config/game/g_18_cz.rb
```diff
@@ -1133,10 +1133,6 @@ module G18CZ
                         0,
                         40
                     ],
```
benjaminxscott, line 1135: deleting coords from corp configs means that the card looks wonky, since the home coordinates went poof

Screenshot 2021-01-19 at 8 23 29 PM

lib/engine/config/game/g_18_cz.rb
```diff
@@ -169,7 +169,7 @@ module G18CZ
                 "8896": {
                     "count": 1,
                     "color": "red",
-                    "code": "city=revenue:yellow_20|green_30|brown_40|gray_50;path=a:0,b:_0,terminal:1;path=a:1,b:_0,terminal:1;icon=image:18_cz/50"
+                    "code": "city=revenue:yellow_20|green_30|brown_40|gray_50;path=a:0,b:_0,terminal:1;path=a:1,b:_0,terminal:1;icon=image:18_cz/50;label=SX"
```
benjaminxscott, line 172: do we want to add requisite labels to the other special offboards as well?
Screenshot 2021-01-19 at 8 28 33 PM

benjaminxscott: actually I'd suggest we don't label the tiles, since it looks a bit odd when there's also the corp's token on it

Screenshot 2021-01-19 at 8 30 27 PM

benjaminxscott: check this commit out, which adds those labels and removes the need for a constant lookup dict (since we can choose tiles by their label)
https://github.com/benjaminxscott/18xx/commit/b0d98824ce2f61748a698c90a30e731de93d18d6

lib/engine/config/game/g_18_cz.rb
benjaminxscott: check this commit out which uses coordinate lookups instead of hardcoding

https://github.com/benjaminxscott/18xx/commit/bedbef8021ef8f2d704940a5f148179059065cc0

lib/engine/step/g_18_cz/home_track.rb
Zwergenpunk: I removed all unused code and refactored most of it away so its just the code I need.

lib/engine/config/game/g_18_cz.rb
Zwergenpunk: I rewrote the placing of home_tile to use the coordinates direct. Thanks for the info that they are important

lib/engine/config/game/g_18_cz.rb
Zwergenpunk: thanks. thats better

tobymao (approved):

PR #3415, opened 2021-01-19
1817/1867/18CO: Make common the tile lay restriction that

jenf:
you cannot upgrade the same hex.

Also make the UI say if you can upgrade/lay at that point


lib/engine/step/tracker.rb
```diff
@@ -26,6 +27,7 @@ def get_tile_lay(entity)
         action[:lay] = !@upgraded if action[:lay] == :not_if_upgraded
         action[:upgrade] = !@upgraded if action[:upgrade] == :not_if_upgraded
         action[:cost] = action[:cost] || 0
+        action[:upgrade_same_hex] = true if action[:upgrade_same_hex].nil?
```
tobymao, line 30: unless action[:upgrade}

tobymao (approved):

ryandriskel:
Nice find Jen!

jenf:
Don't merge this checking an issue

lib/engine/step/tracker.rb
jenf: that doesn't work since action[:upgrade_same_hex] unless i flip the tense to cannot_reuse_same_hex

tobymao: ok no problem this is fine then

tobymao (approved):

lib/engine/step/tracker.rb
jenf: I'd already done it, which would you prefer?

tobymao:
don't care

PR #3427, opened 2021-01-20
1828: Bug fixes

crericha:


lib/engine/step/g_1828/merger.rb
```diff
@@ -60,7 +60,7 @@ def process_merge(action)
         def process_choose(action)
           raise GameError, 'Invalid action' unless @player_choice
           raise GameError, 'Not your turn' unless action.entity == @players.first
-          raise GameError, 'Invalid choice' unless @player_choice.choices.include?(action.choice)
+          raise GameError, 'Invalid choice' if @player_choice.choices.empty? { |c| c.include?(action.choice) }
```
tobymao, line 63: i don't think empty takes a block

crericha: Good catch. Not a fan of that inconsistency within the Array class. Regardless, fixed now.

tobymao (approved):

PR #3433, opened 2021-01-20
Add possibility to specify a selection of corporations for exchange

perwestling:
Added an attribute 'corporations' that takes an array of corporation names.

If 'corporation' attribute is not specified, the 'corporations' is used instead.

This is a replacement for part of #3393.

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -136,7 +136,7 @@ module View
 
         @game.companies.each do |company|
           @game.abilities(company, :exchange) do |ability|
-            next if ability.corporation != @corporation.name && ability.corporation != 'any'
+            next unless ability.corporations(@game).include?(@corporation)
             next unless company.owner == @current_entity
 
             if ability.from.include?(:ipo)
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -136,7 +136,7 @@ module View
 
         @game.companies.each do |company|
           @game.abilities(company, :exchange) do |ability|
-            next unless ability.corporations(@game).include?(@corporation)
+            next unless @game.exchange_corporations(ability).include?(@corporation)
             next unless company.owner == @current_entity
 
             if ability.from.include?(:ipo)
--- a/lib/engine/ability/exchange.rb
+++ b/lib/engine/ability/exchange.rb
@@ -5,20 +5,11 @@ require_relative 'base'
 module Engine
   module Ability
     class Exchange < Base
-      attr_reader :from
+      attr_reader :from, :corporations
 
       def setup(corporations:, from:)
         @corporations = corporations
         @from = Array(from).map(&:to_sym)
-      end
-
-      def corporations(game)
-        candidates = if @corporations == 'any'
-                       game.corporations
-                     else
-                       @corporations.map { |c| game.corporation_by_id(c) }
-                     end
-        candidates.reject(&:closed?)
       end
     end
   end
--- a/lib/engine/ability/exchange.rb
+++ b/lib/engine/ability/exchange.rb
@@ -5,11 +5,20 @@ require_relative 'base'
 module Engine
   module Ability
     class Exchange < Base
-      attr_reader :corporation, :from
+      attr_reader :from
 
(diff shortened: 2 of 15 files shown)


crericha:
I'm not a fan of having two options and the complexity it brings. Instead,  Is it possible to change corporation to corporations and update the other games to specify an array containing one corporation?

lib/engine/ability/exchange.rb
```diff
@@ -7,9 +7,12 @@ module Ability
     class Exchange < Base
       attr_reader :corporation, :from
 
-      def setup(corporation:, from:)
+      def setup(from:, corporation: nil, corporations: nil)
         @corporation = corporation
+        @corporations = corporations
```
tobymao, line 12: @corporation = corporation || corporations[0]
@corporations = corporations || [@corporation]

lib/engine/step/exchange.rb
```diff
@@ -42,17 +42,23 @@ def can_exchange?(entity, bundle = nil)
         owner = entity.owner
         return can_gain?(owner, bundle, exchange: true) if bundle
 
-        corporations =
-          ability.corporation == 'any' ? @game.corporations : [@game.corporation_by_id(ability.corporation)]
-
         shares = []
-        corporations.each do |corporation|
+        get_corporations(ability).each do |corporation|
```
tobymao, line 46: don't think you need this method, i think you can do

ability.corporation == 'any' ? @game.corporations : ability.corporations.map ...

lib/engine/ability/exchange.rb
perwestling: That would have been better. Have now changed to just using corporations which removed this need.

lib/engine/step/exchange.rb
perwestling: Rubocop was not pleased as it would not fit the tenary into one line (too long).

But instead I put this method into the Exchange ability class. And also added a closed? check which I think is needed. Please check the new attempt.

perwestling:
@crericha @tobymao I have now made a third attempt.... where I just use one attribute, corporations, and take the consequences of this. That did require more changes in the titles, but it does make for a better code overall. Moving handling of "any" to the Exchange ability class I think is good as well.

This passes all existing tests. It should not break any existing games. Might require some manual testing in the different affected titles to ensure it does not break anything.

perwestling:
Check when each title has been manually verified with this change
- [ ] 1828.Games
- [ ] 1836Jr30
- [ ] 1860
- [ ] 1882
- [ ] 1889

1824 and 1835 is still not in pre-alpha so these does not have to be verified at this time.

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -136,7 +136,7 @@ def render_exchanges
 
         @game.companies.each do |company|
           @game.abilities(company, :exchange) do |ability|
-            next if ability.corporation != @corporation.name && ability.corporation != 'any'
+            next unless ability.corporations(@game).include?(@corporation)
```
crericha, line 139: Just a thought, would it make more sense to be @game.exchange_corporations(abillity)? That seems to be the general pattern.

crericha:
> @crericha @tobymao I have now made a third attempt.... where I just use one attribute, corporations, and take the consequences of this. That did require more changes in the titles, but it does make for a better code overall. Moving handling of "any" to the Exchange ability class I think is good as well.
> 
> This passes all existing tests. It should not break any existing games. Might require some manual testing in the different affected titles to ensure it does not break anything.

Nice. Thanks for cleaning this up. Once the PR gets merged, I'll convert 1828 from 'any' to a list of corporations, so we can remove that aspect too. Its probably the only game that would use 'any'.

assets/app/view/game/buy_sell_shares.rb
tobymao: yes it should be game.... please change and we can merge

perwestling:
@crericha @tobymao 
Have now refactored to move the method from Exchange class to game base.
Have also tested exchange during SR for 1836Jr and 1889, and it worked fine.

assets/app/view/game/buy_sell_shares.rb
perwestling: Fixed

tobymao (approved):

PR #3444, opened 2021-01-21
18Mag: rule changes, fix tiles, tweak display of minors

### LINKED ISSUE #3434: [18Mag]  Suggestions for graphics re: first round
I'm finding a lot of flipping between the tabs in the first round:  The Entities Tab to see who own what, the Map tab to see where minors are and the Game tab in order to act.

Would it be possible to put unclaimed minors as grey/shaded markers on the map and then have them go to black once they are claimed?  Also, listing minors in the game tab player cards would help keep track of who has claimed what.  

(On a certain level, I want the minors to be color-coded to players, maybe even only as an option like coordinates, but I also feel like that goes against the design aesthetics, so while it might be helpful, it's probably not a great idea.)

### LINKED ISSUE #3435: 18Mag: rule changes and corrections
- New end of game mechanism: Gray phase is triggered if three consecutive ORs go by without a train purchase.
- Lowest value on stock market should be 55, not 37
- The 237 and 238 tiles should be on opposite sides of each other.

- The minors should appear in numerical order on the entities page

roseundy:
- New rule: advance phase if no trains have been bought in the last three ORs
- Rule change: lowest market value now 55 instead of 37
- Fix 237/238 tiles to be double-sided
- Add minor logos to player cards
- Fix ordering of minors on entity page
- Use city reservations on starting map for minors to make it more obvious which minors are still available for drafting.

Fixes #3434 
Fixes #3435 

Common file edits:
- UI::entities now calls @game method to sort corporations and minors
- UI::player now displays minors (determined by method in @game)
- Game::base - support of above two, plus more generic selection of entities for city reservations

player_card_minor
18Mag_Reservations


assets/app/view/game/player.rb
```diff
@@ -32,6 +32,10 @@ def render
           divs << h(Companies, owner: @player, game: @game, show_hidden: @show_hidden)
         end
 
+        if (minors = @game.player_card_minors(@player)).any?
```
tobymao, line 35: unless empty

lib/engine/game/g_18_mag.rb
```diff
@@ -93,10 +95,10 @@ def setup
         # start with first minor tokens placed (as opposed to just reserved)
         @mine = @minors.find { |m| m.name == 'mine' }
         @minors.delete(@mine)
-        @minors.each do |minor|
-          hex = hex_by_id(minor.coordinates)
-          hex.tile.cities[minor.city || 0].place_token(minor, minor.next_token)
-        end
+        # @minors.each do |minor|
```
tobymao, line 98: delete this if not used

assets/app/view/game/player.rb
roseundy: I keep forgetting to avoid any if possible. Will do.

lib/engine/game/g_18_mag.rb
roseundy: D'Oh!

PR #3445, opened 2021-01-21
add companies inside entity page

carlorusso:


tobymao:
what does this look like

@jenf could you review this?

carlorusso:
> what does this look like
> 
> @jenf could you review this?

On desktop it's like this
image

jenf:
Is this trying to show privates that are owned by the bank?

carlorusso:
> Is this trying to show privates that are owned by the bank?

no, only the privates without any owner

assets/app/view/game/entities.rb
```diff
@@ -47,13 +47,19 @@ def render
           h(Bank, game: @game),
           h(GameInfo, game: @game, layout: 'upcoming_trains'),
           *@game.corporations.select(&:receivership?).map { |c| h(Corporation, corporation: c) },
+          render_unsold_companies,
```
jenf, line 50: Given that 1860 should probably show the unowned company if the bank owns this 'unsold' is incorrect and probably should be generalized.

render_unowned_companies would be better

assets/app/view/game/un_owned_companies.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'view/game/company'
+
+module View
+  module Game
+    class UnOwnedCompanies < Snabberb::Component
```
jenf, line 7: UnownedCompanies

lib/engine/game/base.rb
```diff
@@ -1278,6 +1278,11 @@ def company_sellable(company)
         !company.owner.is_a?(Corporation)
       end
 
+      # Called by View::Game::Entities to determine if the company should be shown on entities
+      def unsold_purchasable_companies(_company)
```
jenf, line 1282: unowned

jenf:
In addition, please see if you can sort the formatting so that it is fixed width ala the player & corporation cards

assets/app/view/game/entities.rb
carlorusso: done

assets/app/view/game/un_owned_companies.rb
carlorusso: done

lib/engine/game/base.rb
carlorusso: done

carlorusso:
> In addition, please see if you can sort the formatting so that it is fixed width ala the player & corporation cards

is ok now?

lib/engine/game/base.rb
```diff
@@ -1278,6 +1278,11 @@ def company_sellable(company)
         !company.owner.is_a?(Corporation)
       end
 
+      # Called by View::Game::Entities to determine if the company should be shown on entities
+      def unowned_purchasable_companies(_company)
```
ryandriskel, line 1282: Current Entity usually isn't a company. Should this be _entity ?

assets/app/view/game/entities.rb
```diff
@@ -47,13 +47,19 @@ def render
...
           *extra_bank,
         ].compact)
 
         children.concat(bankrupt_players.map { |p| h(:div, [h(Player, player: p, game: @game)]) })
 
         h('div#entities', div_props, children)
       end
+
+      def unowned_companies
+        companies = @game.unowned_purchasable_companies(@current_entity)
```
ryandriskel, line 60: Is @current_entity defined in this context?

ryandriskel:
18CO could benefit from some way to show this information as privates not bought during the auction can later be bought from the bank by corporations, but the only place to find out which ones are available is in the hidden "Show Companies of Other Players" area during an OR.

carlorusso:
> 18CO could benefit from some way to show this information as privates not bought during the auction can later be bought from the bank by corporations, but the only place to find out which ones are available is in the hidden "Show Companies of Other Players" area during an OR.

Yes, I’m doing this feature for 18ZOO, but it should be useful for other title too

jenf:
Please update picture if the UI has been changed to fix the width

carlorusso:
> Please update picture if the UI has been changed to fix the width

I'm waiting https://github.com/tobymao/18xx/pull/3494 to update this too

assets/app/view/game/entities.rb
carlorusso: yes, it's a player or a corporation

carlorusso:
@jenf updated the image too; maybe the UI can improved, but I don't know how to do, so if you are ok I should prefer to go with this for now

PR #3446, opened 2021-01-21
18CZ Added Rusting of Trains

Zwergenpunk:
Normal buying of trains only rust the same variant.

Exporting a train rusts all possible variants


tobymao:
i'd rather not override phase instead move rust_trains to a game call in base and override it there

Zwergenpunk:
done

Zwergenpunk:
Closed for new PR

Issue #2207, opened 2020-11-05
1817 - Make it easier to see who owns what


PedroRCSequeira:
both could be optional based on user preferences:

1) show company name (S/ME/A&A) instead of the logos on the map

2) circle logo/name with a color for each players. see image.
each player could go to preferences to activate this, set a color for himself, constant in all games, and a color for 2nd players, 3rd, etc, which would be assigned to the remaining players during setup. It could also be used to highlight each players holdings on the entities/game/sheet tabs, and also in the log so it's easier to see who is making what decisions

![image](https://user-images.githubusercontent.com/30501602/98238435-4631fc80-1f5e-11eb-93cf-b02dafe1f7dd.png)

ps: I wrote it as 1817 specific but it could be useful for other games I guess, this is just the first i'm having a hard time figuring out who owns what since there are so many companies

akh911007:
I had a similar thought and suggested the idea for mouse-over text:  https://github.com/tobymao/18xx/issues/2200

The color border idea is interesting too, since that would be useful on mobile.

PedroRCSequeira:
I actually mentioned the mouse-over option to a friend just a few days ago, and yesterday he showed me your idea, I guess I should write there too, it sounds really cool, and both would help, you can see who owns what and then easily get the info instead of swaping back and forth on the tabs

daniel-sousa-me:
That sounds like a very nice friend x)

jenf:
@tobymao what's your opinion of assigning each player a color and changing the border of the tokens to that corps owner color?

PedroRCSequeira:
One extra thing that would help is, players could pick their colors, and also the color of the player to his left, and the next, etc

So when looking at the map, he always knows the *blue* player is just after him in the SR for instance. The one on his right would change depending on player count, but if they play mostly 4p games then that would also give an extra helpful hint

tobymao:
i guess that would be an enhancement over on the table play. if it's trivial to do, i'm ok with it, but i don't think it's that necessary so i wouldn't spend much time on it

giovandc:
Suggestions no. 2 starts from the assumption that all games a fixed stock round order, and this might not be the case as more titles get implemented.
A keyboard shortcut or on-screen button on the map tab to show/hide companies acronym could be useful

benjaminxscott:
Generically I like the idea of a "legend" for tokens / company names

Perhaps in the legend, it could also show which player is President and be alphasorted by player name? 

[...]

PR #3447, opened 2021-01-21
1817/18Mag: Make it easier to see who owns which corps by highlightin…

jenf:
…g them in different colors

Follow up PR would address following improvements before considering #2207 complete

1) Extend 'track' colors to add up to the 12 players for 1817
(I've checked it doesn't crash at present)
2) Improve text on 'track' colors to refer to players as well as track
3) Add player colors onto Player cards and maybe the entity bar
4) Allow the user to override their own color

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -23,7 +23,7 @@ module View
         needs :reservation, default: nil
         needs :game, default: nil, store: true
         needs :city_render_location, default: nil
-        needs :player_colors, default: nil
+        needs :player_colors, default: nil, store: true
 
         RESERVATION_FONT_SIZE = {
           1 => 22,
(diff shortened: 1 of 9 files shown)


assets/app/view/game/part/city_slot.rb
```diff
@@ -23,6 +23,7 @@ class CitySlot < Base
         needs :reservation, default: nil
         needs :game, default: nil, store: true
         needs :city_render_location, default: nil
+        needs :player_colors, default: nil
```
tobymao, line 26: why don't you just make this a store, then you don't have to pass it down everywhere it's unnneded?

jenf: hopefully a bit simplier

tobymao (approved):

PR #3468, opened 2021-01-22
18CZ - More Implementation

Zwergenpunk:
Added Rusting of trains (fixed from last PR)
city for ATE home token can be chosen
B9 can only be tokened when ATE has home token
Added revenue for E trains  and offboard bonus

### DIFF
--- a/lib/engine/config/game/g_18_cz.rb
+++ b/lib/engine/config/game/g_18_cz.rb
@@ -1559,7 +1559,7 @@ module Engine
                                       "nodes": [
                                           "town"
                                       ],
-                                      "pay": 0,
+                                      "pay": 3,
                                       "visit": 99
                                   }
                               ],
@@ -1616,7 +1616,7 @@ module Engine
                                       "nodes": [
                                           "town"
                                       ],
-                                      "pay": 0,
+                                      "pay": 3,
                                       "visit": 99
                                   }
                               ],
@@ -1666,7 +1666,7 @@ module Engine
                                       "nodes": [
                                           "town"
                                       ],
-                                      "pay": 0,
+                                      "pay": 3,
                                       "visit": 99
                                   }
                               ],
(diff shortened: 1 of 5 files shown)


lib/engine/config/game/g_18_cz.rb
```diff
@@ -1559,7 +1559,7 @@ module G18CZ
                                       "nodes": [
                                           "town"
                                       ],
-                                      "pay": 0,
+                                      "pay": 3,
                                       "visit": 99
```
benjaminxscott, line 1563: Is this right? It seems like this would turn a 3E into a 3+3

(With the ability to skip towns, but it still gets paid for three towns with this config)

Zwergenpunk: This is manually overridden in the code. It runs the the top 3 locations. With this 3 it counts the towns as a stop. 
In the revenue_for all except the top 3 stops are removed

benjaminxscott:
Ran into an issue when testing the train rusting,

lib/engine/config/game/g_18_cz.rb
benjaminxscott: Thanks I totally didn't grok that ha

lib/engine/step/g_18_cz/token.rb
```diff
@@ -0,0 +1,22 @@
...
+module Engine
+  module Step
+    module G18CZ
+      class Token < Token
+        def process_place_token(action)
+          ensure_ate_home_token_is_free(action)
+          super
+        end
+
+        def ensure_ate_home_token_is_free(action)
```
tobymao, line 14: seems like this method is not necessary, can move it in line

and also to clean it up

action.city.tile.cities.flat_map(&:tokens).any?

tobymao:
merge conflict

Zwergenpunk:
Yeah i had the same problem. have started to look at the ebuy but it is missing so much that i ignored it for now. Taking loans as a player etc is alot of nee mechanics

lib/engine/step/g_18_cz/token.rb
Zwergenpunk: Yeah i am a fan of more smaller functions but sure i will move it inline

tobymao (approved):

PR #3477, opened 2021-01-23
1856: Simplified logos

ventusignis:
This is a roundabout.

The 1856 logos came first.
I used them as a template to make the alternate logos for 67/17
Then I compressed and cleaned them.
Now these are getting compressed and cleaned like the ones in 67/17


ventusignis:
See https://github.com/tobymao/18xx/pull/3480

PR #3479, opened 2021-01-23
remove LSL icon appropriately after relevant tiles are upgraded

benjaminxscott:
[Sample game](https://gist.github.com/benjaminxscott/aa25da23db6e496a0d2e99a104f27993)

This clears the `lsl` icon from a given hex after it has been upgraded to green

This is rules-accurate since LSL can be used for a free upgrade (specifically from yellow to green), and LSL goes poof in brown phase (so it can never upgrade from green to brown)

Note that it's safe to clear this icon since LSL doesn't block hexes

Before:
before

After:
after

### DIFF
--- a/lib/engine/step/g_1846/track_and_token.rb
+++ b/lib/engine/step/g_1846/track_and_token.rb
@@ -15,7 +15,7 @@ module Engine
 
         def lay_tile_action(action)
           super
-          return unless @game.lake_shore_line && @game.class::LSL_HEXES.include?(action.hex.id) && @upgraded
+          return unless @game.lake_shore_line && @upgraded && @game.class::LSL_HEXES.include?(action.hex.id)
 
           action.tile.icons.reject! { |icon| icon.name == @game.class::LSL_ICON }
         end
--- a/lib/engine/step/g_1846/track_and_token.rb
+++ b/lib/engine/step/g_1846/track_and_token.rb
@@ -12,6 +12,13 @@ module Engine
         def buying_power(entity)
           @game.track_buying_power(entity)
         end
+
+        def lay_tile_action(action)
+          super
+          return unless @game.lake_shore_line && @game.class::LSL_HEXES.include?(action.hex.id) && @upgraded
+
+          action.tile.icons.reject! { |icon| icon.name == @game.class::LSL_ICON }
+        end
       end
     end
   end
--- a/lib/engine/step/g_1846/track_and_token.rb
+++ b/lib/engine/step/g_1846/track_and_token.rb
@@ -15,7 +15,7 @@ module Engine
 
         def lay_tile_action(action)
           super
-          return unless @game.lake_shore_line && @upgraded && @game.class::LSL_HEXES.include?(action.hex.id)
+          return if !@game.lake_shore_line || !@upgraded || !@game.class::LSL_HEXES.include?(action.hex.id)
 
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_1846/track_and_token.rb
```diff
@@ -12,6 +12,13 @@ class TrackAndToken < TrackAndToken
         def buying_power(entity)
           @game.track_buying_power(entity)
         end
+
+        def lay_tile_action(action)
+          super
+          return unless @game.lake_shore_line && @upgraded && @game.class::LSL_HEXES.include?(action.hex.id)
```
tobymao, line 18: don't use unless with multiple conditions, switch it to an iff

benjaminxscott: I had this: 
```
if @game.lake_shore_line && @upgraded && @game.class::LSL_HEXES.include?(action.hex.id)
            action.tile.icons.reject! { |icon| icon.name == @game.class::LSL_ICON }
end
  ```
  
but [the cops complained about guard clauses](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause)

Screenshot 2021-01-23 at 5 21 20 PM

Comment (ryandriskel) [reply lib/engine/step/g_1846/track_and_token.rb:18]: You can also use multiple returns

Comment (tobymao) [reply lib/engine/step/g_1846/track_and_token.rb:18]: yea, do an early return

Comment (benjaminxscott) [reply lib/engine/step/g_1846/track_and_token.rb:18]: Hmm, I'm not sure what the expected patch is

Comment (tobymao) [reply lib/engine/step/g_1846/track_and_token.rb:18]: return super if !game.lsl || !@upgraded || ...

action.tile.icons.rject

Comment (benjaminxscott) [reply lib/engine/step/g_1846/track_and_token.rb:18]: It looks like [the proposed diff](https://github.com/benjaminxscott/18xx/tree/1846_clear_lsl_test) results in 3 test fails

```
rspec './spec/assets_spec.rb[1:1:40:1]' # Assets #html 1846 22071 renders after_lay 451
rspec './spec/lib/engine/games/game_spec.rb[1:1:1]' # 18 Los Angeles hs_srwgrtvq_1602711223 matches result exactly
rspec './spec/lib/engine/games/game_spec.rb[16:1:1]' # 1846 22071 matches result exactly
```

It doesn't seem to result in the expected outcome, based on [a sample game](https://gist.github.com/benjaminxscott/aa25da23db6e496a0d2e99a104f27993)

Screenshot 2021-01-25 at 1 02 11 PM

tobymao: so fix the logic? if ... but with the right conditions

benjaminxscott: I should clarify, the link above is to a commit where I made that modification
(the `...` was just a shorthand, not being used literally)
https://github.com/benjaminxscott/18xx/commit/7ea16a9a0c639d9d27c6457444e44e0bab30862e

tobymao: yea because the logic is different

it should be

super
return if !x || !y

benjaminxscott: OK I think I get it

tobymao (approved):

PR #3480, opened 2021-01-23
1856 Destinations - Show on map

ventusignis:
This adds a `Destination` part which is intended to function as an Icon that has a corporation's logo. They can be toggled on and off
image
image
image
The destinations for unstarted corporations are removed on the last 4train (new corporations started after that don't need destinations)
This also includes the simplified logos for 1856:
```
This is a roundabout.

The 1856 logos came first.
I used them as a template to make the alternate logos for 67/17
Then I compressed and cleaned them.
Now these are getting compressed and cleaned like the ones in **67/17**
```


ventusignis:
I'm thinking these Destination parts can also be used to represent the Hotels in 18FL

tobymao:
does this button show up in all games? probably don't want it in games with no destinations, also conflict

ventusignis:
I've fixed the conflict and a game needs to override `destinations_toggle?` to be able to get the button to appear

assets/app/view/game/part/icons.rb
```diff
@@ -7,42 +7,88 @@ module View
   module Game
     module Part
       class Icons < Base
+        needs :show_destinations, store: true
```
tobymao, line 10: you should probably have a default value here

assets/app/view/game/tile.rb
```diff
@@ -18,6 +18,7 @@ class Tile < Snabberb::Component
 
       needs :show_coords, default: nil
       needs :show_location_names, default: true
+      needs :show_destinations
```
tobymao, line 21: this isn't needed

lib/engine/g_1856/corporation.rb
```diff
@@ -5,7 +5,7 @@
 module Engine
   module G1856
     class Corporation < Corporation
-      attr_accessor :escrow, :presidents_share
+      attr_accessor :escrow, :presidents_share, :destinations
```
tobymao, line 8: we should be moving stuff away from the 1856 specific corporation,

i think other games have destinations as well like 18AL

lib/engine/game/base.rb
```diff
@@ -905,6 +905,10 @@ def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil
         log_share_price(corporation, price) if self.class::SELL_MOVEMENT != :none
       end
 
+      def destinations_toggle?
```
tobymao, line 908: i think you should be able to figure out if i game has destinations automatically by just doing

def destinations?
  @destinations ||= !@corporations.flat_map(&:destinations).empty?

lib/engine/game/g_1856.rb
```diff
@@ -236,6 +276,27 @@ def setup
         # 1 of each right is reserved w/ the private when it gets bought in. This leaves 2 extra to sell.
         @available_bridge_tokens = 2
         @available_tunnel_tokens = 2
+
+        create_destinations DESTINATIONS
```
tobymao, line 280: always use parens with method calls

lib/engine/game/g_1856.rb
```diff
@@ -76,12 +76,52 @@ class G1856 < Base
         1 => { 3 => 10, 4 => 8, 5 => 7, 6 => 6 },
       }.freeze
 
+      DESTINATIONS = {
```
tobymao, line 79: why isn't this a property on the corporation?

lib/engine/game/g_1856.rb
```diff
@@ -236,6 +276,27 @@ def setup
         # 1 of each right is reserved w/ the private when it gets bought in. This leaves 2 extra to sell.
         @available_bridge_tokens = 2
         @available_tunnel_tokens = 2
+
+        create_destinations DESTINATIONS
+      end
+
+      def create_destinations(destinations)
```
tobymao, line 283: why is this created here? seems like this could be more generic

lib/engine/game/g_1856.rb
```diff
@@ -392,13 +453,30 @@ def grant_right(corporation, type)
...
         # The tokens reserved for the company's buyer are sent to the bank if closed before being bought in
         @available_bridge_tokens += 1 if bridge.owned_by_player?
         @available_tunnel_tokens += 1 if tunnel.owned_by_player?
         super
       end
 
+      def remove_destinations
+        @corporations.select { |corp| corp.capitalization != :escrow && corp != national }.each do |corp|
+          corp.destinations.each do |dest|
+            hex_by_id(dest).tile.icons.reject! { |icon| icon.destination? && icon.name == corp.id }
```
tobymao, line 475: why do we even need a destination? why can't it just be an icon?

lib/engine/part/destination.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Part
+    class Destination < Base
+      # Functionally is an extension of Icon
```
tobymao, line 8: but you don't extend icon?

can we remove destination and just use icon? what does 18AL do?

lib/engine/game/g_1856.rb
ventusignis: I figured this is the clearer way to do it for 1856 to be able to support the alternate destinations optional rule. Making it part of the corporation in the config would, in the case of using alternate destinations, require unsetting the destinations in the map and the corporations before setting new ones. This way I can have the game pick one of the static hash maps based on the rules and load that into the game

assets/app/view/game/tile.rb
ventusignis: Thanks! I was able to remove the needs from here and in hex. I'm pushing up a commit for this shortly

assets/app/view/game/part/icons.rb
ventusignis: I'm pushing up a commit for this shortly

lib/engine/game/g_1856.rb
tobymao: ok, but you can still hive it set the corporation destiantions, i wouldn't make the 1856 version be the defacto standard so

lib/engine/game/g_1856.rb
ventusignis: Thanks, fixed

lib/engine/part/destination.rb
ventusignis: It extends Icon now
The reason behind creating a Destination part instead of just using Icon is so that the Destination part can be used for logic, particularly for 1856 escrow-release logic

tobymao: that would be a reason not to have a destitanion part, having logic in the models makes it less extensible unless the logic is super generic

ventusignis: `Destination`s as they are here are generic - the logic would only live in the game / step code.
For example, 18FL may override `calculate_revenue`to increase the value of towns which a corporation has a hotel (`Destination`) in
1856 will use them to represent the hexes to connect on the map and have them visually toggle-able
1822 and 1870 can have them on the map to indicate destinations for corporations and remove them once the corporation connects and places its station in the city 
18AL can also represent them on the map and could have implemented the HIstorical Objective logic by checking if the hex had the corporation's destination on it in the Token step

This way in the model, a `Destination` is simply an `Icon` that:
 * is associated with a corporation or a minor
 * can be toggled on or off in the map display independently of other Icons
 
 If you're OK with leaving `Destination`s as its own Part, something I want to do next in this PR is:
 Add a method to Game that lets a game decide what they are called (18FL: hotel; 18AL: Historical Objective; default: Destination)
 Add methods in `Game::Base` to add and remove destinations
 Add them to 18AL's config as an example of them being able to be added that way

tobymao: I don't want the destinations to be renamed just like i don't want corporations to be renamed to share companies. Everything should just be called a destination.

tobymao:
Let's remove the destination specific icon for now, I don't think it's currently adding very much. Stick to an icon, you already have a Dictionary mapping each corporation to the destination. Let's also remove the toggle to simplify this PR, I don't think it's really necessary to turn it on and off as you pretty much always want to see them. There are already too many toggles on the map

lib/engine/game/base.rb
ventusignis: button is gone

lib/engine/part/destination.rb
ventusignis: Destination as a Part is gone; it was removed in a single commit so it could be revived if wanted later.

lib/engine/game/g_1856.rb
ventusignis: Destination as a Part has been removed

lib/engine/g_1856/corporation.rb
ventusignis: I removed this from the PR altogether. Once I'm able to put together destination checks what `destinations` looks like will be more clear

lib/engine/game/g_1856.rb
ventusignis: This now exists solely to add destination help text to the corporation charters and to add the icons to the map based on which set of destinations is passed to it. This might also change somewhat later when I work on testing for connectivity, especially in the alternate destination case of the TGB which connects against a 2-hex offboard

tobymao (approved):

PR #3486, opened 2021-01-23
1824: implement more part 2

perwestling:
This PR contains 4 commits:
1. Export a normal train at the end of any OR set
2. Use status_str to show information instead of using base type abilities, plus show type there
3. Support simplified logos for regional and staatsbahn
4. Improve log text for coal/pre-staatsbahn dividend


perwestling:
Standard logos looks like this:
image

Simplified logos:
image

lib/engine/game/g_1824.rb
```diff
@@ -307,12 +327,33 @@ def setup
         end
       end
 
+      def timeline
+        @timeline ||= ['At the end of each OR set, the cheapest train in bank is exported.'].freeze
+        @timeline
```
tobymao, line 332: extra timeline is not necessary

tobymao:
compact your svgs

lib/engine/game/g_1824.rb
perwestling: Fixed

perwestling:
> compact your svgs

Fixed.
Should I compact all the other svgs of 1824 as well?

PR #3489, opened 2021-01-24
Move map controls above the map instead of below

ventusignis:
Why? If the controls are below the map, if you try to zoom in or out multiple times in a row
 the map will resize and will move the buttons with it. If the buttons are above instead,
 when the map resizes the buttons stay in place
 
 There are still some cases where the controls will move but this is more stable than what is currently live and it doesn't trample much in the code

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -133,8 +133,7 @@ module View
             position: 'relative',
           },
         }
-
-        h(:div, [h(:div, props, children), render_controls])
+        h(:div, [render_controls, h(:div, props, children)])
       end
 
       def map_x


ryandriskel:
I haven't loaded it up, but I'm concerned with how far I'm going to scroll up and down on a phone to see the map.

yzemaze:
The controls definitely need to be improved*, but I wouldn’t consider moving them to the top of the map, because it’s really bad on mobile:
image

I’d like a zoom overlay like used on more or less any map service out there.

* fixed button width so they don’t jump when toggled, maybe toggle buttons’ opacity instead of using hide/show, prevent wrap/jump dependent on if show last route is available or not

ventusignis:
Thanks for the feedback @yzemaze and @ryandriskel!

I'm closing this as this is a regression for mobile users

PR #3494, opened 2021-01-24
1824: Improve private rendering during first stock round

perwestling:
Removed BuyCompaniesAtFaceValue and unsold_purchasable_companies
as this was only used by 1824.

By setting owner to @bank for all companies, the render_bank_companies
will render the privates in a better way. There are some minor
differences (button text and location) but they are OK. I suppose it
is good to have consistency between titles.

Hide bought coal companies, and corporations that cannot be bought,
during SR 1.

Added corporations attribute to bank. This is needed when buying
companies that bank own. (Which is what is required to get stock
view to work in 1824.) Now that is some 18xx (e.g. 1825) that allow
companies to be sold to bank so it might be useful for other titles
as well.


perwestling:
This is how the first Stock Round now looks like in 1824. In the example to Coal railways have been bought. As can be seen anything that cannot be bought (e.g. Staatsbahn) is hidden during this SR. 
image

For other SRs coal railways will appear as they might be exchangable.

perwestling:
There is a discussion in a thread if the BuyAtFaceValue code should be kept.

perwestling:
Location of privates is a separate issue so that will be handled in another PR (if we want to change this). 
1822 seems to do a special solution for its bidding boxes.
18Zoo has another PR that will solve this for 18Zoo.
1860 is not affected by this PR.
So I think we can go ahead with this if is approved.

tobymao (approved):

PR #3496, opened 2021-01-24
tile-manifest: full view

yzemaze:
The margins are plain annoying when viewing tiles in a second window.

Alternative:
`@media only screen and (min-width: 900px) {
  #tile_manifest {
    margin: 2vw;
  }
}
`


yzemaze:
… and I totally forgot about the tile selector … darn

PR #3504, opened 2021-01-24
Allow automatic actions to save recompution

jenf:
Here's an alternative approach to #3223

@ventusignis for review as well to see if it'd work.

I've tested server based games, migrations (although there's a flaw if the culprit is the action inside auto actions)

In theory this is all working now


jenf:
The stored data would look like
` {
      "type": "lay_tile",
      "entity": "ICR",
      "entity_type": "corporation",
      "id": 647,
      "created_at": 1611507962,
      "derived": [
        {
          "type": "pass",
          "entity": "ICR",
          "entity_type": "corporation",
          "created_at": 1611507962
        }
      ],
      "hex": "H12",
      "tile": "9-12",
      "rotation": 0
    },`

lib/engine/game/base.rb
```diff
@@ -644,13 +675,11 @@ def process_action(action)
         end
 
         @last_processed_action = action.id
-        self
```
tobymao, line 647: why are these removed?

lib/engine/step/g_1867/token.rb
```diff
@@ -7,6 +7,18 @@ module Engine
...
     module G1867
       class Token < Token
+        def derived_actions(entity)
+          Engine::Action::Pass.new(entity) unless can_really_place_token?(entity)
+        end
+
+        def can_really_place_token?(entity)
+          # Cheaper to do the graph first, then check affordability
+          current_entity == entity &&
+            (tokens = available_tokens(entity)).any? &&
```
tobymao, line 17: !empty

tobymao: unless tokens contains nils

lib/engine/round/base.rb
```diff
@@ -130,6 +130,10 @@ def active_step(entity = nil)
         @active_step ||= @steps.find { |step| step.active? && step.blocking? }
       end
 
+      def derived_actions
```
tobymao, line 133: do we need derived actions? can't we just do @game.add_x_action(action)

def add_x_action(action)
  raise if @loading

lib/engine/game/base.rb
```diff
@@ -597,16 +598,46 @@ def initialize_actions(actions)
...
 
+        # Process the action we came here to do first
+        process_single_action(action)
+
+        # Process any derived actions
+        # derived actions cannot be undos so we don't need to worry about state
+        action.derived.each { |a| process_single_action(a) }
+        any_derived = !action.derived.empty?
+
+        unless @loading_internal
```
tobymao, line 621: is this whole section necessary? 

can add_action(x) just mutate the state of the actions?

and then we just do 

while (action = action.derived_actions.pop)

tobymao: we can rename to auto action

lib/engine/game/base.rb
jenf: they were originally used for the undo functionality, since this isn't the process action it makes no sense

lib/engine/game/base.rb
jenf: not really, if you have chained auto actions this wouldn't work since the step can change

tobymao: i think it should work if it's done in a while looop?

lib/engine/round/base.rb
jenf: as discussed, auto_actions need to be able to be doable on a transition between steps so this is probably the cleanest way of doing it.

lib/engine/game/base.rb
```diff
@@ -704,9 +734,7 @@ def next_turn!
       end
 
       def action_from_h(h)
-        Object
-          .const_get("Engine::Action::#{Action::Base.type(h['type'])}")
-          .from_h(h, self)
+        Engine::Action::Base.action_from_h(h, self)
```
ventusignis, line 737: Should we remove this method and replace its calls with just creating the action where it's used?

jenf: good call

lib/engine/step/g_1867/token.rb
```diff
@@ -7,6 +7,18 @@ module Engine
   module Step
     module G1867
       class Token < Token
+        def auto_action(entity)
+          Engine::Action::Pass.new(entity) unless can_really_place_token?(entity)
```
jenf, line 11: One slight flaw with how this is done is it'll log a pass rather than skipping

A solution would be to store if the Pass was a "skip" and tweak the debug appropriately

lib/engine/game/base.rb
jenf: as mentioned in chat pop mutates the list which is needed outside to actually store the auto_actions.

I'll try and see if there's any tidying up but my initial attempts don't seem to be going anywhere

jenf:
I think this is as close as I can get it right now

lib/engine/round/base.rb
```diff
@@ -130,6 +130,10 @@ def active_step(entity = nil)
         @active_step ||= @steps.find { |step| step.active? && step.blocking? }
       end
 
+      def auto_action
+        active_step(nil)&.auto_action(current_entity)
```
tobymao, line 134: you can just call active_step

lib/engine/step/base.rb
```diff
@@ -32,6 +32,8 @@ def actions(_entity)
         []
       end
 
+      def auto_action(_entity); end
```
tobymao, line 35: should this be named auto_actions

jenf: yep

routes/game.rb
```diff
@@ -83,7 +83,7 @@ class Api
 
                 r.params['user'] = user.id
 
-                engine = engine.process_action(r.params)
+                engine = engine.process_action(r.params, add_autoactions: false)
```
tobymao, line 86: why is this false? isn't this the only place you'd want it to be true? maybe i don't understand what this param is doing

jenf: The way it's designed is that only the client actually adds the autoactions and so the server won't then add them. This would cause inconsistency between the server and the client if we're mid-deploy.
I could allow the server to add them and hope the consistency remains the same, but I'm wary of there being races here

PR #3509, opened 2021-01-24
18FL Initial commit

ventusignis:
There's not too much to 18FL outside of the hotels and initial auction. I included the 5 -> 10 share conversions in this PR because it was pretty simple once it was implemented and working

### DIFF
--- a/assets/app/view/game/convert.rb
+++ b/assets/app/view/game/convert.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+
+module View
+  module Game
+    class Convert < Snabberb::Component
+      include Actionable
+
+      def render
+        click = lambda do
+          process_action(Engine::Action::Convert.new(@game.current_entity))
+        end
+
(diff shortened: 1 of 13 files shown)


assets/app/view/game/convert.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+
+module View
```
tobymao, line 5: why do you need a new view for this? does this mess up 1817?

ventusignis: The conversion happens at the start of a corporations operating turn instead of a separate round like 1817 so the conversion button needs to be hooked into the operating round view.

The new Convert view is nice because conversion is simply a yes/no choice, if it's a choice at all.

I haven't checked 1817 on this branch yet but I will later this evening

ventusignis: image
1817 still works as it did

tobymao (approved):

Issue #1332, opened 2020-08-04
Separate count in spreadsheet for neutral tokens


luigicoolGH:
Please describe your idea and why we should consider it.
For 1882, spreadsheet showing separate counts for regular and neutral tokens.  Either a new column or some other way to see it at a glance.

KenKuhn:
I'll give this one a whirl

KenKuhn:
I think this issue is a bit too advanced for me. I'm releasing it back into the ether ;)

PR #3521, opened 2021-01-25
[1882] add separate count for neutral tokens in spreadsheet

erperreault:
Small addition to spreadsheet.rb to count and display neutral tokens. Added special case for CN to count neutral tokens as its own. Other companies show neutral token as "+N".

Issue https://github.com/tobymao/18xx/issues/1332

## Current behavior (can be misleading, CPR vs others for example):
before

## New behavior:
after

## CN counts neutral tokens as its own:
CNafter

### DIFF
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -363,7 +363,16 @@ module View
           h('td.padded_number', @game.format_currency(corporation.cash)),
           h('td.left', order_props, operating_order_text),
           h(:td, corporation.trains.map(&:name).join(', ')),
-          h(:td, "#{corporation.tokens.map { |t| t.used ? 0 : 1 }.sum}/#{corporation.tokens.size}"),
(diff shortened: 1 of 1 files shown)


assets/app/view/game/spreadsheet.rb
```diff
@@ -363,7 +363,16 @@ def render_corporation(corporation, operating_order, current_round, index)
           h('td.padded_number', @game.format_currency(corporation.cash)),
           h('td.left', order_props, operating_order_text),
           h(:td, corporation.trains.map(&:name).join(', ')),
-          h(:td, "#{corporation.tokens.map { |t| t.used ? 0 : 1 }.sum}/#{corporation.tokens.size}"),
+          if corporation.name == 'CN'
```
jenf, line 366: This would break 1867 since the corporation is also called 'CN'

tobymao:
closing since this is not generic

PR #3522, opened 2021-01-25
1828: Augmented the corporation card for systems

crericha:
Augmented the corporation card by including the logo for each shell and a separate line for the trains in each shell. Below are a few examples. @yzemaze, can you please review?

### Desktop
image
image
image

### Mobile
Portrait orientation
image

Landscape orientation
image

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -124,7 +124,7 @@ module View
       def render_title
         title_row_props = {
           style: {
-            grid: '1fr / auto auto',
+            grid: '1fr / auto auto auto',
             padding: '0.2rem 0.4rem',
             background: @corporation.color,
             color: @corporation.text_color,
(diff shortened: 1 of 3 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -147,14 +147,23 @@ def render_title
...
+                    h('div.title', name_props, @corporation.full_name)]
 
-        h('div.corp__title', title_row_props, [
-          h(:img, logo_props),
-          h('div.title', name_props, @corporation.full_name),
-        ])
+        if @corporation.system?
+          logo_props[:attrs][:src] = logo_for_user(@corporation.corporations.last)
+        else
+          logo_props[:attrs][:src] = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7'
```
tobymao, line 159: what is this????

yzemaze: Seems to be an empty placeholder. Bad style @crericha 
The element should only be there `if @corporation.system?` so corp title gets centered either way. Test against 1889.

assets/app/view/game/corporation.rb
```diff
@@ -124,7 +124,7 @@ def render
       def render_title
         title_row_props = {
           style: {
-            grid: '1fr / auto auto',
+            grid: '1fr / auto auto auto',
```
yzemaze, line 127: just FYI: `repeat(3, auto)` or `grid-auto-flow: column` would do the same.

assets/app/view/game/corporation.rb
```diff
@@ -147,14 +147,23 @@ def render_title
           style: {
             color: 'currentColor',
             display: 'inline-block',
-            justifySelf: 'start',
+            justifySelf: 'center',
```
yzemaze, line 150: already covered in main.css

assets/app/view/game/corporation.rb
crericha: Nice. I used repeat(auto-fit, auto) to only create two columns for corporations and a third for systems.

assets/app/view/game/corporation.rb
crericha: Yes, an empty image. It was the recommended solution on stackoverflow. I wasn't too sure about it myself and one of the reasons I wanted yzemaze's input here. My CSS skills are at the beginner level.  `grid: '1fr / repeat(auto-fit, auto)'` solved my column problem.

I cloned an 1889 game and verified the corporation cards look the same on 18xx.games and my local version.

assets/app/view/game/corporation.rb
```diff
@@ -124,7 +124,8 @@ def render
       def render_title
         title_row_props = {
           style: {
-            grid: '1fr / auto auto',
+            grid: '1fr / repeat(auto-fit, auto)',
+            gridAutoFlow: 'column',
```
yzemaze, line 128: Both lines have the same effect. In CSS later beats earlier, so the first line will be ignored:
image

PR #3533, opened 2021-01-26
1824 mainly: Make Coal Railways minors

perwestling:
The first attempt was to make Coal Railways corporations but they function
in all ways as a minor so this commits makes them minors.

The main reason for having them as corporations was that these had
5 different purchase prices (120G - 200G) so using the parring worked.

But it turned out to cause problems down the line so I now go back
to having them as minors.

Now, it still needs to support buying for different prices. And I have
now added support for having multiple prices by adding an interval
attribute to companies. In case this is set (default nil) the stock
view will render buy buttons similar to par buttons, when buying
private from bank.
In case interval is not specified as attribute this change has no
effect.

This is how the Stock view looks like in interval attribute specified for the selected company:
image

### DIFF
--- a/assets/app/view/game/round/stock.rb
+++ b/assets/app/view/game/round/stock.rb
@@ -305,6 +305,7 @@ module View
         def render_buy_input(company)
           return [] unless @current_actions.include?('buy_company')
           return [] unless @step.can_buy_company?(@current_entity, company)
+          return render_buy_input_interval(company) if company.interval
 
           buy = lambda do
             process_action(Engine::Action::BuyCompany.new(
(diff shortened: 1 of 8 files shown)


perwestling:
Although I believe this might be useful for general cases this is the specific use case:

In 1824 there is no initial auction, but instead privates, minors and corporations are sold during SR1. (Any privates and minors not sold during SR1 is removed from the game.)

There are several different types of minors. One of these are Coal Railways. These can be bought for 120G, 140G, 160G, 180G or 200G, and the price paid will be the starting treasury of this minor.

Now, minors are often represented as a private (compare A-C minors in 18MEX) which when bought gives the buyer control of the minor, while appearing as a company on the player charter. So when a player wants to buy a Coal Railway they actually buy a private that triggers the minor to become owned by the player.

And to make it possible to select the amount to buy the private for, we need some way of making this choice.
After trying a couple of different ways this was the best one. It makes for a quite good GUI which should be somewhat intuitive.

tobymao (approved):

PR #3535, opened 2021-01-26
18FL train buy updates

ventusignis:
3E must hit at least 2 paying revenue centers
One train of a type / OR like 36, except 6/3E are exempt from this
A few bug fixes

### DIFF
--- a/lib/engine/step/g_18_fl/buy_train.rb
+++ b/lib/engine/step/g_18_fl/buy_train.rb
@@ -26,7 +26,7 @@ module Engine
 
           @depot_trains_bought << action.train.sym
 
-          pass! unless buyable_trains(action.entity).any?
+          pass! if buyable_trains(action.entity).empty?
         end
       end
     end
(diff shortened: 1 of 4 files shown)


lib/engine/step/g_18_fl/buy_train.rb
```diff
@@ -0,0 +1,34 @@
...
+        def process_buy_train(action)
+          # Since the train won't be in the depot after being bought store the state now.
+          from_depot = action.train.from_depot?
+          super
+
+          return unless from_depot
+
+          @depot_trains_bought << action.train.sym
+
+          pass! unless buyable_trains(action.entity).any?
```
tobymao, line 29: if empty?

ventusignis: I fixed this in 1836jr30 as well as that's where I copied this step from

tobymao (approved):

PR #3540, opened 2021-01-26
[1822] First pass at the first stockround

magnusfalt:
First pass at the stock round which implementes a new type of auction, bidbox.


assets/app/view/game/player.rb
```diff
@@ -101,6 +101,13 @@ def render_info
               h('td.right', @game.format_currency(@player.cash - committed)),
             ]),
           ]) if committed.positive?
+
+          trs.concat([
+             h(:tr, [
+               h(:td, 'Bidding tokens'),
+               h('td.right', "#{@game.active_step.bidding_tokens(@player)} / #{@game.bidding_token_per_player}"),
```
crericha-qualitrol, line 108: Using round state for bidding_tokens would be better than active_step.

assets/app/view/game/player.rb
```diff
@@ -101,6 +101,13 @@ def render_info
...
+             ]),
+           ]) if @game.active_step&.current_actions&.include?('bidding_tokens')
```
crericha-qualitrol, line 110: Suggest using the bid action and gate on respond_to?(:bidding_tokens)

crericha: Suggest using the bid action and gate on respond_to?(:bidding_tokens)

assets/app/view/game/player.rb
crericha: Using round state for bidding_tokens would be better than active_step.

assets/app/view/game/round/stock.rb
```diff
@@ -57,6 +57,12 @@ def render
 
           children.concat(render_buttons)
           children.concat(render_failed_merge) if @current_actions.include?('failed_merge')
+          children << render_bidbox(@game.bidbox_minors) if @current_actions.include?('bidbox_minors') &&
+            @game.respond_to?(:bidbox_privates)
+          children << render_bidbox(@game.bidbox_concessions) if @current_actions.include?('bidbox_concessions') &&
+            @game.respond_to?(:bidbox_concessions)
+          children << render_bidbox(@game.bidbox_privates) if @current_actions.include?('bidbox_privates') &&
+            @game.respond_to?(:bidbox_privates)
```
crericha, line 65: Why do these need to be actions? Isn't the action always bid?

assets/app/view/game/player.rb
crericha, line 108: Using round state would be better than active_step for managing bidding tokens.

assets/app/view/game/round/stock.rb
```diff
@@ -350,6 +356,68 @@ def render_buy_input_interval(company)
             *buy_buttons,
           ])]
         end
+
+        def render_bidbox(companies)
```
crericha, line 360: I wonder if a special render is needed. Other games include privates in the stock round (1860 and 1824), but for purchasing. It would reduce complexity if we did it the same way here and only added the bid button logic. I suggest taking a look.

assets/app/view/game/company.rb
```diff
@@ -57,6 +57,8 @@ def render
             marginBottom: '0.5rem',
             fontSize: '90%',
           }
+          header_style[:background] = @company.header_color if @company.respond_to?(:header_color)
+          header_style[:color] = @company.header_text_color if @company.respond_to?(:header_text_color)
```
crericha, line 61: Would it make sense to add header_color and header_text_color to company and default to the standard colors? That way we don't have to add any additional respond_to? nonsense, simplifying the code.

lib/engine/g_1822/company.rb
```diff
@@ -0,0 +1,24 @@
...
+      attr_reader :type, :removed, :header_name, :header_color, :header_text_color
+
+      def initialize(game, **opts)
+        @type = opts[:type]&.to_sym || :private
+        @header_name = opts[:header_name] || 'PRIVATE COMPANY'
+        @header_color = game.registered_color(opts[:header_color]) || :yellow
+        @header_text_color = opts[:header_text_color] || :black
+        @removed = false
+
+        super(**opts)
```
crericha, line 17: Recommend calling super first, so that your instance vars don't accidently get overridden.

crericha: I know that's not the case now, but if someone adds one of those types to the base class, it will be an issue.

lib/engine/g_1822/company.rb
```diff
@@ -0,0 +1,24 @@
...
+  module G1822
+    class Company < Company
+      attr_reader :type, :removed, :header_name, :header_color, :header_text_color
+
+      def initialize(game, **opts)
+        @type = opts[:type]&.to_sym || :private
+        @header_name = opts[:header_name] || 'PRIVATE COMPANY'
+        @header_color = game.registered_color(opts[:header_color]) || :yellow
+        @header_text_color = opts[:header_text_color] || :black
+        @removed = false
```
crericha, line 15: Do you need to track removed or simply remove in the game class when the action happens? `@companies.delete(company)`

lib/engine/g_1822/minor.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require_relative '../minor'
+
+module Engine
+  module G1822
+    class Minor < Minor
+      attr_accessor :type, :removed, :share_price, :max_share_price
+      attr_writer :par_price
```
crericha, line 9: 1867 has minors with share prices. Is there a way we can standardize instead of duplicating?

lib/engine/game/g_1822.rb
```diff
@@ -58,12 +78,68 @@ def operating_round(round_num)
...
+        self.class::COLORS[color]
+      end
+
       private
 
+      def setup_companies
+        # Randomize from preset seed to get same order
+        @companies.sort_by! { rand }
+
+        p1 = @companies.find { |c| !c.nil? && c.id == :P1 }
```
crericha, line 130: I think you want 'P1' here. Symbols == Strings in JS land, but not so in Ruby once you hit the server. (Note: Hotseat games are all in JS, so that's why you wouldn't see it yet.)

lib/engine/game/g_1822.rb
```diff
@@ -58,12 +78,68 @@ def operating_round(round_num)
...
+        @companies.delete(p1)
+        @companies.unshift(p1)
+
+        c1 = @companies.find { |c| !c.nil? && c.id == :C1 }
```
crericha, line 134: 'C1'

lib/engine/round/g_1822/stock.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          @game.bidbox_privates.map do |private_company|
+            bid = highest_bid(private_company)
+            buy_company(bid) unless bid.nil?
+          end
+
+          super
+        end
+
+        def setup
```
crericha, line 30: Don't need to override if only calling super.

lib/engine/round/g_1822/stock.rb
```diff
@@ -0,0 +1,76 @@
...
+          @game.companies.delete(company)
+          @log << "#{player.name} buys #{company.name} for #{@game.format_currency(price)}"
+        end
+
+        def float_minor(bid)
+          player = bid.entity
+          company = bid.company
+          price = bid.price
+
+          return unless (minor = @game.minors.find { |m| m.id == company.id })
```
crericha, line 51: You've detected a broken game here. Raise an error to reject the last action rather than silently continue on.

crericha: In general, raise an error when something happens that shouldn't be possible.

lib/engine/round/g_1822/stock.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          player.spend(price, @game.bank)
+
+          minor.owner = player
+          minor.float!
+          @game.bank.spend(100, minor)
+
+          hex = @game.hex_by_id(minor.coordinates)
+          hex.tile.cities[minor.city || 0].place_token(minor, minor.next_token)
```
crericha, line 60: If I recall correctly, token placement is part of first turn housekeeping. It shouldn't block another minor that runs earlier from running through the city.

lib/engine/round/g_1822/stock.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          share_price = @game.stock_market.par_prices.find { |pp| pp.price == 50 }
+          @game.stock_market.set_par(minor, share_price)
+
+          @game.companies.delete(company)
```
crericha, line 65: Why is the company being deleted? Its still in the game.

lib/engine/step/g_1822/bidbox_auction.rb
```diff
@@ -0,0 +1,65 @@
...
+          high_bid = highest_bid(company)
+          (high_bid ? high_bid.price + min_increment : company.min_bid)
+        end
+
+        def max_bid(player, _company)
+          player.cash - committed_cash(player)
+        end
+
+        def remove_from_auction(entity)
+          super
```
crericha, line 23: Don't override methods that you only call super in. There are more below.

lib/engine/step/g_1822/bidbox_auction.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        def active_auction
+          company = @auctioning
+          bids = @bids[company]
+          yield company, bids if bids.size > 1
+        end
+
+        def bids_for_player(player, only_highest_bid_committed = false)
+          @bids.values.map do |bids|
+            if only_highest_bid_committed
```
crericha, line 54: Since this is in 1822 code, maybe rename the argument to `only_committed_bids`. Also, prefer keyword arguments (`only_highest_bid_committed: false`) over defaulting arguments, so that order doesn't matter.

lib/engine/step/g_1822/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,79 @@
...
+        include BidboxAuction
+
+        attr_accessor :bidders, :bid_actions
+
+        def actions(entity)
+          actions = super
+          actions << 'bidbox_minors'
+          actions << 'bidbox_concessions'
+          actions << 'bidbox_privates'
+          actions << 'bidding_tokens'
```
crericha, line 19: Yeah, the extra actions feel unnecessary. Maybe I'm missing something.

assets/app/view/game/round/stock.rb
magnusfalt: Sure i can remove the action, bidbox_.. and just use bid and still check the rspond_to?

lib/engine/step/g_1822/buy_sell_par_shares.rb
crericha: Also, don't include if the player can't lay or increase a bid due to available cash. This is needed to auto-skip that player if they have no actions.

assets/app/view/game/round/stock.rb
magnusfalt: I check how all other have done it, and also talked with Per who implements 1824. I want to seperate the bidboxes in its own row.

lib/engine/step/g_1822/buy_train.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module Engine
+  module Step
+    module G1822
+      class BuyTrain < BuyTrain
+        def actions(entity)
+          return [] if entity != current_entity
+
+          return %w[sell_shares buy_train] if president_may_contribute?(entity)
+          return %w[buy_train pass] if can_buy_train?(entity)
```
crericha, line 13: Is this for minors? Does 1867 do the same?

assets/app/view/game/round/stock.rb
crericha: Either `respond_to?` or something like a `can_bid?(player, entity)` method that is always false in the base class.

assets/app/view/game/company.rb
magnusfalt: Since im new here i dont know how much i can add to the "base" code. Its no problem to add it to the base company.

assets/app/view/game/round/stock.rb
magnusfalt: And per also wanted to seperate the minor bids from the rest of the privates.

lib/engine/g_1822/company.rb
magnusfalt: Sure i can do that. But i have looked and super is really all over the place in the code.

lib/engine/g_1822/company.rb
magnusfalt: It was a beginning of a thought. I do remove it, therefor i can remove the remove attribute.

lib/engine/g_1822/minor.rb
magnusfalt: Same as color, dont know much i can add to the base code.

lib/engine/game/g_1822.rb
magnusfalt: Ahh ok, will change it.

lib/engine/round/g_1822/stock.rb
magnusfalt: Im miles away before im done with that file, just added because its gonna med populated with code soon.

assets/app/view/game/company.rb
crericha: Its definitely a judgement call and should be done carefully as it affects all games. My general rule of thumb is if the functionality would be useful to other games, I add it to the base class. If its only useful to my game, I add it to my specific game classes. In this case, I could see other games taking advantage of it. I'll probably use this for 1828 if it becomes available in the code base.

lib/engine/round/g_1822/stock.rb
magnusfalt: True will fix, shouldnt need to do the return.

lib/engine/g_1822/minor.rb
crericha: Code duplication is bad and should be avoided where possible.

lib/engine/round/g_1822/stock.rb
magnusfalt: Thats true, but as you see in the rest of the method par price and other stuff is hardcoded. It is not finished, just places the token in that part of the code just not to break the game.

lib/engine/round/g_1822/stock.rb
magnusfalt: No the company for the minor is a placeholder to get the bidding to work without to rework the existing bidding mechanics. When the minor floats the company for the minor is removed. In the real game you just put the minors in its biddingboxes.

lib/engine/step/g_1822/bidbox_auction.rb
magnusfalt: Sure will change it.

lib/engine/step/g_1822/buy_sell_par_shares.rb
magnusfalt: As commented above, i can change just to have the bid left.

lib/engine/step/g_1822/buy_train.rb
magnusfalt: Was a long time since i played 1867 so i dont remember. The code is not done, i just took the base function and removed the
`return [] if entity.minor?`
so the game doesnt crash.

lib/engine/round/g_1822/stock.rb
crericha: I add TODO comments for those types of things, so I don't forget. That's just a personal preference though.

lib/engine/step/g_1822/buy_train.rb
crericha: The good news is you may be able to use a lot of Jen's 1867 code for minors. I haven't looked into how she implemented it, so don't know for sure.

lib/engine/step/g_1822/buy_sell_par_shares.rb
magnusfalt: Yes good ide. There are a few more places where autopass can be added.

lib/engine/round/g_1822/stock.rb
magnusfalt: Im not used to add TODO that much. But since its hard for the reviews to know how i think i might add a bit more todo's code that is not finished. But in this state most of the code isnt finished :)

crericha: Sure. If you don't like the TODOs, including comments in the PR can provide context to reviewers too. Whatever works best for you and your workflow. :-)

assets/app/view/game/company.rb
tobymao: remove this stuff

assets/app/view/game/company.rb
```diff
@@ -97,8 +99,9 @@ def render
           end
           props[:style][:display] = @display
 
+          header_name = (@company.respond_to?(:header_name) ? @company.header_name : 'PRIVATE COMPANY')
           children = [
-            h(:div, { style: header_style }, 'PRIVATE COMPANY'),
+            h(:div, { style: header_style }, header_name),
```
tobymao, line 104: remove this

assets/app/view/game/round/stock.rb
tobymao: what is all of this?

assets/app/view/game/round/stock.rb
tobymao: if you can do it the generic way, please remove all of this code

lib/engine/g_1822/company.rb
tobymao: remove this class

assets/app/view/game/company.rb
magnusfalt: Remove all the header_ code, so the minors and consession is the yellow color you mean @tobymao ?

lib/engine/g_1822/minor.rb
tobymao: remove this class

lib/engine/g_1822/company.rb
magnusfalt: Hmm i still need the name for the company. If we gonna remove the coloring name is still needed 'PRIVATE COMPANY' in the view.

lib/engine/g_1822/minor.rb
magnusfalt: Do you mean move the share_price, par_price and max_share_price to base class?

assets/app/view/game/company.rb
tobymao: either make it all generic, or remove it, if you want to keep the PR simple, i'd remove all of this specialized stuff for now

crericha: @tobymao, 1822 needs a way for the player to know who can use privates - majors only, players only, majors and minors. The print version uses the header color to distinguish. Seems like an unobtrusive way to distinguish here too. Just my two cents.

roseundy: Toby has beaten into my head the necessity to keep UI code out of the engine code. Putting a header_color method into company violates that. Better to have a method that indicates who may use a company and have the UI code look at that to decide on header colors.

magnusfalt: Im all for keeping ui code from the engine. But we have color on the major companies, just wanted it on the minors, concession to differ them during the bid. But i will remove it.

lib/engine/g_1822/company.rb
magnusfalt: Ok will remove it.

lib/engine/g_1822/minor.rb
tobymao: you don't need those things in a minor. if you have shares, then it's a corporation not a minor, thus you should remove this 1822 specific minor

assets/app/view/game/company.rb
tobymao: if you need the color, then allow every game to have color and have it be a part of the config, don't have respond_to etc 

happy to chat more on slack if you want to discuss

lib/engine/step/g_1822/buy_train.rb
jenf: 1867 minors aren't actually minors but real corporations with a type set to :minor, :major's are major's

jenf: 1867 minors needed share price and not going down this approach was going to be quite painful.

crericha: @magnusfalt, I recommend going the route @jenf did. It sounds like it'll make for an easier implementation. The Minors class is intended for non-share minors like in 1846.

magnusfalt:
The changes in the last commit:

* Removed the derived class Company, move the color and text_color to the base Company class for all other to enjoy.
* With the removal of Company the type was also removed, therefore in places where i need to sperate the 3 different companytypes i look at the first char in the id (sym) property.
* Removed the derived class Minor, this was needed to get a minor to have a stock value in the stockmarket. Instead the minor will float as a corporation with the type minor.
* Removed the bidbox logic from the stock view. Im using the render_bank_companies instead with a small addition of the 'bid' action and the method render_bid_input.
* The method render_bid_input now uses the Bid component.
* Changed the Bid component to also accept a company, this is done to get the correct property of the return array which calls the process_bid method.

lib/engine/game/g_1822.rb
```diff
@@ -33,19 +33,35 @@ class G1822 < Base
...
+      def entity_can_use_company?(_entity, company)
+        # Setting bidding companies owner to bank, make sure the abilities dont show for theese
+        company.owner != @bank
+      end
+
+      def init_round
+        stock_round
+      end
+
+      def operating_order
```
jenf, line 62: maybe we should separate this and sorted_corporations out into an include since it's the same for 1867 and 1822

lib/engine/step/g_1822/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,76 @@
...
+
+        def actions(entity)
+          actions = super
+          actions << 'bidding_tokens'
+          actions << 'bid'
+          actions
+        end
+
+        def description
+          'Bid, convert consession or buy/sell shares'
```
jenf, line 22: concession

lib/engine/game/g_1822.rb
magnusfalt: That would be a good idea, i found more code that we can share in a include. i e with 1817 i have copied:

`      def format_currency(val)
        # On dividends per share can be a float
        # But don't show decimal points on all
        return format('$%.1<val>f', val: val) if val.is_a?(Float) && (val % 1 != 0)

        self.class::CURRENCY_FORMAT_STR % val
      end`

I can wait to extract them into an include until we see how much more i need from any specific title that isnt already in an include.

assets/app/view/game/bid.rb
```diff
@@ -8,14 +8,16 @@ module Round
       class Bid < Snabberb::Component
         include Actionable
         needs :entity
-        needs :corporation
+        needs :corporation, default: nil
+        needs :company, default: nil
```
crericha, line 12: Generalize corporation and company into one type.

assets/app/view/game/bid.rb
```diff
@@ -29,6 +31,7 @@ def render
             process_action(Engine::Action::Bid.new(
               @entity,
               corporation: @corporation,
+              company: @company,
```
crericha, line 34: Again, generalize here. Under-the-hood, you can use `corporation?` and `company?` to store the type for serialization purposes.

assets/app/view/game/player.rb
crericha: Still would like to see `@round.bidding_tokens` using the `round_state` method. Ping me if you aren't sure what I mean.

assets/app/view/game/round/stock.rb
```diff
@@ -350,6 +354,16 @@ def render_buy_input_interval(company)
             *buy_buttons,
           ])]
         end
+
+        def render_bid_input(company)
+          return [] if @step.max_bid(@current_entity, company) < @step.min_bid(company)
+          return [] if @current_actions.include?('bidding_tokens') &&
+            @step.highest_player_bid?(@current_entity, company)
```
crericha, line 361: Is this true, you can't increase your bid if winning? I understand why you wouldn't want to, but is it against the rules?

assets/app/view/game/round/stock.rb
```diff
@@ -350,6 +354,16 @@ def render_buy_input_interval(company)
...
+          return [] if @current_actions.include?('bidding_tokens') && @step.find_bid(@current_entity, company).nil? &&
+            @step.bidding_tokens(@current_entity).zero?
```
crericha, line 363: The UI shouldn't have to make these logic decisions. Can we replace with `@step.can_bid?(@current_entity, company)`?

lib/engine/game/base.rb
```diff
@@ -203,6 +203,8 @@ class Base
 
       CORPORATIONS = [].freeze
 
+      MINORS = [].freeze
```
crericha, line 206: Is this still needed?

assets/app/view/game/round/stock.rb
magnusfalt: Yes it is

assets/app/view/game/round/stock.rb
magnusfalt: Yep that i can do.

assets/app/view/game/round/stock.rb
magnusfalt: Yes you cant move the cude if you are the highest:

4.9.1 Move a previously placed bidding token to a higher-valued empty space on the same item. This is only allowed if
another player has placed a higher bid.

lib/engine/game/base.rb
magnusfalt: I was getting an warning that is wasnt declared, but i can revert it.

lib/engine/step/g_1822/bidbox_auction.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        protected
+
+        def active_auction
+          company = @auctioning
+          bids = @bids[company]
+          yield company, bids if bids.size > 1
+        end
+
+        def bids_for_player(player, only_committed_bids = false)
```
crericha, line 44: Prefer keyword arguments for optional arugments - `only_committed_bids: false`

assets/app/view/game/bid.rb
magnusfalt: Sure, im have mostly programmed in strong type languages. Not used to use the same variable for different objects.

assets/app/view/game/bid.rb
magnusfalt: Do we have a good name for the combination of corporation and company the same variable. Entity is already used for the player.

crericha: Perhaps change entity to bidder and then use entity?

magnusfalt:
Changed the things pointed out by Jennifer and Chris.

assets/app/view/game/buy_trains.rb
```diff
@@ -103,9 +103,11 @@ def render
         else
           store(:active_shell, nil, skip: true)
         end
+        puts 'Render buy trains 5'
```
crericha: Remove debug in this file

assets/app/view/game/round/stock.rb
```diff
@@ -357,12 +357,9 @@ def render_buy_input_interval(company)
 
         def render_bid_input(company)
           return [] if @step.max_bid(@current_entity, company) < @step.min_bid(company)
-          return [] if @current_actions.include?('bidding_tokens') &&
-            @step.highest_player_bid?(@current_entity, company)
-          return [] if @current_actions.include?('bidding_tokens') && @step.find_bid(@current_entity, company).nil? &&
-            @step.bidding_tokens(@current_entity).zero?
+          return [] if @current_actions.include?('bidding_tokens') && !@step.can_bid?(@current_entity, company)
```
crericha, line 360: Why do we need the bidding_tokens action? Can't we use bid? Isn't the player card the only place in the UI that needs to know about bidding_tokens?

assets/app/view/game/player.rb
crericha, line 110: This is the only other place using the bidding_tokens action. Change to `@game.active_step.respond_to?(:bidding_tokens)`, so we don't need a special action.

assets/app/view/game/round/stock.rb
crericha: Refactor line 359 to use `can_bid?` and 1822 bidding is common with other game bidding within the UI hooray.

magnusfalt: The can_bid? method isnt in the normal Auctioner module which uses the 'bid' action. Any game that uses bid in the stockround will then throw an error when trying to access can_bid?.  bidding_tokens is also used in the player.rb to see how many tokens you have left to bid. Dont know how to show them otherwise?

magnusfalt: if we cant use @step.respond_to?(:can_bid?) then if we gonna use bid, we need to have a dummy method in Auctioner with can_bid? not to break for every other game.

crericha: What I meant by refactor is move line 359 to be a method called `can_bid?` within Auctioner. Keep the logic and game differentiation to the engine code as much as possible. Working on the UI can often include a bunch of refactoring to more generalize the code to include your use cases.

See my other comment on the player card.

magnusfalt: I can remove bidding token on the player and print a log message if this is a better solution.

magnusfalt: Sure you can count your bids, as well. Just wanted the user to see in one place. Committed cash, Available cash and how many tokens he/she have left.

magnusfalt: Ok i see i have made lots of comments without seeing your new ones :)

crericha: Definitely keep the bidding tokens on the player card. Its important info that they shouldn't have to calculate for each player.

assets/app/view/game/round/stock.rb
```diff
@@ -296,15 +296,19 @@ def render_bank_companies
...
+              inputs.concat(render_buy_input(company)) if @current_actions.include?('buy_company')
+              inputs.concat(render_bid_input(company)) if @current_actions.include?('bid')
+              children << h('div.margined_bottom', { style: { width: '20rem' } }, inputs)
+            end
             h(:div, props, children)
           end
         end
 
         def render_buy_input(company)
-          return [] unless @current_actions.include?('buy_company')
```
crericha, line 307: Why did you remove this line?

magnusfalt: I moved it to row 303 to be more cohesive with the bid action i added.

```ruby
inputs.concat(render_buy_input(company)) if @current_actions.include?('buy_company')
inputs.concat(render_bid_input(company)) if @current_actions.include?('bid')
```

lib/engine/step/auctioner.rb
```diff
@@ -77,6 +77,10 @@ def max_bid(_entity, _company)
         raise NotImplementedError
       end
 
+      def can_bid?(_entity, _company)
+        true
```
crericha, line 81: Add `return false if max_bid(entity, company) < min_bid(company)` to maintain the old logic.

Also, make sure to test at least one other game to verify your changes didn't break the bidding UI.

magnusfalt: Hmm there was not old logic its a new function. But yes it makes sense to have that locig there.

crericha: The logic was moved out of the UI in the refactor, but it was not placed here. That's what I'm referring too. We still need to maintain the same behavior.

magnusfalt: I really dont follow, the render_bid_input method is new to this PR. Therefore the logic in it can not be old.

magnusfalt: But i agree the default behaviour should be the logic of `max_bid(entity, company) < min_bid(company)`

crericha: My apologies. I confused the file with ` assets/app/view/game/bid.rb`. Thanks for clarifying. 

Since 1822 is the only current game to access bid from the stock round, you can either remove `can_bid?` from the base auctioner or add in the basic logic outlined above. Whichever way you prefer.

magnusfalt: There are more games to use bid in the stockround, just not the way i do it. In method `render_pre_ipo` there is a line 

`children << h(Bid, entity: @current_entity, corporation: corporation) if @current_actions.include?('bid')`

Im guessing its from 1817.

magnusfalt: Since in the base auctioneer the methods `min_bid` and `max_bid` do a `raise NotImplementedError`. Shall i do the same for `can_bid?`

crericha (approved):

lib/engine/game/g_1822.rb
```diff
@@ -62,13 +78,94 @@ def operating_round(round_num)
...
+        Round::G1822::Stock.new(self, [
+          Step::DiscardTrain,
+          Step::Exchange,
+          Step::SpecialTrack,
+          Step::G1822::BuySellParShares,
+        ])
+      end
+
+      def bidbox_minors
+        @companies.select { |c| c.id[0] == 'M' && (c.owner.nil? || c.owner == @bank) }.map.first(4)
```
tobymao, line 101: what's this map here for?

don't call owner.nil?

&& !c.owner || c.owner == @bank

lib/engine/game/g_1822.rb
```diff
@@ -62,13 +78,94 @@ def operating_round(round_num)
...
+        @companies.select { |c| c.id[0] == 'P' && (c.owner.nil? || c.owner == @bank) }.map.first(3)
+      end
+
+      def init_bidding_token
+        self.class::BIDDING_TOKENS[@players.size.to_s]
+      end
+
+      def setup_bidboxes
+        # Set the owner to bank for the companies up for auction this stockround
+        bidbox_minors.map do |minor_company|
```
tobymao, line 118: call this a minor

lib/engine/game/g_1822.rb
```diff
@@ -62,13 +78,94 @@ def operating_round(round_num)
...
+          minor_company.owner = @bank
+        end
+
+        bidbox_concessions.map do |concessions|
+          concessions.owner = @bank
+        end
+
+        bidbox_privates.map do |private_company|
```
tobymao, line 126: call this a company

lib/engine/round/g_1822/stock.rb
```diff
@@ -0,0 +1,109 @@
...
+
+        def setup
+          super
+          @game.setup_bidboxes
+        end
+
+        def finish_round
+          @game.bidbox_minors.map do |minor_company|
+            bid = highest_bid(minor_company)
+            if bid.nil?
```
tobymao, line 19: if (bid = highest_bid...)
else

lib/engine/round/g_1822/stock.rb
```diff
@@ -0,0 +1,109 @@
...
+              minor_company.owner = nil
+            else
+              float_minor(bid)
+            end
+          end
+
+          @game.bidbox_concessions.map do |concessions|
+            bid = highest_bid(concessions)
```
tobymao, line 27: same

lib/engine/round/g_1822/stock.rb
tobymao: why are you using map? you should be using each

lib/engine/step/g_1822/bidbox_auction.rb
```diff
@@ -0,0 +1,57 @@
...
+        def committed_cash(player, _show_hidden = false)
+          bids_for_player(player, only_committed_bids: true).sum(&:price)
+        end
+
+        def find_bid(player, company)
+          @bids[company]&.find { |b| b.entity == player }
+        end
+
+        def highest_player_bid?(player, company)
+          return false if find_bid(player, company).nil?
```
tobymao, line 31: return false unless ...

lib/engine/step/g_1822/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,82 @@
...
+          action.entity.unpass!
+          add_bid(action)
+        end
+
+        def setup
+          setup_auction
+          super
+
+          @bid_actions = 0
+          @bidders = (@round.bidders.nil? ? Hash.new { |h, k| h[k] = [] } : @round.bidders)
```
tobymao, line 44: no nil

lib/engine/step/g_1822/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,82 @@
...
+          @bids = @round.bids unless @round.bids.nil?
```
tobymao, line 45: same

lib/engine/step/g_1822/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+
+        def bidding_tokens(player)
+          @game.bidding_token_per_player - (bids_for_player(player)&.size || 0)
+        end
+
+        def can_bid?(entity, company)
+          return false if max_bid(entity, company) < min_bid(company) || highest_player_bid?(entity, company)
+
+          !(find_bid(entity, company).nil? && bidding_tokens(entity).zero?)
```
tobymao, line 55: nil

lib/engine/game/g_1822.rb
magnusfalt: Just a lack of knowledge of ruby, i probably saw code something like it and did it about the same way.

magnusfalt:
Improved the code with Tobys suggestions

lib/engine/game/g_1822.rb
```diff
@@ -62,13 +78,94 @@ def operating_round(round_num)
...
+        @companies.select { |c| c.id[0] == 'P' && (!c.owner || c.owner == @bank) }.first(3)
+      end
+
+      def init_bidding_token
+        self.class::BIDDING_TOKENS[@players.size.to_s]
+      end
+
+      def setup_bidboxes
+        # Set the owner to bank for the companies up for auction this stockround
+        bidbox_minors.map do |minor|
```
tobymao, line 118: you didn't fix these

lib/engine/game/g_1822.rb
```diff
@@ -62,13 +78,94 @@ def operating_round(round_num)
...
+          minor.owner = @bank
+        end
+
+        bidbox_concessions.map do |concessions|
+          concessions.owner = @bank
+        end
+
+        bidbox_privates.map do |private|
```
tobymao, line 126: don't call it a private it's a reserved keyword

PR #3542, opened 2021-01-26
1817: Add increased loans variant

daniel-sousa-me:
This is not included in the rulebook, but it's a variant available in rr18xx. Some players on the 1817NA tournament Discord were saying they prefer playing with it.


tobymao:
closing for now as we haven't been adding non official variants

PR #3543, opened 2021-01-26
add neutral tokens to spreadsheet (for 1882)

erperreault:
Issue https://github.com/tobymao/18xx/issues/1332

** Making this generic. 

---

Add spreadsheet functionality to count and display neutral tokens.

## Current behavior (can be misleading, CPR vs others for example):
before

## New behavior:
after

## CN counts neutral tokens as its own:
CNafter


erperreault:
Maybe it would look better as: 

1/1, N

jenf:
> Maybe it would look better as:
> 
> 1/1, N

I'd still suggest a separate column as this avoids confusion.

ryandriskel:
I'd like to request the column not exist if the game doesn't support neutral tokens.

scottredracecar:
I find the +N notation with no extra column to be easy to understand.

erperreault:
Set it up as a separate column, here's how it looks:

column

jenf:
Now that I've seen it, I think the previous implementation was probably the better looking. Apologies.

We should add some sort of key to say N=Neutral Token.

erperreault:
- Moved back to +N in the Tokens column.
- Added note below tables.

Questions:  Do we need the note, or is the N clear enough? What about "2/2, N" vs "2/2+N"?

Picture showing both:
note

ventusignis:
At the risk of bikeshedding a possible way to do this is in a fully generic fashion to make a `token_string(entity)` method in game and put the"count + N" logic into 1882 into an implementation of this. Then on top of that make a `def spreadsheet_legend; []; end` method in base and override that in 1882 and use that to build the legend in the view

1822 could later use this fully generic version to display the home, destination, exchange, and spare tokens for the corporations

scottredracecar:
Yes, making it universal sounds like a great idea, but implementing this now and then changing it to universal later when it is needed for 1822 also seems reasonable.

lib/engine/game/g_1882.rb
```diff
@@ -204,6 +204,10 @@ def action_processed(action)
         @corporations.reject! { |c| c.id == 'SC' }
         @sc_company = nil
       end
+
+      def neutral_tokens(corporation)
```
jenf, line 208: Sorry to keep changing this on you but ventusignis suggestion of a custom method in the game class seemed a good one.

Probably worth having a token_note method that allows a string to be added for 1882 and in future other games.

jenf:
Change looks good, remove draft and I'll approve

tobymao (approved):

PR #3549, opened 2021-01-26
1860: refactor step/tracker to use common code; optimize halt usage

### LINKED ISSUE #3208: 1860 default for halts
https://18xx.games/game/22068?action=240

The FYN runs for 90 revenue plus up to 2 halts. The system defaults to L90 + L10. It should most definitely default to L90 + L20. There's almost no circumstance where it would make sense to run for L90 + L0 or L90 + L10 in this situation.

Also, the interface doesn't entirely make clear what is going on here. The L10 button and the +L10 indicator in the Submit button don't show that those values will be paid to treasury. It would be nice if there was more information to show that this money is headed to the treasury and won't be part of the dividend.

Screen Shot 2021-01-10 at 2 17 21 PM

roseundy:
- Refactored Step::G1860::Tracker to use more of base class
- Step::Tracker - added "save_old" method to allow above
- Engine::Route - clear route.halts on touch_node. This allows calculate_stops logic to optimize revenue and halts when building a route

Fixes #3208

### DIFF
--- a/lib/engine/step/g_1860/tracker.rb
+++ b/lib/engine/step/g_1860/tracker.rb
@@ -10,7 +10,6 @@ module Engine
 
         def setup
           @laid_city = false
-          @tr_distance = nil
           @saved_revenues = []
           super
         end
@@ -36,6 +35,7 @@ module Engine
             raise GameError, "#{action.hex.id} cannot be layed as this hex was already layed on this turn"
           end
 
+          @saved_revenues = revenues(action.hex.tile, action.entity)
           lay_tile(action, extra_cost: tile_lay[:cost])
           @upgraded = true if action.tile.color != :yellow
           @laid_city = true if action.tile.cities.any?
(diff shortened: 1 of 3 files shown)


lib/engine/step/g_1860/tracker.rb
```diff
@@ -30,149 +32,51 @@ def lay_tile_action(action)
...
-                " for the #{ability.terrain} tile built by #{company.name}"
-            end
-          end
+          @tr_distance = @game.biggest_train_distance(entity)
+          @saved_revenues = if old_tile.color == :white
+                              []
+                            else
+                              old_tile.nodes.select { |n| reachable_node?(entity, n, @tr_distance) }
+                                .map(&:max_revenue).sort
+                            end
```
ryandriskel, line 55: It might read easier if you split this out to a function

```
@saved_revenues = old_revenues(entity, old_tile)

function old_revenues(entity, old_tile)
  return [] if old_tile.color == :white

  old_tile.nodes.select { |n| reachable_node?(entity, n, @tr_distance) }
                          .map(&:max_revenue).sort
end
```

I didn't make any change here, but you might consider combining the map and select.

roseundy: I can break it out into a function, but I'll have to not do the @tr_distance calculation or save.

What do you mean "combining the map and select"? Aren't they already combined?

roseundy: I'm changing my approach a little bit. I found a way to not modify the base Tracker at all.

tobymao (approved):

PR #3553, opened 2021-01-26
Refactor handle_partial

### LINKED ISSUE #3546: 1849 - Double share causing 110%
In this game https://18xx.games/game/22466 the AFG has 110% of itself in existence. When the presidency changed one player was given the double share from the market instead of a single share.

tysen:
Fixes #3546

Some duplicated mode moved into a function and then overridden for 1849
to avoid giving the outgoing president the double cert from the market.

### DIFF
--- a/lib/engine/share_pool.rb
+++ b/lib/engine/share_pool.rb
@@ -198,9 +198,7 @@ module Engine
         @log << "President's share sold to pool. #{corporation.name} enters receivership"
         return unless bundle.partial?
 
-        difference = bundle.shares.sum(&:percent) - bundle.percent
-        num_shares = difference / corporation.share_percent
-        num_shares.times { move_share(shares_of(corporation).first, owner) }
+        handle_partial(bundle, self, owner)
         return
       end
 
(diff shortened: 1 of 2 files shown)


lib/engine/share_pool.rb
```diff
@@ -251,9 +247,14 @@ def transfer_shares(bundle, to_entity,
 
       return unless bundle.partial?
 
+      handle_partial(bundle, self, owner)
```
ryandriskel, line 250: Several of the other places where this is added have a "return" afterwards. Are they redundant or should this have a return?

tysen: Those other `returns` are control flow to break the execution of the function. This instance, you'll notice, is at the very end of the function, which already has an implicit `return`.

ryandriskel: Silly me, I saw the "end" and thought it was the close of a function, not a conditional, as the preceding lines were not visible..

tobymao (approved):

Issue #3550, opened 2021-01-26
[18Mag] unable to upgrade oo city tile from yellow to green with Major Token in place


Xurgon:
Please fill out the bug report title with the game name and description

e.g. "[18Chesapeake] Incorrect revenue for route from Baltimore to DC"

**What's your Game ID?**

https://18xx.games/game/24044

If your game is "Multiplayer", provide the link to the game. Copy the link from your browser which will look something like `https://18xx.games/game/31337`

- (optional) Provide a link to the specific action that triggered the bug using the `Game` tab and `History` navigation arrows. Copy the resulting link from your browser which will look something like `https://18xx.games/game/31337?action=42`

If your game is "Hotseat", provide a link to your raw game data by using the `Tools` tab and `Copy Game Data`.

Paste that data [into a github gist](https://gist.github.com/) and share the resulting link which will look something like `https://gist.github.com/[your_name]/[some_letters]`.

**How to Reproduce the Bug?**

What steps cause the unexpected behavior?

Unable to upgrade city tile located at G10.
Gives error - no Implicit conversion of NilClass into Integer
Tile is currently yellow with a major token in the non connected token spot

Please include screenshots (you can drag and drop or upload them into this issue)

**Which Browser / Platform?**
- Browser Version
  - On Chrome, see [chrome://version](chrome://version)
  - On Firefox, see [about:support](about:support)
  - On Safari, see [these methods](https://www.lifewire.com/check-the-version-number-of-safari-446563)
- Platform (Mobile, Tablet, Desktop)

PR #3558, opened 2021-01-27
Fix Hex.lay to map cities more robustly

roseundy:
Modify the logic in Hex.lay(tile) to map cities without exits more robustly.

Fixes: #3550 

The current logic fails when mapping a two-city tile with one city having track and the other not, to an upgraded tile:

18mag_OO_upgrade

### DIFF
--- a/lib/engine/hex.rb
+++ b/lib/engine/hex.rb
@@ -128,15 +128,8 @@ module Engine
       @tile.cities.each_with_index do |old_city, index|
         next if city_map[old_city]
 
-        new_city = if !new_cities.include?(tile.cities[index])
-                     tile.cities[index]
-                   elsif !new_cities.include?(tile.cities[0])
-                     tile.cities[0]
-                   else
-                     tile.cities.find { |city| !new_cities.include?(city) }
-                   end
-        # failsafe
-        new_city ||= tile.cities[index] || tile.cities[0]
+        new_city = tile.cities[index]
+        new_city = tile.cities.find { |city| !new_cities.include?(city) } if new_cities.include?(new_city)
 
         city_map[old_city] = new_city
         new_cities << new_city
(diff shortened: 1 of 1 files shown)


lib/engine/hex.rb
```diff
@@ -110,20 +110,38 @@ def lay(tile)
           @tile.cities.zip(tile.cities).to_h
         # if @tile is not blank, ensure connectivity is maintained
         else
-          @tile.cities.map.with_index do |old_city, index|
+          @tile.cities.map do |old_city|
             new_city = tile.cities.find do |city|
               # we want old_edges to be subset of new_edges
               # without the any? check, first city will always match
               old_city.exits.any? && (old_city.exits - city.exits).empty?
```
tobymao, line 117: change any to empty here

roseundy: OK. Looks reasonable otherwise?

lib/engine/hex.rb
```diff
@@ -110,20 +110,38 @@ def lay(tile)
...
+      @tile.cities.each_with_index do |old_city, index|
+        next if city_map[old_city]
+
+        new_city = if !new_cities.include?(tile.cities[index])
+                     tile.cities[index]
+                   elsif !new_cities.include?(tile.cities[0])
+                     tile.cities[0]
+                   else
+                     tile.cities.find { |city| !new_cities.include?(city) }
+                   end
```
tobymao, line 137: the code smells to me, why would the fail safe ever hit?
new_city = tile.cities[index]
new_city = tile.cities.find { |city| !new_cities.include?(city) if new_cities.include?(new_city)

roseundy: I think the only time it could hit is when we map from N cities on the old tile to less then N cities on the new tile.

Since I don't know the reason why the original code had
new_city ||= (tile.cities[index] || tile.cities[0])
I was hesitant to change it.

tobymao (approved):

PR #3562, opened 2021-01-27
18CO - Show bank-owned companies (for later purchase)

ryandriskel:
Taking advantage of the work done in https://github.com/tobymao/18xx/pull/3445

In 18CO, a private does not need to be purchased in the initial auction. It remains in the bank and can be purchased at a later time by a corporation. Prior, there was no way to see which privates had not been purchased except for the log.

Screen Shot 2021-01-27 at 7 43 37 AM

### DIFF
--- a/lib/engine/game/g_18_co.rb
+++ b/lib/engine/game/g_18_co.rb
@@ -622,6 +622,10 @@ module Engine
         end
       end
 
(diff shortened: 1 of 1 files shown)


jenf (approved):

tobymao (approved):
don't need the nil can just do !owner

PR #3567, opened 2021-01-27
Add push notifications on the browser with service workers

daniel-sousa-me:
You will need to add the VAPID keys to the server environment config.

To generate VAPID keys in Ruby, you just need to run:
```
require 'webpush'
Webpush.generate_key()
```

I've put some default keys for the dev environment, because it seemed harmless and more pratical.

Each browser uses their own push service and informs us with a unique endpoint URL, which is stored in user settings. Push notifications are sent more or less anytime an email notification is sent (if they're both enabled).

This is how the settings look like:
image

image

image

We don't have much control over the notifications themselves and the look changes wildly across OSs and browsers.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem 'tilt'
 gem 'uglifier'
 gem 'unicorn'
 gem 'unicorn-worker-killer'
+gem 'webpush'
 
 group :development do
   gem 'pry-byebug'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -15,7 +15,10 @@ GEM
       rake
     get_process_mem (0.2.7)
       ffi (~> 1.0)
+    hkdf (0.3.0)
+    jwt (2.2.2)
     kgio (2.11.3)
+    libv8 (8.4.255.0)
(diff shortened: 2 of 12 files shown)


Zwergenpunk:
Nice! :)

tobymao:
this looks a bit invasive, i think i'd rather go with regular notifications api through message bus

Zwergenpunk:
Regular notifications don't work really on mobile phones. This is like an app that pushes that it is your turn. Even when the browser is not openend etc.

tobymao:
I understand that, didn't realize it was possible. Not sure how I feel about it. If people feel strongly about getting this in, it needs work to be simplified

daniel-sousa-me:
> this looks a bit invasive, i think i'd rather go with regular notifications api through message bus

They are complementary. The push notifications will not fire if there is a connection (like email notifications)

Having some kind of notifications is a highly requested feature. That's the reason why people keep making Discord and Telegram bots, or Slack and WhatsApp groups.

Maybe the text could be more clear in stating that these are background notifications.

What do you want to be simplified?

jenf:
For myself personally I'd more likely use this than web notifications.

PR #3582, opened 2021-01-28
1860: update route.halts even if no halts are desired

### LINKED ISSUE #3581: 1860 unable to choose halts
https://18xx.games/game/23350?action=259

The 4+2 train should have the option to choose to use halts instead of towns. The only optional rule enabled in the game is the 2/3 player map.

Screen Shot 2021-01-28 at 12 29 39 PM

roseundy:
Previous change to halts meant that selecting a route that had enough towns to fill the train allowance prevented ever selecting non-zero halts.

With this change, route.halts will always be written if there are halts are visited on the route, except for loaner trains and when the southern forms (after sr).

Fixes #3581

No pins should be required because it just affects route selection via UI.

### DIFF
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -1095,7 +1095,7 @@ module Engine
         end
 
         # update route halts
-        route.halts = num_halts if (num_halts.positive? || route.halts) && !loaner_new_rules?(route)
+        route.halts = num_halts if (halts.any? || route.halts) && !loaner_new_rules?(route) && !ignore_halts?
 
         stops
       end


lib/engine/game/g_1860.rb
```diff
@@ -1095,7 +1095,7 @@ def compute_stops(route)
         end
 
         # update route halts
-        route.halts = num_halts if (num_halts.positive? || route.halts) && !loaner_new_rules?(route)
+        route.halts = num_halts if (halts.any? || route.halts) && !loaner_new_rules?(route) && !ignore_halts?
```
tobymao, line 1098: empty?

Issue #3374, opened 2021-01-17
1860: Bank Broke after Southern Formed, perhaps misleading


PaulKT:
https://18xx.games/game/20624

I get that in 1860 Southern can form and British Rail not form thus end game via nationalisation may not actually happen. Even if that end game trigger does form the market can still end the game straight out. But if BR forms the bank is unlimited. However, Southern can form and the bank break as per our game. I just wanted to note that it may be confusing.

[20:57]Southern Railway Forms; End of game triggered (via Nationalization). I think it should say "End Game possible via Nationalisation" maybe?

Then when BR forms it should then be "End of game triggered (via Nationalization)."

As some of our players thought "[20:57]Southern Railway Forms; End of game triggered (via Nationalization)." mean the bank was unlimited.

The game was fine, just maybe a bit misleading?

roseundy:
This should have been linked to the PR with the change.

PR #3585, opened 2021-01-29
1860: event_southern_forms text change

ryanespin:
Addresses #3374 

Changed text for SR endgame trigger to be less ambiguous.

### DIFF
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -357,7 +357,7 @@ module Engine
       end
 
       def event_southern_forms!
-        @log << 'Southern Railway Forms; End of game triggered (via Nationalization).'
+        @log << 'Southern Railway Forms; End Game possible via Nationalisation.'
         @southern_formed = true
       end
 
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -357,7 +357,7 @@ module Engine
       end
 
       def event_southern_forms!
-        @log << 'Southern Railway Forms; End Game possible via Nationalisation.'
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1860.rb
```diff
@@ -357,7 +357,7 @@ def event_relax_cert_limit!
       end
 
       def event_southern_forms!
-        @log << 'Southern Railway Forms; End of game triggered (via Nationalization).'
+        @log << 'Southern Railway Forms; End Game possible via Nationalisation.'
         @southern_formed = true
```
roseundy, line 361: I disagree somewhat - the game is going to end soon. If you want to warn that the game can still end via the bank breaking, you should put that in.

roseundy: I can live with "Southern Railway Form; Nationalization will be triggered when all players’ companies have at least one train"

ryanespin: Sure, so something like "Southern Railway Forms; End of game still possible via bank breaking".

If we want to include all end game condtions, maybe something like this: "Southern Railway Forms; End of game possible via Nationalisation, Bank Breaking, or Company hitting max stock value."

And lastly, should I go with nationalisation or nationalization?  The rulebook indicates nationalisation which is the UK spelling, but wasn't sure which to go with for the wording.

ryanespin: > I can live with "Southern Railway Forms; Nationalization will be triggered when all players’ companies have at least one train"

Updated! :)

tobymao (approved):

PR #3590, opened 2021-01-29
1860: add share price movement chart to route_selector

### LINKED ISSUE #3213: 1860 multi bump payouts suggestions
At this screen, it would be nice if there was a tip box (perhaps to the right of the Select Routes tool, perhaps on the corporation charter) that indicated the amount needed to single/double/triple/quad jump on the stock market.

https://18xx.games/game/22068?action=469

Screen Shot 2021-01-10 at 3 38 49 PM

roseundy:
Add a chart to UI::route_selector to give a player the same info they'd see playing a F2F game by looking at the game board:

1860_chart

Also, tweaked dividend table for 18Mag in UI::dividend.

Fixes #3213

### DIFF
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -171,7 +171,7 @@ module View
 
         props = {
           style: {
-            border: '1px solid black',
+            border: '1px solid',
           },
         }
 
@@ -186,7 +186,8 @@ module View
           style: {
             margin: '0.5rem 0 0.5rem 0',
             textAlign: 'left',
-            border: '1px solid black',
+            border: '1px solid',
+            borderColor: color_for(:font),
             borderCollapse: 'collapse',
           },
         }
@@ -194,7 +195,7 @@ module View
         header_props = {
           style: {
             backgroundColor: color_for(:bg2),
-            border: '1px solid black',
+            border: '1px solid',
           },
         }
 
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 
+require 'lib/color'
+require 'lib/settings'
 require 'view/game/actionable'
 
 module View
   module Game
     class Dividend < Snabberb::Component
       include Actionable
+      include Lib::Color
+      include Lib::Settings
 
       needs :routes, store: true, default: []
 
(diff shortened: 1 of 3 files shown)


assets/app/view/game/dividend.rb
```diff
@@ -180,18 +184,25 @@ def dividend_chart
...
             textAlign: 'left',
             border: '1px solid black',
             borderCollapse: 'collapse',
           },
         }
 
+        header_props = {
+          style: {
+            backgroundColor: color_for(:bg2),
+            border: '1px solid black',
```
tobymao, line 197: should this be a user preference collor for the dark mode users?

assets/app/view/game/route_selector.rb
```diff
@@ -270,6 +271,52 @@ def actions(render_halts)
           h(:button, { style: submit_style, on: { click: submit } }, 'Submit ' + revenue + subsidy),
         ])
       end
+
+      def dividend_chart
+        step = @game.active_step
+        return nil unless step.respond_to?(:chart)
```
tobymao, line 277: if you return nil here does the method calling this call compact?

assets/app/view/game/route_selector.rb
```diff
@@ -270,6 +271,52 @@ def actions(render_halts)
           h(:button, { style: submit_style, on: { click: submit } }, 'Submit ' + revenue + subsidy),
         ])
       end
+
+      def dividend_chart
```
tobymao, line 275: i thought had made one of these before or am i misremembering

assets/app/view/game/dividend.rb
roseundy: Dunno. Is there an example I can look at?

assets/app/view/game/route_selector.rb
roseundy: Yes. It works fine if step.chart is not implemented.

assets/app/view/game/route_selector.rb
roseundy: I have something very similar I added to the dividend step for 18Mag.

assets/app/view/game/dividend.rb
roseundy: For what it's worth, it's no different than how the shading is done at the top and bottoms of the corporate cards.

roseundy: Isn't that what color_for() does?

tobymao:
what about the border

assets/app/view/game/dividend.rb
roseundy: Got it. Removed hard-coded 'black'

tobymao (approved):

PR #3591, opened 2021-01-29
[1822] Half dividend and local trains

magnusfalt:
**The problem:**
The rules states the following description of a Local train.
5.5.3 L (local) trains run in a city which has a station token of the owning (minor or major) company. They can additionally run to a single small station, but are not required to do so. They can thus be considered 1 (+1) trains. Only one L train may operate on each station token.

**The solution:**
I really tried to see as many solutions as possible to touch the base code as little as possible. But In the end i had to include some, there was no support for a route when pressing the first city or method in base game class  i could override to get it to work.

First there is a new method the train, local?. Here i try to do it in the most generic way. I look at the distance of the train, is it 1 then its a local. If distance is an array, the node city with a visit of 1 most be included to be a local train.

In the route we start with the touch_node method where the ui calls when a node have been pressed. Since there is no connection object on the hex i can look when you press a city i create a new one when you press the first city, with the same node on left and right but with no path. The code then handle this well, all the checks with home token, revenue and so on works as it is.

To get the last_route to work i first modify the connection_hexes method a bit. Since there are no path in the connection_data object no data was returned from the id method on in the connection object. Here i opted to save the local connection info with a format of ['local', <selectedhex>.id]. The action entry now looks the same for the local as it does for any normal path object, with the exception for the 'local'.

The 'local' is need in the connection_data method when we play back the actions, here i check if the train is local? and the connected_hex array includes a 'local'. If so create the same connection object as what i did in touch_node method. With this previous route now also works.

As it is now, the town is optional, but if any other game will add a "1+1" where the town isnt optional then we have to change this a little bit.

### DIFF
--- a/lib/engine/game/g_1817.rb
+++ b/lib/engine/game/g_1817.rb
@@ -183,9 +183,9 @@ module Engine
       def format_currency(val)
         # On dividends per share can be a float
         # But don't show decimal points on all
-        return format('$%.1<val>f', val: val) if val.is_a?(Float) && (val % 1 != 0)
+        return super unless val.is_a?(Float) && (val % 1 != 0)
 
-        self.class::CURRENCY_FORMAT_STR % val
+        format('$%.1<val>f', val: val)
       end
 
       def maximum_loans(entity)
--- a/lib/engine/game/g_1817.rb
+++ b/lib/engine/game/g_1817.rb
@@ -183,7 +183,7 @@ module Engine
       def format_currency(val)
         # On dividends per share can be a float
         # But don't show decimal points on all
-        return super unless val.is_a?(Float) && (val % 1 != 0)
+        return super if (val % 1).zero?
 
         format('$%.1<val>f', val: val)
       end
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -51,7 +51,7 @@ module Engine
       attr_accessor :bidding_token_per_player
 
       def can_run_route?(entity)
-        entity.trains.any? { |t| t.name == 'L' } ? true : super
+        entity.trains.any? { |t| t.name == 'L' } || super
       end
 
       def check_overlap(routes)
@@ -83,13 +83,10 @@ module Engine
         return [] unless (l_trains = runnable_trains.select { |t| t.name == 'L' })
 
         corporation = l_trains.first.owner
-        help = []
-        help << "L (local) trains run in a city which has a #{corporation.name} token."
-        help << 'They can additionally run to a single small station, but are not required to do so. '\
-                'They can thus be considered 1 (+1) trains.'
-        help << 'Only one L train may operate on each station token.'
-
-        help
+        ["L (local) trains run in a city which has a #{corporation.name} token.",
+         'They can additionally run to a single small station, but are not required to do so. '\
+         'They can thus be considered 1 (+1) trains.',
+         'Only one L train may operate on each station token.']
       end
 
       def init_round
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -50,15 +50,44 @@ module Engine
 
       attr_accessor :bidding_token_per_player
 
+      def can_run_route?(entity)
+        entity.trains.any? { |t| t.name == 'L' } ? true : super
+      end
+
       def entity_can_use_company?(_entity, company)
         # Setting bidding companies owner to bank, make sure the abilities dont show for theese
         company.owner != @bank
+      end
+
+      # TODO: Make include with 1817
+      def format_currency(val)
+        # On dividends per share can be a float
+        # But don't show decimal points on all
+        return format('$%.1<val>f', val: val) if val.is_a?(Float) && (val % 1 != 0)
+
+        self.class::CURRENCY_FORMAT_STR % val
+      end
+
+      def train_help(runnable_trains)
+        help = []
+
+        l_trains = runnable_trains.select { |t| t.name == 'L' }
+        if l_trains.any?
+          corporation = l_trains.first.owner
+          help << "L (local) trains run in a city which has a #{corporation.name} token."
+          help << 'They can additionally run to a single small station, but are not required to do so. '\
+                  'They can thus be considered 1 (+1) trains.'
+          help << 'Only one L train may operate on each station token.'
+        end
+
+        help
       end
 
       def init_round
         stock_round
       end
 
+      # TODO: Make include with 1861, 1867
       def operating_order
         minors, majors = @corporations.select(&:floated?).sort.partition { |c| c.type == :minor }
         minors + majors
@@ -72,7 +101,7 @@ module Engine
           Step::Track,
           Step::Token,
           Step::Route,
-          Step::Dividend,
+          Step::G1822::Dividend,
           Step::DiscardTrain,
           Step::BuyTrain,
         ], round_num: round_num)
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -73,26 +73,21 @@ module Engine
         company.owner != @bank
       end
 
-      # TODO: Make include with 1817
       def format_currency(val)
-        # On dividends per share can be a float
-        # But don't show decimal points on all
-        return format('$%.1<val>f', val: val) if val.is_a?(Float) && (val % 1 != 0)
+        return super unless val.is_a?(Float) && (val % 1 != 0)
 
-        self.class::CURRENCY_FORMAT_STR % val
+        format('£%.1<val>f', val: val)
       end
 
       def train_help(runnable_trains)
-        help = []
+        return [] unless (l_trains = runnable_trains.select { |t| t.name == 'L' })
 
-        l_trains = runnable_trains.select { |t| t.name == 'L' }
-        if l_trains.any?
-          corporation = l_trains.first.owner
-          help << "L (local) trains run in a city which has a #{corporation.name} token."
-          help << 'They can additionally run to a single small station, but are not required to do so. '\
-                  'They can thus be considered 1 (+1) trains.'
-          help << 'Only one L train may operate on each station token.'
-        end
+        corporation = l_trains.first.owner
+        help = []
+        help << "L (local) trains run in a city which has a #{corporation.name} token."
+        help << 'They can additionally run to a single small station, but are not required to do so. '\
+                'They can thus be considered 1 (+1) trains.'
+        help << 'Only one L train may operate on each station token.'
 
         help
       end
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -54,6 +54,20 @@ module Engine
         entity.trains.any? { |t| t.name == 'L' } ? true : super
       end
 
+      def check_overlap(routes)
+        super
+
+        # Check local train not use the same token more then one time
+        local_token_hex = []
+        routes.each do |route|
+          local_token_hex << route.head[:left].hex.id if route.train.local? && !route.connections.empty?
+        end
+
+        local_token_hex.group_by(&:itself).each do |k, v|
+          raise GameError, "Local train can only use the token on #{k[0]} once." if v.size > 1
+        end
+      end
+
       def entity_can_use_company?(_entity, company)
         # Setting bidding companies owner to bank, make sure the abilities dont show for theese
         company.owner != @bank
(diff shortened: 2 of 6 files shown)


lib/engine/train.rb
```diff
@@ -78,6 +79,18 @@ def buyable
       @buyable && !@obsolete
     end
 
+    def local?
+      return @local unless @local.nil?
```
crericha, line 83: !@local

lib/engine/train.rb
```diff
@@ -78,6 +79,18 @@ def buyable
...
+
+      @local = false
+      if @distance.is_a?(Numeric)
+        @local = @distance == 1
+      else
+        distance_city = @distance.find { |n| n['nodes'].include?('city') }
+        @local = distance_city['visit'] == 1 if distance_city
+      end
+    end
+
```
crericha, line 93: Is there a reason that @local shouldn't be determined at initialization?

lib/engine/train.rb
magnusfalt: No since its a boolean you cant use !@local

lib/engine/train.rb
magnusfalt: No reason, i can do an init_local if that fits better.

crericha:
I've always wondered why in the UI the route doesn't show until you click on the second revenue center. Is there a technical reason? Something else? Do you know @magnusfalt @tobymao? I haven't been in this code, so I have no idea.

If there isn't a reason, is this a good opportunity to have the UI support a single revenue center and simply show invalid route unless an 'L' train?

lib/engine/train.rb
crericha: Yeah, that would be better. Then it can simply be an attr_reader var.

magnusfalt:
Every route is based on a connection object which is located on the hex. The code doesnt sugest a route until it can find a connection. This means a connection between to nodes. I "fake" a node to get the localtrain to work. But i dont think the norm should be that.

lib/engine/train.rb
```diff
@@ -80,15 +90,7 @@ def buyable
     end
 
     def local?
```
crericha, line 92: Remove method. Use local attr.

magnusfalt: I have to disagree, is it a ruby standard to have a ? in the end when you have a method that will return a boolean?

This was the reason behind the first version of that method.

crericha: Its true that including a ? in the method is an often used idiom of Ruby; however, the file you are working in isn't using that idiom. See:

      @buyable = true
      @rusted = false
      @obsolete = false
      @operated = false

Prefer conformance to the source, so it has an unified style.

crericha:
> Every route is based on a connection object which is located on the hex. The code doesnt sugest a route until it can find a connection. This means a connection between to nodes. I "fake" a connection to get the localtrain to work. But i dont think the norm should be that.

Toby may have some suggestions. I don't think he's going to like the idea of faking it.

General advice: if you feel like you need to fake or hack something to get it to work, first talk it over with the dev team in the dev channel and/or talk it over with Toby. This will save you time in the long run if the approach you took isn't accepted.

lib/engine/game/g_1822.rb
```diff
@@ -50,15 +50,58 @@ class G1822 < Base
...
       def entity_can_use_company?(_entity, company)
         # Setting bidding companies owner to bank, make sure the abilities dont show for theese
         company.owner != @bank
       end
 
+      # TODO: Make include with 1817
+      def format_currency(val)
+        # On dividends per share can be a float
+        # But don't show decimal points on all
+        return format('$%.1<val>f', val: val) if val.is_a?(Float) && (val % 1 != 0)
```
tobymao, line 80: can we use the other syntax ? '$%.1 ' % val

lib/engine/game/g_1822.rb
```diff
@@ -50,15 +50,58 @@ class G1822 < Base
...
+
+        self.class::CURRENCY_FORMAT_STR % val
```
tobymao, line 82: just c all super?

tobymao: return super if ... && ....

'special " % val

lib/engine/game/g_1822.rb
```diff
@@ -50,15 +50,58 @@ class G1822 < Base
...
+      end
+
+      def train_help(runnable_trains)
+        help = []
+
+        l_trains = runnable_trains.select { |t| t.name == 'L' }
+        if l_trains.any?
```
tobymao, line 89: unless (l_trains = ...).empty?

tobymao: return [] unless (l_trains = ...)empty?

[
'a'
'b'
'c'
]

lib/engine/route.rb
```diff
@@ -222,7 +230,9 @@ def revenue
       @revenue ||=
         begin
           visited = visited_stops
-          raise GameError, 'Route must have at least 2 stops' if connection_data.any? && visited.size < 2
+          if connection_data.any? && visited.size < 2 && !@train.local?
```
tobymao, line 233: connectionsdata.empty?

lib/engine/train.rb
tobymao: if it's just one attr boolean, don't use the ? if it's more complicated use the question mark, in your case i think get rid of init_local

change to

def local?
  return @local if defined?(@local)

  @local = if @distance.is_a?(numeric)
   @distance == 1
else 
  @distance.select { |d| d['nodes'].include?('city') }.sum { | |d| d['visit']  == 1
end

lib/engine/route.rb
```diff
@@ -123,16 +123,22 @@ def touch_node(node)
             connection_data << segment(connection, left: tail[:right])
           end
         end
+        connection_data.pop if @train.local? && connection_data.size == 2
       elsif @last_node == node
         @last_node = nil
+        connection_data.clear if @train.local? && !connection_data.empty?
```
tobymao, line 129: do you need to check connection_data here? why not just always clear it?

lib/engine/route.rb
```diff
@@ -123,16 +123,22 @@ def touch_node(node)
             connection_data << segment(connection, left: tail[:right])
           end
         end
+        connection_data.pop if @train.local? && connection_data.size == 2
```
tobymao, line 126: why do you pop off a connection if the data is 2? 

how can you ever run a city to a town?

lib/engine/route.rb
```diff
@@ -123,16 +123,22 @@ def touch_node(node)
...
       elsif @last_node == node
         @last_node = nil
+        connection_data.clear if @train.local? && !connection_data.empty?
       elsif @last_node
+        connection_data.clear if @train.local? && !connection_data.empty?
```
tobymao, line 131: always clear?

lib/engine/route.rb
```diff
@@ -123,16 +123,22 @@ def touch_node(node)
...
         if (connection = select(@last_node, node)[0])
           a, b = connection.nodes
           a, b = b, a if @last_node == a
           connection_data << { left: a, right: b, connection: connection }
         end
       else
         @last_node = node
+        if @train.local? && connection_data.empty?
```
tobymao, line 139: just do @connection_data = { ... Connection.new}

lib/engine/route.rb
```diff
@@ -285,6 +299,13 @@ def connection_data
       @connection_data = []
       return @connection_data unless @connection_hexes
 
+      if @train.local? && @connection_hexes.one? && @connection_hexes[0].include?('local')
```
tobymao, line 302: why do you need to check include? local what does that do

lib/engine/route.rb
```diff
@@ -285,6 +299,13 @@ def connection_data
...
+        city_node = @game.hex_by_id(@connection_hexes[0][1]).tile.nodes.find do |n|
+          @game.city_tokened_by?(n, corporation)
+        end
+        return @connection_data << { left: city_node, right: city_node, connection: Connection.new }
```
tobymao, line 306: let's make a helper method for this, second time you're calling it

lib/engine/train.rb
crericha: Good news is that's what you had and can revert. My read on the style was different than Tobys.

lib/engine/game/g_1822.rb
magnusfalt: If i try the syntax above i get an RuboCop message:

RuboCop: Favor `format` over `String#%`. [Style/FormatString]

lib/engine/game/g_1822.rb
magnusfalt: Good idea. But i took the code from 1817 as i stated in the TODO to make an include since both games uses the same logic of halfpay. Shall i change it in 1817 as well or make an include?

lib/engine/route.rb
magnusfalt: No its true we can always clear it here.

lib/engine/route.rb
magnusfalt: Yes you can the rules states as following:

L (local) trains run in a city which has a station token of the owning (minor or major) company. They can additionally run to a single small station, but are not required to do so. They can thus be considered 1 (+1) trains. And if you run the lokal to a town you get a correct connection object. Remove the one created when you just pressed the city.

lib/engine/route.rb
magnusfalt: Yes i think we can.

lib/engine/route.rb
magnusfalt: I explaned the whole reason for the 'local' in the comment above. But realised it might be a wall of text :)

To get the last_route to work i first modify the connection_hexes method a bit. Since there are no path in the connection_data object no data was returned from the id method on in the connection object. Here i opted to save the local connection info with a format of ['local', hex.id]. The action entry now looks the same for the local as it does for any normal path object, with the exception for the 'local'.

The 'local' is need in the connection_data method when we play back the actions, here i check if the train is local? and the connected_hex array includes a 'local'. If so create the same connection object as what i did in touch_node method. With this previous route now also works.

lib/engine/route.rb
magnusfalt: Done.

lib/engine/route.rb
magnusfalt: Made a helper method as stated below.

lib/engine/game/g_1822.rb
tobymao: yes fix code as you see it

lib/engine/game/g_1822.rb
```diff
@@ -50,15 +50,53 @@ class G1822 < Base
 
       attr_accessor :bidding_token_per_player
 
+      def can_run_route?(entity)
+        entity.trains.any? { |t| t.name == 'L' } ? true : super
```
tobymao, line 54: any? || super

lib/engine/game/g_1822.rb
```diff
@@ -50,15 +50,53 @@ class G1822 < Base
...
+        return super unless val.is_a?(Float) && (val % 1 != 0)
+
+        format('£%.1<val>f', val: val)
+      end
+
+      def train_help(runnable_trains)
+        return [] unless (l_trains = runnable_trains.select { |t| t.name == 'L' })
+
+        corporation = l_trains.first.owner
+        help = []
```
tobymao, line 86: you can just do an array literal here

lib/engine/route.rb
```diff
@@ -257,11 +265,19 @@ def corporation
...
+      @connection_hexes ||= if @train.local? && connection_data.one? && connection_data[0][:connection].paths.empty?
+                              [['local', connection_data[0][:left].hex.id]]
+                            else
+                              connections&.map(&:id)
+                            end
     end
 
     private
 
+    def add_node_connection(node)
```
tobymao, line 277: make this a bit more explicit

def add_single_node_connection
def add_solo_connection
etc

tobymao:
looks good, small style things and i'll merge tomorrow

lib/engine/game/g_1817.rb
```diff
@@ -183,9 +183,9 @@ def interest_change
       def format_currency(val)
         # On dividends per share can be a float
         # But don't show decimal points on all
-        return format('$%.1<val>f', val: val) if val.is_a?(Float) && (val % 1 != 0)
+        return super unless val.is_a?(Float) && (val % 1 != 0)
```
tobymao, line 186: don't use unless with multiple values, use an if statement

PR #3597, opened 2021-01-30
1867: clarify private company nationalization event

etowle:
This change makes it more explicit that the `nationalize_companies` event only concerns privates. Minors, majors, and privates are all exclusively referred to as "companies" in the rules.


tobymao:
18xx.games explicitly refers to private companies as companies,  minor companies as minors and major companies as corporations. so the terminology is consistent, but private company doesn't hurt

jenf (approved):
Being a Brit I'll still refer to corps as companies occasionally, so worth it.

PR #3613, opened 2021-01-31
Add 1861 map

jenf:
Generalize the Timmins bonus
Remove neutral tokens for 1861
Change the RSR token behavior for 1861
Screenshot from 2021-01-31 17-09-04

### DIFF
--- a/lib/engine/game/g_1861.rb
+++ b/lib/engine/game/g_1861.rb
@@ -32,7 +32,7 @@ module Engine
         unipoed = @corporations.reject(&:ipoed)
         minor = unipoed.select { |c| c.type == :minor }
         major = unipoed.select { |c| c.type == :major }
-        ["#{major.size} major", minor + [@national]]
+        ["#{major.size} major", [@national] + minor]
       end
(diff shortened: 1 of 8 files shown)


lib/engine/game/g_1861.rb
```diff
@@ -14,15 +14,64 @@ class G1861 < G1867
...
+
       def self.title
         '1861'
       end
 
       def unstarted_corporation_summary
-        summary, _corps = super
-        [summary, [@national]]
+        unipoed = @corporations.reject(&:ipoed)
+        minor = unipoed.select { |c| c.type == :minor }
```
tobymao, line 33: partition ?

tobymao (approved):

lib/engine/game/g_1861.rb
jenf: They're not quite the same since there's minor, major and national in the list, so partition wouldn't give the same outcome

I've modified it slightly since having the national first is probably worthwhile since it'll be run.

PR #3617, opened 2021-01-31
1824: Support exchange Mountain and Coal Railways

perwestling:
Mountain Railways is exchanged during SR in phase 3 by selecting
a regional, and click on Exchange. This exchange is for a 10% share.

Coal Railway is done as a Special Buy during SR in phase 3 or 4.
This exchange is for a 20% presidency share.
(The reason for separate ways is that the presidency share is set
to No Buy, and implementing support for doing normal exchange is
complicated. Think this way is good enough.)

When phase 5 is started the exchange of any remaining Coal Railways
is triggered.

Only change in general code is to show Special Buy during stock
round if any step responds to special_buy action.

Some minor changes of details (e.g. names) discovered during
play testing.

### DIFF
--- a/lib/engine/game/g_1824.rb
+++ b/lib/engine/game/g_1824.rb
@@ -133,7 +133,7 @@ module Engine
           # So set them as closed and removed so that they do not appear
           # Affected: Coal Railway C4 (SPB), Regional Railway BH and SB, and possibly UG
           corporations.each do |c|
-            if (%w[SB BH].include?(c.name) || (two_player? && c.name == 'UG'))
+            if %w[SB BH].include?(c.name) || (two_player? && c.name == 'UG')
               c.close!
               c.removed = true
             end
--- a/lib/engine/game/g_1824.rb
+++ b/lib/engine/game/g_1824.rb
@@ -120,13 +120,6 @@ module Engine
       def init_corporations(stock_market)
         corporations = CORPORATIONS.dup
 
-        if option_cisleithania
-          # Remove Coal Railway C4 (SPB), Regional Railway BH and SB, and possibly UG
-          corporations.reject! do |c|
-            (%w[SB BH].include?(c[:sym]) || (two_player? && c[:sym] == 'UG'))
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_1824.rb
```diff
@@ -127,13 +127,27 @@ def init_corporations(stock_market)
...
           )
         end
+
+        return corporations unless option_cisleithania
+
+        # Some corporations need to be removed, but they need to exists (for implementation reasons)
+        # So set them as closed and removed so that they do not appear
+        # Affected: Coal Railway C4 (SPB), Regional Railway BH and SB, and possibly UG
+        corporations.reject! do |c|
+          next unless %w[SPB SB BH].include?(c.id) || (two_player? && c.id == 'UG')
```
tobymao, line 144: don't call unless with multiple conditions

lib/engine/step/g_1824/buy_sell_par_exchange_shares.rb
```diff
@@ -0,0 +1,97 @@
...
+module Engine
+  module Step
+    module G1824
+      class BuySellParExchangeShares < BuySellParShares
+        EXCHANGE_ACTIONS = %w[buy_shares].freeze
+        BUY_ACTION = %w[special_buy].freeze
+
+        def actions(entity)
+          if entity.company?
+            return can_exchange?(entity) ? EXCHANGE_ACTIONS : []
```
tobymao, line 14: one liner?

lib/engine/game/g_1824.rb
perwestling: Fixed. Was actually a bug as well so I rewrote this method.

lib/engine/step/g_1824/buy_sell_par_exchange_shares.rb
perwestling: Fixed

tobymao (approved):

PR #3619, opened 2021-01-31
18CZ Added BuyCorporation Step

Zwergenpunk:
This PR includes the Buy Corporation for 18CZ

A corporation can buy another corporation from 50% to 150% of share value of bought corporation. The shares are payed out by the buying corporation

All Tokens are transfered to new corporation. double tokens can be placed again for a price of 100

Trains are moved to new corporation. All new trains can kept, discarded or upgraded. (I added a restrain to train limit)

### DIFF
--- a/assets/app/view/game/buy_corporation.rb
+++ b/assets/app/view/game/buy_corporation.rb
@@ -2,6 +2,7 @@
 
 require 'view/game/actionable'
 require 'view/game/corporation'
+require 'view/game/buy_value_input'
 
 module View
   module Game
@@ -41,43 +42,11 @@ module View
 
       def render_input
         min_price, max_price = @step.price_range(@corporation, @selected_corporation)
-        input = h(:input, style: { marginRight: '1rem' }, props: {
-                    value: @selected_corporation.share_price.price,
-                    type: 'number',
-                    min: min_price,
-                    max: max_price,
-                    size: max_price,
-                  })
 
-        buy_click = lambda do
-          price = input.JS['elm'].JS['value'].to_i
-          buy = lambda do
-            process_action(Engine::Action::BuyCorporation.new(
-              @corporation,
-              corporation: @selected_corporation,
-              price: price,
-            ))
-            store(:selected_corporation, nil, skip: true)
-          end
-
-          if @selected_corporation.owner == @corporation.owner
-            buy.call
-          else
-            check_consent(@selected_corporation.owner, buy)
-          end
-        end
-
-        props = {
-          style: {
-            textAlign: 'center',
-            margin: '1rem',
-          },
-        }
-
-        h(:div, props, [
-          input,
-          h(:button, { on: { click: buy_click } }, 'Buy'),
-        ])
+        h(BuyValueInput, value: @selected_corporation.share_price.price, min_value: min_price,
+                         max_value: max_price,
+                         size: max_price,
+                         selected_entity: @selected_corporation)
       end
     end
   end
--- a/assets/app/view/game/buy_corporation.rb
+++ b/assets/app/view/game/buy_corporation.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+require 'view/game/corporation'
+
+module View
+  module Game
+    class BuyCorporation < Snabberb::Component
+      include Actionable
+      needs :selected_corporation, default: nil, store: true
+
+      def render
+        @step = @game.active_step
+        @corporation = @game.current_entity
+        children = []
+
+        @game.corporations.select { |item| @step.can_buy?(@corporation, item) }.each do |item|
+          children << h(Corporation, corporation: item)
+          children << render_input if item == @selected_corporation
+        end
+
+        h(:div, children)
+      end
+
+      def render_input
+        max_price = (@selected_corporation.share_price.price * 1.5).ceil
+        min_price = (@selected_corporation.share_price.price * 0.5).ceil
+        input = h(:input, style: { marginRight: '1rem' }, props: {
+                    value: @selected_corporation.share_price.price,
+                    type: 'number',
+                    min: min_price,
+                    max: max_price,
+                    size: max_price,
+                  })
+
+        buy_click = lambda do
+          price = input.JS['elm'].JS['value'].to_i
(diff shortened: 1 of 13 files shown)


tobymao:
can you just use the acquisition stuff from 1817?

Zwergenpunk:
Its completely different context. in 1817 multiple player can bid and in 18cz one player can choose one price for a corporation. 
It is not an extra round but just a step. 

Of course there is similar code, but I think that's not worth it to mix it.

benjaminxscott:
The acquirer also gets the private companies ("local railways" in this game)
e.g. https://github.com/benjaminxscott/18xx/blob/master/lib/engine/step/g_18_co/takeover.rb#L167

Zwergenpunk:
> The acquirer also gets the private companies ("local railways")
> We could add something like in CO e.g. https://github.com/benjaminxscott/18xx/blob/master/lib/engine/step/g_18_co/takeover.rb#L167

really? I did not find it in the rules. Also getting the companies are just 2 lines of code

assets/app/view/game/buy_corporation.rb
```diff
@@ -0,0 +1,85 @@
...
+          children << h('button',
+                        { on: { click: -> { store(:show_other_players, false) } } },
+                        'Hide corporations from other players')
+        end
+
+        h(:div, children)
+      end
+
+      def render_input
+        max_price = (@selected_corporation.share_price.price * 1.5).ceil
```
tobymao, line 43: you should not have any business logic here

assets/app/view/game/buy_corporation.rb
```diff
@@ -0,0 +1,85 @@
...
+          buy = lambda do
+            process_action(Engine::Action::BuyCorporation.new(
+              @corporation,
+              corporation: @selected_corporation,
+              price: price,
+            ))
+            store(:selected_corporation, nil, skip: true)
+          end
+
+          if @selected_corporation.owner == @corporation.owner
```
tobymao, line 64: this stuff looks copy and pasted from the company, can you instead reuse the buy companies view

Zwergenpunk: Yeah it is. So should I create a new component thats referenced by both?

tobymao: try to get buy_company to with corporations

Zwergenpunk: I think it is not really the best idea to mix the buy company and buy corporation view. There are some differences and in the end it would just be a if(corporation) render_buy_corporation else render_buy_companies. Then it is clearer to have 2 seperate classes. 

I grouped the input into an own component, which shares nearly all the code except the input value parameters.

assets/app/view/game/buy_value_input.rb
```diff
@@ -0,0 +1,67 @@
...
+                    type: 'number',
+                    min: @min_value,
+                    max: @max_value,
+                    size: @size,
+                  })
+
+        buy_click = lambda do
+          price = input.JS['elm'].JS['value'].to_i
+          buy = lambda do
+            if @selected_entity.is_a?(Engine::Corporation)
```
tobymao, line 29: i think you can just do

@selected_entity.corporation?

assets/app/view/game/upgrade_or_discard_trains.rb
```diff
@@ -0,0 +1,66 @@
...
+          h('div.margined', train_options)
+        end
+        children = []
+
+        children << h(:div, block_props, [
+          h(Corporation, corporation: corporation),
+          h(:div, trains),
+        ])
+
+        children << h(Map, game: @game) if @game.round.is_a?(Engine::Round::Operating)
```
tobymao, line 57: round.operating?

lib/engine/step/g_18_cz/upgrade_or_discard_train.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        def round_state
+          {
+            bought_trains: [],
+          }
+        end
+
+        def active?
+          buying_entity && trains.any?
```
tobymao, line 28: !trains.empty?

lib/engine/step/g_18_cz/upgrade_or_discard_train.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+
+        def process_pass(action)
+          trains.clear
+          super
+        end
+
+        def upgrade_infos(train, corporation)
+          variant = train.variants.values.find { |item| @game.train_of_size?(item, corporation.type) }
+          return nil, nil if variant.nil?
```
tobymao, line 82: unless variant

lib/engine/step/token_merger.rb
```diff
@@ -37,17 +37,17 @@ def round_state
         }
       end
 
-      def move_tokens_to_surviving(surviving, others)
+      def move_tokens_to_surviving(surviving, others, price_for_new_token = 0, check_tokenable = true)
```
tobymao, line 40: make these kwargs

Zwergenpunk: not sure if that is what you wanted

lib/engine/step/g_18_cz/buy_corporation.rb
```diff
@@ -52,7 +52,7 @@ def process_buy_corporation(action)
           }
 
           remove_duplicate_tokens(entity, corporation)
-          tokens = move_tokens_to_surviving(entity, corporation, 100, false)
+          tokens = move_tokens_to_surviving(entity, corporation, { price_for_new_token: 100, check_tokenable: false })
```
tobymao, line 55: you don't need the hash

lib/engine/step/token_merger.rb
```diff
@@ -37,17 +37,17 @@ def round_state
         }
       end
 
-      def move_tokens_to_surviving(surviving, others, price_for_new_token = 0, check_tokenable = true)
+      def move_tokens_to_surviving(surviving, others, **opt)
```
tobymao, line 40: no, price_for_new_token: 0, check_tokenable: true

Zwergenpunk: ah ok. thanks

PR #3626, opened 2021-02-01
18zoo add tickets

carlorusso:
Addition to 18ZOO:
- add ZOOTickets on game creation
- companies can be bought from players
- companies (only OR or ZOOTicket from owner player) can be bought by corporation
- ZOOTicket can be sold by corporation (price to player, difference to corporation)
- ZOOTickets value changes every round
- Player orders is higher cash + current order (but not on Draft)

### DIFF
--- a/lib/engine/game/g_18_zoo.rb
+++ b/lib/engine/game/g_18_zoo.rb
@@ -257,13 +257,6 @@ module Engine
 
       SMALL_MAP = %i[map_a map_b map_c].freeze
 
-      CERT_LIMIT = {
-        2 => { 5 => 10, 7 => 12 },
-        3 => { 5 => 7, 7 => 9 },
-        4 => { 5 => 5, 7 => 7 },
-        5 => { 5 => 6, 7 => 6 },
-      }.freeze
-
       CERT_LIMIT_INCLUDES_PRIVATES = false
 
       STOCKMARKET_COLORS = {
@@ -492,6 +485,13 @@ module Engine
 
       MARKET_SHARE_LIMIT = 80 # percent
 
+      ZOO_TICKET_VALUE = {
+        1 => { 0 => 4, 1 => 5, 2 => 6 },
+        2 => { 0 => 7, 1 => 8, 2 => 9 },
+        3 => { 0 => 10, 1 => 12, 2 => 15, 3 => 18 },
+        4 => { 0 => 20 },
+      }.freeze
+
       attr_reader :available_companies, :future_companies
 
       def setup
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_zoo.rb
```diff
@@ -546,6 +546,42 @@ def self.title
         '18ZOO'
       end
 
+      def purchasable_companies(entity = nil)
+        entity ||= @round.current_operator
+        return [] unless entity && (entity.corporation? || entity.player?)
```
tobymao, line 551: don't use unless with multiple conditoions change it to an if

lib/engine/step/buy_company.rb
```diff
@@ -104,6 +99,16 @@ def round_state
       def setup
         @blocks = @opts[:blocks] || false
       end
+
+      def pay(entity, owner, price, company)
+        entity.spend(price, owner.nil? ? @game.bank : owner)
```
tobymao, line 104: owner || @game.bank

lib/engine/step/g_18_zoo/buy_company.rb
```diff
@@ -0,0 +1,47 @@
...
+            @log << "#{corporation.name} earns #{@game.format_currency(company.value - price)}"
+
+            company.close!
+          else
+            super
+          end
+        end
+
+        def pay(entity, owner, price, company)
+          entity.spend(price, owner.nil? ? @game.bank : owner) if price.positive?
```
tobymao, line 39: owner || banke

lib/engine/step/g_18_zoo/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative '../buy_sell_par_shares'
+
+module Engine
+  module Step
+    module G18ZOO
+      class BuySellParShares < BuySellParShares
+        def actions(entity)
```
tobymao, line 9: why did you need to implement this method?

carlorusso: the original contains
`actions << 'buy_company' unless purchasable_companies(entity).empty?`
but the `purchasable_companies` is used for both showing the buyable companies on the top and on the bottom
Screenshot 2021-01-31 at 21 49 06
I don't want to add `can_buy_companies_from_other_players` but without that I'll never be able to show the companies.

Maybe we could change the behaviour to be able to buy from bank without `can_buy_companies_from_other_players` ?

lib/engine/game/g_18_zoo.rb
carlorusso: done in 955e5710

lib/engine/step/buy_company.rb
carlorusso: done in 955e5710

lib/engine/step/g_18_zoo/buy_company.rb
carlorusso: done in 955e5710

tobymao (approved):

PR #3628, opened 2021-02-01
1867: Previous corp nationalizes causes skip of train buy

### LINKED ISSUE #3625: 1867 - New company without train doesnt get opportunity to buy train
Original game is https://18xx.games/game/23040
Copied game to hotseat is https://18xx.games/hotseat/hs_uwbvphic_23040

GWR operated by packrat, skips the train purchase phase (eg. would buy a train over from packrat's C&O).

Nothing in the rules indicate that this is what should occur and that GWR should be able to purchase another company's train.

jenf:
In game 23040 NYC was nationalized since it didn't buy a train,
this causes it to skip to the new corps turn GWR but the unpass occurs before
the pass, move the nationalization to after the pass. (fixes #3625)

### DIFF
--- a/lib/engine/step/g_1867/buy_train.rb
+++ b/lib/engine/step/g_1867/buy_train.rb
@@ -24,7 +24,7 @@ module Engine
 
         def pass!
           super
-          @game.nationalize!(current_entity) if current_entity.trains.none?
+          @game.nationalize!(current_entity) if current_entity.trains.empty?
         end
 
         def discountable_trains_allowed?(_entity)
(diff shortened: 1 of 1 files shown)


lib/engine/step/g_1867/buy_train.rb
```diff
@@ -23,8 +23,8 @@ def available_cash(_player)
         end
 
         def pass!
-          @game.nationalize!(current_entity) if current_entity.trains.none?
           super
+          @game.nationalize!(current_entity) if current_entity.trains.none?
```
tobymao, line 27: empty?

tobymao (approved):

PR #3630, opened 2021-02-01
zoom controls on map

yzemaze:
Follow-up to #3489

100 % zoom (default):
image

62 % zoom:
image

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -35,8 +35,7 @@ module View
         @start_pos = [@cols.first, @rows.first]
         @layout = @game.layout
 
-        store(:map_zoom, @map_zoom || Lib::Storage['map_zoom'] || 1, skip: true)
-        @scale = SCALE * @map_zoom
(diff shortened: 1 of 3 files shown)


assets/app/view/game/map.rb
```diff
@@ -35,7 +35,8 @@ def render
         @start_pos = [@cols.first, @rows.first]
         @layout = @game.layout
 
-        @scale = SCALE * map_zoom
+        store(:map_zoom, @map_zoom || Lib::Storage['map_zoom'] || 1, skip: true)
```
tobymao, line 38: why did you change this? don't store with the storage, it gets hard

yzemaze: Well, I wasn’t sure about this, hence its own commit. I thought it’s better to store once per render than to read 7 times (iirc) from localStorage.

tobymao: well you have to revert it

tobymao (approved):

PR #3638, opened 2021-02-01
[1822] Upgrade L train to a 2 train.

magnusfalt:
**Problem:**
The first traincard is double sided, on one side there is an L train. On other side there is a 2-train. You can choose which one you want to buy. If you choose the L train then in the buytrain step you can upgrade the L to a 2 by flipping the card. The game begins in phase 1, phase 2 will begin be first 2 i either bought from depot of upgraded from an L train.

**Solution:**
I needed a way to use a variant on a train to flip it from a L to 2. There was code for buying variant trains, but not upgrade. I had to override the depot-class just to get the upgrade option under Exchange.

When you buy a variant train, every propery on the trainclass is overridden by the variant properties except the sym. I understand this is a must to get the deserialize from the h_to_args with game.train_by_id to work properly. But this was a problem for med beacause the method game.phase.buying_train! is looking on the train.sym to determine if there is a phase change. In my case i wanted a phase change if you buy the variant or upgrade a L to a 2. I had to do a little create solution to get it to work because i couldnt change the buying_train! method to check name, this would have broken the other games which uses a variant, i e 1846. The phasechange and train-rusting is relying on the base train sym.

Since i had to save the orginal sym when upgrading from an L to a 2 i had to change a little bit in the code i did in the route. Since the id for the train is the same for a L and the upgraded 2 the previous route would have caused an error without a little extra check.


lib/engine/g_1822/depot.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require_relative '../depot'
+
+module Engine
```
tobymao, line 5: avoid creating sub classes of base models, switch the depot to call an overridable method on game

lib/engine/game/g_1822.rb
```diff
@@ -93,6 +98,17 @@ def init_round
         stock_round
       end
 
+      # This is need to handle the upgrade from L -> 2 train.
+      def init_train_handler
```
tobymao, line 102: delete this

lib/engine/step/g_1822/buy_train.rb
```diff
@@ -0,0 +1,55 @@
...
+          check_spend(action)
+          if action.exchange
+            upgrade_train_action(action)
+          else
+            buy_train_action(action)
+          end
+          pass! unless can_buy_train?(action.entity)
+
+          # Special case when we are in phase 1, and first 2 train is bought or upgraded
+          return unless @game.phase.name.to_i == 1 && action.train.name == '2'
```
tobymao, line 27: change to an if

lib/engine/depot.rb
```diff
@@ -111,6 +111,8 @@ def discountable_trains_for(corporation)
           discount_info
         end.compact
       end
+      other_discounts = @game.discountable_trains_for(corporation)
```
tobymao, line 114: move this whole method there. this is not providing much benefit if people want to override the top part,  also now there is a performance hit (probably negligble) of creating new arrays with discounts + []

tobymao (approved):

PR #3658, opened 2021-02-03
Fix invite page var name and auto refresh

daniel-sousa-me:
Previously the invitation page did not refresh automatically, which was confusing to a lot of users. I took the chance and created a channel to listen to general game metadata changes to make the home page update automatically and avoid the need to refresh.

I've increased the refresh timeout, since it's less important now. It is still there because the new channel does not publish when an action is submitted. I think that would require a more targeted approach to avoid flooding all the online users.

PS: I'm sorry for lashing out the other day. My life has been rough, but that's no excuse for my reaction. I really appreciate Toby's selfless work (as everyone else's)

### DIFF
--- a/assets/app/game_manager.rb
+++ b/assets/app/game_manager.rb
@@ -54,6 +54,13 @@ module GameManager
     end
   end
 
+  def get_game(id)
+    @connection.safe_get("/game/#{id}") do |data|
+      data[:loaded] = @game_data[:loaded] if @game_data
+      update_game(data)
+    end
+  end
+
   def create_game(params)
     @connection.safe_post('/game', params) do |data|
       store(:games, [data] + @games)
@@ -166,7 +173,7 @@ module GameManager
     @games += [game] if @games.none? { |g| g['id'] == game['id'] }
     @games.reject! { |g| g['id'] == game['id'] } if game['deleted']
     @games.map! { |g| g['id'] == game['id'] ? game : g }
-    store(:game, game, skip: true) if @game&.[]('id') == game['id']
+    store(:game_data, game, skip: true) if @game_data&.dig('id') == game['id']
     store(:games, @games.sort_by { |g| g['id'] }.reverse)
   end
 end
--- a/assets/app/view/invite_game.rb
+++ b/assets/app/view/invite_game.rb
@@ -8,17 +8,45 @@ module View
     include GameManager
 
     needs :user
+    needs :refreshing, default: false, store: true
 
     def render
+      game_refresh
+
       children = [h(:h2, 'This game has not started yet')]
       if @user
-        children << h(:div, [h(GameCard, gdata: @game, user: @user)])
+        children << h(:div, [h(GameCard, gdata: @game_data, user: @user)])
       else
         children << h(:h3, 'You need to login before joining a game:')
         children << h(View::User, type: :login)
       end
 
-      h('div', children)
+      destroy = lambda do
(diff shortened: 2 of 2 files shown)


tobymao:
welcome back @daniel-sousa-me 

the page refresh is there because it's much more robust. instead of this approach, can you instead just have the join game page refresh as well?

daniel-sousa-me:
But should I remove the message bus from the home?
There are 2 instances where having only refresh is not ideal:
1. If you're creating a live game with specific people. It can take up to 10s for it to update
2. When a tab is on the background sometimes the browser freezes it. When the user comes back to it, currently it doesn't update instantly. With the message bus it's always up to date.

tobymao:
yes, please remove the message-bus that you added

daniel-sousa-me:
Done

assets/app/game_manager.rb
```diff
@@ -54,6 +54,15 @@ def get_games(params = nil)
     end
   end
 
+  def get_game(id = nil)
```
tobymao, line 57: why is there a default value here?

assets/app/game_manager.rb
```diff
@@ -54,6 +54,15 @@ def get_games(params = nil)
...
+    id ||= @game_data['id']
+
+    @connection.safe_get("/game/#{id}") do |data|
+      data[:loaded] = @game_data[:loaded] if @game_data
```
tobymao, line 61: i forgot how this code works, can you explain what this is doing?

assets/app/view/invite_game.rb
```diff
@@ -8,17 +8,45 @@ class InviteGame < Snabberb::Component
...
+        children << h(:div, [h(GameCard, gdata: @game_data, user: @user)])
       else
         children << h(:h3, 'You need to login before joining a game:')
         children << h(View::User, type: :login)
       end
 
-      h('div', children)
+      destroy = lambda do
+        `clearTimeout(#{@refreshing})`
+        store(:refreshing, nil, skip: true)
```
tobymao, line 26: do you need store(:refreshing) here? i'm curious if it's needed -- not sure

assets/app/game_manager.rb
daniel-sousa-me: The api returns the object with loaded=true (I don't know why). When the game starts the page automatically enters the game (same url), but if I left loaded=true it doesn't create everything that needs to be created.

assets/app/view/invite_game.rb
daniel-sousa-me: I think the settimeout function returns an id and not the object itself, but I didn't check. This is just copy-pasted from Home

assets/app/game_manager.rb
daniel-sousa-me: For no special reason. I just thought it would make the code more robust

assets/app/game_manager.rb
tobymao: ok i understand, i think we can clean this up a bit,

when you hit /game/:id, this is the full game with all actions which is not needed for the game invite,

what you need is an API to get the game metadata only, so perhaps you can do /game/:id?meta_only=1 to not load the actions and then loaded would not be true

daniel-sousa-me: This is only for the invitation page, so the actions will be empty (or at worst they will be very few only once)

tobymao (approved):

PR #3659, opened 2021-02-03
18Mag: Auto-lay token on yellow OO tiles

### LINKED ISSUE #3650: 18Mag: No choice of token position when laying OO yellow tiles
Currently, when a minor lays the first (yellow) tile on a OO hex, the player must select the city to place the home token. According to Lonny, this is incorrect - the token must go in the city that has track connected to it.

This will require pinning almost all existing games.

roseundy:
Minor tokens should always be placed on city index 0 when laying a yellow OO according to Lonny.
Before this change, the player had to select a city via HomeToken.

This change will require pinning most 18Mag games due to now superfluous place_token actions at the start of games.

Fixes #3650

### DIFF
--- a/lib/engine/step/g_18_mag/track.rb
+++ b/lib/engine/step/g_18_mag/track.rb
@@ -104,6 +104,21 @@ module Engine
           end
           @log << "#{@game.sik.name} earns #{@game.format_currency(cost - extra_cost)}"
         end
+
+        # handle yellow OO tile
+        def update_token!(_action, _entity, tile, old_tile)
+          cities = tile.cities
+          if old_tile.paths.empty? &&
+              !tile.paths.empty? &&
+              cities.size > 1 &&
+              cities.flat_map(&:tokens).any?
+            token = cities.flat_map(&:tokens).find(&:itself)
+
+            # always move token to city with index 0
+            token.move!(cities[0])
+            @game.graph.clear
+          end
+        end
       end
     end
   end
--- a/lib/engine/step/g_18_mag/track.rb
+++ b/lib/engine/step/g_18_mag/track.rb
@@ -111,8 +111,7 @@ module Engine
           if old_tile.paths.empty? &&
               !tile.paths.empty? &&
               cities.size > 1 &&
-              cities.flat_map(&:tokens).any?
-            token = cities.flat_map(&:tokens).find(&:itself)
+              (token = cities.flat_map(&:tokens).find(&:itself))
 
             # always move token to city with index 0
             token.move!(cities[0])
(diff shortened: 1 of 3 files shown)


lib/engine/step/g_18_mag/track.rb
```diff
@@ -104,6 +104,21 @@ def pay_tile_cost!(entity, tile, rotation, hex, spender, cost, extra_cost)
...
           @log << "#{@game.sik.name} earns #{@game.format_currency(cost - extra_cost)}"
         end
+
+        # handle yellow OO tile
+        def update_token!(_action, _entity, tile, old_tile)
+          cities = tile.cities
+          if old_tile.paths.empty? &&
+              !tile.paths.empty? &&
+              cities.size > 1 &&
+              cities.flat_map(&:tokens).any?
```
tobymao, line 114: (token = cities.flat_map(&:tokens).find(&:itself))

then yo ucan delete the second one

lib/engine/step/g_18_mag/track.rb
```diff
@@ -104,6 +104,21 @@ def pay_tile_cost!(entity, tile, rotation, hex, spender, cost, extra_cost)
...
+            token = cities.flat_map(&:tokens).find(&:itself)
+
+            # always move token to city with index 0
+            token.move!(cities[0])
+            @game.graph.clear
```
tobymao, line 119: how often does this happen?

lib/engine/step/tracker.rb
```diff
@@ -178,6 +165,24 @@ def pay_tile_cost!(entity, tile, rotation, hex, spender, cost, _extra_cost)
           "#{tile.location_name.to_s.empty? ? '' : " (#{tile.location_name})"}"
       end
 
+      def update_token!(action, entity, tile, old_tile)
+        cities = tile.cities
+        if old_tile.paths.empty? &&
+            !tile.paths.empty? &&
+            cities.size > 1 &&
+            cities.flat_map(&:tokens).any?
```
tobymao, line 173: fix this section too

lib/engine/step/g_18_mag/track.rb
roseundy: Every time a yellow tile is placed in a OO hex. Pretty much every single 18Mag game.

lib/engine/step/g_18_mag/track.rb
roseundy: OK.

lib/engine/step/tracker.rb
roseundy: Existing code, but I'll fix.

PR #3666, opened 2021-02-03
[1822] End of stock round

magnusfalt:
* End of stock round, remove trains if no bids on minors.
* Minors bidding par price, startorder and treasury.
* First turn housekeeping, minors can buy a L train if one still in the depot as first action in its first operation round it floated.
* Some minor code cleanup.

### DIFF
--- a/lib/engine/round/g_1822/stock.rb
+++ b/lib/engine/round/g_1822/stock.rb
@@ -131,7 +131,7 @@ module Engine
         def remove_l_trains(count)
           @game.log << "#{count} minors with no bids. If available up to #{count} L trains will be removed"
           while (train = @game.depot.upcoming.first).name == 'L' && count.positive?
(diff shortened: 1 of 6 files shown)


lib/engine/round/g_1822/stock.rb
```diff
@@ -97,6 +128,29 @@ def highest_bid(company)
           @bids[company]&.max_by(&:price)
         end
 
+        def remove_l_trains(count)
+          @game.log << "#{count} minors with no bids. If available up to #{count} L trains will be removed"
+          while (train = @game.depot.upcoming.first).name == 'L' && count.positive?
+            @game.depot.remove_train(train)
```
tobymao, line 134: change all depot methods to game methods,

this should be @game.remove_train

tobymao (approved):

Issue #3621, opened 2021-02-01
Remove share counts and use percentage everywhere


tobymao:
Share counts are confusing, I'd rather just have percentages. I think corporation symbols should be used only as opposed to the names.

From a list of JCL's feedback

Other beefs: 
- Lose the:
    - Currency symbols.
    - Share %ages instead of counts (which you do in some places and not others)
    - Last run value.
    - Value and liquidity for players.
    - And their repeats on the spreadsheet.
    - Legend on the market (players should know already).
    - %age left to float (the players should already know).
    - the company names (use symbols everywhere).
All those gone, everywhere, gone.
- Colour the  trains (I don't care what their definition is, I care where they are in the phasing, which is pretty much the only thing I ever care about for a train)
- Get rid of the row of tokens.  Just one with a count remaining and price for current.
- No, I don't care what the coords of the tokens are.  Ever.  And if I did the map would tell me.
- I've already whined on the mobile optimisations that prevent simpler/smaller interfaces (eg hover).
- Alternating between labels below and above values and then mixing with titles to the left is bad.  Pick.
- Not everything needs a label.
- And an odd one: surface current game phase (colour).

dfannius:
I disagree with most of jcl's beefs (unsurprising, given that we've spent a while iterating to get to the point that we're at). So I'd want a chance to defend anything on that "lose the" list before a decision was made to actually lose them.

I do think that train colors would be nice, though. I especially miss them in 1846.

tobymao:
agreed

chrispday:
The only 1 I agree with is surfacing the current game phase.

The rest is typical 18xx gate-keeping, that without new players would be absolutely lost.

tysen:
I think getting rid of the currency symbols might be good. That's mostly noise and f2f I (very close to) never say e.g. 'yen' instead of saying 'dollars', 'bucks', 'smackers', 'bones', 'clams' or omitting the currency signifier entirely. Could be a user setting.

Share count has a different use than percentage ownership (which mostly matches where you see one or the other in the UI). Mostly share count is useful in aggregate to compare revenue/growth potential (say, in a set of ORs) whereas percentage ownership is more valuable when looking at a particular corp.

perwestling:
Share count is kind of a problem in games with unusual share sizes. Maybe it should show 3 10%, 1 20%, 1 5% instead of 5.5?

When it comes to currency I do enjoy it as it gives some flavor. Now I know that is a personal choice. 

[...]

PR #3668, opened 2021-02-03
game menu: phase color

yzemaze:
In #3621 some folks expressed the need for always visualizing phase color. I think this works fine, isn’t too hard on the eyes and the right kind of subtle. (ymmv, personal color settings!)

image
image
image
image

image
image
image
image


scottredracecar:
Looks nice!

tobymao (approved):

scottredracecar:
@yzemaze First, I like this way of doing it. Another way to do it would be to color the 18xx.games logo instead.
image

yzemaze:
@scottredracecar 
That wouldn’t work that well, because the game menu scrolls over it. On mobile it isn’t visible most of the time. (Hence screenshots without it. They did show the top of the screen.)

PR #3679, opened 2021-02-04
Extract g_1889.js and deps.js from main bundle; lay groundwork for extracting all games to separate JS files

michaeljb:
`main.js.gz` has been split up into `main.js.gz`, `deps.js.gz`, and `g_1889.js.gz`. Locally I have the changes needed to also split out `g_18_mex`, `g_18_chesapeake`, and `g_18_chesapeake_off_the_rails`, but it seemed better for this PR to stick with a single title for now.

With the new game "schema", all game-specific code (e.g., for 1889) lives under `lib/engine/game/g_1889/`. In that subdirectory, `game.rb` contains the main `Game` class. `meta.rb` contains a `Meta` module; this module is `include`d in the `Game` class, but is also built into `main.js.gz` instead of `g_1889.js.gz`. It has the info necessary for creating a new game without needing to load in the whole class.

To separate a game whose `Game` class is either a parent or child of another game from `main.js.gz` to its own JS file, both games must be separated, i.e., `g_18_chesapeake.js.gz` and `g_18_chesapeake_off_the_rails.js.gz` must exist together (or at least, it would not be worth the trouble to try and make it work with one and not the other). When a game of 18Chesapeake is entered, `g_18_chesapeake.js.gz` must be loaded; when a game of 18Cheseapeake: Off the Rails is entered, both  `g_18_chesapeake.js.gz` and `g_18_chesapeake_off_the_rails.js.gz` must be loaded.

`rake precompile` still compiles a single bundle file to `public/pinned/`. It should continue to work for all alpha+ games, whether they are bundled in `main.js.gz` or separately (and they should all be separated ASAP). Games with a `DEV_STAGE` of `:prealpha` are now excluded from this bundle.

-----

TODO for this PR
- [x] create new multiplayer game
- [x] create new hotseat game
- [x] create new game by Importing JSON
- [x] enter multiplayer game by directly hitting URL
- [x]  enter hotseat game by directly hitting URL
- [x] enter multiplayer game by clicking "Enter" on game card
- [x] enter hotseat game by clicking "Enter" on game card
- [x] create new hotseat game by cloning a multiplayer game

TODO post merge:
* create PR for 18Chesapeake + 18Chesapeake Off the Rails
* create PR for 18MEX
* create PR for 1846 + 18 Los Angeles
* update dev wiki for new game "schema", pointing to games that use the new layout as examples


lib/engine.rb
```diff
@@ -8,19 +8,63 @@
...
+
+    game_meta = GAMES_BY_TITLE[title]
+    game = get_game(game_meta)
+
+    if game.nil? && RUBY_ENGINE == 'opal'
+      game_by_title(game_meta::DEPENDS_ON) if game_meta::DEPENDS_ON
+
+      src = "/assets/#{game_meta.fs_name}.js"
+
+      # add <script> tag to DOM to load the target game file
```
michaeljb, line 55: This approach for injecting a `<script>` tag to the DOM to load in the game file as needed doesn't seem to work consistently, though at times I've certainly thought it was consistently working.

I believe one reason why it isn't working regularly is that loading in the remote JS is asynchronous, and the only way to be sure that code is executed once the file has been loaded is with the `onload` callback function, but we can't make `game_by_title()` return what we want form an asynchronous callback.

It probably needs to be removed, and when entering a game by clicking "Enter" on the game card, the new page needs to be loaded in a different way than it is now, so that the server can return prerendered HTML including the `<script>` tag for the given game.

Any help on how to approach this particular problem would be greatly appreciated 😄

assets/app/game_manager.rb
```diff
@@ -115,7 +115,8 @@ def enter_game(game)
       end
 
       store(:game_data, game_data.merge(loaded: true), skip: true)
-      store(:app_route, hs_url(game, game_data)) unless @app_route.include?(hs_url(game, game_data))
+      game_url = hs_url(game, game_data)
+      `window.location = #{game_url}` unless @app_route.gsub('%20', ' ').include?(game_url)
```
michaeljb, line 119: For hotseat games; this forces the page to reload, and includes the game title in the URL so that when the client sends the server the request, the server knows if a game js needs to be inclued in the prerendered HTML; see https://github.com/tobymao/18xx/pull/3679/files#r570001184

lib/engine.rb
michaeljb: note: solving this is the main thing needed to solve the unchecked items in the first "TODO" list in the description above

lib/js_context.rb
```diff
@@ -3,14 +3,15 @@
...
+    @files = files
+    combined_files = files.map { |f| File.read(f, encoding: 'UTF-8').to_s }.join
+    @snapshot = MiniRacer::Snapshot.new(combined_files)
   end
 
   def eval(script)
     MiniRacer::Context
       .new(snapshot: @snapshot)
-      .eval(script, filename: @file)
+      .eval(script, filename: @files.join('|'))
```
michaeljb, line 15: I could not figure out from MiniRacer's docs what exactly `filename` was supposed to be here but what I did here is probably wrong.

Previously this context was just one file (`main.js`) but now we need multiple files (`main.js`, `deps.js`, and `g_1889.js`) and I doubt I set up the MiniRacer Context or Snapshot correctly.

lib/engine.rb
michaeljb: Instead of `enter_game`, this should perhaps act more like a normal link would?

https://github.com/tobymao/18xx/blob/e6c6b0bf444f1f3ae033af072f954fe2bd191050/assets/app/view/game_card.rb#L69

assets/app/game_manager.rb
tobymao, line 119: can we get rid of all this gsub %20 stuff instead making the title matching easier in general by doing something more fuzzy

assets/app/game_manager.rb
```diff
@@ -161,11 +162,12 @@ def url(game, path = '')
 
   def hs_url(game, game_data)
     pin = game_data&.dig('settings', 'pin')
+    title = game_data['title']
 
     if pin
       "/hotseat/#{game['id']}?pin=#{pin}"
     else
-      "/hotseat/#{game['id']}"
+      "/hotseat/#{game['id']}?title=#{title}"
```
tobymao, line 170: why do we need to pass in title at all? the game data should have that

lib/assets.rb
```diff
@@ -32,39 +35,76 @@ def html(script, **needs)
     context.eval(Snabberb.html_script(script, **needs))
   end
 
-  def build
-    return [@bundle_path] if @precompiled
+  def game_builds
```
tobymao, line 38: i think we might want to get rid of the source maps and instead make source maps for the uglifier

lib/engine.rb
```diff
@@ -8,19 +8,63 @@
...
 module Engine
+  @games = {}
+
+  # Game or Meta
   GAMES = Game.constants.map do |c|
     klass = Game.const_get(c)
-    next if !klass.is_a?(Class) || klass == Game::Base
 
-    klass
+    if c.start_with?('G')
```
tobymao, line 17: this doesn't seem safe

lib/engine.rb
```diff
@@ -8,19 +8,63 @@
...
+
+  # Game
+  def self.game_by_title(title)
+    return @games[title] if @games[title]
+    return @games[title] = GAMES_BY_TITLE[title] if GAMES_BY_TITLE[title].is_a?(Class)
+
+    game_meta = GAMES_BY_TITLE[title]
+    game = get_game(game_meta)
+
+    if game.nil? && RUBY_ENGINE == 'opal'
```
tobymao, line 50: !game

lib/engine.rb
```diff
@@ -8,19 +8,63 @@
...
+      game_by_title(game_meta::DEPENDS_ON) if game_meta::DEPENDS_ON
+
+      src = "/assets/#{game_meta.fs_name}.js"
+
+      # add <script> tag to DOM to load the target game file
+      `var s = document.createElement('script');
```
tobymao, line 56: should this be here? why isn't this in the app layer?

lib/engine.rb
```diff
@@ -8,19 +8,63 @@
...
+       s.type = 'text/javascript';
```
tobymao, line 57: could you instead do this with snabbdom?

michaeljb: @tobymao I'm now working on getting this to work with snabbdom in the app layer, will re-open the PR when I've got it working.

michaeljb: Moved to the app layer, but not with snabbdom.

The problem with snabbdom is that adding a script to the page will load it asynchronously, and to execute some code when it's done, the `onload` property can be used to define a script the browser calls when it's done. When this is done via a snabberb (ie `h(:script ...)`), `onload` doesn't get rendered to JS correctly.

assets/app/game_manager.rb
michaeljb: I think by making `Engine.game_by_title()` do the fuzzy matching for whatever string is passed in, that would work. Do you want that change on this PR or separately?

michaeljb:
Made a lot of changes in 2c638a4def9e8be0372625e8e1043de67d996fa2, including adding `rake move_game` which uses `scripts/move_game.rb` to move files around so a particular game can be bundled separately. It's not perfect but gets a lot of the job done.

9c58947c6e50043378b0cd9116e6c42e8007559c is the result of running `rake move_game[g_1889]`, and 67e8c6b9e1a68b6c5f952112edfab498ede69a64 was the manual work needed to get the rest of everything working. I've done this locally as well for 18Chesapeake, 18MEX, and 1846 and the output of the script is a big help.

michaeljb:
I did run into some problems debugging, the particular problem was with this line:

https://github.com/tobymao/18xx/pull/3679/files#diff-042ae7bda8741c1de33c019b9d3747ff0540a3e0b1572ff8461bcf11f4c89aa3R79

Originally I didn't have it "returning" `[]`, it was just `next [] unless (game_class = load_game_class(title))` and the error is not very useful. I'm thinking it has to do with either the source maps or what I pointed out here - https://github.com/tobymao/18xx/pull/3679#discussion_r570016639

lib/assets.rb
```diff
@@ -5,66 +5,135 @@
...
+            'files' => [@server_path],
+          },
+          **game_builds,
+        }
+      else
+        opal = compile_lib('opal')
+        deps = compile_lib('deps', 'assets')
+        engine = compile('engine', 'lib', 'engine')
+        app = compile('app', 'assets/app', '')
+        game_files = game_builds.values.flat_map { |g| g['files'] }
```
tobymao, line 78: do

tobymao (approved):

PR #3687, opened 2021-02-04
1856 does destination checks

ventusignis:
This PR will have 1856 check after every tile lay if any escrow corporations have destinated.
This is used to find which hexes are reachable from the home hex for a corporation. 

Later PRs will:
* Make this process use auto actions and not happen when @loading
* Use the results to determine if a corporation has destinated and if so release escrow

### DIFF
--- a/lib/engine/game/g_1856.rb
+++ b/lib/engine/game/g_1856.rb
@@ -312,7 +312,7 @@ module Engine
         POST_NATIONALIZATION_CERT_LIMIT[num_corporations][@players.size]
       end
 
-      def destination_check
+      def destination_check!
         @corporations.each do |corp|
           # The capitalization method of unparred corporations is nil
           next unless corp.capitalization == :escrow
--- a/lib/engine/game/g_1856.rb
+++ b/lib/engine/game/g_1856.rb
@@ -281,6 +281,7 @@ module Engine
       end
 
       def create_destinations(destinations)
+        @destinations = {}
         destinations.each do |corp, dest|
           dest_arr = Array(dest)
           end_a = dest_arr.first
@@ -294,6 +295,7 @@ module Engine
           dest_arr.each do |d|
             hex_by_id(d).original_tile.icons << Part::Icon.new("../logos/1856/#{corp}")
           end
+          @destinations[corp] = [end_b, end_a].freeze
         end
       end
 
@@ -307,6 +309,16 @@ module Engine
         return PRE_NATIONALIZATION_CERT_LIMIT[@players.size] unless @post_nationalization
 
         POST_NATIONALIZATION_CERT_LIMIT[num_corporations][@players.size]
+      end
+
+      def destination_check
+        @corporations.each do |corp|
+          # The capitalization method of unparred corporations is nil
+          next unless corp.capitalization == :escrow
+
+          reachable_hexes = @graph.ghost_train(corp, [hex_by_id(@destinations[corp.id].first)]).map { |h| h[0].id }
+          @log << "#{corp.name} destination check - hexes: #{reachable_hexes}"
+        end
       end
 
       #
--- a/lib/engine/game/g_1856.rb
+++ b/lib/engine/game/g_1856.rb
@@ -316,9 +316,26 @@ module Engine
           # The capitalization method of unparred corporations is nil
           next unless corp.capitalization == :escrow
 
-          reachable_hexes = @graph.ghost_train(corp, [hex_by_id(@destinations[corp.id].first)]).map { |h| h[0].id }
-          @log << "#{corp.name} destination check - hexes: #{reachable_hexes}"
+          destinated!(corp) if hexes_connected?(@destinations[corp.id].first, @destinations[corp.id].last)
         end
+      end
+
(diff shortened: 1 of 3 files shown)


tobymao:
this change is too invasive

can you make it a bit more generic and put the 1856 specific stuff somewhere else? 

as an example ( don't know if this fits your use case )

let's say you just need to know if two hexes are connected, allow graph to do that, and then make 1856 call this to it's purpose

or another example

if graph already works but needs a fake corp, 1856 could pass that in.

ventusignis:
After I cloned the method and ripped out everything that wasn't directly related to seeing if certain hexes were connected, there was barely anything left so I put everything into the game file

lib/engine/game/g_1856.rb
```diff
@@ -309,6 +311,33 @@ def cert_limit
         POST_NATIONALIZATION_CERT_LIMIT[num_corporations][@players.size]
       end
 
+      def destination_check
```
tobymao, line 314: might as well put a ! here because it calls a destinate!

lib/engine/game/g_1856.rb
```diff
@@ -309,6 +311,33 @@ def cert_limit
...
+        @corporations.each do |corp|
+          # The capitalization method of unparred corporations is nil
+          next unless corp.capitalization == :escrow
+
+          destinated!(corp) if hexes_connected?(@destinations[corp.id].first, @destinations[corp.id].last)
+        end
+      end
+
+      def hexes_connected?(start_hex_id, goal_hex_ids)
+        tokens = hex_by_id(start_hex_id).tile.cities.map { |city| [city, true] }.to_h
```
tobymao, line 324: why is this a hash? doesn't seem like you use it excepts as an array?

lib/engine/game/g_1856.rb
```diff
@@ -309,6 +311,33 @@ def cert_limit
         POST_NATIONALIZATION_CERT_LIMIT[num_corporations][@players.size]
       end
 
+      def destination_check
+        @corporations.each do |corp|
+          # The capitalization method of unparred corporations is nil
+          next unless corp.capitalization == :escrow
+
```
tobymao, line 318: start_hex, end_hex = @destinations[cop.id]

lib/engine/game/g_1856.rb
```diff
@@ -309,6 +311,33 @@ def cert_limit
...
+          destinated!(corp) if hexes_connected?(@destinations[corp.id].first, @destinations[corp.id].last)
```
tobymao, line 319: goal_hex_ids is one hex or an array?

lib/engine/step/g_1856/track.rb
```diff
@@ -35,6 +35,12 @@ def old_paths_are_preserved(old_paths, new_paths)
 
           old_paths.all? { |path| new_paths.any? { |p| path <= p } }
         end
+
+        def process_lay_tile(action)
+          super
+          # TODO: Replace 'true' with @loading check & implement auto action usage
+          @game.destination_check # if true
```
tobymao, line 42: don't forget to do this or better to just comment it out for now

lib/engine/game/g_1856.rb
```diff
@@ -294,6 +295,7 @@ def create_destinations(destinations)
           dest_arr.each do |d|
             hex_by_id(d).original_tile.icons << Part::Icon.new("../logos/1856/#{corp}")
           end
+          @destinations[corp] = [end_b, end_a].freeze
```
tobymao, line 298: this seems to indicate that end_a and end_b are just single elements

lib/engine/game/g_1856.rb
tobymao: seems like it should just be start_hex and end_hex, but you have goal_hex_ids.include?(...)

ventusignis: It's an array to handle the 2-hex offboard destination (for CV iirc) in the alternate destinations for 1856

lib/engine/game/g_1856.rb
ventusignis: I replaced this with `if hexes_connected?(*@destinations[corp.id])` and that worked

lib/engine/game/g_1856.rb
ventusignis: It's because `node.walk` needs `visited` to be a hash

lib/engine/game/g_1856.rb
ventusignis: Ok, it's changed now

lib/engine/game/g_1856.rb
ventusignis: The cleanup of this method should make it more clear the second side could be an array

lib/engine/step/g_1856/track.rb
ventusignis: I'm going to see if I can get this to work in this PR

ventusignis: Adding the auto-action to handle destinating/escrow-release looks like it would be complicated enough to warrant a followup PR.

tobymao (approved):

PR #3690, opened 2021-02-04
Restore stockview in OR (for 1824) but corrected

perwestling:
The Stock view appeared by mistake in titles that had a redeem
possibility for a corporation. By restricting the view to
active players only it should just get the player options.


jenf (approved):
Approve, but want a glance from toby

tobymao (approved):

PR #3694, opened 2021-02-04
1860: relocate bankruptcy check

### LINKED ISSUE #3685: [1860] Insolvent company didn't go bankrupt during OR
[1860] Insolvent company didn't go bankrupt during OR

Game ID 25562

https://18xx.games/game/31337?action=710

- During Operating Round 8.3, it was S&C's turn to operate.
- S&C was Insolvent, market share price was £7 and thus, went bankrupt.
- Operating Round should have continued until it is completed.
- Next company that was supposed to operate was FYN.
- FYN was also Insolvent, market share price was also £7 and should have gone bankrupt also.
- However, operation of FYN was skipped and game went directly into Stock Round 9.

Played on mobile.
OS Android 10 (Huawei CLT-L04)
Chrome 88.0.4324.141 (32-bit)

screenshot game 25562

roseundy:
Move bankrupt corporation check from after_action to after the two places where a stock price can drop: dividends and selling shares.

This makes handling a bankruptcy that is not at the end of an OR more robust.

Fixes #3685 

A very small percentage of games may require pins.

### DIFF
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -579,10 +579,12 @@ module Engine
         super
       end
 
-      def action_processed(_action)
-        @corporations.each do |corporation|
-          make_bankrupt!(corporation) if corporation.share_price&.type == :close
-        end
+      def action_processed(_action); end
+
+      def check_bankruptcy!(entity)
+        return unless entity.corporation?
+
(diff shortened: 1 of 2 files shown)


roseundy:
It has the side effect of speeding up most actions.

tobymao (approved):

PR #3707, opened 2021-02-05
18 fl hotels

ventusignis:
image

### DIFF
--- a/assets/app/view/game/hex.rb
+++ b/assets/app/view/game/hex.rb
@@ -141,6 +141,15 @@ module View
         case @role
         when :map
           return process_action(Engine::Action::Assign.new(@entity, target: @hex)) if @actions.include?('assign')
+
+          # passing the entity to active_step yields EndGame as the active step for some reason, so we pass nil
+          step = @game.round.active_step(nil)
+          if @actions.include?('hex_token')
+            return if step.available_tokens(@entity).empty?
+
+            next_token = step.available_tokens(@entity)[0].type
+            return process_action(Engine::Action::HexToken.new(@entity, hex: @hex, token_type: next_token))
+          end
           return unless @actions.include?('lay_tile')
 
           if @selected && (tile = @tile_selector&.tile)
--- a/assets/app/view/game/hex.rb
+++ b/assets/app/view/game/hex.rb
@@ -142,8 +142,7 @@ module View
         when :map
           return process_action(Engine::Action::Assign.new(@entity, target: @hex)) if @actions.include?('assign')
 
-          # passing the entity to active_step yields EndGame as the active step for some reason, so we pass nil
-          step = @game.round.active_step(nil)
+          step = @game.round.active_step
           if @actions.include?('hex_token')
             return if step.available_tokens(@entity).empty?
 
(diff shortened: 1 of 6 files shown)


assets/app/view/game/hex.rb
```diff
@@ -141,6 +141,15 @@ def on_hex_click
         case @role
         when :map
           return process_action(Engine::Action::Assign.new(@entity, target: @hex)) if @actions.include?('assign')
+
+          # passing the entity to active_step yields EndGame as the active step for some reason, so we pass nil
```
tobymao, line 145: uhhh what? sounds like a hack, can you figure this out and do it properly?

lib/engine/action/hex_token.rb
```diff
@@ -0,0 +1,33 @@
...
+module Engine
+  module Action
+    class HexToken < Base
+      attr_reader :hex, :token
+
+      def initialize(entity, hex:, token_type: nil)
+        super(entity)
+        @hex = hex
+        # token may be nil because when you upgrade someone's 00
+        # and place their token, you pretend to be them and you may not have a token
```
tobymao, line 14: sounds scary can you explain

lib/engine/step/g_18_fl/token.rb
```diff
@@ -0,0 +1,58 @@
...
+        def actions(entity)
+          return [] unless entity == current_entity
+          return [] unless can_place_token?(entity)
+
+          ACTIONS
+        end
+
+        def can_place_token?(entity)
+          current_entity == entity &&
+            (tokens = available_tokens(entity)).any? &&
```
tobymao, line 21: !empty?

lib/engine/action/hex_token.rb
ventusignis: This comes from `Engine::Action::PlaceToken`. Git blame shows it relates to 1836jr
```
      def initialize(entity, city:, slot:, token_type: nil)
        super(entity)
        @city = city
        @slot = slot
        # token may be nil because when you upgrade someone's 00
        # and place their token, you pretend to be them and you may not have a token
        @token = @entity.find_token_by_type(token_type&.to_sym)
      end
```

lib/engine/step/g_18_fl/token.rb
ventusignis: Fixed

Also I did a regex search and I see `any?` used instead of `empty?` a fair bit throughout the codebase; would it be worth submitting a refactor PR that changes this for all base and alpha/beta/production games?

assets/app/view/game/hex.rb
ventusignis: I did some poking around and it looks like the "normal" way of invoking `active_step` is to pass nothing (which defaults `entity` to `nil` anyways) and the times that it is given a parameter it is given a `Company` so the `EndGame` actions are empty

```
      def actions(entity)
        return [] if entity.company?
```

tobymao: ok can you remove the comment and just call it the normal way? round.active_step

ventusignis: Oops! The push I made in the morning with the change failed on invalid password, it should be fixed now

lib/engine/action/hex_token.rb
tobymao: there's no reason to copy the comment

ventusignis: I removed the comment, thanks!

tobymao (approved):

PR #3713, opened 2021-02-05
18CZ - Added Ebuy

Zwergenpunk:
Added Ebuy and automatic player debt and penalty interest


assets/app/view/game/player.rb
```diff
@@ -127,6 +127,21 @@ def render_info
             h('td.right', @game.format_currency(@game.bidding_power(@player))),
           ])
         end
+
+        if @game.respond_to?(:debt)
```
tobymao, line 131: is this needed? how do other games do this?

lib/engine/action/payoff_all_debt.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
```
tobymao, line 5: this doesn't seem necessary

tobymao: i wonder if you accomplish this with special buy instead

assets/app/view/game/player.rb
Zwergenpunk: I only found in 1849 something similiar. There it is just not shown.
I'm not sure if we want to show it always or just if > 0 because its probably happening rarely that a player gets a debt.

lib/engine/action/payoff_all_debt.rb
Zwergenpunk: I could add 1 or 1 for Each player loan to the game which is just a placeholder for the normal payyoff loan action

assets/app/view/game/player.rb
tobymao: ok delete it

lib/engine/action/payoff_all_debt.rb
tobymao: not sure what that means but if it's simpler sure

Zwergenpunk: I went with the special buy way. That worked really well

assets/app/view/game/player.rb
Zwergenpunk: I'm not a fan of having hidden game states, but for now I removed it. We'll have to see how often in happens in a real game.

tobymao (approved):

PR #3719, opened 2021-02-06
store scroll position on tab change

### LINKED ISSUE #3454: All titles: Game log window should persist scroll location
Currently, if you scroll backwards in the game log, then switch to a different tab, and switch back to the game log, the log has repositioned itself back to the latest move.  This is a PITA when you are reviewing previous moves and you are moving between tabs (e.g. the Game, Map, and Spreadsheet tabs).

I am requesting that the game log window 'remember where it is' when you tab away and then tab back.

All titles would be great, but 1846 would be my top priority.

yzemaze:
resolves #3454

### DIFF
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -14,7 +14,7 @@ module View
       needs :follow_scroll, default: true, store: true
       needs :selected_action_id, default: nil, store: true
       needs :limit, default: nil
-      needs :scroll_pos, default: nil, store: true
+      needs :scroll_pos, default: nil
 
       def render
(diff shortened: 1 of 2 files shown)


assets/app/view/game/game_log.rb
```diff
@@ -14,6 +14,7 @@ class GameLog < Snabberb::Component
       needs :follow_scroll, default: true, store: true
       needs :selected_action_id, default: nil, store: true
       needs :limit, default: nil
+      needs :scroll_pos, default: nil, store: true
```
tobymao, line 17: if you're going to pass it in don't put store true here

PR #3731, opened 2021-02-07
18CZ added Selling ability to companies

Zwergenpunk:
In 18CZ a company can be sold from the corporation to the bank for face value

### DIFF
--- a/assets/app/view/game/abilities.rb
+++ b/assets/app/view/game/abilities.rb
@@ -8,9 +8,6 @@ module View
     class Abilities < Snabberb::Component
       include Actionable
 
-      needs :game, store: true
-      needs :user, default: nil
-      needs :selected_company, default: nil, store: true
       needs :show_other_abilities, default: false, store: true
 
       ABILITIES = %i[tile_lay teleport assign_hexes assign_corporation token exchange sell_company].freeze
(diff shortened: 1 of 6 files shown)


assets/app/view/game/abilities.rb
```diff
@@ -1,16 +1,19 @@
...
 require 'lib/truncate'
+require 'view/game/actionable'
 
 module View
   module Game
     class Abilities < Snabberb::Component
-      needs :game
+      include Actionable
+
+      needs :game, store: true
```
tobymao, line 11: this is not necessary if it's actionable

assets/app/view/game/abilities.rb
```diff
@@ -85,6 +89,18 @@ def render_actions
       def actions_for(company)
         @game.round.actions_for(company)
       end
+
+      def render_sell_company_button
+        sell = lambda do
+          process_action(Engine::Action::SellCompany.new(
```
tobymao, line 95: can other games do this? it seems like the action already exists

looks like 1860 allows this already, can you do it in the same way 1860 does it?

Zwergenpunk: In 1860 the player can sell the company. In 18cz the corporation can sell the company after it was bought from the player.

Zwergenpunk: And another reason is that in 18CZ it has to be possible at every possible moment, because the money could be needed  to buy a new train etc.

tobymao (approved):

PR #3741, opened 2021-02-08
18FL privates

ventusignis:
This implements the 4 privates for 18FL, including how the powers are used by closing them

This adds a "bull token" to compliment the "cheater token". Unlike the cheater token, a bull token does not occupy a "traditional" slot. If you use a "bull token" on an empty single-slot city, there will still be an empty slot.


lib/engine/ability/README.md
```diff
@@ -254,6 +256,10 @@ Modified station token placement
   effectively increasing the city's size by one. (See 18 Los Angeles's optional
   company "Dewey, Cheatham, and Howe" or the corporations which get removed in
   1846 2p Variant for examples). Default nil.
+- `bull`: Simlar to `cheater` except this token does not take a slot -
```
tobymao, line 259: can you rename bull to extra_slot or something

why did you name it bull?

lib/engine/action/place_token.rb
```diff
@@ -7,19 +7,21 @@ module Action
     class PlaceToken < Base
       attr_reader :city, :slot, :token
 
-      def initialize(entity, city:, slot:, token_type: nil)
+      def initialize(entity, city:, slot:, tokener: nil, token_type: nil)
```
tobymao, line 10: why do you need tokener? can't you just infer it when the company lays the token?

lib/engine/part/city.rb
```diff
@@ -7,15 +7,21 @@ module Engine
...
       def initialize(revenue, **opts)
         super
         @slots = (opts[:slots] || 1).to_i
         @tokens = Array.new(@slots)
+        # Bull tokens are tokens in a city that don't go in a city slot
+        @bull_tokens = []
         @reservations = []
       end
 
+      def total_slots
```
tobymao, line 21: do you need this? could you instead call slots with an arg?

lib/engine/ability/README.md
ventusignis: I named it bull because it made me think of the bull tokens in 18C2C where merger corps get a bull token they can place in tokened out cities this way. I'll rename it `extra_slot`

lib/engine/action/place_token.rb
ventusignis: `Action` doesn't have access to the game outside of `self.h_to_args(h, game)`, passing `tokener` seemed less intrusive than giving `Action` full access to the game

lib/engine/part/city.rb
ventusignis: I changed it to the way you suggested

tobymao (approved):

daniel-sousa-me:
The extra tokens were already implemented for 1870 :/
There was a PR with them, but Toby rejected it because he wanted me to do it another way (even though you did the same as I did...)

PR #3743, opened 2021-02-08
Initial programmed actions framework which progresses buys until floats

jenf:
Relates to #209

This is the initial version.
Next steps would be:
* Separate the logging to be per player
* Deactivate programmed moves at end of round (in a way that also works for 1817)
* Add the UI


tobymao:
lgtm

tobymao (approved):

PR #3748, opened 2021-02-08
[1822] Fixed the map, tiles and gray areas

magnusfalt:
* Fixed the S and the T tiles. They upgrade from white to yellow with a standard yellow city. In green they have a labeled tile.
* When upgrading hex D35 (S-tile) the town is destoyed and a standard yellow tile 57 will replace it.
* The change in track_node_path was to get the lanes to show correcly if you have mutiple lanes with terminal. The coordinate was hardcoded to center of the edge. There is an offboard city with 8 entrances, thats why i needed it to show correct.
* Changed the backgroundcolor of the minor token to black. Was very hard to see the minors which has floated agaist the reserved minor spots.


lib/engine/step/g_1822/track.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require_relative '../base'
+require_relative 'tracker'
+
+module Engine
+  module Step
+    module G1822
+      class Track < Track
+        include Tracker
```
tobymao, line 10: what is the point of this class?

magnusfalt: As it is now its just to include the Tracker module where i override the legal_tile_rotation? method to be able to remove a town from a tile when upgrading to a yellow city.

I still have a few things to do with tracklay and a few methods to override. I have the English chanel, the river crossings cost money when you build into it. So it cost money from both sides not only when you connect.

There a re few private companies that modify the track a well.

But yes as it is now i could have put the method in the tracker.rb in the track.rb class and removed the tracker.rb class.

PR #3753, opened 2021-02-08
store log position only when scrolled up at least one line

yzemaze:
Difference between scrollHeight and scrollTop is offsetHeight. Exact match would lead to problems, adding 20 (line-height) seems to work fine.

Log still does not scroll to the bottom when taking an action (== pre #3719 behavior).

### DIFF
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -218,7 +218,8 @@ module View
       change_anchor = lambda do
         unless route_anchor
           elm = Native(`document.getElementById('chatlog')`)
-          store(:scroll_pos, elm.scrollTop < elm.scrollHeight - 220 ? elm.scrollTop : nil, skip: true)
+          # only store when scrolled up at least one line (20px)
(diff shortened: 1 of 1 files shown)


assets/app/view/game_page.rb
```diff
@@ -216,7 +216,10 @@ def menu
 
     def item(name, anchor = '')
       change_anchor = lambda do
-        store(:scroll_pos, `document.getElementById('chatlog').scrollTop`, skip: true) unless route_anchor
+        unless route_anchor
+          elm = Native(`document.getElementById('chatlog')`)
+          store(:scroll_pos, elm.scrollTop < elm.scrollHeight - 220 ? elm.scrollTop : nil, skip: true)
```
ryandriskel, line 221: perhaps set 220 as a constant, so future us know why 220?

yzemaze: Nope, replacing it with offsetHeight right now, because I neglected chatlog height change by user …

tobymao (approved):

PR #3757, opened 2021-02-09
[1830] add fancy logos and fixup corp colors

benjaminxscott:
Adds logos for 1830 corporations [from this source repo](https://github.com/18xx-maker/18xx-maker/tree/master/src/data/logos/1830)

Updates corp colors to match logos (except for Erie, since its greenish-yellow is unappetizing)

Maintains text-only logos as "simple logos"

Screenshot 2021-02-09 at 7 10 26 PM

Screenshot 2021-02-09 at 7 10 42 PM

Spreadsheet:
Screenshot 2021-02-08 at 5 15 57 PM


tobymao:
possible to compress the logos a bit? look at the wiki cc @yzemaze

yzemaze:
Some of them are already present in other titles’ folders (18Chesapeake, 1846, maybe others). Please copy those whenever possible as they should be optimized already. Replace colors with text editor (not inkscape!) where necessary.

scottredracecar:
I am not sure if it is just the screenshot, but matching the svgs to the current colors would be quite nice.

benjaminxscott:
@scottredracecar - I'd agree with making changes to tokens, however perhaps we should have a separate PR for that ? 
(since that change would diverge from the canonical tokens/colors)

scottredracecar:
What is the reference for canon? These svg colors don't look canonical to me with respect to any version I am aware of. The current colors at least match the AH version well. The Lookout edition has totally different logos/colors for a few corps.

benjaminxscott:
@scottredracecar 
I was basing off these sources:

https://drive.google.com/drive/folders/1kS7sxewZjRStaHCiDcqaEpi9Q-YXQQzh
https://github.com/18xx-maker/18xx-maker/tree/master/src/data/logos/1830

Personally have no issue w/ modifying logos however we want to be readable, just wanted to check if there was a standing convention

How do you feel about 1036112 and the updated screenshots in the top comment?
This makes the Erie and CO brighter, and PRR darker, using the Erie/CO logos and colors from 1846 and PRR from 18Chesapeake

tobymao:
if you have shared logos, don't add them, just reuse them

scottredracecar:
Looks okay to me now. Thanks!

benjaminxscott:
> if you have shared logos, don't add them, just reuse them

It looks like the cases where we reference other games' logos are all variants (e.g. ChesaOTR), do we want to do that for a mainline game like this one?

Not against it, just want to make sure it doesn't make things more complex later (i.e. "hey i'd like to change this logo, whoops it's actually a different game's logo")

scottredracecar:
I think those logos are unlikely to change significantly

tobymao (approved):

PR #3759, opened 2021-02-09
1856: Step & Auto Action to check if corporations destinate

ventusignis:
To use `auto_actions` for escrow release I needed an Action that just signified which corporations destinated. To do this I created a general `Engine::Action::Auto` that just stores a string `details`


lib/engine/step/g_1856/escrow.rb
```diff
@@ -0,0 +1,32 @@
...
+    module G1856
+      class Escrow < Base
+        # Since *any* corporation could destinated after a given corporation does a tile lay, we need
+        # to check for *all* corporations, and since multiple corporations could destinate at once
+        # we need to be able to support multiple destinating at once
+        def auto_actions(entity)
+          [
+            Engine::Action::Auto.new(
+              entity,
+              details: @game.corporations.select { |c| @game.destinated?(c) }.compact.map(&:id).join('/')
```
jenf, line 16: Assuming it works, this would be better as a list

lib/engine/game/g_1856.rb
```diff
@@ -312,13 +312,8 @@ def cert_limit
...
       end
 
-      def destination_check!
-        @corporations.each do |corp|
-          # The capitalization method of unparred corporations is nil
-          next unless corp.capitalization == :escrow
-
-          destinated!(corp) if hexes_connected?(*@destinations[corp.id])
-        end
+      def destinated?(corp)
```
jenf, line 315: destinated seems weird. 
destination_connected?

lib/engine/action/auto.rb
```diff
@@ -0,0 +1,28 @@
...
+require_relative 'base'
+
+# The Auto action type exists for auto_actions
+# This is to be used when the action that is automatically being done
+# is a basic check or otherwise not worth creating an Action class
+# Details is a string.
+module Engine
+  module Action
+    class Auto < Base
+      attr_reader :details
```
jenf, line 12: I think I'd prefer this as a specific DestinationConnected with a list of corps

ventusignis: I renamed it to DestinationConnection as it's vague enough to be reused elsewhere :)

lib/engine/game/g_1856.rb
ventusignis: Renamed

lib/engine/step/g_1856/escrow.rb
ventusignis: It's a list now

lib/engine/step/g_1856/escrow.rb
```diff
@@ -0,0 +1,32 @@
...
+          [
+            Engine::Action::DestinationConnection.new(
+              entity,
+              corporations: @game.corporations.select { |c| @game.destination_connected?(c) }
+            ),
+          ]
+        end
+
+        def actions(_entity)
+          %w[destination_connection]
```
tobymao, line 22: constant

ventusignis: fixed

tobymao (approved):

PR #3766, opened 2021-02-09
18ZOO: Adding tests.

carlorusso:
Making the SR BuySell step a blocking when powers can be choose or zoo ticket companies can be sold.
Locking 9 and 12 par on green and brown phase.
Adding "near_family" logic.
Handling companies "Leprechaun pot of gold", "Too much responsibility"


lib/engine/game/g_18_zoo.rb
```diff
@@ -586,6 +586,81 @@ def unowned_purchasable_companies(_entity)
...
+      def it_s_all_greek_to_me
+        @it_s_all_greek_to_me ||= company_by_id('IT_S_ALL_GREEK_TO_ME')
+      end
+
+      def whatsup
+        @whatsup ||= company_by_id('WHATSUP')
+      end
+
+      def apply_custom_ability(company)
+        if company.sym.to_sym == :TOO_MUCH_RESPONSIBILITY
```
tobymao, line 645: why do convert these to syms? that's unnecessary

carlorusso: ok, but if I don't do it the rspec using this piece of code will fail (because is trying to compare a string with a symbol) :(
any suggestion?

tobymao: why don't you do company.nym == 'TOO_MUCH_RESPONSIBILITY' then?

carlorusso: sure, done

lib/engine/step/g_18_zoo/sell_company.rb
```diff
@@ -0,0 +1,43 @@
...
+    module G18ZOO
+      module SellCompany
+        def sell_price(entity)
+          return 0 unless can_sell_company?(entity)
+
+          entity.value
+        end
+
+        def can_sell_any_companies?(entity)
+          sellable_companies(entity).any?
```
tobymao, line 14: !empty?

lib/engine/step/g_18_zoo/buy_sell_par_shares.rb
```diff
@@ -27,6 +34,33 @@ def process_buy_company(action)
           super
 
           @game.available_companies.delete(action.company)
+          @game.apply_custom_ability(action.company)
+        end
+
+        def get_par_prices(_entity, _corp)
+          super
```
tobymao, line 41: one pass

super.reject do
 .. && ..

end

lib/engine/step/g_18_zoo/sell_company.rb
carlorusso: done in 4800b1d3

lib/engine/step/g_18_zoo/buy_sell_par_shares.rb
carlorusso: done in 4800b1d3

PR #3768, opened 2021-02-09
[1822] Destination tokens

magnusfalt:
* Fixed the border cost of the rivers.
* Fixed a bug if you upgrade a L train to a 2 train, it still considered as a local and could run only one station.
* Placing the destination token should now work for all corporations.


lib/engine/step/g_1822/destination_token.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          [destination_token]
+        end
+
+        def can_place_token?(entity)
+          return false if !entity.corporation? || (entity.corporation? && entity.type != :major)
+          return false if available_tokens(entity).empty?
+
+          destination_hex = @game.class::DESTINATIONS[entity.id]
+          parts = @game.graph.connected_nodes(entity).keys
```
tobymao, line 30: you cannot call this unless you hide it under a @game.loading check or you use auto actions

it's too expensive

lib/engine/step/g_1822/tracker.rb
```diff
@@ -8,6 +8,28 @@ module G1822
       module Tracker
         include Step::Tracker
 
+        def border_cost(tile, entity)
+          hex = tile.hex
+          types = []
+
+          total_cost = tile.borders.dup.sum do |border|
+            next 0 unless (cost = border.cost)
```
tobymao, line 16: why does border_cost method mutate tile.borders?

magnusfalt:
* Renamed border_cost to remove_border_calculate_cost! since it was manipulating the tiles it self by removing the border cost when its paid.
* In destination token, added a "pass" option and moved the call to @game.graph.connected_nodes wraped inside a unless @game.loading in the process_ method. Its more inline what the place token do. With tobys recommendation i will check out Jen's auto_action later to see if we can optimize this step a bit.

tobymao (approved):

PR #3769, opened 2021-02-09
1817: Add expandable full loan table

### LINKED ISSUE #3186: [1817x] Add loan schedule
Add to the info tab a way to visualise the loan tokens.

This is more or less straight forward in 1817, but in 1817NA and 1817WO it's messier. It becomes hard to calculate the interest rate when there is a corp with 10 loans in liquidation in 1817WO

jenf:
fixes #3186

Also change 1817NA and 1817WO to use common code for loan calculations

### DIFF
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -8,6 +8,7 @@ module View
       include Lib::Settings
 
       needs :game
+      needs :show_loan_table, default: false, store: true
 
       def render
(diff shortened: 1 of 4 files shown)


ventusignis:
Will this be extendable to support 18USA's high player count loan chart?
image

jenf:
Yep

tobymao (approved):

PR #3774, opened 2021-02-10
1828 Simple logos, flag for non-buyable IPO shares, and remove cert limit from spreadsheet if no cert limit

crericha:


crericha:
Renamed OS&H to OSH to better fit on the logo and name on the charter. This will require all 1828.Games games to be pinned.

tobymao (approved):

tobymao (approved):

PR #3781, opened 2021-02-10
18fl token cost

ventusignis:
In theory this is the pentultimate PR to alpha - all that's left after this is the private auction and testing this in not-hotseat before alpha

### DIFF
--- a/lib/engine/game/g_18_fl.rb
+++ b/lib/engine/game/g_18_fl.rb
@@ -111,6 +111,14 @@ module Engine
         @phase.status.include?('hotels_doubled') ? 20 : 10
       end
 
+      # TODO: implement
+      # How far is the city_or_town from one of the corporation's stations?
+      # This is by track, not as the crow flies.
+      # city_or_town is either a City or a Town Part
+      def distance_to_station(_corporation, _city_or_town)
+        1
+      end
+
       # Event logic goes here
       def event_close_port!
         @log << 'Port closes'
--- a/lib/engine/game/g_18_fl.rb
+++ b/lib/engine/game/g_18_fl.rb
@@ -120,11 +120,10 @@ module Engine
       #  and can use the exits on tiles, except....
       # The disjoint cities in Jacksonville are a problem though
       def distance_to_station(corporation, start_hex)
-        raise GameError, 'Cannot calculate token cost of Jax' if start_hex.id == 'H23'
+        raise GameError, 'Cannot calculate token cost of Jax' if start_hex.id == 'B23'
 
         goal_hexes = corporation.tokens.select(&:city).map { |t| [t.city.hex, t.city] }
         # # If there is a goal token in non-connected (gray) jax, then we need special logic.
-        # special_jax = jax_token && hex_by_id('H23').tile.color != :gray
         distance = 0
         visited_hexes = []
         start_hexes = [[start_hex, start_hex.tile.cities.first]]
@@ -137,7 +136,7 @@ module Engine
           start_hexes = []
           hexes_to_visit.each do |hex_c|
             visited_hexes << hex_c
-            hex_c.first.neighbors.each do |e, neighbor|
+            valid_neighbors(hex_c).each do |e, neighbor|
               # Ignore this neighbor if they don't connect to each other
               next unless hex_c.first.tile.exits.include?(e) && neighbor.tile.exits.include?((e + 3) % 6)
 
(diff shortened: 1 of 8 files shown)


lib/engine/game/g_18_fl.rb
```diff
@@ -111,6 +111,66 @@ def hotel_value
...
+      # How far is the start_hex from one of the corporation's stations?
+      # This is by track, not as the crow flies.
+      # start_hex is either a City or a Town Part. Cannot be Jacksonville
+      #
+      # NOTE: 18FL does not have forks in plain track, only junctions
+      # Also because of this, we only need to return the hexes visited
+      #  and can use the exits on tiles, except....
+      # The disjoint cities in Jacksonville are a problem though
+      def distance_to_station(corporation, start_hex)
+        raise GameError, 'Cannot calculate token cost of Jax' if start_hex.id == 'H23'
```
tobymao, line 123: make sure this method is not called when game is loading

raise GameError unless @game.loading

lib/engine/game/g_18_fl.rb
```diff
@@ -111,6 +111,66 @@ def hotel_value
...
+
+        goal_hexes = corporation.tokens.select(&:city).map { |t| [t.city.hex, t.city] }
+        # # If there is a goal token in non-connected (gray) jax, then we need special logic.
+        # special_jax = jax_token && hex_by_id('H23').tile.color != :gray
+        distance = 0
+        visited_hexes = []
+        start_hexes = [[start_hex, start_hex.tile.cities.first]]
+        until start_hexes.empty?
```
tobymao, line 131: how difficult is it to create a generic hex distance calculator in graph? i feel like you should be using graph here

ventusignis: It would require a bit more thought and effort; this one works in the case that only track is lawson track. Actually, after some testing it doesn't even work properly on Jacksonville's disconnected cities and I haven't found an obvious reason why yet either

lib/engine/game/g_18_fl.rb
ventusignis: I will have to add the calculated token cost to the Token action to make it work when the game is loading

lib/engine/game/g_18_fl.rb
ventusignis: I was able to fix the logic in this

tobymao: can we put this in graph?

lib/engine/game/g_18_fl.rb
tobymao: doens't the token action have a cost?

ventusignis: No, It uses the price of the token from the corporation. In the case of 1861/1867, when it is time to place a token, the cost is easily calculable from the newly tokened hex and all of the pre-existing tokens (it's x distance as crows flies)

```
      def args_to_h
        {
          'city' => @city.id,
          'slot' => @slot,
          'tokener' => @tokener&.id,
          'token_type' => @token&.type == :normal ? nil : @token&.type,
        }
      end
```

ventusignis: Is there anything preventing us from simply adding a mutable field to the Token action (cost) and modifying that in the step's logic?

lib/engine/game/g_18_fl.rb
ventusignis: I removed explicit references to 18FL, added comments, and moved it to graph

lib/engine/game/g_18_fl.rb
tobymao: nope, although why would you have to mutate it? can't you just create it with the right price?

ventusignis: I was able to put something together, let me know if that looks good

lib/engine/graph.rb
```diff
@@ -89,6 +89,69 @@ def reachable_hexes(corporation)
...
+    #
+    # This implementation assumes:
+    #  - Lawson Track
+    #  - No double dits
+    #  - OO tiles should work? NY tile works
+    def distance_to_station(corporation, start)
+      goal_hexes = corporation.tokens.select(&:city).map { |t| [t.city.hex, t.city] }
+      distance = 0
+      visited_hexes = []
+      start_hexes = if start.respond_to?(&:city?) && start.city?
```
tobymao, line 104: why? 

what is start?

can't you just do start.is_a?(City)

lib/engine/graph.rb
```diff
@@ -89,6 +89,69 @@ def reachable_hexes(corporation)
...
 
+    # How far is the start_hex from one of the corporation's stations?
+    # This is by track, not as the crow flies.
+    # start_cy is either a City or nil (plain track or town)
+    #
+    # This implementation assumes:
+    #  - Lawson Track
+    #  - No double dits
+    #  - OO tiles should work? NY tile works
+    def distance_to_station(corporation, start)
```
tobymao, line 100: can you just use node.walk or path.walk to generate all the possible paths

these assumptions you made shouldn't be necessary and it's weird that you implemented your own bfs

ventusignis: I couldn't figure out how `node.walk` or `path.walk` work or what the inputs do, but since they looked recursive I figured it used a DFS. 

Since I needed a BFS to find the shortest path I tried make a simple one for 18FL, which leads to the assumptions which are cases that aren't handled in the implementation

lib/engine/graph.rb
ventusignis: `start` is where to start searching from (the city or hex to be tokened {PlaceToken or HexToken))

I figured using .city? would be more canonical but I went and changed it to `.is_a?(Engine::Part::City)`

lib/engine/graph.rb
tobymao: i'm not satisfied with this implementation, i think we need to put some more thought into this to do in a more robust manner.

tobymao: let's chat on slack to figure out the best algorithm / approach to make this generically useful

ventusignis:
Determining shortest distance needs to be implemented in a significantly different way, closing this for now so to not pollute the PR list

PR #3786, opened 2021-02-11
1873: map and tiles

roseundy:
First check-in for Harzbahn 1873

I'm not happy with how narrow track is being rendered, but I'm not worrying about that right now.

Common file changes:
config/tiles: added red frame to 1873-specific tiles
game/base: added support for multiple home cities
UI::city_slot: added support for multiple home cities

1873_map_18xx


lib/engine/game/base.rb
```diff
@@ -1774,10 +1774,19 @@ def init_hexes(companies, corporations)
 
         reservations = Hash.new { |k, v| k[v] = [] }
         reservation_corporations.each do |c|
-          reservations[c.coordinates] << {
-            entity: c,
-            city: c.city,
-          }
+          if c.coordinates.is_a?(Array)
+            c.coordinates.each_with_index do |coord, idx|
```
tobymao, line 1778: how hard is it to just change coordinates to always be an array?

roseundy: It would mean either
1. editing the the config files for all other games, or
2. adding the conversion to game/base when the CORPORATIONS variable is set from data['corporations'] (and for MINORS too).

However, I'd also have to comb through the code base to see where *.coordinates is used and modify them.

roseundy: That being, said, I can simplify this to not use an if.

roseundy:
Updated game/base to simplify reservation code

PR #3787, opened 2021-02-11
Refactor teleport to be an activated ability

crericha:
### High-level change summary
- Move teleport to an activated ability, like token and tile lay abilities.
- Move track laying/token variables (e.g. num tiles laid, upgraded, hexes laid, and tokened) to round, so it can be shared across steps.
- Small bug fixes found while refactoring teleport

### Testing
I ran all games against the 02/09 database snapshot and there were no failures.

I hand tested 18Chesapeake, 1830, and 1828.Games with all the permutations I could think of. I'll leave it to our inventive player base to find any others.


ventusignis:
This might break the Terminal Company in 18FL. I'll be able to look at that case on friday most likely

crericha:
I tested the 'token' ability in 1846 and it worked as expected.

lib/engine/step/special_token.rb
```diff
@@ -31,27 +51,53 @@ def process_place_token(action)
...
         return unless entity&.company?
 
-        @game.abilities(entity, :token)
+        @game.abilities(entity, :token) do |ability, _company|
+          return ability
+        end
+
+        @game.abilities(entity, :teleport) do |ability, _company|
+          return ability if ability.used?
+        end
```
tobymao, line 100: let's return nil explicitly at the end here just to be clear

lib/engine/step/special_track.rb
```diff
@@ -70,41 +93,50 @@ def available_hex(entity, hex)
...
+      def abilities(entity, **kwargs, &block)
         return unless entity&.company?
 
         if @round.respond_to?(:just_sold_company) && entity == @round.just_sold_company
           ability = @game.abilities(entity, :tile_lay, time: 'sold', **kwargs, &block)
+          return ability if ability
+        end
+
+        %i[tile_lay teleport].each do |type|
+          ability = @game.abilities(entity,
```
tobymao, line 126: new line style

abilities(
  entities,
  x,
)

lib/engine/step/special_track.rb
```diff
@@ -70,41 +93,50 @@ def available_hex(entity, hex)
...
-        ability ||
-          @game.abilities(
-            entity, :tile_lay, time: %w[special_track %current_step% owning_corp_or_turn], **kwargs, &block
-          )
+        nil
       end
 
       def check_connect(_action, ability)
         hex_ids = ability.hexes
+        return if ability.type == :teleport
```
tobymao, line 139: is this method only called when game is not loading?

lib/engine/step/special_token.rb
crericha: Fixed.

lib/engine/step/special_track.rb
crericha: Fixed.

lib/engine/step/special_track.rb
crericha: Looks like it. I added "unless @game.loading" where it is called.

tobymao: it might be safer if this method always returns when @Game.loading then you don't put the onus on the caller

crericha: Agree. I didn't look closely enough to see all the method did was throw the game error. Fixed.

tobymao (approved):

PR #3788, opened 2021-02-11
18fl private auction

ventusignis:
It's a thing now.

I created an option for a game to use choos-ey buttons instead of a number box. Having to use the number box to choose between only 3 valid choices was annoying

image

### DIFF
--- a/lib/engine/step/g_18_fl/buy_cert.rb
+++ b/lib/engine/step/g_18_fl/buy_cert.rb
@@ -1,0 +1,167 @@
+# frozen_string_literal: true
+
+require_relative '../base'
+require_relative '../../action/par'
+
+module Engine
+  module Step
+    module G18FL
+      class BuyCert < Base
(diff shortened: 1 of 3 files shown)


lib/engine/step/g_18_fl/buy_cert.rb
```diff
@@ -0,0 +1,175 @@
...
+
+        def setup_auction
+          @bids.clear
+          @first_player = current_entity
+
+          refresh_bids
+        end
+
+        def resolve_auction
+          return if @bids.any? { |_e, b| b.nil? }
```
tobymao, line 143: change all the b.nil? to be !b
and you an change the lambda to be |_, bid|

ventusignis: I didn't realize `_` was a valid variable name for a lambda, I went and altered all of the lambdas in the file that were affected

PR #3791, opened 2021-02-11
full size tile manifest & tile selector improvements

yzemaze:
- tile manifest with reduced margins => mobile and 2nd window friendly
- edges/corners of maps/tile manifest: arcs/fans of tiles instead of full circles moved towards center
- tiles are rendered with less distance to clicked hex
- map: tiles cover seams instead of hexes (in most cases)
- darker drop-shadow => better differentiation from background

Mobile edge cases (360px, 320px) …

image.image

image.image

… and map examples:

image.image

image.image

image.image

### DIFF
--- a/assets/app/lib/radial_selector.rb
+++ b/assets/app/lib/radial_selector.rb
@@ -26,7 +26,7 @@ module Lib
         bottom: "#{bottom}px",
         width: "#{size}px",
         height: "#{size}px",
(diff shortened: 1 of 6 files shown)


dfannius:
Looks great! If it's straightforward, it would be nice if in the manifest, each upgrade possibility were rotated to match the rotation of the base tile. (I think this is already happening on the map.)

yzemaze:
Yep, it’s happening on the map, although just for available tiles. I’ll check.

yzemaze:
@dfannius 
image

PR #3796, opened 2021-02-11
Programmed actions: Automatically disable and log when used

jenf:
* Move to a player_log for things that should only be seen by a player (not yet filtered to the player)
* Automatically remove programmed actions on round change

### DIFF
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -281,9 +281,7 @@ module Engine
         end
         share = corporation.ipo_shares.first
         if can_buy?(entity, share.to_bundle)
-          action = Action::BuyShares.new(entity, shares: share)
-          action.user = 'programmed'
-          [action]
+          [Action::BuyShares.new(entity, shares: share)]
         else
           [Action::ProgramDisable.new(entity, reason: "Cannot buy #{corporation.name}")]
         end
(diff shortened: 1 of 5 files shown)


lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -281,7 +281,9 @@ def activate_program_buy_shares(entity, program)
         end
         share = corporation.ipo_shares.first
         if can_buy?(entity, share.to_bundle)
-          [Action::BuyShares.new(entity, shares: share)]
+          action = Action::BuyShares.new(entity, shares: share)
+          action.user = 'programmed'
```
tobymao, line 285: can we do this another way?
like programmed = true

jenf: Sure, do you want the JSON to change then?

tobymao: i wonder if this is necessary at all. i think people might not want it to be any different from a normal action. thoughts on just having it appear as an instant user action?

jenf: It feels like it might confuse people since it's instant

tobymao: i think it would be fine, thoughts about just trying that for now until people complain?

jenf: fair enough, will remove until someone complains

tobymao (approved):

Issue #792, opened 2020-06-20
18EU Special Rules Summary


scottredracecar:
**Rules**
http://www.deepthoughtgames.com/games/18EU/Rules.pdf

**Setup**

- [x] Each minor starts with a 2T (which uses up all 15 of the 2Ts)

- [x] **Initial Sale Round**

- The game begins with this round which takes the place of an auction, it is a kind of dutch auction
- PD goes first and selects a minor to sell (let’s call this player the auctioneer). Once that minor is sold, the next auctioneer is to the left round the table and so on. 
- PD does not actually move during this round. PD takes the first action in SR2.
- The auctioneer, decides whether to bid on the minor in multiples of 5 (min 105) or decline to purchase. If they bid, it is a normal round the table 
- If they do not wish to bid on the item, they can decline to bid, and the next player can bid. If we get back around the table to the auctioneer, the price drops 10 to 90 and the auctioneer can now buy for 90 or decline and repeat the process of dropping the price by 10 per round. Once the price is zero, the auctioneer must buy it for zero.
- After the current minor is sold, the turn passes to the next player after the auctioneer to auction another item.
- Players must always select an item on their turn because they may be able to win it for zero (no passing being auctioneer).

**Operating Rounds**

- [x] During OR1 only, each of the minor company may lay two yellow tiles.
- [x] Operating order is minors run in order then majors
- [x] Minors can’t upgrade
- [x] Red to red bonus (start and end in a red offboard) 0/10/20/30 per token in yellow/green/brown/gray (similar rule in 1836Jr)
- [x] Not allowed to run same hex twice with a train (Paris, B-V)
- [x] Minors always pay half
- [x] Add minor train limit of 2 until phase 4, then limit 1
- [x] The 120 mountains become 60 mountains after tile is laid on them (there is a token to mark this, but we can handle it with terrain cost)
- [x] Tokens are free to place (paid upon formation)
- [x] Move down for issue
- [x] Add trains definition (Scott), run N cities, add dots for free
- [x] The ports on the map are treated like towns (free stop)
- [x] Issue/Redeem from pool
- [x] Fix issue/redeem timing, happens at the end of the turn, issue until 50% in pool, redeem any in pool, both at CMV, stock price drops one per share issued, can't issue in first operation, may not both issue and redeem.
- [x] Issues lower value, may not issue in first or, may not issue and redeem
- [x] Return token to company if a city merges and two tokens in the same city.

**Forced Purchase/EMR**

[...]

PR #3800, opened 2021-02-11
18EU Initial Re-commit

ryandriskel:

This PR adds an initial implementation of 18EU with the auction and first operating rounds of minors supported.

### Changes in common code
**round/auction**: The input and place bid inputs were not coded into the view for minors, since 18Mag is a draft. They have been added. I also moved the lambda function into its own distinct function for readability. There is probably some refactoring that could be done here to reduce duplication of code, but I wanted to save that for its own PR.

**Auctioner / Passable Bid Steps**: It was necessary to add `|| bid.minor` to make the auction code generically support selecting minors. I wanted to eliminate having to worry about this in future auctions and thus created the `bid_target` function.

### Change to 18Mag
18Mag was the only other title implementing minors in the initial auction. However, 18Mag is a draft, thus the input and place bid inputs were not coded into the view at all. It was necessary to add the _auctioneer_ function as a way to indicate to the view that it should not show the newly supported inputs.

Screen Shot 2021-02-11 at 4 31 56 PM

Screen Shot 2021-02-11 at 4 30 26 PM

Screen Shot 2021-02-11 at 4 30 52 PM

Screen Shot 2021-02-11 at 4 29 25 PM


g3lutz:
Okay I am in.
I will read some more this afternoon and this weekend.

lib/engine/step/g_18_eu/modified_dutch_auction.rb
```diff
@@ -0,0 +1,229 @@
...
+          next_entity!
+        end
+
+        def resolve_bids
+          return unless @auctioning
+
+          target = @auctioning
+
+          return unless @active_bidders.one?
+          return unless @bids[@auctioning].any?
```
tobymao, line 207: empty

lib/engine/step/g_18_eu/modified_dutch_auction.rb
```diff
@@ -0,0 +1,229 @@
...
+        end
+
+        def starting_bid(_entity)
+          @base_starting_bid - @current_reduction
+        end
+
+        # TODO: Decrease minimum bid if all passed
+        def min_bid(entity)
+          return unless entity
+          return starting_bid(entity) unless @bids[entity].any?
```
tobymao, line 99: empty

tobymao:
svgs kinda big can they be compressed?

ryandriskel:
> svgs kinda big can they be compressed?

I have run them through https://vecta.io/nano ... there are still a couple larger ones due to complexity. These are the original complex logos and _could_ be simplified if necessary.

PR #3801, opened 2021-02-11
18CZ Fixed all abilities

Zwergenpunk:
Now all abilities of companies work together


lib/engine/step/g_18_cz/sell_company_and_special_track.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          @game.skip_default_track unless @game.purple_tile?(action.tile)
+
+          abilities = @game.abilities(action.entity, :tile_lay, time: 'any')
+          abilities.each(&:use!)
+        end
+
+        def hex_neighbors(entity, hex)
+          return unless (abilities = tile_lay_abilities(entity))
+          return if abilities.all? { |ability| ability.hexes&.any? && !ability.hexes&.include?(hex.id) }
```
jenf, line 50: any? -> !empty

lib/engine/step/g_18_cz/sell_company_and_special_track.rb
```diff
@@ -0,0 +1,85 @@
...
+          return if abilities.all? { |ability| ability.reachable && !@game.graph.connected_hexes(operator)[hex] }
+
+          @game.hex_by_id(hex.id).neighbors.keys
+        end
+
+        def potential_tiles(entity, hex)
+          return [] unless (abilities = tile_lay_abilities(entity))
+
+          abilities_for_hex = abilities.select do |ability|
+            ability.hexes&.none? || ability.hexes&.include?(hex.coordinates)
```
jenf, line 62: empty

lib/engine/step/g_18_cz/sell_company_and_special_track.rb
```diff
@@ -0,0 +1,85 @@
...
+          end
+
+          all_possible_tiles = abilities_for_hex.flat_map(&:tiles)
+
+          tiles = all_possible_tiles.map { |name| @game.tiles.find { |t| t.name == name } }
```
jenf, line 67: Can't this be merged with the step below and just use a select here?

lib/engine/step/g_18_cz/sell_company_and_special_track.rb
Zwergenpunk: ;) . I'm so used to c# linq any, I will make that mistake again and again

lib/engine/step/g_18_cz/sell_company_and_special_track.rb
Zwergenpunk: sure. Just overlooked

benjaminxscott:
After using a private's ability, it looks like it still shows the description - is there a way to denote that the power has been used?
image

benjaminxscott:
terrain discount s.b. `40` for large abilities, see https://github.com/Zwergenpunk/18xx/pull/2

Zwergenpunk:
Oh. Thanks i added all mountain hexes but did not change discount

PR #3811, opened 2021-02-12
hotkeys / keyboard shortcuts

### LINKED ISSUE #495: Add hotkeys
One of my favorite features of Board18 is the implementation of hotkeys. This makes navigating the screens easy. 18xx.games could benefit from hotkeys.

I propose the following which is just a first draft, but these (mostly) make sense to me.

History
- [⌘/Ctrl + ←] previous action
- [⌘/Ctrl + →] next action
- [⌘/Ctrl + b] jump to beginning
- [⌘/Ctrl + e] jump to end

Tabs
- [G]ame
- [P]layers
- [C]orporations
- [M]ap
- M[A]rket
- Trains/[P]hases
- [T]iles
- [R, "pRivate"]Companies
- [S]preadsheet
- T[O]ols

yzemaze:
resolves #495

- Implemented on game_page for tab change, chat-activation/-leave and map zoom only – for now. If this turns out ok, more shortcuts/hotkeys and even systemwide ones (h = home, p = profile, n = next game to act …) shouldn’t be a big deal.
- Shortcuts function only if `div#game` or `body` is focussed. There should be no interference with the chat/input. Downside: on some browsers one click is necessary to focus (almost? anywhere but button, inputs, chat and top menu bar) before shortcuts work as intended. `div#game` is set to `autofocus`, but this is a hack and doesn’t work reliably on all browsers. (While testing Chrome did fine, FF didn’t. ymmv)
- Titles of anchors and buttons do work as tooltips, although a dedicated tooltip-lib might be better suited for this in the long-run.
- Shortcuts aren’t underlined on game menu anchors (like in most programs), because it would look pretty ugly there. [If people insist on them, I don’t mind changing this.]
- Modifiers are catched, so there shouldn’t be any interference on OS-/browser shortcuts.
- If `search for text when you start typing` feature of browser is activated, hotkeys won’t work.
- I chose both a + k for Market, because k feels more natural to me (and it’s the only k as of now), but a was proposed by Scott in the linked issue.

In addition I refactored use of keyCode to key, because keyCode is deprecated. Added benefits: source code with key is way easier to grasp and problems on international keyboard locales are prevented [`-` => `slash` on a German layout e.g.].
I’m not sure why `code &&` in `if code && code == 13` e.g. might have been necessary previously, but my tests didn’t suggest any regressions.

keycode test just for the curious ones: https://keycode.info/

### DIFF
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -168,10 +168,15 @@ module View
     end
 
     def hotkey_check(event)
+      # 'search for text when you start typing' feature of browser prevents execution
+      # only execute when no modifier is pressed to not interfere with OS shortcuts
+      event = Native(event)
+      return if event.getModifierState('Alt') || event.getModifierState('AltGraph') || event.getModifierState('Meta') ||
+        event.getModifierState('Control') || event.getModifierState('OS') || event.getModifierState('Shift')
+
       active = Native(`document.activeElement`)
       return unless active.id == 'game' || active.localName == 'body'
 
-      event = Native(event)
       case event['keyCode']
       when 71 # g
         change_anchor('')
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -197,6 +197,7 @@ module View
         change_anchor('#tools')
       when 'c'
         `document.getElementById('chatbar').focus()`
+        event.preventDefault
       when '-', '0', '+'
         map = `document.getElementById('map')`
         `document.getElementById('zoom'+#{key}).click()` if map
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -241,14 +241,14 @@ module View
       }
 
       menu_items = [
-        item('Game'),
-        item('Entities', '#entities'),
-        item('Map', '#map'),
-        item('Market', '#market'),
-        item('Info', '#info'),
-        item('Tiles', '#tiles'),
-        item('Spreadsheet', '#spreadsheet'),
-        item('Tools', '#tools'),
+        item('Game', '', 'g'),
+        item('Entities', '#entities', 'e'),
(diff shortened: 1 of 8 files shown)


dfannius:
The hotkeys that I really want are -> and <- (or similar if that causes conflicts) for walking through history. I find it much less annoying to tap a key multiple times than to click a mouse or touchpad multiple times.

yzemaze:
> walking through history

If I had read the issue before starting yesterday (evening/night) and not just a few minutes ago, I would’ve already implemented them ;-) Next on my list.

yzemaze:
So preventing browser defaults works fine, but it begs the question if it’s ok to override default behavior of these keys. Especially left/right arrow are questionable, because they could be/are useful to horizontally scroll the map if it doesn’t fit the browser window. We could use arrow up/down instead for history navigation, but then there’d be no way to scroll vertically with the keyboard … Ad hoc it seems to be better to just use a modifier for the history buttons (ctrl should work, shift shouldn’t interfere even less with browsers and plugins) – even though this would need a slightly different approach and definitely more convoluted code.

assets/app/view/game_page.rb
```diff
@@ -147,7 +154,57 @@ def render
...
+
+    def hotkey_check(event)
+      # 'search for text when you start typing' feature of browser prevents execution
+      # only execute when no modifier is pressed to not interfere with OS shortcuts
+      event = Native(event)
+      return if event.getModifierState('Alt') || event.getModifierState('AltGraph') || event.getModifierState('Meta') ||
+        event.getModifierState('Control') || event.getModifierState('OS') || event.getModifierState('Shift')
+
+      active = Native(`document.activeElement`)
+      return unless active.id == 'game' || active.localName == 'body'
```
tobymao, line 178: don't use unless with multiple statements change to if

tobymao (approved):

PR #3813, opened 2021-02-12
18Mag: Fix corporation token lays.

crericha:


jenf (approved):

roseundy:
What changed? I believe this used to work.

roseundy:
Looks like this bug was introduced when teleport was refactored.

PR #3817, opened 2021-02-12
full text search & games menu

### LINKED ISSUE #762: clicking Next on last page of games_row => it disappears
open https://www.18xx.games/?finished=2000 (exact number is irrelevant, just needs to be bigger than actually available pages to show last page of games)
hit "Next" on finished games

=> game_row disappears

Works across browsers and for other categories, too.

PS: Usually there’s no "Next" button on the last page. But if you click Next in quick succession (think FPS-style) the effect is reproducible without URL-manipulation.

### LINKED ISSUE #782: Ability to search for game type/user (New Feature for the site)
I would like to request the ability to search/filter games by user or game.  Both would be very helpful in joining new games and to review old games.

### LINKED ISSUE #3595: Homepage: Sort finished games by date of last action rather than game ID
Currently, when scanning the homepage for what games have finished recently, the items are sorted in descending order by game ID. This prioritizes live games that were started more recently.

I think it would be better if this list was sorted descending by the date of last action instead (don't consider comments, just game actions).

This has two advantages:
1. Async games would get more equal homepage exposure in this list
2. The games shown on the list would cycle more because every game would be listed on the homepage upon completion and for a duration afterwards. Currently, old game ids may never show up on the first view of the homepage.

Similarly, the active games may be preferable to sort by "updated" date/time (rather than ID).

yzemaze:
finally resolves #782 and #4778, fixes #3595, fixes #762

Some things that are possible with full text search are listed in the Wiki: https://github.com/tobymao/18xx/wiki/FAQ#search If this turns out to be too limited it’ll get its own page. If you have questions, please ask.

For this too work easily games are split into sections (your: active, new, archive, hotseat – all/others’: active, new, archive) and only one of these is shown. Switching is easy and possible without page reloads …
image
… and from everywhere, because the top menu is sticky. (Game Menu is still able to scroll over it.) Hotkeys/Shortcuts would be possible of course. (I think) I managed to kill all other page reloads (login, profile, home etc.), too. (Snappier site!)
Search in finished games for 60 in titles only =>
image

There’s for sure some optimization to be done on DB-queries, triggers and the DB-migration script.
Limit is set to 12 for testing purposes only. In regards to performance upping this to 50-ish should be no problem at all, but this would result in more scrolling especially on mobile devices. => I’d just let users decide (in a range from 12 to 48), likely by storing the limit in localStorage to allow different values per device.

No search for hotseat games as of now – and I wouldn’t prioritize implementing it.

I used `scrollIntoView()` a lot, because `#anchors` turned out to be problematic. They just don’t work (well) with a sticky menu and caused problems with url-params, too.

Fire away!

### DIFF
--- a/public/assets/main.css
+++ b/public/assets/main.css
@@ -31,7 +31,7 @@ button, .button_link {
   box-sizing: border-box;
   text-align: center;
 }
-button.small {
+button.small, .button_link.small {
   padding: 0.2rem 0.5rem;
   margin-right: 0.5rem;
 }
--- a/public/assets/main.css
+++ b/public/assets/main.css
@@ -13,11 +13,6 @@ a {
 nav a {
   text-decoration: none;
 }
-nav li {
-  float: left;
-  margin: 0 0.5rem;
-  list-style: none;
-}
 
 button, .button_link {
   font-size: 14px;
@@ -113,7 +108,13 @@ input + label, select + label {
   margin-right: 1rem;
 }
 
+nav ul {
+  margin: 0;
+  padding: 0;
+}
 nav li {
+  float: left;
+  padding: 0 0.5rem;
   list-style: none;
 }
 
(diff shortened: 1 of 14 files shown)


public/assets/main.css
```diff
@@ -159,6 +172,36 @@ nav#game_menu {
   font-weight: bold;
 }
 
+nav .submenu {
```
tobymao, line 175: why did you add so much css

yzemaze: tl;dr because it’s easier, flexible and most of it plain necessary for the menus to function the way they should.

100 – 108: consolidation of nav menus (cf. changes to game_page). This could be moved back inline and added to navigation – would result in more and duplicated code. (… and of course even more so, if more menus would get implemented down the road.)
175 – 183 only apply if items are rendered in a submenu. games_link and menu_item are used for items in menu as well as submenu, adding props with this differentiation there wouldn’t work as easy as adding a few lines css.
185 – 200 is due to the menu differences depending on screen size. What’s displayed at > 600px gets moved into a submenu on <= 600px, hence the media queries.
202 – 204 looks like a vestige from concurrent profile review. I’ll check.

Some screenshots with opened menus:

600px logged in:
image

<= 600px logged in:
image

<= 600px logged in:
image

320px logged out:
image

320px logged in, additional menu rendered over game menu:
image

tobymao:
closing for now

PR #3827, opened 2021-02-13
Alternate spreadsheet

ventusignis:
I put together this PR because I find it hard to easily distinguish full pays from half pays and withholds in the OR history.

This PR adds, as an user setting, an alternate OR history that mimics rr18xx style (|x| for halfpay, [x] for withhold)

image

### DIFF
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -21,7 +21,6 @@ module Lib
     SETTINGS = {
       notifications: true,
       simple_logos: false,
-      alternate_spreadsheet: false,
       red_logo: false,
       bg: DARK ? '#000000' : '#ffffff',
       bg2: DARK ? '#dcdcdc' : '#d3d3d3',
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -21,6 +21,7 @@ module Lib
     SETTINGS = {
       notifications: true,
       simple_logos: false,
+      alternate_spreadsheet: false,
       red_logo: false,
       bg: DARK ? '#000000' : '#ffffff',
       bg2: DARK ? '#dcdcdc' : '#d3d3d3',
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -147,7 +147,7 @@ module View
             link_h = history_link(revenue_text,
                                   "Go to run #{x} of #{corporation.name}",
                                   hist[x].dividend.id - 1,
-                                  { 'text-decoration' => 'none' })
+                                  { textDecoration: 'none' })
             h(:td, props, [link_h])
           else
             h(:td, props, revenue_text)
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -122,36 +122,24 @@ module View
           []
         else
           or_history(@game.all_corporations).map do |x|
-            if @user&.dig('settings', 'alternate_spreadsheet')
-              alternate_or_history(hist, corporation, x)
-            else
-              standard_or_history(hist, corporation, x)
-            end
+            render_or_history_row(hist, corporation, x)
           end
         end
       end
 
-      def alternate_or_history(hist, corporation, x)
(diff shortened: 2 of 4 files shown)


assets/app/lib/settings.rb
```diff
@@ -21,6 +21,7 @@ module Settings
     SETTINGS = {
       notifications: true,
       simple_logos: false,
+      alternate_spreadsheet: false,
```
yzemaze, line 24: No need for an option. The dotted underlining doesn’t really work with linked actions anyway, so please revise this PR, get rid of dotted underlining and implement your changes without user choice.

assets/app/view/game/spreadsheet.rb
```diff
@@ -121,38 +122,84 @@ def render_history(corporation)
...
+          props = {
+            style: {
+              opacity: case hist[x].dividend.kind
+                       when 'withhold'
+                         '0.5'
+                       when 'half'
+                         '0.75'
+                       else
+                         '1.0'
+                       end,
```
yzemaze, line 154: One case-statement with two return values would be nice.

yzemaze (changes requested):

assets/app/lib/settings.rb
ventusignis: I removed the choice to use the old style

assets/app/view/game/spreadsheet.rb
ventusignis: I was able to pull that off

assets/app/view/game/spreadsheet.rb
```diff
@@ -16,6 +16,7 @@ class Spreadsheet < Snabberb::Component
       include Actionable
 
       needs :game
+      needs :user, default: nil, store: true
```
yzemaze, line 19: still necessary? (btw: why store?)

yzemaze:
I just checked out your branch and after seeing it live I’m not really convinced by using | | for half-pay. The visual difference to [ ] is just to minuscule. What do you think about using : : instead?
image

What’s your take on the underlined links in general? Any way we could get rid of the visual clutter but stil convey these are links to actions? It would look a lot cleaner …
image

ventusignis:
@yzemaze What if we removed the underlines and swapped in blue text?
The colons look good. I used |x| because that's what rr18xx used for half pays

yzemaze:
> What if we removed the underlines and swapped in blue text?
I hate blue text and it collides with opacity and user defined bg color.

> The colons look good. I used |x| because that's what rr18xx used for half pays
Ok, then apply : and please check the other comment re: `needs: user`.

daniel-sousa-me:
> I hate blue text and it collides with opacity and user defined bg color.

This. Plus using color to convey information is not color-blind-friendly (8% of male population)

I'm all for removing underlines. Underlining is generally considered bad typography.
I'm conflicted about which chars to use. I like the idea of being consistent with rr18xx, but I agree that the square brackets and the pipes are not very easy to distinguish in the first image. But maybe it becomes better without the underlines? (plus, there's a difference in opacity to help). I personally don't like the colons there aesthetically.

Since our unicode support is pretty consistent, we can also make use of that. Instead of using the ascii pipe, we could use some other unicode char that looks similar, but farther from brackets (longer, shorter, broken, double, thicker, thinner...)

yzemaze:
> I'm all for removing underlines.

Well, do we need to highlight the fact that these are links or could we just get rid of the underlines and rely on (generally savvy) spreadsheet users knowing this by now and others just discovering it? Maybe adding a title/tooltip to the header would help. btw: Linking round numbers to actions might be useful, too.

> Instead of using the ascii pipe, we could use some other unicode char that looks similar, but farther from brackets (longer, shorter, broken, double, thicker, thinner...)

Broken bar (U+000A6) looks indeed nicer than colon:
image

ventusignis:
I've uploaded changes
image

assets/app/view/game/spreadsheet.rb
ventusignis: No, removed

yzemaze (approved):

PR #3837, opened 2021-02-13
1873: start auctions

roseundy:
Add initial premium auction
Add start auction (draft) of mines and starting concessions
Tweak to map
Load "extended" minor and company data (to avoid new classes)
Created "dummy" private companies based on mine (minors) and RR corps
Floating RRs add all home tokens and sell IPO shares to market

Common code change:
UI::Company - allow overriding "PRIVATE COMPANY" label

### DIFF
--- a/lib/engine/step/g_1873/draft.rb
+++ b/lib/engine/step/g_1873/draft.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+require_relative '../base'
+
+module Engine
+  module Step
+    module G1873
+      class Draft < Base
+        attr_reader :companies, :choices
+
+        ACTIONS = %w[bid pass].freeze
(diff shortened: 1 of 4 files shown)


lib/engine/step/g_1873/draft.rb
```diff
@@ -0,0 +1,124 @@
...
+        def name
+          'Draft'
+        end
+
+        def description
+          'Draft Private Mines and Concessions'
+        end
+
+        def finished?
+          @companies.empty? || entities.all?(&:passed?) & @game.premium.zero?
```
tobymao, line 49: && ? 

put in parens order is not obvious

roseundy: Yep, typo that sorta works.

PR #3863, opened 2021-02-14
Programmed actions UI

jenf:
This is built on top of #3796, hopefully it'll update assuming that goes in


lib/engine/game/g_18_mag.rb
```diff
@@ -3,6 +3,9 @@
 module Engine
   module Game
     module G18Mag
+      def available_programmed_actions
```
ryandriskel, line 6: I don't think this is the right place for this.

jenf: No, I'll fix after Toby reviews

roseundy: Kinda confused why 18Mag is in here at all.

tobymao (approved):

lib/engine/game/g_18_mag.rb
jenf: 18mag does actually float corps

PR #3869, opened 2021-02-15
[1828] fix place name Copper county/country

benjaminxscott:
Based on the [live rules link](https://kanga.nu/~claw/1828/1828-Rules.pdf)


tobymao (approved):

perwestling:
I gave the wrong issue, 3869 instead of 3860. So ignore that,

PR #3893, opened 2021-02-15
Developing on Fedora

mavit:
This pull request makes some changes to allow the development environment to run on Fedora.

- Docker doesn’t yet support Control Group v2 so doesn’t run out of the box on Fedora.  Allow Podman to be used instead.
- By default, Podman runs as the user rather than as root, and maps UIDs within the container to different UIDs outside the container.  Apply this mapping when creating the db/data/ directory.
- Fedora use SELinux by default, which means we need to relabel some volumes to allow them to be accessed from within the containers.

This was sufficient for me to be able to run `make` and access http://localhost:9292/.  I haven’t tried running any of this on a non-Fedora system, so it’s probably worth doing that before merging.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -97,6 +97,15 @@ Compose documentation:
 - https://docs.docker.com/compose/
 - https://docs.docker.com/compose/compose-file/
 
+#### Can I use [Podman](https://podman.io/) instead of Docker?
+
+Yes.
+
+```
+make CONTAINER_ENGINE=podman …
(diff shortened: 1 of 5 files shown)


tobymao:
Is there a way to get docker to work on Fedora instead?

mavit:
It’s possible to disable cgroups v2 as a boot-time kernel option.  However, I understand that this will break rootless Podman containers, which I already use for other things, so it’s not a terribly appealing approach.

Issue #3890, opened 2021-02-15
1849: show in market when par prices activate


dfannius:
At the beginning of the game, the only pars available are L.68 and L.100. When the green phase is reached, the L.144 box becomes available, and when the brown phase is reached, the L.216 box becomes available. It would be nice to call these out in the market by shading them differently. The par boxes are already yellow, so we could make the L.144 one light green and L.216 a light brown. ([They're shaded in the physical game.](https://boardgamegeek.com/image/5984052/1849-game-sicilian-railways))

I don't know what other titles, if any, have this sort of mechanism; it would be useful to have a list of them to ensure that if this feature is implemented it's done in a way useful for other titles.

crericha:
1828.Games has this functionality. It's straightforward to do using the par groups.

dfannius:
Great, I'll do it myself with that as a model.

dfannius:
Implemented with #3900.

PR #3900, opened 2021-02-15
1849: Color-code par boxes in stock market (#3890)

dfannius:


lib/engine/game/g_1849.rb
```diff
@@ -42,9 +42,14 @@ class G1849 < Base
...
 
-      MARKET_TEXT = Base::MARKET_TEXT.merge(phase_limited: 'Can only enter during phase 16').freeze
+      MARKET_TEXT = Base::MARKET_TEXT.merge(phase_limited: 'Can only enter during phase 16',
+                                            par: 'Yellow phase par',
+                                            par_1: 'Green phase par',
+                                            par_2: 'Brown phase par').freeze
       STOCKMARKET_COLORS = {
         par: :yellow,
+        par_1: :green,
+        par_2: :brown,
```
jenf, line 52: Need to check this'll actually show up in par values. See 1867 buy_sell_par_shares.rb

dfannius: I followed the 1828 model, which doesn't have a `get_all_par_prices` method and seems to work fine. When I tested it, I was not able to use the green par box at the start of the game but was able to once we entered the green phase. I was mostly just copying from 1828 code so I'm not at all certain that that's sufficient; let me know if you still think it isn't.

crericha: I'll convert both games to use `get_all_par_prices`, rather than overriding `StockMarket`

crericha: The conversion isn't tied to dfannius's PR, but certainly will make it easier!

Issue #3093, opened 2021-01-06
18CO


elkrajico:
Game 18CO
Link: https://18xx.games/game/20628
On SR when anybody pass, it shows:
"Illegal action Blocking step Buy Shares cannot process action 814"
Refresh site is worthless.
So, we END the game.

IMHO. When I clone the game, i was able to pass without any error and normal finish the game.

jenf:
Probably impacted during the migration, loads fine now

elkrajico:
Still sam problem. 
"Illegal action Blocking step Buy Shares cannot process action 830"

jenf:
we have a fix in review, will probably need them to wake up 
#3089

dup of #3094

Issue #3903, opened 2021-02-15
[1849:K2S] Company Associated with RSA does not Correspond to the First Company in Order


Maxwell-Frank:
Something has broken with the randomization of company order and association with the RSA Private in Kingdom of Two Sicilies. 

In two games linked below, the company associated with the RSA private is not the first in line and is thus unavailable to be parred. 

https://gist.github.com/Maxwell-Frank/2a861778386491526aee52c51a4c1838

https://gist.github.com/Maxwell-Frank/05b31465b0a2eca78671ebc0634a2a03

tobymao:
@dfannius

dfannius:
Fixed with #3910.

PR #3910, opened 2021-02-16
Fix 1849-K2S RSA (#3903)

dfannius:
Also removes the "delay IFT" option from 1849-K2S.


dfannius:
Whoops, the commit message claims to fix #3093, but it is actually fixing #3093.

PR #3928, opened 2021-02-16
1870: Fix bug preventing token from moving "right" when hitting a wall

daniel-sousa-me:


jenf (approved):
Approved, but code is frozen.

jenf:
(can be considered a bug so fixed)

PR #3929, opened 2021-02-16
1870: Fix share selling when above 60%

daniel-sousa-me:
If a player owns more than 60% of a corp and they want to sell shares, they need to sell enough to comply with the 60% limit

### DIFF
--- a/lib/engine/step/g_1870/buy_sell_par_shares.rb
+++ b/lib/engine/step/g_1870/buy_sell_par_shares.rb
@@ -55,6 +55,10 @@ module Engine
           entity.cash >= bundle.price && redeemable_shares(entity).include?(bundle)
         end
 
+        def can_sell?(entity, bundle)
+          super && bundle.corporation.holding_ok?(entity, -bundle.percent)
+        end
+
         def process_buy_shares(action)
           super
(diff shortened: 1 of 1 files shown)


crericha:
I implemented this differently for 1828. Should we unify? I'll link the diff this evening.

daniel-sousa-me:
Does 1828 also have this rule? I thought you were able to sell shares as many shares as you wanted

daniel-sousa-me:
To make it clear with an example: Under some specific circumstances you're allowed to own 80% of a corp. Unlike other games (and like 1828), you're not forced to sell those extra 20% later. But in 1860 you can't sell 10%. If you want to sell any share on that corp, you must sell enough to comply with the usual 60% limit. Afaict in 1828 you can only sell 10% to stay with 70%.

crericha:
Thought this was for not forcing selling when above 60%. If for something
else, then disregard my comment.

tobymao (approved):

daniel-sousa-me:
> Thought this was for not forcing selling when above 60%

No, for that I used the `can_hold_above_limit?` option you created ;)
This is adding a constrain that I think is pretty specific to 1870.

PR #3930, opened 2021-02-16
1870: Change corporation colors so that they are closer to the print version

daniel-sousa-me:


jenf:
You'd need to add it to the color map as well... however, note that the code is presently frozen as we're trying to remove the json files. When it's done resubmit this once we've migrated.

daniel-sousa-me:
Aren't those just CSS colors?

jenf:
partially, but we generally replace them with the rgb in case we need to do math

daniel-sousa-me:
Ok, I'll redo this with more standard colors, then x)

jenf:
you can add whatever colours you want, just needs to add them to the mapping

PR #3940, opened 2021-02-16
Operating info fix

daniel-sousa-me:

PR #3941, opened 2021-02-16
Revert "Operating info fix"

tobymao:
Reverts tobymao/18xx#3940


tobymao:
nope, still broken

19148, 20864, 20794

PR #3942, opened 2021-02-16
[1822] Dividend and private companies

magnusfalt:
* Made a new action and a new ui page, Aquire Company. I started with the buy_company view, but i had to modify it with multiple respond_to?, after i while choose to make the new view instead of clutter down the buy company code to much. I assume all other 22 variants need this as well.
* With the private 2P you gain a train in the corporation. They dont count towards the trainlimit or as a train in the mandatory train check. With this train you can also choose a different dividend then the other train, meaning you can ie withhold with the 2P and payout with the others. You can have any combination of the (payout, pay half and withhold)
* Destination bonus, if a train runs a route with both home token and destination token. You get the destination tokens city value one more time. Only one train can have this bonus.
* Mail contracts if the corporation have a mailcontract or both. They take half base value of the start node and half base value of the end node and add it to the corporations treasury. If a corporation have both, they have to have valid routes on 2 different trains to get the bonus from both.
* E-train they can run on the same track as the other trains. They can have no distance restiction, but they only get revenue from cities where corporation have a token. This revenue is doubled, this will also double the destination bonus. One city can give 4x its revenue.
* Some minor bug fixes and clean up

### DIFF
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -101,7 +101,7 @@ module Engine
         'NER' => 3,
       }.freeze
 
-      # Theese trains doesnt count against train limit, they also doesnt count as a train
+      # These trains don't count against train limit, they also don't count as a train
       # against the mandatory train ownership. They cant the bought by another corporation.
       EXTRA_TRAINS = %w[2P P+ LP].freeze
       EXTRA_TRAIN_PULLMAN = 'P+'
@@ -226,10 +226,11 @@ module Engine
       end
 
       def compute_other_paths(routes, route)
-        routes
-          .reject { |r| r == route }
-          .select { |r| train_type(route.train) == train_type(r.train) }
-          .flat_map(&:paths)
+        routes.flat_map do |r|
+          next if r == route || train_type(route.train) != train_type(r.train)
+
+          r.paths
+        end
       end
 
       def crowded_corps
@@ -322,7 +323,7 @@ module Engine
 
       def must_buy_train?(entity)
         !entity.rusted_self &&
-          entity.trains.reject { |t| extra_train?(t) }.empty? &&
+          entity.trains.none? { |t| !extra_train?(t) } &&
           !depot.depot_trains.empty?
       end
 
@@ -377,7 +378,9 @@ module Engine
                     super
                   else
                     entity = route.train.owner
-                    stops.select(&:city?).sum do |stop|
+                    stops.sum do |stop|
+                      next 0 unless stop.city?
+
                       stop.tokened_by?(entity) ? stop.route_revenue(route.phase, route.train) : 0
                     end
                   end
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -101,6 +101,12 @@ module Engine
         'NER' => 3,
       }.freeze
 
+      # Theese trains doesnt count against train limit, they also doesnt count as a train
+      # against the mandatory train ownership. They cant the bought by another corporation.
+      EXTRA_TRAINS = %w[2P P+ LP].freeze
+      EXTRA_TRAIN_PULLMAN = 'P+'
+      EXTRA_TRAIN_PERMANENTS = %w[2P LP].freeze
+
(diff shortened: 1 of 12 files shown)


lib/engine/game/g_1822.rb
```diff
@@ -101,6 +101,12 @@ class G1822 < Base
         'NER' => 3,
       }.freeze
 
+      # Theese trains doesnt count against train limit, they also doesnt count as a train
```
tobymao, line 104: These trains don't

tobymao: they also don't count as a train

lib/engine/game/g_1822.rb
```diff
@@ -150,19 +181,64 @@ def can_run_route?(entity)
...
+      def company_bought(company, entity)
+        # On acquired abilities
+        # Will add more here when they are implemented
+        on_acquired_train(company, entity) if self.class::PRIVATE_TRAINS.include?(company.id)
+        on_aqcuired_remove_revenue(company) if self.class::PRIVATE_REMOVE_REVENUE.include?(company.id)
+      end
+
+      def compute_other_paths(routes, route)
+        routes
+          .reject { |r| r == route }
```
tobymao, line 230: routes.flat_map do |r|
  next if r == route || ...
  paths
end

lib/engine/game/g_1822.rb
```diff
@@ -223,6 +320,12 @@ def init_round
         stock_round
       end
 
+      def must_buy_train?(entity)
+        !entity.rusted_self &&
+          entity.trains.reject { |t| extra_train?(t) }.empty? &&
```
tobymao, line 325: trains.any? ...

lib/engine/game/g_1822.rb
```diff
@@ -259,6 +363,51 @@ def place_home_token(corporation)
...
+            acquire_private_company?(entity, company)
+        end
+      end
+
+      def revenue_for(route, stops)
+        revenue = if train_type(route.train) == :normal
+                    super
+                  else
+                    entity = route.train.owner
+                    stops.select(&:city?).sum do |stop|
```
tobymao, line 380: stops.sum do |stop| 
  next 0 unless stop.city?

lib/engine/step/g_1822/dividend.rb
```diff
@@ -1,23 +1,136 @@
...
         DIVIDEND_TYPES = %i[payout half withhold].freeze
-        include HalfPay
 
         def actions(entity)
-          return [] if entity.corporation? && entity.type == :minor
+          return [] if !entity.corporation? || (entity.corporation? && entity.type != :major)
 
-          super
+          @extra_train_choice ||= ''
+          actions = []
```
tobymao, line 15: actions = super

lib/engine/step/g_1822/dividend.rb
```diff
@@ -1,23 +1,136 @@
...
+          train = find_extra_train(current_entity)
+          return '' unless train
+
+          route = routes.find { |r| r.train == train }
+          "Choose #{train.name}'s dividend (#{@game.format_currency(route.revenue)}) if "\
+          'this is to be different from the other trains'
+        end
+
+        def choices
+          choices = Hash.new { |h, k| h[k] = [] }
```
tobymao, line 31: default array but none of your choices are arrays?

magnusfalt:
With acquire you dont have any price, you just get the company. I started to modifiy the buy_company view. I had to add a few respond_to?, one for the the map not to show in the operations.rb and not to have the "limit_width" set.

The call:
companies = @game.purchasable_companies.sort_by do |company|

i need it to be 
companies = @game.purchasable_companies(company).sort_by do |company|
I tried it, what i could see this should work bacause buy_company action is only set in can_buy? to somthing like !@game.purchasable_companies(company).empty? . But not in very game, so i was afraid i will break something here.

And need a respond_to for the render_input, here i wanted just a button not the whole BuyValueInput.

If its better to do the respond_to methods i can change it.

tobymao (approved):

PR #3943, opened 2021-02-16
Operating info fix

daniel-sousa-me:
Third time's a charm (I hope...)

### DIFF
--- a/lib/engine/game/g_1860.rb
+++ b/lib/engine/game/g_1860.rb
@@ -388,7 +388,7 @@ module Engine
       end
 
       def get_or_revenue(info)
-        info.dividend.kind == 'withhold' ? 0 : info.revenue
+        !info.dividend.is_a?(Action::Dividend) || info.dividend.kind == 'withhold' ? 0 : info.revenue
       end
 
       # OR has just finished, find two lowest revenues and nationalize the corporations
--- a/lib/engine/step/g_1860/dividend.rb
+++ b/lib/engine/step/g_1860/dividend.rb
@@ -27,7 +27,7 @@ module Engine
 
           entity.operating_history[[@game.turn, @round.round_num]] = OperatingInfo.new(
             routes,
-            action,
+            (@game.insolvent?(entity) ? @game.actions.last : action),
             revenue
           )
 
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -132,15 +132,16 @@ module View
 
       def render_or_history_row(hist, corporation, x)
         if hist[x]
-          revenue_text, opacity = case hist[x].dividend.kind
-                                  when 'withhold'
-                                    ["[#{hist[x].revenue.abs}]", '0.5']
-                                  when 'half'
-                                    @halfpaid = true
-                                    ["¦#{hist[x].revenue.abs}¦", '0.75']
-                                  else
-                                    [hist[x].revenue.abs.to_s, '1.0']
-                                  end
+          revenue_text, opacity =
+            case (hist[x].dividend.is_a?(Engine::Action::Dividend) ? hist[x].dividend.kind : 'withhold')
+            when 'withhold'
+              ["[#{hist[x].revenue.abs}]", '0.5']
+            when 'half'
+              ["¦#{hist[x].revenue.abs}¦", '0.75']
+            else
+              [hist[x].revenue.abs.to_s, '1.0']
+            end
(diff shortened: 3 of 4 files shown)


tobymao:
@roseundy please review

lib/engine/step/g_1860/dividend.rb
```diff
@@ -27,7 +27,7 @@ def process_dividend(action)
 
           entity.operating_history[[@game.turn, @round.round_num]] = OperatingInfo.new(
             routes,
-            action,
+            (@game.insolvent?(entity) ? @game.actions.last : action),
```
roseundy, line 30: Why are you doing this? An insolvent company has a withhold dividend action. Why is this needed?

lib/engine/game/g_1860.rb
```diff
@@ -388,7 +388,7 @@ def check_nationalize?
       end
 
       def get_or_revenue(info)
-        info.dividend.kind == 'withhold' ? 0 : info.revenue
+        !info.dividend.is_a?(Action::Dividend) || info.dividend.kind == 'withhold' ? 0 : info.revenue
```
roseundy, line 391: What problem is this solving? Why wouldn't a dividend not have a dividend action?

lib/engine/step/g_1860/dividend.rb
daniel-sousa-me: It has a phantom action that doesn't have an id

lib/engine/game/g_1860.rb
daniel-sousa-me: When the Dividend step is skipped (ie, no train) the OperatingInfo gets the previous action so that the links on the spreadsheet work (see the mentioned issue)

lib/engine/step/g_1860/dividend.rb
crericha: Maybe a dumb question, but why can't 1860 use the same logic as `Dividend::skip!`? Why is the phantom action needed here but not there?

roseundy: Because unlike normal dividend skip operations, there is a route being run for an insolvent company, but the corp is forced to withhold (and the amount isn't zero).

lib/engine/game/g_1860.rb
roseundy: But the amount isn't 0 in this case. It's the revenue the leased train runs for.

lib/engine/step/g_1860/dividend.rb
roseundy: Who is using the id on the actions stored in operating_history?

lib/engine/game/g_1860.rb
tobymao: you can jump to the action in the spreadsheet

daniel-sousa-me: How? Where is that action stored? It just looks like a dangling object

lib/engine/step/g_1860/dividend.rb
tobymao: meant to put this here, but you can jump to an action in the spreadsheet which is why it needs an id

roseundy: I think I'm OK with these changes then.

lib/engine/game/g_1860.rb
tobymao: @roseundy @daniel-sousa-me can you guys clarifiy this particular line? why is the dividend not an action::dividend, what type is it?

roseundy: It's whatever @game.actions.last is on line 30 of step/1860/dividend.rb

daniel-sousa-me: Yeah, it's the action that triggered a skip on the Dividends

daniel-sousa-me: This is not specific to 1860 at all. It is also changed in `step/dividend.rb`. 1860 just was problematic because I didn't notice it was completely bypassing the parent code

roseundy: Hmm, pretty sure I borrowed that code from a different game.

lib/engine/step/g_1860/dividend.rb
roseundy: What revenue gets reported on the spreadsheet?

roseundy (approved):

Issue #1638, opened 2020-09-10
Smart sound if a player is „on turn“


Herbert62:
When I have an opened main view in my browser, I recognize from time to time the colour switching to orange telling me „it is your turn“.
Would it be possible to play a smart sound in the browser window while switching the colour? Configurable in the profile view?

edfactor29:
Would be nice to make it a train whistle.

Illotus:
This would be great improvement.

zurfarosa:
I came here to make this request myself :)

scottredracecar:
We have webhook notifications now and those can be configured to make a sound. Please use that type of solution.

PR #3945, opened 2021-02-16
Optionally play a sound on change of active player

mavit:
Addresses #1638.

Caveats:
- The sound effect that plays when it is another player’s turn was intended as a placeholder for testing, but I grew to like it so I’ve left it in for now.  Nevertheless, we might want to replace it with something less intrusive, disable it by default, or even remove it entirely.
- The sounds do not play immediately if you happen to be reviewing the game history when another player takes their turn.
- I’ve never used Opal before and have barely used Ruby, so, when reviewing, please imagine that meme of the Labrador doing chemistry.


bridger15:
If it is not already build this way, I suggest a specific, more obvious/attention grabbing sound for when it is your own turn, and a less obvious (quieter?) sound when play passes to another player.

mavit:
> If it is not already build this way, I suggest a specific, more obvious/attention grabbing sound for when it is your own turn, and a less obvious (quieter?) sound when play passes to another player.

Agreed.  For the other player sound I had in mind the “ker-chunk ker-chunk” noise of a train as it moves along the rails, but I’m not sure where to find such as sound isolated from background noise.

tobymao:
closing for now

mavit:
> closing for now

May I ask why?

bridger15:
> > closing for now
> 
> May I ask why?

I'd also like to know.  This is a super useful feature for everybody playing live and I still don't understand what's taking so long to make it happen.  Should we make a bigger stink about it? Slip some $20s under the door? :P

dfannius:
> I still don't understand what's taking so long to make it happen.

It has not been a priority for anyone who is spending their free time contributing to the site. PRs are always welcome, of course.

> Should we make a bigger stink about it?

No.

PR #3967, opened 2021-02-17
spreadsheet tokens cleanup, opacity if operated, title added

### LINKED ISSUE #3958: Better way to tell which corporations already operated
This is particularly for 1817, but would apply to any game. In face-to-face play, it's easy to tell which corporations have already operated and which are still on deck - the operated corporation tokens are flipped upside down. It's not so easy to see that here, for example:

Screen Shot 2021-02-16 at 5 56 42 PM

It would be nice to have some sort of an indicators - greyed out, or outlined?

yzemaze:
cleanup
+ opacity 0.4 if operated
+ title for all tokens (order could be included if !operated, but I didn’t see much value)

closes #3958
also helps in some cases of #1122 (which is still not solved perfectly)

image
image

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -504,14 +504,14 @@ module View
       def render_operating_order
         round = @game.round
         order =
-          if @game.round.operating?
-            @game.round.entities.index(@corporation)
+          if round.operating?
+            round.entities.index(@corporation)
           else
             @game.operating_order.index(@corporation)
           end
 
         if order
-          m = round.entities.index(round.current_entity) if @game.round.operating?
+          m = round.entities.index(round.current_entity) if round.operating?
           span_class = '.bold' if order && m && order >= m
           [h(:div, { style: { display: 'inline' } }, [
             'Order: ',
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -140,6 +140,7 @@ module View
           props[:attrs][:title] = "#{corporation.name} has operated"
           props[:style][:opacity] = '0.4'
         end
+
         props
       end
 
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -111,6 +111,30 @@ module View
         end
       end
 
+      def operated?(corporation)
+        return unless @game.round.operating?
+
+        order = @game.round.entities.index(corporation)
+        idx = @game.round.entities.index(@game.round.current_entity)
+        order < idx if order && idx
+      end
+
+      def token_props(corporation)
+        props = {
+          attrs: {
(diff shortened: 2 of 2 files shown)


crericha:
This is a great add! For what's it worth, the opacity is hard for me to see when looking at a glance. I have to look closely and compare back and forth. Have you experimented with other techniques? Perhaps a border or color inversion?

yzemaze:
I didn’t. Borders aren’t ideal because we already use them to show ownership on map and there’s an issue asking to do the same on the market.
Additional logos are way too much work, but with [css-filters](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) one could invert colors:
image
image
Not that great. Adding a clip-path on top of opacity-change works better e.g.:
image
image

scottredracecar:
Wow, the clip path one looks quite nice. I don't think you need the opacity change if you do that one.

dfannius:
Agreed with Scott, plus reduced opacity doesn't work well with overlapping tokens like we get in a 2D market.

yzemaze:
> I don't think you need the opacity change if you do that one.

There are some diamond shaped logos (18CZ e.g.). If they have the same background as the stockmarket cell (gray = user defined!) you’d need another way to distinguish them (if operated), so I’d keep the opacity as is or we’d need another clip-path that works for all logo shapes …

crericha:
Agree with all. The diamond shape is fantastic. Great idea.

Why is the WC logo smaller in the example?

yzemaze:
> Why is the WC logo smaller in the example?

Because 1817 features 3 logo sizes afaik.

crericha-qualitrol:
> Because 1817 features 3 logo sizes iirc.

I think that's a bug. The logos should all be the same size AFAIK: https://boardgamegeek.com/image/3547529/1817.

dfannius:
The 3 logo sizes are intentional; they distinguish 2-share, 5-share, and 10-share corporations.

yzemaze:
It’s in the code and looks like it’s on purpose:
`width: "#{TOKEN_SIZES[@game.corporation_size(corporation)]}px"`

crericha:
> The 3 logo sizes are intentional; they distinguish 2-share, 5-share, and 10-share corporations.

Okay. Didn't know that was a feature. 

Is it just me or does the WC look pixelated on the entity charter? Also missing the text. 

image

yzemaze:
The outline looks blurry, but isn’t. https://18xx.games/logos/1817/WC.svg

tobymao (approved):

jenf:
The 0.4 opacity is a bit too light for me, probably 0.6 would be better

crericha:
> The outline looks blurry, but isn’t. https://18xx.games/logos/1817/WC.svg

The SVG is missing the text.

PR #3970, opened 2021-02-17
Link to BoardGameGeek in the Game Info section of the Info tab

mavit:
Put the BoardGameGeek link next to the existing “more info” link to the wiki.

If we don’t have a link to the wiki, linking to the Google front page seems unhelpful, so let’s not show the link in that case.

Since GAME_INFO_URL will always point to the wiki, make that clearer by renaming it to GAME_WIKI_URL.

### Motivation

Iʼd like to include these and future games in my online boardgame directory at <https://www.mavit.org.uk/boardgames-online/>.  If thereʼs metadata to scrape, I can set things up so that happens automatically.

### DIFF
--- a/assets/app/view/game/game_meta.rb
+++ b/assets/app/view/game/game_meta.rb
@@ -63,9 +63,33 @@ module View
       end
 
       def render_more_info
-        return [] unless @game.class::GAME_INFO_URL
+        return [] unless @game.class::GAME_WIKI_URL || @game.class::GAME_BGG_ID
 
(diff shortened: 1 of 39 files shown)


roseundy:
Why can't we just put the BGG link in the wiki?

tobymao:
there's a code freeze right now, also the wiki is fine for this

mavit:
> Why can't we just put the BGG link in the wiki?

My thinking was that, given the free-form nature of a wiki, it doesn’t make for a terribly reliable source for scraping from.  For example, some wiki pages feature links to more than one game on BGG, or link to a BGG forum post rather than a game page.

I felt a direct link was useful enough to be worth having, but maybe it does go against the minimal aesthetic of the site.

PR #3992, opened 2021-02-18
Migrate 1817, 1817NA, 1817WO, 1877, and 18USA to new structure 

michaeljb:
Also fix 1822 imports--why this threw errors for me after changing the 1817 files, I have no idea


tobymao:
beacuse of import order probably

PR #3993, opened 2021-02-18
Convert 1828.Games and 1849 to use par groups for phased par values

crericha:
This PR refactors the phased par value logic shared by 1828.Games and 1849, to instead use par groups. Par groups functionality was added after these games were implemented, but has has recently been added to both to display the phased par prices as different colors in the stock market.

Additional rationale for the refactoring is 1) data classes (stock_market) shouldn't contain logic and 2) unifies par price behavior with how other games handle variable par prices.

Tested 1828.Games and 1849 in hot seat games, up through the green phase, to verify the par price logic.

### DIFF
--- a/lib/engine/game/g_1828/game.rb
+++ b/lib/engine/game/g_1828/game.rb
@@ -965,7 +965,7 @@ module Engine
 
         def new_auction_round
           Engine::Round::Auction.new(self, [
-            Engine::Step::CompanyPendingPar,
+            G1828::Step::CompanyPendingPar,
             G1828::Step::WaterfallAuction,
           ])
         end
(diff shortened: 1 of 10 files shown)


jenf:
Nice :)

jenf (approved):

tobymao (approved):

tysen (approved):
This works fine but I question the wisdom of having a `par_prices` method in both `Game` and `StockMarket`. Why not just in `Game`?

crericha:
I agree having par prices only in the game class is a good improvement. The
intent here was to bring 28 and 49 in-line with 61, 67, CO, etc., so that
we didn't have two implementations of the same thing.

tysen:
Yeah, I agree. I meant that we ought to remove `par_prices` from `StockMarket` in a separate PR.

PR #3997, opened 2021-02-18
[1822] English channel and Minor 14

magnusfalt:
- Only one of the 2 tracks connecting Merthyr Tydfil and Pontypool can be used each OR.
- Lay a yellow tile on London counts as an upgrade. In the real game, this is a wooden token you flip.
- A corporation now can have more than one token in London.
- English channel, must have connection to France to be able to run train to english channel. France doesnt count as a stop.
- When upgrading english channel to brown, there are 2 lanes to france. Upgrade the france tile to a tile with 2 lanes as well.
- Fixed off board graphics to handle lanes.
- Exposed the parameter from city.place_token "check_tokenable: true" in the tokener.place_token method. I need to bypass the check if a corporation already have a token on a tile. In London you can have multiple tokens.
- Minor 14, this minor shares a city in london with another corporation. When placing home token you have the choise of any of the 6 cities in london and spot in that city will open up. When a major acquires minor 14 they have the option of replace its token or remove it. If they remove the token the newly opened slot will become permantent so other corporations can token there. If a major i phase 5+ choose to acquire minor 14 from the bidbox they have the option of placing a token in any of the cities in london they have connection to. This also opens up a new slot in the selected city.
- There a alot of special cases in the code just for minor 14. I tried to make it as clean as possible.
- London is a real mess in the gui, 6 cities, city revenue, upgrade cost, a label and location name. If minor 14 starts one of the cities also gets 2 slots. What i can test everything works, but there are lot of hit boxes to press and you need some precision to press the right one. Mobile user will have to zoom quite a bit :)

After this PR i will try to migrate to the new structure.

### DIFF
--- a/lib/engine/game/g_1822.rb
+++ b/lib/engine/game/g_1822.rb
@@ -110,11 +110,20 @@ module Engine
 
       LIMIT_TOKENS_AFTER_MERGER = 9
 
+      LONDON_HEX = 'M38'
+      ENGLISH_CHANNEL_HEX = 'P43'
+      FRANCE_HEX = 'Q44'
+      FRANCE_HEX_BROWN_TILE = 'offboard=revenue:yellow_0|green_60|brown_90|gray_120,visit_cost:0;path=a:2,b:_0,lanes:2'
+
       MAJOR_TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false }].freeze
+
+      MERTHYR_TYDFIL_PONTYPOOL_HEX = 'F33'
 
       MINOR_START_PAR_PRICE = 50
       MINOR_BIDBOX_PRICE = 200
       MINOR_GREEN_UPGRADE = %w[yellow green].freeze
+
+      MINOR_14_ID = '14'
 
       PRIVATE_COMPANIES_ACQUISITION = {
         'P1' => { acquire: %i[major], phase: 5 },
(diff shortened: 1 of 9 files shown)


lib/engine/step/g_1822/pending_token.rb
```diff
@@ -0,0 +1,31 @@
...
+          true
+        end
+
+        def process_place_token(action)
+          entity = action.entity
+          city = action.city
+          if action.city.tokened_by?(entity)
+            hex = city.hex
+            city_string = city.hex.tile.cities.size > 1 ? " city #{city.index}" : ''
+            raise GameError, "Cannot place token on #{hex.name}#{city_string} because #{entity.id} cant have 2 "\
```
jenf, line 19: can't

lib/engine/game/g_1822.rb
```diff
@@ -478,6 +532,25 @@ def acquire_private_company?(entity, company)
...
+        return unless city.hex.name == self.class::LONDON_HEX
+
+        # Save the extra token city index in london. We need this if we acquire the minor 14 and chooses to remove
+        # the token from london. The city where the 14's home token used to be is now open for other companies to
+        # token. If we do an upgrade to london, make sure this city still is open.
+        @london_extra_city_index = city.tile.cities.index { |c| c == city }
+      end
+
+      def after_lay_tile(hex, tile)
+        # If we upgraded lodon, check if we need to add the extra slot from minor 14
```
jenf, line 545: london

lib/engine/step/g_1822/minor_acquisition.rb
```diff
@@ -183,13 +197,32 @@ def process_merge(action)
...
+            if (!minor.owner || minor.owner == @bank) && minor.id == @game.class::MINOR_14_ID
+              # Trying to acquire minor 14 from the bank. You still have to have connection to london.
+              # You have a option to place a "cheater" token in one of the cities you have connection to.
+              # A small note, if a corporation already have a token in london it need a clear path to another node
+              # in london. This means you cant just acquire to get an exchange token if you dont have a valid
+              # path to another node in london.
+              # Try all 6 cities in london to see if there is atleast one connection
+              found_connected_city = false
+              connected_nodes = @game.graph.connected_nodes(entity)
+              @game.hex_by_id(@game.class::LONDON_HEX).tile.cities.each do |c|
```
jenf, line 209: How about use cities.any? rather than the break

lib/engine/step/g_1822/token.rb
```diff
@@ -20,9 +20,18 @@ def process_place_token(action)
               raise GameError, "Cannot place token on #{city.hex.name}, that hex is reserved for destination token. "\
                                'Please undo this move and place your destination token'
             end
+
+            if city.tokened_by?(entity)
+              hex = city.hex
+              city_string = city.hex.tile.cities.size > 1 ? " city #{city.index}" : ''
+              raise GameError, "Cannot place token on #{hex.name}#{city_string} because #{entity.id} cant have 2 "\
```
jenf, line 27: can't

jenf (approved):
There's a few 'cant' which should be can't, please correct on a subsequent pr

PR #4008, opened 2021-02-18
Add options to graph and node for 1873

roseundy:
1873 has special requirements for the Graph:
* RR corps have to ignore normal-gauge track when laying tiles and tokens
* private mines need to ignore tokens when laying lines and calculating connectivity to the state railroad track (normal-gauge)
* private mines need to trace track starting from their home hex even though they have no tokens

Engine::Graph now supports three new options: home_as_token, no_blocking, and skip_track to support this
Engine::Part::Node now has a skip_track option to support this

### DIFF
--- a/lib/engine/graph.rb
+++ b/lib/engine/graph.rb
@@ -107,7 +107,7 @@ module Engine
       end
 
       if @home_as_token
-        home_hexes =  Array(corporation.coordinates).map { |h| @game.hex_by_id(h) }
+        home_hexes = Array(corporation.coordinates).map { |h| @game.hex_by_id(h) }
         home_hexes.each do |hex|
           hex.tile.city_towns.each do |ct|
             hex.neighbors.each { |e, _| hexes[hex][e] = true }
--- a/lib/engine/graph.rb
+++ b/lib/engine/graph.rb
@@ -107,9 +107,8 @@ module Engine
       end
 
       if @home_as_token
-        @game.hexes.each do |hex|
-          next unless Array(corporation.coordinates).include?(hex.id)
-
+        home_hexes =  Array(corporation.coordinates).map { |h| @game.hex_by_id(h) }
+        home_hexes.each do |hex|
           hex.tile.city_towns.each do |ct|
             hex.neighbors.each { |e, _| hexes[hex][e] = true }
             nodes[ct] = true
--- a/lib/engine/graph.rb
+++ b/lib/engine/graph.rb
@@ -4,7 +4,7 @@ require_relative 'part/path'
 
 module Engine
   class Graph
-    def initialize(game)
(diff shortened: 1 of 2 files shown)


lib/engine/graph.rb
```diff
@@ -103,6 +106,17 @@ def compute(corporation, routes_only: false)
...
 
+      if @home_as_token
+        @game.hexes.each do |hex|
+          next unless Array(corporation.coordinates).include?(hex.id)
+
+          hex.tile.city_towns.each do |ct|
+            hex.neighbors.each { |e, _| hexes[hex][e] = true }
+            nodes[ct] = true
+          end
+        end
```
michaeljb, line 117: Instead of looping over all hexes, does this work?

```suggestion
        hex = @game.hex_by_id(corporation.coordinates)

        hex.tile.city_towns.each do |ct|
          hex.neighbors.each { |e, _| hexes[hex][e] = true }
          nodes[ct] = true
        end
```

michaeljb: LGTM other than this

michaeljb: Or if the corporation can have multiple homes (I think that's why you'd throw it in an `Array`?), would this work?

```suggestion
        hexes =  Array(corporation.coordinates).map { |h| @game.hex_by_id(h) }
        hexes.each do |hex|
          hex.tile.city_towns.each do |ct|
            hex.neighbors.each { |e, _| hexes[hex][e] = true }
            nodes[ct] = true
          end
        end
```

roseundy: Yes, but I did want to make it flexible enough to handle coordinates being an Array.

roseundy: Yes, I'll use that latter code.

michaeljb: Yeah I figured this was minor compared to other Graph perf stuff 🤷‍♂️

michaeljb: Oh whoops looks like the line I typed had an extra space after `home_hexes =`

roseundy: Crap. What I get for not running rubocop :-(

tobymao:
i will review later today

lib/engine/graph.rb
```diff
@@ -103,6 +106,16 @@ def compute(corporation, routes_only: false)
         end
       end
 
+      if @home_as_token
+        home_hexes = Array(corporation.coordinates).map { |h| @game.hex_by_id(h) }
```
tobymao, line 110: probably could be done with one loop
Array(cords).each do |hex_id|
  hex = @game.hex_by_id(hex_id)
  ...

tobymao (approved):

PR #4011, opened 2021-02-18
18CZ Fixed more bugs

### LINKED ISSUE #3994: [18CZ] Buying a corporation with token in same hex blocks game
https://gist.github.com/Zwergenpunk/d9fce0f6fee969cca1f8d2a84ebea26e

The step to remove the tokens is activated but the token in Praha is already removed.

Maybe regression bug? I think it worked 2 days ago

### LINKED ISSUE #4004: [18CZ] Company unable to cross-buy train
https://18xx.games/game/28590?action=259

The EKJ has no train and wants to cross-buy the 4T out of the BN, but for some reason this is not allowed.

Zwergenpunk:
closes #4004
closes #3994

### DIFF
--- a/lib/engine/step/g_18_cz/buy_corporation.rb
+++ b/lib/engine/step/g_18_cz/buy_corporation.rb
@@ -58,7 +58,8 @@ module Engine
           end
 
           remove_duplicate_tokens(entity, corporation)
-          if tokens_in_same_hex(entity, corporation)
+          tokens_to_clear = tokens_in_same_hex(entity, corporation)
+          if tokens_to_clear
             @round.corporations_removing_tokens = [entity, corporation]
           else
             move_tokens_to_surviving(entity, corporation, price_for_new_token: @game.new_token_price,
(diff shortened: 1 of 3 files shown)


lib/engine/step/g_18_cz/buy_corporation.rb
```diff
@@ -58,7 +58,8 @@ def process_buy_corporation(action)
           end
 
           remove_duplicate_tokens(entity, corporation)
```
benjaminxscott, line 60: do we still need to do the remove_dupe_tokens call if we're handling that below?
(notably we'd want to move the token price logic into here if we get rid of that function call)

Zwergenpunk: yeah the duplicate removes automatic if there are 2 tokens in 1 city.

lib/engine/step/g_18_cz/reduce_tokens.rb
```diff
@@ -6,6 +6,10 @@ module Engine
   module Step
     module G18CZ
       class ReduceTokens < ReduceTokens
+        def description
```
benjaminxscott, line 9: This doesn't seem to be showing up in the step description, in the testcase

PR #4019, opened 2021-02-19
[1846] Fix LSL invalid tile lay

crericha:
- Clear the selected tile when toggling an ability
- Verify a matching ability was found when a company lays a tile
- Remove LSL icons after using its ability
- Show available hexes to token after both tiles have been laid
- Minor performance improvement

@Zwergenpunk, this may also fix CZ.

**Please validate against all games before deploying in case their is an edge case I missed where its valid to not have an ability**

### DIFF
--- a/assets/app/view/game/abilities.rb
+++ b/assets/app/view/game/abilities.rb
@@ -57,7 +57,10 @@ module View
         companies.map do |company|
           props = {
             on: {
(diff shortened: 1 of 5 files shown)


tobymao (approved):

tobymao:
awaiting validation, reverted

PR #4024, opened 2021-02-19
Revert "Convert 1828.Games and 1849 to use par groups for phased par values"

tobymao:
Reverts tobymao/18xx#3993


tobymao:
awaiting validation, reverted

PR #4048, opened 2021-02-20
18SJ: Fix bug that caused nationalization to be dealyed (fixes #3437)

### LINKED ISSUE #3437: [18SJ] -- Nationalization Delayed
https://18xx.games/game/22906

Nationalization does not occur at the beginning of Phase 4, it happens after the subsequent corporation runs.  See timestamp for Phase 4.  This issue only happens with one scenario in that game, where the dividends were paid out and the private for 1/2 price train was used.  It did not happen if the dividends were withheld and the private was not used.

perwestling:
18SJ needed its own special train buy to add the needed checks.

If this break any game delete, pin or archive as usual. Cannot fix via migration as the bug means a corporation might gets to act even though it should have been closed.

Fixes #3437

### DIFF
--- a/lib/engine/game/g_18_sj.rb
+++ b/lib/engine/game/g_18_sj.rb
@@ -271,7 +271,7 @@ module Engine
           Step::G18SJ::BuyTrainBeforeRunRoute,
           Step::G18SJ::Route,
           Step::G18SJ::Dividend,
-          Step::SpecialBuyTrain,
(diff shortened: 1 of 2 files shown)


michaeljb:
I have pinned these games:

```
[21442, 21453, 21984, 22290, 23284, 24458, 25735, 27337, 27436] 
```

PR #4075, opened 2021-02-21
Fix for opal loaded multiple times in dev

magnusfalt:


michaeljb:
There's probably something more that can be done to keep deps from being recompiled every time other files are recompiled, but this is definitely better than having the file expand endlessly!

Thanks!

magnusfalt:
Yes this was just a quick fix to get the file to stop expand. One solution can be to have the combined file of deps and opal have a different name i e dependencies.js. When combining it wont combine with it self.

PR #4080, opened 2021-02-21
The player associated with an action must be the one to make it

### LINKED ISSUE #1558: [1889] port tile illegal placement

### LINKED ISSUE #1725: 1889: Cannot use Mitsubishi Ferry ability if not president of any corporation

michaeljb:
* if a player onws a private company with an ability that may be used out of
  turn, prevent the current player from using the ability (e.g., Mitsubishi
  Ferry and Dougo Railway in 1889)
* Master Mode can still override this

Doesn't add a new step for Mitsubishi Ferry, but does let another player activate the ability at the start of a corporation's OR turn, without that player using master mode.

[Fixes #1558] [Fixes #1725]


assets/app/view/game/actionable.rb
```diff
@@ -44,6 +44,11 @@ def check_consent(player, click)
         store(:confirm_opts, opts, skip: false)
       end
 
+      def valid_actor?(action)
+        @valid_actors = @game.valid_actors(action)
+        @valid_actors.map(&:id).include?(@user['id'])
```
tobymao, line 49: @valid_actions.any? { |actor| actor.id == @user['id'] }

one pass

tobymao (approved):

PR #4092, opened 2021-02-21
scroll to top & game_menu into view on game load / tab change

yzemaze:
Only downside to this minHeight-hack is the ever-present vertical scroll bar on desktops. A small price to pay for a consistent game menu position.

### DIFF
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -183,6 +183,7 @@ module View
       base = @app_route.split('#').first
       new_route = base + anchor
       new_route = base if @app_route == new_route
+      scroll_to_game_menu
       store(:app_route, new_route)
     end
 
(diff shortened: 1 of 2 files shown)


assets/app/view/game_page.rb
```diff
@@ -265,6 +274,9 @@ def menu
           role: 'navigation',
           'aria-label': 'Game',
         },
+        hooks: {
+          postpatch: scroll_to_game_menu,
+        },
```
yzemaze, line 279: There’s a problem with this hook. As far as I understood it hooks fire if something happens to "their" DOM node. So the postpatch hook here should fire after menu was patched. Yet, it fires e.g. on any use of a map controls button. Why?

tobymao: i think it happens anything is rerendered

tobymao (approved):

PR #4096, opened 2021-02-21
[1822] Issue, redeem and loans

magnusfalt:
**Changes to base code**
- View::Game::BuyTrains. A player cant go bankrupt, instead he/she must take a loan with 50% intrest. I added a small check if player must take a loan.
- Engine::Action::PayoffPlayerDebt. A new action to pay off the loan taken above. I couldnt use the current PayoffLoan action, since it requires a loan object, and only corporations can have thoose.
- View::Game::RoundStock. As an optional action in the stock round you can pay of the loan taken. Here I added an button connected to the payoff_player_debt action.
- View::Game::Player. Here I added a check if the player have a @game.player_debt(@player), if so show the current amount. I located that both 18CZ and 1849 can also use this method to show thier player dept on the player card. Not sure if there are more games with player dept. I did a @game.respond_to?(:player_debt), i can make it a base method and remove the respond_to here if we want to implement it for thoose other games.
- Engine::Game. Here I added a method company_status_str(_company), this mimic the corporation status_str(_corporation) but is used on the company card instead. The reason for this is when your not in the stockround you need to see the remaining unsold minors, concessions and privates in their correct order. You also need to see which minors, concessions and privates is in thier respectively bid boxes. I did this with the same status we use on the corporation for such things. I did this as a generic if any other games would like to use it.
- View::Game::Company. Implement of the small statusbar at the bottom of the company card.

**1822 specific**
- Added the last step in the operating steps, issue and redeem shares.
- If in the end of stock round, a player still have a dept. An intrest of 50% is added to its current dept.
- On the entities page, list the remaing minors, concessions and privates.
- On Scotts recommendation i've updated all the icons to simpler icons.
- Fixed some minor bugs


jenf:
72 files is a bit of a big change, could you make it a bit smaller (easiest one is putting the svg's as a separate PR)

assets/app/view/game/buy_trains.rb
```diff
@@ -53,7 +53,18 @@ def render_president_contributions
                               "#{@game.format_currency(share_funds_allowed)}.")
         end
 
-        if @must_buy_train && share_funds_possible < share_funds_required
+        must_take_loan = false
```
tobymao, line 56: i don't think you need to initailze this

assets/app/view/game/company.rb
```diff
@@ -80,6 +84,18 @@ def render
             width: '100%',
           }
 
+          status_props = {
```
tobymao, line 87: can you put this in the block so it's not initialize every time?

tobymao (approved):

PR #4097, opened 2021-02-21
[18CZ] Add original par price to corporation model, use when floating companies

### LINKED ISSUE #4073: [18CZ] Floated Corporations Should Receive 10x Price the Director's Share was Purchased at, not CMV
https://gist.github.com/Maxwell-Frank/684d8cc94f1a68fc88c2099e4cda3ae0

The president's Share of the BY was purchased for $120. An intervening sale dropped the price to $115, but the company should still receive $1200 when it floats. 

See clarification [here](https://boardgamegeek.com/thread/1882303/article/27421085#27421085).

Zwergenpunk:
Adds the original par value to the corporation (Maybe some other games need the property as well)

closes #4073

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -23,7 +23,7 @@ module Engine
     include Spender
 
     attr_accessor :ipoed, :par_via_exchange, :max_ownership_percent, :float_percent, :capitalization, :second_share,
-                  :type, :floatable
+                  :type, :floatable, :original_par_price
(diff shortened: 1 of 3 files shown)


tobymao:
why do you need this vs par_price?

Zwergenpunk:
par_price is changing when a player sells a share. (Because we have market value == par value) 
And the corp has to float with original_par value * 10.

PR #4103, opened 2021-02-21
18MEX remove P2's ability if Copper Canyon is laid [Fixes #2855]

### LINKED ISSUE #2855: 18MEX - track step blocks even though there's no valid track action

michaeljb:
I have a local migration for this that fixes 55/60 of the broken cases. The last 5 can't be fixed by that migration and are all finished anyway; they'll need more fixing anyway for the changes I have in progress for the [rest of the issues I'm working on](https://github.com/tobymao/18xx/issues?q=is%3Aopen+is%3Aissue+assignee%3Amichaeljb+label%3A%22ability+timing%22).

### DIFF
--- a/lib/engine/game/g_18_mex/game.rb
+++ b/lib/engine/game/g_18_mex/game.rb
@@ -1090,6 +1090,17 @@ module Engine
           super + Array(abilities(entity, :train_limit)).sum(&:increase)
         end
 
+        def action_processed(action)
+          case action
+          when Action::LayTile
+            return unless action.hex.id == 'F5'
+            return if p2_company.closed? || action.entity == p2_company
+
+            p2_company.remove_ability(p2_company.all_abilities.first)
+            @log << "#{p2_company.name} loses the ability to lay F5"
+          end
+        end
+
         private
 
         def handle_no_mail(entity, trainless: true)
--- a/lib/engine/game/g_18_mex/step/special_track.rb
+++ b/lib/engine/game/g_18_mex/step/special_track.rb
@@ -14,8 +14,7 @@ module Engine
 
             ability if ability &&
               entity.owner == @game.round.current_entity &&
-              @game.round.active_step.respond_to?(:process_lay_tile) &&
-              copper_canyon_hex_empty?(ability)
+              @game.round.active_step.respond_to?(:process_lay_tile)
           end
 
           def process_lay_tile(action)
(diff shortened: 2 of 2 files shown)


tobymao (approved):

perwestling:
It should be noted the the KCM&O private should be closed when ability is used. That is according to the information from the designer:
https://boardgamegeek.com/thread/535347/error-kcmo-private-company

perwestling:
@michaeljb This has also been clarified in the Wiki page (https://github.com/tobymao/18xx/wiki/18MEX#rules-clarifications) and fixed 2020-11-22 at 18xx.games.

perwestling:
Mea culpa....

You are right. The ability should be removed if someone has laid a tile in the Copper Canyon hex. 
So yes, this is a fine commit.

lib/engine/game/g_18_mex/game.rb
```diff
@@ -1090,6 +1090,17 @@ def train_limit(entity)
           super + Array(abilities(entity, :train_limit)).sum(&:increase)
         end
 
+        def action_processed(action)
+          case action
```
perwestling, line 1094: Should it not be super first, and then case?

perwestling (commented):
I am just wondering about a detail.

tobymao (approved):

lib/engine/game/g_18_mex/game.rb
michaeljb: Hmm maybe it should. I was aping how 18Chesapeake does `action_processed`, which also doesn't call `super`. In `Game::Base`, `action_processed` just checks for corporations to close, which doesn't apply to Chesapeake or MEX, so it'll work without this, but we'll run into trouble if `Game::Base#action_processed` gets more functionality put in.

PR #4104, opened 2021-02-21
[18CZ] only shows the cheapest private of each type

Zwergenpunk:
Only shows the cheapest private of each type. More like the real game. 

image

That breaks all games where user bought a more expensive private...like in my fixture =/

### DIFF
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -103,12 +103,12 @@ module View
 
           @selected_company = @step.auctioning if @step.auctioning
 
-          companies = @step.available.select(&:company?)
-          if @step.respond_to?(:tiered_auction?) && @step.tiered_auction?
-            companies.group_by(&:value).values.map do |tier|
+          if @step.respond_to?(:tiered_auction_companies)
+            @step.tiered_auction_companies.map do |tier|
               h(:div, { style: { display: 'table' } }, tier.map { |company| render_company(company) })
             end
           else
+            companies = @step.available.select(&:company?)
             companies.map { |company| render_company(company) }
           end
         end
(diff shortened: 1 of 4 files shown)


benjaminxscott:
Conveying some of the sentiments in other discussions, and agreeing with them - IMO we should list all privates,

(Maintaining the columnar format as implemented in this PR)

https://github.com/tobymao/18xx/issues/4107

tobymao:
@benjaminxscott @Zwergenpunk is this ready to mereg?

benjaminxscott:
IMO should edit to show all privates and while using  the columnar format in this PR

Zwergenpunk:
Ok we change to list all but grouped. Sounds reasonable

yzemaze:
From the manual:
> Each player may buy **any** one Local Railway for the stated price or pass.

Showing only the cheapest would therefore be wrong.

Zwergenpunk:
Sigh...ok i will change to same as 1828 just with revenue

benjaminxscott:
Thanks Zwergenpunk and thanks Toby for checking in

assets/app/view/game/round/auction.rb
```diff
@@ -104,16 +104,38 @@ def render_companies
...
+          if @step.respond_to?(:unavailable_companies) && @step.unavailable_companies
+            rendered_unavailable_companies = @step.unavailable_companies.map do |company|
+              render_company(company, false)
             end
+            unvailable = h(:div, [
+              h(:div, {
+                  style: {
+                    "font-weight": 'bold',
+                    "font-size": '1.1rem',
+                    "padding-top": '3rem',
```
yzemaze, line 126: There’s `.bold` in main.css, but maybe you just want h3 instead of div => bold + font-size of 1.1rem and it’s semantic, too.
Also please use camelCase for css-_attributes_: `paddingTop`.

Zwergenpunk: its removed either way but thanks ;)

PR #4116, opened 2021-02-22
[1822] Pullmans and full capitalisation

magnusfalt:
**Changes to base code**
- Engine::Train. Move the properties :name and :distance from attr_reader to attr_accessor. When you attach a pullman train to another train, that train doesnt count the towns as stops but count its revenue. Easiest way was just to modify the distance on the train. And the name was just to see which train you attached your pullman to.

**1822 Specific**
- Loan money are "outside" money, they doesnt affect bank money.
- After reading a bit more in the rules and bgg i changed they way you can acquire minor 14 from the bank.
- Override the discard_train step to added a check when you discard a train, only current owner of the corporation can do this.
- In the route step you now can choose to attach a pullman, if you have this private in the corporation, to another train. Giving it the power not count towns as stops.
- In phase 6, major companies now receives full cap when it floats. Rest of the shares is transferred into the market.

With this PR all base systems are now in place. Trains, mailcontract and pullmans from the privates is also done. 11 private abilities and 1 major corporation ability is whats left. And im guesnning some bug fixes :)


lib/engine/game/g_1822/game.rb
```diff
@@ -2696,9 +2729,11 @@ def remove_exchange_token(entity)
         end
 
         def take_player_loan(player, loan)
+          # Give the player the money. The money for loans is outside money, doesnt count towards the normal bank money.
+          player.cash += loan
+
           # Add intrest to the loan, must atleast pay 150% of the loaned value
-          loan += player_loan_intrest(loan)
-          @player_debts[player] += loan
+          @player_debts[player] += player_loan_intrest(loan)
```
jenf, line 2736: Typo interest

lib/engine/game/g_1822/step/minor_acquisition.rb
```diff
@@ -201,13 +201,13 @@ def process_merge(action)
               if (!minor.owner || minor.owner == @bank) && minor.id == @game.class::MINOR_14_ID
                 # Trying to acquire minor 14 from the bank. You still have to have connection to london.
                 # You have a option to place a "cheater" token in one of the cities you have connection to.
-                # A small note, if a corporation already have a token in london it need a clear path to another node
-                # in london. This means you cant just acquire to get an exchange token if you dont have a valid
-                # path to another node in london.
+                # A small note, if a corporation already have a token in london and no clear path to another node.
+                # They can still choose to acquire and still gets both options of place or exchange. We dont want
+                # to an extra check of conencted_nodes. This is very costly.
                 # Try all 6 cities in london to see if there is atleast one connection
```
jenf, line 207: Typo

lib/engine/game/g_1822/step/route.rb
```diff
@@ -0,0 +1,85 @@
...
+            return [] if !entity.operator? || @game.route_trains(entity).empty? || !@game.can_run_route?(entity)
+
+            @pullman_train ||= nil
+            actions = ACTIONS.dup
+            actions << 'choose' if !@pullman_train && find_pullman_train(entity)
+            actions
+          end
+
+          def attach_pullman
+            @orginal_train = @pullman_train.dup
```
jenf, line 20: Unless the detach has to be done immediately this won't work as the route step will be different in the next OR and member variables will be cleared.

Also this probably needs to take into account the case of train discards and ensure the Pullman gets removed.

magnusfalt: Yes i know, you have to attach the pullman the next or. But this how it should be done since you have the option to attach it to another train.

magnusfalt: The pullman detaches before you leave the route. So it should be fine.

jenf: Yep sorry didn't read closely enough

lib/engine/game/g_1822/step/route.rb
```diff
@@ -0,0 +1,85 @@
...
+            @pullman_train = @game.route_trains(entity)[action.choice.to_i]
+            @log << "#{entity.id} chooses to attach the pullman to the #{@pullman_train.name} train"
+
+            attach_pullman
+          end
+
+          def process_run_routes(action)
+            super
+
+            detach_pullman if defined?(@pullman_train) && @pullman_train
```
jenf, line 72: defined? shouldn't be necessary

jenf: Apart from this looks good to me.

magnusfalt: No you are correct. I still have some problem with not defining variables as you have to do in a compiling programming language :)

jenf: As a C++ programmer it doesn't sit well with me either :)

lib/engine/game/g_1822/step/discard_train.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/discard_train'
+
+module Engine
+  module Game
+    module G1822
+      module Step
+        class DiscardTrain < Engine::Step::DiscardTrain
+          def process_discard_train(action)
```
jenf, line 10: I'm a bit confused about why this is necessary

magnusfalt: In base DiscardTrain there is no check to see if the player discarding a train on a corporation is accually the owner. I got it when 2 corporations had to discard trains. Player 1 could discard a train in corporation 2 which had player 2 as an owner.

Didnt want to change the base code if this was intentional.

jenf: Odd, it should've been player 2's turn at that point.

magnusfalt: When you comes to DiscardStep all corporations with an excess of train will show. Even if its player 1's turn both corporations with excess train will show.

magnusfalt: I just added the check so when a player click on a train in the step he/she have to be the owner of that specific corporation

jenf: Can we separate the Discard train from this PR, I'd like to understand it more as it should only show one players trains.

magnusfalt: Sure

magnusfalt: The code for selecting which corps is crowded, this from base.

```ruby
      def crowded_corps
        @crowded_corps ||= corporations.select do |c|
          trains = self.class::OBSOLETE_TRAINS_COUNT_FOR_LIMIT ? c.trains.size : c.trains.count { |t| !t.obsolete }
          trains > train_limit(c)
        end
      end
```
There is nothing that stats it should be just for one entity

magnusfalt: And in DiscardTrain

```ruby
      def active_entities
        crowded_corps.take(1)
      end
```

jenf: Yeah, other games immediately interrupt one player at a time to discard trains. So one corp does it then next and so on.

Is 1822 rules different?

jenf: ahh, this should be a fix to the base class then and tested appropriately (and should be a different PR)

magnusfalt: Yep, i removed my overriden class.

jenf (approved):

PR #4123, opened 2021-02-23
Reorder train purchase operations, so that the train is purchased before events are triggered

crericha:
Fixes issue when the same corporation that buys the 12H and is closed by the Messina Earthquake.

Ran against the 2/21 db without any failures.


tobymao:
@crericha awesome thank you!

Issue #3981, opened 2021-02-18
[18CZ] autopass during the ISR if funds are insuficient


skizelo:
If someone doesn't have enough money to draft a local railway, the client still waits for them to Pass (their only legal move). I noticed it in this game here https://18xx.games/game/27953?action=26 where Silvergoose had $45 and the only remaining local costs $80.

benjaminxscott:
I have a branch for this after the feature freeze thaws (only bugfixes for now)
https://github.com/benjaminxscott/18xx/tree/18cz_autopass_draft

PR #4135, opened 2021-02-23
[18CZ] autopass draft if player doesn't have cash for any privates

benjaminxscott:
closes https://github.com/tobymao/18xx/issues/3981


lib/engine/step/g_18_cz/draft.rb
```diff
@@ -51,6 +51,11 @@ def finished?
         def actions(entity)
           return [] if finished?
 
+          unless @companies.find { |c| current_entity.cash < min_bid(c) }.nil?
```
jenf, line 54: It'd be better doing this as part of process_bid/process_pass (the duplicated code should be removed)

benjaminxscott: Could you help me understand, are `process_$action` called when those buttons are clicked?

(my limited understanding is we return an empty array of actions so the player is given no valid actions and their turn gets skipped)

jenf: Yeah they're what are called.
We've strayed away from doing state changes in the actions method unless there's no other choice as it should really be stateless.

tobymao: if @companies.any? { |c| ... }

benjaminxscott: Wouldn't this need to be in `actions()` since it is autopassing the player?
(i.e. it can't fire on the `process_` calls, since they won't be called)

tobymao: waiting for this to be cleaned up

jenf: it's probably good enough doing it this way. otherwise it'd need a full rewrite

tobymao (approved):

PR #4146, opened 2021-02-24
stock market: scroll grid only, horizontal and ordered legend

yzemaze:
Bank and legend stay in place while grid is scrollable horizontally:
image

Legend inline and ordered by appearance in market (left to right & top to bottom, reversed if not 1d). This should work better than before for most titles.

ante
image

post
image
image

### DIFF
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -266,7 +266,13 @@ module View
         }
 
         children << h(:div, props, [h(Bank, game: @game)].compact)
-        children.concat(grid)
+        grid_props = {
+          style: {
+            width: '100%',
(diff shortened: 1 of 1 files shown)


assets/app/view/game/stock_market.rb
```diff
@@ -275,29 +281,32 @@ def render
           types_in_market = @game.stock_market.market.flatten.compact.flat_map(&:types)
           .uniq.map { |p| [p, colors[p]] }.to_h
 
-          type_text.each do |type, text|
-            next unless types_in_market.include?(type)
+          legend_items = types_in_market.map do |type, color|
```
tobymao, line 284: needs a compact

tobymao (approved):

Issue #4143, opened 2021-02-24
1849 SMS unable to place token in Marsala


ChrisTheCat:
https://18xx.games/game/29606?action=233

In a 3 player game, the CTL was removed from the game during setup. Later, the AFG was trying to use the SMS power to teleport a tile/token to a coastal city hex ("all cities except Caltanisetta and Ragusa"). AFG was not offered the opportunity to place the token in Marsala, presumably because it was still "reserved" for CTL.

Note that the player didn't intend to place the token there, it's obviously not the best play, we just noticed it wasn't highlighted.

crericha:
There isn't a reservation for CTL. You would see it in the city.

Teleport powers must list the possible destination hexes. My guess is the Marsala hex isn't in the list. Should be a one line fix.

crericha:
https://github.com/tobymao/18xx/blob/fab8291ebf7356de7afa39325730cbbba5081e79/lib/engine/game/g_1849/game.rb#L598

PR #4165, opened 2021-02-26
1849 bug fixes

crericha:


crericha:
The 49 failures for can't sell share are expected. They are all trying to sell a corporation share for a corporation that hasn't operated yet.

tobymao (approved):

lib/engine/game/g_1849/game.rb
```diff
@@ -595,7 +595,7 @@ class Game < Game::Base
 
         AFG_HEXES = %w[C1 H8 M9 M11 B14].freeze
         PORT_HEXES = %w[a12 A5 L14 N8].freeze
-        SMS_HEXES = %w[B14 C1 C5 H12 J6 M9 M13].freeze
+        SMS_HEXES = %w[B14 C1 C5 E1 H12 J6 M9 M13].freeze
```
tysen, line 598: I don't understand why this hex was added. I think the only time this could possibly be used by the SMS is in some convoluted scenario involving the advanced rules for token purchasing, with some corporation closures involved.

dfannius: The issue that prompted that change was #4143. It seems legit to me, but I'm fairly new to 1849.

tysen: Oh, yeah. Makes sense now.

PR #4172, opened 2021-02-26
[1822] More privates

magnusfalt:
Sorry for the big PR, but this concludes the last privates. All base functionality is now done, except of one todo with minors bid box refilling. I will now do a few days of bughunting and try to get a live play before an alpha release.

**Changes to base code**
- Engine::Abilities. Moved the statement where start_count was set into a method. This method is called both from init_abilities and from add_ability. start_count is only used at one placed, company card when its bought/acquired by a corporation. It shows how many uses you have. This wasnt updated if you added abilities in the code.
- View::Game::Company. Made it so game.company_status_str is shown on the render_company_on_card method as well.
- Engine::Action:AbilityChoose. A new action, when you select a company ability in a view there now can be choices on that company.
- View::Game::Abilities. If we have the action ability_choose, when you select a company it looks for the method ability_choices(entity) to see if the selected company will have any choose buttons.
- View::Game::Abilities. Changed the order which the game.companies.select selected companies that can be used. Checked owner before actions_for. This is a great perfomance increase for 1822 since there is lots of companies.

**1822 Specific**
- Shortened the Private company names, this to be able to see how many uses a company have.
- Made the purple of the LR corporation a little brighter.
- The major corporation SECR now have its ability to lay a free token in the english channel.
- P5 English Channel. You have the option to do a teleport token into the english channel. If there are not slots, you then will have the option to upgrade the tile if a freespot will open up. This will be in green -> brown and brown -> gray. And then put down your teleport token. If you dont want to do any of that you then have the option to move an exchangetoken to available. Im using the new ability for this.
- P9 Midland and Great Northern Joint Railway. If held by a player you have the option during you turn in the stockround to declare that your holdings in cash doubles when determing the player order in the next stockround. If a player activates this ability, i use the company_status_str so the users can see that this is done. 
- P9 Midland and Great Northern Joint Railway. If held by a corporation its revenue changes with the phase color.
- P15 Highland Railway. When held by a player its revenue is 10 * current phase, and the same amount is paid to the company charter. Since companies cant hold money, i created a dummy player which hold the money. You can also see the amount with the company_status_str. When you acquire this company with a corporation. It gets the money accumulated on the "company charter" and it closes. It closes on phase 7 if not acquired before that, and money goes back to the bank.
- P16 Off-Shore Tax Haven. During the stock round you have the option of buying a share and put in the tax haven. Neither the share bought and the tax haven counts towards your cert limit. And the share doesnt count toward you owning 60% in a corp. The solution is to have another dummy player just to hold the share, and will recieve revenue when the corp is running its train.
- P17 Lancashire Union Railway. When a corporation owns this, you will have to option to select one concession, private company, or minor company from the relevant stack of certificates, excluding those items currently in the bidding boxes, and move it to the top or the bottom of the stack. Here i again use the ability_choose option with loads of buttons. Of all my plays this ability is hardly used, might be my group though. Anyway I feel solution is fine. I also put it so you only can use in the last step which is issue and redeem shares.
- P18 Cromford and High Peak Railway. Here you have the option to move an exchange token to available, or vice versa.


magnusfalt:
Some bugs found and fixed in a live play.

**Changes to base code**
- Engine::StockMarket. Included an extra check to see that we got an shareprice when moving up after sold out in the stockround. This because of the strange layout of the stockmarket in 1822. We got a row above but it can be nil.

**1822 Specific**
- Fixed so you can run the 24 minor to the town in the same hex.
- Fixed a null pointer check in company_status_str
- When you upgrade to a brown city, you have to use the city with the most exits.
- Make sure you stop on the issue/redem step if you have the LUR private even if you are sold out in shares.

magnusfalt:
And of course the alpha tag :)

assets/app/view/game/abilities.rb
```diff
@@ -101,6 +102,27 @@ def render_sell_company_button
 
         h(:button, { on: { click: sell } }, "Sell company (#{@game.format_currency(@selected_company.value)})")
       end
+
+      def render_ability_choice_buttons
+        ability_choice_buttons = @game.round.active_step.ability_choices(@selected_company).map do |choice, label|
+          label ||= choice
+          click = lambda do
+            process_action(Engine::Action::AbilityChoose.new(
```
tobymao, line 110: call it ChooseAbility

PR #4174, opened 2021-02-26
Fix render issue in certain 1817 bankrupcy end game cases

### LINKED ISSUE #4173: [1817] Simultaneous bankruptcy
This is an odd situation, that I manufactured to see how the implementation works. I think it's unlikely it ever happens in a real games.

The rules don't specify an order for cash crisis to be resolved, so I'd say they're happening simultaneously. It also states that if only one player is left in the game, they win. But naturally this implementation had to set an order for the cash crisis to be resolved. 

I was curious to find out what happened if all the remaining players in the game went bankrupt at the same time. Would the game end before resolving the last cash crisis giving a "random" player the victory? Or would it go through all the cash crisis before checking for the end game condition and show all players as bankrupt?

The answer was a bit underwhelming. It turns out the game just breaks.

https://gist.github.com/daniel-sousa-me/a3d5d0555e903d9b37e1546f22aefed9

jenf:
fixes #4173


tobymao:
please deploy

PR #3183, opened 2021-01-10
fix "when" config for 18Chesapeake and Off the Rails

michaeljb:
Time to put #2947 to real use!

Prevent using B&S and C-P outside of owning corporation's step track

[Fixes #1284]
[Fixes #1583]
[Fixes #2723]

-----

[Here's my local migration script](https://gist.github.com/michaeljb/18ffb09762aaccf10764b5b1c21ea48c). This evening I got communication with the server's db setup so I can use my machine's cores to migrate/validate. Here's how well it does with a fairly recent copy of the db:

## 18Chesapeake

2488 games

* 1707 broken
    * 255 pinned
    * 1452 migrated

## 18Chesapeake: Off the Rails

92 games

* 45 broken
    * 16 pinned
    * 29 migrated

PR #4191, opened 2021-02-27
updates to fix ability timing issues 

### LINKED ISSUE #1725: 1889: Cannot use Mitsubishi Ferry ability if not president of any corporation

### LINKED ISSUE #1793: Bug 1836Jr30 - System allows private company power for tile placement to be exercised out of operational order when two corporations are owned by one player.

### LINKED ISSUE #2105: 18MS: Tile laying private ability skips token lay.

### LINKED ISSUE #3783: [1830] Allowing President of owning corporation to lay the C&StL private tile during the stock round
https://18xx.games/game/26856

We are in SR4 and the game is presenting the President of the NYC with the option to lay the C&StL tile.  The tile would have to be laid in the NYCs turn in an OR.  Not sure when it appeared, but it is there now, as well.  Just noticed it this SR.

### LINKED ISSUE #3998: 1836Jr: Can use private company owned by one of my other corporations
Consider https://18xx.games/game/27484?action=278

In this game CGL is first to operate in the OR. The player that owns CGL also owns Nord. Nord owns the CdH ability.

It was now possible for the director of CGL to use Nord's CdH ability to lay tile and place token, and use Nord treasury to pay for it. I believe only Nord should be able to use its private.

michaeljb:
Many files touched are very minor changes, such as: company config `when` changes, updates to function call arguments, and a migration applied to fixtures. Substantive logic changes are only in these 6 files:
*  lib/engine/ability/teleport.rb 
* lib/engine/game/base.rb
* lib/engine/round/operating.rb
* lib/engine/step/special_track.rb
* lib/engine/step/tracker.rb
* lib/engine/game/g_18_mex/game.rb and lib/engine/game/g_1828/game.rb each have small changes to make sure that minors don't block on the tile laying step if there is a tile-laying company to buy, since minors are not allowed to buy companies in that game (this could probably be refactored out to a game class config constant)

My local migration script is the one used in #3183 with a few more tweaks.

-----

* remove unused ABILITIES constant from abilities view

* update README

* remove ability times `special_track` and `other_or`
    * abilities previously using `special_track` now use `owning_corp_or_turn`
    * `other_or` was really a placeholder to avoid breaking games where
      abilities had been used out of turn; now it's gone and we can migrate/pin
      the affected games

* make all `teleport` abilities have `@when` of `['track']` and `@passive` of
  `false`

* make ability_right_time always (unless `strict_time` is false) check if
  `owning_corp_or_turn` (and others) are correct, in order to properly
  disqualify abilities belonging to corporations that are not active
  [Fixes #1793] [Fixes #3783] [Fixes #3998]

    * add param `strict_time` so that the 1846 Assign step for Steamboat Company
      can avoid an infinite loop; if `strict_time` is false, then a time that is
      passsed to the abilities call that matches an ability's time won't be
      further validated; the default strict behavior is to make sure the time
      makes sense, i.e., if the caller gives a time of `owning_corp_or_turn`,
      with strict checking, the code makes sure it actually _is_ the OR turn of
      the owning corporation

* don't pass on Tracker steps if there is a private company that the active
  corporation could buy in order to lay track during this step [Fixes #2105]

* I have a local migration ready that fixes over 3400 games broken by this
  branch, leaving only ~130 to pin/archive; the migration has been applied to
  all relevant spec fixtures

Specific company ability changes:

* 1817 family

    * Pittsburgh Steel Mill / Denver Telecommunications description says "during
      tile-laying", so set `when` to `track` instead of `owning_corp_or_turn`

* 1889

    * fix Mitsubishi Ferry; allow it to work in stock rounds, on the owning
      player's OR turns, and "in between" OR turns, but not after a corporation
      owned by a player has acted on its OR turn [Fixes #1725]
        * add ability times `or_between_turns` and `stock_round`
        * Round::Operating - add attr `current_operator_acted` to enable new
          ability timing of `or_between_turns`

* 18AL

    * restrict BLC to `when: 'track'`, matching its text

* 1MEX

    * restrict KCMO to `when: 'track'`, matching its text


michaeljb:
Keeping this as a draft; in some manual testing I've encountered a problem with the 1846 steamboat Assign step that for whatever reason isn't breaking the assets spec.

michaeljb:
> Keeping this as a draft; in some manual testing I've encountered a problem with the 1846 steamboat Assign step that for whatever reason isn't breaking the assets spec.

Got that fixed and rebased, removing draft status now

tobymao (approved):

PR #4215, opened 2021-02-28
Bold the most recent stock action

### LINKED ISSUE #1399: Stock Rounds: Add styling to most recent action
**Issue:** I often find it difficult to keep track of what everyone is doing during (later) stock rounds, because it's mostly text that updates.

**Suggestion:** I'd love if, during the stock rounds, there was some styling applied to the buys/sells from the most recent person to act. This would make it much easier to see what actions are taking during the Stock Round. The log at the top of the screen is helpful, but it's not where my eyes are always focused. (My eyes focuses on the companies & players.)

**Example:** Here's an example of this idea. I used bolding to show the most recent buy/sells by a player:
image

jenf:
fixes #1399

This is mostly so I can implement the auto-pass functionality which needs to track if other players have sold.

There's a slight issue to tidy up in a subsequent PR in that it doesn't show what you did last turn, but I think that's best done separately.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -324,15 +324,16 @@ module View
         result.index('.') ? result : "#{result}.0"
       end
 
-      def render_shares
-        player_info = @game.players.map do |player|
+      def entities_rows(entities)
+        entity_info = entities.map do |entity|
           [
-            player,
-            @corporation.president?(player),
-            player.num_shares_of(@corporation, ceil: false),
-            @game.round.active_step&.did_sell?(@corporation, player),
-            !@corporation.holding_ok?(player, 1),
-            player.shares_of(@corporation).any?(&:double_cert),
+            entity,
+            @corporation.president?(entity),
+            entity.num_shares_of(@corporation, ceil: false),
+            @game.round.active_step&.did_sell?(@corporation, entity),
+            @game.round.active_step&.last_acted_upon?(@corporation, entity),
+            !@corporation.holding_ok?(entity, 1),
+            entity.shares_of(@corporation).any?(&:double_cert),
           ]
         end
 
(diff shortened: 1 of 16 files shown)


lib/engine/game/g_1824/game.rb
```diff
@@ -1289,6 +1289,7 @@ def exchange_coal_railway(company)
           end
 
           float_corporation(regional) if regional.floated?
+          regional
```
michaeljb, line 1292: ?

assets/app/view/game/corporation.rb
```diff
@@ -342,39 +343,34 @@ def render_shares
...
+      def render_shares
+        shares_props = {
+          style: {
+            paddingRight: '1.3rem',
+          },
+        }
+
+        player_rows = entities_rows(@game.players)
+
+        other_corp_rows = entities_rows(@game.corporations.reject { |c| c == @corporation })
```
michaeljb, line 373: yay for moving repeated code into a function 🎉

lib/engine/game/g_1824/game.rb
michaeljb: oh never mind I see it now in the step logic

michaeljb:
I see a couple `process_*` actions that don't call `track_action` in some games' `BuySellParShares`, e.g., 1860's `process_buy_company` and `process_sell_company`, and 1817WO's `process_choose`. Is that an oversight, or is it intentional and there's some way that you've determined which actions need to call `track_action` and which ones don't?

jenf:
I'll double check, I probably didn't bother doing them because they relate to buying companies.

jenf:
I didn't do choose since it's a consequence of process_par and so adding another thing didn't seem necessary

michaeljb (approved):

PR #4221, opened 2021-03-01
[18CZ] remove unused yellow value from large offboards

benjaminxscott:
[testcase - full game](https://gist.github.com/benjaminxscott/67d6827b148d7439576e2296d8d5d220)

The production game doesn't have a yellow value, since it's impossible to lay the tiles in yellow
Screenshot 2021-02-28 at 7 48 36 PM

after this change:
Screenshot 2021-02-28 at 7 47 25 PM


Zwergenpunk:
image

This is not true. That is the physical copy of a friend.

jenf (approved):

PR #4226, opened 2021-03-01
18zoo game be played from beginning to end

carlorusso:


lib/engine/game/g_18_zoo/step/dividend.rb
```diff
@@ -0,0 +1,96 @@
...
+              share_times: 1,
+              divs_to_corporation: 0,
+            }
+          end
+
+          def payout(entity, revenue)
+            {
+              corporation: 0,
+              per_share: payout_per_share(entity, revenue),
+              share_direction: share_price_change(entity, revenue),
```
michaeljb, line 46: I find it strange to read `share_price_change`, since in this game's case it is only used for paying out, not withholding. Since `payout` is the only caller, and `share_price_change` is a one-liner, would deleting `share_price_change` and using this here work instead?

```suggestion
              share_direction: revenue >= threshold(entity) ? :right : nil,
```

lib/engine/game/g_18_zoo/game.rb
```diff
@@ -1273,15 +1327,22 @@ def update_zoo_tickets_value(turn, round_num = 1)
...
+          @round.bonus_tracks = additional_tracks
         end
 
         def event_new_train!
-          # @new_train_brought = true # TODO: enable again after adding G18ZOO::Step::BuyTrain
+          @round.new_train_brought = true
+        end
+
+        def event_rust_own_3s_4s!
+          @log << '3S long and 4S owned by current player are rusted!'
```
michaeljb, line 1344: Something like this would be more consistent with other event log messages (the `--` will automatically make it bold):

```suggestion
          @log << '-- Event: 3S long and 4S owned by current player are rusted! --'
```

michaeljb (approved):

lib/engine/game/g_18_zoo/game.rb
carlorusso: done in 6e5282e3

lib/engine/game/g_18_zoo/step/dividend.rb
carlorusso: done in 6e5282e3

Issue #4214, opened 2021-02-28
1830 Share price typo in stock market


MalkD:
One of the entries in the Stock Market looks wrong.

I don't own this game, but checked with a picture of the board on BGG, and this value should be 20 rather than 12.

![1830 stock price typo](https://user-images.githubusercontent.com/75179756/109433231-38215380-7a07-11eb-906f-2a21cb139933.png)

PR #4228, opened 2021-03-01
fixes typo in 1830 market

underhilllabs:
* fixed a minor typo in 1830 market.  #4214 

Was incorrectly labelled $12, correct amount is $20.


benjaminxscott:
I think it's possible this causes pinned games, if someone bought/sold stock at that price
(not a blocker, just to note)

jenf (approved):

PR #4244, opened 2021-03-02
18CZ: Put the half-box at the end of the stock-market on the appropriate row.

crericha:


crericha:
@Zwergenpunk FYI...not sure if this was on your radar or not

michaeljb (approved):

Zwergenpunk:
Ah cool. Yeah i knew it but it was invertes bit I did not think it was a problem

PR #4245, opened 2021-03-02
Fixes #4189

### LINKED ISSUE #4189: 18WO Question Game 30061 - Is there a bug with Nieuw Zeeland?  Seems like I never got a chance to place 2nd token, so I cannot lay track
DO NOT PASTE THE GAME JSON! Please use [gist](https://gist.github.com/) instead.

Please fill out the bug report title with the game name and description

e.g. "[18Chesapeake] Incorrect revenue for route from Baltimore to DC"

**What's your Game ID?**

If your game is "Hotseat", provide a link to your raw game data by using the `Tools` tab and `Copy Game Data`.

Paste that data [into a github gist](https://gist.github.com/) and share the resulting link which will look something like `https://gist.github.com/[your_name]/[some_letters]`.

**How to Reproduce the Bug?**

e.g.
- During Operating Round 2.1 the B&O ran a 2 train from Baltimore to DC
- B&O has a token in Baltimore, and both cities are yellow tiles
- Revenue was less than the $60 that I expected

**Which Browser / Platform?**
- Browser Version
  - On Chrome, see [chrome://version](chrome://version)
  - On Firefox, see [about:support](about:support)
  - On Safari, see [these methods](https://www.lifewire.com/check-the-version-number-of-safari-446563)
- Platform (Mobile, Tablet, Desktop)

ventusignis:
The problem was that in the case when there is no choice for the size of the IPOed corporation nor any contest for the bid the home token (NZ) gets laid AFTER attempting to purchase tokens. The fix here is to add a hook to the HomeToken step in the stock round so that the second token can be laid then
Fixes #4189

### DIFF
--- a/lib/engine/game/g_1817_wo/game.rb
+++ b/lib/engine/game/g_1817_wo/game.rb
@@ -622,8 +622,16 @@ module Engine
           hexes.any? { |h| h.tile.cities.any? { |c| c.tokens.count(&:nil?).positive? } }
         end
 
+        def can_place_second_token(corporation)
+          return false if !tokenable_location_exists? || !corp_has_new_zealand?(corporation)
+
+          # Does the corp have a second token already?
+          corporation.tokens[1] && !corporation.tokens[1].city
+        end
+
+        # This must be idempotent.
         def place_second_token(corporation)
-          return unless tokenable_location_exists?
+          return unless can_place_second_token(corporation)
 
           hex = hex_by_id(corporation.coordinates)
 
@@ -717,7 +725,7 @@ module Engine
 
           G1817::Round::Stock.new(self, [
             Engine::Step::DiscardTrain,
-            Engine::Step::HomeToken,
+            G1817WO::Step::HomeToken,
             G1817WO::Step::BuySellParShares,
           ])
         end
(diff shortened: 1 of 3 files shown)


michaeljb:
At the moment game 30061 is the only one affected by this. No migration is possible, so the game must be pinned. However, it has already completed, does that mean we should archive it instead? @jenf @tobymao

michaeljb (approved):

lib/engine/game/g_1817_wo/game.rb
```diff
@@ -622,8 +622,16 @@ def tokenable_location_exists?
           hexes.any? { |h| h.tile.cities.any? { |c| c.tokens.count(&:nil?).positive? } }
         end
 
+        def can_place_second_token(corporation)
+          return false unless tokenable_location_exists? && corp_has_new_zealand?(corporation)
```
tobymao, line 626: change to an if statement

lib/engine/game/g_1817_wo/game.rb
```diff
@@ -622,8 +622,16 @@ def tokenable_location_exists?
...
+
+          # Does the corp have a second token already?
+          corporation.tokens[1] && corporation.tokens[1].city.nil?
```
tobymao, line 629: corporation.tokens[1] && !...city

lib/engine/game/g_1817_wo/game.rb
ventusignis: Changed

lib/engine/game/g_1817_wo/game.rb
ventusignis: Changed

PR #4257, opened 2021-03-02
[18CZ] Fixed scrapping of trains

### LINKED ISSUE #4211: [18CZ] should not be able to purchase a scrapped train
NWB bought in BCB and scrapped both its `2a` trains

EKJ is able to buy in those `2a` trains from NWB on its turn, even though they have been scrapped

[test case](https://gist.github.com/benjaminxscott/e90674ef9a3f35d6b329aae33abb1e14)
edit: fixed json

Zwergenpunk:
Needed to nil the owner

closes #4211

could break games, but probably very unlikely

### DIFF
--- a/lib/engine/game/g_18_cz/game.rb
+++ b/lib/engine/game/g_18_cz/game.rb
@@ -2756,7 +2756,7 @@ module Engine
             next unless rust?(t)
 
             rusted_trains << t.name
-            owners[t.owner.name] += 1
+            owners[t.owner.name] += 1 unless t.owner.nil?
             rust(t)
           end
           return if rusted_trains.none?
--- a/lib/engine/game/g_18_cz/game.rb
+++ b/lib/engine/game/g_18_cz/game.rb
@@ -2756,7 +2756,7 @@ module Engine
             next unless rust?(t)
 
             rusted_trains << t.name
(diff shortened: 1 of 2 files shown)


benjaminxscott:
Perhaps [remove_train](https://github.com/tobymao/18xx/blob/master/lib/engine/depot.rb#L71) could generically set owner to nil?

Zwergenpunk:
you linked the depot. Thats the wrong one. We call the game.remove_train.

Yeah maybe, but I have no idea how much that breaks everything.

lib/engine/game/g_18_cz/game.rb
```diff
@@ -2756,7 +2756,7 @@ def rust_trains!(train, entity)
             next unless rust?(t)
 
             rusted_trains << t.name
-            owners[t.owner.name] += 1
+            owners[t.owner.name] += 1 unless t.owner.nil?
```
tobymao, line 2759: if t.owner

Zwergenpunk: When I remove .nil? the tests fail for some reason =(

Zwergenpunk: Oh i am an idiot...

PR #4259, opened 2021-03-02
[18CO] removes villages from hexes array for GJGR company #4208

### LINKED ISSUE #4208: 18CO: Grand Junction private upgrade should be limited to "towns" only.
18CO has both "towns" and "villages". The special ability of the Grand Junction and Grand River Valley Railway is to upgrade a yellow "town" to a green city. However, it allows upgrading a town or village. This should be restricted to just towns.

underhilllabs:
- removes villages from hexes array for GJGR company
- removed connected check as in the rules it states the small village does not need to be connected

fixes #4208 

Here's json for a game to test: https://gist.github.com/underhilllabs/a1e430bf314cd6fbdeb8cb2e1c2673ff

### DIFF
--- a/lib/engine/game/g_18_co/game.rb
+++ b/lib/engine/game/g_18_co/game.rb
@@ -457,28 +457,13 @@ module Engine
                 count: 1,
                 special: true,
                 tiles: %w[14 15],
-                hexes: %w[B10
-                          B22
-                          C7
-                          C9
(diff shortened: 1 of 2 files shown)


tobymao:
@ryandriskel please review

ryandriskel (approved):

PR #4261, opened 2021-03-02
[1882] Fix minor typo in logging in init_train_handler.

mrkvm:
Remove the word "currently" from log stating which is the NWR train.


mrkvm:
Caught this little logging typo when playing 1882 the other day. I hope it was OK to just create a PR rather than an issue given how simple the change is.

PR #4284, opened 2021-03-03
1870: Show destinations on map

daniel-sousa-me:
Trying to split the connection run PR. The rest (using auto actions) will come when this one is merged.

This is a bit complicated because some tiles are really busy and I had to create new was for assignments to be rendered:

image

image

image

### DIFF
--- a/lib/engine/operator.rb
+++ b/lib/engine/operator.rb
@@ -7,13 +7,15 @@ module Engine
     include Entity
 
     attr_accessor :coordinates
-    attr_reader :color, :city, :loans, :logo, :simple_logo, :operating_history, :text_color, :tokens, :trains
+    attr_reader :color, :city, :loans, :logo, :logo_filename, :simple_logo,
+                :operating_history, :text_color, :tokens, :trains
 
     def init_operator(opts)
       @cash = 0
       @trains = []
       @operating_history = {}
-      @logo = "/logos/#{opts[:logo]}.svg"
+      @logo_filename = "#{opts[:logo]}.svg"
+      @logo = "/logos/#{@logo_filename}"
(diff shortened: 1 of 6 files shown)


lib/engine/operator.rb
```diff
@@ -7,13 +7,15 @@ module Operator
...
-    attr_reader :color, :city, :loans, :logo, :simple_logo, :operating_history, :text_color, :tokens, :trains
+    attr_reader :color, :city, :loans, :logo, :logo_filename, :simple_logo,
+                :operating_history, :text_color, :tokens, :trains
 
     def init_operator(opts)
       @cash = 0
       @trains = []
       @operating_history = {}
-      @logo = "/logos/#{opts[:logo]}.svg"
+      @logo_filename = "#{opts[:logo]}.svg"
```
tobymao, line 17: what's the point of this change?

daniel-sousa-me: It's because of this: https://github.com/tobymao/18xx/blob/cc807c445ef2e35fb2b6df83f5d943bc0fd458a0/lib/engine/game/g_1870/game.rb#L906

The method was already merged by accident.

I generated a 2nd set of svg logos for the small icons (I don't remember exactly what I did to them, but the originals popped too much visually)

yzemaze: > I generated a 2nd set of svg logos for the small icons (I don't remember exactly what I did to them, but the originals popped too much visually)

Sounds too complicated. Try [filter](https://developer.mozilla.org/en-US/docs/Web/CSS/filter), grayscale, saturate or sepia might work.

daniel-sousa-me: This was precisely to avoid having complicated code on the views that's specific to this game...
There is an icon in the icons dir for each assignment. As it is done for every other assignment.

I just changed the assignment_tokens from a constant to a method, to avoid listing all the filenames twice.

PR #4293, opened 2021-03-04
Walk refactor

### LINKED ISSUE #4287: [18Chesapeake] Game allows placement of a tile that does not extend a route
The [game allows](https://18xxgames.slack.com/archives/C012K0CNY5C/p1614812216385000) placing of a tile that fails to "extend a route of the company", at least from [this position](https://18xx.games/game/30071?action=470), where the C&A is allowed to place tile 4 with rotation 0 in I5.

Screen Shot 2021-03-03 at 5 53 28 PM

ventusignis:
Fixes #4287

### DIFF
--- a/lib/engine/part/node.rb
+++ b/lib/engine/part/node.rb
@@ -42,12 +42,13 @@ module Engine
       #
       # visited: a hashset of visited Nodes
       # visited_paths: a hashset of visited Paths
+      # visited_hexedges: a hashset of crossed HexEdges (2-size Arr of hex and edge to represent hex edge crossings)
       # on: see Path::Walk
       # corporation: If set don't walk on adjacent nodes which are blocked for the passed corporation
       # skip_track: If passed, don't walk on track of that type (ie: :broad track for 1873)
-      # 
+      #
       # This method recursively bubbles up yielded values from nested Node::Walk and Path::Walk calls
-      def walk(visited: nil, on: nil, corporation: nil, visited_paths: {}, skip_track: nil)
+      def walk(visited: nil, on: nil, corporation: nil, visited_paths: {}, visited_hexedges: {}, skip_track: nil)
         return if visited&.[](self)
 
         visited = visited&.dup || {}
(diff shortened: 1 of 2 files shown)


ventusignis:
I was working on making it so that it is possible to get ALL valid paths from a node but I noticed that there wasn't anything to prevent something like the example I made in the linked ticket so I'm putting up a PR to fix that issue first

lib/engine/part/node.rb
```diff
@@ -36,7 +36,19 @@ def select(paths, corporation: nil)
...
+      #
+      # visited: a hashset of visited Nodes
+      # visited_paths: a hashset of visited Paths
+      # visited_hexedges: a hashset of crossed HexEdges (2-size Arr of hex and edge to represent hex edge crossings)
+      # on: see Path::Walk
+      # corporation: If set don't walk on adjacent nodes which are blocked for the passed corporation
+      # skip_track: If passed, don't walk on track of that type (ie: :broad track for 1873)
+      #
+      # This method recursively bubbles up yielded values from nested Node::Walk and Path::Walk calls
+      def walk(visited: nil, on: nil, corporation: nil, visited_paths: {}, visited_hexedges: {}, skip_track: nil)
```
tobymao, line 51: visited_edges

ventusignis: I've changed this and it now only stores 1 side of each edge crossed

lib/engine/part/node.rb
```diff
@@ -45,7 +57,7 @@ def walk(visited: nil, on: nil, corporation: nil, visited_paths: {}, skip_track:
         paths.each do |node_path|
           next if node_path.track == skip_track
 
-          node_path.walk(visited: visited_paths, on: on) do |path, vp|
+          node_path.walk(visited: visited_paths, on: on, visited_edges: visited_edges) do |path, vp, vhe|
```
tobymao, line 60: vhe -> ve ?

ventusignis: Thanks, I've updated the PR

PR #4299, opened 2021-03-04
[1822] 1822+ expansion

magnusfalt:
- 6 new minors, not much change except the corporations and company array
- P19 Avonside Engine Company. Added a permanent L train, ive prepered for this so just needed to add some code for english channel which it cant run. And you cant attach a pullman to it.
- P20 Canterbury & Whitstable Railway. Same as P15 but with phase x 5. Refactored the player to hold the cash into a hash instead to hold both privates cash.
- P21 Humber Suspension Bridge Company. 2 free tile lay on the Hull / Grimby hexes. You can lay both tiles even if its a minor. And one of the tile lay may be an upgrade.
- Added a few more constants to help with setup. This is also preperation to have the 1822MRS/NRS easily override the base 1822 class.
- 1822+ categorized the privates into 3 stacks and each stack is connected to a bidbox.

This shouldnt affect any of the current games.

### DIFF
--- a/lib/engine/game/g_1822/step/special_track.rb
+++ b/lib/engine/game/g_1822/step/special_track.rb
@@ -76,6 +76,7 @@ module Engine
             return nil unless connected
 
             return connected if @game.company_ability_extra_track?(entity)
+            return connected if entity.id == @game.class::COMPANY_HSBC
 
             tile_lay = get_tile_lay(operator)
             return nil unless tile_lay
--- a/lib/engine/game/g_1822/step/special_track.rb
+++ b/lib/engine/game/g_1822/step/special_track.rb
@@ -151,6 +151,13 @@ module Engine
                 tile.color == :brown && phase_color != :gray
             end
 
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_1822/step/special_track.rb
```diff
@@ -150,6 +151,13 @@ def abilities(entity, **kwargs, &block)
                 tile.color == :brown && phase_color != :gray
             end
 
+            if entity.id == @game.class::COMPANY_HSBC && entity.owner&.corporation?
+              hsbc_token = entity.owner.tokens
```
tobymao, line 155: can be a one liner, any? { |t| t.used && ... }

KevinR1000:
P21 (HSBC) is not functioning as described here.
Major with token on yellow in Hull cannot build green in Hull and yellow in Grimsby.
See https://18xx.games/game/40231?action=371
New bug report coming.

PR #4301, opened 2021-03-04
1870: Connection runs (last big commit until 1870 is ready)

daniel-sousa-me:
This includes a few lines from #4284 (834 - 841 of `game.rb`), but they should merge well with or without each other

### DIFF
--- a/lib/engine/abilities.rb
+++ b/lib/engine/abilities.rb
@@ -9,6 +9,8 @@ end
 
 module Engine
   module Abilities
+    attr_reader :abilities
+
     def init_abilities(abilities)
       @abilities = []
 
--- a/lib/engine/game/g_1870/round/operating.rb
+++ b/lib/engine/game/g_1870/round/operating.rb
@@ -15,6 +15,12 @@ module Engine
 
             @river_special_tile_lay = nil
           end
+
+          def active_entities
+            return @connection_runs.keys unless @connection_runs.empty?
+
+            super
+          end
         end
       end
     end
--- a/lib/engine/game/g_1870/step/connection.rb
+++ b/lib/engine/game/g_1870/step/connection.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1870
+      module Step
+        module Connection
+          def override_entities
+            @round.connection_runs.keys
+          end
+
+          def current_entity
+            @round.connection_runs.keys.first
+          end
+
+          def context_entities
+            @round.entities
+          end
+
+          def active_context_entity
+            @round.entities[@round.entity_index]
+          end
+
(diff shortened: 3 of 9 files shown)


lib/engine/game/g_1870/round/operating.rb
```diff
@@ -8,13 +8,19 @@ module Game
...
+          attr_accessor :river_special_tile_lay, :connection_steps, :connection_runs
 
           def start_operating
             super
 
             @river_special_tile_lay = nil
           end
+
+          def active_entities
+            return @connection_runs.keys if @connection_runs.any?
```
tobymao, line 20: unless runs.empty?

tobymao:
conflict and also can you delete your old code that checks for destinations?

lib/engine/game/g_1870/round/operating.rb
daniel-sousa-me: Thanks. Old code :|

daniel-sousa-me:
Done. I think the code you're referring to was never merged.

daniel-sousa-me:
This failed because github had some trouble downloading gems. The branch is passing all the tests.

ventusignis:
I had a gems issue on my PR as well; I was able to coax github into re-running the tests by force-pushing a wording change to a comment and the tests passed after that

scottredracecar:
I got mine to pass when I clicked the re-run jobs button.
image

ventusignis:
IIRC only people with permissions to directly modify this repo are allowed to rerun jobs

lib/engine/game/g_1870/step/connection.rb
```diff
@@ -0,0 +1,31 @@
...
+          def context_entities
+            @round.entities
+          end
+
+          def active_context_entity
+            @round.entities[@round.entity_index]
+          end
+
+          def active?
+            @round.connection_runs.any? && !passed?
```
tobymao, line 25: !connection_runs.empty?

lib/engine/abilities.rb
```diff
@@ -39,6 +39,10 @@ def remove_ability_when(time)
       end
     end
 
+    def own_abilities
```
tobymao, line 42: what is this? can't you just call abilities?

daniel-sousa-me: Because it was not accessible. I could have made it readable, but the same name is also used in Game as an alias for all_abilities, so I thought I might as well make it more clear. But I can make it that way, if you prefer.

tobymao: yea just add an attr_reader to abilities

daniel-sousa-me: Done

tobymao (approved):

PR #4307, opened 2021-03-04
18usa: stock value movement, metropolitian cities, offboards

CaptainObi:
First round of changes for 18USA. I've fixed the offboard randomization, added metropolitan randomization, and added the different stock value movement. I've also moved it to the new structure.

### DIFF
--- a/lib/engine/game/g_18_usa/game.rb
+++ b/lib/engine/game/g_18_usa/game.rb
@@ -985,6 +985,17 @@ module Engine
           }
         end
 
+        def stock_round
+          close_bank_shorts
+          @interest_fixed = nil
+
+          G18USA::Round::Stock.new(self, [
+            Engine::Step::DiscardTrain,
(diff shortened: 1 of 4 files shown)


CaptainObi:
(I'm not sure what this bug is, could I get some insight? @ventusignis )

magnusfalt:
In lib/engine/game/g_18_usa/step/loan.rb the name is of the module is 1817, i think you overrides the normal 1817 module and then all the rspec for 1817 will fail.

CaptainObi:
That sounds right

CaptainObi:
I'm clearly doing something wrong. I'm going to redo this whole mess.

magnusfalt:
A good idea is to run "docker-compose exec rack rake" on your machine before you commit. Then you can fix the rubocop errors before it hits github.

CaptainObi:
> docker-compose exec rack rake

I did, but it clearly didn't work.

PR #4310, opened 2021-03-04
Add protection for shenanigans on auto-buys and add auto-pass

jenf:
Another step on #209


daniel-sousa-me:
This seems to be very very conservative. Are you planning on having an option to disable some of these protections? I think most of the times I'll enable auto actions, I don't care if people are selling shares. Otoh, I may care if someone helps me float, because that frees up some money that I may want to spend early.

jenf:
It's intentionally conservative. Eventually there might be some options, but for now the important thing is it doesn't do the wrong thing

daniel-sousa-me:
Indeed I didn't mean to inform you that it was conservative :P

tobymao (approved):

PR #4313, opened 2021-03-04
1870: Add the PriceProtection step to the OR

daniel-sousa-me:
When there's an ebuy, if shares are sold, the president can also protect
them. Had to add the step to the OR and that required changing some
assumptions.

### DIFF
--- a/lib/engine/game/g_1870/step/price_protection.rb
+++ b/lib/engine/game/g_1870/step/price_protection.rb
@@ -26,7 +26,9 @@ module Engine
           end
 
           def active_entities
-            @round.sell_queue.map(&:president)
+            return [] if @round.sell_queue.empty?
+
+            [@round.sell_queue.first.president]
           end
 
           def purchasable_companies(_entity = nil)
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1870/step/price_protection.rb
```diff
@@ -22,7 +22,11 @@ def description
           end
 
           def round_state
-            { sell_queue: [] }
+            super.merge(sell_queue: [])
+          end
+
+          def active_entities
+            @round.sell_queue.map(&:president)
```
jenf, line 29: This should only return one active entity.

lib/engine/game/g_1870/step/price_protection.rb
```diff
@@ -60,24 +66,23 @@ def process_buy_shares(action)
               price: price
             )
 
-            @round.goto_entity!(player)
+            @round.goto_entity!(player) if @round.entities[@round.entity_index].player?
```
jenf, line 69: Why this change? Or is it to stop the entity changing in an OR?

daniel-sousa-me: If player B sells a share of a corp of player A, and player A protects the share, the game continues to the player next to player A (instead of the one next to player B!)

But this only makes sense in the SR

jenf: Please add a comment explaining.

jenf (approved):

PR #4315, opened 2021-03-04
Group the game list on CreateGame by stage

daniel-sousa-me:
I'm sorry for the excessive amount of PRs. I'm not sure if it was worth withholding small stuff like this.

Before:
image

After:
image

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -55,15 +55,22 @@ module View
       @min_p = {}
       @max_p = {}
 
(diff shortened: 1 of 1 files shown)


daniel-sousa-me:
It would also be cool to have the name of the location more spaced from the title of the game, but I'm afraid of messing with CSS on select box, because they're shown very differently in some devices.

scottredracecar:
I like it! Might also provide some incentive to squash bugs to promote games faster too.

tobymao (approved):

PR #4316, opened 2021-03-05
1870: Fix P tiles

daniel-sousa-me:
Oof, this was a tricky one.

I promise I'm done with PRs for the day. After these are merged, 1870 should be ready for alpha!

I'll do some more testing after all this is merged and then I'll do a PR changing the stage

### DIFF
--- a/lib/engine/game/g_1870/game.rb
+++ b/lib/engine/game/g_1870/game.rb
@@ -83,14 +83,19 @@ module Engine
           '145' => 2,
           '146' => 2,
           '147' => 2,
-          '170' => 4,
+          '170P' =>
+          {
+            'count' => 4,
+            'color' => 'brown',
+            'code' => 'city=revenue:50,slots:2;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0',
+          },
           '171K' =>
           {
             'count' => 1,
             'color' => 'gray',
             'code' =>
             'city=revenue:60,slots:3;path=a:0,b:_0;path=a:1,b:_0;'\
-            'path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:5,b:_0;label=K',
+            'path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:5,b:_0',
           },
           '172L' =>
           {
@@ -98,7 +103,7 @@ module Engine
             'color' => 'gray',
             'code' =>
             'city=revenue:60,slots:2;path=a:0,b:_0;path=a:1,b:_0;'\
-            'path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:5,b:_0;label=L',
+            'path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:5,b:_0',
           },
         }.freeze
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1870/game.rb
```diff
@@ -932,13 +937,13 @@ def legal_tile_rotation?(_entity, hex, tile)
             end
         end
 
-        def upgrades_to?(from, to, _special = false)
-          return to.name == '170' if from.color == :green && P_HEXES.include?(from.hex.name)
+        def upgrades_to?(from, to, special = false)
+          return false if (to.name == '171K') ^ (from.color == :brown && from.hex.name == 'B11')
```
jenf, line 941: Do we need xor here?

daniel-sousa-me: It was a decision between that or repeating all the huge lines with slightly different connectives. I opted for xor because I thought it was slightly more informative of the intent (iff) and more compact.

For example, that line would become
```
          return false if (to.name != '171K') && (from.color == :brown && from.hex.name == 'B11')
          return false if (to.name == '171K') && (from.color != :brown || from.hex.name != 'B11')
```

Or some variation of this, which makes it easier to follow from examples, but makes the high-level idea more opaque.

Maybe I could add a comment like _`return if p ^ q` means something like 'assert p <=> q'`

jenf: I'll have to wait for @tobymao's input since we haven't used XORs in the code before.

The other changes look fine.

daniel-sousa-me: There's no hurry :)

It's good we try to keep this as readable as possible, because this code is always tricky (I can't tell you how many iterations I went through until I figured out a clean way to catch all the cases without rewriting the main method...)

tobymao: what are you trying to do here?

daniel-sousa-me: In the case of this line I'm saying that the tile 171K can only be placed on B11 and that if B11 is brown, it can only be upgraded to the tile 171K.

Now that I'm reading the lines I've wrote expanding the xor, they can indeed be simplified:
```
return false if to.name != '171K' && (from.color == :brown && from.hex.name == 'B11')
return false if to.name == '171K' && from.hex.name != 'B11'
```

I'll leave it however you prefer...

daniel-sousa-me: This would normally be done with Labels, but the labelled cities use normal yellow and green tiles and I couldn't find a reasonable way to use them.

Other games with similar situation copy the entire original method (~20 lines) and make the necessary adjustments in place, but I don't think that's a very future-proof option

tobymao: if tile == '171.k'
  return (from.color == brown && from.hex )  || ...
end

daniel-sousa-me: Well, this specific tile doesn't have any other restrictions, but for the brown P tiles, it still needs to check if the paths are respected. At this point we're just copying the entire `upgrades_to?` method multiple times, might as well just override it completely and not call `super`

tobymao: i don't have enough context to make the call, but it seems like there's some simplification you can do

daniel-sousa-me: It would become:

```
        def upgrades_to?(from, to, special = false)
          # correct color progression?
          return false unless Engine::Tile::COLORS.index(to.color) == (Engine::Tile::COLORS.index(from.color) + 1)

          # honors pre-existing track?
          return false unless from.paths_are_subset_of?(to.paths)

          # correct label?
          return false if to.label && !from.label
          return false if to.label == 'K' && from.hex.name != 'B11'
          return false if to.label == 'L' && from.hex.name != 'C18'

          true
        end
```

daniel-sousa-me: I just realised that setting `special = true` skips the labels check and all the other checks that it skips are irrelevant for 1870, so I'll submit a new version using that

daniel-sousa-me: See if you like this version: #4329

PR #4324, opened 2021-03-05
zigzag stock market: fix box_height

### LINKED ISSUE #4321: [zigzag markets] When more than 3 medium tokens are on the same market price in the top row, the 4th token is obscured by the bottom row
https://18xx.games/game/28521?action=416#market

For example:
Screen Shot 2021-03-04 at 11 12 13 PM

yzemaze:
fixes #4321

### DIFF
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -36,8 +36,9 @@ module View
 
       # 1D markets
       VERTICAL_TOKEN_PAD = 4                      # vertical space between tokens
-      MIN_NUM_TOKENS = 4                          # guarantee space for this many tokens
+      MIN_NUM_TOKENS = 3                          # guarantee space for this many tokens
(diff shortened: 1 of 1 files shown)


assets/app/view/game/stock_market.rb
```diff
@@ -72,6 +72,13 @@ def box_style_1d
         }
       end
 
+      def box_height_1d
+        token_height = @game.stock_market.market.first.map do |p|
+          p.corporations.sum { |c| TOKEN_SIZES[@game.corporation_size(c)] + VERTICAL_TOKEN_PAD }
+        end.push(MIN_NUM_TOKENS * (TOKEN_SIZE + VERTICAL_TOKEN_PAD)).max
```
tobymao, line 78: i don't like this style and don't use push

maybe something like
 (map { p.sum } + [constan]).max

yzemaze: just copy & pasted, will check later

PR #4340, opened 2021-03-05
Add Simple Tokens to Lots of Games

scottredracecar:
This adds simple logos for the following games:
- 18FL
- 18GA
- 18Los Angeles
- 18Mag
- 18AL
- 18Chesapeake
- 18CZ
- 18EU
- 18MS
- 18MEX
- 18SJ
- 18TN
- 18USA (same as 1817)
- 1835
- 1836Jr30
- 1846
- 1848
- 1856
- 1860 
- 1861
- 1870
- 1873
- 1882
- 1893


scottredracecar:
This is half simple token references and the other half the new simple tokens themselves, plus the script to generate the references in the scripts folder.

roseundy:
1860 already has simple tokens. What does this do?

scottredracecar:
It makes the simple tokens render like the other simple tokens on the site.

The current 1860 tokens render like this on my mac.
image

While the normal simple tokens render like this.
image

(sorry I cannot do a pure AB comparison as I would need to load up my other machine)

I think it is good for the simple tokens to all follow the same baseline design so I included them, even for cases like this.

Issue #4278, opened 2021-03-03
Option to enable simple corporation logos from within games individually. 


philcampeau:
For the most part, I like using the fancier corporation logos in game, but there are a couple of games whose tokens are just impossible to distinguish. 

Rather than having to switch back & forth depending on my game, it would be nice to have the option to enable the feature one game at a time Via each game’s Tools tab.

dfannius:
I really want this now that 1846 has simple logos; I prefer simple logos on other games so I have the global switch turned on but I prefer the fancy logos on 1846, which now I don't get anymore.

scottredracecar:
I think how this should work is that the system has a per-user record of the last "token state" the user had the game in for that title. 

So if I like simple logos in 1817, I would toggle that on in the map. Then if I like standard logos in 1846, I toggle that on in the map as well. The system remembers that I last had 1817 in simple and 1846 in standard. 

There are no settings/checkboxes, it just remembers.

philcampeau:
At a bare minimum, I think it would be nice to be able to activate it
manually in each game, the same way I can enable Master Mode in a game and
it stays enabled for that game.

>

PR #4349, opened 2021-03-06
show player colors per title

yzemaze:
This allows overriding show_player_colors per title. For now in parallel to the general option, so it’s still just one click to show/hide player colors for all titles – unless one had configured many titles beforehand.
I wouldn’t mind getting rid of the general option completely, because most people (on Slack back then) said they only use it on select titles anyway.

[Addition to lib/storage and & due to specs failing otherwise.]

PS: The same MO would work for #4278

### DIFF
--- a/assets/app/lib/storage.rb
+++ b/assets/app/lib/storage.rb
@@ -4,7 +4,7 @@ module Lib
   module Storage
     def self.[](key)
       value = `localStorage.getItem(#{key})`
-      JSON.parse(value) if value
+      JSON.parse(value) if value && !value.empty?
     end
 
     def self.[]=(key, value)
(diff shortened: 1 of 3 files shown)


assets/app/lib/storage.rb
```diff
@@ -4,7 +4,7 @@ module Lib
   module Storage
     def self.[](key)
       value = `localStorage.getItem(#{key})`
-      JSON.parse(value) if value
+      JSON.parse(value) if value && !value.empty?
```
tobymao, line 7: value&.any?

yzemaze: It’s a string, not an array.

Issue #3739, opened 2021-02-08
Option to suppress currency symbols


ChrisTheCat:
I would prefer if $ and ₽ and Ft symbols weren't displayed. They add visual clutter and make it harder for me to parse numbers at a glance. It would be nice if there was a profile checkbox to toggle them on and off.

bridger15:
I think I agree.  Our current game of 18Mag looks like this: https://i.imgur.com/0Bq7P8i.png

It's got the extra line for ONLY the 'Ft' and if that was disabled then it would look much more clean.

dfannius:
Currency format strings should use non-breaking spaces to avoid the sort of issue in that picture. (Though I'm not sure what it would look like afterwards given that the layout thinks it needs a linebreak.)

yzemaze:
> Though I'm not sure what it would look like afterwards given that the layout thinks it needs a linebreak.

It’s fine. Mock-up with @Zwergenpunk ’s nbsp included and 320px width:
![image](https://user-images.githubusercontent.com/33390595/110209386-9d07fe00-7e8c-11eb-8b28-3e1d7eccfa93.png)

Zwergenpunk:
So for now we have a &nbsp; between the number and the Symbol. Every new game has to remeber that a unicode has to be copied in and no white space. 

An other solution for that would be to adjust the currency_format function to replace a space with a nbsp but thr code for that would look really confusing 😀

The best solution would be to have an own component that uses css nowrap to display the currency, but that would require alot of work. But with that solution it would be possible to give the user settings to disable the currency symbol.

PR #4351, opened 2021-03-06
Added &nbsp; to currency formats with space

Zwergenpunk:
In 18CZ and 1836jr is a unneeded line break

Before:
image

After:
image

### DIFF
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -88,7 +88,7 @@ module View
             textAlign: 'left',
           },
         }
-        share_props = { style: { width: '2.9rem' } }
+        share_props = { style: { width: '2.7rem' } }
 
         if corporation_interest_penalty?(entity)
           corporation_penalty = "#{entity.name} has " +
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -88,7 +88,7 @@ module View
             textAlign: 'left',
           },
         }
-        share_props = { style: { width: '2.7rem' } }
+        share_props = { style: { width: '2.9rem' } }
 
         if corporation_interest_penalty?(entity)
           corporation_penalty = "#{entity.name} has " +
--- a/lib/engine/game/g_1836_jr30/game.rb
+++ b/lib/engine/game/g_1836_jr30/game.rb
@@ -9,7 +9,7 @@ module Engine
       class Game < Game::Base
         include_meta(G1836Jr30::Meta)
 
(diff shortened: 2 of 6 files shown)


yzemaze (commented):
This won’t help, because the td has 1rem padding-right. With 2.9rem th width:
image

If you’re gonna deal with this check .padded_number, else I’ll do.

Zwergenpunk:
yeah I did not check the "long" numbers.
I could also just use a unicode &nbsp; in the currency sting

yzemaze:
Do it right or don’t do it :-)
image
… seems to work very well ;-)

yzemaze:
SJ & Mag, too.

Zwergenpunk:
added

dfannius:
I lobbied in #3739 for currency format strings to have non-breaking spaces. I don't think it's a hack; it really should be an inherent property of currency strings that they not be broken across lines.

yzemaze:
@dfannius "hack" was meant in a complimentary way ;-) And you’re right, although this PR might complicate the other issue …

Zwergenpunk:
The only other solution that really fixes the problem is that the currency_format function replaces whitespace with a nbsp character.
To solve it with css is impossible.

Is the other issue to just show the number? Wouldn't that be pretty easy, to change the currency_format function to just replace the number if the settings is disabled

dfannius:
It would indeed be pretty easy, I imagine; the question is whether it's a good idea (and if anyone who thinks it is has the motivation to do it). But that conversation is probably better had over in that issue.

I like the idea of the non-breaking space substitution being done directly in `currency_format`.

yzemaze:
> The only other solution that really fixes the problem is that the currency_format function replaces whitespace with a nbsp character.

I’d prefer that, because this eliminates the problem of future people implementing games not knowing about space/nbsp. Additionally I wouldn’t count on all editors playing along nicely with unicode …

> To solve it with css is impossible.

Wrong.

Zwergenpunk:
> > The only other solution that really fixes the problem is that the currency_format function replaces whitespace with a nbsp character.
> 
> I’d prefer that, because this eliminates the problem of future people implementing games not knowing about space/nbsp. Additionally I wouldn’t count on all editors playing along nicely with unicode …

Sure that seems fixable
> 
> > To solve it with css is impossible.
> 
> Wrong.

I know that is theoretically possible, but you need to find every spot a currency is displayed and change it. Or you use a a own component for currency format and replace all occurences

tobymao (approved):

PR #4384, opened 2021-03-07
OO tile layer decides token placement

crericha:
- OO tile layer decides token placement
- For 1828, make it the token owner instead and skip if OO is filled with blocking tokens


crericha:
A handful of 1830 games have the wrong entity placing the token.

tobymao:
can you resolve this cnoflict?

crericha:
> can you resolve this cnoflict?

Resolved

PR #4393, opened 2021-03-08
Scottredracecar boot merge

yzemaze:
This didn’t quite work out as expected … It’s worth merging after Scott’s PR is merged, because of the optimized logos commit anyway.


tobymao:
confilct now

PR #4400, opened 2021-03-08
Show Progressbar for 18MS

### LINKED ISSUE #4392: 18MS: Improved timeline
18CZ has an improved timeline. Maybe something similar could be made in 18MS?

image

Zwergenpunk:
Adds Progress bar for 18 MS
The train icons are in 18CZ for export, but I think it works for 18MS as well for the rusting. 
image

With OR 11 optional rule there is one more OR in the progress bar
closes #4392

### DIFF
--- a/lib/engine/game/g_18_ms/game.rb
+++ b/lib/engine/game/g_18_ms/game.rb
@@ -731,6 +731,34 @@ module Engine
           super
         end
 
+        def show_progress_bar?
+          true
+        end
+
+        def progress_information
+          base_progress = [
+            { type: :PRE },
+            { type: :SR },
+            { type: :OR, name: '1' },
+            { type: :OR, name: '2' },
+            { type: :SR },
+            { type: :OR, name: '3' },
+            { type: :OR, name: '4', exportAfter: true },
+            { type: :SR },
+            { type: :OR, name: '5' },
+            { type: :OR, name: '6', exportAfter: true },
+            { type: :SR },
+            { type: :OR, name: '7' },
+            { type: :OR, name: '8', exportAfter: true },
+            { type: :SR },
+            { type: :OR, name: '9' },
+            { type: :OR, name: '10' },
+          ]
+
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_ms/game.rb
```diff
@@ -731,6 +731,34 @@ def president_assisted_buy(corporation, train, price)
...
+            { type: :OR, name: '6', exportAfter: true },
+            { type: :SR },
+            { type: :OR, name: '7' },
+            { type: :OR, name: '8', exportAfter: true },
+            { type: :SR },
+            { type: :OR, name: '9' },
+            { type: :OR, name: '10' },
+          ]
+
+          base_progress.push({ type: :OR, name: '11' }) if @optional_rules&.include?(:or_11)
```
tobymao, line 758: base_progress <<

perwestling:
It looks pretty good but I am not sure I do agree with the use of colors.
In 18MS green phase starts when OR 2 starts. As far as I know this is the only title that use the timeline to change phase.
So I believe OR 2 should be green, and OR 1 should be yellow. 
OR3+ I am not sure what color they should have, but how about having all other boxes without color?
(You cannot know which boxes would be brown as you cannot know when the 6 train is bought, or gray when D is bought.)

The alternative is to put some icon/mark on OR2.

The overcrossed trains for rusting is fine. The alternative would be to put some icon/mark on OR4, OR6, OR8. 
And another possibility is to add 2+, 3+, 4+ but I guess that might put too much info in the timeline?

BTW, maybe END should be light blue? Like the END on the stock market in some games?

Zwergenpunk:
I don't know how good it is to map the OR with a color to a phase, when not all phases are mappable to a OR.

I think we could add a 2+, 4+ etc to the train icon, and I also like the blue for end. Will fix that.

Zwergenpunk:
Changed image in top comment

scottredracecar:
I think the yellow and green should be some neutral color like white rather than one associated with a phase.

tobymao (approved):

PR #4401, opened 2021-03-08
Allow Tile Income ability to work for all tile lays

underhilllabs:
This updates steps/tracker.rb
- it allows for a tile_income ability for ALL tile lays
- I pulled the tile_income feature into 2 methods to simplify tile_lay method
-  also updates 18Neb company abilities.

thanks @crericha for help with the tracker code.


lib/engine/game/g_18_neb/game.rb
```diff
@@ -375,7 +389,13 @@ class Game < Game::Base
             revenue: 5,
             desc: '$5 revenue each time ANY tile is laid or upgraded.',
             sym: 'P5',
-            # abilities: [{ type: 'blocks_hexes', owner_type: 'player', hexes: ['C7'] }],
+            abilities: [
+              {
+                type: 'tile_income',
+                terrain: nil,
```
crericha, line 395: Don't include terrain in the configuration, its confusing. Not specifying will default to nil.

underhilllabs: ah, that makes sense. I've removed terrain.

crericha (approved):

crericha:
I lied. Looks like you need to also default terrain kwag to nil in the tile_income class. Once you update that and the rake task passes, I'll merge.

crericha:
Maybe I wasn't clear. You don't need to add back in the terrain config parameter. What you need to do is default it to nil in the setup function declaration.

def setup(terrain: **nill**, income:, owner_only: false)

crericha:
Also, make sure you run "docker-compose exec rack rake" before updating the PR. If it passes locally, it'll pass here too. Right now it is failing for you.

underhilllabs:
Thanks, @crericha! Sorry about all the rubocop validation errors.

crericha:
> Thanks, @crericha! Sorry about all the rubocop validation errors.

It happens to all of us. Rubocop is pretty particular about what it likes.

Also, the terrain parameter is still in the game config. By your previous commit message, I think you meant to take that out?

PR #4412, opened 2021-03-09
[1822] Refactor graph walk

### LINKED ISSUE #4346: [1822] Game #31338 is Incredibly Laggy
https://18xx.games/game/31338

Near the end of the game #31338 the game started becoming so slow and laggy that we were unable to finish. We all started receiving "page is non-responsive" errors from our browsers and single actions took minutes to register.

### LINKED ISSUE #4354: [1822] Game is stuck
We're almost at the end and the game is now stuck - won't let NER do any action. Try upgrading Derby or just skipping place token: https://18xx.games/game/31338

Could be related to #4346

magnusfalt:
**Changes to base code**
- Engine::Part::Node. Added a new method graph_walk. I opted to make a new method just for the walk called from the graph. Reason for this is in the first step i didnt want to mess with the walk which is called from different games in example 1856 and 1860.
- Engine::Part::Path. Added a new method graph_walk. Reason for the new method is the same as above. The orginal walk is aslo called when you lay a tile with the chain parameter. I didnt want to touch theese in the first round of refactor.
- Engine::Graph. Calling the new graph_walk method. path_node is a mutable hash to pass down if we are starting from multiple token nodes. No need to traverse paths we already is done with.

**1822 Specific**
- No changes here.

fixes #4346
fixes #4354

This have no effect on current games.

### DIFF
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -117,7 +117,9 @@ module Engine
       #  * If `chain` is nil `walk` will yield to a block, providing this Part and the visited Paths including this one
 
       # on: HashSet of Paths that we want to *exclusively* walk on
-      def walk(skip: nil, jskip: nil, visited: nil, visited_edges: nil, on: nil, chain: nil)
+      def walk(skip: nil, jskip: nil, visited: nil, visited_edges: nil, on: nil, chain: nil, skip_track: nil,
+               corporation: nil, connections: nil, path_chain: [], visited_nodes: nil, current_node: nil,
+               max_nodes: nil)
         return if visited&.[](self)
 
         visited = visited&.dup || {}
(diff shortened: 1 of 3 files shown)


tobymao:
what's the harm of replacing walk for all the other games? shouldn't this still work?

magnusfalt:
No there is no harm, some of the calls to path walk expect it to just get the path and the stop when it hits a node. Ive redone it to have the path_tree in the normal walk method.

tobymao:
so can we just replace it all then?

magnusfalt:
Yes, commiting the modified soon.

daniel-sousa-me:
How long will this take? 1870 is basically ready to go to alpha, but apparently has a bugfix waiting for this.

The more work I do, the farther I feel I'm from being able to release it.

daniel-sousa-me:
Ah, does this already incorporate the change I needed to make? That's nice

lib/engine/part/path.rb
```diff
@@ -126,10 +128,33 @@ def walk(skip: nil, jskip: nil, visited: nil, visited_edges: {}, on: nil, chain:
         if chain
           chained = chain + [self]
           yield chained if chain.empty? ? @nodes.size == 2 : @nodes.any?
+        elsif path_tree
+          path_chained = path_chain + [self]
+          if @nodes.any? { |n| n != current_node }
+            path_chained_id = path_chained.map { |p| "#{p.hex.id}-#{p.index}" }.join(':')
+            return if path_tree[path_chained_id]
```
tobymao, line 135: you don't need to do this, you can just store the paths in the path tree?

magnusfalt: No you right. I had a sort before on the id so it didnt matter which why you came from into the paths. This only works for Lawson tiles. But just storing the path_chained will work.

tobymao (approved):

PR #4424, opened 2021-03-10
[2038] Cleanup tsi launch

jason-johnson:


lib/engine/game/g_2038/ability/launch.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative '../../../ability/base'
+
+module Engine
+  module Ability
+    class Launch < Base
+      attr_reader :style, :corporation
```
tobymao, line 8: doesn't seem like style is being used? are there other abilities that could do this?

jason-johnson: Style is not yet used.  But there are several kinds of launching and I expect to handle them from this class.

There are currently no abilities explicitly for this activity.  It would probably be possible to abuse existing activities in some way to accomplish this functionality but it would make the code less clear because anyone reading the code would be expecting something else than what will actually happen.

lib/engine/game/g_2038/meta.rb
```diff
@@ -16,6 +16,14 @@ module Meta
         GAME_RULES_URL = 'https://boardgamegeek.com/filepage/135017/2038-english-rules-and-supplements'
 
         PLAYER_RANGE = [3, 6].freeze
+
+        OPTIONAL_RULES = [
+          {
+            sym: :optional_short_game,
+            short_name: 'Short Game',
+            desc: 'Play a shorter game',
```
michaeljb, line 24: Could the description be more specific? In the rules file I found it says "Remove $6,000 and 2 Phase II ships from the game." Could that text be used? Or if we want to paraphrase instead, "Play with a $4,000 bank and only two 4dc3 ships/trains" or something like that?

michaeljb: Just found the spot with the full Short Game summary, maybe the description should just say to see the bottom right of page 3 for the short game rules?

michaeljb (approved):

lib/engine/game/g_2038/meta.rb
jason-johnson: Sure, I didn't do much with it yet because the only thing even considered at the moment is the par'ing of TSI.  If you need the description fixed first I can do that.

tobymao (approved):

PR #4432, opened 2021-03-10
[18ZOO] Refactor to expose all powers as choice_ability

carlorusso:


carlorusso:
Sorry for the huge PR, I'm moving from choice to choice_ability to be able to show the abilities in each step in the proper way

carlorusso:
I'm doing more modification and I'm changing something I already changed here in a different way.
I'll' re-open a new PR with smaller changes later

PR #4438, opened 2021-03-10
New game page: hide upload input and replace it with a normal button

daniel-sousa-me:


assets/app/view/create_game.rb
```diff
@@ -45,10 +45,12 @@ def render_content
           }
         end
 
+        inputs << h(:br)
```
yzemaze, line 48: Please don’t. Use `display: block` or something else.

assets/app/view/create_game.rb
```diff
@@ -45,10 +45,12 @@ def render_content
...
+        inputs << render_button('Upload button') { `document.getElementById('import_json_file').click()` }
```
yzemaze, line 49: Just `Upload` should be fine ;-)

assets/app/view/create_game.rb
```diff
@@ -45,10 +45,12 @@ def render_content
...
         inputs << h(:input, {
                       attrs: { type: :file, id: :import_json_file, accept: '.json' },
                       on: { change: upload },
-                      style: { background: :unset },
+                      style: { visibility: :hidden },
```
yzemaze, line 53: The element will take up space nonetheless, which likely will result in blank line and possibly even overflow on mobile.
Please test thoroughly.

assets/app/view/create_game.rb
daniel-sousa-me: It's already changed. Why did you ask for this?

assets/app/view/create_game.rb
daniel-sousa-me: I meant "Upload file" 😂

I added a word, because just "upload" was the same size as "submit" and it somehow looked weird

assets/app/view/create_game.rb
daniel-sousa-me: I wasn't afraid of the possible extra line, because I was already in doubt if there should be even more space :P

In any case, I moved it to the bottom to avoid that. It can't be hidden with `display: none`, because that prevents the event from triggering.

assets/app/view/create_game.rb
yzemaze: What are you talking about?
I reviewed your changes. You force-pushed the commit removing this line half an hour later.

assets/app/view/create_game.rb
yzemaze: I see no trace of a width-limit. Have you tested for overflow on mobile?

daniel-sousa-me: It is just a single element on the bottom, it shouldn't be problematic.

assets/app/view/create_game.rb
yzemaze: It’s just really bad style. `br` should only be used for texts (in p) and even that’s not a very good idea because of screen-readers.

assets/app/view/create_game.rb
yzemaze: Sounds like "No". Test it. If it overflows (which I expect it does) fix it.

daniel-sousa-me: I didn't test in all phones, no.

If some device overflows with this, it will overflow with any other file input. But feel free to propose an alternative.

yzemaze: Dude …
It’s not about the file dialog. That will be handled natively and definitely not overflow.
It’s about the invisible html-element that takes up space nonetheless. It’s too wide for the page => overflow => wiggling page on mobile => no go!
Just limit the element’s width and test that the overflow is gone afterwards. No need to do this on "all phones" or any phone at all. You could just use responsive design mode (web-dev) in your preferred browser and set width to 320 px …

daniel-sousa-me: Errr... Where did I write anything about the dialog? wtf...

The width will depend on the browser you're using (as you've just shown in the other PR). That's what I meant by "all phones": not all devices, but all configurations (operating systems + browsers + whatever)

> You could just use responsive design mode (web-dev) in your preferred browser and set width to 320 px
Usually I just go for the lazy option and resize the browser xD

PR #4442, opened 2021-03-11
Walk get all paths

ventusignis:
Before, `Node.path ` would yield all path segments reachable from a node. Now it also yields all possible paths (visited paths and visited edges)


michaeljb:
Looks like the merge conflict is just a comment, should be easy to resolve.

michaeljb:
Does this spot need updating with the visited and paths and visited edges?

https://github.com/tobymao/18xx/blob/2a8ff3ef5f0d70ec93cd827a6b05857a657f777c/lib/engine/graph.rb#L158

tobymao:
this needs a rebase

tobymao:
closing for now, can reopen when you're ready

PR #4459, opened 2021-03-12
[1822] Sold-out companies

### LINKED ISSUE #4343: [1822] Sold-out Companies on the Top Row Should Move Down and to the Right
https://18xx.games/game/31338?action=964

LNWR was sold out at the end of the SR and should have moved diagonally downwards to the right, since it was on the top row, but the stock price did not move at all. 

4.11.1:
`At the end of the stock round, if 100% of any company is in players’ hands the stock market token for each such company rises one space vertically on the stock market, if possible. If the token is already at the top of the chart, it moves diagonally one space down to the right.`

magnusfalt:
- At the end of the stock round, if 100% of any company is in players hands the stock market token for each such company rises one space vertically on the stock market, if possible. If the token is already at the top of the chart, it moves diagonally one space down to the right.

fixes #4343

This change might affect games that bought a share with the wrong price.

### DIFF
--- a/lib/engine/game/g_1822/stock_market.rb
+++ b/lib/engine/game/g_1822/stock_market.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require_relative '../../stock_market'
+
+module Engine
+  module Game
+    module G1822
+      class StockMarket < Engine::StockMarket
+        def move_up(corporation)
+          r, c = corporation.share_price.coordinates
+
+          if r.positive? && share_price(r - 1, c)
+            r -= 1
+          else
+            r += 1
+            c += 1
+          end
+          move(corporation, r, c)
+        end
+      end
+    end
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1822/stock_market.rb
```diff
@@ -0,0 +1,23 @@
...
+      class StockMarket < Engine::StockMarket
+        def move_up(corporation)
+          r, c = corporation.share_price.coordinates
+
+          if r.positive? && share_price(r - 1, c)
+            r -= 1
+          else
+            r += 1
+            c += 1
+          end
```
crericha, line 17: Do you need to account for this condition?

> At the end of the first operating round following a stock round in which a major company’s stock market value
reaches the game end value of the stock market.

crericha (approved):

PR #4471, opened 2021-03-12
18CZ Added log for ability used

### LINKED ISSUE #4334: [18cz] Small Local Companies abilities triggered without benefit
-https://18xx.games/game/31296 

- When the players went to try to use the Small Local Company, they were not able to, as the game showed the uses of (0/1).   A player who had a Small and Medium Local Company in a medium operating company was able to activate the Medium Local Company for a green tile lay, but the Small Local Company showed as being used (even though he doesn't recall using it).

Our only guess would be if the Local Company's ability to pay for terrain is being triggered without notification or request?  We couldn't find a spot in the log where this happened.   It happened to another player as well, who didn't recall using his small local company's ability, but the game showed it as being used.

Zwergenpunk:
Makes it more clear that an ability was used

Resolves #4334

### DIFF
--- a/lib/engine/game/g_18_cz/step/sell_company_and_special_track.rb
+++ b/lib/engine/game/g_18_cz/step/sell_company_and_special_track.rb
@@ -44,6 +44,8 @@ module Engine
 
             abilities = @game.abilities(action.entity, :tile_lay, time: 'any')
             abilities.each(&:use!)
+
+            @log << "The ability of #{action.entity.name} was used"
           end
 
           def hex_neighbors(entity, hex)
--- a/lib/engine/game/g_18_cz/step/sell_company_and_special_track.rb
+++ b/lib/engine/game/g_18_cz/step/sell_company_and_special_track.rb
@@ -38,14 +38,20 @@ module Engine
                       else
                         @game.current_entity
                       end
+
+            unless @game.purple_tile?(action.tile)
+              discount = action.hex.tile.upgrades.sum(&:cost)
+              @log << "#{action.entity.owner.name} receives a discount of "\
+              "#{@game.format_currency(discount)} from "\
+              "#{action.entity.name}"
+            end
+
             lay_tile(action, spender: spender)
 
             @game.skip_default_track unless @game.purple_tile?(action.tile)
 
             abilities = @game.abilities(action.entity, :tile_lay, time: 'any')
             abilities.each(&:use!)
-
-            @log << "The ability of #{action.entity.name} was used"
           end
 
           def hex_neighbors(entity, hex)


crericha:
Aren't abilities logged as "Corporation (Company name) lays tile..."? Is that not clear enough? Added another line to the log seems redundant.

Zwergenpunk:
Yeah I don't know. The only difference that the discount ability was used is that there is no price in the log.

"corporation lays tile ..." vs "corporation lays tile xx for 10K"

I'm not really sure if it really needed, but it was easy to add and if it helps some people it could be good enough

crericha:
Perhaps the root of the matter is discounts should be logged more explicitly? That should be straight-forward to enhance.

Zwergenpunk:
Yeah but it is not a discount it is a free tile lay. It is a little bit complicated in 18cz. But maybe that could be logged better.

crericha:
Free tile lays are definitely logged as Corporation (Company) lays...

Its discounts I don't know about.

Zwergenpunk:
I changed it to the default discount output

lib/engine/game/g_18_cz/step/sell_company_and_special_track.rb
```diff
@@ -38,6 +38,14 @@ def process_lay_tile(action)
...
                         @game.current_entity
                       end
+
+            unless @game.purple_tile?(action.tile)
+              discount = action.hex.tile.upgrades.sum(&:cost)
+              @log << "#{action.entity.owner.name} receives a discount of "\
+              "#{@game.format_currency(discount)} from "\
+              "#{action.entity.name}"
+            end
+
```
crericha, line 48: This line in tracker.rb adds the discount info to the log. Is it not working for CZ?

https://github.com/crericha/18xx/blob/master/lib/engine/step/tracker.rb#L231

michaeljb: I've just found something, a few lines up it's looking for a `:tile_discount` ability, but the abilities for these companies here in 18CZ are `:tile_lay` abilities so that's why it's not getting picked up by the existing code for logging the discounts.

The current tile laying code is kind of hard to follow everywhere with the different places costs get calculated and logged and everything, but improving that will take a pretty big effort, and is clearly beyond the scope of this PR 🙂

michaeljb (approved):

crericha:
I think the underlying code will work. The CZ specific code needs a few
tweaks. If I had an example game to work off of, I can fix it up in 10-15
minutes.

Zwergenpunk:
I post an example game with a purple tile and a discoubt placement later

Zwergenpunk:
Fwiw here a gist with the last 2 actions are the abilities used:
https://gist.github.com/Zwergenpunk/a37757c8da2ec0814fc47e919fc04a54

PR #4472, opened 2021-03-12
Added Option to log all players after reorder

### LINKED ISSUE #3959: 18CZ log doesn't include full setting of priority
https://18xx.games/game/27956?action=159

[17:38]OFE's share price changes from K50 to K53
[17:38]beardbru has priority deal
[17:38]-- Operating Round 3.1 (of 1) --

The log only shows who was given priority, not who got position 2, 3, etc. The log should probably just say:

[time]New priority order: beardbru, ChrisTheCat, papavk, Corian09, tgg

Zwergenpunk:
Made it a little bit more generic. Maybe other games can use the default implementation more

CHanged the display of prioty buttons for most_cash orderings to the 1828 style

closes #3959

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2061,7 +2061,7 @@ module Engine
         self.class::NEXT_SR_PLAYER_ORDER
       end
 
-      def reorder_players(order = nil)
+      def reorder_players(order = nil, log_all_players: false)
         order ||= next_sr_player_order
         case order
         when :after_last_to_act
@@ -2076,7 +2076,11 @@ module Engine
           current_order = @players.dup
           @players.sort_by! { |p| [p.cash, current_order.index(p)] }
         end
-        @log << "#{@players.first.name} has priority deal"
+        @log << if log_all_players
+                  "Priority order: #{@players.reject(&:bankrupt).map(&:name).join(', ')}"
+                else
+                  "#{@players.first.name} has priority deal"
+                end
       end
 
       def new_auction_round
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2061,7 +2061,7 @@ module Engine
         self.class::NEXT_SR_PLAYER_ORDER
       end
 
-      def reorder_players(order = nil, log_all_players: false)
+      def reorder_players(order = nil, log_player_order: false)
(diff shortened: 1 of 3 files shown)


lib/engine/game/base.rb
```diff
@@ -2076,7 +2076,11 @@ def reorder_players(order = nil)
           current_order = @players.dup
           @players.sort_by! { |p| [p.cash, current_order.index(p)] }
         end
-        @log << "#{@players.first.name} has priority deal"
+        @log << if log_all_players
+                  "Priority order: #{@players.reject(&:bankrupt).map(&:name).join(', ')}"
```
crericha, line 2080: 1822 has this feature. Does it make sense to use that?

crericha:
FYI 1828.Games added turn order to the player cards. It could be something you want to take advantage of. For me personally, I rather look there than scroll up in the log to find the order.

image

Zwergenpunk:
oh wow thanks. I want that :D

crericha:
> oh wow thanks. I want that :D
> 
> Maybe the log is still good, but your feature sounds really good.

One thing to know is the "Next SR" order shows during the previous SR too as that is relevant for 1828. You'll either want to update the next SR order after each stock action or keep the list empty until the end of the stock round (in which case it won't display).

Zwergenpunk:
> > oh wow thanks. I want that :D
> > Maybe the log is still good, but your feature sounds really good.
> 
> One thing to know is the "Next SR" order shows during the previous SR too as that is relevant for 1828. You'll either want to update the next SR order after each stock action or keep the list empty until the end of the stock round (in which case it won't display until you want it to).

Thanks for the info. 
I think it is ok that its empty during SR. It does not give any information which player has which position at a given time during SR.

lib/engine/game/base.rb
```diff
@@ -2061,7 +2061,7 @@ def next_sr_player_order
         self.class::NEXT_SR_PLAYER_ORDER
       end
 
-      def reorder_players(order = nil)
+      def reorder_players(order = nil, log_all_players: false)
```
crericha, line 2064: A bit of a nitpick, but would 'log_player_order' make more sense for the name of the kwarg?

crericha (approved):

PR #4474, opened 2021-03-13
fix /tiles/all, add fuzzy matching to fix other /tiles and /map paths

### LINKED ISSUE #4305: Tiles page is broken
- Opening https://18xx.games/tiles/all return just this json: '{"error":"nil is not a class/module"}'

michaeljb:
* use a fork of the [jaro_winkler](https://github.com/tonytonyjan/jaro_winkler/) gem to do fuzzy matching (see https://github.com/michaeljb/jaro_winkler/commit/a52c5dc09e363e900a0d997dfab0b819272b522f)

* add Engine method `meta_by_title` to get a game meta via fuzzy matching; match
  against the title, class name, location, and new meta values `GAME_ALIASES`
  and `GAME_SUBTITLE`; this is especially useful for `18LA` -> `18 Los Angeles`
    * use this in `GameClassLoader`
    * get rid of all the `gsub('%20')` stuff

* change `/tiles/<game_titles>` route to `/tiles/<game_titles>/all`. With fuzzy
  matching, it is harder to tell whether a given string should be a tile ID or a
  game title, if only path part is given after `/tiles`, always treat it as a
  tile ID.

* always display the game title on `/tiles` and `/map` pages to make it clear what
  the fuzzy matching resolved to

* add a few subtitles/aliases

[Fixes #4305]

### DIFF
--- a/spec/assets_spec.rb
+++ b/spec/assets_spec.rb
@@ -33,7 +33,9 @@ describe 'Assets' do
       expect(render(app_route: '/tiles/all')).to include('Generic Map Hexes')
 
       expect(render(app_route: '/tiles/57')).to include('57')
-      expect(render(app_route: '/tiles/18Chesapeake')).to include('I9')
+
+      expect(render(app_route: '/tiles/18Chesapeake/all')).to include('I9')
+
       expect(render(app_route: '/tiles/18Chesapeake/I9')).to include('I9')
       expect(render(app_route: '/tiles/18Chesapeake/X1')).to include('X1')
       x2_x3 = render(app_route: '/tiles/18Chesapeake/X2+X3')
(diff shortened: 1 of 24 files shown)


spec/assets_spec.rb
```diff
@@ -47,14 +49,14 @@ def render(**needs)
...
-      multiple_games = render(app_route: '/tiles/1889+18Chesapeake')
+      multiple_games = render(app_route: '/tiles/1889+18Chesapeake/all')
       expect(multiple_games).to include('Kouchi')
       expect(multiple_games).to include('Delmarva')
 
       %w[1889 18Chesapeake].each do |title|
-        expect(render(app_route: "/tiles/#{title}")).to include("#{title} Map Hexes")
-        expect(render(app_route: "/tiles/#{title}")).to include("#{title} Tile Manifest")
-        expect(render(app_route: "/tiles/#{title}")).not_to include('TODO')
+        expect(render(app_route: "/tiles/#{title}/all")).to include("#{title} Map Hexes")
```
tobymao, line 57: cache this? 

html = render(all) so you don't compute this 3x

tobymao:
jaro-winkler is MIT, instead of using a fork, can you just copy the ruby code into our repo?

tobymao:
include the LICENSE.txt

michaeljb:
@tobymao I copied the jaro-winkler code to `lib/engine/jaro_winkler/`. My first attempt had it at `lib/jaro_winkler/` but I couldn't quickly get that to work with how the code is built in `lib/assets.rb`

tobymao (approved):

PR #4483, opened 2021-03-13
1861/67: Add auto-skips where a merge is not possible or buy company …

### LINKED ISSUE #3070: [1867] "Skip Companies" button appearing before paying interest/loans even when the company has no loans.
Also, maybe the action should be "Pay Interest/Loans" instead. I understand that the action was put in because companies can be bought between paying dividends and paying interest/loans, but if a company is train tight with a loan, you get two consecutive "Skip Companies" actions.

### LINKED ISSUE #3076: 1867: Merger button appears when no merge is possible
The merge button appears when no merge is possible

This is due to a performance improvement

jenf:
…before loans if no loans

fixes #3070, fixes #3076

### DIFF
--- a/lib/engine/game/g_1867/step/buy_company_preloan.rb
+++ b/lib/engine/game/g_1867/step/buy_company_preloan.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/buy_company'
+
+module Engine
+  module Game
+    module G1867
+      module Step
+        class BuyCompanyPreloan < Engine::Step::BuyCompany
+          def auto_actions(entity)
+            # If the entity has no loans then passing here makes no difference
+            [Engine::Action::Pass.new(entity)] if entity.loans.empty?
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_1867/step/buy_company_preloan.rb
```diff
@@ -0,0 +1,18 @@
...
+require_relative '../../../step/buy_company'
+
+module Engine
+  module Game
+    module G1867
+      module Step
+        class BuyCompanyPreloan < Engine::Step::BuyCompany
+          def auto_actions(entity)
+            # If the entity has no loans then passing here makes no difference
+            [Engine::Action::Pass.new(entity)] if entity.loans.empty?
```
tobymao, line 12: better to do like 

return [] if entity.loans.empty?
or return nil if ...

jenf: wouldn't that be the opposite of the intended behavior?

tobymao: i meant the style

return [] unless empty?

PR #4490, opened 2021-03-13
Fixed Next SR Position for not Pass 

Zwergenpunk:
@crericha I fixed the issue that the priority for next SR is shown in the current SR. The passing order is only acknowledged for first_pass.

I checked for 1828 and 18CZ and both work as expected.

### DIFF
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -158,8 +158,8 @@ module View
         order = @game.next_sr_player_order
         trs << render_priority_deal(priority_props) if order == :after_last_to_act &&
                                                        @player == @game.priority_deal_player
-        trs << render_next_sr_position(priority_props) if %i[first_to_pass most_cash].include?(order) &&
-                                                          @game.next_sr_position(@player)
+        trs << render_next_sr_position(priority_props, order) if %i[first_to_pass most_cash].include?(order) &&
+                                                          @game.next_sr_position(@player, order)
 
         h(:table, trs)
       end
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -2052,8 +2052,12 @@ def priority_deal_player
         end
       end
 
-      def next_sr_position(entity)
-        player_order = @round.current_entity&.player? ? @round.pass_order : @players
+      def next_sr_position(entity, order)
```
crericha, line 2055: The game class knows the order, so you don't need to pass it in. Call next_sr_player_order directly from here.

lib/engine/game/base.rb
```diff
@@ -2052,8 +2052,12 @@ def priority_deal_player
...
+        player_order = if @round.current_entity&.player?
+                         order == :first_to_pass ? @round.pass_order : []
```
crericha, line 2057: Is this necessary? I believe pass order is [] unless used.

crericha: At my computer now and pass_order is not [].

Zwergenpunk: I want to use passed players only with first-pass. With most_cash I think it would be best to never show the next priority in SR. 

I also could just plain return nil in that case. Maybe that would be clearer

crericha: Its fine as is. I marked this comment as resolved.

assets/app/view/game/player.rb
```diff
@@ -158,7 +158,7 @@ def render_info
         order = @game.next_sr_player_order
         trs << render_priority_deal(priority_props) if order == :after_last_to_act &&
                                                        @player == @game.priority_deal_player
-        trs << render_next_sr_position(priority_props) if %i[first_to_pass most_cash].include?(order) &&
+        trs << render_next_sr_position(priority_props, order) if %i[first_to_pass most_cash].include?(order) &&
```
crericha, line 161: Need to remove order here too

Zwergenpunk: 🙈

crericha (approved):

PR #4494, opened 2021-03-14
Show game info on Create New Game page.

### LINKED ISSUE #4270: Show more info when creating a game?
The Game Info section on the Info tab credits the publisher and designer of a game, and links to the rules and the wiki.  This seems like the kind of information that would be helpful to access before setting up a game.  Should it show also on the Create New Game page?

mavit:
Fixes #4270.

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -12,6 +12,7 @@ module View
     needs :flash_opts, default: {}, store: true
     needs :user, default: nil, store: true
     needs :visible_optional_rules, default: nil, store: true
+    needs :selected_game, default: nil, store: true
 
     def render_content
       @label_style = { display: 'block' }
@@ -25,12 +26,14 @@ module View
         inputs << h(:label, { style: @label_style }, 'Game Options')
         inputs << render_input('Invite only game', id: 'unlisted', type: :checkbox,
                                                    container_style: { paddingLeft: '0.5rem' })
+        inputs << render_game_info
       elsif @mode == :hotseat
         inputs << h(:label, { style: @label_style }, 'Player Names')
         @num_players.times do |index|
           n = index + 1
           inputs << render_input('', id: "player_#{n}", attrs: { value: "Player #{n}" })
         end
+        inputs << render_game_info
       elsif @mode == :json
         inputs << render_upload_button
         inputs << render_input(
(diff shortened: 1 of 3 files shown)


assets/app/view/create_game.rb
```diff
@@ -272,6 +279,8 @@ def update_inputs
         store(:num_players, range.value.to_i)
       end
 
+      store(:selected_game, selected_game)
```
tobymao, line 282: store rerenders the view unless you do skip: true, so change this one and the store(:num_players to have skip: true

mavit: Done, thanks.

assets/app/view/game/game_meta.rb
```diff
@@ -3,7 +3,13 @@
 module View
   module Game
     class GameMeta < Snabberb::Component
-      needs :game
+      needs :game, default: nil
+      needs :game_class, default: nil
+
+      def initialize(*)
+        super
+        @game_class = @game.class unless defined? @game_class
```
tobymao, line 11: what's going on here... there's got to be a better way,

@michaeljb suggestions?

mavit: `View::Game::GameMeta` expects `game` to be an `Engine::Game::…::Game` object, but in `View::CreateGame` we have an `Engine::Game::…::Meta` class, not an instantiated object.  Since `View::Game::GameMeta` mostly accesses class variables that are common between the `Game` and the `Meta` classes, I wanted to be able to be able to pass it a class rather than an instance.

The obvious thing I tried was:

```ruby
needs :game, default: nil
needs :game_class, default: @game.class
```

However, this doesn’t work because `@game` doesn’t get set early enough.

The other obvious thing I tried was to instantiate the `Meta` class before passing it to `View::Game::GameMeta`, but this throws, e.g., ``undefined method `new' for Engine::Game::G1817::Meta``.  I admit I don’t understand why that is, and whether it would be reasonable to “fix” it.

michaeljb: `Meta` is a module, not a class, so that's why it cannot be instantiated with `new`.

Looks like the only thing in `GameMeta` that needs an actual `Game` instance is the optional rules bit--it wants to show which optional rules were used in creating the current game.

Probably can add a little bit of logic in `GameMeta#render` to determine if `@game` is a `Game` instance or a `Meta` module, I'll play around with it a bit.

michaeljb: @mavit I've made a PR onto your branch, if you merge it, this PR will be updated automatically

https://github.com/mavit/18xx/pull/1

This adds `.meta` and `.game_instance?` methods to `Game` classes, their instances, and `Meta` modules, so the `Meta` is always easy to reach now.

Screenshot 2021-03-15 092032

michaeljb (approved):

PR #4498, opened 2021-03-14
Add player colors on entity page

### LINKED ISSUE #2207: 1817 - Make it easier to see who owns what

jenf:
Also add that Track 1 = Your color

fixes #2207

### DIFF
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -43,7 +43,7 @@ module View
 
       def render_title
         bg_color = if setting_for(:show_player_colors, @game)
-                     player_colors(@game.players)[@player]
+                     player_colors(@game.players)[@player] || color_for(:bg2)
                    else
                      color_for(:bg2)
                    end
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -9,6 +9,7 @@ module View
     class Player < Snabberb::Component
       include Lib::Settings
       include Lib::Text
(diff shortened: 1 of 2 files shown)


yzemaze (approved):

assets/app/view/game/player.rb
```diff
@@ -41,11 +42,16 @@ def render
       end
 
       def render_title
+        bg_color = if setting_for(:show_player_colors, @game)
+                     player_colors(@game.players)[@player] || color_for(:bg2)
```
yzemaze, line 46: I don’t really care about it, but something like `convert_hex_to_rgba(route_prop(idx % ROUTE_COLORS.size, 'color'), 0.5)` for players 9+ in `player_colors` might work.
=> no merge, because I’m not sure if you want to invest more time into this.

jenf: Nor me :)

PR #4500, opened 2021-03-14
1882: Fix NWR private and OO tile if it already has a token on it

### LINKED ISSUE #4379: [1882] Cannot choose token location after green upgrade in OO with two tokens
https://18xx.games/game/27623?action=348

After laying the green tile, CPR cannot choose to place its token in the location occupied by SC. It should be able to choose the location of both tokens.

jenf:
The OO tile can have SC, this causes two tokens
to be present on the tile in the token choice condition which must be selected which token goes where.

fixes #4379

### DIFF
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -204,16 +204,18 @@ module Engine
         if old_tile.paths.empty? &&
             !tile.paths.empty? &&
             cities.size > 1 &&
-            (token = cities.flat_map(&:tokens).find(&:itself))
-          actor = entity.company? ? entity.owner : entity
-          @round.pending_tokens << {
-            entity: actor,
-            hexes: [action.hex],
-            token: token,
-          }
-          @log << "#{actor.name} must choose city for token"
+            !(tokens = cities.flat_map(&:tokens).select(&:itself)).empty?
+          tokens.each do |token|
+            actor = entity.company? ? entity.owner : entity
+            @round.pending_tokens << {
+              entity: actor,
+              hexes: [action.hex],
+              token: token,
+            }
+            @log << "#{actor.name} must choose city for token"
 
-          token.remove!
+            token.remove!
+          end
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/engine/step/tracker.rb
```diff
@@ -204,16 +204,18 @@ def update_token!(action, entity, tile, old_tile)
...
-          actor = entity.company? ? entity.owner : entity
-          @round.pending_tokens << {
-            entity: actor,
-            hexes: [action.hex],
-            token: token,
-          }
-          @log << "#{actor.name} must choose city for token"
-
-          token.remove!
+            !(tokens = cities.flat_map(&:tokens).select(&:itself)).empty?
```
tobymao, line 207: cities.any? { |c| c.tokens.any?  }

cities.flat_map(&:tokens).any?

michaeljb: Neither of those will properly set the value of `tokens`.

Here's something lengthier than what Jen wrote but it passes over the list just once, and avoids using `any?`-- since we use avoid using `any?` to mean `.size > 0`, in some places, I think it would be weird to use it to check for non-nil here.

```
(tokens = cities.reduce([]) { |t, c| t.concat(c.tokens.compact) }).empty?
```

michaeljb: Or for something a touch closer to the current code:

```
(tokens = cities.reduce([]) { |t, c| t.concat(c.tokens.select(&:itself)) }).empty?
```

tobymao: i’d don’t realize you were using a setter

couldn’t you do cities.flatmap(&tokenz).compact

michaeljb: yeah that should work too

michaeljb:
```suggestion
            !(tokens = cities.flat_map(&:tokens).compact).empty?
```

PR #4505, opened 2021-03-14
[1846, bug fix] fix closing corporations during bankruptcy

### LINKED ISSUE #4079: [1846] Corporation that closed during bankruptcy still got an OR turn
The C&O's president went bankrupt (while operating PRR) here - https://18xx.games/game/28841?action=354

Expected behavior: C&O is not in the list of corporations to act this OR

Actual behavior: C&O has a share price of $0 but is the active corporation - https://18xx.games/game/28841?action=376

michaeljb:
[Fixes #4079]

Also, at game end, make sure the Message step is usable so chat can continue; in game 16904 (added as a fixture here), the game ends when one player goes bankrupt while operating the GT; the GT's price was so low that when the player is compelled to sell all shares and go bankrupt, the GT's price drops to $0, closing it. I'm not 100% sure what's happening but the active entity doesn't get moved to the next one, and all steps (including Message) get passed.

One game needs to be pinned/archived (it has already finished):

* https://18xx.games/game/28158?action=366 - ERIE enters receivership at $0 but should have closed; it buys a 4/6 train then closes, removing that 4/6 from the game
* https://18xx.games/game/28158?action=460 - NYC has $739, its president contributes $61 to buy the first grey train, a 6 train; with the new code, a 4/6 train is still in the depot, so NYC has enough cash to buy that train outright


lib/engine/game/base.rb
```diff
@@ -2041,7 +2041,10 @@ def game_ending_description
 
       def action_processed(_action)
         @corporations.dup.each do |corporation|
-          close_corporation(corporation) if corporation.share_price&.type == :close
+          if corporation.share_price&.type == :close
+            close_corporation(corporation)
+            corporation.close!
```
michaeljb, line 2046: calling `corporation.close!` inside of `close_corporation()` instead of here breaks a few 1861/1867 games at the merger step. I'm not familiar with those games yet so I didn't dig deeper to see if this should be fixed another way.

1849 and 18CO both also have code that calls `close_corporation(corporation)` and then separately calls `corporation.close!`.

michaeljb:
This fix is incomplete, found a couple cases where a corporation drops to $0 and should close but doesn't, then proceeds to buy a 4/6 train and then closing.

michaeljb:
> This fix is incomplete, found a couple cases where a corporation drops to $0 and should close but doesn't, then proceeds to buy a 4/6 train and then closes.

Got this fixed, this PR is ready to go now.

lib/engine/game/base.rb
```diff
@@ -907,6 +907,7 @@ def end_game!
         @round_counter += 1
         scores = result.map { |name, value| "#{name} (#{format_currency(value)})" }
         @log << "-- Game over: #{scores.join(', ')} --"
+        @round.steps.find { |s| s.is_a?(Engine::Step::Message) }.unpass!
```
tobymao, line 910: can you just make message step unpassable???

tobymao: how can someone even pass message?

michaeljb: On this branch, it happens in game 16904. Actually it happens that way on master as well every time the active corporation closes, but usually the next one becomes active right away so it doesn't matter (on this branch and master, `force_next_entity!`, called [here](https://github.com/tobymao/18xx/blob/c069a914d0e9bedf222a0139e7868467a700f068/lib/engine/game/base.rb#L1308) passes all the steps [here](https://github.com/tobymao/18xx/blob/c069a914d0e9bedf222a0139e7868467a700f068/lib/engine/round/operating.rb#L49))

* the bankruptcy leaves one active player, so it ends the game
* the active corporation also closes during this bankruptcy--at the point where the president must declare bankruptcy, the active corp's price is $10; as part of going bankrupt, the president sells shares they're normally not allowed to sell, thus dropping the price to $0 and closing it
* `force_next_entity!` passes the step
* the game is over, so there's no new entity/round/step to be active

I'll change it to make message unpassable, that sounds better, and goes along with how Message is also unskippable

michaeljb: Fixed in 16904

Issue #4181, opened 2021-02-27
1846 bankruptcy--when selling shares that normally can't be sold, the prices should not be adjusted


michaeljb:
#4176 is somehow having an effect on the order that shares are sold when a player goes bankrupt, which matters for operating order.

https://18xx.games/game/13423?action=250 -- Thirdman sells ERIE then IC on the live site (currently at cbc5a44f),  so that's their operating order the next round. With 0c9cb6d87 checked out locally (PR 4176 is the only new change at that point), IC gets sold first, and thus operates first in the next OR.

https://github.com/tobymao/18xx/blob/194dc8a3dca812df27d2cb6ea51b5481b48dfefe/lib/engine/game/g_1846/step/bankrupt.rb#L38-L60 ^this is the code that should be updated; `shares_by_corporation` returns a hash so I guess the order is not well-defined there; we should iterate over sorted corporations so that everything is sold in operating order so that it is maintained.

game 13423 in gist form - https://gist.github.com/michaeljb/c0903fca504c5a9566eae2d8ba05f2a0

michaeljb:
> returns a hash so I guess the order is not well-defined there

ruby hashes are ordered, so it is well-defined, it's just not defined in a way that matches operating order

michaeljb:
https://boardgamegeek.com/thread/2621786/article/37259976#37259976

michaeljb:
> #4176 is somehow having an effect on the order that shares are sold when a player goes bankrupt, which matters for operating order.

That "somehow" must be that the `can_sell_any?` added a call to `shares_by_corporation` someplace in the stack, and getting the order of the keys changed up.

michaeljb:
From Tom Lehmann in https://boardgamegeek.com/thread/2621786/article/37259976#37259976:

> This was addressed in the 2nd Edition 1846 Rules (currently sitting with 2nd Edition 1846 at the printer):
>
> 7.11 After a player has gone bankrupt due to a forced train purchase (see 6.89), put the cash the player raised in the corporation’s treasury. Sell all remaining shares owned by this player to the Stock Market without adjusting stock prices, even if this puts more than 50% of a corporation’s stock or a President certificate there. Add the money raised to the corporation’s treasury. This may result in a Presidency change in the corporation currently operating. If so, the new President must immediately purchase a train for it. (emphasis added)

Updated ticket title to reflect this.

PR #4521, opened 2021-03-16
[1817, 1846, many others] during bankruptcy, sell remaining shares based on operating order

michaeljb:
If sort is true, then the hash's keys are sorted so that when the caller
iterates through the corporations, they iterate in operating order. This is
relevant for games that do not end in bankruptcy, where the order the
bankrupting player sells the shares can impact future operating order.

Technically the current site behavior for 1817 does not violate the rules, but
as #4181 describes, the return value of `shares_by_corporation` can potentially
be reordered by a seemingly unrelated code change. Getting a well defined order
in place for selling these shares avoids that being a problem in the future.

----

This breaks breaks about 6% of 1817-family games. No other games are broken, though many will see a change in the logged order of shares sold during bankruptcy.


michaeljb:
will reopen when I have a replacement 1817NA fixture

tobymao (approved):

PR #4531, opened 2021-03-16
[1822] Lur

### LINKED ISSUE #4522: [1822] Move Card Private is Broken or Needs a UX Overhaul
https://18xx.games/game/32904?action=1176

The NER owns the "Move Card" private, but we were unable to make it work. Numerous players tried in game, and in cloned hotseat games, and we were unable to figure out when/where to click to move a card.  

Either it's broken or it needs to be easier to use.

magnusfalt:
- Made the LUR ability shows itself during the track lay and token step as well.

fixes #4522

This change does not affect any of the currect games.

### DIFF
--- a/lib/engine/game/g_1822/game.rb
+++ b/lib/engine/game/g_1822/game.rb
@@ -3451,7 +3451,8 @@ module Engine
         end
 
         def company_choices_lur(company, time)
-          return {} if !(time == :token || time == :track || time == :issue) || !company.owner&.corporation?
+          return {} if time != :token && time != :track && time != :issue
+          return {} unless company.owner&.corporation?
 
           exclude_minors = bidbox_minors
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1822/game.rb
```diff
@@ -3451,7 +3451,7 @@ def company_choices_lcdr(company, time)
         end
 
         def company_choices_lur(company, time)
-          return {} if time != :issue || !company.owner&.corporation?
+          return {} if !(time == :token || time == :track || time == :issue) || !company.owner&.corporation?
```
jenf, line 3454: would probably split the two sides of the or into two to make it clearer
return {} unless company.owner&.corporation?

magnusfalt: Sure it makes sense :)

PR #4532, opened 2021-03-16
[1822MRS] 1822 Medium Regional Scenario

magnusfalt:
This is the 1822 Medium Regional Scenario

- Inherited the 1822 base class.
- Changes cert limit, starting cash, map, tile and train constants.
- Changes the cost for upgrade L -> 2, first one to do it costs £80. After we entered phase 2 it costs £70.
- Advanced variant where you play with P1-12 and start with M14 in bid box 1.

### DIFF
--- a/lib/engine/game/g_1822_mrs/game.rb
+++ b/lib/engine/game/g_1822_mrs/game.rb
@@ -654,6 +654,12 @@ module Engine
 
         UPGRADE_COST_L_TO_2_PHASE_2 = 70
 
+        def bidbox_start_minor
+          return self.class::BIDDING_BOX_START_MINOR_ADV if optional_advanced?
+
+          self.class::BIDDING_BOX_START_MINOR
+        end
+
         def discountable_trains_for(corporation)
           discount_info = []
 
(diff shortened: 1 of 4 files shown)


jenf (approved):

jenf:
My only slight concern is the use of
self.class.const_set
Want to check with Toby

lib/engine/game/g_1822_mrs/game.rb
```diff
@@ -0,0 +1,688 @@
...
+                         end
+          corporation.trains.select { |t| t.name == 'L' }.each do |train|
+            discount_info << [train, train, '2', upgrade_cost]
+          end
+          discount_info
+        end
+
+        def init_companies(players)
+          # Make sure we have the correct starting companies
+          self.class.const_set(:STARTING_COMPANIES, self.class::STARTING_COMPANIES_ADVANCED) if optional_advanced?
```
tobymao, line 673: don't use const set, just turn this into an instance method with an override

magnusfaltinbooks: Ok will do

tobymao (approved):

PR #4543, opened 2021-03-16
18Ireland: Add stock market and corporations

jenf:
(also attempting to see how splitting config would work)


underhilllabs:
I like this new format! I think game.rb will be much cleaner, pulling the entity config out into it's own module.

lib/engine/game/g_18_ireland/misc_config.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Engine
```
tobymao, line 3: do we really need two configs?

tobymao (approved):

lib/engine/game/g_18_ireland/misc_config.rb
jenf: it's probably worth the discussion with others, personally i like things split into categories so i can quickly find them. But having multiple files can be awkward as well.

jenf: discussion complete

PR #4544, opened 2021-03-17
Update game.rb

philcampeau:
Corrected top right corner of the market. It could be relevant in a situation where a company is sold out, gets a price bump into the sold out section, and then still runs once more.


michaeljb:
dup of #4547

PR #4551, opened 2021-03-17
[18Neb] fixes train, changes module name to 18Neb

underhilllabs:
- fixes 5/7 train
- changes module name to 18Neb
- adds track cost for second tile lay


underhilllabs:
Reopening as the tests worked locally. I think there was just an issue with Shikoku 1889.

tobymao (approved):

PR #4562, opened 2021-03-17
18 neb new config

underhilllabs:
I mostly did this pull request to see what Jen's config changes would look like if everything was pulled into a different config file. It's probably too much. Although I do like having a separate map config with HEXES. TILES. LOCATIONS, etc.

As I said in Jen's pull request I think game.rb looks better with the large config entities and arrays pulled out.

@jenf


jenf:
I think that's too many files, but having entities, map and a 'misc' as a separate file works for me. but i'm not sure if toby agrees

jenf:
Let us discuss it for a while and get back to you.

jenf:
Basically we went with a simpler approach just moving a few long winded elements:

map.rb - Tiles and maps constants needed for the game (just constants)
entities.rb - Corporations and companies needed for the game (just constants)

PR #4577, opened 2021-03-18
18fl alpha

ventusignis:
Tokens cost 20 x distance
Game is alpha


lib/engine/game/g_18_fl/step/token.rb
```diff
@@ -42,19 +47,85 @@ def process_hex_token(action)
...
+
           def tokened(hex, entity)
             hex.tile.icons.any? { |i| i.name == entity.id }
           end
+
+          # How far is the start_hex from one of the corporation's stations?
+          # This is by track, not as the crow flies.
+          # start is either a city or hex (plain track or town)
+          def distance_to_station(corporation, start)
+            goal_hexes = corporation.tokens.select(&:city).map { |t| [t.city.hex, t.city] }
```
ventusignis, line 77: I'm going to try re-implementing this w/ the second value I had `node.walk` yield; I think it might work but I'm not sure yet

ventusignis: It works

lib/engine/game/g_18_fl/step/token.rb
```diff
@@ -42,16 +48,36 @@ def process_hex_token(action)
...
 
+          def token_cost_override(entity, city_hex, _slot, token, _tokener)
+            node = city_hex.respond_to?(:city?) ? city_hex : city_hex.tile.towns.first
+            min_distance = INFINITE_DISTANCE
+            goal_cities = entity.tokens.select(&:city).map(&:city)
+
+            node.walk(corporation: entity) do |path, visited_paths|
+              if goal_cities.include?(path.city)
+                # minus one to account for the tokened hex getting included in the count
+                distance = visited_paths.select { |_k, v| v }.keys.map(&:hex).uniq.size - 1
```
tobymao, line 73: visited_paths.uniq { |k, _| k.hex }.size - 1

v is always true

lib/engine/game/g_18_fl/step/token.rb
```diff
@@ -42,16 +48,36 @@ def process_hex_token(action)
...
-            @game.log << "#{entity.name} places a hotel on #{hex.name}"
+            @game.log << "#{entity.name} places a hotel on #{hex.name} for #{@game.format_currency(cost)}"
+            entity.spend(cost, @game.bank)
 
             entity.tokens.delete(token)
             hex.tile.icons << Part::Icon.new("../logos/18_fl/#{entity.id}")
             pass!
           end
 
+          def token_cost_override(entity, city_hex, _slot, token, _tokener)
```
tobymao, line 65: is this only run once ever? does it use auto actions  or is it not calculated when cost is set?

ventusignis: It is run once when creating the action in the view

lib/engine/game/g_18_fl/step/token.rb
ventusignis: Thanks, I 've updated this

lib/engine/game/g_18_fl/step/token.rb
ventusignis: I made it so the step verifies the cost (only when not @loading) so the server can verify the cost
This increases it to running twice total, but it's only when clicking to place a token

tobymao (approved):

PR #4579, opened 2021-03-18
Allow linking to a specific game title to be created

mavit:
Would be handy for, e.g., linking to a specific game from the wiki or elsewhere on the web.


michaeljb:
@mavit FYI this has some conflicts with my recent changes in #4573, I'm looking into how to resolve them

michaeljb:
Conflicts resolved, now I just need to make this work with game "variants" in the new variant checkbox system 😁

PR #4588, opened 2021-03-18
18ZOO - refactor

carlorusso:
Use Abilities to show the powers (instead of choice)
Refator trains to be numeric
Override every step to use the module ChooseAbilityOnOr for choose_ability and process_choose_ability


lib/engine/game/g_18_zoo/step/dividend.rb
```diff
@@ -5,6 +5,8 @@ module Game
     module G18ZOO
       module Step
         class Dividend < Engine::Step::Dividend
+          include Engine::Game::G18ZOO::ChooseAbilityOnOr
```
michaeljb, line 8: Why does this need to be added on all the steps?  Is it not enough for the `BuyOrUsePowerOnOr` step to have this module? Or does each step in the OR actually process the power actions?

michaeljb: FWIW I removed this include line from all the step classes other than `BuyOrUsePowerOnOr` and `bundle exec rspec -e ZOO` still passes

spec/fixtures/18ZOO/or_power.a_spoonful_of_sugar.not_yet_working
```diff
@@ -0,0 +1,38 @@
+{
```
michaeljb, line 1: I don't think we need these "not yet working" fixtures checked in.

spec/fixtures/18ZOO/sr_power.midas.json
```diff
@@ -0,0 +1,106 @@
+{
```
michaeljb, line 1: I just pushed e04c9112f to fix accessing these fixtures in the browser; this fixture fails to render at action 9

http://localhost:9292/fixture/18ZOO/sr_power.midas?action=8 works, then hitting next in the history controls breaks

michaeljb: actually looks like more than of these fixtures are failing to render, with the same error:

Screenshot 2021-03-20 215324

michaeljb: these are the others that fail:
* sr_power.too_much_responsibility
* sr_power.leprechaun_pot_of_gold
* sr_power.it_is_all_greek_to_me

lib/engine/game/g_18_zoo/step/dividend.rb
carlorusso: yes, the problem is that BuyOrUsePowerOnOr is not a blocking step, so if you are for example in "buy_train" step, and you try to execute any of the "choose_ability" action, it will fail with a "choose_ability not defined inside the current step". I found the problem here:
abilities.rb
`
      def render_ability_choice_buttons
        ability_choice_buttons = @game.round.active_step.choices_ability(@selected_company).map do |choice, label|
`
this piece of code call active_step without passing the selected company, so the next method find only if active and blocking

round/base.rb
`
      def active_step(entity = nil)
        return @steps.find { |step| step.active? && step.actions(entity).any? } if entity

      end
`

In another PR I would like to change this to pass also the @selected_company to the active_step, but that can change the behaviour of multiple game, so I need to test it really better

spec/fixtures/18ZOO/or_power.a_spoonful_of_sugar.not_yet_working
carlorusso: ok, I'll remove, thanks

lib/engine/game/g_18_zoo/step/dividend.rb
carlorusso: I just added also a test `sell_ticket_zoo_in_any_step` to check that choose_actions works in any step

spec/fixtures/18ZOO/sr_power.midas.json
michaeljb: Confirmed fixtures are fixed by 28cff0187 👍

michaeljb (approved):

lib/engine/game/g_18_zoo/step/choose_ability_on_or.rb
```diff
@@ -0,0 +1,261 @@
...
+        # TODO: add additional logic
+        # def can_choose_sugar?(corporation)
+        #   @game.a_spoonful_of_sugar.owner == corporation &&
+        #     corporation.trains.any? { |train| !%w[2J 4J].include? train.name } &&
+        #     corporation.all_abilities.none? { |a| a.type == :increase_distance_for_train }
+        # end
+
+        def choices_for_zoo_ticket(company)
+          corporation = @game.current_entity
+          name = company.owner.name
```
crericha, line 136: Do you need to verify that the corporation is owned by the same person as the company?

lib/engine/game/g_18_zoo/step/choose_ability_on_or.rb
```diff
@@ -0,0 +1,261 @@
...
+          player = company.owner
+          corporation = @game.corporation_by_id(action.choice['corporation'])
+          price_for_corporation = action.choice['price']
+          price_for_player = company.value - price_for_corporation
+
+          @log << "-- #{corporation.name} sells #{company.name} (owned by #{player.name}) "\
+            "for #{@game.format_currency(company.value)} --"
+
+          @game.bank.spend(price_for_player, player) if price_for_player.positive?
+          @log << "#{player.name} earns #{@game.format_currency(price_for_player)}" if price_for_player.positive?
```
crericha, line 210: encapsulate in one if statement instead of checking if price_for_player is positive twice

lib/engine/game/g_18_zoo/step/choose_ability_on_sr.rb
```diff
@@ -0,0 +1,153 @@
...
+          return can_choose_greek?(entity) if company == @game.it_is_all_greek_to_me
+
+          false
+        end
+
+        def can_choose_any_ability?(entity)
+          return false unless entity.player?
+
+          return true if @game.zoo_tickets?(entity)
+          return can_choose_midas?(entity) if @game.midas.owner == entity
```
crericha, line 27: Can one player own multiple abilities? If so, do you need to check those too before returning?

crericha: Also from experience, I've found its better to have this type of logic in the game class instead of the steps. Something to consider as you develop the game.

lib/engine/game/g_18_zoo/step/choose_ability_on_or.rb
carlorusso: done in 2edf5ad23

lib/engine/game/g_18_zoo/step/choose_ability_on_or.rb
carlorusso: done in 2edf5ad23

lib/engine/game/g_18_zoo/step/choose_ability_on_sr.rb
carlorusso: done in 2edf5ad23

carlorusso: I didn't move into game class, because for this specific case I'm using them only a single Step.
for the choose_ability_on_or I'll move on game later, as you suggest, because I will need them for more than one step

PR #4594, opened 2021-03-19
Highlight laid tiles

crericha:
Highlights tiles laid this turn. Also highlights last tiles laid for other corporations via the  the "Show Last Route and Track For" drop-down.

For tiles with a border, uses contrast instead of black for the highlight.

image

### DIFF
--- a/assets/app/lib/hex.rb
+++ b/assets/app/lib/hex.rb
@@ -30,7 +30,6 @@ module Lib
       black: '#000000',
       red: '#ec232a',
       blue: '#35A7FF',
-      purple: '#800080',
     }.freeze
 
     def self.points(scale: 1.0)
--- a/assets/app/lib/hex.rb
+++ b/assets/app/lib/hex.rb
@@ -33,9 +33,9 @@ module Lib
     }.freeze
 
     def self.points(scale: 1.0)
-      "#{X_R * scale},#{Y_M * scale} #{X_M_R * scale},#{Y_B * scale} "\
-      "#{X_M_L * scale},#{Y_B * scale} #{X_L * scale},#{Y_M * scale} "\
-      "#{X_M_L * scale},#{Y_T * scale} #{X_M_R * scale},#{Y_T * scale}"
+      "#{(X_R * scale).round(3)},#{(Y_M * scale).round(3)} #{(X_M_R * scale).round(3)},#{(Y_B * scale).round(3)} "\
+      "#{(X_M_L * scale).round(3)},#{(Y_B * scale).round(3)} #{(X_L * scale).round(3)},#{(Y_M * scale).round(3)} "\
+      "#{(X_M_L * scale).round(3)},#{(Y_T * scale).round(3)} #{(X_M_R * scale).round(3)},#{(Y_T * scale).round(3)}"
     end
   end
 end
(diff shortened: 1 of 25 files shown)


crericha:
@yzemaze FYI

assets/app/lib/hex.rb
```diff
@@ -30,6 +30,7 @@ module Hex
       black: '#000000',
       red: '#ec232a',
       blue: '#35A7FF',
+      purple: '#800080',
```
yzemaze, line 33: This allows users to config the color, because user.rb iterates over Lib::Hex::COLOR. The config wouldn’t have any effect though (not saved, not accessed in view/game/hex). It works, but isn’t ideal.
So I’d either remove this line and change "purple" to hex value in 18CZ’s and 1873’s config or render frame in user selected color. The latter would entail using named colors for possible tile frames in the future, too.

assets/app/view/game/hex.rb
```diff
@@ -106,6 +115,18 @@ def hex_outline
         end
       end
 
+      def hex_highlight
+        polygon_props = { attrs: { points: HIGHLIGHT_POINTS } }
+        polygon_props[:attrs]['fill-opacity'] = 0
+        polygon_props[:attrs]['stroke-dasharray'] = 10
+        polygon_props[:attrs]['stroke-width'] = HIGHLIGHT_STROKE_WIDTH
+        if (color = @tile&.frame&.color)
+          polygon_props[:attrs]['stroke'] = contrast_on(color_for(color))
```
yzemaze, line 124: `color_for` only works for named colors and will fail for hex values.

assets/app/view/game/map_controls.rb
```diff
@@ -81,13 +87,15 @@ def route_controls
...
+            store(:historical_laid_hexes, last_laid_hexes(operator))
           else
             store(:historical_routes, [])
+            store(:historical_laid_hexes, nil)
           end
         end
 
         @route_input = render_select(id: :route, on: { input: route_change }, children: operators)
-        h('label.inline-block', ['Show Last Route For:', @route_input])
+        h('label.inline-block', ['Show Last Route and Track For:', @route_input])
```
yzemaze, line 98: I’d prefer "Tile", because tiles consist of more than tracks. (The same for comments in other files.)

crericha: Changed Track to Tile

assets/app/lib/hex.rb
crericha: Moved to tile config in game classes

yzemaze:
@crericha Sorry, didn’t intend to push to here …
Anyway, ante/post:
image.image

image.image

crericha:
Looks good. Thanks.

assets/app/view/game/map_controls.rb
```diff
@@ -81,13 +87,15 @@ def route_controls
           operator = all_operators.find { |o| o.name == operator_name }
           if operator
             store(:historical_routes, generate_last_route(operator))
+            store(:historical_laid_hexes, last_laid_hexes(operator))
```
tobymao, line 90: you need to add skip: true to both of these so you don't rerender twice

crericha: Good catch. Fixed.

PR #4597, opened 2021-03-19
[1822NRS] North Regional Scenario

magnusfalt:
This is the 1822 North Regional Scenario

- Inherited the 1822 base class, and added as a variant
- Changes cert limit, starting cash, map, tile and train constants.
- Changes the cost for upgrade L -> 2, first one to do it costs £80. After we entered phase 2 it costs £70.
- No special minor in bid box 1

### DIFF
--- a/lib/engine/game/g_1822_nrs/game.rb
+++ b/lib/engine/game/g_1822_nrs/game.rb
@@ -1,0 +1,520 @@
+# frozen_string_literal: true
+
+require_relative '../g_1822/game'
+require_relative 'meta'
+
+module Engine
+  module Game
+    module G1822NRS
+      class Game < G1822::Game
+        include_meta(G1822NRS::Meta)
+
+        BIDDING_BOX_START_MINOR = nil
+
+        CERT_LIMIT = { 3 => 16, 4 => 13, 5 => 10, 6 => 9, 7 => 8 }.freeze
+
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_1822_nrs/game.rb
```diff
@@ -0,0 +1,520 @@
...
+              },
+            ],
+            num: 2,
+            price: 0,
+          },
+        ].freeze
+
+        UPGRADE_COST_L_TO_2_PHASE_2 = 70
+
+        def discountable_trains_for(corporation)
```
michaeljb, line 481: I see this function is the same as 1822MRS. Are their rules similar enough that it would make sense for NRS to be a subclass of MRS?

scottredracecar: Besides the setup differences (different corps, privates, minors), there is one fewer 3 train and some of the tiles are different. The stock market is reused directly. Seems like there are enough differences that it would be best as is? Not sure.

magnusfalt: Its just that specific method and const that is the same. Everything else is different. Its working great with your new variant as well so i would recommend to keep them as seperate.

michaeljb (approved):

tobymao (approved):

PR #4604, opened 2021-03-19
improve fuzzy matching

michaeljb:
- match against `full_title` and `SUPERTITLE` (in addition to `title` and
  `SUBTITLE`)
- once list of candidates is compiled, also use individual words from each
  candidate (e.g., put "Wight" in the list for 1860)
- create `fuzzy_candidates` hash just once
- in `closest_title()`, don't upcase the given title if it matches a real
  title (avoids doing unnecessary checks for "18Chesapeake" for example; old
  code was converting it to "18CHESAPEAKE" then failing to find a match in the
  initialized `fuzzy_titles`, so then it had to calculate the distance against
  all candidate strings from all games)
- remove unnecessary period from end of 1817WO's location
- add a couple `closet_title()` specs for individual subtitle words

### DIFF
--- a/lib/engine.rb
+++ b/lib/engine.rb
@@ -51,26 +51,35 @@ module Engine
 
   def self.closest_title(title)
     return VISIBLE_GAMES.first.title unless title
+    return @fuzzy_titles[title] if @fuzzy_titles[title]
 
     title = title.upcase
-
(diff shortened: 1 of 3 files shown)


michaeljb:
This was nagging my mind when I woke up this morning, will get back to looking at 18ZOO later today 😉

PR #4610, opened 2021-03-19
[2038] Fix stock round corporation grouping

jason-johnson:
In the standard game only "Group A" corporations are initially available.  After all corporations in that group IPO the next group becomes available.  The same repeats again until the final group C become available.

Also fixed the corporation colours to match the real game.


lib/engine/game/g_2038/game.rb
```diff
@@ -540,7 +540,7 @@ class Game < Game::Base
             simple_logo: '1830/PRR.alt',
             tokens: [60, 100, 60, 100, 60, 100, 60, 100, 60, 100],
             coordinates: 'K9',
-            color: '#32763f',
+            color: '#40b1b9',
             type: 'groupA',
```
tobymao, line 544: change to snake_case

lib/engine/game/g_2038/game.rb
```diff
@@ -652,6 +652,52 @@ def game_hexes
...
+
+          @b_group_corporations, @c_group_corporations = @b_group_corporations.partition do |corporation|
+            corporation.type == :groupB
+          end
+        end
+
+        def event_group_b_corps_available!
+          @log << 'Group B corporations are now available'
+
+          @corporations += @b_group_corporations
```
tobymao, line 682: @corporations <<

lib/engine/game/g_2038/game.rb
```diff
@@ -652,6 +652,52 @@ def game_hexes
...
+          @b_group_corporations = []
+          @available_corp_group = :groupB
+        end
+
+        def event_group_c_corps_available!
+          @log << 'Group C corporations are now available'
+
+          @corporations += @c_group_corporations
```
tobymao, line 690: same?

lib/engine/game/g_2038/game.rb
jason-johnson: @b_group_corporations is an array, I wouldn't want to append that, not insert it, right?

tobymao: ok, then do concat

tobymao (approved):

PR #4611, opened 2021-03-19
show prealpha games on dev-env by default, hide on prod completely

### LINKED ISSUE #4612: [All] Impossible to create Hotseat prealpha game
**How to Reproduce the Bug?**

1. Create a new game
2. Add "?all" to the URL so prealpha games show up
3. Select Hotseat

Here you see the player count, etc. get in a strange state because the Meta is still from the selected game but the title is now 1817.

image

**Which Browser / Platform?**

Safari 14.0.3 / Mac OSX 11.2.3

yzemaze:
=> `?all` is gone for create_game on all dev-envs.

[Fixes #4612]

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -3,8 +3,6 @@
 require 'game_manager'
 require 'view/form'
 
-VISIBLE_GAMES = (ENV['RACK_ENV'] == 'production' ? Engine::VISIBLE_GAMES : Engine::GAME_METAS).sort
-
 module View
   class CreateGame < Form
     include GameManager
@@ -18,6 +16,7 @@ module View
     needs :game_variants, default: nil, store: true
     needs :selected_variant, default: nil, store: true
     needs :title, default: nil
+    needs :production, default: nil
 
     def render_content
       @label_style = { display: 'block' }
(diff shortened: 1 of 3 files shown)


assets/app/view/create_game.rb
```diff
@@ -3,6 +3,8 @@
 require 'game_manager'
 require 'view/form'
 
+VISIBLE_GAMES = (ENV['RACK_ENV'] == 'production' ? Engine::VISIBLE_GAMES : Engine::GAME_METAS).sort
```
michaeljb, line 6: Looks like Opal can't access `ENV['RACK_ENV']` here--I started up a local "production" server and `ENV` is `{}` so `VISIBLE_GAMES` here gets set to all of the games.

`api.rb` sets a `PRODUCTION` constant ([here](https://github.com/tobymao/18xx/blob/90baf49a2a3619aa62dbff1077eb2ac3114312f9/api.rb#L3)), that could be added to the needs [here](https://github.com/tobymao/18xx/blob/90baf49a2a3619aa62dbff1077eb2ac3114312f9/api.rb#L139-L146) and [here](https://github.com/tobymao/18xx/blob/90baf49a2a3619aa62dbff1077eb2ac3114312f9/assets/app/app.rb#L29-L30)

michaeljb: https://github.com/yzemaze/18xx/pull/3/files

PR #4625, opened 2021-03-20
Add "selected_company" to "all_potential_upgrades"

carlorusso:


tobymao:
why are some of these selected_company: and some selected_company = ? make it consistent

carlorusso:
> why are some of these selected_company: and some selected_company = ? make it consistent

done

carlorusso:
@tobymao is ok to merge this?

.rubocop.yml
```diff
@@ -78,6 +78,10 @@ Lint/UnexpectedBlockArity:
 Lint/UnmodifiedReduceAccumulator:
   Enabled: true
 
+Lint/UnusedMethodArgument:
```
tobymao, line 81: why did you change this?

carlorusso: to simplify this:
      def all_potential_upgrades(tile, tile_manifest: false) # rubocop:disable Lint/UnusedMethodArgument
to this:
       def all_potential_upgrades(tile, tile_manifest: false, selected_company: nil)

we have 4 not used reference inside the code like this, and the keyword parameter cannot be rewritten with a underscore prefix so rubicon complains about that.
If you think is better to avoid to have keywords parameters not used I can change all of them to normal parameters with default

lib/engine/game/g_1873/game.rb
```diff
@@ -1388,7 +1388,6 @@ def sellable_bundles(player, corporation)
           bundles.select { |bundle| @round.active_step.can_sell?(player, bundle) }
         end
 
-        # rubocop:disable Lint/UnusedMethodArgument
```
roseundy, line 1391: Why was this removed?

tobymao: he removed the lint check

PR #4627, opened 2021-03-20
1830: No multiple brown buys from IPO by default (fixes #4456)

### LINKED ISSUE #4456: 1830 - Brown
Players are able to buy brown shares from market and IPO.

This should be at a bare minimum an optional rule to turn on/off at the creation of game,.

dfannius:
Added an option to use the other interpretation of the original Avalon Hill wording.

One of the fixtures actually depended on the old behavior so I enabled that option for it.

### DIFF
--- a/lib/engine/game/g_1822/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1822/step/buy_sell_par_shares.rb
@@ -60,7 +60,7 @@ module Engine
             cash = available_cash || available_cash(entity)
             cash >= bundle.price &&
               !@round.players_sold[entity][corporation] &&
-              (can_buy_multiple?(entity, corporation) || !bought?) &&
+              (can_buy_multiple?(entity, corporation, bundle.owner) || !bought?) &&
               can_gain?(entity, bundle, exchange: exchange)
           end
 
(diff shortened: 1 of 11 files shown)


dfannius:
Hmm, I wonder why rake works fine locally but fails with `uninitialized constant Engine::Game::G1830::Step` here. A Ruby vs Opal thing?

roseundy:
You probably need to specifically require your steps. ex:
 
require_relative 'step/track'

daniel-sousa-me:
They should be running the same thing, since it's a docker. I'll check.

Thank you very much for taking care of this, which has been causing some grief.

daniel-sousa-me:
> Due to non-deterministic order of reading your files

So fun

dfannius:
After a few minutes of research, it's still not obvious to me which `require_relative` line I have to add and where. My`buy_sell_par_shares.rb` does `require_relative '../../../step/buy_sell_par_shares'`. `g_1830/game.rb` does fail on the line where it refers to `G1830::Step::BuySellParShares`, but I don't see the equivalent of the `require_relative` line I might expect in other titles' `game.rb` files, and I don't want to just add random lines in a cargo-cult way until something works. If it's more clear to anyone what the exact fix is, I'd appreciate it.

daniel-sousa-me:
It turns out the problem is much simpler than expected. You forgot to include the file `lib/engine/game/g_1830/step/buy_sell_par_shares.rb`

dfannius:
Ha ha, indeed I did. Thanks.

tobymao (approved):

Issue #1374, opened 2020-08-06
Make it possible to add/show a longer description to a corporation ability


perwestling:
For companies owned by a corporation you can click on it to get a more extensive explanation what it does.

But for abilities in corporation there is no such extensive description available. In some cases a more full text can be helpful so you do not have to lookup the rules. 
Of course some abilities is more or less self explanatory so abilities without extra description should just ignore clicks.

perwestling:
Example of abilities there this can be helpful:
- 18AL, the historical objective. The text could describe that the corporation receives $100 when its historical objective is tokened.
- 18TN, civil war. Hard to give a short description what a Civil War means in this case.

yzemaze:
@perwestling 
I’m not really keen on adding the company solution here. Simplest solution would be to just allow linewrap for abilities. Would this work?
![image](https://user-images.githubusercontent.com/33390595/108185407-8f245000-710c-11eb-9832-0283a15eaf8d.png)
(dummy text)

perwestling:
@yzemaze What is the problem to use a similar behavior for company, corporation, minor? Adding an always present longer text wouldn't it clutter down the GUI?

It could also be noted that I have started to use abilities for corporation (and minors) more and more.... I do believe this is good UX as the players have this information handy without having to look through the rules, or lookup the wiki page.

yzemaze:
It was just a question. If 2 or 3 lines aren’t sufficient (or if there are multiple abilites with longer texts), that option is dead.
Just point me to examples of these longer texts and I’ll think about it.

PR #4632, opened 2021-03-20
[1846] [18LA] clarify when token reservations go away

### LINKED ISSUE #2021: Token reservations removed at Phase IV - 18 Los Angeles and 1846
Token reservations go away at the beginning of phase IV, though the discount on placement remains. This should be reflected on the Info page, in the Events. The word "reserved" should be removed from company charters at the beginning of phase IV.

benjaminxscott:
Clarifies when corp token reservations go away, adds log msg

addresses #2021 somewhat

image

edit: initial PR also changed corp ability text, but I pulled that out

### DIFF
--- a/lib/engine/game/g_1846/game.rb
+++ b/lib/engine/game/g_1846/game.rb
@@ -225,7 +225,11 @@ module Engine
                         price: 900,
                       },
                     ],
-                    events: [{ 'type' => 'remove_tokens' }],
+                    events: [
+                      { 'type' => 'remove_tokens' },
+                      { 'type' => 'remove_reservations' },
+],
+
                   }].freeze
 
         COMPANIES = [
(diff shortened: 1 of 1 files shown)


yzemaze:
I don’t think this is the way to go. Per had suggested in #1374 to add a way to toggle longer descriptions (cf. company descriptions). I was reluctant, but adding this many lines (and bottom-padding) to corp cards doesn’t work well enough either.

benjaminxscott:
@yzemaze I like that idea to have the ability name clickable to see a longer description, especially since there's prior art with private descriptions

i.e. in the 46 case, the short ability description could be "Reserved token", with longer description with all the details

It'll be useful in other games where they have longer/more involved descriptions for things like destinating or foreign investors in 1880

yzemaze:
> It'll be useful in other games where they have longer/more involved descriptions for things like destinating or foreign investors in 1880

I’ve assigned myself to the issue for a reason, but my knowledge of most games is severly lacking. So I need help. Give me examples, ideally with links incl. actions to have test cases, and I’ll take care of it.

benjaminxscott:
I can pull out the rendering changes from this PR, which will then just add the event text to focus on #2021 

The other ability description / rendering stuff can be a different patch for #1374

I'd suggest to integrate keep some of the ability description changes from this diff, but no strong opinion on that

PR #4678, opened 2021-03-21
18Ireland: Initial implementation of the merger voting

jenf:
This is a first pass at implementing the mergers for 18Ireland
which using a voting system.

@todo's label what needs to be improved later.


perwestling (approved):
I think it looks fine. Some @todos so is this done?

Looking at the changes I think 1893 could be refactored a bit. Had it heavy in the round class instead of merger step.

jenf:
it's not really done done, but it's such a big amount of code i wanted a checkpoint.

PR #4681, opened 2021-03-21
Update auto action button

thepackrat:
This fixes a confusion I was having wtih the UI that somehow suggested I hit "Save" after "Enable"


assets/app/view/game/auto.rb
```diff
@@ -138,7 +138,11 @@ def render_merger_pass(settings)
             h(:ul, { style: { 'list-style': 'none' } }, subchildren),
           ])
 
-          subchildren = [render_button(settings ? 'Save' : 'Enable') { enable_merger_pass(form, passable, rounds) }]
+          subchildren = [
+            render_button(
+              settings ? 'Update existing' : 'Enable'
```
yzemaze, line 143: urgh, why with "existing"? That makes for a long button and doesn’t say more than "update" anyway or can you update non-existing things? ;-)

thepackrat: This was because Update, like Save, suggests that it's the thing you need to hit after "Enable" to make it stick. We went around a few times on this and couldn't come up with something better.

tobymao: Update or Save is fine

thepackrat: Updated to update.

Issue #4054, opened 2021-02-20
18CZ need previous train information for acquisitions


ChrisTheCat:
When acquiring a small or medium corporation, the acquiring corporation may convert a train to a larger type that is represented on the same train card (a physical card in the game with three types of trains printed on it). For example, a 3d could be converted to a 3+3d or a 3Ed.

At this phase of a game (https://18xx.games/game/27956?action=470#info) the info tab does not have this information, because the previous train roster has been removed.

It would be nice if there was an easy way to find out, prior to going through the motions of an acquisition, what the potential train conversions are. The workaround for now is to go back to the start of the history and look at the train roster there, but that's non-intuitive for most users. (You could also look in the rules, but that's even more inconvenient.)

Zwergenpunk:
Any idea what would be easy doable? We could show in the buy Corporation step all trains of corps that can be bought and show upgrad paths and costs

jenf:
Fixed by #4739

PR #4739, opened 2021-03-23
Show previous trains to make it easier to value corps in 1817

### LINKED ISSUE #3563: 1817: Show train information for previous trains
In 1817 you may be buying corps with left over trains, it's useful to know the original price of the train in order to value it correctly.

jenf:
fixes #3563

1822 - Hide trains that are on privates (Pullmans etc)
1882 - Hide the SC trains until they are used
1861/67 - Forget the trains existed for 2 player games

### DIFF
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -30,7 +30,7 @@ module View
       end
 
       def render_body
-        children = upcoming_trains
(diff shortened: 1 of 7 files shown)


benjaminxscott:
Would this also address 18CZ in https://github.com/tobymao/18xx/issues/4054 ?

PR #4749, opened 2021-03-24
Allow other steps to have actions during the stock round.

crericha:
Next step in being able to break-out some of the complexity of buy_sell_par_shares step. Allows for minor bidding to take place in the separate step.

Ran against today's db and no failures.

### DIFF
--- a/assets/app/view/game/bid.rb
+++ b/assets/app/view/game/bid.rb
@@ -11,7 +11,7 @@ module View
         needs :corporation
 
         def render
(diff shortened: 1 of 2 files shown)


tobymao:
there's no point running this change against the db, it doesn't touch the db at all

crericha:
Hmm...that's true. In that case I will test quickly.

crericha:
Upon cursory check: normal stock actions look fine and so do auctions 1867 and 1817.

PR #4765, opened 2021-03-24
rubocop 1.11.0, new cops, auto-corrections

yzemaze:
Because more new cops are useful than not, I enabled NewCops. This triggered a config pruning. Now it’s more in line with default settings and should hopefully save some time on future updates.
There seems to be no way to automatically clear entries that don’t change default settings. So after a while another pruning might be in order to remove disabled cops that are pending now and might become disabled by default.
There are still 4 cops marked as todo in there for someone with deeper knowledge of Ruby & app.


.rubocop.yml
```diff
@@ -26,10 +24,6 @@ Layout/LineLength:
 Layout/MultilineMethodCallIndentation:
   Enabled: false
 
-Layout/MultilineHashBraceLayout:
-  Enabled: true
-  EnforcedStyle: new_line
```
michaeljb, line 31: https://docs.rubocop.org/rubocop/cops_layout.html#layoutmultilinehashbracelayout shows that `symmetrical` rather than `new-line` is the default style, so can we keep this config around?

michaeljb: Actually `symmetrical` seems fine, if that's what you intended. If the closing _and_ opening braces could require a new line I would prefer that, but `new_line` apparently allows this, and I'd definitely prefer `symmetrical` over this:

```
{ a: 1,
  b: 2
}
```

assets/app/view/tiles_page.rb
```diff
@@ -39,7 +39,7 @@ def render
         else
           # apparently separating rotations in URL with '+' works by passing ' '
           # to split here
```
michaeljb, line 41: pretty sure this comment was originally mine, doesn't really make sense with `split` taking no arg now
```suggestion
```

lib/engine/game/g_1849/game.rb
```diff
@@ -460,10 +462,9 @@ def track_type(paths)
           types = paths.map(&:track).uniq
           raise GameError, 'Can only change track type at station.' if types.include?(:broad) && types.include?(:narrow)
 
-          case
-          when types.include?(:narrow)
+          if types.include?(:narrow)
             :narrow
-          when types.include?(:broad)
```
michaeljb, line 466: I didn't even know you could write a `case` statement like that 🤪

michaeljb (approved):

.rubocop.yml
yzemaze: Combination of Layout/FirstHashElementLineBreak (enabled) && Layout/MultilineHashBraceLayout (default: symmetrical) && others forces
```
{
  a: 1,
  b: 2,
}
```
… which was intended mostly to pull props into shape ;-) The config now shouldn’t change behavior though, at least there were just 2 auto-corrections.

PR #4802, opened 2021-03-26
validate username & email

### LINKED ISSUE #2357: Better error message on duplicate username
If I try to change my username to one that exists, I get the error "PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "users_lower(name)_index" DETAIL: Key (lower(name))=(...) already exists."

yzemaze:
fixes #2357

image
image

### DIFF
--- a/models/user.rb
+++ b/models/user.rb
@@ -21,12 +21,7 @@ class User < Base
   ]).freeze
 
   def update_settings(params)
-    if params['name']
-      self.name = params['name']
-      raise 'Name cannot be empty' if name.empty?
-
-    end
-
+    self.name = params['name']
     params.each do |key, value|
       settings[key] = value if SETTINGS.include?(key)
     end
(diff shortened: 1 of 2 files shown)


models/user.rb
```diff
@@ -68,4 +63,11 @@ def to_h(for_user: false)
   def inspect
     "#{self.class.name} - id: #{id} name: #{name}"
   end
+
+  def validate
+    super
+    validates_unique(:name, :email, { message: 'is already registered' })
+    validates_format /^[^\s]+$/, :name, message: 'may not be empty'
+    validates_format /^[^@\s]+@[^@\s]+\.[^@\s]+$/, :email
```
benjaminxscott, line 71: You may be able to use this, which is apparently a standard import

URI::MailTo::EMAIL_REGEXP

(since hand-rolled email regexes are tricky)

yzemaze: I remember talking with Toby about the regex last year. tl;dr the simple one was good enough. But this one didn’t cross my path. Thanks.

tobymao (approved):

models/user.rb
tobymao: up to you, don't care

PR #4804, opened 2021-03-26
1862: map update

roseundy:
Improved map and also fixed issue with first edition-based map.

Needed minor changes to UI::Part::Partition and Engine::Tile

new_1862_map

### DIFF
--- a/assets/app/view/game/part/partitions.rb
+++ b/assets/app/view/game/part/partitions.rb
@@ -40,6 +40,8 @@ module View
               :blue
             when :impassable
               :red
+            when :divider
+              :black
             end
 
           setting_for(color)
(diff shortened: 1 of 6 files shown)


jenf (approved):

assets/app/view/game/part/partitions.rb
```diff
@@ -53,7 +55,7 @@ def render_part
           children = []
 
           @tile.partitions.each do |partition|
-            next if partition.blockers.none? do |blocker|
+            next if partition.type != :divider && partition.blockers.none? do |blocker|
```
tobymao, line 58: blockers.empty?

tobymao (approved):

assets/app/view/game/part/partitions.rb
roseundy: I think none? is needed.

PR #4811, opened 2021-03-26
18ireland stock round

jenf:
Since 18Ireland stock round is 90% the same as 1867, separate
the reusable code and reuse.

Also add
Tile laying definitions
Company & Corporation setup
Add events
Fix timing of the merger round
Add end game triggers

### DIFF
--- a/lib/engine/game/g_18_ireland/game.rb
+++ b/lib/engine/game/g_18_ireland/game.rb
@@ -16,7 +16,13 @@ module Engine
 
         CAPITALIZATION = :incremental
         HOME_TOKEN_TIMING = :float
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_ireland/game.rb
```diff
@@ -165,6 +243,28 @@ def next_round!
               new_stock_round
             end
         end
+
+        def event_corporations_can_merge!
+          # All the corporations become available, as minors can now merge/convert to corporations
+          @corporations += @future_corporations
```
tobymao, line 249: concat

tobymao (approved):

PR #4829, opened 2021-03-27
allow zigzag markets to be flipped; match 18CZ's market to physical board

michaeljb:
1860 and 1862 markets are unaffected by this change.

Before:

Screenshot 2021-03-27 111740

After:

Screenshot 2021-03-27 111532

Physical Board:

File_000


roseundy (approved):
Looks reasonable to me.

PR #4859, opened 2021-03-29
1870: St. Louis was losing the P label when the private was used

daniel-sousa-me:


daniel-sousa-me:
When this is merged, I'll add a fixture with a real game (which now fails)

daniel-sousa-me:
This will archive some games

jenf:
Is this in reference to a bug? Is it upgradable tiles that get the label added? If so, that will cause problems as the tile will retain the label after being upgraded

daniel-sousa-me:
The label is being added and removed on step/track, but I forgot to add here. There's no need to remove on the special track, because there's no ability allowing upgrades

jenf (approved):

PR #4862, opened 2021-03-29
Adding Conn. column to spreadsheet for 1870

jeroenh:
The Conn. column shows the connection run that corporations can make
when they are able to reach their destination.


assets/app/view/game/spreadsheet.rb
```diff
@@ -156,6 +156,32 @@ def render_or_history_row(hist, corporation, x)
...
+        return [h(:td)] unless @game.connection_run[corporation]
+
+        x, c_run = @game.connection_run[corporation]
+        revenue_text, alpha =
+          case (c_run.dividend.is_a?(Engine::Action::Dividend) ? c_run.dividend.kind : 'withhold')
+          when 'withhold'
+            ["[#{c_run.revenue}]", '0.5']
+          else
+            [c_run.revenue.to_s, '1.0']
+          end
```
yzemaze, line 168: Looks complicated. Wouldn’t `if c_run.dividend.kind == 'withhold'` do the same?

assets/app/view/game/spreadsheet.rb
```diff
@@ -156,6 +156,32 @@ def render_or_history_row(hist, corporation, x)
...
+
+        props = {
+          style: {
+            color: convert_hex_to_rgba(color_for(:font2), alpha),
+          },
+        }
+
+        link_h = history_link(revenue_text,
+                              "Go to connection run of #{corporation.name} (in #{x})",
```
yzemaze, line 177: `x.join('.')` would fit the style used everywhere else.
(bonus points if you fix L147, too ;-) )
Don’t use "x" if there are better var names available.

assets/app/view/game/spreadsheet.rb
```diff
@@ -181,6 +207,9 @@ def render_titles
         end
         @extra_size = extra.size
 
+        extra_connection_title = @game.respond_to?(:connection_run) ? [h(:th, th_props[1], 'Conn.')] : []
+        extra_connection_sub = @game.respond_to?(:connection_run) ? [h(:th, '')] : []
```
yzemaze, line 209: Why the extra-prefix if it’s called "connection_run" everywhere else?

assets/app/view/game/spreadsheet.rb
jeroenh: It probably would. Most of this function was inspired by the render_or_history_row function. but 1870 doesn't have a half-pay, so I just removed that.
The connection runs also don't have many of the corner cases that or_history can run into, so it could probably be simplified.

assets/app/view/game/spreadsheet.rb
jeroenh: Again, inspired on render_or_history_row doing the same thing. Will try to use the other style.

assets/app/view/game/spreadsheet.rb
jeroenh: It's meant to convey that this is an optional column, but I could not use the "extras" variable as I wanted to place it just before the OR history. So I settled on using the extra-prefix to continue that style.

yzemaze: Sure, if you think this is necessary => extra_connection_run… (it’s a run, not a connection!)

assets/app/view/game/spreadsheet.rb
```diff
@@ -118,22 +118,22 @@ def render_player_or_history
...
         end
       end
 
-      def render_or_history_row(hist, corporation, x)
-        if hist[x]
+      def render_or_history_row(hist, corporation, turn_round)
+        if hist[turn_round]
           revenue_text, alpha =
-            case (hist[x].dividend.is_a?(Engine::Action::Dividend) ? hist[x].dividend.kind : 'withhold')
+            case (hist[turn_round].dividend.is_a?(Engine::Action::Dividend) ? hist[turn_round].dividend.kind : 'withhold')
```
jeroenh, line 129: This line is too long now, and an alternative is needed. what would that look like?

dfannius: How about assigning `div = hist[turn_round].dividend` on the previous line and then using that? It would make the code look nicer in any case.

yzemaze: That would fail. But there are other ways.
https://18xxgames.slack.com/archives/C012K0CNY5C/p1617108056426900?thread_ts=1617103417.422800&cid=C012K0CNY5C

Issue #4857, opened 2021-03-29
[1828.games] Merger procedure does not correctly prompt the player about which share is the "odd" share.


mlatuprosa:
**Issue**: when a player has an odd number of shares in two companies that merge, and some shares in both, depending on the order of the merger, they may not be prompted about which share to keep, and when they are prompted the prompt does not do anything.

Link to game JSON: https://gist.github.com/mlatuprosa/158a64f10d50ade791cb95bef80cbd65

**How to Reproduce the Bug?**

- Obtain two presidencies of floated companies with an odd total number of shares. In the example game, player 2 has 60% of PRR and 30% of NW and they haven't merged yet.
- Merge the two companies in either order.
- If you merge the company with fewer shares held into the company with more shares held (NW into PRR in the example), you will be prompted about which share to keep, but regardless of which one you pick you will be prompted to sell or exchange as if you called the share in the 30% company (NW in the example) the "odd" share. 
- If you merge the company with more shares held into the company with fewer shares held (PRR into NW in the example), you are not prompted about which share to keep at all, and it considers the share in the 60% company to be the "odd" share. 

A live game with a similar issue was https://18xx.games/game/32302?action=793 where the same behavior can be seen upon merging B&M (60% held, freshly floated) into C&O (50% held, not floated) vs. merging C&O into B&M.

**Which Browser / Platform?**
- Firefox 85.02

crericha:
The player is only prompted if it is possible for them to have a choice. The 2:1 exchange sources are 1) the player's hand, 2) discard, 3) other players, 4) IPO, and 5) bank pool. This means that if the number of 2:1 exchanges the player will do is greater or equal to the number of shares they have, in the corp that initiated the merge, then they can not opt to keep a share of that corp. 

For example, the player has 3 NW and 6 PRR shares. If the PRR merges into the NW, then the 3 NW will be needed for the exchange, so a PRR share must be the share kept. On the other hand, if NW merges into PRR, a minimum of 4 PRR shares are needed for the exchange, so the player has a choice of which share to keep.

> but regardless of which one you pick you will be prompted to sell or exchange as if you called the share in the 30% company (NW in the example) the "odd" share.
This is a bug.

mlatuprosa:
Why do shares of the company that initiated the merge have to be involved in every 2:1 conversion? This does not seem to be implied by the phrasing of 7.5.2.4 (a) to me.

crericha:
[...]

PR #4872, opened 2021-03-30
[1828.Games] Fix odd share player choice

crericha:
Substring check was backwards.


jenf:
Which issue do this fix?

crericha:
Its related to https://github.com/tobymao/18xx/issues/4857, but I didn't link because there may be a second, unrelated bug in that issue and I don't want to close it.

jenf (approved):

PR #4873, opened 2021-03-30
1873: Diesel fixes; More colors

### LINKED ISSUE #4865: [Harzbahn 1873] Need more colors for routes
Example game: https://18xx.games/game/33770?action=1689

Trying to create a 9th route results in this error: "undefined method `match' for nil"

I think we need at least 16 colors.

### LINKED ISSUE #4867: [Harzbahn 1873] Diesel route affecting revenue from normal trains
Game https://18xx.games/game/33770?action=1725

When the Diesel routes are added, the revenue produced by the "normal" trains drop by the amount the diesels are running for.

roseundy:
Touches assets/app/lib/settings.rb

Fixes #4865 
Fixes #4867 

Before:
colors_before

After:
colors_after

### DIFF
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -6,8 +6,25 @@ require_relative 'hex'
 module Lib
   module Settings
     DARK = `window.matchMedia('(prefers-color-scheme: dark)').matches`.freeze
-    # http://mkweb.bcgsc.ca/colorblind/ 12 color palette
-    ROUTE_COLORS = %i[#A40122 #008DF9 #00FCCF #FF5AAF #8400CD #FF6E3A #FFB2FD #FFC33B].freeze
+    # http://mkweb.bcgsc.ca/colorblind/ 15 color palette, with some substitutions + 1 additional
+    ROUTE_COLORS = %i[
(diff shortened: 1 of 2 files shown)


yzemaze:
before is this:
image

(color 7 was changed recently #4561)

assets/app/lib/settings.rb
```diff
@@ -6,8 +6,25 @@
...
+      #FFDC3D
+      #EF0096
+      #7CFFFA
+      #005FCC
+      #F60239
+      #00E307
+      #FFCFE2
+      #AFFF2A
+      #E8D2AB
+    ].freeze
```
yzemaze, line 27: hyperbole ;-) 2 lines would suffice.

roseundy: It was a lot easier to reorder this way.

roseundy:
We already had enough pinks, I think.

assets/app/lib/settings.rb
roseundy: OK, fixed.

roseundy:
This shouldn't require pins.

PR #4878, opened 2021-03-30
1870: Close SLSF's private company after parring

daniel-sousa-me:
In the actual game this private does not actually exist, but here it makes things easier. But it was counting toward the value, which it shouldn't.


jenf (approved):

jenf (changes requested):
Want to check this further.

jenf (approved):

jenf:
Shouldn't impact game play

PR #4898, opened 2021-04-01
1870: Only allow brown multiple buy from the market

daniel-sousa-me:
This probably should be generalised, but I don't know of other games that have this rule (other than 1830) and the 1830 optional rule makes it a nontrivial generalisation.


jenf:
Check failure doesn't seem to be you, will look at lunch.

jenf:
I'd suggest it should be generalized.
Have a class constant in the base class call
MULTIPLE_BUY_ONLY_FROM_MARKET with set to false
Add method called multiple_buy_only_from_market? which allows it to be overriden by game options

daniel-sousa-me:
Superseeded by #4977

PR #4906, opened 2021-04-01
[18CZ] Added company size to ability button

Zwergenpunk:
Now looks like that

image

@yzemaze is there something I'm missing. In the function I changed is a "truncate" of the name and I'm not sure if I break something


yzemaze:
It’s more or less just styling and no problem if some buttons might be a tad wider.

PR #4919, opened 2021-04-02
1873: fix machine scrapping for indie mines

### LINKED ISSUE #4918: [Harzbahn 1873] Indie Mines can have multiple machines
Due to a bug in the scrap_mine_train method, when an independent mine buys a machine from the bank, it won't scrap an existing machine. Fundamentally, this is due to the fact that the machine "variant" hasn't been assigned yet to the new train being bought at the point the scrapping is done.

The code is written with the assumption that there is only one machine in the trains list for mines and this bug violates that. I'm moderately surprised that this bug fix didn't kill the three games I've looked at.

roseundy:
Fixes #4918

### DIFF
--- a/lib/engine/game/g_1873/step/buy_train.rb
+++ b/lib/engine/game/g_1873/step/buy_train.rb
@@ -252,10 +252,10 @@ module Engine
           end
 
           def scrap_mine_train(entity, new_train)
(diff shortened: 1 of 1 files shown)


roseundy:
This may require pins, although spot-checking three games didn't show the need.

PR #4922, opened 2021-04-02
Add announcement for Season 5 of the 1846 League to the page

daniel-sousa-me:
@tobymao is it ok if we announce the 1846 league once again on the site for a couple of days?

I've seen more people than usual asking about it and wouldn't like if they missed the small-ish registration window.


tobymao:
yes

Issue #4710, opened 2021-03-23
18MEX: Implement Variant Package #1 - The Baja-Copper Connection


perwestling:
An official variant of 18MEX.

If you need details contact @perwestling

michaeljb:
@philcampeau was talking about working on this on slack (looks like I can’t assign him on the ticket though)

philcampeau:
Tbh I’m still figuring out the ins and outs of how the code works. If someone else wants to take this on, it’s pretty straight forward.

perwestling:
@philcampeau I created this intending to assign it to you. Take your time...

philcampeau:
Will do!

PR #4932, opened 2021-04-03
[18LA, 18MEX] token abilities: fix 'from_owner', rename 'extra' -> 'extra_action'

### LINKED ISSUE #4874: [18LA] Dewey, Cheatham, and Howe is adding an extra token instead of using one from charter, and not charging the company
Turns out `extra` on the token ability is behaving in an overloaded way--sometimes it's meaning an extra action (`Action::SpecialToken`), other times meaning a token in addition to the corporation's tokens. This makes it incompatible with `from_owner: true`.

This also impacts the 18MEX Baja Variant (in development, #4710), which should also use `Action::SpecialToken` and take a token from the corporation's charter.

michaeljb:
* `extra` was overloaded, sometimes being used to mean an extra action (in
  addition to a corporation's normal token action), and sometimes meaning the
  token itself was in addition to the tokens found on the corporation's charter
* add more documentation for the token ability properties
* add fixture for both LA Title and Dewey, Cheatham, and Howe; replace
  previous LAT test; add DC&H test for new logic (which fails against master)

[Fixes #4874]

----

Will need to pin some 18 Los Angeles games where Dewey, Cheatham, and Howe was processed for free and not using a token from the charter, so the owning corp incorrectly had more cash and tokens to play with.

This should unblock #4710 (18MEX Baja Variant)


lib/engine/game/g_1882/game.rb
```diff
@@ -291,6 +291,7 @@ class Game < Game::Base
                 teleport_price: 0,
                 when: 'owning_corp_or_turn',
                 count: 1,
+                from_owner: true,
```
michaeljb, line 294: config change here

lib/engine/step/tokener.rb
```diff
@@ -125,20 +126,17 @@ def adjust_token_price_ability!(entity, token, hex, city, special_ability: nil)
...
-            end
+          if ability.neutral
+            neutral_corp = Corporation.new(
+              sym: 'N',
+              name: 'Neutral',
+              logo: 'open_city',
+              tokens: [0],
+            )
+            token = Engine::Token.new(neutral_corp, type: :neutral)
+          elsif ability.owner.company? && !ability.from_owner
```
michaeljb, line 137: This is the only real logic change in this PR.

1882 had a config change for one private - https://github.com/tobymao/18xx/pull/4932/files#diff-f3da9b37ea0b71dfa9bf440659066596d6d31095abec184956ec760b361473e3R294

A test for the DC&H company was added, and the LAT test was changed so that they could use the same fixture - https://github.com/tobymao/18xx/pull/4932/files#diff-04a1776aafdbd418b15bf5daac0af05a9f64cee365f0cfd8a4a00d8c751a8f26

All other Ruby changes in this PR are about renaming `extra` to `extra_action`

PR #4945, opened 2021-04-03
1870: Fix rounding of reissue par price

daniel-sousa-me:
I thought it always rounded down, but it turns out that it rounds to the closest value


daniel-sousa-me:
I forgot to warn earlier. This may archive some games

PR #4948, opened 2021-04-04
tiles tab/page: toggle tile name

### LINKED ISSUE #4868: Displaying tile counts in a more visually intuitive way
When playing online I find it hard to get a sense of how many of each tile remain. When playing in real life this isn't a problem, as you can easily get a sense of how many remain from the depth of the pile. 

I wonder if the tiles could be represented visually in a 'pile' using a 3D overlapping effect, as in this [mock-up](https://docs.google.com/drawings/d/1mwCdXbFAw8B1VBBBddqs2mXya3rrRRrcVW2_HIbTiIk/edit?usp=sharing), which displays a pile of 4 and 2 tiles respectively. 

In addition, in the interests of reducing information overload, perhaps there could be a checkbox allowing users to display only the tile count, and not the roster number,so "\#3 x 2" becomes simply "2".

yzemaze:
- tile name toggleable on tiles tab and /tiles (partly closes #4868)
- tile name + count rendering revised
- fixed rendering of double sided tiles (18Mag) on /tiles
- refactoring of double sided tiles rendering

ante
image

post
image.image

### DIFF
--- a/assets/app/view/game/tile_manifest.rb
+++ b/assets/app/view/game/tile_manifest.rb
@@ -1,12 +1,19 @@
 # frozen_string_literal: true
 
+require 'lib/settings'
 require 'view/tiles'
 
 module View
   module Game
     class TileManifest < Tiles
+      include Lib::Settings
+
       needs :game
       needs :tile_selector, default: nil, store: true
+
+      def render
+        h(:div, [render_tile_manifest, render_toggle_button])
+      end
 
       def render_tile_selector(remaining, tile, shift: 0)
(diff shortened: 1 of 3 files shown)


tobymao:
i feel like the toggle is not necessary

tobymao (approved):

yzemaze:
> i feel like the toggle is not necessary

New text layout works better than old one, but without tile names the table is way cleaner and tile counts should be a lot easier to grasp:
image
image

I’d even get rid of cell borders …

PR #4962, opened 2021-04-05
1873: 17+ routes, tokening checks reserved tile, player value

### LINKED ISSUE #4936: [Harzbahn 1873] Laying track and building stations on other companies' concession routes...
**What's your Game ID?**

https://18xx.games/game/33488

**How to Reproduce the Bug?**

- During Operating Round 8.1, the SHE built a station in E6 (undone afterwards). That was possible and was correct, although the station was built on an NWE concession hex with the NWE not yet operating, since there was another free station circle on the reserved tile 970.
- This was action https://18xx.games/game/33488?action=877
- But previously, during the same Operating Round 8.1, the WBE, after building the tile 962 on D7 (action https://18xx.games/game/33488?action=870), tried to build also a station there. That was not allowed by the system although it should have been allowed since on the reserved tile 975 there is indeed a free station circle so that such a station building by the WBE did "not prevent the [concession] company from building a station in the same hex of its concession route during its first OR" (see Rule 4.4.2).

By the way: could it be checked please if the following is implemented correctly: 
- If  company A builds on the concession route of company B a tile of exactly the reserved tile type: does it actually use the reserved tile (this should be the case), or does it use another one of the same type, thus "freeing" the reserved tile.
- This is important since if this is not correctly implemented, this would mean that only company B could build this tile if all other tiles of this type were currently placed on the map.

Klaus Kiermeier

### LINKED ISSUE #4952: Harzbahn 1873 Seems to "hang" on routing 8th diesel in game 34033
Game: https://18xx.games/game/34033

OR 9.3, operation of HBE, first company to operate in round.

The HBE acquired a new 5T in previous OR and scrapped a 3T.  After assigning routes to the new 5T, the game seems to hang and becomes unresponsive when trying to assign a route to another diesel (a diesel that did not have a route in the previous OR.

Screen shot just prior to route assignment to diesel:
image

Attempting to assign 8th diesel route F11 - G12 - H13 screen shot (clicked on F11 and then G12):
image

The game responds to the "clear train" button, but it won't assign the route.

### LINKED ISSUE #4955: [Harzbahn 1873] Face value of independent mines is not counted towards the "Value" of the players
As said above, the "Values" of the players do not include the face values of their independent mines.
This does not affect game play as such, but if at the end of a game there were still independent mines owned by players, their face values would contribute to the fortunes of their owners, according to the rules.

Klaus Kiermeier

roseundy:
Fixes #4936 
Fixes #4952 
Fixes #4955 

Common file change: UI:Lib:Settings - unlimited routes by cycling through colors

### DIFF
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -62,7 +62,7 @@ module Lib
     end
 
     def route_prop_string(index, prop)
-      "r#{index}_#{prop}"
+      "r#{index % ROUTE_COLORS.size}_#{prop}"
     end
 
     def change_favicon(active)
--- a/lib/engine/game/g_1873/game.rb
+++ b/lib/engine/game/g_1873/game.rb
@@ -996,12 +996,21 @@ module Engine
           entity&.corporation? && @corporation_info[entity][:type] == :railway
         end
 
+        # determine if a token lay is blocked by the need to keep a slot open
+        # for a pending concession
         def concession_blocks?(city)
           hex = city.hex
           return false unless (exits = CONCESSION_ROUTE_EXITS[hex.id])
           return false unless concession_incomplete?(@concession_route_corporations[hex.id])
           # take care of OO tile. Only care about city along concession route
           return false if (city.exits & exits).empty?
+
+          # if there is a reserved tile that has a city with the same connections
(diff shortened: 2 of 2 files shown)


yzemaze (approved):
change to lib/settings is fine

PR #4983, opened 2021-04-06
18Ireland: Implement hex trains, dividends and free track lays for na…

jenf:
…rrow gauge

### DIFF
--- a/lib/engine/game/g_18_ireland/step/track.rb
+++ b/lib/engine/game/g_18_ireland/step/track.rb
@@ -15,6 +15,45 @@ module Engine
             super
             @game.clear_narrow_graph
           end
+
+          def check_track_restrictions!(entity, old_tile, new_tile)
+            return if @game.loading || !entity.operator?
+
+            connected_paths = if @game.tile_uses_broad_rules?(old_tile, new_tile)
+                                @game.graph_for_entity(entity).connected_paths(entity)
+                              else
+                                # Must update the graph now.
+                                @game.clear_narrow_graph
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_ireland/step/track.rb
```diff
@@ -15,6 +15,45 @@ def process_lay_tile(action)
...
           end
+
+          def check_track_restrictions!(entity, old_tile, new_tile)
+            return if @game.loading || !entity.operator?
+
+            connected_paths = if @game.tile_uses_broad_rules?(old_tile, new_tile)
+                                @game.graph_for_entity(entity).connected_paths(entity)
+                              else
+                                # Must update the graph now.
+                                @game.clear_narrow_graph
```
roseundy, line 26: Doesn't this need to also happen when the normal graph is cleared, or is this the only time narrow_graph is referenced?

jenf: normal graph gets cleared as part of the lay tile a few lines before this gets called.

roseundy: The only reason I ask is because since this is in check_track_restrictions, the graph won't get cleared when loading.

tobymao (approved):

lib/engine/game/g_18_ireland/step/track.rb
jenf: It gets cleared again anyway at the end of process_lay_tile

PR #4991, opened 2021-04-07
[18ZOO] Add Hole

carlorusso:
Adding "Hole" company custom ability


carlorusso:
> Please could you separate the migration of code to the separate entities.rb and map.rb from actual code changes. it's a bit hard to review in it's present form.

@jenf ok, but consider that I changed also all the maps (adding the hole I moved everything, like all A -> B, ...) do you really think it worth to split anyway?

jenf:
ahh, that wasn't clear from the description, I'll review as is.

lib/engine/game/g_18_zoo/game.rb
```diff
@@ -1483,6 +704,40 @@ def share_price_updated(entity, revenue)
...
+          distance = route.visited_stops.sum(&:visit_cost)
+          return distance if @holes.empty?
+
+          distance + ((@holes & route.all_hexes).size == 2 ? 1 : 0)
+        end
+
+        def assign_hole(entity, target)
+          @holes << target
+
+          return unless entity.all_abilities.empty?
```
jenf, line 717: Apart from this method lgtm.

Shouldn't the entity check it has the right ability?

jenf: What do the 'holes' actually do?

PR #5000, opened 2021-04-08
1862 Track

roseundy:
Mostly implements all of the track step
Also implements home token lays, including mandatory home hex upgrade in cases

Still to do for track:
- Special Ipswitch and Harwich upgrade rules (since there wasn't room for two labels)
- Hiding tiles used to implement adding towns to plain track


lib/engine/game/g_1862/step/token.rb
```diff
@@ -0,0 +1,23 @@
...
+      module Step
+        class Token < Engine::Step::Token
+          def actions(entity)
+            return [] if entity.corporation? && entity.receivership?
+
+            super
+          end
+
+          def description
+            'add token'
```
jenf, line 17: Sentence case

lib/engine/game/g_1862/step/track.rb
```diff
@@ -0,0 +1,122 @@
...
+            @game.tiles << old_base_tile
+          end
+
+          def potential_tiles(entity, hex)
+            colors = @game.phase.tiles
+            if normal_available_hex(entity, hex)
+              @game.tiles
+                .select { |tile| colors.include?(tile.color) }
+                .uniq(&:name)
+                .select { |t| @game.upgrades_to?(hex.tile, t) }
```
tobymao, line 50: doesn't seem like you need to copy and paste this, only need to change the lambda

lib/engine/game/g_1862/step/track.rb
```diff
@@ -0,0 +1,122 @@
...
+
+            # Here on out only used for adding towns
+            # The assupmption is that this is not performance critical
+            old_ctedges = hex.tile.city_town_edges.map(&:sort)
+            old_exits = hex.tile.exits.sort
+
+            new_exits = tile.exits.sort
+            new_ctedges = tile.city_town_edges.map(&:sort)
+
+            extra_ctedges = (new_ctedges - old_ctedges).flatten.sort
```
tobymao, line 72: why do you need to sort, aren't you just doing set operations anyways?

roseundy: It's because of this: 
`extra_ctedges == (old_simple_exits - new_simple_exits)`

[0,1] is not equal to [1,0]

It's also because
`extra_ctedges = (new_ctedges - old_ctedges).flatten.sort`
doesn't work if new_ctedges is [[0, 1] , [2, 3]] and old_ctedges is [[1, 0], [3, 2]] (i.e. the component arrays must be sorted).

I could use Set if that would be more elegant.

lib/engine/game/g_1862/step/track.rb
roseundy: OK.

lib/engine/game/g_1862/step/track.rb
roseundy: I did find two superfluous sorts that I removed.

Also, I can remove the sorts if I replace the equality check with an all?

roseundy:
Removed as many sorts as I could.
I also removed the duplicated code between normal_available_hex and available_hex

jenf (approved):

PR #5001, opened 2021-04-08
1873: prohibit all illegal compulsory train buys

### LINKED ISSUE #4997: [Harzbahn 1873] NWE must not buy 1T as its compulsory train

https://18xx.games/game/35738

https://18xx.games/game/35738?action=469

**How to Reproduce the Bug?**

- In the above action the NWE bought a 1T train from the HBE.
- This should not have been allowed since at this step the NWE was obliged to buy, as its first train purchase, a train that qualifies as its compulsory train, and the 1T doesn't (see the respective "Special feature" in Rule 4.4.7).
- All the remainder of the insolvency and recapitalization procedures are correct.

roseundy:
Fixes #4997

### DIFF
--- a/lib/engine/game/g_1873/step/buy_train.rb
+++ b/lib/engine/game/g_1873/step/buy_train.rb
@@ -115,13 +115,15 @@ module Engine
             (depot_trains + switchers + other_trains).compact
           end
 
+          def illegal_concession_buy?(train, entity)
+            @game.concession_pending?(entity) &&
+              (@game.train_is_switcher?(train) ||
+               @game.diesel?(train) ||
+               (train.distance < 2 && entity == @game.nwe))
+          end
+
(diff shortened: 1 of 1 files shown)


roseundy:
There is a slight chance this will require pins for some games.

lib/engine/game/g_1873/step/buy_train.rb
```diff
@@ -127,6 +133,13 @@ def illegal_depot_buy?(train, entity)
           end
 
           def illegal_other_buy?(train, entity)
+            if @game.concession_pending?(entity) &&
```
jenf, line 136: can we de-duplicate this code?

roseundy: Good point. Yes.

PR #5007, opened 2021-04-09
18Ireland bonus payouts for narrow gauge routes

jenf:
The effect of narrow gauge track is to add the value of town stations that
are exclusively on narrow gauge track to the first broad gauge station
they are connected to at the end (or both ends) of the narrow
gauge line. Thus, the companies in the game benefit from the construction
of narrow gauge track even though they do not operate it
directly. It is possible to string together multiple towns on a single
narrow gauge line or network.


roseundy (approved):
lgtm

PR #5012, opened 2021-04-09
1873: sanity check on share forfeit

### LINKED ISSUE #5010: [Harzbahn 1873] Company should become insolvent, but doesn't

https://18xx.games/game/35738

**How to Reproduce the Bug?**

- During Operating Round 5.2, the GHE has an income of 170M, 0M in its treasury, but 570M maintenance costs are due.
- So it should become insolvent.
- But when clicking on "Submit 170M" nothing happens; only the following message appears:
- "All shares must be from the same corporation."
- This makes no sense at all.

roseundy:
Fixes #5010


roseundy:
No pins required.

PR #5013, opened 2021-04-09
18FL: Show hex tokens on charter

ventusignis:
Changes to Corporation view because hex tokens don't have a city

Fixes 5003

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -281,7 +281,7 @@ module View
             if i.zero? && @corporation.coordinates
               @corporation.coordinates.is_a?(Array) ? @corporation.coordinates.join('/') : @corporation.coordinates
             else
-              token.city ? token.city.hex.name : token.price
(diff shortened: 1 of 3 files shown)


roseundy:
Sounds like hex tokens are somewhat similar to the markers in 1873.

michaeljb:
@roseundy doing a git grep for "marker" isn't finding anything in 1873 files. Is there something here that should be generalized?

roseundy:
Sorry, I call large icons "markers". Large icons are already in the common code. I use them in 1873 for mines, and in 1862 to mark the "home" hex for corps before they have run.

https://github.com/roseundy/18xx/blob/1233d5f60b6973554cc2331de1ec2282c19f0304/lib/engine/game/g_1862/game.rb#L476

ventusignis:
image
Hex tokens are a little different, they're tokens used from the charter to be placed on non-city-slots (in this case towns) and are represented as icons

michaeljb (approved):

PR #5016, opened 2021-04-09
18Ireland privates: Board of Works & Drumglass Colliery Railway

jenf:
Fix Irish mail, P2 and P5 for narrow gaugeness


lib/engine/step/track.rb
```diff
@@ -39,18 +39,7 @@ def process_lay_tile(action)
...
-        tile_lay = get_tile_lay(entity)
-        return nil unless tile_lay
-
-        color = hex.tile.color
-        return nil if color == :white && !tile_lay[:lay]
-        return nil if color != :white && !tile_lay[:upgrade]
-        return nil if color != :white && tile_lay[:cannot_reuse_same_hex] && @round.laid_hexes.include?(hex)
-
-        connected
+        base_available_hex(entity, hex)
```
roseundy, line 42: Why this change?

jenf: tracker.rb which step uses doesn't have this code and 18ireland needs to effectively use it for the Board of Works private.

i'm not entirely sure about the naming, hopefully Toby can have a look

jenf: @tobymao could do with some opinions

lib/engine/game/g_18_ireland/entities.rb
```diff
@@ -42,6 +42,17 @@ module Entities
...
             sym: 'BoW',
+            abilities: [
+              {
+                type: 'tile_lay',
+                hexes: [],
+                tiles: [],
+                when: 'owning_corp_or_turn',
+                owner_type: 'corporation',
+                free: true,
+                count: 2,
```
michaeljb, line 53: The text says "The company closes once this ability has been used twice", so I think you want to add `closed_when_used_up` here?

lib/engine/step/track.rb
michaeljb: Does `reachable: true` on the ability definition solve this? From the description in the abilities README it sounds like it raises an error after placing the tile, so I don't know if it affects which hexes are highlighted (which I think is what you're trying to accomplish here?)

Maybe the `Engine::Step::SpecialTrack` available_hex logic could be modified so that if `reachable` is true, then only reachable hexes are highlighted? (i.e., generalize what you have in `G18Ireland::Step::SpecialTrack#available_hex`)

Instead of `base_available_hex`, maybe `tracker_available_hex`?

lib/engine/game/g_18_ireland/game.rb
```diff
@@ -291,17 +291,21 @@ def tile_uses_broad_rules?(old_tile, tile)
...
-                                  end
-          # Must be connected for the tile to be layable
-          return false unless connection_directions
+          new_tile_paths.all? { |path| path.track == :broad || old_paths.any? { |p| path <= p } }
+        end
+
+        def legal_tile_rotation?(entity, hex, tile)
+          # TIM, DR and TDR can lay irrespective of connectivity.
+          if !entity.company? || !%w[TIM DR TDR].include?(entity.id)
+            corp = entity.owner.corporation? ? entity.owner : entity
```
michaeljb, line 300:
```suggestion
            corp = entity.corporation
```

lib/engine/game/g_18_ireland/step/special_track.rb
```diff
@@ -1,30 +1,56 @@
...
+          end
+
+          def available_hex(entity, hex)
+            return base_available_hex(entity.owner, hex) if entity.id == 'BoW'
+
+            super
+          end
+
           def hex_neighbors(entity, hex)
+            operator = entity.owner.corporation? ? entity.owner : entity
```
michaeljb, line 37:
```suggestion
            operator = entity.corporation
```

lib/engine/game/g_18_ireland/step/special_track.rb
```diff
@@ -1,30 +1,56 @@
...
+          include NarrowTrack
           def potential_tiles(entity, _hex)
             return super unless entity.id == 'TIM'
             return [] unless (tile_ability = abilities(entity))
 
             tile_ability.tiles.map { |name| @game.tiles.find { |t| t.name == name } }
           end
 
+          def get_tile_lay(entity)
+            entity = entity.owner if entity.company?
```
michaeljb, line 20:
```suggestion
            entity = entity.corporation
```

lib/engine/game/g_18_ireland/step/special_track.rb
```diff
@@ -1,30 +1,56 @@
...
-            super
-          end
 
-          def process_lay_tile(action)
-            super
-            @game.clear_narrow_graph
+            return unless (ability = abilities(entity))
+            return if !ability.hexes&.empty? && !ability.hexes&.include?(hex.id)
+
+            if ability.type == :tile_lay && ability.reachable && !(@game.graph.connected_hexes(operator)[hex] ||
```
michaeljb, line 48: I notice the `reachable` check here, but none of the 18Ireland entities definitions use `reachable: true` (this comment might be a little redundant with https://github.com/tobymao/18xx/pull/5016/files#r615523144)

lib/engine/step/track.rb
tobymao: i need more context? can you hit me up on slack about it?

lib/engine/game/g_18_ireland/entities.rb
jenf: yep

lib/engine/step/track.rb
jenf: @michaeljb unfortunatly you can't do that generalization as this is effectively a player choice for a discount on terrain costs, but is done using their normal placement rules. 1822 uses reachable without hexes to mean an extra lay.

We could add an attribute which would be 'counts towards lays' which would also tidy up 1817's privates a bit

lib/engine/game/g_18_ireland/step/special_track.rb
jenf: done

lib/engine/game/g_18_ireland/entities.rb
jenf: fixed

lib/engine/step/track.rb
jenf: I'm probably gonna rework this after some mild refactoring in 1817 to make this a lot easier.

jenf: should be good now.

tobymao (approved):

michaeljb (approved):

PR #5021, opened 2021-04-10
1849K2S balance changes

scottredracecar:
change to 4x 4H; starting capital 5p 330, 6p 300


scottredracecar:
This changes the train roster so the spec won't pass

jenf (approved):

PR #5028, opened 2021-04-11
1858: Initial config and auction

daniel-sousa-me:


lib/engine/step/dividend.rb
```diff
@@ -82,7 +82,9 @@ def log_run_payout(entity, kind, revenue, action, payout)
         end
       end
 
-      def share_price_change(_entity, revenue)
+      def share_price_change(entity, revenue)
+        return {} if entity.minor?
```
jenf, line 86: override this class rather than altering

lib/engine/game/g_1858/game.rb
```diff
@@ -0,0 +1,1236 @@
...
+
+        STARTING_CASH = { 3 => 500, 4 => 375, 5 => 300, 6 => 250 }.freeze
+
+        CAPITALIZATION = :incremental
+
+        MUST_SELL_IN_BLOCKS = true
+
+        LAYOUT = :flat
+
+        TILES = {
```
jenf, line 39: Split this file into map and entities.rb (see 18ireland/1849 for an example

lib/engine/game/base.rb
```diff
@@ -1105,7 +1105,7 @@ def all_companies_with_ability(ability_type)
       end
 
       def payout_companies
-        companies = @companies.select { |c| c.owner && c.revenue.positive? }
+        companies = @companies.select { |c| c.owner && c.owner != bank && c.revenue.positive? }
```
jenf, line 1108: Override rather than alter.

lib/engine/game/base.rb
```diff
@@ -2229,7 +2229,7 @@ def president_assisted_buy(_corporation, _train, _price)
       end
 
       def round_description(name, round_number = nil)
-        round_number ||= @round.round_num
+        round_number ||= @round&.round_num || 1
```
jenf, line 2232: Ditto

lib/engine/config/tile.rb
```diff
@@ -322,6 +322,7 @@ module Tile
         '105' => 'city=revenue:40,slots:3;path=a:0,b:_0,track:dual;path=a:1,b:_0,track:dual;path=a:2,b:_0,track:dual;label=BM',
         '106' => 'junction;path=a:0,b:_0,track:dual;path=a:2,b:_0,track:dual;path=a:4,b:_0,track:dual;path=a:3,b:_0,track:dual',
         '107' => 'junction;path=a:0,b:_0,track:dual;path=a:1,b:_0,track:dual;path=a:3,b:_0,track:dual;path=a:4,b:_0,track:dual',
+        '108' => 'junction;path=a:0,b:_0,track:dual;path=a:1,b:_0,track:dual;path=a:2,b:_0,track:dual;path=a:3,b:_0,track:dual',
```
jenf, line 325: This should be in the game config.

lib/engine/game/base.rb
daniel-sousa-me: This is just to prevent spurious message. It's not specific for this game at all

lib/engine/config/tile.rb
daniel-sousa-me: It is a standard tile. All other similar ones are here, why shouldn't this one be?

lib/engine/step/dividend.rb
daniel-sousa-me: Isn't this applicable to all minors in all games?

jenf: We have specific code for all other games that have minors, so it's best to stick with the same pattern.

lib/engine/game/base.rb
jenf: then this should be a separate PR then

lib/engine/config/tile.rb
jenf: fair enough... tis fine.

PR #5030, opened 2021-04-11
[1822] Wierd Dividend Choice

### LINKED ISSUE #5029: 1822 game has uncaught exception in console
I'm not sure what exactly caused this but this game no longer loads after an action was taken and there is an exception in the browser console

https://18xx.games/game/37076

magnusfalt:
- Added an extra check when choosing dividend.

fixes #5029

This shouldnt break any games.

### DIFF
--- a/lib/engine/game/g_1822/step/dividend.rb
+++ b/lib/engine/game/g_1822/step/dividend.rb
@@ -115,8 +115,7 @@ module Engine
               when 'half'
                 'half pay'
               else
-                @extra_train_choice = 'payout'
-                'pay out'
+                raise GameError, "#{action.choice} is an illegal choice"
               end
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1822/step/dividend.rb
```diff
@@ -114,6 +114,9 @@ def process_choose(action)
                 'withhold'
               when 'half'
                 'half pay'
+              else
+                @extra_train_choice = 'payout'
```
jenf, line 118: This should raise instead (and the broken game undone to the right point)

magnusfalt:
Its the last action 593 that needs to be undone.

jenf (approved):

PR #5041, opened 2021-04-12
1862: tile rendering fixes, hidden tiles

roseundy:
- Add "hidden" attribute to tiles and hide those tiles on the manifest
- Treat tiles that have only one town like multi-town tiles if there are other paths on the tile (i.e. don't center them in the track)
- When a tile has over two towns all with the same revenue, only render a single revenue

This is all pretty much common code changes except for the 1862 tiles.

Old tile rendering for 1862:
62_manifest_before

New tile rendering:
62_manifest_after

### DIFF
--- a/assets/app/view/game/tile.rb
+++ b/assets/app/view/game/tile.rb
@@ -39,6 +39,9 @@ module View
 
         return false if revenue.uniq.size > 1
 
+        # avoid obcsuring track with revenues
+        return true if @tile.cities.empty? && @tile.city_towns.size == 2 && @tile.exits.size > 4
+
         return false if @tile.cities.sum(&:slots) < 3 && @tile.city_towns.size == 2
 
         true
@@ -55,7 +58,10 @@ module View
         render_revenue = should_render_revenue?
         children << render_tile_part(Part::Track, routes: @routes) if !@tile.paths.empty? || !@tile.stubs.empty?
         children << render_tile_part(Part::Cities, show_revenue: !render_revenue) unless @tile.cities.empty?
-        children << render_tile_part(Part::Towns, routes: @routes) unless @tile.towns.empty?
+
+        unless @tile.towns.empty?
+          children << render_tile_part(Part::Towns, routes: @routes, show_revenue: !render_revenue)
+        end
 
         borders = render_tile_part(Part::Borders) if @tile.borders.any?(&:type)
         # OO tiles have different rules...
--- a/assets/app/view/game/tile.rb
+++ b/assets/app/view/game/tile.rb
@@ -39,7 +39,7 @@ module View
 
         return false if revenue.uniq.size > 1
 
-        # avoid obcsuring track with revenues
+        # avoid obscuring track with revenues
         return true if @tile.cities.empty? && @tile.city_towns.size == 2 && @tile.exits.size > 4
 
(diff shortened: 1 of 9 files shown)


michaeljb:
> When a tile has over two towns all with the same revenue, only render a single revenue

Your wording here sounds like the intent is for this to apply only to tiles with three or more towns, but in the screenshots, it applies to two or more, e.g., `780_2b`.

lib/engine/game/g_1862/map.rb
```diff
@@ -14,120 +14,139 @@ module Map
...
-            'code' => 'town=revenue:20;path=a:0,b:_0;path=a:2,b:_0;path=a:1,b:3',
+            'hidden' => 1,
+            'code' => 'town=revenue:20,loc:0;path=a:0,b:_0;path=a:2,b:_0;path=a:1,b:3',
           },
           '16_1b' =>
           {
             'count' => 2,
             'color' => 'green',
-            'code' => 'town=revenue:20;path=a:1,b:_0;path=a:3,b:_0;path=a:0,b:2',
+            'hidden' => 1,
```
michaeljb, line 24: Why is this tile still visible in the "new tile rendering" screenshot? Is that a screenshot of the `/tiles/1862/all` page, whereas the tile manifest tab in a game would be different?

roseundy: Exactly.

roseundy: I'm not sure it's a bug or a feature.

roseundy:
> > When a tile has over two towns all with the same revenue, only render a single revenue
> 
> Your wording here sounds like the intent is for this to apply only to tiles with three or more towns, but in the screenshots, it applies to two or more, e.g., `780_2b`.

Yeah, bad comment.

lib/engine/game/g_1862/map.rb
michaeljb: I might consider making a separate section for "hidden" tiles on the `/tiles/*/all` page, or maybe sorting them to the end or something. Not a big deal for now though.

assets/app/view/game/tile.rb
```diff
@@ -39,6 +39,9 @@ def should_render_revenue?
 
         return false if revenue.uniq.size > 1
 
+        # avoid obcsuring track with revenues
+        return true if @tile.cities.empty? && @tile.city_towns.size == 2 && @tile.exits.size > 4
```
michaeljb, line 43: > 
> 
> > > When a tile has over two towns all with the same revenue, only render a single revenue
> > 
> > 
> > Your wording here sounds like the intent is for this to apply only to tiles with three or more towns, but in the screenshots, it applies to two or more, e.g., `780_2b`.
> 
> Yeah, bad comment.


This is the LOC that makes this make sense now, I just noticed that `16_2` shows two circles, so I was confused about `780_2b` only showing one. So it's more like "When a tile has towns with 3+ towns all with the same revenue, or 2 towns and a bunch of track, only render a single revenue"

assets/app/view/game/tile.rb
```diff
@@ -39,6 +39,9 @@ def should_render_revenue?
 
         return false if revenue.uniq.size > 1
 
+        # avoid obcsuring track with revenues
```
michaeljb, line 42:
```suggestion
        # avoid obscuring track with revenues
```

PR #5046, opened 2021-04-13
[1866] Initial

magnusfalt:
- Initial commit of map, tiles and privates for 1866.


jenf (approved):
Changes look fine, dunno about the actual content :)

PR #5054, opened 2021-04-14
1856 false presidency

ventusignis:
Implements the nationalization scenario where there is a 1-share presidency

Also fixes nationalization (it was broken by other changes at some point)
Also fixes national being able to borrow a train if it has a 4T


roseundy (approved):
Looks fine to me.

PR #5058, opened 2021-04-14
1817DE Alpha

scottredracecar:


lib/engine/game/g_1817_de/meta.rb
```diff
@@ -8,15 +8,17 @@ module G1817DE
...
-        DEV_STAGE = :prealpha
+        DEV_STAGE = :alpha
         DEPENDS_ON = '1817'
 
         GAME_DESIGNER = 'Scott Petersen'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/1817DE'
         GAME_LOCATION = 'Germany'
         GAME_RULES_URL = {
-          '1817DE' =>
+          '1817DE Rule Differences' =>
```
jenf, line 18: is this supposed to be locked?

scottredracecar: Okay, wiki link is live now.

lib/engine/game/g_1817_de/meta.rb
```diff
@@ -8,15 +8,17 @@ module G1817DE
...
                           'https://docs.google.com/document/d/1lH3TcQc6etptZDWMbX1oZe7ne2VbbinAYkX8W_lpu-c/edit',
```
michaeljb, line 19: @scottredracecar I believe @jenf meant this link, it still seems to lock people out - https://docs.google.com/document/d/1lH3TcQc6etptZDWMbX1oZe7ne2VbbinAYkX8W_lpu-c/edit

Looks like we could use https://github.com/tobymao/18xx/wiki/1817DE instead?

scottredracecar: Yes, that makes more sense. Fixed it now.

PR #5070, opened 2021-04-16
1817DE Balance updates/fix home token

scottredracecar:
@jenf I added these two to the operating_round, but I don't really know if both are needed to make Erie style OOs work:
            Engine::Step::SpecialToken,
            Engine::Step::HomeToken,

Otherwise it is a copy/paste from 1817 operating_round, but at least it works now (it of course did not work before without these).


jenf:
Just home token, special token relates to private abilities

scottredracecar:
Okay, thanks, removed that now

scottredracecar:
Also, I have no idea if games will need to be archived. I suspect not as most are either completed or not far enough along to matter.

PR #5071, opened 2021-04-16
[1866] Map, entities and logos

magnusfalt:
- More work on the map, initial array of all the corporations with logos.


magnusfalt:
Updated this PR with some setup code and a few more logos.

tobymao (approved):

PR #5081, opened 2021-04-18
1822: fix tile numbers, add slot for gray S

scottredracecar:
Some tiles were numbered wrong in 1822v1 (Fixed in new printing)
145 -> 768 ("X")
146 -> 767 ("K")
147 -> 769 ("bird foot")

Also add slot for gray S tile to match both printings

@magnusfalt FYI


magnusfalt (approved):
This will break most of the 1822 games.

scottredracecar:
Yes, but better to fix it now than wait for more games to be made

michaeljb:
It may actually be possible to do a migration for this if we want to preserve the existing 1822 games. I'll try and take a look at that after dinner.

michaeljb:
The migration worked for all the affected 1822 games, and I found a couple spots that were missing the tile number changes (fixed in 106bcf1288f2761935f0a5c8fad9a26065af97fa).

Deploying now 🚀

michaeljb:
for posterity, this is the script I used to do the migration

```
#!/usr/bin/env ruby

require 'json'
load 'validate.rb'

IDS = JSON.load(File.read('ids.json'))
TILES = {"145"=>"768", "146"=>"767", "147"=>"769"}

IDS.each do |id|
  game = Game[id]

  game.actions.each do |action|
    next unless (action_tile = action.action['tile'])

    old_tile, tile_id = action_tile.split('-')
    next unless (new_tile = TILES[old_tile])

    action.action['tile'] = "#{new_tile}-#{tile_id}"
    action.save
  end
  game.save
end
```

magnusfalt:
Thank you @michaeljb. I have to learn how to do the migration scripts as well :)

michaeljb:
I didn't use the normal migration script here--that actually plays the games and has to re-play them whenever there's an error and it tries to repair the error by inserting/deleting/rearranging actions.

This case was easier (and much faster) because the actual game state at the actions doesn't matter, I just had to rewrite one attribute on the actions.

PR #5082, opened 2021-04-18
1862: implement routes and trains

roseundy:
Implement 1862 trains and route rules.
- Routes must intersect, one route must included home token (same as 1860)
- Track can be reused between routes, but not within a route
- Revenue locations only count once across all routes (same as 1873).
- Split assignment of revenues and bonuses across trains.
- Freight trains: hex trains. Much complexity due to the rule that multiple non-permanent freight trains have to be linked end-to-end and then treated as a single, long train. Permanent trains may or may not be linked. Had to build "sets" of these trains and detect violations of this as well as weird branching cases.
- Local and Express trains: fairly normal, except the M/N trains are much harder to optimize in 1862 due to the fact revenue centers only count once across all trains. I had to create a method that optimizes all trains at once. Revenue_str indicates which location was skipped.

Example of local trains:
1862_local

Example of express trains:
1862_express

Example of freight trains:
1862_freight

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -191,6 +191,7 @@ module Engine
             price: 100,
             rusts_on: '3F',
             num: 7,
+            no_local: true,
             variants: [
               {
                 name: '2L',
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1862/game.rb
```diff
@@ -888,6 +935,442 @@ def check_intersection(routes)
...
+          end
+        end
+
+        def create_oneway_set(visited, route, set, stops)
+          set << route
+          end_a = stops[route].first
+          end_b = stops[route].last
+          set.concat(get_set_node(end_a, route, stops)) if end_a != visited
+          set.concat(get_set_node(end_b, route, stops)) if end_b != visited
+          set.compact.uniq
```
tobymao, line 1073: why will there be nils here or dups? if this is a set there would be no dups, but this must not be a set

roseundy: I left these in due to paranoia. You are right, these are not true Ruby sets. I will experiment to make sure they aren't needed. In the "normal" case they aren't, but I was worried about being fed degenerate routes.

roseundy: Yes, both are needed. There can be dups and nils.

roseundy:
I found a redundant method that I've now deleted (remove_stops).

PR #5087, opened 2021-04-19
1862: disambiguate multiple offboards on one hex

roseundy:
I added these nifty red hexes with lines through them and two offboards to make the 1862 map look correct.
The problem is that the existing code can't deal with multiple offboards on one hex for the purpose of selecting a route. This is because it doesn't know which node to use for the touch_node method in route.

This PR fixes that by creating a "disambiguation" front end to touch_node for hexes with two offboards.

Hopefully, more than just 1862 can make use of it.

Example:
disambiguation

### DIFF
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -24,6 +24,7 @@ module Engine
       @node_chains = {}
       @connection_data = nil
       @last_node = nil
+      @last_offboard = []
       @stops = nil
     end
 
@@ -42,6 +43,7 @@ module Engine
       @halts = nil
       @connection_data = nil
       @last_node = nil
+      @last_offboard = []
     end
 
     def cycle_halts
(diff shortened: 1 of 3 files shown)


lib/engine/route.rb
```diff
@@ -163,6 +165,37 @@ def touch_node(node)
...
+    # hex with multiple offboards was clicked - determine which one to
+    # connect to the route
+    #
+    # we pick the one next to the first/last node in connected_data
+    # If both nodes could connect, cycle through them
+    def disambiguate_node(nodes)
+      onodes = nodes.select(&:offboard?)
+
+      # find first and last nodes in current route
+      list = connection_data.any? ? [head[:left], tail[:right]] : [@last_node]
```
tobymao, line 177: empty? and switch

lib/engine/route.rb
```diff
@@ -163,6 +165,37 @@ def touch_node(node)
...
+      return if list.empty?
```
tobymao, line 178: how can the list every be empty?

roseundy: If @last_node is nil

lib/engine/route.rb
roseundy: OK

PR #5098, opened 2021-04-20
1817: Add ability for base special track to consume a lay

jenf:
This will allow 18ireland's special track to be simpler

also fixes:

* Needing to Pass if Pittsburgh/Denver is unused
* Not paying $20 if using Pittsburgh second
* Somehow laying mines on green tiles


lib/engine/step/tracker.rb
```diff
@@ -172,9 +176,9 @@ def pay_all_tile_income(company, ability)
             " for the tile built by #{company.name}"
       end
 
-      def pay_terrain_tile_income(company, ability, terrain, entity)
+      def pay_terrain_tile_income(company, ability, terrain, entity, spender)
         return unless terrain.include?(ability.terrain)
-        return if ability.owner_only && company.owner != entity
+        return if ability.owner_only && (company.owner != entity && company.owner != spender)
```
crericha, line 181: Are the parens necessary?

crericha (approved):

crericha (approved):

PR #5106, opened 2021-04-21
[1866] IAR and ISR

magnusfalt:
**Changes to base code**
- View::Game::Par -1866 have multiple par values which is the same value. I needed a way to make them different when you get the par buttons. So instead of just "#{@game.format_currency(share_price.price)}" it will call a method in the base game file "#{@game.par_price_str(share_price)}" which i override in the 1866 file.

**1826 Specific**
- Initial Auction Round - this is a selection auction.
- Initial Stock Round - Mostly basic buy and sell shares stuff with a small exception of the buying of Stock Turn Tokens.


lib/engine/game/g_1866/game.rb
```diff
@@ -241,27 +251,131 @@ class Game < Game::Base
...
+                    when 1
+                      'M'
+                    else
+                      'B'
+                    end
+          "#{format_currency(share_price.price)}#{row_str}"
+        end
+
+        def phase_par_type
+          par_types = {
```
tobymao, line 368: should be a constant

tobymao (approved):

PR #5115, opened 2021-04-22
1822: Automatically determine destination connectivity

### LINKED ISSUE #4680: 1822 add auto actions for laying destination token (or skipping lay destination token)
Since there is no "choice" when laying the destination token, the user should not be prompted to either skip or lay the destination token. It should be done with an auto action.

In this example, LYR must lay its destination token. The token should be automatically laid with no user interaction (log the placement as normal).
https://18xx.games/game/33787?action=231

Conversely, in this action, LYR has not reached its destination so an auto action should skip the destination token placement automatically (no comment).
https://18xx.games/game/33787?action=191

jenf:
fixes #4680

### DIFF
--- a/lib/engine/game/g_1822/step/destination_token.rb
+++ b/lib/engine/game/g_1822/step/destination_token.rb
@@ -16,6 +16,14 @@ module Engine
             ACTIONS
           end
 
+          def auto_actions(entity)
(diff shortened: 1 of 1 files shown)


magnusfalt (approved):
Looks good :)

PR #5116, opened 2021-04-22
1862: buy trains

roseundy:
Implement most of the buy_trains step for 1862, except for the emergency train buying rules.
- separate limits for each train type (freight/local/express) until phase G
- trains are now lettered A-H with the actual trains being options
- trains bought across companies are fixed price
- add train warranties that can be bought with new trains from depot
- warranties are kept track of by appending a number of *s to the name, every OR one * is removed
- implement deferred rusting due to warranties

This PR has several common file changes:
- UI::GameInfo, Engine::Game::Base - I needed to add control over how the train types and prices were displayed
- UI::BuyTrains - above, plus UI element for warranty purchases, fixed prices for used trains
- Action::BuyTrains - warranties

Old game info:
old_info

New game info:
new_info

New buy train UI:
new_buy_trains

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -1441,8 +1441,7 @@ module Engine
         end
 
         def info_train_name(train)
-          train.names_to_prices.keys.find { |n| /^[A-H]$/ =~ n } + ': ' +
-            train.names_to_prices.keys.reject { |n| /^[A-H]$/ =~ n }.join(', ')
+          train.sym + ': ' + train.names_to_prices.keys.reject { |n| n == train.sym }.join(', ')
         end
 
         def info_train_price(train)
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -12,6 +12,7 @@ require_relative 'step/track'
 require_relative 'step/token'
 require_relative 'step/route'
 require_relative 'step/dividend'
+require_relative 'step/buy_train'
 
 module Engine
   module Game
@@ -21,7 +22,7 @@ module Engine
         include Entities
         include Map
 
-        attr_accessor :chartered, :base_tiles
+        attr_accessor :chartered, :base_tiles, :deferred_rust
 
         register_colors(black: '#000000',
                         orange: '#f48221',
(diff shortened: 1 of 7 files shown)


roseundy:
I also noticed and fixed an independent UI bug: #5120

lib/engine/game/g_1862/game.rb
```diff
@@ -1365,6 +1427,57 @@ def revenue_str(route)
...
+        def train_limit_by_type(entity)
+          @phase.train_limit(entity)
+        end
+
+        def used_train_price(train)
+          train.sym == @phase.name ? train.price : (train.price / 2).to_i
+        end
+
+        def info_train_name(train)
+          train.names_to_prices.keys.find { |n| /^[A-H]$/ =~ n } + ': ' +
```
tobymao, line 1444: what is this stuff

lib/engine/game/g_1862/step/buy_train.rb
```diff
@@ -0,0 +1,116 @@
...
+          def fixed_price(train)
+            @game.used_train_price(train)
+          end
+
+          def warranty_text
+            'Warranties (each)'
+          end
+
+          def warranty_max
+            /A|D/.match?(@depot.depot_trains.first.sym) ? 2 : 3
```
tobymao, line 102: why are you using regex

lib/engine/game/g_1862/game.rb
roseundy: There are 8 train levels, A-H, that correspond to each of the phases. Within each level, there are 3 train types that can be bought (freight, local and express). I started out with this "stuff" but the info page was confusing. The first C train, for instance will rust all A trains, but the train/phase roster only showed that a 3F would rust a 1F. This method will show the train level followed by the trains at that level.

In addition, I collapse all of the prices for a single level down to one, since they are all the same.

lib/engine/game/g_1862/step/buy_train.rb
roseundy: Since the phases/train levels are named A-H, it seemed the easiest way to check against multiple values.

Are regexps bad? As a former perl coder, I was happy to find that Ruby supported them.

roseundy: OK, removed.

lib/engine/game/g_1862/game.rb
roseundy: Removed regexps

tobymao (approved):

PR #5123, opened 2021-04-23
1870: there were five copies of tiles 5 and 6 instead of 2 (fixes #5112)

### LINKED ISSUE #5112: 1870 tile manifest incorrect
The yellow city tiles have incorrect amounts ... There should be only two #5s, two #6s, and four #57s ... Not five of each.

Thanks

Randall

daniel-sousa-me:
May archive games


jenf:
Please find a new spec to replace the existing one

daniel-sousa-me:
Yeah, I'll try to fix all the bugs today and then I'll find one that still works. Marking as draft for now

jenf:
Nearly all of them break.
Best option is https://18xx.games/game/34110

daniel-sousa-me:
I have updated the fixture to that game, but it's likely that some other fix will break it

jenf (approved):

PR #5126, opened 2021-04-24
[1840] Company Round

Zwergenpunk:
Adds most of the company round.

The round is a little bit complicated, because it technically consists of multiple rounds.
The corporations don't finish all their actions and then pass, but each corporation does one or two things and then its the next corporations turn etc.

I can imagine that the Round handling gets refactored again, when the full Line Rounds gets added. 

Most of the code is just adding an auction for corporations, which are handled like a merge + bidding in the selection auction style.

image

And the switch train view is extended to support reassigning trains between arbitrary corporations.

image

I'm not sure if i add a new Action + view for that reassigning or if it is ok there. The slot magic is a little bit hacky, but I do have to extract the corporationId and trainId in some way, and that works pretty good for now.


lib/engine/action/switch_trains.rb
```diff
@@ -14,7 +14,11 @@ def initialize(entity, slots: nil)
 
       def self.h_to_args(h, _game)
         {
-          slots: h['slots']&.map { |m| m.to_i },
+          slots: h['slots']&.map do |m|
+            Integer(m)
+          rescue StandardError
```
tobymao, line 19: what is this

assets/app/view/game_page.rb
```diff
@@ -407,7 +407,11 @@ def render_action
       when Engine::Round::Auction
         h(Game::Round::Auction, game: @game, user: @user)
       when Engine::Round::Merger
-        h(Game::Round::Merger, game: @game)
+        if current_entity_actions.include?('buy_train')
```
tobymao, line 410: why do you need this? is this gonna break any other game?

lib/engine/game/g_1840/game.rb
```diff
@@ -161,6 +252,77 @@ def all_major_corporations_ipoed?
         def can_par?(corporation, parrer)
           super && corporation.type == :major
         end
+
+        def show_progress_bar?
+          true
+        end
+
+        def progress_information
+          [
```
tobymao, line 261: constant

lib/engine/game/g_1840/round/company_operating.rb
```diff
@@ -0,0 +1,63 @@
...
+            'CR'
+          end
+
+          def name
+            'Company Round'
+          end
+
+          def select_entities
+            entites = @game.operating_order
+            entites = entites.reject { |item| item.type == :minor }
```
tobymao, line 27: can you do rejecT! here?

lib/engine/game/g_1840/step/acquisition_auction.rb
```diff
@@ -0,0 +1,220 @@
...
+            'Acquire Corporations'
+          end
+
+          def pass_description
+            'Pass (Bid)'
+          end
+
+          def log_pass(entity)
+            message = "#{entity.name} passes bidding"
+            message += "on #{@auctioning.name}" if @auctioning
```
tobymao, line 35: <<

lib/engine/game/g_1840/step/acquisition_auction.rb
```diff
@@ -0,0 +1,220 @@
...
+          end
+
+          def log_skip(entity)
+            @log << "#{entity.name} cannot afford bidding on #{@auctioning.name}"
+          end
+
+          def process_bid(action)
+            corporation = action.corporation
+
+            if @bids[corporation].any?
```
tobymao, line 46: empty

lib/engine/game/g_1840/step/acquisition_auction.rb
```diff
@@ -0,0 +1,220 @@
...
+          end
+
+          def show_bidding_corporation?
+            true
+          end
+
+          def min_bid(company)
+            return unless company
+
+            return starting_bid(company) unless @bids[company].any?
```
tobymao, line 112: empty

lib/engine/game/g_1840/step/reassign_trains.rb
```diff
@@ -0,0 +1,66 @@
...
+              new_corporation.trains << train
+
+              reassignments << "#{train.name} ➝ #{new_corporation.name}"
+            end
+
+            @log << "#{entity.owner.name} reassignes trains: #{reassignments.join(', ')}"
+          end
+
+          def trains(entity)
+            entity.trains + @game.tram_owned_by_corporation[entity].flat_map(&:trains)
```
tobymao, line 56: more efficient to do flat_map(&trains(.concat(entitytrains)

lib/engine/action/switch_trains.rb
Zwergenpunk: The 1873 implementations needs an int but I have a string Token and thats the only way in ruby tontry parse an int.

I could of course create a new action but I thought I reuse as much as possible

assets/app/view/game_page.rb
Zwergenpunk: I need to buy a train during merge round. If someone else had this action it should have shown something. But without that nothing is displayed. 

We can add the buy train view to merge as well, but I think showing the map etc is better and thats the easiest way

lib/engine/action/switch_trains.rb
Zwergenpunk: I think I'm creating a new view and a new action. The more I think about it the less I like it.

tobymao: ok so you will redo this?

Zwergenpunk: New view + Action is available

tobymao (approved):

Issue #4308, opened 2021-03-04
Game Log: report stock price and token movement


tysen:
https://18xx.games/game/28128?action=163

GT is fully owned and so moves up one space in the market. HBR as well. The log only reports HBR though, I assume because the share price values for the space GT left and the space GT entered are equal.

yzemaze:
> I assume because the share price values for the space GT left and the space GT entered are equal.

Exactly:
https://github.com/tobymao/18xx/blob/68f86a0f08dec9f28c7a4a620c73d4dcada49dd7/lib/engine/game/base.rb#L859-L865

As it’s designed to report stock price _changes_ and not _movements_, it’s not a bug.
Do you think reporting movement and price changes would be worthwhile?

dfannius:
I do.

- It's useful information.
- We call out all other movement on the market.
- The current lack of message is confusing enough that once a month or so someone "erroneously" reports a bug (e.g., "this stock didn't move even though it was sold out").

dfannius:
One proposal is to change the message from "AAA's share price changes from $X to $Y" to "AAA's share price moves [up/down/left/right] from $X to $Y". This also would make it clearer to newer players how market movement works, since right now it's easy for them not to pay attention to up vs right, or down vs left.

It would also be nice to call out ledging.

thepackrat:
I can take this on a general extension to the multijump reporting.

PR #5131, opened 2021-04-25
Report stock movement

cfcurtis:
Reports stock movement and ledging/ceiling instead of just price change as suggested in #4308.

This is my first contribution to this project, so I am open to suggestions/critiques and would appreciate explanations if I have done something wrong. Thank you!

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -896,19 +896,17 @@ module Engine
       end
 
       def log_share_price(entity, from)
-        msg = entity.share_price.report_last_move()
+        msg = entity.share_price.report_last_move
+        return unless msg
 
-        if msg
-          msg = "#{entity.name}'s share price " + msg
-          to = entity.share_price.price
-          if from != to
-            msg += "from #{format_currency(from)} "\
-              "to #{format_currency(to)}"
-          else
-            msg += "and stays at #{format_currency(from)}"
-          end
-          @log << msg
-        end
+        msg = "#{entity.name}'s share price " + msg
+        to = entity.share_price.price
+        msg += if from != to
+                 "from #{format_currency(from)} to #{format_currency(to)}"
+               else
+                 "and remains at #{format_currency(from)}"
+               end
+        @log << msg
       end
 
       def can_run_route?(entity)
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -896,11 +896,19 @@ module Engine
       end
 
       def log_share_price(entity, from)
-        to = entity.share_price.price
-        return unless from != to
+        msg = entity.share_price.report_last_move()
 
-        @log << "#{entity.name}'s share price changes from #{format_currency(from)} "\
-          "to #{format_currency(to)}"
+        if msg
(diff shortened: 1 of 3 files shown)


roseundy:
Has this been tested on games with linear markets?

Also, this seem pretty complicated. Wouldn't it be easier to just check the before/after price objects rather than the price.price values?

cfcurtis:
> Has this been tested on games with linear markets?

I've tested on some 1817 and 1846 games but it's certainly not exhaustive.

> Also, this seem pretty complicated. Wouldn't it be easier to just compare the before/after price objects rather than the price.price values?

It did get quite complicated. Somehow changing all the calls to log_share_price to take a price object rather than a value felt like a bigger change, but I can do that if it would be better. I think I'd still need to keep track of ledging in stock_market.rb though.

tobymao:
yea maybe that would be better

lib/engine/stock_market.rb
```diff
@@ -55,15 +55,25 @@ def move_up(corporation)
       return move_right(corporation) if one_d?
```
crericha, line 55: Many games override stock_market movement. They will also need to be looked at.

lib/engine/share_price.rb
```diff
@@ -107,5 +110,31 @@ def normal_movement?
...
+        end
+        if c0 != c1
+          msg += 'and ' if r0 != r1
+          msg += "#{(c1 - c0).abs} #{c1 < c0 ? 'left' : 'right'} "
+        end
+      elsif @limit_hit
+        msg += "is #{@limit_hit == :ledge ? 'ledged' : 'at ceiling'} "
+      end
+
+      @last_reported = @coordinates
```
crericha, line 135: Since the last reported coordinates are being set in the stock market move method, why does it need to be set here? I am concerned that having a report method change state will produce unpredictable behavior in edge cases.

lib/engine/game/base.rb
```diff
@@ -896,11 +896,17 @@ def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil
       end
 
       def log_share_price(entity, from)
-        to = entity.share_price.price
-        return unless from != to
+        msg = entity.share_price.report_last_move
```
crericha, line 899: The stock market moves multiple times at the end of the stock round in some games (e.g. 1828). Make sure that case is handled properly.

lib/engine/stock_market.rb
cfcurtis: Good to know, thank you. I'm starting to suspect this isn't actually a great first issue to have chosen.

lib/engine/share_price.rb
cfcurtis: Following @roseundy's suggestion I'm working on a better version that modifies log_share_price to compare two share_price objects instead of keeping track of last_reported coordinates, so hopefully this issue won't appear.

cfcurtis:
Closing this PR for now because the implementation isn't ideal and merging this in would probably cause more harm than good. I appreciate the reviews and will try again, as it's still a feature I'd like to see.

PR #5159, opened 2021-04-28
18MEX Baja variant fixes

### LINKED ISSUE #5044: [18MEX Baja] Uncaught error for CdB action when out of tokens
When you try to use the CdB action, but the company is out of tokens, you get an error message "undefined method 'price=' for nil"

### LINKED ISSUE #5099: 18MEX Baja private needs more text
https://18xx.games/game/38278?action=292

The Compagnie Du Boleo charter should state A) it can be bought by a corporation in phase 2 and B) it can only be bought for 50-100% of face value (rather than 50-150 for the other companies).

michaeljb:
* better baja private wording [Fixes #5099]

* if owning corp has no tokens left on charter, ability tokens with `from_owner`
  should cause `ability_usable?` to return `false` [Fixes #5044]

    * exception: in 1882, the NWR ability moves a token on the map so it still
      works if the owning corp has none left

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2437,6 +2437,8 @@ module Engine
             end
           return true unless corporation
 
+          return false unless token_ability_from_owner_usable?(ability, corporation)
+
           tokened_hexes = []
 
           corporation.tokens.each do |token|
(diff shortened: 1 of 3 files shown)


michaeljb:
@ChrisTheCat what do you think of this wording?

> May play token in Santa Rosalia at no cost and in addition to regular token lay. May be purchased by a corporation for $15-$30, from Phase 2.

ChrisTheCat:
Works for me.

tobymao (approved):

PR #5162, opened 2021-04-28
18Ireland: Show corporations in IPO order

jenf:
Also add them to the timeline to make it easier to see

### DIFF
--- a/lib/engine/game/g_18_ireland/game.rb
+++ b/lib/engine/game/g_18_ireland/game.rb
@@ -394,6 +394,20 @@ module Engine
           super
         end
 
+        def unstarted_corporation_summary
+          unipoed = @corporations.reject(&:ipoed)
+          minor = unipoed.select { |c| c.type == :minor }
+          major = unipoed.select { |c| c.type == :major }
+          ["#{major.size} major", minor]
+        end
+
+        def timeline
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_ireland/game.rb
```diff
@@ -394,6 +394,20 @@ def upgrade_cost(old_tile, hex, entity)
           super
         end
 
+        def unstarted_corporation_summary
+          unipoed = @corporations.reject(&:ipoed)
+          minor = unipoed.select { |c| c.type == :minor }
+          major = unipoed.select { |c| c.type == :major }
```
crericha, line 400: Use partition?

lib/engine/game/g_18_ireland/game.rb
```diff
@@ -394,6 +394,20 @@ def upgrade_cost(old_tile, hex, entity)
...
+          ["#{major.size} major", minor]
+        end
+
+        def timeline
+          timeline = []
+          minors = @corporations.reject(&:ipoed).select { |c| c.type == :minor }.map(&:name)
```
crericha, line 406: Check both conditions in the select, so you only have to iterate through the array once.

PR #5164, opened 2021-04-28
[1870] fix valid hexes for P3 (SCC)

### LINKED ISSUE #5151: [1870] SCC Token Cannot be placed in Houston
1870 game 38345.

I am unable to place the Southern Cattle Company token in Houston [M6] (which should be allowed), but I am able to place it in Baton Rouge [M14] (which should be illegal).

benjaminxscott:
closes #5151 

References:

> "P3 (SCC) token may be placed in any city west of the Mississippi. Cities located in the same hex as any portion of the Mississippi river are not allowed."

Rules: http://www.hexagonia.com/rules/MFG_1870.pdf
Map: https://18xx.games/map/1870

### DIFF
--- a/lib/engine/game/g_1870/game.rb
+++ b/lib/engine/game/g_1870/game.rb
@@ -289,7 +289,7 @@ module Engine
             abilities: [
               {
                 type: 'assign_hexes',
-                hexes: %w[B9 B11 D5 E12 F5 H13 J3 J5 L11 M2 M6 N1 N7],
+                hexes: %w[B9 B11 D5 E12 F5 H13 J3 J5 L11 M2 M6 N7],
                 when: 'owning_corp_or_turn',
                 count: 1,
                 owner_type: 'corporation',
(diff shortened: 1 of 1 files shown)


benjaminxscott:
May want to hold-merge until confirmation that `N1` is also valid placement

tobymao:
confirmatino from who

benjaminxscott:
Asked [in slack](https://18xxgames.slack.com/archives/C012K0CNY5C/p1619903962378500)

Since it's alpha IMO it's OK to merge and fix later if it's incorrect

michaeljb:
From page 3 in the linked rulebook:

> Small cities are represented on the map [...]
>
> Large cities are represented on the map [...]
>
> There are four red off-board locations on the 1870 map. They represent destinations that are not on the map, but were important to railway development in the region. The black triangles are their connecting points to track from adjacent hexes.

As in most other 18xx games, the red offboard region is not limited to the hex that we call N1 - https://boardgamegeek.com/image/120350/1870-railroading-across-trans-mississippi-1870

My reading is that N1 is not a valid target; it is an "off-board location" rather than a "city".

benjaminxscott:
Updated this PR to remove N1

Issue #5168, opened 2021-04-29
[1822] Do not display Close Message for Highland/mini-Highland if they are not open


Maxwell-Frank:
https://18xx.games/game/39386?action=696

Currently, when a grey train is purchased, an event appears in the log that reads `Event: P15-HR and P20-C&WR now closes and its money returned to the bank --`

This causes some confusion because it could imply that one, or both, of the privates were still open (i.e. in player hands) and actually closed.  This message appears even in a 22MRS scenario where neither private were even in the game.

KevinR1000:
Relevant code is in `/lib/engine/game/g_1822/game.rb ` at about line 729.
`def event_phase_revenue!`

dfannius:
Fixed by #5173 (as brought to my attention by @andypymont).

PR #5173, opened 2021-04-29
[1822] Better logging when phase revenue companies close

KevinR1000:
Companies P15-HR and P20-C&WR close at beginning of phase 7 if player-owned (in `def event_phase_revenue!`).
Rather than generic statement, report to log the balance and owner of each closing company.
See issue https://github.com/tobymao/18xx/issues/5168 .

### DIFF
--- a/lib/engine/game/g_1822/game.rb
+++ b/lib/engine/game/g_1822/game.rb
@@ -733,7 +733,7 @@ module Engine
             next if !company || company&.closed? || !@phase_revenue[company_id]
 
             @log << "#{company.name} closes"
-              # "(#{format_currency(@phase_revenue[company_id].cash)}, #{company.owner.name}) "
+            # "(#{format_currency(@phase_revenue[company_id].cash)}, #{company.owner.name}) "
             if @phase_revenue[company.id].cash.positive?
               @phase_revenue[company.id].spend(@phase_revenue[company.id].cash, @bank)
             end
--- a/lib/engine/game/g_1822/game.rb
+++ b/lib/engine/game/g_1822/game.rb
@@ -727,13 +727,12 @@ module Engine
         end
 
         def event_phase_revenue!
-          @log << '-- Event: P15-HR and P20-C&WR now close with money returned to the bank --'
+          @log << '-- Event: Phase revenue companies now close with money returned to the bank --'
           self.class::PRIVATE_PHASE_REVENUE.each do |company_id|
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1822/game.rb
```diff
@@ -727,11 +727,14 @@ def event_full_capitalisation!
...
         def event_phase_revenue!
-          @log << '-- Event: P15-HR and P20-C&WR now closes and its money returned to the bank --'
+          @log << '-- Event: Phase revenue companies now close with money returned to the bank --'
           self.class::PRIVATE_PHASE_REVENUE.each do |company_id|
             company = @companies.find { |c| c.id == company_id }
             next if !company || company&.closed? || !@phase_revenue[company_id]
 
+            @log << "#{company.name} "\
+              "(#{format_currency(@phase_revenue[company_id].cash)}, #{company.owner.name}) "\
+              'closes'
```
crericha, line 737: Why include cash and owner name? I don't believe any other private closing log message includes things like that. my personal preference is for consistency unless good reason.

lib/engine/game/g_1822/game.rb
```diff
@@ -727,11 +727,14 @@ def event_full_capitalisation!
         end
 
         def event_phase_revenue!
-          @log << '-- Event: P15-HR and P20-C&WR now closes and its money returned to the bank --'
+          @log << '-- Event: Phase revenue companies now close with money returned to the bank --'
```
crericha, line 730: "Phase revenue companies" is misleading as P9 also provides phase revenue.

KevinR1000:
"Phase revenue" is the name the Info page uses for the event.
`Phase revenue | P15-HR and P20-C&WR closes if not acquired by a major company`

Probably no good reason to call out the player who made a mistake (abandoning money), but it saves me from having to look back in the history to see who made the mistake.

I have no objection to reverting the current log entry and stripping details out of the new entries.

KevinR1000:
A possibly good reason to use the "phase revenue companies" language:
-  This code is apparently used in 1822 variants where one or both of P15-HR and P20-C&WR do not exist.  See issue.
-  If the language is generic, this code could be used in other 1822 games where similar companies exist.

Neither of those is overwhelming, however.

crericha:
If phase revenue is called out as P15 and P20 on the info page, then its fine.

Yeah, let's not call out the player. We can secretly judge them. ;-)

crericha (approved):

PR #5175, opened 2021-04-29
[1840] Added Basic Line Operation Round

Zwergenpunk:
It is now possible to lay orange framed tiles, default tiles and tokens in one step

Also added the fixed rotation for the orange tiles

OR Handling should now be complete (just no end yet)

Added Spender to place_token

I'm busy over the weekend, so don't be suprised if nothing happens :)

### DIFF
--- a/lib/engine/game/g_1840/round/line_operating.rb
+++ b/lib/engine/game/g_1840/round/line_operating.rb
@@ -8,7 +8,7 @@ module Engine
       module Round
         class LineOperating < Engine::Round::Operating
           def select_entities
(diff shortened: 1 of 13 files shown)


lib/engine/game/g_1840/round/line_operating.rb
```diff
@@ -18,6 +18,15 @@ def self.short_name
           def name
             'Line Round'
           end
+
+          def after_process(action)
+            entity = @entities[@entity_index]
+            case action
```
tobymao, line 24: not sure you need a case here


process_action if action.is_a?(x) && !action.routes.empty?

PR #5177, opened 2021-04-29
1860: Disable auto actions if the SR restarts

### LINKED ISSUE #5170: Bankruptcy in 1860 doesn't cancel auto actions
Selling shares normally cancels auto actions. But if a player sells shares, driving a company bankrupt, the SR immediately ends, and auto-actions are never cancelled.

A bankruptcy is an important *Event*. It should cancel auto actions, rather than somehow preserving the previous SR's auto actions.

We ran into this in game 39135

jenf:
fixes #5170

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2017,6 +2017,10 @@ module Engine
           end
       end
 
+      def clear_programmed_actions
+        @programmed_actions.clear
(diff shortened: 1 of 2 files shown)


crericha (approved):

roseundy:
FWIW, this looks good.

PR #5178, opened 2021-04-29
18Ireland: If no minor is started in SR, close a minor

jenf:
Also hide majors until phase 6

### DIFF
--- a/lib/engine/game/g_18_ireland/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_ireland/step/buy_sell_par_shares.rb
@@ -35,6 +35,7 @@ module Engine
 
             @auctioning = nil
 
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_ireland/step/buy_sell_par_shares.rb
```diff
@@ -69,6 +70,12 @@ def ipo_type(entity)
               'Minor is not the first available minor'
             end
           end
+
+          def round_state
+            super.merge({
```
tobymao, line 75: unnened formatting

tobymao: super.merge(minor_started: false) i think works

tobymao (approved):

Issue #5011, opened 2021-04-09
1868 Wyoming check list


michaeljb:
1868: Boom and Bust in the Coal Mines and Oil Fields of Wyoming

Should have started taking notes right when I started reading through the rules; will fill in all stuff before 11.3.4 later

# New Stuff

Issues found / things changed after the original list was implemented

## Bugs

- [x] undoing or using history nav breaks rendering uranium town revenue (it shows the base revenue instead of the revenue with the uranium bonus)
- [ ] development tokens are often rendered on top of boom city revenue

## New Features

- [x] new private companies (holding for more IRL playtesting)
- [x] three stacks of corporations, only top of each can be started
- [x] more corporations, including some with purple homes
- [ ] autopass for development rounds

# Original List

## Tile Rendering

- [x] boom towns
- [x] boom cities
- [x] bb/Bb/bB tiles
- [x] east/west bonuses
- [x] city name positioning/wrapping for K11 - "Atlantic City & South Pass City"
- [x] uranium icon (currently using port as placeholder)
- [x] better purple for Rapid City and Chadron, custom brown for Yellowstone
    - need to make `register_colors(purple: '#6f30a0')` work for a hex that is colored `:purple` (and same thing for `:yellowstone`)
- [x] better white for boom town/city revenues
- [x] offboard revenue placement

## Tiles

- [x] boomtown/boomcity tiles need to be implemented as proper double-sided tiles

## Stock Market

- [x] better blue for par values

## Company Abilities
- [x] P3/P5/P6 - choose one of the three connected to it
- [x] P1 Hell on Wheels
    - [x] close when golden spike is complete
- [x] P2 Wylie Permanent Camping
- [x] P3a Union Pac Coal
    - [x] 11.2.1.2 - UP coal DT
    - [x] place at same time other coal DT would be placed
    - [x] may be moved, terrain costs paid each time
    - [x] cube goes away, private closes at phase 5
- [x] P3b Buffalo Bill's Bonanza Oil
    - [x] 11.2.1.3 - BZ oil DT, placed in phase 5
    - [x] $0 base cost to place, $0 to maintain
    - [x] may be moved
    - [x] must pay terrain to place
- [x] P3c Fremont Expedition
    - [x] 11.2.1.4 - $20 discount from terrain cost for DT placement
- [x] P4a Grenville M Dodge
    - [x] one-time-use extra tile lay, 2 or 3 yellows, ignore terrain costs, then close
- [x] P4b Edward Gillette
[...]

PR #5181, opened 2021-04-30
make all 1868WY towns dots; fix rect towns on 1860 tiles

### LINKED ISSUE #5174: 1860 Cowes tile displays small town differently from other tiles
All of the 1860 tiles follow the convention used by the physical printing of the game, that small towns and halts are dots when they appear at a junction and perpendicular dashes when they appear along a track segment. Except for the yellow, green, and brown Cowes tile #787-789. It has a small town that appears as a black dot instead of the dash that you see on 747, 748, 749, 764, 765, 766.

The small town on tiles 787-789 should be a dash, not a dot.

michaeljb:
progress on #5011, fixes #5174

### DIFF
--- a/lib/engine/game/g_1868_wy/game.rb
+++ b/lib/engine/game/g_1868_wy/game.rb
@@ -33,6 +33,18 @@ module Engine
           %w[20b 30b 40o 50o 55y 60y 64 68 72],
           ['', '20b', '30b', '40o', '50o', '55y', '60y'],
         ].freeze
+
+        def init_tiles
+          super.each do |tile|
+            tile.towns.each { |town| town.style = :dot }
+          end
+        end
+
+        def init_hexes(companies, corporations)
+          super.each do |hex|
+            hex.tile.towns.each { |town| town.style = :dot }
+          end
+        end
       end
     end
   end
--- a/lib/engine/part/town.rb
+++ b/lib/engine/part/town.rb
@@ -5,6 +5,7 @@ require_relative 'revenue_center'
 module Engine
   module Part
     class Town < RevenueCenter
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_1868_wy/game.rb
```diff
@@ -33,6 +33,22 @@ class Game < Game::Base
           %w[20b 30b 40o 50o 55y 60y 64 68 72],
           ['', '20b', '30b', '40o', '50o', '55y', '60y'],
         ].freeze
+
+        def init_tiles
+          tiles = super
```
tobymao, line 38: super.each do 
end

may work? cuz it returns self

lib/engine/part/town.rb
```diff
@@ -23,11 +25,13 @@ def town?
...
-      # it has any paths and either it's not in the center or it is in the center
-      # and has less than two exits and less than three paths
+      # render as a dot or rectangle; if @style is set to 'rect', then render a
+      # rectangle; if @style is set to something else, render a dot; if @style is
+      # unset, render with a rectangle if it has any paths and either it's not
+      # in the center or it is in the center and has less than two exits and
+      # less than three paths
       def rect?
-        paths.any? && paths.size < 3 && (@loc != 'center')
+        @style ? (@style == :rect) : (paths.any? && paths.size < 3)
```
tobymao, line 34: !paths.empty?

tobymao (approved):

lib/engine/game/g_1868_wy/game.rb
michaeljb: ah yep, that works

PR #5198, opened 2021-05-01
[1868WY] basic configuration changes

michaeljb:
* fix IPO name
* can buy companies from phase 3 for $1 to face
* various constant-based rules
* unlimited tiles except for C, L, G, and gray tiles
* hide ghost town (GT) tiles from manifest
* remove the plain ghost town tile, can just destroy the town instead
* remove the wind river canyon tile (and WRC label from the map hex), can just
  use a plain yellow tile and turn it gray after its laid

[Issue #5011]


michaeljb:
Not really any code here to review, but wanted to get the check mark for build stuff passing before merging it into master.

PR #5217, opened 2021-05-02
1860: check for bankruptcy in process_dividend

### LINKED ISSUE #5213: 1860 Game doesn’t end if final company is about to go bankruot
https://18xx.games/game/34878

A company hit max stock value and so the game ended at the end of that OR, however the final company to operate VYSC dropped to 0 and went bankrupt.

after the “skip train” action for the previous company in the train buying step there are no more actions to be taken by players in the game but the game does not end. Has to be ended manually.

roseundy:
Fixes #5213


roseundy:
There is a small chance this will require pins.

michaeljb (approved):
I found 3 games (out of 1006) that will need pinning, the one in issue #5213 (or the end_game action could just be deleted from that one), and apparently 2 more where more routes were run after the game was supposed to be over.

PR #5226, opened 2021-05-02
18Ireland: Donegal shouldn't force connectivity to normally laid tiles

### LINKED ISSUE #5193: 18Ireland Id: Dongal tries to force other tiles to be connected
https://18xx.games/game/39559
DR power fails while laying track in F4 and E3 see SC

Screenshot_20210501-105913_Chrome
Screenshot_20210501-110010_Chrome
Screenshot 2021-05-01 11123618Ireland

Windows 10 goole Chrome

 one tile lay it is accepted

Ire

not sure if the following is also a bug?
teh small track bonus does not appear any longer after Colloney (E7) has been upgraded?

Ire1

playing one more OR the small track bonus is calculated right
Ire3

this constellation fails also:
Screenshot_20210501-120140_Chrome

jenf:
fixes #5193

### DIFF
--- a/lib/engine/step/special_track.rb
+++ b/lib/engine/step/special_track.rb
@@ -57,6 +57,7 @@ module Engine
           lay_tile_action(action, spender: owner)
         else
           lay_tile(action, spender: owner)
+          ability.laid_hexes << action.hex.id
           @round.laid_hexes << action.hex
           check_connect(action, ability)
         end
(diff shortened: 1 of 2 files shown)


lib/engine/step/special_track.rb
```diff
@@ -165,7 +166,7 @@ def check_connect(_action, ability)
         return if ability.hexes.size < 2
         return if !ability.start_count || ability.start_count < 2 || ability.start_count == ability.count
 
-        paths = ability.hexes.flat_map do |hex_id|
+        paths = ability.laid_hexes.flat_map do |hex_id|
```
michaeljb, line 169: With a change to one of the base steps, I was a little bit nervous but validated against the full DB and it's all good. 👍

michaeljb (approved):

PR #5233, opened 2021-05-04
Add option to manually set player order

daniel-sousa-me:
I don't think any normal game will care about this feature, but it is pretty useful for tournaments.

image

Only the owner can move players around and only if that option was selected when creating the game

I changed the visual of the user list a bit, because it was getting confusing with too many buttons. Now instead of commas, it's using a left border in the same way of the entities list.
This is how the player list looks like when you don't have permission to change the order:

image

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -189,37 +189,10 @@ module View
         },
       }
 
-      p_elm = players.map.with_index do |player, index|
-        short_name = player['name'].truncate
-        if owner? && new? && player['id'] != @user['id']
-          button_props = {
-            on: { click: -> { store(:confirm_kick, [@gdata['id'], player['id']]) } },
-            attrs: { title: "Kick #{player['name']}!" },
-            style: {
-              padding: '0.1rem 0.3rem',
-              margin: '0 0.3rem 0.1rem 0.3rem',
-            },
-          }
-
-          elm = if @confirm_kick != [@gdata['id'], player['id']]
-                  h(:button, button_props, "#{short_name} ❌")
-                else
-                  button_props['on'] = { click: -> { kick(@gdata, player) } }
-                  h(:button, button_props, 'Kick! ❌')
-                end
-
-        else
-          player_props = { attrs: { title: player['name'].to_s } }
-
-          elm = h(:span, [
-            h(acting?(player) ? :em : :span, player_props, short_name),
-            index == (players.size - 1) || owner? && new? ? '' : ', ',
-          ])
-        end
-        elm
-      end
+      p_elm = players.map.with_index { |player, index| render_player(player, index) }
 
       children = [h(:div, [h(:strong, 'Id: '), @gdata['id'].to_s])]
+      children << h(:div, [h(:strong, 'Player order set manually')]) if @gdata.dig('settings', 'player_order')
       children << h(:div, [h(:i, 'Invite only game')]) if @gdata['status'] == 'new' && @gdata.dig('settings',
                                                                                                   'unlisted')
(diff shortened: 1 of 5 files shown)


assets/app/view/game_card.rb
```diff
@@ -191,35 +191,62 @@ def render_body
...
         short_name = player['name'].truncate
+        player_props = { attrs: { title: player['name'].to_s } }
+        button_style = {
+          padding: '0.1rem 0.3rem',
+          margin: '0 0.3rem 0.1rem 0.3rem',
+        }
+
+        p_content = []
+
+        if index.positive? && (owner? && new? && !@gdata.dig('settings', 'player_order').nil?)
```
tobymao, line 202: you don't need the nil

assets/app/view/game_card.rb
```diff
@@ -191,35 +191,62 @@ def render_body
...
+            p_content << h(:button, button_props, '❌')
+          else
+            button_props['on'] = { click: -> { kick(@gdata, player) } }
+            p_content << h(:button, button_props, 'Kick! ❌')
+          end
+        end
 
-        else
-          player_props = { attrs: { title: player['name'].to_s } }
+        if index != players.size - 1 && owner? && new? && !@gdata.dig('settings', 'player_order').nil?
```
tobymao, line 229: maybe make this a helper? again, get rid of the nil

routes/game.rb
```diff
@@ -26,6 +26,9 @@ class Api
             halt(400, 'Cannot join because game has started') unless game.status == 'new'
 
             GameUser.create(game: game, user: user)
+            game.settings['player_order'] << user.id unless game.settings['player_order'].nil?
```
tobymao, line 29: if game.settinsg...

tobymao:
if this is only used for you / tournaments, i think this is too much. can we instead just give you the ability to set the seed? i think that'll be a much easier change

daniel-sousa-me:
The seed is a bit messy, because I'd need to be randomly throwing seeds until I get the order I want (1846 has a variable setup).

But I can strip all the UI and only keep the line on the ordered players model and the api call

daniel-sousa-me:
I'll fix up this version and create a separate PR with the stripped version and you see what you prefer

assets/app/view/game_card.rb
daniel-sousa-me: I don't even need the dig. I just thought it would make things more clear

dfannius:
> if this is only used for you / tournaments, i think this is too much. can we instead just give you the ability to set the seed? i think that'll be a much easier change

I haven't looked at the code to assess how risky/complicated of a change this is, but this feature is vital for tournaments and can be useful outside of them too (although of course I don't know how often it will actually be used in practice), so it would be nice to find a way to incorporate some form of it. Continually resetting the random seed and starting over again until we get a result that occurs with 1/4! probability feels pretty icky and wouldn't scale if we wanted to make more random things settable by hand in the future (say, to match a face-to-face game).

daniel-sousa-me:
I can see some instances where people would want this. That's why I did the interface at all, because I'm not going to use it.

Some details that may be important: When creating a game, there's a new option for manually setting player order. If it is not selected, none of this will show up and the API will refuse to manually set the order. If it is chosen, then all users see the message on the card saying that the order will be set manually.

tobymao:
the seed guesser is only on @daniel-sousa-me side of things. you can write a script to find the correct seed for a particular order and then create a game with the correct seed

daniel-sousa-me:
I'd be grateful if my side didn't need to be messy, but I'll take what I can get. Did you look at #5263? It's only 2 more lines then a version where the seed can be changed.

tobymao:
didn't see that one thanks, let's use that one

PR #5234, opened 2021-05-04
1870: Any share price is a valid par price for reissuing (fixes #5060)

### LINKED ISSUE #5060: 1870
Two issues: 1) when two tokens occupy the same space, and the top company has a share sold, but doesn't move because it's on the ledge, it swaps places with the bottom token. Only when a token moves into a shared spade does it go to the bottom of the stack

2) when reissuing a share with a CMV of $300, the new par price should be $225. The site has the share reissued at $200.

daniel-sousa-me:
May archive some games


jenf (approved):
looks good, but needs merge

PR #5235, opened 2021-05-04
1870: Fix stock movements (fixes #5067)

### LINKED ISSUE #5067: [1870] Sold out companies in top row should move right and down
https://18xx.games/game/37666?action=502

See ATSF

> Up one row if all shares are owned by players or the company at
the end of a stock round. If the token is at the top of the market, it moves,
instead, one row downward and one column to the right.

daniel-sousa-me:
Preserve order when hitting the soft ledge (don't go down to come back up)
When hitting the ceiling, move right and down instead

### DIFF
--- a/lib/engine/game/g_1870/stock_market.rb
+++ b/lib/engine/game/g_1870/stock_market.rb
@@ -6,6 +6,15 @@ module Engine
   module Game
     module G1870
       class StockMarket < Engine::StockMarket
+        def move_up(corporation)
+          r, c = corporation.share_price.coordinates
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1870/stock_market.rb
```diff
@@ -6,6 +6,28 @@ module Engine
   module Game
     module G1870
       class StockMarket < Engine::StockMarket
+        def move_down(corporation, twice = false)
```
jenf, line 9: The twice logic shouldn't be here

PR #5238, opened 2021-05-04
18Ireland: Merger connectivity isn't blocked by tokens

### LINKED ISSUE #5218: 18Ireland merger round—valid mergers ignoring tokens not allowed
**What's your Game ID?**

https://18xx.games/game/39912?action=338

**How to Reproduce the Bug?**

Excalabur should be able to merge GNWR and DDR, as they are connected, even though the DER token intervenes.  (Intervening tokens are ignored for mergers.)

jenf:
fixes #5218

### DIFF
--- a/lib/engine/game/g_18_ireland/step/merge.rb
+++ b/lib/engine/game/g_18_ireland/step/merge.rb
@@ -204,6 +204,7 @@ module Engine
 
             @round.merging = nil
             @round.vote_outcome = nil
+            @corp_connectivity.clear
             pass!
           end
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_ireland/step/merge.rb
```diff
@@ -248,13 +249,39 @@ def mergeable_type(entity)
...
+                visited[path] = true
+
+                path.nodes.each do |p_node|
+                  next unless p_node.city?
+                  next if visited[p_node]
+
+                  visited[p_node] = true
+
+                  a = p_node.tokens.select { |t| t&.corporation&.type == :minor }.map(&:corporation)
+                  connected.concat(a)
```
tobymao, line 270: this looks like it's gonna be expensive

p_node.tokens.each do |...|
connected << corp

lib/engine/game/g_18_ireland/step/merge.rb
```diff
@@ -248,13 +249,39 @@ def mergeable_type(entity)
             end
           end
 
+          def corporation_connected?(c1, c2)
+            # Are corporations connected (blocking by tokens doesn't matter)
+            unless @corp_connectivity[c1]
+              new_home = c1.tokens.first.city
+              connected = []
+              visited = {}
+              new_home.walk(skip_track: :narrow, tile_type: @game.class::TILE_TYPE) do |path, _, _|
```
tobymao, line 258: is this expensive? do you think we should add a non blocking graph?

jenf: in terms of being expensive, it is done in the UI (and when not loading), and it's only done once per each unique network, which is probably as cheap as it can be

tobymao (approved):

PR #5242, opened 2021-05-05
[1822] Minors shouldnt be able to run E trains

### LINKED ISSUE #5079: 1822 - Minors shouldn't be able to run E trains
Link: https://18xx.games/game/34665?action=908

Per rule 5.11.24 in the AAG rulebook, minor railroads should not be able to run an E-train. In the link above, if M12 tries to run P35-O36, the displayed route value is £40.

magnusfalt:
- If a minor only have an E-train, skip the run routes step.

fixes #5079

This will probably break a game or two if there was a minor who owned a E-Train.

### DIFF
--- a/lib/engine/game/g_1822/step/route.rb
+++ b/lib/engine/game/g_1822/step/route.rb
@@ -9,6 +9,7 @@ module Engine
         class Route < Engine::Step::Route
           def actions(entity)
             return [] if !entity.operator? || @game.route_trains(entity).empty? || !@game.can_run_route?(entity)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1822/step/route.rb
```diff
@@ -56,6 +57,10 @@ def detach_pullman
             @pullman_train = nil
           end
 
+          def only_e_train?(entity)
+            @game.route_trains(entity).reject { |t| t.name == @game.class::E_TRAIN }.empty?
```
jenf, line 61: reject.empty? is probably slower than all?

jenf (approved):

PR #5246, opened 2021-05-05
[1822] P21-HSBC is not working as described

### LINKED ISSUE #5241: [1822] P21-HSBC is not working as described.
In the implementation description (https://github.com/tobymao/18xx/pull/4299), P21 is described as follows:
- P21 Humber Suspension Bridge Company. 2 free tile lay on the Hull / Grimby hexes. You can lay both tiles even if its a minor. And one of the tile lay may be an upgrade.

The [linked game](https://18xx.games/game/40231?action=368) is at a point where I have been trying to use P21.
Major company owns P21 and has token in Hull.
Hull has a yellow tile and Grimsby has no tile.
Using the ability button, I can build a green tile in Hull for free.
But when I try to use the ability button again to build a yellow in Grimsby, I receive a "Cannot lay a yellow now" error.
Pictures follow, showing selected ability, un-greyed map, tile offering, and error message.

Screen Shot 2021-05-04 at 20 18 54
Screen Shot 2021-05-04 at 20 19 11
Screen Shot 2021-05-04 at 20 19 41
Screen Shot 2021-05-04 at 20 20 48

I have previously observed (but not reported) that a minor company is permitted to place only one tile with this card.

magnusfalt:
- Fixed so you can upgrade and lay yellow tile if you use P21.

fixes #5241

This shouldnt break any games.

### DIFF
--- a/lib/engine/game/g_1822/step/tracker.rb
+++ b/lib/engine/game/g_1822/step/tracker.rb
@@ -25,21 +25,10 @@ module Engine
           super
         end
 
-        def get_tile_lay(entity)
-          corporation = if entity.id == @game.class::COMPANY_HSBC
-                          entity
-                        else
-                          entity.company? ? entity.owner : entity
-                        end
-          action = @game.tile_lays(corporation)[@round.num_laid_track]&.clone
-          return unless action
+        def get_tile_lay_corporation(entity)
+          return entity if entity.id == @game.class::COMPANY_HSBC
 
-          action[:lay] = !@round.upgraded_track if action[:lay] == :not_if_upgraded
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1822/step/tracker.rb
```diff
@@ -25,6 +25,23 @@ def check_track_restrictions!(entity, old_tile, new_tile)
           super
         end
 
+        def get_tile_lay(entity)
```
jenf, line 28: Might be best modifying the base code to allow a get tile lay without the entity.company? stuff

magnusfalt: Yes good call.

jenf (approved):

PR #5252, opened 2021-05-05
1870: Fix error on closing corporation (fixes #4987)

### LINKED ISSUE #4987: Unable to close company in 1870
I have a company on the verge of closing in 1870, but when I'm taking it there I get an error message. The game is 

https://18xx.games/game/35230

The stock value is 10, so it would go left and close. I've never played 1870 before, but the stock market says "Corporation closes", and I get the message:

Undefined message pass for nil

daniel-sousa-me:
The pass! on PriceProtection may be called during the Stock Round
while on the turn closed corporation and it inherits BuySellParShares

force_next_entity! tries to pass! every step, but we still want to run
through the CheckConnection

### DIFF
--- a/lib/engine/game/g_1870/step/check_connection.rb
+++ b/lib/engine/game/g_1870/step/check_connection.rb
@@ -53,13 +53,17 @@ module Engine
             ACTIONS
           end
 
+          # This step should not be passed
+          # This prevents the force_next_entity! from trying
+          def pass!; end
+
           def process_destination_connection(action)
             action.corporations.each do |corporation|
               @game.log << "-- #{corporation.name} can connect to its destination --"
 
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1870/step/check_connection.rb
```diff
@@ -53,13 +53,17 @@ def actions(_entity)
             ACTIONS
           end
 
+          # This step should not be passed
```
jenf, line 56: it'd be better to override operating.rb to change the behaviour to skip force next entity on check_connection.

daniel-sousa-me: That requires copy-pasting the entire function, which is very likely that someone will want to change in the future

tobymao (approved):

PR #5259, opened 2021-05-05
1860, 1862: highlight home token

### LINKED ISSUE #4946: 1860 QoL - Differ home tokens on map
In 1860, each company must include its home token for some part of its runs. So it would be helpful if the home tokens were visibly identifiable on the map. Something clean and simple would go a long way.

roseundy:
Both 1860 and 1862 have the requirement that at least one route must contain the home token. It is helpful to highlight this token on the map.

Fixes #4946 

Examples:
1860_highlight

token_hightlight

token_highight_players

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -59,13 +59,10 @@ module View
             fill: color,
           }
 
-          if @game&.highlight_token?(@token)
+          if (highlight = @game&.highlight_token?(@token))
             radius -= 3
-            token_attrs.merge!(
-              stroke: 'white',
-              'stroke-width': '3px',
-              filter: 'drop-shadow(8px 8px 2px #444)',
-            )
+            token_attrs[:stroke] = 'white'
+            token_attrs[:'stroke-width'] = '3px'
           end
 
           children = [h(:circle, attrs: token_attrs)]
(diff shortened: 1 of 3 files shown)


daniel-sousa-me:
There's a shadow also in 1870 (in this case for destination tokens), which appears 10 lines bellow. Probably they could use the same code and visual.

assets/app/view/game/part/city_slot.rb
```diff
@@ -54,7 +54,21 @@ def render_part
...
 
-          children = [h(:circle, attrs: { r: @radius, fill: color })]
+          token_attrs = {
+            r: @radius,
+            fill: color,
+          }
+
+          if @game&.highlight_token?(@token)
+            radius -= 3
+            token_attrs.merge!({
```
tobymao, line 64: i don't like the format of this

merge!(
  x: '',
  y'',
)

roseundy: OK

roseundy:
> There's a shadow also in 1870 (in this case for destination tokens), which appears 10 lines bellow. Probably they could use the same code and visual.

Perhaps, but that code applies to the entire group, rather than just the token.

roseundy:
I combined my highlight shadow effect with the extra token code to deal with the (so far, unneeded) case where both are attempted.

daniel-sousa-me:
For 1870 I don't really care the details of how it looks. It just should be somehow different than a normal token. I would just merge them.

daniel-sousa-me:
So, just making 1870/game.rb use the highlight token and not having anything special for the extra token

daniel-sousa-me:
But don't let this bikesheding prevent you from merging the commit. This is not important at all

roseundy:
OK, latest commit now has highlight higher priority than extra_token, so 1870 can switch at its leisure.

PR #5263, opened 2021-05-06
Add API call to manually set player order

daniel-sousa-me:
Same as #5233, but without any UI, just the API calls.

### DIFF
--- a/routes/game.rb
+++ b/routes/game.rb
@@ -139,6 +139,14 @@ class Api
           # POST '/api/game/<game_id>/kick
           r.is 'kick' do
             game.remove_player(r.params['id'])
+
+            game.to_h
+          end
+
+          r.is 'player_order' do
+            game.settings['player_order'] = r.params['player_order']
+            game.save
+
             game.to_h
           end
         end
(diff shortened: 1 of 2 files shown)


routes/game.rb
```diff
@@ -162,6 +171,7 @@ class Api
             max_players: r['max_players'],
             settings: {
               seed: Random.new_seed % 2**31,
+              player_order: r['player_order'] ? [] : nil,
```
tobymao, line 174: player_order: r['player_order'] don't need the nil

routes/game.rb
```diff
@@ -139,6 +139,15 @@ class Api
           # POST '/api/game/<game_id>/kick
           r.is 'kick' do
             game.remove_player(r.params['id'])
+
+            game.to_h
+          end
+
+          halt(400, 'You are not allowed to change the player order in this game') unless game.settings['player_order']
```
tobymao, line 146: remove this halt

daniel-sousa-me: Then I'll also remove the line from the game card and the option from create_game, because this way any game can have the order manually changed.

tobymao (approved):

PR #5270, opened 2021-05-07
update 1829

volker18xxdev:
100% Ownership, upgrade towns to cities, insert draft-round without Bids


volker18xxdev:
this offense:
Offenses:

queue.rb:27:18: C: [Correctable] Performance/RegexpMatch: Use match? instead of =~ when MatchData is not used.
    next true if user.email =~ /@(msn hotmail outlook live passport)/
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
doesn't come not from my files.

Zwergenpunk:
The problem with the rubocop should be fixed. Just rebase and push again :)

volker18xxdev:
done

lib/engine/game/g_1829/Step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Game
+    module G1829
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParShares
+          def actions(entity)
+            return [] unless entity == current_entity
+
+            actions = super
+
+            actions << 'pass' unless actions.include?('pass') || actions.empty?
```
tobymao, line 15: don't use unless with multiple conditions

lib/engine/game/g_1829/Step/close_company.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1829
+      module Step
+        class CloseCompanyVoluntary < Engine::Step::BuyTrain
+          def process_close_company(_action)
+            @log << "#{company.name} is closed."
```
tobymao, line 9: why do you need 2 logs? just have one

tobymao:
the Step/draft should not be capitalized

volker18xxdev:
please give a hint, which file? (i'll copied the draft file from 18CZ, so i don't know, where is the mistake.)

PR #5272, opened 2021-05-07
1870: Fix bug that broke game rendering at the end of the auction.

daniel-sousa-me:
Moved sell_queue from round to game, because it now is checked at all times (and was already available almost anywhere anyway)


daniel-sousa-me:
Please merge asap, because most games will break.

The check is failing because of bc14ca69a2f03edb0beb1f8e1faf1ba9d3571273 (rubocop issue)

PR #5277, opened 2021-05-07
1862: handle missed obligations

roseundy:
Add player penalty (with possible user interaction) for missed Obligations.
Also includes several bug fixes found with a multiplayer full test game.

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -630,7 +630,7 @@ module Engine
         def enforce_obligations
           @players.each do |player|
             remaining_fine = 0
-            player.companies.each do |company|
+            player.companies.dup.each do |company|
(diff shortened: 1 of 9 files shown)


lib/engine/game/g_1862/game.rb
```diff
@@ -623,6 +626,94 @@ def add_obligation(entity, corporation)
           @chartered[corporation] = true
         end
 
+        # called at end of stock round
+        def enforce_obligations
+          @players.each do |player|
+            remaining_fine = 0
+            player.companies.each do |company|
```
tobymao, line 633: dup since you're mutating this?

roseundy: Yep. I missed this because my test case only had one missed obligation. Two missed obligations on a player would be quite the dumpster fire, but theoretically possible.

PR #5309, opened 2021-05-10
Fix token locations not showing.

### LINKED ISSUE #5127: Token locations in entity displays are showing costs paid instead of locations
There seems to be something weird going on with the token locations in the entity displays.  I've only noticed this in the last couple days. 

Here's a screen shot from my current 1846 game (https://18xx.games/game/36387#entities), but I've also peeked around at a few other games of different titles, and it seems to be pervasive.

What's up with the token locations circled in red, which seem to be showing the cost paid for them instead of their current board locations?  :)

image

jenf:
Hexes were not being set correctly as the newly placed tile did not
have the hex assigned.

fixes #5127

### DIFF
--- a/lib/engine/hex.rb
+++ b/lib/engine/hex.rb
@@ -151,6 +151,9 @@ module Engine
         new_city.groups = old_city.groups
       end
 
+      @tile.hex = nil
+      tile.hex = self
+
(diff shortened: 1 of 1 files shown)


roseundy (approved):
LGTM

PR #5334, opened 2021-05-13
Make prototype games more obviously prototypes

### LINKED ISSUE #5330: Separate prototypes more cleanly
Presently we have prototypes stuck in alpha, as the maps and rules may change at any point.

We should add a new flag which says prototype.

'Production' prototypes would move to after the regular 'production' games, with an appropriately notice saying the rules may change.

In alpha & beta it wouldn't be separated but show similar to how 18SJ shows '18SJ (Sweden and Norway - Prototype)'

jenf:
Provides a flag, separates production prototype games from production non-prototype games.

Makes 1849: K2S production

fixes #5330


roseundy:
Would 18CO also be a prototype? I guess the designer needs to clarify. @ryandriskel

jenf:
Possibly, but that can be done retroactively.

scottredracecar:
I think he has said he does not consider it to be a prototype.

tobymao (approved):

PR #5335, opened 2021-05-13
18Ireland: Disallow repeating the same merger proposal

### LINKED ISSUE #5223: 18Ireland: Can re-propose previously proposed proposals
If a proposed merger is reject-
ed, an identical proposal may not be made again by any player
during that merger round.

jenf:
fixes #5223


roseundy (approved):
LGTM

PR #5390, opened 2021-05-18
[1840] Added Train count per Player and Player Order choosing for first SR

Zwergenpunk:
Also Adds selling of companies.

This is applied with the choice since it needs to be possible at any time and that is a real simple solution


tobymao:
conflicts

Zwergenpunk:
Mhm not quite sure why this is failiing =( have to look at it later

Zwergenpunk:
Added selling of companies

tobymao (approved):

PR #5391, opened 2021-05-18
1856: CGR may not close & may not change price until it owns a permanent train

ventusignis:


lib/engine/game/g_1856/stock_market.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require_relative 'stock_market'
+
+module Engine
+  module Game
+    module G1856
+      class StockMarket < Engine::StockMarket
+        attr_writer :game
+
+        def move(corporation, row, column, force: false)
```
tobymao, line 11: do you need this method? you already overidde dividend and stock

ventusignis: I believe so
This is needed for share selling price changes and stock only handled end-of-round stock price bubbles which I removed as it's effictively duplicated here.
dividend should stay so that the "1 left" and "1 right" messages are reflected correctly in the UI

PR #5398, opened 2021-05-19
[18ireland] corrects IR6 tile count 

philcampeau:
In the current iteration there is only one green BC (IR6) tile. There should be two.


jenf:
@philcampeau 
could you use 'fixes #5389' in the commit message in future as it links the tickets appropriately.

philcampeau:
noted!

PR #5410, opened 2021-05-20
Fix webhooks from discordapp.com

### LINKED ISSUE #5289: [Webhooks] Custom Discord webhooks not recognized.
Attempting to add a discord webhook currently results in the error "webhook_url only supports Slack and Discord, let us know if you want others" A discord webhook should look something like this: https://discordapp.com/api/webhooks/840703942626115634/urgw6SuZkhbu0SdYXeNZZi0w0cgRGScBKNekjBY2yBuxmZRez4_enXRa8w9P3nIz4OjV 
If it is looking for discord.com, it may be getting confused by the "discordapp.com" in the link.

It would be useful to include this link in the wiki page so people know how to create a webhook in discord: https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks

...also currently webhooks send notifications for every action, which I assume is more than intended.

gpanders:
Fixes #5289

### DIFF
--- a/lib/hooks.rb
+++ b/lib/hooks.rb
@@ -14,7 +14,7 @@ module Hooks
 
     req.body =
       case uri.host
-      when 'discord.com'
+      when 'discord.com', 'discordapp.com'
         JSON.generate(
           content: message,
           allowed_mentions: { parse: ['users'] },


Zwergenpunk:
Discordapp.com is the legacy domain of discord.

PR #5416, opened 2021-05-20
1862: change freight route distance calc; add alt logos

### LINKED ISSUE #5379: [1862] Toggling Simple Logos Causes Token Images to not Load
When Simple Logos are turned on, company tokens do not load. 

Simple Logos

roseundy:
Fixes #5379 

After some debate, simplified Freight route hex-distance bonus calculation:

New_HexDist

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -1580,30 +1580,14 @@ module Engine
           false
         end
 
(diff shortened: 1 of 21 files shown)


roseundy:
This shouldn't need archiving, since revenue is pre-validated.

tobymao (approved):

PR #5421, opened 2021-05-21
[2038] Update Initial Auction handling

jason-johnson:
This PR improves the initial auction for 2038 to be closer to how it is supposed to function.  Also made some fixes for the short game variant.


assets/app/view/game/round/auction.rb
```diff
@@ -152,9 +152,10 @@ def render_company_actions(company)
                       size: @current_entity.cash.to_s.size,
                     })
 
+          bid_str = @step.respond_to?(:bid_str) ? @step.bid_str(company) : 'Place Bid'
```
tobymao, line 155: shouldn't this be inside the if statement?

lib/engine/game/g_2038/game.rb
```diff
@@ -653,10 +654,50 @@ def game_hexes
...
+                new_stock_round
+              end
+            when init_round.class
+              init_round_finished
+              reorder_players
+              new_stock_round
+            end
+        end
+
+        HOME_TOKEN_TIMING = :never
```
tobymao, line 699: put this with the other constants at the top

lib/engine/game/g_2038/step/waterfall_auction.rb
```diff
@@ -0,0 +1,90 @@
...
+          end
+
+          def placement_bid(bid)
+            @purchasing_first_minor = bid.company && bid.company == @companies.first
+            super
+            @purchasing_first_minor = false
+          end
+
+          def minor?(company)
+            @game.minors.find { |m| m.id == company.id }
```
tobymao, line 42: makes more sense to do any? since you just expect a boolean

jason-johnson: Ok, I was avoiding any? because I was under the impression from the wiki that Opal didn't produce good code for it but checking just now it only mentions the GC which is probably the same with this statement anyway.  Will change.

assets/app/view/game/round/auction.rb
jason-johnson: I copied this line from line 142 from the same file as I needed this same functionality for bidding.  I've also tested and this is working correctly for my game and it didn't change anything for the games I tested (nor would I expect it to since none of them will implement #bid_str right now).

tobymao: it's only used in the context of the if statement on line 157, you should put it in there

PR #5425, opened 2021-05-21
[1846] Implement 2E Private Companies, add 1E Optional Rule

### LINKED ISSUE #4732: [1846] Add new private companies from 2nd printing
With the second printing coming out the designer has also included two new privates. The deliveries are scheduled for April so I imagine the privates are finalized.

Details on them are not publicly available as of yet, but if the designer does give the specifics would the original implementation agreement cover the new privates?

Some details here:

https://boardgamegeek.com/thread/2589981/1846-gmt-reprint-2nd-edition-now-printer

michaeljb:
Boomtown - +$20 marker on Cincinnati

Little Miami - if Cincinnati and Dayton are not yet connected, may lay/upgrade
an extra tile on one or both of them for free; each tile laid must contribute
track to a new connection between the cities; per further details provided by
Tom via email (and presumably spelled out in the rulebook), any blocking tokens
between Cincinnati and Dayton are ignored

To avoid graph checking on every action, Little Miami's ability is *not*
automatically removed when a connection is completed by another entity laying
track. If the connection already exists, Little Miami's ability will simply
raise an error when its first track lay is made.

Rather than determining which tile lays + rotations could possibly create the
Cincinnati-Dayton connection, all rotations are permitted, and connections
checked after Little Miami's second action (which could be a pass if only one
tile lay is needed to complete the connection).

Little Miami is added to the Orange Group, Boomtown to the Blue Group; during
initial setup each group has one additional company removed (so that the number
of companies in play during the game remains the same).

At Tom's request, including both of these private companies is the default
behavior, with a new optional rule to only use the companies from the 1st
Edition.

[Resolves #4732]

-----

Fixtures were updated to use the first edition optional rule with this
script (which also happens to minify them):

```

require 'json'

Dir.glob('spec/fixtures/1846*/*.json').each do |fixture_path|
  game_data = JSON.parse(File.read(fixture_path))

  game_data['settings'] ||= {}
  game_data['settings']['optional_rules'] ||= []
  unless game_data['settings']['optional_rules'].include?('first_ed')
    game_data['settings']['optional_rules'] << 'first_ed'
  end

  File.write(fixture_path, game_data.to_json)
end
```

-----

Migration script to add the first edition optional rule to all existing games in
the database (took about 10 seconds in my local trial run):

```
require_relative 'models'

Dir['./models/**/*.rb'].sort.each { |file| require file }
Sequel.extension :pg_json_ops
require './lib/engine'

games = Game.where(title: ['1846', '1846 2p Variant']).all
games.each do |game|
  game.settings['optional_rules'] ||= []
  game.settings['optional_rules'] << 'first_ed'
  game.save
end
```

-----

JS script which users can paste into their console to update their hotseat
games (will add this to the wiki):

```
function add_1e_option(game_id) {
  const game_data = JSON.parse(localStorage[game_id]);

  if (!['1846', '1846 2p Variant'].includes(game_data.title)) {
    return;
  }

  if (!game_data.settings.optional_rules) {
    game_data.settings.optional_rules = [];
  }

  if (!game_data.settings.optional_rules.includes('first_ed')) {
    game_data.settings.optional_rules.push('first_ed');

    localStorage[game_id] = JSON.stringify(game_data);
    console.log('Added optional rule "1st Edition Private Companies" to game: ' + game_id);
  }
}

// included so that if the user runs this script after having already started a 2E hotseat game,
// that 2E game can be fixed
function remove_1e_option(game_id) {
  const game_data = JSON.parse(localStorage[game_id]);

  if (!['1846', '1846 2p Variant'].includes(game_data.title)) {
    return;
  }

  if (game_data.settings.optional_rules.includes('first_ed')) {
    const filter = function (value, index array) {
      return value != 'first_ed';
    };

    game_data.settings.optional_rules = game_data.settings.optional_rules.filter(filter)

    localStorage[game_id] = JSON.stringify(game_data);
    console.log('Removed optional rule "1st Edition Private Companies" from game: ' + game_id);
  }
}

function fix_all_games() {
  Object.entries(localStorage).forEach(function (entry) {
    const game_id = entry[0];
    if (game_id.match(/^hs_[a-z]+_[0-9]+$/) == null) {
      return;
    }

    add_1e_option(game_id);
  });
}

fix_all_games();
```


michaeljb:
Here's my first pass at updating the wiki for this release -https://github.com/tobymao/18xx/wiki/1846/00c6bda37ff0406694e0430ac4f0af5f83448d52 

The changes have been reverted so they do not appear on https://github.com/tobymao/18xx/wiki/1846 at the moment.

michaeljb:
Should we add an "LM" icon to H12 and G13, similar to the "LSL" icon on D14 and D17?

lib/engine/game/g_1846/game.rb
```diff
@@ -741,6 +783,111 @@ def train_buying_power(entity)
...
+          @little_miami_hexes ||= little_miami.all_abilities.first.hexes
+        end
+
+        def little_miami_router
+          @little_miami_router ||=
+            Engine::Corporation.new(name: 'LM Corp', sym: 'LM Corp', tokens: [], coordinates: little_miami_hexes.first)
+        end
+
+        def compute_little_miami_graph
+          graph = Graph.new(self, no_blocking: true, home_as_token: true)
```
tobymao, line 801: can we add !loading checks in all these graph places?

michaeljb: 983bf5ca5

michaeljb:
> Should we add an "LM" icon to H12 and G13, similar to the "LSL" icon on D14 and D17?

Added icon in 3d7c7b0f8, added logic to remove it when used/closed/removed during setup, just like Lake Shore Line, in 39e48e82a

michaeljb:
Whoops, was cleaning up some other git branches, did not mean to delete this one.

tobymao (approved):

michaeljb:
Will merge and release this on Tuesday, per an email from Tom.

Issue #5432, opened 2021-05-22
[1862] lower left half of G14 doesn’t count as port for freight 


philcampeau:
![38DFDE25-B0C8-46A7-A419-E768E35B0DF9](https://user-images.githubusercontent.com/26125362/119229365-ad239700-bae5-11eb-9069-d7ff40afd2f3.jpeg)
![EF62E4B5-B27D-4E52-99D2-7B8E7D9AA320](https://user-images.githubusercontent.com/26125362/119229366-adbc2d80-bae5-11eb-94c5-d84f9cd91064.jpeg)

In the first pic above, a freight train runs for 130, but in the second it runs for 140. We suspect it’s not adding the port bonus in the first image. 

game is: https://18xx.games/game/42532

roseundy:
The problem was the port just had the wrong revenues defined for it in the code.

PR #5437, opened 2021-05-22
[1862] Correct value of Holland offboard.

KevinR1000:
G14 and F15 had different values for Holland offboard.
See https://github.com/tobymao/18xx/issues/5432


roseundy:
I would appreciate a chance to fix my own bugs, please. Having the code change on me while I am fixing multiple bugs is a pain.

PR #5455, opened 2021-05-24
1862: add refinance as an option; add missing stock value

### LINKED ISSUE #5449: 1862 Refinancing bug: refinance not available when shares in treasury.
**What's your Game ID?**

https://18xx.games/game/43189?action=158

**How to Reproduce the Bug?**

A company has the option to *choose* between refinancing and selling treasury stock when it is unable to afford a train.  This appears to be implemented incorrectly: The FDR is not given the option to refinance, and is only prompted to sell treasury shares.  Checking the implementation checklist, this appears to be deliberate but incorrect rules implementation.  The correct implementation would expose both share sales and refinancing to the user when funds were short for a new train.

### LINKED ISSUE #5453: [1862] Wrong stock value
495 should be 475

roseundy:
Fixes #5449 
Fixes #5453 

Modified UI::Convert to allow specifying text on button

Unlikely this will require archiving games. If it invalidates a lot of games, then something is wrong.

### DIFF
--- a/assets/app/view/game/convert.rb
+++ b/assets/app/view/game/convert.rb
@@ -19,7 +19,9 @@ module View
           on: { click: click },
         }
 
-        h(:div, [h('button', props, @game.round.active_step.description)])
+        step = @game.round.active_step
+        text = step.respond_to?(:convert_text) ? step.convert_text(@game.current_entity) : step.description
+        h(:div, [h('button', props, text)])
       end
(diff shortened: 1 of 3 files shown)


tobymao:
it's invalidating basically every single game

tobymao:
ah maybe because i had strict: true, oopsies

PR #5462, opened 2021-05-24
18scan early build

philcampeau:
Skeleton framework for 18Scan.


tobymao:
did you code this?

philcampeau:
Yes. Still early in the process.

tobymao:
it looks like you just copied and pasted stuff from 1860, i'd rather things not be just copied and pasted

philcampeau:
The only element I copied from 1860 was the opening auction code, since it works the same way. It still needs to be tweaked too.

tobymao:
code should never be copied, if it's truly the same, it should be modularized, perhaps @roseundy can help you, or if you want to just submit this PR as the config I'll accept that.

philcampeau:
Sure I can do it that way if it’s better.

PR #5506, opened 2021-05-28
1856: CGR can only buy permanent trains

ventusignis:


michaeljb:
Could this be done by implementing `G1856::Step::BuyTrain#buyable_trains` instead of adding a new ability? Something like this:

```
def buyable_trains(entity)
  trains = super
  trains.reject!(&:rusts_on) if entity == @game.national
  trains
end
```

ventusignis:
Yeah that's a much simpler way to get the same effect

michaeljb (approved):

PR #5596, opened 2021-06-03
1856: Fix CGR losing tokens in the case of duplicate tokens in city (toronto, OO)

### LINKED ISSUE #5571: 1856: CGR loses tokens if duplicate tokens in a hex and max tokens
In this game the CGR had tokens in each OO in Kitchener and after everything settled it only had 9 tokens
https://18xx.games/game/44516?action=659

ventusignis:
Fixes #5571

I think this doesn't archive any games because the broken actions are just that token actions were skipped and there's a default migration for that?

### DIFF
--- a/lib/engine/game/g_1856/game.rb
+++ b/lib/engine/game/g_1856/game.rb
@@ -1666,9 +1666,14 @@ module Engine
 
           # There won't be any duplicates (OO or NY) that need deduplicating where a home city is involved
           # because that case is automatically resolved above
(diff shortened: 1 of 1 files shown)


jenf:
unfortunately will be archive

jenf (approved):

ventusignis:
I think we'll have to live with these early alpha games getting the ultimate form of nationalization then :(

tobymao (approved):

PR #5605, opened 2021-06-05
1888: Maps, tiles, entities, etc

roseundy:
First PR for 1888

Common code changes:
* UI::LargeIcons - check game before use (to allow display of map)
* UI::Upgrade - new terrain type: wall


assets/app/view/game/part/large_icons.rb
```diff
@@ -124,7 +124,7 @@ def render_part
             decor = h(:circle, attrs: { r: LARGE_RADIUS, fill: 'white', cx: dx, cy: dy })
 
             show_player_colors = setting_for(:show_player_colors, @game)
-            if (decoration = @game.decorate_marker(icon))
+            if (decoration = @game&.decorate_marker(icon))
```
tobymao, line 127: why would game be nil

roseundy: It's nil when using a map route directly: 18xx.games/map/1888

PR #5619, opened 2021-06-06
1862: highlight in yellow trains a corporation doesn't have permits for at train buy

ventusignis:
A
image

B
image

C
image

D (multi permit)
image

@roseundy Does this look reasonable?

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -10,6 +10,7 @@ module View
       include Actionable
       include EmergencyMoney
       include AlternateCorporations
+      include Lib::Settings
       needs :show_other_players, default: nil, store: true
       needs :corporation, default: nil
       needs :active_shell, default: nil, store: true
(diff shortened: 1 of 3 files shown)


wheresvic:
@ventusignis this is pretty cool - may I make a suggestion - can we invert it and highlight the trains you have permits for in green?

Or change from yellow to red. I think those colors make more sense (at least from a general web perspective where success is associated with green and failure with red).

ventusignis:
@wheresvic I can't really make it so the trains you do have permission for are in green because the UI for buying trains is shared across games and having trains become green in all other games would become useless noise. I also chose yellow instead of red because it *is* possible to buy unlicensed trains

assets/app/view/game/buy_trains.rb
```diff
@@ -229,12 +230,17 @@ def from_depot(depot_trains)
                 warranties: warranties,
               ))
             end
-
+            train_props = { style: {} }
+            unless @game.licensed_to_operate?(corporation, train, name)
```
roseundy, line 234: I would prefer a more generic name for this method that is not tied to the 1862 concept of licenses.

ventusignis: I've changed it to able_to_operate which should be generic enough

tobymao (approved):

PR #5623, opened 2021-06-06
[1840] Fixed some bugs, typos and added some infos

### LINKED ISSUE #5606: [1840] Undo and/or Restart Last Turn  fails;  game iD 44873
1840_20210605-122522_Chrome

### LINKED ISSUE #5615: [1840] List of several bugs/suggestions
Bug: Restart and undo function has unintended effects on the stadtbahn tiles. Laying a stadtbahn tile and then undoing that action allows you to lay a stadtbah tile in the hex that was previously connected to the "undone" tile lay. Restarting the turn during the stadtbah tilelay step also allows your first tilelay to be a stadtbahn tile (this should not be the case according to rules 1X.4 - specifying this first tile needs to connect to your token). Undone stadtbahn tilelays also appear to no longer being placeable after a restart, since the game no longer recognizes them to be the next in the chain. This extends to future turns as well, bugging the ordering of the stadtbahn connections for the rest of the game.

Bug: Private companies pay no dividends in CR1/CR2 - Edit: seems they never pay dividends?

Bug/suggestion: The red and purple bonus tilelays can only be laid if the operating line is connected to any of those possible tiles. While the rules are not very specific, both the yellow bonus tile and the upgrade to green bonus action specify that they follow the normal track laying rules of IX.4 or IX.5. It goes to reason these tiles should be placeable even without connectivity, because all the bonus actions for red/purple are very close to the starting stations of the stadtbahns. Rarely will these tiles be connected this early in the game, and having all these tiles unplaced makes the game a lot less interesting since a large portion of the board becomes unpassable very early on.

Suggestion: The info tab should probably specify when phases change. It now does not tell you when the phase transitions and green tiles become available, neither under gamephases or on the timeline. (possible fix being to color-code the timeline as is done on the physical board?)

Bug: After a company buys a private, and it is consecutively sold to the bank, the money should go to the company. Currently, it still goes to the player, allowing the player to get money for it twice.

Bug(?): Tile L30b currently provides no revenue for non-stadtbahn lines. Should the village/dot in the centre not count as a halt and provide 30 revenue?

### LINKED ISSUE #5617: [1840] gameID 44985 route: reenter hex is not possible & rev calc
a) reenter Hex

1840a 2021-06-06 153952
1840b  2021-06-06 154058
1840c2021-06-06 154142

b) revenue calculation  Nordbahnhof = 30 ok  |  Bahnhof Hauptzollamt = zero??

1840d 2021-06-06 160832
1840e 2021-06-06 161207

### LINKED ISSUE #5642: [1840] last "Stadtbahn" tile lay is not possible, gameID 45364
SJE has enough money and tile "L2" is avail, but F12 is no option

1840m 2021-06-07 184824

Zwergenpunk:
closes #5615
closes #5617
closes #5606
closes #5642

### DIFF
--- a/lib/engine/game/g_1840/entities.rb
+++ b/lib/engine/game/g_1840/entities.rb
@@ -146,7 +146,7 @@ module Engine
               sym: 'Karlskirche',
               value: 20,
               revenue: 10,
-              desc: 'Increase income by 20 if tram route runs to or through E21',
+              desc: 'Increase income by 20 if tram route runs to or through E21 (when red tile is placed)',
               abilities: [
                 {
                   type: 'hex_bonus',
@@ -178,7 +178,7 @@ module Engine
               sym: 'Hofburg',
               value: 40,
               revenue: 20,
-              desc: 'Increase income by 20 if tram route runs to or through E19',
+              desc: 'Increase income by 20 if tram route runs to or through E19 (when red tile is placed)',
               abilities: [
                 {
                   type: 'hex_bonus',
@@ -194,7 +194,7 @@ module Engine
               sym: 'Stephansdom',
               value: 50,
               revenue: 25,
-              desc: 'Increase income by 20 if tram route runs to or through D20',
+              desc: 'Increase income by 20 if tram route runs to or through D20 (when red tile is placed)',
               abilities: [
                 {
                   type: 'hex_bonus',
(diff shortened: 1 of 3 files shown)


jenf (approved):

Zwergenpunk:
Added mini bugfix

PR #5628, opened 2021-06-07
Update user.rb

mwayned:
Removes check for 'slack' or 'discord' in URLs allowing webhooks to be sent to other sites.

### DIFF
--- a/models/user.rb
+++ b/models/user.rb
@@ -83,7 +83,7 @@ class User < Base
     end
 
     # rubocop:disable Style/GuardClause
-    if (url = settings['webhook_url']) && !/slack|discord/.match?(URI.parse(url).host)
+    if (url = settings['webhook_url'])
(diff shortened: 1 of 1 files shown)


tobymao:
what sites and what formats do they take?

gpanders:
The webhook code [already supports arbitrary URLs](https://github.com/tobymao/18xx/blob/master/lib/hooks.rb#L23); it just sends a POST with:

```json
{"text": <message>}
```

This means anyway can stand up their own webhook listener that accepts that message format.

mwayned:
I am playing with a site on a raspberry pi (got the domain
18xxconductor.fun for cheap if you can believe that). I plan on basically
just handling the JSON the site already sends to trigger other webhooks
allowing different discord channels in this case to be alerted for
different games.

I believe I messed up my edit as it is so I can update the request to do
what I intended instead of what is submitted.

Thanks,

Matthew

PR #5633, opened 2021-06-07
18ZOO: change grey tile to three different tiles with own icon

carlorusso:


jenf (approved):

jenf (changes requested):
What issue does this fix?

carlorusso:
> What issue does this fix?

The three grey tiles were available only for TI, BB and GI home, but there was no way to understand it; with this fix we have this new tiles:
image

as you can see now is clear that the tile is for specific corporation

jenf (approved):

PR #5638, opened 2021-06-07
1856: Waterloo cannot be used outside of a corporation's OR now

### LINKED ISSUE #5627: 1856 Waterloo & Saugeen can be (incorrectly) be used in SR
https://18xx.games/game/44700?action=183#tools

ventusignis:
Fixes #5627

### DIFF
--- a/lib/engine/game/g_1856/game.rb
+++ b/lib/engine/game/g_1856/game.rb
@@ -363,6 +363,7 @@ module Engine
                         {
                           type: 'tile_lay',
                           owner_type: 'corporation',
+                          when: 'track',
                           consume_tile_lay: true,
                           count: 1,
                           hexes: ['I12'],


jenf (approved):

jenf:
This might cause a lot of games to be archived, can someone check the database?

jenf (changes requested):
please run the validation script on, so we know if we need a migration or not.

I think 'track' forces an option to buy corporations

ventusignis:
If this will cause such wide archiving, is there a way for me to make a migration script to mitigate this assuming people aren't abusing this private?

ventusignis:
Also the fix for #5639 when I'm able to put that together will surely cause archiving so maybe merging this should be delayed until it can be batched with other archiving PRs come up?

Issue #4904, opened 2021-04-01
1822: Implement 2-player rules for MRS/NRS


TBarchet:
It would be cool if the 2-player variant was available for 1822.  I've copied the rules below for reference.  Given the "human readable" instructions for private companies, I would suggest following the recommendation to only use P1-P12.  I'd also probably skip the SW Peninsula variant (13.4.4) unless it's easy to implement.

**2-player variant for the regional scenarios**
13.4.1 This 2-player variant can be used with either regional scenario, but the Medium Regional Scenario is recommended.
13.4.2 Overview:

- 7 bidding tokens per player (combine two colors)
- £750 starting capital
- Certificate limit is 27
- Bank is £12,000

13.4.3 Private companies:

- Recommended: P1 - P12.
- If substituting other private companies, then replace like with like to avoid too great an impact on cash flow and train progression. For example, if introducing the permanent L-train then remove one of the 2P-trains. Or, if introducing the Highland Railway, then remove a mail contract or the cash doubling company). Players may wish to avoid using the tax haven to maintain balance between the players.

13.4.4 Southwest Peninsula variant: add minor companies M22 and M23, and remove two others at random. Both of these minor companies will use the Southwest England off-board area for their home stations.

PR #5657, opened 2021-06-09
[1822MRS] Added 2-player Variant

TBarchet:
Adds the 2 player variant to 1822MRS.  Variant is automatically activated when 2-players are selected.  Checkbox for Advanced still overrides Box 1 Minor placement (M14 instead of M24)

### DIFF
--- a/lib/engine/game/g_1822/game.rb
+++ b/lib/engine/game/g_1822/game.rb
@@ -311,6 +311,7 @@ module Engine
         BIDDING_BOX_START_PRIVATE = 'P1'
 
         BIDDING_TOKENS = {
+		  '2': 7,
           '3': 6,
           '4': 5,
           '5': 4,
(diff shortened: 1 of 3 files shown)


TBarchet:
Adds feature [#4904](https://github.com/tobymao/18xx/issues/4904)

jenf (commented):
Doesn't the bank limit need to be changed?
• Bank is £12,000

TBarchet:
> Doesn't the bank limit need to be changed?
> • Bank is £12,000

For some reason it is specifically listed in the rules for MRS/NRS and the 2-player scenario, despite the fact that 1822 normally has a £12k bank.  I didn't notice until I started to put it into code that there is no bank change for any of the UK 1822s 😄

jenf (approved):

tobymao (approved):

PR #5661, opened 2021-06-09
[1846, 18LA] remove icons when appropriate

### LINKED ISSUE #1762: 1846 - remove unused icons
Companies/Corporations that can be removed from the game that currently leave icons behind:

* Steamboat Company
* Meat Packing Company
* Lake Shore Line
* Illinois Central (can only be removed in 2p Variant)

benjaminxscott:
Several changes related to icon cleanup

1 When a private company is removed (during setup or private closure), also remove its icons
e.g. The "hog" icon for Boomtown private will be removed from H12 if Boomtown is removed during setup

Note that this also fixes a bug in "first edition" variants where the Boomtown and Little Miami private icons were still present even after being removed at setup-time

This also applies to 18LA, for the equivalent Citrus and Port companies

Examples where privates are removed:

no_steam
no_meat
no_meat2
no_boom

2 When the IC corporation is removed in a two-player variant context, remove its icon on related hexes 

Example when IC is removed in a 2P game
no_ic

3 In 18LA, remove the label for `SBL` private after it is used (this is equivalent to LSL in 1846)

before:
before_la_patch
after:
after_la_patch

--
Note that the hex test spec is changed, since the privates added in the second printing of 1846 means there will always be at least one private removed from each Blue and Orange group, which may non-deterministically remove LSL / steamboats / meat

closes #1762

### DIFF
--- a/lib/engine/game/g_1846/game.rb
+++ b/lib/engine/game/g_1846/game.rb
@@ -177,8 +177,16 @@ module Engine
         LITTLE_MIAMI_ICON = 'lm'
 
         MEAT_HEXES = %w[D6 I1].freeze
+        MEAT_ICON = 'meat'
+
         STEAMBOAT_HEXES = %w[B8 C5 D14 I1 G19].freeze
+        STEAMBOAT_ICON = 'port'
+
         BOOMTOWN_HEXES = %w[H12].freeze
+        BOOMTOWN_ICON = 'boom'
+
+        IC_HEXES = %w[E5 F6 G5 H6 J4].freeze
+        IC_ICON = 'ic'
 
         MEAT_REVENUE_DESC = 'Meat-Packing'
 
@@ -245,6 +253,9 @@ module Engine
             @round.active_step.companies.delete(company)
           end
           remove_from_group!(blue_group, @companies) do |company|
+            remove_boomtown_icons if company == boomtown
+            remove_steamboat_icons if company == steamboat
+            remove_meat_icons if company == meat_packing
             company.close!
             @round.active_step.companies.delete(company)
           end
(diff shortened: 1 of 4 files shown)


michaeljb:
@benjaminxscott have you tested 18 Los Angeles with this? The 18LA code doesn't redefine any of the relevant behavior and the Citrus company (the Meat-Packing analog) still uses "meat" in the code and hexes so I think it should work.

lib/engine/game/g_1846/game.rb
```diff
@@ -631,6 +641,30 @@ def remove_lm_icons
           end
         end
 
+        def remove_boomtown_icons
+          self.class::BOOMTOWN_HEXES.each do |hex|
+            hex_by_id(hex).tile.icons.reject! { |icon| icon.name == self.class::BOOMTOWN_ICON }
+          end
+        end
```
michaeljb, line 648: These `remove_*_icons` method are a little repetitive, could they be changed to something like this?

```suggestion
        def remove_icons(hex_ids, icon_name)
          hex_ids.each do |hex_id|
            hex_by_id(hex_id).tile.icons.reject! { |icon| icon.name == icon_name}
          end
        end

        def remove_boomtown_icons
          remove_icons(hex_ids,  self.class::BOOMTOWN_ICON)
        end
```

benjaminxscott: Sure, I agree w/ the DRY violation there

Do you know a ruby-ish way to obtain `hex_ids` directly from the company instead of having some duplicative vars / hash?
e.g. `steamboat.ability['assign_hexes'].hexes`

benjaminxscott:
It seems to work as expected in LA, removing the port and "meaty citrus" markers

Screenshot 2021-06-10 10 18 48 AM
Screenshot 2021-06-10 10 17 57 AM

PR #5663, opened 2021-06-10
18KA initial commit

ventusignis:
18KA is my prototype of a bigger 1856 on mars with more shenanigans
This has the map, corporations, privates, and trains


scottredracecar:
Wow, cool!

lib/engine/game/g_18_ka.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1836Jr56
```
tobymao, line 5: name

ventusignis: Fixed

tobymao (approved):

Issue #5634, opened 2021-06-07
[18ZOO] train run should now have the $N inside the texts


carlorusso:
18ZOO has two currencies:
$N are for only money, train revenue doesn't contain anything

![image](https://user-images.githubusercontent.com/1408567/120980125-96806f80-c776-11eb-81df-7e55ef66586c.png)
![image](https://user-images.githubusercontent.com/1408567/120980185-a8faa900-c776-11eb-97e3-e45b7bafd7da.png)

Please remove $N from the previous points

serpentium:
To clarify, should look like
GI runs a 2S train for **50N**: L4-K9
GI pays out **50N** = 1$N per share (2$N to Player 1)

PR #5672, opened 2021-06-10
18ZOO: fixes https://github.com/tobymao/18xx/issues/5634

carlorusso:
fixes https://github.com/tobymao/18xx/issues/5634

breaking changes

### DIFF
--- a/lib/engine/game/g_18_zoo/game.rb
+++ b/lib/engine/game/g_18_zoo/game.rb
@@ -698,7 +698,7 @@ module Engine
             end
           end
 
-          revenue
+          { type: :revenue, value: revenue }
         end
 
         def zoo_ticket?(company)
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_zoo/game.rb
```diff
@@ -844,9 +846,66 @@ def routes_subsidy(routes)
...
+          val.is_a?(Integer) ? super : "#{val[:value]} nuts"
+        end
+
+        def compute_stops(route)
+          visits = route.visited_stops
+          distance = route.train.distance
+          return visits if distance.is_a?(Numeric)
+          return [] if visits.empty?
+
+          # distance is an array of hashes defining how many locations of
```
tobymao, line 865: it looks like you copy and pasted code, let's not do that

lib/engine/game/g_18_zoo/game.rb
```diff
@@ -844,9 +846,66 @@ def routes_subsidy(routes)
           subsidy
         end
 
+        def routes_revenue(routes)
+          value = routes.sum { |route| route.revenue.is_a?(Integer) ? route.revenue : route.revenue[:value] }
+          { type: :revenue, value: value }
+        end
```
tobymao, line 852: what's going on here? this is not consistnent with any other game which seems problematic

PR #5674, opened 2021-06-11
18fl final touches

ventusignis:
Fixed a very minor bug I found (erroneous browser console logging) and added the 70% ownership optional rule for 2p

### DIFF
--- a/lib/engine/game/g_18_fl/step/token.rb
+++ b/lib/engine/game/g_18_fl/step/token.rb
@@ -83,6 +83,7 @@ module Engine
             node = city_hex.respond_to?(:city?) ? city_hex : city_hex.tile.towns.first
             min_distance = INFINITE_DISTANCE
             goal_cities = entity.tokens.select(&:city).map(&:city)
+            return unless node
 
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_fl/step/token.rb
```diff
@@ -83,6 +83,7 @@ def token_cost_override(entity, city_hex, _slot, token)
             node = city_hex.respond_to?(:city?) ? city_hex : city_hex.tile.towns.first
```
tobymao, line 83: return unless (node = )

ventusignis: Wouldn't that be a syntax error? I could also do `return if !node`

tobymao: it is?

ventusignis: I've changed it to `return if node.nil?`

tobymao: return unless node

PR #5681, opened 2021-06-11
18ZOO: fixes https://github.com/tobymao/18xx/issues/5634

carlorusso:


carlorusso:
Replace https://github.com/tobymao/18xx/pull/5672 with a simpler solution

tobymao (approved):

PR #5691, opened 2021-06-12
Make it possible to override orientation check

perwestling:
This is needed for 1893, when upgrading Leverkusen hex, from
yellow to green.

Yellow Leverkusen hex has two towns, each on a separate track.
The green Leverkusen hex hex have track at same edges, but the
track makes an X with a city in the middle. It does remove the
towns.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1366,6 +1366,11 @@ module Engine
         true
       end
 
+      def override_legal_tile_rotation?(_entity, _hex, _tile)
+        # Return true to skip further general checks for legal tile rotation
+        false
+      end
+
       def can_par?(corporation, parrer)
         return false if corporation.par_via_exchange && corporation.par_via_exchange.owner != parrer
         return false if corporation.needs_token_to_par && corporation.tokens.empty?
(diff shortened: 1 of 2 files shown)


roseundy:
If you have an overridden Step::Track, couldn't you just override the legal_tile_rotation? method?

perwestling:
@roseundy I do have a overridden Step::Track. And Step::Track do include tracker.rb
If I have a legal_tile_rotation? there how do I use it to call the legal_tile_rotation? implementation in tracker.rb?
The legal_tile_rotation? in my subclass do overshadow the one in tracker.rb.

Like this
module A
  def foo
     ...
     <do some checks>
     ...
  end
end

class B
  include A
  def foo
     return true if some_check

     call foo in module A
  end
end

tobymao (approved):

roseundy:
Just call super

perwestling:
I think you are right when it comes to this case:
Track include Tracker
MySubclass subclass Track
Method in MySubclass will override method in Track, and call super will use superclass implementation

But my example does not work, as B does not have a super. Now that is a different case. And I do think after all this is not needed. So I will probably revert my change above and do as you suggest instead.

PR #5692, opened 2021-06-12
1888: add destination to corp status

### LINKED ISSUE #5678: [1888] Show destination information on charter
Wiki page (https://github.com/tobymao/18xx/wiki/1888) states: "If a train’s route contains both home and destination the bonus shown on the company sheet is added to the income."

This information is not shown in-game in the 18xx.games version. I would suggest add this - compare e.g. 18AL on one way of doing it.
(In essence: having a corporation specific ability that show the destination, and the bonus.)

roseundy:
Fixes #5678 

1888_dest_status

### DIFF
--- a/lib/engine/game/g_1888/game.rb
+++ b/lib/engine/game/g_1888/game.rb
@@ -265,6 +265,15 @@ module Engine
           bonus = destinated?(route.train.owner, route.stops) ? ' (dest)' : ''
           super + bonus
         end
+
+        def destination_str(corp)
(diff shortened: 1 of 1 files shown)


perwestling:
I was thinking of having that as an Ability instead. (It is possible to use Base ability so a new one does not need to be created.)
Advantage of Ability is that one can have it clickable if the room is not enough, and there one can add extra information.

I guess advantage of having it as above is simplicity.

tobymao (approved):

PR #5708, opened 2021-06-14
18ZOO: show debt inside Entity view

carlorusso:
Before the modification:
image

After the modification:
image


carlorusso:
let me know if do you think is better in this way, or I should always call a method to get the text

tobymao:
just leave it as loan

PR #5714, opened 2021-06-14
1856: Can't gain full cert if only half cert space

### LINKED ISSUE #5712: 1856: Allowed to full cert when only half cert below limit
https://18xx.games/game/46189?action=845#tools

ventusignis:
Fixes #5712


dagibbs:
This issue seems to have recurred/come back.

Game: https://18xx.games/game/100214 after action 920 (I would guess action 921) -- Dagrim is at 10.5 of 11 certificates, but is not allowed to buy another share.

The linked rules, in the "CGR Share Limits" section, include the line: "Players may hold one half certificate above the current stock limits, but players may not purchase a half share if they are as their limit."

At 10.5 of 11, I should not be at my limit -- and so I should be able to purchase a full share, and end up 1/2 share above my limit.

-David

PR #5734, opened 2021-06-17
18ZOO: fixes https://github.com/tobymao/18xx/issues/5702

carlorusso:


perwestling (changes requested):
bundle.shares.size == 1

can be changed to

bundle.shares.one?

carlorusso:
> bundle.shares.size == 1
> 
> can be changed to
> 
> bundle.shares.one?

done in e506d8977

perwestling (approved):

PR #5736, opened 2021-06-17
Fix issue that stalled game/44700

ventusignis:
I'm not sure why `can_par?` was implemented this way when this new way is far simpler. Maybe it was used for something else earlier in 18xx.games but I couldn't find a reason for it not to simply be that you cannot par the national


perwestling (approved):
I do not understand the original code. What has national.ipoed to do with it? What use case was it for being parred if national.ipoed was true?

The new code looks straightforward and what I would expect for 1856.

PR #5747, opened 2021-06-18
18ZOO: fix following bug on https://github.com/tobymao/18xx/issues/5702

carlorusso:


lib/engine/game/g_18_zoo/game.rb
```diff
@@ -197,8 +197,6 @@ class SharedGame < Game::Base
 
         SELL_BUY_ORDER = :sell_buy
 
-        NEXT_SR_PLAYER_ORDER = :most_cash # TODO: check if a bug
```
carlorusso, line 200: I checked and I don't use this override

lib/engine/game/g_18_zoo/step/buy_train.rb
```diff
@@ -62,10 +62,10 @@ def process_buy_train(action)
             @game.first_train_of_new_phase = false
           end
 
-          def sellable_bundle?(bundle)
```
carlorusso, line 65: using just single item bundle was breaking the text for "you can get $N selling shares", so designer decided we can ignore for online version

lib/engine/game/g_18_zoo/step/buy_train.rb
```diff
@@ -62,10 +62,10 @@ def process_buy_train(action)
...
-            return false unless bundle.shares.one?
-
+          def process_sell_shares(action)
```
carlorusso, line 65: player should be able to sell 1 tiger, then another tiger

PR #5754, opened 2021-06-19
18Carolinas: re-checkin

roseundy:
First checkin for 18Carolinas (OK, technically the 2nd).

Common file changes:
- UI::Hex
- Engine::Tile, Engine::Part::Frame
Add ability to show a 2nd color on a frame.


roseundy:
updated_map
updated_tiles

assets/app/view/game/hex.rb
```diff
@@ -62,6 +62,19 @@ def render
             points: FRAME_COLOR_POINTS,
           }
           children << h(:polygon, attrs: attrs)
+
+          if (color2 = @tile&.frame&.color2)
+            attrs = {
+              stroke: color2,
+              'stroke-width': FRAME_COLOR_STROKE_WIDTH,
```
tobymao, line 69: we do style strkeWidth: now

roseundy: OK. Was just cut and pasting from existing code in the same file. I'll update all the old style attributes in the file.

roseundy: So, strokeWidth: FRAME_COLOR_STROKE_WIDTH doesn't actually work.

roseundy: Not sure SVG styles work the same.

PR #5768, opened 2021-06-19
Possibly show Price in Minor charter

perwestling:
Show price if minor has a value and no owner.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -314,7 +314,7 @@ module View
       end
 
       def render_minor_price
-        price_str = "Price: #{@game.format_currency(@corporation.value - @corporation.discount)}"
+        price_str = "Price: #{@game.format_currency(@corporation.min_bid)}"
         h(:div, { style: { textAlign: 'center' } }, price_str)
       end
 
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -209,7 +209,9 @@ module View
         }
 
         holdings =
-          if !@corporation.corporation? || @corporation.floated? || @corporation.trains.any?
+          if @corporation.minor? && !@corporation.owner && !@corporation.value.zero?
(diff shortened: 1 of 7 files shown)


perwestling:
If the minor has attribute 'value' non-zero, do show Price in charter:
image

If discount is non-zero, present Price as value - discount.

If value is zero, or owner is set, do not change current view. So it will looks like this:
image

perwestling:
Commit also contains some updates to use the new value and discount attributes. As far as I know it is so far only 1893 that uses these.

assets/app/view/game/corporation.rb
```diff
@@ -311,6 +313,11 @@ def render_tokens
         end)
       end
 
+      def render_minor_price
+        price_str = "Price: #{@game.format_currency(@corporation.value - @corporation.discount)}"
```
tobymao, line 317: should this be min_bid?

tobymao:
why do minors have value and discount? shouldn't they be companies then?

assets/app/view/game/corporation.rb
perwestling: Fixed

perwestling:
> why do minors have value and discount? shouldn't they be companies then?

@tobymao Minors is in between Company and Corporation.
They do operate - but does not have any shares - but they have an owner.
They cannot be sold. They can have abilities. etc

In latter titles (1893, and maybe 1824) I have moved towards buying the minor instead of buying a company which is connected to a minor and indirectly taken over it.  

So I introduced possibility to buy Minor during the auction. 
But the Minor did not show the value/cost. If you clicked on them a Buy button appear, but you cannot do this for all.

In all it makes for an easier implementation, to use Minors directly instead of Proxy privates.

tobymao:
but don't all other games like 1846 use company and then convert to minor?

perwestling:
Ok, I will throw the changes with minors out, by making a commit that uses privates instead.

PR #5769, opened 2021-06-20
18 Carolinas / 1860: Move to common buy_sell_par_shares step

roseundy:
18Carolinas and 1860 have almost the exact same stock round rules.  This PR creates a common step: buy_sell_par_shares_companies that will be used by both of the games.

This finishes the changes needed for the 18Carolinas stock round.

This should have zero impact on 1860 games, but I would appreciate this being checked before being deployed.


lib/engine/step/buy_sell_par_shares_companies.rb
```diff
@@ -0,0 +1,174 @@
...
+        return ['sell_shares'] if must_sell?(entity)
+
+        actions = []
+        actions << 'buy_shares' if can_buy_any?(entity)
+        actions << 'par' if can_ipo_any?(entity)
+        actions << 'buy_company' if can_buy_any_companies?(entity)
+        actions << 'sell_shares' if can_sell_any?(entity)
+        actions << 'sell_company' if can_sell_any_companies?(entity)
+
+        actions << 'pass' if actions.any?
```
tobymao, line 23: unless empty

lib/engine/step/buy_sell_par_shares_companies.rb
```diff
@@ -0,0 +1,174 @@
...
+      def can_buy_company?(player, company)
+        !did_sell?(company, player)
+      end
+
+      def can_buy_any_companies?(entity)
+        return false if bought? ||
+          !entity.cash.positive? ||
+          @game.num_certs(entity) >= @game.cert_limit
+
+        @game.companies.select { |c| c.owner == @game.bank }.reject { |c| did_sell?(c, entity) }.any?
```
tobymao, line 52: companies.any? { ...}

lib/engine/step/buy_sell_par_shares_companies.rb
roseundy: Yep

lib/engine/step/buy_sell_par_shares_companies.rb
roseundy: OK

Issue #5717, opened 2021-06-15
[18ZOO] Special track cannot lay invalid tiles


carlorusso:
https://gist.github.com/carlorusso/3f2c7049d05c41c3c83565c9db26d301

**How to Reproduce the Bug?**

Rabbit lays an invalid track

PR #5793, opened 2021-06-22
18ZOO: fix https://github.com/tobymao/18xx/issues/5717

carlorusso:


carlorusso:
please archive_alpha_games

perwestling (approved):

lib/engine/game/g_18_zoo/step/tracker.rb
```diff
@@ -41,6 +47,13 @@ def update_tile_lists(tile, old_tile)
           end
 
           def check_track_restrictions!(entity, old_tile, new_tile)
+            # handle edge cases where the new ends override olds but original paths are not touched
+            raise GameError, 'New track must override old one' if INVALID_TRACK_UPDATES.any? do |props|
```
tobymao, line 51: why are these edge cases?

carlorusso: easier to explain with some images:
original:
image
track to be invalid but considered valid:
image
another with thee same problem:
image

the 5th problematic cases are all like this:
I think the reason is related to the way Path is created (splitting by the town)

tobymao: does this fail on other games as well? that's why i ask about your approach. I think these are caught in the general code so you must have an error in your code?

carlorusso: yes, I was thinking the same, but the only failing tracks are custom green tile with two cities - only 981 and 991 have the same style
1829 / carolinas has 1a (but yellow doesn't suffer for this problem)
1840 has narrow / normal track
1893 has K1 (but is yellow)
so no other game have this kind of track, I'm not able to test in other game
and I also checked the base method "check_track_restrictions", it works for normal case (like when you upgrade but you don't use the new track), and it fails only when you don't have the original track, you have 1 cities -> 2 cities and both ends or original tile are in two different path in the new one

tobymao: i'd rather fix the base method to support this

carlorusso: sure, I could try, but I decided to do this solution because I didn't get how to solve there; the Paths in both old_tile and new_tile doesn't show anything useful (my knowledge of that part is really poor)
can you suggest me anything to try to solve it there?

tobymao: i guess you need to make sure that city is still connected to the same exits

carlorusso: I removed this check here

carlorusso:
@tobymao following your suggestion I updated the shared code, now it check that all the original city have at least the same exits in the new tile

carlorusso:
@tobymao any news on this? I miss this PR, and a PR with. some text change to move to beta

PR #5806, opened 2021-06-23
Fix being unable to buy a 5% share of CGR to reach cert limit in 1856

### LINKED ISSUE #5803: 1856 buy the last 1/2 share of CGR
1856 game id 44700

i own 15.5 shares of 16 and i can not purchase a 1/2 share of CGR.

ventusignis:
Fixes #5803

### DIFF
--- a/lib/engine/game/g_1856/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1856/step/buy_sell_par_shares.rb
@@ -89,11 +89,10 @@ module Engine
 
             corporation = bundle.corporation
 
-            bundle_size = bundle.respond_to?(:shares) ? bundle.shares.sum(&:cert_size) : bundle.cert_size
-
             corporation.holding_ok?(entity, bundle.percent) &&
               !attempt_cgr_action_while_not_floated?(bundle) &&
-              (!corporation.counts_for_limit || exchange || @game.num_certs(entity) + bundle_size <= @game.cert_limit)
+              (!corporation.counts_for_limit || exchange ||
+                @game.num_certs(entity) + bundle.shares.sum(&:cert_size) <= @game.cert_limit)
           end
 
           def attempt_cgr_action_while_not_floated?(bundle)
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1856/step/buy_sell_par_shares.rb
```diff
@@ -89,9 +89,11 @@ def can_gain?(entity, bundle, exchange: false)
 
             corporation = bundle.corporation
 
+            bundle_size = bundle.respond_to?(:shares) ? bundle.shares.sum(&:cert_size) : bundle.cert_size
```
tobymao, line 92: why wouldn't bundle have shares?

ventusignis: If there is only one share available, instead of being a ShareBundle it is a Share at this point

tobymao: why? shouldn't everything always be a share bundle? all shares should be share.to_bundle, did you ovveride something?

ventusignis: I added `puts 'base cb', entity, bundle` to `can_buy?` in `Engine::Step::BuySellParShares` and that shows sometimes being single shares being passed to `can_buy?` (and then to `can_gain?`)in unrelated games (tested in 1889)

image


From searching the codebase it looks like the default behavior is:
Market/Treasury/IPO shares: Share
Par new Corporation: Bundle
Corporate Share Buy: Bundle

tobymao: ok, can we cahnge market treasure ipo to be bundle?

ventusignis: I changed them to use `.to_bundle` and the tests passed locally and buying the 0.5 cert worked in 1856 after adjusting

PR #5813, opened 2021-06-23
18ZOO: fix https://github.com/tobymao/18xx/issues/5717 and invert no_…

carlorusso:
…local logic


carlorusso:
following @KevinR1000 suggestion

PR #5816, opened 2021-06-23
[1893] Correct presidency change (fixes #5807)

### LINKED ISSUE #5807: 1893 AGV giving a share to the wrong player
When I bought a 5th share of AGV, it gave it to another player in game#46583. It shows me as president with 40% and the other player has 50% though.

perwestling:


lib/engine/game/g_1893/share_pool.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative '../../share_pool'
+
+module Engine
+  module Game
+    module G1893
+      class SharePool < Engine::SharePool
+        def shares_for_presidency_swap(shares, num_shares)
```
tobymao, line 9: is there a way around this?

perwestling: The problem is that when presidency is exchanged, and the new president own a double certificate, then special handling is needed. 
The alternatives? I suppose:

1. Do the change in the base class - so that it would apply to any title with a 20% non-president share. BUT! It would probably be tricky in some cases - e.g. having a more complex share structure.
2. Change base class to call @game.shares_for_presidency and define it in base game.rb - and override it in 1993 game.rb

@tobymao Should I go for number 2?

PS. The only game I know of with a similar problem is 1849. I do not know if the problem has been solved there - as the 20% non-president certificate also always is the last one it probably will never happen.
Just checked - 1849 is the only game, besides 1893, that have double cert. And it has no special code for this case.

tobymao: ok let's do #2

perwestling: @tobymao Fixed

Issue #5749, opened 2021-06-18
[18ZOO] - 1S cannot handle single city + multiple towns


carlorusso:
https://gist.github.com/carlorusso/6665d3c9bf7d737020405173d8fc4302
http://localhost:9292/fixture/18ZOO/or_power.patch?action=88

select local on click works:
<img width="1234" alt="image" src="https://user-images.githubusercontent.com/1408567/122560486-f3b2d580-d040-11eb-8f75-e5448811abb9.png">

select a city + a town works:
<img width="1233" alt="image" src="https://user-images.githubusercontent.com/1408567/122560546-03321e80-d041-11eb-9e23-9027c09c87ad.png">

cannot select a 2nd town:
<img width="1224" alt="image" src="https://user-images.githubusercontent.com/1408567/122560589-104f0d80-d041-11eb-8c4c-63de8b1fcf1f.png">

tobymao:
this change doesn't really make sense to me, what does a local train mean?

carlorusso:
"local train" is used for trains that can touch a city (and maybe other town)

tobymao:
how does it differ from 1822?

carlorusso:
1822 has the L train that can touch a city and a single town.
and the current code works perfectly to handle just a single town.
I need to be able to touch any number of town (other than a single city); and with the current code, when you click the second town, it will remove the city

tobymao:
isn't that just a normal train? for example in 1830, a train can run from a city to a town, how is that different from a local train?

carlorusso:
local train are able to run:
- single city (will use "local")
- city + towns
! local train are not able to touch a single city (they need at least 2 node to be considered)

PR #5817, opened 2021-06-23
18ZOO: fix https://github.com/tobymao/18xx/issues/5749

carlorusso:


lib/engine/route.rb
```diff
@@ -130,7 +131,7 @@ def segment(chain, left: nil, right: nil)
     end
 
     def touch_node(node)
-      if connection_data.any?
+      if connection_data.any? && !local_connection?
```
tobymao, line 134: !connect_data.empty?

lib/engine/route.rb
```diff
@@ -5,7 +5,7 @@
 module Engine
   class Route
     attr_accessor :halts, :routes
-    attr_reader :last_node, :phase, :train, :abilities
+    attr_reader :last_node, :phase, :train, :abilities, :local_length
```
tobymao, line 8: why does this need to be an attribute? no one uses it

lib/engine/route.rb
```diff
@@ -20,6 +20,7 @@ def initialize(game, phase, train, **opts)
       @subsidy = opts[:subsidy]
       @halts = opts[:halts]
       @abilities = opts[:abilities]
+      @local_length = @game.respond_to?(:local_length) ? @game.local_length : 2
```
tobymao, line 23: remove the respond_to

lib/engine/route.rb
carlorusso: done in 68d946131

lib/engine/route.rb
carlorusso: removed; I put there because originally I was logging it and ruby code was failing :)

lib/engine/route.rb
carlorusso: removed in 68d946131

Issue #5818, opened 2021-06-23
[1840] Missing playable hex compared to published map


McGill31:
The 18xx.game implementation of 1840 has no hex in position D2.

In the published physical game, D2 is a playable hex, allowing track to run West from Dornbach into either C3 or E3.

Can the 18xx.game implementation be changed to match the physical map?

PR #5820, opened 2021-06-24
[1840] Update map.rb to add missing hex

philcampeau:
Added empty hex D2, which does appear on the published map. Also added Wien name to hex I11, as it's on the published board. Fixes bug #5818


Zwergenpunk:
This tile has to be removed for the 2p snd small 3p map

philcampeau:
Indeed. Updated.

Zwergenpunk (approved):

Zwergenpunk:
Actually, your second change is not needed, since the hex is already removed for the 3p map
https://github.com/tobymao/18xx/blob/a0150b4306ed2aa5d3776bc5e09ba54322a4b237/lib/engine/game/g_1840/game.rb#L236

I'm sorry for the uneeded work. I did not assume I remove a not existing hex.

tobymao:
you need to fix the build

philcampeau:
Should be good now

PR #5827, opened 2021-06-24
This fixes subsidy handling not being able to handle revenue_for throwing an error

ventusignis:
I messed around with 18VA last night to see how much effort it would be to implement it. It looks like it wouldn't be too much effort using 18FL as a base and I got pretty far with it. I'm splitting up the progress into several PRs

Without this, if a game that has a subsidy throws a GameError during `revenue_for` (for example 1836jr does this to enforce legal use of the port) will log it out to the console instead to the train view as it normally would be

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -270,7 +270,11 @@ module View
           '(Invalid Route)'
         end
 
-        render_halts ||= @game.respond_to?(:routes_subsidy) && @game.routes_subsidy(active_routes).positive?
+        render_halts ||= begin
+          @game.respond_to?(:routes_subsidy) && @game.routes_subsidy(active_routes).positive?
+        rescue Engine::GameError
+          false
+        end
         subsidy = render_halts ? " + #{@game.format_currency(@game.routes_subsidy(active_routes))} (subsidy)" : ''
         buttons = [
           h('button.small', { on: { click: clear } }, 'Clear Train'),


tobymao:
1836jr doesn't call routes_subsidy, i don't get this change?

ventusignis:
18VA does though (in the other PR). I'm splitting up my 18VA branch into smaller PRs to avoid putting up very large PRs

tobymao:
can we avoid raising in subsidy and raise in revenue?

ventusignis:
It's not being raised in subsidy_for; it's being raised in `revenue_for`.
`subsidy` in route calls this
`      @subsidy ||= @game.subsidy_for(self, stops)`

and stops is defined as
`      @stops ||= @game.compute_stops(self)`

And the base game `compute_stops` uses `revenue_for` on the stops because the trains in 18VA are not number trains so it needs to sort them

Do you think would be better to put the rescue in `compute_stops`?
I could do something like
```ruby
        max_num_stops.downto(1) do |num_stops|
          # to_i to work around Opal bug
          stops, revenue = visits.combination(num_stops.to_i).map do |stops|
            # 1) At least one stop must have a token
            next if stops.none? { |stop| stop.tokened_by?(route.corporation) }
...

            begin
              revenue = revenue_for(route, stops)
            rescue Engine::GameError
              next
            end
            [stops, revenue]
          end.compact.max_by(&:last)
```

tobymao:
yea, put the junk in 18va

ventusignis:
I'm going to put this in 18VA in after the initial commit is merged

PR #5828, opened 2021-06-24
18VA initial commit

ventusignis:
This has the map, trains, corporations, and privates. This does *not* have logos. They are shamelessly stolen from other games (like the B&O) or even incorrectly used from other games if the color matches (FECR from 18FL being used for RFPR here)

image
image


lib/engine/game/g_18_va/game.rb
```diff
@@ -0,0 +1,708 @@
+# frozen_string_literal: true
+
+require_relative 'meta'
+require_relative '../base'
+
+if RUBY_ENGINE == 'opal'
+  require_tree 'engine/game/g_18_va/step'
```
tobymao, line 7: why do you need to require this stuff? do other games do this?

ventusignis: Only 18FL (which I cloned 18VA from). I removed it from both and it worked. I also removed the Assign step (for now) since that's not needed to make the build work anymore

PR #5830, opened 2021-06-24
18Carolinas: track rules

roseundy:
Implement all track rules, including track conversion operation.
Make all "Southern" track dual gauge instead of narrow gauge (so Graph will work correctly for mixed segments).
Finish implementing dual-sided tiles

Common file changes:
UI::Round::Operating - add optional "mode" button, alternate route selector view
UI::TrackConversion - new view
Engine::Route - moved token check to game
Engine::Game::Base - move an Engine::Route check method here
Engine::Step::Route - added default alternate_selector? method

Mode Button:
mode_button

Track Conversion Mode:
Revamped_Conversion

### DIFF
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -37,8 +37,10 @@ module View
           left = []
           left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
-          left << h(RouteSelector) if @current_actions.include?('run_routes') && !@step.alternate_selector?
-          left << h(TrackConversion) if @current_actions.include?('run_routes') && @step.alternate_selector?
+          if @current_actions.include?('run_routes')
+            left << h(RouteSelector) unless @step.conversion?
+            left << h(TrackConversion) if @step.conversion?
+          end
           left << h(Dividend) if @current_actions.include?('dividend')
           left << h(Convert) if @current_actions.include?('convert')
           left << h(SwitchTrains) if @current_actions.include?('switch_trains')
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -38,8 +38,7 @@ module View
           left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
           if @current_actions.include?('run_routes')
-            left << h(RouteSelector) unless @step.conversion?
-            left << h(TrackConversion) if @step.conversion?
+            left << (@step.conversion? ? h(TrackConversion) : h(RouteSelector))
           end
           left << h(Dividend) if @current_actions.include?('dividend')
           left << h(Convert) if @current_actions.include?('convert')
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -38,8 +38,7 @@ module View
           left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
           if @current_actions.include?('run_routes')
-            left << h(RouteSelector) unless @step.conversion?
-            left << h(TrackConversion) if @step.conversion?
+            left << @step.conversion? ? h(TrackConversion) : h(RouteSelector)
           end
           left << h(Dividend) if @current_actions.include?('dividend')
           left << h(Convert) if @current_actions.include?('convert')
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -38,7 +38,8 @@ module View
           left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
           if @current_actions.include?('run_routes')
-            left << @step.conversion? ? h(TrackConversion) : h(RouteSelector)
+            left << h(RouteSelector) unless @step.conversion?
+            left << h(TrackConversion) if @step.conversion?
           end
           left << h(Dividend) if @current_actions.include?('dividend')
           left << h(Convert) if @current_actions.include?('convert')
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -24,6 +24,7 @@ module View
     module Round
       class Operating < Snabberb::Component
         needs :game
+        needs :mode, store: true, default: nil
 
         def render
           round = @game.round
@@ -34,6 +35,7 @@ module View
           entity = entity.owner if entity.company? && !round.active_entities.one?
 
           left = []
+          left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
           left << h(RouteSelector) if @current_actions.include?('run_routes')
           left << h(Dividend) if @current_actions.include?('dividend')
@@ -124,6 +126,14 @@ module View
           ]
 
           h(:div, children)
+        end
+
+        def render_mode_button
+          pressed = lambda do
+            store(:mode, @step.change_mode)
+          end
+
+          h('button.small', { on: { click: pressed } }, @step.mode_text)
         end
       end
     end
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -37,7 +37,8 @@ module View
           left = []
           left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
-          left << h(RouteSelector) if @current_actions.include?('run_routes')
+          left << h(RouteSelector) if @current_actions.include?('run_routes') && !@step.alternate_selector?
+          left << h(TrackConversion) if @current_actions.include?('run_routes') && @step.alternate_selector?
           left << h(Dividend) if @current_actions.include?('dividend')
           left << h(Convert) if @current_actions.include?('convert')
           left << h(SwitchTrains) if @current_actions.include?('switch_trains')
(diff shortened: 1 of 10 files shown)


assets/app/view/game/route_selector.rb
```diff
@@ -183,23 +189,35 @@ def render
           },
         }
 
-        instructions = 'Click revenue centers, again to cycle paths.'
-        instructions += ' Click button under Revenue to pick number of halts.' if render_halts
+        if @step.respond_to?(:instructions)
```
tobymao, line 192: avoid doing respond_to, please remove all of them and add implementation in base game

assets/app/view/game/route_selector.rb
```diff
@@ -183,23 +189,35 @@ def render
...
+          instructions = @step.instructions
+        else
+          instructions = 'Click revenue centers, again to cycle paths.'
+          instructions += ' Click button under Revenue to pick number of halts.' if render_halts
+        end
+
+        h3_text = @step.respond_to?(:section_header) ? @step.section_header : 'Select Routes'
+
+        table_header = if @step.respond_to?(:show_table_header?) && !@step.show_table_header?
```
tobymao, line 201: why

assets/app/view/game/route_selector.rb
```diff
@@ -258,31 +276,70 @@ def actions(render_halts)
...
           marginTop: '1rem',
           padding: '0.2rem 0.5rem',
         }
 
-        revenue = begin
-          @game.format_revenue_currency(@game.routes_revenue(active_routes))
-        rescue Engine::GameError
-          '(Invalid Route)'
-        end
+        revenue = if @step.respond_to?(:total_str)
```
tobymao, line 305: too much garbage can we avoid?

assets/app/view/game/route_selector.rb
```diff
@@ -258,31 +276,70 @@ def actions(render_halts)
...
         render_halts ||= @game.respond_to?(:routes_subsidy) && @game.routes_subsidy(active_routes).positive?
         subsidy = render_halts ? " + #{@game.format_currency(@game.routes_subsidy(active_routes))} (subsidy)" : ''
-        buttons = [
-          h('button.small', { on: { click: clear } }, 'Clear Train'),
-          h('button.small', { on: { click: clear_all } }, 'Clear All'),
-          h('button.small', { on: { click: reset_all } }, 'Reset'),
-        ]
-        if @game_data.dig('settings', 'auto_routing') || @game_data['mode'] == :hotseat
+
+        show_ct_button = !@step.respond_to?(:show_clear_train_button?) || @step.show_clear_train_button?
```
tobymao, line 322: why!?

lib/engine/game/g_18_carolinas/game.rb
```diff
@@ -297,14 +343,62 @@ def operating_round(round_num)
...
+            G18Carolinas::Step::Track,
             Engine::Step::Token,
             Engine::Step::Route,
             Engine::Step::Dividend,
             Engine::Step::BuyTrain,
           ], round_num: round_num)
         end
 
+        def upgrades_to?(from, to, special = false, selected_company: nil)
+          standard = to.paths.any? { |p| p.track == :broad }
```
tobymao, line 355: partition

lib/engine/game/g_18_carolinas/game.rb
```diff
@@ -297,14 +343,62 @@ def operating_round(round_num)
...
+          southern = to.paths.any? { |p| p.track != :broad }
+
+          north = @north_hexes.include?(from.hex)
+          south = @south_hexes.include?(from.hex)
```
tobymao, line 359: do you really need both? aren't they opposites?

lib/engine/game/g_18_carolinas/step/track.rb
```diff
@@ -0,0 +1,102 @@
...
+      module Step
+        class Track < Engine::Step::Track
+          LAY_ACTIONS = %w[lay_tile pass].freeze
+          CONVERT_ACTIONS = %w[run_routes pass].freeze
+          ALL_ACTIONS = %w[lay_tile run_routes pass].freeze
+
+          def actions(entity)
+            return [] unless entity == current_entity
+            return [] if entity.company? || !can_lay_tile?(entity) && @mode == :new_track
+            return ALL_ACTIONS if @game.loading && @game.phase.available?('5')
```
tobymao, line 17: game.loading? this doesn't seem right

assets/app/view/game/route_selector.rb
roseundy: Unfortunately, this has be specific to a step, not a game, since I need route_selector to work normally for the actual run_routes step.

assets/app/view/game/route_selector.rb
roseundy: Because the header is confusing for track conversion.

assets/app/view/game/route_selector.rb
roseundy: Yes, I think I need to create a new view for track_conversion.

assets/app/view/game/route_selector.rb
roseundy: Removing buttons because it's confusing for track conversion. I will build a new view/action.

lib/engine/game/g_18_carolinas/game.rb
roseundy: No. There are some hexes with both.

lib/engine/game/g_18_carolinas/game.rb
roseundy: OK

lib/engine/game/g_18_carolinas/step/track.rb
roseundy: It's needed because of the modality. When loading mode is set to track laying mode, so when it encounters a run_routes it fails.

lib/engine/game/g_18_carolinas/game.rb
roseundy: Hmmm. Except standard and southern are booleans, not lists of paths. standard needs to be true if any path has broad track. southern needs to be true if any path isn't broad.

roseundy:
OK, completely new approach with new UI View for track_conversion.

assets/app/view/game/round/operating.rb
```diff
@@ -34,8 +35,10 @@ def render
           entity = entity.owner if entity.company? && !round.active_entities.one?
 
           left = []
+          left << render_mode_button if @step.respond_to?(:mode_enabled?) && @step.mode_enabled?
           left << h(SpecialBuy) if @current_actions.include?('special_buy')
-          left << h(RouteSelector) if @current_actions.include?('run_routes')
+          left << h(RouteSelector) if @current_actions.include?('run_routes') && !@step.alternate_selector?
```
tobymao, line 40: switch this to an if else?

tobymao: why would you call it alternate_selector when it's track conversion

assets/app/view/game/track_conversion.rb
```diff
@@ -0,0 +1,170 @@
...
+      include Lib::Settings
+
+      needs :routes, store: true, default: []
+      needs :selected_route, store: true, default: nil
+
+      # Get routes that have a length greater than zero
+      # Due to the way this and the map hook up routes needs to have
+      # an entry, but that route is not valid at zero length
+      def active_routes
+        @routes.select { |r| r.chains.any? }
```
tobymao, line 19: routes.reject chains.empty?

assets/app/view/game/track_conversion.rb
```diff
@@ -0,0 +1,170 @@
...
+        end
+
+        @routes.each(&:clear_cache!)
+
+        trains = trains.flat_map do |train|
+          onclick = lambda do
+            unless (route = @routes.find { |t| t.train == train })
+              route = Engine::Route.new(@game, @game.phase, train, abilities: @abilities, routes: @routes)
+              @routes << route
+              store(:routes, @routes)
```
tobymao, line 49: skip

assets/app/view/game/track_conversion.rb
```diff
@@ -0,0 +1,170 @@
...
+
+          invalid_props = {
+            attrs: {
+              colspan: '4',
+            },
+            style: {
+              padding: '0 0 0.4rem 0.4rem',
+            },
+          }
+          train_name = 'Segment:'
```
tobymao, line 98: don't need this var

assets/app/view/game/track_conversion.rb
```diff
@@ -0,0 +1,170 @@
...
+
+        h(:div, div_props, [
+          h(:h3, h3_text),
+          h('div.small_font', instructions),
+          train_help,
+          h(:table, table_props, [
+            h(:tbody, trains),
+          ]),
+          actions,
+        ].compact)
```
tobymao, line 130: why do you need compact here

assets/app/view/game/round/operating.rb
roseundy: OK, bad name. It was an alternative to the normal selector, but it still allows one to select a route. I'll come up with a better name.

assets/app/view/game/track_conversion.rb
roseundy: OK

assets/app/view/game/track_conversion.rb
roseundy: No reason. Just left over from leveraged code.

assets/app/view/game/round/operating.rb
roseundy: Switching to a flat if/else is difficult since I need to check for "run_routes" before I know I can call the alternate_selector (or whatever) method.

assets/app/view/game/track_conversion.rb
roseundy: I'm afraid I don't understand. You want a skip in the lambda?

assets/app/view/game/track_conversion.rb
roseundy: OK

assets/app/view/game/track_conversion.rb
roseundy: No, I think you want the page to be redrawn when you select a new route (this is the lamba associated with clicking on a train - the highlighting needs to change from one route to another).

tobymao: but you already call store(:selected_route) right under it?

roseundy: OK. Got it.

PR #5851, opened 2021-06-27
[1846] add "optional rule" to ensure Boomtown, Little Miami, and C&O are used

michaeljb:


lib/engine/game/g_1846/game.rb
```diff
@@ -319,6 +319,12 @@ def remove_from_group!(group, entities)
           # modify code which includes rand() w/o breaking existing games
           return if removals.empty?
 
+          if @optional_rules.include?(:second_ed_co)
+            until (removals & ['Boomtown', 'Little Miami', 'C&O']).empty?
```
tobymao, line 323: why don't you just remove them and then sort and take?

michaeljb: Fixed this, thanks

tobymao (approved):

PR #5852, opened 2021-06-28
Make a set of consistent and correct simple logos for 18VA

ventusignis:
Creates simple logos for 18VA
image


roseundy (approved):
Looks fine.

tobymao (approved):

PR #5872, opened 2021-06-30
Adds a Certificate Limit table to the info tab for games that provide the information

### LINKED ISSUE #5711: 1856: Show cert limit table for current player count
The cert limit in 1856 is dynamic after the first 6T is bought and changes based on the # of players and # of open corporations; this table should be displayed somewhere

ventusignis:
By default this is not enabled. It is enabled for 1856 to so that this fixes #5711 and 1836jr56 turns it back off.

Pictures (note, the table is filtered down to the player count)
image
image

### DIFF
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -33,8 +33,31 @@ module View
         children.concat(discarded_trains) if @depot.discarded.any?
         children.concat(phases)
         children.concat(timeline) if timeline
+        children.concat(cert_limit_table) if @game.cert_limit_table
         children.concat(endgame)
         children << h(GameMeta, game: @game)
+      end
+
+      def cert_limit_table
+        [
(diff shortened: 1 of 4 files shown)


ventusignis:
I could imagine 1817 and the like could use this to show how the number of remaining players affects the certificate limit, or how a corporation closing in 1846 lowers the certificate limit, or how no corporation nationalizing in 18MEX raises the cert limit

crericha:
Isn't this what rulebooks are for? I'm not saying I'm against this, but think it may be a bridge too far.

ChrisTheCat:
> Isn't this what rulebooks are for? I'm not saying I'm against this, but think it may be a bridge too far.

Another option might be the game wiki.

tobymao:
i think we should just have this in the wiki, don't think this info is that critical

PR #5877, opened 2021-06-30
Make it so that 1856 highlights when a corporation doesn't have treasury to pay interest

ventusignis:
The default behavior in the view is to append ' (Liquidate)' if the corporation doesn't have enough cash to pay the interest but that doesn't apply to 1856 so I refactored that into a method in base
Now 1856 highlights when interest is > treasury
image

1817** (tested in 1817WO) still works
image


tobymao (approved):

assets/app/view/game/dividend.rb
```diff
@@ -52,8 +52,8 @@ def render
               'None'
             end
 
-          if entity.loans.any? && !@game.can_pay_interest?(entity, corp_income)
-            text += ' (Liquidate)'
+          if entity.loans.any? && !@game.can_pay_interest?(entity, corp_income) && @game.cannot_pay_interest_warning_str
```
tobymao, line 55: mabye just call it cannot_pay_interest_str

ventusignis: I've changed the name

PR #5880, opened 2021-06-30
18Carolinas: routes and dividends

roseundy:
Implement running trains and dividends.
Adds concept of "variable" train lists for a corp. User controls train mix being run subject to constraint of current "train power" (which is hard-coded for the time being).

Common file changes:
UI::RouteSelector - add UI elements to add and delete trains, and to change the size of the currently selected train.
Engine::Game::Base - default methods for the above.

Example:
variable_trains

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -238,6 +238,7 @@ module View
         end
 
         reset_all = lambda do
+          @game.reset_adjustable_trains!(@routes)
           @selected_route = nil
           store(:selected_route, @selected_route)
           @routes.clear
@@ -256,6 +257,26 @@ module View
             routes: @routes.reject { |r| r.paths.empty? },
           )
           store(:routes, @routes)
+        end
+
+        add_train = lambda do
+          @game.add_route_train(@routes)
+          store(:routes, @routes)
+        end
+
+        delete_train = lambda do
+          @game.delete_route_train(@selected_route)
+          store(:routes, @routes)
+        end
+
+        increase_train = lambda do
+          @game.increase_route_train(@selected_route)
+          store(:selected_route, @selected_route)
+        end
+
+        decrease_train = lambda do
+          @game.decrease_route_train(@selected_route)
+          store(:selected_route, @selected_route)
         end
 
         submit_style = {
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_carolinas/game.rb
```diff
@@ -6,6 +6,8 @@
 require_relative 'map'
 require_relative 'step/buy_sell_par_shares_companies'
 require_relative 'step/track'
```
tobymao, line 8: i don't think you need any of these requires do you?

lib/engine/game/g_18_carolinas/game.rb
```diff
@@ -431,6 +454,100 @@ def check_route_token(route, token)
...
+        def delete_route_train(route)
+          train = route.train
+          @corporation_trains[train.owner].delete(train)
+        end
+
+        def increase_route_train(route)
+          train = route.train
+          corp = train.owner
+          return if train.distance == MAX_TRAIN
+          return if route.routes.map(&:train).sum(&:distance) >= @corporation_power[corp]
```
tobymao, line 537: routes.sum train.distance

lib/engine/game/g_18_carolinas/step/route.rb
```diff
@@ -0,0 +1,62 @@
...
+
+          def variable_distance?(_entity)
+            true
+          end
+
+          def add_train(routes)
+            @game.add_train(routes[0].corporation)
+          end
+
+          def remove_train(route)
```
tobymao, line 47: doesn't lok like these methods are used anywhere

assets/app/view/game/route_selector.rb
```diff
@@ -258,6 +259,26 @@ def actions(render_halts)
...
+          store(:routes, @routes)
+        end
+
+        delete_train = lambda do
+          @game.delete_route_train(@selected_route)
+          store(:routes, @routes)
+        end
+
+        increase_train = lambda do
+          @game.increase_route_train(@selected_route)
```
tobymao, line 273: did you implement these methods? i don't see them, also how can the UI change the game state, you need an action representing this don't you

lib/engine/game/g_18_carolinas/game.rb
roseundy: I wasn't sure. I'll delete.

lib/engine/game/g_18_carolinas/game.rb
roseundy: Yep.

lib/engine/game/g_18_carolinas/step/route.rb
roseundy: Yep. Vestige from previous iteration.

assets/app/view/game/route_selector.rb
roseundy: Yes. They are implemented in game.rb

The UI only changes game state indirectly. It manipulates the trains and their distances attached to the routes submitted to the RunRoutes action. When the action is processed the game state train distances are updated via the game method update_route_trains.

tobymao (approved):

PR #5881, opened 2021-06-30
18va private auction

ventusignis:


lib/engine/game/g_18_va/step/waterfall_auction.rb
```diff
@@ -0,0 +1,65 @@
...
+
+            @game.log << 'Players are reordered based on remaining cash'
+
+            # players are reordered from most remaining cash to least with prior order as tie breaker
+            current_order = @game.players.dup.reverse
+            @game.players.sort_by! { |p| [p.cash, current_order.index(p)] }.reverse!
+          end
+
+          def all_passed!
+            companies_without_bids = @companies.reject { |c| @bids.key?(c) && !@bids[c].empty? }
```
tobymao, line 40: @bids[c]

lib/engine/game/g_18_va/step/waterfall_auction.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/waterfall_auction'
+
+module Engine
+  module Game
+    module G18VA
+      module Step
+        class ForcedBid
```
tobymao, line 9: what is this

ventusignis: The waterfall auction code and view are expecting the bids for the corporations to be Bid actions (or at least objects that look like Bid actions) so this is a stubbed out version without the core Action code used. It's needed because when a company is reduced to a price of $0 in the auction, instead of the next player being forced to buy the company, they're forced to bid on it so this is used to represent their forced bid

lib/engine/game/g_18_va/step/waterfall_auction.rb
ventusignis: fixed

tobymao (approved):

PR #5887, opened 2021-07-01
1836jr56: Prevent buying a 3 then a 3'

ventusignis:
This might break some games, is there a way for me to verify this?


tobymao:
you'd need to run the full db

PR #5892, opened 2021-07-01
1889 Beginner Game rules

ventusignis:
For the additional tiles, I would've preferred to add tiles but it looks like the canonical way to use `optional_tiles` is to add the optional tiles to the roster and remove them in the method.

image
image
image

Side effect of having the beginner tiles in the manifest the way they are: upgrades are slightly more complicated:
image


scottredracecar:
Great idea to add this.

tobymao (approved):

PR #5896, opened 2021-07-01
1873: rounding issue on half dividend withholds

### LINKED ISSUE #5876: [Harzbahn 1873] Profit distribution is rounding incorrectly
Hotseat: https://gist.github.com/physic/9ea72a741f49691608ea6d789e9026e5

action=198

Rounding is never performed for monies withheld.
Rounding is performed after calculating the total dividend to a player.

See 4.3.2 Profit Distribution
> The director determines whether the profit is withheld,
> half distributed or completely distributed. When withheld,
> the bank pays the entire profit into the company’s
> treasury. In the case of half distribution, the bank pays
> half of the profit to the company’s treasury and the other
> half is distributed to the shareholders. Dividends are paid
> to the shareholders according to their respective shares in
> the company. If the distribution amount for a player is not
> an integer, the bank rounds in favour of the player to the
> next whole number.

Scenario: 2-share HW runs for 230 and pays half.
Expected:
The company receives **$115**.
A player with one share receives $57.5 rounded up = $58.
A player with both shares receives **$115**, no rounding.

Observed:
The company receives **$114**.
A player with one share receives $58.
A player with both shares receives **$116**.

roseundy:
Fixes #5876 

This seemingly minor change (+/- 1M to players/corps) killed two of my hotseat tests.

### DIFF
--- a/lib/engine/game/g_1873/step/dividend.rb
+++ b/lib/engine/game/g_1873/step/dividend.rb
@@ -155,6 +155,10 @@ module Engine
             @log << "#{entity.owner.name} receives #{@game.format_currency(revenue)}"
             @game.bank.spend(revenue, entity.owner)
           end
+
+          def half_pay_withhold_amount(entity, revenue)
+            revenue / 2
+          end
         end
       end
     end
--- a/lib/engine/game/g_1873/step/dividend.rb
+++ b/lib/engine/game/g_1873/step/dividend.rb
@@ -156,7 +156,7 @@ module Engine
             @game.bank.spend(revenue, entity.owner)
           end
 
-          def half_pay_withhold_amount(entity, revenue)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1873/step/dividend.rb
```diff
@@ -155,6 +155,10 @@ def payout_shares(entity, revenue)
             @log << "#{entity.owner.name} receives #{@game.format_currency(revenue)}"
             @game.bank.spend(revenue, entity.owner)
           end
+
+          def half_pay_withhold_amount(_entity, revenue)
+            revenue / 2
```
tobymao, line 160: does this work in ruby and javascript?

roseundy: It should (at least for 1873) since the revenue is always even.

Issue #2041, opened 2020-10-26
18GA: Consider adding Optional Rules


perwestling:
18GA has a plethoria of optional rules, and some are even just suggestions for the players to decide on the exact amount. So we might not want to support all 6. But adding support for some of the rules might give the game more replayability.

perwestling:
@ryandriskel I got an BGG email from the designer with some feedback. Regarding optional rules he wrote:

" I would appreciate it if you could consider adding an optional rule allowing for a soft rusting (delayed obsolescence) of the 4trains only. That was a major change in the version 2 rules (as a standard rule) and the hard rust option was removed. Another optional rule that would help is allowing all railroads the option of laying two yellow tiles on their first turn of operations (not just those RRs that float in SR1). I'm pretty sure I added that as a standard rule in version 2, but adding the soft rule is more important in my mind, at least."

ryandriskel:
I see both of these have been added. Are there others we want to add? I'd do it but I need specifics.

<img width="518" alt="Screen Shot 2022-02-27 at 8 24 06 AM" src="https://user-images.githubusercontent.com/15675400/155888493-b04d1354-2610-4b18-9881-9072e30eb8fc.png">

tobymao:
don’t need this issue

PR #5899, opened 2021-07-01
[18GA] Adds optional rules from rulebook

philcampeau:
adds optional rules from rulebook as per issue #2041


lib/engine/game/g_18_ga/meta.rb
```diff
@@ -16,6 +16,18 @@ module Meta
         GAME_RULES_URL = 'http://www.18xx.net/18GA/18GAr.txt'
 
         PLAYER_RANGE = [3, 5].freeze
+        OPTIONAL_RULES = [
+          {
+            sym: :soft_rust_t4,
```
ventusignis, line 21: should this be `:soft_rust_4t`

tobymao:
also build is broken

lib/engine/game/g_18_ga/meta.rb
philcampeau: fixed this

PR #5903, opened 2021-07-02
[1846] remove_from_group! - don't mutate the passed in group

### LINKED ISSUE #5883: [1846] Guaranteed C&O, Boomtown, and Little Miami not removing proper number of privates
When using the guarantee option, the game removes the normal number of privates for the 1st edition of the game (none for 5p, 1 each of blue & orange for 4p, 2 of each for 3p) then adds in Boomtown and Little Miami, instead of removing an additional one of each at every player count, then adding in the guaranteed privates. Removal of Corporations appears correct.

michaeljb:
[Fixes #5883]

### DIFF
--- a/lib/engine/game/g_1846/game.rb
+++ b/lib/engine/game/g_1846/game.rb
@@ -314,9 +314,10 @@ module Engine
         end
 
         def remove_from_group!(group, entities)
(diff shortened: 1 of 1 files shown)


michaeljb:
This breaks the existing games that have used the option, they'll need to be pinned.

tobymao (approved):

PR #5929, opened 2021-07-03
[1828.Games] Systems get 10x share price in their treasury

### LINKED ISSUE #5911: [1828.games] Merged company starts with incorrect treasury if floated in the exchange process after merging
https://18xx.games/game/48104?action=94

The CPR/MC system has floated, but its treasury should have $940, 10x the IPO, not $670. It appears that the code checks to see if the system has floated after the exchange process, because if in hotseat you do not exchange enough shares for the resulting system to float, its treasury is discarded and later when a 6th share is bought it is floated with $940, as expected. However, because when the merger takes place CPR/MC does not have enough shares to float, its treasury should be discarded before the share exchange process, and when a share is exchanged to float it, it should get the new IPO for its treasury.

crericha:
Provided that one of the merging corporations hasn't yet floated. Fixes #5911


crericha:
42531 needs a pin.

PR #5945, opened 2021-07-05
[18VA] add fancy logos, match BO and CO colors

benjaminxscott:
Added fancy logos and added the current simple logos behind the "simple logos" flag

New logos taken from copyright-kosher sources: [Virginian](https://github.com/18xx-maker/18xx-maker/blob/main/src/data/logos/Virginian.svg), [Western Maryland](https://github.com/18xx-maker/18xx-maker/blob/main/src/data/logos/1831/WMR.svg), and [Richmond, Fredericksburg & Potomac Railroad](https://en.wikipedia.org/wiki/Richmond,_Fredericksburg_and_Potomac_Railroad)

Also compressed the new SVGs

Before:
image

After:

image

image

image


ventusignis:
Thank you!

PR #5961, opened 2021-07-07
1830+ PRR Home - Halt implementation method (preview)

ventusignis:
I'm looking for feedback on this before I get too far on it

Autorouting looks like it works for PRR now and tokens work too

image
After clicking the halt for the second train
image

Halts are automatically skipped for other corps and autorouting works as expected
image

I tried some prototypes of using a city (megaltoona?) connected to all the exits on Altoona, but it gets far too crowded on Altoona on brown tiles that aren't the 3 pointed triangle. Having four halts on a brown tile isn't nearly as crowded and it may be possible to make the halts slimmer and PRR themed later if I go down this path

### DIFF
--- a/lib/engine/game/g_1830_plus/game.rb
+++ b/lib/engine/game/g_1830_plus/game.rb
@@ -9,6 +9,8 @@ module Engine
     module G1830Plus
       class Game < G1830::Game
         include_meta(G1830Plus::Meta)
+        attr_reader :prr_graph
+
         CERT_LIMIT = { 2 => 32, 3 => 27, 4 => 20, 5 => 16, 6 => 14, 7 => 13 }.freeze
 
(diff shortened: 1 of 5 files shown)


benjaminxscott:
Would the 'halt' rendering (the '0' thing) for payouts be removed?
(presumably that would just be a rendering flag for something like `fake_halt`)

Perhaps the "token spots" (the halts) could have the small PRR logo on them?

i.e. maybe we can make the token slot smaller so it fits better, or is that too much work to reinvent?

ventusignis:
I was thinking it may be possible to make a "home-halt" node specifically for this case where it's rendered like the PRR icon, but on the track. It might look like the "home-halts" block which would be incorrect though

tobymao:
close this out

PR #5962, opened 2021-07-07
Fix 1856 not allowing dits to downgrade

### LINKED ISSUE #5959: [1856] unable to replace town with plain track in endgame - 'New track must override old one' bug?
Per title, In an 1856 game I'm unable to replace single dit tiles with yellow plain track, for example placing #7 in K16 in the json example below.  

I understand that the rules allow such a placement, but the implementation currently gives the warning: 'New track must override old one'

game json: https://gist.github.com/tom-blight/c774c384d4d94165eb3c4c0ba2872863
game: https://18xx.games/game/45471
browser: chrome, windows

ventusignis:
A check was recently added to `tracker.rb` that makes sures that towns aren't removed during tile upgrades which of course prevents us from replacing town tiles with plain track tiles. Instead of altering the base code with flags I just copied it over into 1856's Track and removed the check
Fixes #5959

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -294,6 +294,10 @@ module Engine
 
       VARIABLE_FLOAT_PERCENTAGES = false
 
+      # Setting this to true is neccessary but insufficent to allow downgrading town tiles into plain track
+      # See 1856 for an example
+      ALLOW_REMOVING_TOWNS = false
+
       CACHABLE = [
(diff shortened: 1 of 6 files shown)


tobymao:
i think we should probably juts add a flag

ventusignis:
I've changed this to use a common flag

tobymao (approved):

Issue #5809, opened 2021-06-23
[1893] 'd' for 20% share is not shown on player charter


perwestling:
https://18xx.games/game/46448?action=128

Consider if a 'd' should be displayed for the 20% non-president share on the player charter.
This might be a general feature, not just 1893.

![image](https://user-images.githubusercontent.com/2631151/123088769-6f889580-d426-11eb-92e7-a13d82a691cb.png)

benjaminxscott:
What does `d` stand for in this case?

perwestling:
@benjaminxscott 'd' stands for double, ie the 20% non-president share, in an otherwise all 10% shares.
That is used in 1893 and 1849.
In 1893 the double share is used during merge. In 1849 the double share is the last one.

benjaminxscott:
Sounds like a good usecase, it seems a bit tricky since 

Currently the asterisk to denote president's share is pretty easy to grok, since it is "always the case" even if you have more than the presidency (e.g. `B&O* 40%` is easy to read that you have the presidency and two extra shares)

Perhaps similarly add the `d` after the asterisk for such cases?

So `B&O d 20%` or `B&O*d 40%` if they have the presidency

--
Related but different, perhaps games with special share sizes could define `status_str (corp)` which provides a note under the corp card

e.g. [in ches 2P](https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_18_chesapeake/game.rb#L563), there's a note that the presidency is a 30% share

![image](https://user-images.githubusercontent.com/1711810/124798901-c4d5a400-df08-11eb-9aa9-f03a23c3d864.png)

perwestling:
The d/* is presented on the corporation charters, see e.g. AGV in the screen shot above. I do not think it will be a problem to use a similar notation on the player charter as there are on the corporation charter.

Note! Size of presidency share does not matter as this is always shown as "*". It is just if there is double certificates besides the president's share.

benjaminxscott:
> The d/* is presented on the corporation charters
Ah, I was proposing to use `status_str` to denote "which share in the company will be a double share"

For instance, if I start the game how do I know that the last share is a double share?
(I could dig thru the rules, but that's the same deal with 2P Chesapeake, it's easier to just put it in `status_str`)

PR #5971, opened 2021-07-07
[1849] add note that last share is a double

benjaminxscott:
Proposal to add a note on each corp using the `status_str` method to denote that the last share of each corp is a double

This makes this mechanic more apparent to players

Checking with @perwestling @scottredracecar first

Presumably this would apply to other games, like 1893

Screenshot 2021-07-07 3 52 40 PM

Related to #5809


tobymao:
i don't think this is necessary

PR #5973, opened 2021-07-08
1817 perform sort, uniq on purchase options

### LINKED ISSUE #5934: 1817: Sort public company bids by value
As [seen here](https://18xx.games/game/46967?action=125) and [with gist](https://gist.github.com/benjaminxscott/440e4da71fd996b0921333a779cd752f), if a player has a restricted set of bids due to cash+privates the error message does not show the possible bids in order.  In the example given the message says possible bids are $160, $200, $150, etc. but since the message goes away fairly quickly it's easy to miss that the lowest possible bid is $150, not $160.

benjaminxscott:
Screenshot 2021-07-07 9 50 01 PM

closes #5934


benjaminxscott:
no idea what happened here

PR #5978, opened 2021-07-09
[1893] Add possibility to purchase EVA from other player

perwestling:
This commit makes it possible for the active player to purchase
the EVA company (for any amount - upto cash limit) at two times:
- During the stock round (is not counted as an action)
- During train purchase step

The purchaser is the active player, not the minor/corporation.
The purchaser pay.

Note! The rules does allow for buy/sell at any time, and this
would also be regarded as an action, for determination of first
player, but this implementation is a simplication that should
be good enough to at least give the players an option to buy EVA.

### DIFF
--- a/lib/engine/game/g_1893/buy_company.rb
+++ b/lib/engine/game/g_1893/buy_company.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require_relative '../../step/buy_company'
+
+module Engine
+  module Game
+    module G1893
+      module Step
+        class BuyCompany < Engine::Step::BuyCompany
+          def actions(entity)
+            return ['buy_company'] if purchasable_companies(entity).any?
+
+            []
+          end
+
+          def process_buy_company(action)
+            buyer = spender(action.entity)
+            company = action.company
+            company.min_price = 1
+            company.max_price = buyer.cash
+
+            buy_company(buyer, company, action.price, company.owner)
+          end
+
+          def purchasable_companies(entity = nil)
+            entity ||= @game.current_entity
+            @game.purchasable_companies(spender(entity))
+          end
+
+          # Who should pay for bought company? (Assuming company is EVA)
+          def spender(entity)
+            return unless entity
+
+            entity.player? ? entity : entity.player
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_1893/game.rb
+++ b/lib/engine/game/g_1893/game.rb
@@ -584,7 +584,7 @@ module Engine
           },
           {
             sym: 'EVA',
-            name: 'Eisenbahnverkehrsmittel Aktiengesellschaft',
+            name: 'EVA (Eisenbahnverkehrsmittel Aktiengesellschaft)',
             value: 150,
             revenue: 30,
             desc: 'Leaves the game after the purchase of the first 6-train. This private cannot be sold to '\
@@ -739,6 +739,14 @@ module Engine
           end
         end
 
+        # EVA can be bought from other player
+        def purchasable_companies(entity)
+          return [] if eva.closed? || eva.owner == @bank
+          return [eva] if eva.owner != entity
+
+          []
+        end
+
         # Show minors in spreadsheet
         def all_corporations
           @minors + @corporations
@@ -859,6 +867,7 @@ module Engine
             Engine::Step::Bankrupt,
             Engine::Step::DiscardTrain,
             Engine::Step::HomeToken,
+            G1893::Step::BuyCompany,
             G1893::Step::Track,
             Engine::Step::Token,
             Engine::Step::Route,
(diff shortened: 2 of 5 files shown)


lib/engine/game/g_1893/buy_eva_from_other_player.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module BuyEvaFromOtherPlayer
+  def buy_eva_from_other_player(entity, company, price)
+    raise GameError, "Only #{@game.eva.name} can be bought from other player" unless company == @game.eva
+
+    buyer = spender(entity)
+    @log << "Buyer #{buyer.name}, Entity #{entity.name}, Company #{company.name}"
+    owner = company.owner
+    company.owner = buyer
+    owner&.companies&.delete(company)
```
tobymao, line 11: why would owner be nil, if owner is nil line 15 will crash

lib/engine/game/g_1893/step/buy_train.rb
```diff
@@ -1,12 +1,31 @@
...
+
+            possible_actions << 'buy_company'
+          end
+
+          include BuyEvaFromOtherPlayer
+
+          # Buy EVA from other player
+          def process_buy_company(action)
+            company = action.company
+            raise GameError, "Only #{@game.eva.name} can be bought from other player" unless company == @game.eva
```
tobymao, line 24: why do you do this check twice

lib/engine/game/g_1893/buy_eva_from_other_player.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module BuyEvaFromOtherPlayer
+  def buy_eva_from_other_player(entity, company, price)
```
tobymao, line 4: why do you need this module? why don't just override process_buy_company

lib/engine/game/g_1893/step/buy_train.rb
```diff
@@ -1,12 +1,31 @@
...
         class BuyTrain < Engine::Step::BuyTrain
+          def actions(entity)
+            possible_actions = super
+
+            return [] if possible_actions.empty?
+
+            possible_actions << 'buy_company'
+          end
+
+          include BuyEvaFromOtherPlayer
```
tobymao, line 19: doesn't seem necessary

lib/engine/game/g_1893/buy_eva_from_other_player.rb
perwestling: Probably some cut & paste code from a context where company might have no owner. 
Right, removed &.

lib/engine/game/g_1893/step/buy_train.rb
perwestling: Right, not needed there.
Having the check in buy_eva_from_other_player might still be good as that method can be called from stock round code that might have other companies you can buy.

I put some guards in actions in buy_train.rb so that action is only available if a purchase is possible.

lib/engine/game/g_1893/buy_eva_from_other_player.rb
perwestling: Hm.... I think my plan was to use it from buy_sell_par_shares.rb in 1893 as well, but that does not seem to need it? Right, need to look into my solution again to see if I can do as you recommend.

lib/engine/game/g_1893/step/buy_train.rb
perwestling: I'll revisit my solution, and update the PR.

perwestling:
@tobymao Now I remember why I could not use process_buy_company directly. The view code is hardcoded to use current_entity as buyer. Which means that if player tries to buy private from other player, it will be the acting corporation that buys, and not the player.

So to make it possible to use I need to change this. One way of doing it is to change BuyValueInput to take buyer as an input parameter instead of using @game.current_entity. Then the step supporting player to player buy have its spender method which means it can set spender as buyer, while the other cases can set current entity as buyer.
(Unless there is some case where a player pays but a corporation buys? Not that I know of.)

tobymao:
yes that seems cleaner than this, this seems like a hack

perwestling:
@tobymao New attempt pushed

lib/engine/game/g_1893/buy_company.rb
```diff
@@ -0,0 +1,40 @@
...
+
+require_relative '../../step/buy_company'
+
+module Engine
+  module Game
+    module G1893
+      module Step
+        class BuyCompany < Engine::Step::BuyCompany
+          def actions(entity)
+            return ['buy_company'] if purchasable_companies(entity).any?
```
tobymao, line 11: unless empty

tobymao (approved):
much cleaner

PR #5987, opened 2021-07-10
Allow setting the seed manually when creating a game (api only)

daniel-sousa-me:
We want to use this to have the same setup for everyone in the 18Zoo tournament

### DIFF
--- a/routes/game.rb
+++ b/routes/game.rb
@@ -171,7 +171,7 @@ class Api
             description: r['description'],
             max_players: r['max_players'],
             settings: {
-              seed: Random.new_seed % 2**31,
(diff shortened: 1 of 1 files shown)


routes/game.rb
```diff
@@ -171,7 +171,7 @@ class Api
             description: r['description'],
             max_players: r['max_players'],
             settings: {
-              seed: Random.new_seed % 2**31,
+              seed: r['seed'] || Random.new_seed % 2**31,
```
tobymao, line 174: make it (r['seed'] || Random.new_seed) % 2 ** 31 so that we can sanitize the input a bit

PR #5990, opened 2021-07-10
Implementation of Bond concept

perwestling:
- Issuer is a class on the same level as Corporation, Company
- Bond is a class on the same level as Share
- An Issuer can have a number of Bonds
- Presented on stock view if title has any Issuer
- Presented on Entity view if title has any Issuer
- Bonds Bought/Sold following share rules, using its price
  but no limit on how many bonds a player can own or sell
- Pays revenue to player owner each OR
- Bonds and Issuers shown on Spreadsheet, if title has any Issuer
- Bonds (if any own) shown on player chart

### DIFF
--- a/assets/app/view/game/auto.rb
+++ b/assets/app/view/game/auto.rb
@@ -23,6 +23,7 @@ module View
         if @game.players.find { |p| p.name == @user&.dig('name') }
           types = {
             Engine::Action::ProgramBuyShares => ->(settings) { render_buy_shares(settings) },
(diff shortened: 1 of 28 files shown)


perwestling:
1893 have one Issuer, AdSK.

Game view in 1893 during SR:
image

Entity view at the same time:
image

Spreadsheet that shows Bond count for the players, and a row for AdSK:
image

Note that Value for players includes value of Bonds.

Example on revenue payout at start of OR:
image

Example how the log looks like when buying/selling bonds during SR:
image

PR #5993, opened 2021-07-10
Apply freight bonus in 18VA to port CITY as well

### LINKED ISSUE #5991: [18VA] Doubling of Port/bonus revenue calculated incorrectly
The value of a port bonus and the city it is connected to should be doubled when a G train runs to them. Currently, .games only doubles the value of the port bonus, not the city itself. In this example https://18xx.games/game/48858?action=246 B&O should earn $110 from the 2G (G4-H5-I6) (30 + 2(20+20)), but it instead earns 90 because Annapolis is not doubled, only the port on I6.

Relevant rules section 4.5.4.2: 
"If the Corporation has a station token in the Port, the value of the City is increased by the Port bonus shown as a “+”
amount. Additionally, if the Corporation has a station token in the Port, and is running a Freight Train (1G, 2G, 3G, 4G or
5G) to the Port, the total value of the City, including the Port bonus, is doubled."

ventusignis:
Fixes #5991

### DIFF
--- a/lib/engine/game/g_18_va/game.rb
+++ b/lib/engine/game/g_18_va/game.rb
@@ -738,11 +738,19 @@ module Engine
             revenue += offboard_stop.route_revenue(@phase, train)
           end
           # Freight and doubler trains double ports
-          revenue += port_stop.route_revenue(@phase, train) if port_stop && %i[freight doubler].include?(train_type)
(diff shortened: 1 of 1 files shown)


pstone1:
Was this fix applied? It isn't working in https://18xx.games/game/48858 and it does not say it is pinned

PR #6007, opened 2021-07-13
1862: fix display of PD in Parliament rounds

### LINKED ISSUE #6004: 1862 - Parliament Round Priority Deal
If a player acts in the parliament round, it changes where the priority deal marker sits on the players entities cards. During parliament round PD should not change as it reverts back to the original Priority owner at the start of the stock round.

roseundy:
Fixes #6004 

UI only, no pins required.

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -637,7 +637,6 @@ module Engine
           players = @players.reject(&:bankrupt)
 
           if @round.is_a?(Engine::Game::G1862::Round::Stock)
-            puts "@round: #{@round}, class: #{@round.class.name}"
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1862/game.rb
```diff
@@ -633,6 +633,26 @@ def remove_marker(corporation)
           hex.tile.icons.delete(marker) if marker
         end
 
+        def priority_deal_player
+          players = @players.reject(&:bankrupt)
+
+          if @round.is_a?(Engine::Game::G1862::Round::Stock)
+            puts "@round: #{@round}, class: #{@round.class.name}"
```
tobymao, line 640: puts

PR #6020, opened 2021-07-14
Add Shares to the player card based off the spreadsheet calculation

ventusignis:
image
image
image

### DIFF
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -3,12 +3,14 @@
 require 'lib/settings'
 require 'lib/text'
 require 'view/game/companies'
+require 'view/share_calculation'
 
 module View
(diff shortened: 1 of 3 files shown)


assets/app/view/game/player.rb
```diff
@@ -252,6 +256,12 @@ def render_minors(minors)
         }
         h('div', outer_props, [h('div', inner_props, minor_logos)])
       end
+
+      def num_shares_of(entity, corporation)
```
tobymao, line 260: don't copy and paste code, make this a module

ventusignis: I made this a module underneath `app/view` where both of the uses are from

tobymao (approved):

Issue #5966, opened 2021-07-07
[1862] Highlight trains that match company permit


PilotBob:
**What's your idea?**

It seems that the train selection is inverted during buy Trains. It makes more sense that the trains a company has a permit for should be highlighted rather than the inverse of this.

benjaminxscott:
Hmm, I personally prefer the way it is based on:
- in other games, no highlighting on a train implies it's fine to buy
- adding highlighting implies a warning when buying, so that's good

Combined, those seem like they work pretty well here, but I'm not the decider just my 2 satoshis

jenf:
Closing as won't do.

dfannius:
My suggestion would be to change the color of the highlight away from yellow, which feels a little more like a good thing than a bad thing. A red/pink sort of color has unfortunately already been taken by the Priority Deal marker, but a gray "highlight" might make it more clear that these trains are a bad idea. Another possibility would be to just gray out the text itself.

PR #6029, opened 2021-07-15
[1862, others] make inoperable trains gray while buying

benjaminxscott:
I like this suggestion from @dfannius

Screenshot 2021-07-15 8 02 11 AM

related to #5966


roseundy:
Fine with me.

ventusignis:
Would expanding this to 18CZ be useful?

benjaminxscott:
Yeah, CZ just has to implement `able_to_operate?` which could return `train_of_size?(train, entity.type)` afaik

PR #6032, opened 2021-07-16
18CZ: Highlight trains a corp cannot run

ventusignis:
This should help prevent the trap of people buying a `4e` when they want a `3E` ;)
image
image

### DIFF
--- a/lib/engine/game/g_18_cz/game.rb
+++ b/lib/engine/game/g_18_cz/game.rb
@@ -325,6 +325,10 @@ module Engine
           @timeline.append("Next set of Operating Rounds will have #{OR_SETS[@turn - 1]} ORs")
         end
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_cz/game.rb
```diff
@@ -325,6 +325,21 @@ def timeline
           @timeline.append("Next set of Operating Rounds will have #{OR_SETS[@turn - 1]} ORs")
         end
 
+        def train_type_by_name(name)
```
Zwergenpunk, line 328: There is a function to map train to corp type. Im just mobile so I cannot link direkt, but toby will not approve regexes

ventusignis: Thanks, I found the TRAINS_FOR_CORPORATIONS map

PR #6036, opened 2021-07-16
In timing issues w/ 1856, if an incorrect pass gets given to Track, escrow does not get passed.

### LINKED ISSUE #6035: Game not loading - 47224
**How to Reproduce the Problem?**
Try going to game 47224

I used the Undo button and the screen failed to refresh.

e.g.
- Trying to login with my username results in "Access Denied"

**Which Browser / Platform?**
- Browser Version
  - On Chrome, see [chrome://version](chrome://version)
  - On Firefox, see [about:support](about:support)
  - On Safari, see [these methods](https://www.lifewire.com/check-the-version-number-of-safari-446563)
- Platform (Mobile, Tablet, Desktop)

ventusignis:
This provides a description for Escrow which notifies the user to undo. This is somewhat possible in 1856
because the game takes somewhat long (~2 seconds by the end of the game) to replay for an undo so a race
condition with the server is possible.

FIxes #6035 
This adds a description to a step so it will not need pins or archiving games.

### DIFF
--- a/lib/engine/game/g_1856/step/escrow.rb
+++ b/lib/engine/game/g_1856/step/escrow.rb
@@ -20,6 +20,10 @@ module Engine
             ]
           end
 
+          def description
+            'Escrow Calculation - Undo if you see this'
+          end
+
           def actions(_entity)
             ACTIONS
           end


tobymao:
Why don't you disallow an illegal action from happening?

ventusignis:
I changed it to validate in the Track action instead; when merging this can you also delete the last action in game #47224 from the DB? (action ID 1079)

ventusignis:
I reverted the change because the auto action sometimes the auto action goes on the `Tile` step's action, and sometimes it goes on the `Token` step's action. An example of this is if the Token step is skipped due to there being nowhere a token could legally be placed, the Tile step gets the auto action. To have the Tile action properly validate, it would need to know if the token step would get skipped or not

 At this point, letting the player back out in this unlikely scenario seems relatively clean.

tobymao (approved):

Liturgista:
Thanks. Problem fixed!

PR #6040, opened 2021-07-17
18 usa privates batch2

ventusignis:
This batch is the bridge, oil, coal, and iron privates in 18USA

I made some special tiles for coal, oil, and iron in the game because the `tile_lay` ability was more natural to work with than `assign` when dealing with how the iron token "can be flipped with a tile upgrade action" in green phase and tiles with oil tokens can be upgraded and automatically update their revenue. 

image
image

### DIFF
--- a/lib/engine/game/g_18_usa/step/track.rb
+++ b/lib/engine/game/g_18_usa/step/track.rb
@@ -12,30 +12,7 @@ module Engine
           include Engine::Step::UpgradeTrackMaxExits
 
           def check_track_restrictions!(entity, old_tile, new_tile)
-            return if @game.loading || !entity.operator?
-
-            graph = @game.graph_for_entity(entity)
-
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_usa/step/track.rb
```diff
@@ -0,0 +1,44 @@
...
+  module Game
+    module G18USA
+      module Step
+        class Track < Engine::Step::Track
+          include Engine::Step::UpgradeTrackMaxExits
+
+          def check_track_restrictions!(entity, old_tile, new_tile)
+            return if @game.loading || !entity.operator?
+
+            graph = @game.graph_for_entity(entity)
```
tobymao, line 17: what's different about this ?

ventusignis: This is to support the case of upgrading the iron tiles where there is no town/city to change revenue and no new track to connect to.  I've replaced this with a simpler version that references `super`

tobymao (approved):

PR #6067, opened 2021-07-21
18Carolinas: more pre-alpha fixes

roseundy:
After multiple playthroughs with Scott, I made several minor fixes, plus:
- I redesigned how "variable" train rosters work. To be consistent with the rest of the code base, I now manipulate the corporation.trains list, moving trains to/from the depot as needed. Along with this, the depot now contains the maximum number of trains possibly needed at each size. This makes the autorouter much happier and also avoids the need to change train distance on the fly. This required a new method in Route to set the train.
- I redesigned how track conversion mode works. It is now a "choice" action and a new step. Visually, it looks much the same, but the code is cleaner and works with the server much better.
- When I discovered that the game needed dual-gauge track for its intended purpose, I switched to using narrow gauge track for Southern track, which led to the following item
- I added the ability via instance variables in Path to ignore track type (gauge) when walking paths and comparing paths. This affected Path and Tile.

### DIFF
--- a/lib/engine/game/g_18_carolinas/game.rb
+++ b/lib/engine/game/g_18_carolinas/game.rb
@@ -484,6 +484,7 @@ module Engine
             G18Carolinas::Step::Bankrupt,
             Engine::Step::HomeToken,
             G18Carolinas::Step::Track,
+            G18Carolinas::Step::ConvertTrack,
             G18Carolinas::Step::Token,
             G18Carolinas::Step::Route,
             G18Carolinas::Step::Dividend,
@@ -702,6 +703,7 @@ module Engine
           routes.each do |route|
             next if route.visited_stops.empty?
 
+            route.train.owner = entity
             entity.trains << route.train
           end
           @corporation_trains[entity] = nil
@@ -819,6 +821,7 @@ module Engine
             raise GameError, 'Route must have Southern track' unless route.paths.any? { |p| p.track != :broad }
           else
             raise GameError, 'Train below minimum size' if route.train.distance < min_train
+            raise GameError, 'Train w/o owner' unless route.train.owner
 
             if route.routes.sum { |r| r.train.distance } > loan_or_power(route.train.owner)
               raise GameError, 'Train sizes exceed train power'
--- a/lib/engine/game/g_18_carolinas/game.rb
+++ b/lib/engine/game/g_18_carolinas/game.rb
@@ -434,12 +434,10 @@ module Engine
           @power_progress = 0
 
           @bankrupted = {}
-        end
 
-        def init_graph
-          Engine::Part::Path.ignore_gauge_walk = true
-
-          super
+          @all_tiles.each { |t| t.ignore_gauge_walk = true }
+          @_tiles.values.each { |t| t.ignore_gauge_walk = true }
+          @graph.clear_graph_for_all
         end
 
         def trains
@@ -604,7 +602,8 @@ module Engine
 
           return unless @phase.name == '5'
 
-          Engine::Part::Path.ignore_gauge_compare = true
+          @all_tiles.each { |t| t.ignore_gauge_compare = true }
+          @_tiles.values.each { |t| t.ignore_gauge_compare = true }
           upgrade_c_hexes
           @graph.clear_graph_for_all
         end
--- a/lib/engine/game/g_18_carolinas/game.rb
+++ b/lib/engine/game/g_18_carolinas/game.rb
@@ -279,6 +279,8 @@ module Engine
         MAX_PROGRESS = 15
 
         C_TILES = %w[C1 C2 C3 C4 C5 C6 C7 C8 C9].freeze
+        CHARLOTTE_HEX = 'D10'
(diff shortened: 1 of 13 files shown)


lib/engine/game/g_18_carolinas/step/track.rb
```diff
@@ -8,76 +8,75 @@ module G18Carolinas
...
-          def conversion?
-            true
-          end
-
-          def available_hex(entity, hex)
-            return super if @mode == :new_track
-
-            @game.graph_for_entity(entity).reachable_hexes(entity)[hex]
+          def legal_tile_rotation?(entity, hex, tile)
+            return super unless @game.phase.available?('5')
```
tobymao, line 59: this looks copy and pasted, what did you change?

lib/engine/part/path.rb
```diff
@@ -153,10 +156,10 @@ def walk(
           neighbor.paths[np_edge].each do |np|
             next if on && !on[np]
             next unless lane_match?(@exit_lanes[edge], np.exit_lanes[np_edge])
-            next unless tracks_match?(np, dual_ok: true)
+            next unless any_track || tracks_match?(np, dual_ok: true)
```
tobymao, line 159: don't use unless with multiple statements

lib/engine/route.rb
```diff
@@ -4,8 +4,8 @@
 
 module Engine
   class Route
-    attr_accessor :halts, :routes
-    attr_reader :last_node, :phase, :train, :abilities
+    attr_accessor :halts, :routes, :train
```
tobymao, line 7: when the train is set do you need to clear the cache?

lib/engine/game/g_18_carolinas/step/track.rb
roseundy: This line: old_paths.all? { |path| new_paths.any? { |p| @game.path_subset?(path, p) } }

roseundy: I was hesitant to modify tracker to pull this out as an overrideable method, but I can if you think it best.

lib/engine/part/path.rb
roseundy: OK

lib/engine/route.rb
roseundy: No, I don't think it's required, but it can't hurt either.

lib/engine/game/g_18_carolinas/step/track.rb
tobymao: i think when you do copy and pasting and have to override other places like track_match? it probably indicates you're doing something wrong

all of this stuff is really just because you're trying to skip tracks_match? you could make this all much cleaner by setting a new path on variable to ignore tracks_match?, then you wouldn't have to change path walk and also copy and paste this method

let's make this change

lib/engine/route.rb
tobymao: in that case remove the accessor and do

def train=(new_train)
   ...

lib/engine/game/g_18_carolinas/game.rb
```diff
@@ -410,7 +510,46 @@ def upgrades_to?(from, to, special = false, selected_company: nil)
           # handle C tiles specially
           return false if from.label.to_s == 'C' && to.color == :yellow && from.cities.size != to.cities.size
 
-          super
+          # correct color progression?
+          return false unless Engine::Tile::COLORS.index(to.color) == (Engine::Tile::COLORS.index(from.color) + 1)
```
tobymao, line 514: again, don't copy and paste huge chunks of code like this

lib/engine/game/g_18_carolinas/step/track.rb
roseundy: Except I don't always want to skip tracks_match?, or I'm not understanding your suggestion.

lib/engine/route.rb
roseundy: OK

lib/engine/game/g_18_carolinas/game.rb
roseundy: OK, I'll make this check overriddable.

lib/engine/part/path.rb
```diff
@@ -62,13 +62,29 @@ def initialize(a, b, terminal: nil, lanes: LANES, track: :broad, ignore: nil)
         separate_parts
       end
 
+      def ignore_gauge_compare
```
tobymao, line 65: no, we shouldn't use class variables

roseundy:
Revamped to avoid code duplication or modifying Node, Route and Graph by adding class variables to Path to allowing ignoring track (gauge). Separately controllable for Path walks vs. Path compares.

lib/engine/part/path.rb
roseundy: What's the alternative? Instance variables were not working - too hard to find all Path instances being used.

roseundy:
And revamped again to avoid class variables.

tobymao (approved):

PR #6068, opened 2021-07-21
[18Rhl] Disallow double visit to Rhine Metropolis

perwestling:
When Duisburg, Düsseldorf and Köln are yellow hexes, they have three
separate city slots. Two of these are on one side of Rhine, and the
third is on the other. It is not allowed to include the two city
slots on one side of Rhine in the same route (as it is the same city).
Having a route from one side of Rhine to the other side of Rhine
and visiting the different sides of the same hex, is allowed though.

When the Rhine Metropolis hex is upgraded to green (or later) the
hex will be a normal "one city - three slots" and then the normal
game logic will stop the hex being visited twice in one route.

This is a 2nd attempt to implement this feature - the 1st attempt
was reverted, as it need to be reviewed, and the 1st attempt also
did not work as it made routes through green Rhine metropolis cities
illegal.

### DIFF
--- a/lib/engine/game/g_18_rhl/game.rb
+++ b/lib/engine/game/g_18_rhl/game.rb
@@ -737,6 +737,8 @@ module Engine
 
         RGE_HEXES = %w[A4 A6 L11 L13].freeze
 
+        RHINE_METROPOLIS_HEXES = %w[D9 F9 I10].freeze
+
         def num_trains(train)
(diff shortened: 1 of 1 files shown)


perwestling:
I did revert the previous commit for this check, and have now made a new attempt.

tobymao (approved):

PR #6069, opened 2021-07-21
[18Rhl] Implement all route bonuses

perwestling:
Also correct so that 8 train does not count towns.
And use a Trajekt icon for the green Rhine Metropolis cities.


lib/engine/game/g_18_rhl/game.rb
```diff
@@ -1029,7 +1077,73 @@ def check_distance(route, visits)
...
 
-          raise GameError, 'Route for 8 trains cannot include any off-board hexes besides RGE ones'
+          raise GameError, 'Route for 8 trains cannot include any off-board hexes besides the RGE ones'
+        end
+
+        def revenue_for(route, stops)
+          revenue = super
+          corporation = route.train.owner
+
+          [off_board_out_tokened_penalty(route, stops, corporation),
```
tobymao, line 1087: this line and 1105, maybe turn this into a method so you can reuse in both places

perwestling: Done

PR #6071, opened 2021-07-21
If an entity has a train_discount ability, factor that in for train min_price

### LINKED ISSUE #5983: [18VA] train buying step auto-skipped even with P2 train discount
**edit: if the company has zero dollars going into the train-buying step**, it looks like the P2 Tredegar Iron Works one-time discount of $200 on a train purchase from the bank is not being factored into the logic that decides whether to auto-skip the train-buying step.

At [this step](https://gist.github.com/benjaminxscott/86ccc0506724499cfa7ec8a448fe1f36) the N&W has converted to a ten-share company (so its train limit is 4), currently has two trains, and purchased P2 before the train-buying step. It buys in the P2 for all of its money (50 bucks)

The train-buying step is being auto skipped since N&W has zero treasury money, even though it could obtain the next train from the depot for free using its $200 P2 discount

ventusignis:
Fixes #5983

### DIFF
--- a/lib/engine/train.rb
+++ b/lib/engine/train.rb
@@ -72,8 +72,11 @@ module Engine
       "#{@sym}-#{@index}"
     end
 
-    def min_price
-      from_depot? ? @price : 1
+    # if set ability must be a :train_discount ability
+    def min_price(ability: nil)
+      return 1 unless from_depot?
+
+      ability&.discounted_price(self, @price) || @price
     end
 
     def from_depot?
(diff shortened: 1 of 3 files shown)


lib/engine/train.rb
```diff
@@ -72,8 +72,11 @@ def id
       "#{@sym}-#{@index}"
     end
 
-    def min_price
-      from_depot? ? @price : 1
+    # if set ability must be a :train_discount ability
+    def min_price(ability: nil)
+      return from_depot? ? @price : 1 unless ability
```
tobymao, line 77: return 1 unless from_depot?

ability&.discounted_price(self, @price) || @price

ventusignis: changed

tobymao (approved):

PR #6080, opened 2021-07-22
Allow for colored borders; use it in 1836jr to make impassable borders blue

### LINKED ISSUE #3269: [1836jr30] Show red borders from waterways in blue
The game of 1836jr30 is set in the Low Countries Belgium and The Netherlands. The board has some blue tiles showing port connections.

There are also some borders on the board that cannot be crossed, shown in red. These are between C7-9, and between F4-6 and G3-5.

These borders actually represent fairly big waterways that cannot be crossed by bridges. Is it possible to change these red lines to blue lines to show they are waterways? Maybe with a thicker line to denote they can not be crossed?

ventusignis:
I also made impassable borders slightly thicker to make them look different from passable borders. We may want to emphasize the difference a little more in a future PR

Fixes #3269 
(I think I have different colors set up on my local v 18xx.games hence the difference in shades of blue)
before
image

after
image

### DIFF
--- a/assets/app/view/game/part/borders.rb
+++ b/assets/app/view/game/part/borders.rb
@@ -54,6 +54,8 @@ module View
         }.freeze
 
         def color(border)
+          return setting_for(border.color) if border.color
+
           color =
             case border.type
             when :mountain
@@ -65,6 +67,10 @@ module View
             end
 
           setting_for(color)
+        end
+
+        def border_width(border)
+          border.type == :impassable ? '10' : '8'
         end
 
         def render_cost(border)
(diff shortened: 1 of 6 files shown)


perwestling:
How to separate a passable water border with an impassable? 

Hm, I have this in 18Rhl, but I suppose I can change the 
"30 upgrade for water, plus a stub towards a blue - passable border" to a
"0 upgrade for water, plus a stub towards a blue - passable border with 30 cost when built to". 
Yes, that should work. 

Unless one could have different blue colors? Or dashed? 
Nah, I think I will try my idea (as soon as this is on master).

(It is for 18Rhl, BTW, which have a possible bridge betwee C8 and B9 (Wesel), while all other blue edges are impassible.)

PR #6082, opened 2021-07-23
[18Rhl] Complete implementation of all company abilities

perwestling:
Some minor changes was needed in general code to cover the case when
a player owned company performs a tokening - as the tokener is
then not the comany, or company owner, but the current entity
(corporation) that is operating. This should not affect existing
titles, as a check is done if owner is player.

### DIFF
--- a/lib/engine/game/g_18_mex/game.rb
+++ b/lib/engine/game/g_18_mex/game.rb
@@ -69,7 +69,7 @@ module Engine
           '470' =>
           {
             'count' => 1,
-            'color' => 'orange',
+            'color' => 'yellow',
             'code' =>
             'town=revenue:20,loc:0;path=a:0,b:_0;path=a:3,b:_0;path=a:3,b:4;path=a:4,b:_0;label=CC',
           },
--- a/lib/engine/game/g_18_mex/game.rb
+++ b/lib/engine/game/g_18_mex/game.rb
@@ -69,7 +69,7 @@ module Engine
           '470' =>
           {
             'count' => 1,
-            'color' => 'yellow',
+            'color' => 'orange',
             'code' =>
             'town=revenue:20,loc:0;path=a:0,b:_0;path=a:3,b:_0;path=a:3,b:4;path=a:4,b:_0;label=CC',
           },
--- a/lib/engine/game/g_18_rhl/step/lay_tile_checks.rb
+++ b/lib/engine/game/g_18_rhl/step/lay_tile_checks.rb
@@ -5,6 +5,7 @@ module LayTileChecks
     super
 
     @game.eastern_ruhr_connection_check(action.tile.hex)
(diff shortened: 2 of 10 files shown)


lib/engine/game/g_18_mex/game.rb
```diff
@@ -69,7 +69,7 @@ class Game < Game::Base
           '470' =>
           {
             'count' => 1,
-            'color' => 'yellow',
+            'color' => 'orange',
```
tobymao, line 72: 18mex?

lib/engine/game/g_18_rhl/step/lay_tile_checks.rb
```diff
@@ -5,5 +5,13 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
     super
 
     @game.eastern_ruhr_connection_check(action.tile.hex)
+    return if action.tile.hex.name != 'E14' || !@game.prinz_wilhelm_bahn
+
+    # Remove hex block
+    comp = @game.prinz_wilhelm_bahn
+    comp.all_abilities.each do |a|
```
tobymao, line 12: do you need to dup here or no

lib/engine/step/special_token.rb
```diff
@@ -9,6 +9,7 @@ class SpecialToken < Base
       include Tokener
 
       def actions(entity)
+        return [] unless ability(entity)
```
tobymao, line 12: why did you add this, it's the same as line 13

lib/engine/step/tokener.rb
```diff
@@ -33,6 +33,8 @@ def token_cost_override(_entity, _city_hex, _slot, _token)
 
       def available_tokens(entity)
         token_holder = entity.company? ? entity.owner : entity
+        return [] if token_holder.player?
```
tobymao, line 36: is this correct?

lib/engine/game/g_18_mex/game.rb
perwestling: Yeah, 18MEX has a similar orange special tile, but it broke 18MEX so that can be added in a separate commit. Maybe.
Will add a commit that removes this change.

lib/engine/step/special_token.rb
perwestling: Right - removed.

lib/engine/step/tokener.rb
perwestling: Does not seem to be needed after all, so I have now removed this change.

lib/engine/game/g_18_rhl/step/lay_tile_checks.rb
perwestling: Probably safe in this case as it only contains one ability... but sure adding dup is fine, so I am doing that.

PR #6089, opened 2021-07-24
[18Rhl] Implement RhE bonus when link exists Aachen-Duren-Cologne

perwestling:
Also add a reference game.

### DIFF
--- a/lib/engine/game/g_18_rhl/game.rb
+++ b/lib/engine/game/g_18_rhl/game.rb
@@ -843,12 +843,12 @@ module Engine
 
         def operating_round(round_num)
           G18Rhl::Round::Operating.new(self, [
-            G18Rhl::Step::RheBonusCheck,
             G18Rhl::Step::Bankrupt,
             Engine::Step::HomeToken,
             G18Rhl::Step::SpecialToken, # Must be before any track lay (due to private No. 4)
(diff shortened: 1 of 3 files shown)


tobymao (commented):
you need to keep the return if loading since connection check hits graph, you need to use auto actions

perwestling:
@tobymao If I keep the "return if loading" this breaks the code. 

What happens is that RhE lays the last tile that satifies the bonus, the check returns without checking the bonus, and now the corproation does not have money to do a token - so the game is out of synch.  It fails on action 201.

I am not sure how to solve this - can I get it to wait until graph has been loaded? 
What do you mean by auto action?

It need to take place between lay tile action and token action.

tobymao:
There’s a concept of auto actions for this use case, where you persist an
action so that you don’t need to check the graph. See 1870 destination as
an example.

perwestling:
Ok, I will check it out.

perwestling:
@tobymao Have now made an attempt. It is not perfect, as it does not do the check directly after the tile lay, so it wont give RhE the money direct after it creates the link, but at least it gives close to the required. RhE will have to wait until next OR to spend the money.

tobymao:
why not do it correctly?

tobymao (approved):

PR #6106, opened 2021-07-26
[18SJ] Add special 2 player variant

perwestling:
This 2 player variant adds a 3rd player, a bot, that is
partly automatic, and partly handled by the players, via
master mode. The bot is excluded from the value/result.

More details in the 18SJ rule book.

### DIFF
--- a/lib/engine/game/g_18_sj/step/buy_train.rb
+++ b/lib/engine/game/g_18_sj/step/buy_train.rb
@@ -20,16 +20,6 @@ module Engine
 
           def do_after_buy_train_action(_action, _entity); end
 
-          def buyable_trains(entity)
-            if entity.player == @game.edelsward
-              result = @depot.depot_trains[0..1]
-              puts result.to_s
-              result
-            end
-
-            super
-          end
-
           def must_buy_train?(entity)
             return false if entity.player == @game.edelsward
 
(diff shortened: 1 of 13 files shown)


lib/engine/game/g_18_sj/step/buy_train.rb
```diff
@@ -19,6 +19,28 @@ def actions(entity)
           end
 
           def do_after_buy_train_action(_action, _entity); end
+
+          def buyable_trains(entity)
+            if entity.player == @game.edelsward
+              result = @depot.depot_trains[0..1]
+              puts result.to_s
```
tobymao, line 26: remove debug

lib/engine/game/g_18_sj/step/buy_train.rb
```diff
@@ -19,6 +19,28 @@ def actions(entity)
           end
 
           def do_after_buy_train_action(_action, _entity); end
+
+          def buyable_trains(entity)
+            if entity.player == @game.edelsward
+              result = @depot.depot_trains[0..1]
```
tobymao, line 25: result is never used?

lib/engine/game/g_18_sj/step/buy_train.rb
perwestling: Fixed

lib/engine/game/g_18_sj/step/buy_train.rb
perwestling: This code was a remain from when I tried to get this to work.

The final way was that as part of the cleanup after dividend, a corporation owned (requisited 100%) by the bot (Edelswärd) automatically gets 0-2 trains without using buy_trains. So a bot will not use that step.

PR #6138, opened 2021-07-29
[18SJ] Correct bot check (fixes #6137)

### LINKED ISSUE #6137: [18SJ] Corporations requisit trains even when not bot
https://18xx.games/game/51087?action=51

In this game KFJ requisit a train. But it is only bot that should do this (Edelswärd, in two player variant)

perwestling:
A serious bug in 18SJ. If this correction breaks any games those games are not salvable, and will have to be deleted.

### DIFF
--- a/lib/engine/game/g_18_sj/game.rb
+++ b/lib/engine/game/g_18_sj/game.rb
@@ -846,6 +846,10 @@ module Engine
           @gc ||= company_by_id('GC')
         end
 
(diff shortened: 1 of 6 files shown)


perwestling:
Am working on (re)adding a reference game for 18SJ - to stop bugs like this from being introduced.

tobymao (approved):

PR #6155, opened 2021-08-01
18usa privates batch4

ventusignis:
Adds the two train running shenanigans privates to the game.
This will make it 18/30 done
This also adds the pullman train and the ability to add it to a train as an ability (like the two new privates)

### DIFF
--- a/lib/engine/graph.rb
+++ b/lib/engine/graph.rb
@@ -10,6 +10,7 @@ module Engine
       @connected_nodes = {}
       @connected_paths = {}
       @reachable_hexes = {}
+      @jump_connected_hexes = {}
       @tokenable_cities = {}
       @routes = {}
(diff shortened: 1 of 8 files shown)


lib/engine/graph.rb
```diff
@@ -23,7 +24,7 @@ def clear
       @connected_hexes.clear
       @connected_nodes.clear
       @connected_paths.clear
-      @reachable_hexes.clear
```
tobymao, line 26: why did you delete this

tobymao: can you undo these changes and just use a new graph like 1873?

ventusignis: I've undid the changes and made a separate graph for this case

tobymao (approved):

PR #6163, opened 2021-08-02
[18CZ] add phases to all trains

### LINKED ISSUE #6161: [18CZ] Not all train names have phase attached
Here's the current train table:

image

All train names have their corresponding phase (a through j) as a suffix except for 5E, 6E, and 8E, which for consistency should be 5Eh, 6Ei, and 8Ej respectively.

benjaminxscott:
closes #6161


Zwergenpunk:
you probably need to updae the mapping from train to corporation size

dfannius:
Also the `rusts_on` fields.

Issue #5909, opened 2021-07-02
18CO president offered illegal stock issue choices during EMR


ChrisTheCat:
https://18xx.games/game/40783?action=605

The president is offered the opportunity to issue 1, 2, 3, or 4 shares of UP. However, issuing 3 or 4 shares results in an error.

ryandriskel:
If for some reason someone else takes this on:

It is correct that issuing 3 / 4 shares is illegal as 2 shares is enough to buy the 6 train. We need to update the UI to not show options that aren't legal ( 3 / 4 shares in this case ).

ryandriskel:
I know what needs to be done here, but I borked some games with my initial implementation. Attempt two coming soon™

PR #6191, opened 2021-08-08
Fix 5909 - remove 18CO EMR extra options

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/5909

The return is "all bundles that do not allow buying any train, plus the one bundle that allows buying the cheapest train, if it exists."

I ran tests on this to ensure I did not wind up with unexpected nils.

### DIFF
--- a/lib/engine/game/g_18_co/game.rb
+++ b/lib/engine/game/g_18_co/game.rb
@@ -1495,7 +1495,9 @@ module Engine
         end
 
         def emergency_issuable_bundles(entity)
-          issuable_shares(entity)
+          eligible, remaining = issuable_shares(entity)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_co/game.rb
```diff
@@ -1495,7 +1495,9 @@ def emergency_issuable_cash(corporation)
         end
 
         def emergency_issuable_bundles(entity)
-          issuable_shares(entity)
+          bundles = issuable_shares(entity)
+          blocks = bundles.partition { |bundle| bundle.price + entity.cash < @depot.min_depot_price }
```
tobymao, line 1499: eligible, remaining = issuable_shares(entity).partition {}

eligible.concat(remaining.take(1))

tobymao (approved):

PR #6192, opened 2021-08-08
18EU - Only exchange connected minor

ryandriskel:

A corporation must be connected to the minor in order for it to be eligible for exchange.

Map:

Screen Shot 2021-08-07 at 10 26 02 PM

Before:

Screen Shot 2021-08-07 at 10 26 48 PM

After (active player owns 13 but not 15):

Screen Shot 2021-08-07 at 10 25 56 PM

### DIFF
--- a/lib/engine/game/g_18_eu/game.rb
+++ b/lib/engine/game/g_18_eu/game.rb
@@ -991,6 +991,13 @@ module Engine
 
           all_minor_hexes(corporation)
         end
+
+        def exchange_corporations(exchange_ability)
+          return super if !exchange_ability.owner.minor? || @loading
+
+          parts = graph.connected_nodes(exchange_ability.owner).keys
+          parts.select(&:city?).flat_map { |c| c.tokens.compact.map(&:corporation) }
+        end
       end
     end
   end


lib/engine/game/g_18_eu/game.rb
```diff
@@ -991,6 +991,13 @@ def home_token_locations(corporation)
 
           all_minor_hexes(corporation)
         end
+
+        def exchange_corporations(exchange_ability)
+          return super unless exchange_ability.owner.minor?
+
+          parts = graph.connected_nodes(exchange_ability.owner).keys
```
tobymao, line 998: this is expensive, can we put a return if game.loading? 

if not, then we need to turn this into an auto action

ryandriskel: Indeed we can. I forgot that's a thing!

ryandriskel: I directed it to super because that will evaluate all corps being legal, which I feel is more correct during load.

tobymao (approved):

PR #6198, opened 2021-08-10
Add context of who is laying tile to logic to determine if tile upgrades are legal

ventusignis:
This is needed for my approach on handling 18USA's special tile rules - particularly that in phase 5 and beyond, on it's first turn a corporation may, as a bonus tile lay, lay or upgrade its home to brown (skipping intermediate tile lays).

To know if it's legal to lay the given brown tile the `all_potential_upgrades` and `upgrades_to?` methods need to know which corporation is trying to lay in the given hex


tobymao:
can't you just raise in the step?

ventusignis:
I'm going to try something else with 18USA to see if this change can be avoided (at least for now)

Issue #5297, opened 2021-05-09
Make it obvious how to ping a player whose turn it is


slapphappyjazz:
I wanted to "ping" the current player in an async game, to remind them it was their turn.  (I suspected they had missed the automatic notification, and it turned out this was indeed the case).  I couldn't see how to ping them, so I resorted to e-mail outwith the game.  rr18xx has a button on one of its tabs to remind a player that it's their turn.

After trawling the docs this morning, I found that typing "@user" in the chat will send them a message.  But it's not obvious from the UI.  Here are some suggestions for making this easier.

1 - add a sentence to the help or the tools page

2 - add help text to the chat box itself, that is deleted when you start typing in the box

3 - add a "notify current player" button, either next to the chat box or on the Tools page.  Selecting this would switch to the chat box and enter "@user" - so the player selecting this option will quickly learn how to notify from the chat.

3a - a for option 3, but use with a "notify player" button that brings up a drop-down list to select the player to notify.

4 - make the player name on the Game page into a menu, with options to notify the player (as above) or view profiler.

marndt93:
What if a mail icon was added next to each player name on the game page.  Clicking this would add the ping message to the chat.
![image](https://user-images.githubusercontent.com/28715646/151258474-4c3efc6d-b9f9-4931-abb0-163f4aab77ac.png)

Alternatively a placeholder could be used to show the notify player command
![image](https://user-images.githubusercontent.com/28715646/151259207-e808da3c-4815-4f2c-b636-278e82fe351e.png)

tobymao:
let's go with placeholder, very elegant and clear

slapphappyjazz:
Looks good, but the message should use the same word in both places, either
Use @player command to ping a player, or
Use @user command to ping a user.

marndt93:
I was planning to use "Use @username command to ping a user" to remain consistent with the message in the Tools tab.

slapphappyjazz:
Sounds good to me.

PR #6199, opened 2021-08-10
player notification info sentence in tools tab

erperreault:
Screenshot from 2021-08-10 09-32-57

For #5297

I thought for a while about the other suggestions but not sure they'd be better - default text in the input box wouldn't be needed after one or two appearances, and the button would create two disconnected ways to do the same thing.

### DIFF
--- a/assets/app/view/game/tools.rb
+++ b/assets/app/view/game/tools.rb
@@ -29,7 +29,7 @@ module View
 
       def player_notification
         h('div.margined', [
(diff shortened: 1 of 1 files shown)


tobymao:
closing due to inactivity

erperreault:
Should I have opened this as a regular PR instead of a draft? I was hoping that marking it draft would invite opinions if this wasn't an acceptable fix, or if it's the right idea then someone would let me know I'm clear to change it to an regular PR.

tobymao:
yep, open up as a regular PR

PR #6201, opened 2021-08-10
1825: unit 1 & 2 maps, entities; option parsing

roseundy:
* split map and entities out from game.rb
* add options for all Unit combinations
* add checking for inconsistent options and raise OptionError
* add Unit 1 and Unit 2 maps, tiles, and entities.
* combine tiles based on options
* combine maps based on options
* combine entities based on options

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -287,9 +287,8 @@ module View
       game_params = params
       if @mode == :multi
         begin
-          game = create_game(params)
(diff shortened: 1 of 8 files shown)


roseundy:
Crap. Why in the world does 1825 have specs?

tobymao:
just delete them

roseundy:
Unfortunately, it was in the generic spec tests

PR #6217, opened 2021-08-13
[18carolinas] Update map.rb

philcampeau:
Corrects Richmond (A23) connection to dual gauge.


philcampeau:
Scott isn’t quite sure how he wants this to develop.

PR #6228, opened 2021-08-16
18Texas: Add new auction type: SimultaneousAuction

roseundy:


lib/engine/game/g_18_texas/step/simultaneous_auction.rb
```diff
@@ -0,0 +1,129 @@
...
+          end
+
+          def starting_bid(company)
+            [0, company.value].max
+          end
+
+          def min_bid(company)
+            return unless company
+
+            return starting_bid(company) unless @bids[company].any?
```
tobymao, line 80: if empty

roseundy: Hey, I ripped that off from 1817.

tobymao: fix it in both places and stop copying code :)

Issue #6074, opened 2021-07-22
[18zoo] Allowed to buy fourth "power" despite rules limit


KevinR1000:
This is an issue where 18xx.games is overly permissive relative to the rules, and (it seems) unintentionally so.

Example:  https://18xx.games/game/50339?action=27

To purchase Hole, I used the "Buy HOLE from The Bank" button at the bottom (above the market chart), and then confirmed that The Bank had agreed to the sale (responding to the warning).
I also see now that it applied the wrong price, using the 1$N from the box there.

I suspect that the "Buy XX from The Bank" buttons at the bottom are unnecessary.  Regardless, those buttons seem to bypass the "three powers" check.

carlorusso:
thanks @KevinR1000 , I'm the developer behind 18ZOO, I'll check this :)

carlorusso:
hi @KevinR1000 , thanks again for this submission :)
I found the problem, I have a working patch, but there is a tournament (https://18xx.tools/tournament/18ZOO/) and I prefer to wait for that before releasing this.

PR #6247, opened 2021-08-20
[18ZOO] fix https://github.com/tobymao/18xx/issues/6074

carlorusso:


lib/engine/game/g_18_zoo/step/buy_sell_par_shares.rb
```diff
@@ -54,7 +61,7 @@ def log_pass(entity)
             @log << "#{entity.name} declines to buy shares"
           end
 
-          def purchasable_companies(_entity)
+          def purchasable_companies?
```
tobymao, line 64: purchasable_companies? should return a boolean, the question mark always implies a boolean, so this needs to change

PR #6252, opened 2021-08-21
Prevent Safari treating chat input as username field

### LINKED ISSUE #4656: [Main chat] Safari browser on mobile assumes chat input is a login field
Visiting the site on my iPhone (Safari / iOS 14.4.1) today, I got a popup asking if I wanted to login using touch id. Doing so I ended up typing my email adress into the chat. Noticed that several other users also posted their email in the main chat earlier.

dlpierce:
Attempts to fix #4656 again after recent occurrences.

Adds a form tag around the chat input on the home page. This appears to prevent Safari from detecting the input as part of a login form when certain keywords are in the chat log.

Steps to replicate:
- Use MacOS Safari to create an account, electing to save your password. (Thanks BrowserStack)
- Logout
- Login using the saved password.
- Enter a chat message with the string "username"
- Refresh the page
- Click on the chat input and observe the option to fill in the form with your username.

Screenshot_20210821_040130

### DIFF
--- a/assets/app/view/chat.rb
+++ b/assets/app/view/chat.rb
@@ -42,6 +42,13 @@ module View
         end
       end
 
+      prevent_default = lambda do |event|
+        event = Native(event)
+        event.preventDefault
+      end
+
+      form_props = { on: { submit: prevent_default } }
+
       chatbar_props = {
         attrs: {
           autocomplete: 'off',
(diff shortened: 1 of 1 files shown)


dlpierce:
Looks like this can also happen with the game chat.

PR #6263, opened 2021-08-24
18Texas - Add private company powers, revise map

scottredracecar:
Company powers are activated during owning president's turn so new owning_player_track timing is added.

@roseundy provided code to perform a teleport action when a player owns the private company

### DIFF
--- a/assets/app/view/game/part/city_slot.rb
+++ b/assets/app/view/game/part/city_slot.rb
@@ -140,10 +140,11 @@ module View
             # If there's a choice of tokens of different types show the selector, otherwise just place
             next_tokens = step.available_tokens(entity)
             if next_tokens.size == 1 && actions.include?('place_token')
+              token_owner = step.respond_to?(:token_owner) && step.token_owner(@selected_company || @game.current_entity)
               action = Engine::Action::PlaceToken.new(
                 @selected_company || @game.current_entity,
                 city: @city,
-                tokener: @selected_company&.owned_by_player? ? @game.current_entity : nil,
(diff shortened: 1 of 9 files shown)


assets/app/view/game/part/city_slot.rb
```diff
@@ -140,10 +140,11 @@ def on_click(event)
             # If there's a choice of tokens of different types show the selector, otherwise just place
             next_tokens = step.available_tokens(entity)
             if next_tokens.size == 1 && actions.include?('place_token')
+              token_owner = step.respond_to?(:token_owner) && step.token_owner(@selected_company || @game.current_entity)
```
tobymao, line 143: put this method in game and don't use respond to

lib/engine/step/special_token.rb
```diff
@@ -53,7 +53,7 @@ def process_place_token(action)
         raise GameError, "Cannot place token on #{hex.name}#{city_string}" unless available_hex(entity, hex)
 
         place_token(
-          entity.owner,
+          entity.owner.player? ? @round.teleport_tokener : entity.owner,
```
tobymao, line 56: why not call token_owner

assets/app/view/game/part/city_slot.rb
roseundy: OK

lib/engine/step/special_token.rb
roseundy: Yep.

tobymao (approved):

PR #6297, opened 2021-08-30
Skip undone actions when navigating history and don't reload entire game when navigating forward

crericha:
Store all game actions, even future ones, in the game class. Add helper methods to determine previous and next game actions. Add a helper method to process actions from the current state to the specified action.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -622,9 +622,12 @@ module Engine
       # Initialize actions respecting the undo state
       def initialize_actions(actions, at_action: nil)
         @loading = true unless @strict
-
         filtered_actions, active_undos = self.class.filtered_actions(actions)
-        @raw_filtered_actions = filtered_actions.compact
+
+        # Store all actions for history navigation
+        @raw_all_actions = actions
+        filtered_actions.each.with_index { |action, index| @raw_all_actions[index]['skip'] = true unless action }
+
         @undo_possible = false
         # replay all actions with a copy
         filtered_actions.each.with_index do |action, index|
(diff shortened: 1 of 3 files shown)


lib/engine/game/base.rb
```diff
@@ -622,14 +620,19 @@ def self.filtered_actions(actions)
...
 
       # Initialize actions respecting the undo state
-      def initialize_actions(actions)
+      def initialize_actions(actions, at_action: nil)
         @loading = true unless @strict
-
         filtered_actions, active_undos = self.class.filtered_actions(actions)
+
+        # Store all actions for history navigation
+        @raw_all_actions = actions
```
tobymao, line 628: what's the difference between raw_actions and this

lib/engine/game/base.rb
```diff
@@ -764,6 +767,35 @@ def last_game_action_id
         @last_game_action_id || 0
       end
 
+      def previous_game_action_id_from(action_id)
```
tobymao, line 770: rename all of these to be without the game, so previos_action_id(action)

lib/engine/game/base.rb
```diff
@@ -764,6 +767,35 @@ def last_game_action_id
...
+        if (action = @raw_all_actions.reverse.find { |a| a['id'] < action_id && !a['skip'] && a['type'] != 'message' })
```
tobymao, line 771: @raw.rverse.find {...}&.['id']

all of these can be cleaned up

lib/engine/game/base.rb
crericha: raw_actions is all the actions up to the current game state. For example, if the player is reviewing the game history, it will only include all actions to the point at which they are in the game history. Conversely, raw_all_actions is all the actions taken, even future ones. 

A lot of logic depends on raw_actions, so I created a new array to not break that logic.

lib/engine/game/base.rb
crericha: I feel like previous_action_id is misleading as this will return the previously processed player move, ignoring any messages and undone actions in between the current move and the last move, so it isn't the previous action. I worry without adding another descriptive word, the method will be misleading to others. That's why I added the word "game" to the method name. Thoughts?

lib/engine/game/base.rb
crericha: Ahh, cool. I didn't realize you can use "." with operator methods. Will clean up.

lib/engine/game/base.rb
tobymao: the game in it doesn't help that goal imo, it just makes it more verbose.

lib/engine/game/base.rb
crericha: Confirmed you can't use "&." with an operator method. I did find another hash method (fetch) that retrieved the value.

PR #6310, opened 2021-09-06
Auto independent mines

magnusrk-ch:
This adds an auto program for Harzbahn 1873 to automate independent mines, which currently are a big drag on game speed due to their large number and mostly trivial actions.

The first commit contains the program, the second commit extends the UI to allow disabling each of the three actions a mine can perform separately.

Code passed rake, but I'm a total beginner at ruby and github, so there's probably room for improvement still.

### DIFF
--- a/assets/app/view/game/auto.rb
+++ b/assets/app/view/game/auto.rb
@@ -23,6 +23,7 @@ module View
         if @game.players.find { |p| p.name == @user&.dig('name') }
           types = {
             Engine::Action::ProgramBuyShares => ->(settings) { render_buy_shares(settings) },
+            Engine::Action::ProgramIndependentMines => ->(settings) { render_independent_mines(settings) },
             Engine::Action::ProgramMergerPass => ->(settings) { render_merger_pass(settings) },
             Engine::Action::ProgramSharePass => ->(settings) { render_share_pass(settings) },
           }.freeze
@@ -330,6 +331,46 @@ module View
                          'Please read this for more details when it will deactivate')])
 
         subchildren = [render_button(settings ? 'Update' : 'Enable') { enable_share_pass }]
+        subchildren << render_disable(settings) if settings
+        children << h(:div, subchildren)
+
+        children
+      end
+
+      def enable_independent_mines(form)
+        settings = params(form)
+
+        indefinite = settings['indefinite']
+
+        process_action(
+          Engine::Action::ProgramIndependentMines.new(
+            sender,
+            indefinite: indefinite
+          )
+        )
+      end
+
+      def render_independent_mines(settings)
+        form = {}
+        text = 'Auto Independent Mines'
+        text += ' (Enabled)' if settings
+        children = [h(:h3, text)]
+        children << h(:p,
+                      'Automatically skip closing independent mines, and optionally also laying track.'\
+                      ' This will deactivate itself in the next SR, unless set to indefinite.'\
+                      ' It will also deactivate itself when a mine has negative income.')
+
+        indefinite = settings ? settings&.indefinite : false
+        children << h(:div, [render_input(
+          'Indefinite (normally stops after one OR set)',
+          id: 'indefinite',
+          type: 'checkbox',
+          inputs: form,
+          attrs: {
+            checked: indefinite,
+          })])
+
+        subchildren = [render_button(settings ? 'Update' : 'Enable') { enable_independent_mines(form) }]
         subchildren << render_disable(settings) if settings
         children << h(:div, subchildren)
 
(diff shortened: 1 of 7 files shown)


roseundy:
Maybe I'm missing it, but I don't see where render_independent_mines is being called.

Also, please make sure you test this on non-1873 games. Running rake is not sufficient since rake will not catch UI bugs.

magnusrk-ch:
> Maybe I'm missing it, but I don't see where render_independent_mines is being called.
> 
> Also, please make sure you test this on non-1873 games. Running rake is not sufficient since rake will not catch UI bugs.

Yeah, that line ended up in a different branch for auction automation. Looks like the same happened to the 1873 config file that enables the automation for 1873. I'll add them back once I've sorted out what else got lost in the split.

I think the change itself is quite safe. The only edits are to 1873 files, the program (only enabled by 1873) and the program UI (only invoked if the program is enabled). I'll do some manual testing just in case, though.

assets/app/view/game/auto.rb
```diff
@@ -336,6 +336,67 @@ def render_share_pass(settings)
...
+        form = {}
+        text = 'Auto Independent Mines'
+        text += ' (Enabled)' if settings
+        children = [h(:h3, text)]
+        children << h(:p,
+                      'Automatically skip independent mine actions.'\
+                      ' This will deactivate itself in the next SR, unless set to indefinite.'\
+                      ' It will also deactivate itself when a mine has negative income.')
+
+        skip_track = settings ? settings&.skip_track : true
```
tobymao, line 380: you could remove all of these and just do 

render_checkbox(... settings&.skip_track : true)

assets/app/view/game/auto.rb
```diff
@@ -336,6 +336,67 @@ def render_share_pass(settings)
...
+        skip_buy = settings ? settings&.skip_buy : true
+        skip_close = settings ? settings&.skip_close : true
+        children << h(:div, [
+          render_checkbox('Skip track lay', 'skip_track', form, skip_track),
+          render_checkbox('Skip switchers', 'skip_buy', form, skip_buy),
+          render_checkbox('Skip close mine', 'skip_close', form, skip_close),
+        ])
+        indefinite = settings ? settings&.indefinite : false
```
tobymao, line 388: this one can be render_checkbox(... settings&.indefinite)

lib/engine/game/g_1873/step/buy_train.rb
```diff
@@ -345,6 +345,19 @@ def ebuy_president_can_contribute?(_corporation)
           def president_may_contribute?(_corporation, _active_shell)
             false
           end
+
+          def auto_actions(entity)
```
tobymao, line 349: these are all copy and pasted, maybe make a module and include it?

magnusrk-ch:
I've recovered the missing deltas and applied Toby's suggested changes. Ran through two ORs of 1873 to test it, and verified it doesn't show up in 1860.

roseundy:
Make sure you run it against a more mainline game like 18Chesapeake. 1860 overrides too many things.

lib/engine/game/g_1873/step/buy_train.rb
magnusrk-ch: I'm not sure exactly what you mean here. If it's about the auto_actions method, that's common to all classes that support programs. That seems like it should be addressed separately. If it's about the activate method, they're only identical in the first commit. The second commit changes them to read different program properties. Let me know if I've misunderstood.

magnusrk-ch:
18Chesapeake appears fine as well.

lib/engine/game/g_1873/step/buy_train.rb
tobymao: activate_program_independent_mines  is not correct? whta's the 'second commit' you mean you plan to have a different PR?

magnusrk-ch: activate_program_independent_mines is correct (AFAIK). The "second commit" is the second commit in the PR branch. In the combined change the methods are similar, but not identical.

tobymao: available_actions = actions(entity)
            return unless available_actions.include?('pass')
            return unless entity.minor?
            return unless program.skip_buy

            [Action::Pass.new(entity)]


this looks identical to me? what is different about them across the different steps?

magnusrk-ch: program.skip_buy is replaced by program.skip_close and program.skip_track in the other classes.

tobymao: got it thanks

PR #6311, opened 2021-09-06
1825: rendering fixes; K1, K6 and D1

roseundy:
* Implement "Sepia" tiles to distinguish pre-printed non-upgradable track from gray tiles. A nice side-effect of this is that I no longer need stubs on the gray borders in order to lay track against them.
* Fix rendering of "Halt" tile 11
* Minor tile tweaks
* Implement final kits K1, K6 and D1

sepia_map


lib/engine/config/tile.rb
```diff
@@ -490,13 +490,17 @@ module Tile
         '1200' => 'city=revenue:10;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:4,b:_0;path=a:5,b:_0',
       }.freeze
 
+      BROWNSEPIA = {
```
tobymao, line 493: just call it sepia?

roseundy: No. It's a striped tile with both brown and sepia

PR #6356, opened 2021-09-15
1825: fix OO tiles, 3 rusting, another constant mutation

### LINKED ISSUE #6349: 1825 - Double City Tiles
https://18xx.games/game/56469
Game start errors out with a Tile #10 inconsistency.
-Optional Rules: BigBank; D1; K2; K3; K5; K7; R1; R2; R3; U3+; Units 1+2+3

Need to standardize the double city spots on the board and tiles.

Original game maps and tiles do not have OO on them.

Unit 1 has OO on tiles and on the board
Unit 2 has OO on tiles and on the board
Unit 3 has OO on tiles and without on the board
D1 has #10 tiles without OO
R2.1 has OO on tiles and on the board
Screenshot 2021-09-14 113618
image (2)

### LINKED ISSUE #6352: [1825 ] Phase 4: 7-trains rust 3-trains
Kit K3  "Phase 4":  Phase 4 starts with the pucrchase of the first 6 train leaving 3-trains still "in use". Bur with the purchase of the first 7-train, 3-trains are no longer "in use". What means, in my understanding, that 3-trains are rusted by the 7-trains.

### LINKED ISSUE #6353: [1825 Unit 1] Cannot lay tiles with GWR unless game is reloaded
https://18xx.games/game/56408

Could not reproduce the issue in a hotseat game

During OR 3.1 when doing the tile lay for  GWR cannot lay at 12T. I get the pictured error. Tried closing the tab and reloading but cannot move forward with second tile lay.

Also happens when trying to use chat.

Windows 10
Firefox 92.0 64-bit

image

roseundy:
Fixes #6349 
Fixes #6352 
Fixes #6353 

This shouldn't invalidate games.

### DIFF
--- a/lib/engine/game/g_1825/game.rb
+++ b/lib/engine/game/g_1825/game.rb
@@ -189,12 +189,38 @@ module Engine
           },
         }.freeze
 
+        ALL_TRAINS_K3 = {
+          '2' => { distance: 2, price: 180, rusts_on: '5' },
+          '3' => { distance: 3, price: 300, rusts_on: '7' },
+          '4' => { distance: 4, price: 430 },
+          '5' => { distance: 5, price: 550 },
+          '3T' => { distance: 3, price: 370, available_on: '3' },
+          'U3' => {
+            distance: [{ 'nodes' => ['city'], 'pay' => 3, 'visit' => 3 },
+                       { 'nodes' => ['town'], 'pay' => 99, 'visit' => 99 }],
+            price: 410,
+            available_on: '3',
+          },
+          '6' => { distance: 7, price: 650 },
+          '4T' => { distance: 4, price: 480, available_on: '4' },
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1825/game.rb
```diff
@@ -189,12 +189,38 @@ def game_phases
...
+            available_on: '4a',
+          },
+        }.freeze
+
         def build_train_list(thash)
           thash.keys.map do |t|
             new_hash = {}
             new_hash[:name] = t
             new_hash[:num] = thash[t]
-            new_hash.merge!(ALL_TRAINS[t])
```
tobymao, line 197: is this constant mutation?

roseundy: I don't think so since it's just merging a constant hash into a new hash. I'm not changing any of the hash values being merged.

PR #6364, opened 2021-09-22
1825: better handling of options

roseundy:
The modifies the create_game view to add optional rule sanity checking, and mutually exclusive options (essentially radio buttons).
It also changes the way optional rules are "hidden" based on player count - instead of removing them, it changes the checkboxes to disabled. It also fixes issues the code had when options became hidden to due player count

This also modifies the game_card view to dynamically determine the minimum number of players bases on optional rules

Finally, the meta for 1825 is modified to take advantage of these changes.

### DIFF
--- a/lib/engine/game/g_1825/meta.rb
+++ b/lib/engine/game/g_1825/meta.rb
@@ -119,6 +119,118 @@ module Engine
             desc: 'Do not increase bank size based on number of minors and kits',
           },
         ].freeze
+
+        MUTEX_RULES = [
+          %i[unit_1 unit_2 unit_3 unit_12 unit_23 unit_123],
+          %i[big_bank strict_bank],
+        ].freeze
+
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1825/meta.rb
```diff
@@ -119,6 +119,118 @@ module Meta
             desc: 'Do not increase bank size based on number of minors and kits',
           },
         ].freeze
+
+        MUTEX_RULES = [
```
tobymao, line 123: exclusive_rules, standalone rules?

tobymao (approved):

lib/engine/game/g_1825/meta.rb
roseundy: sorry, MUTEX is a verilog term for "mutually exclusive'

PR #6418, opened 2021-10-10
[18 tokaido] initial prototype implementation

doubt72:
Initial prototype implementation.

Needs playtesting; I don't think there are any major bugs at this point (it's not really that complicated and have run lots of test games to work out the kinks), but have _absolutely no idea_ if it's really well balanced; needs real people to pound on it.


tobymao:
have you tested a non hot seat game?

doubt72:
Yes.  And it did indeed find some arity issues.

Issue #6374, opened 2021-09-25
[1840] Grinzing and Heiligenstadt errors on 2P/3P map


orome:
Two map errors on 2P/3P:

- [Grinzing](https://www.grinzland.com/top-attractions) is present on 2P overlay but it is absent from the printed overlay 
- Heiligenstadt has an extra interchange on the 3P map not present on the printed map

Both should be removed.

Zwergenpunk:
The Heiligensttadt extra interchange requires alot of rewriting to be removed. So that is going to stay until someone has a lot of time.
Grinzing is just a bug. Thanks for finding it ;)

orome:
As stated:
> Grinzing is present on **2P overlay** but it is absent from the printed overlay

To be clear: Grinzing _should be on the 3P_ board, just not on the 2P.

orome:
@benjaminxscott I see it removed from both 3P and 2P (so presumably it's removed from the 2P because it's removed from 3P?). It should only be removed from 2P.

orome:
[Those changes](https://github.com/tobymao/18xx/pull/6439/files) look right to me (though I'm not a Ruby guy).

orome:
@Zwergenpunk Were both Grinzing and Heiligenstadt fixed, or just Grinzing?

Zwergenpunk:
just Grinzing. heiligenstadt will probably never be fixed, because the order of the cities is relevant for alot of stuff in the game. and the city to remove has index 0 =/

orome:
KK Just making sure to split off Heiligenstadt (#6488) for bookkeeping.

PR #6437, opened 2021-10-15
[1840] remove Grinzing hex in small 3P map

v01dpointer:
partially addresses #6374 
Screenshot 2021-10-15 1 23 14 PM
Screenshot 2021-10-15 1 24 58 PM


benjaminxscott:
The fixture uses `A13` which was removed, so the now-deprecated fixture broke the build

PR #6452, opened 2021-10-19
Show all finished games on the home page

magnusrk-ch:
Currently games the logged-in user was part of are excluded from the home page. It makes sense to have a separate list of your own games on the profile page, but I don't see any reason we can't also show them on the home page. They're already being fetched and downloaded, just filtered out.

No changes to the profile page.

### DIFF
--- a/assets/app/view/home.rb
+++ b/assets/app/view/home.rb
@@ -16,16 +16,15 @@ module View
     needs :refreshing, default: nil, store: true
 
     def render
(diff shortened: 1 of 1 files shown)


magnusrk-ch:
While working on a related change I discovered this change introduces a bug. Your own games don't paginate, so as you paginate through finished games more and more of your own games accumulate and push out other games.

PR #6477, opened 2021-11-01
Add 1822MX initial files

outsidePasser:


outsidePasser:
Files are based on those from 1822 MRS with changes for 1822 MX.

outsidePasser:
Moved map and tiles to map.rb

tobymao (approved):

PR #6478, opened 2021-11-03
[18NY] More operation round functionality

crericha:
- Implement connection bonus tokens
- Place coal tokens with assignment
- Implement hex trains
- Implement tokens ($20 * shortest legal path from another station)

### DIFF
--- a/lib/engine/game/g_18_ny/game.rb
+++ b/lib/engine/game/g_18_ny/game.rb
@@ -147,12 +147,32 @@ module Engine
             ['Capitalization Round', 'Special Capitalization Round before next Stock Round'],
         ).freeze
 
+        ASSIGNMENT_TOKENS = {
+          'connection_bonus' => '/icons/18_ny/connection_bonus.svg',
+          'coal' => '/icons/18_ny/coal.svg',
+        }.freeze
+
+        CONNECTION_BONUS_HEXES =
+          %w[A13 A19 A23 B12 C11 C23 C25 D0 D18 D20 E9 F10 F12 G9 G13 G19 G21 G25 I19 I23 J18 J22 J26 K19].freeze
+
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_ny/game.rb
```diff
@@ -383,6 +403,72 @@ def upgrade_cost(tile, _hex, entity, spender)
...
+          if (num_bonuses = route.stops.count { |stop| stop.hex.assigned?('connection_bonus') }).positive?
+            str += " + #{num_bonuses} Connection Bonus#{num_bonuses == 1 ? '' : 'es'}"
+          end
+
+          str
+        end
+
+        def routes_revenue(routes)
+          revenue = super
+          revenue += connection_bonus(routes.first.corporation) if routes.any?
```
tobymao, line 437: unless empty

lib/engine/game/g_18_ny/step/route.rb
```diff
@@ -0,0 +1,21 @@
...
+require_relative '../../../step/route'
+
+module Engine
+  module Game
+    module G18NY
+      module Step
+        class Route < Engine::Step::Route
+          def process_run_routes(action)
+            super
+            action.routes.map(&:stops).flatten.select { |stop| stop.hex.assigned?('connection_bonus') }.each do |stop|
```
tobymao, line 12: flat_map(&:stops).each do |stop|
   next if stop..

lib/engine/game/g_18_ny/game.rb
crericha: Okay. I thought we determined performance was the same though?

tobymao: performance is not the same

crericha: Hmm...I thought Jen ran it both ways and determined it was. Maybe I'm misremembering.

PR #6514, opened 2021-11-14
[18NY] Implement mergers

crericha:
Also buying privates in the stock round and change minor tokens to black.

Common code changes
- Made the Merge button customizable based on which entity is selected. The button also appears disabled if nothing is selected.
- Added buy company for bank owned privates to base stock round. It was original subclassed by many games.
- Fixed 1822 assumption that if buy_company and bid actions are present, then companies are biddable. For 18NY the biddable entity is the minors (auction).

Tested against the entire database and UI was spot tested in affected games.

### DIFF
--- a/assets/app/view/game/round/stock.rb
+++ b/assets/app/view/game/round/stock.rb
@@ -369,7 +369,7 @@ module View
         end
 
         def render_bid_input(company)
-          return [] unless @step.can_bid?(@current_entity, company)
+          return [] unless @step.respond_to?(:can_bid_company?) && @step.can_bid_company?(@current_entity, company)
 
           [h(Bid, entity: @current_entity, corporation: company)]
         end
--- a/assets/app/view/game/round/merger.rb
+++ b/assets/app/view/game/round/merger.rb
@@ -191,7 +191,6 @@ module View
               store(:flash_opts, 'Select a corporation to merge with')
             end
           end
-          puts @selected_corporation
           h(:button, { attrs: { disabled: !@selected_corporation }, on: { click: merge } },
             @step.merge_name(@selected_corporation))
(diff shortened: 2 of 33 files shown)


assets/app/view/game/round/stock.rb
```diff
@@ -369,7 +369,7 @@ def render_buy_input_interval(company)
         end
 
         def render_bid_input(company)
-          return [] unless @step.can_bid?(@current_entity, company)
+          return [] unless @step.respond_to?(:can_bid_company?) && @step.can_bid_company?(@current_entity, company)
```
tobymao, line 372: why was this changed?

tobymao: this looks like a very different change, why not continue to call it can_bid?

tobymao: why use respond_to?

crericha: 18NY hit view code that was originally was only hit by 1822 and used methods that only exist in the 1822 code base. The 1822 can_bid? method collided with the 18NY can_bid? method, which had a different signature and caused issues. This view code assumed that if you can 'buy_company' and 'bid', then you must be able to bid on companies (bad assumption and needed to be fixed).

Upon investigation into the code base, the rest of the code base matched the 18NY can_bid? signature, so I left that one alone and instead changed the 1822 specific can_bid? to can_bid_company?, which is what the method is really trying to determine (vs bidding to start a corporation).  Since only 1822 uses that method, I included respond_to?, so that other games weren't required to implement the method unless they needed it. This is standard practice throughout the view code for game specific behavior.

tobymao (approved):

PR #6527, opened 2021-11-21
[18NY] NYC Formation

crericha:
Also a few bug fixes

Comments inline explaining the common code changes.

I pulled the DB a few days ago and I ran against all the games. All games passed except 18TN, which needs to be pinned. The existing games have a withhold actions for when the train doesn't run due to the civil war. These are now handled automatically without the need for the player to submit the action. @perwestling FYI in case you want to take a look I also made a small tweak to 18MS.


lib/engine/game/base.rb
```diff
@@ -1306,6 +1306,8 @@ def place_home_token(corporation)
               home_token_locations(corporation)
             end
 
+          return unless hexes
```
crericha, line 1309: National railways may not have a home token to lay. Continue on without placing a token if they have no home hex.

lib/engine/game/base.rb
```diff
@@ -1537,9 +1539,9 @@ def can_go_bankrupt?(player, corporation)
       end
 
       def total_emr_buying_power(player, corporation)
```
crericha, line 1541: Allow method to be used for a player when their is no associated corporation.

lib/engine/game/g_18_ireland/round/stock.rb
```diff
@@ -17,6 +17,10 @@ def finish_round
             end
             super
           end
+
+          def corporations_to_move_price
```
crericha, line 21: Default for minors changed to not float at the end of the stock round. 18Ireland is special in that minors do float, so put the special case here instead of in the mainline code.

lib/engine/game/g_18_ms/game.rb
```diff
@@ -365,8 +365,10 @@ def revenue_for(route, stops)
         end
 
         def routes_revenue(routes)
+          return 0 if routes.empty?
+
           if @round.round_num == 2
-            active_step.current_entity.trains.each do |t|
+            routes.first.corporation.trains.each do |t|
```
crericha, line 371: Changed dividend step to determine action on whether there is revenue or not, not on routes. This caused an infinite loop for 18MS since current_entity leads to actions being called, which in turn calls routes_revenue...Get the corporation from the route instead of current_entity to break the infinite loop.

lib/engine/round/stock.rb
```diff
@@ -83,7 +83,7 @@ def finish_round
       end
 
       def corporations_to_move_price
-        @game.corporations.select(&:floated?)
+        @game.corporations.select { |c| c.floated? && c.type != :minor }
```
crericha, line 86: In most games, minors' stock prices do not move at the end of the stock round. Make this the default, rather than have most games need to override.

lib/engine/share_pool.rb
```diff
@@ -232,6 +232,7 @@ def transfer_shares(bundle, to_entity,
 
       # skip the president's share swap if the initiator is already the president
       # or there was no previous president. this is because there is no one to swap with
+      previous_president ||= corporation if bundle.owner == corporation && !bundle.presidents_share
```
crericha, line 235: For nationals, it is common the presidents share isn't acquired when the corporation starts, but through mergers, etc.. Add this line, so that the president's share is swapped for correctly when a player acquires a second share and now has enough to obtain the president's share.

lib/engine/step/dividend.rb
```diff
@@ -10,7 +10,7 @@ class Dividend < Base
       ACTIONS = %w[dividend].freeze
 
       def actions(entity)
-        return [] if entity.company? || routes.empty?
+        return [] if entity.company? || @game.routes_revenue(routes).zero?
```
crericha, line 13: Trigger the dividend step for games that have revenue that comes from other places beyond their trains.

lib/engine/step/minor_half_pay.rb
```diff
@@ -5,12 +5,13 @@ module Step
     module MinorHalfPay
       def actions(entity)
         return [] if entity.minor?
+        return [] if entity.corporation? && entity.type == :minor
```
crericha, line 8: Allow minor's that are based on the corporation class to also take advantage of the module.

spec/fixtures/18TN/7818.json
```diff
@@ -1927,14 +1927,6 @@
...
       "original_id": 193
     },
-    {
-      "type": "dividend",
-      "entity": "GMO",
-      "entity_type": "corporation",
-      "id": 203,
-      "kind": "withhold",
-      "original_id": 194
-    },
```
crericha, line 1937: Removal of actions no longer required for 18TN and are instead handled dynamically.

lib/engine/game/g_18_ny/game.rb
```diff
@@ -352,6 +423,26 @@ def operating_order
...
+          @buffalo_corp ||=
+            Engine::Corporation.new(name: 'Buffalo', sym: 'BUF', tokens: [], coordinates: 'E3')
+
+          non_blocking_graph.clear_graph_for(@buffalo_corp)
+          non_blocking_graph.connected_hexes(@buffalo_corp).key?(albany_hex)
+        end
+
+        def home_token_locations(corporation)
+          return super unless corporation == nyc_corporation
+          return nil unless corporation.tokens.select(&:used).empty?
```
tobymao, line 440: tokens.any?(&:used)

lib/engine/game/g_18_ny/game.rb
```diff
@@ -257,37 +316,50 @@ def event_close_companies!
...
         end
 
         def event_nyc_formation!
+          return if @nyc_formation_state
+
           @log << "-- Event: #{EVENTS_TEXT['nyc_formation'][1]} --"
+          @nyc_formation_state = :round_one
+
+          @log << 'No further minor corporations may be started'
+          @corporations.select { |c| c.type == :minor && !c.floated? && !c.closed? }.each do |c|
```
tobymao, line 352: remove the select and just do an next if in the each

lib/engine/game/g_18_ny/game.rb
```diff
@@ -352,6 +423,26 @@ def operating_order
...
+        def non_blocking_graph
+          @non_block_graph ||= Graph.new(self, no_blocking: true, home_as_token: true)
+        end
+
+        def albany_and_buffalo_connected?
+          @buffalo_corp ||=
+            Engine::Corporation.new(name: 'Buffalo', sym: 'BUF', tokens: [], coordinates: 'E3')
+
+          non_blocking_graph.clear_graph_for(@buffalo_corp)
+          non_blocking_graph.connected_hexes(@buffalo_corp).key?(albany_hex)
```
tobymao, line 435: connected_hexens[albany_hex]

lib/engine/game/g_18_ny/game.rb
```diff
@@ -751,6 +819,187 @@ def complete_acquisition(_entity, corporation)
...
+        end
+
+        def minors_connected_to_albany
+          non_blocking_graph.clear
+
+          active_minors.select do |minor|
+            # Minor 1 and 2 are always considered connected
+            next true if %w[1 2].include?(minor.id)
+
+            non_blocking_graph.connected_hexes(minor).key?(albany_hex)
```
tobymao, line 924: connected_hexes(minor)[albany_hex]

tobymao:
rerun with strict: true as well to check game loading failures

lib/engine/game/g_18_ny/game.rb
crericha: Fixed

lib/engine/game/g_18_ny/game.rb
crericha: Fixed

lib/engine/game/g_18_ny/game.rb
crericha: Fixed

lib/engine/game/g_18_ny/game.rb
crericha: Fixed

tobymao (approved):
merge whenever you're ready

crericha:
> rerun with strict: true as well to check game loading failures

Picked up a few 1825 illegal tile rotations, but that wouldn't have anything to do with my changes.

Remember that 18TN will need pins when you deploy

PR #6528, opened 2021-11-22
[18NY] Implement capitalization round and bug fixes

crericha:
Ran against the whole db of games with strict: true. No new failures were seen.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1295,10 +1295,9 @@ module Engine
             @round.clear_cache!
             return
           end
+
           # If the tile does not have any paths at the present time, clear up the ambiguity when the tile is laid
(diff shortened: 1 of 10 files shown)


lib/engine/game/base.rb
```diff
@@ -984,6 +984,10 @@ def can_buy_presidents_share_directly_from_market?
         false
       end
 
+      def can_swap_for_presidents_share_directly_from_corporation?
+        true
```
crericha, line 988: 1856 hard codes who gets the president's share when the CGR is formed, rather than use share_pool to do it. It worked because of a bug in share_pool. Now that it is fixed, 1856 needs a way to tell share_pool not to hand out the president's share when someone gets enough shares to obtain it.

tobymao (approved):

PR #6532, opened 2021-11-23
add error messages if unable to token

### LINKED ISSUE #6530: [1846] Better error for token placement error
**What's your idea?**

When IC attempts to place a token in Terre Haute (G7) here https://18xx.games/game/61221?action=235
You get an error message "IC cannot place a token 619-1-0 G7", a better message would be "IC has insufficient funds for token placement"

benjaminxscott:
closes #6530 

Tested with cases:
1 - IC attempts to token Terra Haute, which it can't afford
https://gist.github.com/benjaminxscott/4526e3d9603ea5d9df8218f254e5c2c5

2 - NYC attempts to token Erie, where it already has a token
https://gist.github.com/benjaminxscott/d94992f37484341daa179eb9c262112d

3 - GT attempts to token Erie, which has a reserved slot so it has no open slots
https://gist.github.com/benjaminxscott/14743b143adca8253140b918607a37d6

### DIFF
--- a/lib/engine/part/city.rb
+++ b/lib/engine/part/city.rb
@@ -127,26 +127,25 @@ module Engine
         if check_tokenable && !tokenable?(
             corporation, free: free, tokens: token, cheater: cheater, extra_slot: extra_slot, spender: spender
           )
-          raise GameError, "#{corporation.name} cannot lay token - has no tokens left" if @error == :no_tokens
 
-          if @error == :existing_token
+          case @error
+          when :no_tokens
+            raise GameError, "#{corporation.name} cannot lay token - has no tokens left"
+          when :existing_token
             raise GameError,
                   "#{corporation.name} cannot lay token - already has a token on #{tile.hex&.id}"
-          end
-          if @error == :blocked_reservation
+          when :blocked_reservation
             raise GameError,
                   "#{corporation.name} cannot lay token - remaining token slots are reserved on #{tile.hex&.id}"
-          end
-          if @error == :no_money
+          when :no_money
             raise GameError,
                   "#{corporation.name} cannot lay token - cannot afford token on #{tile.hex&.id}"
-          end
-          if @error == :no_slots
+          when :no_slots
             raise GameError,
                   "#{corporation.name} cannot lay token - no token slots available on #{tile.hex&.id}"
+          else
+            raise GameError, "#{corporation.name} cannot lay token on #{id} #{tile.hex&.id}"
           end
-
-          raise GameError, "#{corporation.name} cannot lay token on #{id} #{tile.hex&.id}"
         end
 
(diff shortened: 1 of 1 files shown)


lib/engine/part/city.rb
```diff
@@ -110,6 +110,28 @@ def place_token(corporation, token, free: false, check_tokenable: true, cheater:
...
+            raise GameError,
+                  "#{corporation.name} cannot lay token - cannot afford token on #{tile.hex&.id}"
+          end
+          unless extra_slot || get_slot(
+token.corporation, cheater: cheater
+)
+            raise GameError,
+                  "#{corporation.name} cannot lay token - no token slots available on #{tile.hex&.id}"
+          end
+
```
crericha-qualitrol, line 134: Since you mentioned DRY in slack, here is an idea of how to do that:

Create a method that returns the error string or reason (Ex. :reserved) if tokenable? fails and then have the various methods that make up tokenable? set that error string/reason?

I'm not entirely sure if this is a good idea or not, but I do think it is worth avoiding duplication and dependency between tokenable? and place_token.

lib/engine/part/city.rb
```diff
@@ -110,6 +127,25 @@ def place_token(corporation, token, free: false, check_tokenable: true, cheater:
         if check_tokenable && !tokenable?(
             corporation, free: free, tokens: token, cheater: cheater, extra_slot: extra_slot, spender: spender
           )
+          raise GameError, "#{corporation.name} cannot lay token - has no tokens left" if @error == :no_tokens
+
+          if @error == :existing_token
```
tobymao, line 132: case @error

crericha:
Posted on wrong account. Moved here.

Since you mentioned DRY in slack, here is an idea of how to do that:

Create a method that returns the error string or reason (Ex. :reserved) if tokenable? fails and then have the various methods that make up tokenable? set that error string/reason?

I'm not entirely sure if this is a good idea or not, but I do think it is worth avoiding duplication and dependency between tokenable? and place_token.

lib/engine/part/city.rb
```diff
@@ -78,14 +78,31 @@ def city?
       def tokenable?(corporation, free: false, tokens: corporation.tokens_by_type, cheater: false,
                      extra_slot: false, spender: nil)
         tokens = Array(tokens)
-        return false if !extra_slot && tokens.empty?
+        @error = :generic
+        if !extra_slot && tokens.empty?
+          @error = :no_tokens
```
tobymao, line 83: why don't you just always return the error and switch it to nil or error

lib/engine/part/city.rb
benjaminxscott: https://github.com/tobymao/18xx/pull/6532/commits/7eccffb8121f545867d6efbfc518af904fc5aaa9

tobymao (approved):

PR #6533, opened 2021-11-23
1822 Create hook so subclasses can override updating private revenue

outsidePasser:


lib/engine/game/g_1822/game.rb
```diff
@@ -953,6 +953,12 @@ def operating_round(round_num)
         end
 
         def payout_companies
+          # Functionality extracted so it can be over-ridden by subclasses
```
crericha-qualitrol, line 956: I don't think having the rationale of why the logic was split out will be helpful to future developers. I suggest removing.

outsidePasser: I don't mind removing it, but I thought it would be better to explain why a method contained only a single call along with "super".

crericha: My opinion is that its a common idiom used throughout the code base, so its not worth noting in this one place.

tobymao: yea, no need to have comments like this

outsidePasser:
I'll pull the comment out in a bit.

tobymao (approved):

PR #6534, opened 2021-11-24
[18NY] Playtest bug fixes

crericha:


assets/app/view/game/game_info.rb
```diff
@@ -225,7 +225,7 @@ def trains
               event_text << if index.zero?
                               event_name
                             else
-                              "#{event_name}(on #{ordinal(index + 1)} train)"
+                              "#{event_name}(on #{ordinal(train2.index + 1)} train)"
```
crericha, line 228: The relative index will change as trains of that type are bought. It should be the train's absolute index that matters.

lib/engine/train.rb
```diff
@@ -28,7 +28,7 @@ def initialize(name:, distance:, price:, index: 0, **opts)
       @rusted = false
       @obsolete = false
       @operated = false
-      @events = (opts[:events] || []).select { |e| @index == (e[:when] || 0) }
+      @events = (opts[:events] || []).select { |e| @index == (e[:when] || 1) - 1 }
```
crericha, line 31: Event definitions shouldn't need to know train indexes are zero based. For example, if :when = 3, it is expected that the event happens on the 3rd train, not the 4th.

lib/engine/train.rb
```diff
@@ -76,8 +76,9 @@ def id
     # if set ability must be a :train_discount ability
     def min_price(ability: nil)
       return 1 unless from_depot?
+      return @price unless ability
 
-      ability&.discounted_price(self, @price) || @price
+      Array(ability).map { |a| a.discounted_price(self, @price) }.min
```
crericha, line 81: Add support for private companies that have multiple train discount abilities.

tobymao (approved):

PR #6545, opened 2021-11-27
18Carolinas balance changes

scottredracecar:
-Add some tiles per playtesting feedback
-Change EMR to 1.5x penalty instead of 2

Requires pins

@roseundy is the EMR change okay? Does this need to be changed elsewhere?


scottredracecar:
Of course Steve has thought of everything ahead of time and had a spec for checking if the back sides match the front sides!

PR #6546, opened 2021-11-27
18Texas - Rule change, always allow two yellow tiles

scottredracecar:
I am trying to make this update via a website code change -- apologies if this fails and I will remove it and redo it later.


scottredracecar:
I am not sure if this requires pins, but it would be nice to preserve the old games for playtesting reference.

Issue #6169, opened 2021-08-02
1870 Company share redemption does not disable auto pass


zalasta524:
Not sure if this is an enhancement request or a bug, but I am in an 1870 game and had put on auto pass.  Prior to my turn someone performed a share redemption on a certificate I owned, shouldn’t that have disabled the auto pass as it was a direct action that affected me?  Furthermore, I am now unable to undo the pass because it will roll back everyone else that auto passed as well, which means I lost my turn to be able to buy?

magnusrk-ch:
There's a number of scenarios here.

1) Share redemption from yourself (non-president). Clearly shenanigans.
2) Share redemption from yourself (president). Clearly not shenanigans, but also irrelevant since it can't happen while you're auto passing.
3) Share redemption from someone else. Shenanigans, since it's functionally a sale by that player, and sales are shenanigans.
4) Share redemption from market. Probably not shenanigans, but I'm sure we could come up with a scenario.

So we have three real cases, two of which are shenanigans and one maybe shenanigans. Since share redemption is limited to once per SR I think we should keep it simple and always treat share redemption as shenanigans.

magnusrk-ch:
There's also a related scenario where a company buys shares _other than its own_ from a player. This is also functionally a sale and should be shenanigans.

I'm not sure this is actually possible in any implemented titles, but it would be easy to support as part of the same change.

crericha:
I agree with @magnusrk-ch's proposal. It's the safer route. We can always remove scenario 4 later if people complain.

Issue #6529, opened 2021-11-23
1830 did not deactivate autopass


misterallentrains:
Game #62394.

I set autopass in SR3 and it did not deactivate at the start of OR3.1.  It didn't deactivate until OR4.1 started, which meant that I missed all of SR4 (and I had company starting money, too!). 

I'm not complaining, I just don't want this glitch to happen again, ha ha.

Thanks for running such a great site!
Michael

magnusrk-ch:
Based on the action log, the following happened.

* 10407 (ConstantineXIII) has an auction action to buy NYC. Then they pass (103) and set up an auto pass (104).
* 105: 2419 (MeepleSteve) passes and 9007 (Orgak) auto passes.
* 106: 3799 (SwanofTrumpeter) passes.
* 107: 2522 (misterallen) sets up auto pass, which immediately triggers a pass, and then 10407's (ConstantineXIII) auto pass triggers.
* **At this point we're at all pass, SR is over, OR has begun. The auto action is cleared and should have produced a log message saying so.**
* **108: 2522 (misterallen) sets up auto pass again.**
* 109: NYC operates
* [...]
* No further round changes until next SR starts, and auto action is still valid.

My guess to what happened is that you
1) set up your auto action the first time
2) noticed it was not enabled 
3) didn't realize that it had immediately taken effect, triggered an all pass and then cleared itself because the SR was over 
4) set it again to fix it

tobymao:
closing then as user error

crericha:
I suggest we highlight the auto menu when the player has an auto order activated, so that it's obvious. Maybe change the color from black to green? Thoughts?

magnusrk-ch:
There is a green checkmark next to Auto when you have a program enabled.

Issue #6367, opened 2021-09-23
Auto pass in stock round cancels incorrectly


curlett:
If someone sells a stock then you turn on auto pass I would expect it to pass when it gets back to your turn.

Instead it cancels sure to shares sold, even if the share was sold after you turned on auto pass.

![Screenshot_20210922-184535_Chrome](https://user-images.githubusercontent.com/11777262/134443853-4e4a9922-4493-41a5-a43a-5453b263b1e0.jpg)

magnusrk-ch:
This looks like a duplicate of #5215

Issue #6357, opened 2021-09-16
Auto pass disabled by take_loan in SR on 1817NA. 


joflynn:
In SR 1. One player is maxing loans to drive up interest rates before OR 1.1. 

Would like to auto pass through the SR. 

Log reports:

[14:07]-- Disabled programmed action due to 'Unknown action take_loan disabling for safety’

Expected action, take loan doesn’t disable auto pass. 

game id 56011

perwestling:
This is basically the same issue as #5861 although that is for 1817.

magnusrk-ch:
Can someone familiar with the 1817 family (and ideally other games that support loans) comment on whether taking a loan should ever be considered shenanigans? Otherwise it's trivial to stop it from cancelling auto pass.

scottredracecar:
Yes, 1817 loans should certainly be considered shenanigans.

magnusrk-ch:
So another example of a situation that's definitely shenanigans, but the user is aware of it and would like the option to ignore the shenanigans and keep auto-passing.

joflynn:
Would it still be possibly shenanigans on SR1? That was the concern.

scottredracecar:
Whether yes or no, I don’t think the autopass rules should be sophisticated.

1877 (which us based off 1817) is definitely yes in SR1. This particular play, maybe not. Possibly sometimes yes though.

There should perhaps be a “hard autopass.” That completely ignores all shenanigans.

PR #6561, opened 2021-12-01
Add "Auto pass" button in stock rounds.

magnusrk-ch:
The button enables the auto pass in stock round program, same as the auto tab, but more accessible. The motivation is half to encourage more use of auto pass and half to make it less annoying when auto pass keeps getting cancelled over and over because someone sells and buys and sells and buys...

Only displayed for the active player, only in stock around, and does not show up in hotseat games.

autopass

### DIFF
--- a/assets/app/view/game/pass.rb
+++ b/assets/app/view/game/pass.rb
@@ -11,7 +11,10 @@ module View
 
       def render
         children = []
-        children << h(PassButton, before_process_pass: @before_process_pass) if @actions.include?('pass')
+        if @actions.include?('pass')
+          children << h(PassButton, before_process_pass: @before_process_pass)
+          children << h(PassAutoButton) if @game.round.stock? && @game.active_players_id.index(@user['id'])
+        end
         h(:div, children.compact)
       end
     end
--- a/assets/app/view/game/pass_auto_button.rb
+++ b/assets/app/view/game/pass_auto_button.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+
+module View
+  module Game
+    class PassAutoButton < Snabberb::Component
+      include Actionable
+
+      def render
+        props = {
+          on: {
+            click: lambda do
+              process_action(Engine::Action::ProgramSharePass.new(@game.current_entity))
+            end,
+          },
+        }
+
+        h(:button, props, 'Auto pass')
+      end
+    end
+  end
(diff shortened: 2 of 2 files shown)


crericha-qualitrol:
I've seen a number of issues posted regarding auto pass not working properly. There are warnings about using it in the "Auto" tab, so I feel okay about the use at your own risk nature of the button there. Promoting the button to the main screen hides all those warnings.

I'm not necessarily opposed to the change, but I do think the above should be considered before pulling the trigger.

crericha-qualitrol:
Or first fix these issues first:
https://github.com/tobymao/18xx/issues/5215
https://github.com/tobymao/18xx/issues/6169
https://github.com/tobymao/18xx/issues/6324
https://github.com/tobymao/18xx/issues/6357
https://github.com/tobymao/18xx/issues/6367

magnusrk-ch:
My thinking here is that the button doesn't promise anything. It doesn't say "auto pass unless X, Y and Z" or link to details. The only way it can fail to meet expectations is by _not_ passing. That's a safe way to fail, and also self-explaining (there's a comment in the log saying why it was cancelled).

In the case of someone who has read the auto tab and actually expects the automation to stop after shenanigans, they're no worse off with the new button than they were using the auto tab directly.

crericha-qualitrol:
Take a look at the linked bugs. A few of them are related to auto pass passing when it shouldn't.

crericha-qualitrol:
It does look like the warnings about using auto pass have been removed from the Auto menu. And I'm not saying don't promote the button to the main page. I'm saying think through if it is ready for that or if these bugs should be addressed first.

scottredracecar:
I took a look at those issues and most of them are the autopass being
overly conservative and turning itself off. The few that are most
relevant/problematic are game specific, such as 1870 price protection. I
support adding the button as is.

tobymao:
i have no preference or opinion, @crericha-qualitrol i'll leave the decision to you

crericha:
@magnusrk-ch, good comments on the issues I linked. Would you be able to fix #6169? If that one is resolved, I'd be okay with this PR.

magnusrk-ch:
I've added a proposal for how to fix it, let's see if anyone has other opinions.

jasonab:
There's also #6529 as an autopass issue

magnusrk-ch:
#6583 and #6575 propose fixes for two of the above issues.

PR #6575, opened 2021-12-03
Don't disable auto actions for shenanigans that happened before the p…

### LINKED ISSUE #5215: Do not disable auto-pass, if the conditions haven't been met since enabling it
In this game: https://18xx.games/game/39177 I turned on auto-pass, then someone passed. It was my turn then and my auto-pass had been disabled:
```
[09:58]-- Enabled programmed action Buy Shares
[18:03]Miranda passes
[18:03]-- Disabled programmed action due to 'Shares were sold'
```
 I'm guessing that happened, because the engine considers every action since the user's last action. It would be nice though, if it only considered the actions taken after enabling auto-pass.

magnusrk-ch:
…rogram was enabled.

Fixes the case where someone e.g. sells shares, you enable auto-pass, and your auto-pass gets cancelled because there was shenanigans since your last turn.

Fixes #5215

### DIFF
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -375,13 +375,16 @@ module Engine
         end
       end
 
-      def should_stop_applying_program(entity, share_to_buy)
+      def should_stop_applying_program(entity, program, share_to_buy)
         # check for shenanigans, returning the first failure reason it finds
         @round.players_history.each do |other_entity, corporations|
           next if other_entity == entity
 
           corporations.each do |corporation, actions|
             actions.each do |action|
+              # ignore shenanigans that happened before the program was enabled
+              next if action.id < program.id
+
               reason = action_is_shenanigan?(entity, other_entity, action, corporation, share_to_buy)
               return reason if reason
             end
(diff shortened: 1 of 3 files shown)


lib/engine/step/buy_sell_par_shares.rb
```diff
@@ -375,13 +375,18 @@ def action_is_shenanigan?(entity, other_entity, action, corporation, share_to_bu
...
+        program_created_at = Time.at(program.created_at)
+
         # check for shenanigans, returning the first failure reason it finds
         @round.players_history.each do |other_entity, corporations|
           next if other_entity == entity
 
           corporations.each do |corporation, actions|
             actions.each do |action|
+              # ignore shenanigans that happened before the program was enabled
+              next if Time.at(action.created_at) < program_created_at
```
tobymao, line 388: why don't you just compare action_ids?

magnusrk-ch: Done

magnusrk-ch:
It seems possible that this could break games. Rake doesn't show anything, though.

magnusrk-ch:
Investigating a crash caused by this in 1822. Actions without id.

PR #6493, opened 2021-11-07
[18NY] Coal tokens and a bit more

crericha:
- Claim coal tokens (auto actions)
- Enhance auto actions to use the active step after processing the last action, not before processing
- Minor fix-ups to connection bonuses
- Implement dividend step

PR #6603, opened 2021-12-08
step_for should stop when it reaches a blocking step. Fixes #6601

### LINKED ISSUE #6601: Error when undoing tile lay
https://18xx.games/game/63872?action=244

Lay a yellow tile in L27 south toward Miami, click undo, try to lay an upgrade in J27. Results in this error. Reproducible if you clone the game. Occurs in multiple titles, not just 18FL.

Screen Shot 2021-12-07 at 13 14 07

crericha:


KevinR1000:
Digging through recent commits, this may have been related to the following pull requests.
It might be worth revisiting these holistically to see whether there is a better way.

These made changes to assets/app/view/game/part/city_slot.rb related to blocking actions:
https://github.com/tobymao/18xx/pull/6539
https://github.com/tobymao/18xx/pull/6553

Also, #6493, #6494, and #6497 dealt with auto actions and modified lib/engine/round/base.rb.  I don't think this is related, however.

tobymao (approved):

PR #6608, opened 2021-12-09
Add option to automatically switch auto-buy to auto-pass when done.

### LINKED ISSUE #6483: Automatically Auto-pass after float
Currently, when the "Auto Buy Shares" system is done buying, it just sits there waiting for the player to manually pass.  Under many circumstances, this will be the last action the player intends to take during that SR, and so they immediately just turn on "Auto Pass" instead.

It would be convenient to have an extra checkbox in the auto-buy section for "Automatically pass after purchases" (or something to that effect - that wording may be a bit too ambiguous) where it'll convert the auto-buy order into an auto-pass order once the target has been reached.

magnusrk-ch:
The switch only happens if the auto-buy program completes successfully, not if it gets interrupted by shenanigans or inability to buy the specified shares.

Fixes #6483

### DIFF
--- a/lib/engine/action/base.rb
+++ b/lib/engine/action/base.rb
@@ -5,6 +5,7 @@ require_relative '../helper/type'
 module Engine
   module Action
     class Base
+      include Comparable
       include Helper::Type
 
       attr_reader :entity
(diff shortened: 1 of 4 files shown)


lib/engine/action/base.rb
```diff
@@ -75,6 +75,11 @@ def copy(game)
       def free?
         false
       end
+
+      def happened_before?(other)
```
tobymao, line 79: can you instead do 

def <=> 

to change the sorting, and then you can remove the usage of Time.at in the other section?

magnusrk-ch: Okay, done.

I already removed the old code from should_stop_applying_program in the original commit, are you referring to some other section?

tobymao: that's what i meant

PR #6612, opened 2021-12-09
[18Texas] plain track is nearly unlimited

### LINKED ISSUE #6611: [18Texas] Plain Yellow Track Should be Unlimited
The end of rule 12.2 stipulates that plain yellow track should be unlimited, but we've run out! (https://18xx.games/game/63174?action=405)

benjaminxscott:
closes #6611


scottredracecar:
Don't merge this. I don't have a good way to check how many tiles are being
used if it's actually unlimited (unless I were to count them, yuck).

Maybe just bump it to 30 which is "practically unlimited" and then I can
review games to see how many are getting used. Worst case scenario seems to
be around action 400-450.

benjaminxscott:
You're keeping track of how many tiles are used for the physical printing?

scottredracecar:
Please revert this. I changed the rule now so that the tile mix is intended to be a hard limit.

Which I think was my original intention anyway because the tile mix is exactly the same as 1851 and has no provision like this (a remnant I reused from another rulebook).

tobymao:
done @scottredracecar

PR #6621, opened 2021-12-11
[1889] use fancy logos

benjaminxscott:
Added optimized SVGs

Since it looks like the physical printing of 89 is using the fancy logos, could we use the fancy logos in the online game?
image

AFAIK previously we decided not to use fancy logos since it wasn't clear at the time what the physical version was going to use

If people want simple logos they can still use the 'Simple Logos' option to have the lettered logos as before

Screenshot 2021-12-11 12 19 52 PM
Screenshot 2021-12-11 12 20 01 PM
Screenshot 2021-12-11 12 20 15 PM


scottredracecar:
Agree, now with simple logos it is a different story.

PR #6625, opened 2021-12-12
[1893] fix dusseldorf to be untokenable and fix value

### LINKED ISSUE #6623: [1893] Dusseldorf/Nuess red area is incorrect
In my copy of 1893:Cologne, the Dusseldorf/Neuss red area in A4/A5 has a value of 20/30/50, not 10/30/50; and it doesn't have a station circle!  The 18xx.games shows this area as a station, which allows a corporation to place a station marker here.  That isn't what the physical map shows.
My copy of the game is version 2.0 and was purchased this year.
IMG_20211212_173729177

benjaminxscott:
Needs pins 

Checked [on BGG](https://boardgamegeek.com/image/2089180/1893-cologne) that this is the case

Fixed revenue in yellow to be 20, Made it a town so it's still a revenue location but is not tokenable

closes #6623

image


benjaminxscott:
The fixture was broken by this change, as the revenue changed

PR #6635, opened 2021-12-15
[18USA] Implement P15 and P20

crericha:


lib/engine/game/g_18_usa/step/track.rb
```diff
@@ -21,6 +21,18 @@ def actions(entity)
             actions
           end
 
+          def can_lay_tile?(entity)
+            super || can_place_token_with_p20?(entity)
+          end
+
+          def can_place_token_with_p20?(entity)
+            entity.companies.include?(@game.company_by_id('P20')) &&
+            !entity.tokens.reject(&:used).empty? &&
```
tobymao, line 30: entity.tokens.any?&:used) or whatever is the correct logic

tobymao (approved):

lib/engine/game/g_18_usa/step/track.rb
crericha: Fixed.

PR #6637, opened 2021-12-15
Check for auto-buy completion before checking available actions.

magnusrk-ch:
Auto-buy programs would not mark themselves as complete if buy_shares was not an available action (a common scenario when spending all your money to float). Formerly only a nuisance, but with the option to transition to auto-pass it actually breaks that functionality. Easy fix, just check for completion before checking if buying is allowed.

This *will* change behavior for buy X shares programs in buy-sell titles. They would previously generate a pass action after the last buy and terminate on the next turn. Now they will terminate after the last buy and not generate a pass (unless conversion to auto-pass is enabled). This seems like an improvement since the option to auto-pass exists, and the default behavior has been aligned with buy until float.

I've fixed the one fixture that was affected by this. My understanding is that real games won't need patching because the generated pass action remains valid even if the code would no longer generate it.

### DIFF
--- a/spec/fixtures/1889/hs_lhglbbiz_17502.json
+++ b/spec/fixtures/1889/hs_lhglbbiz_17502.json
@@ -374,7 +374,8 @@
       "created_at": 1615311276,
       "corporation": "IR",
       "until_condition": 1,
-      "from_market": true
+      "from_market": true,
+      "auto_pass_after": true
     },
     {
       "type": "pass",
@@ -393,6 +394,21 @@
             "IR_4"
           ],
           "percent": 10
+        },
+        {
+          "type": "program_disable",
+          "entity": 671,
+          "entity_type": "player",
+          "created_at": 1615311279,
+          "reason": "1 share(s) bought in IR, end condition met"
+        },
+        {
+          "type": "program_share_pass",
+          "entity": 671,
+          "entity_type": "player",
+          "indefinite": false,
+          "created_at": 1615311279,
+          "unconditional": false
         },
         {
           "type": "pass",
@@ -451,7 +467,7 @@
           "entity": 671,
           "entity_type": "player",
           "created_at": 1615311545,
-          "reason": "1 share(s) bought in IR, end condition met"
+          "reason": "LenaC bought on corporation TR and is unsecure"
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


spec/fixtures/1889/hs_lhglbbiz_17502.json
```diff
@@ -394,6 +395,21 @@
...
+          "entity": 671,
+          "entity_type": "player",
+          "created_at": 1615311279,
+          "reason": "1 share(s) bought in IR, end condition met"
+        },
+        {
+          "type": "program_share_pass",
+          "entity": 671,
+          "entity_type": "player",
+          "indefinite": false,
```
magnusrk-ch, line 409: Question: would it be better to not output fields that have the default value?

tobymao: sure no need

magnusrk-ch: Done

lib/engine/action/base.rb
```diff
@@ -58,7 +58,7 @@ def to_h
           'created_at' => @created_at.to_i,
           'auto_actions' => @auto_actions.empty? ? nil : @auto_actions.map(&:to_h),
           **args_to_h,
-        }.reject { |_, v| v.nil? }
```
tobymao, line 61: is this right? i think the intent was to explicitly check for nil, it’s not necessarily true you want to get rid of false here

magnusrk-ch: It's true that someone could introduce an action param which defaults to true. However, that's a bad idea and I think it's fine not to support it. (Better to invert the semantics of the param to default to false.)

Just to be sure, I scanned through the entire database dump I received. The only params to appear with value false are program params: auto_pass_after, from_market, indefinite, skip_buy, skip_track, skip_close, unconditional. These are all prepared to handle nil/absent values. (Values like auto_pass_after already have to deal with nil at runtime because they were extensions to an already existing program.)

Filtering false values has the same benefits as filtering nil, e.g. smaller payloads, less boilerplate in the representation, fewer diffs in fixtures when adding/changing/removing fields.

tobymao: i don't think it's safe to do this, let's get rid of this change

magnusrk-ch: Okay, reverting.

PR #6638, opened 2021-12-15
1825: 4+4 fix, leasing fixes

### LINKED ISSUE #6518: [1825] Company in receivership cannot upgrade a tile
Game: https://18xx.games/game/60178
Action: https://18xx.games/game/60178?action=538

LNWR is in receivership and it is it's turn to operate.
Try and upgrade hex R12 with any tile, you get the error message 'Cannot upgrade a city/town without a train', I believe the company should be able to lease a train on it's turn in order to use that for the purposes of upgrading a tile.  This error is given on almost all upgrades you attempt.  Interestingly it will allow an upgrade of hexag S13 but I'm assuming this is because it is a lay not a promotion (that space has yet to be upgraded)

I think the correct behaviour is that the train lease should be permitted from the start of the turn to allow LNWR to perform upgrades within that train length then it can run that train, must withhold revenue and then will pay £10 for the train lease out of it's cash.

### LINKED ISSUE #6519: [1825] Company in receivership given limited choice of train lease when Advanced Trains are in play
Game and action: https://18xx.games/game/60178?action=615

LNWR is in receivership and is about to run it's trains, it's only being offered the 3T for lease.
The rules aren't very clear here but from the base rules:

6.4.3 A Company in Receivership may lease a Train from the unsold stock in the Bank for £10 per turn, This Train remains the property of the Bank at all times. Only one Train may be leased by a Company. (A Train may still be leased even if the Company already owns a Train or Trains.) The leased Train will be the next one on offer for sale. If the Train is of a new type it will not calise a Phase change. (The Train may still run its full distance.)

Advanced Trains rules:
These trains are available for the price shown on the card after the relevant phase has been introduced as described in 4.7.  They are alternatives to whatever other type of Train is currently on offer.

I see two possible interpretations:
All currently available for purchase trains are options for leasing (I believe this is the intent despite the slightly wooly rules)
OR
The next 'mainline' train is available for leasing and the advanced trains are ignored (I don't think this is a good interpretation)

Certainly what is being offered now (LNWR is only offered a 3T for lease) is wrong, why the 3T and not the U3?  Those are made available at the same time.

### LINKED ISSUE #6629: [1825] 4+4E trains aren't doubling correctly
https://18xx.games/game/60178?action=1877

Pick ANY 4 legal cities for the 4+4E, it will behave like an E in that it can run to an unlimited number of cities, but it's not correctly handling the + part of the train and doubling the resultant income

roseundy:
Fixes #6518 
Fixes #6519 
Fixes #6629 

Leasing trains now allowed in track/token step. Leasing now allows choice of available trains.

### DIFF
--- a/lib/engine/game/g_1825/step/track_and_token.rb
+++ b/lib/engine/game/g_1825/step/track_and_token.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative 'lease_train'
 require_relative '../../../step/track_and_token'
 
 module Engine
@@ -7,10 +8,20 @@ module Engine
     module G1825
       module Step
         class TrackAndToken < Engine::Step::TrackAndToken
+          include LeaseTrain
+
           def actions(entity)
             return [] if @game.silent_receivership?(entity)
+            return [] if entity != current_entity
 
-            super
+            actions = []
+            if can_lay_tile?(entity)
+              actions << 'lay_tile'
+              actions << 'special_buy' if entity.receivership? && !@round.leased_train && !@game.leaseable_trains.empty?
+            end
+            actions << 'place_token' if can_place_token?(entity)
+            actions << 'pass' if actions.any?
+            actions
           end
 
           def description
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1825/step/track_and_token.rb
```diff
@@ -1,16 +1,27 @@
...
+            return [] if entity != current_entity
 
-            super
+            actions = []
+            if can_lay_tile?(entity)
+              actions << 'lay_tile'
+              actions << 'special_buy' if entity.receivership? && !@round.leased_train && !@game.leaseable_trains.empty?
+            end
+            actions << 'place_token' if can_place_token?(entity)
+            actions << 'pass' if actions.any?
```
tobymao, line 23: unless actions.empty?

PR #6644, opened 2021-12-17
use "Shikoku 1889" as 1889's display title

michaeljb:
* update 1889 references in announcements, new player recommendations, and tutorial
* simplify 1889's location to "Japan" so that the string in the new game dropdown
  is not so repetitive
* Add `GAME_DISPLAY_TITLE` constant to game meta, set 1889's display title to 'Shikoku 1889'
* reorganize constants in the base `Game::Meta` module so all the title stuff has
  its own section
* add `GAME_DISPLAY_TITLE` and corresponding functions `display_title` and
  `Engine.closest_display_title`
    * use `display_title` for game cards, home page filter, new game dropdown,
      game page header, tab title
* remove `GAME_SUPERTITLE`; 1889 was the only one using it, now it uses
  `GAME_DISPLAY_TITLE` to ditch the colon (i.e., 'Shikoku: 1889' -> 'Shikoku
  1889')

Before the Kickstarter went live, Josh messaged me asking if the colon could be removed from the title on the new game page, and if the title could be updated everywhere. I finally got around to setting up 18xx.games on my new machine today.

~While this change would not _require_ migrating all past "1889" games to "Shikoku 1889", that would probably be desired so that all the games in the database would be internally consistent, and one would not need to query both for '1889' and 'Shikoku 1889' to find all the plays for this title.~


tobymao:
i don’t want to change the name in the database, this pr should only change the display name and not the title

michaeljb:
With 45d01bd6c, 1889 will appear as "Shikoku 1889" in the UI but will be unchanged in the database, and no migrations needed.

lib/engine/game/meta.rb
```diff
@@ -53,10 +54,15 @@ def title
...
+        def display_title
+          @display_title ||= (self::GAME_DISPLAY_TITLE || title)
+        end
+
         def full_title
           @full_title ||=
             self::GAME_FULL_TITLE ||
-            [self::GAME_SUPERTITLE, title, self::GAME_SUBTITLE].compact.join(': ')
+            self::GAME_DISPLAY_TITLE ||
+            [title, self::GAME_SUBTITLE].compact.join(': ')
```
michaeljb, line 65: It's just occurred to me that with `display_title` being a thing now, `full_title` would be better described as `long_display_title`.

Would you have any issue with me renaming the `full_title` constants and variables in the code?

tobymao (approved):

PR #6652, opened 2021-12-19
Fix ability_right_time? (on_phase check)

Galatolol:
I think I found an error. The way on_phase checking was handled resulted in such ability:

```
              {
                type: 'revenue_change',
                revenue: 0,
                on_phase: 'Brown',
                owner_type: 'corporation',
              },
```

triggering upon the sale of the company. Indeed the whole line was skipped since `on_phase` was `nil`.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2559,7 +2559,7 @@ module Engine
       def ability_right_time?(ability, time, on_phase, passive_ok, strict_time)
         return true unless @round
         return true if time == 'any' || ability.when?('any')
-        return (on_phase == ability.on_phase) || (on_phase == 'any') if on_phase
+        return (on_phase == ability.on_phase) || (on_phase == 'any') if ability.on_phase
         return false if ability.passive && !passive_ok
         return true if ability.passive && ability.when.empty?
 


lib/engine/game/base.rb
```diff
@@ -2559,7 +2559,7 @@ def ability_usable_this_or?(ability)
       def ability_right_time?(ability, time, on_phase, passive_ok, strict_time)
         return true unless @round
         return true if time == 'any' || ability.when?('any')
-        return (on_phase == ability.on_phase) || (on_phase == 'any') if on_phase
+        return (on_phase == ability.on_phase) || (on_phase == 'any') if ability.on_phase
```
magnusrk-ch, line 2562: Shouldn't it be ability.on_phase == 'any' as well then?

Galatolol: I did test it, but it provoked some unexpected problems: the following ability didn't work.
```
              {
                type: 'close',
                on_phase: 'never',
                owner_type: 'corporation',
              },
```
Possibly removing the whole second part, i.e. ` || (on_phase == 'any')` could make sense though. I'll check later how it behaves.

tobymao: will you check this?

Galatolol: I plan to do it tomorrow, not having time today.

Galatolol: I tested it. Both

 `return (on_phase == ability.on_phase) if ability.on_phase`

and

`return (on_phase == ability.on_phase) || (ability.on_phase == 'any') if ability.on_phase`

result in the following ability not working:
```             
              {
                type: 'close',
                on_phase: 'never',
                owner_type: 'corporation',
              },
```

Whilst with

`return (on_phase == ability.on_phase) || (on_phase == 'any') if ability.on_phase`

both that ability and the ability I mentioned in the description of the PR work.

magnusrk-ch: My initial comment was based on a misunderstanding. Abilities never use `'any'` as an on_phase value, only actual phases or `'never'`.

Galatolol: @tobymao I've checked this with several different companies and they all worked as expected.

tobymao (approved):

lib/engine/game/base.rb
roseundy: This change broke 18Texas. It uses "on_phase: %w[3 4]," in it's ability definition for several companies.

roseundy: I think 18Texas relied on this not working correctly.

roseundy: So why was this changed again?

roseundy:
I think this fix is incorrect or incomplete. It looks like this fix assumes on_phase will only be set for abilities that trigger on phase changes. 18Texas uses on_phase to qualify _during_ a phase. I had to add an override to ability_right_time? to get this to work. This fix breaks it again.

tobymao:
i have no idea, is 18texas broken or this patch? if 18texas is the only game that this breaks for, maybe 18texas is wrong

roseundy:
This patch broke 18Texas. I have a fix for that (doesn't touch any common code), but I feel that the "on_phase" documentation is misleading - especially with this patch.

roseundy:
Also, given what you said about on_phase really only being used for phase transitions, does this patch make sense?

magnusrk-ch:
I'm leaning towards the right thing being checking *both* `on_phase` and `ability.on_phase`, but more details about the original breakage would be good.

PR #6668, opened 2021-12-23
1832 Inital Files

CaptainObi:
These are the inital commits for 1832, sorry for the large commit but it's mainly config and such.


tobymao:
cleanup the svgs

CaptainObi:
done

PR #6674, opened 2021-12-25
Add rake action spec_parallel for running rspec in parallel.

magnusrk-ch:
Also set as the default rake action instead of regular spec.

Runtime on my machine (quad core) is reduced by 25-50%, depending on overall system load.

50% is the maximum possible speedup as there's a single spec that takes half the overall time, and parallel_tests won't split up individual specs. Splitting up the large specs (assets_spec and game_spec) is one option, investigating e.g. parallel_split_tests is another.

### DIFF
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -1,0 +1,5 @@
+-I assets/app/
+--format progress
+--format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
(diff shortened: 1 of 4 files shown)


magnusrk-ch:
PR got messed up somehow, looking.

magnusrk-ch:
> PR got messed up somehow, looking.

Should be sorted now.

PR #6678, opened 2021-12-28
18SJ: Fix side effects for E trains

### LINKED ISSUE #6676: [18SJ] Non E trains allowed to go through SJ token.
https://18xx.games/game/65110?action=656

My Corp has an E train and a 6 train. However if you look at the route, the 6 train is able to go through an SJ token while I believe that only the E train should be able to do it.

perwestling:
Contains two fixes:
- Making SJ tokens passable is now done after the token step
  so that current entity cannot affect track or token through
  a blocked city, even if SJ token is there.
- If current entity makes a non-E route it cannot pass through
  a blocked city, even if SJ token is there.

Fixes #6676

### DIFF
--- a/lib/engine/game/g_18_sj/game.rb
+++ b/lib/engine/game/g_18_sj/game.rb
@@ -1602,13 +1602,10 @@ module Engine
 
           # Check if a stop (excluding start and stop) does not allow pass-through
           # due to temporary passable SJ token
-          passthrough = stops.dup
-          passthrough.shift
-          passthrough.pop
-          passthrough.each do |s|
+          stops[1...-1].each do |s|
             next if !s.city? || s.tokened_by?(route.train.owner)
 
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_sj/game.rb
```diff
@@ -1595,6 +1596,21 @@ def player_count
           # to handle setup of cash and cert limit so that it includes the bot.
           two_player_variant ? 3 : @players.size
         end
+
+        def ensure_route_does_not_passthrough_blocked_city(route, stops)
+          return if stops.size < 3 || !owns_electric_train?(route.train.owner) || route.train.name == 'E'
+
+          # Check if a stop (excluding start and stop) does not allow pass-through
+          # due to temporary passable SJ token
+          passthrough = stops.dup
```
tobymao, line 1605: stops[1…-1]

perwestling: Ah....
Fixed with new commit.

PR #6680, opened 2021-12-28
18NewEngland: add majors, fix map, ISR, SR

roseundy:
Add major corps
Fix missing tiles on map
Move to physical map coordinates
Break entities and map out of game file
Implement first stock round (snake order, reservations, etc)
Implement stock round


tobymao (approved):

assets/app/view/game/corporation.rb
```diff
@@ -66,7 +66,7 @@ def render
 
         children = [render_title, render_holdings]
 
-        unless @corporation.minor?
+        unless @game.corporation_is_minor?(@corporation)
```
tobymao, line 69: why?

crericha: Does it break a lot of games if you add `corp.type == :minor` to the `corporation.minor?` logic?

roseundy: This game uses the 1822 style of minors. This means that corporation? is true, minor? is false, but type is set to :minor

I want these to display like minors, but I didn't want to affect how 1822 worked, so I created and used a game method corporation_is_minor? to adjust the appearance in the UI.

crericha: There are a ton of games that use corp.type == :minor, not just 1822. I worry this method will increase the confusion between the two styles of minors. How about creating a method like show_full_coporation_card?(corporation) or something like that to determine if the share price, ownership, etc. is shown on the card. I think that's more specific to what you are trying to do.

tobymao: that doesn’t make sense, why isn’t it just a minor then?

roseundy: @tobymao - I can't speak to the other games, but in 18NewEngland, I need a minor that has a share price and could be parred instead of merely bought. Please bear in mind, I didn't create this type of minor. It already existed and I'm just leveraging it.

@crericha - Is your main concern about the method name? I can change it.

scottredracecar: 1822-style minors that have a share price and various train buying/owning features are the norm, not 46-style private company minors.

roseundy: I followed @crericha's suggestion and renamed the corporation_is_minor? method

crericha: Looks good to me.

The other thing I asked is if you tried to modify the minor? method by adding the type == :minor logic to it.

roseundy: I thought about it, but that would alias the two minor variations together. I didn't want to break "type" minors in other games.

roseundy: This may be a good idea, but I don't want to bog down this game with refactoring a bunch of other games.

roseundy: To answer your question specifically, that change would affect how minors in 1822 are displayed.

PR #6688, opened 2021-12-30
18NewEngland North - Revise companies, map, add logos

scottredracecar:


roseundy (approved):
Looks good to me

roseundy:
Need to compress all of the svgs, though. I've an easy flow through Inkscape if you want me to do it.

roseundy:
Ignore that last comment. These are already optimized. I missed that they aren't using text objects.

PR #6694, opened 2022-01-01
18NY - Update colors/logos for v2

scottredracecar:


lib/engine/game/g_18_ny/entities.rb
```diff
@@ -311,7 +311,7 @@ module Entities
             always_market_price: true,
             float_percent: 20,
             coordinates: 'J20',
-            color: '#ED1C24',
+            color: 'purple',
```
crericha, line 314: Please use the hex code to stay consistent with the rest of the implementation. The purple you emailed me is #8C4776. Apply to the SVGs too.

lib/engine/game/g_18_ny/entities.rb
```diff
@@ -351,7 +351,7 @@ module Entities
             always_market_price: true,
             float_percent: 20,
             coordinates: 'F26',
-            color: '#A7A9AC',
+            color: 'red',
```
crericha, line 354: Please use the hex code to stay consistent with the rest of the implementation. The B&A logo uses #E21F27. Apply to the SVGs too.

lib/engine/game/g_18_ny/entities.rb
scottredracecar: ok

lib/engine/game/g_18_ny/entities.rb
scottredracecar: ok

scottredracecar:
@crericha Okay, I added those hex codes. I prefer in general to use the named web safe colors as they are more vibrant on the screen. The purple hex code is a CMYK color converted to RGB and are more "dull." Let's go with this for now and then consider a more thorough standardization some other time. Other logos like D&H also don't match right now. I'd suggest "darkblue" for that one, for example.
image

crericha (approved):

PR #6709, opened 2022-01-05
Partial implementation of 18ChristmasEve

lkingsford:
Implementation includes:
- Map.
- Code to copy frames to new tiles. Frames are used to show rooms - happy for feedback if there was a more preferred way.
- Implementation of new private powers.

Implementation is presently missing the 'gift' rules.

### DIFF
--- a/lib/engine/game/g_18_christmas_eve/game.rb
+++ b/lib/engine/game/g_18_christmas_eve/game.rb
@@ -18,12 +18,43 @@ module Engine
         # rubocop:disable Layout/LineLength
         HEXES = {
           white: {
-            %w[C8 D11 D3 D7 D9 E4 F5 F11 H11 I2 I4 I8 J5 J7 K4 K12 L1 L3 L9] => '',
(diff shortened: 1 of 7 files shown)


lkingsford:
I'll note that it isn't in the approved spreadsheet yet - but I'm the designer, and for explicitness's sake, I do so approve.

lib/engine/game/g_18_christmas_eve/game.rb
```diff
@@ -0,0 +1,332 @@
...
+
+        def most_rooms?(route)
+          most = route.routes.max_by { |r| rooms_in_route(r) }
+          route == most
+        end
+
+        def nog_express?(route)
+          bar = route.visited_stops.find { |s| s.tile&.location_name == 'Bar' }
+          dc = route.visited_stops.find { |s| s.tile&.label&.to_s == 'DC' }
+          (bar and dc)
```
tobymao, line 313: don't use and

lkingsford: Fixed. Good spot.

PR #6732, opened 2022-01-09
18ChristmasEve Gift Shares and move to alpha

lkingsford:
I'm keen to know if there's another UI way I should have addressed the gift share buttons - they show all the time (as opposed to when you have selected that share). Happy for feedback. I also wasn't sure if the gift should have been something I added in 'action' rather than addressing the logic in the step.

I wasn't certain if there was anything else you needed before being comfortable going into alpha. I've run through a couple of games from start to end. I have a full JSON of one of those - but I didn't include as a spec (yet) as the game is still a prototype which will likely have future (hopefully minor) map/rule changes that will break those tests.

### DIFF
--- a/lib/engine/game/g_18_christmas_eve/game.rb
+++ b/lib/engine/game/g_18_christmas_eve/game.rb
@@ -278,6 +278,13 @@ module Engine
 
         }.freeze
 
+        def stock_round
+          Round::Stock.new(self, [
+            Engine::Step::DiscardTrain,
+            G18ChristmasEve::Step::BuySellParGiftShares,
+          ])
+        end
+
         def operating_round(round_num)
           Engine::Round::Operating.new(self, [
             Engine::Step::Bankrupt,
--- a/lib/engine/game/g_18_christmas_eve/game.rb
+++ b/lib/engine/game/g_18_christmas_eve/game.rb
@@ -302,9 +302,10 @@ module Engine
         end
 
         def status_array(corporation)
-          [] unless @round.respond_to?(:presidencies_gifted)
-          if @round.presidencies_gifted.include? corporation then
-            ["Can not gift presidents certificate again this round."]
+          return [] unless @round.respond_to?(:presidencies_gifted)
+
+          if @round.presidencies_gifted.include? corporation
+            ['Can not gift presidents certificate again this round.']
           else
             []
           end
--- a/lib/engine/game/g_18_christmas_eve/game.rb
+++ b/lib/engine/game/g_18_christmas_eve/game.rb
@@ -304,7 +304,7 @@ module Engine
         def status_array(corporation)
           return [] unless @round.respond_to?(:presidencies_gifted)
 
-          if @round.presidencies_gifted.include? corporation
+          if @round.presidencies_gifted.include?(corporation)
             ['Can not gift presidents certificate again this round.']
           else
             []
--- a/lib/engine/game/g_18_christmas_eve/game.rb
+++ b/lib/engine/game/g_18_christmas_eve/game.rb
@@ -14,7 +14,16 @@ module Engine
 
         SELL_BUY_ORDER = :sell_buy_sell
 
-        GAME_END_CHECK = { bankrupt: :immediate, stock_market: :current_round, bank: :full_or }.freeze
+        GAME_END_CHECK = {
+          bankrupt: :immediate,
+          stock_market: :current_round,
+          bank: :full_or,
+          custom: :one_more_full_or_set,
+        }.freeze
+
+        GAME_END_REASONS_TEXT = Base::GAME_END_REASONS_TEXT.merge(
+          custom: 'First D Purchased'
+        )
 
         # rubocop:disable Layout/LineLength
         HEXES = {
@@ -278,6 +287,53 @@ module Engine
 
         }.freeze
 
+        # Reimplement 18Ches trains for D end game trigger
+        TRAINS = [
+          {
+            name: '2',
+            distance: 2,
+            price: 80,
+            rusts_on: '4',
+            num: 7,
+          },
+          {
+            name: '3',
+            distance: 3,
+            price: 180,
+            rusts_on: '6',
+            num: 6,
+          },
+          {
+            name: '4',
+            distance: 4,
+            price: 300,
+            rusts_on: 'D',
+            num: 5,
+          },
+          {
+            name: '5',
+            distance: 5,
+            price: 500,
+            num: 3,
+            events: [{ 'type' => 'close_companies' }],
+          },
+          { name: '6', distance: 6, price: 630, num: 2 },
+          {
+            name: 'D',
+            distance: 999,
+            price: 900,
+            num: 20,
+            available_on: '6',
+            discount: { '4' => 200, '5' => 200, '6' => 200 },
+            events: [{ 'type' => 'd_trigger' }],
+          },
+        ].freeze
+
+        EVENTS_TEXT = Base::EVENTS_TEXT.merge(
+          'd_trigger' => ['D Trigger',
+                          'After the purchase of the first D, game ends end of the following OR set'],
+        ).freeze
+
         def stock_round
           Round::Stock.new(self, [
             Engine::Step::DiscardTrain,
@@ -342,6 +398,17 @@ module Engine
           rev_str += ' + Hat' if route.train.owner.companies.include?(hat) && most_rooms?(route)
           rev_str += ' + Nog' if route.train.owner.companies.include?(nog) && nog_express?(route)
           rev_str
+        end
+
+        @d_trigger = false
+
+        def event_d_trigger!
+          @log << 'First D purchased. Game will end at end of next set of ORs' unless @d_trigger
+          @d_trigger = true
+        end
+
+        def custom_end_game_reached?
+          @d_trigger
         end
       end
     end
--- a/lib/engine/game/g_18_christmas_eve/game.rb
+++ b/lib/engine/game/g_18_christmas_eve/game.rb
@@ -301,6 +301,15 @@ module Engine
           ], round_num: round_num)
         end
 
+        def status_array(corporation)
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_christmas_eve/game.rb
```diff
@@ -294,6 +301,16 @@ def operating_round(round_num)
           ], round_num: round_num)
         end
 
+        def status_array(corporation)
+          return [] unless @round.respond_to?(:presidencies_gifted)
+
+          if @round.presidencies_gifted.include? corporation
```
tobymao, line 307: missing parens

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+              }
+            )
+          end
+
+          def actions(entity)
+            return [] unless entity == current_entity
+
+            actions = super
+            actions << 'choose' if can_gift_any?
+            actions << 'pass' unless actions.empty? && actions.include?('pass')
```
tobymao, line 27: don't have multiple conditions with an unless

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+            actions
+          end
+
+          def choice_available?(entity)
+            return false unless @game.sellable_turn? && entity.corporation?
+
+            @choices = gift_choices(entity)
```
tobymao, line 34: you shouldn't be assigning @choices within a question mark method

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+            return [] unless entity == current_entity
+
+            actions = super
+            actions << 'choose' if can_gift_any?
+            actions << 'pass' unless actions.empty? && actions.include?('pass')
+            actions
+          end
+
+          def choice_available?(entity)
+            return false unless @game.sellable_turn? && entity.corporation?
```
tobymao, line 32: multiple conditions unless

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+
+            @choices = gift_choices(entity)
+            @choices != {}
+          end
+
+          def gift_choices(entity)
+            return {} unless entity&.corporation? && entity&.president?(current_entity)
```
tobymao, line 39: same

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+
+            choices = {}
+            # Cert only
+            # If we own 20%, then we've only got the pres cert to gift
+            if current_entity.percent_of(entity) != 20
+              choices.merge!(@game.players
```
tobymao, line 45: don't mix ! with non mutation methods, it's string

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+                .map.with_index { |p, i| [p, i] }
+                .reject { |p, _i| p == current_entity }
```
tobymao, line 47: don't combine so many rejects, can probably just use one

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+                .reject { |_p, _i| @round.presidencies_gifted&.include? entity }
+                .select { |p, _i| p.percent_of(entity).zero? }
+                .to_h { |p, i| ["prez_#{i}_#{entity.id}", "Presidency to #{p.name}"] })
+            end
+
+            choices
+          end
+
+          def can_gift_any?
+            @game.sellable_turn? && @game.corporations.any? { |e| gift_choices(e).any? }
```
tobymao, line 69: don't use any? for gift_choices, use empty?

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,101 @@
...
+          end
+
+          def choice_name
+            'Gift'
+          end
+
+          attr_reader :choices
```
tobymao, line 76: don't put this here

lib/engine/game/g_18_christmas_eve/game.rb
lkingsford: As in, `if (@round.presidencies_gifted.include? corporation)` or `@round.presidencies_gifted.include?(corporation)`?

tobymao: latter, x.y?()

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
lkingsford: I think I agree. I'm trying to come up with a better way - but in `stock.rb`:160, `choices` is called for corporations in line 11 of `choose.rb` without any parameters. `choice_available?` is the only function called with each corporation as a parameter - meaning that unless I change `choose.rb` and `stock.rb` to include an entity parameter, the choices have to be the same for each corporation (albeit - they can be shown on a per-corporation basis). 

In this case - I was trying to limit the impact of my changes outside of 18ChEv's classes as much as possible. In principal - would me making the change to `stock.rb` and `choose.rb` be acceptable? Or, should I add a 'company specific choice' to `stock.rb`, or was there something else - like another action - I could have used other than choices that would have been more appropriate?

tobymao: i don't understand why you need to set @choices, i don't have much context

lkingsford: A president of a corp may gift a cert to another player. They may gift only to a player that has none of that corp. They may gift a presidents certificate, if that player will have the most or equal most shares after gifting. The result of this is that the choices for one corp may be ['Gift cert to Player 1', 'Gift cert to Player 2'] and another corp may be ['Give prez to Player 3']. Choices shows what buttons are available for the `choose` action. If I don't change `@choices` each time (or change it into a function that accepts an entity), then the same buttons will appear for each company.

lkingsford: For instance:
image

tobymao: shouldn't the buttons only appear when you click on it though? you shouldn't have buttons appearing on corps that aren't selected

lkingsford: You are right - I thought they should too, but didn't seem to be a supported way of doing it with `choose`/`choice` - and so I thought, at least while alpha, this might be acceptable. The 1828 use of `choose` for merge uses merge specific code hiding in the view. Was trying to limit the scope of my changes (in part because I was nervous that putting company `choice` options in the corporation hidey button space would effect another game) - but I'll spend some time and see whether it will effect any other games.

lkingsford: I've attempted to resolve both of these in https://github.com/tobymao/18xx/pull/6732/commits/a344f4bd905881f5c04f6399f40e5329646489fd 
I think it's an appropriate approach - but happy to hear if you think (for instance) that there should be a whole new class of `choose` action, and want me to back this back out. 
I was honestly surprised that adding specific actions to the stock round for specific companies was novel in the code - I'd still be glad to hear that I missed something, and there was something else I could have used.

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
lkingsford: This because `!entity_choices(e).empty?` - I hope that's what you mean

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
```diff
@@ -0,0 +1,98 @@
...
+              }
+            )
+          end
+
+          def actions(entity)
+            return [] unless entity == current_entity
+
+            actions = super
+            actions << 'choose' if can_gift_any?
+            actions << 'pass' if actions.any? && !actions.include?('pass')
```
tobymao, line 27: !empty

lib/engine/game/g_18_christmas_eve/game.rb
lkingsford: Resolved

lib/engine/game/g_18_christmas_eve/step/buy_sell_par_gift_shares.rb
lkingsford: Resolved

lkingsford:
I'm sorry for pushing in that last change - I realised that the end game trigger rule wasn't implemented yet, and thought it was better before it went to alpha. If you'd prefer, I can take the commit back out of this PR for a future one.

lkingsford:
@tobymao Not sure if you were still waiting on me, but those changes you requested are all done

tobymao:
haven’t had time to review

tobymao (approved):

PR #6734, opened 2022-01-09
1825: move optional trains to end of train list

### LINKED ISSUE #6733: [1825] Unsold U3 trains incorrectly block purchase of 6
GameID; https://18xx.games/game/64648

**How to Reproduce the Bug?**

During a game of 1825 including phase 4 (K3) + advanced trains (K2) , having a company with enough cash to purchase a 6 (>650) while there was a U3 train left unsold, did not offer the 6 as a train available to purchase, blocking phase 4. In this particular game, the company is SECR.

From the printed 1825 K2.1 rules (it is identical in the K2 rules)

"Unsold trains of type '3T' and 'U3' do not prevent the purchse of a '6' train and the consequent introduction of Phase Four. Any unsold Trains of these types then remain available for purchase as alternatives"

**Which Browser / Platform?**

Mac and iOs, but this is a game logic bug.

roseundy:
Fixes #6733

### DIFF
--- a/lib/engine/game/g_1825/game.rb
+++ b/lib/engine/game/g_1825/game.rb
@@ -169,6 +169,8 @@ module Engine
           '3' => { distance: 3, price: 300 },
           '4' => { distance: 4, price: 430 },
           '5' => { distance: 5, price: 550 },
+          '6' => { distance: 6, price: 650 },
+          '7' => { distance: 7, price: 720 },
           '3T' => { distance: 3, price: 370, available_on: '3' },
           'U3' => {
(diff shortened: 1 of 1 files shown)


thepackrat:
I've got a different fix which allows trains to be marked as non-blocking so that they stay in the expected game order.

thepackrat:
take a look at https://github.com/tobymao/18xx/pull/6735 for the no-blocking attribute fix. I believe it's useful in other games also and then the trains come out in a list-lineear order.

roseundy:
Is there anything wrong with this fix?

PR #6735, opened 2022-01-09
Adds a does_not_block attribute to trains so they can be optional in purchae order

thepackrat:
This enabled 1825 to skip optional trains in train purchases.


roseundy:
This is not needed.

thepackrat:
The available_on would require a fake phase to be created that isn't in the rules so that 6 would be available only after all 5s were bought. This is similar to some alternate trains in EU and 24 (and other not in flight games)

PR #6765, opened 2022-01-12
Refactor buying_power

scottredracecar:
Replaces PR #6765

This PR:
- replaces base's buying_power with a generic version
- removes the now redundant generic def from 1822/18CO/18Texas/1849/18Ireland
- moves 1846's buying_power definition to its game
- removes the 18NewEngland fixture that broke because it relied on the old buying_power def

@tobymao Please check if this breaks any games. It should not.

@roseundy tested this for 18NewEngland and it will archive most games (please pin them)


tobymao:
bravo

roseundy:
FWIW, looks good.

PR #6784, opened 2022-01-14
[18NY] Move Philadelphia revenue left one hex, so that the offboard entrances are visible

crericha:


tobymao (approved):

lib/engine/game/g_18_ny/map.rb
```diff
@@ -182,6 +182,8 @@ module Map
         }.freeze
 
         SECOND_EDITION_LOCATION_NAMES = {
+          'K15' => 'Philidelphia',
```
benjaminxscott, line 185: note typo

crericha: Good catch

PR #6795, opened 2022-01-15
Linkify game references in chat.

### LINKED ISSUE #6773: Linkify game IDs in the chat
People often reference games in the chat, e.g. inviting players to join. We should convert game IDs to links.

Specifically:
* Linkify `https://18xx.games/game/123`
* Linkify `https://18xx.games/game/123?action=456`
* Don't linkify other URLs, to be on the safe side. Can add more patterns later as needed.
* Linkify `game 123`
* Linkify `#123`
* Don't linkify `123`, too many false positives.
* Links open in a new tab.

Bonus:
* Make the URL detector use the current domain instead of hardcoding `https://18xx.games`.

magnusrk-ch:
fixes #6773

chat_links

### DIFF
--- a/assets/app/view/chat.rb
+++ b/assets/app/view/chat.rb
@@ -82,7 +82,32 @@ module View
       h('div#chat.half', props, children)
     end
 
+    private
+
+    QUOTED_BASE_URL = Regexp.quote(`window.location.origin || ''`)
+    GAME_LINK_RE = Regexp.new("((?:game #?|#|#{QUOTED_BASE_URL}/game/)(\\d+(?:\\?action=\\d+)?))")
+
+    def parse_message(message)
+      message_parts = message.split(GAME_LINK_RE)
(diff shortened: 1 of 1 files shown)


assets/app/view/chat.rb
```diff
@@ -82,7 +82,32 @@ def render
       h('div#chat.half', props, children)
     end
 
+    private
+
+    QUOTED_BASE_URL = Regexp.quote(`window.location.origin || ''`)
```
tobymao, line 87: what happens if this is empty string?seems dangerous

magnusrk-ch: If the base URL is empty the third pattern reduces to `/game/`. This is not ideal but it neither causes a degenerate pattern nor is likely to produce many false positives.

The empty string only occurs in miniracer, and doesn't propagate to the actual rendered page because chat is loaded lazily post-load.

PR #6796, opened 2022-01-15
[1866] Run nationals

magnusfalt:
- Now you can run the national infinite train within its region.


lib/engine/game/g_1866/game.rb
```diff
@@ -510,6 +587,11 @@ def national_corporation?(corp)
           self.class::NATIONAL_CORPORATIONS.include?(corp.name)
         end
 
+        def national_within_region?(entity, visits)
+          hexes = self.class::NATIONAL_REGION_HEXES[entity.id]
+          visits.count { |v| hexes.include?(v.hex.name) } == visits.length
```
tobymao, line 592: size

PR #6804, opened 2022-01-16
[1866] Dividend

magnusfalt:
- Minior nationals only pays half and dont move on the stock market.
- Major nationals always payout full and always move one to the right on the stock market.
- National corporations can opnly lay tiles within its region, and the same when it run its infinite train.
- Corporation need operating rights to operate outside its own region.


lib/engine/game/g_1866/game.rb
```diff
@@ -482,6 +585,22 @@ def par_price_str(share_price)
...
 
+        def place_home_token(corporation)
+          return super unless corporation.id == 'PLM'
+          return if corporation.tokens.first&.used
+
+          corporation.coordinates.each do |coord|
+            hex = hex_by_id(coord)
+            tile = hex&.tile
+            cities = tile.cities
+            city = cities.find { |c| c.reserved_by?(corporation) } || cities.first
```
tobymao, line 596: cities[0]

lib/engine/game/g_1866/game.rb
```diff
@@ -563,33 +682,63 @@ def tile_lays(entity)
           self.class::TILE_LAYS
         end
 
+        def train_help(_entity, runnable_trains, _routes)
+          return [] if runnable_trains.empty?
+
+          entity = runnable_trains.first.owner
```
tobymao, line 688: runnable_trains[0].owner

tobymao (approved):

PR #6812, opened 2022-01-18
18NewEngland: allow merge with colocated minor

### LINKED ISSUE #6811: 18NewEngland: Not allowed to merge with minor in same hex
https://18xx.games/game/67962?action=187

HNH should be able to merge with NLN because it is co-located on the same hex, but isn't allowed to

roseundy:
Fixes #6811 

This could potentially break games. @tobymao Can you pin these?


tobymao (approved):

roseundy:
Never mind, I don't think this will break any games.

PR #6817, opened 2022-01-19
1848 Enable COM Start and Home Rules

lkingsford:
Includes a bug fix which could potentially (but probably hasn't) effected other games when the first company in an OR is skipped, or when all companies in an OR is skipped. See c2cae9a.

### DIFF
--- a/lib/engine/game/g_1848/step/operating.rb
+++ b/lib/engine/game/g_1848/step/operating.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative '../../../round/operating'
+
+module Engine
+  module Game
+    module G1848
+      module Round
+        class Operating < Engine::Round::Operating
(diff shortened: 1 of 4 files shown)


lkingsford:
Alternative (which I'd be ok with, if you wanted to reduce the blast radius) is just to do the skip fix in 1848. It'll still be there later if needed.

lib/engine/game/g_1848/step/operating.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require_relative '../../../round/operating'
+
+module Engine
+  module Game
+    module G1848
+      module Round
+        class Operating < Engine::Round::Operating
+          def skip_entity?(entity)
+            return super if entity.name != :COM
```
tobymao, line 11: should you use entity.sym or name?

Issue #6640, opened 2021-12-17
Clicking "invite" from within existing game url creates url in clipboard with redundant game/<id> appended to existing game url


ryanarn:
**How to Reproduce the Problem?**

Create a game, press "invite" button. Open new tab, paste url in clipboard, open site, e.g.,

`https://18xx.games/game/65327`

From there, repeat. Press "invite" button. Open new tab, paste url in clipboard, open site, e.g.,

`https://18xx.games/game/65327/game/65327`

This will go on indefinitely.

What steps cause the unexpected behavior?

Inviting from within a game URL gives a new url with game/<number> appended to the previous url.

Please include screenshots (you can drag and drop or upload them into this issue)

![1846](https://user-images.githubusercontent.com/1122885/146467371-c10e1059-2282-4630-befb-ce7744761001.png)

**Which Browser / Platform?**
**Chrome Version 96.0.4664.110 (Official Build) (64-bit) (cohort: 96_win_110)**

In game_card.rb the function:

```
    def render_invite_link
      msg = 'Copied invite link to clipboard; you can share this link with '\
            'other players to invite them to the game'

      invite_url = url(@gdata)
      flash = lambda do
        `navigator.clipboard.writeText((window.location + invite_url).replace('//game', '/game'))`
        store(:flash_opts, { message: msg, color: 'lightgreen' }, skip: false)
      end
      render_link(invite_url, flash, 'Invite')
    end
```
The js looks to be expecting to append the invite_url to the hostname but is appending to the full url (which in the case of the second invite button press is the full URL of the game). The fix is likely to use window.location.hostname:

 `navigator.clipboard.writeText((window.location.hostname+ invite_url).replace('//game', '/game'))`

I'm having problems setting up WSL, so I couldn't test it:

benjaminxscott:
closed by https://github.com/tobymao/18xx/pull/6820

PR #6820, opened 2022-01-19
invite link should use window location host

marndt93:
The invite link should use window.location.host so that the appended invite_url is not continuously added to the end of the url.
https://github.com/tobymao/18xx/issues/6640

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -188,7 +188,7 @@ module View
 
       invite_url = url(@gdata)
       flash = lambda do
-        `navigator.clipboard.writeText((window.location + invite_url).replace('//game', '/game'))`
+        `navigator.clipboard.writeText((window.location.host + invite_url).replace('//game', '/game'))`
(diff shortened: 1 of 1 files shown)


tobymao:
thanks!

assets/app/view/game_card.rb
```diff
@@ -188,7 +188,7 @@ def render_invite_link
 
       invite_url = url(@gdata)
       flash = lambda do
-        `navigator.clipboard.writeText((window.location + invite_url).replace('//game', '/game'))`
+        `navigator.clipboard.writeText((window.location.host + invite_url).replace('//game', '/game'))`
```
magnusrk-ch, line 191: This needs to be window.location.origin, not host. The host property doesn't include the scheme, so now the links are just "18xx.games/game/123" and don't linkify properly.

tobymao: can you please fix @magnusrk-ch ? thanks for identifying

Issue #3469, opened 2021-01-23
Add button to copy human readable transcript text


thepackrat:
It would be great to have a button to copy the human readable transcript text from the history box (including chat message) into the clipboard so that it could be pasted somewhere as a record. 

This would appear next to undo/redo or in the a corner of the transcript box and could be an icon rather than text to save space.

Maxwell-Frank:
Maybe "copy transcript" could be added on the Tools tab with the "copy game data" buttons.

PR #6823, opened 2022-01-20
add copy transcript button

marndt93:
resolves: #3469
Adds a copy transcript button which copies game log to clipboard.
image

image

### DIFF
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -240,7 +240,7 @@ module View
               sender = line.entity.name || line.user
               line = "#{sender}: #{line.message}"
             end
-            log_text.push(time ? "#{time} #{line}" : line)
+            log_text << (time ? "#{time} #{line}" : line)
           end
         end
         `navigator.clipboard.writeText(log_text.join('\n'))`
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -230,7 +230,7 @@ module View
 
       def copy_log_transcript
         actions = @game.actions.to_h { |a| [a.id, a] }
-        logText = String.new
+        log_text = String.new
         @game.log.map.group_by(&:action_id).flat_map do |action_id, entries|
           action = actions[action_id]
           entries.flat_map do |entry|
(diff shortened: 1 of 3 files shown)


tobymao:
how does this look on mobile?

assets/app/view/game/game_log.rb
```diff
@@ -226,8 +228,36 @@ def render_date_banner(time)
         h('div.chatline', { style: { textAlign: :center } }, date)
       end
 
+      def copy_log_transcript
+        actions = @game.actions.to_h { |a| [a.id, a] }
+        logText = ''
```
tobymao, line 233: snake_case

marndt93:
> 

what is the min screen size supported for mobile?

assets/app/view/game/game_log.rb
tobymao: change this to String.new and do string << x or an array and do [].join()

tobymao:
that's probably fine

assets/app/view/game/game_log.rb
```diff
@@ -226,8 +228,36 @@ def render_date_banner(time)
...
+        logText = String.new
+        @game.log.map.group_by(&:action_id).flat_map do |action_id, entries|
+          action = actions[action_id]
+          entries.flat_map do |entry|
+            time = action&.created_at ? "[#{Time.at(action.created_at || Time.now).strftime('%R')}]" : ""
+            line = entry.message
+            if line.is_a?(Engine::Action::Message)  
+              sender = line.entity.name || line.user
+              line = "#{sender}: #{line.message}"
+            end
```
tobymao, line 242: log_text << z

marndt93: I do not know this syntax/ruby.  Is this code valid?
image
image

tobymao: oh poop, ok just use an array then and then finish with a join

marndt93: also, using String.new fails the linter:
image

assets/app/view/game/game_log.rb
```diff
@@ -230,32 +230,32 @@ def render_date_banner(time)
...
+            time = action&.created_at ? "[#{Time.at(action.created_at || Time.now).strftime('%R')}]" : ''
             line = entry.message
-            if line.is_a?(Engine::Action::Message)  
+            if line.is_a?(Engine::Action::Message)
               sender = line.entity.name || line.user
               line = "#{sender}: #{line.message}"
             end
-            log_text += time ? "#{time} #{line}\n" : "#{line}\n"
-          end  
+            log_text.push(time ? "#{time} #{line}" : line)
```
tobymao, line 243: log_text << not push

PR #6830, opened 2022-01-20
auto-router use bitfields for conflict detection with pure js build-combo loop

JoeB989:
Two major points
1. I left auto_router.compute unmolested, but no longer called
2. I left both the ruby and javascript combo-builder functions in place, in case it makes sense to go back to the ruby one while still using bitfield compares


tobymao:
@JoeB989 thanks for the PR, do not leave any unused code, delete everything, git history can be used to bring things back

assets/app/view/game/route_selector.rb
```diff
@@ -251,14 +251,37 @@ def actions(render_halts)
         end
 
         auto = lambda do
+          # Two quick fallbacks if router.test_compute breaks
```
tobymao, line 254: delete all comments like this

assets/app/view/game/route_selector.rb
```diff
@@ -251,14 +251,37 @@ def actions(render_halts)
...
-          @routes = router.compute(
+          @routes = router.test_compute(
             @game.current_entity,
+            use_js_algorithm: true,
             routes: @routes.reject { |r| r.paths.empty? },
           )
           store(:routes, @routes)
         end
 
+        # test_auto = lambda do
```
tobymao, line 266: delete this

assets/app/view/game/route_selector.rb
```diff
@@ -303,6 +326,8 @@ def actions(render_halts)
         ]
         if @game_data.dig('settings', 'auto_routing') || @game_data['mode'] == :hotseat
           buttons << h('button.small', { on: { click: auto } }, 'Auto')
+          # buttons << h('button.small', { on: { click: test_auto } }, 'TEST Auto')
```
tobymao, line 329: delete

tobymao:
i will review once this pr is cleaned up, remove everything that is unnecessary, comments, dead code etc

JoeB989:
Do you want me to remove the old compute() as well?  Not leave it as a quick fallback if catastrophic problems?

tobymao:
> Do you want me to remove the old compute() as well? Not leave it as a quick fallback if catastrophic problems?

right. that’s what git is for, we can revert

JoeB989:
> right. that’s what git is for, we can revert
One last question: I got the impression you wanted the algorithm able to run in a ruby-only environment (point 2 in my first comment).  Do you still want that, or should I remove the ruby bitfield code?

tobymao:
> > right. that’s what git is for, we can revert
> 
> One last question: I got the impression you wanted the algorithm able to run in a ruby-only environment (point 2 in my first comment). Do you still want that, or should I remove the ruby bitfield code?

as much as i would like for it to run in both places, it's probably not useful and adds to the tech debt, let's only keep the bare necessities, so javascript versions

JoeB989:
@tobymao pinging in case you didn't see it's ready for re-review

tobymao:
i’ve been busy and this will take some time

JoeB989:
ok, no problem

lib/engine/auto_router.rb
```diff
@@ -16,9 +16,12 @@ def compute(corporation, **opts)
       route_limit = opts[:route_limit] || 1_000
 
       connections = {}
+      # Sort trains longest first, in case that improves combo calculations (some spotty evidence that it might)
+      # HACK: sort by train price, since some trains have multiple distance portions - or consider adding distance portions
+      trains = @game.route_trains(corporation).sort_by { |train| -train.price }
```
tobymao, line 21: if there's no real evidence of this, we should just delete it

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
+
       now = Time.now
 
       skip_paths = static.flat_map(&:paths).to_h { |path| [path, true] }
 
+      train_routes = Hash.new { |h, k| h[k] = [] }    # map of train to route list
+      path_abort = Hash.new { |h, k| h[k] = false }   # each train has opportunity to abort a branch of the path walk tree
+      route_counter = Hash.new { |h, k| h[k] = 0 }
+
+      trains.each do |train|
```
tobymao, line 46: there's no point to do this, you already created hashes on line 43 with default values...

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
+        path_abort[train] = false
+        route_counter[train] = 0
+      end
+
+      hexside_bits = Hash.new { |h, k| h[k] = 0 } # map of hexside_id to bit number
+
+      # table of bit 0-31 to bitmask
+      bit_to_bitmask = []
```
tobymao, line 54: bit_to_bitmask = (0..31).map { |i| 1 << i }

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
         else
-          puts "Path search: #{nodes.index(node)} / #{nodes.size}"
+          puts "Path search: #{nodes.index(node)} / #{nodes.size} - paths starting from #{node.hex.name}"
         end
 
+        node_now = Time.now
+
+        walk_counter = 0
+        counter = 0
+        route_counter.each { |train, _| route_counter[train] = 0 }
```
tobymao, line 73: pointless line?

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
+        node_abort = false
+
         node.walk(corporation: corporation, skip_paths: skip_paths) do |_, vp|
+          next if node_abort
```
tobymao, line 77: should this be next :abort if node_abort?

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
+
           paths = vp.keys
 
+          abort = nil
+          walk_counter += 1
+
+          if Time.now - node_now > node_timeout
+            # TODO: this is not a complete node.walk abort, find somthing more than :abort but less than break.
```
tobymao, line 85: what do you mean

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
           id = chains.flat_map { |c| c[:paths] }.sort!
           next if connections[id]
 
-          connections[id] = chains.map do |c|
```
tobymao, line 91: why was this line changed?

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
           puts 'Path timeout reached'
           break
         else
-          puts "Path search: #{nodes.index(node)} / #{nodes.size}"
+          puts "Path search: #{nodes.index(node)} / #{nodes.size} - paths starting from #{node.hex.name}"
         end
 
+        node_now = Time.now
+
+        walk_counter = 0
```
tobymao, line 71: delete thes counter variables, they look like debug code

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
-            @game,
-            @game.phase,
-            train,
-            connection_data: connection,
-          )
-          route.revenue
-          train_routes[train] << route
-        rescue GameError # rubocop:disable Lint/SuppressedException
+          # each train has opportunity to vote to abort a branch of this node's path-walk tree
+          path_abort.each { |train, _| path_abort[train] = false }
```
tobymao, line 144: looks like another unecessary line

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+                @game.phase,
+                train,
+                connection_data: connection,
+                bitfield: bitfield,
+              )
+
+              route.revenue # raises various errors if bad route
+              train_routes[train] << route
+              route_counter[train] += 1
+            end
```
tobymao, line 161: your indentation looks off here

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
-          )
-          route.revenue
-          train_routes[train] << route
-        rescue GameError # rubocop:disable Lint/SuppressedException
+          # each train has opportunity to vote to abort a branch of this node's path-walk tree
+          path_abort.each { |train, _| path_abort[train] = false }
+
+          # build a test route for each train, use route.revenue to check for errors, keep the good ones
+          trains.each do |train|
+            unless path_abort[train]
```
tobymao, line 148: just put the rescue inside of the each instead of keeping track of path_abort

lib/engine/auto_router.rb
```diff
@@ -28,21 +31,63 @@ def compute(corporation, **opts)
...
+      # which can cause all train routes to conflict with each other
+      node_timeout = path_timeout / nodes.size
+
       now = Time.now
 
       skip_paths = static.flat_map(&:paths).to_h { |path| [path, true] }
 
+      train_routes = Hash.new { |h, k| h[k] = [] }    # map of train to route list
+      path_abort = Hash.new { |h, k| h[k] = false }   # each train has opportunity to abort a branch of the path walk tree
+      route_counter = Hash.new { |h, k| h[k] = 0 }
```
tobymao, line 44: looks unused, delete

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+          rescue RouteTooLong
+            # ignore for this train, and abort walking this path if ignored for all trains
+            path_abort[train] = true
+            if path_abort.values.all?
+              abort = :abort # this path is dead for all trains, don't walk it further
+            end
+          rescue NoToken, RouteTooShort
+            # keep extending this connection set
+          rescue ReusesCity, RouteBlocked
+            abort = :abort # this path is dead, don't walk it further
```
tobymao, line 174: comments like these are not useful

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+              route_counter[train] += 1
+            end
+
+          # These all result in the route not being added to train_routes[train],
+          # but the nature of the error determines how to continue or terminate processing of the path walk
+          rescue RouteTooLong
+            # ignore for this train, and abort walking this path if ignored for all trains
+            path_abort[train] = true
+            if path_abort.values.all?
+              abort = :abort # this path is dead for all trains, don't walk it further
```
tobymao, line 169: same here

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+            end
+          rescue NoToken, RouteTooShort
+            # keep extending this connection set
+          rescue ReusesCity, RouteBlocked
+            abort = :abort # this path is dead, don't walk it further
+          rescue GameError => e
+            # an unhandled route error that probably should be handled
```
tobymao, line 176: unnecessary comemnts

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+            puts e
+          end
+          abort
         end
+        puts ' Node timeout reached' if node_abort
       end
-      puts "Pruned paths to #{train_routes.map { |k, v| k.name + ':' + v.size.to_s }.join(', ')} in: #{Time.now - now}"
+
+      # Check that there are no duplicate hexside bits (algorithm error)
+      mismatch = hexside_bits.length - hexside_bits.uniq.length
```
tobymao, line 185: don't use length, use size

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
-            combo += [route]
-            route.routes = combo
-            route.clear_cache!(only_routes: true)
-            counter += 1
-            if (counter % 1000).zero?
-              puts "#{counter} / #{limit}"
-              raise if Time.now - now > route_timeout
-            end
+      # final sanity check on best combos: recompute each route.revenue in case it needs to reject a combo
+      bad_route = nil
```
tobymao, line 206: this is not necessary

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+      end || []
+
+      max_routes.each { |route| route.routes = max_routes }
+    end
 
-            route.revenue
-            possibilities << combo
-            combo
-          rescue GameError # rubocop:disable Lint/SuppressedException
+    def print_bad_route(routes, bad_route)
```
tobymao, line 225: this is not necessary

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+    # returns:
+    #   the bitfield (array of ints) representing all hexsides in the connection path
+    # updates:
+    #   new hexsides are added to hexside_bits
+    def bitfield_from_connection(connection, hexside_bits, bit_to_bitmask)
+      bitfield = [0]
+      connection.each do |conn|
+        paths = conn[:chain][:paths]
+        index = 1
+        stop = paths.size
```
tobymao, line 246: size

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+        node2 = paths[0]
```
tobymao, line 247: why is this called node2 and not node1

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+        while index < stop
```
tobymao, line 248: why not just do

paths.size.times do |i|
... 

i doubt this hand rolled loop is much faster?

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
 
-        combos.compact!
-      rescue RuntimeError
-        puts 'Route timeout reach'
-        break
+    def check_edge_and_set(bitfield, hexside_edge, hexside_bits, bit_to_bitmask)
+      if hexside_bits.include?(hexside_edge)
+        set_bit(bitfield, hexside_bits[hexside_edge], bit_to_bitmask)
+      else
+        newbit = hexside_bits.length.positive? ? hexside_bits.map { |_hexside, bit| bit }.max + 1 : 0
```
tobymao, line 285: size

tobymao, line 285: map.max is probably inefficient

lib/engine/auto_router.rb
tobymao: why is this even necessary? why would the hexside bits not be unique? you made hexside bits yourself

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+        set_bit(bitfield, newbit, bit_to_bitmask)
       end
+    end
 
-      puts "Found #{possibilities.size} possible routes in: #{Time.now - now}"
+    # bitfield is an array of integers, can be expanded by this call if necessary
+    # bit is a bit number, 0 is lowest bit, 32 will jump to the next int in the array, and so on
+    def set_bit(bitfield, bit, bit_to_bitmask)
+      entry = (bit / 32).to_i
+      mask = bit_to_bitmask[bit & 31]
```
tobymao, line 295: do we even need bit_to_bitmask? can you just do 1 << bit

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+      add_count = entry + 1 - bitfield.size
```
tobymao, line 296: this code looks convoluted, are you trying to to do n.times.each { bitfield << 0 } 

or bitfield.concat([0] * n)

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+      end
+      bitfield[entry] |= mask
+    end
 
-      max_routes = possibilities.max_by do |routes|
-        routes.each { |route| route.routes = routes }
-        @game.routes_revenue(routes)
-      end || []
+    # does testroute conflict with any route in the combo?
+    def route_bitfield_conflicts(combo, testroute)
```
tobymao, line 305: this looks unused

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
 
-      max_routes.each { |route| route.routes = max_routes }
+          index -= 1
+        end
+      end
+      false
     end
+
+    # The js-in-Opal algorithm
+    def js_evaluate_combos(rb_sorted_routes, _route_timeout)
```
tobymao, line 319: is route_timeout unused? why is it prefixed

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+      rb_possibilities = []
+      possibilities_count = 0
+      conflicts = 0
+      limit = rb_sorted_routes.map(&:size).reduce(&:*) # rubocop:disable Lint/UselessAssignment
```
tobymao, line 323: why are you getting a lint error here? if it's useless, delete it

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+      now = Time.now
+
+      if RUBY_ENGINE == 'opal'
```
tobymao, line 326: delete this if statement

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+        %x(
+        let possibilities = []
+        let combos = [];
+        let counter = 0;
+        let max_revenue = 0;
+
+        // marshal Opal objects to js for faster/easier access
+        js_sorted_routes = [];
```
tobymao, line 334: you're minnig a definition here, you can make it const as well if it's not changing

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+        Opal.send(rb_sorted_routes, 'each', [], function(rb_routes)
```
tobymao, line 335: why do you have rb_sorted_routes and js_sorted_routes? you should probably only have one, please clean this up

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+              js_routes.push( { route: rb_route, bitfield: rb_route.bitfield, revenue: rb_route.revenue } );
+            });
+            js_sorted_routes.push(js_routes);
+          });
+        let js_limit = limit
+
+        // init combos with first train's routes
+        for (r=0; r < js_sorted_routes[0].length; r++)
+        {
+          let route = js_sorted_routes[0][r];
```
tobymao, line 349: i think this can be a const, look over all the code and fix any lets to const

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+          combos.push(combo);
+          possibilities_count += 1;
+
+          // accumulate best-value combos, or start over if found a bigger best
+          if (combo.revenue >= max_revenue)
+          {
+            if (combo.revenue > max_revenue)
+            {
+              possibilities = []
+              max_revenue = combo.revenue;
```
tobymao, line 361: be consistent, don't use semi colons, remove all of them

lib/engine/auto_router.rb
```diff
@@ -88,81 +133,328 @@ def compute(corporation, **opts)
...
+            Opal.send(rb_routes, 'each', [], function(rb_route)
+            {
+              js_routes.push( { route: rb_route, bitfield: rb_route.bitfield, revenue: rb_route.revenue } );
+            });
+            js_sorted_routes.push(js_routes);
+          });
+        let js_limit = limit
+
+        // init combos with first train's routes
+        for (r=0; r < js_sorted_routes[0].length; r++)
```
tobymao, line 347: javascript style sohuld be

for (...) {
} 

change all instances

lib/engine/route.rb
```diff
@@ -27,6 +27,8 @@ def initialize(game, phase, train, **opts)
       @last_node = nil
       @last_offboard = []
       @stops = nil
+
+      @bitfield = opts[:bitfield] # array of ints used only by auto-routing algorithm
```
tobymao, line 31: do you even use bitfield in ruby? if not delete all related code

lib/engine/game_error.rb
```diff
@@ -3,4 +3,21 @@
...
+  class NoToken < GameError
+  end
+
+  class RouteTooShort < GameError
+  end
+
+  class RouteTooLong < GameError
+  end
+
+  class RouteBlocked < GameError
```
tobymao, line 18: i don't see these error classes ever used?

lib/engine/auto_router.rb
```diff
@@ -7,6 +7,7 @@ module Engine
   class AutoRouter
     def initialize(game)
       @game = game
+      @nextbit = 0
```
tobymao, line 10: can you explain nextbit?

lib/engine/auto_router.rb
```diff
@@ -91,78 +116,281 @@ def compute(corporation, **opts)
           connections[id] = chains.map do |c|
             { left: c[:nodes][0], right: c[:nodes][1], chain: c }
           end
-        end
-      end
 
-      puts "Found #{connections.size} paths in: #{Time.now - now}"
-      puts 'Pruning paths to legal routes'
+          counter += 1
```
tobymao, line 120: counter seems unused

lib/engine/auto_router.rb
```diff
@@ -28,21 +30,44 @@ def compute(corporation, **opts)
...
           puts 'Path timeout reached'
           break
         else
-          puts "Path search: #{nodes.index(node)} / #{nodes.size}"
+          puts "Path search: #{nodes.index(node)} / #{nodes.size} - paths starting from #{node.hex.name}"
         end
 
+        node_now = Time.now
+
+        counter = 0
```
tobymao, line 56: unused var?

lib/engine/auto_router.rb
```diff
@@ -28,21 +30,44 @@ def compute(corporation, **opts)
...
+        node_abort = false
+
         node.walk(corporation: corporation, skip_paths: skip_paths) do |_, vp|
+          next :abort if node_abort
+
           paths = vp.keys
 
+          if Time.now - node_now > node_timeout
```
tobymao, line 64: would this be simplified as node_abort = Time.now - node_now > node_timeout
no need to do next :abort since it's already done in the first part of the loop

lib/engine/auto_router.rb
tobymao: this is unaddressed

tobymao: but up ahead you do  path_abort = Hash.new { |h, k| h[k] = false }   which makes each key default false

tobymao:
@JoeB989 the last thing to fix is adding

        @flash&.call('Auto route selection failed to complete (timeout)')

there was a previous PR that added the flash call, can you add this when the javascript times out?

tobymao:
i fixed a conflict so you should pull

lib/engine/auto_router.rb
JoeB989: Yeah the code was sloppy.  I've tried to clarify it.  Up top, declare and populate a hash of train to <something>, only setting keys (but having to provide initial values).  In this loop here, set all trains' path_abort to false, so only if they all raise RouteTooLong on the connection will it abort the path.  Path_aborts are cleared for every connection.  And remove the unless path_abort[train], since that was dead code (all flags are falsed right before the loop)

Previously, the code only declared the shape of the hash up top, and here on the first connection loop the path_abort had no entires, but each "unless path_abort[train]" caused the creation of the train,false kvp.  Subsequent connection loops would proceed with all kvp's in place.  That's why I call that sloppy.

Note: I had tried to make path_abort longer-lived as the connections build, but had some logic problem that I never figured out.  The unless is actually unneeded for now, given that they're all falsed before the trains.each,; I had been keeping it here in case I get back to path_abort surviving path.walk yields.  (E.g a 5 train will abort on a 6 route, but a 2nd 10 train will keep accumulating 7,8,9,10,11 routes while the path_abort[5] would prevent 5-route.revenue on 7,8,9,10,11 routes)

JoeB989:
I rebased and added flash.call.

tobymao (approved):

PR #6836, opened 2022-01-22
[1866] Loans

magnusfalt:
- Fixed C and B tiles
- Loans, both personal and corporation
- If a company holds any loans it cannot sell trains for less than face value and cannot buy trains for more than face value.
- Some more code cleanup
- Some rules changes

### DIFF
--- a/lib/engine/game/g_1866/stock_market.rb
+++ b/lib/engine/game/g_1866/stock_market.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative '../../stock_market'
+
+module Engine
+  module Game
+    module G1866
+      class StockMarket < Engine::StockMarket
+        def move_down(corporation)
+          r, c = corporation.share_price.coordinates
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_1866/stock_market.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative '../../stock_market'
```
tobymao, line 3: is this class necessary? i try to avoid having too many of these base classes, but if it's the best way we can keep it

magnusfalt: I think it is necessary, i override the move_down and move_up methods. The stock market in 1866 is very special, i dont see any other game have any usefulness for that code. Atleast the move down code, its very specific. We can solve it with a bunch of constants or base-methods calls, but i think that would clutter down the base code more then just to keep the overridden class.

Issue #6684, opened 2021-12-29
Auto-complete timeout should give a message to the UI


slapphappyjazz:
https://18xx.games/game/66498?action=478

At this point in the game, the auto-routing algorithm only gaves routes for 2 of the 3 trains that the company owned.  The third train was left blank.  This seemed weird and it wasn't obvious what was going on.  A similar thing happened a few times later in the game as well, when the algorithm only reported a route for one train out of two that the company owned.

On investigating this later, I found a message in the browser console saying that the auto-routing algorithm had timed out.

I think the UI should give an indication when the auto-routing algorithm times out.  an error message at the top of the screen would probably work, or an entry in the game log - whichever you think best.  There should be some indication to the user as to what has happened.

(We played badly and there weren't enough companies opened or tokens on the board, which probably gave the algorithm too large a network to analyse, but the point I'm making is about the feedback to the user.  I've no problem with the algorithm timing out as such).

PR #6837, opened 2022-01-22
display message if auto-router times out

marndt93:
resolves: #6684

### DIFF
--- a/lib/engine/auto_router.rb
+++ b/lib/engine/auto_router.rb
@@ -153,9 +153,7 @@ module Engine
         combos.compact!
       rescue RuntimeError
         puts 'Route timeout reach'
-        if @flash
-          @flash.call('Auto route selection failed to complete (timeout)')
-        end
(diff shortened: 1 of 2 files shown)


lib/engine/auto_router.rb
```diff
@@ -5,8 +5,9 @@
 
 module Engine
   class AutoRouter
-    def initialize(game)
+    def initialize(game, flash)
```
tobymao, line 8: let this be optional

tobymao: flash = none

flash&.call

lib/engine/auto_router.rb
```diff
@@ -152,6 +153,9 @@ def compute(corporation, **opts)
         combos.compact!
       rescue RuntimeError
         puts 'Route timeout reach'
+        if @flash
```
tobymao, line 156: you can just do @flash&.call right

PR #6838, opened 2022-01-22
21Moon: SRs and CRs

roseundy:
Common code changes:

* UI::BuySellShares - allow custom button names when selling corporate shares
* UI::Corporation - handle treasury_as_holding
* Engine::Corporation - add treasury_as_holding option to treat all corporate-held shares the same
* Engine::Step::BuySellParShares - allow overriding of can_dump? method

### DIFF
--- a/lib/engine/corporation.rb
+++ b/lib/engine/corporation.rb
@@ -26,7 +26,7 @@ module Engine
                   :type, :floatable, :original_par_price, :reservation_color, :min_price, :ipo_owner,
                   :always_market_price
     attr_reader :companies, :name, :full_name, :fraction_shares, :id, :needs_token_to_par,
-                :presidents_share, :price_multiplier
+                :presidents_share, :price_multiplier, :treasury_as_holding
     attr_writer :par_price, :share_price
 
     SHARES = ([20] + Array.new(8, 10)).freeze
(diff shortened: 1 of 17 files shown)


lib/engine/corporation.rb
```diff
@@ -26,7 +26,7 @@ class Corporation
                   :type, :floatable, :original_par_price, :reservation_color, :min_price, :ipo_owner,
                   :always_market_price
     attr_reader :companies, :name, :full_name, :fraction_shares, :id, :needs_token_to_par,
-                :presidents_share, :price_multiplier
+                :presidents_share, :price_multiplier, :treasury_as_holding
```
tobymao, line 29: what does treasure_as_holding mean

tobymao (approved):

lib/engine/corporation.rb
roseundy: Normally, for stock owned by the corporation, corporations distinguish between shares of their own corporation ("treasury" shares) and shares of other corporations ("holdings"). The "treasury_as_holding" is a flag that indicates that all shares are treated the same.

roseundy: Like you said, good naming is hard.

Issue #6740, opened 2022-01-09
UI - show both number of shares and percentage of ownership


slapphappyjazz:
In playing 18CZ today, we noticed that it isn't easy to see both the number of shares you own in a company and the percentage of the company owned.  This is particularly relevant to 18CZ because it has three different sizes of company.  In the physical game, the share certificates show the percentage ownership, so you can see both pieces of information in the same place.
In 18xx.games, the player cards show you the percentage owned in each company, but not the number of shares.  The company cards show the number of shares owned but not the percentage.   In games like 18CZ (and perhaps 1817?) it would be helpful to display both.
For 1830-style games, this is much less relevant.  Possibly this could be a display configuration to be chosen by the game implementor.

PR #6839, opened 2022-01-24
show share ownership percentage

marndt93:
resolves: #6740
image

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -338,7 +338,7 @@ module View
           type = entity.player? ? 'tr.player' : 'tr.corp'
           type += '.bold' if last_acted_upon
           name = entity.player? ? entity.name : "© #{entity.name}"
-          show_percent = @game.instance_variable_get(:@show_share_percent_ownership)
+          show_percent = @game.class::SHOW_SHARE_PERCENT_OWNERSHIP
           percent_shares = num_shares * @corporation.share_percent
           percent_shares_str = percent_shares.positive? && show_percent ? " (#{percent_shares}%)" : ''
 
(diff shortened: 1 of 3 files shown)


marndt93:
Couple questions regarding this PR:
1. should we show percentage for other rows?
2. should this feature only be available for specific games (issue was opened specifically for 18cz which has atypical shares)?

roseundy:
I don't think this should be the default. It needs to be configurable by game.

assets/app/view/game/corporation.rb
```diff
@@ -338,10 +338,13 @@ def entities_rows(entities)
           type = entity.player? ? 'tr.player' : 'tr.corp'
           type += '.bold' if last_acted_upon
           name = entity.player? ? entity.name : "© #{entity.name}"
+          show_percent = @game.instance_variable_get(:@show_share_percent_ownership)
```
roseundy, line 341: @game.class::SHOW_SHARE_PERCENT_OWNERSHIP

lib/engine/game/base.rb
```diff
@@ -453,6 +456,7 @@ def initialize(names, id: 0, actions: [], at_action: nil, pin: nil, strict: fals
         @round_counter = 0
 
         @optional_rules = init_optional_rules(optional_rules)
+        @show_share_percent_ownership = self.class::SHOW_SHARE_PERCENT_OWNERSHIP
```
roseundy, line 459: Not needed

assets/app/view/game/corporation.rb
marndt93: Thanks, couldn't figure out how to access the class property.  kept getting undefined method error...

tobymao (approved):

PR #6848, opened 2022-01-26
Issue#6650 show upcoming trains

### LINKED ISSUE #6650: [ALL] Suggestion: Display Upcoming Trains in Game tab during stock rounds
Place the train information under player info next to The Bank information. Sometimes knowing how many trains of a particular type are available can influence whether to start a new company or not and at what price to par, and clicking between tabs is sometimes too cumbersome a task to bear.

Example of knowing available trains to influence: in 1856, the number of shares required to float the company changes with the available train *at the time that the company would operate*. So if during the Stock Round the available train is a 3T, only three shares will float the company, but when it's the company's turn in the OR and the 4T is available because all the 3Ts were bought, then the company does not, in fact, float with only three shares. So if one can calculate the likelihood of this happening at a glance, rather than clicking between tabs, one could either opt to not start the company, or par at a lower price to afford an extra share.

marndt93:
resolves #6650 
image

### DIFF
--- a/assets/app/view/game/train_schedule.rb
+++ b/assets/app/view/game/train_schedule.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 require 'lib/settings'
 
 module View
--- a/assets/app/view/game/train_schedule.rb
+++ b/assets/app/view/game/train_schedule.rb
@@ -60,11 +60,10 @@ module View
           tds = [h(:td, @game.info_train_name(trains.first)),
                  h("td#{price_str_class}", @game.info_train_price(trains.first)),
                  h('td.right', "×#{trains.size}")]
-          tds << h('td.right', events) if events.size.positive?
+          tds << h('td.right', events) unless events.empty?
 
           h(:tr, tds)
         end
-        trs ||= 'None'
 
         h('div#upcoming_trains.card', [
           h('div.title', title_props, 'Upcoming Trains'),
--- a/assets/app/view/game/train_schedule.rb
+++ b/assets/app/view/game/train_schedule.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+require 'lib/settings'
+
+module View
+  module Game
+    class TrainSchedule < Snabberb::Component
+      include Lib::Settings
+
+      needs :game
+
+      def price_str_class
+        max_size = @game.depot.upcoming.group_by(&:name).map do |_name, trains|
+          trains.first.names_to_prices.keys.size
+        end.max
+        max_size == 1 ? '.right' : ''
+      end
+
+      def rust_obsolete_schedule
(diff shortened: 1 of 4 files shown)


roseundy:
Why is this being attached to the stock market?

scottredracecar:
Bank being attached to the stock market makes more sense in some games,
such as 1817 where it also contains the loan information that is key for
market decisions.
https://18xx.games/game/70158#market

A typical bank (with no extra information) isn't intuitive to me that it
shows up in the market tab and not the info tab.

This is all pretty subjective.

assets/app/view/game/train_schedule.rb
```diff
@@ -0,0 +1,78 @@
...
+          events << h('div.left', "rusts #{rust_schedule[name].join(', ')}") if rust_schedule[name]
+          events << h('div.left', "obsoletes #{obsolete_schedule[name].join(', ')}") if obsolete_schedule[name]
+          tds = [h(:td, @game.info_train_name(trains.first)),
+                 h("td#{price_str_class}", @game.info_train_price(trains.first)),
+                 h('td.right', "×#{trains.size}")]
+          tds << h('td.right', events) if events.size.positive?
+
+          h(:tr, tds)
+        end
+        trs ||= 'None'
```
tobymao, line 67: trs will never be nil, what was the intent of ||= here?

assets/app/view/game/train_schedule.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        rust_schedule, obsolete_schedule = rust_obsolete_schedule
+        trs = @game.depot.upcoming.group_by(&:name).map do |name, trains|
+          events = []
+          events << h('div.left', "rusts #{rust_schedule[name].join(', ')}") if rust_schedule[name]
+          events << h('div.left', "obsoletes #{obsolete_schedule[name].join(', ')}") if obsolete_schedule[name]
+          tds = [h(:td, @game.info_train_name(trains.first)),
+                 h("td#{price_str_class}", @game.info_train_price(trains.first)),
+                 h('td.right', "×#{trains.size}")]
+          tds << h('td.right', events) if events.size.positive?
```
tobymao, line 63: unless events.empty?

assets/app/view/game/train_schedule.rb
tobymao: i see you didn't write the code, but please change it as it looks off

marndt93: I think the intent here is to make sure an empty card is not displayed for the upcoming trains.  If you are sure that this is not possible, then it can probably be removed.

tobymao: i think it's a bug because in ruby [] is not falsey, and trs will never bi nil

PR #6867, opened 2022-01-29
Fix auto-buy allowing to buy receivership shares without the buy option.

### LINKED ISSUE #6861: [Harzbahn 1873] Owner of U set to nil due to autopurchase issue
**What's your Game ID?**

https://18xx.games/game/68415

**How to Reproduce the Bug?**

Go to the link. The game is stuck at "Loading game..."

This game worked just fine until right now, start of the last SR.

Javascript console says:
```
Object { name: "player?", message: "undefined method `player?' for nil", cause: {…}, args: [], backtrace: (13) […], stack: "  from https://18xx.games/assets/deps.js:1:14303:in `p'\n  from https://18xx.games/assets/main.js:2:27960:in `Opal.modules[\"engine/game/base\"]/<'\n  from https://18xx.games/assets/main.js:62:103394:in `Opal.modules[\"view/game/company\"]/<'\n  from https://18xx.games/assets/main.js:62:99091:in `Opal.modules[\"view/game/company\"]/<'\n  from https://18xx.games/assets/deps.js:25:61655:in `Opal.modules[\"snabberb/component\"]/<'\n  from https://18xx.games/assets/main.js:62:95323:in `l'\n  from https://18xx.games/assets/deps.js:1:16009:in `d.yield1'\n  from https://18xx.games/assets/deps.js:15:20262:in `Opal.modules[\"corelib/enumerable\"]/g/<'\n  from https://18xx.games/assets/deps.js:1:16009:in `d.yield1'\n  from https://18xx.games/assets/deps.js:15:67506:in `V'\n  from https://18xx.games/assets/deps.js:1:18347:in `d.send2'\n  from https://18xx.games/assets/deps.js:1:18173:in `d.send'\n  from https://18xx.games/assets/deps.js:1:21082:in `ne'" }
```

**Which Browser / Platform?**

Happens on Chrome and Firefox and both on desktop and mobile. So I guess it's not platform dependent. Can provide more details if there is new evidence that it is.

magnusrk-ch:
Works by moving the restart logic from corporation_available? to can_buy?.

1873 was the only company that had corporation_available? return false after it had previously returned true, which seems counter to the general purpose of that method, and also doesn't get applied in many parts of the game since corporation_available? is assumed to be a UI convenience.

By putting the logic in can_buy? instead it gets applied in the same place as corporation_available? (for showing buy buttons), but also applies to auto-buy and other parts of share buying.

New result on receivership auto-buy: 'Disabled programmed action 'Buy U from market until 2 shares' due to 'Cannot buy U from market''

Fixes #6861

### DIFF
--- a/lib/engine/game/g_1873/game.rb
+++ b/lib/engine/game/g_1873/game.rb
@@ -896,8 +896,6 @@ module Engine
 
         def corporation_available?(entity)
           return false unless entity.corporation?
-          return true if entity == @mhe
-          return can_restart?(entity, @round.active_step.current_entity) if entity.receivership?
 
           entity.ipoed || can_par?(entity, @round.active_step.current_entity)
         end
--- a/lib/engine/game/g_1873/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1873/step/buy_sell_par_shares.rb
@@ -29,6 +29,13 @@ module Engine
             end
           end
 
+          def can_buy?(entity, bundle)
+            corporation = bundle.corporation
+            return @game.can_restart?(corporation, entity) if corporation.receivership? && corporation != @game.mhe
+
+            super
+          end
+
           def can_buy_multiple?(entity, corporation, _owner)
             return unless corporation.corporation?
 


magnusrk-ch:
This will not by itself fix the broken game from #6861. The last action in the game (1344) already has invalid auto actions attached. This action would have to be deleted or somehow manually undone.

roseundy:
I have the correct fix. The fix will _not_ break the game.

PR #6873, opened 2022-01-30
18MO - Initial commit

scottredracecar:


lib/engine/game/g_18_mo/game.rb
```diff
@@ -0,0 +1,247 @@
...
+          route.visited_stops.each do |stop|
+            hex = stop.hex
+            raise GameError, 'Route cannot run to multiple cities in a hex' if visited_hexes[hex]
+
+            visited_hexes[hex] = true
+          end
+        end
+
+        # Two lays with one being an upgrade, second tile costs 20
+        TILE_LAYS = [
```
tobymao, line 203: this shouldn't be here, it should be above

scottredracecar: ok

PR #6876, opened 2022-01-30
18EU Red to Red Bonus

ryandriskel:

A train beginning and ending its route in different red hexes gains a bonus based on how many of the corporation's tokens it has run through, limited to a maximum by the phase.

Screen Shot 2022-01-30 at 9 58 24 AM

On route selector:

Screen Shot 2022-01-30 at 9 59 23 AM

Resulting log:

Screen Shot 2022-01-30 at 9 56 37 AM

### DIFF
--- a/lib/engine/game/g_18_eu/game.rb
+++ b/lib/engine/game/g_18_eu/game.rb
@@ -49,6 +49,15 @@ module Engine
           ]
         ).freeze
 
+        RED_TO_RED_BONUS = {
+          '2' => [0, 0],
+          '3' => [10, 10],
+          '4' => [10, 10],
+          '5' => [20, 80],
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_eu/game.rb
```diff
@@ -182,23 +191,35 @@ def player_sort(entities)
           (minors.sort_by { |m| m.name.to_i } + majors.sort_by(&:name)).group_by(&:owner)
         end
 
-        # def revenue_for(route, stops)
-        # revenue = super
+        def route_ends_red?(stops)
+          return false unless stops.length > 1
```
tobymao, line 195: size

PR #6878, opened 2022-01-30
Some fixes for 18Rhl

perwestling:
1. RhE corrections
 * When setting par after buying private 6, the par can only be
   70, 75 or 80.
 * Money for presidency will be the winning bid of private 6.
 * Private 6 text updated.
2. Remove (incorrectly located) fixture from pre-alpha testing
 of 18Rhl as changes breaks this fixture. As it is already
 paring of RhE after private auction it is not possible to fix.
 A new fixture will be added (but maybe not until after alpha
 started?).
3. Split out entities (and some related methods) to an entities file.
4. Split out map and tiles (and some related methods) to a map file.
5. Correct special track for privates 2-4 (and 1 if using variant)
 so that these can be use to lay a track between regular track build
 and train buy.


lib/engine/game/g_18_ms/game.rb
```diff
@@ -175,14 +175,6 @@ class Game < Game::Base
         HEXES_FOR_GRAY_TILE = %w[C9 E11].freeze
         COMPANY_1_AND_2 = %w[AGS BS].freeze
```
tobymao, line 177: why did this change

perwestling: I removed line 178-184 as I noticed that these were not used anywhere. 
I can remove that change from the commit if you wish. Should I do that?

perwestling: @tobymao I have now removed the changed 18MS file - I can do that in a separate commit if I think it is good. Or are there some tool one can use to discover unused code?

perwestling:
Let me know if I should make this into two commits - one with restructuring of maps/entities, and one with the missing features?

lib/engine/game/g_18_ms/game.rb
tobymao: yo udidn't need to remove it, i was just asking

tobymao (approved):

tobymao:
add it back and merge when ready

perwestling:
I think it is fine to leave that change out as it has nothing to do with 18Rhl. I can do a separate commit for that.
So OK to merge then?

Issue #6538, opened 2021-11-25
Set minimum players when creating game


ChrisTheCat:
When creating a game, you can set the maximum number of players but not the minimum. Thus, if you only want to play a five player game, it will be advertised as 3-5 players.

And a player who wants to join a game with exactly five players might end up in three player game.

It would be nice if players could set both the minimum and maximum number of players when creating games.

Zwergenpunk:
maybe it would even be better to just say a fixed number of players. 
While game creation a check box could be shown which says something like "exactly  player count"

PR #6880, opened 2022-01-30
choose min players in game setup

marndt93:
resolves: #6538


migrate/0001_initial.rb
```diff
@@ -33,6 +33,7 @@
       String :description, null: false
 
       String :title, null: false
+      Integer :min_players, null: false, default: 3
```
tobymao, line 36: you can't mutate a migration like this, you have to add a brand new one

models/game.rb
```diff
@@ -127,6 +127,7 @@ def to_h(include_actions: false, player: nil)
       description: description,
       user: user.to_h,
       players: ordered_players.map(&:to_h),
+      min_players: defined?(min_players) ? min_players : nil,
```
tobymao, line 130: what's going on here this looks bad

models/game.rb
```diff
@@ -127,6 +127,7 @@ def to_h(include_actions: false, player: nil)
       description: description,
       user: user.to_h,
       players: ordered_players.map(&:to_h),
+      min_players: min_players || nil,
```
tobymao, line 130: why do you need an || nil

migrate/0006_add_min_players_game_setting.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    add_column :games, :min_players, Integer, null: false, default: 3
```
tobymao, line 5: would it make more sense to move max_players and min_players into settings?

models/game.rb
marndt93: I am not sure if it is necessary, but I was worried about games created before this property was added.

tobymao: no it's not necessary

migrate/0006_add_min_players_game_setting.rb
marndt93: possibly.  Would you like this change made?

You are talking about the settings json obj that is already in the games db table?

tobymao: right, in the settings, the migration will be a bit more involved

tobymao:
i'm expecting a migration of max_players

marndt93:
> i'm expecting a migration of max_players

max_players column should be removed, right?

tobymao:
yes

tobymao:
you need to iterate every game and move the deleted settings into the json, please test this out on games in your db, if it’s too involved, i’d rather just close this pr

marndt93:
> you need to iterate every game and move the deleted settings into the json, please test this out on games in your db, if it’s too involved, i’d rather just close this pr

hmm okay.  I can do this, but can you tell me where the local db is, so I can view the contents to verify?

If you want, we can close this PR for now and I can re-open it when I am finished with the migration

tobymao:
just make a couple of games on your local server

tobymao:
what do you mean view local?

marndt93:
> what do you mean view local?

I need to make some games with the old db configuration, then write and run a migration that moves the max_players to the settings and then drops the max_players column.

Wouldn't it be ideal to open the database and view the tables to verify that the data was moved successfully?  The docker script creates a postgres db locally, doesn't it?

tobymao:
look at the docker compose for the port, you can use any postgres client to look at the local db

vandamm:
@marndt93 something like this should be sufficient :
```ruby
Sequel.migration do
  transaction
  up do
    run 'update games set settings = jsonb_set(settings, '{max_players}', to_jsonb(max_players))'

    alter_table :games do
      drop_column :max_players
    end
  end

  down do
    alter_table :games do
      add_column :max_players, Integer, null: false, default: 6
    end  

    run 'update games set max_players = (settings->>'max_players' || 6)::integer;'
  end
end
```

marndt93:
@vandamm does this drop the max_players column and then add it back again?

vandamm:
@marndt93 it drops in direct migration, but adds back if you want to rollback

marndt93:
@vandamm I was just testing this today, but I am getting some errors.  I am not too experienced with migrations, so not too sure what the issue is.  Should we be using jsonb_insert instead of jsonb_set?

This is the error:
`rack_1   | E, [2022-03-13T18:33:29.918603 #8] ERROR -- : PG::UndefinedColumn: ERROR:  column "{max_players}" does not exist
rack_1   | LINE 1: ...pdate games set settings = jsonb_insert(settings, "{max_play...
rack_1   |                                                              ^
rack_1   | HINT:  Perhaps you meant to reference the column "games.max_players".: update games set settings = jsonb_insert(settings, "{max_players}", to_jsonb(max_players))`

PR #6886, opened 2022-01-31
1825/18EU Allow for a train that does not require a token.

ryandriskel:
This need is shared between 1825 and 18EU, so I wanted to make the implementation more generic. The original implementation was tied to the train being an MN type, which did not apply to 18EU.

1825 Tested:

Screen Shot 2022-01-31 at 10 43 32 AM

### DIFF
--- a/lib/engine/train.rb
+++ b/lib/engine/train.rb
@@ -9,7 +9,7 @@ module Engine
 
     attr_accessor :obsolete, :operated, :events, :variants, :obsolete_on, :rusted, :rusts_on, :index, :name,
                   :distance, :reserved
-    attr_reader :available_on, :discount, :multiplier, :sym, :variant
+    attr_reader :available_on, :discount, :multiplier, :sym, :variant, :requires_token
     attr_writer :buyable
 
     def initialize(name:, distance:, price:, index: 0, **opts)
@@ -30,6 +30,7 @@ module Engine
       @operated = false
       @events = (opts[:events] || []).select { |e| @index == (e['when'] || 1) - 1 }
       @reserved = opts[:reserved] || false
+      @requires_token = opts[:requires_token].nil? ? true : opts[:requires_token]
       init_variants(opts[:variants])
     end
 
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -231,8 +231,6 @@ module Engine
 
       ALLOW_TRAIN_BUY_FROM_OTHERS = true # Allows train buy from other corporations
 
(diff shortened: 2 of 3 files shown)


lib/engine/train.rb
```diff
@@ -30,6 +30,7 @@ def initialize(name:, distance:, price:, index: 0, **opts)
       @operated = false
       @events = (opts[:events] || []).select { |e| @index == (e['when'] || 1) - 1 }
       @reserved = opts[:reserved] || false
+      @requires_token = opts[:requires_token].nil? ? true : opts[:requires_token]
```
roseundy, line 33: I'm not a Ruby expert, but couldn't you just add "requires_token: true" to the def initialize for the train class?

roseundy: Otherwise, looks OK to me.

ryandriskel: > requires_token: true

Sure can. I was following the pattern of the reserved variable, but that is cleaner.

ryandriskel:
Going to hold this PR while we look for a better test case.

roseundy (approved):
LGTM

lib/engine/train.rb
```diff
@@ -9,10 +9,10 @@ class Train
 
     attr_accessor :obsolete, :operated, :events, :variants, :obsolete_on, :rusted, :rusts_on, :index, :name,
                   :distance, :reserved
-    attr_reader :available_on, :discount, :multiplier, :sym, :variant
+    attr_reader :available_on, :discount, :multiplier, :sym, :variant, :requires_token
     attr_writer :buyable
 
-    def initialize(name:, distance:, price:, index: 0, **opts)
+    def initialize(name:, distance:, price:, index: 0, requires_token: true, **opts)
```
tobymao, line 15: just put it in opts

ryandriskel: I originally had it there, but Steve advised this made setting the default true much cleaner. I can move it back, though.

lib/engine/train.rb
ryandriskel: @tobymao Do you have a cleaner way to write this line that I removed?

lib/engine/train.rb
tobymao: yea let's move it back

lib/engine/train.rb
tobymao: !!opts[:requireds_token]

ryandriskel: Wouldn't that result in a default value of false?

tobymao: ok then how you have it isfine

tobymao (approved):

Issue #6882, opened 2022-01-31
Auto-routing gives invalid results


kepakiano:
Given this game and the route for one 4T and the 3T already chosen:
![image](https://user-images.githubusercontent.com/1355029/151776192-261d23eb-a643-44b7-a1ba-ff9cadab52c5.png)
auto-routing yields a rather funny result:
![image](https://user-images.githubusercontent.com/1355029/151776389-c1549a7e-9af2-4784-a803-a3d53fefe01c.png)

Here is the link: https://18xx.games/game/68601?action=730

Without pre-choosing routes it now super quickly finds 10 more than I could find by hand, so that's awesome :D

scottredracecar:
I tried again on a different machine (first was my phone) and got the same result as the phone. I suggest this be closed as "no bug" unless others can reproduce. If this is a problem, others should be along to report their findings shortly via other tickets or this one.

![image](https://user-images.githubusercontent.com/29106780/151836209-88ecb9ef-8aba-4f07-abf2-79e0ce255eae.png)

kepakiano:
I'd appreciate if you didn't do that, as I can reproduce this behavior on two computers (Firefox + Chromium) and my phone.

scottredracecar:
Some other discussion here:
https://github.com/tobymao/18xx/issues/6887#issuecomment-1026108613

tysen:
I just tried locking in two of the trains as originally described and the auto button produced this result:

<img width="1072" alt="image" src="https://user-images.githubusercontent.com/1501322/151860168-6a5bae2f-1928-4a80-a0d6-17d296f9a5a0.png">

Issue #6890, opened 2022-01-31
Autoroute chooses invalid route for two trains on local dit


benjaminxscott:
When autorouting two 2 trains for a single city which has a dit on the same hex, it attempts to run both trains to the same dit, which is invalid

https://18xx.games/game/69695?action=135

https://gist.github.com/benjaminxscott/486b6bc834f215be477dbdb68477ca32

cc @JoeB989 
![IMG_5089](https://user-images.githubusercontent.com/1711810/151878168-dcd4da85-7585-489f-9082-c0b01ef1fd24.jpg)
![IMG_5090](https://user-images.githubusercontent.com/1711810/151878177-28cfc8b0-edf9-461e-93a2-7eda889c5c89.jpg)

JoeB989:
Looks like an auto-router bug because the routes don't cross any hexsides.  I'll look into

JoeB989:
Ok I can fix this.  But can you also send me a more developed map where there are longer routes through that tile?  I'm probably not handling city + dit inside a longer route

benjaminxscott:
In this game, the tile will upgrade to a regular city so I don't think it's a good example

I can grab an 18Mex game where it should have longer routes through such "dit cities"

Here's a game where P13 +dit is the terminal of a train, which seems to work fine via Auto
https://gist.github.com/benjaminxscott/fe8a1642cb65487165e68f2e343b0df9

JoeB989:
Ok, thanks, I'll add to my collection.

PR #6894, opened 2022-01-31
Fixed auto-routing when user supplies some routes

JoeB989:
Fixes issue #6882.
And fixed tiny intra-tile route (issue #6890)


lib/engine/auto_router.rb
```diff
@@ -34,21 +34,26 @@ def compute(corporation, **opts)
...
       # which can cause all train routes to conflict with each other
       node_timeout = path_timeout / nodes.size
+      path_walk_timed_out = false
 
       now = Time.now
 
       skip_paths = static.flat_map(&:paths).to_h { |path| [path, true] }
+      # if only routing for subset of trains, keep track of which ones we'll assemble routes for
+      skip_trains = static.flat_map(:train).to_a
+      look_trains = trains - skip_trains
```
tobymao, line 44: can you jsut do trains = trains - skip_trains

lib/engine/auto_router.rb
```diff
@@ -155,13 +161,19 @@ def compute(corporation, **opts)
         puts ' Node timeout reached' if node_abort
       end
 
+      @flash&.call('Auto route path walk failed to complete (PATH TIMEOUT)') if path_walk_timed_out
```
tobymao, line 164: does this actually work given that the calls are blocking?

lib/engine/auto_router.rb
JoeB989: Sure

lib/engine/auto_router.rb
JoeB989: It displays, but the combo loop is normally so fast I can't tell if the flash blocks until function ends.

tobymao: i'd put it at the end anyways

JoeB989: I just verified that it won't flash until the function ends.  If both timeouts occur, I'm not sure if user will see both flashes.

tobymao: just combine them into one

JoeB989: ok

lib/engine/auto_router.rb
JoeB989: Actually it turns out I can't, there's some weird crossover side effect afterwards when I click a train to clear it (an exception where '5' is not a train).  It's like `trains = trains - skip_trains` is modifying the actual train collection in the RR company.

tobymao (approved):

PR #6903, opened 2022-02-01
18EU Remove redundant tokens after upgrade

ryandriskel:

"These additional station tokens may be retained as long as they are in non-contiguous City circles: that is, until these cities are upgraded using the brown tiles designated for the purpose. When this occurs, any tokens beyond one in either Berlin or Vienna are removed from the map and returned to the Corporation for reuse: any tokens beyond one in each of “northern” and “southern” Paris are likewise removed from the map and returned to the Corporation for reuse."

Screen Shot 2022-02-01 at 12 21 01 AM

Screen Shot 2022-02-01 at 12 21 11 AM

### DIFF
--- a/lib/engine/game/g_18_eu/step/track.rb
+++ b/lib/engine/game/g_18_eu/step/track.rb
@@ -15,6 +15,23 @@ module Engine
 
             super
           end
+
+          def update_token!(_action, _entity, tile, old_tile)
+            super
+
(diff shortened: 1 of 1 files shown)


ryandriskel:
Crud, sorry Toby ... you hadn't reviewed this one yet. I thought I was on the Double city PR.

PR #6917, opened 2022-02-02
[1873] Remove restart logic from corporation_available?

magnusrk-ch:
1873 is the only company that has corporation_available? return false after it has previously returned true, which seems counter to the general purpose of that method, and also doesn't get applied in many parts of the game since corporation_available? is assumed to be a UI convenience.

Also fixes a line that would return nil in a method that's supposed to return a boolean.

### DIFF
--- a/lib/engine/game/g_1873/game.rb
+++ b/lib/engine/game/g_1873/game.rb
@@ -896,8 +896,6 @@ module Engine
 
         def corporation_available?(entity)
           return false unless entity.corporation?
-          return true if entity == @mhe
-          return can_restart?(entity, @round.active_step.current_entity) if entity.receivership?
 
           entity.ipoed || can_par?(entity, @round.active_step.current_entity)
         end
--- a/lib/engine/game/g_1873/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1873/step/buy_sell_par_shares.rb
@@ -31,7 +31,7 @@ module Engine
 
           def can_buy?(entity, bundle)
             corp = bundle.corporation
-            return if corp.receivership? && !@game.can_restart?(corp, entity)
+            return @game.can_restart?(corp, entity) if corp.receivership?
 
             super
           end


roseundy:
Why are you making these changes?

roseundy:
nil _is_ a boolean

roseundy:
Why is corporation_available? going false a problem?

magnusrk-ch:
There's no change in appearance or behavior. The check for MHE/can_restart? is simply unnecessary now that can_buy? correctly limits who can buy a company in receivership.

nil is falsey, but it's not a boolean. The difference is minor, but I don't see any reason not to return a proper boolean, and the code is slightly simpler to boot.

roseundy:
In other words, this is a gratuitous change because you don't like my code. My code is functional and you've proposed a change that is not correct. I will not approve this as it stands.

tobymao:
@magnusrk-ch i don't understand how your code in equivalent, seems different

roseundy:
I believe your first edit does makes a change in appearance: unavailable corporations have a dotted line outline which would now go away for receivership companies that can't be restarted by the current player. I'm more ambivalent about that change, but the existing code makes it more clear which companies have no actions.

magnusrk-ch:
> @magnusrk-ch i don't understand how your code in equivalent, seems different
Returning false from corporation_available? disables all interaction with that corporation. However, all the individual interactions for receivership companies are also disabled by their own specific conditions (e.g. can_buy? for buying the market shares), so either way the result is no actions allowed.

> I believe your first edit does makes a change in appearance: unavailable corporations have a dotted line outline which would now go away for receivership companies that can't be restarted by the current player. I'm more ambivalent about that change, but the existing code makes it more clear which companies have no actions.
No, the dotted line remains. It is based on whether there are any actions available, and that doesn't change.

roseundy:
The second edit you made is not equivalent

roseundy:
Sorry about that, I didn't realize my code-specific comments weren't visible until now.

PR #6919, opened 2022-02-02
only show bank when relevant

benjaminxscott:
The `undefined` element is no longer rendered if the bank is infinite, as in CZ
image

Games with a bank such as 46 still have it rendered
Screenshot 2022-02-02 3 37 47 PM


scottredracecar:
This is a good idea, but would it be better to show some kind of reminder
that the bank is infinite?

benjaminxscott:
This is mostly a bugfix rather than new feature

I think it used to show a bank when it was infinite (would show 99,999), and at some point that was abstracted to be hidden if it were infinite

tobymao (approved):

Issue #3527, opened 2021-01-25
[18MS] possible bug with extra tile lay in company's first OR?


DookTibs:
See https://18xx.games/game/23704?action=76

![first_or](https://user-images.githubusercontent.com/6773412/105759098-a4df9480-5f15-11eb-9dab-ff7d3f26c66a.png)

I was president of L&N with board state as shown in that link and the above screenshot. My understanding of the rules is that in a public company's first OR, they get a free extra yellow tile lay.

In this game the interface would allow me to place two yellows, for instance:

![twoyellows](https://user-images.githubusercontent.com/6773412/105758691-271b8900-5f15-11eb-89e8-7d007f27df43.png)

But if I first upgraded a hex, for instance Selma, then it would not give me an option to do the free yellow tile lay:

![onegreen](https://user-images.githubusercontent.com/6773412/105758755-3dc1e000-5f15-11eb-8a8d-dd22dfd88ac1.png)

I may well have misunderstood this rule; maybe you can't combine an upgrade with the free tile lay; if that's the case then there's no bug at all.

This was on Google Chrome 88.0.4324.96 (Official Build) (x86_64), OSX 10.13.6.

Maxwell-Frank:
You should definitely be allowed to upgrade and then lay a yellow. I swear this used to work. Maybe one of the recently added games broke it?

tobymao:
@jenf

jenf:
Probably not me, the code is quite explicit
[{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false }]

jenf:
The rule is within
11.4.2 LAYING YELLOW TILES
so it looks like it's only allowed if you lay a yellow tile.

@perwestling could you confirm?

Maxwell-Frank:
11.3.: 2. [companies may] Lay or upgrade one track tile, as described in 11.4. On its first operating turn only, a major company may lay an extra yellow tile before or after its normal tile lay. (Optional)

ikelly:
I'm running into the reverse issue in https://18xx.games/hotseat/hs_entxotne_41281 where L&N has just laid its initial yellow tile at C9 and now wishes to upgrade into D10 but isn't allowed. Since 11.3.2 says the extra yellow tile is "before or after its normal tile lay" I thought that this would be allowed.

ikelly:
For posterity from the game referenced above:

[...]

PR #6927, opened 2022-02-03
18ms double

thepackrat:
Fix for  https://github.com/tobymao/18xx/issues/3527

Adds the yellow + upgrade behaviour for first OR from 18MS rules clarification


perwestling (approved):

perwestling:
As a general I would recommend squashing commits if they are the same fix. Easily done using "git rebase -i".

This commit has a conflict so it cannot be rebased. Maybe apply the changes as one commit on latest master, and push again?

PR #6931, opened 2022-02-03
18EU Pullman Trains

ryandriskel:

This train can be run to a single stop a different train is already visiting.

- Must have a train to purchase a Pullman
- Forced discarded if it is the corporation's only train
- Forced discarded if the corporation is over train limit
- Cannot be bought over
- Can be optionally discarded before buying trains

Route Running

Screen Shot 2022-02-03 at 12 00 47 PM

Screen Shot 2022-02-03 at 12 04 27 PM

Optional Discard

Screen Shot 2022-02-03 at 12 00 47 PM

Forced discard

Screen Shot 2022-02-03 at 12 02 48 PM


lib/engine/game/g_18_eu/game.rb
```diff
@@ -322,6 +323,58 @@ def after_par(corporation)
 
           corporation.spend(TOKENS_FEE, @bank)
         end
+
+        def check_overlap(routes)
+          super
+
+          pullman_stop = routes.find { |r| pullman?(r.train) }&.visited_stops&.first
+          return if pullman_stop.nil?
```
tobymao, line 331: return unless pullman

tobymao (approved):

Issue #2548, opened 2020-12-04
[1848] Rules Implementation


benjaminxscott:
[rules pdf ](http://ohley.de/english/1848/1848-rules.pdf)

[My latest dev branch](https://github.com/benjaminxscott/18xx/tree/1848) (likely unstable)

### Capital
- Supports 3 - 6 players, receiving {840, 630, 510, 430} 
- Bank starts with 10,000 dollarydoos

### Auction
- [x] Dutch auction with rules:

- On a player's turn, they have two actions: buy a private or reduce the price a private
- Once a player owns a private, they have an additional action: passing
- If all privates have been reduced to their minimum price and nobody has bought a private yet, the currently active player must buy one
- If any privates are left and all players have passed in a row, an "OR" happens where privates pay and the auction continues

### Stock Rounds
- Sell any number of shares, then buy up to one share, or pass
- Priority order works as normal (left of the last player to take an action in SR)

#### Certificate Limits
- [ ] In 3p game, can hold 70% of corp, else 60%
- [x] Privates do not count against cert limit
- [ ] After a company goes into receivership, the cert limit for all players goes down 

| Players | 0 Receiverships | 1 | 2 | 3 | 4 | 5+ |
|---|---|---|---|---|---|---|
| 3      | 20 |  18 |16 |14 |12 |10|
| 4      | 17 |15 |13 |11 |10 |9|
| 5      | 14 |13 |12 |10 |9 |8 |
| 6      | 12 |11 |10 |9 |8 |7 |

#### Launching Corporations
- 60% to float, full-cap
- May be parred at {60,70,80,90,100}
- [x] CAR is always parred at 100 (Presidency is granted by P6)

- [x] COM receives two free home tokens, in Sydney and Adelaide
- [x] COM is parred by whomever buys its Presidency, and may only operate once 60% of its shares have been bought and Sydney is connected to to Adelaide (e.g. if a D train ignoring tokens could connect the two cities)
- [ ] From Phase 6 on, COM can operate without the above requirement
- [ ] If all the conditions for the launch of the COM are fulfilled before it would be the COM's turn to operate in the relevant OR (according to share price order), it starts to operate in the same OR. If the second condition (the Sydney-Adelaide link or the purchase of the first 6/6+ train) is fulfilled for the first time by a company with a lower share price, the COM will operate for the first time in the next OR.
[...]

PR #6933, opened 2022-02-03
1848 Trains as variants and 2E Buy Rules

lkingsford:
See issue #2548 (@benjaminxscott - if you want to update some of those tasks)

- Can buy current Train or Plus-Train now, rather than all train, then all plus-trains, then the next train.
- 2E available when you buy first 5, doesn't count against train limit, doesn't count for compulsory purchase, can only buy 1. Transfers are not yet prevented.

Trains are formatted that takes up more space, but copied the 1846 styling with its variants (and then changed remaining lines for consistency). I hope that's OK.

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -264,6 +264,7 @@ module Engine
             distance: 2,
             price: 200,
             num: 6,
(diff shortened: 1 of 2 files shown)


benjaminxscott:
ok

PR #6943, opened 2022-02-04
 Count future unstarted corps (fixes #5285) 

### LINKED ISSUE #5285: [18Ireland] Display Number of Unstarted Majors from Start of Game
https://18xx.games/game/40986?action=82

For those of us without great memories, it might be nice to know how many major companies are available in the game from the beginning, even though they're not yet available to start. Currently the "unstarted majors" displays as 0.

dfannius:


lib/engine/game/g_1867/game.rb
```diff
@@ -1174,7 +1174,7 @@ def sorted_corporations
         end
 
         def unstarted_corporation_summary
-          unipoed = @corporations.reject(&:ipoed)
+          unipoed = (@corporations + @future_corporations).reject(&:ipoed)
```
tobymao, line 1177: why is this copy and pasted from 1861, isn't 67 a subclass of 61? do we even need this method?

same with ireland, the ireland code is better because it uses partition, can you clean that up?

dfannius: This is the first time I've looked at any of three three titles and I don't know their rules; I was just trying to clean up an easy issue. The 1861 and 1867 `unstarted_corporation_summary` methods are slightly different in the way they present the information (the string they return); someone who knows the titles might be able to say why. It doesn't seem that 1861 or 1867 can simply use `partition` because they have a third type called `:national`.

I could make a 1867 helper method that does almost all of the work and then separate methods for 1861 and 1867 that format the results differently as a string if you would prefer that. I think Ireland has to stay separate.

Issue #6899, opened 2022-02-01
Add async/live game setting and game card tag


marndt93:
It is common for users to tag their games as "live" or "async" in the game description.  It can be confusing if a game is not tagged in the description whether the game is intended to be played live or async.  This could be solved by adding radio buttons to select live/async options when creating a new multiplayer game (default async).

example mock up:

![image](https://user-images.githubusercontent.com/28715646/151916417-47169213-1544-4d98-ac85-de0744f87cb5.png)

scottredracecar:
Maybe even more apparent by coloring the yellow header instead and having the text under the Join button?

marndt93:
> Maybe even more apparent by coloring the yellow header instead and having the text under the Join button?

Do you mean something similar to this?
![image](https://user-images.githubusercontent.com/28715646/151918852-ab73b0ad-9a31-48a4-9984-cbf10c71aeb3.png)

tobymao:
oh god no

marndt93:
> oh god no

lol, not a fan of doing it this way either.  The header looks too cluttered

scottredracecar:
Well, conceptually, yes, but not those colors of course. The current yellow is fine for async and then for live, you could use #DAF7A6 for example. I can help pick different alternative colors too.

I don't like that the extra word makes the header bigger so I would suggest to put that in the button "Join Async" or "Join Live"

marndt93:
So you're thinking like this:
![image](https://user-images.githubusercontent.com/28715646/152049956-3908346d-e6c3-4e49-bd9d-3bc9c3d76da3.png)

scottredracecar:
Yes, but to pick a color that is a bit more distinct than yellow, perhaps
try #91F0FF?

scottredracecar:
Could you show a screenshot using #91F0FF? I really like this idea.

marndt93:
@scottredracecar 
![image](https://user-images.githubusercontent.com/28715646/152464068-4ccff715-3c7e-4304-996b-6ee38eabe4b1.png)

scottredracecar:
Looks good to me.

marndt93:
should I proceed with this design then?
@tobymao @scottredracecar

scottredracecar:
yes

PR #6945, opened 2022-02-04
add async/live game setting

marndt93:
resolves: #6899


assets/app/view/create_game.rb
```diff
@@ -328,6 +330,25 @@ def mode_input(mode, text)
...
     end
 
+    def render_game_type
+      h(:div, { style: { padding: '0.5rem' } }, [
+        render_input(
+          'Async',
+          id: 'async',
+          type: 'radio',
+          attrs: { name: 'is_async', checked: @is_async == true },
+          on: { click: -> { @is_async = !@is_async } }
```
tobymao, line 340: are you sure this is right? shouldn't this use store?

tobymao (approved):

PR #6946, opened 2022-02-04
cleanup autorouter

tobymao:


lib/engine/auto_router.rb
```diff
@@ -64,18 +61,13 @@ def compute(corporation, **opts)
...
 
         node.walk(corporation: corporation, skip_paths: skip_paths) do |_, vp|
-          next :abort if node_abort
-
-          paths = vp.keys
-
-          if Time.now - node_now > node_timeout
-            # TODO: this is not a complete node.walk abort, find somthing more than :abort but less than break.
-            # Or wait til node.walk speed is improved and this may become less important
+          if node_abort || Time.now - node_now > node_timeout
```
JoeB989, line 64: Once node timeout hits, node_abort is true many times through this node.walk.  So any code that executes when node_abort true is unnecessary.  Granted you're only executing a few lines, but this change risks losing that knowledge "don't do work if node_abort true".

I had a `puts "node_abort in node.walk" if node_abort `and it would print hundreds of times in some cases.

lib/engine/auto_router.rb
```diff
@@ -288,130 +270,129 @@ def js_evaluate_combos(_rb_sorted_routes, _route_timeout)
       conflicts = 0
       now = Time.now
 
-      %x(
```
JoeB989, line 291: It seems like only spacing changed.  Is there a way to suppress spacing diffs in the compare?  Or at least see them side by side?

JoeB989 (approved):

lib/engine/auto_router.rb
```diff
@@ -125,45 +117,35 @@ def compute(corporation, **opts)
...
-            # keep extending this connection set
+            path_abort.delete(train)
           rescue ReusesCity
-            abort = :abort
-          rescue GameError => e
-            puts e
+            path_abort.clear
+          rescue NoToken, RouteTooShort, GameError # rubocop:disable Lint/SuppressedException
           end
-          abort
```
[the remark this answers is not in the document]
tobymao: abort doesn’t do anything you made it a variable = true

lib/engine/auto_router.rb
tobymao: corrext

PR #6953, opened 2022-02-04
1825: add 2nd Midland reservation when using units 1 and 2

roseundy:
Fixes issue raised by @volker18xxdev 
Required minor refactoring of '25 code to handle multiple "blocks_hexes" abilities
Also adds fixture for '25

This should not require pins

### DIFF
--- a/lib/engine/game/g_1825/game.rb
+++ b/lib/engine/game/g_1825/game.rb
@@ -1176,6 +1176,11 @@ module Engine
           entity.close!
         end
 
+        def hex_blocked_by_ability?(_entity, abilities, hex)
+          abilities = Array(abilities)
+          abilities.any? { |ability| ability.hexes.include?(hex.id) }
+        end
+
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1825/game.rb
```diff
@@ -1176,6 +1176,11 @@ def check_bankrupt!(entity)
           entity.close!
         end
 
+        def hex_blocked_by_ability?(_entity, abilities, hex)
+          abilities = Array(abilities)
+          abilities.any? { |ability| ability.hexes.include?(hex.id) }
```
tobymao, line 1181: can you make this a one liner Array(...)any same with the one below

tobymao (approved):

lib/engine/game/g_1825/game.rb
roseundy: Yep

PR #6960, opened 2022-02-05
18Rhl: Corrections of privates and map

### LINKED ISSUE #6948: [18Rhl] p4 implemented incorrectly
https://18xx.games/game/70978

“ As director of a corporation the owner may upgrade *one* of the yellow hexes of Köln / Düsseldorf / Duisburg for free. This tile placement is in addition to the corporation's normal tile lay. The corporation may place a station token there in the same OR by paying the appropriate costs. There need not be a link to the corporation's network. If the station marker is not placed in the same Operating Round but later, the corporation must have a track connection to that hex. After the placement of the station token the corporation may not place a tile even when it has not used its normal tile lay (placing a station marker is always the step after the tile lay!). Note! Token can only be used together with the upgrade to green and not as a token-only action. If all applicable hexes are green the special ability is no longer usable.”

 p4 is supposed to allow you to upgrade for free and pay for a token, but it is charging for the upgrade and the token is free.

perwestling:
P1 (Ratingen): Set attribute that using this does not consume regular
tile lay. Also, restrict so this P1 cannot be used during yellow phase.

P2: Bug fixes: Make tile lay free. Also fixed a crash.
    And updated description.

P3: Set attribute that using this does not consume regular
tile lay.

P4: Bug fixes: Make tile lay free, and token cost is not free.
    Also fixed a crash. And updated description. This fixes https://github.com/tobymao/18xx/issues/6948

P6: Correct description.

Finally, also do not allow any tile laying of any kind after
a tokening has taken place.

Also have transformed some unused red hexes to gray.

### DIFF
--- a/lib/engine/game/g_18_rhl/entities.rb
+++ b/lib/engine/game/g_18_rhl/entities.rb
@@ -167,6 +167,7 @@ module Engine
                           special: true,
                           count: 1,
                           when: %w[owning_player_or_turn],
+                          consume_tile_lay: false,
                         }],
           },
           {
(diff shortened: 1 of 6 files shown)


perwestling:
If this breaks any (alpha) games, archive them.

tobymao (approved):

Issue #6961, opened 2022-02-05
1860: Indicate shortened trains during nationalisation phase


thepackrat:
When the game moved into the nationalisation phase, trains are shortened to remove the small station quota, so a 9+5 becomes a 9 train, with small stations optional but counting against main length.

It would be good to indicate on the train name in the routing display this new length rather than the current behaviour which shows the original name (8+4) and then reuses the route counter to count small and large stations, for e.g. 6+2 as used. e.g

train | used 
8+4 | 5+3 | £260 | J8-J6-J4-I5-H4-H4-G3-G5-F6-E5-D6-C5-B4

i suggest changing the train name to either
8+0, or 8 (8+4)

and having the used display reflect this to better indicate what's going on in that state.

thepackrat:
Discussed with Steve, going with 8 (8+4) as the shortened name.

PR #6962, opened 2022-02-05
1860: Displays the shortened trains in nationalization

thepackrat:
In the nationalization N+m trains run as N train only. This displays a hint for this in the nationalisation phase.

Fixes https://github.com/tobymao/18xx/issues/6961


roseundy (approved):
Looks good

PR #6963, opened 2022-02-05
refactor check_connected to not use node.select

### LINKED ISSUE #6157: [18Ireland] Companies can run through tokened-out Belfast
https://18xx.games/game/50876?action=1432

It looks like companies can run through Belfast, if that tile is tokened out, when specifically running through the H8 tile twice. In the pasted game:
- Belfast has been tokened by the CBSCR and the MRNCC
- The GSWR during its operating turn is able to run a diesel train through Belfast when using the southwestern and southern pieces of track
- The run is counted as legal, and paid out accordingly

This was replicated by other companies, but not used in the game. Using the GNRI I tried to enter through the northwest entrance and exit through the south but was notified that the route was not connected. 

Playing around in hot-seat mode, it might depend on the gray plain-tile tracks. See following screen-shots:
Screen Shot 2021-08-01 at 1 18 36 PM

Screen Shot 2021-08-01 at 1 18 54 PM

In the first, by passing through H8 twice the GNRI can legally run through Belfast to the I9 city. In the second, by going directly from Belfast to I9 the route is considered not connected.

Further, when I run through Belfast and stop short of crossing H8, the route is illegal:
Screen Shot 2021-08-01 at 1 32 05 PM
but if I continue and connect to another stop through H8 it is deemed legal:
Screen Shot 2021-08-01 at 1 32 11 PM

At first we thought it was specific to the southwest and south entrances, but I am able to get a legal route using the northwest and south track:
Screen Shot 2021-08-01 at 1 36 33 PM

I'm running the game on Safari, but we've been playing in Chrome and on phones too, so I'm not sure if it is a browser issue.

roseundy:
Fixes #6157 

Completely removes Node::select and Path::select methods
Removes "on" parameter from Node::walk and Path::walk
Refactors Step::SpecialTrack to avoid Path::select

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1160,9 +1160,9 @@ module Engine
         end
       end
 
-      def check_connected(route, _token)
-        corporation = route.corporation
-        return if route.ordered_paths.each_cons(2).all? { |pair| pair[0].connects_to?(pair[1], corporation) }
+      def check_connected(route, token_city)
+        corporation = token_city ? route.corporation : nil
+        return if route.ordered_paths.each_cons(2).all? { |a, b| a.connects_to?(b, corporation) }
 
         raise GameError, 'Route is not connected'
       end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1160,10 +1160,9 @@ module Engine
         end
       end
 
-      def check_connected(route, token)
-        paths_ = route.paths.uniq
-
-        return if token.select(paths_, corporation: route.corporation).size == paths_.size
+      def check_connected(route, _token)
+        corporation = route.corporation
+        return if route.ordered_paths.each_cons(2).all? { |pair| pair[0].connects_to?(pair[1], corporation) }
 
         raise GameError, 'Route is not connected'
       end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1160,8 +1160,7 @@ module Engine
         end
       end
 
-      def check_connected(route, token_city)
-        corporation = token_city ? route.corporation : nil
+      def check_connected(route, corporation)
         return if route.ordered_paths.each_cons(2).all? { |a, b| a.connects_to?(b, corporation) }
 
         raise GameError, 'Route is not connected'
--- a/lib/engine/game/g_1860/game.rb
+++ b/lib/engine/game/g_1860/game.rb
@@ -1558,25 +1558,7 @@ module Engine
           visits[1..-2].any? { |node| node.city? && custom_blocks?(node, corporation) }
         end
 
-        # change all bankrupt flipped tokens to neutral
-        def flipped_to_neutral
-          @bankrupt_corps.each do |corp|
-            corp.tokens.each do |token|
-              token.type = :neutral if token.status == :flipped
-            end
-          end
-        end
-
-        # change all bankrupt neutral tokens to flipped
-        def neutral_to_flipped
-          @bankrupt_corps.each do |corp|
-            corp.tokens.each do |token|
-              token.type = :normal if token.status == :flipped
-            end
-          end
-        end
-
-        def check_connected(route, token)
+        def check_connected(route, _token)
           visits = route.visited_stops
           blocked = nil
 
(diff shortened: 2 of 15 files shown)


lib/engine/game/base.rb
```diff
@@ -1160,10 +1160,9 @@ def check_overlap(routes)
         end
       end
 
-      def check_connected(route, token)
-        paths_ = route.paths.uniq
-
-        return if token.select(paths_, corporation: route.corporation).size == paths_.size
+      def check_connected(route, _token)
+        corporation = route.corporation
+        return if route.ordered_paths.each_cons(2).all? { |pair| pair[0].connects_to?(pair[1], corporation) }
```
tobymao, line 1165: all? {|a, b| a.connects_to?(b)

lib/engine/game/g_1860/game.rb
```diff
@@ -1590,17 +1572,11 @@ def check_connected(route, token)
...
           end
 
-          paths_ = route.paths.uniq
-          token = blocked if blocked
-
-          flipped_to_neutral
-          if token.select(paths_, corporation: route.corporation).size != paths_.size
-            neutral_to_flipped
+          # no need to check whether cities are tokened out because of the above
+          unless route.ordered_paths.each_cons(2).all? { |pair| pair[0].connects_to?(pair[1], nil) }
```
tobymao, line 1576: a, b

lib/engine/game/g_18_usa/game.rb
```diff
@@ -753,10 +753,8 @@ def check_connected(route, token)
             end
           end
 
-          paths_ = route.paths.uniq
-          token = blocked if blocked
-
-          return if token.select(paths_, corporation: route.corporation).size == paths_.size
+          # no need to check whether tokened out because of the above
+          return if route.ordered_paths.each_cons(2).all? { |pair| pair[0].connects_to?(pair[1], nil) }
```
tobymao, line 757: couldn't this be cleaned up? just be like route.connected? with an arg to pass in corp or nil,

lib/engine/part/path.rb
```diff
@@ -104,21 +104,28 @@ def ends
...
-
-        walk(on: on) do |path|
-          on[path] = 1 if on[path]
+      def connects_to?(other, corporation)
+        [@a, @b].each do |part|
+          if part.edge?
+            edge = part.num
+            if (neighbor = hex.neighbors[edge])
+              np_edge = hex.invert(edge)
+              return true if neighbor == other.hex && other.edges.find { |e| e.num == np_edge }
```
tobymao, line 113: edges.any? { |... }

lib/engine/game/base.rb
roseundy: Didn't know that was a thing. Will fix.

lib/engine/game/g_1860/game.rb
roseundy: OK

lib/engine/game/g_18_usa/game.rb
roseundy: Yes, I meant to do this, but got side-tracked.

lib/engine/part/path.rb
roseundy: OK

lib/engine/game/g_1860/game.rb
roseundy: Actually, this will be cleaned up the same as 18USA

lib/engine/game/base.rb
```diff
@@ -1160,10 +1160,9 @@ def check_overlap(routes)
         end
       end
 
-      def check_connected(route, token)
-        paths_ = route.paths.uniq
-
-        return if token.select(paths_, corporation: route.corporation).size == paths_.size
+      def check_connected(route, token_city)
+        corporation = token_city ? route.corporation : nil
```
tobymao, line 1164: corporation = route.corporation if token_city

tobymao: and you can do token_city = nil so that's it's not required

tobymao: this method is weird now, it used to take in the token_city, but not anymore, you don't actually use the token city at all, can you clean this up?

lib/engine/game/g_1888/step/special_track.rb
```diff
@@ -27,37 +27,6 @@ def legal_tile_rotation?(entity, hex, tile)
 
             tile.rotation.zero?
           end
-
-          # needed to deal with non-plain track connections
-          #
-          def check_connect(_action, ability)
```
tobymao, line 33: out of curiosity how is it possible you deleted this

roseundy: The base method works fine now. Before, it would work with non-plain tiles.

lib/engine/game/base.rb
roseundy: Yeah, I realized over night I should have done that.

roseundy: I will pass corporation instead of token to check_connected

scottredracecar:
should also fix 3476

dfannius:
Yes, that is totally the same issue, nice find.

PR #6968, opened 2022-02-05
fix price protect cert limit check

### LINKED ISSUE #6001: 1870: Share Count Incorrectly Calculated during price protect
https://18xx.games/game/40469?action=883

At this point in the game, it says that Roller has 9/16 shares.

They hold:

SP (60%*) (0 - in the yellow area)
FW*	50% (4)
ATSF*50% (4)
MKT*40% (3)
SLSF 10% (1)
GMO 10% (1)

This adds up to 13. When they price protect 4 shares, their share count jumps up to 17/16, which is correct but shouldn't be allowed as price protection isn't an option if it will push you over share limit.

JoeB989:
Fixes #6001


JoeB989:
need to resubmit PR

Issue #6704, opened 2022-01-03
1870 K and L icons don't persist


ChrisTheCat:
https://18xx.games/game/59924?action=928

On the starting map, both the L and P icons appear on St. Louis. Once the brown P tile is laid, the L disappears. Same issue for Kansas City.

<img width="259" alt="Screen Shot 2022-01-03 at 14 05 09" src="https://user-images.githubusercontent.com/8732435/147985581-d2acb9e1-57eb-4e91-94f8-4f27909b5101.png">

<img width="340" alt="Screen Shot 2022-01-03 at 14 05 27" src="https://user-images.githubusercontent.com/8732435/147985587-bf530879-34e3-439d-b8f7-1625ca570142.png">

JoeB989:
Are the K and L still respected for the grey upgrade?

ChrisTheCat:
I believe the answer is yes.

thepackrat:
This matches the physical game, no?

Well, we can do better. We can't currently fix this properly because the pointy render isn't great with multiple labels and complex cities, but we can fix it so it's playable and then come back and render it prettily later.

PR #6970, opened 2022-02-05
1870: Fixes all the multi-label punches for tile lays

thepackrat:
This deliberately doesn't use multiple labels because the pointy
weights don't currently render multiple labels for cities like a 14.

Fixes https://github.com/tobymao/18xx/issues/6704


thepackrat:
Stumbled across a promotion corner case in the label wrangling. Fix in flight.

PR #6974, opened 2022-02-05
Adds an announcement of multi-jumps,

thepackrat:
Ads a specific fix for 1860 and 1862 which have folded flat markets, all other games don't need a fix.

multi-step share price increases now look like this:  IOW's share price changes from £68 to £100 (4 steps)
single steps are unchanged.


thepackrat:
Note, this works as expected for 22 and 46 out of the box

PR #6984, opened 2022-02-06
1877 - Add mine tiles, add "No Mine Tiles" optional variant

scottredracecar:
Requires pins please


tobymao:
how are you handling extra revenue

tobymao (approved):

PR #7000, opened 2022-02-06
[18Rhl] should be one fewer 2 train

### LINKED ISSUE #6993: 18Rhl: Wrong number of 2 Trains
Checking the 18Rhl rulebook it says that there are 28 trains. But it does not specify the number per train type.

Checking my physical (Marflow) copy of 18Rhl this seem to be the distribution:

- 6 2T
- 1 optional 2T
- 5 3T
- 3 4T
- 3 5T
- 6 6T
- 4 8T

Adding this up makes it 28.

Checking 18Rhl at 18xx.games it show this distribution:
image

So it seems like 18Rhl @ 18xx.games does have one 2T too many.

benjaminxscott:
closes #6993

without optional 2train added:
image

### DIFF
--- a/lib/engine/game/g_18_rhl/game.rb
+++ b/lib/engine/game/g_18_rhl/game.rb
@@ -214,7 +214,7 @@ module Engine
         def num_trains(train)
           return train[:num] unless train[:name] == '2'
 
-          optional_2_train ? 8 : 7
+          optional_2_train ? 7 : 6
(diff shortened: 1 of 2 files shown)


perwestling:
It is a correct change but I waited to do it, as it would break most started 18Rhl games when it is deployed. But yes, it is alpha games.

PR #7002, opened 2022-02-06
price-protection fixes and tests

### LINKED ISSUE #6001: 1870: Share Count Incorrectly Calculated during price protect

JoeB989:
Fixed #6001
Fixed cert counting of white/yellow shares during price protection 
Added in-game test spec and price-protect fixtures


spec/lib/engine/games/game_spec.rb
```diff
@@ -106,4 +106,31 @@ module Engine
       expect(game.round.active_step).to be_instance_of(Engine::Step::Route)
     end
   end
+
+  # active-game tests: run all .ingame files found in spec/fixtures/,
+  # verify that the result and last "test_last_actions" actions match those in the JSON
+  Find.find(FIXTURES_DIR).select { |f| File.basename(f) =~ /.ingame/ }.each do |fixture|
```
tobymao, line 112: can you just do the normal one and delete this method?

JoeB989: The normal one expects completed game, this one expects partial game stopped where the actions to be tested are.  I could remove the` expect(rungame.finished).to eq(true)` above, but I didn't know what other side effects I might run into.

tobymao: can you just end your games with a end game action?

JoeB989: ah hopefully.  I 'll try

tobymao (approved):

PR #7018, opened 2022-02-07
Simplify receivership and trainless-company-shares behaviour to make it easier to use in multiple games

andypymont:
- Make the base game's `init_share_pool` method use the `PRESIDENT_SALES_TO_MARKET` constant - therefore games no longer need to override both the constant and the method to enable presidencies to be sellable
- Move the shared behaviour of trainless corporation shares being worth 50% in 1860, 1862, and 18GB into a single module for improved maintainabillity

### DIFF
--- a/lib/engine/game/trainless_shares_half_value.rb
+++ b/lib/engine/game/trainless_shares_half_value.rb
@@ -26,16 +26,8 @@ module TrainlessSharesHalfValue
     end
   end
 
-  def player_shares_value(player)
-    trainless_shares, train_shares = player.shares.partition { |s| s.corporation.trains.empty? }
-    train_shares.sum(&:price) + trainless_shares.sum { |s| (s.price / 2).to_i }
-  end
-
   def player_value(player)
-    player.cash + player_shares_value(player) + player.companies.sum(&:value)
-  end
-
-  def liquidity(player)
-    player.cash + player_shares_value(player)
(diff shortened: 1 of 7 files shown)


lib/engine/game/trainless_shares_half_value.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+#
+# This module makes the shares of companies without a train worth half the usual value
```
roseundy, line 4: The player value here also assumes a player can sell a presidents share to the market - which is somewhat orthogonal to trainless companies having price shares. So far both have been true together, but that might not necessarily be always true.

andypymont: Great catch, thanks Steve - I will get this to look for the flag on the game and react accordingly.

roseundy:
This looks good to me. I really should have done this when I implemented 1862.

lib/engine/game/trainless_shares_half_value.rb
andypymont: The original liquidity calculation is already using `bundles_for_corporation`, so this turned out to actually be quite a simplification - just needed to account for `PRESIDENT_SALES_TO_MARKET` in one more place. Testing now and will add a commit.

tobymao (approved):

PR #7033, opened 2022-02-08
21Moon: add more grays; new icons; test play bug fixes

roseundy:
Almost done!
- added 1846 style bankruptcy support
- renamed base camp to home base (many places)
- renamed "transport" to "train" (many places)
- implemented rule about not swapping the same stock certificate twice
- reserved one out of the seven starting corps
- home base highlighting
- allow over 50% holding when using P6
- map now uses shades of gray rather than "white" hexes
- updated most of the icons

Common file edits:
- assets/app/lib/hex.rb: new gray colors
- lib/engine/config/tile.rb: new gray colors
- lib/engine/round/base.rb: use self.class:: before DEFAULT_STEPS constant (I wound up not needed this, but it still seems like a good idea)


roseundy:
The map:
21moon_map_latest

tobymao (approved):

PR #7040, opened 2022-02-08
prevent Miss River crossing tile lays if Bridge private not owned by a corporation

### LINKED ISSUE #6331: [1870] Can lay double dit tile across river before Mississippi Bridge company is bought in
https://18xx.games/game/55344?action=250

I was able to lay a straight in A14 and the double dit #2 in A16 across the river even though I think we aren't supposed to be allowed to do that until the MS Bridge private has been bought in.

### LINKED ISSUE #6659: 1870 crossing the Mississippi River before the bridge private is purchased
https://18xx.games/game/64871?action=160#map

Screen Shot 2021-12-22 at 06 48 29

MP crossed the river while Hobie still owns the Mississippi River Bridge private company.

JoeB989:
Fixes #6331
Fixes #6659


lib/engine/game/g_1870/game.rb
```diff
@@ -791,8 +791,12 @@ def legal_tile_rotation?(_entity, hex, tile)
 
           (tile.exits & hex.tile.borders.select { |b| b.type == :water }.map(&:edge)).empty? &&
             hex.tile.partitions.all? do |partition|
-              tile.paths.all? do |path|
-                (path.exits - partition.inner).empty? || (path.exits - partition.outer).empty?
+              if partition.restrict != ''
```
tobymao, line 794: this seems strange to me, what if restrict is nil?

JoeB989: It wasn't in my testing, but I can add a check.  Is there a shorthand way to test nil or blank?

tobymao: not in default ruby

tobymao (approved):

PR #7067, opened 2022-02-10
[1866] Playtest fixes

magnusfalt:
- Redid the whay last train is ignored. Couldnt do this other than orerride the depot class
- Fixed a transist hub bug
- Modified Stock Turn Tokens end game trigger

### DIFF
--- a/lib/engine/game/g_1866/depot.rb
+++ b/lib/engine/game/g_1866/depot.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require_relative '../../depot'
+
+module Engine
+  module Game
+    module G1866
+      class Depot < Engine::Depot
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1866/depot.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require_relative '../../depot'
+
+module Engine
+  module Game
+    module G1866
+      class Depot < Engine::Depot
+        def depot_trains(clear: false)
```
tobymao, line 9: is this needed? can you just use available on? what exactly is this class doing

magnusfalt:
Closing until the get the go this is what Simon wants.

PR #7083, opened 2022-02-11
use pill in game card to tag live games

marndt93:
image
@scottredracecar you had a lot of input about the colors for the original design.  Any input?  I just used the same color I used in the mock up here.

@tobymao


assets/app/view/game_card.rb
```diff
@@ -237,7 +232,11 @@ def render_body
         elm
       end
 
-      children = [h(:div, [h(:strong, 'Id: '), @gdata['id'].to_s])]
+      row_styles = { style: { display: 'flex', flexDirection: 'row', justifyContent: 'space-between' } }
+      pill_styles = { style: { background: 'darkred', borderRadius: '30px', padding: '0px 5px' } }
+      id_row = [h(:div, [h(:strong, 'Id: '), @gdata['id'].to_s])]
+      id_row << h(:div, pill_styles, 'Live') unless @gdata['settings']['is_async'] || @gdata['settings']['is_async'].nil?
```
tobymao, line 238: don't use unless with multiple conditions

marndt93: you just prefer a regular if statement in this case?

tobymao (approved):

PR #7084, opened 2022-02-11
show correct active player if there are multiple

### LINKED ISSUE #5847: [1870] Simultaneous Connection Runs tell multiple players it's their turn
https://18xx.games/game/37799?action=598

3 companies with 3 different presidents could reach their destination through a single tile lay. The UI which tells players to take their turn was on for all three of those players, even though only one of them was actually up to play.
1870 bug 1
1870 player order 2
1870 player order 3

### LINKED ISSUE #7091: [1817] - Confirm the Market consents to share redemption?
https://18xx.games/game/63692?action=942

Redeeming shares from the market triggered a dialog with "Click confirm if Market has already consented to this action." Clicking confirm worked and allowed the game to proceed, but that shouldn't have shown up.

Screenshot 2022-02-11 07 38 26

JoeB989:
Use current_entity instead of active_players_id in case there are multiple "active" players
Fixes #5847
Fixes #7091
1870 destination connections may cause multiple railroads and players to be "active", but of course only one of them is really the active player.  The Choose buttons work properly, erroring for non-active player. Only the game header showed the active color for all destinating players.

### DIFF
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -284,7 +284,7 @@ module View
     def active_player
       @game_data[:mode] != :hotseat &&
         !cursor &&
-        @game.active_players_id.include?(@user&.dig('id'))
+        @game.current_entity.player.id == @user['id']
     end
 
     def menu
--- a/assets/app/view/game_page.rb
+++ b/assets/app/view/game_page.rb
@@ -284,7 +284,7 @@ module View
     def active_player
       @game_data[:mode] != :hotseat &&
         !cursor &&
-        @game.current_entity.player.id == @user['id']
+        @game.current_entity.player.id == @user&.dig('id')
     end
 
     def menu
--- a/assets/app/view/game/issue_shares.rb
+++ b/assets/app/view/game/issue_shares.rb
@@ -44,7 +44,7 @@ module View
             end
 
             # confirm if redeeming from a different player
-            if (bundle.owner != @game.bank) && (bundle.owner != @game.current_entity)
+            if (bundle.owner != @game.bank) && (bundle.owner != @game.current_entity) && bundle.owner.player?
               check_consent(bundle.owner, process_redeem)
             else
               process_redeem.call


assets/app/view/game_page.rb
```diff
@@ -284,7 +284,7 @@ def render_title
     def active_player
       @game_data[:mode] != :hotseat &&
         !cursor &&
-        @game.active_players_id.include?(@user&.dig('id'))
+        @game.current_entity.player.id == @user['id']
```
tobymao, line 287: if @user is none this will crash,

so why are we current_entity? it should be active_players

JoeB989: In my test there are three active players.  All three got the active banner - that's the bug.  Should it be this?
`@game.current_entity.player.id == @user&.dig('id')`

JoeB989: Only one of them is allowed to act, although all 3 get the orange "your turn" banner.  The game operates properly no matter who hits which button ("not your turn" error as appropriate), it's just misleading UI.

JoeB989: I made this change and tested some games through it.

tobymao:
issues still haven’t been addressed

tobymao:
in 1870, if connection allows multiple corporations to destinate, then isn't the UI correct -- multiple people are active players and so should be notified as such

JoeB989:
Well the UI confused the players, they thought they were allowed to control other player's railroad.  Corporations are supposed to run sequentially per the rules.  It's kind of like a mini OR.  The implementer just chose to compress that mini OR into a single step.

JoeB989:
Here's what it looks like without the fix.  All three look like they are destinating ATSF
3 active players

tobymao:
then 1870 is wrong and should be fixed, not the generic code

JoeB989:
Canceling PR since will require an 1870 design change

PR #7087, opened 2022-02-11
Flip positions of Minors 6 and 11

### LINKED ISSUE #7081: [18EU] Minors 11 and 6 and flipped
Chris Schaffer provided a snippit of the map and you can see that minors 11 and 6 should be flip flopped so that 11 is on the bottom and 6 is on top.

kepakiano:
Fixes #7081


kepakiano:
hot-seated until green and it seems to work as expected:
image

PR #7094, opened 2022-02-11
Run routes through correct nodes when deserializing (fixes #5279)

### LINKED ISSUE #5279: [18Ireland] pre filled track jumps to wrong city on DD tile (I9)
https://18xx.games/game/40351?action=804

If a company's track uses the small dit on DD tile (I9), the pre filled track jumps to the big city, whether the company runs this track or not.

dfannius:
When serializing a route, save off its nodes (both the hex and the spot on the tile).

When deserializing a route, if the saved nodes are present and there's a choice of nodes to use for some tile, use the one that matches the saved nodes.

Tiles that can fail to deserialize routes correctly without this code include IR7, IR8, and 457-464.

If saved nodes are not present then the deserialization is the same as before, so it's backward compatible.

This code is fine with respect to the comment at https://github.com/tobymao/18xx/issues/5279#issuecomment-1034000192, since if the deserialization code can't find a chain that hits the saved nodes, it just returns the first possibility, as before.

### DIFF
--- a/lib/engine/part/base.rb
+++ b/lib/engine/part/base.rb
@@ -14,7 +14,7 @@ module Engine
       end
 
       def full_id
-        "#{hex.id} #{id}"
+        "#{hex.id}-#{tile.name}-#{index}"
       end
 
       def hex
--- a/lib/engine/part/base.rb
+++ b/lib/engine/part/base.rb
@@ -13,6 +13,10 @@ module Engine
         @id ||= "#{tile.id}-#{index}"
       end
 
+      def full_id
+        "#{hex.id} #{id}"
+      end
+
       def hex
         @tile&.hex
       end
--- a/lib/engine/route.rb
+++ b/lib/engine/route.rb
@@ -20,6 +20,7 @@ module Engine
       @subsidy = opts[:subsidy]
       @halts = opts[:halts]
       @abilities = opts[:abilities]
+      @saved_nodes = opts[:nodes] # node.full_id for every node in the route
       @local_length = @game.local_length
 
       @node_chains = {}
@@ -46,6 +47,7 @@ module Engine
       @paths = nil
       @stops = nil
       @subsidy = nil
+      @saved_nodes = nil
       @visited_stops = nil
       @check_connected = nil
       @check_distance = nil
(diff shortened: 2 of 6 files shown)


lib/engine/route.rb
```diff
@@ -380,18 +386,32 @@ def add_single_node_connection(node)
...
           left = (a[:nodes] - middle)[0]
           right = (b[:nodes] - middle)[0]
-          return [a, b, left, right, middle[0]]
+          candidates.append([a, b, left, right, middle[0]])
+        end
+      end
+
+      return [] if candidates.empty?
+
+      return candidates[0] if candidates.length == 1
```
tobymao, line 403: size

tobymao:
i did not do this approach because node ids are not deterministic, i'm afraid if you try this, it's going to cause subtle bugs, we'll have to really think closely about this

dfannius:
What about node IDs is not deterministic?

Upgrading tiles is fine in the general case because it will always pick some chain, and almost always there is only one possibility. In 18Ireland it will never turn out to be an issue (there won't be a meaningful choice that then has to be reflected in the upgraded tile). I don't know about tiles 457-464.

dfannius:
Also note that the only way in which this can change existing behavior is on tiles with more than one way to get from one exit to another (e.g., IR7), which are currently already broken anyway (#5279).

tobymao:
@dfannius the node id is tied to the tile id, but the order in which you place tiles (duplicates of the same one) may not be deterministic, and thus node ids don't map across reloads

dfannius:
In my current code, `Node.full_id` looks something like `I9 IR8-0-1` (hex, tile type, which copy of that tile type, node index on that tile type). What if I just make it `I9 IR8-1` (leaving out the tile copy number)? It's not like it matters which copy of IR8 it is.

lib/engine/part/base.rb
```diff
@@ -13,6 +13,10 @@ def id
         @id ||= "#{tile.id}-#{index}"
       end
 
+      def full_id
+        "#{hex.id}-#{tile.name}-#{index}"
```
tobymao, line 17: hex can be nil, this will crash

PR #7099, opened 2022-02-11
18usa pullman scrapping

ventusignis:
18USA Pullmans:
* Can be scrapped any time an owning corporation can act
* Cannot be bought *from the bank* during buy_trains, but can be bought from the bank before running
* A corporation can only own one pullman, owning more than one forces a discard/scrap
* Do not satisfy the requirement to own a train
* Count towards train limit
1:
image
2:
image
3: Player 2 is acting, so ATSF cannot scrap the pullman
image
4:
image
5: Over train limit with 2 pullmans, may only discard pullmans
image
6: More train limit discard/scrapping
image
7: Can scrap a pullman as a corporate SR action
image
8: 
image


lib/engine/game/g_18_usa/step/buy_pullman.rb
```diff
@@ -0,0 +1,57 @@
...
+  module Game
+    module G18USA
+      module Step
+        class BuyPullman < Engine::Step::Base
+          include Engine::Step::Train
+          include ScrapTrainModule # Why would you scrap a pullman so you can buy a pullman!?
+          def actions(entity)
+            return [] if entity != current_entity
+
+            if @game.depot.upcoming.any? { |t| t.name == 'P' } && can_buy_train?(entity) && @game.pullmans_available? && \
```
ventusignis, line 16: There's no reason for me to use @game.pullman_train? one line below, but not here

lib/engine/game/g_18_usa/step/buy_sell_par_shares.rb
```diff
@@ -11,6 +11,44 @@ class BuySellParShares < G1817::Step::BuySellParShares
...
+
+          def can_scrap_train?(entity)
+            return false unless entity.corporation?
+            return false unless entity.owned_by?(current_entity)
+
+            entity.trains.find { |t| @game.pullman_train?(t) }
+          end
+
+          def scrappable_trains(entity)
+            entity.trains.select { |t| t.name == 'P' }
```
ventusignis, line 36: helper method

assets/app/view/game/round/merger.rb
```diff
@@ -54,6 +54,7 @@ def render
           buttons << render_offer(entity, auctioning_corporation) if actions.include?('assign')
 
           buttons << render_merge(entity) if actions.include?('merge')
+          buttons << h(ScrapTrains, corporation: entity) if actions.include?('scrap_train')
```
tobymao, line 57: why isn't this just done where discard trains is added? game_page.rb

lib/engine/game/base.rb
```diff
@@ -2760,6 +2760,18 @@ def train_limit_header
       def train_power?
         false
       end
+
+      # This is a bit of a hack to get around how 1840 and 18USA have different and incompatible opinions on what the
+      #  option to use the scrap_trains action in a merger round means
+      def use_1840_style_merger_round_scrap_trains?
```
tobymao, line 2766: uhhhhh can we fix the hack?

lib/engine/game/base.rb
```diff
@@ -2760,6 +2760,18 @@ def train_limit_header
...
+        true
+      end
+
+      # If true, instead of doing a relatively big view with a header and corporation context for scrapping trains, only a
+      # button [{scrap}] will be displayed, where {scrap} is taken from the current step's scrap_info(train) method
+      def use_compact_scrap_trains_view
```
tobymao, line 2772: why? can we just unify it?

lib/engine/game/g_18_usa/step/special_buy_train.rb
```diff
@@ -1,12 +1,13 @@
...
 
 require_relative '../../../step/special_buy_train'
-
+require_relative 'scrap_train_module'
 module Engine
   module Game
     module G18USA
       module Step
         class SpecialBuyTrain < Engine::Step::SpecialBuyTrain
+          include ScrapTrainModule
```
tobymao, line 10: why is this includede in every single step? this smells funny

assets/app/view/game/round/merger.rb
crericha: Maybe. It's a bit tricky in that this is always an available (if you have a pullman) optional action, much like "take loan". In my opinion optional actions, should be at the bottom of the displayed buttons to minimize the attention they draw. They are there if you want them, but not demanding your attention.

lib/engine/game/base.rb
crericha: It's definitely a hack that probably shouldn't have been in the code base. I'd prefer to write an issue to have it resolved as neither of us know enough about 1840 to make the change. Maybe @Zwergenpunk has thoughts?

lib/engine/game/g_18_usa/step/special_buy_train.rb
crericha: This should be a non-blocking step, like loan.

lib/engine/game/base.rb
crericha: Matthew is trying to add a button under the company (aka corporate stock round actions). The default is a whole scrap trains section, much like buy trains. I don't think the two are very compatible, but open to ideas.

lib/engine/game/g_18_usa/game.rb
```diff
@@ -630,20 +630,78 @@ def operating_round(round_num)
...
+        def use_1840_style_merger_round_scrap_trains?
+          false
+        end
+
+        def crowded_corps
+          @crowded_corps ||= corporations.select do |c|
+            trains = self.class::OBSOLETE_TRAINS_COUNT_FOR_LIMIT ? c.trains.size : c.trains.count { |t| !t.obsolete }
+            trains > train_limit(c) || c.trains.count { |t| pullman_train?(t) } > 1
+          end
+        end
```
crericha, line 674: Is there a reason this crowded_corps logic can't be common? It looks generic. The last line may be something like super || c.trains.count { |t| pullman_train?(t) } > 1.

lib/engine/game/g_18_usa/game.rb
```diff
@@ -630,20 +630,78 @@ def operating_round(round_num)
...
+          'Scrap Pullman'
+        end
+
+        def use_compact_scrap_trains_view
+          true
+        end
+
+        def use_1840_style_merger_round_scrap_trains?
+          false
+        end
```
crericha, line 667: The base class default should be false and then defined in 1840.

The name could also be like use_operating_round_for_merger_round_train_actions, so its more generic.

lib/engine/game/g_18_usa/game.rb
```diff
@@ -630,20 +630,78 @@ def operating_round(round_num)
...
+          "Scrap Pullman for #{format_currency(pullman_scrap_value)}"
+        end
+
+        def scrap_button_text
+          'Scrap Pullman'
+        end
+
+        def use_compact_scrap_trains_view
+          true
+        end
```
crericha, line 663: These there methods should be in the scrap_train step (the non-blocking one I mentioned).

lib/engine/game/g_18_usa/game.rb
```diff
@@ -630,20 +630,78 @@ def operating_round(round_num)
...
+
+          scrap_train(train)
+        end
+
+        # Do error checking before calling this.
+        def scrap_train(train)
+          @bank.spend(pullman_scrap_value, train.owner)
+          @log << "#{train.owner.name} scraps a pullman for #{format_currency(pullman_scrap_value)}"
+          @depot.reclaim_train(train)
+        end
```
crericha, line 703: These 3 methods should also be looked at to see if they fit better in the scrap train step.

lib/engine/game/g_18_usa/game.rb
```diff
@@ -654,24 +712,29 @@ def next_round!
               reorder_players
               new_operating_round
             when Engine::Round::Operating
+              # The normal export logic which plays nicely with privates exists in first_turn_housekeeping but it does not
+              #  work in the edgecase where there are 0 entities acting in an operating round - this exists to cover this
+              #  edge case. Since there are no corporatiosn, there is nobody who could use a private to save a train from
+              #  rusting prematurely so this is fine.
+              export_train if @round.entities.empty?
```
crericha, line 719: I fixed this too. Its already in master.

My fix was to move the logic to @round.finished? instead of @round.after_process.

crericha: You can remove this one.

lib/engine/game/g_18_usa/step/acquire.rb
```diff
@@ -0,0 +1,75 @@
...
+              'Pass (Offer for Sale)'
+            elsif @auctioning
+              'Pass (Bid)'
+            elsif @buyer && can_take_loan?(@buyer)
+              'Pass (Take Loan)'
+            elsif @buyer && can_payoff?(@buyer)
+              'Pass (On payoff Loan)'
+            elsif @buyer
+              'Pass (Scrap Train)'
+            end
```
crericha, line 31: When you want to append functionality, its best if you call super and add your new logic around it. Try to keep the code DRY, so that rework and bugs are easier to resolve across the code base.

lib/engine/game/base.rb
Zwergenpunk: oh crap :D
in 1840 the merger round is the "buy tram lines". There is a "real" buy and a scrap train step in it.

I am not sure how to resolve that other than chris suggested. I need a real view and chris does not need a real view, just the action.
Maybe the name could be clearer, or we could add a "show_action_view?" to the step. This would put the logic more into the steps and away from the game.

Did that somehow help? ;)

assets/app/view/game/round/merger.rb
ventusignis: I was thinking similarly to crericha, since the pullman scrapping is an optional action that can be done nearly any time like taking a loan I thought it would be best to keep the button near that

lib/engine/game/g_18_usa/step/special_buy_train.rb
ventusignis: There's some logic that looks at the current step for the methods in `ScrapTrainModule`, I can look into if making that logic search for a ScrapTrains step instead of the current step would work

crericha: Yeah, let's rework the view to find the step. For all other games, it's likely the buy train step, so active step == scrap train step for them.

crericha: We'll just have to test those games out to verify.

lib/engine/game/g_18_usa/game.rb
ventusignis: I replaced it with
```

        def crowded_corps
          @crowded_corps ||= super & corporations.select { |c| c.trains.count { |t| pullman_train?(t) } > 1 }
        end

```

lib/engine/game/g_18_usa/step/special_buy_train.rb
ventusignis: Using `        step = @game.round.step_for(@corporation, 'scrap_train')` instead of `@game.round.active_step` worked well as a replacement, I can test 18Mag but I'm not sure how to play some of the other games like 1840

assets/app/view/game/discard_trains.rb
```diff
@@ -7,6 +7,10 @@ module Game
     class DiscardTrains < Snabberb::Component
       include Actionable
 
+      def current_entity_actions
+        @current_entity_actions ||= @game.round.actions_for(@game.round.active_step&.current_entity) || []
+      end
```
crericha, line 12: There is only one method (render). The call can go directly into render.

assets/app/view/game/scrap_trains.rb
```diff
@@ -22,27 +23,32 @@ def render
             alignItems: 'center',
           },
         }
-
-        h(:div, [
-          h(:h3, 'Trains to Scrap'),
-          h(:div, div_props, scrap_trains(scrappable_trains)),
-        ])
+        if @game.use_compact_scrap_trains_view
```
crericha, line 26: This choice may vary from step to step. For example, buy_train may want to show the entire scrap train display. Can you add the method to the step instead?

I also recommend calling it something like `scrap_train_button_only`.

assets/app/view/game/scrap_trains.rb
```diff
@@ -22,27 +23,32 @@ def render
...
               @corporation,
               train: train,
             ))
           end
-
-          [h(:div, train.name),
-           h('div.nowrap', train.owner.name),
-           h('div.right', step.scrap_info(train)),
-           h('button.no_margin', { on: { click: scrap } }, step.scrap_button_text(train))]
+          if @game.use_compact_scrap_trains_view
```
crericha, line 44: Calling the method twice doesn't make sense. Just make a separate method that renders the button and have the full render call that method.

lib/engine/game/base.rb
crericha: Let's rename to what the method actually does "train_actions_always_use_operating_round_view".

lib/engine/game/base.rb
crericha, line 2772: Move to step

lib/engine/game/g_18_usa/game.rb
```diff
@@ -630,20 +630,76 @@ def operating_round(round_num)
...
             G18USA::Step::SpecialBuyTrain,
             G18USA::Step::Assign,
             G18USA::Step::Track,
             G18USA::Step::DenverTrack,
             G18USA::Step::Token,
+            G18USA::Step::BuyPullman,
             G18USA::Step::Route,
             G18USA::Step::Dividend,
-            Engine::Step::DiscardTrain,
+            G18USA::Step::DiscardTrain,
```
crericha, line 644: The OR discard train step will need to be near the top of the list. There is a private that allows a company to buy a train anytime on their turn and that will need to trigger this step.

lib/engine/game/g_18_usa/game.rb
crericha: Looks like I've given two naming ideas. Names are hard. :-)

lib/engine/game/g_18_usa/game.rb
```diff
@@ -654,24 +710,29 @@ def next_round!
...
               @log << "-- #{round_description('Merger and Conversion', @round.round_num)} --"
               G1817::Round::Merger.new(self, [
                 G18USA::Step::ReduceTokens,
-                Engine::Step::DiscardTrain,
+                G18USA::Step::DiscardTrain,
                 G1817::Step::PostConversion,
-                G1817::Step::PostConversionLoans,
-                G1817::Step::Conversion,
+                G18USA::Step::PostConversionLoans,
+                G18USA::Step::Conversion,
```
crericha, line 726: Can the merger round steps be reverted and then add the scrap_train step?

lib/engine/game/g_18_usa/game.rb
```diff
@@ -654,24 +710,29 @@ def next_round!
...
             when G1817::Round::Merger
               @log << "-- #{round_description('Acquisition', @round.round_num)} --"
               G1817::Round::Acquisition.new(self, [
                 G18USA::Step::ReduceTokens,
                 G1817::Step::Bankrupt,
                 G1817::Step::CashCrisis,
-                Engine::Step::DiscardTrain,
-                G1817::Step::Acquire,
+                G18USA::Step::DiscardTrain,
+                G18USA::Step::Acquire,
```
crericha, line 735: Same comment. Can these be reverted and add the scrap_trains step?

lib/engine/game/g_18_usa/step/acquire.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
```
crericha, line 1: Revert

lib/engine/game/g_18_usa/step/buy_pullman.rb
```diff
@@ -0,0 +1,57 @@
...
+module Engine
+  module Game
+    module G18USA
+      module Step
+        class BuyPullman < Engine::Step::Base
+          include Engine::Step::Train
+          def actions(entity)
+            return [] if entity != current_entity
+
+            if @game.depot.upcoming.any? do |t|
```
crericha, line 14: Order by the lowest cost check first and then work up to the highest cost check.

crericha: Also, I think it would read more clearly if you use braces instead of a block for your lambda.

crericha: Is can_buy_train?(entity) equivalent to can buy pullman train? We only want it to return true if so.

lib/engine/game/g_18_usa/step/buy_pullman.rb
```diff
@@ -0,0 +1,57 @@
...
+               end && can_buy_train?(entity) && @game.pullmans_available? && \
+                entity.runnable_trains.none? { |t| @game.pullman_train?(t) }
+              return %w[buy_train pass]
+            end
+
+            []
+          end
+
+          def description
+            'Buy Pullman (Early)'
```
crericha, line 25: I don't think (Early) is needed.

lib/engine/game/g_18_usa/step/buy_pullman.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+require_relative '../../../step/train'
+module Engine
+  module Game
+    module G18USA
+      module Step
+        class BuyPullman < Engine::Step::Base
```
crericha, line 9: Rather than reimplement much of the buy train step, is it possible to use buy train, but limit to just pullmans?

crericha: If its because of the repay loan assuming buy train happens only right before repay loan, then fix that instead of duplicating logic. You can fix it by setting a round variable or something along those lines.

lib/engine/game/g_18_usa/step/buy_sell_par_shares.rb
```diff
@@ -11,6 +11,44 @@ class BuySellParShares < G1817::Step::BuySellParShares
...
+            actions = []
+            if @round.current_actions.none?
+              actions << 'take_loan' if @game.can_take_loan?(entity) && !@corporate_action.is_a?(Action::BuyShares)
+              actions << 'buy_shares' unless @game.redeemable_shares(entity).empty?
+              actions << 'scrap_train' if can_scrap_train?(entity)
+            end
+            actions
+          end
+
+          def can_scrap_train?(entity)
```
crericha, line 28: Maybe use a module, so you don't have to duplicate these methods between here and the scrap_train step?

lib/engine/game/g_18_usa/step/buy_train.rb
```diff
@@ -13,6 +13,9 @@ def should_buy_train?(entity)
 
           def buyable_trains(entity)
             buyable_trains = super
+            # Cannot buy pullmans from the bank/depot? # TODO: Confirm if this is true and if false, remove this line.
+            # https://github.com/tobymao/18xx/issues/7097
+            buyable_trains.reject! { |t| @game.pullman_train?(t) && t.from_depot? }
```
crericha, line 18: I recommend posting the question of bgg

lib/engine/game/g_18_usa/step/conversion.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
```
crericha, line 1: Revert and add scrap_train step.

lib/engine/game/g_18_usa/step/discard_train.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
```
crericha, line 1: Revert and add scrap_train step

lib/engine/game/g_18_usa/step/post_conversion_loans.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
```
crericha, line 1: Revert and add scrap_train step

lib/engine/game/g_18_usa/step/scrap_train.rb
```diff
@@ -0,0 +1,46 @@
...
+          def actions(entity)
+            actions = []
+            actions << 'scrap_train' if entity == current_entity && entity.corporation? &&
+                entity.trains.any? { |t| @game.pullman_train?(t) }
+            actions << 'pass' if blocks?
+            actions
+          end
+
+          def blocks?
+            @round.paid_loans[current_entity] && can_scrap_train?(current_entity)
```
crericha, line 20: Hmm...we don't want both repay loan and scrap trains to block, so that you have to pass both. Can you make this step before repay loans and then rely on it blocking for the player to get a final opportunity before their turn ends?

lib/engine/game/g_18_usa/step/scrap_train.rb
```diff
@@ -0,0 +1,46 @@
...
+require_relative 'scrap_train_module'
+module Engine
+  module Game
+    module G18USA
+      module Step
+        class ScrapTrain < Engine::Step::Base
+          include ScrapTrainModule
+          def actions(entity)
+            actions = []
+            actions << 'scrap_train' if entity == current_entity && entity.corporation? &&
```
crericha, line 13: call can_scrap_train? here instead of duplicating the logic.

lib/engine/game/g_18_usa/step/scrap_train.rb
```diff
@@ -0,0 +1,46 @@
...
+            'Scrap Pullman'
+          end
+
+          def pass_description
+            'Skip Scrap Pullman'
+          end
+
+          def can_scrap_train?(entity)
+            return false unless entity.corporation?
+            return false unless entity.owned_by?(current_entity)
```
crericha, line 33: What is this conditional doing? If it's the corporation's turn, won't this eject early?

crericha: If you remove blocking? and let the loan step do that, you can remove this check as the actions method will gate the calls for you.

lib/engine/game/g_18_usa/step/scrap_train_module.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G18USA
+      module Step
+        module ScrapTrainModule
+          def scrappable_trains(entity)
```
crericha, line 8: The module and step both do this. I recommend putting most of the logic in the module (so it can be used by the buy_sell_par_shares step) and keep the scrap_trains step light.

crericha:
Don't forget that rural junctions do not pay pullmans.

lib/engine/game/g_18_usa/game.rb
ventusignis: These are migrated now

lib/engine/game/g_18_usa/game.rb
ventusignis: These are also put into scrap train module

lib/engine/game/g_18_usa/step/acquire.rb
ventusignis: I just force pushed the latest set of changes. I don't think that for the acquisition round (probably for merger but I need to look at that before I can say it with the say confidence) that reverting this and adding a scrap_train step makes sense.

In the 1817 Acquisition round, it is implemented mainly in the Acquisition step as  a series of possible actions with a few other steps for interrupting steps (bankruptcy, cash crisis, forced train discard, forced token reduction) so I believe that the best way here is to insert the ability to scrap a pullman when a corporation is up to act in the round (as opposed to a player)

assets/app/view/game/discard_trains.rb
ventusignis: I've inlined the method

assets/app/view/game/scrap_trains.rb
ventusignis: Done

assets/app/view/game/scrap_trains.rb
ventusignis: It's now called from the step, I've added the method to the steps in the games that use it instead of the base step

lib/engine/game/base.rb
ventusignis: Done

lib/engine/game/g_18_usa/step/buy_sell_par_shares.rb
ventusignis: Since `can_scrap_train?` is the same between `buy_sell_par_shares` and `scrap_train` I pulled it into the `scrap_train_module` but `process_scrap_train` is slightly different between the two (one calls `scrap_train_by_owner`, the other calls `scrap_train_by_corporation`)

lib/engine/game/g_18_usa/step/scrap_train.rb
ventusignis: I've consolidated this logic into the scrap_train module and I was able to consolidate the `scrap_train_by_<foo>` methods there because <foo>.owned_by?(<foo>) is true if it's a corporaton

lib/engine/game/g_18_usa/step/scrap_train_module.rb
ventusignis: Done in an earlier round of cleanup today

lib/engine/game/g_18_usa/step/buy_pullman.rb
ventusignis: Removed the Early

lib/engine/game/g_18_usa/step/buy_pullman.rb
ventusignis: Resolving the above comment obsoletes this one

lib/engine/game/g_18_usa/step/buy_pullman.rb
ventusignis: This inherits from BuyTrain now and I added a check to 18USA's operating round to get around the loan interest issue

lib/engine/game/g_18_usa/game.rb
ventusignis: I moved it near ScrapTrain

lib/engine/game/g_18_usa/step/scrap_train.rb
ventusignis: Instead I included ScrapTrainModule into 18USA::Step::Loan and removed the ScrapTrain step from the operating round so that the Loan step can be the only blocking step - if the company can do a loan or scrapping action

lib/engine/game/g_18_usa/step/scrap_train.rb
ventusignis: scrap_train.rb is removed in the latest version of the PR

lib/engine/game/g_18_usa/game.rb
ventusignis: I've rebased the branch against master and removed this

assets/app/view/game/discard_trains.rb
```diff
@@ -35,7 +35,9 @@ def render
             h(:div, trains),
           ])
         end
-
+        overflow << h(ScrapTrains, corporation: @current_entity) if @game.round.actions_for(
+            @game.round.active_step&.current_entity
```
crericha, line 39: active_step is assigned to step above. Can you use that instead of triggering active_step again?

assets/app/view/game/scrap_trains.rb
```diff
@@ -6,14 +6,36 @@ module View
...
 
+        if step.scrap_trains_button_only?
+          render_button(scrappable_trains)
+        else
+          render_section(scrappable_trains)
+        end
+      end
+
+      def render_button(scrappable_trains)
+        step = @game.round.step_for(@corporation, 'scrap_train')
```
crericha, line 26: This section is a duplicate of line 53 and on. Can you break out into a method both call?

lib/engine/game/base.rb
```diff
@@ -2775,6 +2775,18 @@ def train_power?
...
       end
+
+      # This is a bit of a hack to get around how 1840 and 18USA have different and incompatible opinions on what the
+      #  option to use the scrap_trains action in a merger round means
+      # If a game overrides this to true, then if the possible actions for the current entity include any of:
+      # * buy_train
+      # * scrap_train
+      # * reassign_train
+      # then the Operating view will be used instead of the Merger round view in a merger round.
+      # See https://github.com/tobymao/18xx/issues/7169
```
crericha, line 2786: I don't know if the history is important. Maybe it could be simplified to something like "Renders the operating round view for train actions, even when in another round type."

assets/app/view/game/scrap_trains.rb
```diff
@@ -6,14 +6,36 @@ module View
...
       def render
-        @corporation = @game.round.active_step.current_entity
-        step = @game.round.active_step
+        @corporation ||= @game.round.active_step.current_entity
+        step = @game.round.step_for(@corporation, 'scrap_train')
 
         scrappable_trains = step.scrappable_trains(@corporation)
         return nil if scrappable_trains.empty?
 
+        if step.scrap_trains_button_only?
```
crericha, line 18: Can you use `step.respond_to?(:scrap_trains_button_only?) && step.scrap_trains_button_only?`, so the other games aren't required to implement the method?

Then you can remove the method from all the other games and not require future games to implement it either.

lib/engine/game/g_18_usa/step/loan.rb
```diff
@@ -1,12 +1,20 @@
...
 
 require_relative '../../g_1817/step/loan'
+require_relative 'scrap_train_module'
 
 module Engine
   module Game
     module G18USA
       module Step
         class Loan < G1817::Step::Loan
+          include ScrapTrainModule
```
crericha, line 11: Why is scrap train needed here?

lib/engine/game/g_1840/step/buy_train.rb
```diff
@@ -47,6 +47,10 @@ def process_scrap_train(action)
             @game.scrap_train(action.train, action.entity)
           end
 
+          def scrap_trains_button_only?
```
crericha, line 50: Remove from here and other games as per the comment in the scrap_trains view file

lib/engine/game/g_18_usa/step/scrap_train_module.rb
```diff
@@ -0,0 +1,63 @@
...
+
+            scrap_train(train)
+          end
+
+          # Do error checking before calling this.
+          def scrap_train(train)
+            @game.bank.spend(pullman_scrap_value, train.owner)
+            @game.log << "#{train.owner.name} scraps a pullman for #{@game.format_currency(pullman_scrap_value)}"
+            @game.depot.reclaim_train(train)
+            @game.reset_crowded_corps
```
crericha, line 44: What's the purpose of this action? In the discard_train step, the reclaim_train action is all that process_discard_train does. Why is this different?

lib/engine/game/g_18_usa/step/scrap_train_module.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+
+          def can_scrap_train?(entity)
+            return false unless entity
```
crericha, line 48: a nice trick here is `return false unless entity&.corporation?` It will handle both cases.

lib/engine/game/g_18_usa/step/scrap_train_module.rb
ventusignis: I did some testing and digging and depot.reclaim_train(_) invokes @game.remove_train(_) which clears the @crowded_corps cache so this method is unnecessary so I'm removing it.
I added it originally so that the "have to discard one of two pullmans" case doesn't prevent the subsequent "still over train limit" scenario from triggering

lib/engine/game/g_18_usa/step/loan.rb
ventusignis: By putting scrap train here, there is only one blocking step in the18USA operating round instead of two

assets/app/view/game/discard_trains.rb
ventusignis: I've updated this block

assets/app/view/game/scrap_trains.rb
ventusignis: I was able to do it

lib/engine/game/base.rb
ventusignis: I trimmed down the comment

lib/engine/game/g_18_usa/step/scrap_train_module.rb
ventusignis: Done

assets/app/view/game/scrap_trains.rb
ventusignis: I've changed this and the related files

lib/engine/game/g_18_usa/step/buy_train.rb
ventusignis: The new version of the rules clarifies this case

crericha (approved):

Issue #7110, opened 2022-02-12
18 Los Angeles 2 checklist


michaeljb:
- [x] map changes
- [x] public company (corporation) changes
- [x] private company changes
  - [x] for 2E, CHE's _total_ discount should be $20, never $40; Alhambra has water and mountain, but plenty of other hexes have multiple waters or multiple mountains, and the discount in those cases is meant to be $20 as well
- [x] new private companies
    - [x] Angeles Public Dump
    - [x] LA Paving
    - [x] Redondo Junction
    - [x] RKO Pictures
- [x] corporation removal
- [x] private company removal
- [x] home station placement
- [x] two player corporation removal
- [x] PER home chosen when parred

michaeljb:
will check the boxes above when PRs are merged, I think this is all that's missing from my local code:

* setting PER's home when parred
* new private company abilities
* manual testing
* validate existing 1846 and 18LA1 games

michaeljb:
Update: will not being to alpha without the new private abilities

michaeljb:
screenshot for my current approach for implementing Redondo Junction (will include this on the PR) ![Screenshot from 2022-03-05 17-27-12](https://user-images.githubusercontent.com/1045173/156904896-8b2a8045-4ae5-4096-8b1d-7741bb02d8ba.png)

michaeljb:
My branch `18LA2-dump` has the necessary fixes, but will have merge conflicts with the CHE changes so I'll wait for #7413 to be merged before opening a new PR (since I've already force-pushed the updates, GitHub does not allow me to simply reopen #7412)

PR #7111, opened 2022-02-12
getting started on 18 Los Angeles 2nd Edition

michaeljb:
* `g_18_los_angeles1/` and `G18LosAngeles1` will now contain code for the 1st
  Edition

* `g_18_los_angeles/` and `G18LosAngeles` will now contain code for the 2nd
  Edition

* `G18LosAngeles` has `GAME_TITLE = '18 Los Angeles 2'` while `G18LosAngeles1`
  has `GAME_TITLE = '18 Los Angeles'`; this means no DB migrations will be needed
  for the 1st Edition to continue to work

* `G18LosAngeles1` inherits from `G18LosAngeles` so that when the 2nd Edition is
  production ready, we can easily combine the two editions back into a single
  game entry, with the first edition being turned into a checkmark option. When
  that happens, the currently commented out lines in the two `meta.rb` files will
  be uncommented.

* as I make the actual changes necessary for the 2nd Edition, code that is
  specific to the 1st Edition will be moved from `g_18_los_angeles/` to
  `g_18_los_angeles1/`; starting with all the code in `g_18_los_angeles/`
  minimizes the diff for this initial change

#7110

### DIFF
--- a/lib/engine/game/g_18_los_angeles/meta.rb
+++ b/lib/engine/game/g_18_los_angeles/meta.rb
@@ -8,7 +8,7 @@ module Engine
       module Meta
         include Game::Meta
 
-        DEV_STAGE = :production
+        DEV_STAGE = :prealpha
         DEPENDS_ON = '1846'
 
         GAME_DESIGNER = 'Anthony Fryer'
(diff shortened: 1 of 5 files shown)


tobymao (approved):

tobymao:
while you're at it, can you make 18zoo maps variants?

PR #7112, opened 2022-02-12
18GB Train buying

andypymont:
- If a corporation does not own a train, does not buy one from another company, and does not have sufficient money to buy the cheapest available new train, it enters emergency money raising
- A five-share corporation in emergency money raising may convert to a ten-share company, dropping three steps instead of two
- After converting during EMR, the corporation must buy a new train if it can afford one
- Corporations in receivership buy the most expensive available train if there is more than one available
- Corporations in receivership which are 5-share corporations will convert if they cannot afford the cheapest train in the depot

### DIFF
--- a/lib/engine/game/g_18_gb/game.rb
+++ b/lib/engine/game/g_18_gb/game.rb
@@ -572,6 +572,16 @@ module Engine
           @_shares[share.id] = share
         end
 
+        def emergency_convert_bundles(corporation)
+          return [] if corporation.trains.any?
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_gb/game.rb
```diff
@@ -572,6 +572,16 @@ def add_new_share(share)
           @_shares[share.id] = share
         end
 
+        def emergency_convert_bundles(corporation)
+          return [] if corporation.trains.any?
```
tobymao, line 576: unless empty?

PR #7138, opened 2022-02-13
[1866] Playtest fixes

magnusfalt:
[1866] Playtest fixes
- Changes to the stock market
- Changes from obsolete trains to hard rusted trains
- Use the node walk skip path to really make sure the operations right work as intended
- Minor plays test fixes

**Changes to base code**
- Engine::Graph - Added code to be able to use the skip_paths variable that the node walk is using.
- Base - Added a method which is called tom the graph to fetch if there is any skip paths that we want the node_walk to exclude


lib/engine/game/g_1866/game.rb
```diff
@@ -866,6 +861,16 @@ def graph_for_entity(entity)
           national_corporation?(entity) ? @national_graph : @graph
         end
 
+        def graph_skip_paths(entity)
+          skip_paths = {}
+          @hexes.each do |hex|
+            add_skip_paths = (national_corporation?(entity) && !hex_within_national_region?(entity, hex)) ||
```
tobymao, line 867: why is national_corporation checked in the hot loop

lib/engine/game/g_1866/game.rb
```diff
@@ -866,6 +861,16 @@ def graph_for_entity(entity)
...
+              (corporation?(entity) && !hex_operating_rights?(entity, hex))
```
tobymao, line 868: same here with corporation

tobymao (approved):

Issue #7160, opened 2022-02-15
[18EU] Some Minor Companies can't do anything in the Operations round.


hachi-18xx:
**What's your Game ID?**
https://18xx.games/game/72624

**How to Reproduce the Bug?**
- No.1 Minor Company can't do anything in the Operations round.
- It looks like there is no home token.
- Similarly, 6, 9 and 11 do not have tokens on the board.
- 1, 6, 9 and 11 were  purchased at a discount on the Initial sale round.

ryandriskel:
Fix has been submitted. Your game will work again once it is deployed.

ryandriskel:
I checked your game and you should be good to resume play!

PR #7161, opened 2022-02-15
18EU Fix 7160 Purchased Minor should place token

ryandriskel:
Fix https://github.com/tobymao/18xx/issues/7160

Minors that went down in price were missing the call to place their home token.

Before
Screen Shot 2022-02-14 at 8 13 29 PM

After
Screen Shot 2022-02-14 at 8 13 20 PM

### DIFF
--- a/lib/engine/game/g_18_eu/step/modified_dutch_auction.rb
+++ b/lib/engine/game/g_18_eu/step/modified_dutch_auction.rb
@@ -189,6 +189,7 @@ module Engine
             bidder.spend(price, @game.bank) if price.positive?
             @log << "#{bidder.name} purchases #{target.name} for #{@game.format_currency(price)}"
 
+            place_initial_token(target)
             reset_auction(bidder, target)
           end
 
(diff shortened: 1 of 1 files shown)


tobymao (approved):

lib/engine/game/g_18_eu/step/modified_dutch_auction.rb
```diff
@@ -258,6 +257,12 @@ def reset_auction(winner, target)
             @auction_triggerer = nil
             next_entity!
           end
+
+          def place_initial_token(minor)
+            hex = @game.hex_by_id(minor.coordinates)
+            city = minor.city.to_i || 0
```
tobymao, line 263: wait what, city.to_i when would that be false? if city is nil, to_i will crash

ryandriskel: That's not new code, just moved, so to be honest I don't remember why it is there. However, it won't crash if nil:

`2.6.3 :001 > a = nil
 => nil 
2.6.3 :002 > a.to_i
 => 0 `

ryandriskel: If anything it makes the || 0 seem redudant?

tobymao: right

tobymao: so let's remove teh ||0

PR #7163, opened 2022-02-15
[18USA] Randomize privates, 1817 trains optional rule, and simple tokens

crericha:


crericha:
Now also includes simple tokens

lib/engine/game/g_18_usa/game.rb
```diff
@@ -119,6 +119,14 @@ class Game < G1817::Game
           },
         ].freeze
 
+        def game_phases
+          phases = self.class::PHASES
+          return phases unless @optional_rules.include?(:seventeen_trains)
+
+          phases = phases.dup
```
tobymao, line 126: there's no reason to dup, reject makes a copy

tobymao (approved):

lib/engine/game/g_18_usa/game.rb
crericha: Fixed.

PR #7174, opened 2022-02-15
[1866] Playtest fixes

magnusfalt:
- Fixed a token check if you sell a national share
- Reframe the tile when you lay them to reflect the frame color of the map
- Stripe a used yellow C tile

**Changes to base code**
- Engine::Tile.reframe! - Added a check if color1 is nil, clear the frame
- Engine::Tile.restripe! - New method to expose the stripe part on the tile

### DIFF
--- a/lib/engine/tile.rb
+++ b/lib/engine/tile.rb
@@ -527,7 +527,11 @@ module Engine
     end
 
     def reframe!(color1, color2 = nil)
-      @frame = Engine::Part::Frame.new(color1, color2)
+      @frame = color1.nil? ? nil : Engine::Part::Frame.new(color1, color2)
+    end
+
+    def restripe!(color)
+      @stripes = color.nil? ? nil : Part::Stripes.new(color)
     end
 
     def available_slot?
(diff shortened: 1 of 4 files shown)


lib/engine/tile.rb
```diff
@@ -527,7 +527,11 @@ def restore_borders(edges = nil)
     end
 
     def reframe!(color1, color2 = nil)
-      @frame = Engine::Part::Frame.new(color1, color2)
+      @frame = color1.nil? ? nil : Engine::Part::Frame.new(color1, color2)
+    end
+
+    def restripe!(color)
+      @stripes = color.nil? ? nil : Part::Stripes.new(color)
```
tobymao, line 534: color ? new(color) : nil

lib/engine/tile.rb
```diff
@@ -527,7 +527,11 @@ def restore_borders(edges = nil)
     end
 
     def reframe!(color1, color2 = nil)
-      @frame = Engine::Part::Frame.new(color1, color2)
+      @frame = color1.nil? ? nil : Engine::Part::Frame.new(color1, color2)
```
tobymao, line 530: same

lib/engine/tile.rb
magnusfalt: Fixed

tobymao (approved):

PR #7215, opened 2022-02-17
fix route node serializing

roseundy:
Fixes server vs hotseat issue when serializing route nodes

Also makes import script more robust with user names containing spaces.


roseundy:
#7223 supersedes this

roseundy:
Nevermind, I forgot this also fixed the script.

PR #7223, opened 2022-02-17
Allow find_pairwise_chain to use saved_nodes even with tile upgrade

### LINKED ISSUE #7205: [1852] Saved Route does not remember how St. Louis Works
https://18xx.games/game/73507?action=384

When the route run last OR is auto-populated, the system gets...creative...with which path through St. Louis was used.

roseundy:
Fixes #7205 

This changes node.full_id -> node.partial_id so that Engine::Route can use @saved_nodes even across tile upgrades.

### DIFF
--- a/lib/engine/action/run_routes.rb
+++ b/lib/engine/action/run_routes.rb
@@ -51,7 +51,7 @@ module Engine
             'subsidy' => route.subsidy,
             'halts' => route.halts,
             'abilities' => route.abilities,
-            'nodes' => route.nodes.map(&:full_id),
+            'nodes' => route.nodes.map(&:partial_id),
           }.select { |_, v| v }
         end
 
--- a/lib/engine/action/run_routes.rb
+++ b/lib/engine/action/run_routes.rb
@@ -51,7 +51,7 @@ module Engine
             'subsidy' => route.subsidy,
             'halts' => route.halts,
             'abilities' => route.abilities,
-            'nodes' => route.nodes.map(&:partial_id),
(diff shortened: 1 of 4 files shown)


dfannius:
The code is fine but I am not a fan of the new name `partial_id`. The point of `full_id` was that it identifies any node on the map uniquely (`id` doesn't distinguish between different instances of the same tile); this is still the case, so I would personally keep the name as is. I won't insist on it though.

tobymao:
> 

i agree with @roseundy, full_id was a strange name, because it wasn't full

dfannius:
OK, do we have a better name for something that uniquely identifies a node on the map than `partial_id`?

roseundy:
How about "route_id"?

dfannius:
`signature`?

tobymao:
signature is fine

roseundy:
I was thinking hash or key, but signature is better.

tobymao:
@roseundy and @dfannius how does this work? i didn't think upgrades preserved city indices

tobymao:
additionally, how will this affect previously serialized formats? will it be ok?

roseundy:
> @roseundy and @dfannius how does this work? i didn't think upgrades preserved city indices

In general, they don't. But there is very limited set at tiles that this would even apply to: tiles that have multiple nodes connected to a given edge. The upgrades to those tiles need to make sure the indices match.

roseundy:
> additionally, how will this affect previously serialized formats? will it be ok?

They will be ignored, so it will be OK. Open games might not benefit from the fix, however.

PR #7242, opened 2022-02-19
[18Rhl] Teleport token by private 4 should not be free (Fixes #7157)

### LINKED ISSUE #7157: [18Rhl] Private No. 4 allows placement of free station token
**What's your Game ID?**

https://18xx.games/game/72882?action=96

Player uses private no. 4 to place tile and then a token and isn't required to pay for the token, even though the text for the private says token should be paid for.
Perhaps there is confusion with Private no. 2 which does allow a free token?

perwestling:


perwestling:
This might break some 18Rhl games as an action now cost money.

PR #7250, opened 2022-02-20
18SJ: Refactoring of Sveabolaget

perwestling:
Simplified Sveabolaget assignment implementation as it uses general
functionality.

Renamed assign class for Sveabolaget, to prepare for version 0.9 of 18SJ.


perwestling:
This should not break any existing games. If it do I suppose I have introduced some bug but I did test this on an existing game.

Issue #7275, opened 2022-02-22
[18NY] v2 update is paying a $30 bonus per token hit for 2H trains as well as diesel.


philcampeau:
The v2 update is supposed to give a $30 bonus to each token hit by a diesel, but currently it appears to be giving the bonus to all trains hitting their company's token.

![Capture](https://user-images.githubusercontent.com/26125362/155232507-3c5bf32c-c685-47e6-bce9-657040be3e28.JPG)
![capture2](https://user-images.githubusercontent.com/26125362/155232509-a16f2c8c-b847-4884-bf4e-fc226ab9e53c.jpg)

we tested right up to the D&H running two trains, and it got 2x$30 bonus added to its run.

philcampeau:
found the issue in line 779 of the game.rb file:

`          if second_edition? && (route.train.name == 'D' || route.train.name == '2H')`

philcampeau:
So not all trains, but 2H and Diesel, instead of only diesel.

crericha:
Oops that was test code.

crericha:
Fixed by #7276

PR #7276, opened 2022-02-22
fixes 2H train receiving Diesel bonus

philcampeau:
fixes bug #7275
2H should not receive $30 per token bonus, only the diesel


tobymao:
remove the parens

crericha (approved):

PR #7281, opened 2022-02-23
18USA: Fix train upgrades so they work on server as well as hotseat

ventusignis:
Also fixes P19

### DIFF
--- a/lib/engine/game/g_18_usa/step/route.rb
+++ b/lib/engine/game/g_18_usa/step/route.rb
@@ -25,8 +25,8 @@ module Engine
             @upgrade_train_assignments[upgrade] = train
             @round.train_upgrade_assignments[train] = [] unless @round.train_upgrade_assignments[train]
             @round.train_upgrade_assignments[train] << upgrade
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_usa/step/route.rb
```diff
@@ -68,6 +68,7 @@ def train_choices(entity, allow_permanents)
             @game.route_trains(entity).reject do |t|
               (!allow_permanents && !t.obsolete_on && !t.rusts_on) || t.variant['name'] == 'P'
```
crericha, line 69: I know this line wasn't changed, but can you use @game.pullman_train?(t) instead of == 'P'?

crericha: What's the purpose of these lines if the result is being ignored on line 71?

ventusignis: Line 71 was added when I was debugging the issue but I forgot to remove that line. I changed 69 to use @game.pullman_train?(t)

crericha (approved):

PR #7283, opened 2022-02-23
[18USA] Private and subsidy bug fixes

crericha:


lib/engine/game/g_18_usa/step/p11_track.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G18USA
+      module P11Track
```
ventusignis, line 6: Does this combine nicely with P7(Track Engineers) and/or P17 (Great Northern Railway)?

crericha: Good idea to test this. I made a small update to allow P7 and P11 to coexist. P17 is activated by being selected and is not affected.

crericha:
> I already did the same thing in the 1866 code, so i removed it from there. But you still need to clear it in the `@round.num_upgraded_track = 0` in the setup method.

Good catch. I fixed it and removed `@round.num_upgraded_track = 0` from 1866. 

In the future, please consider adding generalized functionality that you implement, like this one, to the base classes, so other games can reuse it. Any functionality in the current base classes, beyond 1889/18Chesapeake implementations, is because others have done just that.

PR #7284, opened 2022-02-23
[18NY] $30 bonus per station only applies to D trains.

### LINKED ISSUE #7275: [18NY] v2 update is paying a $30 bonus per token hit for 2H trains as well as diesel.

crericha:
Fixes #7275


crericha:
Already fixed by @philcampeau

PR #7296, opened 2022-02-23
[1840] Added new fixture

Zwergenpunk:
this would not have helped to prevent the issues but maybe it will prevent other problems ;)


crericha:
It was a view only problem, so you'd have to check for the buttons in the view if you want to prevent in the future.

Issue #7194, opened 2022-02-16
18EU Auto Actions for Auction


ryandriskel:
The 18EU auction is absurdly long for async games. Often, players have a good idea of how high/low they wish to bid/buy. We could significantly decrease the time to auction if players could pre-program their limits for each minor.

scottredracecar:
See also #5336 and #4928

Issue #5336, opened 2021-05-14
Auto bid in auctions


rclyne:
In a lot of games there is a single item auctioned at a time.  Sometimes you want to bid up to an amount.  Could a max-bid item be added to the Auto tab where you can add a Max Bid for the item and then the system will automatically raise your bid by the minimum if you are loosing the auction.  I suspect it could not be automated in games like 1822 where multiple items are auctioned simultaneously.

PR #7301, opened 2022-02-24
18EU and Waterfall Auto Auction

ryandriskel:
Closes https://github.com/tobymao/18xx/issues/7194
Partial implementation of https://github.com/tobymao/18xx/issues/5336

This provides functionality for automatic bidding on 18EU and Waterfall style auctions. I added to 18MT as a "safer" testing ground for this functionality, since that title is still in alpha. It should work for all waterfalls, though.

Changes to **form.rb**

I moved the render_checkbox method here as it mirrors other functionality within the form class. This allows it to be used by more than just the auto class.

Changes to **auto.rb**, and addition of **auto_action/base.rb and auto_action/auction_bid.rb**

I wanted to separate my functionality so defined a new class to hold specific auto action views. I propose refactoring the current auto actions into these specific files in a future PR, as it makes functionality much easier to find.

**auto_action/base.rb and auto_action/auction_bid.rb**

I am not confident on if I am using the "needs" appropriately, but it doesn't work without them. Please advise if changes are needed.

---

Screen Shot 2022-02-23 at 6 57 59 PM

Screen Shot 2022-02-23 at 3 43 09 PM

Screen Shot 2022-02-23 at 3 43 20 PM

Screen Shot 2022-02-23 at 3 47 09 PM

Screen Shot 2022-02-23 at 3 47 53 PM

Screen Shot 2022-02-23 at 3 49 02 PM

---

Screen Shot 2022-02-23 at 6 58 08 PM

Screen Shot 2022-02-23 at 7 00 50 PM

Screen Shot 2022-02-23 at 3 51 24 PM

### DIFF
--- a/assets/app/view/game/auto_action/auction_bid.rb
+++ b/assets/app/view/game/auto_action/auction_bid.rb
@@ -23,10 +23,7 @@ module View
 
           form = {}
 
-          children = [h(:h3, name)]
-          children << h(:p, description)
-
-          children << h(:div, [render_entity_selector(form)])
+          children = [h(:h3, name), h(:p, description), h(:div, [render_entity_selector(form)])]
 
           children << h(Corporation, corporation: selected) if selected&.corporation? || selected&.minor?
           children << h(Company, company: selected) if selected&.company?
--- a/assets/app/view/game/auto_action/auction_bid.rb
+++ b/assets/app/view/game/auto_action/auction_bid.rb
@@ -1,0 +1,170 @@
+# frozen_string_literal: true
+
+require 'view/game/auto_action/base'
+
+module View
+  module Game
+    module AutoAction
+      class AuctionBid < Base
+        needs :bid_target, store: true, default: nil
+
+        def name
+          "Auto bid in Auction Round#{' (Enabled)' if @settings}"
+        end
+
+        def description
+          'Automatically bid, buy or pass in the auction round. '\
+            'Bids increase by the minimum increment. '\
+            'Buys only when the current price matches your specified price.'
+        end
+
+        def render
+          return [] unless @game.round.auction?
+
+          form = {}
+
+          children = [h(:h3, name)]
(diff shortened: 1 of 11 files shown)


assets/app/view/game/auto_action/auction_bid.rb
```diff
@@ -0,0 +1,170 @@
...
+            'Bids increase by the minimum increment. '\
+            'Buys only when the current price matches your specified price.'
+        end
+
+        def render
+          return [] unless @game.round.auction?
+
+          form = {}
+
+          children = [h(:h3, name)]
```
tobymao, line 26: you can make this just a static array up to line 29 because it's all the same

childen = [...]

lib/engine/action/program_auction_bid.rb
```diff
@@ -0,0 +1,59 @@
...
+require_relative 'program_enable'
+
+module Engine
+  module Action
+    class ProgramAuctionBid < ProgramEnable
+      attr_reader :bid_target, :enable_maximum_bid, :maximum_bid, :enable_buy_price, :buy_price, :auto_pass_after
+
+      def initialize(entity, bid_target:, maximum_bid:, buy_price:, enable_maximum_bid: false,
+                     enable_buy_price: false, auto_pass_after: false)
+        super(entity)
```
tobymao, line 13: why are you calling super

lib/engine/action/program_auction_bid.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require_relative 'base'
```
tobymao, line 3: why are you including base

lib/engine/action/program_auction_bid.rb
ryandriskel: All 6 other "program_..." action classes already defined do it.

PR #7341, opened 2022-02-27
[18NY] Always label special hexes

### LINKED ISSUE #7085: 18NY add simple labels for special locations
Add simple labels for the board for all special locations, and have them punch through yellow and green tiles that lack the special placement restriction symbols. 

This would be inline with the implementation of this feature in 1870.

thepackrat:
This change allows special hexes in 18NY to always be labelled, with the labels present on plain tiles laid in yellow or green on the labelled hexes to make it clearer where the green/brown/grey special tiles will go. 

Fixes #7085


lib/engine/game/g_18_ny/game.rb
```diff
@@ -659,7 +667,25 @@ def home_token_locations(corporation)
...
+              new_tile.label = old_tile.label.to_s
+            end
+
+            # remove the label when we remove a temporaily labelled tile.
+            if PLAIN_SYMBOL_HEXES.include?(old_tile.color) &&
+               PLAIN_SYMBOL_HEXES[old_tile.color].include?(new_tile.hex.id)
+
+              old_tile.label = nil
+            end
+          end
```
crericha, line 687: If you are trying to write something generic for all titles, 674 - 687 make sense. In 18NY (and probably most games), there aren't temporarily labled tiles. Something similar to the example below would work fine since the validation was already done elsewhere (upgrades_to_correct_label?).

`new_tile.label = old_tile.label.to_s if old_tile.label && !new_tile.label`

thepackrat: Yes, those interlocks work, I was just being super conservative because of how pernicious extra labelled tiles are. Will upload the simplified version you suggested. I may look at a generic thing at some point.

thepackrat: Done, thanks.

crericha: My original comment was to remove 674 - 687 and replace with the one line. Taking a look at the code, it looks like much of that is still there. Is it needed and I'm just missing something?

thepackrat: I've removed the interlocks for adding a label, but you need to know which labels need to be removed from plain tiles on tile upgrade, which is why the second  block of logic is still there.

crericha: Ahh, yes. That's what I was missing.

If you need to keep `if old_tile.label`, then you don't need to check that condition again on the new line. Otherwise, looks good.

thepackrat: Done, thanks.

crericha (approved):

PR #7395, opened 2022-03-05
18GB Alpha changes

andypymont:
- Update the Info page's rules link to point to 2nd edition rules rather than 1st edition (consistent with wiki)
- The MC private should cost £50, not £60 (2nd edition rules change)


andypymont:
Because the MC private cost is a reduction rather than an increase, I believe this hopefully won't break any existing games. The private is only used in high player-count games and I tested one of them locally and this change did not break the game (it just means someone bids £60 when they could have bid £50).

PR #7398, opened 2022-03-05
1822MX: Fix P5, hide cube tiles

outsidePasser:


lib/engine/game/g_1822_mx/game.rb
```diff
@@ -53,6 +53,9 @@ class Game < G1822::Game
         PRIVATE_PHASE_REVENUE = %w[].freeze # Stub for 1822 specific code
         P7_REVENUE = [0, 0, 0, 20, 20, 40, 40, 60].freeze
 
+        # Don't run 1822 specific code for the LCDR
+        COMPANY_LCDR = ''
```
michaeljb, line 57: We more commonly use `nil` to indicate an unset value
```suggestion
        COMPANY_LCDR = nil
```

michaeljb (approved):

PR #7412, opened 2022-03-08
18LA2: private company Angeles Public Dump

michaeljb:
the Dump places a -$20 station token which blocks everyone--including the owning
corporation--and reduces the city's revenue by $20 for everyone's routes

Westminster is the only $10 target for the Dump, and if the Dump is placed there
the total revenue is $0 instead of -$10. Westminster is also a target for the
Steamship (adding $40), and if both are placed there, then the Dump's full
penalty applies for a net revenue of $30


michaeljb:
just found a bug with this interacting with the RJ private, will reopen when I can fix it

PR #7413, opened 2022-03-08
18LA2: changes for private company Chino Hills Excavation

michaeljb:
1E: $20 discount to all mountain costs, can apply to more than mountain in a
single tile lay, e.g., two separate borders, or a border and the main tile
upgrade cost

2E: $20 discount for all terrain costs, but limited to one discount per tile lay

The existing `tile_discount` ability has no configurable mechanism for being
limited to a single use, so a tile's `borders` and `upgrades` are checked and raise
the flag if any terrain is found, in which case the CHE discount can be used.

In entities files, treat 1E/2E versions as completely different companies, instead of one
where the 1E definition overrides the 2E definition, so that the different IDs
can be used in the game logic to apply the different "ability" separately

#7110

### DIFF
--- a/lib/engine/game/g_18_los_angeles/entities.rb
+++ b/lib/engine/game/g_18_los_angeles/entities.rb
@@ -69,28 +69,6 @@ module Engine
             color: nil,
           },
           {
-            name: 'Chino Hills Excavation',
-            value: 50,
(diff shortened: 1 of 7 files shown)


michaeljb:
closing for test failures

PR #7418, opened 2022-03-09
Rolling Stock: Acquisition and Closing Rounds

roseundy:
Add "unordered" round support, including new UI elements UI::Round::Unordered, UI::Propose
Add Propose and Response actions

Common file edits:
- UI::Entities - call @game.player_entities instead of @game.players
- UI::PassButton - call @game.pass_entity instead of @game.current_entity
- UI::Map - no map if layout is :none
- UI::GamePage - Omit Map and Tiles tabs if layout is :none, call Round::Unordered if needed, set current_entity_actions based on @user/hotseat if Unordered round
- Engine::Game::Base - add pass_entity and player_entity methods
- Engine::Round::Base - add unordered? method


assets/app/view/game/propose.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+require 'view/game/actionable'
+
+module View
+  module Game
+    class Propose < Snabberb::Component
```
tobymao, line 7: offer?

lib/engine/action/propose.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Action
+    class Propose < Base
```
tobymao, line 7: offer?

lib/engine/action/propose.rb
```diff
@@ -0,0 +1,34 @@
...
+      attr_reader :entity, :corporation, :company, :price
+
+      def initialize(entity, corporation:, company:, price:)
```
tobymao, line 10: do you need corporation? isn't entity the corporation?

lib/engine/game/g_rolling_stock/entities.rb
```diff
@@ -440,7 +440,7 @@ module Entities
             text_color: 'white',
             tokens: [],
             shares: [10, 10, 10, 10, 10, 10],
-            float_percent: 10,
+            float_percent: 50,
```
tobymao, line 443: what's up with float percent?

lib/engine/action/propose.rb
roseundy: No, entity is the player.

assets/app/view/game/propose.rb
roseundy: OK

lib/engine/action/propose.rb
roseundy: OK

lib/engine/game/g_rolling_stock/entities.rb
roseundy: It's the number that causes it to float correctly because the stock in RS Stars doesn't add up to 100%

PR #7421, opened 2022-03-09
Auto-router 1828 coalfield fix

### LINKED ISSUE #6365: [1828.games] Autoroute still creates routes ending in Coalfields without Coal Marker
https://github.com/tobymao/18xx/issues/6327 is not solved.

> Any Game, e.g. 55044, 54684, 56936. Chrome, Android, Windows 10 21H1x64 / Server 2019
> 
> Part 1: It's half bug, half enchantment:
> 
> The game only shows an error when you begin/end a route in the coalfields when you submit a route not during defining the route. Only after submitting it tells you you can't go through the Coalfields and doesn't accept the route.

=> This part is solved.

> Part 2: It's a bug:
> 
> Autoroute also creates such illegal routes.

=> This part is not solved:
The Interface now doesn't accept coalfields without coal marker - but autorouting still creates routes beginning/ending in coalfields anyway - so the problem is still not fixed. (Worse; you now cannot unclick the coalfield "directly" after using autoroute.)
Coalfields

JoeB989:
Fixes #6365


lib/engine/game/g_1828/game.rb
```diff
@@ -1513,6 +1513,18 @@ def system_market_price(corporations)
...
+        def check_other(route)
+          if route_includes_coalfields?(route.routes) && !@game.coal_marker?(current_entity)
+            raise GameError, 'Cannot run to Virginia Coalfields without a Coal Marker'
+          end
+
+          super
+        end
+
+        def route_includes_coalfields?(routes)
+          routes.flat_map(&:connection_hexes).flatten.include?(VA_COALFIELDS_HEX)
```
tobymao, line 1526: it's gonna be more efficient probably if you just something like

routes.any? do | route|
  route.connection_hexes.include?(VA...)

JoeB989: Ok.  I was also thinking I should only test the one route - I'll change that too.

JoeB989: connection_hexes contains things like this
   0: (2) ['G23', 'G21']
so I think the flatten is needed

PR #7423, opened 2022-03-09
18MEX: Handle NdM 5% corner case (fixes #7360)

### LINKED ISSUE #7360: [18MEX] Cannot buy a 5% share of NdM while at certificate limit
Game ID: hs_uzkanjlr_74628
Game Data: https://gist.github.com/mlatuprosa/9e85dfc3c2ec0d732386677dd02faeb0

Reproduction: have a 5% share of NdM in the open market and have a player at certificate limit with 55% held of NdM. (I am not sure whether 55% held is specifically relevant here.) In a stock round turn where that player (Latro in the linked game) has not yet done anything, click on NdM. 
Expected behavior: being given the option to buy the 5% share, per 3.3(b)
Actual behavior: all options given are sales. The option to buy a 5% share is restored upon falling below certificate limit.
image
Browser: Firefox 97.0.1

perwestling:
When at cert limit, you are allowed to by yet another 5% share in NdM
as they do not count towards cert limit. But you are not allowed to
buy swap a 5% for a 10%.

### DIFF
--- a/lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
@@ -9,15 +9,24 @@ module Engine
       module Step
         class BuySellParShares < Engine::Step::BuySellParShares
           def can_buy?(entity, bundle)
-            super && !attempt_ndm_action_on_unavailable?(bundle)
+            return false if attempt_ndm_action_on_unavailable?(bundle)
+
+            super
           end
 
           def can_sell?(entity, bundle)
-            super && !attempt_ndm_action_on_unavailable?(bundle)
+            return false if attempt_ndm_action_on_unavailable?(bundle)
+
+            super
           end
 
-          def can_gain?(entity, bundle)
-            super && !attempt_ndm_action_on_unavailable?(bundle)
+          def can_gain?(entity, bundle, exchange: false)
+            return false if attempt_ndm_action_on_unavailable?(bundle)
+            return if !bundle || !entity
+
+            corporation = bundle.corporation
+
+            corporation.holding_ok?(entity, bundle.percent) && (exchange || room_to_gain?(entity, bundle, exchange))
           end
 
           include SwapBuySell
@@ -35,6 +44,17 @@ module Engine
 
           def attempt_ndm_action_on_unavailable?(bundle)
             bundle.corporation == @game.ndm && @game.phase.status.include?('ndm_unavailable')
+          end
+
+          # Need to allow buying 5% shares in NdM even if at cert limit as these shares are not
+          # counted towards cert limit (but they still count for 60% corporation limit).
+          def room_to_gain?(entity, bundle, exchange)
+            return true if bundle.corporation == @game.ndm &&
+                           @game.num_certs(entity) == @game.cert_limit &&
+                           bundle.percent == 5 &&
+                           !exchange
+
+            @game.num_certs(entity) < @game.cert_limit
           end
         end
       end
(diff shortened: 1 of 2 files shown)


michaeljb (approved):
Approved, with a disclaimer: I haven't played 18MEX and don't have time to test out the fix right now, but the code changes at least _look_ reasonable to me based on the description of #7360 🙂 

Also, this will require pins, right? It changes the actions available to players holding a 5% NdM share, therefore there are potentially games out there where a player might have been auto-skipped, and now they would have the option to buy a share?

lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
```diff
@@ -9,15 +9,24 @@ module G18MEX
...
+            return false if attempt_ndm_action_on_unavailable?(bundle)
+
+            super
           end
 
-          def can_gain?(entity, bundle)
-            super && !attempt_ndm_action_on_unavailable?(bundle)
+          def can_gain?(entity, bundle, exchange: false)
+            return false if attempt_ndm_action_on_unavailable?(bundle)
+            return if !bundle || !entity
```
dfannius, line 25: The `!bundle` check should go first, right?

lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
```diff
@@ -36,6 +45,17 @@ def can_buy_any?(entity)
...
+
+          # Need to allow buying 5% shares in NdM even if at cert limit as these shares are not
+          # counted towards cert limit (but they still count for 60% corporation limit).
+          def room_to_gain?(entity, bundle, exchange)
+            return true if bundle.corporation == @game.ndm &&
+                           @game.num_certs(entity) == @game.cert_limit &&
+                           bundle.percent == 5 &&
+                           !exchange
+
+            @game.num_certs(entity) < @game.cert_limit
```
dfannius, line 57: I personally think it's clearer to do the normal check (`@game.num_certs(entity) < @game.cert_limit`) first and then deal with the remaining exception, but it's not that important.

lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
michaeljb: AFAICT, everything after the `attempt_ndm_action_on_unavailable?` line is based on [`Engine::Step::ShareBuying#can_gain?`](https://github.com/tobymao/18xx/blob/master/lib/engine/step/share_buying.rb#L36-L43), with a change to the last line.

So having the `!bundle` check second is I think is basically the same as making it `!attempt_ndm_action_on_unavailable?(bundle) && super`

But on second glance, it does seem like if `bundle` is `nil`, putting the `!bundle` check first would catch it, but with the `attempt_ndm_action_on_unavailable?(bundle)` call first, it is prone to a "nil pointer" error.

perwestling:
> Approved, with a disclaimer: I haven't played 18MEX and don't have time to test out the fix right now, but the code changes at least _look_ reasonable to me based on the description of #7360 🙂
> 
> Also, this will require pins, right? It changes the actions available to players holding a 5% NdM share, therefore there are potentially games out there where a player might have been auto-skipped, and now they would have the option to buy a share?

I did try the included game and it seemed to work as it should, handling the corner case, and also not allowing swap of 5% for a 10% in market. 

It might break games, yes, so better do pin.

lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
perwestling: Changed to make it bundle, entity, ndm-unavailable

It is correct that new code is based on super method. Needed to copy it to change the cert limit check. And I then removed one check which I think did not apply to 18MEX.

lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
perwestling: Agreed, have updated now.

dfannius (approved):

PR #7409, opened 2022-03-07
Added 1877: Stockholm Tramways map, tiles, market and corporations.

lbrande:
Slightly modified the weights for centered location names.

PR #7429, opened 2022-03-10
Fixed incorrect region_weights for labels introduced in 6af9cd7cb5fc3…

lbrande:
…dc464c34dfa7c63e78b43e253dd. I've validated the change on all games using `:pointy` (both tiles and hexes) and specifically checked 1866 which was messed up by the previous change (see https://github.com/tobymao/18xx/pull/7409#discussion_r823423885).


magnusstjarnhild:
Great Love!

PR #7442, opened 2022-03-11
Implemented 1877: Stockholm Tramways train buying rules.

lbrande:


assets/app/view/game/buy_trains.rb
```diff
@@ -19,7 +19,7 @@ def render_president_contributions
         player = @corporation.owner
         children = []
 
-        verb = @must_buy_train ? 'must' : 'may'
+        verb = @must_buy_train || @step.must_buy_train_if_able_to?(@corporation) ? 'must' : 'may'
```
tobymao, line 22: why is this needed

assets/app/view/game/emergency_money.rb
```diff
@@ -14,13 +14,15 @@ def render_emergency_money_raising(player)
             verticalAlign: 'top',
           },
         }
-        player.shares_by_corporation.each do |corporation, shares|
-          next if shares.empty? || @game.sellable_bundles(player, corporation).empty?
+        if @game.round.actions_for(entity).include?('sell_shares')
```
tobymao, line 17: why is this needed?

lbrande: 1877 Stockholm Tramways does allow the president to contribute but not by selling shares. Without this the sell shares actions were visible (but not possible to perform because they are not legal actions).

assets/app/view/game/buy_trains.rb
lbrande: 1877: Stockholm Tramways only requires corporations to buy a train if it's possible to buy the cheapest train from the depot (using president's contribution). Thus `@must_buy_train` may be `false` while I still want to tell the player how much he "must" contribute to buy the cheapest train from the depot.

However, I realize now that the current code is wrong on one point, if a corporation decides not to buy a train when it has none, the president still has to contribute as much as he can, something which is not reflected in the code or in the text prompt, i.e. this code has to be fixed either way before merging. What I need instead is to make an option for this function to output a prompt that tells the player he must either buy a train or contribute all his cash to the company.

lbrande:
I've now fixed it so that the UI reflects the fact that a train must be bought or the president has to contribute all his cash (and that will now happen also). As such the pull request is ready to be merged.

tobymao (approved):

Issue #7453, opened 2022-03-11
[18NY] Minor uses the West Point Foundry power without it closing. 


akh911007:
In this game we noticed that Minor 6 was able to buy a 4H train for 100 (half price) without actually using the power or closing the private. If he clicks the private to directly use its power, the private closes correctly. 

https://gist.github.com/akh911007/f494e5428c9dc0975291acd8d5d20d12

We undid the turn and used the power to continue the game but I wanted to get a bug ticket put in.

crericha:
Broken by another game.

PR #7454, opened 2022-03-11
Revert "18JP-T: Allow corporation to have a train discount ability"

crericha:
This reverts commit e1320ef8c7cf3c9a36cf7e6da0e47a2642e894d8.

The commit broke train discounts in two ways:
1. It made them automatic. This should not be the case for discounts that have a "count" of how many times it can be used
2. It did not call use! on the ability to, so that "count" was updated.


crericha:
Fixes #7453

tobymao:
@vandamm

PR #7462, opened 2022-03-12
18MEX: Allow buying from IPO for "yellow" corps (fixes #7446)

### LINKED ISSUE #7446: [18MEX] Unable to buy yellow shares at cert limit
https://18xx.games/game/76345

Unable to buy yellow certs after hitting the cert limit.

perwestling:
When a corporation is in "yellow" on stock market, it is allowed
to be bought regardless of cert limit, even if buying from IPO.

I do not know if this will break any game but it potentially might make some player having a buy action that they have not had before.

### DIFF
--- a/lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
@@ -19,8 +19,7 @@ module Engine
           def can_gain?(entity, bundle, exchange: false)
             return if !bundle || !entity || attempt_ndm_action_on_unavailable?(bundle)
 
-            corporation = bundle.corporation
-            corporation.holding_ok?(entity, bundle.percent) && (exchange || room_to_gain?(entity, bundle, corporation))
+            bundle.corporation.holding_ok?(entity, bundle.percent) && (exchange || room_to_gain?(entity, bundle))
           end
 
           include SwapBuySell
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
```diff
@@ -40,13 +40,14 @@ def attempt_ndm_action_on_unavailable?(bundle)
             bundle.corporation == @game.ndm && @game.phase.status.include?('ndm_unavailable')
           end
 
-          def room_to_gain?(entity, bundle)
+          def room_to_gain?(entity, bundle, corporation)
```
dfannius, line 43: Since `corporation` is always equal to `bundle.corporation` I wouldn't add a new argument that adds no additional information to the function call (having it makes it seem like it's possible for `corporation` to not be deducible from `bundle`). It's not like dereferencing `bundle.corporation` another time or two will have a performance impact.

perwestling: @dfannius Have now refactored and removed that parameter.

lib/engine/game/g_18_mex/step/buy_sell_par_shares.rb
```diff
@@ -40,14 +39,17 @@ def attempt_ndm_action_on_unavailable?(bundle)
             bundle.corporation == @game.ndm && @game.phase.status.include?('ndm_unavailable')
           end
 
-          def room_to_gain?(entity, bundle, corporation)
+          def room_to_gain?(entity, bundle)
             return true if @game.num_certs(entity) < @game.cert_limit
-            return true if corporation.share_price && !corporation.share_price.counts_for_limit
+
+            # There is room ff this is a corporation in "yellow" zone in the stock market
```
dfannius, line 45: Since I'm already here, "ff" is a typo. Otherwise looks great.

dfannius (approved):

PR #7464, opened 2022-03-12
18JP-T: Make corporation train discount work only on passive abilities

vandamm:


assets/app/view/game/buy_trains.rb
```diff
@@ -227,6 +227,13 @@ def from_depot(depot_trains, corporation)
                   entity = @selected_company
                 end
               end
+            else
+              # Handle a corporation having train discount ability
+              @game.abilities(@corporation, :train_discount, time: @step.ability_timing) do |ability|
+                next if ability.count&.positive?
```
crericha, line 233: Looking at the ability base class, count will be nil when not used, so `next if ability.count` should be used.

crericha: Good to go after that fix.

crericha (approved):

PR #7489, opened 2022-03-18
fix: allow train buys from other players in 1871

kelsin:


tobymao (approved):

kelsin:
Breaks all of our specs of course. That means I think games will break on the site right?

kelsin:
Game used to autopass when now it needs to show a train buy step etc

PR #7494, opened 2022-03-18
18JP-T: Fix special tile upgrades

vandamm:
This will certainly break some games in progress as it changes upgrade paths.


lib/engine/game/g_18_jpt/game.rb
```diff
@@ -269,13 +267,14 @@ def tile_lays(entity)
...
-            return to.name == 'J14' if %w[E77 F94].include?(from.hex.name)
-            return to.name == 'J15' if from.hex.name == 'J88'
-          end
+          # Allow merging of two separate cities into one with two slots
+          return to.name == '611' if from.hex.coordinates == 'F92' && from.color == :green
 
-          return false if RESERVED_TILES.include?(to.name)
+          # KU hexes have default upgrades in all colors but brown so check without label for simplicity
+          if %w[E77 F94].include?(from.hex.name) && from.color != :brown
+            from = from.dup
```
tobymao, line 275: what's the point of duping here?

vandamm: @tobymao these hexes have custom upgrades only from brown so there are two ways:
1) describe every possible legal upgrade here (eg write 6, 57 for white, 14,15 for yellow etc)
2) check for upgrade eligibility ignoring label on this hex

I chose second option. And as I don't know if `from` is a real tile that is on the map, or its duplicate, I dup it to avoid my label removal persisting to the map.

tobymao: i don't think it's doing what you're doing, dup is a shallow copy of from, if you do labels.clear, that's probably the same reference to the underlying list, are you user that this code is doing what you intend?

vandamm: Well, it seems to work as I intend it to, but I think I'll do more tests of you're concerned

vandamm: @tobymao done this in other way

tobymao (approved):

PR #7499, opened 2022-03-19
18Rhl: Fix Trajekt handling in Rhine Metropoles (fix #7834)

perwestling:
Redesigned the three green Rhine Metropole hexes to be 2+1/1+2
cities in appearance, but work as one city with 3 slots in
practice. This means that a green such hex that has one or
more slots untokened are passable by everyone. And that such
green hex that are fully tokened are passable for the 3
corporations that have a token in any slot on the hex.

To accomplish this there is a special "make neutral/blocked"
procedure at suitable times during the operation round.
I.e. at start/end of OR, and then any of these 3 hexes
are tokened/upgraded.

### DIFF
--- a/lib/engine/game/g_18_rhl/entities.rb
+++ b/lib/engine/game/g_18_rhl/entities.rb
@@ -237,7 +237,7 @@ module Engine
               {
                 type: 'teleport',
                 owner_type: 'player',
-                tiles: %w[921 922 923 924 925 926],
+                tiles: %w[921X 922X 923X 924X 925X 926X],
                 hexes: %w[D9 F9 I10],
                 count: 1,
                 when: %w[owning_player_or_turn],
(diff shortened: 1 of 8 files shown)


perwestling:
I believe this will break all 18Rhl games. There are currently 10 ongoing games, in OR 4 or later. So these could be pinned.
(It can be good to let them complete in case they discover any other bugs. Otherwise there have not been any changes to the code in a couple of weeks so it is pretty stable.)

This fix is probably/hopefully the last fix before title can be set in beta; but this should be tested over a period to see that it does work. I might have missed to test some case.

And I would like to add a fixture from alpha tests before I put this in beta.

perwestling:
Ready to review.

Still breaks existing 18Rhl games.

perwestling:
I noticed some problems with 8 trains that I need to investigate.

perwestling:
I will get back with a new pull request later.

PR #7500, opened 2022-03-19
fix: 1871 exchange shares blocking buyable treasury shares

kelsin:
This allows games to have unbuyable treasury shares in a corporation without blocking buyable treasury shares from showing up properly. Later on in the process we first call `can_buy` on these shares anyway so filtering for buyable ones when getting the list should be 100% the same except in games like 1871 that has reserved shares sitting around.

I also reorder the "but for other player" buttons to match the order of the normal buttons and fix the same bug as above in a 1871 log line.

Also relabel the C tile to CX as intended and made sure the logic works.

### DIFF
--- a/assets/app/view/game/buy_sell_shares.rb
+++ b/assets/app/view/game/buy_sell_shares.rb
@@ -16,13 +16,13 @@ module View
         @step = @game.round.active_step
         @current_entity = @step.current_entity
 
-        @ipo_shares = @corporation.ipo_shares.group_by(&:percent).values
+        @ipo_shares = @corporation.ipo_shares.select(&:buyable).group_by(&:percent).values
           .map(&:first).sort_by(&:percent).reverse
 
         @treasury_shares = if @corporation.ipo_is_treasury?
                              []
                            else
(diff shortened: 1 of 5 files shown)


assets/app/view/game/buy_sell_shares.rb
```diff
@@ -16,13 +16,13 @@ def render
         @step = @game.round.active_step
         @current_entity = @step.current_entity
 
-        @ipo_shares = @corporation.ipo_shares.group_by(&:percent).values
+        @ipo_shares = @corporation.ipo_shares.select(&:buyable).group_by(&:percent).values
```
tobymao, line 19: can you check that this doesn't break other games? shared view code is very dangerous these days

kelsin: One of the first things that the farther down loops do is check if each share is buyable with one of the `can_buy` methods. I can do further testing if you want of course, but I think we're good here. The main thing it does is that if there are unbuyable shares on the top of the stack with buyable shares beneath we didn't show the button. I don't know of a game that relies on that (actually checking the `buyable` field, not just buyable due to limits etc).

PR #7504, opened 2022-03-19
18NewEngland: keep minors first when resorting (fixes #7502)

### LINKED ISSUE #7502: 18NewEngland: re-sorting operating order after force-selling shares doesn't keep minors first
**What's your Game ID?**
https://18xx.games/game/77411

Game data:
https://gist.github.com/jtamplin/d770d704140423b8ded4e9a4faa6b31c

**How to Reproduce the Bug?**

After Peter sold shares to fund buying a 4 train, the B&A showed operating before the GR, which shouldn't be the case.  We went back and did things differently so the order was correct.

**Which Browser / Platform?**
Chrome 99.0.4844.74
Windows laptop

dfannius:
Games that were already affected by this will need to be pinned.

### DIFF
--- a/lib/engine/game/g_18_new_england/game.rb
+++ b/lib/engine/game/g_18_new_england/game.rb
@@ -312,7 +312,7 @@ module Engine
         end
 
         def operating_round(round_num)
-          Engine::Round::Operating.new(self, [
+          G18NewEngland::Round::Operating.new(self, [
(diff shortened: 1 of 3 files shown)


jtamplin (approved):
LGTM

roseundy (commented):
Yes, this looks good.

PR #7543, opened 2022-03-27
Changed how upgrading track and ensuring max exits works

outsidePasser:
Games now define which hex types (cities and/or regular track) require upgrade with max exits.

Needed because 1822MX requires max exits for upgrades of cities, but not for regular track (adding a builder cube doesn't use the maximum numbers of exits).


lib/engine/step/tracker.rb
```diff
@@ -338,14 +338,24 @@ def potential_tiles(entity, hex)
...
+        tiles = potential_tiles(entity, hex).map do |tile|
           tile.rotate!(0) # reset tile to no rotation since calculations are absolute
           tile.legal_rotations = legal_tile_rotations(entity, hex, tile)
           next if tile.legal_rotations.empty?
 
           tile.rotate! # rotate it to the first legal rotation
           tile
         end.compact
+
+        if (hex.tile.cities.any? && @game.class::TILE_UPGRADES_MUST_USE_MAX_EXITS.include?(:cities)) ||
```
tobymao, line 350: !cities.empty?

lib/engine/step/tracker.rb
```diff
@@ -338,14 +338,24 @@ def potential_tiles(entity, hex)
...
+          (hex.tile.cities.empty? && @game.class::TILE_UPGRADES_MUST_USE_MAX_EXITS.include?(:track))
+          tiles.group_by(&:color).values.flat_map do |group|
```
tobymao, line 352: you don't need values, you can just do flat_map do | _, group|

lib/engine/step/tracker.rb
```diff
@@ -338,14 +338,24 @@ def potential_tiles(entity, hex)
...
+            max_edges = group.map { |t| t.edges.length }.max
```
tobymao, line 353: edges.size

lib/engine/step/tracker.rb
outsidePasser: Changed

lib/engine/step/tracker.rb
outsidePasser: Changed

lib/engine/step/tracker.rb
outsidePasser: Changed

PR #7554, opened 2022-03-30
Allow minors to use ports to connect to majors for acquisition

outsidePasser:


lib/engine/game/g_1822_mx/step/minor_acquisition.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require_relative '../../g_1822/step/minor_acquisition'
+
+module Engine
+  module Game
+    module G1822MX
+      module Step
+        class MinorAcquisition < Engine::Game::G1822::Step::MinorAcquisition
+          def connected_to_port?(entity)
```
tobymao, line 10: when is this called? maybe this should be an auto action

outsidePasser: My apologies for resolving - I had a response and I must have accidently removed it.  

connected_to_port? is called from entity_connects? inside the MinorAcquisition step.  entity_connects? is called for each minor that could possibly be acquired to see if it is connected with the active major.  Because of the connectivity using the ports in the Gulf of California, we also check to make sure both the major and the minor are connected to one of those ports.

tobymao: yea, you need to use an auto action probably, it's too expensive to call graph like this

outsidePasser: 1822MX does inherit from 1822, which does these calculations inside an auto action, I believe.  From minor_aquisition.rb:

          def auto_actions(entity)
            return [Engine::Action::Pass.new(entity)] if mergeable(entity).empty?
          end
          
mergable is what calls entity_connects?

The part that concerns me right now is that because I wasn't doing a graph.clear_graph_for(entity) in connected_to_port?, the cached values could be incorrect - but adding a graph.clear_graph_for(entity) would definitely affect performance, and the way this is written now it would be recalculating graphs way longer than it should.  

Would it be ok if the auto action recalculated the graphs for each major/minor once when it enters MinorAcquisition?

tobymao: i don't have enough context into what you're doing to understand, @crericha do either of you know ?

crericha: I don't have the context either. I told @outsidePasser to reach out to me if he wants to talk through it.

outsidePasser: Put back a change to clear the graph when needed.  I'll talk to @crericha to run it by him in the next day or so.

outsidePasser: Changed to use the game graph instead of a special graph per @crericha.  In order for that to work, the main game graph needed to have home_as_token set.  Not clear if that will break anything else in 22MX.  I will keep an eye on it.

outsidePasser: I think this one should be good to go.

lib/engine/game/base.rb
```diff
@@ -1073,6 +1073,13 @@ def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil
           stock_market.move_down(corporation) if was_president
         when :left_block_pres
           stock_market.move_left(corporation) if was_president
+        when :left_per_if_pres_else_left_one
```
tobymao, line 1076: what does the naming mean? left_per?

lib/engine/game/g_1822_mx/game.rb
```diff
@@ -590,8 +647,15 @@ def close_p16
           company.close!
         end
 
-        def company_status_str
-          ''
+        def company_status_str(company)
+          bidbox_minors.each_with_index do |c, index|
```
tobymao, line 651: (bidbox_minors.with_index + bidbox_connessions.with_index).flatten.each ...

lib/engine/game/g_1822_mx/step/minor_acquisition.rb
tobymao: this isn't doing auto actions how is this good to go?

outsidePasser: I'm very confused then.  auto_actions from 1822 minor_acquisition (which this inherits from) calls mergable, which calls entity_connects?, which calls connected_to_port?.  There's no other place this is called from except from inside auto actions.

lib/engine/game/base.rb
outsidePasser: I agree this is confusing.  left_per_10_if_pres_else_left_one is clearer.

lib/engine/game/g_1822_mx/step/minor_acquisition.rb
tobymao: ok so it is using auto actions? that's fine then

lib/engine/game/g_1822_mx/game.rb
outsidePasser: Changed to something cleaner

lib/engine/game/base.rb
outsidePasser: Changed

PR #7581, opened 2022-04-06
Add support for cities in graph's home_as_token option

crericha:
I ran this change against the entire database with strict=true. No games that use home_as_token failed.


crericha:
@outsidePasser FYI

tobymao (approved):

PR #7588, opened 2022-04-08
18Rhl: Speed up route process when the route is illegal

ventusignis:


perwestling (changes requested):
The termini checks ("Route cannot begin/end in a town" and the RGE (8 train) must start/end in specific off-boards) is just inbetween messages that makes increased routes possible. So does this not mean that some increases in routes become discarded, even though they would be legal?

lib/engine/game/g_18_rhl/game.rb
```diff
@@ -210,6 +210,21 @@ class Game < Game::Base
           },
         ].freeze
 
+        def compute_stops(route, train = nil)
+          return super unless route.train.name == '8' && check_distance_throws_game_error(route)
```
tobymao, line 214: don't use unless with multiple statements

perwestling:
One possible fix for the 8 problem and towns is to not give the error for 8 trains. For non-8 trains this error message is not a problem as their routes are much shorter.

ventusignis:
> The termini checks ("Route cannot begin/end in a town" and the RGE (8 train) must start/end in specific off-boards) is just inbetween messages that makes increased routes possible. So does this not mean that some increases in routes become discarded, even though they would be legal?

Looking at how 18Rhl's `check_distance` is implemented, it looks like every case where it raises `GameError`, the set of visits is outright illegal. All that `compute_stops` is doing is picking which of those visits to actually count towards the run. In the cases where the set of visits is outright illegal, no set of picked stops is going to make it legal again. The user is able to keep on clicking additional nodes to make the route legal

I'm not sure how this affects autorouting

JoeB989:
> I'm not sure how this affects autorouting

Auto-routing will continue extending the route when anything raises `GameError`.  So any invalid route that could become valid with more stops should raise `GameError`.  Invalid routes that can't become valid if extended should return a unique error that AR will detect and abort extensions, something like `RouteTooLong` or you could create a new one.

A quick look at `check_distance `seems like most errors should/could abort route extension (return `RouteTooLong` or new error) to shorten AR without losing valid routes.  Only "Route cannot end in a town" (split out from cannot begin/end in a town) needs to still return `GameError` so AR continues extending past the town.

RouteTooLong etc. are derived from GameError so your basic `rescue GameError` is sufficient for non-AR flows.

JoeB989:
To be clear, if you're swallowing GameErrors in AR code flows, that will mess up auto-router processing.  AR needs the to see the raised errors or it will likely not work properly.

JoeB989:
> and the RGE (8 train) must start/end in specific off-boards)

Note: for AR specifically, all routes will be built from both ends, so AR can consider this only as "must begin at".  So the code could return `UniqueError` if !begins, and `GameError` if !begins && ! ends

PR #7592, opened 2022-04-08
Implement variant DB2 improving the SECR

damerell:
I have not implemented this as two separate options since AFAIK anyone
who uses them uses them together, and '25 has _quite a lot_ of options
already.

### DIFF
--- a/lib/engine/game/g_1825/entities.rb
+++ b/lib/engine/game/g_1825/entities.rb
@@ -561,7 +561,7 @@ module Engine
             secr[:abilities] << { type: 'blocks_hexes', owner_type: nil, hexes: ['W21'] }
             PAR_BY_CORPORATION['SECR'] = 67
             lbsc = corps.find { |corp| corp[:sym] == 'LBSC' }
-            lbsc[:abilities] = nil
+            lbsc[:abilities] = []
           end
           add_entities(corps, R3_CORPORATIONS) if @regionals[3]
           add_entities(corps, K5_CORPORATIONS) if @kits[5]
--- a/lib/engine/game/g_1825/entities.rb
+++ b/lib/engine/game/g_1825/entities.rb
@@ -164,6 +164,7 @@ module Engine
             color: '#ffef00',
             text_color: '#000000',
             reservation_color: nil,
+            abilities: [],
           },
           {
             sym: 'LBSC',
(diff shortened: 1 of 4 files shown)


damerell:
I should perhaps have said that I am a Perl-writing sysadmin who knew no Ruby a week ago, so I expect you will need to correct me about some of what I have done.

lib/engine/game/g_1825/entities.rb
```diff
@@ -554,6 +555,14 @@ def game_corporations
             gwr[:tokens] = [0, 0, 40, 100, 100, 100, 100]
             gwr[:coordinates] = %w[V14 Y7]
           end
+          # Modify SECR / LBSC in variant DB2
+          if @optional_rules.include?(:db2)
+            secr = corps.find { |corp| corp[:sym] == 'SECR' }
+            secr[:abilities] << { type: 'blocks_hexes', owner_type: nil, hexes: ['W21'] }
+            PAR_BY_CORPORATION['SECR'] = 67
+            lbsc = corps.find { |corp| corp[:sym] == 'LBSC' }
+            lbsc[:abilities] = nil
```
roseundy, line 564: Shouldn't this be [] instead of nil ?

lib/engine/game/g_1825/meta.rb
```diff
@@ -175,6 +181,8 @@ def self.check_options(options, num_players)
           regionals[2] = true if optional_rules.include?(:r2)
           regionals[3] = true if optional_rules.include?(:r3)
 
+          dbvariants[2] = true if optional_rules.include?(:db2)
```
roseundy, line 184: Is dbvariants ever used? Why do you need this?

lib/engine/game/g_1825/entities.rb
damerell: Corrected.

lib/engine/game/g_1825/meta.rb
damerell: My mistake. I started thinking of a scheme like regionals[] etc but the Dave Berry variants don't really have much in common other than the author.

roseundy:
Actually, I believe Ruby was inspired in part by Perl.

damerell:
So do I, but (alas) being a mediocre Perl programmer doesn't help that much. :-/

PR #7599, opened 2022-04-10
Implement variant DB1: additional upgrade paths for double dits

damerell:
Unfortunately Dave Barry's kit DB1 is itself full of optional rules.
This implements the rule he "uses in practice", giving the 887/888
tiles to Units 1 and 3 but not 2 where they are primarily for the
more complex variant which he doesn't use in practice.

As with DB2 I have hence only implemented what I think is the most
popular version since 1825 has many options already.

### DIFF
--- a/lib/engine/game/g_1825/map.rb
+++ b/lib/engine/game/g_1825/map.rb
@@ -489,18 +489,22 @@ module Engine
         }.freeze
         # rubocop:enable Layout/LineLength
 
+        # This includes upgrades for the DB1 kit tiles 887/888. This does not
+        # affect the game without it since none of them are present.
         DIT_UPGRADES = {
           # gentle curve to three curves with a halt
           '8' => %w[11],
           # yellow double-dit to green K or X city
-          '1' => %w[14],
-          '2' => %w[15],
-          '55' => %w[14],
-          '56' => %w[15],
+          '1' => %w[14 888],
+          '2' => %w[15 887],
+          '55' => %w[14 888],
+          '56' => %w[15 887],
           '69' => %w[119],
-          '114' => %w[15],
+          '114' => %w[15 887],
           '198' => %w[119],
           '199' => %w[119],
+          '887' => %w[63 166],
+          '888' => %w[63 166],
           # yellow single-dit to green city (also brown/green city)
           '3' => %w[12 14 15 119],
           '4' => %w[14 15 119],
@@ -544,6 +548,20 @@ module Engine
           append_game_tiles(gtiles, K5_TILES) if @kits[5]
           append_game_tiles(gtiles, K6_TILES) if @kits[6]
           append_game_tiles(gtiles, D1_TILES) if @optional_rules.include?(:d1)
+          gtiles = db1_tiles(gtiles) if @optional_rules.include?(:db1)
+          gtiles
+        end
+
+        def db1_tiles(gtiles)
+          gtiles.delete('87')
+          gtiles.delete('88')
+          eightysevens = ((@units[1] ? 2 : 0) + (@units[3] ? 1 : 0))
+          gtiles['887'] = eightysevens
+          gtiles['888'] = eightysevens
+          if @units[3]
+            gtiles['14'] -= 1
+            gtiles['15'] -= 1
+          end
           gtiles
         end
 
--- a/lib/engine/game/g_1825/map.rb
+++ b/lib/engine/game/g_1825/map.rb
@@ -487,6 +487,11 @@ module Engine
           },
           '200' => 2,
         }.freeze
+
+        DB1_UNIT3_ANTITILES = {
+          '14' => -1,
+          '15' => -1,
+        }.freeze
         # rubocop:enable Layout/LineLength
 
         # This includes upgrades for the DB1 kit tiles 887/888. This does not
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1825/step/track_and_token.rb
```diff
@@ -74,6 +74,10 @@ def lay_tile_action(action, entity: nil, spender: nil)
             if tile_lay[:cannot_reuse_same_hex] && @round.laid_hexes.include?(action.hex)
               raise GameError, "#{action.hex.id} cannot be layed as this hex was already layed on this turn"
             end
+            if (tile.name == '888' || tile.name == '887') &&
+               action.hex.id == 'Q13' && (entity.name != 'NS')
```
roseundy, line 78: Inconsistent use of parens

damerell: Oh bother. Rubocop likes to make you take them out, whereas I like to put them in. I think that's how this happened. Correcting.

lib/engine/game/g_1825/map.rb
```diff
@@ -544,6 +548,20 @@ def game_tiles
...
+
+        def db1_tiles(gtiles)
+          gtiles.delete('87')
+          gtiles.delete('88')
+          eightysevens = ((@units[1] ? 2 : 0) + (@units[3] ? 1 : 0))
+          gtiles['887'] = eightysevens
+          gtiles['888'] = eightysevens
+          if @units[3]
+            gtiles['14'] -= 1
+            gtiles['15'] -= 1
```
roseundy, line 563: I'm sure this works, but it might be nice to generalize append_game_tiles to deal with negative counts and possibly create a "delete_game_tiles" method. You might get weird behavior if gtile[tile] was decremented to zero (i.e. it would show up in the Tiles tab).

lib/engine/game/g_1825/step/track_and_token.rb
```diff
@@ -74,6 +74,10 @@ def lay_tile_action(action, entity: nil, spender: nil)
             if tile_lay[:cannot_reuse_same_hex] && @round.laid_hexes.include?(action.hex)
               raise GameError, "#{action.hex.id} cannot be layed as this hex was already layed on this turn"
             end
+            if (tile.name == '888' || tile.name == '887') &&
+               (action.hex.id == 'Q13') && (entity.name != 'NS')
+              raise GameError, 'Only the NS can place tiles 887/888 on Q13'
+            end
```
roseundy, line 80: This works, but it will show 888 and 887 as possible upgrades for Q13 for the other companies. This should probably go into @game.upgrades_to?

damerell: I've tried that, but on slapping in some debugging output, upgrades_to never seems to be called in 1825 with selected_company other than nil. I guess I could work out everywhere it gets called from and which of the base methods I need to replace but it seems like that might make the implementation otherwise more fragile because if those base methods change, g_1825 won't keep up.

(My terminology may be rather confused; sorry)

lib/engine/game/g_1825/map.rb
damerell: I'm not sure if I should read "it might be nice" as "you should do this to get the PR accepted". I would rather not if I don't have to because as I may have mentioned I am desperately cargo-culting my way through Ruby and - while I don't have a deep knowledge of many titles - I'm not aware of other cases where it would be useful and 1825's tile manifest means the zero case won't arise.

roseundy: I would strongly prefer it (Toby can make the final call). I'll just point out that having the helper methods makes adding optional rules easier.

lib/engine/game/g_1825/step/track_and_token.rb
roseundy: Yeah, that's unfortunate. Never mind then.

lib/engine/game/g_1825/map.rb
tobymao: yep let's do @roseundy's suggestion

lib/engine/game/g_1825/step/track_and_token.rb
roseundy: I think you can put this in upgradable_tiles, which does get the current entity as a parameter. It is the ultimate cause of upgrades_to?

roseundy: Your call

lib/engine/game/g_1825/map.rb
damerell: OK, I'll have a crack at it tomorrow.

lib/engine/game/g_1825/step/track_and_token.rb
damerell: I implemented this - hence the appearance of a protect_ns method in the update I'm about to push - but the result is that when you go to upgrade Q13, the 887/888s don't appear and there is no feedback as to why; you just have to know that rule in variant DB1 or infer that it was necessary. This strikes me as worse than being able to try it and getting a clear message as to why it's not permitted. Therefore, I unimplemented it.

lib/engine/game/g_1825/map.rb
```diff
@@ -544,6 +560,17 @@ def game_tiles
           append_game_tiles(gtiles, K5_TILES) if @kits[5]
           append_game_tiles(gtiles, K6_TILES) if @kits[6]
           append_game_tiles(gtiles, D1_TILES) if @optional_rules.include?(:d1)
+          gtiles = db1_tiles(gtiles) if @optional_rules.include?(:db1)
+          gtiles
```
tobymao, line 564: why do you return g_tiles here?

tobymao: why do you reassign gtiles? you mutate it in db1_tiles?

damerell: I reassign it because the way everything even function arguments is a reference was a bit of a mystery to me. Fixed.

I return gtiles because that's what game_tiles did before I started messing around with it.

tobymao (approved):

PR #7602, opened 2022-04-10
Fixed issue where owners were not removed from share_holders

outsidePasser:


roseundy:
Why is this needed? What's wrong with corporation.share_holders having a zero value?

outsidePasser:
In 22MX, players who are shareholders in NDEM can lay track.  Checking this by looking only at player_share_holders doesn't work, because players who used to own a share but sold it still are in there.  I worry that not removing players from the list when they are not actually share holders could cause subtle bugs.

tobymao:
this is unneeded just check that the value is greater than 0.

PR #7606, opened 2022-04-11
[18CZ] gray normal cities only have 2 slots

### LINKED ISSUE #7583: [18CZ] grey tile for normal city
The grey tile for normal city should get only 2 spaces and not 3 spaces

benjaminxscott:
closes #7583 

From fwtwr and user report
image


Zwergenpunk:
When I read the issue I thought about pinging you to fix thr issur sincr I'm on vacation 😆 thanks for fixing :)

@toby this will pin alot of games

PR #7633, opened 2022-04-16
1825: Correct typo making F8 into a single dit not a double dit

damerell:
There was a semicolon missing in the definition of hex F8.


roseundy:
It's curious that this didn't cause the game to bomb out.

PR #7636, opened 2022-04-17
1825: implement Francis Tresham's unpublished Unit 4 map

damerell:
Francis Tresham never got around to selling this but he distributed
various copies either as envelope stiffeners or at conventions, so
the layout is known.

Fortunately it does not introduce any more companies, trains, money etc
so there's no need to add three options (units 3+4, 2+3+4, 1+2+3+4).

It is intentional that this implementation doesn't add a @units[4];
there's a lot of 'case units.keys.sort.map(&:to_s).join' which would
break.

### DIFF
--- a/lib/engine/game/g_1825/game.rb
+++ b/lib/engine/game/g_1825/game.rb
@@ -706,15 +706,18 @@ module Engine
         end
 
         def location_name(coord)
-          unless @location_names
-            @location_names = LOCATION_NAMES.dup
-            if optional_rules.include?(:unit_4)
-              @location_names['A5'] = 'Inverness'
-              @location_names['C3'] = 'Fort William'
-              @location_names.delete('B8')
-            end
-          end
+          @location_names ||= game_location_names
           @location_names[coord]
+        end
+
+        def game_location_names
+          locations = LOCATION_NAMES.dup
+          if optional_rules.include?(:unit_4)
+            locations['A5'] = 'Inverness'
+            locations['C3'] = 'Fort William'
+            locations.delete('B8')
+          end
+          locations
         end
 
         def upgrades_to?(from, to, special = false, selected_company: nil)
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1825/game.rb
```diff
@@ -704,6 +705,18 @@ def tile_color_valid_for_phase?(tile, phase_color_cache: nil)
...
+            @location_names = LOCATION_NAMES.dup
+            if optional_rules.include?(:unit_4)
+              @location_names['A5'] = 'Inverness'
+              @location_names['C3'] = 'Fort William'
+              @location_names.delete('B8')
+            end
+          end
+          @location_names[coord]
+        end
+
```
roseundy, line 719: @locations_name || = ....

damerell: I'm sorry, I'm not really clear on what you're suggesting here, beyond "implement this differently". I copied what 1824 does. Is there another game I should emulate, please?

roseundy: Assigning a "caching" variable has a pretty standard usage pattern on 18xx.games. Look at the definition of bank_by_options in 1825 for an example. Or just git grep for '||='.

roseundy: Given what you are doing, this is fine.

I would do it like this (mostly just personal preference, same effect as what you have):
```
def location_name(coord)
  @locations_names ||= game_location_names
  @locations_names[coord]
end

def game_location_names
  locations = LOCATION_NAMES.dup
  if optional_rules.include?(:unit_4)
 ...
  end
end
```

damerell: I haven't forgotten about this, I'll do it when I've got some spare time.

roseundy: Like I said, I think what you have is fine. If you don't have time, don't worry about it.

tobymao (approved):

PR #7686, opened 2022-05-03
1822MX ndem auction

outsidePasser:


assets/app/view/game_page.rb
```diff
@@ -410,6 +410,10 @@ def render_action
           h(Game::Round::Merger, game: @game)
         elsif current_entity_actions.include?('buy_shares') && @game.current_entity&.player?
           h(Game::Round::Stock, game: @game)
+        elsif current_entity_actions.include?('bid')
+          h(Game::Round::Auction, game: @game, user: @user)
+        elsif current_entity_actions.include?('choose')
+          h(Game::Round::Choices, game: @game)
```
roseundy, line 416: Won't this break existing games that have a "choose" action in an OR? There are several that do. Not sure about "bid"

roseundy: If you want a "Choices" round, why are you doing it from an Operating round?

outsidePasser: When NdeM privatization occurs, the NdeM has one more operating round that runs normally, and then there is a phase where the NdeM tokens are auctioned off.  I am using "Choices" to allow the current player to pick which NdeM token they want to start an auction for.  I may have not been understanding the code, but it looked like it was not limited to using "Game::Round::foo" within "Engine::Round::foo".  

If there are other games that have something that chooses a token from the map from within an OR, I can conform to what they did.

ventusignis: 1856 does this during CGR formation (mid-OR) - if the CGR forms with too many tokens the president must choose which ones to remove

https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1856/step/remove_tokens.rb

outsidePasser: Thank you - I will take a look.

roseundy: My point is that you are forcing a different UI::Round when "choose" is an action. That will break games that rely on "choose" still using UI::Round::Operating

outsidePasser: As it turns out - I had already changed it to remove_token - I had just forgotten.  Removing "choose".  I do not think that what I have would work if I removed the "bid" one.  What would be the best way to make sure I didn't break anything else with that change?

roseundy: Look at UI::Round::Operating and see if it checks for a "bid" action. If it does, you are very likely breaking other games.

outsidePasser: Do you mean Game::Round::Operating?  If so, it doesn't check for a "bid" action.

roseundy: View::Game::Round::Operating (assets/app/view/game/round/operating.rb)

roseundy: It's a little strange to use View::Game::Round::Auction for an operating round.

outsidePasser: Would the alternative be to implement "bid" inside View::Game::Round::Operating?  I did it this way because I thought there would be a lot of duplication between the code I'd be adding to Operating and Auction.

outsidePasser: I was also basing it on the code using View::Game::Round::Merger and View::Game::Round::Stock inside an operating round that was already there.

roseundy: It's probably OK, I'm just being paranoid.

outsidePasser: Paranoia is good.  Is there anything else I should address for this PR?

lib/engine/game/g_1822_mx/game.rb
```diff
@@ -506,13 +510,23 @@ def company_tax_haven_bundle(choice); end
...
+        end
+
         def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil)
           return super unless bundle.corporation == corporation_by_id('NDEM')
 
           @share_pool.sell_shares(bundle, allow_president_change: false, swap: swap)
         end
 
         def operating_order
+          # print("operating_order")
```
roseundy, line 529: Sorry, I missed this before. Remove the comment.

roseundy (approved):

PR #7693, opened 2022-05-05
18GB Second edition playtest variant

andypymont:
Add a new pre-game variant for 18GB (default to on for new games) which contains a bundle of proposed second-edition changes for playtesting:
- GN and MC privates provide additional station markers rather than using an existing marker of the corporation
- Home stations of all corporations are placed on the board at the start of the game
- Green area of the market is changed: instead of the current ability, this area allows a corporation president to purchase 2 shares per stock round turn
- N-S bonus is the same as the E-W bonus (20/30/40 in yellow/blue/gray phases)
- 3+1 trains are replaced by 3+2 trains and 5+2 trains are replaced by 3X trains
- No token placements or share sales after an OR ends with 3 or fewer trains remaining)


lib/engine/game/g_18_gb/step/special_token.rb
```diff
@@ -28,12 +40,14 @@ def adjust_token_price_ability!(_entity, token, hex, _city, special_ability: nil
...
 
           def switch_for_expensive_token(token)
+            return token if @game.second_ed_playtest?
             return token unless token.price.zero?
 
             current_entity.tokens.reject(&:used).max_by(&:price)
           end
 
           def process_place_token(action)
+            puts "from_owner: #{ability(action.entity).from_owner}"
```
roseundy, line 50: Debug statement

roseundy (approved):

PR #7695, opened 2022-05-06
1822: Use FutureLabels for S & T tiles

ventusignis:
I updated 1822 to use FutureLabels instead of special upgrade logic. The yellow S tile upgrade logic is still needed because the town is removed in yellow.
I also found and fixed a JS v Ruby bug when testing this which is now fixed

image
image
image
image
image

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1512,7 +1512,7 @@ module Engine
 
       def upgrades_to_correct_label?(from, to)
         # If the from tile has a future label and the to tile is the color for it use that, otherwise use the from's label
-        return from.future_label.label == to.label.to_s if from.future_label && to.color == from.future_label.color
+        return from.future_label.label == to.label.to_s if from.future_label && to.color.to_s == from.future_label.color
 
         from.label == to.label
       end
--- a/lib/engine/game/g_1822/game.rb
+++ b/lib/engine/game/g_1822/game.rb
@@ -529,13 +529,6 @@ module Engine
 
         attr_accessor :bidding_token_per_player, :player_debts
 
-        def all_potential_upgrades(tile, tile_manifest: false, selected_company: nil)
-          upgrades = super
-          return upgrades unless tile_manifest
-
-          upgrades
-        end
-
         def bank_sort(corporations)
           corporations.reject { |c| c.type == :minor }.sort_by(&:name)
         end
--- a/lib/engine/game/g_1822/game.rb
+++ b/lib/engine/game/g_1822/game.rb
@@ -522,8 +522,6 @@ module Engine
         UPGRADABLE_S_YELLOW_CITY_TILE = '57'
         UPGRADABLE_S_YELLOW_ROTATIONS = [2, 5].freeze
         UPGRADABLE_S_HEX_NAME = 'D35'
-        UPGRADABLE_T_YELLOW_CITY_TILES = %w[5 6].freeze
-        UPGRADABLE_T_HEX_NAMES = %w[B43 K42 M42].freeze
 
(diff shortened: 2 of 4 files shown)


lib/engine/game/base.rb
```diff
@@ -1512,7 +1512,7 @@ def upgrades_to_correct_color?(from, to, selected_company: nil)
 
       def upgrades_to_correct_label?(from, to)
         # If the from tile has a future label and the to tile is the color for it use that, otherwise use the from's label
-        return from.future_label.label == to.label.to_s if from.future_label && to.color == from.future_label.color
+        return from.future_label.label == to.label.to_s if from.future_label && to.color.to_s == from.future_label.color
```
roseundy, line 1515: Why does `to` requires `to_s`, but not `from`?

ventusignis: `from`is a `tile`, which (effectively) validates the string `color` when parsing the tile config and converts `color` to a `symbol` when the tile is being created, whereas in `future_label`, `color` is just a `string` causing this to `==` in JS, but `!=` in Ruby without the `to_s

```ruby
module Engine
  class Tile
    include Config::Tile

    attr_accessor :blocks_lay, :hex, :icons, :index, :legal_rotations, :location_name,
                  :name, :opposite, :reservations, :upgrades, :color, :future_label
    attr_reader :borders, :cities, :edges, :junction, :nodes, :labels, :parts, :preprinted, :rotation, :stops, :towns,
                :offboards, :blockers, :city_towns, :unlimited, :stubs, :partitions, :id, :frame, :stripes, :hidden

    ALL_EDGES = [0, 1, 2, 3, 4, 5].freeze

    def self.for(name, **opts)
      if (code = WHITE[name])
        color = :white
      elsif (code = YELLOW[name])
        color = :yellow
        ...
```

perwestling:
Interesting. Would be nice to implement for more titles (e.g. 18MEX) if this works well.

One thing I have not checked if this functionality work for the tile page? That is show potential upgrades based on these future labels? E.g. in 18MEX if clicking on green 619, 14 or 15 it will show both the standard brown 63 and the special Guadalajara brown 480?
In the same way clicking on standard yellow city tiles in 1822 would show green S and T as well as the standard green 2-city tiles.

lib/engine/game/g_1822/game.rb
```diff
@@ -535,10 +533,6 @@ def all_potential_upgrades(tile, tile_manifest: false, selected_company: nil)
           upgrades = super
           return upgrades unless tile_manifest
 
-          upgrades |= [@green_s_tile] if self.class::UPGRADABLE_S_YELLOW_CITY_TILE == tile.name
```
magnusstjarnhild, line 538: Looks like whole method  all_potential_upgrades could be removed?

ventusignis:
> One thing I have not checked if this functionality work for the tile page? That is show potential upgrades based on these future labels? E.g. in 18MEX if clicking on green 619, 14 or 15 it will show both the standard brown 63 and the special Guadalajara brown 480?
In the same way clicking on standard yellow city tiles in 1822 would show green S and T as well as the standard green 2-city tiles.

This way of doing it doesn't show them as possible upgrades on the tile page. 
A possible set of enhancements would be to:

-  Add a switch to the tile page to show all current map tiles
- Show future labels & sticky icons on upgraded tiles in tile upgrade choices
- Allow iteratively showing future upgrade paths

lib/engine/game/g_1822/game.rb
ventusignis: It's removed now

magnusstjarnhild (approved):

PR #7702, opened 2022-05-07
Add variant DB3

damerell:
As with the other DB variants, I have implemented only what I think
is the most commonly used set of changes, since 1825 has many options
already.

This implementation does (purely cosmetically) add town names to hex
I13 even in non-DB3 games since that seems harmless.


tobymao:
does this variant do anything?

damerell:
Yes. The changes to I9, J6, G13, H10, F4, and F2 have game mechanical effects, as does the addition of two tiles. The only cosmetic changes are that I13 is named and G7 renamed.

roseundy (approved):

PR #7704, opened 2022-05-07
[18NY] Bug fixes and enhancements

### LINKED ISSUE #7075: Show upgrades for yellow town tiles from the start of the game
18xx.games usually shows the upgrades that will become available in later phases.  You can see this by cliicking on a tile in the Tiles tab or on a tile on the map during a company's turn.
However, it doesn't show the upgrades available in 18NY for yellow town tiles, until the first 4H train has been purchased.  I.e. it only shows the upgrades after they have become available.

See e.g. https://18xx.games/game/71704?action=86

### LINKED ISSUE #7539: 18NY corporation offered special company ability incorrectly
https://18xx.games/game/76110?action=600

The NH should not be offered the West Point Foundry discount, which is only available to its owner the HR.

### LINKED ISSUE #7584: [18NY] Wrong stock movement for minors
The minors need to pay out atleast its share value to move to the right. The rules 

7.5.1 Stock Price Movement
Minor Companies consider the total revenues for stock price movement, while Share Companies consider the amount paid as dividends.
If that amount is greater than or equal to the stock price, the stock price moves one space to the right. If earnings are withheld or the amount paid is less than half the stock price, the stock price moves one space to the left. If the amount paid is at least half but less than the stock price, the stock price does not change.

Example game:

https://18xx.games/game/79491?action=85

And if we look at the code it only looks if the revenue is positive.

```ruby
            return { share_direction: :right, share_times: 1 } if entity.type == :minor && revenue.positive?
```

crericha:
Fixes #7584 
Fixes #7539 
Fixes #7075

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -90,7 +90,7 @@ module View
                 if real_tile
                   tiles.delete(real_tile)
                   [real_tile, nil]
-                elsif !@game.tile_color_valid_for_phase?(tile, phase_color_cache: phase_colors)
+                elsif !@game.tile_valid_for_phase?(tile, hex: @tile_selector.hex, phase_color_cache: phase_colors)
                   [tile, 'Later Phase']
                 elsif @game.tiles.none? { |t| t.name == tile.name }
                   [tile, 'None Left']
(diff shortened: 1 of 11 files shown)


crericha:
@roseundy there are some common code refactoring in the PR. Can you please review them?

roseundy (approved):

PR #7714, opened 2022-05-08
[18USA] Allow pullman scrapping, even when there are no loan actions

crericha:


crericha:
There are no games in my copy of the production db that need a pin, but there may be one of the latest version of the production db.

michaeljb (approved):

Issue #7752, opened 2022-05-12
1822MX Minor 18 issue


ScottDrone:
1822MX Hotseat Game

https://gist.github.com/ScottDrone/39def40555cc68272f17419e27d70b47

Same game as this morning I found an anomaly when reviewing certificates in hand.  The first 7 train was just purchased, the person that purchased the FCM/18 combo at the beginning of the game still shows a "MINOR: 18 Ferrocarril Mexicano" in the section of player holdings typically occupied by private companies.  Minor 18 was long since acquired by the FCM but this card still seems to be occupying a certificate slot towards the player limit and providing $100 in value to the players total score.

I would not expect this item to persist after minor 18 is acquired by a major company.

![Screenshot 2022-05-12 174211](https://user-images.githubusercontent.com/105426708/168179261-c6ca4474-4fe0-42d5-8fa6-0e85bf15fca8.png)

Issue #7736, opened 2022-05-11
[22MX] cannot sell share when emergence finance


familiar-hk:
As title

outsidePasser:
Do you have a game where this occurred?

familiar-hk:
hotseat game

familiar-hk:
![image](https://user-images.githubusercontent.com/98611238/167809596-a5e740db-d851-401f-85f6-22d97963c428.png)

outsidePasser:
Can you dump out the game data and post a link to it?

familiar-hk:
[hs_rlruzqrm_1652241896.zip](https://github.com/tobymao/18xx/files/8681669/hs_rlruzqrm_1652241896.zip)

outsidePasser:
Thank you for the game data.  Fix incoming.

Issue #7732, opened 2022-05-11
22MX - orange highlight when when it’s your turn not working 


fsaglio:
The orange ribbon to alert you that it is your turn doesn’t appear to work consistently in 22MX.   In both games I have tried it has either permanently been there regardless of whose turn it is, or hasn’t ever turned orange.  

Has happened to the other players in both games as well.  

https://18xx.games/game/83635

PR #7754, opened 2022-05-13
[1822MX] Bug fixes

outsidePasser:
Fix problem where player cannot sell shares to forced purchase trains

Fixes: #7736

Fix problem where player is given M18 minor as a private

Fixes: #7752

Fix problem where active_player is not reset between rounds

Fixes: #7732

### DIFF
--- a/lib/engine/game/g_1822_mx/game.rb
+++ b/lib/engine/game/g_1822_mx/game.rb
@@ -544,8 +544,9 @@ module Engine
         end
 
         def active_players
-          return super if @round.active_step&.current_entity != ndem
-          return [@round.active_step.ndem_acting_player] if @round.active_step.respond_to?(:ndem_acting_player)
+          if current_entity == ndem && @round.active_step.respond_to?(:ndem_acting_player)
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_1822_mx/game.rb
```diff
@@ -544,7 +544,8 @@ def operating_order
         end
 
         def active_players
-          return [@ndem_acting_player] if @ndem_acting_player
+          ndem_acting_player = @round.active_step.respond_to?(:ndem_acting_player) ? @round.active_step.ndem_acting_player : nil
+          return [ndem_acting_player] if ndem_acting_player
```
crericha, line 548: return [@round.active_step.ndem_acting_player] if @round.active_step.respond_to?(:ndem_acting_player)

crericha: Also, should you check if current_entity == ndem?

outsidePasser: Yep, changes incoming

crericha (approved):

PR #7755, opened 2022-05-13
[1866] IAR Rework

magnusstjarnhild:
Based on playerfeedback, observing played games and more playtesting Simon now have decided on the following changes.

**1866 Specific**
- Removed the option of 7'th player. Valid playercount are 3 to 6.
- Removed preprinted Paris tile on the map, its now a normal C hex.
- Added a new C tile, for yellow, green, brown and gray.
- Used the CitiesPlusTownsRouteDistanceStr include to show seperate cities and towns when calculate the route.
- The 200 place on the stockmarket is now a normal gray spot.
- Changed the price for the palacecar from 80 to 75.
- The palacecar only counts the tokened cities for the E-trains.
- Change the way how the stock turn token moves when a player passes, buy and sells shares.
- When a corporation closes the bank will take take of the loans. The player is no longer liable for the loans.
- When buying a Stock Turn token in the OR, it will operate if its later in the operating order.
- Removed the Stock Turn Token end game trigger line on the stock market. Stock Turn Tokens now trigger engame by enter the blue area.
- Reworked the Initial Auction Round. You now have the option of bid for the company, buy a stock turn token, start a corporation or pass.

**Changes to base code**
- View::Game::Round::Auction.render - Made a respond_to? check for the action par and the method companies_pending_par
- View::Game::Round::Auction.render - Added call to new methods render_par_corporations and render_stock_market
- View::Game::Round::Auction.render_par_corporations - New method to show corporations that you can start during the auction. Need action par and method par_corporations(entity) in the step.
- View::Game::Round::Auction.render_stock_market - New method to show the stock market on the auction page. Need method show_stock_market? and set to true to show it.

There are major changes how the IAR works and the stock turn tokens move so we will need to archive all of the 1866 games.


assets/app/view/game/round/auction.rb
```diff
@@ -320,6 +323,26 @@ def render_corporations
...
+
+          props = {
+            style: {
+              display: 'inline-block',
+              verticalAlign: 'top',
+            },
+          }
+
+          corporations = []
+          @step.par_corporations(@current_entity).map do |corporation|
```
tobymao, line 337: should be each, you're not using the map functionality

lib/engine/game/g_1866/step/single_item_auction.rb
```diff
@@ -78,11 +132,15 @@ def description
 
           def get_par_prices(entity, corp = nil)
             par_type = @game.phase_par_type
-            par_prices = @game.par_prices_sorted.select do |p|
-              p.types.include?(par_type) && p.price <= entity.cash && @game.can_par_share_price?(p, corp)
+            @game.par_prices_sorted.select do |p|
+              multiplier = corp.nil? ? 1 : 2
```
tobymao, line 136: i don't like nil? can you do 

corp ? 1 : 2

Issue #7756, opened 2022-05-13
1822MX - Build cubes


ScottDrone:
Another hotseat 1822MX bug report

https://gist.github.com/ScottDrone/d06187fe26649255e01a995741df9792

A major corporation upgraded tile L19 for the first tile lay in green phase after a 3 train was purchased

The same major was then allowed to place 4 build cubes in a row without being moved to the build station step, the Done(track) button must be pressed to move forward.  The major does not own a private that would enable them to lay 3 additional cubes.

PR #7761, opened 2022-05-14
[1822MX] Check to make sure company can lay tile before accepting builder cube

outsidePasser:
Fixes: #7756


lib/engine/game/g_1822_mx/step/track.rb
```diff
@@ -49,6 +49,9 @@ def pass!
           def process_lay_tile(action)
             @log << "Tile placement for NDEM by #{@ndem_tile_layers.first.name}" if current_entity == @game.ndem
             if action.tile.id == 'BC-0'
+              tile_lay = get_tile_lay(action.entity)
+              raise GameError, 'Cannot lay a builder cube now' unless tile_lay && tile_lay[:lay]
```
tobymao, line 53: don't use unless with multiple conditions

PR #7762, opened 2022-05-14
18SJ: Add spec file and some minor fixes

### LINKED ISSUE #7727: [18SJ] Need to clear graph after merger
While regression testing a Graph optimization I'm making, I discovered a failure in 18SJ that only occurs when validating with `strict: true`

After debugging, I discovered that the BuyCompany step doesn't clear graph after performing the merge.

perwestling:
Contains a spec file for the latest version (0.92).

Also, includes a few minor fixes:
 * Correct a minor error in a log message
 * Change name to the new official
 * Add clear of graph after KHJ merge (fixes #7727)

### DIFF
--- a/lib/engine/game/g_18_sj/meta.rb
+++ b/lib/engine/game/g_18_sj/meta.rb
@@ -11,10 +11,10 @@ module Engine
         DEV_STAGE = :production
         PROTOTYPE = true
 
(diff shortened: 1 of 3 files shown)


perwestling:
@roseundy I added a line to clear the graph. Does it looks OK?

PR #7790, opened 2022-05-17
18LA2: private company Angeles Public Dump, move to alpha

michaeljb:
The Dump places a -$20 station token which blocks everyone--including the
corporation that placed it--and reduces the city's revenue by $20 for all routes

Westminster is the only $10 target for the Dump, and if the Dump is placed there
the net revenue is $0 instead of -$10. Westminster is also a target for the
Steamship (adding $40), and if both are placed there, then the Dump's full
penalty applies for a net revenue of $30.

In order to allow a corporation owning Angeles Public Dump to place the Dump
token in a city where they have a regular token, token abilities have a new
attribute, `check_tokenable`, which can be set to `false` to bypass the usual
tokening requirements; at the time the check is made, the token still belongs to
the owning corporation, rather than the dummy DUMP corporation, and since a new
Token is created when using a special ability in `adjust_token_price_ability!`,
`check_tokenable` was needed to get around the usual token requirement of only
one token per city.

### DIFF
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -18,6 +18,7 @@ module View
     def render_notification
       message = <<~MESSAGE
 
+        <p><a href='https://github.com/tobymao/18xx/wiki/18LosAngeles#2nd-edition'>18 Los Angeles 2</a> is in alpha.</p>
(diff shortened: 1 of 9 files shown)


michaeljb:
closing to fix failing tests

PR #7795, opened 2022-05-18
Fix buy_share to handle preferred shares with bundles correctly

### LINKED ISSUE #7787: [Harzbahn 1873] Game stuck in loading loop forever
Hi,

Game 78782 won't load - it just sits in "Loading game..." forever.  Is there any way to rescue this game?

Thanks,
Eric

https://18xx.games/game/78782

roseundy:
Fixes #7787

### DIFF
--- a/assets/app/view/game/button/buy_share.rb
+++ b/assets/app/view/game/button/buy_share.rb
@@ -31,7 +31,7 @@ module View
           text += " #{@partial_percent}% of" if @partial_percent
           text += " #{bundle.percent}%" if show_percentage
           text += " #{@source}"
-          text += ' Preferred' if @share.preferred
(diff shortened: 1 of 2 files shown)


tobymao (approved):

tobymao:
wait a second, why would share not have the preferred method?

tobymao (changes requested):
something is fishy here

roseundy:
Because share could be a bundle, which doesn't have that method.

tobymao:
can it have that method?

roseundy:
The bug is that Button::BuyShare was modified to handle preferred shares without recognizing that bundles can also be bought.

tobymao (approved):

PR #7809, opened 2022-05-21
Draft version of 2D market arrows.

### LINKED ISSUE #7078: Add arrows to 2D stock markets
All (or at least most) physical copies of 18xx games with 2D stock markets have arrows at the left and the right to show that tokens will move down or up when they reach the end of the row.  It would be helpful to have this visual cue in 18xx.games as well.

andrewzwicky:
Fixes #7078 .

Currently not working for 2D markets that slope downward (see pictures).  I am checking to see that I'm even remotely in the right area of code, or if this should go in the individual market specifications or somewhere else.

image
image

### DIFF
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -259,7 +259,7 @@ module View
                 align = { left: 0, bottom: 0 }
                 arrow = '⭣'
               # last cell on right, not top row
-              elsif ((col_i + 1) == row_prices.length) && !row_i.zero?
+              elsif ((col_i + 1) == row_prices.size) && !row_i.zero?
                 align = { right: 0, top: 0 }
                 arrow = '⭡'
               else
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -245,7 +245,7 @@ module View
         # Need to peek at row below to know if sitting on ledge.
         @game.stock_market.market.push([]).each_cons(2).each_with_index.map do |rows, row_i|
           row_prices, next_row = rows
-          first_price=true
+          first_price = true
 
           row = row_prices.each_with_index.map do |price, col_i|
             if price
@@ -256,8 +256,8 @@ module View
 
               # first cell on left, not on bottom row, has price in cell below
               if first_price && !next_row.empty? && next_row[col_i]
-                  align = { left: 0, bottom: 0 }
-                  arrow = '⭣'
+                align = { left: 0, bottom: 0 }
+                arrow = '⭣'
               # last cell on right, not top row
               elsif ((col_i + 1) == row_prices.length) && !row_i.zero?
                 align = { right: 0, top: 0 }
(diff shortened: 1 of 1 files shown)


andrewzwicky:
For the purposes of getting proof of concept, this assumes all 2D markets function this same way, but for actually merging I'd need to check which ones might not match this approach.

assets/app/view/game/stock_market.rb
```diff
@@ -242,17 +242,29 @@ def grid_zigzag(zigzag)
...
-          row = row_prices.map do |price|
+        @game.stock_market.market.each_with_index.map do |row_prices, i|
+          row = row_prices.each_with_index.map do |price, j|
             if price
               corporations = price.corporations
               num = corporations.size
               spacing = num > 1 ? (RIGHT_TOKEN_POS - LEFT_TOKEN_POS) / (num - 1) : 0
               tokens = corporations.map.with_index { |corp, index| h(:img, token_props(corp, index, num, spacing)) }
 
+              if j.zero? && ((i + 1) != @game.stock_market.market.length)
```
tobymao, line 253: change length to size

tobymao:
looks fine

roseundy:
At some point, I'd like this to be extended to allow for arrows in the middle of a row for games where parts of the market are out of play until some point in the game (`*cough* `1841 `*cough*`)

andrewzwicky:
image
image
image

assets/app/view/game/stock_market.rb
```diff
@@ -242,17 +242,37 @@ def grid_zigzag(zigzag)
       end
 
       def grid_2d
-        @game.stock_market.market.flat_map do |row_prices|
-          row = row_prices.map do |price|
+        # Need to peek at row below to know if sitting on ledge.
+        @game.stock_market.market.push([]).each_cons(2).each_with_index.map do |rows, row_i|
```
tobymao, line 246: you don't need to do a flat map?

andrewzwicky: I'm just learning ruby, so please correct me if I'm misunderstanding.

I don't think there's anything to flatten here.  Outer loop is rows, inner loop is columns.  Each column loop creates an array based on the cells, but the row loop doesn't actually see that array, it only gets the `max-content` div.  I don't really know what's going on deep in these `h` objects, but I think `map` and `flat_map` have the same result here.

andrewzwicky: I should amend that to say that `flat_map` and `map` should be expected to produce the same thing, because the result only has 1 axis.

After you posted this, I tried `flat_map` and it doesn't work.  The details of why it doesn't work are beyond me right now, but I'm trying to figure it out.

"  from <embedded>:1523:35:in `constructor.method_missing_stub'"

tobymao: but why did the previous code use a flat_map?

andrewzwicky: I don't think the old code needed it either.  Using the code from master, I printed `grid.size` and `grid.flatten.size` for a handful of 2D markets, and the number was always the same.  I don't have an issue trying to get `flat_map` to work, but I'd need some help with that `method_missing_stub` thing

Issue #7808, opened 2022-05-21
1822MX Major can't close P16 Stock drop private


Ticketautomat1:
https://18xx.games/game/84867

outsidePasser:
This is a result of using the base 1822 SpecialChoose step, which doesn't include the general SpecialChoose step.  This had been working until a previous PR regrading selling shares broke it.

PR #7822, opened 2022-05-25
[1822MX] Fix inability for a company to close P16.

outsidePasser:
This bug was introduced with the fix for issue 7759

Fixes: #7808


lib/engine/step/special_choose.rb
```diff
@@ -9,6 +9,8 @@ class SpecialChoose < Engine::Step::Base
       ACTIONS = %w[choose_ability].freeze
 
       def actions(entity)
+        return [] unless entity.company?
```
tobymao, line 12: does this break any other games?

Issue #7803, opened 2022-05-20
1822mx - can’t buy NDEM when at cert limit 


fsaglio:
Game link: https://18xx.games/game/84565

we cannot buy shares of the NDEM while at cert limit.   NDEM shares don’t count against limit, so should be able to be purchased.

Issue #7781, opened 2022-05-16
1822MX: Ferry connections missing from map


waaayback:
The dashed ferry connection lines to Baja California per rule 6.4.7 are not on the map.

You might also want to check if these connections (for minor acquisition) have been coded (i.e. is it just a map oversight)

outsidePasser:
The connections have been coded.  It is just a map oversight, because nothing exists yet to make these lines easily.

outsidePasser:
Anchor icons are being added to the map to signify the connections instead of the dashed lines.  Note added to wiki

PR #7824, opened 2022-05-25
[1822MX] Bug fixes

outsidePasser:
Make NDEM corporation purchasable if a player is at the share limit 

Fixes: #7803

Add anchor icons to signify ferry connections in Gulf of California

Fixes: #7781

Fix number of slots in gray tiles

Fixes: #7828


outsidePasser:
I tried just this first:

image

but rubocop doesn't like it:

image

PR #7838, opened 2022-05-26
18GB 2E Updates

andypymont:
Updates as per designer Dave Berry:

- The LM private company should give a £20 bonus to Liverpool, not £10.
- The second-edition playtest end-game variant is being updated to trigger with 2 or fewer (instead of 3 or fewer) trains remaining at the end of an OR, and to only block token placement and not also share sales as per the previous version of this rule.


andypymont:
The change to the 2E end-game rules will be a breaking change for games which already reached the (previous) end-game restrictions. Checking live games at the time of writing, this should hopefully only be games that are already finished, but that could change by the time of deployment.

PR #7844, opened 2022-05-27
18usa - fix toronto gray value

scottredracecar:
References:
https://boardgamegeek.com/image/3477275/18usa
https://boardgamegeek.com/image/6718800/18usa


ventusignis (approved):
It looks like since Toronto's gray value changes and 18USA has revenue-dependent multi-jumps, some games may need to be pinned although I don't know how many as it's only the endgame that's affected

PR #7853, opened 2022-05-30
[18NY, 18USA] Bug fixes

### LINKED ISSUE #7777: 18USA: Autopass does not work
Stock round autopass does not pass in 18USA

### LINKED ISSUE #7800: [18NY] Minors shouldn't be able to scrap their last train and then pass

### LINKED ISSUE #7801: [18NY] Remove river cost from offboard hexes.
The offboard hex is never upgraded, so the cost only needs to be on the adjoining hex.

### LINKED ISSUE #7821: [18NY v2] Issueing shares during EMR
**What's your Game ID?**
https://18xx.games/game/85175?action=479

**How to Reproduce the Bug?**
NY&H does not have enough money to afford a new train ($180 in company, train costs $300.)
EMR can be triggered.
This can be done in multiple ways (rules 7.8.2).
One option is to issue shares as needed, at current share value:
> Sell shares from the Company Treasury, subject to the normal rules that no more shares may be in the Bank Pool than are held by players, to a maximum of five shares. The stock price is adjusted as usual. **The Company may not sell shares beyond those used to buy the train.**

NY&H needs to raise $120 to afford the next available train from the bank.
Current share price is $90.

Expected result:
NY&H is allowed to issue up to 2 shares to buy the most expensive available train.

Observed result:
NY&H is able to issue up to 3 shares:
image

Expected Root Cause:
This is likely due to the EMR rules allowing issuing of shares to buy a more expensive train than the cheapest one currently available:
From the rule clarification notes on the Wiki:
> Corporations must buy the cheapest train available during Emergency Money Raising if they take a loan or the president contributes cash; however, if they only issue shares, then they can buy any available train.

However in this situation there is no cheaper train, only the 6H in the bank pool for 300.
This is likely an artifact from the previous edition where scrapped/discarded trains would go to the bankpool, in 2E these trains are removed from the game instead and so never will become available in the bankpool.

### LINKED ISSUE #7850: [18USA] Perhaps a problem starting companies
In the game https://18xx.games/game/82814, I have just started ATSF and I had not the option to use privates as part of the starting capital. This created a strange situation, where I have negative cash.

crericha:
Fixes #7821 
Fixes #7801 
Fixes #7800 
Fixes #7777
Fixes #7850

18NY will need pins. 18USA does not.

### DIFF
--- a/lib/engine/game/g_18_ny/game.rb
+++ b/lib/engine/game/g_18_ny/game.rb
@@ -1003,8 +1003,11 @@ module Engine
 
         def emergency_issuable_bundles(corp)
           bundles = bundles_for_corporation(corp, corp)
+          cash_needed = @depot.min_depot_train.variants.map { |_, v| v[:price] }.max - corp.cash
+          return [] if cash_needed.negative?
 
-          num_issuable_shares = [5, corp.num_player_shares].min - corp.num_market_shares
+          max_issuable_shares = [5, corp.num_player_shares].min - corp.num_market_shares
+          num_issuable_shares = [max_issuable_shares, (corp.share_price.price / cash_needed.to_f).ceil].min
           bundles.reject { |bundle| bundle.num_shares > num_issuable_shares }.sort_by(&:price)
         end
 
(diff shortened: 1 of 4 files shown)


H1llsman:
Hey to confirm with the 18NY emergency issue fix:
Is this only for the 2nd Edition variant, or also for the base game?

PR #7877, opened 2022-06-03
1877: Stockholm Tramways implemented auctioning of corporations and starting phases of corporations.

lbrande:
I had to fix a bug in `buy_sell_par_shares_via_bid.rb` to make it also work with games with a `HOME_TOKEN_TIMING` other than `:par`. This required changing the `HOME_TOKEN_TIMING` of 1861/1867 to `:par`, as it was incorrectly set to `:float` and it only worked correctly due to the bug I fixed.

### DIFF
--- a/lib/engine/game/g_1877_stockholm_tramways/game.rb
+++ b/lib/engine/game/g_1877_stockholm_tramways/game.rb
@@ -4,6 +4,9 @@ require_relative 'meta'
 require_relative '../base'
 require_relative 'entities'
 require_relative 'map'
+require_relative 'step/buy_sell_par_shares'
+require_relative 'step/buy_train'
+require_relative 'step/dividend'
 
 module Engine
   module Game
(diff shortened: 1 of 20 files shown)


perwestling (commented):
What is s1?

lib/engine/game/g_1877_stockholm_tramways/game.rb
```diff
@@ -139,10 +143,58 @@ class Game < Game::Base
            'sl_trigger' => ['SL Trigger', 'SL will form at end of OR, game ends at end of following OR set'],
          ).freeze
 
+        def setup
+          @sl = nil
```
perwestling, line 147: What is s1?

lbrande: SL is the name of the end of game event, like LNER in 1862. I have however not implemented it's logic yet.

lbrande:
> What is s1?

See my answer above.

Issue #7738, opened 2022-05-11
FutureLabels for everyone


ventusignis:
![image](https://user-images.githubusercontent.com/2993555/167867414-3e382fd1-46cc-4bdf-83de-4202bd878fd5.png)

- 1822 : https://github.com/tobymao/18xx/pull/7695
- [x] T cities
- [x] S city
- [x] 1822MRS
- 1822mx
- [x] T cities
- 1856 https://github.com/tobymao/18xx/pull/7739
- [x] P cities
- [ ] H hex
- [ ] rendering issue
- [x] B-L -> Bar/Lon
- 1836jr56
- [x] P cities
- [x] H hex
- [x] B-L -> Bar/Lon
- 18Dixie https://github.com/tobymao/18xx/pull/7683
- [x] Like, all the cities
- 18VA https://github.com/tobymao/18xx/pull/7740
- [x] Normal -> P -> Wash / Rich, or nothing
- 18JP-T
- [x] Gray KU
- 18Scan
- [ ] plain yellow -> Y
- [ ] Oslo?
- 18MEX
- [x] Guadalajara
- 18AL
- [x] Montgomery
- 18TN
- [x] Nashville
- [x] C/N -> P
- 1870
- [x] Kansas City
- [x] St Louis
- 1861
- [x] Karkhov Y -> Kh
- 18MS
- [x] Mobile -> Mob
- [x] Birmingham/Montgomery -> BM
- 18NY
- [x] Brooklyn -> BR
- [x] Syracuse -> S
- 1862
- [x] Ipswich -> I
- [x] Harwich -> H
- 1871
- [x] T
- [x] CX

Assorted issues:
1856: Adding a future label to Hamilton leads to rendering issues
1870: could use double-future labels for plain -> P -> KC double-fork
![image](https://user-images.githubusercontent.com/2993555/167870296-5d78cc77-0909-4cc9-ad58-a774e7e44514.png)

outsidePasser:
1822MX was completed with #7760

andypymont:
1861 and 1867 completed in #7944

crericha:
This issue is mostly done; however, it has grown stale. Closing for now. Someone can reopen if they want to work on the remaining items.

PR #7898, opened 2022-06-10
Use future_labels for 1870

daniel-sousa-me:
#7738

### DIFF
--- a/lib/engine/game/g_1870/step/track.rb
+++ b/lib/engine/game/g_1870/step/track.rb
@@ -2,6 +2,7 @@
 
 require_relative '../../../step/tracker'
 require_relative '../../../step/track'
+require_relative '../../../part/future_label'
 
 module Engine
   module Game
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1870/step/track.rb
```diff
@@ -18,21 +19,15 @@ def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
...
 
-            if action.tile.hex.id == 'B11'
-              action.tile.label = 'K P'
-            elsif action.tile.hex.id == 'C18'
-              action.tile.label = 'P L'
-            elsif %w[J3 J5 N17].include?(action.tile.hex.id)
-              action.tile.label = 'P'
+            case action.tile.hex.id
+            when 'B11'
+              action.tile.future_label = Engine::Part::FutureLabel.new('K', 'gray')
```
ventusignis, line 28: I merged your changes into the 18Mex future labels PR and rack rake passes with both sets together.

I looked back at 1870 and the only way I can see to make it to avoid this kind of logic in game code would require rendering changes to make it possible to display more than one LargeIcon and show more than one FutureLabel for Kansas City and St. Louis.

In the meantime it might be worth changing the K and L labels to "KC" and "StL" to help link the gray tiles to which map hexes they end up going on before the brown P tile gets laid

tobymao:
@ventusignis what's the status of this what am i supposed to do

ventusignis:
The PR looks good to merge to me, I just had some nitpicking thoughts on it

PR #7923, opened 2022-06-14
Create New Game page: replace dropdown with a table, add input to filter by keyword

### LINKED ISSUE #6900: Create new game displays wrong game info
There is a very minor bug where the wrong game info will display when creating a new game.

Repro steps:
1. click "create a new game"
2. change the game title to any game other than the default
3. navigate to previous page (browser back button)
4. click "create a new game" again
5. observe wrong game variant/optional rules and game info are displayed for the default 1817 game

image

### LINKED ISSUE #7493: All prototype games should be in prototype section
In the game creation dropdown, some prototype titles are in the prototype section, while others are in the alpha section. That's because the sorting code in [`create_game.rb`](https://github.com/tobymao/18xx/blob/e2b672d9ce33569d8024461ddee1806ef8312338/assets/app/view/create_game.rb#L89) only puts titles in the prototype section if `DEV_STAGE == production` (which seems at odds with being a prototype anyway!).

I suggest that all prototype games be in the prototype section, and that those games' entries can get an additional `[alpha]` or `[beta]` (instead of the current `[Prototype]`) if additional information really needs to be conveyed (although honestly anyone playing a prototype already should know what they're in for).

michaeljb:
* no title is selected by default
    * fixes #6900; when taking the steps to reproduce that issue, the new
      behavior is the last title that was selected is still selected, instead of
      1817 being selected but showing the wrong game variants / optional rules
* game setting inputs (async/live, player count, etc) are not shown until a
  title is selected
* the long dropdown is replaced with a table showing game titles, location, and
  status (prod/beta/alpha, and prototype)
* keyword search allows users to narrow games shown in table not just by title,
  but by publisher, designer, etc.
* group all prototype games at the end, fixes #7493
* when a game is selected, scroll to top to start picking game settings

## Screenshots
Screenshot from 2022-06-14 11-15-11

Search by "1817" to find 1817 and games inheriting from it:

Screenshot from 2022-06-14 11-15-39

Search by designer:

Screenshot from 2022-06-14 11-15-49

Search by variant name:

Screenshot from 2022-06-14 11-16-07

Selected game is highlighted in the table list:

Screenshot from 2022-06-14 11-16-29

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -576,21 +576,7 @@ module View
       end.compact
     end
 
-    def render_game_row_title(title, game_meta)
-      style = {
-        width: '150px',
-        margin: '0',
-      }
-      if selected_game && game_meta == (selected_game::GAME_IS_VARIANT_OF || selected_game)
-        style['background-color'] = 'lightblue'
-      end
-
-      render_button(title, style: style) do
-        select_game(game_meta)
-      end
-    end
-
-    def render_games_table(cols = %w[location status])
+    def render_games_table(cols = %w[title location status])
       selected_games = game_rows_data.sort_by { |g| [g['meta']::PROTOTYPE ? 1 : 0, g['meta']] }
       selected_games = selected_games.select { |g| %i[alpha beta production].include?(g['meta']::DEV_STAGE) } if @production
 
@@ -606,11 +592,17 @@ module View
         end
       end
 
-      props = { style: { 'vertical-align': 'middle' } }
+      td_props = { style: { 'vertical-align': 'middle' } }
       game_rows = selected_games.map do |game|
-        h(:tr, [
-            h(:td, props, [render_game_row_title(game['title'], game['meta'])]),
-          ] + cols.map { |col| h(:td, props, game[col].to_s) })
+        row_style = { cursor: 'pointer' }
+        if selected_game && game['meta'] == (selected_game::GAME_IS_VARIANT_OF || selected_game)
+          row_style['background-color'] = 'lightblue'
+          row_style['color'] = 'black'
+        end
+
+        row_props = { on: { click: -> {select_game(game['meta'])}}, style: row_style }
+
+        h(:tr, row_props, cols.map { |col| h(:td, td_props, game[col].to_s) })
       end
 
       props = { style: { 'text-align': 'left' } }
@@ -619,7 +611,7 @@ module View
         {},
         [
           h(:thead, [
-              h(:tr, [h(:th, { style: { 'text-align': 'center' } }, 'Title')] + cols.map { |col| h(:th, props, col.capitalize) }),
+              h(:tr, cols.map { |col| h(:th, props, col.capitalize) }),
             ]),
           h(:tbody, game_rows),
         ]
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -478,8 +478,8 @@ module View
         min_players = min_p
       else
         # Letters resolve to 0 when converted to integers
-        max_players = max_players_elm.value.to_i.zero? ? nil : max_players_elm.value.to_i if max_players_elm
-        min_players = min_players_elm.value.to_i.zero? ? nil : min_players_elm.value.to_i if min_players_elm
+        max_players = (val = max_players_elm&.value.to_i).zero? ? nil : val
+        min_players = (val = min_players_elm&.value.to_i).zero? ? nil : val
       end
 
       if max_players
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -339,7 +339,7 @@ module View
 
     def mode_input(mode, text)
       click_handler = lambda do
-        store(:mode, mode, skip: !!selected_game_or_variant)
+        store(:mode, mode, skip: !selected_game_or_variant.nil?)
         update_inputs
       end
 
@@ -470,15 +470,13 @@ module View
       title = selected_game_or_variant.title
       max_p = @max_p[title]
       min_p = @min_p[title]
-      if selected_game_or_variant
-        max_players_elm = Native(@inputs[:max_players])&.elm
-        min_players_elm = Native(@inputs[:min_players])&.elm
-      end
+      max_players_elm = Native(@inputs[:max_players])&.elm
+      min_players_elm = Native(@inputs[:min_players])&.elm
 
       if title_change
         max_players = max_p
         min_players = min_p
-      elsif selected_game_or_variant
+      else
         # Letters resolve to 0 when converted to integers
         max_players = max_players_elm.value.to_i.zero? ? nil : max_players_elm.value.to_i if max_players_elm
         min_players = min_players_elm.value.to_i.zero? ? nil : min_players_elm.value.to_i if min_players_elm
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -579,14 +579,13 @@ module View
     end
 
     def render_game_row_title(title, game_meta)
-      style =
-        if selected_game && game_meta == (selected_game::GAME_IS_VARIANT_OF || selected_game)
-          {
-            'background-color': 'lightblue',
-          }
-        else
-          {}
-        end
+      style = {
+        width: '150px',
+        margin: '0',
+      }
+      if selected_game && game_meta == (selected_game::GAME_IS_VARIANT_OF || selected_game)
+        style['background-color'] = 'lightblue'
+      end
 
       render_button(title, style: style) do
         select_game(game_meta)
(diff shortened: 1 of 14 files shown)


roseundy:
@michaeljb can you check to see if 1825, 1888 and Rolling Stock still work with the refactored code? Those all required changes to create_game. I don't see any obvious problems, but still...

michaeljb:
I think Rolling Stock looks good, checking out 1888 and 1825 next....

1) search for "rolling stock"
Screenshot from 2022-06-14 11-34-49

2) select "Rolling Stock", the main title shown is "Rolling Stock Stars", and the variant for Stars is selected

Screenshot from 2022-06-14 11-34-55

3) uncheck the Stars checkbox, the main title is now "Rolling Stock", and the "Short Game" option is now available

Screenshot from 2022-06-14 11-35-06

michaeljb:
@roseundy 1888 has the north map checked by default, unchecking it shows the warning

michaeljb:
@roseundy I've done a little bit of clicking around with 1825 and it seems responsive, I don't know it well enough to thoroughly test it though. Various combinations of units/regionals/kits/DB variants are showing various info/errors. Anything specific I should try out?

crericha:
I recommend a status for the production games too. In my opinion it's weird they are blank.

michaeljb:
@crericha pushed in 9993519f4, here's how that looks:
Screenshot from 2022-06-14 11-51-09


Screenshot from 2022-06-14 11-51-18

roseundy:
> @roseundy I've done a little bit of clicking around with 1825 and it seems responsive, I don't know 1825 well enough to thoroughly test it though. Various combinations of units/regionals/kits/DB variants are showing various info/errors. Anything specific I should try out?

Nope, I think you've covered it.

public/assets/main.css
```diff
@@ -425,3 +425,24 @@ div#game:focus {
     float: left;
   }
 }
+
+table.create-game-table thead tr th {
```
tobymao, line 429: why did you need to add this?

tobymao:
closing this for now, michael feel free to reopen when you're ready

public/assets/main.css
michaeljb: Without `text-align: left`:

Screenshot 2022-11-27 122410

With it:

Screenshot 2022-11-27 122312

michaeljb:
I've rebased onto the latest master, got rid of a few things from my original PR that are now covered/fixed by #8133 and #8144, so all of the 1825/etc stuff should be good.

public/assets/main.css
tobymao: why didn't you just add it inline? why does it need to be in the css?

assets/app/view/create_game.rb
```diff
@@ -545,5 +520,146 @@ def update_inputs(title_change: false)
...
+      selected_games = game_rows_data.sort_by { |g| [g['meta']::PROTOTYPE ? 1 : 0, g['meta']] }
+      selected_games = selected_games.select { |g| %i[alpha beta production].include?(g['meta']::DEV_STAGE) } if @production
+
+      if @keywords&.size&.positive?
+        searches = @keywords.split(/[:, ]+/).map(&:upcase)
+
+        selected_games = selected_games.select do |game|
+          searches.all? do |search|
+            game['meta'].keywords.any? do |keyword|
+              keyword =~ /#{search}/
```
tobymao, line 606: would jaro winkler be better here?

lib/engine/game/g_1867/meta.rb
```diff
@@ -17,6 +17,8 @@ module Meta
         GAME_PUBLISHER = :grand_trunk_games
         GAME_RULES_URL = 'https://boardgamegeek.com/filepage/212807/18611867-rulebook'
 
+        KEYWORDS = ['1861'].freeze
```
crericha, line 20: Why inside of 1867?

assets/app/view/create_game.rb
```diff
@@ -370,7 +339,7 @@ def mode_selector
 
     def mode_input(mode, text)
       click_handler = lambda do
-        store(:mode, mode, skip: true)
+        store(:mode, mode, skip: !!selected_game_or_variant)
```
crericha, line 342: is this clearer than `selected_game_or_variant.nil?`?

assets/app/view/create_game.rb
```diff
@@ -494,33 +465,37 @@ def filtered_rule?(rule)
...
 
     def update_inputs(title_change: false)
+      return unless selected_game_or_variant
+
       title = selected_game_or_variant.title
       max_p = @max_p[title]
       min_p = @min_p[title]
-      max_players_elm = Native(@inputs[:max_players]).elm
-      min_players_elm = Native(@inputs[:min_players]).elm
+      if selected_game_or_variant
```
crericha, line 473: Returns on line 468 if selected_game_or_variant is false, so I don't think this check is needed.

assets/app/view/create_game.rb
```diff
@@ -494,33 +465,37 @@ def filtered_rule?(rule)
...
+        max_players_elm = Native(@inputs[:max_players])&.elm
+        min_players_elm = Native(@inputs[:min_players])&.elm
+      end
 
       if title_change
         max_players = max_p
         min_players = min_p
-      else
+      elsif selected_game_or_variant
```
crericha, line 481: Same comment as above.

assets/app/view/create_game.rb
```diff
@@ -494,33 +465,37 @@ def filtered_rule?(rule)
...
         # Letters resolve to 0 when converted to integers
-        max_players = max_players_elm.value.to_i.zero? ? nil : max_players_elm.value.to_i
-        min_players = min_players_elm.value.to_i.zero? ? nil : min_players_elm.value.to_i
+        max_players = max_players_elm.value.to_i.zero? ? nil : max_players_elm.value.to_i if max_players_elm
+        min_players = min_players_elm.value.to_i.zero? ? nil : min_players_elm.value.to_i if min_players_elm
```
crericha, line 484: max_players_elm and min_players_elm will always exist if that if on line 473 can be removed

crericha: Also the &.value below (492 and 498)

crericha:
FWIW I think the game buttons are kind of ugly. What if the title wasn't a button and instead when you click on a row, it is highlighted somehow (border maybe)?

Also, when a game is selected (say way down in the list), what happens? Does the page move to the top, so the user can enter the game details and hit the create button?

lib/engine/game/g_1867/meta.rb
michaeljb: In the code 1861 inherits from 1867, but in the real world, 1867 derived from 1861 so I wanted 1867 to show up when you search for 1861, and this seemed the easiest way to do that.

Maybe it makes more sense to always include a game when you search for any game it derives from, don't know if that would be too complicated to set up yet, but it would let me get rid of this one-off `KEYWORDS`. With the current state of the code, searching for "1846" will show you (among other games) 18 Los Angeles, but searching for "18 Los Angeles" won't show you 1846.

assets/app/view/create_game.rb
```diff
@@ -545,5 +520,146 @@ def update_inputs(title_change: false)
...
+
+      uncheck_optional_rules
+      @optional_rules = []
+
+      preselect_variant(@selected_game)
+      preselect_optional_rules(@selected_game)
+
+      update_inputs(title_change: true)
+
+      `window.scroll(0, 0)`
```
michaeljb, line 541: re: https://github.com/tobymao/18xx/pull/7923#issuecomment-1331657847

> FWIW I think the game buttons are kind of ugly. What if the title wasn't a button and instead when you click on a row, it is highlighted somehow? That's what the spreadsheet and info tab do, so it's a little more on brand.

I like this suggestion, I'll try that out.

> Also, when a game is selected (say way down in the list), what happens? Does the page move to the top, so the user can enter the game details and hit the create button?

Yep, here at the end of `select_game()` the user is sent to the top of the page

public/assets/main.css
michaeljb: Original PR was too long ago for me to remember the why, but it works just fine inline 🙃 Fixed in 729f04a55

assets/app/view/create_game.rb
michaeljb: If a game is selected in multiplayer/hotseat, then the mode is changed to "Import hotseat game", when switching the mode back to multiplayer or hotseat, `min_players_elm` and `max_players_elm` are both `nil`.

They're also `nil` when the new game page is first loaded, and no game is yet selected; when the first game is selected, without the `&.value` on 492 and 498 those lines error out. This is new behavior since on the current `master` there is always a game selected.

I pushed up a little bit nicer version of 483 and 484 in 59ce2b9a3

tobymao (approved):

assets/app/view/create_game.rb
michaeljb: Here's how it looks with the button removed and the row highlighting the active game:

Screenshot 2022-12-03 202637

assets/app/view/create_game.rb
michaeljb: I think not; jaro winkler would assign a score to how far off each keyword is, so we'd also have to come up with some minimum score a game must reach in order to filter the list. With the current logic every game is just a match or it isn't.

assets/app/view/create_game.rb
michaeljb: With this setup the whole row is clickable, but it doesn't really look like it unless you're on a computer and can see the cursor change on hover. 🤷

michaeljb: @crericha I'll go ahead and merge like this, we can make further UI tweaks later if necessary

PR #7956, opened 2022-06-18
Break shareholder ties by presidency transfer order.

### LINKED ISSUE #5374: Displayed order of shareholders should match presidency transfer
I want the displayed order of shareholders on the corporation card to match the order in which presidency would transfer? (and I think that when there’s a tie in number of shares, this order is usually priority order?)

When it is not, although the rules are followed, I find the display confusing. For example:

If I see a company box that looks like this:
A: 5
B: 2
C: 2
I expect that B and C are ordered by who gets the presidency next. Meaning if A sells 4-5 shares, B gets presidency. In this game, C got the presidency. Note that this is correct. C Should have gotten the presidency, but I would have expected the order to display like this:
A: 5
C: 2
B: 2
instead.

See https://18xx.games/game/40790 as a situation where cwmassey sold 5 shares of GNRI and made kaelistus president. Display order was confusing and I didn't expect that.

andrewzwicky:
fixes #5374

For the player rows portion only, re-order based on priority order.  This means that the listed order should reflect the way that the presidency would be transferred if the president sold.

Tested with sample games from:
1889
1846
Rolling Stock Stars
1849
1822
1862
21Moon

Old:
image

New:
image

(This is just showing how ties are broken, I know it can't actually transfer in this case because two shares, etc.)


assets/app/view/game/corporation.rb
```diff
@@ -311,14 +311,28 @@ def share_number_str(number)
...
+      def entities_rows(entities, pres_transfer_order = false)
         step = @game.round.active_step
+
+        if pres_transfer_order
+          # For player rows, create an array of names, rotate that array so
+          # the president is at index 0, and use the resulting order to sort
+          # player names when they have the same number of shares. This results
+          # in the listed order showing the correct transfer of presidency.
+          players_in_order = entities.map(&:name)
+          unless @corporation.owner.nil?
```
tobymao, line 323: if @corporation.owner

assets/app/view/game/corporation.rb
```diff
@@ -311,14 +311,28 @@ def share_number_str(number)
...
+            player_index = players_in_order.find_index(@corporation.owner.name)
```
tobymao, line 324: index

assets/app/view/game/corporation.rb
```diff
@@ -311,14 +311,28 @@ def share_number_str(number)
...
+            players_in_order = players_in_order.rotate(player_index) unless player_index.nil?
```
tobymao, line 325: if player_index

PR #7963, opened 2022-06-20
1849: Added notes on share price movement for shares in pool, shares sold, …

drewww:
…or sold out.


drewww:
It's my first PR -- let me know if this is bad etiquette or whatever. Just getting my feet wet.

PR #7991, opened 2022-06-23
Adding price movement charts for 1822 + variants

drewww:


lib/engine/game/g_1822/game.rb
```diff
@@ -1086,7 +1086,7 @@ def route_trains(entity)
           entity.runnable_trains.reject { |t| pullman_train?(t) }
         end
 
-        def setup
+         def setup
```
tobymao, line 1089: is your indentation off?

docker-compose exec rack rubocop -A

drewww: Ah, yes. Fixed + will run the linter before PRs going forward. Missed that note in the PR guidelines. My bad.

PR #8007, opened 2022-06-25
18SJ: Implement changes for alpha

perwestling:
1. Meta changes:
   - Change to alpha
   - remove version from title
   - Set as AAG game and use pre-version of rules
2. Update region to Sweden
3. Base game: Full cap corporation may now redeem
4. 2 172 tiles replaced with 1 171 tile.
5. 2 player variant is now automatic if 2 player chosen
   and not available as a choice
6. Remove broken spec game (tile change break game)


perwestling:
I suspect this commit to break most 18SJ games. As 18SJ is prototype status I do not think we need to bother that much with them, so archive finished games, pin non-finished with problems, would be more than enough.
Right now it is 22 games active, and some of them haven't gone that far.

PR #8008, opened 2022-06-25
refactor game opening button text "Enter" => 'View'

benjaminxscott:
Before:
image

After:
image

### DIFF
--- a/assets/app/view/chat.rb
+++ b/assets/app/view/chat.rb
@@ -32,7 +32,7 @@ module View
       enter = lambda do |event|
         event = Native(event)
 
-        if event['key'] == 'Enter'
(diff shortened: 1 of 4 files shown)


crericha:
@benjaminxscott, why did you change the button for the player's active games to be 'View'? I think 'Enter' still makes sense for the games that the player is in and 'View' for active games the player is not in.

benjaminxscott:
Summary of convo in slack:
- Proposal from @crericha to revert this PR and implement a new feature to differentiate b/t games that a player is in in order to have different text for those (e.g. "Enter" if player is in the game, "View" otherwise)
- @dfannius , myself, and I believe @crericha would prefer the above behavior

PR #8016, opened 2022-06-26
[1871] Players couldn't run companies for UB (fixes #7987)

### LINKED ISSUE #7987: [TOP 1871] UB owner doesn't have control to run a UB-owned corporation
I have never played TOP 1871, so I don't know how accurate this is, just entering it on behalf of someone else.

Apparently when a corporation owned by the UB has an OR turn, the player who owns the UB is not given control in order to run the corporation. The workaround is to use master mode.

dfannius:


dfannius:
I just noticed that even though the UB owner can now make moves for UB, those moves show up as having been executed in master mode for other players. Just a visual wart but should be fixed as part of this PR. Looking into that now.

dfannius:
OK, that part is fixed.

lib/engine/game/base.rb
```diff
@@ -584,7 +584,7 @@ def pass_entity(_user)
       end
 
       def active_players
-        players_ = @round.active_entities.map(&:player).compact
+        players_ = @round.active_entities.map(&:player).map { |p| acting_for_player(p) }.compact
```
tobymao, line 587: don't do a double map, just map once

PR #8024, opened 2022-06-28
Add support for Google Chat webhook.

magnusrk-ch:
Google Chat requires the format <users/123> to notify users rather than the more common <@123>.


tobymao:
i assume you tested this?

magnusrk-ch:
Yes, I commented out the notification disabling on my local server, triggered some notifications.

Actually, on that note, what do you think about allowing webhook notifications from local servers by default? It doesn't require any config or quota like email notifications do.

tobymao:
id rather not

PR #8026, opened 2022-06-29
[1828] Explain choices during merge

dfannius:
A few choices had no explanation of what choice you were making. Also made punctuation more consistent.

### DIFF
--- a/lib/engine/game/g_1828/step/merger.rb
+++ b/lib/engine/game/g_1828/step/merger.rb
@@ -332,7 +332,7 @@ module Engine
                 "#{c.name} (#{@game.format_currency(c.share_price.price)})"
               end
               @log << "#{entity.name} chooses which odd share to retain for trading up"
-              @player_choice = PlayerChoice.new(step_description: 'Choose share to retain after 2:1 exchange',
+              @player_choice = PlayerChoice.new(step_description: 'Choose odd share to retain for trading up',
                                                 choice_description: 'Choose share',
                                                 choices: choices)
               nil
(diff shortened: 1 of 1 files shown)


crericha:
The choices are in the step description

```
          def description
            description = 'Merger - '
            description += @player_choice.step_description if @player_choice
            description += "Select corporation to merge with #{mergeable_entity.name}" if @state == :select_target
            description
          end
```

lib/engine/game/g_1828/step/merger.rb
```diff
@@ -331,6 +331,7 @@ def identify_odd_share(entity)
               choices = merging_corporations.map do |c|
                 "#{c.name} (#{@game.format_currency(c.share_price.price)})"
               end
+              @log << "#{entity.name} chooses which odd share to retain for trading up"
               @player_choice = PlayerChoice.new(step_description: 'Choose share to retain after 2:1 exchange',
```
crericha, line 335: Can you update the step_description to be similar to the log message?

lib/engine/game/g_1828/step/merger.rb
```diff
@@ -365,8 +366,9 @@ def exchange_singles(entity)
               @exchange_selection = nil
 
               @log << "#{entity.name} discards 1 #{share.corporation.name} share and receives " \
-                      "#{@game.format_currency(share.price)} from the bank."
+                      "#{@game.format_currency(share.price)} from the bank"
             elsif entity.player? && !@exchange_selection
+              @log << "#{entity.name} chooses whether to trade up or sell single #{share.corporation.name} share"
               exchange_cost = share.price - @system.share_price.price
               @player_choice = PlayerChoice.new(step_description: 'Exchange or sell share',
```
crericha, line 373: Can you update the step_description to be similar to the log message?

dfannius:
Aha, despite having the code in front of me, I didn't realize that the information was in the step description above the log. I never thought to look all the way up there for information about what the generic instruction "Choose:" actually means, while I'm always looking at the log to see what's going on.

How would you feel about expanding the "Choose:" button slightly (e.g., "Choose odd share to keep", "Choose to trade up or sell"), without any other changes (so removing these log changes)?

crericha:
My first implementation attempt was putting the text in the buttons, but I felt it looked weird. The buttons were really long. If you want to make another attempt at it, that's fine by me.

My personal preference is to add the log messages and sync up the step description to have similar wording. The new log messages are better than what I wrote.  That way, wherever the player looks, they'll get context for the decision.

dfannius:
Ah, okay, I thought you were implicitly telling me that my added log lines were redundant and I should just update the step description. I'll try to find something that feels good to me.

dfannius:
`step_description`s updated.

crericha (approved):

PR #8031, opened 2022-06-30
Validate correct auto actions on the server

### LINKED ISSUE #7813: Game state not fully reset after GameError for the player taking the illegal action
https://18xx.games/game/84989
Was placing a token in NY. It said the cost was $80 and failed
Placed a token in Albany and it worked but the game hung and now won't load
We have stopped playing the game, so don't need it recovered or working

crericha:
Fixes #7813

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -709,7 +709,7 @@ module Engine
             auto_actions.concat(actions)
           end
           if validate_auto_actions
-            raise GameError, "Auto actions do not match" unless auto_actions_match?(action.auto_actions, auto_actions)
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -766,6 +773,16 @@ def maybe_raise!
         self
       end
 
+      def auto_actions_match?(a, b)
+        return false unless a.size == b.size
```
tobymao, line 777: actions_a.zip(actions_b).all? do |a, b|
  a.to_h&.except('created_at') == b.to_h&.except('created_at')
end

crericha: Fixed

tobymao (approved):

PR #8036, opened 2022-07-01
1870: Add upwards arrow on left of the soft ledge

daniel-sousa-me:


daniel-sousa-me:
The code became a quite ugly. Suggestions to improve readability are welcome.

In case the message is not clear, here's the market with the new arrows:

image

andrewzwicky:
Does this happen to address this comment as well?  I didn't want to make the original implementation too complicated so I left that out in the first attempt.

https://github.com/tobymao/18xx/pull/7809#issuecomment-1135368820

daniel-sousa-me:
Maybe we can create a method in the game that given coordinates only says if it's the leftmost space.

This could also be used for the movement itself. Iirc currently each game has to reimplemen tthe entire stock move method.

PR #8052, opened 2022-07-06
Refactor 1870 soft ledge move. Use same logic for 1849

daniel-sousa-me:
The methods `move_down` and `move_up` on `lib/engine/game/g_1849/stock_market.rb` now are only there to write stuff to the log. Can we remove them?

### DIFF
--- a/lib/engine/stock_market.rb
+++ b/lib/engine/stock_market.rb
@@ -40,14 +40,18 @@ module Engine
       corporation.original_par_price = share_price
     end
 
+    def right_ledge?(r, c)
+      c + 1 >= @market[r].size
+    end
+
     def move_right(corporation)
       r, c = corporation.share_price.coordinates
 
-      if c + 1 < @market[r].size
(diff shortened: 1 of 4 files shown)


lib/engine/stock_market.rb
```diff
@@ -40,14 +40,18 @@ def set_par(corporation, share_price)
       corporation.original_par_price = share_price
     end
 
+    def reroute_right_move(r, c)
```
tobymao, line 43: i don’t understand this method name, what does it mean?

daniel-sousa-me: Given coordinates on the stock market it tells if we're at the rightmost spot, which means we should move up instead of right.

tobymao: def right_ledge?

daniel-sousa-me: Thanks. Updated

tobymao (approved):

PR #8055, opened 2022-07-07
[1848] - implement p5 and p6 logic

bentziaxl:
P5 -> The owner receives a 10% share in the QR. Cannot be bought by a corporation
P6 -> The owner receives a Director's Share share in the CAR, which must start at a par value of £100. Cannot be bought by a corporation

issue #2548 is showing "CAR is always parred at 100 (Presidency is granted by P6)" as implemented, but i don't think it was

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -249,6 +249,8 @@ module Engine
             discount: -170,
             revenue: 25,
             desc: 'The owner receives a 10% share in the QR. Cannot be bought by a corporation',
+            abilities: [{ type: 'shares', shares: 'QR_1' },
+                        { type: 'no_buy' }],
           },
           {
             sym: 'P6',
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1848/game.rb
```diff
@@ -435,6 +439,22 @@ def must_buy_train?(entity)
...
+
+        def after_buy_company(player, company, _price)
+          # share_price = 100
+          # # NOTE: This should only ever be P6
+          abilities(company, :shares) do |ability|
+            ability.shares.each do |share|
+              if share.president
+                stock_market.set_par(share.corporation, stock_market.par_prices.find { |p| p.price == 100 })
+                share_pool.buy_shares(player, share, exchange: :free)
+                after_par(share.corporation)
```
bentziaxl, line 451: after_par(share.corporation) checks if it's incremental cap game (it's not)
 and does floating operations (in this game you need more than prez share to float), so irrelevant
 So I could remove this line, but saw it in other games, wanted to be consistent

crericha: The code base changes quickly and after_par may grow functionality in the future that you'll need, so its best to leave it in there even if nothing happens for now.

bentziaxl: ok great, tx

PR #8056, opened 2022-07-07
[1848] implement some boe features

bentziaxl:
BOE is a corporation of type bank that has no president share, all the shares start in the pool, doesn't have a president but pays dividends to shareholders.

Implemented no president, starting shares in pool, 2000 starting capital

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -17,6 +17,8 @@ module Engine
 
         BANK_CASH = 10_000
 
+        BOE_START_CASH = 2000
+
         CERT_LIMIT = { 3 => 20, 4 => 17, 5 => 14, 6 => 12 }.freeze
 
         STARTING_CASH = { 3 => 840, 4 => 630, 5 => 510, 6 => 430 }.freeze
@@ -375,6 +377,7 @@ module Engine
               allow_president_change: false
             )
             @boe.owner = @share_pool
+            @boe.cash = BOE_START_CASH
           end
         end
 
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -364,7 +364,6 @@ module Engine
           @sydney_adelaide_connected = false
 
           @boe = @corporations.find { |c| c.type == :bank }
-
           @boe.ipoed = true
           @boe.ipo_shares.each do |share|
             @share_pool.transfer_shares(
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1848/game.rb
```diff
@@ -17,6 +17,8 @@ class Game < Game::Base
 
         BANK_CASH = 10_000
 
+        BOE_START_CASH = 2000
```
crericha, line 20: Nit: Recommend naming BOE_STARTING_CASH to match STARTING_CASH. I'll leave it up to you if you want to follow my recommendation.

lib/engine/game/g_1848/game.rb
```diff
@@ -361,6 +364,21 @@ class Game < Game::Base
...
+          @boe.ipoed = true
+          @boe.ipo_shares.each do |share|
+            @share_pool.transfer_shares(
+              share.to_bundle,
+              share_pool,
+              spender: share_pool,
+              receiver: @bank,
+              price: 0,
+              allow_president_change: false
+            )
```
crericha, line 378: You should only need the first two arguments.

crericha: Maybe allow_president_change: false too.

lib/engine/game/g_1848/game.rb
```diff
@@ -369,6 +387,15 @@ def new_auction_round
           ])
         end
 
+        def stock_round
+          Engine::Round::Stock.new(self, [
+            Engine::Step::DiscardTrain,
```
crericha, line 392: Is there a situation in 1848 where a corp could or would need to discard a train during the stock round?

lib/engine/game/g_1848/game.rb
```diff
@@ -369,6 +387,15 @@ def new_auction_round
...
+            Engine::Step::Exchange,
+            Engine::Step::SpecialTrack,
```
crericha, line 394: Also, is there a case for laying track in the SR?

crericha: Found it in the rules.

lib/engine/game/g_1848/game.rb
crericha: Found it in the rules.

lib/engine/game/g_1848/game.rb
bentziaxl: where?

lib/engine/game/g_1848/game.rb
bentziaxl: where in the rules? I just copied the super method and modified one step, didn't want to touch the others in case they are relevent

bentziaxl:
I'll need to implement not president double share later, it worked with all extra the params in the setup method, but i'm not sure why

PR #8061, opened 2022-07-10
[1848] add boe stock row and initial price

bentziaxl:
added bank of england stock row, and a custom method to find boe share price.
added some cosmetic changes such as game trigger text.
fixed a minor bug in setup where the boe owner and starting cash was defined inside the share selling for loop.
fixed boe double share

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -114,7 +114,7 @@ module Engine
 
         GAME_END_REASONS_TEXT = {
           bank: 'The bank runs out of money',
-          stock_market: 'Corporation hit max stock value or Bank of England has given 16 of more loans',
+          stock_market: 'Corporation hit max stock value or Bank of England has given 16 or more loans',
           custom: 'Fifth corporation is in receivership',
         }.freeze
 
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -530,9 +530,8 @@ module Engine
         end
 
         def lookup_boe_price(p)
-          @stock_market.market[BOE_ROW].size.times do |i|
-            next unless @stock_market.share_price(BOE_ROW, i)
-            return @stock_market.share_price(BOE_ROW, i) if @stock_market.share_price(BOE_ROW, i).price == p
+          @stock_market.market[BOE_ROW].each do |i|
+            return i if i.price == p
           end
         end
       end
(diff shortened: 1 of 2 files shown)


benjaminxscott:
typo "of more loans" 
https://github.com/tobymao/18xx/pull/8061/files#diff-f9b14b22c95c8c97908b4b327ff152658d3d149bdf56462e59d5b65812680356R117

bentziaxl:
fixed

bentziaxl:
ignore latest commit and revert, by accident

lib/engine/game/g_1848/game.rb
```diff
@@ -487,6 +528,13 @@ def after_buy_company(player, company, _price)
             end
           end
         end
+
+        def lookup_boe_price(p)
+          @stock_market.market[BOE_ROW].size.times do |i|
```
crericha, line 533: .each instead of .size.times

bentziaxl: i copied this method from somewhere else, using each is way easier, i wonder why it's not used everywhere for price lookups.

lib/engine/game/g_1848/game.rb
```diff
@@ -530,9 +530,8 @@ def after_buy_company(player, company, _price)
         end
 
         def lookup_boe_price(p)
-          @stock_market.market[BOE_ROW].size.times do |i|
-            next unless @stock_market.share_price(BOE_ROW, i)
-            return @stock_market.share_price(BOE_ROW, i) if @stock_market.share_price(BOE_ROW, i).price == p
+          @stock_market.market[BOE_ROW].each do |i|
```
crericha, line 533: It's best practice to be more descriptive in the block args to give context of what is being iterated. For example, here I'd used sp for share_price.

crericha: Let me know if you want to make that change before I merge.

crericha (approved):

lib/engine/game/g_1848/game.rb
crericha: I'm not sure how the .times snuck in. Usually Toby (the primary reviewer) is good at catching those.

lib/engine/game/g_1848/game.rb
bentziaxl: done. can merge

crericha: I was referring to the block variable - i, which represents an array of share prices. It makes it easier to maintain when the block variable gives context as to what it is. the method param was fine -- though price would have been more descriptive there.

bentziaxl: done

crericha (approved):

PR #8069, opened 2022-07-13
[1848] Implement privates, BoE market row, extra tile lay on phase 3

bentziaxl:
Implement privates, BoE market row, extra tile lay, and loans

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -3,6 +3,7 @@
 require_relative 'meta'
 require_relative 'map'
 require_relative '../base'
+require_relative '../../loan'
 
 module Engine
   module Game
@@ -162,7 +163,8 @@ module Engine
         TRAINS = [
           {
             name: '2',
-            distance: 2,
+            distance: [{ 'nodes' => %w[city offboard], 'pay' => 2, 'visit' => 2 },
+                       { 'nodes' => ['town'], 'pay' => 99, 'visit' => 99 }],
             price: 100,
             rusts_on: '4',
             num: 6,
@@ -172,17 +174,21 @@ module Engine
           },
           {
             name: '3',
-            distance: 3,
+            distance: [{ 'nodes' => %w[city offboard], 'pay' => 3, 'visit' => 3 },
+                       { 'nodes' => ['town'], 'pay' => 99, 'visit' => 99 }],
             price: 200,
             rusts_on: '6',
             num: 5,
             variants: [
               { name: '3+', distance: 3, price: 230 },
             ],
+            events: [{ 'type' => 'take_out_loans' },
+                     { 'type' => 'lay_second_tile' }],
           },
           {
             name: '4',
-            distance: 4,
+            distance: [{ 'nodes' => %w[city offboard], 'pay' => 4, 'visit' => 4 },
+                       { 'nodes' => ['town'], 'pay' => 99, 'visit' => 99 }],
             price: 300,
             rusts_on: '8',
             num: 4,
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1848/game.rb
```diff
@@ -172,17 +174,21 @@ class Game < Game::Base
...
           {
             name: '3',
-            distance: 3,
+            distance: [{ 'nodes' => %w[city offboard], 'pay' => 3, 'visit' => 3 },
+                       { 'nodes' => ['town'], 'pay' => 99, 'visit' => 99 }],
             price: 200,
             rusts_on: '6',
             num: 5,
             variants: [
               { name: '3+', distance: 3, price: 230 },
```
crericha, line 183: Did you forget the + variants or something you plan to do later?

lib/engine/game/g_1848/game.rb
```diff
@@ -249,21 +267,29 @@ class Game < Game::Base
...
+                      discount: 40,
+                      hexes: %w[B7 B9 C2 C4 C8 E6 E8],
+                      tiles: %w[7 8 9],
+                      count: 1,
+                      reachable: true,
+                      consume_tile_lay: true,
+                      owner_type: 'player',
+                      when: 'owning_player_or_turn',
+                    },
+                  ],
```
crericha, line 292: FYI when one of the abilities is used, you'll need a way to remove the other (or set the count to 0).

lib/engine/game/g_1848/stock_market.rb
```diff
@@ -0,0 +1,17 @@
...
+
+require_relative '../../stock_market'
+
+module Engine
+  module Game
+    module G1848
+      class StockMarket < Engine::StockMarket
+        def move_down(corporation)
+          r, c = corporation.share_price.coordinates
+          r += 1 if (r + 1 < @market.size || r + 1 != Game::BOE_ROW) && share_price(r + 1, c)
```
crericha, line 11: I don't believe Game::BOE_ROW will work. You'll need to pass the constant in via the constructor or something along those lines.

lib/engine/game/g_1848/game.rb
bentziaxl: something i plan to do later. In fact I already did in my forked repo :-)

lib/engine/game/g_1848/game.rb
bentziaxl: I'm aware, i put it in the do list.
My plan of attack is make the privates work, then fix the edge cases

lib/engine/game/g_1848/stock_market.rb
bentziaxl: I test all my changes. This works. Do you want me to change it to something else?

lib/engine/game/g_1848/game.rb
bentziaxl: The distance change came after a rule that i noticed, all the trains are unlimited plus trains, including the variant train.
The variant (+ train) just adds one free gauge change

lib/engine/game/g_1848/stock_market.rb
crericha: @tobymao how is Game::BOE_ROW working?

tobymao: import order, it’s bad , can you import it explicitly?

bentziaxl: image

bentziaxl: would G1848::Game::BOE_ROW be better? Or should I define a method to read the constant?
Or make it a readable attribute and do @game.boe_row instead?

bentziaxl: the rubocop style guide references using double colon to access class instance constants here:
https://github.com/rubocop/ruby-style-guide#double-colons

SomeModule::SomeClass::SOME_CONST

tobymao: yes and make sure to require the fike

crericha: game.rb requires stock_market.rb, so I don't think stock_market.rb can require game.rb - circular reference. I believe you'll need to pass BOE_ROW in as a constructor arg. 

Why pass in BOE_ROW instead of the game instance? We try to keep the game instance out of the basic classes to prevent too much game logic being stuffed into them.

tobymao: you can do circular references in ruby

bentziaxl: fixed

crericha (approved):

PR #8076, opened 2022-07-14
update docker cmd

benjaminxscott:
t looks like this is the modern behavior, as compose is included with Docker engine now
ref: https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command

tested on my local setup (debian 11, docker 20.10.17), worked fine


tobymao:
seems like this is unnecessary and production is only on 20.10.7 which doesn't work

benjaminxscott:
Is prod still running a version where this is the case? docker is up to `23.x` now

(noting this since I have to manually edit the Makefile to build locally)

tobymao:
haven't updated docker in prod

PR #8077, opened 2022-07-15
[1848] BOE dividend payout, more privates impl

bentziaxl:
Adding more 1848 impl:
1- p3 and the ability to place the tasmania token in one of two hexes on the map
2- define min max prices for privates
3-  boe dividend math (base pay + for each tokened city)
4- boe price doesn't rise when sold out, not market limit, etc

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -9,7 +9,7 @@ module Engine
   module Game
     module G1848
       class Game < Game::Base
-        attr_reader :sydney_adelaide_connected
+        attr_reader :sydney_adelaide_connected, :boe
 
         include_meta(G1848::Meta)
         include Map
@@ -255,6 +255,8 @@ module Engine
             name: "Melbourne & Hobson's Bay Railway Company",
             value: 40,
             discount: 10,
+            min_price: 1,
+            max_price: 40,
             revenue: 5,
             desc: 'No special abilities.',
           },
@@ -262,6 +264,8 @@ module Engine
             sym: 'P2',
             name: 'Sydney Railway Company',
             value: 80,
+            min_price: 1,
+            max_price: 80,
             discount: 10,
             revenue: 10,
             desc: 'Owning Public Company or its Director may build one (1) free tile on a desert hex (marked by'\
@@ -296,18 +300,53 @@ module Engine
             name: 'Tasmanian Railways',
             value: 140,
             discount: 30,
+            min_price: 1,
+            max_price: 140,
             revenue: 15,
             desc: 'The Tasmania tile can be placed by a Public Company on one of the dark blue hexes. This is in'\
                   " addition to the company's normal build that turn.",
+                  abilities: [
+                    {
+                      type: 'tile_lay',
+                      hexes: %w[I8 I10 D5],
+                      tiles: %w[241,1],
+                      owner_type: 'corporation',
+                      when: 'owning_corp_or_turn',
+                      count: 1,
+                      free: true,
+                    },
+                  ],
+
           },
           {
             sym: 'P4',
             name: 'The Ghan',
             value: 220,
             discount: 50,
+            min_price: 1,
+            max_price: 220,
             revenue: 20,
             desc: 'Owning Public Company or its Director may receive a one-time discount of £100 on the purchase'\
                   ' of a 2E (Ghan) train. This power does not go away after a 5/5+ train is purchased.',
+            abilities: [
+                    {
+                      type: 'train_discount',
+                      discount: 100,
+                      trains: ['2E'],
+                      count: 1,
+                      owner_type: 'corporation',
+                      when: 'any',
+                    },
+                    {
+                      type: 'train_discount',
+                      discount: 100,
+                      trains: ['2E'],
+                      count: 1,
+                      owner_type: 'player',
+                      when: 'any',
+                    },
+                  ],
+
           },
           {
             sym: 'P5',
@@ -490,6 +529,11 @@ module Engine
         def init_stock_market
           G1848::StockMarket.new(game_market, self.class::CERT_LIMIT_TYPES,
                                  multiple_buy_types: self.class::MULTIPLE_BUY_TYPES)
+        end
+
+        def operating_order
+          boe, others = @corporations.select(&:floated?).sort.partition { |c| c.type == :bank }
+          boe + others
         end
 
         def upgrades_to?(from, to, _special = false, selected_company: nil)
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -336,7 +336,7 @@ module Engine
                       trains: ['2E'],
                       count: 1,
                       owner_type: 'corporation',
-                      when: 'any',
+                      when: 'buying_train',
                     },
                     {
                       type: 'train_discount',
@@ -344,7 +344,7 @@ module Engine
                       trains: ['2E'],
                       count: 1,
                       owner_type: 'player',
-                      when: 'any',
+                      when: 'owning_player_or_turn',
                     },
                   ],
 
@@ -522,7 +522,7 @@ module Engine
             Engine::Step::Token,
             Engine::Step::Route,
             G1848::Step::Dividend,
-            Engine::Step::DiscardTrain,
+            Engine::Step::SpecialBuyTrain,
             G1848::Step::BuyTrain,
             [Engine::Step::BuyCompany, { blocks: true }],
           ], round_num: round_num)
@@ -531,6 +531,10 @@ module Engine
         def init_stock_market
           G1848::StockMarket.new(game_market, self.class::CERT_LIMIT_TYPES,
                                  multiple_buy_types: self.class::MULTIPLE_BUY_TYPES)
+        end
+
+        def init_share_pool
+          G1848::SharePool.new(self)
         end
 
         def operating_order
@@ -598,6 +602,10 @@ module Engine
         # for 3 players corp share limit is 70%
         def corporation_opts
           @players.size == 3 ? { max_ownership_percent: 70 } : {}
+        end
+
+        def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil)
+          super(bundle, allow_president_change: pres_change_ok?(bundle.corporation), swap: nil)
         end
 
         def pres_change_ok?(corporation)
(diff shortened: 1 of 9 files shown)


lib/engine/game/g_1848/game.rb
```diff
@@ -492,8 +533,18 @@ def init_stock_market
                                  multiple_buy_types: self.class::MULTIPLE_BUY_TYPES)
         end
 
+        def init_share_pool
+          G1848::SharePool.new(self)
+        end
+
+        def operating_order
+          boe, others = @corporations.select(&:floated?).sort.partition { |c| c.type == :bank }
+          boe + others
```
crericha, line 542: @corporations.select(&:floated?).sort.partition { |c| c.type == :bank }.flatten

lib/engine/game/g_1848/game.rb
```diff
@@ -492,8 +533,18 @@ def init_stock_market
...
+        end
+
         def upgrades_to?(from, to, _special = false, selected_company: nil)
           return %w[5 6 57].include?(to.name) if (from.hex.tile.label.to_s == 'K') && (from.hex.tile.color == 'white')
+          return ['241'].include?(to.name) if selected_company && selected_company.sym == 'P3'
```
crericha, line 547: if selected_company&.sym == 'P3'

lib/engine/game/g_1848/step/dividend.rb
```diff
@@ -0,0 +1,68 @@
...
+    module G1848
+      module Step
+        class Dividend < Engine::Step::Dividend
+          BOE_BASE_PAYOUT = { '2' => 0, '3' => 100, '4' => 100, '5' => 200, '6' => 200, '8' => 300 }.freeze
+
+          def actions(entity)
+            return super unless entity == @game.boe
+
+            # boe always pays if it has any revenue
+            process_dividend(Action::Dividend.new(entity, kind: 'payout'))
```
crericha, line 16: Do this as an auto_action. State should not change in the actions method.

lib/engine/game/g_1848/step/dividend.rb
```diff
@@ -0,0 +1,68 @@
...
+          end
+
+          def change_share_price(entity, payout)
+            return super unless entity == @game.boe
+          end
+
+          def dividend_options(entity)
+            return super unless entity == @game.boe
+
+            revenue = calculate_boe_revenue
```
crericha, line 31: Do you have dividend options for BOE? It always pays out. I think you can skip the options step.

crericha: I think 1840 does this.

lib/engine/game/g_1848/step/dividend.rb
```diff
@@ -0,0 +1,68 @@
...
+
+            payout_shares(entity, calculate_boe_revenue) if payout[:per_share].positive?
+            pass!
+          end
+
+          def calculate_boe_revenue
+            current_phase_name = @game.phase.current[:name]
+            base_pay = BOE_BASE_PAYOUT[current_phase_name]
+            cities_revenue = get_token_cities_total_revenue(@game.boe)
+            base_pay + cities_revenue
```
crericha, line 54: BOE_BASE_PAYOUT[current_phase_name] + get_token_cities_total_revenue(@game.boe)

lib/engine/game/g_1848/step/dividend.rb
```diff
@@ -0,0 +1,68 @@
...
+          end
+
+          def get_token_cities_total_revenue(corporation)
+            @game.hexes.sum do |hex|
+              hex.tile.cities.sum do |city|
+                city.tokened_by?(corporation) ? city.revenue[hex.tile.color] : 0
+              end
```
crericha, line 61: Going through all the hexes is time consuming and slow down game load time. If you cache the boe tokens, it can be much more efficient.

lib/engine/game/g_1848/step/operating.rb
```diff
@@ -10,6 +10,7 @@ class Operating < Engine::Round::Operating
           def skip_entity?(entity)
             return super if entity.name != :COM
 
+            @log << 'COM does not operate, Sydney and Addelaide are not connected' unless @game.sydney_adelaide_connected
```
crericha, line 13: Will it operate if the phase passes that requires the connection?

lib/engine/game/g_1848/step/special_track.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require_relative '../../../step/special_track'
+
+module Engine
+  module Game
+    module G1848
+      module Step
+        class SpecialTrack < Engine::Step::SpecialTrack
+          def potential_tiles(entity, hex)
+            return @game.tiles.select { |tile| tile.color == 'blue' } if entity.sym == 'P3'
```
crericha, line 11: The tile is in the ability. I don't think you need this special code.

lib/engine/game/g_1848/game.rb
bentziaxl: that's what happens when you start developing in a language you're not familiar with. 
I had no ides what &. meant.  Now i know Thanks!

lib/engine/game/g_1848/step/dividend.rb
bentziaxl: I rewrote it to this:
 def auto_actions(entity)
            return super unless entity == @game.boe
            [Action::Dividend.new(entity, kind: 'payout')]
          end
          
 and i'm getting an infinte loop. any ideas why? what am i missing?

lib/engine/game/g_1848/step/special_track.rb
bentziaxl: yeah, you're right. It wasn't working because tile_valid_for_phase was returning false.
I removed this and added the method in games.rb

lib/engine/game/g_1848/step/operating.rb
bentziaxl: up until phase 5 (i'll need to add that logic), COM is skipped unless the two cities are connected.
If it happens in the OR prior to COM operating turn , COM will operate.
if it happens after, COM will operate the next OR.

lib/engine/game/g_1848/step/dividend.rb
bentziaxl: found a better way of doing this. I implemented how boe gets tokens in my local repo, and so i can iterate over corporation.tokens,  instead of the map, much more efficient.

lib/engine/game/g_1848/step/dividend.rb
crericha: actions() is likely returning an action for BoE, which will trigger auto_actions to be called.

1) Is actions() gated by entity == current_entity?
2) After the dividend is processed, is the step passing?

lib/engine/game/g_1848/share_pool.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require_relative '../../share_pool'
+
+module Engine
+  module Game
+    module G1848
+      class SharePool < Engine::SharePool
```
tobymao, line 8: let's avoid having a share pool if possible, perhaps by having params on the pool or the corporation

lib/engine/game/g_1848/step/dividend.rb
bentziaxl: done, adding super after the action worked

lib/engine/game/g_1848/share_pool.rb
bentziaxl: by adding params to pool or corporation you mean the base game files?
Or copy corporation into g_1848 and modify it there?

bentziaxl: https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_18_ireland/share_pool.rb, 
does the exact thing i'm after. One corporation (in ireland it's a class of corporation), that has a different market limit.

Is it because it's one company in 1848 you prefer I avoid overwriting share pool?

bentziaxl: Would this work?:
in https://github.com/tobymao/18xx/blob/master/lib/engine/share_pool.rb#L153 
I change @game.class::MARKET_SHARE_LIMIT into @game.market_share_limit(corporation)
in the base game def market_share_limit(corporation) to return MARKET_SHARE_LIMIT, and in 1848 game.rb i overwrite this method to return the company specific share limit, in this case 100 for boe, 50 for rest?

lib/engine/game/g_1848/step/dividend.rb
bentziaxl: hmm, removing both actions and auto_actions, seems to still work. BOE pays out dividends when it has any, otherwise skips.
I have no idea why, thoughts?

crericha: Likely the skip! method. That's how automatic dividends were done before auto_actions existed. See minor_half_pay.rb as an example.

bentziaxl: that's nasty. Yeah the skip! calls process_dividend with subtype withold, which i ignored. So regardless of wether process_dividend was called via skip or not, dividends were being paid.
Why would you name a method skip, if it still calls the procss_dividend action?

bentziaxl: fixed, added auto actions

lib/engine/game/g_1848/share_pool.rb
bentziaxl: done, as discussed in slack.

bentziaxl: rookie mistake, didn't run test prior to checkin in my changes. re-trying locally

lib/engine/game/g_1848/step/dividend.rb
```diff
@@ -0,0 +1,57 @@
...
+  module Game
+    module G1848
+      module Step
+        class Dividend < Engine::Step::Dividend
+          BOE_BASE_PAYOUT = { '2' => 0, '3' => 100, '4' => 100, '5' => 200, '6' => 200, '8' => 300 }.freeze
+
+          def actions(entity)
+            return super unless entity == @game.boe
+
+            ['dividend']
```
crericha, line 15: Make a frozen constant

lib/engine/game/g_1848/step/special_track.rb
crericha: Okay. The diff is still showing the special code. Was it not removed or am I looking at an old copy?

bentziaxl: i removed reference to the file in game.rb, but not the actual file. just submitted it

PR #8079, opened 2022-07-16
Fixes 5639, providing a migration, and moving 1856 to production

### LINKED ISSUE #5639: [1856] Extra Pass required even though no action is possible
During operation, when you take a loan early the game still requires you to pass last, even though you are not allowed to take more.
And that even if you do not have funds to repay any loan.

So the game should skip pass in these situations.

https://18xx.games/game/44377?action=133

image

ventusignis:
Fixes #5639

### DIFF
--- a/lib/engine/game/g_1856/meta.rb
+++ b/lib/engine/game/g_1856/meta.rb
@@ -8,7 +8,7 @@ module Engine
       module Meta
         include Game::Meta
 
-        DEV_STAGE = :beta
+        DEV_STAGE = :production
 
         GAME_DESIGNER = 'Bill Dixon'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/1856'
--- a/lib/engine/game/g_1856/step/loan.rb
+++ b/lib/engine/game/g_1856/step/loan.rb
@@ -11,9 +11,9 @@ module Engine
             return [] if !entity.corporation? || entity != current_entity
 
             actions = []
-            actions << 'payoff_loan' if can_payoff?(entity)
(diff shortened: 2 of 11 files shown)


migrate_game.rb
```diff
@@ -263,6 +274,19 @@ def repair(game, original_actions, actions, broken_action)
   elsif broken_action['type'] == 'dividend' and broken_action['entity_type'] == 'minor'
     actions.delete(broken_action)
     return
+  
+  elsif (broken_action['type'] != 'pass') &&
```
ventusignis, line 278: Is this reasonable? The extra pass gets caught up in a few different ways but I'm wondering if this is too permissive

migrate_game.rb
```diff
@@ -182,6 +182,16 @@ def repair(game, original_actions, actions, broken_action)
         return
       end
     end
+    if game.active_step.is_a?(Engine::Game::G1856::Step::BuySellParShares) ||
```
tobymao, line 185: we're not gonna run this, and will just pin, so let's get rid of it

ventusignis: That sounds good to me

assets/app/view/game_page.rb
```diff
@@ -22,7 +22,7 @@ class GamePage < Snabberb::Component
     APP_PADDING_BOTTOM = '2vmin'
 
     def render_broken_game(e)
-      inner = [h(:div, "We're sorry this game cannot be continued due to #{e}")]
```
ventusignis, line 25: Removing this change

tobymao (approved):

PR #8080, opened 2022-07-16
Rolling Stock - add investment round programmed auto-pass

### LINKED ISSUE #7934: [RSS] Auto pass in investment phase
When a player does not have enough cash to purchase a private and does not want to sell their company shares, it would be nice if they could auto pass.

roseundy:
Fixes #7934 

This required refactoring @programmed_actions code to allow multiple outstanding programmed actions for the same player.
I've run a few 1830 and Harzbahn cases to make sure I didn't break things, however, since this involves UI code changes in server mode, I couldn't check that as completely as I'd like. I did spot check a few games.

Does not require pinning/archiving.

### DIFF
--- a/assets/app/view/game/auto.rb
+++ b/assets/app/view/game/auto.rb
@@ -40,11 +40,10 @@ module View
 
           if !(available = @game.available_programmed_actions).empty?
             available.each do |type|
-              enabled = @game.programmed_actions[sender].find { |a| a.is_a?(type) }
-              if (method = types[type])
-                settings = enabled if enabled
-                children.concat(method.call(settings))
-              end
+              next unless (method = types[type])
+
+              settings = @game.programmed_actions[sender].find { |a| a.is_a?(type) }
+              children.concat(method.call(settings))
             end
           else
             children << h('p.bold', 'No auto actions are presently available for this game.')
--- a/assets/app/view/game/auto.rb
+++ b/assets/app/view/game/auto.rb
@@ -39,10 +39,10 @@ module View
           }.freeze
 
           if !(available = @game.available_programmed_actions).empty?
-            enabled = @game.programmed_actions[sender]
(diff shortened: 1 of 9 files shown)


assets/app/view/game/auto.rb
```diff
@@ -39,10 +39,10 @@ def render_content
           }.freeze
 
           if !(available = @game.available_programmed_actions).empty?
-            enabled = @game.programmed_actions[sender]
             available.each do |type|
+              enabled = @game.programmed_actions[sender].find { |a| a.is_a?(type) }
               if (method = types[type])
```
crericha, line 44: Since you are already refactoring, I think you can do something like "next unless (method = types[type])" at the top of the block.

assets/app/view/game/auto.rb
```diff
@@ -39,10 +39,10 @@ def render_content
...
-                settings = enabled if enabled.is_a?(type)
+                settings = enabled if enabled
```
crericha, line 45: It looks like settings is nil if line 45 doesn't evaluate to true, so I believe the gating if statement is unnecessary.

assets/app/view/game/auto.rb
```diff
@@ -39,10 +39,10 @@ def render_content
...
                 children.concat(method.call(settings))
```
crericha, line 46: I wonder why we call this if settings is nil.

roseundy: To show the "Enable" button (instead of the "Disable" button).

assets/app/view/game/auto.rb
roseundy: Yep. This can be written better.

assets/app/view/game/auto.rb
roseundy: Yep.

PR #8087, opened 2022-07-17
Use TILE_UGPRADES_MUST_USE_MAX_EXITS in 1862

ventusignis:
image

1862 couldn't use a future label for Ipswich because it's already using a LargeIcon for ESR's home, but since the special brown Y and brown N are just a Y and N with one less spoke, the `TILE_UPGRADES_MUST_USE_MAX_EXITS ` setting could be used to get the same effect of having custom labels for the tiles without needing custom logic

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -68,6 +68,7 @@ module Engine
         CAPITALIZATION = :full
 
         MUST_SELL_IN_BLOCKS = true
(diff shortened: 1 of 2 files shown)


roseundy:
The logic looks fine, but I'm not very happy about the tiles looking different from the print version of this game (which uses the 'I' and 'H' labels).

ventusignis:
> The logic looks fine, but I'm not very happy about the tiles looking different from the print version of this game (which uses the 'I' and 'H' labels).

That's a fair point. I'm closing this PR.

PR #8104, opened 2022-07-21
18SJ: Put A label on Stockholm hex/tiles (fixes #8100)

### LINKED ISSUE #8100: Stockholm gray tile can be used by brown non-Stockholm cities
Tile 611 seems to have incorrect tile upgrade to Stockholm gray.

image

perwestling:
Without the label an illegal upgrade on regaular cities might occur.


perwestling:
Might potentially broke any game where tile X5 has been put in the wrong hex, but shouldn't affect games that hasn't reached gray phase.

Did some checks and it seems a lot of recently finished games had been pinned so maybe this wont affect any existing games?

Issue #8103, opened 2022-07-21
18USA Friendly Sale Amount Mismatch


Narvinyar:
MP goes up for sale for $120, but it's share price is only $59 so it should be going for $118 instead. Same goes for UP that is worth $46 but is only put up for $90 as the minimum bid; it should be $92. 

Chapter 14.2.3 states that minimum bid is current share price x company size 

**What's your Game ID?**

https://18xx.games/game/89952

bentziaxl:
see #2165 , bids have to be multiple of 10 above the minimum bid in 1817/18usa.
Therefore if MP is valued at 118, the lowest legal bid 120, so the game handled this correctly.

However, UP opening bid should be 100 (multiple of $10 that is at least the minimum bid of 92)

PR #8115, opened 2022-07-24
fix use of round instead of ceil for 1817 bids

bentziaxl:
fixing #8103 

The rules are that the starting bid is "the bid may be any multiple of $10 that is at least the minimum bid
and above the previous bid." and minimum bid is share value * stock price.

if a company has 2 shares at 46 (2*46=92), round method will round it down to 90. Rules require it to be 100. 
ceil is the correct method, always rounding up.

This will probably break some games (games where someone took the starting bid on companies with stock value * shares in the 0-5 last digit values.


dfannius (approved):

crericha:
Round behaves differently in the client vs server. We'll need to rethink the solution.

PR #8118, opened 2022-07-25
move add_neutral_tokens to init_hexes. correcting starting map

### LINKED ISSUE #3080: 1867 National Railway token should be on starting map
When the "show starting map" button is clicked, the national railway token should remain on the map.

bentziaxl:
fix #3080

### DIFF
--- a/lib/engine/game/g_1861/game.rb
+++ b/lib/engine/game/g_1861/game.rb
@@ -912,7 +912,7 @@ module Engine
           minors + majors + [@national]
         end
 
-        def add_neutral_tokens
+        def add_neutral_tokens(_hexes)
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1861/game.rb
```diff
@@ -912,7 +912,7 @@ def operating_order
           minors + majors + [@national]
         end
 
-        def add_neutral_tokens
+        def add_neutral_tokens(_hexes)
```
bentziaxl, line 915: without this, 1861 breaks, as it calls for init_hexes from 1867, which calls add_neutral_tokens with the extra params.

I could use .airity to figure out if 1867 or 61 is calling this method, but it seems messier then this.

Issue #8020, opened 2022-06-26
Align 18SJ tile setup with AAG standard


perwestling:
Tiles 7/8/9 should be unlimited. (Compare 1817)

The following tiles are 18SJ specific and should be renamed:
440 -> X1
298SJ -> X2
466 -> X3
299SJ ->X4
131 -> X5

scottredracecar:
Also 171 -> X6

PR #8121, opened 2022-07-26
Remove unneeded local method upgrades_to in 18sj 

### LINKED ISSUE #8120: [18SJ] can't place #X5 on Stockholm
Can place it on other locations surprisingly but not on G10 where it should go.

https://18xx.games/game/91159

Kind regards
Michal

bentziaxl:
fix #8120:

#8020 updated the tile names. Stockholm gray was changed from 131 to X5.
The local method of upgrades_to?, returns non existent tile 131 when trying to upgrade to gray. The super method handles the upgrade fine, no need to override it.

### DIFF
--- a/lib/engine/game/g_18_sj/game.rb
+++ b/lib/engine/game/g_18_sj/game.rb
@@ -823,8 +823,8 @@ module Engine
 
         def upgrades_to?(from, to, _special = false, selected_company: nil)
           # Handle upgrade to Stockholm gray tile
-          return to.name == '131' if from.color == :brown && from.hex.name == 'G10'
-          return false if to.name == '131'
+          return to.name == 'X5' if from.color == :brown && from.hex.name == 'G10'
+          return false if to.name == 'X5'
 
           super
         end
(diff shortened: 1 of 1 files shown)


perwestling (approved):
I tried it out and it worked fine. Yes, this was unneeded code from the start as the upgrade of Stockholm hex always was legal.

perwestling:
If this breaks any game, pin them.

bentziaxl:
the tiles had some labels that were changed, added and removed. Maybe some existing label broke the regular upgrade_to? logic and required an override at some point

dolfyner:
Hi!
Just wondering if I will get a notice when the issue is fixed? Seeing as my original ticket was closed/merged into this one?

tobymao:
issue is fixed but probably not deployed yet

dolfyner:
Ok! Will there be a notice of deployment here? Will the deployment break the existing game or can we wait for the deployment and continue it? 
Tx for the help!

tobymao:
no notice, if your game breaks, it can be pinned

tobymao:
it's deployed

dolfyner:
Excellent! Works! Tx again 😁

PR #8129, opened 2022-07-31
[1848] Fix issues

bentziaxl:
fix privates values 
fix boe acting when it shouldn't 
add logic to place tasmania tile when companies close and not used 
add multiple boe tokens in one city 
add boe payout tile on map 
add boe can buy/sell stocks at all time without restrictions 
eloan only when must buy train, 
fix 2e issue when corp has less than cost of train, but can buy it with discount, it should still show up 
privates can be used by both player and corp, when used, need to remove the other ability listed to avoid double dipping


assets/app/view/game/buy_trains.rb
```diff
@@ -302,7 +302,7 @@ def from_depot(depot_trains, corporation)
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             entity = @corporation
 
-            if @selected_company&.owner == @corporation
+            if @selected_company&.owner == @corporation || @selected_company&.owner == @corporation.owner
```
crericha, line 305: Is [@corporation, @corporation.owner].include?(@selected_company&.owner) more DRY? Up to you.

assets/app/view/game/corporation.rb
```diff
@@ -82,6 +82,7 @@ def render
         extras = []
         if @game.corporation_show_loans?(@corporation)
           extras.concat(render_loans) if @game.total_loans&.nonzero?
+          extras.concat(render_interest) if @game.class::SHOW_INTEREST && @game.total_loans&.nonzero?
```
crericha, line 85: Why did you go with a constant rather than a method?

crericha: Also, don't repeat if @game.total_loans&.nonzero?. Put both line 84 and 85 within that if statement.

lib/engine/game/g_1848/game.rb
```diff
@@ -588,7 +592,7 @@ def operating_order
         end
 
         def upgrades_to?(from, to, _special = false, selected_company: nil)
-          return %w[5 6 57].include?(to.name) if (from.hex.tile.label.to_s == 'K') && (from.hex.tile.color == 'white')
+          return %w[5 6 57].include?(to.name) if (from.hex.tile.label.to_s == 'K') && (from.hex.tile.color == :white)
```
crericha, line 595: @ventusignis has a whole scheme for handling label upgrades. Talk to him about how he did it for other games.

lib/engine/game/g_1848/game.rb
```diff
@@ -606,6 +610,10 @@ def sar
           @sar ||= @corporations.find { |corporation| corporation.name == 'SAR' }
         end
 
+        def tasmania
+          @tasmania ||= @companies.find { |c| c.sym == 'P3' }
```
crericha, line 614: There is a company_by_id method

lib/engine/game/g_1848/game.rb
```diff
@@ -665,7 +673,7 @@ def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil
         end
 
         def pres_change_ok?(corporation)
-          return false if corporation == @boe
+          true unless corporation == @boe
```
crericha, line 676: Isn't is just corporation != @boe?

lib/engine/game/g_1848/game.rb
```diff
@@ -716,11 +724,12 @@ def market_share_limit(corporation = nil)
           MARKET_SHARE_LIMIT
         end
 
-        def can_take_loan?(entity)
+        def can_take_loan?(entity, ebuy = false)
```
crericha, line 727: kwarg

lib/engine/game/g_1848/game.rb
```diff
@@ -716,11 +724,12 @@ def market_share_limit(corporation = nil)
...
+          return ebuy if ebuy
+
           entity.corporation? &&
             entity.loans.size < maximum_loans(entity) &&
-            !@loans.empty? &&
-            @take_out_loan_triggered
+            @loans.any? && @take_out_loan_triggered
```
crericha, line 732: Looks like this got reverted from the last PR

assets/app/view/game/corporation.rb
bentziaxl: 1- based on how coupled the previous logic was, i assumed it will be easer to follow if render_loans and render_interest were separated, rather then introduce the logic inside render_loans. LMK if you think i should change it
2- I thought it would be cleaner and more consistent to have a visual toggle as a constant

lib/engine/game/g_1848/step/buy_train.rb
```diff
@@ -33,13 +38,26 @@ def buyable_trains(entity)
...
             trains_to_buy = trains_to_buy.reject { |t| t.name == '2E' } if owns_2e?(entity)
-            trains_to_buy
+            fix_2e_issue(entity, trains_to_buy)
           end
 
           def owns_2e?(entity)
             entity.trains.any? { |t| t.name == '2E' }
           end
 
+          def fix_2e_issue(entity, trains_to_buy)
```
crericha, line 48: Discounts are handled implicitly. Why do you need this?

lib/engine/game/g_1848/step/special_track.rb
```diff
@@ -8,12 +8,13 @@ module G1848
       module Step
         class SpecialTrack < Engine::Step::SpecialTrack
           def process_lay_tile(action)
-            ability = abilities(action.entity)
             super
+            # track lay ability is both on player and corporate, remove all left over abilities
+            company = action.entity
+            company.all_abilities.each { |ab| company.remove_ability(ab) }
             return unless @game.private_closed_triggered
```
crericha, line 15: If you are going to do this pattern multiple times, you can create a module to handle.

lib/engine/game/g_1848/step/tasmania_tile.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
```
crericha, line 1: why do you need a special step for tasmania tile?

lib/engine/game/g_1848/step/track.rb
```diff
@@ -9,6 +9,12 @@ module G1848
       module Step
         class Track < Engine::Step::Track
           include Engine::Game::G1848::Tracker
+
+          def actions(entity)
+            return super unless entity == @game.boe
```
crericha, line 14: You can also create a module for boe to skip all the steps it shouldn't run rather than repeating the code.

lib/engine/step/special_buy_train.rb
```diff
@@ -22,7 +22,7 @@ def blocks?
 
       def process_buy_train(action)
         company = action.entity
-        corporation = company.owner
+        corporation = @game.current_entity
```
crericha, line 25: I prefer to use @round.current_operator because current_entity can also be a player in some circumstances.

lib/engine/step/tracker.rb
```diff
@@ -83,7 +83,7 @@ def abilities(entity, **kwargs, &block)
         @game.abilities(entity, :tile_lay, **kwargs, &block)
       end
 
-      def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
+      def lay_tile(action, extra_cost: 0, entity: nil, spender: nil, ignore_abilities: nil)
```
crericha, line 86: What is the problem you're trying to solve here?

lib/engine/game/g_1848/step/buy_train.rb
bentziaxl: the Ghan costs $200, the privates give $100 discount, If a corp has 100-199 left in the bank the 2e doesn't show up despite having the ability to pay for it

lib/engine/game/g_1848/step/tasmania_tile.rb
bentziaxl: Special case in the rule book.
If the privates close and the Tasmania tile hasn't placed (an ability of one of the privates) the owning player or corp must place it immediately. This step is to handle this scenario.

lib/engine/step/tracker.rb
bentziaxl: The special Tasmania scenario i described above.
At that point either the player or owning corp is placing the tile. I could temporarily add the ability to either with checks, I thought this was easier.

lib/engine/game/g_1848/step/special_track.rb
bentziaxl: do you have a working example on how to do this?

lib/engine/game/g_1848/step/track.rb
bentziaxl: do you know of a working example of this?

assets/app/view/game/corporation.rb
crericha: 1 - Okay
2 - Since corporation_show_loans? is already a method, I think we should make SHOW_INTEREST constant into a method too. I've found the constants often need to get refactored into a method anyhow as there is nuance in many titles.

lib/engine/game/g_1848/step/buy_train.rb
crericha: That capability is built-in. You shouldn't need special code for it. See American Locomotive private in 18USA as an example.

crericha: Example of American Locomotive discounting: https://18xx.games/game/92368?action=322

lib/engine/game/g_1848/step/special_track.rb
crericha: The InterestOnLoans module is a good example of functionality being encapsulated into a module. In your case, you may have a module with a method with the signature ability_used!(company). This could also be a game method instead, if you don't want to go the module way. Half a dozen of one, 6 of another.

Also, all_abilities.each { || company.remove_ability() } will not work the way you want. You can't iterate over a list of abilities that you are removing items from. Instead, use all_abilities.dup.each.

lib/engine/game/g_1848/step/tasmania_tile.rb
```diff
@@ -0,0 +1,71 @@
...
+          def description
+            'Tasmania tile'
+          end
+
+          def blocks?
+            @game.tasmania
+          end
+
+          def active?
+            active_entities.any?
```
crericha, line 30: !active_entities.empty?

Why? active_entities.any? loops through blocks and results in 100's of lines of JS. active_entities.empty? maps to array.length == 0 (1 line of code). Game load performance is significantly better when .empty? is used instead of .any?. The exception is when you need to pass in a block to any? as the block looping is then needed and .empty? won't work.

lib/engine/game/g_1848/step/tasmania_tile.rb
crericha: Okay

lib/engine/game/g_1848/step/tasmania_tile.rb
```diff
@@ -0,0 +1,71 @@
...
+            super
+            @active_entity = nil
+          end
+
+          def available_hex(_entity, hex)
+            @game.tasmania.abilities.first.hexes&.include?(hex.id)
+          end
+
+          def potential_tiles(_entity, _hex)
+            @game.tiles.select { |tile| tile.color == 'blue' }
```
crericha, line 49: The tiles are listed in the ability. It's cleaner if you use that code over special code.

lib/engine/game/g_1848/step/tasmania_tile.rb
```diff
@@ -0,0 +1,71 @@
...
+          end
+
+          def upgradeable_tiles(_entity, ui_hex)
+            super(@game.tasmania, ui_hex)
+          end
+
+          def abilities(_entity, **_kwargs)
+            @game.tasmania.abilities.first
```
crericha, line 57: This isn't likely to break; however, it's more robust if you look through abilities for the tile laying ability.

lib/engine/game/g_1848/step/track.rb
crericha: module skip_boe
  def actions(entity)
     return [] if @game.boe == entity

     super
  end
end

Then in the step:
include skip_boe

lib/engine/game/g_1848/step/buy_train.rb
bentziaxl: I'll try to figure out how to fix this bug in the base code

lib/engine/step/tracker.rb
crericha: Workarounds/hacks like this are bad in the common code. For example, if every game added their workarounds here, the code would be unreadable. It's more acceptable if the workaround is in game specific code.

I'm trying to think of an implemented game that allows the player to act on behalf of the company for a tile lay, but I came up empty. The closest I came up with was the Steamboat private in 46 that can be assigned by the owner at the beginning of the OR.

If the entity is the tasmania company, does it not work?

bentziaxl: found a better way of doing this, in tasmania I add a track laying ability to the owning player, and use the regular flow.

assets/app/view/game/buy_trains.rb
```diff
@@ -302,7 +302,7 @@ def from_depot(depot_trains, corporation)
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             entity = @corporation
 
-            if @selected_company&.owner == @corporation
+            if [@corporation, @corporation.owner].include?(@selected_company&.owner)
```
crericha, line 305: Will this break other games where the corp owner can't activate the power? I wonder if you need to check who is allowed to play the ability?

lib/engine/game/g_1848/game.rb
```diff
@@ -716,28 +721,30 @@ def market_share_limit(corporation = nil)
           MARKET_SHARE_LIMIT
         end
 
-        def can_take_loan?(entity)
+        def can_take_loan?(entity, ebuy: nil)
+          return ebuy if ebuy
```
crericha, line 725: return true if ebuy?

lib/engine/game/g_1848/step/buy_sell_par_shares.rb
```diff
@@ -15,6 +15,18 @@ def process_buy_shares(action)
...
           end
+
+          def can_sell?(entity, bundle)
+            return true if bundle.corporation == @game.boe
+
+            super
+          end
+
+          def can_buy?(entity, bundle)
+            return true if bundle.corporation == @game.boe
```
crericha, line 26: Does this handle not going over the cert limit?

bentziaxl: can_buy? doesn't handle cert limit even in the parent method, it's handled in must_sell?

assets/app/view/game/buy_trains.rb
bentziaxl: if the ability is passive, that it shouldn't have an impact.
The game will need to enable the ability to show up when owned by player, and only then will this code path execute.
Checking for active abilities

bentziaxl: doesn't break 18usa. Tries two other passive games.

lib/engine/game/g_1848/step/buy_sell_par_shares.rb
crericha: can_buy? calls can_gain?, which checks the cert limit. You'll need to check cert limit here.

bentziaxl: i removed the overridden method.
It's not clear from the rules wether you can or can't buy a share if you've already sold in the same round.
I will keep the behaviour the same as other 18xx games.

dfannius: > **12.3.2** If a player has sold shares in a company in a SR, he may not buy shares in the same company during this same SR.

So it's the same as the usual.

If you're talking about the Bank of England, then at least in the original rules, it's specified that it counts as a company except for explicit exceptions.

bentziaxl: ok, so my understanding is correct. I was working off a checklist someone else made, where it said no restrictions on boe shares at all. Current code is accurate with the rules

lib/engine/game/g_1848/game.rb
```diff
@@ -773,6 +806,12 @@ def check_distance(route, visits, _train = nil)
           raise GameError, 'Route must include Alice Springs'
         end
 
+        def route_distance_str(route)
```
ventusignis, line 809: There's a common module used in several games you can use instead of this, see 18FL as an example

require_relative '../cities_plus_towns_route_distance_str'

lib/engine/game/g_1848/game.rb
```diff
@@ -794,7 +833,8 @@ def edge_is_a_border(edge)
         end
 
         def revenue_for(route, stops)
-          k_sum = stops.count { |rl| rl.hex.tile.label.to_s == 'K' }
+          k_sum = stops.count { |rl| rl.hex&.tile&.label&.to_s == 'K' || rl.hex&.tile&.future_label&.label.to_s == 'K' }
```
ventusignis, line 836: Thought: I wonder if future labels would be better if they were a form of label, that just applied later on with separate rendering rules. I'm not sure how often logic like this will be needed to distinguish between future labels and labels

lib/engine/game/g_1848/game.rb
```diff
@@ -879,6 +919,25 @@ def ghan_visited?(visited_node)
...
         end
+
+        def entity_can_use_company?(entity, company)
+          # company abilities only work once they can be bought
+          return false unless can_use_compnay_ability?
+
+          super
+        end
+        
+        def can_use_compnay_ability?
```
ventusignis, line 930: Nitpick: typo in method name, should be company not compnay

lib/engine/game/g_1848/step/tasmania_tile.rb
```diff
@@ -0,0 +1,93 @@
...
+  module Game
+    module G1848
+      module Step
+        class TasmaniaTile < G1848::Step::SpecialTrack
+          def actions(_entity)
+            return [] if !@game.private_closed_triggered || @game.tasmania.closed?
+
+            if @active_entity.nil?
+              @active_entity = @game.tasmania
+              @game.log << "#{@game.tasmania.name} must lay tasmania tile"
```
ventusignis, line 15: Nitpick: would Tasmania be better capitalized? I'm not sure how it's referred to in the rest of the rules and descriptions in the game. I'm ok either way

lib/engine/game/g_1848/game.rb
```diff
@@ -270,8 +292,8 @@ class Game < Game::Base
           {
             sym: 'P1',
             name: "Melbourne & Hobson's Bay Railway Company",
-            value: 40,
-            discount: 10,
+            value: 30,
+            discount: 0,
```
ventusignis, line 296: Nitpick: If the discount is just 0 I believe you can simply omit the discount entirely

lib/engine/game/g_1848/game.rb
bentziaxl: It is a bit of a future label situation, the K is on the map, but not on the yellow tile, but is on the green tile.
Normal label won't work (or will require extra logic, which i removed so i can use future label instead).
Visually, it's clear that those are k tiles, i'm not sure if this needs rework

lib/engine/game/g_1848/step/tasmania_tile.rb
bentziaxl: it's a variable referring to to the Tasmania private company. in the log it will show up properly

lib/engine/game/g_1848/game.rb
bentziaxl: agreed, but i don't want to touch the future label code at this point

lib/engine/game/g_1848/game.rb
bentziaxl: cool!

lib/engine/game/g_1848/game.rb
bentziaxl: yeah, i don't know of a game that would use future labels, but not treat them as regular label, but maybe there are some out there.

assets/app/view/game/corporation.rb
```diff
@@ -81,7 +81,10 @@ def render
 
         extras = []
         if @game.corporation_show_loans?(@corporation)
-          extras.concat(render_loans) if @game.total_loans&.nonzero?
+          if @game.total_loans&.nonzero?
```
crericha, line 84: In general positive? is better to use because non-zero may be -1. Obviously in this case, -1 would be  bug in itself.

lib/engine/game/g_1848/step/buy_sell_par_shares.rb
```diff
@@ -15,6 +15,12 @@ def process_buy_shares(action)
                        allow_president_change: @game.pres_change_ok?(action.bundle.corporation))
             track_action(action, action.bundle.corporation)
           end
+
+          def can_sell?(entity, bundle)
+            return true if bundle.corporation == @game.boe
```
crericha, line 20: Do BoE shares go back to BoE instead of the market?

lib/engine/game/g_1848/step/buy_train.rb
```diff
@@ -32,14 +37,36 @@ def buyable_trains(entity)
...
+              trains_to_buy << ghan_train
+            end
+            trains_to_buy.uniq
           end
 
           def owns_2e?(entity)
             entity.trains.any? { |t| t.name == '2E' }
           end
 
+          def can_buy_2e?(entity)
```
crericha, line 52: If this method is going to only apply to the discount, maybe change the name to can_buy_2e_with_discount?(entity)?

lib/engine/game/g_1848/step/tracker.rb
```diff
@@ -9,10 +9,7 @@ module Tracker
         include Engine::Step::Tracker
 
         def lay_tile_action(action)
-          # Yellow cities on K tiles need the K label added
-          add_k = action.hex.tile.label.to_s == 'K'
           super
-          action.hex.tile.label = 'K' if add_k
 
           was_connected = @game.sydney_adelaide_connected
           now_connected = @game.check_sydney_adelaide_connected
```
crericha, line 15: You do not want to check graph on every tile lay. This is a HUGE performance hit. Take a look at the CheckNYCFormation  step for an example of how to use auto actions.

crericha (approved):

lib/engine/game/g_1848/step/buy_train.rb
bentziaxl: it doesn't, i'll need to use it for a bug i found. When a company is at train limit, it can buy a 2e currently. room?(entity) needs to check when at limit if the corp can buy the 2e, and will rely on this method

lib/engine/game/g_1848/step/buy_sell_par_shares.rb
bentziaxl: there's no boe market/ipo distinction. In the phyiscal game the shares go into a pile on top of the BOE placemat.
In this implementation i just move all the shares into the market at setup of game.

lib/engine/game/g_1848/step/tracker.rb
crericha: To be fixed in next PR

PR #8133, opened 2022-08-04
Implement minimum player counts

### LINKED ISSUE #6538: Set minimum players when creating game

crericha:
Fixes #6538

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -10,7 +10,8 @@ module View
     include Lib::WhatsThis::AutoRoute
 
     needs :mode, default: :multi, store: true
-    needs :num_players, default: 3, store: true
+    needs :min_players, default: nil, store: true
+    needs :max_players, default: nil, store: true
     needs :flash_opts, default: {}, store: true
     needs :user, default: nil, store: true
     needs :visible_optional_rules, default: nil, store: true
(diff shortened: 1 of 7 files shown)


crericha:
@roseundy, please verify that using max player count for 1825 unit selection and incorporating min player count into filtered rules meets your intentions.

assets/app/view/create_game.rb
```diff
@@ -342,7 +358,7 @@ def mode_selector
 
     def mode_input(mode, text)
       click_handler = lambda do
-        store(:mode, mode, skip: true)
+        store(:mode, mode) # , skip: true)
```
tobymao, line 361: why was this changed? should you remove the comment

tobymao (approved):

assets/app/view/create_game.rb
crericha: To hide the minimum players input when hotseat is selected. I will verify it is still needed and if so remove the commented out portion.

crericha: Not needed. Reverted.

PR #8134, opened 2022-08-04
[1848] Fix overuse of graph

bentziaxl:
fix other issues:
- privates counting for cert limit
- boe shares sellable after buy
- no loans after dividends in later game stages
- bow token takeover broken


lib/engine/game/g_1848/game.rb
```diff
@@ -596,6 +604,7 @@ def operating_round(round_num)
             G1848::Step::SpecialBuyTrain,
             G1848::Step::BuyTrain,
             [G1848::Step::BuyCompany, { blocks: true }],
+            [G1848::Step::Loan, { blocks: true }],
```
crericha, line 607: Do you want to make the person pass twice in a row? There are better ways to only block once at the end of the turn. Also, will the loan step block if the company has already taken its loan?

lib/engine/game/g_1848/game.rb
```diff
@@ -642,17 +651,20 @@ def adelaide
           @adelaide ||= hex_by_id('G6')
         end
 
-        def check_sydney_adelaide_connected
-          return @sydney_adelaide_connected if @sydney_adelaide_connected
+        def sydney_adelaide_connected?
+          return true if @sydney_adelaide_connected
 
           graph = Graph.new(self, home_as_token: true, no_blocking: true)
           graph.compute(sar)
```
crericha, line 658: Can the graph be removed from here and just return @sydney_adelaide_connected?

crericha: Basically, make a separate method called, check_for_sydney_adelaide_connection.

lib/engine/game/g_1848/game.rb
bentziaxl: one pass is better than two :-). How do you only block once?

lib/engine/game/g_1848/game.rb
bentziaxl: it's only used once
image

bentziaxl: I added the first line as a defense in case anyone accidentally calls this method, when they just want to check sydney_adelaide_connected value
 i can rename to check_... if that's the convention

lib/engine/game/g_1848/game.rb
bentziaxl: you're right, it doesn't. It allows a company to take two loans in one OR, which isn't not allowed. Oops. Good catch

lib/engine/game/g_1848/step/buy_train.rb
```diff
@@ -68,6 +68,8 @@ def ghan_private_owned?(entity)
           end
 
           def room?(entity)
+            return true if entity.trains.count { |t| t.name != '2E' } == @game.train_limit(entity) && can_buy_2e?(entity)
```
crericha, line 71: Couldn't this whole method be collapsed down to:

entity.trains.count { |t| t.name != '2E' } <= @game.train_limit(entity) && can_buy_2e?(entity)

if can_buy_2e? works with and without the ghan private

cost = ghan.train.price
cost -= 100 if ghan_private_owned?(entity) && @game.phase.available?(ghan_train.available_on)
cost <= entity.cash

lib/engine/game/g_1848/step/check_com_formation.rb
```diff
@@ -0,0 +1,47 @@
...
+            @connected
+          end
+
+          def auto_actions(entity)
+            return [] unless @connected
+
+            [Engine::Action::DestinationConnection.new(entity)]
+          end
+
+          def process_destination_connection(_action)\
```
crericha, line 39: Is the \ a typo?

bentziaxl: yes

lib/engine/game/g_1848/step/buy_train.rb
bentziaxl: entity.trains.count { |t| t.name != '2E' } <= @game.train_limit(entity) && can_buy_2e?(entity):

Wouldn't it return false if a corp bought a 2e, but is not at train limit yet?

bentziaxl: as for the second part, reading the value of the discount from the ability leads to nicer code.
100 needs explanation, and assumes knowledge of how much the discount is.

ghan_train.price - ghan_private_ability.discount is self documenating

bentziaxl: yes, they don't have to apply the discount.

lib/engine/game/g_1848/game.rb
crericha: Common practice is to tie both capabilities into one step (See payoff loan in 1817). Maybe EndOfTurnActions?

lib/engine/game/g_1848/game.rb
crericha: Got it. Maybe rename so that its clear that the method is for making the check, not querying state.

lib/engine/game/g_1848/step/buy_train.rb
bentziaxl: I realized what you're saying. can_buy_2e forces player to have the ghan private, but that's not correct.
You can buy it without the discount while being full

crericha: My bad, I was thinking of only the 2E case. It should be || not &&.

bentziaxl: i made some more changes, review and let me know if you have any other comments

lib/engine/game/g_1848/step/buy_train.rb
```diff
@@ -50,25 +50,31 @@ def owns_2e?(entity)
...
 
+          def at_train_limit?(entity)
+            entity.trains.count { |t| t.name != '2E' } == @game.train_limit(entity)
+          end
+
           def room?(entity)
-            return true if entity.trains.count { |t| t.name != '2E' } == @game.train_limit(entity) && can_buy_2e?(entity)
+            return true if at_train_limit?(entity) && can_buy_2e?(entity)
 
             entity.trains.count { |t| t.name != '2E' } < @game.train_limit(entity)
```
crericha, line 79: New methods added simplify this

def room?(entity) 
  !at_train_limit?(entity) || can_buy_2e?(entity)
end

lib/engine/game/g_1848/step/take_loan_buy_company.rb
```diff
@@ -7,42 +7,37 @@ module Engine
...
           def process_take_loan(action)
-            @loan_taken = true
             entity = action.entity
             @game.take_loan(entity, action.loan)
           end
 
           def setup
             # you can only take one loan per OR turn
-            @loan_taken = false
+            @game.loan_taken = false
```
crericha, line 40: Make this a round variable. See buy_sell_par_shares for examples. (def round_state and def setup)

crericha (approved):

PR #8159, opened 2022-08-09
[1848] Fix issues

bentziaxl:
- discount on variant D not showing
- kwargs failure when ebuying
- 2E failing when running across borders
- When ebuying showing discard trains when it shouldn't
- lower amount of "unlimited" trains. 999 trains causes major slowdown (duh!)
- don't show shares to sell when ebuying and can_ebuy_sell_shares is false
- can buy more expensive trains in 1848

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1841,6 +1841,7 @@ module Engine
 
       def discountable_trains_for(corporation)
         discountable_trains = @depot.depot_trains.select { |train| train.discount || train.variants.any? { |_, v| v[:discount] } }
+
         corporation.trains.flat_map do |train|
           discountable_trains.flat_map do |discount_train|
             discount_info = []
@@ -1860,6 +1861,7 @@ module Engine
                 discount_info << [train, discount_train, v[:name], price]
               end
             end
+
             discount_info
           end.compact
         end
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1840,24 +1840,26 @@ module Engine
       end
 
       def discountable_trains_for(corporation)
-        discountable_trains = @depot.depot_trains.select(&:discount)
-
+        discountable_trains = @depot.depot_trains.select { |train| train.discount || train.variants.any? { |_, v| v[:discount] } }
         corporation.trains.flat_map do |train|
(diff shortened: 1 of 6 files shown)


lib/engine/game/base.rb
```diff
@@ -1840,22 +1840,26 @@ def buy_train(operator, train, price = nil)
...
 
         corporation.trains.flat_map do |train|
           discountable_trains.flat_map do |discount_train|
+            discount_info = []
             discounted_price = discount_train.price(train)
-            next if discount_train.price == discounted_price
-
-            name = discount_train.name
-            discount_info = [[train, discount_train, name, discounted_price]]
+            if discount_train.price > discounted_price
```
crericha, line 1849: >=

lib/engine/game/base.rb
```diff
@@ -1840,22 +1840,26 @@ def buy_train(operator, train, price = nil)
...
+              name = discount_train.name
+              discount_info = [[train, discount_train, name, discounted_price]]
```
crericha, line 1851: remove name = discount_train.name and use discount_train.name directly since its only used once

lib/engine/game/g_1848/step/buy_train.rb
```diff
@@ -37,6 +37,7 @@ def buyable_trains(entity)
             # Cannot buy 2E if one is already owned, can't by non 2e if at limit
             trains_to_buy = at_train_limit?(entity) ? [] : super
             trains_to_buy = trains_to_buy.select(&:from_depot?) unless @game.can_buy_trains
+            trains_to_buy = reject_unaffordable_depot_trains(trains_to_buy, entity) if must_buy_train?(entity)
```
crericha, line 40: Why do you need to remove these? Won't the buy action fail if they try to buy it? In practice its good to show the train you could buy after enough loans.

bentziaxl: bad reading of the rules on my part. Discarded trains are removed from the game.

crericha (approved):

lib/engine/game/base.rb
bentziaxl: it's used bellow, L1856 as well

crericha (approved):

PR #8161, opened 2022-08-10
[1848] Fix issues

bentziaxl:
fix minor issues:
- use a dummy corp instead of SAR to check if sydney-adelaide are connected
- do not show log that com can operate via connectivity if it already can via phase change
- if corp goes into receivership, and boe gets a token on a white tile. OR 0 in that case when calculating revenue


lib/engine/game/g_1848/game.rb
```diff
@@ -636,9 +636,9 @@ def tile_valid_for_phase?(tile, hex: nil, phase_color_cache: nil)
           super
         end
 
-        def sar
```
bentziaxl, line 639: I checked the code, @game.sar isn't used anywhere except here

PR #8167, opened 2022-08-11
[1822PNW] Merge round and forests

outsidePasser:


lib/engine/game/g_1822_pnw/game.rb
```diff
@@ -391,7 +442,7 @@ def setup_companies
 
           minors = @companies.select { |c| c.id[0] == self.class::COMPANY_MINOR_PREFIX }
           minor_6, minors = minors.partition { |c| c.id == 'M6' }
-          minors_assoc, minors = minors.partition { |c| MINOR_ASSOCIATIONS.key?(c.id) }
+          minors_assoc, minors = minors.partition { |c| MINOR_ASSOCIATIONS.key?(c.id[1..-1]) }
```
crericha, line 445: What's this for?

lib/engine/game/g_1822_pnw/game.rb
```diff
@@ -604,6 +656,25 @@ def regional_railway?(entity)
...
+          @corporations.select { |c| c.floated? && c.type == :minor && !MINOR_ASSOCIATIONS.include?(c.id) }
+        end
+
+        def associated_major(minor)
+          corporation_by_id(MINOR_ASSOCIATIONS[minor.id])
+        end
+
+        def forest?(tile)
+          tile.parts.each do |part|
+            return true if part.is_a?(Engine::Part::Upgrade) && (part.terrains[0] == :forest)
```
crericha, line 674: What do you think about associating a forest icon instead of using terrain? It might be more straight-forward.

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -0,0 +1,214 @@
...
+              'Choose par value for new company'
+            when :selecting_shares
+              'Choose number of shares'
+            when :selecting_token
+              'What to do with the token'
+            end
+          end
+
+          def min_exchange_shares(par, minors_value, minors_cash)
+            ((minors_value - minors_cash - 1) / par).floor + 1
```
crericha, line 54: I believe If you don't have a float, integer math will be used.

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -0,0 +1,214 @@
...
+
+          def max_exchange_shares(par, minors_value, player_cash)
+            ((minors_value + [player_cash, (par - 1)].min) / par).floor
+          end
+
+          def possible_exchanged_shares(par, minors_cash, minors_value, player_cash)
+            (2..10).to_a.filter do |n|
+              n >= min_exchange_shares(par, minors_value,
+                                       minors_cash) && n <= max_exchange_shares(par, minors_value,
+                                                                                player_cash) && (6 * par) >= minors_value
```
crericha, line 65: min and max don't change. Calculate before the block?

crericha: Also, if (6 * par) >= minors_value if false, will the block always return false?

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -0,0 +1,214 @@
...
+          end
+
+          def can_par_at?(par, minors_cash, minors_value, player_cash)
+            return true unless possible_exchanged_shares(par, minors_cash, minors_value, player_cash).empty?
+          end
+
+          def valid_par_prices(minor_one, minor_two)
+            minors_value = (minor_one.share_price.price + minor_two.share_price.price) * 2
+            minors_cash = minor_one.cash + minor_two.cash + 100
+            @game.stock_market.par_prices.map(&:price).sort.filter do |par|
```
crericha, line 76: Use select instead of filter. Apply elsewhere filter is used.

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -0,0 +1,214 @@
...
+              choices['exchange'] = 'Move an exchange token to available token area on the corporation chart'
+            end
+            choices
+          end
+
+          def process_merge(action)
+            @associated_minor = action.entity
+            @unassociated_minor = action.corporation
+
+            if !@unassociated_minor || !mergeable(@associated_minor).include?(@unassociated_minor)
```
crericha, line 115: Put an @loading check here since you are using graph in merageable

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -0,0 +1,214 @@
...
+            transfer_posessions(@associated_minor, @new_corporation)
+            transfer_posessions(@unassociated_minor, @new_corporation)
+
+            @merge_state = :selecting_par
+          end
+
+          def process_choose(action)
+            case @merge_state
+
+            when :selecting_par
```
crericha, line 133: methods for each case may make the code more readable.

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -0,0 +1,214 @@
...
+          def transfer_posessions(minor, corporation)
+            receiving = []
+
+            if minor.cash.positive?
+              receiving << @game.format_currency(minor.cash)
+              minor.spend(minor.cash, corporation)
+            end
+
+            companies = @game.transfer(:companies, minor, corporation).map(&:name)
+            receiving << "companies (#{companies.join(', ')})" if companies.any?
```
crericha, line 195: !companies.empty?

lib/engine/game/g_1822_pnw/step/special_track.rb
```diff
@@ -20,8 +20,11 @@ def legal_tile_rotation?(_entity, _hex, tile)
             return true if tile.id == 'BC-0'
           end
 
-          #          def available_hex(entity, hex)
-          #          end
+          def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
+            raise GameError, 'Cannot upgrade forests' if action.hex.assigned?('forest')
```
crericha, line 24: The forest is assigned here, but terrain in the game class?

lib/engine/game/g_1822_pnw/step/track.rb
```diff
@@ -33,10 +33,18 @@ def process_lay_tile(action)
               @round.num_laid_track += 1
               @round.laid_hexes << action.hex
             else
+              forest = @game.forest?(action.hex.tile)
```
crericha, line 36: If you are going to reuse, consider a module

lib/engine/game/g_1822_pnw/game.rb
outsidePasser: [1..-1] takes the first character off the of the string of the id (which is always 'M') so that it can match up against the IDs in MINOR_ASSOCIATIONS.  Most of the places this is checked it's being done against the corporation ID (which does not contain the "M") instead of against the company ID (which does).

lib/engine/game/g_1822_pnw/game.rb
outsidePasser: I went with terrain because those hexes have a terrain cost of $10 when you put a tile on them - so it seemed more appropriate.

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: It will in ruby, yes - but since it is just taking the floor at the end, it doesn't actually matter.

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: Yes, they can be calculated before the block.  I'll change it.

It is the intent to always return false in that case - it would be better if that was a check up front.  I'll change it.

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: Ok, changed

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: Done

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: Changed

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: Done

lib/engine/game/g_1822_pnw/step/special_track.rb
outsidePasser: On the base map, hexes can contain the terrain "forest" - this is similar to hills, mountains, etc., and is put in within the map.rb file (e.g. 'upgrade=cost:10,terrain:forest', ).  Once a tile gets placed on the hex, the forest gets put on that tile, similar to how coal works in 1817 - so I patterned it after that.

lib/engine/game/g_1822_pnw/step/track.rb
outsidePasser: There's a total of three lines that are re-used - is that worth making a module for?

lib/engine/game/g_1822_pnw/game.rb
crericha: Okay. A well-named helper method will make the code more readable for others.

lib/engine/game/g_1822_pnw/game.rb
crericha: Okay, if I understand. The forest terrain informs which tiles the lumber ability can be placed on. Once placed, a forest icon is assigned to the tile. Is that correct?

Tile has a method called terrain. Why doesn't that method work for you?

lib/engine/game/g_1822_pnw/step/track.rb
crericha: To me? Yes. Up to you for you.

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -112,7 +113,7 @@ def process_merge(action)
             @associated_minor = action.entity
             @unassociated_minor = action.corporation
 
-            if !@unassociated_minor || !mergeable(@associated_minor).include?(@unassociated_minor)
+            if !@game.loading || !@unassociated_minor || !mergeable(@associated_minor).include?(@unassociated_minor)
```
crericha, line 116: !@game.loading && (other conditions)

lib/engine/game/g_1822_pnw/game.rb
outsidePasser: Ok.  Made helper method.

lib/engine/game/g_1822_pnw/game.rb
outsidePasser: Because I didn't see it. :)

Changed to:
          tile.terrain.include?(:forest)

lib/engine/game/g_1822_pnw/step/merge.rb
outsidePasser: Fixed

lib/engine/game/g_1822_pnw/step/track.rb
outsidePasser: I don't think I will for that.

PR #8171, opened 2022-08-11
Convert game result from user name to id, remove game messages for users not in the game

crericha:
- Game result hash keys converted from user name to id
- Added finished_at field to game, so when game json is updated due to development changes, it doesn't look like the game just ended
- Remove game messages for users not in the game

I have a script I'll need to run to do the game result conversion once this is deployed. Game cards for finished games will not render properly during that time.


assets/app/view/game/actionable.rb
```diff
@@ -95,6 +95,7 @@ def process_action(action)
 
         if game.finished
           @game_data[:result] = game.result
+          @game_data[:finished_at] = Time.now.to_i
```
tobymao, line 98: what is this? is this a one time set?

lib/engine/game/base.rb
```diff
@@ -1179,16 +1183,21 @@ def discard_discount(train, price)
...
         @finished = true
+        @ended_by_player = player_initiated
         store_player_info
         @round_counter += 1
-        scores = result.map { |name, value| "#{name} (#{format_currency(value)})" }
+        scores = result.map { |id, value| "#{@players.find { |p| p.id == id.to_i }&.name} (#{format_currency(value)})" }
         @log << "-- Game over: #{scores.join(', ')} --"
       end
 
+      def manually_ended?
```
tobymao, line 1197: probably not necessary and you can just use a attr_reader

migrate/0007_add_elo_ratings_to_games.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    add_column :games, :elo, :jsonb, null: false, default: '{}'
```
tobymao, line 5: i don't know that we want to store elo in games, let's chat through that?

assets/app/view/game/actionable.rb
crericha: Yes, a one-time set. This is to maintain the the game finish date, while allowing dev changes that cause the json to change. Otherwise, every game will appear like it finished when the json change is applied.

crericha: This particular location is because of hot seat games.

lib/engine/game/base.rb
crericha: So just make an attr_reader than a method?

tobymao: yea

PR #8196, opened 2022-08-14
[1848] minor improvements + perth bug fix

### LINKED ISSUE #8186: 1848 Perth not correct
In the 18xxgames version, Perth appears as connecting to the three surrounding hexes, and at least in our game, a train can run through it.  On the hardcopy map, and in the rules, Perth only has 3 arrowing pointing into it, so can only be start (or end) of a route.

### LINKED ISSUE #8197: [1848] 2E is counting dits/towns when it should not
Rules state in 8.2.5 that "The Ghan” train only counts Alice Springs and the city where it starts." It is also counting dits/towns which it should not do.

https://18xx.games/game/93112?action=597

bentziaxl:
add blocking loan step for cases where a company has 0 trains, one step away from receivership. Should able to put down token and get a loan
add k bonus string revenue
fix perth, fix #8186 
check for player cert limit not game cert limit in must_sell?
fixed #8197 
hide show interest in spreadsheet

### DIFF
--- a/lib/engine/game/g_1848/step/blocking_loan.rb
+++ b/lib/engine/game/g_1848/step/blocking_loan.rb
@@ -16,7 +16,7 @@ module Engine
             return [] if entity == @game.boe
 
             actions = []
-            actions += %w[take_loan pass] if check_blocking_loan(entity)
+            actions << %w[take_loan pass] if check_blocking_loan(entity)
             actions
           end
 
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_1848/step/blocking_loan.rb
```diff
@@ -0,0 +1,56 @@
...
+        class BlockingLoan < Engine::Step::Base
+          include SkipBoe
+          # Used when a company is one step away from receivership & places a token.
+          # Should be able to take a loan before receivership
+          def actions(entity)
+            return [] if !entity.corporation? || entity != current_entity
+            return [] if entity == @game.boe
+
+            actions = []
+            actions += %w[take_loan pass] if check_blocking_loan(entity)
```
tobymao, line 19: <<

PR #8198, opened 2022-08-15
Daily calculation of user stats

crericha:


lib/bus.rb
```diff
@@ -25,11 +34,15 @@ def self.[](key)
...
+    redis.pipelined do |pipeline|
+      pipeline.set(key, value)
+      pipeline.expire(key, ttl_for(key))
     end
 
     value
   end
+
+  def self.ttl_for(key)
+    TTL.find { |k, _v| key.include?(k) }&.last || 0
```
tobymao, line 46: why not just TTL[key] is key some kind of dict?

lib/user_stats.rb
```diff
@@ -0,0 +1,80 @@
...
+  K_NEW = 60.0
+  YEARS_INCLUDED = 4
+  BATCH_SIZE = 1000
+
+  def self.calculate_stats
+    user_stats = Hash.new { |hash, key| hash[key] = Hash.new([1200, 0]) }
+    on_or_after = Date.today - (365 * YEARS_INCLUDED)
+    offset = 0
+
+    loop do
```
tobymao, line 16: i think there's a sequel method that let's you iterate in batches. checkout paged_each(rows_per_fetch: 1000)

lib/user_stats.rb
```diff
@@ -0,0 +1,80 @@
...
+  def self.k_for(num_players, num_plays)
+    (num_plays <= 30 ? K_NEW : K) / (num_players - 1)
+  end
+
+  def self.update_stats(player, elo_change, category, stats)
+    old_elo, plays = stats[player][category]
+    stats[player][category] = [old_elo + elo_change[category], plays + 1]
+  end
+
+  def self.store_stats(user_id, stats)
```
tobymao, line 77: this stores one at a time maybe add a pipeline on top so it's only one connection?

lib/bus.rb
crericha: Key has a unique id in it. Example, if the user's id is 1234, the key would be "USER_STATS:1234"

lib/user_stats.rb
crericha: Awesome. Fixed.

lib/user_stats.rb
crericha: Fixed.

lib/bus.rb
tobymao: this should be explicit, get the key by splitting the string and then doing normal hash lookip

crericha: Fixed.

lib/user_stats.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    Game.eager(:players).where(status: %w[finished archived], manually_ended: false)
+                        .where { finished_at >= on_or_after }
+                        .order(:finished_at)
+                        .paged_each(rows_per_fetch: BATCH_SIZE) do |game|
+      calculate_game_stats(game, user_stats)
+    end
+
+    store_stats(user_stats)
+    true
```
tobymao, line 23: why return true?

lib/bus.rb
```diff
@@ -24,12 +33,28 @@ def self.[](key)
...
 
+  def self.store_keys(keys)
+    redis.pipelined do |pipeline|
+      keys.each do |key, value|
+        pipeline.set(key, value)
+        pipeline.expire(key, ttl_for(key))
+      end
+    end
+
+    true
```
tobymao, line 44: why return true?

lib/bus.rb
```diff
@@ -7,7 +7,16 @@ module Bus
   TTL = 60 * 5
   URL = 'redis://redis:6379'
 
-  USER_TS = 'user_ts:%d'
+  USER_TS_PREFIX = 'user_ts'
```
tobymao, line 10: freeze these as well

tobymao (approved):

lib/bus.rb
```diff
@@ -24,12 +33,28 @@ def self.[](key)
...
+    redis.pipelined do |pipeline|
+      pipeline.set(key, value)
+      pipeline.expire(key, ttl_for(key))
     end
 
     value
   end
+
+  def self.ttl_for(key)
+    prefix = key.split(':').first
```
tobymao, line 57: split(':')[0]

lib/user_stats.rb
crericha: Before it was returning the user_stats hash, which is long. What would you like it to return?

lib/bus.rb
crericha: It succeeded? Would you prefer nothing?

lib/user_stats.rb
tobymao: well if you don't need it to return anything don't return anything

lib/bus.rb
tobymao: do these methods except a return?

lib/bus.rb
crericha: Is [0] more efficient?

lib/user_stats.rb
crericha: So put a nil at the bottom?

lib/bus.rb
tobymao: no, but i prefer it because it's more concise

lib/user_stats.rb
tobymao: no, just don't do anything

crericha: Ruby will return the output of the last line of the method. What's your opinion on what to do if that last line is a really big hash, string, etc.?

tobymao: it's fine, just don't use it, you should only return stuff purposefully if you're going to use the value

PR #8200, opened 2022-08-15
[1822PNW] Private 14

outsidePasser:


lib/engine/ability/assign_hexes.rb
```diff
@@ -5,10 +5,11 @@
 module Engine
   module Ability
     class AssignHexes < Base
-      attr_reader :hexes
+      attr_reader :hexes, :closed_when_used_up
 
-      def setup(hexes:)
+      def setup(hexes:, closed_when_used_up: nil)
         @hexes = hexes
+        @closed_when_used_up = closed_when_used_up || false
```
tobymao, line 12: || false isn't necessary

outsidePasser: Removed

PR #8205, opened 2022-08-16
[1848] Fix issues 8188 & 8203

### LINKED ISSUE #8188: 1848 value of private companies
You have the value of private companies as the starting price (top value) of the auction chart.  This is not exactly correct - if the game were to end before private companies close, private companies have no value.  The rules don't say the private companies may be bought from 1 to 2x face value, and since the amount private companies may be bought are specified for each company, and that the apparent value/price of the companies looks based off of 1830 - I would say the values of the privates would be 20, 40, 70, 110, 165 and 220.

### LINKED ISSUE #8203: [1848] train issues
Trade in should be 800, not 500.
Don't think you can trade up Ghan to D.
Cannot cross buy Ghans.
EMR with multiple loans is not intuitive (should not have to click "buy" on unaffordable train.).

bentziaxl:
- reset company values after end of auction (value only used to keep track of starting price before reductions)
- variant discount logic was broken, it used price of non variant train
- 2e can't be cross bought.
fixes #8188
fixes #8203


lib/engine/game/g_1848/game.rb
```diff
@@ -282,7 +282,7 @@ class Game < Game::Base
                 distance: 999,
                 price: 1100,
                 num: 20,
-                discount: { '4' => 300, '5' => 300, '6' => 300 },
+                discount: { '4' => 300, '4+' => 300, '5' => 300, '5+' => 300, '6+' => 300, '6' => 300 },
```
bentziaxl, line 285: suggestions on how to fix the method in engine/train.rb so i don't have to specify variants?
or maybe it makes sense to explicitly list them, as there might be games where the discount is different across variants of trains?

crericha: I think it makes sense to explicitly list them

lib/engine/game/g_1848/game.rb
```diff
@@ -966,6 +966,15 @@ def first_column?(entity)
           _r, c = entity.share_price.coordinates
           c == 1
         end
+
+        def next_round!
+          reset_company_values if @round.is_a?(Engine::Round::Auction)
```
crericha, line 971: I think this will be called if everyone passes. Please verify.

bentziaxl: I verified, it doesn't. If everyone passes the auction round is still active.
image
 and 
 
image

PR #8222, opened 2022-08-20
Show game messages to owner. Make them nicer

daniel-sousa-me:
Since recently, the messages were being show only to the users in the game, but not to the user that created the game (this is important for the tournaments).

Since a long time ago, when the owner (not in the game) sent a message, the id would show up as the sender and the Master Mode warning would appear. Changed this behaviour.

### DIFF
--- a/models/game.rb
+++ b/models/game.rb
@@ -120,7 +120,9 @@ class Game < Base
 
   def to_h(include_actions: false, logged_in_user_id: nil)
     actions_h = include_actions ? actions.map(&:to_h) : []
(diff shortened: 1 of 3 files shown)


models/game.rb
```diff
@@ -120,7 +120,9 @@ def archive!
 
   def to_h(include_actions: false, logged_in_user_id: nil)
     actions_h = include_actions ? actions.map(&:to_h) : []
-    actions_h.reject! { |a| a['type'] == 'message' } unless players.find { |p| p.id == logged_in_user_id }
+    unless players.find { |p| p.id == logged_in_user_id } || user_id == logged_in_user_id
```
tobymao, line 123: don't use unless with multiple arguments, switch it to an if

PR #8223, opened 2022-08-20
[1822PNW] Private 19

outsidePasser:


lib/engine/game/g_1822_pnw/game.rb
```diff
@@ -627,6 +709,18 @@ def choices_entities
           []
         end
 
+        def check_connected(route, corporation)
+          return if route.ordered_paths.each_cons(2).all? do |a, b|
+                      a.connects_to?(b,
```
tobymao, line 714: i don't like this formatting can you do

a.connects_to(
  b,
)

outsidePasser: Changed formatting - I didn't particular like that formatting either, but that's what rubocop did.

PR #8230, opened 2022-08-21
1871 future labels

ventusignis:
Adds future labels for T,  X, and CX tiles and removes the custom logic in 1871. Also adds a fixture for 1871

image
image

`upgrades_to?` could be removed by making the straight tile power a SpecialTile action, but that would break existing games and be fairly intrusive so I'm not comfortable making that change.
However, since there are no green tiles in the game that the the straight yellow (9) tile can be upgraded to, it is possible to remove the custom `all_potential_upgrades` method without changing any functionality

@dfannius , @kelsin could you please look this over to see if you're ok with these changes? (I'm not sure who is leading/maintaining 1871)


dfannius:
I'm not familiar with the future-labels feature, and all of the code being changed was written by @kelsin, but based on my limited knowledge the commit looks okay to me.

kelsin:
Yeah I'm not in tune with the latest changes but as long as this still works correctly with the right tiles available to the right people at the right times that's fine.

PR #8240, opened 2022-08-23
Show messages as redacted instead of removing them

daniel-sousa-me:
Removing the message actions from the array makes the code reassign action numbers, which then become inconsistent. At the very least it is breaking the history, but it is very likely also breaking games with targeted undos.

It is also breaking some important stuff over at 18xx.tools, because the 18xx.games code starts behaving haphazardly.

Try navigating in the history here: https://18xx.games/game/94459?action=6

### DIFF
--- a/models/game.rb
+++ b/models/game.rb
@@ -126,7 +126,14 @@ class Game < Base
   def to_h(include_actions: false, logged_in_user_id: nil)
     actions_h = include_actions ? actions.map(&:to_h) : []
     if !players.find { |p| p.id == logged_in_user_id } && user_id != logged_in_user_id
-      actions_h.reject! { |a| a['type'] == 'message' }
+      actions_h.map! do |a|
+        if a['type'] == 'message'
+          a['user'] = user_id # Assign all messages to the creator of the game to not leak who sent it
+          a['message'] = '(Message redacted)'
+        end
+
+        a
+      end
     end
     settings_h = settings.to_h
 


models/game.rb
```diff
@@ -126,7 +126,14 @@ def archive!
   def to_h(include_actions: false, logged_in_user_id: nil)
     actions_h = include_actions ? actions.map(&:to_h) : []
     if !players.find { |p| p.id == logged_in_user_id } && user_id != logged_in_user_id
-      actions_h.reject! { |a| a['type'] == 'message' }
+      actions_h.map! do |a|
+        if a['type'] == 'message'
+          a['user'] = user_id
```
tobymao, line 131: why is the user being assigned here?

daniel-sousa-me: Just to hide who sent the message

tobymao: ah ok, can you put a quick comment to signal that

tobymao:
@crericha can you take a look?

crericha:
The 18xx.games engine has been changed over to not assuming actions
increment by one. It looks like you caught a bug and I'd rather fix that
than add messages back in. Redacted is unnecessary clutter.

tobymao:
thanks for the fix and catch @daniel-sousa-me but it looks like @crericha will fix this at the root, closing for now

daniel-sousa-me:
Can you please fix it then? The actions are being assigned new numbers in every game I looked.

crericha:
Yes, I will fix this issue. Please submit a ticket if you haven't done so
already. The couple of games that I tested when fixing history worked
great, so this is probably an edge case I missed (maybe a common one).

PR #8250, opened 2022-08-28
18Rhl: Fix Trajekt handling in Rhine Metropoles (fix #7834)

perwestling:
Redesigned the three green Rhine Metropole hexes to be 2+1/1+2
cities in appearance, but work as one city with 3 slots in
practice. This means that a green such hex that has one or
more slots untokened are passable by everyone. And that such
green hex that are fully tokened are passable for the 3
corporations that have a token in any slot on the hex.

To accomplish this there is a special "make neutral/blocked"
procedure at suitable times during the operation round.
I.e. at start/end of OR, and then any of these 3 hexes
are tokened/upgraded.

Also, fixed upgrade of green Rhine Metropolis hexes.

The methods for route handling, bonuses, distance, revenues
has been rewritten to handle these cases.

Also corrected some minor bugs.

### DIFF
--- a/lib/engine/game/g_18_rhl/game.rb
+++ b/lib/engine/game/g_18_rhl/game.rb
@@ -563,6 +563,9 @@ module Engine
             # Moers can be upgraded to gray 950
             return to.name == '950' if from.color == :brown && from.hex.name == 'D7'
 
+            # Dusseldorf and Cologne can be upgraded to gray 932V
+            return to.name == '932V' if from.color == :brown && %w[F9 I10].include?(from.hex.name)
+
             # Duisburg can be upgraded to gray 929
             return to.name == '949' if from.color == :brown && from.hex.name == 'D9'
           elsif from.color == :brown && %w[D9 F9 I10].include?(from.hex.name)
--- a/lib/engine/game/g_18_rhl/game.rb
+++ b/lib/engine/game/g_18_rhl/game.rb
@@ -643,8 +643,7 @@ module Engine
 
         def illegal_revisit_of_rhine_metropolis_hex?(actual_visits)
           actual_visits.select { |v| RHINE_METROPOLIS_HEXES.include?(v.hex.name) && v.hex.tile.color == :green }
-                       .group_by { |v| v.hex.name }.reject { |_, e| e.size < 2 }
-                       .any?
+                       .group_by { |v| v.hex.name }.any? { |_, e| e.size >= 2 }
         end
 
         def check_distance(route, visits)
@@ -887,7 +886,7 @@ module Engine
         end
 
         def get_token(entity, token)
-          return token unless token.nil?
+          return token if token
 
           # Due to changing the token type, this can cause problems when doing undo.
           # As a fall back assume first available token of type normal/neutral is OK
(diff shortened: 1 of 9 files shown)


perwestling:
At this point in time there are 2 games ongoing, both in SR 8.  I think they can be pinned to let them complete. Any finished games can be archived.

This change do break any existing games that has reached green phase and upgraded a Rhine Metropolis.

lib/engine/game/g_18_rhl/game.rb
```diff
@@ -623,79 +641,167 @@ def event_remove_tile_block!
           yellow_block_hex.tile.icons.reject! { |i| i.name == 'green_hex' }
         end
 
+        def illegal_revisit_of_rhine_metropolis_hex?(actual_visits)
+          actual_visits.select { |v| RHINE_METROPOLIS_HEXES.include?(v.hex.name) && v.hex.tile.color == :green }
+                       .group_by { |v| v.hex.name }.reject { |_, e| e.size < 2 }
+                       .any?
```
tobymao, line 647: it should be group_by.any? {...} not reject? any

lib/engine/game/g_18_rhl/game.rb
```diff
@@ -757,8 +863,50 @@ def player_value(player)
...
+              entity_tokens = tokens.count { |t| t.corporation == entity }
+              type = entity_tokens.positive? ? :neutral : :normal
+            end
+
+            tokens.each { |t| t.type = type }
+          end
+        end
+
+        def get_token(entity, token)
+          return token unless token.nil?
```
tobymao, line 890: if token

lib/engine/game/g_18_rhl/game.rb
```diff
@@ -905,63 +1013,19 @@ def iron_rhine_bonus(stops)
...
-          doubles = 0
-          stops.each do |s|
-            next unless RHINE_METROPOLIS_HEXES.include?(s.hex.name)
-
-            doubles += 1
-            bonus[:revenue] += s.route_revenue(route.phase, route.train)
-          end
-          return bonus unless doubles.positive?
+          doubles = revenue_stops.select { |rs| rs[:rhine_metropolis_bonus] }
+          return bonus unless doubles.any?
```
tobymao, line 1025: if doubles.empty?

lib/engine/game/g_18_rhl/game.rb
perwestling: Fixed

lib/engine/game/g_18_rhl/game.rb
perwestling: Fixed

lib/engine/game/g_18_rhl/game.rb
perwestling: Fixed

perwestling:
@tobymao: Have now fixed your comments.
Also updated Welcome message to make players aware that 18Rhl needs alpha testing.

Right now there is only two active 18Rhl games, and they are both in the last (9th) OR set. So depending on when this is deployed, pinning those + archive the rest, or archive all, might be the proposed action.

PR #8282, opened 2022-09-05
[1848] fix EMR text

### LINKED ISSUE #8263: 1848 incorrect EMR message
The EMR message should not even mention how much cash the president has, nor should it say how much cash the president needs to contribute.

https://18xx.games/game/93503?action=512

Screen Shot 2022-08-31 at 18 37 05

bentziaxl:
fixes #8263

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -281,7 +281,11 @@ module View
           children.concat(render_president_contributions)
         end
 
-        children << h(:div, @step.ebuy_custom_message(@corporation)) if @step.respond_to?(:ebuy_custom_message)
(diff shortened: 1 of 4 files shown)


assets/app/view/game/buy_trains.rb
```diff
@@ -281,6 +281,8 @@ def render
           children.concat(render_president_contributions)
         end
 
+        children << h(:div, @step.ebuy_custom_message(@corporation)) if @step.respond_to?(:ebuy_custom_message)
```
crericha, line 284: I'm not a fan of half solutions. If we need to move ebuy text into steps (which I do think is a good idea), let's move the default text into the train module and call it ebuy_message.

crericha (approved):

tobymao:
this change has been reverted because it was untested and broke the ui

PR #8284, opened 2022-09-05
[18JP-T] Fix train discount for TC once ability is available.  Fixes #8243

### LINKED ISSUE #8243: [18JP-T] Train discount not working for TC corporation.
What's your Game ID?
https://18xx.games/game/89711

**How to Reproduce the Bug?**

==> TC builds to A93 and tokens - this should enable the discount.  This is indicated in the logs as if it was activated.

[18:24] ulukai operates TC
[19:58] TC lays tile #4 with rotation 1 on D90 (Futakotamagawa)
[19:58] TC lays tile #4 with rotation 1 on C91 (Mizonoguchi)
[19:58] TC lays tile #9 with rotation 1 on B92
[19:58] TC places a token on A93 (Nagatsuta) for ¥40
[19:58] -- TC activates its ability: Buy trains at 20% discount --
[19:59] TC runs a 3 train for ¥50: C91-D90-F88
[20:03] TC pays out ¥50 = ¥5 per share (¥30 to ulukai)
[20:03] TC's share price moves right from ¥90 to ¥100

=> However - 6 trains still show at cost of ¥630. Discount should be applied.
Upon purchase discount is not applied:

[09:23] TC buys a 6 train for ¥630 from The Depot
[09:23] -- Phase 6 (Operating Rounds: 3 | Train Limit: 2 | Available Tiles: Yellow, Green, Brown) --
[09:23] -- Event: 3 trains rust ( KC x1, TR x1, The Depot x1, SER x1, KER x1, OER x1, TC x1) --

Thanks

timpelican:
Train discount wasn't being applied, as the '`include?`' in line 305 was evaluating true with no `selected_company`.  Checking that there is a `selected_company` allows the elsif at 312/313 to evaluate, and use the TC company discount.

This still left emergency fund-raising triggered, as the default `depot.min_depot_price` doesn't take discounts into account.  I've created `G18JPT::Step::BuyTrain`, which implements a `cheapest_train_price` that understands the TC's discount ability.

Works on the exported version of the game I hit the bug in, and passes rubocop.

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -302,7 +302,7 @@ module View
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             entity = @corporation
 
-            if @selected_company && [@corporation, @corporation.owner].include?(@selected_company&.owner) \
+            if @selected_company && [@corporation, @corporation.owner].include?(@selected_company.owner) \
               && @step.respond_to?(:ability_timing)
               @game.abilities(@selected_company, :train_discount, time: @step.ability_timing) do |ability|
                 if ability.trains.include?(train.name)
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -302,7 +302,8 @@ module View
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             entity = @corporation
 
-            if [@corporation, @corporation.owner].include?(@selected_company&.owner) && @step.respond_to?(:ability_timing)
+            if @selected_company && [@corporation, @corporation.owner].include?(@selected_company&.owner) \
+              && @step.respond_to?(:ability_timing)
               @game.abilities(@selected_company, :train_discount, time: @step.ability_timing) do |ability|
                 if ability.trains.include?(train.name)
                   price = ability.discounted_price(train, price)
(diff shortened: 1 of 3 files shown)


assets/app/view/game/buy_trains.rb
```diff
@@ -302,7 +302,8 @@ def from_depot(depot_trains, corporation)
             president_assist, _fee = @game.president_assisted_buy(@corporation, train, price)
             entity = @corporation
 
-            if [@corporation, @corporation.owner].include?(@selected_company&.owner) && @step.respond_to?(:ability_timing)
+            if @selected_company && [@corporation, @corporation.owner].include?(@selected_company&.owner) \
```
tobymao, line 305: the & after selected_company&.owner is no longer necessary

tobymao: why is this necessary? how can a corporation have no owner?

timpelican: I hadn't noticed that, thanks.  It WAS needed when `@selected_company` could be nil (no private company selected to use its power), and hence trying to call `nil.owner`, rather than a company having no owner.  Now we're short-circuiting earlier for `@selected_company` actually being a company, that can't happen.  I'll tidy that up.

timpelican: I did some more digging, and the reason the logic goes wrong in the first place looks to be that Engine::Player overrides '==', and that override method can return nil if it's being compared to nil, rather than true or false.  I suspect that's confusing include?, and making it think that @corporation.owner == @selected_company&.owner is true, when @selected_company (and hence @selected_company&.owner) is nil.

@corporation.owner == nil definitely returns nil, rather than false.

It feels like '==' should always return true or false, but I'm enough of a noob that I could be *very* wrong on that.

That feels broader than just addressing this particular bug - what's the best way to discuss?

tobymao: we can chat about it on slack if you want

PR #8305, opened 2022-09-07
Refactor game queries to return only what page uses

crericha:


models/game.rb
```diff
@@ -89,9 +85,19 @@ def self.home_games(user, **opts)
...
+    kwargs[:limit] = 1000
     fetch(user ? LOGGED_IN_QUERY : LOGGED_OUT_QUERY, **kwargs,).all.sort_by(&:id).reverse
   end
 
+  def self.profile_games(user)
+    kwargs = {}
+    kwargs[:user_id] = user.id if user
+    kwargs[:status] = %w[new active archived finished]
+    kwargs[:limit] = 100
+    fetch(USER_QUERY, **kwargs,).all.sort_by(&:id).reverse
```
tobymao, line 98: why is it doing a sort_by reverse here, it can be done in sql, fix that here and line 90?

models/game.rb
```diff
@@ -89,9 +85,19 @@ def self.home_games(user, **opts)
...
     kwargs[:user_id] = user.id if user
     kwargs[:title] = opts['title'] != '' ? opts['title'] : nil
+    kwargs[:status] = %w[new active]
+    kwargs[:limit] = 1000
     fetch(user ? LOGGED_IN_QUERY : LOGGED_OUT_QUERY, **kwargs,).all.sort_by(&:id).reverse
   end
 
+  def self.profile_games(user)
+    kwargs = {}
+    kwargs[:user_id] = user.id if user
```
tobymao, line 95: why wouldn't there be a user?

tobymao: fetch(user_query, user_id: user.id)... we don't need the kwargs building

models/game.rb
crericha: On line 90, its multiple queries combined together. Is it possible to sort those after the unions?

models/game.rb
crericha: copy and paste

crericha: Fixed

models/game.rb
crericha: I don't think the sort order even matters because the page itself orders by updated_at. I just removed them.

PR #8307, opened 2022-09-08
[1848] D Variant Discount on Info Page

### LINKED ISSUE #8261: [1848] Info page should mention discounts for diesels
The info page for 1830 notes the discount in the price of a diesel for upgrading a train.  The info page for 1848 doesn't mention this.

Yes, I know I ought to read the rulebook, but I often rely on the info page instead!

andrewzwicky:
Fixes #8261.  Right now, 1848 is the only title this impacts.

18KA & 18SJ both have a discount on a variant, but the same discount on the whole group, so this is properly displayed.

Currently, this will display the discounted train in brackets at the front of the line, and in the future case where multiple variants have different discounts, it will be separated by a semicolon.

image


assets/app/view/game/game_info.rb
```diff
@@ -169,7 +169,7 @@ def trains
         rust_schedule, obsolete_schedule = rust_obsolete_schedule
 
         show_obsolete_schedule = !obsolete_schedule.keys.empty?
-        show_upgrade = @depot.upcoming.any?(&:discount)
+        show_upgrade = @depot.upcoming.map(&:variants).map(&:values).flatten.any? { |train| train['discount'] }
```
tobymao, line 172: you don't need the map map flatten, just do this in one any?

any| variants|
   any? variants.values

andrewzwicky: I could swap it for something like this?

```ruby
show_upgrade = @depot.upcoming.any? { |train| train.variants.any? { |_k, v| v['discount'] }}
```

I'm not seeing a way to do it with a single any?  `variants` is always populated, even if there is only a single variant, and sometimes there are multiple variants, but none of them have a discount.

andrewzwicky:
I added another commit to fix the Rubocop warning, sorry about that.  It's a really simple fix to add a space.  I don't know the best way to get that merged or just make a new commit or what.

https://github.com/andrewzwicky/18xx/commit/82a49e8ff7c63bcfb6915463eb34d947de671bab

tobymao:
already pushed it

PR #8312, opened 2022-09-08
Explicitly sort home page games

crericha:


tobymao (approved):

assets/app/view/home.rb
```diff
@@ -49,9 +49,9 @@ def render
 
       render_row(children, 'Your Games', your_games, :personal) if @user
       render_row(children, 'Hotseat Games', hotseat, :hotseat) if hotseat.any?
-      render_row(children, 'New Games', grouped['new'], :new) if @user
-      render_row(children, 'Active Games', grouped['active'], :active)
-      render_row(children, 'Finished Games', grouped['finished'], :finished)
+      render_row(children, 'New Games', grouped['new'].sort_by { |g| -g['id'] }, :new) if @user
+      render_row(children, 'Active Games', grouped['active'].sort_by { |g| -g['updated_at'] }, :active)
```
tobymao, line 53: these sort orders look weird, i don't think it will work well because you only get a slice of them

crericha: Added sorts to model too so pagination works

PR #8331, opened 2022-09-12
Fixed typos on welcome page

jbecker7:
Alpha and beta should be lowercase in this context, so I adjusted them.

I also fixed a verb that was missing an "s" and removed an incorrect comma.


perwestling:
I managed to write the wrong issue - so ignore the above.

PR #8356, opened 2022-09-18
18GB: Implement the finalised 2nd Edition

andypymont:
- Incorporate the changes from the 2E test variant that made it into the finalised 2nd Edition into all 18GB games going forward.
- Drop the '2E playtest variant' from the site along with the code associated with the tested changes that didn't make it into 2nd Edition.
- One of the existing test fixtures was updated to account for the changes but the other was too radically impacted and has been removed. A new fixture was added to compensate. It would still be good to add another in future for a large-player-count game and further protect against regression.


andypymont:
I anticipate that games will need to be archived when this PR is merged:
- This will be a breaking change for all 2E playtest variant games (due to the changes to the train types and phases)
- It will also be a breaking change for some non-2E-playtest games, in particular any where tiles were placed through inactive corporation home tiles and any where tokens were placed after end-game restrictions will now come into effect.

Issue #8351, opened 2022-09-17
[1822PNW] Should P19 block mergers between minors?


setrajonas:
https://18xx.games/game/97938?action=453

M2 has P19 and built the mountain pass to connect with M1, but was unable to merge in the subsequent merger round. From the rules it seems like this merger should still be allowed.

setrajonas:
Update: it seems like the mountain pass is glitchy in general, M2 also can't lay track beyond it or run through it.

outsidePasser:
It was deliberately written so that you cannot lay track through it.  There is now clarification from the designed that you can lay track through it.

I have seen it reported that the owner cannot run through it, but I have not been able to reproduce it.

Regardless, a fix is being worked to address both of these issues.

Issue #8340, opened 2022-09-13
[1822PNW] add logos to associated minor charters


RedBrand88:
**What's your idea?**
If you can add the Major company logos to the associated minor's charters it will be more obvious to new players which minors are associated. Minor 20 just has a very dark gray header and it wasn't obvious to me that it was associated with the NP major. Just a thought not completely necessary once you've played a couple of times it wont be as necessary but overall might be a better experience for newer players.

PR #8362, opened 2022-09-21
1822PNW Fix coal token

outsidePasser:
Fixes: #8351

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -174,6 +174,11 @@ module View
           children << h(:img, logo_props)
         end
 
+        if @game.second_icon(@corporation)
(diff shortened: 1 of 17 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -174,6 +174,11 @@ def render_title(bg = nil)
           children << h(:img, logo_props)
         end
 
+        if @game.second_icon(@corporation)
```
tobymao, line 177: why is this needed?

outsidePasser: This was put in so that an icon can be displayed on associated minors to make it clearer which major that minor will grow up into.  It was requested in #8340, and I thought it sounded like a good idea.

outsidePasser:
Decided to not pin - it will break a lot of games, but I don't want to chase bugs that are already fixed.

PR #8382, opened 2022-09-25
[1873] Disable autopass in draft if passing is not a valid action.

### LINKED ISSUE #8345: [Harzbahn 1873] Enabling Pass in Draft while winning Initial Auction prevents other players from passing in the auction.
https://18xx.games/game/97726?action=8

Harzbahn begins with a bid for turn order, followed by a draft of mines and concessions.

Scenario: physicssquirrel bid 90 in the initial auction, and then enabled "Pass in Draft until premium is 20"

Expected: physicssquirrel should be forbidden from setting "Pass in Draft" because he is currently winning the auction and if he wins the auction, he may not pass; he must buy something. 

Actual: Skanadron is presented the option to overbid or pass, but the pass button does nothing and hangs the browser tab.

magnusrk-ch:
The winner of the premium auction must buy something and cannot pass. If a player set autopass in draft before winning the auction, this would trigger an illegal move and break the game.

fixes #8345

### DIFF
--- a/lib/engine/game/g_1873/step/draft.rb
+++ b/lib/engine/game/g_1873/step/draft.rb
@@ -134,6 +134,10 @@ module Engine
               end
             end
 
+            unless actions(entity).include?('pass')
+              return [Action::ProgramDisable.new(entity, reason: 'Auction winner cannot pass')]
+            end
+
             [Action::Pass.new(entity)]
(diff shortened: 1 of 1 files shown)


tobymao:
@roseundy can you take a look

Issue #7891, opened 2022-06-08
Destination tokens do not support simplified logos


perwestling:
Trying out 1822MX. When I toggle simplified logos it is still very hard to make out what some of the tokens are. Is it possible to improve?

![image](https://user-images.githubusercontent.com/2631151/172637465-ddf23928-72d1-4070-8560-3a94aa5e88a4.png)

crericha:
Destination tokens in all games do not support simplified logos.

andrewzwicky:
This can get the 18JP-T label if that matters.  I think the 1822 stuff will benefit from someone more familiar with those games, I'm not sure how the destination tokens work and because those games don't have simplified logos yet, that's probably the first step.

PR #8400, opened 2022-10-02
[1870][18JP-T] Partial fix for simplifed logo destination tokens

andrewzwicky:
part of #7891 .  This does not address 1822 because I need more time to figure out the intricacies of those tokens and they don't use this same code path.

|       | Old | New |
| ----------- | ----------- | ----------- |
| 1870   |  1870_bad  |  1870_fixed | 
| 18JP-T   |  18jp_bad  | 18jp_fixed   |

I couldn't see any obvious changes in the base 1832 map, but it's in pre-alpha and I wasn't able to get far enough in a hotseat game to see any changes.  I don't see any reason it wouldn't work the same as the others as a drop-in replacement.


lib/engine/game/base.rb
```diff
@@ -1988,7 +1988,13 @@ def show_progress_bar?
 
       def progress_information; end
 
-      def assignment_tokens(assignment)
+      def assignment_tokens(assignment, simple_logos = false)
+        if assignment.is_a?(Engine::Corporation)
+          return assignment.simple_logo if simple_logos && !assignment.simple_logo.nil?
```
tobymao, line 1993: if simple_logos && assignment.simple_logo

PR #8419, opened 2022-10-04
Fix river partitions rendering on map URL

### LINKED ISSUE #7387: river partitions don't render on the /map URL
The river partition in 1870 doesn't render on the map at https://18xx.games/map/1870 but it does render in game.

andrewzwicky:
Fixes #7387

1. Tile in RHL had two of the same label on it.

image -----> image

2. In the map view, `@game` is nil, so CitySlot needs to handle a nil game (18Rhl map view was previously throwing an undefined method error)

image

3. In the map view, `@game` is nil, so partitions should render if that's the case, so they appear on the map view.
image

### DIFF
--- a/assets/app/view/game/part/partitions.rb
+++ b/assets/app/view/game/part/partitions.rb
@@ -54,10 +54,7 @@ module View
 
           @tile.partitions.each do |partition|
             next if partition.type != :divider && partition.blockers.none? do |blocker|
-              is_blocked = @game&.abilities(blocker, :blocks_partition)&.blocks?(partition.type)
-              puts @tile.name
-              puts is_blocked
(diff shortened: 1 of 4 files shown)


assets/app/view/game/part/partitions.rb
```diff
@@ -54,7 +54,7 @@ def render_part
 
           @tile.partitions.each do |partition|
             next if partition.type != :divider && partition.blockers.none? do |blocker|
-              @game&.abilities(blocker, :blocks_partition)&.blocks?(partition.type)
+              @game.nil? || @game&.abilities(blocker, :blocks_partition)&.blocks?(partition.type)
```
tobymao, line 57: !@game

PR #8424, opened 2022-10-05
Show variants in rust schedule / properly hide obsolete schedule

### LINKED ISSUE #6987: Show variant trains in rusting schedule
Only the primary train type is shown, not the variant train. For example in 1846:
https://18xx.games/game/71607?action=0#info

image

andrewzwicky:
Fixes #6987

While I was fixing this, I noticed that the obsolete schedule wasn't properly being hidden.  Even if there no obsolete in a game, that dictionary still has `nil` as a key.

| | 1846    |
| ----------- | ----------- |
| Old | image  |   
| New | image |

| | 1822 PNW |   
| ----------- | ----------- |
| Old | image  |   
| New | image |

| | 18Scan |   
| ----------- | ----------- |
| Old | image  |   
| New | image | 

| | 1862|   
| ----------- | ----------- |
| Old | image   |   
| New | image  |

### DIFF
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -151,7 +151,14 @@ module View
         obsolete_schedule = {}
         @depot.trains.group_by(&:name).each do |_name, trains|
           first = trains.first
+
+          base_rust = first.variants.first[1][:rusts_on]
+          base_obsolete = first.variants.first[1][:obsolete_on]
+
           first.variants.each do |name, train_variant|
+            train_variant[:rusts_on] = train_variant[:rusts_on] || base_rust
+            train_variant[:obsolete_on] = train_variant[:obsolete_on] || base_obsolete
+
             unless Array(rust_schedule[train_variant[:rusts_on]]).include?(name)
               rust_schedule[train_variant[:rusts_on]] =
                 Array(rust_schedule[train_variant[:rusts_on]]).append(name)
--- a/assets/app/view/game/game_info.rb
+++ b/assets/app/view/game/game_info.rb
@@ -181,7 +181,7 @@ module View
       def trains
         rust_schedule, obsolete_schedule = rust_obsolete_schedule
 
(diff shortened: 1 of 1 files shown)


assets/app/view/game/game_info.rb
```diff
@@ -151,7 +151,20 @@ def rust_obsolete_schedule
         obsolete_schedule = {}
         @depot.trains.group_by(&:name).each do |_name, trains|
           first = trains.first
+
+          # the first variant in this list should be the base variant
+          # per the train variant initialization.  If other variants
+          # do not have an explicit rust/obsolete, inherit the base
+          # values for display purposes.
+          _, base_variant = first.variants.first
```
tobymao, line 159: variants[0]

assets/app/view/game/game_info.rb
```diff
@@ -168,7 +181,7 @@ def rust_obsolete_schedule
       def trains
         rust_schedule, obsolete_schedule = rust_obsolete_schedule
 
-        show_obsolete_schedule = !obsolete_schedule.keys.empty?
+        show_obsolete_schedule = !obsolete_schedule.keys.compact.empty?
```
tobymao, line 184: you can just do obsolete_scheduler.keys.any?

assets/app/view/game/game_info.rb
andrewzwicky: [0] will be a little more complicated because it will need to check that .variants is not empty.  .first correctly sets base_variant to nil if .variants is empty.

andrewzwicky: which doesn't really matter that much because we won't end up using the nil if .variants is empty anyway.

andrewzwicky: I was confused just now because variants is a hash, not an array. I changed to:
```ruby
base_variant = first.variants.values[0]
```

PR #8453, opened 2022-10-16
[1822PNW] Fix currency format for floats

outsidePasser:
Fixes: #8449

PR #8460, opened 2022-10-21
[1822MX] Fix pound sign in C7 description

vandamm:
* Replace pound sign in C7 description


vandamm:
Initially this PR also included a fix that is done in #8453

crericha (approved):

PR #8465, opened 2022-10-24
18SJ: Correct pay out to corporations [fixes #8381]

### LINKED ISSUE #8381: [18SJ] Full cap company gets revenue from IPO shares instead of treasury
Example: https://18xx.games/game/93437

KFJ is a full cap company with 4 shares in IPO. When it runs, those 4 shares pay to KFJ's treasury. They should not.

Additionally, if KFJ redeems a share, so it has 3 shares in IPO and 1 share in treasury, it gets revenue for 3 shares, meaning the treasury share does *not* pay. It should.

> Dividends for shares in a corporation’s treasury are paid to the treasury. Dividends for shares in the Bank Pool or initial offering are kept by the bank.

perwestling:
In Oscarian age, bank shares pay to corporation, not IPO shares.

In base game, full cap corporations do not get any pay for shares, while incremental cap corporations only get paid for shares in treasury.

### DIFF
--- a/lib/engine/game/g_18_sj/step/dividend.rb
+++ b/lib/engine/game/g_18_sj/step/dividend.rb
@@ -49,6 +49,18 @@ module Engine
 
             super
           end
+
+          def corporation_dividends(entity, per_share)
+            return 0 if entity.minor?
+            # For Oscarian era shares in the Bank pool pay to corporation.
+            # Corporations cannot have shares in Treasury.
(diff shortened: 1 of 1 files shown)


perwestling:
This breaks some games of 18SJ. Cannot be fixed as it affects the treasury of some corporations, as well as potential game end.
So I suggest pinning any active games of 18SJ - at least if broken.
For completed games - any broken games there could be archived.

PR #8490, opened 2022-11-01
Tiles tab: don't show potential upgrades with non-matching cities

### LINKED ISSUE #8488: 18GB - Incorrect Upgrades Shown
Go to this link:
https://18xx.games/game/99513?action=0#tiles

Extra (incorrect) upgrades are shown for the yellow XX and OO tiles. I think the others are correct.

For example, this one should not upgrade to G06 or G11.

image

andypymont:
Update the `upgrades_to?` check in the base game definition to account not just for the quantities of city/towns on the two tiles being compared, but also that each city/town on the old tile is present (in terms of a city/town with the same combination of edge connections) on the new tile.

- This avoids the issue of illegal multi-city tile upgrades being visible in the tile manifest when clicking on a tile to check valid upgrades - which closes #8488 
- To avoid adding an overhead to every upgrade check in all games, this check is only performed for multi-city tiles (e.g. OO tiles) - single-city tiles are already well-covered by the existing check that all edge connections are preserved
- This also means fewer illegal-to-place tiles are passed into the `legal_tile_rotations` check in track-laying step to be filtered at that stage. That check still catches elements such as track being pointed at impassable tiles or the edge of the map.

### DIFF
--- a/lib/engine/tile.rb
+++ b/lib/engine/tile.rb
@@ -395,15 +395,11 @@ module Engine
       ct_edges.values
     end
 
-    def city_town_edges_match?(city_edges, other_city_edges)
-      city_edges.all? { |e| other_city_edges.include?(e) }
-    end
-
     def city_town_edges_are_subset_of?(other_cte)
       cte = city_town_edges
       ALL_EDGES.any? do |rotation|
         rotated = other_cte.map { |city| city.map { |edge| rotate(edge, rotation) } }
-        cte.all? { |city| rotated.any? { |other_city| city_town_edges_match?(city, other_city) } }
+        cte.all? { |city| rotated.any? { |other_city| (city - other_city).empty? } }
       end
     end
 
--- a/lib/engine/tile.rb
+++ b/lib/engine/tile.rb
@@ -398,8 +398,11 @@ module Engine
     def city_town_edges_are_subset_of?(other_cte)
       cte = city_town_edges
       ALL_EDGES.any? do |rotation|
-        rotated = other_cte.map { |city| city.map { |edge| rotate(edge, rotation) } }
-        cte.all? { |city| rotated.any? { |other_city| (city - other_city).empty? } }
+        cte.all? do |city|
+          other_cte.any? do |other_city|
+            city.all? { |edge| other_city.include?(rotate(edge, rotation)) }
+          end
+        end
       end
     end
 
--- a/lib/engine/tile.rb
+++ b/lib/engine/tile.rb
@@ -395,6 +395,18 @@ module Engine
       ct_edges.values
     end
 
+    def city_town_edges_match?(city_edges, other_city_edges)
+      city_edges.all? { |e| other_city_edges.include?(e) }
+    end
+
+    def city_town_edges_are_subset_of?(other_cte)
+      cte = city_town_edges
+      ALL_EDGES.any? do |rotation|
+        rotated = other_cte.map { |city| city.map { |edge| rotate(edge, rotation) } }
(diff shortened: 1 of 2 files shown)


lib/engine/tile.rb
```diff
@@ -395,6 +395,14 @@ def city_town_edges
       ct_edges.values
     end
 
+    def city_town_edges_are_subset_of?(other_cte)
+      cte = city_town_edges
+      ALL_EDGES.any? do |rotation|
```
tobymao, line 400: this seems expensive what are you trying to do?

andypymont: I am checking that each city has a matching city (in terms of edge connections) on the new tile instead of simply the whole tile having matching edges. To try and mitigate the expense I had made sure this was only called for tiles with more than one city, rather than when checking every upgrade. Do you think it's still an issue given this is the case?

(The only alternative route I can see to fixing the issue is to move all the code for checking a valid rotation out of the base Tracker step into the base Game definition, and thereby give the Tiles view access to those checks, so it can check for a valid rotation. This however will require that I touch every game that has overriden any of those methods in the track-laying step or which has some tiles with only certain rotations allowed. I can do this but it's obviously a much bigger change to the codebase.)

lib/engine/tile.rb
```diff
@@ -395,6 +395,14 @@ def city_town_edges
...
+        rotated = other_cte.map { |city| city.map { |edge| rotate(edge, rotation) } }
```
tobymao, line 401: why are you calling a map of maps, isn't this returning an array of arrays?

lib/engine/tile.rb
```diff
@@ -395,6 +395,14 @@ def city_town_edges
...
+        cte.all? { |city| rotated.any? { |other_city| (city - other_city).empty? } }
```
tobymao, line 402: what's city and other city? edges?

lib/engine/tile.rb
tobymao, line 400: can this just be done inline like

ALL_EDGES.any? do |rotation|
  other_cte.any? do | city| 
       city.any ? | edge|
        ( rotate(edge, rotation) - other).empty?

lib/engine/tile.rb
andypymont: Yes, `city_town_edges` (which existed before this PR) returns an array for arrays - each item in the outer array is a city on the tile, each item in the inner array is an edge that city is connected to.

For example, for this tile:
image
it returns `[[0, 3], [2, 4]]`

The map of maps is to translate this result for each possible rotation.

lib/engine/tile.rb
andypymont: Yes, an array of integers containing the edge numbers connected to that city.

lib/engine/tile.rb
andypymont: I was trying to avoid calling `city_town_edges` repeatedly since this will have to recalculate from the cities on the tile, so is the expensive part of the calculation. Therefore I call this once and store in `cte`, and then rotate-and-compare it to the data from the other tile which has been passed in as an argument to this method.

andypymont: We need to make sure all that all edges of all cities are present on the new tile, so I guess it'd be:
```
cte = city_town_edges
ALL_EDGES.any? do |rotation|
    cte.all? do |city|
        other_cte.any? do |other_city|
            city.all? |edge| other_cte.include?(rotate(edge, rotation))
        end
    end
end
```

Is that better? It feels less readable; does the `.map`have a significant overhead we should be avoiding?

tobymao: cool

tobymao (approved):

PR #8509, opened 2022-11-12
show corporation type and escrow in 1856 spreadsheet

zeitlinger:
show corporation type and escrow in 1856 spreadsheet - this makes it very easy to see the amount of capital each company will get when you buy a share

image


crericha:
The column should be called capitalization imo.

zeitlinger:
> The column should be called capitalization imo.

good idea - done

PR #8515, opened 2022-11-14
18Dixie: ICG/SCL Merger start

ventusignis:
I'm opening this up as a draft PR to get feedback before I get too far, particularly on how to propagate merger information between steps. `round_state` works, but is a little clunky and it seems that only one step in a game can specify the method (when I had it defined in icgscl_presidency_exchange.rb, the step didn't recognize the round state variables)

### DIFF
--- a/lib/engine/game/g_18_dixie/step/icgscl_merge_consent.rb
+++ b/lib/engine/game/g_18_dixie/step/icgscl_merge_consent.rb
@@ -14,6 +14,11 @@ module Engine
               merge_consent_secondary_corp: nil,
               merge_consent_pending_corps: [],
               merge_consent_subsidy: nil,
+              merge_presidency_cash_crisis_corp: nil,
+              merge_presidency_cash_crisis_player: nil,
+              merge_presidency_exchange_corps: [],
+              merge_presidency_exchange_merging_corp: nil,
+              merge_presidency_exchange_subsidy: nil,
             }
           end
 
@@ -62,7 +67,7 @@ module Engine
           end
 
           def turn_to_choose(entity)
-            return false unless entity.corporation
+            return false unless entity.corporation?
             return false unless @round.merge_consent_merging_corp
             return false if @round.merge_consent_pending_corps.empty?
             return false unless entity == @round.merge_consent_pending_corps[0]
--- a/lib/engine/game/g_18_dixie/step/icgscl_merge_consent.rb
+++ b/lib/engine/game/g_18_dixie/step/icgscl_merge_consent.rb
@@ -1,0 +1,139 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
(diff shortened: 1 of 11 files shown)


michaeljb:
> particularly on how to propagate merger information between steps. round_state works, but is a little clunky and it seems that only one step in a game can specify the method (when I had it defined in icgscl_presidency_exchange.rb, the step didn't recognize the round state variables)

At the moment I'm not familiar with 18Dixie and I haven't started looking through the PR yet, but one idea for this is instance variables on the game class that can be reset every stock round (or whatever). I'll read the Dixie rules and take a look at the PR to come up with more specific thoughts, but wanted to post that idea first in case it's useful.

lib/engine/game/g_18_dixie/step/icgscl_merge_consent.rb
```diff
@@ -0,0 +1,144 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G18Dixie
+      module Step
+        class MergeConsent < Engine::Step::Base
+          def round_state
```
michaeljb, line 10: After perusing your changes and the [18Dixie Player Aid on BGG](https://boardgamegeek.com/filepage/132102/18dixie-player-aid), this is what I've come up with for the merger stuff.

I think if you didn't want this on the round state, you could have something like this on the game class:

```
attr_accessor :merger

@merger = {
  consent: {
    merging_corp: nil,
    primary_corp: nil,
    secondary_corp: nil,
    pending_corps: [],
    subsidy: nil,
  },
  presidency: {
    cash_crisis: {
      corp: nil,
      cash_crisis_player: nil,
    },
    exchange: {
      corps: [],
      merging_corp: nil,
      subsidy: nil,
    },
  },
}
```

All the steps would be able to access what they needed with `@game.merger[:presidency]` or whatever. Don't know if that's really any less clunky than the round state approach 🤷

lib/engine/game/g_18_dixie/trains.rb
```diff
@@ -73,20 +73,21 @@ module Trains
...
           },
           {
             name: '5D',
             available_on: '7',
             distance: [{ 'nodes' => %w[city offboard town], 'pay' => 5, 'visit' => 5, 'multiplier' => 2 }],
-            events: [], # ICG Merger Chance
+            events: [{ 'type' => 'icg_formation_chance' }],
             price: 900,
             num: 5,
+            variant_event_handling: :individual,
```
michaeljb, line 90: If I'm understanding the changes in `lib/engine/train.rb` correctly, this value prevents other variants from having the same events as this one; I see how this makes sense for the 4D, since it's a variant of the 2D, but I don't get why the 5D needs this.

tobymao:
feel free to reopen

PR #8517, opened 2022-11-14
show order of unfloated companies

zeitlinger:
makes it easier to see who is going to buy which trains

image

### DIFF
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -9,6 +9,10 @@ require 'view/game/stock_market'
 require 'view/game/tranches'
 require 'view/game/actionable'
 
+FLOATED = 2
+UNFLOATED = 1
+UNSTARTED = 0
+
 module View
   module Game
     class Spreadsheet < Snabberb::Component
@@ -337,8 +341,20 @@ module View
             @game.operating_order
           end
 
+        unfloated_corporations =
+          (@game.all_corporations - operating_corporations)
+            .filter { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
+            .sort
+
         result = @game.all_corporations.map do |c|
-          operating_order = (operating_corporations.find_index(c) || -1) + 1
+          operating_order =
+            if (index = operating_corporations.find_index(c))
+              [FLOATED, index + 1]
+            elsif (index = unfloated_corporations.find_index(c))
+              [UNFLOATED, index + 1]
+            else
+              [UNSTARTED, 0]
+            end
           [operating_order, c]
         end
 
@@ -360,7 +376,7 @@ module View
             when :market_shares
               num_shares_of(@game.share_pool, corporation)
             when :share_price
-              [corporation.share_price&.price || 0, -operating_order]
+              [corporation.share_price&.price || 0, operating_order[0], -operating_order[1]]
             when :par_price
               corporation.par_price&.price || 0
             when :cash
@@ -432,7 +448,7 @@ module View
 
         order_props = { style: { paddingLeft: '1.2em' } }
         order_props[:style][:color] =
-          if operating_order.zero?
+          if operating_order[0] == UNSTARTED
             'transparent'
           elsif corporation.operating_history.keys[-1] == current_round
             convert_hex_to_rgba(color_for(:font2), 0.5)
@@ -510,7 +526,9 @@ module View
             corporation.share_price ? @game.format_currency(corporation.share_price.price) : ''),
           h('td.padded_number', @game.format_currency(corporation.cash)),
           *treasury,
-          h('td.padded_number', order_props, operating_order),
+          h('td.padded_number', order_props, if operating_order[0] == UNFLOATED
+                                               "[#{operating_order[1]}]" else
+                                                                           operating_order[1] end),
           h(:td, corporation.trains.map { |t| t.obsolete ? "(#{t.name})" : t.name }.join(', ')),
           h(:td, @game.token_string(corporation)),
           *extra,
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -334,23 +334,25 @@ module View
       end
 
       def sorted_corporations
-        operating_corporations =
+        operating_array =
           if @game.round.operating?
             @game.round.entities
           else
             @game.operating_order
           end
+        operating_corporations = operating_array.each_with_index.to_h
 
         unfloated_corporations =
-          (@game.all_corporations - operating_corporations)
+          (@game.all_corporations - operating_array)
             .select { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
             .sort
+            .each_with_index.to_h
 
         result = @game.all_corporations.map do |c|
           operating_order =
-            if (index = operating_corporations.find_index(c))
+            if (index = operating_corporations[c])
               [FLOATED, index + 1]
-            elsif (index = unfloated_corporations.find_index(c))
+            elsif (index = unfloated_corporations[c])
               [UNFLOATED, index + 1]
             else
               [UNSTARTED, 0]
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -527,8 +527,10 @@ module View
           h('td.padded_number', @game.format_currency(corporation.cash)),
           *treasury,
           h('td.padded_number', order_props, if operating_order[0] == UNFLOATED
-                                               "[#{operating_order[1]}]" else
-                                                                           operating_order[1] end),
+                                               "[#{operating_order[1]}]"
+                                             else
+                                               operating_order[1]
+                                             end),
           h(:td, corporation.trains.map { |t| t.obsolete ? "(#{t.name})" : t.name }.join(', ')),
           h(:td, @game.token_string(corporation)),
           *extra,
--- a/assets/app/view/game/spreadsheet.rb
+++ b/assets/app/view/game/spreadsheet.rb
@@ -343,7 +343,7 @@ module View
 
         unfloated_corporations =
           (@game.all_corporations - operating_corporations)
-            .filter { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
+            .select { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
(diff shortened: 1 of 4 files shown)


zeitlinger:
looks like a flaky test, but I can't find the option to re-run the check

zeitlinger:
apparently, it's failing consistently - but I don't understand the error

benjaminxscott:
You can run `docker-compose exec rack rake` locally to re-run the tests w/o having to force-push or manually restart the job

Based on job log:
https://github.com/tobymao/18xx/actions/runs/3464134055/jobs/5785286705#step:4:1673

The failing game:
https://github.com/tobymao/18xx/blob/master/spec/fixtures/18MEX/17849.json

For context, my understanding is that rake tries to execute each test game against the engine, and if it fails to reach the expected state (in this example, the game was expected to conclude) then it bails

Presumably something in this diff caused the game to hit an error condition

zeitlinger:
> edit: it seems to be complaining that the method is undefined? In the case where it says "missing method stub" usually that's due to it calling a function that's not defined

Thanks a lot - this is also how far I've got.
I'm wondering what the best approach to reproduce the error is.
Can you replay the game locally (how does this work?) and maybe see a better error message in the browser?

benjaminxscott:
I typically import the 'fixture' (aka test game) as a 'hotseat' in your local server running from `make` on your branch

When I do the above, I see a more useful error

> "undefined method `sort_order_key' for <Engine::Minor: A>"

assets/app/view/game/spreadsheet.rb
```diff
@@ -510,7 +523,9 @@ def render_corporation(corporation, operating_order, current_round)
             corporation.share_price ? @game.format_currency(corporation.share_price.price) : ''),
           h('td.padded_number', @game.format_currency(corporation.cash)),
           *treasury,
-          h('td.padded_number', order_props, operating_order),
+          h('td.padded_number', order_props, if operating_order[0] == 1
```
michaeljb, line 526:
```suggestion
          h('td.padded_number', order_props, if operating_order[0] == UNFLOATED
```

assets/app/view/game/spreadsheet.rb
```diff
@@ -337,8 +337,21 @@ def sorted_corporations
...
-          operating_order = (operating_corporations.find_index(c) || -1) + 1
+          index = operating_corporations.find_index(c)
+          sort_order_key = sort_order.find_index(c)
+          operating_order = if index
+                              [2, index + 1]
+                            elsif sort_order_key
+                              [1, sort_order.length - sort_order_key]
+                            else
+                              [0, 0]
+                            end
```
michaeljb, line 354: I found this code a little hard to read, after a bit of testing I think the following code accomplishes the same thing, and eliminates the need for the `operating_order()` function you added below:

```
        floated_order = 0
        unfloated_order = 0

        result = @game.all_corporations.sort.map do |c|
         operating_order =
           if operating_corporations[floated_order] == c
             floated_order += 1
             [FLOATED, floated_order]
           elsif c.ipoed
             unfloated_order += 1
             [UNFLOATED, unfloated_order]
           else
             [UNSTARTED, 0]
           end
          [operating_order, c]
        end
```

Up at the time you'll need to add these constants:

```
      FLOATED = 2
      UNFLOATED = 1
      UNSTARTED = 0
```

Would you mind testing that out to see if it gives you the same results?

assets/app/view/game/spreadsheet.rb
```diff
@@ -616,6 +631,12 @@ def pd_props
 
       private
 
+      def operating_order(c, operating_corporations)
+        sort_order_key = c.sort_order_key if c.respond_to?(:sort_order_key)
+        operating_order = sort_order_key if sort_order_key && !operating_corporations.find_index(c)
+        operating_order
+      end
+
```
michaeljb, line 639:
```suggestion
```

michaeljb (changes requested):

assets/app/view/game/spreadsheet.rb
zeitlinger: Thanks for taking a look - the result looks much cleaner :smile: 

There's a failing test case, because `@game.all_corporations.sort` throws an exception for minors - as they don't implement the compare function (this was covered by `sort_order_key = c.sort_order_key if c.respond_to?(:sort_order_key)`).

The `respond_to` is actually breaking the encapsulation, so it's hacky.

michaeljb: Huh, looks like there are multiple ways different games handle the minors.

* [1822](https://github.com/tobymao/18xx/blob/ca01b6ef29aba08967ad0adf03f9915702a1d1b0/lib/engine/game/g_1822/entities.rb#L862-L864) and [1861](https://github.com/tobymao/18xx/blob/b3484429fc2a709d2a1b4aba60ffcc3b84f022fe/lib/engine/game/g_1861/game.rb#L911) - the minors are defined as `CORPORATIONS` and thus end up in `@corporations` instead of `@minors`
* [1873](https://github.com/tobymao/18xx/blob/b3484429fc2a709d2a1b4aba60ffcc3b84f022fe/lib/engine/game/g_1873/game.rb#L889-L891) and [1893](https://github.com/tobymao/18xx/blob/b3484429fc2a709d2a1b4aba60ffcc3b84f022fe/lib/engine/game/g_1893/game.rb#L745-L748) - `all_corporations` returns `@minors` + `@corporations`

I'll check out that failing test case, I'm thinking it make sense to implement `<=>` for minors like it's done on the [corporations](https://github.com/tobymao/18xx/blob/c15f2071c754697bbe905d4959889dc68d73a410/lib/engine/corporation.rb#L92-L105)

michaeljb: Hmm thinking about that a little more it probably won't work, it would need the new `<=>` method to put things in the same order as `@game.operating_order` which doesn't really work.

zeitlinger: It actually does work - I've moved the spaceship operator into the operator module. WDYT?

michaeljb: What you have here is pretty close to what I came up for that, even the new `1` at the start of the corporation array and the `0` for the minor. I don't really like magic numbers like those, and the corporation's `[1, 0, 0, 0, 0, name]` also made me wince when I typed that out.

I went back to your original `operating_order` function and `sort_order` array. I think with what I have here, `unfloated_corporations` is equivalent to your `sort_order` array. Thoughts?


```
        unfloated_corporations =
          (@game.all_corporations - operating_corporations)
            .filter { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
            .sort

        op_order = 0
        unfloated_order = 0

        result = @game.all_corporations.sort.map do |c|
          operating_order =
            if operating_corporations[op_order] == c
              op_order += 1
              [FLOATED, op_order]
            elsif unfloated_corporations[unfloated_order] == c
              unfloated_order += 1
              [UNFLOATED, unfloated_order]
            else
              [UNSTARTED, 0]
            end
          [operating_order, c]
        end
```

michaeljb: Well I guess this isn't quite right either, it's failing the test `Assets #html 18MEX 17849 renders endgame` because some minors are being compared 🤔

michaeljb: Oh I see, I still have a `sort` in the line starting with `result =`

michaeljb: Ok think this is working now:

```
        unfloated_corporations =
          (@game.all_corporations - operating_corporations)
            .filter { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
            .sort

        op_order = 0
        unfloated_order = 0

        result = @game.all_corporations.map do |c|
          operating_order =
            if (index = operating_corporations.find_index(c))
              [FLOATED, index + 1]
            elsif (index = unfloated_corporations.find_index(c))
              [UNFLOATED, index + 1]
            else
              [UNSTARTED, 0]
            end
          [operating_order, c]
        end
```

assets/app/view/game/spreadsheet.rb
```diff
@@ -360,7 +378,7 @@ def sorted_corporations
             when :market_shares
               num_shares_of(@game.share_pool, corporation)
             when :share_price
-              [corporation.share_price&.price || 0, -operating_order]
+              [corporation.share_price&.price || 0, -operating_order[1]]
```
michaeljb, line 381:
```suggestion
              [corporation.share_price&.price || 0, operating_order[0], -operating_order[1]]
```

Without `operating_order[0]`:

Screenshot 2022-11-26 123049

With it:

Screenshot 2022-11-26 123137

zeitlinger: thanks for catching :smile:

assets/app/view/game/spreadsheet.rb
zeitlinger: good idea :smile:

assets/app/view/game/spreadsheet.rb
```diff
@@ -510,7 +526,9 @@ def render_corporation(corporation, operating_order, current_round)
             corporation.share_price ? @game.format_currency(corporation.share_price.price) : ''),
           h('td.padded_number', @game.format_currency(corporation.cash)),
           *treasury,
-          h('td.padded_number', order_props, operating_order),
+          h('td.padded_number', order_props, if operating_order[0] == UNFLOATED
+                                               "[#{operating_order[1]}]" else
+                                                                           operating_order[1] end),
```
michaeljb, line 531:
```suggestion
          h('td.padded_number', order_props, if operating_order[0] == UNFLOATED
                                               "[#{operating_order[1]}]"
                                             else
                                               operating_order[1]
                                             end),
```

michaeljb (approved):

assets/app/view/game/spreadsheet.rb
```diff
@@ -337,8 +341,20 @@ def sorted_corporations
             @game.operating_order
           end
 
+        unfloated_corporations =
+          (@game.all_corporations - operating_corporations)
+            .filter { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
```
michaeljb, line 346:
```suggestion
            .select { |c| c.respond_to?(:sort_order_key) && c.sort_order_key }
```

michaeljb (approved):

PR #8543, opened 2022-11-27
Add "Shares %" and "Book Value"

### LINKED ISSUE #8438: [1856] Shares in spreadsheet for CGR are wrong
https://18xx.games/game/98691

image

"Shares" should show the total percentage a player has, so 8 means 80%.

For player "Gerd" for example, this should be 2 (20%), because the player owns the 10% director and 2 5% certificates (which count as 0.5 certificates) - but "Shares" actually say "4".

zeitlinger:
Fixes #8438

1. "Shares %" it the total percentage of shares a player owns
2. "Book Value" is `trains + tokens + cash - loans`
3. "Rusted Book Value" is the "Book Value" excluding the trains that will be rusted or obsoleted next

image

### DIFF
--- a/lib/engine/operator.rb
+++ b/lib/engine/operator.rb
@@ -85,7 +85,7 @@ module Engine
           trains = rusted_trains + obsolete_trains
 
           unless trains.empty?
-            rust_value = @trains.filter { |t| trains.include?(t.name) }.sum(&:price)
+            rust_value = @trains.sum { |t| trains.include?(t.name) ? t.price : 0 }
             break
           end
         end
--- a/lib/engine/operator.rb
+++ b/lib/engine/operator.rb
@@ -64,7 +64,11 @@ module Engine
 
     def current_book_value
       trains = @trains.sum(&:price)
-      tokens = @tokens.filter(&:used).sum { |t| @token_book_value_override || t.price }
+      tokens = if @token_book_value_override
+                 @tokens.size * @token_book_value_override
+               else
+                 @tokens.filter(&:used).sum(&:price)
+               end
       loans = @loans.sum(&:amount)
       trains + tokens + @cash - loans
     end
--- a/lib/engine/operator.rb
+++ b/lib/engine/operator.rb
@@ -62,11 +62,32 @@ module Engine
       @tokens.select(&:city)
     end
 
-    def book_value
+    def current_book_value
       trains = @trains.sum(&:price)
       tokens = @tokens.filter(&:used).sum { |t| @token_book_value_override || t.price }
       loans = @loans.sum(&:amount)
       trains + tokens + @cash - loans
+    end
+
+    def rusted_book_value(depot)
+      rust_value = 0
+
+      rust_schedule, obsolete_schedule = depot.rust_obsolete_schedule
+      unless depot.upcoming.empty?
+        depot.upcoming.group_by(&:name).map do |name, _|
+          rusted_trains = rust_schedule[name] || []
+          obsolete_trains = obsolete_schedule[name] || []
+
+          trains = rusted_trains + obsolete_trains
+
+          unless trains.empty?
+            rust_value = @trains.filter { |t| trains.include?(t.name) }.sum(&:price)
+            break
+          end
+        end
+      end
+
+      current_book_value - rust_value
     end
   end
 end
--- a/lib/engine/operator.rb
+++ b/lib/engine/operator.rb
@@ -26,6 +26,7 @@ module Engine
       @text_color = opts[:text_color] || '#ffffff'
       @destination_coordinates = opts[:destination_coordinates]
       @destination_icon = opts[:destination_icon] ? "/icons/#{opts[:destination_icon]}" : ''
+      @token_book_value_override = opts[:token_book_value_override]
     end
 
     def operator?
@@ -59,6 +60,13 @@ module Engine
 
     def placed_tokens
       @tokens.select(&:city)
+    end
+
+    def book_value
+      trains = @trains.sum(&:price)
+      tokens = @tokens.filter(&:used).sum { |t| @token_book_value_override || t.price }
+      loans = @loans.sum(&:amount)
+      trains + tokens + @cash - loans
     end
   end
 end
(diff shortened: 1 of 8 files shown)


lib/engine/operator.rb
```diff
@@ -60,5 +61,37 @@ def unplaced_tokens
     def placed_tokens
       @tokens.select(&:city)
     end
+
+    def current_book_value
+      trains = @trains.sum(&:price)
+      tokens = if @token_book_value_override
+                 @tokens.size * @token_book_value_override
+               else
+                 @tokens.filter(&:used).sum(&:price)
```
tobymao, line 70: don't use filter, use reject/select

lib/engine/operator.rb
```diff
@@ -60,5 +61,37 @@ def unplaced_tokens
...
+      rust_schedule, obsolete_schedule = depot.rust_obsolete_schedule
+      unless depot.upcoming.empty?
+        depot.upcoming.group_by(&:name).map do |name, _|
+          rusted_trains = rust_schedule[name] || []
+          obsolete_trains = obsolete_schedule[name] || []
+
+          trains = rusted_trains + obsolete_trains
+
+          unless trains.empty?
+            rust_value = @trains.filter { |t| trains.include?(t.name) }.sum(&:price)
```
tobymao, line 88: same

tobymao: in this case you can just do  @trains.sum { |t| trains.include?(t.name) ? t.price : 0 } so you only do one iteration

michaeljb:
Could the rendering be updated so that the "Book Value" and "Rusted BV" rows use `@game.format_currency` to get the `$` visible, and the "Shares %" row show the percentage? e.g., "220%" instead of "220" for the farthest right player in your screenshot

lib/engine/operator.rb
michaeljb: As in the other comment, I think it's possible to do just one iteration here as well:

```suggestion
                 @tokens.sum { |t| t.used ? t.price : 0 }
```

crericha:
I see more and more game specific spreadsheet columns being added. IMO, it's time to generalize the spreadsheet view logic for these columns, so that the view code doesn't get unwieldy.

I suggest a base game method that returns game specific columns and their values. Something like `additional_spreadsheet_columns` and have games override it.

zeitlinger:
> I see more and more game specific spreadsheet columns being added. IMO, it's time to generalize the spreadsheet view logic for these columns, so that the view code doesn't get unwieldy.
> 
> I suggest a base game method that returns game specific columns and their values. Something like `additional_spreadsheet_columns` and have games override it.

good idea - but this PR is not game specific.

crericha:
Okay, thanks for clarifying. I thought it was only for 1817, which has the concept of book value in its rulebook.

Taking a general approach will require a lot of per title overriding to avoid having incorrect book values. More than 1817 and 1862. Here are some examples that immediately come to mind:

1. 18CZ allows companies to sell their private companies back to the bank, so these are assets that should be accounted for. 1860 also lets private companies sell back to the bank during the SR. 
2. 18CZ has different rusting schedules for each company/train type pairing. Is that taken into account?
3. 18NY scraps trains when they rust. The scrap value should be part of the rusted book value. There are other games that scrap too and I don't know if they are done in the same way.

I'm sure there are many more nuances to supporting all the titles.

Regarding the implementation, we found that pushing data down into the company, operator, etc. classes doesn't really work because they are designed to be simple data classes with very basic methods on that data and for something nuanced like this, more knowledge ends up being needed. The 1862 averaging in this PR is an example. The common idiom in the code base is to use the game class to compute the value like `@game.book_value(company)`. 

What's also nice about using the game class is you can add book value support for the games you know, omit the others, and someone can come along later and add it.

roseundy:
Is anyone asking for book value? This seems pretty game-specific.

crericha:
Regarding the shares % portion of the PR, the implementation doesn't match how I read the issue. My reading is the submitter wanted the percentage per company, not the overall percentage. IMO overall percentage is meaningless.

Using the example above, for cavour's holding of SR, they were requesting `*2 (100%)`.

zeitlinger:
> Is anyone asking for book value? This seems pretty game-specific.

Yes, as a player I find that quite interesting.
Is there any concern for adding a feature that some people may like to ignore? 
If that is the case, we could find some form of hiding columns in the spreadsheet.

zeitlinger:
> Regarding the shares % portion of the PR, the implementation doesn't match how I read the issue. 

I can answer the the submitter of the issue :smile: 

> My reading is the submitter wanted the percentage per company, not the overall percentage.

This is also an interesting idea - but would made the spreadsheet much harder to read, like the `*2 (100%)` you suggested.

> IMO overall percentage is meaningless.

It it another variant of the total shares you own.

Number of shares gives an indication of the amount of future share price appreciation and shares % gives an indication of the amount future dividends.

zeitlinger:
> Okay, thanks for clarifying. I thought it was only for 1817, which has the concept of book value in its rulebook.

Yes, this is the inspiration :smile: 

> Taking a general approach will require a lot of per title overriding to avoid having incorrect book values; more than 1817 and 1862. Here are some examples that immediately come to mind:

That's a great point. I would like to start with something simple, to hear what other players are thinking about the feature - there are several ways to do this:

1. Start with a very simple implementation for all titles - and ask for feedback from players (this PR)
2. Limit to 1817 initially
3. Hide the feature behind some toggle, button, or similar - indicating that it's an early access feature

> Regarding the implementation, we found that pushing data down into the company, operator, etc. classes doesn't really work

That's great feedback - will change the PR accordingly!

tobymao:
no, we shouldn't have any options, too much complexity

so if this isn't that valuable of a feature, we should just not have it

zeitlinger:
> no, we shouldn't have any options, too much complexity
> 
> so if this isn't that valuable of a feature, we should just not have it

Hiding columns in the spreadsheet is a feature that would improve usability on mobile quite a lot - I often find myself scrolling. 

Something like this:
https://stackoverflow.com/questions/12365162/design-of-html-table-with-many-columns

crericha:
> Number of shares gives an indication of the amount of future share price appreciation and shares % gives an indication of the amount future dividends.

How does total shares % give an indication of the amount of future dividends? Companies run for different amounts, so lumping all shares % into one number doesn't give enough information to determine (or even ballpark) payouts.

tobymao:
closing as it adds too much complexity

PR #8546, opened 2022-11-27
[1861/1867] Endgame fixes: loans dropping share price, endgame timing

### LINKED ISSUE #8520: 1861: endgame calculation is wrong
https://18xx.games/game/102563

In a recently finished game of 1861, a company (GRR) ended the game with two loans outstanding. According to the rules (12.2) this should reduce their net worth one space per loan - in this case, from 200 to 165. But the final values are not taking this into account.

### LINKED ISSUE #8541: 1861: game ends incorrectly with bank busting
1861 Game #103700

As it said when the first 8-train was bought (in OR6.2), the game should have ended in OR7.3. BUT it actually ended in OR7.2, presumably due to the bank busting. This is wrong - once the 8T has been bought, the bank busting is irrelevant.

This is unclear in the GTG edition rules, and much clearer in the AAG version (see section 13, 3rd paragraph).

michaeljb:
* Drop corp share price one step per loan at game end [Fixes #8520]
    * Additionally, override `player_value` so that during the game, it reflects the player's current worth if the game were to end immediately so that the dropping share price at the end doesn't cause any surprises.
* Cache `game_end_check` so bank breaking doesn't override 8-train ending [Fixes #8541]


michaeljb:
This won't cause any pins, but will certainly change some game results, which could affect ELO ratings and tournament results (if there are any 1861/1867 tournaments). I don't know what manual steps that will require.

michaeljb:
The endgame timing fix will require pins, and the share price changing at the end will cause some results to change, which could affect ELO ratings and tournament results, and will affect finish games with results already stored in the DB.

michaeljb:
Hmm interesting, this doesn't actually invalidate any games. Instead, a game that was incorrectly finished will still appear as finished until you open it up:

Screenshot 2022-11-29 190400


Screenshot 2022-11-29 190443

I clicked through to the proper ending, and the stored result was updated:

Screenshot 2022-11-29 191705

So I guess this timing fix gives anyone with a game that ended prematurely the option to fix it, but it will also remain "finished" if left alone.

tobymao (approved):

PR #8548, opened 2022-11-28
2038: Move map code to map.rb file and add some tiles

jason-johnson:
I have moved the map code to map.rb as other titles seem to have done.  I've also begun adding an initial cut of how 2038 tiles might be modelled.  For now I just take the simplest possible thing that will let me test running actual routes with no concern on how it looks.

I also did make a couple of changes to more closely model how "off board" locations work in 2038.


lib/engine/game/g_2038/map.rb
```diff
@@ -0,0 +1,205 @@
...
+            'count' => 2,
+            'color' => 'yellow',
+            'code' => 'town=revenue:yellow_30|gray_60;town=revenue:20;town=revenue:20;path=a:1,b:_0;path=a:2,b:_0;'\
+                      'path=a:4,b:_1;path=a:5,b:_1;path=a:0,b:_2;path=a:3,b:_2',
+          },
+          '2011' =>
+          {
+            'count' => 1,
+            'color' => 'yellow',
+            'code' => 'city=revenue:yellow_30|gray_60,slots:2;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:4,b:_0;'\
```
michaeljb, line 84: Should this and the following tiles also have a `path=a:3,b:_0`?

lib/engine/game/g_2038/map.rb
```diff
@@ -0,0 +1,205 @@
...
+                      'path=a:5,b:_0;label=I',
+          },
+          '2012' =>
+          {
+            'count' => 1,
+            'color' => 'yellow',
+            'code' => 'city=revenue:80,slots:4;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:4,b:_0;'\
+                      'path=a:5,b:_0;label=I',
```
michaeljb, line 92: Is there any difference between tiles 2012 to 2022?

lib/engine/game/g_2038/map.rb
jason-johnson: Ah yes it should.

lib/engine/game/g_2038/map.rb
jason-johnson: Addressed in comment above.

michaeljb (approved):

PR #8562, opened 2022-12-05
[1868WY] private companies can be closed at any time

michaeljb:
[#5011]

### DIFF
--- a/lib/engine/game/g_1868_wy/step/manual_close_company.rb
+++ b/lib/engine/game/g_1868_wy/step/manual_close_company.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G1868WY
+      module Step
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_1868_wy/step/manual_close_company.rb
```diff
@@ -0,0 +1,51 @@
...
+      module Step
+        class ManualCloseCompany < Engine::Step::Base
+          ACTIONS = %w[manual_close_company].freeze
+
+          def description
+            'Manually Close Companies'
+          end
+
+          def actions(entity)
+            return [] unless entity.company? &&
```
tobymao, line 17: don't use unless with multiple conditions

tobymao (approved):

lib/engine/game/g_1868_wy/step/manual_close_company.rb
michaeljb: Hopefully this is the last time you'll need to make this comment that lol #8564

PR #8565, opened 2022-12-05
add dashes to tokens to be color blind friendly

zeitlinger:
add dashes to tokens to be color blind friendly

image

### DIFF
--- a/assets/app/lib/settings.rb
+++ b/assets/app/lib/settings.rb
@@ -86,12 +86,26 @@ module Lib
     end
 
     def player_colors(players)
+      player_route_props(players, ->(index) { route_prop(index, :color) })
+    end
+
+    def player_colors_with_dash(players)
+      player_route_props(players, lambda { |index|
+        {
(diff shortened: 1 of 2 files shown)


roseundy:
How does this play with with token highlighting?

crericha:
> How does this play with with token highlighting?

Yeah, I can see the "spokes" being a problem. Maybe experiment with a small white or black outline between the token and the player color? Basically, some way to distinguish the player color circle instead of reshaping it.

zeitlinger:
> How does this play with with token highlighting?

the problem (for me) is to find a set of colors that are easily distinguishable.
In [another game](https://www.boardgamers.space/boardgame/gaia-project), I've successfully used dashes (added selectively to colors that are similar for me) to make a clear distinction.

Here's a set of colors that would otherwise be difficult for me:

image

jason-johnson:
Could this be added as a profile switch so people who need it can turn it on?

tobymao:
closing for now

PR #8578, opened 2022-12-10
1858 initial implementation

ollybh:
_Closing this pull request, for now at least. This has been split up into multiple smaller patches and is getting gradually merged into the upstream code._

----

This is a mostly working implementation of Ian D Wilson's game, 1858: The Railways of Iberia. It needs a lot more testing, but the core game is working. There are some significant differences between 1858 and other titles, so here's a summary of some of the strange things I have had to do in the code.

### Private companies

Private companies are fairly different in 1858 from other games. They pay an income to the players who own them like most games, but they can also lay track. Once they are owned by a corporation they briefly act as a train, giving revenue in the operating round.

~~The implementation has the companies pretending to be minors, so they can act in operating rounds. They are a subclass of Company but also include Operator, and have other methods to spoof the Minor interface.~~

~~They are added to both the @game.privates and the @game.minors arrays. The are a few places where it has been necessary to add uniq calls on top of (companies + minors + corporations) collations, to avoid duplications. In other places there are tests for entity.company? used to detect non-operating entities that have also had to be changed.~~

~~One concern with this is that there is no inheritance from Minor, so any future changes in that class won't be reflected in the G1858::Company class, so could cause breakage. There will need to be tests to try and catch anything like that.~~

_The private railways are now implemented differently, with a company and a minor entity for each one._

### Graph

There are three types of track available: broad gauge, narrow (metre) gauge and dual gauge. Routes can never include both broad and narrow gauges. To cope with this the game maintains two graphs, `@graph_broad` and `@graph_metre`. ~~Some methods have been added to the Game class for clearing graphs, replacing calls to `@game.graph.clear` with `@game.clear_graph`, so that these methods can be overridden in the G1858::Game class.~~ _[This has been merged, pull request #8651]_

~~I still ended up having to subclass Graph, as the `home_as_token` bit of Graph.compute needed modification. This needs tidying up, at the moment the whole `compute` method has been copied into the subclass. A neater solution would be to extract a couple of methods (home_hexes and home_hex_nodes) in the base class and just override these.~~ _[This has been merged, pull request #8621]_

One issue I ran into is that the interaction of an entities `coordinates` and `city` attributes is unclear when there are cities in several of the home hexes. Different games assume different things. For 1858 I've assumed that the order of arrays in both matches. So the second value in the city array give the index of the city in the hex that is second in the coordinates array.

### Stubs

~~Stubs are used to enforce the direction of yellow tiles in private company home hexes. `StubsAreRestricted.legal_if_stubbed?` used to just check for one stub, it now checks for all that are present.~~ _[This has been merged, pull request #8644]_

~~And there is now a Stubs ability as these restrictions disappear when the private company closes.~~ _[This has been removed]_

### Revenue bonuses

~~This is a purely cosmetic change. It allows for dummy trains to appear in the route selector list, showing additional income. It is used here to show revenue from private companies – when they are owned by major corporations they provide additional revenue, instead of paying income to the corporation's treasury. This makes the corporation's total revenue clearer, without this the player would just see the train revenue. This would also be useful for in 1873 for QLB's bonus income (which _always_ confuses me) and possibly other games too.~~ _[This has been removed]_

### Other things

- ~~There's a new border type, `province`. Other games have used the mountain type for provincial borders, but I thought it would be cleaner to have a dedicated type.~~ [This has been merged, pull request #8620]_
- ~~I've added a `presidency` value for the `from` attribute of an exchange ability. This is for when a share can be exchanged only for a president's certificate. Using `ipo` gives the option of exchanging for a president's certificate or a normal share. There might be a better way to do this.~~ _[This has been removed. There are now checks in the 1858 code to stop the exchange for a normal share.]_
- ~~BuySellParSharesViaBid has been tweaked to allow it to be used for auctioning companies as well as minors.~~ _[This has been merged, pull request #8746]_
- ~~Obsolete trains can be bought across. There's a new constant `ALLOW_OBSOLETE_TRAIN_BUY` (defaults to false) that can be set to allow this.~~ _[This has been merged, pull request #8636]_


assets/app/view/game/alternate_corporations.rb
```diff
@@ -510,6 +510,67 @@ def render_rs_income
           h(:div, { style: { gridColumnStart: 2 } }, [movement_table]),
         ])
       end
+
+      # This is used by 1858 which has private railway companies that act a bit
+      # like minor railways, laying track in operating rounds. This rendering
+      # code is a simplified version of Engine::View::Game::Company.render.
+      def render_private_railway
```
tobymao, line 517: why is this needed? could you just not having this extra UI?

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -214,18 +215,22 @@ def render_short
       # Allow privates or minors to be exchanged for shares if they have the ability
       def render_exchanges
         children = []
-        source_entities = @game.companies + @game.minors
+        # 1858 can have the same entity appear as both a private company and a
+        # minor company. Use uniq to get rid of any potential duplicates.
+        source_entities = (@game.companies + @game.minors).uniq
```
tobymao, line 220: can we simplify this by having a subpar ui?

assets/app/view/game/buy_sell_shares.rb
```diff
@@ -272,6 +277,15 @@ def render_merge
 
         [h(:button, { on: { click: merge } }, 'Merge')]
       end
+
+      def render_convert
```
tobymao, line 281: is there a way to reuse the existing functionality like with abilities or choose?

assets/app/view/game/route_selector.rb
```diff
@@ -167,6 +167,27 @@ def render
           ]
         end
 
+        bonuses = @game.revenue_bonuses(current_entity).map do |bonus|
```
tobymao, line 170: is this necessary?

lib/engine/ability/README.md
```diff
@@ -193,6 +193,16 @@ This company comes with a share of a corporation when acquired.
   randomly selected at game setup.
 - `corporations`: A list of corporations to be used with `"share": "random_share"`
 
+## stubs
```
tobymao, line 196: does this not exist for other games?

lib/engine/game/base.rb
```diff
@@ -871,6 +872,17 @@ def route_trains(entity)
...
       end
 
+      # Revenue bonuses that should be added to the list of train revenues
+      # displayed in the route selector. This method should return an array of
+      # hashes, with each item having these keys:
+      #  - id: The label to show in the `train` column of the route selector.
+      #  - used: Text to show in the `used` column.
+      #  - revenue: The bonus revenue.
+      #  - route: Text to show in the `route` column.
+      def revenue_bonuses(_entity)
```
tobymao, line 882: is this necessary?

lib/engine/game/g_1858/graph.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require_relative '../../graph'
+
+module Engine
+  module Game
+    module G1858
+      class Graph < Engine::Graph
+        # TODO: this method has been cut'n'pasted from Engine::Graph, with a
```
tobymao, line 9: can we do this refactor?

assets/app/view/game/route_selector.rb
ollybh: It's not necessary, but it is very helpful. Here's a screenshot of what this looks like:
route-selector
Without the addition of this bonus revenue, the player would just see the 2M and 2H trains in the route selector list, and a total revenue of Pt125. By adding this extra revenue from the private company (C&B in this case) it makes it clearer what the corporation's revenue will be.
I think that this would be useful in other titles too, for example for the QLB's revenue bonus in 1873.

assets/app/view/game/alternate_corporations.rb
ollybh: This is needed because the private companies in 1858 are operating entities, so get shown in operating rounds and as entities owned by the player. Without this alternate view they will be rendered as if they were minors, but that's a bit of a mess, you get shown their operating order (which never varies) and that the value of their last run (always zero, as they can't own trains). Plus there's an attempt to put in the full name of the entity on the charter, which doesn't fit.

Here's what this looks like with the default rendering:
render-minors

This alternate rendering mimics the default view of a private company, which is what these entities are.
render-privates

lib/engine/ability/README.md
ollybh: No. Other games do use stubs, but those are all included in the map definition. The difference with 1858 is that these stubs are not permanent, and the requirement to lay track linking to the specific edge (or edges) of a hex goes away when a private company closes. So I've implemented these as abilities to allow them to be removed at the appropriate time.

lib/engine/game/base.rb
ollybh: This is the model side of the change you queried in your [comment](https://github.com/tobymao/18xx/pull/8578/files/a441fe31534cee4e17fd4e556a1d3009435a6c66#r1045123208) on the changes in `assets/app/view/game/route_selector.rb`.

The default implementation is to return an empty array. Unless this method is overridden the game behaviour will not change.

assets/app/view/game/buy_sell_shares.rb
ollybh: This was put in as the same entities appear in both the `@game.companies` and the `@game.minors` arrays. These entities also have exchange abilities, and this was leading to the exchange buttons being duplicated.

If this `uniq` call is a problem then it should be possible to work around this by moving the entities between the companies and minors arrays as needed, instead of putting them in both. It will make the 1858 game code a little more complicated, but shouldn't be too difficult.

lib/engine/game/g_1858/graph.rb
ollybh: I've added two commits to do this.

- 2fceee9 extracts two methods from `Engine::Graph.compute`. These are used when `@home_is_token` is true and calculate the home hexes and the home nodes that should be used to start the graphs.
- 26fcea0 overrides `home_hex_nodes` in `G1858::Graph`. This is where the implementation is different from the existing code. Two things are changed, how multiple cities in a hex are handled, and also being able to start a node walk from a section of plain track.

assets/app/view/game/buy_sell_shares.rb
ollybh: This is built on top of the existing functionality for converting corporations, the difference here is that it is a stock round action. As well as the normal selling or buying shares, you can convert a five-share corporation into a ten-share corporation. This UI is to give you a button to do this. Would using abilities make this any simpler?

lib/engine/ability/README.md
michaeljb: I think it would be simpler to just make the stubs part of the map like in other games, and when the company is closed/bought, you can delete the stubs from the tile.

tobymao: can you delete all changes you added for this feature?

lib/engine/part/node.rb
```diff
@@ -57,6 +57,7 @@ def walk(
           node_path.walk(
             visited: visited_paths,
             skip_paths: skip_paths,
+            skip_track: skip_track,
```
michaeljb, line 60: Was this needed for something specific to 1858? If it's a bug fix for general graph walking, let's get this in a separate PR.

lib/engine/ability/README.md
michaeljb: I think it would be simpler and better to just make the stubs part of the map like in other games, and when the company is closed/bought, you can delete the stubs from the tile.

lib/engine/part/stub.rb
```diff
@@ -18,6 +19,10 @@ def stub?
       def track
         :broad
       end
+
+      def inspect
+        "<#{self.class.name} edge=#{@edge}>"
```
michaeljb, line 24: Good idea to add this 👍

lib/engine/game/g_rolling_stock/step/buy_sell_shares_bid_companies.rb
```diff
@@ -148,7 +148,7 @@ def can_buy?(entity, bundle)
             entity.cash >= @game.next_price_to_right(bundle.corporation.share_price).price
           end
 
-          def buy_shares(entity, bundle, exchange: nil, swap: nil, allow_president_change: true, borrow_from: nil)
+          def buy_shares(entity, bundle, exchange: nil, swap: nil, allow_president_change: true, silent: nil, borrow_from: nil)
```
michaeljb, line 151: Since the base `buy_shares` has `silent` default to `nil`, does it need to be added to all these subclass overrides? When they call `super` won't the superclass just use the default by default? It'd be nice to avoid touching other game classes if possible.

lib/engine/hex.rb
```diff
@@ -143,7 +143,7 @@ def lay(tile)
               next unless ability.hex == coordinates
 
               ability.tile = new_city.tile
-              ability.city = new_city.index
+              ability.city = new_city.tile.cities.find_index(new_city)
```
michaeljb, line 146: Why is this necessary?

assets/app/view/game/route_selector.rb
michaeljb: Instead of adding to the train row, could the Submit button just be modified with the grand total from the privates? I'm thinking something along the lines of 1822's mail contract bonus - 
Screenshot 2022-12-12 212739

lib/engine/game/g_1858/company.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+require_relative '../../company'
+require_relative '../../operator'
+
+module Engine
+  module Game
+    module G1858
+      # Private companies in 1858 are a hybrid with elements of other games'
```
michaeljb, line 9: ATM I'm suspicious of trying to treat the 1858 privates as both private companies and minor companies, it seems like it might simplify a few things to make them minors that happen to also pay. I'll tinker with that and hopefully have more firm thoughts later on.

michaeljb (changes requested):
I haven't been able to quite review everything, planning to do more later in the week. I'll also go through more of the 1858 rules for more context since I haven't played it.

lib/engine/game/g_rolling_stock/step/buy_sell_shares_bid_companies.rb
michaeljb: I see I was mistaken with this comment, and the args do need to specified...maybe instead of that, args in the overrides that would default to nil and aren't used in the override can just be sucked in to `**kwargs`?

assets/app/view/game/alternate_corporations.rb
tobymao: i think this PR would be a lot simpler if you just made the companies minors

lib/engine/game/g_1858/company.rb
tobymao: agree, i think this PR should be redone companies as minors instead of this weird hybrid

tobymao:
the proper approach is to make both a company and a minor, don't try to put it all in the company hybrid

1828 as an example

https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1828/step/buy_company.rb

lib/engine/part/node.rb
ollybh: It's a bug fix, though it might not be possible to trigger this bug in other games. In 1858 you're allowed to build broad gauge and narrow gauge track so that they meet at a hex edge, but it's not possible to trace a route or run any trains down a link like this. So you can have something like node, broad gauge track, narrow gauge track, node. As the skip_track setting was not being passed down it was possible for this route to be walked without objecting to the gauge change.

I'll create a pull request that is just for this change.

assets/app/view/game/route_selector.rb
ollybh: Yes, that would be easy enough to do (it's just overriding the `submit_revenue_str` method). I thought that showing the additional revenue sources in the table made it even clearer what has happening, but it's not necessary.

assets/app/view/game/alternate_corporations.rb
ollybh: OK. I'll look into doing that.

lib/engine/part/stub.rb
```diff
@@ -5,10 +5,11 @@
 module Engine
   module Part
     class Stub < Base
-      attr_reader :edge
+      attr_reader :edge, :owner
 
-      def initialize(edge)
+      def initialize(edge, owner: nil)
         @edge = edge
+        @owner = owner
```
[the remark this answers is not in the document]
ollybh: Yes, this was only needed for the Stubs ability.

lib/engine/hex.rb
ollybh: This is a bug fix. new_city.index is the index value from the list of tile parts, not the list of cities. These might match if all the cities are defined at the start of the part list, but I had a tile where the definition was "city, path, city, path, city, part…" This was giving a new_city index of 4, when there were only three cities on the tile. I could have just changed the order of the parts on this tile, but I thought it would be better to fix the bug so that it didn't catch anyone else out.

lib/engine/part/node.rb
ollybh: I've created pull request #8612 for this fix.

lib/engine/hex.rb
ollybh: I've created a separate pull request for this change, #8613.

assets/app/view/game/route_selector.rb
ollybh: This has been done, removing this change.

lib/engine/game/base.rb
ollybh: This has been removed.

lib/engine/part/stub.rb
ollybh: The stubs ability has been removed.

lib/engine/part/stub.rb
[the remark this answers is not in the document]
ollybh: This was removed along with all the other code for the stubs ability.

lib/engine/game/g_rolling_stock/step/buy_sell_shares_bid_companies.rb
ollybh: Your commit 0236a8ff5e meant that these changes were no longer required, so they have been removed.

lib/engine/game/g_1858/company.rb
ollybh: The hybrid company/minor class has been removed.

assets/app/view/game/buy_sell_shares.rb
ollybh: This has been removed, it was no longer needed after the removal of the hybrid company/minor objects.

lib/engine/ability/README.md
ollybh: The stubs ability has been removed.

ollybh:
The hybrid company/minor class has been removed. The private railways are now implemented by having both a company object and a minor object for each one.

The stubs ability has also been removed, replaced by 1858-specific code to create and remove stubs along the routes of the private railway.

And the revenue bonuses shown in the route selector have also gone. Instead of this, the text on the submit button is now adjusted to show additional revenue from private railways.

lib/engine/part/node.rb
ollybh: This pull request has been merged.

lib/engine/hex.rb
ollybh: This pull request has been merged.

PR #8592, opened 2022-12-17
Spreadsheet alternative order - corps first

Galatolol:
An option to have an alternative order of stuff displayed in the spreadsheet.

By default the spreadsheet is exactly the same except for a new button - "Corporations/Players First" - which changes the order to have corporations' holdings before shareholders. A new column ("President") is then displayed as well.


tobymao:
closing this for now

PR #8616, opened 2022-12-31
[18Rhl] Fix CMV bug for RhE (fixes #8608)

### LINKED ISSUE #8608: [18Rhl] Stock value of RhE doesn't rise after the corporation floats
**What's your Game ID?**

https://18xx.games/game/107778?action=50

**How to Reproduce the Bug?**

Perform the private auction. Private auction is part of the first Stock Round. Thus, RhE, which is considered floated after the auction, should have its stock price moved up.

_**At the end of every Stock Round**, check whether 50% of the shares of a corporation have been bought
from the Initial Offering for the first time floated (the GVE-share sold with the Starting Package counts for the
50% as well). If that is the case, the corporation has been. The player who holds the director’s certificate
receives the corporation’s charter and all its station markers. The Bank pays the starting value for all
shares sold into the corporation’s treasury. **Subsequently the stock marker is moved up one space**._

(16.4 Floating a corporation)

**Which Browser / Platform?**

Firefox 103.0.1

perwestling:
Newly floated corporations should get a prise rise (bump) upwards at the end of the stock round it is floated. Now RhE is floated as part of the initial auction, so it did not get this prise rise.

Have now added a special case in the first SR to cover this case. This also broke the reference game so it has been removed. Will add a new reference game later.

Note! This do break existing 18Rhl games and wont be able to fix them so I suggest archiving completed games, and pinning any running game.


perwestling:
I believe this requires completed games to be archived, as the prise rise will affect the money in the game. Not possible to mend the games.

And running games have to be pinned.

Issue #8589, opened 2022-12-15
1861 duplicate tokens during merge


Dave4B:
In 1861 (possibly 1867 - not tested), if two minors with tokens in the same hex (Moscow or Kiev) merge, the player is prompted to remove all but one of the duplicates.
The program then places a token on the major's charter as a zero cost token.
The published rules are unclear as to what should happen in this case, but the designer (Ian D. Wilson) has stated that the second token should cost 20R per hex, the same as forming a major by converting a single minor,

Dave4B:
Just tried this in 1867 with three minors; two in Montreal and one other.
Program required me to remove one of the three tokens; I removed the non-Montreal one.
No response so I removed a Montreal one.
Program continued with one token on the map (Montreal) and a zero price one on the charter.
This is wrong; the non-duplicate token should not be removable.

Dave4B:
Not quite sure whether this should be a separate issue, but the message which appears when you merge exactly two minors in the same hex is "When merging more than 2 minor corporations the new corporation can only keep 2 tokens. Choose which tokens to remove. After merging an additional token will be available on the charter."
This should be a different (new?) message about a corporation not being allowed to have more than one token in a hex/tile.

ollybh:
There are three separate problems here, all related to having more than one token in the same hex:

1. If two tokens are in the same city (not just the same hex) then one gets put back on the minor's charter, and then moved across to the new public company as a zero-cost token.
2. If three or more minors merging it is possible to remove all but one of the tokens.
3. The help message is misleading.

I've got fixes for (1) and (3), I'll submit a pull request  when I've done (2) as well.

Dave4B:
[...]

Issue #8572, opened 2022-12-06
1861: wrong order of minor nationalisation


ianwilson156:
Latest game of 1861
https://18xx.games/game/105949

At the rusting event at the purchase of the first 4T (by K-R), in OR3.1, eight minors were nationalised. The program nationalised them in descending price order from the KR, so OK first then wrapped around. This is wrong - the rule is (section 10, Trainless at the start of Phases 4, 6 and 8)
"...nationalized in descending stock price order".

I've always interpreted this to mean "from the highest to the lowest".

Normally, it doesn't matter, but in this case - when the RSR runs out of tokens - it affects where the RSR ends up with tokens. This may affect where routes get blocked later on.

ollybh:
The minors are being nationalised in the order they are defined in the game file (with yellow minors first, then green minors). This happens for both the nationalisation of the trainless minors at the start of phases 4 and 6, and all remaining minors at the start of phase 8.

The order of major nationalisation is also wrong, this is happening in reverse stock market order.

I've submitted a pull request (#8633) to fix these.

PR #8633, opened 2023-01-05
[1861/1867] Fix nationalisation order

ollybh:
Ian Wilson noted in issue #8572 that minor companies are being nationalised in the wrong order. Where multiple minors are nationalised at the start of phase 4, 6 or 8 this should be done in stock market order. The order can make a difference if the national railway runs out of tokens whilst these minors are being nationalised, it should get the tokens of the first ones to be nationalised.

This pull request fixes this order. It also changes the order of major company nationalisation – this is suppose to happen in stock market order, but it was in reverse stock market order.

There are also tests added for both the minor and major company nationalisation order.

This change has the potential to break a few games as it might change the location of the national railway's tokens, so 1861 and 1867 games will need to be pinned.

I'm submitting this as a draft pull request as there is another open issue (#8589) where the fix will need to pin 1861/1867 games, so it might make sense to push these both through at the same time.

### DIFF
--- a/lib/engine/game/g_1867/game.rb
+++ b/lib/engine/game/g_1867/game.rb
@@ -1627,7 +1627,7 @@ module Engine
             corporation.type != :minor
           end
           @log << 'Minors nationalized' if removed.any?
-          removed.each { |c| nationalize!(c) }
+          removed.sort.each { |c| nationalize!(c) }
           @corporations = corporations
         end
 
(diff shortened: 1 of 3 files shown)


crericha:
Whomever deploys this can use pin cb81a0c50

ollybh:
D'oh, forgot to put 1867 in the pull request subject. @crericha could you (or someone else) please add a label for 1867?

ollybh:
I'm removing the draft status from this one. I'm just added pull request #8689 which fixes issue #8589 and also needs pins for 1861/1867 games.

PR #8636, opened 2023-01-06
Add a flag to allow obsolete trains to be bought between corporations

ollybh:
1858 has wounded trains that run for half revenue and can be discarded during the train buying step. You are allowed to buy these wounded trains across between corporations. That isn't possible with the current implementation of obsolete trains: there's a check in `Engine::Train.buyable` that blocks this for an obsolete train.

This patch adds a flag to the game class, `ALLOW_OBSOLETE_TRAIN_BUY`, that can be set to allow obsolete trains to be bought across. It has a default value of `false` and so does not change any behaviour for other games.

This commit has been extracted from pull request #8578.

### DIFF
--- a/lib/engine/depot.rb
+++ b/lib/engine/depot.rb
@@ -118,7 +118,8 @@ module Engine
 
     def other_trains(corporation)
       all_others = @trains.reject do |train|
-        !train.buyable || [corporation, self, nil].include?(train.owner)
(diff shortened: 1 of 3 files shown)


perwestling:
18MS has this functionality as an optional rule (that is, plus trains rust at end of OR set and these trains cannot be bought between corporations, but there is an optional rule that allows this cross corporation buys).

Do not remember how this was implemented in that game, but it sounds like this change should be used by that title.

crericha (approved):

PR #8637, opened 2023-01-06
Add hook to allow cost of a tile lay to be modified

ollybh:
In 1858 laying a second tile in a turn costs 20, but this cost is halved if one of the tiles adds just narrow gauge track. At the moment there isn't an easy way to modify the cost returned from `Engine::Step::Tracker.get_tile_lay`, without duplicating the entire existing `lay_tile_action` method.

This adds a hook within `lay_tile_action` that can be overridden in a subclass to do this. The default implementation is a no-op.

This is another change extracted from pull request #8578.

### DIFF
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -60,6 +60,7 @@ module Engine
           raise GameError, "#{action.hex.id} cannot be layed as this hex was already layed on this turn"
         end
 
+        tile_lay_cost_override!(tile_lay, action, tile, old_tile)
         extra_cost = tile.color == :yellow ? tile_lay[:cost] : tile_lay[:upgrade_cost]
 
         lay_tile(action, extra_cost: extra_cost, entity: entity, spender: spender)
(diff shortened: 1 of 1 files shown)


lib/engine/step/tracker.rb
```diff
@@ -60,6 +60,7 @@ def lay_tile_action(action, entity: nil, spender: nil)
           raise GameError, "#{action.hex.id} cannot be layed as this hex was already layed on this turn"
         end
 
+        tile_lay_cost_override!(tile_lay, action, tile, old_tile)
```
crericha, line 63: Depending on what you are trying to do, there are already ways to set the tile lay costs. Can you please provide detail on what the game needs to do, so I can steer you in the right direction.

crericha:
This is handled in the game class's `tile_cost_with_discount`, which is called in tracker's `lay_tile`

ollybh:
@crericha Thanks for the advice. I had missed that method. I've redone my code to use that instead.

PR #8641, opened 2023-01-09
[1880] shares, foreign investors, broker fees

bentziaxl:
always market price & no limit on ownership
add float precent change events 
add partial cap during float
add foreign investor share when floating a corp 
foreign leases train
auto withold dividends for foreign investors 
add broker fee on selling shares
remove set SR (in this game SR happen in the middle of an OR when a train type is bought our or removed, will be implemented later)
sold share go to ipo, never to market

Note: Foreign investor owning share doesn't show up in the UI, will need to fix the view files to fix this bug.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -225,6 +225,11 @@ module Engine
       # loans taken during ebuy can lead to receviership
       EBUY_CORP_LOANS_RECEIVERSHIP = false
 
+      # where should sold shares go to?
+      # :bank - bank pool
+      # :corporation - back to corporation/ipo
+      SOLD_SHARES_DESTINATION = :bank
+
       # when is the home token placed? on...
       # par
       # float
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -16,6 +16,9 @@ module Engine
 
         TRACK_RESTRICTION = :permissive
         SELL_BUY_ORDER = :sell_buy
+        SELL_MOVEMENT = :down_per_10
+        MARKET_SHARE_LIMIT = 100
+        EBUY_PRES_SWAP = false
         CURRENCY_FORMAT_STR = '¥%s'
 
         BANK_CASH = 37_860
@@ -23,6 +26,12 @@ module Engine
         CERT_LIMIT = { 3 => 20, 4 => 16, 5 => 14, 6 => 12, 7 => 11 }.freeze
 
         STARTING_CASH = { 3 => 600, 4 => 480, 5 => 400, 6 => 340, 7 => 300 }.freeze
+
+        EVENTS_TEXT = Base::EVENTS_TEXT.merge(
+          'float_30' => ['30% to Float', 'Corporations must have 30% of their shares sold to float'],
+          'float_40' => ['40% to Float', 'Corporations must have 40% of their shares sold to float'],
+          'float_60' => ['60% to Float', 'Corporations must have 60% of their shares sold to float'],
+        ).freeze
 
         STOCKMARKET_COLORS = Base::STOCKMARKET_COLORS.merge(
           pays_bonus: :yellow,
@@ -133,7 +142,14 @@ module Engine
                     rusts_on: '4+4',
                     num: 5,
                   },
-                  { name: '3', distance: 3, price: 180, rusts_on: '6', num: 5 },
+                  {
+                    name: '3',
+                    distance: 3,
+                    price: 180,
+                    rusts_on: '6',
+                    num: 5,
+                    events: [{ 'type' => 'float_30' }],
+                  },
                   {
                     name: '3+3',
                     distance: [{ 'nodes' => %w[city offboard town], 'pay' => 3, 'visit' => 3 },
@@ -142,7 +158,14 @@ module Engine
                     rusts_on: '6E',
                     num: 5,
                   },
-                  { name: '4', distance: 4, price: 300, rusts_on: '8', num: 5 },
+                  {
+                    name: '4',
+                    distance: 4,
+                    price: 300,
+                    rusts_on: '8',
+                    num: 5,
+                    events: [{ 'type' => 'float_40' }],
+                  },
                   {
                     name: '4+4',
                     distance: [{ 'nodes' => %w[city offboard town], 'pay' => 4, 'visit' => 4 },
@@ -151,7 +174,13 @@ module Engine
                     rusts_on: '8E',
                     num: 5,
                   },
-                  { name: '6', distance: 6, price: 600, num: 5 },
+                  {
+                    name: '6',
+                    distance: 6,
+                    price: 600,
+                    num: 5,
+                    events: [{ 'type' => 'float_60' }],
+                  },
                   {
                     name: '6E',
                     distance: [{ 'nodes' => %w[city offboard town], 'pay' => 6, 'visit' => 99 }],
@@ -180,21 +209,16 @@ module Engine
         def next_round!
           @round =
             case @round
-            when Round::Stock
+            when Engine::Round::Stock
               @operating_rounds = @phase.operating_rounds
               reorder_players
               new_operating_round
-            when Round::Operating
-              if @round.round_num < @operating_rounds
+            when Engine::Round::Operating
+              if @round.round_num
                 or_round_finished
                 new_operating_round(@round.round_num + 1)
-              else
-                @turn += 1
-                or_round_finished
-                or_set_finished
-                new_stock_round
               end
-            when Round::Draft
+            when Engine::Round::Draft
               new_stock_round
             when init_round.class
               init_round_finished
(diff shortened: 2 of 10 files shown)


lib/engine/game/g_1880/game.rb
```diff
@@ -203,9 +227,93 @@ def next_round!
             end
         end
 
+        def stock_round
+          G1880::Round::Stock.new(self, [
+            Engine::Step::DiscardTrain,
+            Engine::Step::Exchange,
+            Engine::Step::SpecialTrack,
```
perwestling, line 234: Is there track laid in stock round?

lib/engine/game/g_1880/game.rb
```diff
@@ -203,9 +227,93 @@ def next_round!
...
+            c.shares.last(5).each { |s| s.buyable = false }
+          end
+        end
+
         def p1
           company_by_id('P1')
         end
+
+        def event_float_30!
+          @log << "-- Event: #{EVENTS_TEXT['float_30'][1]} --"
```
perwestling, line 271: A minor comment is that it is possible to combine the code of these three events in one method, and then call the method from each event with suitable parameters, like 
def do_event_float(event_text_name, float_percent)
   ...
end

lib/engine/game/g_1880/step/dividend.rb
```diff
@@ -0,0 +1,30 @@
...
+require_relative '../../../step/dividend'
+require_relative '../../../step/minor_half_pay'
+
+module Engine
+  module Game
+    module G1880
+      module Step
+        class Dividend < Engine::Step::Dividend
+          def actions(entity)
+            return [] if entity.minor?
```
perwestling, line 12: You have both ways of doing minor?

lib/engine/game/g_1880/game.rb
bentziaxl: There will be other events for those trains, which I think means i'll need to play around with the phase.rb code here:
https://github.com/tobymao/18xx/blob/526032884ca315ba29ac875e0edecc7212e3c407/lib/engine/phase.rb#L26

and get it send event_{event_name}! for all other events, and do_event_float for the float events.
That would be pretty messy. Or am i missing something?

lib/engine/game/g_1880/step/dividend.rb
bentziaxl: not sure what you're asking. what do you mean both ways?

perwestling:
Just put in some minor comments. Otherwise it looks good.

A general comment is that I prefer these commits to be squashed into one, unless each commit is about a specific part. So in this case there could be one commit with FI code and one commit with.... rest?  Rubocop, cleanup, etc, could be incorporated.
But yeah, it might be a personal preference.

lib/engine/game/g_1880/game.rb
bentziaxl: no, i'll clean it up

lib/engine/game/g_1880/step/dividend.rb
perwestling: The next line where using corporation, with type minor. So are both ways of doing minors used?

bentziaxl: fixed

lib/engine/game/g_1880/game.rb
bentziaxl: fixed

lib/engine/game/g_1880/game.rb
perwestling: I just saw it as a refactoring opportunity as the code was more or less triplicated. But I did not check in details, and I do not know the final result, so it might be something to look for near the end. 
In this case it is also three simple methods so benefit would be low. The DRY principle is more to avoid code duplication to avoid errors caused by changes not applied in all instances, and also about readability. 

I can recommend "Clean code" books/article if you have not read them. (Not that I am saying that your code is not "clean" just as a general recommendation after doing a lot of code review professionally, and worked a lot with legacy code.)

bentziaxl: Agreed, I'm familiar with clean code and DRY. I just didn't see a way to do it here without introducing a lot more code, and if statements. I was asking if you saw a way to do so. Closing the comment.

perwestling (approved):
Looks OK from my point of view.

lib/engine/game/g_1880/game.rb
```diff
@@ -24,6 +27,12 @@ class Game < Game::Base
 
         STARTING_CASH = { 3 => 600, 4 => 480, 5 => 400, 6 => 340, 7 => 300 }.freeze
 
+        EVENTS_TEXT = Base::EVENTS_TEXT.merge(
+          'float_30' => ['30% to Float', 'Corporations must have 30% of their shares sold to float'],
+          'float_40' => ['40% to Float', 'Corporations must have 40% of their shares sold to float'],
+          'float_60' => ['60% to Float', 'Corporations must have 60% of their shares sold to float'],
+        ).freeze
```
crericha, line 34: Oops, we dupped on this one. I'll merge the two.

lib/engine/game/g_1880/share_pool.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative '../../share_pool'
+
+module Engine
+  module Game
+    module G1880
+      class SharePool < Engine::SharePool
+        def sell_shares(bundle, allow_president_change: true, swap: nil, silent: nil)
```
crericha, line 9: Don't duplicate this method. Add a hook to choose where the share goes. Maybe a new game constant `SOLD_SHARES_DESTINATION = :bank ` and for 80 `:corporation`

lib/engine/game/g_1880/step/dividend.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/dividend'
+require_relative '../../../step/minor_half_pay'
+
+module Engine
+  module Game
+    module G1880
+      module Step
+        class Dividend < Engine::Step::Dividend
```
crericha, line 10: This class is fine; however, there is already a `minor_half_pay` module. A `minor_withhold` module would keep to that spirit and allow reuse. Up to you if you want to go that route.

FYI the `minor_half_pay` module has more code. Take a look and see if you need any of those other methods.

lib/engine/game/g_1880/game.rb
```diff
@@ -203,9 +227,92 @@ def next_round!
...
+
+        def float_corporation(corporation)
+          @log << "#{corporation.name} floats"
+
+          @bank.spend(corporation.par_price.price * 5, corporation)
+          @log << "#{corporation.name} receives #{format_currency(corporation.cash)}"
+
+          # reserve share for foreign investor
+          foreign_investor = @minors.find { |m| m.owner == corporation.owner }
+          return if foreign_investor.shares.any?
```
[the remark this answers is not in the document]
bentziaxl: done

lib/engine/game/g_1880/step/dividend.rb
bentziaxl: added minor_withold

lib/engine/game/g_1880/game.rb
```diff
@@ -16,14 +16,24 @@ class Game < Game::Base
 
         TRACK_RESTRICTION = :permissive
         SELL_BUY_ORDER = :sell_buy
+        SELL_MOVEMENT = :down_per_10
+        MARKET_SHARE_LIMIT = 100
```
crericha, line 20: Is MARKET_SHARE_LIMITED needed if shares go back to the corporation?

lib/engine/game/base.rb
```diff
@@ -225,6 +225,8 @@ class Base
       # loans taken during ebuy can lead to receviership
       EBUY_CORP_LOANS_RECEIVERSHIP = false
 
+      SOLD_SHARES_DESTINATION = :bank
```
crericha, line 228: Add a description and what choices are valid

lib/engine/game/g_1880/share_pool.rb
```diff
@@ -0,0 +1,23 @@
...
+    module G1880
+      class SharePool < Engine::SharePool
+        def log_sell_shares(entity, verb, bundle, price, swap_text)
+          fee = additional_price_adjustments(bundle)
+          @log << "#{entity.name} #{verb} #{num_presentation(bundle)} " \
+                  "of #{bundle.corporation.name} and receives #{@game.format_currency(price)}" \
+                  " (broker fee: #{@game.format_currency(fee)})#{swap_text}"
+        end
+
+        def additional_price_adjustments(bundle)
```
crericha, line 16: Maybe call this method broker_fee?

lib/engine/share_pool.rb
```diff
@@ -145,6 +145,15 @@ def sell_shares(bundle, allow_president_change: true, swap: nil, silent: nil)
                       swap_to_entity: swap_to_entity)
     end
 
+    def log_sell_shares(entity, verb, bundle, price, swap_text)
+      @log << "#{entity.name} #{verb} #{num_presentation(bundle)} " \
+              "of #{bundle.corporation.name} and receives #{@game.format_currency(price)}#{swap_text}"
+    end
+
+    def additional_price_adjustments(_bundle)
```
crericha, line 153: Why does the base class need this?

lib/engine/game/g_1880/game.rb
bentziaxl: not needed. will clean

lib/engine/game/g_1880/share_pool.rb
bentziaxl: i'm overwriting a hook in the base code that is used in sell_shares, if i rename it broker_fee here, i'll need to do it there, additional price adjustment is more generic

lib/engine/share_pool.rb
bentziaxl: I originally duplicated the sell_shares method, as i needed to do 3 things:
send shares to the corporation
reduce broker fee,
add it to the logs.

addressing your comment, I added the 3 hooks
1- SOLD_SHARES_DESTINATION
2- additional_price_adjustments for the broker fee (keeping it generic name wise, not every game that might use this will want it to be called broker fee)
3- pulled the log into a separate method

lib/engine/game/base.rb
bentziaxl: done

lib/engine/share_pool.rb
crericha: Okay, I missed that it was used in the base class.

crericha (approved):

PR #8653, opened 2023-01-12
[1880] SR in the middle of and OR, rocket of china, par order, export all trains

bentziaxl:
SR in the middle of and OR
rocket of china
par order
export all trains

### DIFF
--- a/lib/engine/ability/README.md
+++ b/lib/engine/ability/README.md
@@ -154,8 +154,8 @@ This company may not be bought in.
 
 ## purchase_train
 
-Immediately purchases (for normal amount) the currently available depot train
-for the owning corporation.
+Immediately purchases the currently available depot train for the owning corporation.
+- `free`: If true, the train cost is free, otherwase at cost. Default false.
 
 ## reservation
 
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -15,7 +15,7 @@ module Engine
         include Entities
 
         attr_accessor :train_marker
-        attr_reader :full_cap_event
+        attr_reader :full_cap_event, :par_order
 
         TRACK_RESTRICTION = :permissive
         SELL_BUY_ORDER = :sell_buy
@@ -304,6 +304,7 @@ module Engine
         def setup
           @full_cap_event = false
           @foreign_investors_can_operate = true
+          @par_order = { 100 => [], 90 => [], 80 => [], 70 => [] }
           setup_building_permits
           setup_unsaleable_shares
         end
@@ -326,6 +327,7 @@ module Engine
 
         def new_auction_round
           Engine::Round::Auction.new(self, [
+            G1880::Step::CompanyPendingPar,
             G1880::Step::SelectionAuction,
           ])
         end
@@ -338,9 +340,14 @@ module Engine
           @round =
             case @round
             when Engine::Round::Stock
-              @operating_rounds = @phase.operating_rounds
-              reorder_players
-              new_operating_round
+              if @saved_or_round
+                @log << "--Return to Operating Round--"
+                @saved_or_round
+              else
+                @operating_rounds = @phase.operating_rounds
+                reorder_players
+                new_operating_round
+              end
             when Engine::Round::Operating
               if @round.round_num
                 or_round_finished
@@ -357,7 +364,6 @@ module Engine
 
         def stock_round
           G1880::Round::Stock.new(self, [
-            Engine::Step::DiscardTrain,
             Engine::Step::Exchange,
             G1880::Step::BuySellParShares,
           ])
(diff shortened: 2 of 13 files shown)


lib/engine/ability/README.md
```diff
@@ -154,8 +154,8 @@ This company may not be bought in.
 
 ## purchase_train
 
-Immediately purchases (for normal amount) the currently available depot train
-for the owning corporation.
+Immediately purchases the currently available depot train for the owning corporation.
+- `free`: If true, the train cost is free, otherwase at cost. Default false.
```
bentziaxl, line 159: I used purchase_train ability for rocket of china, as it's the closest ability i found. 
I added the free param, but haven't figure out how to use it in view to change the wording there (It still says buy X train for Y cost). 
If it makes more sense i can add a new ability (exchange_train or similiar) instead of using this one

lib/engine/game/g_1880/step/company_pending_par.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Game
+    module G1880
+      module Step
+        class CompanyPendingPar < Engine::Step::CompanyPendingPar
+          def process_par(action)
+            super
+            corporation = action.corporation
+            @log << "#{corporation.name} selects ABC building permit"
+            corporation.building_permits = 'ABC'
+          end
```
bentziaxl, line 15: Ugly hard coding, kept it for now as 99% of the time when launching BCR the player will pick ABC permit (vs BCD). 
I will fix it later to be a choose action instead

lib/engine/ability/README.md
crericha: This is fine if you update the view accordingly.

lib/engine/game/g_1880/game.rb
```diff
@@ -304,6 +304,7 @@ def event_token_cost_doubled!
         def setup
           @full_cap_event = false
           @foreign_investors_can_operate = true
+          @par_order = { 100 => [], 90 => [], 80 => [], 70 => [] }
```
crericha, line 307: I have a feeling that you'll have to populate this with nils, so someone can pick a specific spot in the order.

lib/engine/game/g_1880/game.rb
```diff
@@ -338,9 +340,14 @@ def next_round!
           @round =
             case @round
             when Engine::Round::Stock
-              @operating_rounds = @phase.operating_rounds
-              reorder_players
-              new_operating_round
+              if @saved_or_round
+                @log << '--Return to Operating Round--'
```
crericha, line 344: I think its a new OR, not a return. The new OR just happens to begin where the last left off.

lib/engine/game/g_1880/game.rb
```diff
@@ -393,9 +400,12 @@ def bcr
         end
 
         def status_array(corporation)
-          return if corporation.minor? || !corporation.floated?
+          return if corporation.minor? || !corporation.original_par_price
 
-          ["Building Permits: #{corporation.building_permits}"]
+          status = ["Building Permits: #{corporation.building_permits}"]
+          par_location = @par_order[corporation.original_par_price.price].find_index(corporation) + 1
+          status << ["Par Price: #{corporation.original_par_price.price}-#{par_location}"]
```
crericha, line 407: Hopefully we can do away with all of this and show visually. Okay for now.

lib/engine/game/g_1880/game.rb
crericha: I was expecting 1880 would detect the train event, set some state in the game, and override operating.finished? to true.

lib/engine/game/g_1880/step/company_pending_par.rb
```diff
@@ -0,0 +1,26 @@
...
+
+          def auto_actions(entity)
+            share = @game.abilities(companies_pending_par.first, :shares).shares.first
+            share_price = @game.stock_market.par_prices.find { |pp| pp.price == 100 }
+            [Engine::Action::Par.new(entity, corporation: share.corporation, share_price: share_price)]
```
crericha, line 20: They have to pick a spot within the 100's.

lib/engine/game/g_1880/game.rb
crericha: The player is allowed to pick any of the 4 spots within each par value

lib/engine/game/g_1880/step/simple_draft.rb
```diff
@@ -39,10 +39,16 @@ def process_bid(action)
 
             @log << "#{player.name} chooses #{minor.full_name} (#{minor.name})"
 
+            assign_bcr_share_to_fi(action.entity, minor) if player.shares.find { |s| s.corporation.id == 'BCR' }
```
crericha, line 42: Why do you need to do this? The shares ability should trigger float_corporation being called.

bentziaxl: float_corporation happens as soon as a player wins the private. At that point they do not own a FI so no share is sent to a FI. Once the player with the BCR picks a FI, then the share is assigned there. 
This code does this during the draft, as it should per the rules

PR #8659, opened 2023-01-14
18WE: Initial commit with map and tileset

andypymont:
Began the implementation of [18WE: Western Europe](https://boardgamegeek.com/boardgame/317048) by Karl Ernst.
- Map implemented
- Tileset implemented
- Other rules currently inherited from the game's predecessor, 1862

### DIFF
--- a/lib/engine/game/g_18_we/map.rb
+++ b/lib/engine/game/g_18_we/map.rb
@@ -1,0 +1,301 @@
+# frozen_string_literal: true
+
+require_relative 'meta'
+require_relative '../base'
+
+module Engine
+  module Game
+    module G18WE
+      module Map
+        TILES = {
+          'WE01' => {
+            'count' => 3,
+            'color' => 'yellow',
+            'code' => 'label=Z;city=revenue:40,slots:2;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0;path=a:4,b:_0',
+          },
+          'WE02' => {
+            'count' => 'unlimited',
+            'color' => 'yellow',
+            'code' => 'label=Y;city=revenue:30,slots:2;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0;path=a:4,b:_0',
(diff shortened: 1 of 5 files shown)


lib/engine/game/g_18_we/map.rb
```diff
@@ -0,0 +1,301 @@
...
+        }.freeze
+
+        LOCATION_NAMES = {
+          'K0' => 'Scandinavia',
+          'J1' => 'Skaggerak',
+          'K2' => 'København',
+          'L3' => 'Denmark Sound',
+          'N3' => 'Baltic Sea',
+          'P5' => 'St. Petersburg',
+          'O4' => 'Konigsburg',
```
ollybh, line 240: Königsberg.

lib/engine/game/g_18_we/map.rb
```diff
@@ -0,0 +1,301 @@
...
+          'E10' => 'Rouen',
+          'I10' => 'Frankfurt',
+          'M10' => 'Breslau',
+          'O10' => 'Krakow',
+          'B11' => 'Brest',
+          'F11' => 'Paris',
+          'H11' => 'Köln',
+          'L11' => 'Praha',
+          'C12' => 'Nantes',
+          'E12' => 'Orleans',
```
ollybh, line 265: Orléans. It doesn't have the acute accent on the 18WE images on BGG, but generally local spellings are being used, so this would be more consistent.

andypymont: As you say, this commit currently reflects the existing board for the game. I have asked Karl as designer for his input on changing things. As well as the two you have highlighted, there are 11 other locations which could be more "localised" - the most striking being Constantinople (which could be rendered as Konstantiniyye or Istanbul).

michaeljb: This is something that's always easy to change later, so I don't think we need to hold up pre-alpha progress for it.

Issue #8673, opened 2023-01-18
1822+ Game locked up and will not load


DFritsch56:
**What's your Game ID?**
https://18xx.games/game/105899

Locked up while laying track.  Tried to click check mark to confirm track selection but game never accepted it. None of us can get game to load. Please help! Restart turn if possible.

David Fritsch

PR #8675, opened 2023-01-19
Fix city tokens and reservations getting out of sync

outsidePasser:
Fixes: #8673


lib/engine/part/city.rb
```diff
@@ -171,7 +171,15 @@ def reset!
 
       def delete_token!(token, remove_slot: false)
         if remove_slot
-          @tokens.delete(token)
+          # This can make the reservations out of sync.  Delete the reservation in the same
+          # position, but add a nil to the end so it still has the correct number
+          position = @tokens.find_index(token)
```
tobymao, line 176: index

michaeljb (approved):

PR #8704, opened 2023-01-24
[18MT] Correct B2/B4 border to impassable

### LINKED ISSUE #8459: 18MT Blue hexside border
There is a blue line between B2 and B4 on the base map.  This appears to have no effect on game play.  Either it should be removed or urge effect added in.  
My assumption is it supposed to be an impassable hexside as there is a large lake in reality at that part of the map.
image

philcampeau:
Fixes #8459

Will require existing games to be pinned.


tobymao (approved):

philcampeau:
Note that my update will change that border from blue to red, but I think the function is more important than the color.

Issue #8142, opened 2022-08-06
Profile page not working


l33twash0r:
I tried to look up on games I have played in the past, but when I click on the Profile page I just get an error:

{"error":"User does not exist"}

tobymao:
what's your user name?

l33twash0r:
https://18xx.games/profile/l33twash0r

Forgot to add that to the issue.

crericha:
Your username has a space at the end. I removed it and your profile page loads now.

tobymao:
@crericha we should add strip() to name creation / editting

crericha:
Yes, my thoughts exactly.

PR #8707, opened 2023-01-25
strip param values in User#update_settings

SylvanSign:
not totally sure if this is the right fix, but we encountered a bug where the custom webhook url wasn't working because somebody accidentally put a trailing space in the Webhook URL input. I also saw a similar, older issue that this should solve https://github.com/tobymao/18xx/issues/8142

### DIFF
--- a/models/user.rb
+++ b/models/user.rb
@@ -23,10 +23,10 @@ class User < Base
   ]).freeze
 
   def update_settings(params)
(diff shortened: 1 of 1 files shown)


tobymao (approved):

tobymao:
reverting this because it broke stuff #8717 

@SylvanSign please test your changes next time

PR #8723, opened 2023-01-25
[1880] fix issues

bentziaxl:
add stub module

 minor typos (boought vs bought, log showing owner of woosong when getting payment, SR 5 vs SR5.1)

 fix offboard not being terminal, and forcing connection if FI and corp are connected via offboard

 FI token removal removed all tokens in the city, instead of just FI's token

### DIFF
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -301,7 +301,7 @@ module Engine
           return if corporation.ipo_shares.size > 5 || corporation.fully_funded
 
           amount = corporation.original_par_price.price * 5
-          @log << "Five shares of #{corporation.name} have been boought. "\
+          @log << "Five shares of #{corporation.name} have been bought. "\
                   "#{corporation.name} receives #{format_currency(amount)}"
           @bank.spend(amount, corporation)
           corporation.fully_funded = true
@@ -713,7 +713,7 @@ module Engine
         def round_description(name, round_number = nil)
           round_number ||= @round.round_num
           description = super
-          description += ".#{round_number}" if @round.is_a?(Engine::Round::Operating)
+          description += ".#{round_number}" if name == self.class::OPERATING_ROUND_NAME
           description += " - Train Marker at #{@train_marker.name}" if @train_marker && !@end_game_triggered
           description
         end
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -13,6 +13,7 @@ module Engine
         include_meta(G1880::Meta)
         include Map
         include Entities
+        include StubsAreRestricted
 
         attr_accessor :train_marker
         attr_reader :full_cap_event, :communism, :end_game_triggered, :saved_or_round, :final_operating_rounds
--- a/lib/engine/game/g_1880/step/check_fi_connection.rb
+++ b/lib/engine/game/g_1880/step/check_fi_connection.rb
@@ -96,7 +96,7 @@ module Engine
           end
 
           def discard_token(fi)
-            fi.tokens.first.city.remove_tokens!
+            fi.tokens.first.remove!
             @game.log << "#{fi.name} token is discarded"
           end
 
--- a/lib/engine/game/g_1880/step/selection_auction.rb
+++ b/lib/engine/game/g_1880/step/selection_auction.rb
@@ -37,6 +37,8 @@ module Engine
                 # It's now free so the next player is forced to take it.
                 @round.next_entity_index!
                 forced_win(current_entity, first)
+                @all_passed_win = true
+                post_win_bid(current_entity, first)
               end
             else
               @game.payout_companies
--- a/lib/engine/game/g_1880/step/selection_auction.rb
+++ b/lib/engine/game/g_1880/step/selection_auction.rb
@@ -37,8 +37,6 @@ module Engine
                 # It's now free so the next player is forced to take it.
                 @round.next_entity_index!
                 forced_win(current_entity, first)
-                @all_passed_win = true
-                post_win_bid(current_entity, first)
               end
             else
               @game.payout_companies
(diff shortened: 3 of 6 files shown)


lib/engine/game/g_1880/stock_market.rb
```diff
@@ -31,6 +31,14 @@ def move_left(corporation)
 
           super
         end
+
+        def sort_market_legend?
+          @sort_market_legend ||= true
```
tobymao, line 36: what's the point of @sort_market_legend ||= true ? same with the false version

assets/app/view/game/stock_market.rb
```diff
@@ -338,9 +338,11 @@ def render
           type_text = @game.class::MARKET_TEXT
           colors = @game.class::STOCKMARKET_COLORS
 
-          types_in_market = @game.stock_market.market.flatten.compact.flat_map(&:types)
-          .uniq.to_h { |p| [p, colors[p]] }
+          types_in_market_map = @game.stock_market.market.flatten.compact.flat_map(&:types).uniq
+          types_in_market_map.sort! if @game.stock_market.sort_market_legend?
+          types_in_market_map.reverse! if @game.stock_market.reverse_market_legend?
```
crericha, line 343: Seems pretty complicated and confusing to use. I recommend using the order as defined in @game::class.MARKET_TEXT. It's easy for a game to change that order if they want to.

lib/engine/game/g_1880/step/selection_auction.rb
```diff
@@ -48,10 +49,21 @@ def all_passed!
...
-            next_entity!
-            @auction_triggerer = current_entity
+
+            if all_passed
+              next_idx = (@round.entity_index + 1) % entities.size
+              @auction_triggerer = entities[next_idx]
+            else
+              next_entity!
+              @auction_triggerer = current_entity
+            end
```
crericha, line 65: What's going on here? I thought all'd you have to do is set @auction_triggerer to to current_entity and then call post_win_bid. Do you really need all this complexity?

bentziaxl: if player_1 auto wins this by all_passing, then the next auction will be started by player_3, and not player_2. 

next_entity is called twice, once in post_win_bid, and once in in process_pass https://github.com/tobymao/18xx/blob/1dd6528a098872c003943bd4dd6506e26c99f482/lib/engine/step/selection_auction.rb#L47 

I could add a round/step param all_passed_win, to avoid the next! in process_pass, would that be cleaner?

crericha: The code you have to work with isn't great. Nothing obvious comes to mind; however, compensating for how process_pass is implemented in your post_win implementation is very brittle code.

assets/app/view/game/stock_market.rb
bentziaxl: done, however this will impact games where the MARKET_TEXT is different than the order of prices as appearing the market (46 for example), the legend will be in the wrong order, but this is easily fixed.

lib/engine/game/g_1880/stock_market.rb
bentziaxl: not needed anymore

lib/engine/game/g_1880/step/selection_auction.rb
bentziaxl: added all_passed_win, is this better?

assets/app/view/game/stock_market.rb
bentziaxl: should i add a flag to use this? or just apply everywhere?

assets/app/view/game/stock_market.rb
```diff
@@ -338,9 +338,10 @@ def render
           type_text = @game.class::MARKET_TEXT
           colors = @game.class::STOCKMARKET_COLORS
 
-          types_in_market = @game.stock_market.market.flatten.compact.flat_map(&:types)
-          .uniq.to_h { |p| [p, colors[p]] }
+          types_in_market_map = @game.stock_market.market.flatten.compact.flat_map(&:types).uniq
+          types_in_market_map.sort_by! { |e| @game.class::MARKET_TEXT.keys.index(e) }
```
bentziaxl, line 343: it also forces games to have a market_text entry for every price type otherwise it will error
should i add || Float::INFINITY to cover for those cases?

bentziaxl: done, however this will impact games where the MARKET_TEXT is different than the order of prices as appearing the market (46 for example), the legend will be in the wrong order, but this is easily fixed. 
Should I add a flag, or apply it everywhere?

crericha:
@bentziaxl please close this PR and open new PRs for the parts that are still relevant

bentziaxl:
> @bentziaxl please close this PR and open new PRs for the parts that are still relevant

I did, i remerged with base, and took out the auction fix, we'll do that later

lib/engine/game/g_1880/step/check_fi_connection.rb
```diff
@@ -96,7 +96,7 @@ def process_token(action)
           end
 
           def discard_token(fi)
-            fi.tokens.first.city.remove_tokens!
+            fi.tokens.first.remove!
```
crericha, line 99: There is more to this one (corp needed to be closed using close_corporation). I've addressed it. Please remove this change from the PR.

lib/engine/game/g_1880/game.rb
```diff
@@ -713,7 +714,7 @@ def lessee
         def round_description(name, round_number = nil)
           round_number ||= @round.round_num
           description = super
-          description += ".#{round_number}" if @round.is_a?(Engine::Round::Operating)
+          description += ".#{round_number}" if name == self.class::OPERATING_ROUND_NAME
```
crericha, line 717: What was wrong with the original way?

PR #8725, opened 2023-01-26
Change 1836jr56 to semi-restrictive track

### LINKED ISSUE #8724: 1836jr56 track should be semi-restrictive
There is no note in the 36jr56 rules saying that track should be permissive.  As per the 1836jr rules, all rules are the same as the main game except where noted, and 1856 uses semi-restrictive track.

philcampeau:
Fixes #8724


tobymao:
you need to fix the build

PR #8727, opened 2023-01-26
[1868WY] P7 "Big Boy" private

michaeljb:
this private provides a `+1 +1` token which boosts a train's city and town values by one

* represent the token as `[+1 +1]`, and when assigned to a train, represent the new number in square brackets to distinguish it; e.g., a `2` train becomes `[3+1]`
* wrap choice buttons in a div so that in the train-buying step, all choices are rendered on the same line
  * before, with first choice on one line and the rest on another line: 
  * after, all on one line: 
* the token can first be assigned when the private is bought (`Choose` step), or during the `BuyTrain` step; `ChooseBigBoy` is implemented as a module so to easily provide the functionality to both possible steps
* the token can be re-assigned to a different train during the `BuyTrain` step

[#5011]

### DIFF
--- a/assets/app/view/game/choose.rb
+++ b/assets/app/view/game/choose.rb
@@ -45,7 +45,7 @@ module View
         div_class = choice_buttons.size < 5 ? '.inline' : ''
         h(:div, [
           h("div#{div_class}", { style: { marginTop: '0.5rem' } }, "#{@game.round.active_step.choice_name}: "),
-          *choice_buttons,
+          h(:div, choice_buttons),
         ])
       end
 
--- a/assets/app/view/game/round/operating.rb
+++ b/assets/app/view/game/round/operating.rb
@@ -45,6 +45,7 @@ module View
           left << h(SwitchTrains) if @current_actions.include?('switch_trains')
           left << h(ReassignTrains) if @current_actions.include?('reassign_trains')
           left << h(DoubleHeadTrains) if @current_actions.include?('double_head_trains')
+          left << h(Choose) if @current_actions.include?('choose')
 
           if @current_actions.include?('buy_train')
             left << h(IssueShares) if @current_actions.include?('sell_shares') || @current_actions.include?('buy_shares')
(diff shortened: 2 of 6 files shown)


tobymao (approved):

crericha:
A mechanism for attaching to a train already exists (1822 family, 18USA, etc.). Did you have a look at how they did it @michaeljb? If so, did it not suit your needs?

michaeljb:
@crericha yeah, in this game the token can only be attached to a train when the private is bought, or during the train-buying step, not during the route step. I did take inspiration from the [`attach_pullman` method](https://github.com/tobymao/18xx/blob/37b1f0a08e4e040891eff08f5ab46dbf6fc5be8f/lib/engine/game/g_1822/step/route.rb#L20-L36), but with the different timing and the fact that the token gets removed when the private closes there were some different things to handle.

Some of the logic is duplicated a little, do you think it's worth trying to make a common module for some of the attach/detach logic?

crericha:
I think it would be good to make a view component that they can share. The step logic is probably okay separate, unless there is low hanging fruit there; however, I'm mostly interested in having a common UI.

I like the bracket notion. Good idea.

michaeljb:
@crericha thanks for the comment, I finally had some time to think about it more and review the rules and so forth.

So right now both 1822 and 1868 Wyoming have the train attachment implemented with the `Choose` action and view, so they inherently share some UI elements. They have to be implemented in different steps because of the games' different rules, and I don't think it would make that much sense to make a completely standalone "train attachment" step and view.

If you're ok with the bit of logic being duplicated, I don't see any other gains that we can get from generalizing anything. Maybe the `choice_name` string, but even then they work differently enough that I don't want to change the +1+1 token's string--instead of being chosen at route runtime like the '22 pullman, it must be set in the train buying phase, which means you run the risk of attaching it to a train that will rust before the next time you run routes, and be stuck running without it actually attached to a train.

1822 pullman, during routes:

Screenshot 2023-02-03 164636

1868 Wyoming +1+1 token, during train-buying:

Screenshot 2023-02-03 164804

PR #8749, opened 2023-01-28
[1858] Operating round structure and track laying

ollybh:
This adds the overall structure of the operating round for 1858, and the code for laying track. 1858 has several differences from other games which have had to be implemented here.

### Broad gauge and narrow gauge routes

Routes for laying track (and placing tokens and running trains) start at a token (or a home hex for a minor) and are traced on either broad/dual gauge track or narrow/dual gauge track. Routes cannot include both narrow and broad gauge track. This is implemented by having two graph objects created, with `skip_track: :narrow` and `skip_track: :broad`. Various methods in the track step check both of these graphs to see if a tile placement is valid on either of them.

### Discounts for narrow gauge track

Terrain costs are halved for yellow narrow gauge tiles. Also the Pt20 cost of a second tile is halved if either the first or the second tile is only adding narrow gauge track. `gauges_added` is included in the round state to record the type of tracks that are added by each tile placement so this can be tracked.

### Private railway home hexes and routes

Private railway companies are implemented using linked company and minor entities. The minors lay track in an operating round and skip all other steps. When laying track they can only place yellow tiles and must be able to trace a route back to on of their home hexes. A subclass of Graph is used to trace routes from home hexes, the default implementation of this class does not work when there is plain track in a home hex that is not connected to a node. A PathNode class is created for this, it pretends that there is a node in these plain track segments so that `Node.walk` can start at these locations.

### Stubs in private railway home hexes

Stubs are used to force yellow tiles to be laid along the routes of the private railways. These are created at the start of the game by the `setup_stubs` method. A `clear_stubs` method is included in this pull request, though nothing here invokes that method. It will be called when a private is acquired by a corporation or it closes.

### Two abilities blocking the same hex

Privates have `blocks_hex` abilities to stop other people from laying track in their home hexes. A few hexes on the map are the home hexes for two private companies, in this case either minor (or a corporation that ones one of the privates) can lay track in the hex. The code in `ability_blocking_hex` allows this to happen.

### President can pay for track

If a corporation or minor cannot afford the cost of a tile lay this will be paid by the president.

### DIFF
--- a/lib/engine/game/g_1858/game.rb
+++ b/lib/engine/game/g_1858/game.rb
@@ -19,6 +19,8 @@ module Engine
         include G1858::Trains
         include StubsAreRestricted
 
+        attr_reader :graph_broad, :graph_metre
+
         GAME_END_CHECK = { bank: :current_or }.freeze
         BANKRUPTCY_ALLOWED = false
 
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1858/step/track.rb
```diff
@@ -0,0 +1,183 @@
...
+            message += "lays tile ##{tile.name}"
+            message += " with rotation #{rotation} on #{hex.name}"
+            message += " (#{tile.location_name})" unless tile.location_name.to_s.empty?
+            @log << message
+          end
+
+          def hex_neighbors(entity, hex)
+            hexes_broad = @game.graph_broad.connected_hexes(entity)[hex]
+            hexes_metre = @game.graph_metre.connected_hexes(entity)[hex]
+            hexes = [hexes_broad, hexes_metre].compact.inject([], :|)
```
michaeljb, line 113: Does this work?
```suggestion
            hexes = [*hexes_broad, *hexes_metre].uniq
```

lib/engine/game/g_1858/step/track.rb
```diff
@@ -0,0 +1,183 @@
...
+            # Need to check twice whether this tile is OK, once using routes along
+            # broad/dual gauge track, and once allow metre/dual gauge.
+            if !valid_tile_lay?(entity, old_tile, new_tile, @game.graph_broad) &&
+                !valid_tile_lay?(entity, old_tile, new_tile, @game.graph_metre)
+              # # The check for private railways home hexes is needed in case a private
+              # # builds plain track that's not connected to a revenue centre, it will
+              # # not be classed as a connected path.
+              # XXX Is this still needed after the graphing changes? XXX
+              # && !(entity.minor? && entity.home_hex?(new_tile.hex))
+              raise GameError, 'Must use new track or change city value'
```
michaeljb, line 134: Should this error message say something about being reachable by one of the track gauges?

Aznd4u69:
Stop emailing me

lib/engine/game/g_1858/game.rb
```diff
@@ -46,6 +48,145 @@ class Game < Game::Base
...
+        end
+
+        def upgrade_cost(tile, hex, entity, _spender)
+          return 0 if tile.upgrades.empty?
+          return 0 if entity.minor? && home_hex?(entity, hex)
+
+          cost = tile.upgrades[0].cost
+          return cost unless metre_gauge_upgrade(tile, hex.tile)
+
+          discount = cost / 2
```
michaeljb, line 161: I think you need to be explicit about rounding here, IIRC this is something where ruby and javascript (opal) can produce different results

michaeljb:
> Stop emailing me

Aznd4u69 I don't know how you got added as a participant on this issue but you can unsubscribe on the right side of the [web page here](https://github.com/tobymao/18xx/pull/8749).

michaeljb:
@Aznd4u69 I see you listed as a watcher of this repo on https://github.com/tobymao/18xx/watchers, you'll want to click "Unwatch" in the top right corner to stop getting emails

lib/engine/game/g_1858/game.rb
ollybh: Is there a safe way of doing this? From my reading of the (very brief) Opal docs [[https://opalrb.com/docs/guides/v1.7.2/unsupported_features](https://opalrb.com/docs/guides/v1.7.2/unsupported_features)] this will be integer division in Opal as the divisor is an integer. The terrain costs at 20, 40, 80 and 120, so I believe this should be fine.

lib/engine/game/g_1858/step/track.rb
ollybh: Good point. I have just reused the existing message for semi-restrictive upgrades, but I'll add something about track gauge.

lib/engine/game/g_1858/step/track.rb
ollybh: Yes, that works and is neater. I hadn't realised that you could use the splat operator on a nil value inside an array definition.

lib/engine/game/g_1858/game.rb
michaeljb: > Is there a safe way of doing this?

There are a few places where the division uses a float for the divisor and then `ceil` to round up:

https://github.com/tobymao/18xx/blob/a35337642a6a6f8ea36e046c21e4b9dfd48b38ad/lib/engine/company.rb#L28

If you call `.to_i` on the integer, Ruby (and I presume Opal) will round that down.

> From my reading of the (very brief) Opal docs [https://opalrb.com/docs/guides/v1.7.2/unsupported_features] this will be integer division in Opal as the divisor is an integer.

> > Integer / Float difference
> >
> > In Opal, both integers and floats belong to same class `Number` (using JavaScript native numbers). So `1 / 4` is `0.25` (not `0`) and `4.0 / 2` is `2` (not `2.0`).

So if the divisor is the integer `2`, and the numerator is _odd_, you'd end up with a floating point in Opal, but an integer in Ruby. Note the `1/4` example.

> The terrain costs are 20, 40, 80 and 120, so I believe this should be fine.

Given that, I think everything should be fine without further changes.

michaeljb (approved):

PR #8679, opened 2023-01-20
[1880] game end conditions, private abilities, 2p trains, force exchange rocket at b3, cross buying trains after 3 train, bcr

bentziaxl:
1880 implementation:
game end conditions
private abilities, 
2p trains, 
force exchange rocket at b3, 
cross buying trains after 3 train, 
bcr choose starting cert,
make corp.all_abilities return also all owning_player abilities by corp owner

PR #8750, opened 2023-01-28
Fix terrain costs in 1889, partial rollback of #8679

### LINKED ISSUE #8744: [Shikoku 1889]: GameError: Blocking step Place a Token cannot process action
- Game:  https://18xx.games/game/110147
- Action: https://18xx.games/game/110147?action=356

**How to Reproduce the Bug?**

Honestly not sure what happened. A past action (356) where TR ran routes seems to have somehow broken the current game state on their next turn. Really strange!

michaeljb:
In 1889, the SMR private gives a discount of ¥80 to mountain terrain, only if there is no river terrain.

Some other approach will be needed for SMR to continue to work properly while allowing 1880's P4 to also work

Fixes #8744
Reverts one change from #8679 

---

This will fix 1889 but break any 1880 games relying on #8679's terrain ability change.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1531,7 +1531,7 @@ module Engine
         end
 
         tile.upgrades.sum do |upgrade|
-          discount = ability && upgrade.terrains.include?(ability.terrain) ? ability.discount : 0
+          discount = ability && upgrade.terrains.uniq == [ability.terrain] ? ability.discount : 0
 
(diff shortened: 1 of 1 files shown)


crericha:
What's the point of fixing one game to break another? I think the right fix is to add another property to the ability to distinguish between only that terrain vs includes that terrain.

michaeljb:
Closing this one, Chris will implement the proper fix, per Slack chat

Issue #8706, opened 2023-01-25
1822PNW Portage tile - dock upgrade tile conflict


fanman80:
Game: https://18xx.games/game/110083

In OR 4.1 minor 17 went first, acquired P-12 Dock Upgrades (small), and placed the tile on hex N9.  
![image](https://user-images.githubusercontent.com/70117294/214469189-56e83543-6a43-4906-953d-454d351f3f97.png)

Later that OR minor 18 acquired P16-Pacific Portage Company, and placed one of the portage tiles on hex N9.
![image](https://user-images.githubusercontent.com/70117294/214469296-65fdfa76-5505-4819-b2d3-d6478de417e9.png)

Apparently the small dock upgrade is no longer accessible.

This seems platform independent.  The same views are seen on my laptop (Windows with Brave browser) and phone (Android also Brave browser).

Edit: Per rule 5.7.13 next to last sentence the portage tile should not have been allowed to be placed in hex N9.

"5.7.13  Offshore and offboard track spurs (adjacent to various hex grid edges) allow greater flexibility in track placement and upgrade. Six of the offboard track spurs have an anchor icon next to them. These six locations, coming off the main map hexes at Vancouver, BC, Bellingham, Olympia, Port Townsend, Tacoma and Portland, are the only spots where a port tile may be laid. Any port location may only have one port tile. **Port tiles and Pacific Portage Company tiles can’t be stacked. The first tile laid blocks the placement of the other.** For more information see the private company descriptions at the end of the rules."

PR #8756, opened 2023-01-29
[1822PNW] Don't allow portage tiles on top of port tiles

outsidePasser:
Fixes: #8706


outsidePasser:
Pushing a fix to use the standard tile numbers.

outsidePasser:
This fix will require 22PNW games to be pinned

crericha (approved):

PR #8761, opened 2023-01-29
[1880] fix 2p not showing up

### LINKED ISSUE #8748: 1880 Unable to buy 2P train and wrong ending company selected
copied and pasted 

[Kenneth Kuehner]

This is for game Id: 111550  The first thing I noticed is the 6E train is selling for 600 and it's supposed to be 700.  A player wanted to buy a 2 train when they became available and was unable to, he did meet all requirements had money and a train space, when the game end was activated instead of the player who activated being the one who trigger the next round it put it on the next player in this case QSR player when the WNR player activated game end and then when the game should of ended it lets players keep playing infinitely. Those are the things I noticed. As far as game play help, it would be nice it the chat would tell how much each player got for the bonus run amounts when they run. That's all I have. Thanks for putting 1880 on the website and the work you do for the website, I know it's a lot of work!

bentziaxl:
fix #8748

### DIFF
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -65,8 +65,7 @@ module Engine
 
           def buyable_trains(entity)
             trains_to_buy = super
-            trains_to_buy.reject! { |t| t.name == '2P' } if owns_restored2?(entity)
-            trains_to_buy << restored2_train if can_buy_restored2?(entity)
+            trains_to_buy.reject! { |t| t.name == '2P' } unless can_buy_restored2?(entity)
             # Can't buy trains from other corporations until train 3
             return trains_to_buy if @game.can_cross_buy?
 
@@ -78,9 +77,10 @@ module Engine
           end
 
           def can_buy_restored2?(entity)
-            return false if !@game.phase.available?(restored2_train&.available_on) || owns_restored2?(entity)
-
-            restored2_train.price <= entity.cash
+            @game.phase.available?(restored2_train&.available_on) &&
+            !owns_restored2?(entity) &&
+            @corporations_sold.empty? &&
+            entity.cash >= restored2_train.price
           end
 
           def restored2_train
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -65,11 +65,26 @@ module Engine
 
           def buyable_trains(entity)
             trains_to_buy = super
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1880/step/buy_train.rb
```diff
@@ -65,12 +65,27 @@ def setup
 
           def buyable_trains(entity)
             trains_to_buy = super
-            trains_to_buy.reject! { |t| t.name == '2P' } if entity.trains.find { |t| t.name == '2P' }
+            trains_to_buy.reject! { |t| t.name == '2P' } if owns_restored2?(entity)
+            trains_to_buy << restored2_train if can_buy_restored2?(entity)
```
crericha, line 69: Why do you have to both reject and add?

bentziaxl: It's fun train logic. Currently the 2p shows up when a player has enough money to buy the cheapest non-2p train.
I need the 2p to show up when the player has enough to pay for it and be removed when they already own one. 
Is there a better way?

crericha: Why isn't it showing up? I couldn't find where it was being removed. Can you point me at that code?

bentziaxl: It's a combo of:
2p not being the cheapest depot train (due to not counting as train for EMR)
and this code:
https://github.com/tobymao/18xx/blob/master/lib/engine/step/train.rb#L152

bentziaxl: this is the bug

crericha: Why don't we set EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST to false and then in buyable_trains, reject 2p if the player's cash is less than the cost of a 2p (or they have sold a share)?

crericha: You know if shares have been sold by `@corporations_sold` not being empty

bentziaxl: it seems convoluted to change EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST value to handle a non ebuy scenario and rely on the fact that this game only has one train on sale at the time from the depot and no rule about ebuy trains and cheapest to make this work. If you think it's a better approach i'll do it, but i don't see how it's an improvement over this

crericha (approved):

PR #8762, opened 2023-01-29
[1880] add stub fix + typos

### LINKED ISSUE #8753: [1880] Beijing exit spaces not protected
It is possible to lay track which blocks a not yet started company from building out of Beijing. This should not be allowed.
[hs_ffzpfsdg_1674941485.json.txt](https://github.com/tobymao/18xx/files/10528373/hs_ffzpfsdg_1674941485.json.txt)

bentziaxl:
add stubs are restricted (fix #8753 )
add terminal to offboards so FI connection doesn't wrongly close FIs
fix typos 
fix SR log showing SR 5.1 when it should be SR 5

### DIFF
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -5,6 +5,7 @@ require_relative 'map'
 require_relative 'entities'
 
 require_relative '../base'
+require_relative '../stubs_are_restricted'
 
 module Engine
   module Game
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -13,6 +13,7 @@ module Engine
         include_meta(G1880::Meta)
         include Map
         include Entities
+        include StubsAreRestricted
 
         attr_accessor :train_marker
         attr_reader :full_cap_event, :communism, :end_game_triggered, :saved_or_round, :final_operating_rounds
@@ -301,7 +302,7 @@ module Engine
           return if corporation.ipo_shares.size > 5 || corporation.fully_funded
 
           amount = corporation.original_par_price.price * 5
-          @log << "Five shares of #{corporation.name} have been boought. "\
+          @log << "Five shares of #{corporation.name} have been bought. "\
                   "#{corporation.name} receives #{format_currency(amount)}"
           @bank.spend(amount, corporation)
           corporation.fully_funded = true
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_1880/game.rb
```diff
@@ -712,7 +713,7 @@ def lessee
         def round_description(name, round_number = nil)
           round_number ||= @round.round_num
           description = super
-          description += ".#{round_number}" if @round.is_a?(Engine::Round::Operating)
+          description += ".#{round_number}" if name == self.class::OPERATING_ROUND_NAME
           description += " - Train Marker at #{@train_marker.name}" if @train_marker && !@end_game_triggered
```
bentziaxl, line 717: needed change, timing issue.
When round_description is called for logging during an SR, it's still the OR, so the previous conditional will resolve to true even when i want to log the SR round. 
https://github.com/tobymao/18xx/blob/master/lib/engine/game/base.rb#L2627

lib/engine/game/g_1880/game.rb
```diff
@@ -13,6 +13,7 @@ class Game < Game::Base
         include_meta(G1880::Meta)
         include Map
         include Entities
+        include StubsAreRestricted
```
ollybh, line 16: Do you also need to `require_relative '../stubs_are_restricted'` before including this module?

lib/engine/game/g_1880/step/check_fi_connection.rb
```diff
@@ -87,12 +87,26 @@ def process_percent(action)
 
           def process_token(action)
             fi = @merging[:fi]
+<<<<<<< HEAD
```
ollybh: This merge conflict needs to be resolved.

lib/engine/game/g_1880/game.rb
bentziaxl: yep, added

lib/engine/game/g_1880/step/check_fi_connection.rb
bentziaxl: i removed it....where are you seeing this?

ollybh: My bad, I'd clicked onto the wrong page and was looking at the first commit, not the full pull request.

crericha (approved):

PR #8763, opened 2023-01-29
Add test to check for typos in tile colours

ollybh:
1822PNW was broken by issue #tobymao/18xx#8732. This was caused by a typo in the colour name part of a tile definition. This adds a rspec test to catch mistakes like this, which are easily made by those of us who think that 'grey' is the correct spelling. It checks that all colour names are part of the Engine::Config::Tile.COLORS list.

### DIFF
--- a/spec/lib/engine/games/config_spec.rb
+++ b/spec/lib/engine/games/config_spec.rb
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 require './spec/spec_helper'
+require './lib/engine/config/tile'
 
 require 'json'
 
 module Engine
+  include Engine::Config::Tile
+  tile_colors = COLORS
   players = ('a'..'z')
 
   Engine::GAME_METAS.each do |game_meta|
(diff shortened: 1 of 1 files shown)


spec/lib/engine/games/config_spec.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 require './spec/spec_helper'
+require './lib/engine/config/tile'
 
 require 'json'
 
 module Engine
+  include Engine::Config::Tile
+  tile_colors = COLORS
```
ollybh, line 10: Is there a cleaner way of getting hold of the `Engine::Config::Tile.COLORS` array? This works but it also includes everything else from the Tile module.

michaeljb: Down in the spec you could just reference `Engine::Config::Tile::COLORS`. You still need to `require` the whole file, but then you don't need to `include` the module.

ollybh: Thanks for the suggestion, that's much better.

tobymao (approved):

PR #8769, opened 2023-01-30
[1880] Fix OR to SR transition

crericha:


bentziaxl (approved):
looks good, tested on a few games

PR #8783, opened 2023-02-02
[1880] fix edge case 4t rocket

### LINKED ISSUE #8781: [1880] Game locks after the 4 trains when rocket is unused, and all 4 trains need to be exported
We're unable to load the game after the 4s were exported: https://18xx.games/game/111402
It just sits on a Loading Game screen.

There was another report in chat with a similar issue, but i don't know if anything was filed.

### LINKED ISSUE #8803: Error in 1880
https://18xx.games/game/111314?active=1

I'm running into an error exchanging the Rocket for a T4. "The server does not accept this action due to wrong number of argument. Given 2, expect 3"

bentziaxl:
fixes #8781

Edge case where rocket of china hasn't been used AND a full operating order passes on the 4 trains.
see here: 
https://boardgamegeek.com/thread/1042314/forced-exchange-p7-rocket-and-removal-unpurchased
"If no one buys a 4T, as you pick them up to get rid of them, when you pick the first one up, the second one zips off to the P7 owner and then all the rest go away. I'll admit this isn't clear." and "The discarding of the 4-trains is one after the other, so the Rocket of China will be exchanged into the second 4-train in case of all 4-trains are discarded.
Lonny"

fixes #8803 as well. @game.phase.buying_train! has 3 params but the code was only passing 2.

### DIFF
--- a/lib/engine/game/g_1880/step/assign.rb
+++ b/lib/engine/game/g_1880/step/assign.rb
@@ -69,14 +69,13 @@ module Engine
 
           def process_assign_rocket(action)
             buying_corp = action.target
-            train = @game.depot.upcoming.first
+            train = @game.rocket_train || @game.depot.upcoming.first
             @log << "#{buying_corp.name} exchanges the #{@game.rocket.name} for a #{train.name} train"
 
             @game.rocket.close!
             @game.buy_train(buying_corp, train, :free)
-            @game.phase.buying_train!(buying_corp, train)
-            @game.depot.export_all!(train.name, silent: true) if @round.special_edge_case_4t_rocket
-            @round.special_edge_case_4t_rocket = false
+            @game.phase.buying_train!(buying_corp, train, @game.depot)
+            @game.depot.export_all!(train.name, silent: true)
           end
         end
       end
--- a/lib/engine/game/g_1880/step/assign.rb
+++ b/lib/engine/game/g_1880/step/assign.rb
@@ -11,7 +11,8 @@ module Engine
           def actions(entity)
             return [] if entity.player?
             return [] unless @game.abilities(entity, :assign_corporation)
-            return [] if current_entity.minor? || (entity == @game.p5 && current_entity.building_permits&.include?('D'))
+            return [] if current_entity.minor? ||
+                        (entity == @game.p5 && current_entity.corporation? && current_entity.building_permits&.include?('D'))
 
             return ACTIONS if @game.forced_exchange_rocket? && entity == @game.rocket
             return ACTIONS_WITH_PASS if p5_block? && current_entity.owner == @game.p5.owner
@@ -74,6 +75,8 @@ module Engine
             @game.rocket.close!
             @game.buy_train(buying_corp, train, :free)
             @game.phase.buying_train!(buying_corp, train)
+            @game.depot.export_all!(train.name, silent: true) if @round.special_edge_case_4t_rocket
+            @round.special_edge_case_4t_rocket = false
           end
         end
       end
--- a/lib/engine/game/g_1880/step/assign.rb
+++ b/lib/engine/game/g_1880/step/assign.rb
@@ -75,7 +75,6 @@ module Engine
             @game.rocket.close!
             @game.buy_train(buying_corp, train, :free)
             @game.phase.buying_train!(buying_corp, train, @game.depot)
-            @game.depot.export_all!(train.name, silent: true)
           end
         end
       end
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -35,9 +35,7 @@ module Engine
 
           def round_state
             super.merge(
-            {
-              bought_trains: false,
-            }
+            { bought_trains: false }
           )
           end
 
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_1880/step/assign.rb
```diff
@@ -11,7 +11,8 @@ class Assign < Engine::Step::Assign
           def actions(entity)
             return [] if entity.player?
             return [] unless @game.abilities(entity, :assign_corporation)
-            return [] if current_entity.minor? || (entity == @game.p5 && current_entity.building_permits&.include?('D'))
+            return [] if current_entity.minor? ||
+                        (entity == @game.p5 && current_entity.corporation? && current_entity.building_permits&.include?('D'))
```
bentziaxl, line 16: game tries to call current_entity.building_permits on a private (the rocket)

lib/engine/game/g_1880/step/buy_train.rb
```diff
@@ -55,7 +58,14 @@ def discard_trains?
           def discard_all_trains(train_name)
             @game.log << "#{train_name} has not been bought for a full operating order, "\
                          "removing all remaining #{train_name} trains"
-            @game.depot.export_all!(train_name, silent: true)
+            return @game.depot.export_all!(train_name, silent: true) unless special_edge_case_4t_rocket?(train_name)
+
+            @game.depot.export!(train_name, silent: true)
+            @round.special_edge_case_4t_rocket = true
+          end
```
bentziaxl, line 65: only export one if we're in 4t rocket, edge case territory.
The assign step will kick in, and prompt the owner of the rocket to pick corp to gain the 4

lib/engine/game/g_1880/step/assign.rb
```diff
@@ -74,6 +75,8 @@ def process_assign_rocket(action)
             @game.rocket.close!
             @game.buy_train(buying_corp, train, :free)
             @game.phase.buying_train!(buying_corp, train)
+            @game.depot.export_all!(train.name, silent: true) if @round.special_edge_case_4t_rocket
+            @round.special_edge_case_4t_rocket = false
           end
```
bentziaxl, line 80: if in special edge cage, export the rest of the trains.

crericha:
Agree this is ugly. I won't have time until Sunday to provide suggestions on any opportunities I see to clean it up.

lib/engine/game/g_1880/step/buy_train.rb
```diff
@@ -55,7 +58,14 @@ def discard_trains?
           def discard_all_trains(train_name)
             @game.log << "#{train_name} has not been bought for a full operating order, "\
                          "removing all remaining #{train_name} trains"
-            @game.depot.export_all!(train_name, silent: true)
+            return @game.depot.export_all!(train_name, silent: true) unless special_edge_case_4t_rocket?(train_name)
+
+            @game.depot.export!(train_name, silent: true)
+            @round.special_edge_case_4t_rocket = true
```
michaeljb, line 64: I think this is a bit more readable, and then I don't think anything here is that ugly:

```suggestion
            if special_edge_case_4t_rocket?(train_name)
              @game.depot.export!(train_name, silent: true)
              @round.special_edge_case_4t_rocket = true
            else
              @game.depot.export_all!(train_name, silent: true) unless 
            end
```

lib/engine/game/g_1880/step/assign.rb
```diff
@@ -74,6 +75,8 @@ def process_assign_rocket(action)
             @game.rocket.close!
             @game.buy_train(buying_corp, train, :free)
             @game.phase.buying_train!(buying_corp, train)
+            @game.depot.export_all!(train.name, silent: true) if @round.special_edge_case_4t_rocket
+            @round.special_edge_case_4t_rocket = false
```
michaeljb, line 79: It seems like if `special_edge_case_4t_rocket` is already `false`, so you don't need to re-set it every time? So maybe this works:

```suggestion
            if @round.special_edge_case_4t_rocket
              @game.depot.export_all!(train.name, silent: true) 
              @round.special_edge_case_4t_rocket = false
            end
```

lib/engine/game/g_1880/step/buy_train.rb
bentziaxl: The ugliness is that I'm exporting one train, and relying on another event defined somewhere else in the code to kick in, take over and handle the rest. 
It makes for hard to maintain/debug code. It's not obvious what's happening here, unless you know the inner working of the system and specifically the 1880 assign.rb class.

michaeljb: Alright, I'm not very familiar with 1880 so I'll leave it to @crericha for code review suggestions.

crericha:
What would be nice is if the edge case detection only happened in one location and then your regular assign step logic took care of the rest.

Maybe something like in game.after_buying_train after the trigger_sr? check:

if train.name == '4' && !@game.rocket.closed? && @game.depot.upcoming.first.name != train.name
  @rocket_train = train
  return
end

and in the assign step's process_assign_rocket:

train = @game.rocket_train || @game.depot.upcoming.first

bentziaxl:
yeah, that works and is much cleaner

crericha (approved):

PR #8791, opened 2023-02-04
validation: Catch all errors in `process_action`

michaeljb:
In `validate.rb`'s `run_game()` function, the game engine is loaded outside of the `begin`/`rescue` block, before calling `maybe_raise!` inside the block, so that an error in the game state can be caught without crashing the whole program.

The changes in the recent PRs #8735, #8756, and #8787 all changed/removed certain game objects like trains and tiles. This caused the game engine to error out when in `process_action` when converting the action to a hash (for example, the 1852: Missouri change removed some trains from the roster, so actions using the removed trains tried calling `@train.id` where `@train` was `nil`) before `process_single_action` was called.

That made the `validate.rb` functions impossible to use to identify which games to archive; the `NoMethodError` on the `nil` train crashed the whole thing.

An alternative change in `run_game()` would be to move the game load into the `begin`/`rescue` block, but when the other kinds of errors are raised, the `engine` ends up being `nil`, causing the code in the `rescue` block to just raise further errors.

https://github.com/tobymao/18xx/blob/94ac0bbc49a3de286a9d74e6b3cb51249ec017ef/validate.rb#L22-L39

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -748,6 +748,9 @@ module Engine
         @last_processed_action = action.id
 
         self
+      rescue StandardError => e
+        rescue_exception(e, action)
+        self
       end
 
       def process_single_action(action)
@@ -774,6 +777,10 @@ module Engine
           end
         end
       rescue Engine::GameError => e
+        rescue_exception(e, action)
+      end
+
+      def rescue_exception(e, action)
         @raw_actions.pop
         @actions.pop
         @exception = e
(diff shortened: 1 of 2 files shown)


scottredracecar:
Michael, this seems like a great and needed feature for how the site is
being used to develop games. What happens to a game with an illegal action
that this catches?

tobymao (approved):

michaeljb:
Hi Scott, illegal actions should be treated the same as before--the game cannot continue, an error message should be displayed showing what action the game was broken on, etc.

This just fixes an issue where the script we use to identify which games need to be archived/pinned was crashing entirely on the first error it encountered, rather than capturing what game and action that error occurred on and moving on to the next game

Issue #7490, opened 2022-03-18
[1852] Half-price token not working with teleport ability?


BellmakerSA:
https://18xx.games/game/76943

CBQ has Half-Price token, they can use it to teleport to KC. But it doesn’t give a discount and then the ability goes to 0/1

scottredracecar:
https://18xx.games/game/76943?action=163

I agree that should work, but it might fall into the category that combining two abilities is unsupported. This affects a few titles--mostly comes up for 1822.

roseundy:
Definitely a bug.

PR #8793, opened 2023-02-05
18MO: Playtest changes

### LINKED ISSUE #8786: 1852 Missouri Playtesting Changes
The following is a list of changes after a round of playtesting and analysis after Chattanooga 2023.

This is what is online now, strikethrough is the changes I would like to make
image

Setup (summarizes table changes)
- Remove one more orange and blue private company for all groups (but don't remove an extra "green group" corporation)
- Change names of 2Y/3Y/3G/4G to Y2/Y3/G3/G4 ("Green 3" rather than "3 Green" which looks like a 3G goods train)
- Change starting capital for three players to 400

Map
- Remove hex edge river between B6/C5
- Change E/W bonus to +50 in A7, J2, K3
- Add a connection from the city to the north hex edge in J4 and K5, change city value to 10 for both
- Add a spike to the SW edge of J2

Rules
- East St Louis dot is an optional stop

roseundy:
Fixes #8786 

This will require archives/pins.

Given that it's been out for almost a year pending games should probably be pinned even if it's still in alpha.

### DIFF
--- a/lib/engine/game/g_18_mo/entities.rb
+++ b/lib/engine/game/g_18_mo/entities.rb
@@ -158,7 +158,7 @@ module Engine
             type: 'train_discount',
             discount: 60,
             owner_type: 'corporation',
-            trains: %w[2Y 3Y 3G 4G 3E 5 6],
+            trains: %w[Y2 Y2 G3 G4 3E 5 6],
             count: 1,
             closed_when_used_up: true,
             when: 'buying_train',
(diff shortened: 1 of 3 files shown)


michaeljb (approved):

michaeljb:
Is #7490 the only thing keeping this game in alpha instead of beta or production?

scottredracecar:
It could go to production, but I was making playtesting changes (last year)
and I didn't know when I would stop. Probably if these new changes are
okay, it could go to prototype production.

PR #8804, opened 2023-02-06
[1858] Stock round: auctions, buying and selling shares

ollybh:
This is the start of the stock round implementation for 1858. Buying and selling shares is fairly normal, the main difference being that you can't directly par a corporation before phase 5. Before that point corporations can only be started by exchanging a private railway company, this is to be added in a later commit.

In a stock round you can also put up a private company for auction. Everyone gets a chance to bid on this company, until only one person is left. This is implemented using the Engine::Step::BuySellParSharesViaBid code from 1861/1867.

### DIFF
--- a/lib/engine/game/g_1858/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1858/step/buy_sell_par_shares.rb
@@ -1,0 +1,114 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/buy_sell_par_shares_via_bid'
+
+module Engine
+  module Game
+    module G1858
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParSharesViaBid
+          # Buy actions that will end a player's turn in the stock round.
+          PURCHASE_ACTIONS = [Action::Bid, Action::BuyShares, Action::Par].freeze
+
+          def actions(entity)
+            return [] unless entity == current_entity
+            return %w[bid pass] if @auctioning
+            return ['sell_shares'] if must_sell?(entity)
+            return [] if bought?
+
+            actions = []
+
+            # Sell actions.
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1858/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,114 @@
...
+          end
+
+          def can_bid_company?(player, company)
+            (@auctioning.nil? || @auctioning == company) &&
+            auctionable_companies.include?(company) &&
+              (min_bid(company) <= player.cash)
+          end
+
+          def log_skip(entity)
+            super unless @round.current_actions.any?
```
tobymao, line 108: if empty?

lib/engine/game/g_1858/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,114 @@
...
+          def auctionable_companies
+            @game.buyable_bank_owned_companies
+          end
+
+          def can_bid_any?(player)
+            auctionable_companies.any? { |company| can_bid_company?(player, company) }
+          end
+
+          def can_bid_company?(player, company)
+            (@auctioning.nil? || @auctioning == company) &&
```
tobymao, line 102: !@auctioning

lib/engine/game/g_1858/step/buy_sell_par_shares.rb
ollybh: Sorry, I thought I'd already got rid of all of these blockless `any?` and `none?` calls. It's gone now.

lib/engine/game/g_1858/step/buy_sell_par_shares.rb
ollybh: Changed.

PR #8816, opened 2023-02-08
[1850] mesabi tokens

bentziaxl:
- clean map 
- add mesabi token logic
- add tile lay definition similar to 18cz (2 lays for first OR of a corp, 1 lay/upgrade after)
- add icons

### DIFF
--- a/lib/engine/game/g_1850/step/buy_company.rb
+++ b/lib/engine/game/g_1850/step/buy_company.rb
@@ -13,7 +13,7 @@ module Engine
             companies = @game.purchasable_companies(entity)
 
             entity == current_entity &&
-            companies.any? &&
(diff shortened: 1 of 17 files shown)


lib/engine/game/g_1850/step/buy_company.rb
```diff
@@ -0,0 +1,23 @@
...
+    module G1850
+      module Step
+        class BuyCompany < Engine::Step::BuyCompany
+          def can_buy_company?(entity)
+            return super unless @game.phase.name == '2'
+
+            companies = @game.purchasable_companies(entity)
+
+            entity == current_entity &&
+            companies.any? &&
```
tobymao, line 16: !companies.empty

PR #8829, opened 2023-02-11
[18PNW] swap A and C regional starts

### LINKED ISSUE #8822: 1822PNW Swap locations of A and C
In the physical game, the locations of regionals A and C are swapped from what is on 18xx.games (top to bottom like the company numbering). 

A should be in Calgary, C should be in Vancouver, WA.

benjaminxscott:
closes #8822


outsidePasser:
I know PNW is in Alpha, but it really shouldn't be at this point (I'll check it in a bit), so please PIN broken games.

PR #8838, opened 2023-02-12
1822 PNW convert to latest rulebook and move to Beta

outsidePasser:


assets/app/view/game/round/stock.rb
```diff
@@ -318,7 +318,8 @@ def render_bank_companies
             },
           }
 
-          @game.buyable_bank_owned_companies.map do |company|
+          companies = @step.respond_to?(:bank_owned_companies) ? @step.bank_owned_companies : @game.buyable_bank_owned_companies
```
tobymao, line 321: why not just override this in your game? respond_to? in general is not a great way to go

lib/engine/game/g_1822_pnw/step/conversion.rb
```diff
@@ -0,0 +1,37 @@
...
+            return [6] if (6 * par) < minors_value && par == 100 && minors_cash >= (minors_value - 600)
+            return [] if (6 * par) < minors_value
+
+            min_shares = min_exchange_shares(par, minors_value, minors_cash)
+            max_shares = max_exchange_shares(par, minors_value, player_cash)
+            (2..10).to_a.select { |n| n >= min_shares && n <= max_shares }
+          end
+
+          def can_par_at?(par, minors_cash, minors_value, player_cash)
+            return true unless possible_exchanged_shares(par, minors_cash, minors_value, player_cash).empty?
```
tobymao, line 31: ! possible_exchanged_shares(par, minors_cash, minors_value, player_cash).empty? 

this is already a boolean right

assets/app/view/game/round/stock.rb
outsidePasser: The step in this case has the knowledge of which company is being formed, but the game doesn't know that.  I'm not sure how to get the game to know about it without game doing the same thing (@step.respond_to?).  I was following the pattern that shows up several other times in stock.rb.

lib/engine/game/g_1822_pnw/step/conversion.rb
outsidePasser: It's not a boolean - it's an array that contains possible values for the number of shares that can be exchanged.

outsidePasser: But with empty? it is a boolean.  Oops. Yes, I'll change that.

assets/app/view/game/round/stock.rb
tobymao: the game knows what step it is though

outsidePasser: Changed to override in game.

PR #8842, opened 2023-02-13
[1858] Operating round: route and dividend steps

ollybh:
There are three types of train in 1858 that have different ways of running routes and earning revenue:

- 2H/4H/6H are hex trains. They can only run on broad and dual gauge track.
- 2M/3M/4M/5M/6M are metre (narrow) gauge trains that cannot run on broad gauge track. Their range is the number of cities and off-board areas visited.
- 5E/6E/7E/5D are express trains. They run on broad and dual gauge track. They skips towns. Their range is the number of cities and off-board area visited. 5D trains earn double revenue.

Trains are wounded before rusting. Obsolete trains earn half revenue. They are not automatically discarded after running.

Privates owned by corporations don't pay income to the corporation, instead this is added to the corporations revenue from trains. This is handled in the `routes_revenue` method.

This private revenue can result in total revenues that don't divide neatly between shareholders even on a full payout. The earnings per share is rounded down and the leftover revenue is paid to the company treasury.

Share price only increases if earnings per share are at least one tenth of the current market price. There are no double jumps.


tobymao (approved):
conflicts

ollybh:
@tobymao The conflict is now resolved. It wasn't much of a conflict, both this pull request and #8804 were adding new methods to `lib/engine/game/g_1858/game.rb` and the merger wasn't sure how to order these in some places.

PR #8849, opened 2023-02-15
separated entities & map content from game.rb

philcampeau:
While I'm figuring out how to do more complex coding steps, I figured it could be helpful do so some basic organisation of the data on the site. This update just creates a map.rb and entities.rb file for 1830, separate from the game.rb file. 

It seems to pass all tests on my machine.


philcampeau:
Still needs some tweaks. Closing for now.

PR #8851, opened 2023-02-16
1880 - fix ui highlighting multiple companies for home token and add track blocking for rocket

bentziaxl:


bentziaxl:
games that do not trade in the rocket right away will break, extra passing step needed

crericha (approved):

PR #8858, opened 2023-02-17
[1880] correct Tile Reservation Block

### LINKED ISSUE #8853: [1880] Token Placement allowed on D12 (Shenyang & Fushun - QSR Home) before QSR started
GameID: https://18xx.games/game/111286

Specific Action Link: https://18xx.games/game/111286?action=379

JHA was able to place a token on D12, which is QSR's home. QSR has not yet started. 1880 rules prohibit placing a token on either the QSR or SCR home until the associated company's home token has been placed (see bottom of page 13 of rules pdf linked in the game Info section).

philcampeau:
Fixes #8853

TILE_RESERVATION_BLOCKS_OTHERS set to :always as per rules.

### DIFF
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -21,7 +21,7 @@ module Engine
                     :foreign_investors_operate, :rocket_train
 
         TRACK_RESTRICTION = :permissive
-        TILE_RESERVATION_BLOCKS_OTHERS = :yellow_only
+        TILE_RESERVATION_BLOCKS_OTHERS = :always
         TILE_UPGRADES_MUST_USE_MAX_EXITS = %i[cities].freeze
         HOME_TOKEN_TIMING = :operating_round
         SELL_BUY_ORDER = :sell_buy


bentziaxl:
It might not be this simple, the rules say that in yellow and green (while the two cities are disjoined on the tile) all other corps are blocked, but in brown and gray other corps can lay a token as long as one space is open.
did you test if :always behaves correctly?

philcampeau:
The only mention of tokening those two hexes that i see in the rules is as follows: 

> For the two companies QSR and SCR, which start on a hex with two large cities (OO-tile) the 
> following additional rules apply: 
> - If the tile in the hex has already been promoted to green, the company may choose 
> in which of the two cities to place its home station. 
> - As long as the home station has not been placed, no other company may place a 
> station marker in the hex.

I'm not seeing the section you're mentioning in either the Lookout or the Double O rules. It's very possible I'm just missing it. Could you quote the section?

bentziaxl:
Maybe we need another bgg thread, but my interoperation is that this rule section only applies to yellow and green, while it's an OO tile, with two distinct cities, as QSR/SCR have a free choice between those two cities. In brown it's a single city with two slots, so the choice isn't there anymore, and regular 18xx rules should apply.

crericha:
These two rules are at odds:

- If a city contains the home base of a company that is not yet in operation, other companies may only build a
station there if that leaves at least one free place on the tile for the not yet operating company to place its station
marker.
- For the two companies QSR and SCR, which start on a hexagon with two large cities (OO-tile) the following
additional rules apply:
  - If the tile in the hexagon has already been upgraded to green, the company may choose in which of the two
cities to place its home station.
  - As long as the home station has not been placed, no other company may place a station marker in the
hexagon

Technically, its clear that the QSR and SCR home hexagons cannot be placed in; however, @bentziaxl brings up a good point of was the intention until the city combined or regardless. I think it's fair to seek errata for the rule before applying a change.

And FWIW, the code was intentionally written as it was because I had the same interpretation as @bentziaxl, though it should exclude yellow and green, not just yellow.

bentziaxl:
posted the question in bgg: 
https://boardgamegeek.com/thread/1787508/tokens-arrows

philcampeau:
Helmut Ohley confirms the reservation is only for yellow/green, so this is the wrong fix.

https://boardgamegeek.com/thread/3030201/article/41751358#41751358

PR #8871, opened 2023-02-18
[1880] fix scr qsr placement blocking

bentziaxl:
fix #8853

PR #8879, opened 2023-02-19
[1880] Fix OO hex tile reservations

crericha:
The current fix in #8871 assumes all tiles with more than one city is an OO. This is not a valid way to identify an OO city. There are many games with more than one extra city on a tile that is not an OO tile and thus the name of the feature is incorrect.

This fix attempts to provide a clearer blocking mechanism that can be reused without providing unexpected behavior.

Also removed yellow_only as it was added for 1880 and not used by any other game.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -198,9 +198,7 @@ module Engine
       # do tile reservations completely block other companies?
       # :never -- token can be placed as long as there is a city space for existing tile reservations
       # :always -- token cannot be placed until tile reservation resolved
-      # :yellow_only -- token cannot be placed while tile is yellow or until the tile reservation is resolved
-      # :oo_only -- token cannot be places while the tile has two or more seperate cities (OO tile),
-      #             if and when the cities merge into one, regular reservation rules apply.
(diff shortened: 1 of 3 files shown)


michaeljb:
@crericha sorry about #8871, my mistake.

michaeljb:
@crericha LGTM but I can let someone else approve/merge (even if it's just you 🙂)

crericha:
> @crericha sorry about #8871, my mistake.

No worries. I encourage reviewers to be super critical of common code changes and to think about them from a the general use case. It's super easy for us developers (myself included) to tunnel-vision on the game we are developing and miss the forest from the trees.

> @crericha LGTM but I can let someone else approve/merge (even if it's just you 🙂)

The currently deployed code will work for 1880, so no rush to merge. I'd like Ben to weigh-in and see if he agrees before merging since this PR reworks his PR.

bentziaxl (approved):

bentziaxl:
Yeah, I definitely had tunnel-vision when fixing this. Thanks @crericha for cleaning my mess.

@michaeljb don't feel bad for approving my fix, it was a clean change, logic was sound, i would have approved it as well.

PR #8843, opened 2023-02-13
[1858] Quick start and two player variants

ollybh:
The quick start variant skips the initial auction of private companies, and instead randomly assigns players predefined batches of privates.

The two-player variant has seven privates unavailable for purchase during the game, but still blocking track building and tokening. The maximum percentage of a corporation that can be bought is increased from 60% to 70%.

PR #8888, opened 2023-02-19
[1858] Operating round: buying/discarding trains and issuing shares

ollybh:
Implements the buy trains and issue shares steps of the operating round.

Non-standard things in 1858:
- Obsolete (wounded) trains can be discarded in the train buying step.
- The last tier of non-permanent trains (6H and 4M trains) are rusted after five grey trains have been purchased (three grey trains in the two-player) variant. The variable  `@phase7_trains_bought` is used to track how many grey trains have been bought from the depot, and a rusting event is triggered once the fifth is bought.
- A corporation does not have to own a train. If it doesn't then its share price moves left.
- Emergency money raising can be done by issuing company shares or the president contributing cash, or both. The president cannot sell shares.
- Even if a corporation can afford a cheaper train from the depot, EMR can be used to buy a more expensive one.


lib/engine/game/g_1858/game.rb
```diff
@@ -283,6 +288,25 @@ def submit_revenue_str(routes, _show_subsidy)
             "#{format_revenue_currency(private_revenue)} private revenue"
         end
 
+        def buy_train(operator, train, price = nil)
+          bought_from_depot = (train.owner == @depot)
+          super
+          return if @phase7_trains_bought >= @phase7_train_trigger
```
bentziaxl, line 294: wouldn't be easier/more robust to check that the train already rusted?

lib/engine/game/g_1858/step/buy_train.rb
```diff
@@ -0,0 +1,115 @@
...
+
+          def process_sell_shares(action)
+            raise GameError, "#{entity.name} cannot sell shares at this time" unless action.entity.corporation?
+
+            @last_share_issued_price = action.bundle.price_per_share
+            @game.share_pool.sell_shares(action.bundle)
+            old_price = action.entity.share_price
+            @game.stock_market.move_left(action.entity)
+            @game.log_share_price(action.entity, old_price)
+          end
```
bentziaxl, line 74: can EMR cause president change? do you need to add code to check for that?

ollybh: The president can't sell shares in EMR in 1858 (`EBUY_CAN_SELL_SHARES = false`). This code is for issuing shares from the corporation treasury, so can't trigger a presidency change.

lib/engine/game/g_1858/game.rb
ollybh: I'd hope that this code doesn't skip any train buys, but it could be changed. Is there an easy way of checking if a class of train has rusted?

tobymao:
conflicts

ollybh:
> conflicts

Now fixed (this was a small conflict with #8843, with both adding lines to the same method).

PR #8889, opened 2023-02-19
[18BF] Map, entity, tile, market and train definitions

ollybh:
[18BF](https://github.com/tobymao/18xx/wiki/18BF-The-Big-Four) is a prototype from Ian D Wilson. It's a game in the 1861/1867/1812 family. It main difference from 1861 is a second set of mergers, with public companies merging to form 20-share systems. It's also got a bigger map, more privates (in three flavours: private railways, ferries and undergrounds), more minors, and more varied trains.

The game is being implemented as a subclass of G1867 as it can reuse a lot of that game's code.

This pull request adds definitions for the game maps and tiles, all the different entities, trains and the stock market.

This only touches the shared code in one place, adding a `type` property to the `Company` class.

The London hex has twelve token slots (two connected to each of the edges). Having this many slots in a single hex does not render at all well. Instead an extra set of hexes for London are dropped into the North Sea, with six hexes with two token slots each. These are stitched onto the main map by setting the various hexes' `neighbours` so that routes into and out of London work correctly.

There's also a little bit of code to set up the game correctly, and to calculate route revenue and bonuses.

image
[Screenshot of the southern half of the map, showing a train running from SE's token in London out through Kent]


tobymao:
why did you did you add type to company

lib/engine/game/g_18_bf/game.rb
```diff
@@ -0,0 +1,153 @@
...
+          setup_company_price_up_to_face
+
+          @show_majors = false
+          setup_london_hexes
+          setup_bonuses
+        end
+
+        def setup_preround
+          # Randomise the privates available and the minor company order.
+          random = Random.new(rand)
```
tobymao, line 40: why are you using Random? you can't do that, you can only use the rand method in the game as that works on both server and client

ollybh:
> why did you did you add type to company

For convenience. There are three different types for private company in 18BF: private railways, ferry companies, and underground railways. There are slightly different rules for each, and having the `type` attribute makes them easier to differentiate.

There are already `type` attributes for corporations and minors, so this would mirror those.

This change to the `Company` class isn't essential, it wouldn't be much of a challenge to add other code to work out the different types of private companies. Let me know if this should be changed.

lib/engine/game/g_18_bf/game.rb
ollybh: I'd misunderstood the use of the Game.rand method. I've now removed the use of the Ruby Random class, and I'm instead using `Array.sort_by { rand }`.

lib/engine/game/g_18_bf/entities.rb
```diff
@@ -0,0 +1,1008 @@
...
+            5
+          when :system
+            10
+          end
+        end
+
+        def setup_companies
+          msg = 'The three private railway companies removed from play are: '
+          rejected = @companies.select { |c| c.type == :railway  }.sort_by { rand }.take(3)
+          rejected.sort_by { |c| @companies.find_index(c) }.each do |company|
```
tobymao, line 975: don't use find_index, always use index

lib/engine/game/g_18_bf/map.rb
```diff
@@ -0,0 +1,336 @@
...
+
+        def bonus_london_offboard(train, stops)
+          return 0 if stops.none?(&:offboard?) ||
+                      stops.none? { |stop| stop.groups.include?('London') }
+
+          revenue_bonus(:london, train)
+        end
+
+        def bonus_mine(train, stops)
+          return 0 if stops.none?
```
tobymao, line 312: stops.empty?

lib/engine/game/g_18_bf/map.rb
```diff
@@ -0,0 +1,336 @@
...
+
+          revenue_bonus(:mine, train)
+        end
+
+        def bonus_scottish_border(train, stops)
+          hexes = stops.map(&:hex)
+          # The border bonus is payable if:
+          # - At least one of the route's stops is in Scotland.
+          # - At least one stop is outside Scotland.
+          return 0 if !hexes.intersect?(@scotland) || hexes.difference(@scotland).none?
```
tobymao, line 322: empty?, replace all none? calls with !empty? or empty?

empty? checks size, none is a N operation and checks for nils

lib/engine/game/g_18_bf/step/special_track.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Game
+    module G18BF
+      module Step
+        class SpecialTrack < Engine::Step::SpecialTrack
+          def process_lay_tile(action)
+            super
+
+            company = action.entity
+            ability = @game.abilities(company, :tile_lay)
+            company.remove_ability(ability) unless ability.laid_hexes.none?
```
tobymao, line 15: do a find and replace all

lib/engine/game/g_18_bf/entities.rb
ollybh: Changed `find_index` to `index`.

lib/engine/game/g_18_bf/map.rb
ollybh: D'oh. Changed to `empty?`.

lib/engine/game/g_18_bf/map.rb
ollybh: Changed.

lib/engine/game/g_18_bf/step/special_track.rb
ollybh: Changed.

PR #8892, opened 2023-02-20
[1850] fix mesabi token buying when not connected

### LINKED ISSUE #8877: [1850] Buying Mesabi Tokens
https://18xx.games/game/114095
In game ID 114095 the RI was able to buy a Mesabi token when there was no connection.

bentziaxl:
fix #8877 8877

### DIFF
--- a/lib/engine/game/g_1850/step/track.rb
+++ b/lib/engine/game/g_1850/step/track.rb
@@ -71,7 +71,7 @@ module Engine
 
           def hex_neighbors(entity, hex)
             connected = super
-            @game.clear_token_graph_for_entity(entity) if entity.tokens.none? {|t| t.city}
+            @game.clear_token_graph_for_entity(entity) if entity.tokens.none?(&:city)
             connected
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1850/step/track.rb
```diff
@@ -69,12 +69,20 @@ def buy_mesabi_token(corporation)
             corporation.mesabi_token = true
           end
 
+          def hex_neighbors(entity, hex)
+            connected = super
+            @game.clear_token_graph_for_entity(entity)
+            connected
```
bentziaxl, line 75: I had to add graph clearing as i ran into an issue:
The hex neighbors check caches the graph when it runs, however it seem to happen prior to a corporation placing its home token when floating, resulting in the corp not being able to lay a tile.
This seems like a bad idea, as it will mean often refreshes of the graph, and bad performance.
Is there a better place to clear the graph? maybe right after home token placement?

PR #8905, opened 2023-02-25
[1862] fix presidency check when selling shares

### LINKED ISSUE #8901: 1862 President sale of share
In [game 113371](https://18xx.games/game/113371?action=435), I was President of the L&H in SR4.  It was valued at 220.  I sold a share, and the value did not drop.  My understanding is that when the President sells shares, the price drops regardless of the zone.

https://gist.github.com/benjaminxscott/341a7dac1792d3abab743a5677964399

benjaminxscott:
fixes #8901

### DIFF
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -1196,7 +1196,7 @@ module Engine
 
           @share_pool.sell_shares(bundle, allow_president_change: allow_president_change, swap: swap)
           num_shares = bundle.num_shares
-          unless corporation.owner == bundle.owner
+          unless corporation.owner == bundle.president
             num_shares -= 1 if corporation.share_price.type == :ignore_one_sale
             num_shares -= 2 if corporation.share_price.type == :ignore_two_sales
           end


HogueEric:
Thank you.

roseundy:
@benjaminxscott 

I'm just catching up on this. Can you explain why bundle.owner doesn't work here?

lib/engine/game/g_1862/game.rb
```diff
@@ -1196,7 +1196,7 @@ def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil
 
           @share_pool.sell_shares(bundle, allow_president_change: allow_president_change, swap: swap)
           num_shares = bundle.num_shares
-          unless corporation.owner == bundle.owner
+          unless corporation.owner == bundle.president
```
roseundy, line 1199: What am I missing? Isn't bundle.president defined to be corporation.owner? Won't this always be true?

roseundy:
Can we please revert this? This is not the correct fix for the bug. This will break games.

lib/engine/game/g_1862/game.rb
benjaminxscott: bundle.owner points to the sharepool object that the bundle is in, afaik

roseundy: My point is that bundle.president always points to corporation.owner, regardless of who owns the bundle. This means that the price will always drop even if it's not the president selling the shares.

The bug is that I don't make this check until after calling sell_shares, and thus the price never drops for one or two shares.

benjaminxscott: Ah I see what you're saying

Further context to my above comment which I think was part of my confusion, `bundle.owner` in the bugged implementation returned a `SharePool` object after `sell_shares`, so the comparison was silently failing

https://github.com/tobymao/18xx/issues/8916

PR #8907, opened 2023-02-25
Harzbahn hex variant

### LINKED ISSUE #8837: [Harzbahn 1873] Provide 100 M cost for hex F7 as an option
On the physical game map for the AA 2nd Edition, hex F7 costs 100 M to build, but 18XX.games shows it costing 50 M.

I would be great for players of that version if the 100 M cost were provided as an option.

philcampeau:
Fixes #8837

Adds variant option to make F7 terrain cost 100, as in the AAG printing and approved by the game designer on bug 8837 comments

### DIFF
--- a/lib/engine/game/g_1873/game.rb
+++ b/lib/engine/game/g_1873/game.rb
@@ -234,6 +234,13 @@ module Engine
           5 => 200,
         }.freeze
 
+        OPTION_REMOVE_HEXES = ['F7'].freeze
+        OPTION_ADD_HEXES = {
+          ['F7'] => 'city=revenue:20;city=revenue:20;path=a:1,b:_0,track:narrow;'\
+                    'path=a:3,b:_1,track:narrow;upgrade=cost:50,terrain:mountain;'\
+                    'icon=image:1873/11_open,sticky:1,large:1',
+        }.freeze
+
         def location_name(coord)
           @location_names ||= game_location_names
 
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1873/game.rb
```diff
@@ -3237,6 +3111,26 @@ def available_programmed_actions
...
+            new_map = self.class::HEXES[color].transform_keys do |coords|
+              coords - OPTION_REMOVE_HEXES
+            end
+            OPTION_ADD_HEXES.each { |coords, tile_str| new_map[coords] = tile_str } if color == :green
+
+            new_hexes[color] = new_map
+          end
+
+          new_hexes
+        end
```
roseundy, line 3133: Why are you doing it this way? Did you see how I changed the map for 2P 18Mag?

roseundy: Rather than swapping out the standard F7 hex with the optional F7 hex, wouldn't it be easier to leave the F7 hex out at first and then just add one or the other? It's why I created game_hexes as a method rather than using the constant HEXES.

lib/engine/game/g_1873/game.rb
```diff
@@ -2877,277 +2884,144 @@ def game_trains
...
-              ] => 'path=a:1,b:3;path=a:5,b:3',
-              %w[
-                H1
-              ] => 'path=a:5,b:3,track:narrow',
-              %w[
-                J5
-              ] => 'path=a:2,b:4',
-            },
-          }
-        end
```
roseundy, line 3150: What was wrong with game_hexes?

philcampeau:
I was curious why you did it the way you did, actually.

You declare the game_hexes variable, but that variable is never called anywhere else in the code. Why not just use HEXES, the standard across the site? 

I used the code in your first question because I felt that 8 lines of code stripping one hex away and replacing it with another was a more elegant solution.

roseundy:
game_hexes is definitely being called, otherwise the game wouldn't work at all. I put the call to it in game/base.rb.

For 18Mag, it was the easiest way to swap one map for another. This is because HEXES is a constant and can't be modified and I needed to change the hexes based on a game option. I just carried it over in Harzbahn because it was the next game I implemented.

PR #8920, opened 2023-02-26
[18mag] add module 3 - new major CIWL

bentziaxl:
added module 3 of the new 18mag expansion. 
New corporation named CIWL, that grants a bonus for a train running red to red.
If both trains run red to red, the bonus counts twice (clarification via email from lonny)
bonus is paid to both corp and private


roseundy:
Looks reasonable to me.

PR #8921, opened 2023-02-26
Create map & entities for titles

philcampeau:
Creates map.rb and entities.rb for games which didn't have them already, and separates out the relevant code into those files.


philcampeau:
I did this as a series of separate commits, one per game, to keep the changes clear. Each iteration was tested with rack rake, rubocop, and setting up the respective game and ensuring it loads and runs correctly.

tobymao:
do one at a time

PR #8940, opened 2023-02-27
[18Ireland] Implements the official larger bank variant

philcampeau:
Implements the official larger bank variant from the AAG rulebook (5k bank instead of 4k)


philcampeau:
image

PR #8946, opened 2023-02-27
[1858] Stock round: exchange and tokening steps

ollybh:
This is the next batch of code for 1858. This is the penultimate chuck of game logic, after this there is just the private closure round to be implemented.

This is the implementation of the stock round exchange and tokening steps.

Private railway companies in 1858 can be exchanged for shares in two different ways:

1. A private with a city in one of its home hexes can be used to start a    new corporation. The par price is taken from the private's face value, with the player paying for half of the president's certificate. The corporation starts with a token in the privates home city (if there are two cities in the private's home hexes then the player gets to choose which one to start in). The corporation becomes the owner of the private company.
2. A private can be exchanged for a treasury share from an existing corporation, if there is a valid route from one of the corporation's tokens to any of the privates home hexes. The corporation may place a token in a city in one of the private's home hexes. The corporation becomes the owner of the private company.

Any privates gained by corporations in this way will close at the end of the corporation's dividend step in the next operating round. The corporation will receive the private's face value as cash to its treasury.

Later in the game, a corporation may be started by simply setting a par price and purchasing the president's certificate. In this case the corporation may start in any open token slot, and pays twice the city's revenue to do this.

### Hacks

There's one place this commit touches common code. This is in `Engine::Part::City.exchange_token`. There are three cities in 1858 which each have two private companies that have a reservation ability for a token slot. However there is only a single slot in these cities before they are upgraded to green. This change to `exchange_token` catches the case where the reservation ability for the non-existent second slot is being used, and switches the two reservations around.

### DIFF
--- a/lib/engine/game/g_1858/game.rb
+++ b/lib/engine/game/g_1858/game.rb
@@ -515,7 +515,7 @@ module Engine
         def delete_icons(company)
           icon_name = company.sym.delete('&')
           @hexes.each do |hex|
-            hex.tile.icons.delete_if { |icon| icon.name == icon_name }
+            hex.tile.icons.reject! { |icon| icon.name == icon_name }
           end
         end
 
--- a/lib/engine/game/g_1858/game.rb
+++ b/lib/engine/game/g_1858/game.rb
@@ -171,13 +171,17 @@ module Engine
             quick_start
             operating_round(1)
           else
+            # The initial stock round isn't *quite* a normal stock round,
+            # you cannot start public companies in the first stock round.
+            # This difference is handled in exchange_corporations().
             stock_round
           end
         end
 
         def stock_round
           Engine::Round::Stock.new(self, [
-            Engine::Step::HomeToken,
+            G1858::Step::Exchange,
+            G1858::Step::HomeToken,
             G1858::Step::BuySellParShares,
           ])
         end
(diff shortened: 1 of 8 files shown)


lib/engine/game/g_1858/game.rb
```diff
@@ -397,6 +511,48 @@ def payout_companies
           super(ignore: exchanged_companies.map(&:id))
         end
 
+        # Removes all of the icons on the map for a private railway company.
+        def delete_icons(company)
+          icon_name = company.sym.delete('&')
+          @hexes.each do |hex|
+            hex.tile.icons.delete_if { |icon| icon.name == icon_name }
```
tobymao, line 518: always use regex

ollybh: @tobymao  I don't understand what you mean here. Use a regexp for testing the strings are the same? Something like:

```
rx = /^#{company.sym.delete('&')}$/
@hexes.each do |hex|
  hex.tile.icons.delete_if { |icon| rx.match?(icon.name) }
end
```
Is there an issue using `==` to test for string equality? AFAIK that works in both Ruby and Javascript.

tobymao: reject… sorry typo

ollybh: Ah, that makes _much_ more sense. I've changed the `Array.delete_if` call to `Array.reject!`.

PR #8947, opened 2023-02-28
[1868WY] Boomtown/Boom City tiles

michaeljb:
More progress for #5011

### Implementation Notes

* **Explanation of Change**

In 1868 Wyoming, coal and oil companies add Development Tokens (DTs) to the board in order to upgrade Boomtowns into Boom Cities, and as phases change, the DTs will rust and Boom Cities may BUST back into Boomtowns or ghost towns.

The physical tiles are double sided and show three revenues:
* BUSTed revenue, $10 - when a new phase rusts old DTs and now there are 2 or less tokens on or adjacent to this tile
* standard city revenue - with 3 DTs on or adjacent to this tile, it makes what a normal city of that color makes
* +$10 - 4+ DTs makes the city worth an extra $10

On the physical tiles, the revenues are colored black/tile color/white, so that's what they look like in the tile manifest here. When actually played on the board, the active revenue is white, and the inactive ones are grayed out.

* **Any Assumptions / Hacks**

    * Although this game is not the first to use unlimited tiles, I re-implemented `add_extra_tile` in this game class because I believe it is the first to use unlimited tiles which are double-sided
    * Over in `g_1868_wy/map.rb`, the white background is defined with `#FFFFFF` instead of the `white` string, since that color usually isn't white (I should have named it something else back at the beginning of development on this site 🙃):

    https://github.com/tobymao/18xx/blob/8fc198d7ab4f720326002e1539d74fc3b2c4a872/lib/engine/game/g_1868_wy/map.rb#L224-L228

    * Also worth noting, not necessarily a hack, but a change to core code is to add a writer method for `Tile#revenue_to_render` to dynamically change the revenue to render, so that users don't have to count up the DTs to see what their Boom City will make.

* **Screenshots**

yellows:

Screenshot 2023-02-27 211705

greens:

Screenshot 2023-02-27 211757

206b was laid, then boomed to a Boom City, but now some of the surrounding DTs have been removed, so it is only making $10 and will BUST to a ghost town in the next BUST round:

Screenshot 2023-02-27 214258


michaeljb:
Just realized `update_opposites` is now duplicated across 18Mag, 18Carolinas, and now 1868 Wyoming, will reopen after I can fix that.

PR #8954, opened 2023-03-01
[18mag] module 1 - supporters

bentziaxl:
Implement module 1 of the new 18mag expansion.
6 supporters (implemented as privates), that are drafted like the minors and majors
adding different abilities that can be used by minors/majors once a round.

### DIFF
--- a/lib/engine/game/g_18_mag/step/route.rb
+++ b/lib/engine/game/g_18_mag/step/route.rb
@@ -35,8 +35,16 @@ module Engine
             raise GameError, 'Must use all purchased rail-car benefits' unless @game.all_railcars_used?(action.routes)
 
             ciwl_income(action.routes) if @game.new_major?
-
             super
+            deconvert_plus_one_train if @game.supporters? && @round.original_train
+          end
+
+          def deconvert_plus_one_train
+            @round.ma_plus_train.name = @round.original_train.name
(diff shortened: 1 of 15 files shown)


lib/engine/game/g_18_mag/step/route.rb
```diff
@@ -35,8 +35,16 @@ def process_run_routes(action)
...
             super
+            deconvert_plus_one_train if @game.supporters? && @round.original_train
+          end
+
+          def deconvert_plus_one_train
+            @round.ma_plus_train.name = @round.original_train.name
+            @round.ma_plus_train.distance = @round.original_train.distance
+
+            @round.original_train = nil
+            @round.ma_plus_train = nil
```
roseundy, line 47: What is this doing? Why are you assigning to @round.ma_plus_train and then making it nil?

bentziaxl: see here: https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1822/step/route.rb#L50

I'm modifying the distance of the train that was converted into +1 town range back to it's original values.
That's why i'm change .name and . distance, then setting it to nil.
Similar code to pullman attachment and removal in 1822

tobymao:
@roseundy is this good to go?

roseundy:
Good to go.

PR #8959, opened 2023-03-02
Refactor stock movement into separate classes per market type

crericha:
- Introduce stock movement class for each type of stock market
- Remove zigzag market workarounds from being based on a 1d market, in 1860, 1862, and 18CZ
- Some methods took coordinates ([r, c]) and others r, c. Standardize on coordinates

### DIFF
--- a/lib/engine/game/g_1860/step/dividend.rb
+++ b/lib/engine/game/g_1860/step/dividend.rb
@@ -63,18 +63,18 @@ module Engine
             if revenue.positive?
               curr_price = entity.share_price.price
               if revenue >= curr_price && revenue < 2 * curr_price
-                { share_direction: :right, share_times: 2 }
+                { share_direction: :right, share_times: 1 }
               elsif revenue >= 2 * curr_price && revenue < 3 * curr_price
-                { share_direction: :right, share_times: 4 }
+                { share_direction: :right, share_times: 2 }
               elsif revenue >= 3 * curr_price && revenue < 4 * curr_price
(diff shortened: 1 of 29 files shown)


Zwergenpunk (approved):
lgtm

lib/engine/game/g_1860/step/dividend.rb
```diff
@@ -101,10 +101,6 @@ def payout_shares(entity, revenue)
             @log << "#{entity.name} pays out #{@game.format_currency(revenue)} = "\
                     "#{@game.format_currency(per_share)} (#{receivers})"
           end
-
-          def movement_str(times, dir)
-            "#{times / 2} #{dir}"
-          end
```
roseundy, line 107: Where did this go? I'm away from my dev machine so I can't tell where this was called.

Zwergenpunk: this is not longer needed, because the stepsize is now calculated correctly (doubling stepsize is not needed anymore)

roseundy: Nevermind. Found the call.

I'm curious what this will look like during the dividend step. Will it now show as "Right/Left" or "Up/Down"?

crericha: Dividend steps use left right movement. Only stock selling or floating at the end of the SR use up/down. 

I also loaded a 62 and CZ and verified the view for dividend step looked correct.

tobymao (approved):

PR #8961, opened 2023-03-02
[18OE] UK/FR Basic Map

Kalnoch:
Initial files to show the basic map for the 18OE UK/France scenario. It contains all the regular map tiles, the offboards, and the 0 cost ferries. It is missing the seazones and ports.

Also added basic market, entities, and tiles. Does not have OE specific tiles yet.

image


michaeljb:
If the 18OE scenarios use just a subset of the map, but companies have the same home locations and so on, it might be better to refer to the hexes by where they are on the full map.

I'm not familiar with 18OE but I can see from this image from the Kickstarter page that there are hexes in Spain west of the UK and France hexes, so presumably on the full map the column numbers for this section will be higher:

18OE map

Kalnoch:
> If the 18OE scenarios use just a subset of the map, but companies have the same home locations and so on, it might be better to refer to the hexes by where they are on the full map.
> 
> I'm not familiar with 18OE but I can see from this image from the Kickstarter page that there are hexes in Spain west of the UK and France hexes, so presumably on the full map the column numbers for this section will be higher:

Yeah, I shifted everything west by 12 hexes as I thought having a map not starting at A:1 would be weird, though the ability to reuse as much code as possible is probably preferable as yes, all the regionals always start in the same spots with only one or two exceptions in the smaller scenarios.

michaeljb (approved):

PR #8967, opened 2023-03-04
[1858] Private closure round implementation

ollybh:
## 1858 private closure round

The private closure round occurs at the end of the operating round in which the first 5E or 4M train triggers has been bought.

In this round any private railways that are still in player hands close, either being exchanged for a share in a connected corporation, or being sold back to the bank for their face value. If a private railway is exchanged for a treasury share then the corporation gets an opportunity to place a free token in a city in the private's home hexes.

### Implementation notes

This does not touch any shared code, everything is in the `lib/engine/game/g_1858` directory.

The closure round is implemented as a subclass of `Engine::Round::Operating`. The entities that act are the minor parts of the private railways.

At the start of the round any still unpurchased privates are closed, as is the H&G private which does not have a minor entity.

Each minor gets a choice step, where the player can choose between shares in connected corporations, or sell the private to the bank (the 'pass' option).

If the private is exchanged for a treasury share, and there is a city in one of the private's home hexes, and the corporation has a token available, then a token will be pushed onto the `@round.pending_tokens` array, causing the `HomeToken` step to occur, giving the corporation the opportunity to place a free token in the city. The reservation ability is moved from the minor to the corporation to ensure that it is still active when the HomeToken step occurs, this is to catch an unlikely corner case where the corporation could be blocked out of a city if a private that closes later in the round also has a reservation on the same slot (issue ollybh/18xx#47).


lib/engine/game/g_1858/step/private_closure.rb
```diff
@@ -0,0 +1,111 @@
...
+            else
+              company.owner = @game.bank
+              player.companies.delete(company)
+            end
+            @game.close_private(company)
+          end
+
+          def exchange_corporations(minor)
+            @game.corporations.select do |corporation|
+              @game.corporation_private_connected?(corporation, minor) &&
```
michaeljb, line 102: I see this method relies on the graph, is it possible to call this only when an action is executed via the UI, instead of when determining available actions? Then it could be skipped when `@game.loading` is true so that the engine could do less graph work when loading the game.

I see it is used for some other actions, so that may be something to tackle in a separate PR. (and apologies if this has been discussed elsewhere and I missed it)

michaeljb: 1. Checking for the corporation having shares first sounds like a good change, anything that can short-circuit graph checks is certainly useful. I wonder if something could be done with auto actions on track lay that look for and cache connections, and then in the private closure step you could just check that cache, but that approach does sound quite a bit more complicated.

2. It sounds like this won't be invoked enough to cause any real issues.

3. I did forget about the fact that normal Track steps already do check `can_lay_tile?` all the time anyway, so no worries there.

michaeljb (approved):

lib/engine/game/g_1858/step/private_closure.rb
michaeljb: Scratch what I said for 1, `auto_actions` should be used so that an initial check for connected corporations and privates is done in the browser. It can return `Pass` when there are no connections. Here are a couple examples:

* https://github.com/tobymao/18xx/blob/d2cc06960023d3dc478da11e520086354cd5daf5/lib/engine/game/g_1822/step/minor_acquisition.rb#L34
* https://github.com/tobymao/18xx/blob/08c4329acc894c4094777a5ec296913420d5898f/lib/engine/game/g_18_ny/step/merge_with_nyc.rb#L18

michaeljb (changes requested):
See the auto_actions comment

lib/engine/game/g_1858/step/private_closure.rb
ollybh: OK, I'll look into auto_actions and make this change.

I'll also see if the graph calls in the track and buy/sell shares steps can be removed in the same way, and submit separate pull requests for those.

ollybh: I've made this change, moving the call to `exchange_corporations` out from `actions` into `auto_actions`.

michaeljb (approved):

PR #8969, opened 2023-03-04
[1822PNW,MX,CA] fix P16 not counting as a cert

### LINKED ISSUE #8965: [1822PNW] P-16 doesn't count as a certificate
https://18xx.games/game/110112

P-16 is not being counted as a certificate.  I'm sure this is because P-16 in base 1822 is the Tax Haven, which doesn't count as a certificate.

It appears that the code for it was not updated when PNW was forked from the base code.

benjaminxscott:
Fixes #8965 

### Implementation Notes

* **Explanation of Change**

Override a base 1822 function that handles the way that P16 `Tax Haven` in the base game does not count against certificate limit

* **Screenshots**
Screenshot 2023-03-04 9 17 56 AM

* **Any Assumptions / Hacks**

This s.b. generalized in 1822 base code so future games don't have to know about this hack-around

### DIFF
--- a/lib/engine/game/g_1822_ca/game.rb
+++ b/lib/engine/game/g_1822_ca/game.rb
@@ -246,9 +246,8 @@ module Engine
 
         # Stubbed out because this game doesn't it, but base 22 does
         def company_tax_haven_bundle(choice); end
-
-        # Stubbed out because this game doesn't it, but base 22 does
         def company_tax_haven_payout(entity, per_share); end
+        def num_certs_modification(_entity) = 0
       end
     end
   end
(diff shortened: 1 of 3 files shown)


outsidePasser:
Thanks.  There's likely more issues like this because of how 22 checks for specific privates.  I don't have a good general way to deal with this yet.  It's on my radar, but kind of in the back.

michaeljb:
> Thanks. There's likely more issues like this because of how 22 checks for specific privates. I don't have a good general way to deal with this yet. It's on my radar, but kind of in the back.

Maybe there's a way for the `cert_size` idea on Shares to be applied to Companies?

https://github.com/tobymao/18xx/blob/0236a8ff5e4139b119ca08419e39554966847d0d/lib/engine/share.rb#L15

michaeljb (approved):

PR #8970, opened 2023-03-05
Fix `add_extra_tile` to work with double-sided tiles

michaeljb:
Missed this in #8951 while splitting up work on 1868 Wyoming (#5011)

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1794,6 +1794,12 @@ module Engine
         @cities.concat(extra_cities)
         extra_cities.each { |c| @_cities[c.id] = c }
 
+        if (opp_name = tile.opposite&.name) && !new_tile.opposite
+          opp_tile = tile_by_id("#{opp_name}-#{new_tile.index}") || add_extra_tile(tile_by_id("#{opp_name}-0"))
+          opp_tile.opposite = new_tile
+          new_tile.opposite = opp_tile
+        end
+
         new_tile
       end
 


roseundy:
How did you catch this? Does this only affect 18Wyoming?

lib/engine/game/base.rb
```diff
@@ -1794,6 +1794,12 @@ def add_extra_tile(tile)
         @cities.concat(extra_cities)
         extra_cities.each { |c| @_cities[c.id] = c }
 
+        if (opp_name = tile.opposite&.name) && !new_tile.opposite
+          opp_tile = tile_by_id("#{opp_name}-#{new_tile.index}") || add_extra_tile(tile_by_id("#{opp_name}-0"))
+          opp_tile.opposite = new_tile
+          new_tile.opposite = opp_tile
+        end
+
```
roseundy, line 1802: Isn't this logic already in update_tile_lists! ?

michaeljb:
> How did you catch this? Does this only affect 18Wyoming?

While Wyoming is the only game with unlimited double-sided tiles, yes.

I caught it from looking at my code which wasn’t merged into master yet—a lot of the work for this game was done a while ago and I’ve been going back to make manageable PRs recently. I saw this bit of code in an overridden method that I missed when doing #8951. I haven’t gone back to test it to see if it is necessary, I expect to have time for that this weekend.

lib/engine/game/base.rb
michaeljb: Yes it is, it doesn't need to be here, but this comment lead to me finding a problem with how I implemented it there. I'll close this and open a new PR.

michaeljb:
See #8996

PR #9000, opened 2023-03-12
[1880] allow cross buy of 2p

### LINKED ISSUE #8966: [1880] Reclaimed 2 should be cross-buyable
@bentziaxl I couldn't find anything in the rules that says this is not allowed. Did I just miss it or was the current implementation based on the assumption that you couldn't, like in 48?

bentziaxl:
Fixes #8966

### DIFF
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -66,6 +66,7 @@ module Engine
           def buyable_trains(entity)
             trains_to_buy = super
             trains_to_buy.reject! { |t| t.name == '2P' } unless can_buy_restored2?(entity)
+            trains_to_buy.reject! { |t| t.name == '2P' && t.from_depot? } unless can_buy_restored2_from_depot?(entity)
             # Can't buy trains from other corporations until train 3
             return trains_to_buy if @game.can_cross_buy?
 
@@ -80,6 +81,10 @@ module Engine
             @game.phase.available?(restored2_train&.available_on) &&
             !owns_restored2?(entity) &&
             @corporations_sold.empty?
+          end
+
+          def can_buy_restored2_from_depot?(entity)
+            can_buy_restored2?(entity) && entity.cash >= restored2_train.price
           end
 
           def restored2_train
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1880/step/buy_train.rb
```diff
@@ -79,8 +79,7 @@ def owns_restored2?(entity)
           def can_buy_restored2?(entity)
             @game.phase.available?(restored2_train&.available_on) &&
             !owns_restored2?(entity) &&
-            @corporations_sold.empty? &&
-            entity.cash >= restored2_train.price
+            @corporations_sold.empty?
```
crericha, line 82: Will this change affect EMR? Can you check if this change lets you buy a 2P with player cash?

crericha: My hunch is we'll need to remove depot owned 2Ps if the corp has less cash than they cost.

bentziaxl: i'm not clear about the rules:
1- can you cross buy a 2p when in EMR? (yes?)
2- can you sell shares in EMR, buy a 2p from the bank, then sell more shares to buy a regular train? 
3- can you sell shares in EMR then cross buy? (no?)

crericha: Under forced train purchase
> The train acquired by Forced Purchase may not be a restored 2-train. 

No other special rules around restored 2-train, except that it doesn't count towards your train requirement

crericha: So
1 - yes, provided it is with company cash
2 - no
3 - no, not for any train. Must be from the depot

If a company has no trains, but 250 in cash, it can buy a reclaimed 2-train and then EMR a train from the depot.

bentziaxl: much clearer now, yeah, i'll need to remove 2p from depot if low on cash

bentziaxl: Fixed. Maybe there's a better/cleaner way to code it?
I can wrap my head around the if/unless conditionals to make it work in one block

crericha (approved):

PR #9006, opened 2023-03-13
[18mag] fix new supporter

bentziaxl:
minor fix to existing supporter. Change tiles, add that special is false

Tile payment was wrong, and didn't pay SIK company if supporter was used. 
I extracted the method into a PayTile module, that is used by both track (regular lay) and special_track (lay via supporter)


tobymao:
@roseundy

roseundy:
lgtm

roseundy (approved):
Seems OK

PR #9007, opened 2023-03-13
Fixes phase issue in 1856

### LINKED ISSUE #8993: [1856] 3rd values for off-board location not taken into account in late phases
**What's your Game ID?**

https://18xx.games/game/111352?action=1057

Detroit-Winsor should be worth 60 and not 50.

**Which Browser / Platform?**
- Browser Version
  - Firefox 110.0.1

philcampeau:
Fixes #8993

My mistake, when I was doing a previous modification I must have accidentally removed "black" from the tiles section of Phase D. This corrects it.


michaeljb:
I think this won't actually require pins since route values are stored in the action, so when validating the routes won't be re-computed

PR #9011, opened 2023-03-14
[1817] Volatility-variant private "Inventor" pays too much on 6 train purchase

### LINKED ISSUE #8988: [1817] Volatility-variant private "Inventor" pays too much on 6 train purchase
**What's your Game ID?**
https://18xx.games/game/106175
https://18xx.games/game/99154

https://18xx.games/game/106175?action=1524
https://18xx.games/game/99154?action=1119

https://gist.github.com/benjaminxscott/d313d464aa2505f14915e475ffc20876

The inventor company should only pay out $60 for the first 6 train purchase/export. Instead it pays $60 and then makes another payment of $30.

benjaminxscott:
Fixes #8988

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

- Sets the `rusted` flag on trains before `remove_train()` is called, as within that function it was not possible to check the `rusted` state otherwise. This behavior was causing a bug as described here https://github.com/tobymao/18xx/issues/8988#issuecomment-1467477512

-  Also Removes an un-necessary "2+ trains provide zero revenue" definition from the lookup hash, since the calling function returns zero anyway if the train type is not in the hash

* **Screenshots**
Screenshot 2023-03-14 12 06 21 AM

* **Any Assumptions / Hacks**

We assume that `remove_train` will operate correctly if the train is already marked as rusted

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1950,9 +1950,9 @@ module Engine
       end
 
       def rust(train)
+        train.rusted = true
         remove_train(train)
         train.owner = nil
-        train.rusted = true
       end
 
       def num_corp_trains(entity)
--- a/lib/engine/game/g_1817/game.rb
+++ b/lib/engine/game/g_1817/game.rb
@@ -871,15 +871,8 @@ module Engine
         end
 
         def remove_train(train)
-          from_depot = train.from_depot?
+          inventor_payout(train)
           super
-
-          return unless inventor_private&.owner&.corporation?
-          return if train.rusted || !from_depot || @depot.trains.find { |t| t.name == train.name } != train
-          return unless (payout = inventor_payout(train)).positive?
-
-          @log << "#{inventor_private.owner.name} collects #{format_currency(payout)} from #{inventor_private.name}"
-          @bank.spend(payout, inventor_private.owner)
         end
 
         def loan_shark_private
@@ -925,8 +918,13 @@ module Engine
         end
 
         def inventor_payout(train)
+          return unless inventor_private&.owner&.corporation?
+
           @payouts ||= { '2' => 20, '3' => 30, '4' => 40, '5' => 50, '6' => 60, '7' => 70, '8' => 80 }
-          @payouts[train.name] || 0
+          return unless (payout = @payouts.delete(train.name))
+
+          @log << "#{inventor_private.owner.name} collects #{format_currency(payout)} from #{inventor_private.name}"
+          @bank.spend(payout, inventor_private.owner)
         end
 
         private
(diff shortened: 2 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -1950,9 +1950,9 @@ def remove_train(train)
       end
 
       def rust(train)
+        train.rusted = true
         remove_train(train)
         train.owner = nil
-        train.rusted = true
```
crericha, line 1955: This is probably okay. Did you check all games that implement remove_train to verify this change doesn't mess them up?

lib/engine/game/g_1817/game.rb
```diff
@@ -873,8 +873,9 @@ def rust(train)
         def remove_train(train)
           from_depot = train.from_depot?
           super
+
           return unless inventor_private&.owner&.corporation?
-          return if !from_depot || @depot.trains.find { |t| t.name == train.name } != train
+          return if train.rusted || !from_depot || @depot.trains.find { |t| t.name == train.name } != train
```
crericha, line 878: I don't remember what this logic is attempting to do and I think it is hard to follow. I think we should go a simpler way of removing the value from @payouts after it pays out. That way, it'll only pay out once, like it should, and we don't have to track any additional state.

To do this, move the log message  and the return unless inventor_private&.owner&.corporation? to inventor_payout  and handle everything there. Then the return if rusted, !depot, etc. logic can be removed.

lib/engine/game/base.rb
benjaminxscott: I looked at '28 and NY which seem to be the only games that shadow this function, and `rusted` doesn't seem to be a flag that they are checking or influencing their logic

benjaminxscott: OK with keeping this change to base code even if the bugfix implementation doesn't require it?

lib/engine/game/g_1817/game.rb
benjaminxscott: Ah, made a commit that should implement this approach, check it out [5db9212](https://github.com/tobymao/18xx/pull/9011/commits/5db92125ab6b40c5c2969d8556d378e7330e3aec)

lib/engine/game/base.rb
crericha: remove_train is the method in question. It is overridden by 17, 22, 28, 77, and NY.

lib/engine/game/g_1817/game.rb
crericha: Looks good, I think.

lib/engine/game/base.rb
benjaminxscott: Ah yes I checked 22 as well which seemed to be fine

'17 is this one so it should work

77 I think is a different function? `def remove_trainless`

benjaminxscott:
@crericha OK to merge?

crericha:
Yeah, looks good to me.

PR #9027, opened 2023-03-16
[18cuba] initial map commit

bentziaxl:


lib/engine/game/g_18_cuba/game.rb
```diff
@@ -0,0 +1,216 @@
...
+          super
+          @tile_groups = init_tile_groups
+          update_opposites
+        end
+
+        def init_tile_groups
+          self.class::TILE_GROUPS
+        end
+
+        def update_opposites
```
michaeljb, line 195: You should remove this method, and add `include DoubleSidedTiles` to use the methods in [double_sided_tiles.rb](https://github.com/tobymao/18xx/blob/087b3018ee5b8e844a9518c1dc5401637d2418ab/lib/engine/game/double_sided_tiles.rb#L8)

tobymao (approved):

PR #9029, opened 2023-03-17
Raise an error if the tile being laid is already on the board

### LINKED ISSUE #9026: 18Mag Minor 12 cannot run through or upgrade H25
https://18xx.games/game/115149

michaeljb:
This bug was introduced by #8951, and fixed by #8996, but in between those two, games could be broken and not found when validating #8996 since the tile mismatch could not have any real impact on the game for many actions--in the case of game 115149, the game first misused tiles at action 630 but was not visibly broken until action 1766.

Closes #9026

----

10 of 11 active 18Mag games are impacted; pinning would prevent them from hitting this new exception but cannot prevent encountering the broken behavior seen in #9026.  
Here are the affected games:

`[111821, 113092, 113389, 114853, 115023, 115024, 115139, 115149, 115319, 116143]`

I've pinned them all. Many have triggered the end and may be able to finish, but could still encounter the issues caused by #8951.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2083,6 +2083,9 @@ module Engine
 
       def update_tile_lists(tile, old_tile)
         add_extra_tile(tile) if tile.unlimited
+
+        raise GameError, "Cannot lay tile #{tile.id}; it is already on hex #{tile.hex.id}" if tile.hex
+
         @tiles.delete(tile)
         @tiles << old_tile unless old_tile.preprinted
       end
--- a/lib/engine/game/double_sided_tiles.rb
+++ b/lib/engine/game/double_sided_tiles.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../game_error'
+
 # This module is used by games which double sided tiles, providing methods for
 # initializing double-sided tiles and keeping tile lists updated appropriately.
 module DoubleSidedTiles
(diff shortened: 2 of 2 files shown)


bentziaxl:
I'm not sure how that fixes the issue. 
In the game that i saw a corp tried to lay tile #9 (straight yellow), and the selector picked a straight yellow that was already laid somewhere else on the map in a previous turn. 
It sounds like this fix will raise an error, and not allow a legal placement.
What am i missing?

michaeljb:
After #8951, some tiles could end up in the available tile pool even if they shouldn't have been since the tile on their opposite side was used.

After #8996, that couldn't happen again, but any games where that situation had already happened couldn't be retroactively fixed.

This PR also can't retroactively fix those games, but if this error raising had been around prior to #8951, then as soon as a game tried to use one of those tiles it shouldn't, the error would have been raised and the game would not have actually entered a broken state.

tobymao (approved):
make sure to run through all games with this one

michaeljb:
> make sure to run through all games with this one

Sounds good, should be able to do that validate and deploy later this week

michaeljb:
Found 5 games broken by this: `[96788, 99175, 99503, 114466, 116812] `. The first four have already finished, the last was encountering another issue that stemmed from this tile-already-laid problem (#9042); I have undone that game to before the breaking action, as described in the comments on #9042.

PR #9039, opened 2023-03-19
[1880] fix 6e route issue

### LINKED ISSUE #9034: [1880] 6E train seem cannot run the northern route (the one with $50 bonus)
As title
please check in the last OR of BZU run in the hotseat file
[hs_qekzcoeg_1679043717.zip](https://github.com/tobymao/18xx/files/11010360/hs_qekzcoeg_1679043717.zip)

### LINKED ISSUE #9038: 1880:  A few minor display bugs
Some UI things I've noticed in my current game (116522):

1)  During the SCR par selection at action (https://18xx.games/game/116522?action=400) the game is telling me I can afford 11 shares:  :)

image

2) On the Info tab, the Train display has two errors:
  a) The "B permit" event is missing (in two places)
  b) The "40% float" event is on the 4's, NOT the 4+4's

image

bentziaxl:
fixes #9038: 
removes the extra float_40 event on 4+4.
sets min of shares purchasable and 10

fixes #9034

### DIFF
--- a/assets/app/view/game/par_chart.rb
+++ b/assets/app/view/game/par_chart.rb
@@ -39,7 +39,7 @@ module View
 
         cash = @current_entity.cash
         prices = @step.get_par_prices(@current_entity, @corporation_to_par)
-        prices.map! { |sp| "#{@game.format_currency(sp.price)} (#{(cash / sp.price.to_f).floor} shares)" }
+        prices.map! { |sp| "#{@game.format_currency(sp.price)} (#{[(cash / sp.price.to_f).floor, 10].min} shares)" }
         text = "#{@current_entity.name} can par #{@corporation_to_par.name} at #{prices.join(', ')}"
         h(:div, props, text)
       end
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -196,7 +196,7 @@ module Engine
                     price: 450,
                     rusts_on: '8E',
                     num: 5,
-                    events: [{ 'type' => 'float_40' }, { 'type' => 'permit_c' }],
+                    events: [{ 'type' => 'permit_c' }],
                   },
                   {
                     name: '6',
@@ -683,10 +683,15 @@ module Engine
 
             unless stops.find { |stop| stop.tokened_by?(route.corporation) }
               stops.pop
-              stops << sorted_stops.find { |stop| stop.tokened_by?(route.corporation) }
+              tokened_stop = sorted_stops.find { |stop| stop.tokened_by?(route.corporation) }
+              stops << tokened_stop if tokened_stop
             end
 
-            stops << sorted_stops.select { |stop| trans_siberian_hexes.include?(stop.hex) } if trans_siberian_bonus?(sorted_stops)
+            if trans_siberian_bonus?(sorted_stops)
+              stops.concat(sorted_stops.select do |stop|
+                             trans_siberian_hexes.include?(stop.hex)
+                           end)
+            end
 
             stops.uniq!
           end


assets/app/view/game/par_chart.rb
```diff
@@ -39,7 +39,7 @@ def par_choices
 
         cash = @current_entity.cash
         prices = @step.get_par_prices(@current_entity, @corporation_to_par)
-        prices.map! { |sp| "#{@game.format_currency(sp.price)} (#{(cash / sp.price.to_f).floor} shares)" }
+        prices.map! { |sp| "#{@game.format_currency(sp.price)} (#{[(cash / sp.price.to_f).floor, 10].min} shares)" }
         text = "#{@current_entity.name} can par #{@corporation_to_par.name} at #{prices.join(', ')}"
```
bentziaxl, line 43: this should probably be changed by code that returns the total number of shares a crop has. 
What's the easiest way to determine that?

lib/engine/game/base.rb
```diff
@@ -1377,12 +1377,12 @@ def compute_stops(route, train = nil)
           stops, revenue = visits.combination(num_stops.to_i).map do |stops|
             # Make sure this set of stops is legal
             # 1) At least one stop must have a token (if enabled)
-            next if train.requires_token && stops.none? { |stop| stop.tokened_by?(route.corporation) }
+            next if train.requires_token && stops.none? { |stop| stop&.tokened_by?(route.corporation) }
```
bentziaxl, line 1381: When starting a route from validvastock, stop is nil and this line throws an exception. No idea why it's nil

lib/engine/game/base.rb
```diff
@@ -1377,12 +1377,12 @@ def compute_stops(route, train = nil)
...
 
             # 2) We can't ask for more revenue centers of a type than are allowed
             types_used = Array.new(distance.size, 0) # how many slots of each row are filled
 
-            next unless stops.all? do |stop|
+            next unless stops.flatten.all? do |stop|
               row = distance.index.with_index do |h, i|
```
bentziaxl, line 1386: When doing a route from Russia (A3) and Vladivostock (A15), the stops variable for some reason combines them into an array of two stops, and stop.type fails. This is the cause of the error reported in the bug.
Flatten obviously works, but i don't know why it combines it into an array

assets/app/view/game/par_chart.rb
crericha: There isn't a great way. Maybe pass off of `@full_cap_event ? 10 : 5`

lib/engine/game/base.rb
crericha: Can you write a bug for this and reference that this is a workaround?

lib/engine/game/base.rb
```diff
@@ -1377,12 +1377,12 @@ def compute_stops(route, train = nil)
...
             # Make sure this set of stops is legal
             # 1) At least one stop must have a token (if enabled)
-            next if train.requires_token && stops.none? { |stop| stop.tokened_by?(route.corporation) }
+            next if train.requires_token && stops.none? { |stop| stop&.tokened_by?(route.corporation) }
 
             # 2) We can't ask for more revenue centers of a type than are allowed
             types_used = Array.new(distance.size, 0) # how many slots of each row are filled
 
-            next unless stops.all? do |stop|
+            next unless stops.flatten.all? do |stop|
```
crericha, line 1385: This bug is in revenue_stops().

`stops << sorted_stops.select { |stop| trans_siberian_hexes.include?(stop.hex) } if trans_siberian_bonus (sorted_stops)
`

should be

`stops.concat(sorted_stops.select { |stop| trans_siberian_hexes.include?(stop.hex) }) if trans_siberian_bonus?(sorted_stops)`

lib/engine/game/base.rb
crericha: See above comment

lib/engine/game/base.rb
bentziaxl: amazing, thanks for finding the source of the bug!

lib/engine/game/base.rb
bentziaxl: found the issue. see bellow

lib/engine/game/g_1880/game.rb
```diff
@@ -684,9 +684,14 @@ def revenue_stops(route)
             unless stops.find { |stop| stop.tokened_by?(route.corporation) }
               stops.pop
               stops << sorted_stops.find { |stop| stop.tokened_by?(route.corporation) }
+              stops = stops.compact
             end
```
bentziaxl, line 688: L686 sometimes adds nil to stops when none of the stations are tokened.
This means that if a user starts the route away from the corp's tokened spots they get errors until the route has a token in it.
compact fixes it. (Is there a one liner that does << unless nil ?)

lib/engine/game/base.rb
bentziaxl: not changing any base classes

assets/app/view/game/par_chart.rb
crericha: Did you want to fix this or good with always saying 10? I don't have a strong preference.

bentziaxl: leaving as is. If another game with par view and has 5 share corps will want to change it, they can.

PR #9051, opened 2023-03-23
[18OE] float minors, implement track rights

Kalnoch:
### Implementation Notes

Enabled floating of minors in the stock round and users cannot pass until they have floated all of their minors. Removes leftover regionals once cap is hit.

Sets minors treasury at auction time, capped at 180

Added in track rights logic to enforce minors/regionals to only be able to place tokens/track in their respective track rights zones. Minors choose their track rights zone from the available ones when they float.

Corrected half-pay to follow 18OE logic and set dividends to not move minor/regional stock price (since they technically don't exist on the market)

- [x]  Branch is derived from the latest master
- [x]  Code passes linter with docker compose exec rack rubocop -a
- [x]  Tests pass cleanly with docker compose exec rack rake


lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,107 @@
...
+            return [] unless entity == current_entity
+            return ['sell_shares'] if must_sell?(entity)
+
+            actions = []
+            actions << 'buy_shares' if can_buy_any?(entity)
+            actions << 'par' if can_ipo_any?(entity) || can_float_minor?(entity)
+            actions << 'buy_company' if !purchasable_companies(entity).empty? || !buyable_bank_owned_companies(entity).empty?
+            actions << 'sell_shares' if can_sell_any?(entity)
+
+            actions << 'pass' if !can_float_minor?(entity) && !bought? && actions.any?
```
tobymao, line 20: don't call any? on it's own intsead do !actions.empty? any? checks for nil and is more expensive

lib/engine/game/g_18_oe/step/dividend.rb
```diff
@@ -0,0 +1,41 @@
...
+            withheld = half_pay_withhold_amount(entity, revenue)
+            { corporation: withheld, per_share: payout_per_share(entity, revenue - withheld) }
+          end
+
+          def half_pay_withhold_amount(entity, revenue)
+            (revenue / 2 / entity.total_shares) * entity.total_shares
+          end
+
+          def share_price_change(entity, revenue = 0)
+            return {} if @game.minor_regional_order.include? entity
```
tobymao, line 24: always use parenthesis.

PR #9055, opened 2023-03-26
[18mag] fix +1 train supporter

### LINKED ISSUE #9045: 18Mag +1 engineer not working
https://18xx.games/game/116943
 
 doesn’t count the town when you use the power, does with autoroute but then there’s a “nil can’t be an integer” error?

bentziaxl:
fixes #9045

one of the supporters converts a train into +1, for example 2+1.
this can combine with the major ability that converts into plus train. (gc major)

Need special logic to handle
2+1 (supporter only)
2+2 (plus train conversion only)
2+3 (supporter and plus train)

added
1- train conversion and train distance use a city_distance_value  in case it's a gc train
2- compute stops needs to use the updated distance, previously used distance without GC. instead of copying the entire content of compute stops i pass a dummy duplicate train with the correct distance (open to other suggestions such as adding a param or copying the entire method)
3- modified route_distance to also work if just supporter is used without gc

### DIFF
--- a/lib/engine/game/g_18_mag/game.rb
+++ b/lib/engine/game/g_18_mag/game.rb
@@ -419,12 +419,15 @@ module Engine
         def buy_train(operator, train, price = nil)
           @train_bought = true if train.owner == @depot # No idea if this is what Lonny wants
           cost = price || train.price
+          full_cost = train.price if train.owner == @depot
           if price != :free && train.owner == @depot
             if multiplayer?
               corp = %w[2 4].include?(train.name) ? @ldsteg : @mavag
-              operator.spend(cost / 2, @bank)
-              operator.spend(cost / 2, corp)
-              @log << "#{corp.name} earns #{format_currency(cost / 2)}"
+              bank_cost = full_cost ? cost - (full_cost / 2) : cost / 2
+              corp_cost = full_cost ? full_cost / 2 : cost / 2
+              operator.spend(bank_cost, @bank)
+              operator.spend(corp_cost, corp)
+              @log << "#{corp.name} earns #{format_currency(corp_cost)}"
             else
               operator.spend(3 * cost / 4, @bank)
               operator.spend(cost / 4, @ldsteg)
@@ -616,31 +619,43 @@ module Engine
           raise GameError, 'Must visit minimum of two non-mine stops' if visits.sum(&:visit_cost) < 2
         end
 
+        def compute_stops(route, train = nil)
+          return super if !gc_train?(route) || other_gc_train?(route)
+
+          train ||= route.train
+          gc_train = train.dup
+          gc_train.distance = gc_train_distance(train.distance)
+          super(route, gc_train)
+        end
+
         def gc_train_distance(route_distance)
           if route_distance.is_a?(Numeric)
             town_distance_value = route_distance
+            city_distance_value = route_distance
           else
             # route has a 1 town value from the supporter
             town_distance = route_distance.find { |n| n['nodes'] == ['town'] }
-            town_distance_value = town_distance['pay'] + route_distance
+            city_distance_value = route_distance.find { |n| n['nodes'].include?('city') }['pay']
+            town_distance_value = town_distance['pay'] + city_distance_value
           end
           [
-              {
-                nodes: %w[city offboard town],
-                pay: route_distance,
-                visit: route_distance,
-              },
-              {
-                nodes: %w[town],
-                pay: town_distance_value,
-                visit: town_distance_value,
-              },
+            {
+              'nodes' => %w[town],
+              'pay' => town_distance_value,
+              'visit' => town_distance_value,
+            },
+            {
+              'nodes' => %w[city offboard town],
+              'pay' => city_distance_value,
+              'visit' => city_distance_value,
+            },
+
             ]
         end
 
         # Change "Stop" displayed if G&C power is used
         def route_distance(route)
-          return super if !gc_train?(route) || other_gc_train?(route)
+          return super if (!gc_train?(route) || other_gc_train?(route)) && route.train.distance.is_a?(Numeric)
 
           n_cities = route.stops.select { |s| s.visit_cost.positive? }.count { |n| n.city? || n.offboard? }
           n_towns = route.stops.count(&:town?)
--- a/lib/engine/game/g_18_mag/step/special_choose.rb
+++ b/lib/engine/game/g_18_mag/step/special_choose.rb
@@ -52,20 +52,22 @@ module Engine
 
             if distance.is_a?(Numeric)
               town_distance_value = 1
+              city_distance_value = distance
             else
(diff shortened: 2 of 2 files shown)


lib/engine/game/g_18_mag/game.rb
```diff
@@ -419,12 +419,16 @@ def must_buy_train?(_entity)
...
+          full_cost = train.price if train.owner == @depot
           if price != :free && train.owner == @depot
             if multiplayer?
               corp = %w[2 4].include?(train.name) ? @ldsteg : @mavag
-              operator.spend(cost / 2, @bank)
-              operator.spend(cost / 2, corp)
-              @log << "#{corp.name} earns #{format_currency(cost / 2)}"
+              bank_cost = full_cost ? (full_cost / 2) - (full_cost - cost) : cost / 2
+              corp_cost = full_cost ? full_cost / 2 : cost / 2
+              puts("here, #{bank_cost} #{full_cost} #{bank_cost + full_cost}")
```
bentziaxl, line 428: code could be cleaner. 
When using the kalman kando supporter, there's a train discount. However the corp still gets the half of the full cost of the train. This code handles that case

lib/engine/game/g_18_mag/game.rb
```diff
@@ -616,31 +620,47 @@ def check_distance(route, visits)
...
+        def compute_stops(route, train = nil)
+          train ||= route.train
+          distance = if gc_train?(route) && !other_gc_train?(route)
+                       gc_train_distance(route.train.distance)
+                     else
+                       route.train.distance
+                     end
+          train_dup = train.dup
+          train_dup.distance = distance
+
```
bentziaxl, line 632: When the train has been converted to +1 using the supporter, stops breaks when also adding gc (for example 2+1 turns into 2+3), the base compute_stops only returns two cities/offboards and 1 town. 
Duplicating the train and setting the correct distance fixes the issue.

Would a better alternative be to modify compute_stops to take distance=nil and pass the correct distance?
maybe copy the entire base method here and modify accordingly?

michaeljb:
At least these games will need pins: `[116455, 116943, 117151, 117170, 117186, 117303]`

bentziaxl:
hmm, looking why those will need to be pinned

michaeljb:
huh, very strange, the validate script finds a `nil can't be coerced into Integer` error for each of those games. I just ran it again and go the same result, but then tried opening each one in the browser and they all load just fine there 🤔

michaeljb:
Just did a bit more testing, with this PR's branch checked out the games I listed do appear to load fine in the browser but then the server rejects any further moves so I think the validate script is reporting things correctly, but it is weird that they don't render as broken in the browser.

The stack trace shows the `nil can't be coerced into Integer` exception coming from `compute_stops`.

michaeljb:
I don't know 18Mag and I can't really tell if it's expected that these changes would break any games, maybe @roseundy can take a look?

lib/engine/game/g_18_mag/game.rb
```diff
@@ -616,31 +619,47 @@ def check_distance(route, visits)
...
+              nodes: %w[town],
+              pay: town_distance_value,
+              visit: town_distance_value,
+            },
+            {
+              nodes: %w[city offboard town],
+              pay: city_distance_value,
+              visit: city_distance_value,
+            },
+
```
roseundy, line 656: Does changing the order between town and city,offboard,town matter?

bentziaxl: so as far as i know the rules of 18mag allow for a 2+2 train (after upgrade from the major) to run 1 city and 3 towns.
If that's the case than the correct order should be town then city/town/offboard.

lib/engine/game/g_18_mag/game.rb
```diff
@@ -419,12 +419,15 @@ def must_buy_train?(_entity)
...
           @train_bought = true if train.owner == @depot # No idea if this is what Lonny wants
           cost = price || train.price
+          full_cost = train.price if train.owner == @depot
           if price != :free && train.owner == @depot
             if multiplayer?
               corp = %w[2 4].include?(train.name) ? @ldsteg : @mavag
-              operator.spend(cost / 2, @bank)
-              operator.spend(cost / 2, corp)
-              @log << "#{corp.name} earns #{format_currency(cost / 2)}"
+              bank_cost = full_cost ? (full_cost / 2) - (full_cost - cost) : cost / 2
```
michaeljb, line 426: I find it hard to read that formula, how about this?

```suggestion
              bank_cost = full_cost ? cost - (full_cost / 2)  : cost / 2
```

lib/engine/game/g_18_mag/game.rb
```diff
@@ -616,31 +619,47 @@ def check_distance(route, visits)
...
+          distance = if gc_train?(route) && !other_gc_train?(route)
+                       gc_train_distance(route.train.distance)
+                     else
+                       route.train.distance
+                     end
+          train_dup = train.dup
+          train_dup.distance = distance
+
+          super(route, train_dup)
+        end
```
michaeljb, line 633:
```suggestion
        def compute_stops(route, train = nil)
          return super if !gc_train?(route) || other_gc_train?(route)
          
          train ||= route.train
          gc_train = train.dup
          gc_train.distance = gc_train_distance(train.distance)

          super(route, gc_train)
        end
```

lib/engine/game/g_18_mag/game.rb
```diff
@@ -616,31 +619,47 @@ def check_distance(route, visits)
...
-              },
+            {
+              nodes: %w[town],
+              pay: town_distance_value,
+              visit: town_distance_value,
+            },
+            {
+              nodes: %w[city offboard town],
+              pay: city_distance_value,
+              visit: city_distance_value,
```
michaeljb, line 654: I found the source of the validation errors, the classic server/client string/symbol issue. The base method expects the keys in these `distance` hashes to be strings: https://github.com/tobymao/18xx/blob/6310b539085311620e4382d7ec38958def286e7e/lib/engine/game/base.rb#L1373

In the compiled JS (client side / hotseat mode), symbols and strings are equivalent, but in the server side Ruby (and in `validate.rb`) they are not, so you need to use strings instead of symbols:

```suggestion
              'nodes' => %w[town],
              'pay' => town_distance_value,
              'visit' => town_distance_value,
            },
            {
              'nodes' => %w[city offboard town],
              'pay' => city_distance_value,
              'visit' => city_distance_value,
```

lib/engine/game/g_18_mag/game.rb
michaeljb: No need to do that, but I think this method can be cleaned up while keeping the same logic--I'll post my suggested code.

michaeljb (changes requested):

michaeljb (approved):

PR #9072, opened 2023-03-30
Add connected option for Token ability

Galatolol:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
* Add an option to the `token` ability that requires that the chosen city is reachable by the corporation
* Necessary for 1847 AE


lib/engine/step/special_token.rb
```diff
@@ -60,11 +60,12 @@ def process_place_token(action)
             true
           end
 
+        connected = special_ability.respond_to?(:reachable) ? special_ability.reachable : false
```
michaeljb, line 63: Would this work instead? I think we want to avoid adding new uses of `respond_to?` where possible.

```suggestion
        connected = special_ability.type == :token ? special_ability.reachable : false
```

lib/engine/step/special_token.rb
```diff
@@ -94,7 +95,13 @@ def teleport_complete
       end
 
       def available_hex(entity, hex)
-        return if !ability(entity).hexes.empty? && !ability(entity).hexes.include?(hex.id)
+        ability = ability(entity)
+
+        if ability.respond_to?(:reachable) && ability.reachable
```
michaeljb, line 100: Same as my other comment--but I do have another question, what other kinds of abilities can end up being called here that wouldn't be a `Token` ability that responds to `reachable`?

Galatolol: Teleport IIRC ;)

lib/engine/step/special_token.rb
Galatolol: Depends on how generic we want that to be I guess. I suspect `special_token` could be used by more abilities in the future. But I can see if `.type == :token` works, I wasn't keen on `respond_to?` either.

lib/engine/step/special_token.rb
```diff
@@ -60,11 +60,12 @@ def process_place_token(action)
             true
           end
 
+        connected = special_ability.type == :token ? special_ability.reachable : false
```
crericha, line 63: In graph, reachable and connected are two different concepts. I believe you want connected here (there is connected track between the destination city and another tokened city), so the ability should be renamed to connected.

lib/engine/step/special_token.rb
```diff
@@ -94,7 +95,13 @@ def teleport_complete
       end
 
       def available_hex(entity, hex)
-        return if !ability(entity).hexes.empty? && !ability(entity).hexes.include?(hex.id)
+        ability = ability(entity)
+
+        if ability.type == :token && ability.reachable
+          return @game.token_graph_for_entity(entity.owner).reachable_hexes(entity.owner)[hex]
```
crericha, line 101: I believe you want connected_hexes here.

crericha:
Add the new parameter to the ability readme in the token section with a description.

lib/engine/step/special_token.rb
```diff
@@ -94,7 +95,13 @@ def teleport_complete
...
+        end
+
+        return if !ability.hexes.empty? && !ability.hexes.include?(hex.id)
```
crericha, line 104: Move this back up to line 99. We need to check that the hex is valid before doing any further checks.

crericha (changes requested):
Please resolve my new comments. Thanks!

lib/engine/step/special_token.rb
crericha: Okay, I have them backwards. Then the kwarg should be called reachable, not connected. If that's easy to fix, can you fix it?

Galatolol: Sure.

Galatolol: Although I'm not sure if that's the right choice. In `process_place_token()` we operate on hex and use `reachable_hexes`. However, `place_token()` (the method with the `connected` kwarg) operates on city and uses `connected_nodes`. So I think that the name of the kwarg makes sense.

crericha: Yeah, I'm good with that.

crericha (approved):

PR #9084, opened 2023-04-05
[1849] Fix 8253 - Allow up movement when blocked from blue zone

### LINKED ISSUE #8253: [1849] Stock marker does not move up when against right edge
See https://18xx.games/game/94686?action=518

1849 has a peculiar rule governing marker movement when dividends are paid.
Rule 10.5.3: 
> If the corporation declares a dividend greater than or equal to its current market value, the stock market token is moved one column to the right. **If the token has already reached the end of a track or the limited zone before Phase 16, it is moved up one row instead (following the printed arrows).**

At the indicated action, the game log reports
"IFT share price blocked from moving right by phase"

As indicated in the screenshot below, IFT is against the barrier, but it has room to move up.  The price should have increased from 240 to 276.
Screen Shot 2022-08-29 at 13 13 49

MrSkinny:
Fixes #8253

First time using CodeSpaces, not sure why it made an empty commit!

### Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1849/stock_market.rb
+++ b/lib/engine/game/g_1849/stock_market.rb
@@ -40,12 +40,10 @@ module Engine
           price = share_price(coordinates).price
 
           # Special case rule: Stock marker UP if not in blue zone
-          if !@game.phase.status.include?('blue_zone') && BLOCKED_RIGHT_PRICES.include?(price)
-            return up(corporation, coordinates)
-          elsif !BLOCKED_RIGHT_PRICES.include?(price) || @game.phase.status.include?('blue_zone')
-            return super
-          end
-
+          return up(corporation, coordinates) if !@game.phase.status.include?('blue_zone') && BLOCKED_RIGHT_PRICES.include?(price)
+            
+          return super if !BLOCKED_RIGHT_PRICES.include?(price) || @game.phase.status.include?('blue_zone')
+            
           @game.log << "#{corporation.name} share price blocked from moving right by phase" if corporation
           coordinates
         end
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1849/stock_market.rb
```diff
@@ -38,7 +38,14 @@ def up(corporation, coordinates)
 
         def right(corporation, coordinates)
           price = share_price(coordinates).price
-          return super if !BLOCKED_RIGHT_PRICES.include?(price) || @game.phase.status.include?('blue_zone')
+
+          # Lower two blocked right prices push stock marker UP if not in blue zone
+          if !@game.phase.status.include?('blue_zone') && BLOCKED_RIGHT_PRICES.slice(0, 2).include?(price)
```
crericha, line 43: Let's not slice the constant. The behavior should work for all the block_right_prices. Up will bounce off the ceiling for the highest stock price.

lib/engine/game/g_1849/stock_market.rb
```diff
@@ -38,7 +38,14 @@ def up(corporation, coordinates)
...
+            @game.log << "#{corporation.name} share price blocked from moving right by phase; moves up instead" if corporation
```
crericha, line 44: These methods are used for more than stock price movement, so the log message shouldn't go here. If you feel a message is needed (I think it's okay without one), then the message would go in the `move` method.

The other log message below is a bug.

lib/engine/game/g_1849/stock_market.rb
MrSkinny: Assuming you mean there's already logic further up the chain that will force it to bounce off the ceiling.

crericha (approved):

crericha:
run `docker-compose exec rack rubocop -A` to fix the formatting failures.

MrSkinny:
> run `docker-compose exec rack rubocop -A` to fix the formatting failures.

Yeah I was just doing edits directly from the Github web interface for those small things, but that doesn't run the linter fixer first.  I'll log back into the Codespace and do it.

PR #9096, opened 2023-04-10
[1858] Remove graph call from Step::Track.actions

ollybh:
The call to `can_lay_tile?` in the actions method meant that the graph had to be evaluated whilst a game's history is loaded, increasing the amount of work to be done.

This call has been removed from actions and instead `auto_actions` is used to generate a pass if it is not possible for a minor to lay any track.

### DIFF
--- a/lib/engine/game/g_1858/step/track.rb
+++ b/lib/engine/game/g_1858/step/track.rb
@@ -9,7 +9,6 @@ module Engine
         class Track < Engine::Step::Track
           def actions(entity)
             return [] unless entity == current_entity
-            return [] if !entity.minor? && !entity.corporation?
 
             ACTIONS
           end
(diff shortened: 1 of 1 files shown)


crericha (approved):

lib/engine/game/g_1858/step/track.rb
```diff
@@ -10,11 +10,17 @@ class Track < Engine::Step::Track
           def actions(entity)
             return [] unless entity == current_entity
             return [] if !entity.minor? && !entity.corporation?
```
crericha, line 12: Do you have operating round entities that aren't minors or corporations? If not, this line is unnecessary.

ollybh: No, you're right. I was being overly cautious/paranoid. I've removed this line.

PR #9101, opened 2023-04-13
18NL pre-alpha implementation

philcampeau:
Passes Rake and Rubocop.
Still not functional, in progress, set to pre-alpha.

Permission from designer:
image

### DIFF
--- a/lib/engine/game/g_18_nl/step/special_token.rb
+++ b/lib/engine/game/g_18_nl/step/special_token.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/special_token'
+
+module Engine
+  module Game
+    module G18NL
+      module Step
+        class SpecialToken < Engine::Step::SpecialToken
+          def special_token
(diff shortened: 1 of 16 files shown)


lib/engine/game/g_18_nl/step/special_token.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require_relative '../../../step/special_token'
+
+module Engine
+  module Game
+    module G18NL
+      module Step
+        class SpecialToken < Engine::Step::SpecialToken
+          def special_token
+            raise GameError, 'Cannot place station on hex or tile with existing track' unless tile.path.empty?
```
michaeljb, line 11: I think this will need to be `tile.paths.empty?`

```suggestion
            raise GameError, 'Cannot place station on hex or tile with existing track' unless tile.paths.empty?
```

philcampeau:
Withdrawing for now. Needs more work and I'm not sure when I'll have time to pursue it.

PR #9102, opened 2023-04-13
[1858] 5-share to 10-share corporation conversion

ollybh:
### Implementation Notes

As a sell-type stock action, corporations (public companies) can convert from a 5-share company to a 10-share company. And, at the start of phase 5, all companies that have not already grown up are automatically converted.

There's also some code added to allow closed corporations to be restarted, and to make sure they restart as 10-share companies if we're after the start of phase 5.

Finally, there's a bit of code churn, with a change in the labels used for corporation type: `5-share` and `10-share` instead of `medium` and `large`. This matches the terminology used in the 1858 rules.

There are no changes to shared code in this pull request. It uses the implementation of stock round convert actions that was recently added by @crericha in #9099.

#### Convert button rendered for A corporation
image

### PR checks


lib/engine/game/g_1858/game.rb
```diff
@@ -461,6 +472,33 @@ def rust_phase4_trains!(purchased_train)
...
+
+          shares = @_shares.values.select { |share| share.corporation == corporation }
+          shares.each { |share| share.percent /= 2 }
+          corporation.share_holders.transform_values! { |percent| percent / 2 }
+
+          new_shares = Array.new(5) { |i| Share.new(corporation, percent: 10, index: i + 4) }
+          new_shares.each do |share|
+            add_new_share(share)
+          end
+          new_shares
```
michaeljb, line 491: I don't see anything that calls `convert!` using this return value, can it be removed?

michaeljb, line 491: I don't see anything that calls `convert!` using this return value, can it be removed?

```suggestion
```

ollybh: Yes, good catch. I've removed this line.

PR #9104, opened 2023-04-15
Raise an error if the tile being laid (or its opposite) is already on the board

michaeljb:
It's #9029 again, but the error is only raised if the tile's hex is the hex from the game. Tiles need to be assigned an `A1` hex belonging to the TileSelector (this is why #9029 broke everyone's games), so it's not enough to check whether a tile has a hex, we need to check if that hex is the game's hex.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2084,6 +2084,9 @@ module Engine
       def update_tile_lists(tile, old_tile)
         add_extra_tile(tile) if tile.unlimited
 
+        # TileSelector creates "fake" A1 hexes that are attached to the tiles,
+        # so here we need to check that tile.hex actually belongs to the Game
+        # object
         if (hex = tile.hex) && (hex == hex_by_id(hex.id))
           raise GameError,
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -2083,6 +2083,12 @@ def bankruptcy_limit_reached?
 
       def update_tile_lists(tile, old_tile)
         add_extra_tile(tile) if tile.unlimited
+
+        if (hex = tile.hex) && (hex == hex_by_id(hex.id))
```
crericha, line 2087: I don't follow the case you are trying to prevent. What circumstances does the tile have a hex, but that hex isn't a valid hex?

crericha, line 2087: Add a comment about the tile selector assigning a fake hex.

crericha (approved):

lib/engine/game/base.rb
michaeljb: Done, thanks!

PR #9109, opened 2023-04-15
Don't let players over the cert limit autopass

### LINKED ISSUE #7142: In 18CZ auto pass in Share Dealing Round jumps over reduction down to certificate limit
Game: 
https://18xx.games/game/70487
Action ID
https://18xx.games/game/70487?action=549

In game 70487 the player Zezzo sells shares and buys a UG. 
The Player I33twash0r automatically passes with 13 of allowed 12 certificates.

The correction of the certificate limit should disable auto pass in SDR.

We cannot correct this in master mode. If there is a possibility, please tell us.

michaeljb:
Fixes #7142

While #7142 only references 18CZ, this can affect all games.

I repro'd the behavior locally and created a game where illegal passes were happening automatically. With this change applied, the validation script did not break on the game with the illegal actions, since they are "snuck" in with auto_actions, so no pins/etc will be needed.

### DIFF
--- a/lib/engine/step/buy_sell_par_shares.rb
+++ b/lib/engine/step/buy_sell_par_shares.rb
@@ -404,6 +404,9 @@ module Engine
       end
 
       def should_stop_applying_program(entity, program, share_to_buy)
+        # don't let players over the cert limit autopass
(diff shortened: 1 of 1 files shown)


Zwergenpunk:
lgtm

PR #9113, opened 2023-04-16
[1846] In track and token step, highlight tokenable hexes

### LINKED ISSUE #9112: [1846] during track and token step, sometimes valid hexes for the token are faded out and look invalid
https://gist.github.com/michaeljb/d5b1033696ddc73a14834aea7231f923

In this gist, NYC has upgraded one tile, but not laid a new yellow tile or placed a token. The only hex lit up for them is the only place they could lay a yellow tile. They can still place a token in Chicago, but this is not clear in the UI.

michaeljb:
Fixes #9112

Screenshots from the hotseat game shared in that bug; NYC has laid a tile upgrade, and can still place a token in Chicago or Cleveland, or lay a yellow tile in South Bend. The new behavior incorporates the normal Token step behavior, where all reachable hexes are highlighted.

Before:

Screenshot 2023-04-16 015019

After:

Screenshot 2023-04-16 015112

### DIFF
--- a/lib/engine/step/token.rb
+++ b/lib/engine/step/token.rb
@@ -25,7 +25,7 @@ module Engine
       end
 
       def available_hex(entity, hex)
-        @game.token_graph_for_entity(entity).reachable_hexes(entity)[hex]
+        tokener_available_hex(entity, hex)
       end
 
       def process_place_token(action)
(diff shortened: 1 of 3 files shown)


Zwergenpunk:
lgtm

PR #9114, opened 2023-04-16
Fix displayed stock price movement at end of market

### LINKED ISSUE #8536: [1817] Stock moves show "1 right" when at end of stock market
image

https://18xx.games/game/86792?action=1835

michaeljb:
No step actually implements `movement_str`, so I removed the check for it when defining `direction`.

Fixes #8536, which is actually applicable to more games than just 1817.

### DIFF
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -37,24 +37,31 @@ module View
 
           corp_income = option[:corporation] + option[:divs_to_corporation]
 
-          new_share = entity.share_price
-
           direction =
-            if new_share && option[:share_direction]
+            if (new_share_price = entity.share_price) && option[:share_direction]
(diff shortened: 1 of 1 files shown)


assets/app/view/game/dividend.rb
```diff
@@ -37,24 +37,31 @@ def render
...
 
-              moves.map do |times, dir|
-                times.times { new_share = @game.stock_market.find_relative_share_price(new_share, dir) }
+              movement_str =
+                moves.map do |times, dir|
+                  real_moves = 0
+                  times.times do
+                    prev_price = new_share_price
+                    new_share_price = @game.stock_market.find_relative_share_price(new_share_price, dir)
+                    real_moves += 1 unless prev_price == new_share_price
```
crericha, line 50: Do you think we should break out if price doesn't change?

PR #9115, opened 2023-04-16
[1822 family] fix acquiring special trains, clean up train discarding

### LINKED ISSUE #8558: [1822 family] DiscardTrain step should force discard when owning multiple pullmans as well as the other special trains
At the moment this bug report is based just on my reading the code, I don't have a broken sample game yet.

* [1822MX has logic to discard multiple Pullmans](https://github.com/tobymao/18xx/blob/a6cb05dc3c2df2bb9f8b31eeba257989aeab4cb3/lib/engine/game/g_1822_mx/step/discard_train.rb), I believe this should be moved into [1822 's step](https://github.com/tobymao/18xx/blob/a6cb05dc3c2df2bb9f8b31eeba257989aeab4cb3/lib/engine/game/g_1822/step/discard_train.rb), so that all the games in the family inherit this logic, instead of just 1822MX
* I'm not certain, but I think the 1822 games should have the 1822 discard train step in the stock rounds, instead of the base discard train step; this was certainly the case for 1822PNW (#8557)
  * [1822](https://github.com/tobymao/18xx/blob/a6cb05dc3c2df2bb9f8b31eeba257989aeab4cb3/lib/engine/game/g_1822/game.rb#L1162)
  * [1822MX](https://github.com/tobymao/18xx/blob/a6cb05dc3c2df2bb9f8b31eeba257989aeab4cb3/lib/engine/game/g_1822_mx/game.rb#L355)
  * [1822PNW](https://github.com/tobymao/18xx/blob/a6cb05dc3c2df2bb9f8b31eeba257989aeab4cb3/lib/engine/game/g_1822_pnw/game.rb#L551)

### LINKED ISSUE #8964: 1822MX company with special train prohibited from acquiring another special train
The rules make it clear that a company with a special train may acquire another and must then discard one. I was prevented from doing so in https://18xx.games/game/112299?action=1014

**1822MX rules** _6.5.15 There are three special trains that are bought as private companies (in addition to the private company P1 5-train, which is a normal 5-train). A company may only own one of these special trains at a time. If a company has one of them and acquires another, it must select one to discard from the game._

Untitled

michaeljb:
* "Special trains" in 1822 games are permanent 2, L, or 3/2 trains that come from private companies. Majors are allowed to own only one, but are allowed to acquire multiples but then discard down to one. Change `can_gain_extra_train?` to accommodate this.
    * `can_gain_extra_train?` still prohibits directly gaining a second Pullman via private company, but a second Pullman can be gained by acquiring a Minor that has one.
* 1822MX had some logic for discarding excess Pullman trains, but the same logic and rules are applicable to other 1822 titles as well, so move that logic up to 1822.

* Fixed a typo in 1822 code, `aqcuired` -> `acquired`

Fixes #8558
Fixes #8964

-----

Broken games - a small handful of 1822, 1822MRS, and 1822PNW games had Majors acquiring a second Pullman, but not being forced to discard it, so these games need to be pinned (or archived if complete):

`[83306, 86393, 92092, 97900, 98269, 109800, 114635, 116539, 117920]`


lib/engine/game/g_1822/game.rb
```diff
@@ -1387,11 +1388,8 @@ def bidbox_start_private
         end
 
         def can_gain_extra_train?(entity, train)
-          if train.name == self.class::EXTRA_TRAIN_PULLMAN
-            return false if entity.trains.any? { |t| t.name == self.class::EXTRA_TRAIN_PULLMAN }
-          elsif self.class::EXTRA_TRAIN_PERMANENTS.include?(train.name)
-            return false if entity.trains.any? { |t| self.class::EXTRA_TRAIN_PERMANENTS.include?(t.name) }
-          end
+          return false if pullman_train?(train) && entity.trains.any? { |t| pullman_train?(t) }
```
crericha, line 1391: Should this be disallowed or only forced discard?

lib/engine/game/g_1822/game.rb
```diff
@@ -1694,6 +1692,15 @@ def extra_train_permanent_count(corporation)
           corporation.trains.count { |train| extra_train_permanent?(train) }
         end
 
+        def pullman_train_count(corporation)
+          corporation.trains.count { |train| pullman_train?(train) }
+        end
+
+        def remove_discarded_train?(action)
```
crericha, line 1699: suggest passing in the train instead of the action

lib/engine/game/g_1822/game.rb
crericha: Rules say "may never own more than one pullman", so I think you can acquire the private, but then have to discard.

PR #9116, opened 2023-04-16
[18OE] float major

### LINKED ISSUE #9032: [18OE] Rules Implementation - UK/France Scenario
General

- [ ]  Trains
- [ ]  Phases
- [x]  Stock market
- [ ]  Map
- [ ]  Tiles
- [x]  Money, cert limit
- [ ]  Logos

Private auction

- [x]  Tiered Waterfall
- [x]  Can purchase anything on top row
- [ ]  Can attempt to purchase minor for greater than face value - causes immediate auction

Entities

- [x]  Ponts et Chaussees
- [ ]  Star Harbor Trading Co
- [ ]  Central Circle Transport Co
- [ ]  White Cliffs Ferry
- [ ]  Minors - C, H, K, M
- [x] Regionals

SR

- [x]  Sell - buy
- [x]  Price goes down one space per share when selling
- [x]  Cannot buy secondary shares until all companies/minors floated (major phase)
- [x]  Cannot pass until all minors floated
- [x]  Close unfloated regionals once max number floated
- [x]  Fixed par prices
- [x]  Home token upon floating
- [x]  Incremental capitalization
- [ ]  May own over 60% for majors/nationals (must pay double from open market)
- [x]  Float Minor
- [x]  Float Major
- [ ]  Trade with player
- [x]  Stock moves up at end of round if sold out
- [x]  Regional Phase
- [ ]  Major Phase
- [ ]  Merge minors into majors
- [ ]  Consolidation Phase

OR

- [x]  Track points based on corporation type
- [x]  Build only in track rights zone (non-majors)
- [ ]  Majors restricted to track rights zone until first merge of minor
- [x]  Semi-restrictive building
- [ ]  Extra cost for cross port/ferry track/token
- [ ]  Patronage tiles
- [ ]  OE route bonus (30/60/100)
- [ ]  First time OE bonus
- [ ]  Ferry crossings
- [ ]  Port and Sea crossings
- [ ]  Port authorities
- [ ]  Pay, Split, Hold (regional/major), Split (minor), Pay (national)
- [ ]  Transfer tokens
- [ ]  2+2s reserved 1 per company
- [ ]  Pullman cars
- [ ]  Form national (train phase 4/6/8)
- [ ]  Issue/Redeem shares

In game wiki:
TBD

Kalnoch:
Fixes #9032 

> Please minimize the amount of changes to shared `lib/engine` code, if possible

> If you are implementing a new game, please break up the changes into multiple PRs for ease of review.

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
Add in the ability to float majors after all minors/regionals have floated. Added more phases/trains/tiles.

* **Any Assumptions / Hacks**
~~Needed to add tokens to the accessor section as it didn't seem to be able to modify an existing corporations tokens otherwise.~~
~~The code does not force a 50% non-president share holder who converts a regional to major to buy a share at the end of the convert step.~~


lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,38 +1,102 @@
...
+            end
+
+            # Set corporation type to :major
+            corporation.type = 'major'
+
+            # Majors are affected by the stock market, set tokens in the correct place
+            @game.stock_market.move_right(corporation)
+            @game.stock_market.move_right(corporation)
+            @game.stock_market.move_up(corporation)
+            corporation.tokens += [40, 60, 60, 80, 80, 80].map { |price| Engine::Token.new(corporation, price: price) }
```
crericha, line 93: corporation.tokens.concat() and then you don't need to make tokens an attr_accessor in operator.rb

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,38 +1,102 @@
...
 require_relative '../../../step/buy_sell_par_shares'
+require_relative '../../../action/convert'
 
 module Engine
   module Game
     module G18OE
       module Step
         class BuySellParShares < Engine::Step::BuySellParShares
           def actions(entity)
+            return corporation_actions(entity) if entity.corporation?
```
crericha, line 12: Do you want to do all corporations or ones owned by the current entity?

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,38 +1,102 @@
...
             return [] unless entity == current_entity
+            return [] if @round.current_actions.any? { |x| x.instance_of?(Engine::Action::Par) }
```
crericha, line 14: Shouldn't the code call pass! after process_par_action, rather than going back here?

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,38 +1,102 @@
...
+            actions << 'convert' if can_convert_any?(entity)
 
-            actions << 'pass' if !can_float_minor?(entity) && !bought? && !actions.empty?
+            actions << 'pass' if !can_float_minor?(entity) && !actions.empty?
             actions
           end
 
+          def corporation_actions(entity)
+            if @round.current_actions[0].instance_of?(Engine::Action::BuyShares) &&
+              @round.current_actions[0].bundle.corporation != entity
```
crericha, line 30: Why actions[0]? Can you not sell first?

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,38 +1,102 @@
...
+          end
+
           def can_buy_any_from_ipo?(entity)
             return unless @game.corporations.all?(&:ipoed)
 
             super
           end
 
+          def can_buy?(entity, bundle)
+            if @round.current_actions[0].instance_of?(Engine::Action::Convert) &&
```
crericha, line 45: Typically games redefine this constant to include additional actions that take the place of the "purchase" action. https://github.com/tobymao/18xx/blob/master/lib/engine/step/buy_sell_par_shares.rb#L16

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,38 +1,102 @@
...
+
+            super
+          end
+
+          def can_convert?(corporation)
+            if !@game.corporations.all?(&:ipoed) || @round.current_actions.any? { |x| x.instance_of?(Engine::Action::Convert) }
+              return
+            end
+
+            return if @round.current_actions[0].instance_of?(Engine::Action::BuyShares) &&
```
crericha, line 58: Then this would be return if any purchase actions

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Kalnoch: You can float major if you have 50% shares but not ownership, so need to check all corps

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Kalnoch: Yeah that would be a better way of doing that

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Kalnoch: So the conversion in 18OE allows a single buy either before or after the conversion and is all one step. So you can't sell, buy, then convert.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
crericha: If the rule is you can buy one share before or after a convert, then why does convert factor in at all? Isn't it simply there isn't a buy action?

lib/engine/game/g_18_oe/game.rb
```diff
@@ -231,6 +457,14 @@ def metropolis_tile?(tile)
              OE18 OE26 OE27 OE28 OE29 OE30 OE37 OE38 OE39 OE40 OE41].include?(tile.name.to_s)
         end
 
+        def must_buy_train?(entity)
+          # must buy the reserved 2+2, otherwise only majors must buy trains
+          # return unless entity.type == 'major'
+          return if depot.depot_trains[0]&.name != '2+2' && entity.type != 'major'
```
ollybh, line 463: The type attribute returns a label, not a string, so you need to be testing for `:major`.
Also, it's clearer to use `Array.first` to return the first item in an array.
```
          return if depot.depot_trains.first&.name != '2+2' && entity.type != :major
```

lib/engine/game/g_18_oe/game.rb
```diff
@@ -267,6 +501,13 @@ def after_par(corporation)
           corporation.spend(TRACK_RIGHTS_COST[region], @bank)
         end
 
+        def issuable_shares(entity)
+          return [] if !entity.corporation? || (entity.corporation? && entity.type != 'major')
```
ollybh, line 505: Again, check for `:major` not `'major'`.
The `entity.corporation?` check on the right-hand side of the operator is redundant.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -1,13 +1,26 @@
...
         class BuySellParShares < Engine::Step::BuySellParShares
+          def setup
+            super
+
+            @can_convert = []
+            @game.corporations.each do |corp|
+              if corp.share_holders[current_entity] >= 50 && corp.type == :regional && @game.corporations.all?(&:ipoed)
+                @can_convert << corp
+              end
+            end
```
ollybh, line 19: As you're returning the items from the list you're iterating over, you could use `Array.select` or `Array.reject`.
```
@can_convert = @game.corporations.select do |corp|
  corp.share_holders[current_entity] >= 50 && corp.type == :regional && @game.corporations.all?(&:ipoed)
end
```

lib/engine/game/g_18_oe/game.rb
Kalnoch: Yeah, I'm still relatively new to Ruby so knowing when I need to use labels vs strings is tricky for me

ollybh: These label/string mismatches don't get picked up when running in hotseat mode – all the code is converted by Opal into Javascript, which only has strings. But when running the game in server mode they will break.
Try creating a game in server mode with a few dummy players. This will flush these errors out.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -267,6 +501,13 @@ def after_par(corporation)
           corporation.spend(TRACK_RIGHTS_COST[region], @bank)
         end
 
+        def issuable_shares(entity)
+          return [] if !entity.corporation? || (entity.corporation? && entity.type != :major)
```
ollybh, line 505:
```
    return [] if !entity.corporation? || entity.type != :major
```
Testing `entity.corporation?` a second time is redundant, it will not be evaluated unless it's already been found to be true.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,69 @@ def actions(entity)
...
+
           def can_buy_any_from_ipo?(entity)
             return unless @game.corporations.all?(&:ipoed)
 
             super
           end
 
+          def can_buy?(entity, bundle)
+            action = @round.current_actions.find { |a| a.is_a?(Engine::Action::Convert) }
+            return false if action && action&.entity != bundle.corporation
```
ollybh, line 48: Could there be multiple convert actions in `@round.current_actions`? If there are, then you will just be checking the first one to see whether it matches `bundle.corporation`.
Would this be better?
```
return false if @round.current_actions.any? { |a| a.is_a?(Engine::Action::Convert) && a.entity == bundle.corporation }
```

Kalnoch: There can only be one convert action done per turn, so there should only be 0 or 1 in current_actions

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,69 @@ def actions(entity)
             actions << 'par' if can_ipo_any?(entity) || can_float_minor?(entity)
             actions << 'buy_company' if !purchasable_companies(entity).empty? || !buyable_bank_owned_companies(entity).empty?
             actions << 'sell_shares' if can_sell_any?(entity)
-
-            actions << 'pass' if !can_float_minor?(entity) && !bought? && !actions.empty?
+            actions << 'convert' unless @can_convert.empty?
```
ollybh, line 29: This line isn't needed. The earlier call to `corporation_actions` will have added convert to the action list if any conversions are possible.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -68,6 +124,29 @@ def visible_corporations
...
           end
 
+          def process_buy_shares(action)
+            super
+            corp = action.bundle.corporation
+            @can_convert = if @can_convert.find { |c| c == corp }
+                             [corp]
+                           else
+                             []
+                           end
```
ollybh, line 134: More concise:
```
@can_convert.select! { |c| c == corp }
```

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,69 @@ def actions(entity)
...
+            return false if action && action&.entity != bundle.corporation
+
+            super
+          end
+
+          def can_sell?(entity, bundle)
+            return unless bundle
+            return if bundle.corporation.type == :regional
+
+            super
```
ollybh, line 57: There needs to be a check here to stop the player selling shares in a major that has just formed.

ollybh (changes requested):
This is mostly correct, but I can spot a couple of problems that could arise if there is a 50/50 ownership split of a regional that is converted to a major:

1. There is no check in `can_sell?` to stop the player from selling a share in the new major after conversion.
2. There is no check to make sure that the player is the president after the conversion. This should probably go in `pass!`, throwing an error if the player attempts to pass without buying a share if they are not the president.

It might be easier to track what is happening if you store a few different items of step/state instead of your `can_convert` array:

- `@converted` – starts as nil, set to the new major when a conversion is started.
- `@bought` – starts as nil, set to the entity if a share is bought.
- `@sold?` – starts as false, set to true if any shares are sold.

From these bits of state:

- `can_sell?` would check the entity is not `@converted`.
- `can_buy?` would check against `@converted`.
- `corporation_actions` would have the calculation currently in `setup`, and also look at `@bought` and `@sold?` (maybe create a `can_convert?` method for the check for each entity).

[Sorry for taking a while to take a closer look at this pull request. It's been several years since I've played 18OE, so I had to relearn the rules.]

Kalnoch:
> This is mostly correct, but I can spot a couple of problems that could arise if there is a 50/50 ownership split of a regional that is converted to a major:
> 
> 1. There is no check in `can_sell?` to stop the player from selling a share in the new major after conversion.
> 2. There is no check to make sure that the player is the president after the conversion. This should probably go in `pass!`, throwing an error if the player attempts to pass without buying a share if they are not the president.
> 
> It might be easier to track what is happening if you store a few different items of step/state instead of your `can_convert` array:
> 
> * `@converted` – starts as nil, set to the new major when a conversion is started.
> * `@bought` – starts as nil, set to the entity if a share is bought.
> * `@sold?` – starts as false, set to true if any shares are sold.
> 
> From these bits of state:
> 
> * `can_sell?` would check the entity is not `@converted`.
> * `can_buy?` would check against `@converted`.
> * `corporation_actions` would have the calculation currently in `setup`, and also look at `@bought` and `@sold?` (maybe create a `can_convert?` method for the check for each entity).
> 
> [Sorry for taking a while to take a closer look at this pull request. It's been several years since I've played 18OE, so I had to relearn the rules.]

Really appreciate the thorough review! I will take a look at these suggestions this weekend

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Kalnoch: I tried that and it always breaks the game, not sure why. My guess is the convert isn't getting added into the player actions list only the corporations?

ollybh: Is this breaking for new games, or loading ones that were created before the code was changed? If it's breaking for loading old games then this might mean that there is a pass action in the game that is no longer needed. I can't immediately see why this would have changed the behaviour.

Kalnoch: Breaks both. If convert is the only action left for a player it auto passes them as `actions.empty?` is true for the player

ollybh: OK, leave this in if it's not working without it. I didn't need to add a convert action here in 1858, which also uses `corporation_actions`, so I didn't think it would be needed for you.

[Apologies for not replying sooner, I missed that you'd added another comment].

lib/engine/game/g_18_oe/game.rb
```diff
@@ -231,6 +457,14 @@ def metropolis_tile?(tile)
              OE18 OE26 OE27 OE28 OE29 OE30 OE37 OE38 OE39 OE40 OE41].include?(tile.name.to_s)
         end
 
+        def must_buy_train?(entity)
+          # must buy the reserved 2+2, otherwise only majors must buy trains
+          # return unless entity.type == 'major'
+          return if depot.depot_trains.first&.name != '2+2' && entity.type != :major
```
ollybh, line 463: Code style: `must_buy_train?` should always return a boolean value, so this should be `return false if…`.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+
           def can_buy_any_from_ipo?(entity)
             return unless @game.corporations.all?(&:ipoed)
 
             super
           end
 
+          def can_buy?(entity, bundle)
+            action = @round.current_actions.find { |a| a.is_a?(Engine::Action::Convert) }
+            return false if action && action&.entity != bundle.corporation
```
ollybh, line 48: You're now setting `@converted` in process_convert`, so can't you just check that value here, instead of digging through the action history?

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+
+            super
+          end
+
+          def can_sell?(entity, bundle)
+            action = @round.current_actions.find { |a| a.is_a?(Engine::Action::Convert) }
+            return unless bundle
```
ollybh, line 55: `return false unless`

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+            return false if bundle.corporation.type == :regional
+            return false if action && action&.entity == bundle.corporation
```
ollybh, line 57:
```suggestion
            return false if bundle.corporation == @converted
```

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+          end
+
           def can_float_minor?(entity)
             return unless entity.player?
 
             !bought? && entity.companies.any? { |company| @game.company_becomes_minor?(company) }
           end
 
+          def can_convert_any?
+            @game.corporations.select { |corp| can_convert?(corp) }.any?
```
ollybh, line 69: You're iterating through the array twice here.
```suggestion
            @game.corporations.any? { |corp| can_convert?(corp) }
```

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+
+            unless entity.president?(current_entity)
+              # does the converter have 50% ownership?
+              return unless entity.share_holders[current_entity] >= 50
+              # does the converter have the ability to buy a share to take the presidency?
+              return unless @bought.nil?
+
+              # does the converter have enough liquidity to become president?
+              new_share_price = @game.stock_market.find_share_price(entity, %i[right right up])
+              return unless @game.liquidity(current_entity) >= new_share_price.price
```
ollybh, line 92: Again, these should all be `return false`.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+
+          def can_convert?(entity)
+            # are we in the major railroad phase?
+            return unless @game.corporations.all?(&:ipoed)
+            # is the current entity a regional?
+            return unless entity.type == :regional
+            # has any shares been sold?
+            return if @sold
+            # has there already been a conversion?
+            return unless @converted.nil?
```
ollybh, line 80: The explicit `nil?` check isn't needed.
```suggestion
            return false if @converted
```

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+            # if anything has been bought, is it the current corp?
+            return if !@bought.nil? && @bought != entity
```
ollybh, line 82:
```suggestion
            return if @bought && @bought != entity
```

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -16,23 +26,101 @@ def actions(entity)
...
+
+            unless entity.president?(current_entity)
+              # does the converter have 50% ownership?
+              return unless entity.share_holders[current_entity] >= 50
+              # does the converter have the ability to buy a share to take the presidency?
+              return unless @bought.nil?
```
ollybh, line 88:
```suggestion
              return false if @bought
```

ollybh (changes requested):
It looks like you've addressed the biggest issues I had before: not being president after a company is converted, and selling shares in a newly converted company.

You are still checking the round's action history in a couple of places to determine the company that is being converted, I believe that this is no longer necessary.

Other than that, it's just some code style changes:
- Methods where the names end in ? should always return a boolean, so the guard clauses should be `return false`.
- You usually don't need to explicitly check `nil?` on a variable in a guard clause. In particular, `return unless @var.nil?` is a double negation, `return if @var` is much clearer to me.

ollybh (approved):
Looks good.

PR #9123, opened 2023-04-17
[1858] Alternate view for minors

ollybh:
The default view for a minor shows:
1. Its name.
2. The trains it owns.
3. The cash in its treasury.
4. Its position in the operating order.
5. The value of its last run.

minor-default

Most of these are irrelevant for 1858, as the minor parts of the private railway companies can't own trains, don't have a treasury, and don't run routes.

This alternate view makes minors look like companies. This shows more relevant information:
- Revenue.
- Home hexes.
- Whether it can be used to start a corporation (public company).
- Face value.

Making the minors look like companies also helps hide the internal split of private railways into a company and a minor, this does not need to be visible to the player.

minor-alternate

This introduces new code into the shared `assets/app/view/game` directory, but it will be ignored by all other games unless they explicitly invoke it from Game::corporation_view.

- ~~Add the `pins` label if this change will break existing games~~

### DIFF
--- a/assets/app/view/game/alternate_corporations.rb
+++ b/assets/app/view/game/alternate_corporations.rb
@@ -510,6 +510,67 @@ module View
           h(:div, { style: { gridColumnStart: 2 } }, [movement_table]),
         ])
       end
+
+      # This is used by 1858 which has private railway companies that act a bit
+      # like minor railways, laying track in operating rounds. This rendering
+      # code is a simplified version of Engine::View::Game::Company.render.
+      def render_private_railway
+        minor = @corporation
+
+        select_minor = lambda do
+          if @selectable
+            selected_corporation = selected? ? nil : minor
+            store(:selected_corporation, selected_corporation)
+          end
+        end
+
+        header_style = {
+          background: minor.color,
+          color: minor.text_color,
+          border: '1px solid',
+          borderRadius: '5px',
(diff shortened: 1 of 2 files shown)


crericha:
How do these minors participate in the operating round? (I'm not familiar with the game)

ollybh:
> How do these minors participate in the operating round? (I'm not familiar with the game)

They can lay one or two yellow tiles in the track step, with any costs (for terrain or the second tile) paid by the player who owns the minor. All other steps (token/routes/dividend/trains) are skipped.

crericha:
Another dumb question: did you consider making the railways privates and adding them to the operating round?

It might be better to chat on slack, so we can go back and forth more quickly.

ollybh:
> Another dumb question: did you consider making the railways privates and adding them to the operating round?

That was my initial approach, back in pull request #8578. It was fairly disruptive to shared code that assumed companies are not operators, and Toby suggested that I split the private railways into pairs of companies and minors. https://github.com/tobymao/18xx/pull/8578#issuecomment-1347773369

crericha (approved):

ianwilson156:
I admit this is all my fault. When I designed 1858, I created entities which are somewhere between a traditional private and a minor. 
Most humans can understand after a game or two, but computers...

PR #9125, opened 2023-04-18
[18Rhl] Fix illegal route

### LINKED ISSUE #8774: [18Rhl] Illegal route allowed
**What's your Game ID?**

https://18xx.games/game/111564?action=95 - got to the Run trains step and hit Auto.

A route going from one part of a Rhine Metropolis to another is allowed which shouldn't be the case.

> **A route comprises of at least two different cities** or off-board destinations that are connected by track.
> Each route must contain somewhere on it a station marker of the corporation using the route. A route may
> neither start nor end at a town. 
> (....)
> When the hexes of the “Rhine Metropolis” have been upgraded in green the Rhine can be crossed there by
> a “Trajekt“ (railroad ferry). The three station spaces on the green Metropolis tiles are regarded as one
> city tile with three station spaces
> {...}
> As long as the “Rhine Metropolis“ hexes have not been upgraded (they show still the yellow base) the two
> sides of the river are treated as separate cities on the same tile.

**How to Reproduce the Bug?**

Have a corporation with a station in a green "Rhine Metropolis". Run a train from one city part to another, through the ferry link.

**Which Browser / Platform?**
Firefox 103.1

Galatolol:
Fixes #8774 

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Raise an error when a route consists of only a Rhine metropolis city.

This may break existing games, so `pins` is needed.

### DIFF
--- a/lib/engine/game/g_18_rhl/game.rb
+++ b/lib/engine/game/g_18_rhl/game.rb
@@ -698,6 +698,10 @@ module Engine
           last = visits.last
           rge = rge_train?(route)
 
+          if route.hexes.uniq.size == 1 && RHINE_METROPOLIS_HEXES.include?(first.hex.id)
+            raise GameError, 'Route cannot consist of only one Rhine metropolis'
+          end
+
           if rge
             if !rge_terminus?(first) && !rge_terminus?(last)
               raise GameError, 'Route for 8 trains must begin and/or end in an RGE hex'


michaeljb:
The game engine doesn't actually re-validate routes when they are processed, so this does not break any games.

michaeljb (approved):

PR #9139, opened 2023-04-23
18NY - fix salvage rendering

benjaminxscott:
addresses regression in #9136, adds tests

> Please minimize the amount of changes to shared `lib/engine` code, if possible

> If you are implementing a new game, please break up the changes into multiple PRs for ease of review.

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/assets/app/view/game/discard_trains.rb
+++ b/assets/app/view/game/discard_trains.rb
@@ -16,6 +16,8 @@ module View
         }
         step = @game.active_step
         overflow = step.crowded_corps.map do |corporation|
+          salvagable = step.trains(corporation)&.any?(&:salvage)
+          @verb = salvagable ? 'Salvage' : 'Discard'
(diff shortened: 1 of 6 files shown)


crericha:
I have a few issues with this PR:
1. The new fields are not tied into the game's salvaging logic. There should not be two sources of truth.
2. The button should change the verb based on which train is selected as there may be a mixture of salvageable and non-salvageable trains.
3. If we are going to have a common concept of salvage, it should be supported in common code and not game specific code.

I personally don't think this reported issue needs to be fixed, but won't oppose a fix if done properly and completely.

benjaminxscott:
@crericha - I don't disagree, will send a slack DM

benjaminxscott:
Will work on generalizing later on

PR #9156, opened 2023-04-30
update game.shares to utilize both corporation and player shares 

Kalnoch:
if update_cache is called on shares it doesn't wipe out shares currently in players hands


michaeljb:
The impact of changing a core method like this could be large, so I'm going to run validation against the full DB.

michaeljb (approved):

PR #9159, opened 2023-05-03
[1847 AE] Add draft and the remaining companies

Galatolol:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Companies are drafted in the following order: Player4, 3, 2, 1, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, ...
If all players pass and there are still undrafted companies, a "short OR" takes place in which companies pay revenue and the L corporation operates if it is floated (5 of its shares are held by players).

Draft step and round based on those for 18CZ.
If the next player doesn't have valid actions, a pass is automatically processed for them, similarly to how it is done for 1871.


lib/engine/game/g_1847_ae/round/draft.rb
```diff
@@ -0,0 +1,70 @@
...
+        class Draft < Engine::Round::Draft
+          def initialize(game, steps, **opts)
+            super
+
+            @players_reordered_for_next_round = false
+          end
+
+          def next_entity_index!
+            # First round of draft is perforemd in reverse player order, then it must be reversed back to normal
+            if @entity_index == @entities.size - 1 && !@game.draft_first_round_finished
```
ollybh, line 18: Instead of adding the `draft_first_round_finished` to the Game class, you could just check `@reverse_order` and set that to false at the end of the first draft round.

lib/engine/game/g_1847_ae/round/draft.rb
```diff
@@ -0,0 +1,70 @@
...
+
+          def reset_entity_index!
+            @game.next_turn!
+            @entity_index = (@entity_index + 1) % @entities.size
+
+            return if @players_reordered_for_next_round
+
+            # Reorder players for the next round of draft
+            @game.reorder_players
+            @players_reordered_for_next_round = true
```
ollybh, line 64: What's the purpose of this `@players_reordered_for_next_round` check? I can't see how if this will ever be true when it is checked, this method should only ever be called once.
It would also be more consistent with the style of other games to have the `reorder_players` call in `next_round!` instead of here.

lib/engine/game/g_1847_ae/step/draft.rb
```diff
@@ -0,0 +1,145 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G1847AE
+      module Step
+        class Draft < Engine::Step::Base
```
ollybh, line 9: If you extend `Engine::Step::SimpleDraft` instead of `Engine::Step::Base` you get several of these methods defined for you.

lib/engine/game/g_1847_ae/step/draft.rb
```diff
@@ -0,0 +1,145 @@
...
+
+            @log << "#{current_entity.name} has no valid actions and passes"
+            @round.process_action(Engine::Action::Pass.new(current_entity), suppress_log: true)
+          end
+
+          def next_player!
+            @round.next_entity_index!
+
+            pass_if_no_valid_action!
+          end
```
ollybh, line 118: This feels over complicated. Can't you put the `@companies_any?` check in `actions`, returning `[]` if the player can't afford anything. The log message would be sent from `log_skip`. This would mean you wouldn't need to have the almost unmodified version of `process_action` in the round class.

Galatolol: That was the starting point since it's done that way in 18CZ, but I couldn't get that to work. First player for the next round sometimes wasn't properly set if players were skipped and if the first player in draft was to be skipped (which can happen here), the round was ending immediately.

lib/engine/game/g_1847_ae/round/draft.rb
Galatolol: It seems to get called N-1 times where N is the number of players, that's why the check was there. 

I moved the reorder to `next_round!`, good suggestion.

lib/engine/game/g_1847_ae/step/draft.rb
ollybh: > That was the starting point since it's done that way in 18CZ, but I couldn't get that to work. First player for the next round sometimes wasn't properly set if players were skipped. If the first player in draft was to be skipped (which can happen here), the round was ending immediately.

Yes, I see this now. This is because the inherited version of `Round::Base.finished?` isn't finding any active steps, and so is ending the round. I think you're going to need to borrow some of the logic from `Round::Stock` to work out when to move to the next player and when to end the round. Let me think about this…

ollybh: Galatolol/18xx#1 is an attempt to move the code controlling the flow of the round out from the Step class into the Round class. See if that makes sense to you.

Galatolol: Thank you. I'm looking into that.

Galatolol: The proposed solution, with some tweaks, did the trick.

lib/engine/game/g_1847_ae/step/draft.rb
```diff
@@ -0,0 +1,88 @@
...
+
+            @log << "#{player.name} buys #{company.name} for #{@game.format_currency(price)}"
+
+            company.abilities.each do |ability|
+              next unless ability.type == :shares
+
+              # Give related share to the player and give money for the share to the corporation
+              ability.shares.each do |share|
+                @game.share_pool.buy_shares(player, share, exchange: :free)
+                share.corporation.cash += share.corporation.par_price.price * share.percent / 10
```
ollybh, line 63: You should be calling `@game.bank.spend` here, otherwise you are creating new money.

lib/engine/game/g_1847_ae/step/draft.rb
```diff
@@ -0,0 +1,88 @@
...
+            company.owner = player
+            player.companies << company
+            player.spend(price, @game.bank)
+
+            @companies.delete(company)
+
+            @log << "#{player.name} buys #{company.name} for #{@game.format_currency(price)}"
+
+            company.abilities.each do |ability|
+              next unless ability.type == :shares
```
ollybh, line 58: I don't think you want to be allocating the shares at this time. Section 4.5.2 of the rules says that the private companies' investor certificates get exchanged for shares as a purchase action in the stock round, after the corporation has been floated.

lib/engine/game/g_1847_ae/game.rb
```diff
@@ -211,10 +243,17 @@ def init_corporations(stock_market)
         end
 
         def setup
+          # Place L's home station in case there is a "short OR" during draft
+          hex = hex_by_id(l.coordinates)
+          tile = hex&.tile
```
ollybh, line 248: Could `hex` be nil here? If it is, then `tile` will be set to nil, and the next line will throw an error.
I can't see why the call to `hex_by_id` would return nil, so the `&` should be removed.

lib/engine/game/g_1847_ae/step/draft.rb
Galatolol: Those aren't investor certificates but presidencies.

ollybh: OK, my mistake.

ollybh (approved):
I'm now happy with the code in this pull request, but someone who knows the codebase better than me might be able to suggest further improvements.

lib/engine/round/draft.rb
```diff
@@ -6,8 +6,11 @@ module Engine
   module Round
     class Draft < Base
       def initialize(game, steps, **opts)
+        # reverse_order: 4, 3, 2, 1; 4, 3, 2, 1; ...
         @reverse_order = opts[:reverse_order] || false
+        # snake_order: 1, 2, 3, 4; 4, 3, 2, 1; 1, 2, 3, 4; 4, ...
         @snake_order = opts[:snake_order] || false
+        # rotating_order: 1, 2, 3, 4; 2, 3, 4, 1; 3, 4, 1, 2; ...
```
michaeljb, line 13: Great comments here! 👍

michaeljb (approved):
LGTM

Issue #9166, opened 2023-05-07
1817 Loan Payoff Bug


jdbehan:
![1817 Bug - game id](https://user-images.githubusercontent.com/98102209/236697265-02b15133-bb3f-4131-b29f-be31c2530880.png)
![1817 Bug - before paying off one loan](https://user-images.githubusercontent.com/98102209/236697269-b6ea9ddb-d48b-4474-8dcf-28bcff2ff0d3.png)
![1817 Bug - after paying off one loan - loans down to 8 not 9](https://user-images.githubusercontent.com/98102209/236697274-440c5df1-c85e-4453-9a6e-0a1bbfbdbfb2.png)

I found a bug.  In the game above, I pay off one loan, but my loan number drops from 10 to 8.  I have included screen shots.

michaeljb:
https://18xx.games/game/120548?action=1637 - [gist](https://gist.github.com/michaeljb/b9a37dfde8789327aefc6e39201371e3)

crericha:
The view passed in the same loan id twice, causing the company to have two of the same loan object. This happened after a lot of undoing, so it's likely the view was out of sync.

PR #9170, opened 2023-05-09
[1817] Handle loans internally, ignoring loan provided from view

crericha:
Better fix for #9166


crericha:
@michaeljb probably doesn't need pins, but if you don't mind, please check against your local instance to confirm.

michaeljb (approved):

PR #9171, opened 2023-05-09
Allow consent checks for share buying/exchanging

ollybh:
This makes it possible to check for consent when buying shares.

This is needed for 1858. It is possible to exchange a private company for a share in a corporation; if these are owned by different people and the share is coming from the corporation's treasury, then the corporation's president needs to consent to this exchange.

A new method, `consenter_for_buy_shares` is added to the game class, with a default implementation. If consent is needed for a share buy, then this method should return the player whose consent is needed.

- [x] ~~Add the `pins` label if this change will break existing games~~


michaeljb (approved):

ianwilson156:
Given Toby's denial of #9177 - and his decree that formalising consent does not fit the "ethos" of the site - should we have this feature in 1858?

michaeljb:
This consent check works the same as the 1861 merger or cross-buying a train does—it’s just a message that the active player has to click past, the system is not enforcing the consent check.

michaeljb:
Just did a bit more looking through the history since I remembered this having been discussed before, and I found #5133 which changed the wording, so I'm sure we can change the wording again. Perhaps something like this?

"This action requires permission from <player(s)>. Click confirm only if you have already received permission from them."

ianwilson156:
Michael - yes, that's better. But the second sentence may be overly complex? How about:

"This action requires permission from <player(s)>. Click confirm if you have permission from them."

(bearing in mind that English is many players second or third language)

Issue #8178, opened 2022-08-13
1830 : overlapping labels on OO tile 59


Drealmer:
When laying tile `#67` with rotation 5 on H18: the "50" income is hard to read under the "OO" label.
<img width="263" alt="tile_info_overlap" src="https://user-images.githubusercontent.com/1515698/184476208-ff678e9e-0b95-4101-a6d7-0bbc1b49afcf.png">

Issue #7765, opened 2022-05-14
Some problems with tile renderings


perwestling:
Here are the tiles from 18MEX:
![image](https://user-images.githubusercontent.com/2631151/168427977-e9c4d302-d0d7-4d84-aa37-33ccca6fb3e8.png)

Some problems:
- The CC tile is not yellow in the physical game, but rather orange. This is also the color chosen in 18Rhl.
- Wonder if the town in the CC tile should be rendered as it is for tile 58 instead of a circle?
- A couple of tiles have the label and revenue share the same location (475, 479P, 485P, 486P)

perwestling:
CC tile is probably the same mistake as was done for 18Rhl, so that can easily be fixed by correcting the town definition. 
But still remains if CC tile should change color - I am not sure having it orange would be a good idea after all.

Issue #6604, opened 2021-12-08
[1888] Mountain Tile L41 Renders Improperly


Maxwell-Frank:
The company treasury bonus covers the revenue. 
<img width="187" alt="L41" src="https://user-images.githubusercontent.com/75334833/145285751-c1e1f55a-86d2-4642-9ec4-160a9bd43d9d.png">

Issue #5981, opened 2021-07-09
[1828.games] H22 private blocker rendering


benjaminxscott:
The H22 tile (Philly) with it's start of game yellow OO has its blocking private C&A rendered behind the name

Presumably we could affect the private blocker weight to move it to region 17 or something

![674C37A7-FB47-4633-8ECA-C688D8B560B3](https://user-images.githubusercontent.com/1711810/125102584-4d328100-e090-11eb-8e8e-07f1c9295c2a.jpeg)

crericha:
No plans to fix

PR #9175, opened 2023-05-11
add `/tiles/test` route to easily reference tiles from different titles and game states

michaeljb:
* can display a tile as it appears on the board at any given point in any given fixture

    * fixtures are processed to the specified action, and the select tiles are
      rendered; a link to the fixture at the action is rendered in the tile block

* include tiles showing the problems reported in the following issues:

    * #4992

    * #5153

    * #5167

    * #5673

        * copy 18Mag fixture 76622 to a new fixture to reproduce

    * #5981

    * #6604

    * #7765

    * #8178

* add completed 1868 Wyoming game as a fixture

    * 1846 and 1868 Wyoming tiles added here were used for the before/after
      screenshots in #9174

* move fixtures from `/spec/fixtures/` into `/public/fixtures/` so the JSON can be accessed via `/fixtures/<title>/<fixture_id>.json`

    * this is why there are over 200 changed files, most of them are not actually changed

* test cases defined in a location accessible by both the view code and server side code (`lib/engine/test_tiles.rb`) so that when the page is initially served up, it already includes the `<script>` tags for the games being tested

    * test cases need a `title`, can also have `title`, `fixture`, `action`, and
      any `kwargs` used by `render_tile_blocks`

----

Screenshot before fixtures finish processing:

loading

Done:

loaded


assets/app/view/tiles_page.rb
```diff
@@ -246,5 +254,108 @@ def render_toggle_button
...
+      return h(:div, [h(:p, 'Loading...')]) unless @connection
+
+      rendered_test_tiles = []
+
+      test_tiles.each do |title, fixtures|
+        if title
+          game_class = load_game_class(title)
+        else
+          fixtures[nil][nil].each do |hex_or_tile, opts|
+            %i[flat pointy].map do |layout_|
```
tobymao, line 273: why are you calling map here? what are you mapping to?

tobymao (approved):

assets/app/view/tiles_page.rb
michaeljb: In an earlier version I was probably returning the rendered stuff directly instead of `concat`-ing to `rendered_test_tiles`. Changed to `each`.

PR #9182, opened 2023-05-14
[1822 family] set `must_lay_together` to true for "extra tile lay" privates

### LINKED ISSUE #9179: [1822] P12 Extra Tile Lay should not be allowed to be used interspersed with owning Major's tile lays
In other words, `must_lay_together` should be set to `true` and enforced.

* [gist](https://gist.github.com/michaeljb/dd5c40f66cfb7537324c5fa09f738f92)
* [BGG ruling](https://boardgamegeek.com/thread/2425653/article/34793831#34793831)

michaeljb:
The extra tile lay privates should have `must_lay_together: true`, and additionally, this sequence should not be allowed for Majors:

1) lay 1 yellow
2) use the extra tile private
3) lay its second yellow

https://boardgamegeek.com/thread/2425653/article/34793831#34793831

Fixes #9179

### Implementation Notes

* change `G1822::Step::SpecialTrack` to inherit from `Engine::Step::SpecialTrack` instead of `Engine::Step::Base` to inherit some of the behavior for blocking for a company with `must_lay_together` * copy-paste in the `Engine::Step::Tracker` implementation of `hex_neighbors` so that changing the parent class doesn't change this method
* add helper method (which is also used for 1822MX and 1822PNW) to set `@company` appropriately so the step blocks for the company
* `@round.num_laid_track` is incrmented if and only if the Major has already laid one tile; if 0, that means the company is going before them, and if 2, that means the companby is going after, either of which works with Simon's linked BGG ruling

Additionally, add the basic migration (simple cases for adding/deleting pass actions) from #9169, and add in one more case where `prev_action` can be deleted if it is a `pass`; works for a small number of the games broken here, found these ids locally:

```
[87425, 89050, 89915, 92424, 106702, 108057, 111160, 117359, 118300, 118941]
```

-----

If this and #9169 are approved at similar times, I can resolve their merge conflicts in `migrate_game.rb` and then apply just one migration (instead of multiple migrations for multiple deployments).


lib/engine/game/g_1822/step/special_track.rb
```diff
@@ -64,9 +60,13 @@ def process_lay_tile(action)
             ability.use!(upgrade: %i[green brown gray].include?(action.tile.color))
             ability.use! if minor_single_use
 
-            if ability.type == :tile_lay && ability.count <= 0 && ability.closed_when_used_up
-              @log << "#{ability.owner.name} closes"
-              ability.owner.close!
+            if ability.type == :tile_lay
+              if ability.count <= 0 && ability.closed_when_used_up
```
tobymao, line 64: ability.size

tobymao (approved):

lib/engine/game/g_1822/step/special_track.rb
michaeljb: This is not the `count` method which can be replaced by `size` if no block is given, but an attr reader for how many times the ability can be used.

https://github.com/tobymao/18xx/blob/faa7f11d904821f1d53ce2e0ce4ed007fcff5253/lib/engine/ability/base.rb#L24

tobymao: woops

PR #9193, opened 2023-05-20
Add a :future track type for paths

ollybh:
Several games have pre-determined routes on the map where track will be built later in the game:

- Harzbahn 1873 has concession routes for the railway companies.
- 1840 has the route for the Stadtbahn companies.
- 18SJ has three mainline routes.

On the physical game boards for these games these are indicated by lines on the map. 18xx.games currently indicates these routes by having icons in the hexes that form the route, but this is less clear as it does not show the edges that the route enters and exits each hex.

This pull adds a :future track type for path tile parts. This gives a dotted line on the game map, allowing the route to be drawn.

Any path tile parts with a track type of :future are not added to the tile's `paths` array, instead they are stored in a separated `future_paths` array. This means that they will not be considered when calculating routes, and are not checked or preserved when a tile is upgraded.

The pull request adds future track to the maps for 1840 (CC: @Zwergenpunk), Harzbahn 1873 (CC: @roseundy) and 18SJ (CC: @perwestling).

Suggestions are welcome for better naming for this track type, or better representation on the map. I'm just creating this as a draft pull request for the moment, to make sure everyone is happy with this before it gets added.

### 1840 Stadtbahn routes
image

### 1873 railway concessions
image

### 18SJ mainline routes
image

## Before clicking create

- [x] ~~Add the `pins` label if this change will break existing games~~

### DIFF
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -35,6 +35,11 @@ module View
             width: 2,
             dash: '12',
           },
+          future: {
+            color: '#000000',
+            width: 2,
+            dash: '6',
+          },
         }.freeze
 
         # width here is added to track width
@@ -91,7 +96,7 @@ module View
           # Array<Array<Path>>
           @routes_paths = @routes.map { |route| route.paths_for(@tile.paths) }
 
-          paths_and_stubs = @tile.paths + @tile.stubs
+          paths_and_stubs = @tile.paths + @tile.stubs + @tile.future_paths
           path_indexes = paths_and_stubs.to_h { |p| [p, indexes_for(p)] }
 
           sorted = paths_and_stubs
@@ -127,7 +132,7 @@ module View
             elsif path.offboard
               pass1 << h(TrackOffboard, offboard: path.offboard, path: path, region_use: @region_use,
                                         border_props: border_props, **props)
-            elsif path.track == :thin
+            elsif path.track == :thin || path.track == :future
               pass1 << h(TrackNodePath, tile: @tile, path: path, region_use: @region_use,
                                         pass: 1, border_props: border_props, inner_props: inner_props, **props)
 
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -36,9 +36,9 @@ module View
             dash: '12',
           },
           future: {
-            color: '#000000',
-            width: 2,
-            dash: '6',
+            color: '#888888',
+            width: 6,
+            dash: '9 3',
           },
         }.freeze
 
--- a/assets/app/view/game/tile.rb
+++ b/assets/app/view/game/tile.rb
@@ -74,7 +74,9 @@ module View
         children = []
 
         render_revenue = should_render_revenue?
-        children << render_tile_part(Part::Track, routes: @routes) if !@tile.paths.empty? || !@tile.stubs.empty?
+        if !@tile.paths.empty? || !@tile.stubs.empty? || !@tile.future_paths.empty?
+          children << render_tile_part(Part::Track, routes: @routes)
+        end
         children << render_tile_part(Part::Cities, show_revenue: !render_revenue) unless @tile.cities.empty?
 
         children << render_tile_part(Part::Towns, routes: @routes, show_revenue: !render_revenue) unless @tile.towns.empty?
--- a/lib/engine/game/g_1840/map.rb
+++ b/lib/engine/game/g_1840/map.rb
@@ -317,14 +317,62 @@ module Engine
             ['B28'] => 'upgrade=cost:40,terrain:water',
             %w[H10 D18 E17] =>
             'city=revenue:0;city=revenue:0;label=OO',
-            %w[D12 I9] => 'city=revenue:0;upgrade=cost:20;frame=color:#ffa500;icon=image:1840/green_hex',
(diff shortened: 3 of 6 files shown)


perwestling:
I think it looks fine (in 18SJ at least).
Now there is another detail in 18SJ to consider: not having this "future feature" in 18SJ I instead solved it by adding an icon, one for each mainline. This icon persist when you put a tile on a mainline hex in case the tile does not complete the mainline on that hex, and is removed when an upgrade do complete a mainline hex.

This is a problem for 18SJ, if playing the same in person - it can be hard to tell from looking at the board that a hex has an avilable mainline upgrade.

Now, if doing this change request, removing these "mainline icons" might be an idea. One could put them in when doing an upgrade, or maybe remove them all together.

I just got an idea on how to solve this - what if when putting a tile in a mainline hex, an icon would appear showing the type of mainline path that remains to do?  So showing a bend, or a straight? Now this would require some work, so I suppose fixing this in 18SJ would be a separate commit after this feature is added.

Zwergenpunk:
I think it looks awesome. 

I'm not sure If future track is the perfect name, since the tracks, at least an 1840 and 1873, are optional. Maybe Something Like predefined or something. 
But I'm also ok with future tracks if I'm the only one with that problem.

ollybh:
> I just got an idea on how to solve this - what if when putting a tile in a mainline hex, an icon would appear showing the type of mainline path that remains to do? So showing a bend, or a straight? Now this would require some work, so I suppose fixing this in 18SJ would be a separate commit after this feature is added.

It would be fairly easy to keep these these lines on the map when a tile is upgraded: check whether the new tile has track that matches the future track start and end points, and copy it to the new tile if not. But this could make the new tile more cluttered, and it could be showing a track route where this is not a legal tile upgrade. So an icon might be the better solution.

ollybh:
> I'm not sure If future track is the perfect name, since the tracks, at least an 1840 and 1873, are optional. Maybe Something Like predefined or something.

I don't have any attachment to this name, so I'm happy to change it if we can get a consensus on a better name.

I was trying to think of a way of expressing the concept of a route which does not currently have track, but may do at some point in time, and there is a benefit from building this route. I couldn't think of a single word that represented all of this.

The other name I considered using was 'none', as this is a path without any track. But I thought that might be more confusing.

Zwergenpunk:
> > I'm not sure If future track is the perfect name, since the tracks, at least an 1840 and 1873, are optional. Maybe Something Like predefined or something.
> 
> I don't have any attachment to this name, so I'm happy to change it if we can get a consensus on a better name.
> 
> I was trying to think of a way of expressing the concept of a route which does not currently have track, but may do at some point in time, and there is a benefit from building this route. I couldn't think of a single word that represented all of this.
> 
> The other name I considered using was 'none', as this is a path without any track. But I thought that might be more confusing.

Yeah none would be even more confusing ;) maybe future is not so bad at all. It makes clear that that the tracks is not layed yet which is the most important factor.

perwestling:
> I'm not sure If future track is the perfect name, since the tracks, at least an 1840 and 1873, are optional. 
In 18SJ the tracks are optional as well. You just get an extra build action if your first action (or if the extra build action) do add a path to the "future track". So you may very well build across a "future track" hex ignoring the tracks.

So if I were to suggest another name it would be something that pointed towards this being preferred, or recommended, or something. But just "preferred track" is not clear that it is about future builds. So maybe "preferred build path"?

Nah, I do not really have anything better. Maybe "future (track)" is best after all as it is highly likely that it will happen, if it has a benificial effect to build it.

perwestling:
> > I just got an idea on how to solve this - what if when putting a tile in a mainline hex, an icon would appear showing the type of mainline path that remains to do? So showing a bend, or a straight? Now this would require some work, so I suppose fixing this in 18SJ would be a separate commit after this feature is added.
> 
> It would be fairly easy to keep these these lines on the map when a tile is upgraded: check whether the new tile has track that matches the future track start and end points, and copy it to the new tile if not. But this could make the new tile more cluttered, and it could be showing a track route where this is not a legal tile upgrade. So an icon might be the better solution.

I had some bad experience of modifying tiles dynamically. But I suppose it was just to ensure that any removal of the tile from the board clean up.

Adding a "future track" to a tile would be really nice, although I have not tried how it looks like.

But for a physical game I think a token might be the best, and then it might just be to mimic this in the online game?

Anyway, I intend to look into this in the future (!). There are at least one other thing I have to fix in 18SJ so I can look into that then. So I think the icons can be kept for now as they are. In other words, from 18SJ point of view it is fine to proceed with your commit without doing any changes.

roseundy (commented):
This seems OK to me.

For 1873, "future" is as good as anything ("concession route" is too long).

I'm curious why you didn't subclass Engine::Part:Path

Did you test the UI at all with the affected games?

roseundy:
Also, for 1873, I would _not_ want the future track to be copied to tiles place on those map hexes. The physical game doesn't require that and It would make the tiles more confusing

ollybh:
> I'm curious why you didn't subclass Engine::Part:Path

I tried that approach too: https://github.com/ollybh/18xx/tree/shadow-paths. I preferred just having a different track type in the existing Path class, it's a little bit simpler.

> Did you test the UI at all with the affected games?

I did with 1873, laying track that followed the concession route and also laying tiles in different directions.

ollybh:
I've converted this pull request from draft to a full pull request. Everyone who has commented seems to think this is a good idea, and I've not yet had any requests to change any implementation details.

PR #9200, opened 2023-05-23
1841: corporation start operations (stock purchase, token placement, …

roseundy:
### Implementation Notes

* **Explanation of Change**

No common code changes.
This implements the 41-specific changes needed when starting a corporation
- After a Par action, allows president to purchase up to 40% of shares, then
- Either auto-places tokens for historical corps, or has the president select token locations (from a set of locations that varies by major/minor, phase, and whether one or two tokens are being placed), then
- Has the president select the number of tokens to buy, where the price is fixed for historical corps or determined by the proximity of the laid token(s) to existing track on the map

* **Screenshots**

* **Any Assumptions / Hacks**


lib/engine/game/g_1841/game.rb
```diff
@@ -363,6 +413,88 @@ def place_home_token(corporation)
...
+          end
+        end
+
+        def hex_connected?(hex)
+          !hex.tile.preprinted ||                                  # any tile has been laid here
+            (!hex.tile.paths.empty? && !hex.tile.cities.empty?) || # pre-printed with track and a city
+            (!hex.tile.paths.empty? && tile_laid_at_exit?(hex)) # pre-printed with track and tile connecting to it
+        end
+
+        def search_hexes(current_hex, hex_list, start_hex)
```
michaeljb, line 448: If I'm understanding correctly, this `search_hexes` will need to be called for token prices to be computed whenever a home token is placed. Is there anything that can be done to mitigate this computation when loading the game, e.g., auto actions for 1822 destination tokens?

roseundy: I thought about optimizing it, but:
1. it's really not very computationally expensive. It's only traversing a maximum of 19 hexes.
2. It will only happen at most once or twice per corporation the entire game.

If you think it will slow down the game, I should be able to skip it when loading.

roseundy: Actually, only doing when not loading is problematic, since the second home token placement may be skipped depending on where the first was placed and that requires doing the search.

I still stand by my statement that this isn't computationally expensive - it's _much_ faster than building a graph.

roseundy: Ignore the last comment, I forgot this wasn't token placement, but token purchase. I could easily change the code such that the token cost is only calculated when not loading. It means that the log won't show the price - only the UI.

roseundy: And...once again, ignore the last comment. I forgot that the choose action doesn't supply enough information (unless I do something ugly like encode both the number of tokens and the price in a single integer).

So I'm back to my assertion that this isn't a performance hit. Also, the token prices only need to be calculated for non-historical companies (about half of them).

michaeljb: Sounds good! Seeing something called `search_hexes` just sent up a red flag in my mind, but since it's a much less big deal than building the whole graph I don't think we need to worry about it. 🙂

michaeljb (approved):
LGTM

PR #9206, opened 2023-05-27
Disable create game button when incompatible optional rules are selected

michaeljb:
Old behavior with bad set of optional rules selected:

* multiplayer: can create a game table, no error until starting the game
* hotseat: taken to error page
* JSON: taken to error page

New behavior with bad set:

* multiplayer and hotseat: 'Create' button is disabled
* JSON: error message flashes and user stays on the page

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -32,7 +32,6 @@ module View
 
     def render_create_button(check_options: true)
       error = check_options &&
-              selected_game_or_variant.respond_to?(:check_options) &&
               selected_game_or_variant.check_options(@optional_rules, @min_players, @max_players)&.[](:error)
       (render_button('Create', { style: { margin: '0.5rem 1rem 1rem 0' }, attrs: { disabled: !!error } }) { submit })
     end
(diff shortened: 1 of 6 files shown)


assets/app/view/create_game.rb
```diff
@@ -408,6 +421,13 @@ def submit
         game_data[:settings][:seed] = game_params[:seed] if game_params[:seed]
       end
 
+      checked_options = Engine.meta_by_title(game_data[:title])
+                          .check_options(game_data[:settings][:optional_rules],
+                                         game_data[:min_players], game_data[:max_players])
```
roseundy, line 426: Why doesn't this need a .respond_to? guard clause?

michaeljb: On the branch where I'm fixing the 1822PNW variants I added it to the base meta class, I tried splitting it out from this branch but that should be here too. And I guess with that method I can get rid of the other `.respond_to(:check_options)` here. What do you think? b89fe89

roseundy: Well, if check_options is now in the base class, then you should definitely get rid of all the repond_to guards.

roseundy (approved):
Looks good to me

PR #9208, opened 2023-05-27
[1822PNW] fix L/2 train roster variants

michaeljb:
The current variants for reducing the number of L/2 trains are not functional. This PR replaces the four non-functional variants with two new ones that match the variants in the latest version of the rulebook.

* four variants were implemented in #8380, merged on 2022-09-26
* the variants were announced [on BGG][0] on 2022-10-03
* the implementation for the variants was deleted in #8443, merged on 2022-10-15, but the `OPTIONAL_RULES` remained in `G1822PNW::Meta`; since this was deployed, 1822PNW games could be created with a variant for fewer L/2 trains selected, but the setup was not actually changed
* the latest (dated 2023-02-22) [draft rules][1] show variants for removing two or three L/2 trains (Chapter 11, page 28)
* since this implementation uses different names for the optional rules, old games created with the non-functional optional rules will not need pins

[0]: https://boardgamegeek.com/thread/2945245/how-many-l2-trains
[1]: https://boardgamegeek.com/thread/2890404/1822pnw-public-rules-review

----

This cannot be deployed without https://github.com/tobymao/18xx/pull/9206, but they didn't have overlapping code so I split them up

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -384,6 +384,13 @@ module Engine
         optional_rules
       end
 
+      def check_optional_rules!
+        min_players = @players.size
+        max_players = @players.size
+        error_string = meta.check_options(@optional_rules, min_players, max_players)&.[](:error)
+        raise OptionError, error_string if error_string
+      end
+
       def setup_optional_rules; end
 
       def log_optional_rules
@@ -568,8 +575,9 @@ module Engine
 
         init_company_abilities
 
-        setup_optional_rules
+        check_optional_rules!
         log_optional_rules
+        setup_optional_rules
         setup
         @round.setup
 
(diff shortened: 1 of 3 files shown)


michaeljb:
Tests were breaking since they actually depended on #9206

lib/engine/game/base.rb
```diff
@@ -384,6 +384,13 @@ def init_optional_rules(optional_rules)
         optional_rules
       end
 
+      def check_optional_rules!
+        min_players = @players.size
+        max_players = @players.size
+        error_string = meta.check_options(@optional_rules, min_players, max_players)&.[](:error)
```
michaeljb, line 390: @roseundy I added this to the `setup` so that for 1822PNW I didn't need to duplicate the optional rule compatibility logic in the meta and the game class. Is this useful for 1825? Or should I move this out of the base game and only into `G1822PNW`?

roseundy: I'd leave it in the base, the way you have it. I do dislike that 1825 is checking options in multiple places.

roseundy (commented):
Looks good to me.

PR #9211, opened 2023-05-27
[1822 family] allow combining `:tile_lay` private company abilities

### LINKED ISSUE #4469: 1822 - Cannot simultaneously use P2 + P10 for a single track lay
https://18xx.games/game/32376

When a major or a minor has multiple private companies it appears that it is impossible to use more than one power at a time.

Specifically this is going to come up frequently with privates P2 + P10.

I have encountered it in an earlier game today.

michaeljb:
* `Action::TileLay` has a new property, `combo_entities`, to indicate the other companies who are being combined with the primary entity on the action; represented in the JSON form of the action as an array of company ID strings

    * it should not make a difference which company is the primary one on an
      action referenced by `entity`, and which are referenced in `combo_entities`

* in the ability configuration, `combos` is a list of strings of company IDs, to indicate which other companies can be combined with the current ability

* in `View::Game::Abilities`, if the `@selected_company`'s tile-laying ability has any `combo_entities`, a new "Combine with" section is rendered, where other companies can be selected by checkbox to join their powers

    * if one company is selected via checkbox that is incompatible with any
      other checkbox-option companies, those other checkbox options will be
      greyed out and unselectable

    * if `@combos_only` is true, then only the combos section will be rendered, instead of the whole abilities view; this is used when the "Extra Tile Lay" privates take over the turn

* various methods in some 1822 and some base classes needed a refactor to work with one entity or many; for backwards compatibility and to avoid rewriting methods for many game classes, the affected methods first have a couple of lines where the given `entity` or `entities` is converted to a single `entity` and an array of `combo_entities`; sometimes the method logic requires acting on all of the `entities` at once, so the `entities` variable contains all of them (for instance, to find if any of the given entities make a tile lay free)

* implement combos for all 1822 family games, add fixtures where the combos happen

Fixes #4469

----

### Quad Combo

Use combos to lay an extra upgrade (P12), to a color beyond the current phase (P11), getting the estuary crossing for free (P10), and removing the town (P2), all in one action.

Map - Before
quad combo before

Map - After
quad combo after 1

Log - Before
quad combo before 2

Log - After
quad combo after 2

Selecting all the companies to combine; after selecting a company the others it can combine with are visible under "Combine with:", and it doesn't matter which one is the "primary" one
quad combo before 3

### Combo conflict

P8 cannot be combined with P10 and P11, but could still be combined with P12
Screenshot 2023-05-27 002221

### Action Data

Whichever company is the primary `entity`, the action should be processed in the same way; these are functionally equivalent:

action A

action B

### Extra Tile Lay, Available Hexes

After P12 lays its first of up to two extra yellow tiles, `SpecialTrack` becomes a blocking step. P12 can still combo with other companies, and selecting something to combine with changes which hexes are highlighted as available.

p12 1
p12 2


roseundy (commented):
I'm not an ability expert, but seems OK to me.

tobymao (approved):

PR #9223, opened 2023-06-01
adding map and entities file to 1828

philcampeau:
### Before clicking "Create"

### Implementation Notes

Added map.rb and entities.rb file to 1828, as per style guide.


michaeljb (approved):
LGTM, passes validation

PR #9225, opened 2023-06-01
1858 map display

ollybh:
# Before clicking "Create"

# Implementation notes

The map for 1858 was cluttered in places, with various tile parts obscured. Several changes have been made to make it clearer:

1. Future track paths have been added to show the routes of the private railway companies.
2. Blockers have been hidden for the hexes where there is a future track paths (they are kept the the locations where the private railway company only has a single home hex).
3. Stubs are hidden (rendered as future track) as they are not adding any information above what is shown by the future paths.
4. Where there are  two icons in a hex these have been split up. This stops them from being rendered on top of other tile parts.

## Core changes

This has a couple of small changes to core code, to allow the rendering of stubs to be changed. Previously these were always rendered as broad gauge track. Now the stub’s track type can be set when the stub is created. This does not have any effect on the stub’s behaviour, just how it is rendered.

### 1858 starting map after these changes
image

### 1858 starting map before changes
image


michaeljb (approved):
Looks good!

PR #9227, opened 2023-06-02
[1828] place blocking token in correct city on hex

### LINKED ISSUE #9202: 1828.Games - Blocking token placed incorrectly in NY hex during system merger
Game in question: [`https://18xx.games/game/121897?action=706#map`](https://18xx.games/game/121897?action=706#map)

Observation:
Upon merging NYH and B&M to form a system, the B&M token was replaced with a blocking token, but that token was placed in the wrong segment of hex G23.

Expectation:
Since both the NYH and B&M tokens were in the northern half of hex G23, the blocking token should also have been placed in the northern half. If I had one had been in the north and one in the south, I would have expected a prompt asking which to replace.

Relevant Rule 7.5.2.3: 

> * If both merging public companies had station markers in the same hexagon:
> · One of the two station markers must be replaced with a blocking station marker (active player’s choice) and the removed station marker is returned to the new system’s charter.
> 

Please let me know if more information is needed

Before system merger:
2023-05-25 12_07_22-_ 1828 Games - 121897 - 18xx Games - Brave

After system merger:
2023-05-25 12_07_47-_ 1828 Games - 121897 - 18xx Games - Brave

crericha:
Fixes #9202

### DIFF
--- a/lib/engine/game/g_1828/game.rb
+++ b/lib/engine/game/g_1828/game.rb
@@ -1130,7 +1130,7 @@ module Engine
           minor.tokens.each do |token|
             city = token&.city
             token.remove!
-            place_blocking_token(city.hex) if block && city
+            place_blocking_token(city.hex, city: city) if block && city
           end
           @graph.clear_graph_for(minor)
           @minors.delete(minor)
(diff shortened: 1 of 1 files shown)


crericha:
May require pins if the location the blocking token was *supposed* to be was later tokened by company.

tobymao (approved):

PR #9230, opened 2023-06-02
1860: Implements Re-use Of Hexes variant

philcampeau:
### Before clicking "Create"

### Implementation Notes

Implements the Re-use Of Hexes variant from the official rules, allowing trains to reenter a hex as long as no track is reused. 

image


roseundy (approved):
Looks good.

PR #9251, opened 2023-06-07
1841: Corporation share operations, corps can ipo corps, handle circular ownership

roseundy:
### Implementation Notes

* **Explanation of Change**

Adding the bulk of corporation share operations, including:

1. Corporations can buy stock in other corporations, and can become president
2. Corporations can sell stock in other corporations and sell their own IPO shares
3. Corporations can start new corporations
4. Corporations can only hold 5 non-IPO certificates
5. Implements "Anti-Trust Law" - a player cannot control more than 60% of a stock - both directly and indirectly
6. Handles circular chains of control (i.e. "frozen" corporations). This only happens after EMR

* **Common Code Changes**

Because corporations owning corporations is a new concept, supporting it required a fair bit of common code changes
1. **Engine: Corporation** - added parameter to allow corporations to be treated like player share holders. This is needed for presidential transfers to work correctly with corporation ownership
2. **Engine: SharePool** - call new game method "possible_presidents" instead of players. Again to allow presidential transfers to work correctly
3. **Engine: Ownable** - rewrite "player" method to avoid infinite recursion with circular ownership
4. **Engine: Game: Base** - add "corporations_can_ipo?", "possible_presidents" and "receivership_corporations" base methods
5. **UI: CorporateBuySellShares** - new view to allow corporate share operations
6. **UI: Corporation** - Allow corporation owners of stock to be listed like player owners if "corporations_can_ipo?" is true
7. **UI: Entities** - Use more general game definition of receivership "receivership_corporations"
8. **UI: Round: Operation** - Show new CorporateBuySellShares view for games that allow corporations to ipo.

* **Screenshots**

Corporations owning Corporations owning Corporations:
1841_subsidiaries

* **Any Assumptions / Hacks**


lib/engine/ownable.rb
```diff
@@ -10,8 +10,19 @@ def owned_by?(entity)
       owner == entity || owner&.owner == entity || owner == entity&.owner
     end
 
+    # avoid infinite recursion for 1841
     def player
-      owner&.player? ? owner : owner&.player
+      owner_chain = [owner]
```
tobymao, line 15: what is this? i can't really understand it

tobymao: maybe call it 

owners = {owner: true}

lib/engine/ownable.rb
```diff
@@ -10,8 +10,19 @@ def owned_by?(entity)
...
+      test_owner = owner
```
tobymao, line 16: current = owner

lib/engine/ownable.rb
```diff
@@ -10,8 +10,19 @@ def owned_by?(entity)
...
+      until test_owner&.player?
+        return nil unless test_owner&.owner
```
tobymao, line 18: so you're going until the current owner is a player, but are you expecting current owner to be nil? why is that?

roseundy: OK, so this is the original code for this method:
`owner&.player? ? owner : owner&.player`

**The purpose of the method is to return the player that ultimately controls (owns) the object (self) or nil if the ultimate owner is not a player.**

The way it works is if the owner is a player, then then it returns that player, if not, it recursively calls the method on the owner. For players, this is redefined to return self. For non-player ultimate owners (e.g. the bank or the share pool), this method is redefined to return nil.

I had to modify this for 1841, because 1841 1) allows corporations to own corporations and 2) "circular" ownership is legally allowed to occur (usually only in EMR scenarios). Circular ownership, (i.e. a cycle) results in infinite recursion with the original code. So I had to modify this.

Basically, I converted a recursive method into a iterative one so I could detect a cycle. If a cycle is detected I return nil. I return the owner if the owner is player. I return nil if not owned. I originally thought about coding this to be iterative for corporations and recursive otherwise, but I thought that was clunky.

roseundy: I could change L18 to be "return test_owner.player unless test_owner&.owner" so it becomes recursive if not owned.

lib/engine/ownable.rb
roseundy: OK, although maybe:

chain = {owner: true}

since a "chain of ownership" is a concept the 1841 rules specifically mention.

lib/engine/ownable.rb
roseundy: Yeah, that's better.

lib/engine/ownable.rb
roseundy: I've revised this. See latest commit.

roseundy:
I've reduced the blast radius of the player method change - it now applies to corporations only.

tobymao (approved):

PR #9253, opened 2023-06-09
[18NL] initial commit

philcampeau:
### Before clicking "Create"

### Implementation Notes

Initial commit for 18NL (still pre-alpha)

To do:

- [x]  fix ability for P2 (should not be able to lay token on a hex with track)
- [x]  price for P2 drops if everyone passes in initial auction.
- [ ]  If P2 is used to token H13, the hex can't be upgraded until NZO has chosen its home station on that hex.

### DIFF
--- a/lib/engine/game/g_18_nl/entities.rb
+++ b/lib/engine/game/g_18_nl/entities.rb
@@ -1,0 +1,198 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G18NL
+      module Entities
+        COMPANIES = [
+          {
+            name: 'Amsterdamsche Omnibus-Maatschappij',
+            sym: 'P1',
+            value: 20,
+            revenue: 0,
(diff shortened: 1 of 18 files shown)


lib/engine/game/g_18_nl/entities.rb
```diff
@@ -0,0 +1,198 @@
...
+    module G18NL
+      module Entities
+        COMPANIES = [
+          {
+            name: 'Amsterdamsche Omnibus-Maatschappij',
+            sym: 'P1',
+            value: 20,
+            revenue: 0,
+            desc: 'Company may build an extra tile on F5 with a ƒ40 discount. Blocks F5 while owned by a player.',
+            abilities: [{ type: 'blocks_hexes', owner_type: 'player', hexes: ['F5'] }],
```
ollybh, line 14: This private also needs a `tile_lay` ability.

ollybh (changes requested):
Most of this looks fine. Just a few small things to fix.

ollybh (approved):

PR #9255, opened 2023-06-10
fix issue importing hotseat games

benjaminxscott:
Fixes issue importing hotseat games

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -398,7 +398,7 @@ module View
         game_params[:max_players] = @max_p[title] if game_params[:max_players].to_i <= 0
         game_params[:seed] = game_params[:seed].to_i
         game_params[:seed] = nil if (game_params[:seed]).zero?
-        return create_game(game_params) if @mode == :multi
+        return create_game(game_params) 
       end
 
       players = game_params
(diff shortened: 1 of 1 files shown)


tobymao (approved):

assets/app/view/create_game.rb
```diff
@@ -392,12 +392,14 @@ def submit
...
-      game_params[:seed] = game_params[:seed].to_i
-      game_params[:seed] = nil if (game_params[:seed]).zero?
 
-      return create_game(game_params) if @mode == :multi
+      if @mode == :multi
+        title = selected_game_or_variant.title
+        game_params[:max_players] = @max_p[title] if game_params[:max_players].to_i <= 0
+        game_params[:seed] = game_params[:seed].to_i
+        game_params[:seed] = nil if (game_params[:seed]).zero?
+        return create_game(game_params) if @mode == :multi
```
perwestling, line 401: Duplicated if as line 396 already assures mode is multi

perwestling:
Tried out this change on latest master, and got another error.
image

assets/app/view/create_game.rb
tobymao, line 401:
```suggestion
        return create_game(game_params) 
```

PR #9269, opened 2023-06-16
[1822] Fixes Hill Discount private

### LINKED ISSUE #9268: [1822MRS] Continuous Hill Discount is applied to rough terrain, but shouldn't be
The Hill Discount seems to discount rough terrain, when it shouldn't. 
My assumptions are that tile I24 (green dome costing 40) is rough terrain and green domes costing 60 are hills.

**How to reproduce:**
In game: https://18xx.games/game/125682?action=389 LYR acquires the Hill Discount, chooses the continuous discount and lays H23 for the regular 40.
Afterwards, it can lay a tile on I24 for 20. However, this should be 40 as I24 is rough terrain and the continuous discount doesn't apply.

philcampeau:
Fixes #9268 

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

P8 is supposed to either pay the full cost of one rough terrain, hill, or mountain cost, or provide an ongoing discount to hills and mountains only. However, P8 is currently programmed to apply the ongoing discount to `terrain:hill` and `terrain:mountain`, but the map.rb file uses `cost:40,terrain:hill` for rough terrain, in order to have it look more like the physical game. 

This change makes it so that the discount applies only to hexes with a 60-cost hill, as well as all mountain hexes. I also updated the language in the private description to indicate this so that we don't get a wave of bug reports. 

* **Screenshots**
image

image


philcampeau:
Hmm that's strange it passed on my end. I must have hit a keystroke by accident. I'll check when I get back home.

michaeljb (approved):
Approved, when someone (maybe me later this week) is able to deploy and pin they can merge.

michaeljb:
Thanks Phil!

PR #9294, opened 2023-07-07
1841: implement mergers

roseundy:
### Implementation Notes

* **Explanation of Change**

This implements the merger mechanism in 1841

Common code changes:
- View::Round::Merger - replaced <entity>.owner with <entity>.player to allow 1841's corporations owning corporations to work
- Game::Engine::Part::Path - fixed issue where skip_paths wasn't passed to subsequent calls to path.walk


tobymao:
sorry, what is the rule in 1841 that requires request undo? why can't you just use the chat box for this?

roseundy:
When performing a merger, there is the possibility that no player or corporation becomes the president of the new corporation. The rules say that the merger must be unwound in that case. Unfortunately, there is no way to detect this case beforehand. So when I detect that it happens, I stall the game and request an undo.

As an alternative, I could just raise an error on the last action before I detect that there is no president. It might be confusing for the player, since all they potentially did was try to cash out a partial share.

tobymao:
yea lets get rid of the complexity and just raise an error with the appropriate message and instruct the players to manually undo

roseundy:
OK

tobymao (approved):

PR #9299, opened 2023-07-11
[1871]  autobuy from market to float

### LINKED ISSUE #9282: 1871 TOP auto buy until float broken
https://18xx.games/game/126153?action=417

When choosing "until float"

Screenshot 2023-06-27 at 07 19 14

you get this error

Screenshot 2023-06-27 at 07 19 33

In this case, "until float" should buy the shares from the market.

benjaminxscott:
Fixes #9282

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

1 - Changes `from_market` to also be writable
2 - Add an override in 1871 to float via a purchase from market

* **Screenshots**

On local, it's `testman3` turn, who sets an auto-action which is immediately executed as it's his turn

Setting auto-action
Screenshot 2023-07-11 4 05 30 PM

Execution of auto-action
Screenshot 2023-07-11 4 05 01 PM

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/action/program_buy_shares.rb
+++ b/lib/engine/action/program_buy_shares.rb
@@ -6,7 +6,8 @@ require_relative 'program_enable'
 module Engine
   module Action
     class ProgramBuyShares < ProgramEnable
-      attr_reader :corporation, :until_condition, :from_market, :auto_pass_after
+      attr_accessor :from_market
+      attr_reader :corporation, :until_condition, :auto_pass_after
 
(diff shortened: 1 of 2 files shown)


benjaminxscott:
Clarification that players will always want to buy from the market:

> TheCat: The reserved share gets exchanged immediately on float, so there's no way for the shares to be sold out before the exchange.

philcampeau:
\o/ \o/

PR #9301, opened 2023-07-12
[1856] fixes train limit increase for CGR

philcampeau:
### Before clicking "Create"

- [x ] Branch is derived from the latest `master`

### Implementation Notes

For some reason the train limit increase for the CGR was set to 99 instead of 1. Given how CGR can only buy trains at face value, this is likely unimportant, but regardless, it should be fixed.


tobymao:
@ventusignis can you look at this?

PR #9392, opened 2023-08-13
1841: deal with pres cert xfer during merger

### LINKED ISSUE #9389: [1841] Child Company Share Disappeared During Tuscan Merge
Game ID, with action: https://18xx.games/game/130760?action=349

Entering Tuscan Merge: SSFL owns 30% of IFAI, no other shareholders of IFAI; SSFL is started, but SFMA is not (status as of action 348). On Tuscan merge, 2 shares of IFAI are moved into SFLi - this should be 3, since SSFL owns 3. After the Tuscan Merge, 2 shares of IFAI are in SFLi, 7 are in IFAI IO, and that's it - one share is missing from the game.

Side note: "dropped a share on the ground and kicked it under the sofa during the Tuscan merge" is a very realistic feature to have in the game, but I'm guessing it was not intended :)

roseundy:
Fixes #9389

### DIFF
--- a/lib/engine/game/g_1841/game.rb
+++ b/lib/engine/game/g_1841/game.rb
@@ -1172,15 +1172,6 @@ module Engine
           @share_pool.transfer_shares(share.to_bundle, new_owner, allow_president_change: false)
         end
 
-        # no payment, only for asset movement
-        def asset_transfer_shares(bundle, new_owner)
-          @share_pool.transfer_shares(bundle, new_owner, allow_president_change: false)
-          return unless bundle.presidents_share
-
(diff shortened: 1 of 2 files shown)


tobymao (approved):

roseundy:
I've reconsidered my point solution and am changing to a more general solution.

It requires touching SharePool to check for the edge case where a president's share swap isn't needed because the president's share is already moving from from the old president to the new president. It kinda worked in SharePool before - but I think more by chance, and it didn't work when more than the president share was being transferred (which is what caused the 41 bug).

tobymao:
looks good but may wanna check all games

roseundy:
Checked all games. Only new failure was the 1841 game that found the bug.

Issue #9419, opened 2023-08-17
1822Africa Checklist


vandamm:
The game is being implemented based on these documents:
- [Map](https://boardgamegeek.com/image/7682366/scottredracecar)
- [Rules differences from 1822](https://docs.google.com/document/d/1tgRiGEqSKZSQrJ92Wq-rbb0kS2xBZAr0hao3RMjKqcI/edit)
- [Private companies](https://docs.google.com/spreadsheets/d/1yX8NeNFsgTf7dLbzZ69ZnESZvaWC_3MlGyvqZKPDEtI/edit?usp=sharing)
- [Phases](https://github.com/tobymao/18xx/pull/9414#issuecomment-1681402941)

#### From 1822
- [x] P1 - Permanent L-train
- [x] P2, P3 - Permanent 2-train
- [x] P9 - Remove Town
- [x] P13 - Station Swap
- [x] P16 - A10x Phase

#### From 1822MX
- [x] P4, P5 - Pullman
- [x] P7 - Extra tile

#### Unique
Most of these privates require a lot of work so they will be implemented after everything else
- [x] P6 - Recycled train
- [x] P8 - Reserve Three Tiles
- [x] P10 - Game Reserve
- [x] P11 - Mountain Rebate
- [x] P12 - Fast Sahara Building
- [x] P14 - Gold Mine
- [x] P15 - Coffee Plantation
- [x] P17 - Bank Share Buy
- [x] P18 - Safari Bonus

## Map
- [x] Custom major company logos
- [x] Game reserves - Special map hex/tile that does not count against the train number (can’t be run with E).

## Stock market
- [x] Minor companies may not advance beyond 90 value. 
- [x] If they would otherwise move right, they remain in the same space and retain their position in the stack.

## Stock round
- [x] Single queue with only three bid boxes
- [x] There are two back-to-back  stock rounds each time there would normally be an 1822 stock round.
- [x] Before each of the first two SRs, find the top-most concession and place all cards before it at the bottom of the deck (preserving the order)
- [x] SR End: Clear bid box cards and export trains at the end of each round like in 1822.
- [x] SR End: Additionally, if there is a concession in bid box 1, clear it (but don’t remove a train). This major is no longer available to start from a concession.
- [x] SR End: If no cards were bought from the bid boxes during the SR, remove each of the cards in the bid boxes from the game before restocking (but after train exporting, if any).

## Operating Round
[...]

PR #9414, opened 2023-08-16
[1822Africa] Implementation part 1

vandamm:
- [x] Branch is derived from the latest `master`
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

### Implementation Notes

* **Explanation of Change**

This is a foundation for implementation of 1822Africa, requested by designer Scott Petersen. All game information can be found in this [BGG thread](https://boardgamegeek.com/thread/3131296/1822africa-design-diary)

[Rules differences from 1822](https://docs.google.com/document/d/1tgRiGEqSKZSQrJ92Wq-rbb0kS2xBZAr0hao3RMjKqcI/edit)
[Privates list](https://docs.google.com/spreadsheets/d/1yX8NeNFsgTf7dLbzZ69ZnESZvaWC_3MlGyvqZKPDEtI/edit?usp=sharing)

* **Screenshots**
Screenshot 2023-08-17 at 00 42 29

PR #9425, opened 2023-08-18
[1822Africa] Implementation p2

vandamm:

### Implementation Notes

* **Explanation of Change**

Continuation of #9414. Progress can be tracked in #9419


lib/engine/game/g_1822_africa/game.rb
```diff
@@ -252,35 +335,47 @@ def setup_companies
...
+          head = @companies[0...next_concession_index]
+          tail = @companies[next_concession_index..-1]
 
-        def put_concession_to_front(companies)
-          first_concession_index = companies.find_index { |c| c.id[0] == self.class::COMPANY_CONCESSION_PREFIX }
+          @companies = tail + head
 
-          head = companies[0...first_concession_index]
-          tail = companies[first_concession_index..-1]
+          unowned_head = head.filter { |c| !c.owner || c.owner == @bank }
```
michaeljb, line 359:
```suggestion
          unowned_head = head.select { |c| !c.owner || c.owner == @bank }
```

lib/engine/game/g_1822_africa/game.rb
```diff
@@ -292,57 +387,128 @@ def timeline
...
+        def stock_round(round_num = 1)
+          G1822Africa::Round::Stock.new(self, [
+            Engine::Step::DiscardTrain,
+            G1822::Step::BuySellParShares,
+          ], round_num: round_num)
+        end
+
+        def total_rounds(name)
+          @operating_rounds if name == self.class::OPERATING_ROUND_NAME
+          2 if name == self.class::STOCK_ROUND_NAME
```
michaeljb, line 451: In ORs this method will return `nil`.

This should use a `case name` block. https://rubystyle.guide/#case-vs-if-else

lib/engine/game/g_1822_africa/game.rb
```diff
@@ -292,57 +387,128 @@ def timeline
...
-          # End if bid boxes cannot be refilled AFTER SR
+        def compute_game_end
+          return %i[custom full_or] if bidbox.length < self.class::BIDDING_BOX_MINOR_COUNT
+          return %i[stock_market current_or] if @stock_market.max_reached?
         end
 
         # Temporary stub
         def setup_exchange_tokens; end
 
         # Stubbed out because this game doesn't it, but base 22 does
```
michaeljb, line 508:
```suggestion
        # Stubbed out because this game doesn't use it, but base 22 does
```

michaeljb (changes requested):
Thanks for the contribution!

As stated in the default PR description, when implementing a new game it's best to keep the changes smaller. I don't mind seeing boiler-plate type things grouped together (so entities, new icons, constant values, base 1822 overrides would have all been fine), but it would have been better for the new logic implementing the special rules of this title to have been split out into other PRs.

No need to restructure that stuff right now, but keep that in mind in the future. I'll merge when my other comments are addressed.

lib/engine/game/g_1822_africa/step/dividend.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Game
+    module G1822Africa
+      module Step
+        class Dividend < Engine::Game::G1822::Step::Dividend
+          def share_price_change(entity, revenue = 0)
+            return { share_direction: :left, share_times: 1 } unless revenue.positive?
+
+            price = entity.share_price.price
+
+            times = (revenue.to_f / price).floor
```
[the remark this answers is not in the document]
scottredracecar: Minors work like other 1822 games, always pay 50/50, always move up (or down) exactly one space

vandamm:
@michaeljb thanks for the review! I’m sorry for submitting such a big PR, but I’ve tried to keep it structured per commit. I’ll make sure to stick to smaller ones in the future. 
I’ll fix your remarks tomorrow.

lib/engine/game/g_1822_africa/game.rb
vandamm: Sorry for this silly mistake! I learned some  Ruby specifically to contribute here so I may miss some conventions or make mistakes :)

michaeljb: If you do ever have a function similar to this but the `if`s are checking different conditions, what you have would be fine if the first line just started with a `return`. You could make both/all lines start with a `return` and let rubocop clean up the last `return`.

lib/engine/game/g_1822_africa/game.rb
vandamm: I'm changing it though I don't quite get the difference, they look identical to me

michaeljb (approved):
Looks good, thanks!

PR #9463, opened 2023-08-25
[1888] fixes market to match final production copy

### LINKED ISSUE #9319: [1888] The share price board
image
image

The board is different for 18xx.games and the real game.

philcampeau:
Fixes #9319

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
Changed market to match with the physical production copy. Also removed the fixture game which no longer passed rake and added two new ones that do. 

We'll need to pin any games that had corps enter the market spaces that are no longer yellow.

* **Screenshots**
Current market on site: 
image

Market on physical copy:
image

* **Any Assumptions / Hacks**


philcampeau:
Note that while the physical copy shows a brown colour for those spaces, the rules confirm that these work as 1830's yellow space.

roseundy (approved):
Looks good.

Issue #9376, opened 2023-08-11
1822CA checklist


michaeljb:
### Alpha blockers (last updated November 30)

- [x] GNWR and NTR destn hexes are explicit city slots. GNWR is the north edge slot, NTR is the SE
- [x] ~[game 13, WRS] autopass was pressed in SR3 for a player during the auction, but the auction came back around and the player as not autopassed.~ A Major was started, so this is actually WAI
- [x] [game 16, ERS] trying to lay tile Q1 in Quebec in OR3.1. getting an "undefined method. exchange token for nil" error. it was finicky. had to place the QMOO home token in the same location reserved for M5, and then was able to place the Q1 tile
    - if QMOO's home is different from M5's, maybe it needs to become a pending token again when a yellow tile is laid instead of it automatically going to one of the new cities
- [ ] [game 17, WRS] SR6 - $0 bids on privates disappearing
- [x] [game 17, WRS] OR 7.1 - GNWR incorrectly prompted about choosing home (related to 18's home being different in WRS than in full game)
- [x] [game 18, full game] OR 8.2 - ICR upgrades Montreal to gray (but only when paying for it instead of using P16) and then appears to have no `available_hexes` for running routes (ie all hexes are dimmed and not selectable)
- [x] [game 19, full game] P21 3-Tile Grant "The National Dream" should be able to lay tiles on Y cities
- [x] [game 21, ERS] infinite recursion error while computing sawmill bonus

-----

I skimmed the 1822CA rulebook and made this checklist mainly from the rules differences section highlighted in red. It looks like just the basic scaffolding/setup is implemented so far.

Terms defined differently in 1822CA rulebook vs 18xx.games code, such as "city", will use the 18xx.games code definition here.

* [1822CA Rulebook on BGG](https://boardgamegeek.com/filepage/238950/1822ca-rules)
* [Errata](https://docs.google.com/document/d/1puHQJV4eLeunOtu_RyqAT-_mBCI93u8dqSBNwWMsAiE/edit)

I'm thinking I'll start by getting the Eastern Regional Scenario playable first, then the Western scenario, and then the full game.

Any changes that are checked here but aren't merged in are pushed to a branch at https://github.com/michaeljb/18xx.games/branches.

### New/Changed Private Companies

Other privates equivalent to privates implemented for 1822, some work may still be needed to "wire" them up correctly.

ERS

- [x] P5-P6 Pullman (train becomes N+2N instead of N+inf)
[...]

PR #9468, opened 2023-08-25
[1822Africa] Fix share movement on issue

vandamm:
### Before clicking "Create"

### Implementation Notes

When issuing shares stock price only moved left one step regardless of the number of shares issues.
Correct behavior is to move one step per share issues.

This change **may** break existing games if they had share issuing

### DIFF
--- a/lib/engine/game/g_1822_africa/game.rb
+++ b/lib/engine/game/g_1822_africa/game.rb
@@ -429,7 +429,7 @@ module Engine
             G1822Africa::Step::MinorAcquisition,
             G1822::Step::PendingToken,
             G1822::Step::DiscardTrain,
-            G1822::Step::IssueShares,
+            G1822Africa::Step::IssueShares,
           ], round_num: round_num)
         end
 
--- a/lib/engine/game/g_1822_africa/step/issue_shares.rb
+++ b/lib/engine/game/g_1822_africa/step/issue_shares.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative '../../g_1822/step/issue_shares'
+
+module Engine
+  module Game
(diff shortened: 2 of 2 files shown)


scottredracecar:
@michaeljb this is how it works in 1822CA too in case you have not modified this from base 22. It's correctly done in 1822MX already.
#9376

Rule from MX is (should be same as CA):
A major company may either issue or redeem its own stock, but not both in one turn. Issued stock is transferred from the company
treasury to the bank pool. Issuing stock causes the company’s price to move one step left per 10% share issued. The company
receives from the bank the final resting price, after dropping the price for each 10% share sold. Issuing stock must not cause more
than 50% of the stock to be in the bank pool. If the token moves to a space on the stock market which is already occupied, it is placed
at the bottom of the pile; if the token was already on the left-most space of the stock market, it keeps its relative position in a pile.

michaeljb:
Thanks for the tip @scottredracecar! Fixed for CA in #9472.

michaeljb:
Now 1822CA, 1822MX, 1822PNW, and 1822Africa can all have this same class duplicated...this is certainly one area making a strong case for me to pursue my idea of creating a new abstract game class, `G1822Base`, which all the 1822 games could inherit from so that the derivative games could stop overriding things that are specific to 1822 Great Britain.

Actually that abstract game class pattern would be good to apply to 1846 as well, would likely help smooth some things over when it comes time to do 1833NE.

scottredracecar:
That is a great idea. For my 1852 Missouri game, I needed to override a
bunch of special rules from 1846 (steamboat symbols on setup, etc). I
followed the model of 18LA to help figure out what I needed to override and
duplicate those overrides. A generic G1846Base would be a better way to do
that.

PR #9479, opened 2023-08-26
[1822Africa] */E train fixes

### LINKED ISSUE #9476: 1822Africa - E-train not counted correctly
In this hotseat game, this (E) route should be counted as:
2* (30+20+40+30*2) = 300
https://gist.github.com/scottredracecar/4ebc80a244f8152652d724b0b0d0c21f

Sorry, I can't figure out what is causing it to count 270.

image

@vandamm

vandamm:
Fixes #9476

### Before clicking "Create"

### Implementation Notes

- Fix Express mode being available if only one token is included in the route
- Fix destination bonus only counting once instead of twice for trains in Express mode

### DIFF
--- a/lib/engine/game/g_1822_africa/game.rb
+++ b/lib/engine/game/g_1822_africa/game.rb
@@ -581,6 +581,7 @@ module Engine
                       express_revenue = revenue_for_express(route, stops)
 
                       return express_revenue if train_over_distance?(route)
+                      return super unless includes_two_tokens?(route)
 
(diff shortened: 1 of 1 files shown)


vandamm:
I cannot add labels, but this may break games

scottredracecar:
added a label, this is the only game far enough for Es to be relevant and it already broke on some other action.
https://18xx.games/game/132623

michaeljb (approved):

PR #9480, opened 2023-08-27
[1822CA] P26-P27 Grain Trains, grain elevators, and ports

michaeljb:
Grain trains can attach to regular trains in order to earn revenue at grain elevators without counting the town as a stop. All standard trains except E-trains can earn bonuses at grain elevators, and an extra bonus if the route includes a grain elevator and a port.

Any train except an E-train can attach a grain train. This is done in the Route step. If a Major owns both a pullman and a grain train, the pullman is attached (or skipped) first.

Details on the revenue bonus rules can be found on page 21 in the [rulebook](https://boardgamegeek.com/filepage/238950/1822ca-rules):

* the table in rule 5.11.27 covers most of it
* 5.11.25 shows the port limit 

Very small, semi-related changes also included here:

* reorder pullman's distance to fix it counting stops; this was broken when enough towns were added since it was originally copied from 1822 pullmans, which have no limit to the number of towns
* after the train object is created, rename the "5P" train to "5"; the different id/name is useful in setup, but once a Major owns the train it is just like any other 5-train (and IMO looks better when renamed from attaching the pullman or grain train)

Core changes:

* extract new method `stop_type()` in `Game::Base`, so that grain elevator towns could be treated as the new type `grain_elevator` instead of `town` for grain trains
* in `Engine::Train`, added `attr_accessor :no_local` so it could be toggled when attaching a grain train; this was necessary for the LP-G train to select more than 2 stops

#9376

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1361,7 +1361,7 @@ module Engine
           h['nodes'].each { |type| type_info[type] << info }
         end
 
-        grouped = visits.group_by(&:type)
+        grouped = visits.group_by { |v| stop_type(v, train) }
 
         grouped.sort_by { |t, _| type_info[t].size }.each do |type, group|
           num = group.sum(&:visit_cost)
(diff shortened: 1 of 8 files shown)


michaeljb:
grain elevators and ports

tobymao (approved):

PR #9495, opened 2023-08-28
[1822CA, 1822MX] if a city has multiple tokens from one corp, return extra to charter

### LINKED ISSUE #9494: [1822MX] excess tokens disappear instead of returning to Major's "Available" tokens
gist - https://gist.github.com/michaeljb/a4442c112337d0e95a858c7a8b10d004

* MC acquired Minors 12 and 13, so it has two tokens on L19
* MC has 1 Available token and 1 Exchange token
* MC upgrades L19 to green, joining the two cities up; one token is removed from the city
* MC still has 1 Available token and 1 Exchange token; the second token from the map is just gone

Rule 6.11.7 (emphasis mine), from page 19:

> In most cases, a company may not have more than one token in a hex. The tiles and off-board areas show the conventional grouping of slots by enclosing them in an oval. However, there are some exceptions in hexes with preprinted track on the board which start with separate unconnected token slots. The upgrade paths for these cities are a single merged city. It is possible, typically via minor company acquisition or via the NdeM privatization token auction, that a company has multiple tokens on the same hex that are not in the same grouping. This is allowed, and when these tokens merge, **extra tokens in excess of one are returned to the “Available” section of the charter for use** following the normal rules.

This will also be necessary for #9376

michaeljb:
* in previous 1822MX implementation, the extra token was simply destroyed instead of being returned

Fixes #9494

#9376

### DIFF
--- a/lib/engine/game/g_1822/step/tracker.rb
+++ b/lib/engine/game/g_1822/step/tracker.rb
@@ -101,6 +101,28 @@ module Engine
         def upgraded_track?(from, _to, _hex)
           from.color != :white
         end
+
+        def update_token!(action, entity, tile, old_tile)
+          tile.cities.each do |c|
+            present_corps = []
+            tokens_to_return = []
+            c.tokens.compact.each do |t|
+              if present_corps.include?(t.corporation)
+                tokens_to_return << t
+              else
+                present_corps << t.corporation
+              end
+            end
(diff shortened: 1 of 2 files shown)


michaeljb:
Possibly breaks some 1822MX games

tobymao (approved):

Issue #9358, opened 2023-08-08
[1880] trains should be purchasable from depot / other corps interchangeably


benjaminxscott:
Ref: https://github.com/tobymao/18xx/issues/9331, a translation error resulted in a mis-implementation of the timing when corps may buy trains "across" or from depot

Corps may buy trains interchangeably from other corps or from the depot

crericha:
@benjaminxscott can you provide more context here? The referenced game is 1848.

philcampeau:
This appears to have been resolved. I'm able to buy trains in either order now. Closing.

PR #9507, opened 2023-08-29
[1880] allows buy from depot after cross buy

philcampeau:
Fixes #https://github.com/tobymao/18xx/issues/9358

### Before clicking "Create"

### Implementation Notes

removed 
```
&&
   corporations_sold.empty?
```

from lines 83 & 84 of buy_trains.rb, which appear to be what was causing the game to end a corp's turn as soon as it bought a train across from another corp. 

* **Explanation of Change**

This coding was due to a mistranslation in the rules, which was only meant to reconfirm that you can only buy trains across during the Buy Trains step of the game.

### DIFF
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -80,8 +80,7 @@ module Engine
 
           def can_buy_restored2?(entity)
             @game.phase.available?(restored2_train&.available_on) &&
-            !owns_restored2?(entity) &&
-            @corporations_sold.empty?
+            !owns_restored2?(entity)
           end
 
           def can_buy_restored2_from_depot?(entity)


lib/engine/game/g_1880/step/buy_train.rb
```diff
@@ -80,8 +80,7 @@ def owns_restored2?(entity)
 
           def can_buy_restored2?(entity)
             @game.phase.available?(restored2_train&.available_on) &&
-            !owns_restored2?(entity) &&
-            @corporations_sold.empty?
```
crericha, line 84: This is incorrect. The president cannot contribute towards purchasing a restored 2T.

crericha (changes requested):
@philcampeau please revert this change. The code that disables cross buys is somewhere else. I looked briefly and couldn't find it. @bentziaxl wrote that code and hopefully remembers where it is.

bentziaxl:
and that code was very messy :-(

PR #9512, opened 2023-08-30
[1822Africa] Only increase loans after SR x.2

### LINKED ISSUE #9493: 1822Africa - Don't apply loan interest after SR X.1 (1 of 2)
I guess I am too good of an 1822 player such that I did not run into major loan problems in my solo testing.

The current implementation is extra harsh as the loan interest is applied after each of the back-to-back SRs. I want to remove the application of loan interest after SRX.1 (1 of 2) and I adjusted the rules doc now to reflect that.

This action is right after SR5.1 and loan interest is applied. 
https://18xx.games/game/132767?action=327

vandamm:
Fixes #9493

### Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1822_africa/round/stock.rb
+++ b/lib/engine/game/g_1822_africa/round/stock.rb
@@ -79,8 +79,8 @@ module Engine
             # Refill the bidbox
             @game.bidbox_refill!
 
-            # Increase player loans with 50% interest
-            @game.add_interest_player_loans!
+            # Increase player loans with 50% interest in SR x.2
+            @game.add_interest_player_loans! if round_num == 2
 
             # Move sold out corps stock value right
             corporations_to_move_price.sort.each do |corp|


roseundy:
Out of curiosity, why are you requesting pins for an alpha game?

michaeljb:
Maybe he didn't know about the archive label? Anyway I've been pinning the 1822Africa games.

michaeljb (approved):

vandamm:
Yep, I don't know how this stuff works unfortunately :(

scottredracecar:
I've been watching progress on these games daily so I'd be happy to have whatever pins can be accommodated, but I know it takes extra effort (although I don't really understand how much). This was a prototype gameplay change (not an "alpha bug"), so I have some concern about losing games and discouraging players from playing. I understand that they started an alpha game with no guarantee of completing it, but my expectation is that there won't be many (any?) more bugs that archive/pin games. The 1822 implementation has been tested a lot and despite appearances, this doesn't stray that far from it ruleswise.

There will be one massive tile update later that will probably pin/archive every game, but that will be in about a month when there are a lot of competed games I can review. That will probably be the final change before moving off prototype status (maybe a month after that). Assuming that is going fine, I won't care about pins so much as I don't expect to be reviewing games regularly after that tile revision is found to be acceptable. I guess the players still appreciate it.

Just wanted to share my thoughts about this in response to Steve's question about pinning alpha games. Otherwise, I agree with the concept of not pinning alpha games. There is just one dude that kind of cares about it in this particular case.

Alex, one bad part about pins is that pinned games don't send turn notifications.

PR #9532, opened 2023-09-02
1822Africa - Update Tiles, Fix setup private company count

scottredracecar:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
- Fixes the number of private companies in the deck setup (should be 12, not 10)
- Revises map tiles to test for use in the physical game (will change back to unlimited plain yellow tiles after playtest period)
- Removes two (now) unnecessary off-map spikes

This will archive almost all existing games due to the deck setup and tile changes and that is okay.

### DIFF
--- a/lib/engine/game/g_1822_africa/game.rb
+++ b/lib/engine/game/g_1822_africa/game.rb
@@ -331,7 +331,7 @@ module Engine
           @companies.clear
           @companies.concat(minors)
           @companies.concat(concessions)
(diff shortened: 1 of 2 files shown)


vandamm (approved):

vandamm:
@scottredracecar do you want to ask for pins or proceed with archiving as usual?

scottredracecar:
No pins for this one.

My new main area of interest is whether the tile mix is right so pinning games won't help address that. Bummer to lose ~30 games in progress though.

PR #9554, opened 2023-09-05
[1822Africa] Fix express trains running over the distance

### LINKED ISSUE #9550: 1822Africa - E train earnings incorrect
https://18xx.games/game/133836?action=369

In this game, the 5/E ran as an E ["Express]", but appears it was counted like an 1830 diesel with 290 earnings.

E-trains should only generate earnings from tokened cities and nothing else, not even the game reserve. The best E run for this turn is two stops, Cape Town (home) to Nairobi (destination), calcluated as 2 * (20 + 30 * 2) = 160

Bonuses are applied after the fact, like the private company bonuses for gold mine or coffee plantation, and thus would add the bonus after the run is totaled (and doubled).

vandamm:
Fixes #9550

### Before clicking "Create"

### Implementation Notes

*/E train running over the distance were using express mode, but still calculating the best possible revenue. So it could be that such a train runs for 10-20 stops, treats itself as express, but gets revenue as a diesel)

**This will definitely archive some games**

### DIFF
--- a/lib/engine/game/g_1822_africa/game.rb
+++ b/lib/engine/game/g_1822_africa/game.rb
@@ -594,6 +594,8 @@ module Engine
           express_revenue = revenue_for_express(route, stops)
           express_revenue += destination_bonus_for(route) * self.class::EXPRESS_TRAIN_MULTIPLIER
 
+          return express_revenue if train_over_distance?(route)
+
           [revenue, express_revenue].max
         end
 


michaeljb (approved):

michaeljb:
> This will definitely archive some games

Actually the routes don't get validated by the engine when they are processed, so this won't pin any games--some will just have routes where if you went back to reproduce the route, the UI would be showing you a smaller amount than what they gained before.

https://github.com/tobymao/18xx/blob/e5e6898ff5c6353d78f0a4ef2f14552c2b6ad34b/lib/engine/step/route.rb#L28-L48

PR #9575, opened 2023-09-09
[1822Africa] Fix Gold Mine bonus crashing routing

vandamm:
### Before clicking "Create"

### Implementation Notes

Fixes the game hanging on routing step

### DIFF
--- a/lib/engine/game/g_1822_africa/game.rb
+++ b/lib/engine/game/g_1822_africa/game.rb
@@ -642,7 +642,7 @@ module Engine
         end
 
         def gold_mine_bonus(_route, stops)
-          return 0 unless stops&.find { |s| s.hex == @gold_mine_token.hex }
+          return 0 if !@gold_mine_token || stops&.none? { |s| s.hex == @gold_mine_token.hex }
 
           self.class::GOLD_MINE_BONUS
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1822_africa/game.rb
```diff
@@ -642,7 +642,7 @@ def gold_mine_corp
         end
 
         def gold_mine_bonus(_route, stops)
-          return 0 unless stops&.find { |s| s.hex == @gold_mine_token.hex }
+          return 0 if !@gold_mine_token || !stops&.find { |s| s.hex == @gold_mine_token.hex }
```
ollybh, line 645:
```suggestion
          return 0 if !@gold_mine_token || stops&.none? { |s| s.hex == @gold_mine_token.hex }
```
You're not using the return value from `find`, apart from converting it into a boolean. Calling `none?` instead is more idiomatic for Ruby.

vandamm: @ollybh thank you for the suggestion, I've changed the code. I'm still not used to Ruby having methods for all possible use-cases :)

michaeljb (approved):

PR #9584, opened 2023-09-10
[18Tokaido] beta candidate; adds waterfall optional rule

doubt72:
### Before clicking "Create"

### Implementation Notes

- Moves 18Tokaido to beta.
- Adds a variant that uses the regular waterfall auction rather than the snake draft
- Changes name of to remove the space from '18 Tokaido' -- don't _think_ this will break anything.  There don't seem to be any games currently in progress anyway.


tobymao:
@michaeljb when you deploy this, can you rename all 18 tokaido games?

tobymao (approved):

doubt72:
I'm going to close the PR in favor of a different one with a few different changes.

PR #9587, opened 2023-09-11
18Tokaido changes/rename/move to beta

doubt72:
### Before clicking "Create"

### Implementation Notes

Opening this in favor of #9584 (figured I might as well make these changes now, since there aren't any active games).

Some code changes, but mostly changes to the map layout, tile inventory, and stock market.

- Map changes
  - Changed some names to be more consistent in scale
  - Changed some offboard values
  - Moved/split one offboard
  - Switched one mountain to empty
  - Switched T tiles to more standard OO naming
- Stock Market changes
  - Split out into own object
  - Made it a bit more complicated/added yellow zone
  - Added option to remove yellow zone
- Tile changes 
  - Split out into own object
  - Pared back tiles a little bit and reorganized order in code by color
  - Added option to add a few tiles
- Adds a variant that uses the regular waterfall auction rather than the snake draft, which also changes other aspects of how privates work (initial prices, how much they can be bought into companies for)
- Moves 18 Tokaido to beta
- Changes the name to 18Tokaido (no space)

Added later:
- Removed unused cert limit code (never mind, that breaks the tests)
- Turned off permissive tile lays


doubt72:
@michaeljb -- also whatever considerations where needed for #9584, i.e., renaming games?

doubt72:
Meh, I broke the PR; constant needs to be there even if it isn't used

michaeljb:
> @michaeljb -- also whatever considerations where needed for https://github.com/tobymao/18xx/pull/9584, i.e., renaming games?

Should be able to deploy this evening and will rename games in the DB.

michaeljb:
Updating with this command in an `irb` console:

```
DB[:games].where(title:'18 Tokaido').update(title: '18Tokaido')
```

PR #9601, opened 2023-09-13
Upgrade to Opal 1.7.3 and Ruby 3.2

michaeljb:
* Opal 1.7 adds support for `Set` and Ruby 3.2, see the [Opal Change Log](https://github.com/opal/opal/blob/master/CHANGELOG.md) for more

* Ruby 3.2 upgrade requires updating bundler and some additional gems: argon, rerun, and their dependencies listen, rack, rb-fsevent

* the method name caching added in Opal 1.7 has a bug with toString; I've submitted a PR to fix this, but have also included a temporary fix to unblock our code, courtesy of hmdne; see https://github.com/opal/opal/issues/2583#issuecomment-1718381113

* the Ruby 3.2 docker image includes a newer version of git, which will show a "dubious ownership" error message and breaks `rake precompile` without the `git config` added in the Dockerfile here; see the [GitHub blog](https://github.blog/2022-04-18-highlights-from-git-2-36/#stricter-repository-ownership-checks) for more detail

* add note in rubocop config about upgrading the configured target version; even without upgrading the rubocop version, upgrading the target ruby version activates more cops, so I'm saving that for another time

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -16,10 +16,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Ruby 3.0
+    - name: Set up Ruby 3.2
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 3.0
+        ruby-version: 3.2
     - name: Build and test with Rake
       run: |
         gem install bundler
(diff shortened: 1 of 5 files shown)


michaeljb:
Will hold this till after I’m back from my trip.

Additionally, there have been some complaints since upgrading to Opal 1.6, so I may roll that back and see if it does address the source of those complaints. If so, I’ll resolve those before pressing on with 1.7

PR #9638, opened 2023-09-19
[1822CA] fix destinating

michaeljb:
bug caused by avoiding var name shadowing with a simple appended underscore--one spot in the loop was missed; this changes the outer var to be named `dest_cities` so the inner one can just be `cities` instead of `cities_`

#9376

### DIFF
--- a/lib/engine/game/g_1822_ca/game.rb
+++ b/lib/engine/game/g_1822_ca/game.rb
@@ -591,14 +591,14 @@ module Engine
         def destination_city(hex, entity)
           return hex.tile.cities[0] unless (exits = entity.destination_exits)
 
-          cities = exits.each_with_object([]) do |exit, cities_|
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1822_ca/game.rb
```diff
@@ -591,14 +591,14 @@ def event_open_detroit_duluth!
         def destination_city(hex, entity)
           return hex.tile.cities[0] unless (exits = entity.destination_exits)
 
-          cities = exits.each_with_object([]) do |exit, cities_|
+          dest_cities = exits.each_with_object([]) do |exit, cities|
             hex.paths[exit].each do |path|
               next unless (city = path.city)
 
-              cities_ << city unless cities.include?(city)
```
michaeljb, line 598: `cities.include?(city)` throws ` "undefined method include for Nil"` because it should have been `cities_.include?(city)`

PR #9644, opened 2023-09-19
[RollingStock] Other companies are not biddable while in the middle of auction

crericha:


roseundy (approved):
LGTM

roseundy:
Why did this issue just pop up now?

crericha:
> Why did this issue just pop up now?

The issue was always there, but the exception happened after the page loaded, so it went unnoticed. A refactor made it happen during load instead of after.

PR #9680, opened 2023-09-25
[1822Africa] Allow placing Gold Mine if corporation is out of tokens

### LINKED ISSUE #9676: 1822Africa - Cannot use P14 Power
https://18xx.games/game/134428?action=458 

Possibly triggered because the company is out of tokens to place. It worked correctly here: https://18xx.games/game/135568?action=463

vandamm:
Fixes #9676

### Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1822_africa/step/token.rb
+++ b/lib/engine/game/g_1822_africa/step/token.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require_relative '../../g_1822/step/token'
+
+module Engine
+  module Game
+    module G1822Africa
+      module Step
+        class Token < G1822::Step::Token
+          def actions(entity)
+            actions = super.dup
+
+            actions << 'choose_ability' if need_choose_ability?(entity, actions) && !actions.include?('choose_ability')
+
+            actions << 'pass' if !actions.empty? && !actions.include?('pass')
+            actions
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1822_africa/step/token.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Game
+    module G1822Africa
+      module Step
+        class Token < G1822::Step::Token
+          def actions(entity)
+            actions = super.dup
+
+            actions << 'choose_ability' if need_choose_ability?(entity, actions) && !actions.include?('choose_ability')
+
+            actions << 'pass' if actions.any? && !actions.include?('pass')
```
michaeljb, line 15: The JS produced by Opal for `any?` without a block is much less efficient than `!empty?`. (Unfortunately there are also a bunch of other places in the code doing this that need to be cleaned up too)

vandamm: Can we make rubicon check for that?

michaeljb: Not with any cops/configurations I’ve been able to find

vandamm: @michaeljb changed it

PR #9697, opened 2023-09-28
[1861/1867] Switch merger round PostMergerShares and ReduceTokens steps

### LINKED ISSUE #9655: 1861: removal of excess tokens during merger (minor)
https://18xx.games/game/135455

During MR3.1, there was a three-way merger between N, MB and KB. We were given the option of buying shares, and then the director had to remove one of the tokens. BUT the rules say the sequence should be the other way around:

12.3 "The Director must chose two Station Markers of the merging companies..."
12.4 "Starting with the Director... buy shares in the Company..."

[This has arisen because the 3rd edition rules have got this wrong.]

I assume 1867 is the same.

ollybh:
When three or more minor companies merge, the Grand Trunk Games rules say that you remove extra tokens *after* players have bought extra shares in the newly created major company. This contradicts the All-Aboard Games rules, which has the tokens removed first. According to the designer (@ianwilson156) the All-Aboard Games rules are correct, so this switches the order of the PostMergerShares and ReduceTokens steps. Fixes #9655.

Changing the order of these steps will break games where three or more minors have been merged into a major. To avoid pinning games, `migrate_games.rb` has code added to switch the actions in these steps. There probably aren't a huge number of games affected by this change (three-way mergers are relatively rare), but I'd rather not have to pin these.


michaeljb:
Code changes look good, and great comments! Looking into the validation and migration locally now to check it out.

michaeljb:
Found breakages in 209 out of 887 games of 1861/1867. I manually checked on just a couple, and as expected, they are breaking at a merger with 3+ companies. Running your migration locally on them now.

michaeljb:
Anyway the migration is running a long, some games are being fixed but it looks like a good number are failing to migrate, getting stuck in a loop. I'll share more details (and a tweak to the migration script to avoid adding `pin: null` in the db) when it finishes running. In the meantime, one sample game that is failing to migrate in this way is 113326.

[This gist](https://gist.github.com/michaeljb/0097a5c356a6ee54d5dba985845a2910) has the ids for the 208 broken games.

michaeljb:
64 of the 208 broken games failed to migrate, I've uploaded their ids to [this gist](https://gist.github.com/michaeljb/5674feb60e8aa7964ce75e77b269b114).

ollybh:
Looking at three random games in this list, they all have auto_actions in the last action of the reduce tokens step. These are passes for the merge step, so I'll need to move them from their original location to the last action in the post merge shares step.

ollybh:
Moving the auto_actions reduced the number of failing games to 15. 12 of the remaining ones were the script getting confused by undo actions. Just three to go…

ollybh:
@michaeljb I hope this is good to go now. I've tested the modified script locally and it can now successfully migrate all the affected games.

michaeljb:
@ollybh awesome! I’ll take another look soon, probably on Tuesday or Wednesday evening

scottredracecar:
Ian, I wouldn't worry about it. People only work on these things if they
want to do it. There is a huge list of other bugs that could be worked on
if there was interest in working on them. Just take it as a pat on the back
that people want to get it right for your game. :-)

Scott

ollybh:
> Scott, thanks for that! It just surprised me that there were so many 100's of games in progress...

@ianwilson156 Most of these games are finished, not in progress. Of the 208 games with 3-way mergers, 10 were in progress, the rest finished.

I'm surprised by how many games there were with these three-way mergers. I don't recall seeing this ever happen in a face-to-face game. Different kinds of groupthink, maybe. And there were three games which each had three three-way mergers.

michaeljb (approved):

PR #9715, opened 2023-10-03
[18Tokaido] color plus name changes to match print version

doubt72:
### Before clicking "Create"

### Implementation Notes

Changes to get in sync with current version of in-progress print prototype.  A couple company (three-letter) code changes to match the prototype map, some tile number changes to match prototype tile sheet, and colors now match 18Chesapeake.

All the code changes are basically cosmetic, but pretty sure the private/company three-letter code changes require a pin.


michaeljb:
Yep, the company `sym` changes and tile name changes will require pins.

michaeljb (approved):

PR #9731, opened 2023-10-06
18ZOO - Update StockMarket to show bigger cells when there is any user info in any cell

### LINKED ISSUE #9693: 18zoo market details
Hi,
I ask if it will be possible to code more numbers in each cell of market tab.
What 18zoo needs is three numbers (share value,revenue, threshold) for each cells like in this pic

:strip_icc()/pic6153038.png)

i know it is the only 18xx with this feature, so I hope I can get the “ok” and ask my coder to see how to do it

carlorusso1984:
Fixes #9693

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
As requested inside the issue, the stock market was updated to support the info, but only for the 2d grid

* **Screenshots**
image

* **Any Assumptions / Hacks**

### DIFF
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -31,24 +31,24 @@ module View
       }.freeze
 
       # All markets
-      PAD = 4                                     # between box contents and border
+      PAD = 4 # between box contents and border
       BORDER = 1
-      WIDTH_TOTAL = 50                            # of entire box, including border
+      WIDTH_TOTAL = 50 # of entire box, including border
       TOKEN_SIZE = 25
       TOKEN_SIZES = { small: 25, medium: 32, large: 40 }.freeze
 
       # 1D markets
-      VERTICAL_TOKEN_PAD = 4                      # vertical space between tokens
-      MIN_NUM_TOKENS = 3                          # guarantee space for this many tokens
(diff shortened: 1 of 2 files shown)


assets/app/view/game/stock_market.rb
```diff
@@ -69,7 +69,7 @@ class StockMarket < Snabberb::Component
         textAlign: 'center',
         position: 'absolute',
         bottom: "#{PAD}px",
-        width: "#{WIDTH_TOTAL - (2 * PAD) - (2 * BORDER)}px",
+        width: "#{WIDTH_TOTAL + PRICE_HEIGHT - (2 * PAD) - (2 * BORDER)}px",
```
carlorusso1984, line 72: I added PRICE_HEIGHT to both width and height to keep the cells as square

tobymao:
i reverted this, does this change the default stock market look?

carlorusso1984:
It should change only the stock market when used as 2D and there are userInfo

serpentium:
> i reverted this, does this change the default stock market look?

Can you check again? Carlo answer it it is only here.

carlorusso1984:
@serpentium I have done this new PR to avoid changing default stock market: https://github.com/tobymao/18xx/pull/9790
I think I'll have that merged this weekend

PR #9785, opened 2023-10-15
[1868WY] fix LHP assignment when phase 5 is triggered by export

### LINKED ISSUE #9714: [1868 Wyoming] Game will not load
**What's your Game ID?** 134425 https://18xx.games/game/134425

**How to Reproduce the Bug?** Try to open the game link, not sure what game actions potentially caused it because I am unable to see the game log.

**Which Browser / Platform?** Google Chrome Version 116.0.5845.187

michaeljb:
Export means the assignment happens in a stock round instead of an operating round. The Assign step has a couple of methods implemented so it can render without error in the stock round, and a newly created `selected_company` method so the LHP is automatically selected without the player choosing it.

Because the LHP is the active entity, it needs methods from the `ShareHolder` module which are called on the entities when rendering a stock round.

Fixes #9714

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**

Adding the `ShareHolder` module to companies feels like a hack, and there is some finickiness with how the new `selected_company` method is called.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -703,6 +703,13 @@ module View
       end
 
       def can_assign_corporation?
+        # reset selected_company if cleared during an undo action
+        @selected_company ||=
+          begin
+            step = @game.round.active_step
+            (step.respond_to?(:selected_company) && step.selected_company) || nil
+          end
+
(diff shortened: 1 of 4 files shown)


michaeljb:
Moving this from the SR to the BUST round would solve some of the SR rendering hackiness.

PR #9787, opened 2023-10-16
enforce  token abilities w/ limited hexes

### LINKED ISSUE #9782: [1882] NWR station teleport allows to place in non-NWR city
NWR private allows to teleport token to non-NWR tile.
in our game NWR private allowed to place token of CPR in Pile o'bones, so not in NWR tile.
It looks like a bug

game id 137524

https://18xx.games/game/137524

benjaminxscott:
Fixes #9782

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
It doesn't look like we were checking when tokens were placed via ability

* **Screenshots**

Expected error thrown when wrong hex is chosen

Screenshot 2023-10-15 11 22 24 PM

Still works if correct hex

Screenshot 2023-10-15 11 22 46 PM

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/step/tokener.rb
+++ b/lib/engine/step/tokener.rb
@@ -48,18 +48,15 @@ module Engine
         hex = city.hex
         extra_action ||= special_ability.extra_action if %i[teleport token].include?(special_ability&.type)
 
+        check_connected(entity, city, hex) if connected
         unless @game.loading
-
-          check_connected(entity, city, hex) if connected
-
           if special_ability&.type == :token && special_ability.city && special_ability.city != city.index
             raise GameError, "#{special_ability.owner.name} can only place token on #{hex.name} city "\
                              "#{special_ability.city}, not on city #{city.index}"
           end
 
-          if special_ability&.type == :token && !special_ability&.hexes&.empty? && !special_ability.hexes.include?(hex.id)
-            raise GameError, "#{special_ability.owner.name} can only place token on #{special_ability.hexes} hexes"\
-                             ", not on #{hex.id}"
+          if special_ability&.type == :token && !@round.active_step.available_hex(special_ability.owner, hex)
+            raise GameError, "#{special_ability.owner.name} cannot place token on hex #{hex.id}"
           end
 
           if special_ability&.type == :teleport &&
--- a/lib/engine/step/tokener.rb
+++ b/lib/engine/step/tokener.rb
@@ -48,30 +48,38 @@ module Engine
         hex = city.hex
         extra_action ||= special_ability.extra_action if %i[teleport token].include?(special_ability&.type)
 
-        check_connected(entity, city, hex) if connected
+        unless @game.loading
 
-        if special_ability&.type == :token && special_ability.city && special_ability.city != city.index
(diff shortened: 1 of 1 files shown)


lib/engine/step/tokener.rb
```diff
@@ -55,6 +55,11 @@ def place_token(entity, city, token, connected: true, extra_action: false,
                            "#{special_ability.city}, not on city #{city.index}"
         end
 
+        if special_ability&.type == :token && special_ability.hexes.any? && !special_ability.hexes.include?(hex.id)
```
crericha, line 58: Prefer `!empty?` over `any?`. The latter is much slower due to the way it is implemented in Opal.

crericha:
The special_token step often has custom logic in `available_hex`. Checking that that method returns true is a more thorough confirmation than looking at the abilities hexes attribute.

lib/engine/step/tokener.rb
crericha, line 58: Only do the check if `!@game.loading`, so that we don't recheck every time the game loads, slowing down the load speed.

benjaminxscott: ah, I was copypasting from https://github.com/benjaminxscott/18xx/blob/master/lib/engine/step/tokener.rb#L132

benjaminxscott: would this check wrap the two other city and hex checks in this function?

crericha: Its wrong there too then :-D

crericha: What do you mean?

benjaminxscott: should the game-loading check wrap all the code from line 52 to ~72?
(the various sanity checks for city/hex)

crericha: Yes, good catch.

michaeljb:
I searched for games which had occurrences of `type: 'token'`, and validated those titles against this branch.

Found only five game IDs from a few titles needing pins:

* game ids: `[107436, 129222, 129223, 136816, 137524]`
* titles: `["1882", "18VA", "18ZOO - Map A", "18ZOO - Map D"]`

The full list of titles validated:

`["18 Los Angeles","18 Los Angeles 2","1822","1822Africa","1822MRS","1822MX","1822NRS","1822PNW","1828","1846","1846 2p Variant","1847 AE","1852: Missouri","1856","1882","18FL","18GA","18GB","18MEX","18MS","18Mag","18NL","18NY","18NY 1st Edition","18SJ","18USA","18VA","18ZOO","18ZOO - Map A","18ZOO - Map B","18ZOO - Map C","18ZOO - Map D","18ZOO - Map E","18ZOO - Map F"] `

benjaminxscott:
> The special_token step often has custom logic in `available_hex`. Checking that that method returns true is a more thorough confirmation than looking at the abilities hexes attribute.

@crericha I refactored to check `available_hex()` as you mentioned, however it breaks a bunch of fixtures - am I approaching this correctly?
https://github.com/tobymao/18xx/pull/9787/files#diff-956c7250f236cdc45811a6aa52c6b19c2c3debad547c417246c26f9df193ed4dR58

PR #9827, opened 2023-10-24
generalize values used for variable dividends

benjaminxscott:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Currently, if one creates a game with a "variable" dividend payout method and a custom 'dividend step', one must also define a specific set of methods in the custom step or the custom step will error out.

This attempts to shave off those rough edges by adding default methods in `step/dividend base`, which are called [from view/dividend](https://github.com/tobymao/18xx/blob/master/assets/app/view/game/dividend.rb#L147) to render the variable payout UI. 

I tried to set sane defaults here, LMK if these need tweaked

This also removes the relevant custom method definitions in 1840 and Rolling Stock which can just use those defaults

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -48,6 +48,18 @@ module Engine
         end
       end
 
+      def variable_share_multiplier(_corporation)
+        1
+      end
+
+      def variable_input_step
+        1
+      end
+
+      def variable_max
+        1
+      end
+
       def process_dividend(action)
         entity = action.entity
         revenue = total_revenue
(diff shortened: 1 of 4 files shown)


lib/engine/step/dividend.rb
```diff
@@ -48,6 +48,18 @@ def dividend_options(entity)
         end
       end
 
+      def variable_share_multiplier(_corporation)
```
ollybh, line 51: Could you add comments to all three of these functions, describing what they do and when they would need to be overridden?

roseundy:
18Mag also has variable dividends.

PR #9829, opened 2023-10-25
[1840] add validity check for chosen dividend

### LINKED ISSUE #9825: [1840] Dividend payout is not limited to Revenue
**What's your Game ID?**
https://18xx.games/game/136326

https://gist.github.com/benjaminxscott/f15deff30698e0698c92405555888e7a

**How to Reproduce the Bug?**
- During the "Company Round 2 - Pay or Withhold Dividends" a player can set the payout amount
- Interface suggests: "Select dividend to distribute to shareholders, between 0 and 100. Leftover will be withholded to corporation.", and the maximum value is pre-filled in the text form.
- The rules state that the Dividend paid must come from the Revenue pile (so maximum of 100 in that case)
- The interface allows you to fill in any number (even more than the total Treasury!)

benjaminxscott:
Fixes #9825

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Make sure the user has chosen a valid dividend payout amount

I also took the liberty of tweaking the explanation language to hopefully be a bit more clear

* **Screenshots**

before:
Screenshot 2023-10-24 2 14 12 PM

after:
Screenshot 2023-10-25 11 30 31 AM

Screenshot 2023-10-25 11 27 29 AM

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_1840/step/dividend.rb
+++ b/lib/engine/game/g_1840/step/dividend.rb
@@ -37,7 +37,12 @@ module Engine
             amount = action.amount || 0
             payout = corp_dividend_options(entity, amount)[kind]
 
-            raise GameError, "Amount must be multiples of #{variable_input_step}" if amount % variable_input_step != 0
+            if (amount > variable_max(entity)) || amount.negative?
+              raise GameError,
+                    "Dividend cannot be larger than total revenue of #{@game.format_currency(variable_max(entity))} "\
+                    "or smaller than #{@game.format_currency(0)}"
+            end
+            raise GameError, "Dividend must be multiples of #{variable_input_step}" if amount % variable_input_step != 0
 
             entity.operating_history[[@game.turn, @round.round_num]] = OperatingInfo.new(
               routes,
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1840/step/dividend.rb
```diff
@@ -37,7 +37,12 @@ def process_dividend(action)
             amount = action.amount || 0
             payout = corp_dividend_options(entity, amount)[kind]
 
-            raise GameError, "Amount must be multiples of #{variable_input_step}" if amount % variable_input_step != 0
+            if (amount > variable_max(entity)) || amount.negative?
+              raise GameError,
+                    "Dividend cannot be larger than total revenue of #{@game.format_currency(variable_max(entity))} "\
+                    "or smaller than #{@game.format_currency(0)}"
```
magnusrk-ch, line 43: Nit: Do we really need to explain that the dividend can't be negative? :)

If we do, how about shortening it to something like "Dividend cannot be negative or larger than total revenue ($total)"

benjaminxscott: I don't have a strong preference, the user is only running into this error if they've done something wonky so it seems OK

PR #9856, opened 2023-11-05
[1850][1870] Fix price not dropping when selling shares results in a presidency transfer.

### LINKED ISSUE #9659: [1850] I sold five shares of MILW and the stock price did not decline. Game Id 132733
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

andrewzwicky:
Fixes #9659 

### Before clicking "Create"

- I don't believe 1870 would need to be pinned, but 1850 likely would be.

### Implementation Notes
1. The 1850 game was not using the 1850 PriceProtection step, so I deleted it.
2. I added two test cases covering the presidency transfer (this fix) and ensuring that priority continues to be passed properly (was working correctly, but wanted to confirm with a test).
3. The PriceProtection step is now identical between 1850 and 1870.  I could not find any differences in the rules for price protection between the two games.
4. The current president will show in the log for price protection when they can't protect due to presidency transfers (affects both 1850 and 1870).

* **Explanation of Change**
I believe the root cause of the issue was the `bundle.corporation.president?(bundle_owner)` checks in the existing 1850 implementation.  When the first check is hit, `bundle.corporation.president?(bundle_owner)` is `True`.  This prevents shares from hitting the `sell_queue` if the president is selling.  The `sell_queue` is what the `PriceProtection` step uses to determine whether price protection happens or not.

    If the sale has triggered a change in presidency, by the time the second check is hit, `bundle.corporation.president?(bundle_owner)` is `False`, because the sale has triggered a change in presidency.  This prevents the `change_price` function from being called.

    I changed `sell_shares_and_change_price` to be identical between 1850 and 1870.  The sell queue will be used in all cases, and `process_pass` will handle the price drop both cases (presidential transfer and no transfer).  This eliminates the `change_price` function, but that logic is covered by `process_pass` in the `PriceProtection` step.

* **Screenshots**
This is from the given game, showing the result of the actions after fixes applied. https://18xx.games/game/132733?action=196
```
[11:14] -- Stock Round 6 --
[12:37] JGould sells 5 shares of MILW and receives $600
[12:37] xyz becomes the president of MILW
[12:37] JGould declines to buy shares
[12:37] xyz can't price protect 5 shares of MILW
[12:37] MILW's share price moves down from $120 to $76
```

* **Any Assumptions / Hacks**
1850 and 1870 price protection rules are identical.

@philcampeau


andrewzwicky:
I don't think I can add labels to this PR.

Rupre:
This is still not fixed in game id 132733.  I just sold the five shares of MILW and the price did not go down.

andrewzwicky:
Are you testing on the production website?  I think it takes a while for the changes to get deployed to the server and go live.  The live version right now says:

`Current version: deployed at Wed Nov 01 2023`

Rupre:
Yes.  I'll try again in a few days.

PR #9883, opened 2023-11-10
[1888] Fix tile manifest in 1888

### LINKED ISSUE #9795: [1888-N] There should be 2 #51 tiles
There is 2 #51 tiles in the physical game and in the rules

andrewzwicky:
Fixes #9795 

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**


roseundy:
This will necessitate pinning games. Also, should this just effect 1888-N and not the prototype version?

andrewzwicky:
Yes it should, I looked for some time, but couldn't find 1888-N, I thought it was a different game, didn't see the optional rule.  I'll revise and resubmit.

PR #9887, opened 2023-11-11
[1880] skip P5 assignment if corp already has D

### LINKED ISSUE #9791: [1880] autopass when player has P5 and company already has a D permit
https://18xx.games/game/135632?action=620

During phase D, at the start of a companies operating turn it was given the option of passing or skipping track lay.
No explanation of why

It had a CD permit, and I had the D permit private: the pass was to not give the company the D permit.  Shouldn't need the option, and if the option is given then it needs to be explained.

Android

benjaminxscott:
Fixes #9791

### Before clicking "Create"

### Implementation Notes

I believe this will require pins, since players could have inserted a 'pass' action for an "Assignment" step which is now skipped

* **Explanation of Change**

Skips the 'P5 assignment step' when the active corp already has a D permit

Adds an error case if the player tries to assign P5 to a corp that already has a D permit

Also made some "happy to glad" changes to the verbiage

* **Screenshots**

The below cases assume that a corp is operating whose president owns the P5

To address the specific case in the issue: in D phase, the Assign step will NOT block (be skipped) if active corp already has a D permit

image

in D phase, an Assign step will block if P5 can be used if active corp does not have a D permit
image

P5 cannot be used if corp already has a D permit
image

P5 can still be used earlier in the game
image

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_1880/step/assign.rb
+++ b/lib/engine/game/g_1880/step/assign.rb
@@ -9,13 +9,11 @@ module Engine
         class Assign < Engine::Step::Assign
           ACTIONS_WITH_PASS = %w[assign pass].freeze
           def actions(entity)
-            return [] if entity.player?
+            return [] if entity.player? || current_entity.minor?
             return [] unless @game.abilities(entity, :assign_corporation)
-            return [] if current_entity.minor? ||
-                        (entity == @game.p5 && current_entity.corporation? && current_entity.building_permits&.include?('D'))
 
             return ACTIONS if @game.forced_exchange_rocket? && entity == @game.rocket
-            return ACTIONS_WITH_PASS if p5_block? && current_entity.owner == @game.p5.owner
+            return ACTIONS_WITH_PASS if p5_block?
 
             super
           end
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1880/step/assign.rb
```diff
@@ -9,13 +9,11 @@ module Step
         class Assign < Engine::Step::Assign
           ACTIONS_WITH_PASS = %w[assign pass].freeze
           def actions(entity)
-            return [] if entity.player?
+            return [] if entity.player? || current_entity.minor?
```
benjaminxscott, line 12: these changes are a no-op with the intent to make the checks a bit more readable

p5_block does the same checks, so the functionality is the same

will be slightly slower since we descend into p5_block instead of short-circuiting earlier

bentziaxl: looks good, thanks for putting it all in p5_block, instead of the existing p5_block || condition.

Very nitpicky, and feel free to ignore, but maybe it should be p5_block? instead of p5_block

bentziaxl (approved):

lib/engine/game/g_1880/step/assign.rb
benjaminxscott: not sure I understand, the current patch has p5_block? as the function name, if that's what you're describing

PR #9889, opened 2023-11-11
[rolling stock] fix broken dividend step

### LINKED ISSUE #9888: [Rolling Stock Stars] dividend step throws error
**Game ID**
138858

https://gist.github.com/benjaminxscott/063ba50d1a4c376838f4eb794d2506f0

**Link**
https://18xx.games/game/138858

**How to Reproduce the Bug?**
Due to the behavior described in the title, I'm not in a position to contribute any further information.

During Turn 6, closing everything went well.  Currently it's my turn for dividends (T6).

**Which Browser / Platform?**

Checked on:
- Chrome Version 119.0.6045.124
- MS Edge Version 119.0.2151.44

benjaminxscott:
Fixes #9888

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Rolling stock's dividend step used to inherit from base.rb instead of base/dividend.rb, for some reason

Also adds a test file

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_rolling_stock/step/dividend.rb
+++ b/lib/engine/game/g_rolling_stock/step/dividend.rb
@@ -76,6 +76,14 @@ module Engine
             @game.max_dividend_per_share(current_entity)
           end
 
+          def variable_share_multiplier(_corporation)
+            1
+          end
+
+          def variable_input_step
+            1
+          end
+
           def chart
             @game.dividend_chart(current_entity)
           end
--- a/public/fixtures/RollingStockStars/dividend.json
+++ b/public/fixtures/RollingStockStars/dividend.json
@@ -1,0 +1,1725 @@
+{
+  "status": "finished",
+  "actions": [
+    {
+      "type": "bid",
+      "entity": 3669,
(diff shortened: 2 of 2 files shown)


roseundy:
Why are you changing this?

roseundy:
If the only reason you are inheriting from Dividend is to define 'variable_share_multiplier', why don't you just define it?

benjaminxscott:
I don't have a strong preference for implementation, it seems like this step is handling dividends so it seemsodd to have it derive from the base step instead

benjaminxscott:
As mentioned in a (deleted?) comment, we could define `variable_share_multiplier` explicitly, which is also fine, I don't have a preference for either approach

roseundy:
That would be my preferred approach, but why this is failing now? I deleted my comment because I saw you already added variable_share_multiple to Rolling Stock in a previous PR (#9827)

benjaminxscott:
no worries, it was my bad -not sure why the tests didn't catch it?

roseundy:
LGTM

PR #9891, opened 2023-11-12
don't override dividend history when skipping dividend step

### LINKED ISSUE #9882: [1856] CGR runs should not be shown as zero while it withholds for a train
After the CGR forms, if it has no train it must 'borrow' one and withhold every turn until it can afford one. This may go on for a number of turns, if, for example, it is saving for a Diesel.

However, while this is going on, the game incorrectly shows the CGR's runs as '0' (zero).  They are NOT zero - the CGR is most defintely operating and withholding, and this is useful information.

See this game for example:
https://18xx.games/game/137667?action=709

The CGR has run twice, most recently for $630. The runs should be shown correctly BOTH in the SPREADSHEET for OR 6.2 and 6.3, AND in the 'Show last route' dropdown on the MAP tab.

benjaminxscott:
Fixes #9882

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

The base dividend `skip!` function was overriding corp operating history to insert a zero payout 'withhold'

I wanted to double-check that in such cases we were still outputting corp operation history, so I checked and every game that overrides `process_dividend()` will still output corp operating history - either implementing a game-specific operating history (i.e. games with "psuedo-corps" like Bank of England) or calling `super`

* **Screenshots**

Screenshot 2023-11-11 5 14 46 PM

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -34,9 +34,6 @@ module Engine
 
       def skip!
         process_dividend(Action::Dividend.new(current_entity, kind: 'withhold'))
-
-        current_entity.operating_history[[@game.turn, @round.round_num]] =
-          OperatingInfo.new([], @game.actions.last, 0, @round.laid_hexes)
       end
 
       def dividend_options(entity)
--- a/lib/engine/step/dividend.rb
+++ b/lib/engine/step/dividend.rb
@@ -33,7 +33,10 @@ module Engine
       end
 
       def skip!
-        process_dividend(Action::Dividend.new(current_entity, kind: 'withhold'))
+        # insert a dummy action so the operating history has something to link to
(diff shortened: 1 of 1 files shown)


roseundy:
Are you sure this won't break games? Did you check to see why it was originally placed there?

benjaminxscott:
You're right, that 'dummy action' is used so the OR history revenue number is clickable, I have another approach I'll push up

https://github.com/tobymao/18xx/commit/27cd6b97d8497c9805a30aa7ea760bc4dd9f8b39#diff-0de8560da1654b67eb2f92fddc1d275aa29c42f92e109bbade13e943ddb685adR29

tobymao:
so are you closing this?

benjaminxscott:
I pushed a new commit

-_Ben

PR #9892, opened 2023-11-12
[1848] output BoE revenue history

benjaminxscott:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Output revenue history for BoE

* **Screenshots**

before:
Screenshot 2023-11-11 5 20 28 PM

after:
Screenshot 2023-11-11 5 05 11 PM

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_1848/step/dividend.rb
+++ b/lib/engine/game/g_1848/step/dividend.rb
@@ -38,6 +38,12 @@ module Engine
             revenue = calculate_boe_revenue
             payout = send(:payout, entity, revenue)
             payout_shares(entity, revenue) if payout[:per_share].positive?
+            entity.operating_history[[@game.turn, @round.round_num]] = OperatingInfo.new(
+          [],
+          action,
+          revenue,
+          @round.laid_hexes
+        )
             pass!
           end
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1848/step/dividend.rb
```diff
@@ -38,6 +38,13 @@ def process_dividend(action)
             revenue = calculate_boe_revenue
             payout = send(:payout, entity, revenue)
             payout_shares(entity, revenue) if payout[:per_share].positive?
+            action.id = @game.actions.last.id
```
bentziaxl, line 41: why is this line needed?

benjaminxscott: `action` needs an 'id' or else it won't render BoE's revenue as a clickable hyperlink when stored in the operating history

The BoE revenue isn't a "real action" that's stored in the state, so we just set it to the last real action. 

I do the same thing here for the case when a company skips its revenue step on its first operating round
https://github.com/benjaminxscott/18xx/blame/57f9f3e26577523aa7ce77b3a2850c8f8f668adf/lib/engine/step/dividend.rb#L36

Issue #9711, opened 2023-10-01
Dependent games frequently are processed twice; the initial fix breaks 18ZOO


michaeljb:
#9710 fixes the problem of dependent games being processed twice (or even thrice for something like 1822CA ERS), but then 18ZOO games fail to load:

> 08:48:25 [nogerwig](https://18xx.games/profile/10573) has anyone else trouble with games that don't load? [#136274](https://18xx.games/game/136274) doesn't for me, 18zoo

18ZOO maps need to become proper variants, as suggested over a year and a half ago in https://github.com/tobymao/18xx/pull/7111#issuecomment-1037700947 🙃 
> while you're at it, can you make 18zoo maps variants?

carlorusso1984:
Hey @michaeljb , should I look for the fix of 18ZOO? Let me know

michaeljb:
@carlorusso1984 I think the fix will be reorganizing the game files for each map so it’s set up like 1822 and 1822MRS/NRS. 

I have a branch where I got started on this but haven’t had time to finish yet. I’ll tag you when I open the PR

PR #9897, opened 2023-11-13
18zoo - Implements fixes asked inside https://github.com/tobymao/18xx/issues/9848

### LINKED ISSUE #9848: 18ZOO new features
Hi, i uploaded new manual version on bgg (vs 10) so here the new features to implement.

three FAQs to check against code
- si può upgradare in verde una traccia inserita con ANCIENT MAPS nello stesso OR
- Work in Progress says it blocks Wings, but it separately says the blocked tile can be opened by upgrading to a tile with two station spots. So if you upgrade to two spots and token the non-blocked one, is wings still blocked? no, can fly over the new tokened station
- verificare che GREEK + PASSA TURNO fa ancora rimanere il giocatore in SR

NEW FEATURES
- add brown nr47,nr44,nr39
- add green nr20,nr16
- change the number of yellows to x13 nr9, x7 nr7, x18 nr8
- change price of MIDAS and GREEK from 2$ a 1$ 
- enforce the new rule (6.2.1 Avoid the “wait-sell in order tecnique”): a non-president selling the same share in the same SR doesn't push down reputation more than the first time (just once). If the player selling it is the president, reputation goes down as usual (even more than once)

carlorusso1984:
Fixes #9848

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
1. Add test to verify Ancient Maps.
2. Add test to verify "Work in Progress doesn't block Wings for two space cities"
3. Added new yellows, green and brown tiles
4. Updated configuration to see the new tiles using http://18xx.games/tiles/18zoo/all
5. Updated price for MIDAS and GREEK abilities
6. Support the (6.2.1 Avoid the “wait-sell in order tecnique”) rule

* **Screenshots**

* **Any Assumptions / Hacks**


lib/engine/game/g_18_zoo/step/buy_sell_par_shares.rb
```diff
@@ -70,6 +70,25 @@ def available_companies
 
             @game.available_companies
           end
+
+          def round_state
+            # An extra field to track whether a share was already sold from any not president since the start of the round.
+            super.merge({ share_sold: [] })
+          end
```
bentziaxl, line 77: there is already a variable that keeps track of sold shares:
@round.players_sold

please use that (you can search for examples on how to use) before introducing new variables

lib/engine/game/g_18_zoo/step/buy_sell_par_shares.rb
```diff
@@ -70,6 +70,25 @@ def available_companies
...
+            raise GameError, "Cannot sell shares of #{shares.corporation.name}" if !can_sell?(entity, shares) && !swap
+
+            if shares.owner == shares.corporation.owner
+              should_move = true
+            elsif @round.share_sold.include?(shares)
+              should_move = false
+            else
+              should_move = true
+              @round.share_sold << shares
+            end
```
bentziaxl, line 89: this if/else can be simplifies to two lines:
should_move = !@round.share_sold.include?(shares)
 @round.share_sold << shares if shares.owner != shares.corporation.owner && !@round.share_sold.include?(shares)

lib/engine/game/g_18_zoo/step/buy_sell_par_shares.rb
carlorusso1984: the variable is used to say if a player sold a specific company; it's used to be able to handle the "cannot buy if sold a company"
what I need to save here is the specific shares, because that specific Share should have a specific logic

bentziaxl (approved):

michaeljb:
Hey @carlorusso1984, unfortunately you'll need to re-apply these changes with the new file structure in #9931, which was done to fix #9711. The main change to look out for is that a straight "18ZOO" game can't be created, the particular map has to be selected, so any new fixtures here will need to specify the right title, and modified fixtures should already be moved into the new subdirectory.

Issue #9839, opened 2023-10-30
18 tokaideo, getting error  "undefined method "sort_order_key" for <Engine::Bank> " , unable to proceed


kshgrshrm:
**What's your Game ID?**  https://18xx.games/game/138956

**How to Reproduce the Bug?** Atomsk player has tried every action combination possible but they are not able to complete their turn by presing "Skip companies". As soon as they do they get the error mentioned in the title

<!--
What steps cause the unexpected behavior? pressing "skip companies" 

**Which Browser / Platform?** all platforms.  firefox 119.0

benjaminxscott:
In hotseat I get the likely related error:

`comparison of `[[[1, <Engine::Bank>], 15, "Sleeper Train"], <Engine::Company:ST>] with [[[1, <Engine::Corporation:YSL>],10,"Sasago Railway Tunnel"], <Engine::Company:SRT>] failed`

https://gist.github.com/benjaminxscott/e16ded0f2b41f1bdc757579a1bf8bb0e

![Screenshot 2023-11-02 10 11 03 AM](https://github.com/tobymao/18xx/assets/1711810/625a274f-1f0f-4094-97aa-7981ab0f552a)

andrewzwicky:
I looked into this a little bit.  In this particular game, the Sleeper Train was left behind when not claimed and remained in the Bank.  I'll write a PR to fix.

### Summary
https://18xx.games/game/138956?action=23 at this point in the game, someone buys the second to last private, no one has any money, so no one has valid actions, and the Sleeper Train private remains in the bank.  Later on in the game, the privates need to pay out and there is sorting code that tries to compare the Bank to a Corporation.

There are two issues I can see.
1. The initial stock round allows players to buy/bid on both private companies and normal corporations, which seems out of the ordinary.
2. The privates that are not claimed are not always properly closed.

The correction is probably to make sure companies are closed properly in all cases.

### Details
Fast Forward to later in the game, at the start of the first operating round where privates have been bought in by corporations.  At the start of OR2.2, the games tries to pay out companies to their respective owners.  In the code for paying out companies, there is some sorting code: https://github.com/tobymao/18xx/blob/40c49f97d2eabc50b4d642779d145d3df539be75/lib/engine/game/base.rb#L1480-L1499

[...]

PR #9903, opened 2023-11-16
18 tokaido fix sort key issue

### LINKED ISSUE #9902: [18Tokaido] GameError: Blocking step Sell then Buy Shares cannot process action 120
**What's your Game ID?**
https://18xx.games/game/138029

**How to Reproduce the Bug?**
https://18xx.games/game/138029?action=119
 It just says sorry this game cannot be continued. Only actions this SR4 have been buying shares.

**Which Browser / Platform?**
Chrome: 119.0.6045.124

doubt72:
Fixes #9902, backs out the breaking change for #9839 and fixes that underlying issue.

### Before clicking "Create"

### Implementation Notes

There's a sort in `payout_companies` that breaks when both the bank and corporations own anything (works fine if it's only one or the other).  @andrewzwicky tracked that down, but his change both broke a bunch of existing games where the corporations weren't already removed, and also...  They're not supposed to be removed.  So I backed out that change and fixed the actual sort issue.

TBH, I think the version of `payout_companies` I put in my game object could go in the main engine (whether you filter out the bank owned companies or short-circuit the payout loop, the result is the same), but in the interests of not potentially breaking anything else, I just did the minimal thing.

### DIFF
--- a/lib/engine/game/g_18_tokaido/game.rb
+++ b/lib/engine/game/g_18_tokaido/game.rb
@@ -299,12 +299,8 @@ module Engine
               reorder_players
               new_stock_round
             when Engine::Round::Auction
-              init_round_finished
               new_stock_round
             when Engine::Round::Stock
-              # If the initial round is a normal stock round,
-              # ensure that companies are closed properly if unclaimed.
-              init_round_finished unless @draft_finished
               @operating_rounds = @phase.operating_rounds
               @need_last_stock_round = false
               reorder_players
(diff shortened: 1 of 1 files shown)


andrewzwicky:
Ouch, apologies for that ☹.  The fact that `init_round_finished` existed already and was used in the other initial rounds made me think it was forgotten.

doubt72:
No need to apologize, you responded to the bug faster than me and did find the underlying problem, so "fixing the fix" was easy.

PR #9919, opened 2023-11-21
[18GB] add reserved leasable 6X for corps in receivership

### LINKED ISSUE #9877: [18GB] game stuck w/ error: undefined method `owner' for nil"
Game will not load

(https://18xx.games/game/136880?action=830)

game will not load. We were in final round. 18GB- so it ends when there are no trains to buy or when you get a stock to the top of the track.
We had two companies which did not have CEO's and it was time for one to run. It went into the NPC's turn and would no longer load the game.

I am using Safari.

benjaminxscott:
Fixes #9877

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Adds a reserved 6X to the roster, to be used when a corp in receivership has to 'lease' a train to run its routes

Since the 'leased' train is not transferred to the corp, it's used for routing directly from the depot and its owner is `@depot` just like the prior implementation of 'leased trains'

Should not require pins, as any game that was in this state was broken anyway, this may actually unblock other games

Does not affect the roster when viewed by a player, the 'reserved' 6X is not visible

* **Screenshots**
image

image

Screenshot 2023-11-21 12 08 51 PM

image

* **Any Assumptions / Hacks**


tobymao:
fix the tests

benjaminxscott:
Ack

-_Ben

andypymont:
Thanks Benjamin for working on this issue and trying to fix my oversight in the original implementation.

I was discussing this issue with Dave Berry, the 18GB designer. His opinion was:

> Regarding the question of leasing a train when none is available, the rules are indeed silent on this.  In a face-to-face game, it really doesn't matter: the game will finish at the end of the OR and nothing the insolvent company can do will affect the outcome.  So, in the rare occasions that it happens, the players can just ignore it.
>
> For 18xx.games, of course, the code needs to handle this case to avoid hanging.  The bug report notes that someone tried to fix it by adding a dummy 6X train and having the company lease that.  It may be simpler just to say that if there are no trains left, the company doesn't run a train at all.  Either way should suffice, as the actions of this company won't affect the outcome of the game.

My suggestion therefore would be that we just have a corporation (== "company" in Dave's text above and the 18GB written rules) not lease a train at all if there isn't one in the bank. This will be a simpler change to implement.

(I'm also more than happy for you to make this pass the tests, especially if you've done it already. Thanks again.)

benjaminxscott:
Ah sure that's an easier fix than the current approach (which should pass
tests now)

I didn't grok that a corp leasing a train in that situation was basically a
noop

-_Ben

benjaminxscott:
closed in favor of #10095

Issue #6940, opened 2022-02-04
[1849] Autopass immediately fails if double share available


KevinR1000:
See https://18xx.games/game/70047?action=95

A double share is available in 1849.
If I try to autopass from the main screen, it immediately fails due to the double share.

![Screenshot_2022-02-03-20-08-49](https://user-images.githubusercontent.com/82781561/152460774-b4b2ea59-1077-4ca9-98bb-0aa3f42496f3.png)

philcampeau:
The problem here is, the method that checks for this has no way of knowing that you don't want that certificate. I don't think there's a simple way of implementing this without adding new checks specifically for 1849, and even then, you'd need one for each corp, since while you might not want the 20% of IFT, maybe you do want the 20% cert of another corp if that becomes available. If you really don't, you can use the "Unconditional pass" option in the Auto menu.

There has been no chatter about this since it was reported almost 4 years ago. I'm going to close this bug as "not planned".

PR #9925, opened 2023-11-23
[1849] autopass tweak

### LINKED ISSUE #8384: [1849] Auto-pass should not disable when double share available in stock I've previously sold
See for example complete game https://18xx.games/game/94678?action=455

This occurred several times in the recent round of the 1849 tournament.  I selected auto-pass in the stock round and the auto-pass was disabled when a double share became available, even though I wasn't able to buy that share because I'd previously sold stock in that company earlier in that stock round.  This kept slowing down the game because I had to log on and reselect the auto-pass option.  

The simple fix would be to add a test to the code that checks for a double share being available, to additionally check that the player hasn't previously sold stock in that company earlier in the SR.

Thinking about edge cases, I guess a player could have an option to act if the double-share had been in the market and they have another share in the company that they could sell to make the double share unavailable for the next player.  I'm not sure that would be worth coding for.  

If you do want to code for this, then I think the pseudocode would be:
```
if a double share available
      if the player has not previously sold stock in that company then autopass=off
      else if the double share is in the treasury then autopass=remain
             else if the player owns stock in the company then autopass=off
                   else autopass=remain
```

philcampeau:
Fixes #8384

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Currently, autopass always disables if the 20% final cert is available for purchase and you can afford it. 

This update first checks if the share is in the treasury, and then verifies if the active player has sold shares this round. If yes, autopass remains enabled. It then checks if the share is available from the bank pool. In that case, autopass is disabled even if the player has previously sold shares, in the event that the active player wants to sell a share to block the purchase of the double share. 

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_1849/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1849/step/buy_sell_par_shares.rb
@@ -71,9 +71,16 @@ module Engine
             if !share_to_buy || !share_to_buy.last_cert
               @game.corporations.each do |corporation|
                 share = corporation.ipo_shares.find(&:last_cert)
+                if share && @game.last_cert_last?(share.to_bundle) &&
+                  corporation.holding_ok?(entity, share.percent) &&
+                  !@round.players_sold[entity][corporation]
+                  return "20% double cert for #{corporation.name} is currently available from the treasury"
+                end
+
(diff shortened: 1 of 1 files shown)


philcampeau:
Notre, this does not fix the related bug #6940

PR #9936, opened 2023-11-26
[1822CA] fix QMOO / Quebec

michaeljb:
QMOO's home can be any Quebec city, but actions of other corporations may end up limiting QMOO's options. Usually when the first yellow tile is laid in Quebec, QMOO will have to choose its home; now the pending token step will activate appropriately. This also adds logic to update the home reservation, to ensure that a specific city is reserved for QMOO if necessary.

Adds a function to the core game class to hide hex reservations for corporations, to declutter the Quebec tile a bit when QMOO is selecting its home.

Adds loc-specific rendering for destinations, for ICR destination on Quebec.

#9376


assets/app/view/game/part/small_item.rb
```diff
@@ -28,8 +28,23 @@ def preferred_render_locations_by_loc
               preferred_render_locations
             end
           else
-            @loc = nil
-            preferred_render_locations
+            case @loc.to_s
```
tobymao, line 31: why do you convert it to a string?

tobymao (approved):

assets/app/view/game/part/small_item.rb
michaeljb: Because the `case` on line 13 also does this.

After checking a bit more, some locs are defined as strings but sometimes they end up as floats - https://github.com/tobymao/18xx/blob/06954baaf8fa7f22746b365dd1e6d575d0ebe84f/lib/engine/tile.rb#L415-L419
It looks like the only code that calls `compute_loc()` is in `compute_city_town_edges()`, which shouldn't interact with "small items" so I think this can be removed. If removing it causes problems I'll add it back with some comments explaining why it's needed.

PR #9958, opened 2023-11-30
[18FL] code cleanup

philcampeau:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

As I was looking through the code to understand how something worked, I noticed that the POSC private had been given two different variables in the game code (both token_company and steamboat). token_company was only used in one other place in the code, so I removed it and updated its name in the other place. 

* **Screenshots**

* **Any Assumptions / Hacks**


philcampeau:
Added a second commit that removed lines 250-253 of game.rb and simplified it. My guess was this was borrowed and modified code from 1846, but there's no need to multiply the revenue by the number of port icons, since all port hexes only have one icon. 

I also added 3 real-game fixture files. None of these changes should affect any existing or completed games.

ventusignis:
> Added a second commit that removed lines 250-253 of game.rb and simplified it. My guess was this was borrowed and modified code from another game, but there's no need to multiply the revenue by the number of port icons, since all port hexes only have one icon.
> 
> I also added 3 real-game fixture files. None of these changes should affect any existing or completed games.

yeah, this private was copied from 1846 originally hence the messiness; thanks for the cleanup!

PR #10015, opened 2023-12-13
[multiple] Adjust CSS to not add extra bar under minor companies.

andrewzwicky:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

    In fbcd3ed `min-height:100px` was added to the `.card` css to ensure that the empty slots in the 1817 auction were not too short (my assumption).  In other games, this adds tiny bars to some of the minor corp cards that don't have much info.  This change reverts the `min-height` and replaces it with a description of a few newlines for those empty slot companies.  To allow this to work, the description style now has the `whiteSpace: 'pre-line',` property, which is identical to `normal` except that it preserves newlines.  I didn't see any newlines in any other company descriptions.

* **Screenshots**

    #### 18EU & 1880
    | Old      | New |
    | ----------- | ----------- |
    | image | image |

    #### 1817 Empty Slot (Changes practically invisible, Empty Slot Card is bigger)
    | Old      | New |
    | ----------- | ----------- |
    | image | image  |

* **Any Assumptions / Hacks**
    My assumption is that 1817 is the only place where cards need to be inflated like this.

### DIFF
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -104,6 +104,7 @@ module View
             fontSize: '80%',
             textAlign: 'left',
             fontWeight: 'normal',
+            whiteSpace: 'pre-line',
           }
 
           value_style = {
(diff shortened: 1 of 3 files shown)


andrewzwicky:
could have sworn it passed on my machine, will fix in a sec

PR #10016, opened 2023-12-13
[1825] corps can only pay multiples of 10 for crossbuys

### LINKED ISSUE #7545: [1825] Smallest denomination of Company Credits is 10 (not 1)
**What's your Game ID?**

https://18xx.games/game/77435

In OR8.1 the FR bought a 5-train from the L&Y for 681 Company Credits.
This is against the Rules since the smallest denomination of Company Credits is 10 so that each purchase of a train from another company must have a price that is a multiple of 10.

philcampeau:
Fixes #7545

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

In 1825, corps are only supposed to buy trains in multiples of £10. This code enforces that. 

Unfortunately, the only fixture file for 1825 had a train buy for a non-multiple of 10, and even looking at 10 different completed games on the site, I didn't find a single instance of a game where this wasn't done. I'll set a reminder for myself to add a couple of fixtures in a month once some new games have been completed with this rule in place. 

* **Screenshots**

* **Any Assumptions / Hacks**


roseundy (approved):
Looks good to me.  It would be nice if there was an existing game that doesn't violate this rule so you could add it as a replacement fixture.

philcampeau:
I'm going to say let's hold off on merging this just yet. I just ran a rake test with a game where all the cross buys were in multiples of 10, and it still gave an error.

philcampeau:
This update is ready to merge, but I don't know how to pin games, or if I even have that access.

Issue #9945, opened 2023-11-27
[18EU] "undefined method `+' for nil" when passing on a minor with no bids


richpri:
Game link is https://18xx.games/game/142349

**What's your Game ID?**  142349? It is a 18EU game

**How to Reproduce the Bug?**
As richprice click on "Pass (Bid)" in game 142349

**Which Browser / Platform?**
Google Chrome / Linux UBUNTU 

Chrome Version 119.0.6045.159 (Official Build) (64-bit)

Ubuntu version:
rich@richdesktop:~$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy

Some additional information  from chat box:

[20:34] richprice: @JGould I am getting an error message when I try to pass this bid.
-- 2023-11-26 --
[08:22] richprice: @JGould @NY_NAFO @CraigBee Due to Bug I cannot make a move. Should we abandon this game? If not then what should we do?
-- 2023-11-27 --
[06:33] richprice: @JGould @NY_NAFO @CraigBee Still getting 'undefined method' error when I select Pass (Bid). The game is definitely broken.
[06:35] richprice: @JGould @NY_NAFO @CraigBee I am going to end this game. I hope that we will play other games in the future.

I can be contacted at richpri46@gmail.com

andrewzwicky:
This was fixed in #10017

PR #10017, opened 2023-12-14
[18EU] show bidders on corporation cards

andrewzwicky:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

    Noticed while looking at #9945, because that minor does have bids on it, but they're not shown, so it's tough to tell.

    In an `Auction` round, if minors are being displayed, show bids if there are any for that minor.  Most applicable for 18EU.  I couldn't find another game with the right set of criteria to benefit from this.  Some don't use actual minors, just corporations with the type "minor", and some use other rounds that don't inherit from Auction.

* **Screenshots**

    #### 18EU
    | Old      | New |
    | ----------- | ----------- |
    | image | image |

* **Any Assumptions / Hacks**

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -166,7 +166,7 @@ module View
             textAlign: 'center',
             fontWeight: 'bold',
             padding: '0.5rem',
-          }
+          },
         }
 
         rows = @bids
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -13,6 +13,7 @@ module View
       include Lib::Settings
 
       needs :user, default: nil, store: true
+      needs :bids, default: nil
       needs :corporation
       needs :selected_company, default: nil, store: true
       needs :selected_corporation, default: nil, store: true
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_eu/step/modified_dutch_auction.rb
```diff
@@ -91,6 +91,9 @@ def pass_auction(entity)
 
           def remove_from_auction(entity)
             @active_bidders.delete(entity)
+            @bids[@auctioning]&.reject! do |bid|
+              bid.entity == entity
+            end
```
andrewzwicky, line 96: Without this change, bids would just linger in the UI until the auction was concluded.

assets/app/view/game/corporation.rb
```diff
@@ -147,6 +149,59 @@ def render
         h('div.corp.card', { style: card_style, on: { click: select_corporation } }, children)
       end
 
+      def render_bidders
```
andrewzwicky, line 152: Is there a way to combine this with render_bidders from the `Company` view?  I didn't see an appropriate file that both included, so I just copy/pasted the code with minor styling differences for padding and center text to match the Company look.

PR #10041, opened 2023-12-26
[18SJ] Add check to Step::AssignGotaKanalbolaget

ollybh:
When selling shares in emergency money raising, `Step::AssignGotaKanalbolaget.actions` was being called with a player entity, causing an error when `@game.abilities` was called. This adds a check to make sure the entity is a corporation.

Fixes tobymao#9759.


ollybh:
I'm not sure why the checks failed for this. Running rake/rubocop locally works fine.

PR #10044, opened 2023-12-27
[1858] Clear round.current_actions after an auction

ollybh:
`@round.current_actions` was not being cleared when there was a private railway company auction. This could lead to the wrong player being given the priority deal if the following sequence occurred:
1. There was a sell or buy shares action.
2. There were one or more auctions.
3. Everyone passes.

After the auction the `{Sell,Buy}Shares` action was still in `@round.current_actions` when the active player became the person to the right of the one who started the auction. This means that the priority deal would be one place to the right of where it should have been.

This is fixed by clearing `current_actions` in `win_bid`.

Fixes tobymao#9968.

There are at least three games that are broken by this change: 139169, 140242 and 143291. There might be a few others that I have missed.


benjaminxscott:
Could you merge master in and re-push so it picks up the new Gemfile to see if that fixes the build issue?

ref: https://github.com/tobymao/18xx/pull/10049

PR #10049, opened 2023-12-28
attempt to fix GH actions build issue

benjaminxscott:
Github Action builds are failing, while local builds were working. 

Possibly due to a compiled dependency missing in Ruby 3.x

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**


benjaminxscott:
After this is merged, we need to retrigger the other PR runs

PR #10053, opened 2023-12-29
[1830] adds 3rd 6-train option

### LINKED ISSUE #10046: 1830 - Missing optional 6 train
Game ID: 146318 /https://18xx.games/game/146318

When playing asyc with two physical copies of the game board, the original AH rules allow for an optional 6 train. We played up to the point where the 3rd 6 or first Diesel needed to be bought, and not having that as an option created a cascade of events durastically altering the remainder of game play.

Could the option of adding a third six train be present in the original game, or as a toggle button option when setting up a new game? Also, is there a way to make it optional where you don’t need to buy it and you can go straight to buying a diesel? 

Filing as a bug because it doesn’t match the original rules, but I can understand if this needs to be filed as an enhancement instead.

philcampeau:
Fixes #10046

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
The Avalon Hill rules did include an option to add a 3rd 6-train. This is also referenced in the Mayfair rules:

image

### DIFF
--- a/lib/engine/game/g_1830/game.rb
+++ b/lib/engine/game/g_1830/game.rb
@@ -158,7 +158,7 @@ module Engine
                     num: 3,
                     events: [{ 'type' => 'close_companies' }],
                   },
-                  { name: '6', distance: 6, price: 630, num: 2 },
+                  { name: '6', distance: 6, price: 630 },
                   {
                     name: 'D',
                     distance: 999,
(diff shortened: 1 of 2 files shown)


philcampeau:
No pins needed

philcampeau:
I got a pic of the optional rules section. You're able to use all three changes listed, or any of the three. I don't think there'll be much demand for the others, but I can implement them too if desired.

image

michaeljb (approved):

PR #10066, opened 2023-12-31
fix random seed with new `seed_v2` db settings field

### LINKED ISSUE #7465: Seed specification not functional.
The game json has the specified seed, but it is not loading the same information on new game creation. Despite using the same seed, I am getting different random shares for the random share ability.

michaeljb:
* past games have a `seed` that is set when the game is started, that is only used for player order
* the new `seed_v2` is used for player order, as well as the in-game `@seed` used by `rand`
* new games don't have `seed`, but old games with `seed` and not `seed_v2` will continue to work as before, with `@seed` being set based on the game's ID (so no migrations needed)

Fixes #7465


tobymao:
what is used for in game seed??

michaeljb:
The original seed is just used for player order, #7465 talks about getting repeatable random elements in the game, for example in 18Chesapeake which corporation's presidents share goes with the Cornelius Vanderbilt company.

tobymao (approved):

crericha:
Overall looks good. 

For what its worth...I think its possible to not replace `seed` with `seed_v2` if you write script that modifies the settings of all existing games to 1) add `player_order` and 2) make `seed` nil.

crericha (approved):

tobymao:
maybe let’s do that instead?

michaeljb:
Player order is an existing field, and I’ve got a script now that I think works for this, I’ll update the PR tomorrow

michaeljb:
Changes were significant enough that I made a new branch instead of modifying this one.
 #10070

PR #10070, opened 2023-12-31
use `settings['seed']` for `@rand` implementation

### LINKED ISSUE #7465: Seed specification not functional.

michaeljb:
* old behavior: the `seed` in the DB settings was only used for randomizing turn order, and the game object's `@seed` was initialized with the numerical part of the game's ID
* new behavior: the `seed` is used for randomizing turn order and initializing `@seed`
* functions in `scripts/migrate_seed.rb` will be needed on the server to update `settings['seed']` and `settings['player_order']` for all currently active or finished games
* functions in `scripts/migrate_seed_hotseat.js` can be used in client browsers to migrate hotseat games
* hotseat games on devices where the browser console cannot be opened, such as mobile phones, cannot be migrated
* add announcement to the welcome page with a [link to the wiki][0]
* the random seed can be copied from the Tools page, and the [wiki link][0] is there as well

Fixes #7465 

[0]: https://github.com/tobymao/18xx/wiki/Random-Seeds


michaeljb:
Forgot I need to migrate the fixtures, will reopen when that's done.

michaeljb:
Fixtures have been updated using `scripts/migrate_seed_fixtures.rb`

tobymao (approved):

assets/app/view/game/game_data.rb
```diff
@@ -74,8 +74,29 @@ def render_game_data_buttons
           buttons << h(:label, 'to play around in hotseat mode')
         end
 
+        buttons.concat(render_random_seed)
+
         h('div.margined', buttons)
       end
+
+      def render_random_seed
+        return [] unless (@seed = @game_data&.[](:settings)&.[](:seed))
```
crericha, line 83: Does dig not work here?

lib/engine/game/base.rb
```diff
@@ -607,12 +620,17 @@ def initialize(names, id: 0, actions: [], at_action: nil, pin: nil, strict: fals
         @log << '----'
       end
 
+      def initialize_seed(seed)
+        @seed = (seed || @id.to_s.scan(/\d+/).first.to_i) % RAND_M
```
crericha, line 624: Should the seed itself be modulo'd by `RAND_M` or only @rand?

assets/app/view/game/game_data.rb
michaeljb: Good call, switched to dig.

lib/engine/game/base.rb
michaeljb: When I did that I thought there would be no difference since the operations in `rand` apply the mod operation, but the turn order shuffling done on the ruby side does use the seed without modding it. So it's a minor difference, but doesn't hurt to move it to `@rand`, done.

crericha (approved):

PR #10071, opened 2023-12-31
rysnc the pinned directory on deploy, clean unused pins in queue

michaeljb:
* rsync `public/pinned` on deployment so the current release always shows up in pins directory, simplifying the process to pin games
* clean up unused pin files daily in `queue.rb` * keep the newest few around so that just-deployed pins aren't cleaned up before they can be used

### DIFF
--- a/queue.rb
+++ b/queue.rb
@@ -3,6 +3,7 @@
 require 'logger'
 require 'rufus-scheduler'
 require 'require_all'
+require 'sequel'
 require_relative 'models'
 require_rel './models'
 require_relative 'lib/assets'
@@ -10,6 +11,8 @@ require_relative 'lib/bus'
 require_relative 'lib/hooks'
 require_relative 'lib/mail'
 require_relative 'lib/user_stats'
+
+Sequel.extension :pg_json_ops
 
 PRODUCTION = ENV['RACK_ENV'] == 'production'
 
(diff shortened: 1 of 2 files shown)


queue.rb
```diff
@@ -21,6 +24,8 @@
 
 scheduler = Rufus::Scheduler.new
 
+PINS_TO_KEEP = 3
```
tobymao, line 27: Shouldn’t you keep all pins that are still being played?

michaeljb: This constant ensures the latest pins are kept even if they're not used, otherwise all unused pins are discarded. I've added a comment and updated the constant name to clarify this.

tobymao (approved):
i don’t think we need to keep any pins that aren’t used

michaeljb:
When a new release is deployed, that pin file won't be used at all, it might not be used till the next release.

If you have some games that need to be pinned to release A when you deploy release B, but you don't pin until after deploying B, there will be a small window where neither A nor B is used at all but you don't want either of them cleaned up, so I think keeping 2 around makes sense.

I set it to 3 as a fallback in case a game needs to be pinned to a version other than the latest, but cases where that happens and we can find the right version are probably rare enough that it's fine to do rebuild and deploy the pin when it does happen. I'll change the constant value down to 2.

PR #10089, opened 2024-01-03
18GB: Prevent special token powers being used illegally

### LINKED ISSUE #9262: [18GB] Company uses GN power without closing it (rules infraction)
In a current game of GB, the LYR used the GN power to lay a token in York without closing the private. (It did not 'parachute' into York, so no fee needed to be paid.) This was in OR2.1
https://18xx.games/game/123733

The rules say (section 9):
P3: Great Northern (GN) Cost: £70 Income: £25/OR
The GN allows a company to lay the GN station token in York (I14) [9.2]. The GN owner may use this ability once per game, **after the GN has closed**, for any company which they control

andypymont:
Fixes #9262

### Implementation Notes

* Private company powers in 18GB can only be used once the company has been closed (by using the `choice` ability on the company) by its owner. Therefore in the `abilities` function in `game.rb` for 18GB, abilities are filtered based on whether or not the company has been closed
* However, the standard `special_token` step (from which 18GB's `special_token` step inherits) was bypassing this check with its own `ability` function, allowing a token to still be placed at some times when it should not
* Therefore, this change updates 18GB's `special_token` step to filter out the abilities based on company status and avoid a token ability being used illegally

### DIFF
--- a/lib/engine/game/g_18_gb/step/special_token.rb
+++ b/lib/engine/game/g_18_gb/step/special_token.rb
@@ -13,6 +13,20 @@ module Engine
             super
           end
 
+          def ability(entity)
+            return unless entity&.company?
+
+            @game.abilities(entity, :token) do |ability, _company|
+              return ability unless entity.value.positive?
+            end
+
+            @game.abilities(entity, :teleport) do |ability, _company|
+              return ability if !entity.value.positive? && ability.used?
+            end
+
+            nil
+          end
+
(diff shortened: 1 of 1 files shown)


andypymont:
I don't have the access to add labels to this PR, but this will be a breaking change for the two games mentioned in #9262 and any others where the power was used with an inappropriate timing.

michaeljb (approved):

PR #10095, opened 2024-01-04
[18GB] receivership corps can't lease a train if there are none

### LINKED ISSUE #9877: [18GB] game stuck w/ error: undefined method `owner' for nil"

benjaminxscott:
Fixes #9877

### Before clicking "Create"

### Implementation Notes

If a corp in receivership would have to lease a train but there are none, it doesn't get to lease one at all

This should be a no-op in practice, as the game ends soon after the final train is purchased - and receivership corps just withhold to buy a train anyway (which never happens as there are none)

Shouldn't require pins as this PR handles an edge case that would have broken existing games

See linked older PR for more details

### DIFF
--- a/lib/engine/game/g_18_gb/game.rb
+++ b/lib/engine/game/g_18_gb/game.rb
@@ -720,7 +720,7 @@ module Engine
         def route_trains(entity)
           return super unless insolvent?(entity)
 
-          [@depot.min_depot_train]
+          @depot.upcoming.any? ? [@depot.min_depot_train] : []
         end
 
         def express_train?(train)
--- a/lib/engine/game/g_18_gb/game.rb
+++ b/lib/engine/game/g_18_gb/game.rb
@@ -720,7 +720,7 @@ module Engine
         def route_trains(entity)
           return super unless insolvent?(entity)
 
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_gb/game.rb
```diff
@@ -720,7 +720,7 @@ def legal_tile_rotation?(_entity, _hex, tile)
         def route_trains(entity)
           return super unless insolvent?(entity)
 
-          [@depot.min_depot_train]
+          @depot.upcoming.any? ? [@depot.min_depot_train] : []
```
ollybh, line 723: `Array.any?` (and `Array.none?`) shouldn't be used without a code block testing the array elements. To test whether an array has any elements you should use `Array.empty?`.
```suggestion
          @depot.upcoming.empty? ? [] : [@depot.min_depot_train]
```

Issue #10023, opened 2023-12-16
18AL - 50% train discount from NDY private not being applied after selection


saccenti:
I'm stuck in an 18AL game: https://18xx.games/game/144868

Specifically at this point: https://18xx.games/game/144868?action=137

I'm not able to purchase a train with New Decatur Yards. If I understand the rules, I should at this moment be able to purchase the 4T for $150.

saccenti:
See also here: https://github.com/tobymao/18xx/issues/9686

saccenti:
Phil, why did you remove "needs triage" label, without any comment? This is probly a misunderstanding on my end(?) but I don't see how this issue has been addressed.

dfannius:
"needs triage" means that the issue hasn't been classified yet. It has now been classified with the labels "18AL" and "abilities".

Issue #9192, opened 2023-05-19
[18MS] Not being able to get apply the Mobile and Ohio private company ability


Velgarath:
Hi!
I am not being able to apply the discount, There is a special button that you can click before the train purchase, but the discount is not applied in the purchase
![pic](https://github.com/tobymao/18xx/assets/134040689/4b7c002a-b73b-48ba-aa87-03ca6a51ea3a)

Just in case you need it, I experienced the problem in game 123063
Thanks

michaeljb:
* https://18xx.games/game/123063?action=74
* [gist](https://gist.github.com/michaeljb/ceae160d1042ea41a39d6125046f6a16)

JH445:
Experiencing the same issue in my current 18MS game (https://18xx.games/game/135267). I'm trying to buy a 4+ train, but there doesn't seem to be an option to apply the private company's discount.
![18ms-clip](https://github.com/tobymao/18xx/assets/11559050/96a05b9a-fb46-498e-b07d-17474bc8c695)

Yuki-yao:
triggered same bug in [gist](https://gist.github.com/Yuki-yao/967d6d5669c1925554e73a37faf09f79)

jtamplin:
This impacted https://18xx.games/game/145138 - at 12:35 in OR 2/10, FR tried to use the discount on a 4+ train.

saccenti:
This bug is still present as of January 2024. Here's an example:

https://18xx.games/game/150322?action=59

PR #10098, opened 2024-01-05
Fix Tredgar in 18VA

### LINKED ISSUE #9686: [18VA] Tredegar Iron Works doesn't give discount
**What's your Game ID?**

https://18xx.games/game/135758

Here's the game data as a gist
https://gist.github.com/dtschirmer/c8e66ccfbea4cbfd56b60705069bee2c

**How to Reproduce the Bug?**

During OR 4.2,  while operating VR, I clicked on the Tredegar Iron Works ability button, and the price of the train I was going to buy did not change (should be -$200).
I'm not sure how the price change is actually supposed to happen in the game, but I wasn't able to purchase a train for a lower value. Looking at another game's data where it did work, the json just showed a lower price. No matter what I tried (clearing browser cache, logging out and in, updating browser, taking off Firefox protection) I couldn't get a lower price.

**Which Browser / Platform?**

Firefox 117.0.1 (64-bit)
Mac OSX 12.6.1

ventusignis:
Fixes #9686 

### Before clicking "Create"

- [ x] Add the `pins` label if this change will break existing games

### Implementation Notes
* **Explanation of Change**

Changing the timing from `buying_train` to `buy_train` fixes Tredgar; looking at the `ability_right_time?` implementation it looks like `buying_train` isn't the name of a step or explicitly stated so it doesn't get past this

https://github.com/tobymao/18xx/blob/master/lib/engine/game/base.rb#L3082
* **Screenshots**

* **Any Assumptions / Hacks**
Are other games that use `when: 'buying_train'` broken?
Looking at this and the blame for game::base, I'm not sure how Tredgar ever worked

### DIFF
--- a/lib/engine/game/g_18_va/entities.rb
+++ b/lib/engine/game/g_18_va/entities.rb
@@ -45,7 +45,7 @@ module Engine
                 trains: %w[2 3 4 5 6 4D],
                 count: 1,
                 closed_when_used_up: true,
-                when: 'buying_train',
+                when: 'buy_train',
               },
             ],
           },


michaeljb:
> Are other games that use `when: 'buying_train'` broken?

Hmm, looks likely: #9192, #10023

michaeljb (approved):

PR #10101, opened 2024-01-05
[18EU] Leading bidder is removed from auction early

andrewzwicky:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

    While investigating #9945 I saw that when a bidder has bid their maximum cash, they are removed from the auction as "can't afford" (because they would not be allowed to bid again).  This is not correct in the case where they had bid all their remaining money, and all other players want to pass, as the second place bidder would win the auction.  This change leaves them in the auction.  If someone else does overbid them, _then_ they are properly removed from the auction. 

    I tested the following scenarios:
    * Everyone passes, maxed out bidder wins ✔
    * Someone outbids later, maxed out bidder is removed ✔

    I imported all active 18EU games and validated them, so I don't think this requires pins.

* **Screenshots**

    |     **Wrong** |
    | ----------- |
    |     pass_removes |

    |     **Correct** |
    | ----------- |
    |         pass_leaves |

* **Any Assumptions / Hacks**


lib/engine/game/g_18_eu/step/modified_dutch_auction.rb
```diff
@@ -102,6 +102,7 @@ def next_entity!
             entity = entities[entity_index]
             return next_entity! if entity&.passed?
             return unless @auctioning
+            return if @bids[@auctioning]&.select { |bid| bid.entity == entity }&.any?
```
ollybh, line 105:
```suggestion
            return if @bids[@auctioning]&.any? { |bid| bid.entity == entity }
```

michaeljb (approved):

PR #10103, opened 2024-01-06
[1832] updates to code

philcampeau:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
More updates to code. adds System corporations to game, free terrain cost for home tile, FECR's Key west token, and more. 

* **Screenshots**

* **Any Assumptions / Hacks**


lib/engine/game/g_1832/game.rb
```diff
@@ -148,8 +148,14 @@ class Game < Game::Base
...
 
-        TILE_LAYS = [{ lay: true, upgrade: true, cost: 0, cannot_reuse_same_hex: true },
-                     { lay: :not_if_upgraded, upgrade: false, cost: 0 }].freeze
+        TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false }].freeze
+        SYSTEM_TILE_LAYS = [{ lay: true, upgrade: true },
+                            { lay: :not_if_upgraded, upgrade: false },
+                            { lay: true, upgrade: :not_if_upgraded, cannot_reuse_same_hex: true }].freeze
+
+        def tile_lays(_entity)
+          @entity.system? ? SYSTEM_TILE_LAYS : TILE_LAYS
```
ollybh, line 157:
```suggestion
          @entity.type == :system ? SYSTEM_TILE_LAYS : TILE_LAYS
```
`entity.system?` is going to be `false` for all corporations, including systems (AFAICT, only 1828’s System class would return `true` for this).

lib/engine/game/g_1832/game.rb
```diff
@@ -242,6 +248,43 @@ def setup
...
+        #         @operating_rounds = @phase.operating_rounds
+        #         reorder_players
+        #         new_operating_round
+        #       end
+        #     when G1832::Round::Merger
+        #       new_operating_round
+        #     when Engine::Round::Operating
+
+        # end
+
```
ollybh, line 287: Please remove this block of commented out code. Having dead code committed would make the git history messier than it needs to be.

lib/engine/game/g_1832/round/merger.rb
```diff
@@ -0,0 +1,72 @@
...
+#           end
+
+#           def force_next_entity!
+#             # called from the close corporation, ignore it
+#           end
+#         end
+#       end
+#     end
+#   end
+# end
```
ollybh, line 72: Adding a minimal version of this class would be better than all this comment-out code. Or you could remove this file entirely for the moment, and change `G1832::Game.merge_rounds` to use `Engine::Round::Merger` until you have implemented this class.
```suggestion
# frozen_string_literal: true

require_relative '../../../round/merger'

module Engine
  module Game
    module G1832
      module Round
        class Merger < Engine::Round::Merger
        end
      end
    end
  end
end
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,57 @@
...
+
+require_relative '../../../step/token'
+
+module Engine
+  module Game
+    module G1832
+      module Step
+        class Token < Engine::Step::Token
+          def actions(entity)
+            actions = super.dup
```
ollybh, line 11:
```suggestion
            actions = super
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,57 @@
...
+            actions.uniq
+          end
+
+          def choice_name
+            'Additional Token Actions'
+          end
+
+          def choices
+            choices = []
+            choices << ['Place Key West Token'] if can_place_miami_token?(current_entity)
```
ollybh, line 23: I'd recommend using the hash version for the choices here, something like:
```suggestion
            choices = {}
            choices['keywest'] = 'Place Key West Token' if can_place_miami_token?(current_entity)
```
The reason is that by using the array version, it's the button label that gets stored in the game JSON for the choice action. If you use the hash version, then it's the choice key, not  the value (button label) that is stored. So you can later tweak the button text without breaking existing games.

It's probably not a big deal here, so feel free to ignore this one.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,57 @@
...
+            choices
+          end
+
+          def process_choose(action)
+            place_miami_token(action.entity) if action.choice == 'Place Key West Token'
```
ollybh, line 28: Slightly more defensive:
```suggestion
            raise GameError, 'Illegal choice' unless action.choice == 'Place Key West Token'
            
            place_miami_token(action.entity)
```
If you have changed the choices used to be a hash rather than an array, you'll need to change the test here to look for the hash key.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,57 @@
...
+            corporation.spend(ability.cost, @game.bank)
+            @log << "#{corporation.name} places the Key West token on #{hex.name} for #{@game.format_currency(ability.cost)}"
+          end
+
+          def skip!
+            pass!
+          end
+
+          def can_place_miami_token?(entity)
+            (ability = @game.abilities(entity, :assign_hexes)) &&
```
ollybh, line 48: If there any chance that `entity` could have more than one `:assign_hexes` ability? If it does, then this code is going to cause errors: `@game.abilities` can either return a single `Ability` or an array of them (or nil). Your code will break if it gets an array.

ollybh (changes requested):

lib/engine/game/g_1832/game.rb
philcampeau: fair! I have a habit of leaving stuff I'm working on like that, but yeah it doesn't help anything for legibility.

lib/engine/game/g_1832/game.rb
philcampeau: How about this: 

```
        def tile_lays(entity)
          return self.class::SYSTEM_TILE_LAYS if system?(entity)

          self.class::TILE_LAYS
        end

        def system?(corporation)
          return false unless corporation

          corporation.type == :system
        end
```
obviously, with type: 'system' in the entities file.

lib/engine/game/g_1832/step/token.rb
philcampeau: sure thing

lib/engine/game/g_1832/step/token.rb
philcampeau: I borrowed this code from elsewhere, hadn't even considered that. Updated to clearly find this ability.

>           def can_place_miami_token?(entity)
>             (abilities = @game.abilities(entity, :assign_hexes)) &&
>             abilities.any? { |ability| ability.cost == 100 } &&
>             entity.cash >= 100 &&
>             (3..6).cover?(@game.phase.name.to_i)
>           end

lib/engine/game/g_1832/game.rb
ollybh: Yes, that would work.

lib/engine/game/g_1832/step/token.rb
ollybh: Now you are assuming that you'll get an array returned from `@game.abilities`. If there's just one matching ability it returns an `Ability` object, not an array.
https://github.com/tobymao/18xx/blob/d5e0e80e7b870eccefd5bef31930818fccc705ff/lib/engine/game/base.rb#L1975-L1996
If you're not absolutely sure whether there will be none, one or more matching abilities returned from `@game.abilities` then convert the returned value to an array:
```
@abilities = Array(@game.abilities(entity, :assign_hexes)
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          def choices
+            choices = {}
+            choices['keywest'] = 'Place Key West Token' if can_place_miami_token?(current_entity)
+
+            choices
+          end
+
+          def process_choose(action)
+            raise GameError, 'Illegal choice' unless action.choice == 'Place Key West Token'
```
ollybh, line 29: As you've followed my suggestion of using the hash version of choices, you need to test here against the choice key, not its value.
```suggestion
            raise GameError, 'Illegal choice' unless action.choice == 'keywest'
```

lib/engine/game/g_1832/step/token.rb
philcampeau: awesome, thanks!

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          def skip!
+            pass!
+          end
+
+          def can_place_miami_token?(entity)
+            @abilities = Array(@game.abilities(entity, :assign_hexes)) &&
+            abilities.any? { |ability| ability.cost == 100 } &&
+            entity.cash >= 100 &&
+            (3..6).cover?(@game.phase.name.to_i)
```
ollybh, line 54: Sorry, there was a typo in my last suggestion (a spurious `@`). This method could also be made clearer by removing the `&&`s.
```suggestion
            return false if entity.cash < 100
            return false unless (3..6).cover?(@game.phase.name.to_i)
            
            abilities = Array(@game.abilities(entity, :assign_hexes))
            abilities.any? { |ability| ability.cost == 100 }
```
Also, checking for the ability cost feels a bit weird. This might be clearer if you check that the ability's `hexes` property includes the Miami hex.

philcampeau:
Fixed all the recommendations, and noticed I had defined CORPORATE_BUY_SHARE_ALLOW_BUY_FROM_PRESIDENT twice, so I removed one of the two.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,61 @@
...
+          def skip!
+            pass!
+          end
+
+          def can_place_miami_token?(entity)
+            return false if entity.cash < 100
+            return false unless (3..6).cover?(@game.phase.name.to_i)
+
+            abilities = Array(@game.abilities(entity, :assign_hexes))
+            abilities.any? { |ability| ability.hexes == 'N16' }
```
ollybh, line 55: `hexes` is an array. It would also be better to define a constant for the Miami hex.
```suggestion
            abilities.any? { |ability| ability.hexes.include?('N16') }
```

ollybh: I see you've already got `MIAMI_HEX` defined, so:
```suggestion
            abilities.any? { |ability| ability.hexes.include?(MIAMI_HEX) }
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,61 @@
...
+module Engine
+  module Game
+    module G1832
+      module Step
+        class Token < Engine::Step::Token
+          def actions(entity)
+            actions = super
+            actions += %w[choose pass] if can_place_miami_token?(entity)
+
+            actions.uniq
```
ollybh, line 14: This can be made a tiny bit simpler, and avoid the need to call `uniq`, by using the array union operator.
```suggestion
            return super unless can_place_miami_token?(entity)

            super | %w[choose pass]
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,60 @@
...
+            place_miami_token(action.entity)
+            pass!
+          end
+
+          def place_miami_token(corporation)
+            ability = @game.abilities(corporation, :assign_hexes)
+            hex = @game.hex_by_id(ability.hexes.first)
+
+            hex.tile.icons.reject! { |icon| icon.name == 'FECR_key_west' }
+            hex.assign!(corporation)
```
ollybh, line 39: I'd missed this one when pointing out the problems with `can_place_miami_token?` – this will also break if a corporation has two abilities of type `assign_hexes`. You need to do something similar here, using `Array()` to make sure that you get a list of abilities back from `@game.abilities`, then finding the Miami ability in those.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,63 @@
...
+            place_miami_token(action.entity)
+            pass!
+          end
+
+          def place_miami_token(corporation)
+            return unless can_place_miami_token?(corporation)
+
+            hex = MIAMI_HEX
+            abilities = Array(@game.abilities(corporation, :assign_hexes))
+            ability = abilities.any? { |a| a.hexes.includes?(MIAMI_HEX) }
```
ollybh, line 39:
```suggestion
            ability = abilities.find { |a| a.hexes.includes?(MIAMI_HEX) }
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -0,0 +1,63 @@
...
+            raise GameError, 'Illegal choice' unless action.choice == 'keywest'
+
+            place_miami_token(action.entity)
+            pass!
+          end
+
+          def place_miami_token(corporation)
+            return unless can_place_miami_token?(corporation)
+
+            hex = MIAMI_HEX
```
ollybh, line 37:
```suggestion
            hex = @game.hex_by_id(MIAMI_HEX)
```

philcampeau: ah yes of course.

PR #10104, opened 2024-01-06
[18Ardennes] Starting major corporations

ollybh:
### Explanation of Change

In 18Ardennes major companies are started by converting minors. There is an auction at the start of each stock round for the rights to start a major, with the players bidding amounts for the rights to start the major, and saying which minor will be converted. The players’ first action(s) in the stock round must be converting the minors into majors.

### Implementation Notes

This is done by first having an auction round to buy (hidden) concession private companies linked to the major corporations. The majors are started in the `BuySellParSharesCompanies` step, in the `process_par` method. A `pledged_minors` hash is added to the game state, to track which minors are going to be used to start the majors.

There are a couple of places where this touches the core rendering code:

1. When placing a bid for the right to start a major, you have to bid an amount and state which minor you will use to start the major. I've implemented this by putting multiple bid buttons next to the bid amount, one for each eligible minor. The array of minor names used to produce these buttons come from a method called `bid_choices`, inspired a similar idea in `G18FL::Step::BuyCert`.
image

2. It's possible (though unlikely) to bankrupt in the stock round. If you've won the right to start a major company then you *must* start it in the following stock round. If you're unable to raise enough cash to do this then you are bankrupt and out of the game. A button to declare bankruptcy is rendered if this has happened, with a description of what went wrong.
image

### Before clicking "Create"

- ~~Add the `pins` label if this change will break existing games~~

### DIFF
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -181,8 +181,15 @@ module View
 
           buttons = []
           if @step.may_bid?(company) && @step.min_bid(company) <= @step.max_place_bid(@current_entity, company)
-            bid_str = @step.respond_to?(:bid_str) ? @step.bid_str(company) : 'Place Bid'
-            buttons << h(:button, { on: { click: -> { create_bid(company, input) } } }, bid_str)
+            if @step.respond_to?(:bid_choices) && (choices = @step.bid_choices(company))
+              choices.map do |choice, label|
+                bid_lambda = -> { create_assigned_bid(company, choice, input) }
+                buttons << h(:button, { on: { click: bid_lambda } }, label)
+              end
+            else
+              bid_str = @step.respond_to?(:bid_str) ? @step.bid_str(company) : 'Place Bid'
+              buttons << h(:button, { on: { click: -> { create_bid(company, input) } } }, bid_str)
+            end
           end
           buttons.concat(render_move_bid_buttons(company, input))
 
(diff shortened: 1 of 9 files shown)


michaeljb:
Could you rebase your branch against the current master? The [commits tab](https://github.com/tobymao/18xx/pull/10104/commits) currently shows many unrelated commits for 1822Africa and 18Tokaido.

ollybh:
> Could you rebase your branch against the current master? The [commits tab](https://github.com/tobymao/18xx/pull/10104/commits) currently shows many unrelated commits for 1822Africa and 18Tokaido.

Gah. I'm not sure what's happened here. The files changed tab only shows the new stuff that should be part of this branch but, like you say, there are lots of extra commits listed. It should just be commits [520987b](https://github.com/tobymao/18xx/pull/10104/commits/520987b9aaa70135c798525d44c3bd0ec11ef323) onwards.

Is there any easy way to do this rebase? I don't understand git rebasing much, I thought `git rebase --onto upstream 520987b^` might be right, but that's trying to rebase all these extra commits as well, giving errors as they have already been merged.

ollybh:
@michaeljb I couldn't work out how to rebase my branch, but I ended up resetting it to match origin/master and cherry-picking the relevant commits. It looks correct now.

assets/app/view/game/round/auction.rb
```diff
@@ -181,8 +181,15 @@ def render_company_actions(company)
 
           buttons = []
           if @step.may_bid?(company) && @step.min_bid(company) <= @step.max_place_bid(@current_entity, company)
-            bid_str = @step.respond_to?(:bid_str) ? @step.bid_str(company) : 'Place Bid'
-            buttons << h(:button, { on: { click: -> { create_bid(company, input) } } }, bid_str)
+            if @step.respond_to?(:bid_choices) && (choices = @step.bid_choices(company))
```
michaeljb, line 184: I see 18FL's auction step has `bid_choices` defined, does it still render properly?

https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_18_fl/step/buy_cert.rb#L16

ollybh: 18FL won't have any situation where this code path would be reached, none of its steps define `may_bid?` or inherit from `Step::Auctioner`.

18FL uses `bid_choices` for a round of blind bids at the start of the game, to determine player order when selecting private companies; `bid_choices` is called from `render_turn_bid`.

That said, it might not have been the best choice to reuse this name. If any future games want have both 18Ardennes's assigned bids and 18FL's blind bids then there will be a conflict. Should I rename this method?

michaeljb: I think it's fine, don't need to future-proof for hypotheticals

PR #10109, opened 2024-01-07
fix 1822 endgame check when bank breaks as stock hits max value

### LINKED ISSUE #10108: [1822] market value endgame should take priority over bank breaking
1822 - Game #146150

In Operation Round 10.1, the LNWR concurrently triggered two end game conditions (i.e. its marker hit the top of the market track while concurrently "breaking the bank"). The way I read rule 1.3, the market hitting the top of the chart should be the determining end game trigger, which would end the game at the end of OR 10.1. However, the game is showing game completion at the end of 10.2. ???

Please correct error or provide rule clarification as appropriate.  Thanks

https://18xx.games/game/146150

-->

benjaminxscott:
Fixes #10108

### Before clicking "Create"

### Implementation Notes

It's possible some games need pinned, if they ran into this edge case previously

* **Explanation of Change**

Check stock value before bank-break, when both are triggered in the same OR

* **Screenshots**

image

* **Any Assumptions / Hacks**
Other 22* games depend on this base code, I looked through PNW/CA/MRS to ensure they didn't override this logic and it didn't seem like they did

Africa doesn't have bank breaking so it wasn't relevant


benjaminxscott:
I can rebase this if https://github.com/tobymao/18xx/pull/10095 gets rejected
(I just had that as the base branch when I checked a new one out)

michaeljb:
I think the existing behavior is correct, see [my comment on the issue](https://github.com/tobymao/18xx/issues/10108#issuecomment-1880199314)

benjaminxscott:
No worries yeah that makes sense as a sequence of actions

-_Ben

PR #10124, opened 2024-01-11
Fix 1822CA's grain bonus computation

### LINKED ISSUE #10123: Slow rendering
https://18xx.games/game/147861?action=1276 - [gist](https://gist.githubusercontent.com/michaeljb/58a5c07956642ff096be2568fc20af3d/raw/73b5b68c21d5d2c394c16bd333b1e3ee25340e24/147861_at_1276.json)

~Computing the graph~ Rendering the route selector here takes long enough for this message to be displayed in my browser:

slowing

This particular game is 1822CA but this behavior has also been observed in other games, including at least 1822 and 18NewEngland.

michaeljb:
Before, rendering https://18xx.games/game/147861?action=1276 takes nearly a minute. After this change, it's down to about 6 seconds.

In game 147861 at action 1276, it is PGE's route step in OR 9.2. The route they ran in OR 9.1 combined a G-train with their 7-train to visit 19 (!) stops. The G-train is not automatically re-attached to the same train they used last time, so that 19-stop route is invalid for the 7-train. The grain bonus calculation was using `stops` instead of `visited_stops`, leading to a bad loop of trying to compute the revenue-generating stops on an invalid route.

Fixes #10123

### DIFF
--- a/lib/engine/game/g_1822_ca/game.rb
+++ b/lib/engine/game/g_1822_ca/game.rb
@@ -468,7 +468,7 @@ module Engine
           sawmill_bonus = sawmill_bonus(route.routes)
           str += " + Sawmill ($#{sawmill_bonus[:revenue]})" if sawmill_bonus && sawmill_bonus[:route] == route
 
-          str += grain_and_port_bonus(route.train, route.stops)[:description]
+          str += grain_and_port_bonus(route.train, route.visited_stops)[:description]
 
           str
         end


lib/engine/route.rb
```diff
@@ -331,7 +333,12 @@ def revenue(suppress_check_other: false, supress_route_token_check: false)
           check_overlap!
           check_connected!
 
-          @game.revenue_for(self, stops)
+          rev = @game.revenue_for(self, stops)
+          @route_error = nil
```
crericha, line 337: Why do you need to set route_error to nil here?

roseundy (commented):
So, you think something in revenue_str is causing the slowdown? Do you know what, specifically?

michaeljb:
I thought this was a problem with the base revenue_str or maybe the 1822 destination bonus computation, but debugging some more it's to do with the sawmill and grain bonuses that are specific to 1822CA. Having the error handler in the base route does catch the problem but I want to fix the underlying problem too, will re-open when that's ready

lib/engine/route.rb
michaeljb: I was thinking that when the route is extended (by clicking in the UI), the `Route` object would be updated, instead of a new one being created, but I don't think that can actually be the case since `@revenue` is cached here. Removed this in 10758570b

michaeljb:
> So, you think something in revenue_str is causing the slowdown? Do you know what, specifically?

Turned out to actually be `stops` vs `visited_stops` in the grain bonus a9dc4ed5a

lib/engine/route.rb
```diff
@@ -342,7 +347,12 @@ def subsidy
     end
 
     def revenue_str
-      @revenue_str ||= @game.revenue_str(self)
+      @revenue_str ||=
+        if @route_error
+          visited_stops.map { |s| s.hex.name }.join('-')
+        else
+          @game.revenue_str(self)
+        end
```
roseundy, line 355: So,  is this needed any more?

lib/engine/route.rb
```diff
@@ -342,7 +347,12 @@ def subsidy
     end
 
     def revenue_str
-      @revenue_str ||= @game.revenue_str(self)
+      @revenue_str ||=
+        if @route_error
+          visited_stops.map { |s| s.hex.name }.join('-')
```
crericha, line 352: Is there a base game method that does just this part? It'd be better if we weren't duplicating logic.

lib/engine/route.rb
michaeljb: Probably not. Since you mentioned the problem where something like this happened in 18NewEngland, I was hoping this would be a catch-all but since we don't have that game to test any more, it's probably best to just stick with the fix that is needed for 1822CA.

lib/engine/route.rb
michaeljb: It's close to what the game `revenue_str` method does. I'm going to delete this and just keep the 1822CA fix.

PR #10127, opened 2024-01-11
[18EU][18MT] Change default Pass optional to conditional in programmed auction bid, was causing infinite loops

### LINKED ISSUE #9615: 18EU - Clicking on button to buy a minor company (#5 for $50) causes a never return timeout.  Game Id 133714
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

andrewzwicky:
Fixes #9615 

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
   Two games allow expanded auto action options during the initial private/minor auction (18EU & 18MT).  In 18EU, in certain cases, a player can be next to "act", have auto actions programmed, actually be unable to pass (not enough money to spend in auction, thus already eliminated OR must select a minor to auction), but the default auto action is Pass.  This triggers an infinite loop of Pass actions which do not change the game state (player cannot Pass), thus freezing the game.

   This change returns an empty array if pass is not a valid action, otherwise it returns Pass.  This accommodates the "Pass after max bid reached / buy price impossible" functionality.

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/step/programmer_auction_bid.rb
+++ b/lib/engine/step/programmer_auction_bid.rb
@@ -53,7 +53,7 @@ module Engine
                                              reason: "Price for #{target.name} exceeded maximum bid")]
         end
 
-        [Action::Pass.new(entity)]
+        []
       end
 
       def auto_buy?(_entity, program)
--- a/lib/engine/step/programmer_auction_bid.rb
+++ b/lib/engine/step/programmer_auction_bid.rb
@@ -53,7 +53,7 @@ module Engine
                                              reason: "Price for #{target.name} exceeded maximum bid")]
         end
 
-        []
+        actions(entity).include?('pass') ? [Action::Pass.new(entity)] : []
       end
 
       def auto_buy?(_entity, program)
--- a/spec/lib/engine/games/g_18eu_spec.rb
+++ b/spec/lib/engine/games/g_18eu_spec.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+require './spec/spec_helper'
+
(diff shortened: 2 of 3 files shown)


michaeljb:
I verified the added 18EU test times out without the `programmer_auction_bid.rb` change, so it's a good reproduction of the loop case--thanks for adding that test!

I think this looks good, but I don't fully understand this:

> This is in draft now to get comments about other ways to fix this, because this approach breaks the "pass after complete" logic. That logic is basically, disable auto-actions unless "pass after complete" is checked.

By "pass after complete", do you mean the checkbox with "Pass after max bid reached / buy price impossible"? Does it still work correctly when the player can legally pass? Should that message be changed?

andrewzwicky:
That message about this being in draft is just outdated.  I'll amend it.

That message is still correct, it will pass if pass is a valid option, otherwise it will return no action.  If pass becomes a valid option again, it should resume passing.  I think the UI is still displaying the correct information.

michaeljb:
> That UI message "pass after complete" is still correct

Sorry, I don't see that UI message, I just see "Pass after max bid reached / buy price impossible" and wanted to clarify if that's what you mean?

Screenshot 2024-01-23 195021

andrewzwicky:
Yes, that's I'm referring to.  The "Pass after max bid reached / buy price impossible" checkbox functions the same as before with the single exception that it won't pass if pass is not a possible action.

The text of the pull request is up-to-date.

michaeljb:
Great, thanks for the explanation!

PR #10134, opened 2024-01-13
[1850] corrects color of one stock market space

philcampeau:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

One space on the stock market was color coded wrong, showing as brown when the space should be orange (see 40 value space in the leftmost column of the screen shot below, though the original market used green instead of orange)

* **Screenshots**

image

* **Any Assumptions / Hacks**


philcampeau:
Since the game is in alpha, I assume we won't pin any games, but I figured I'd mention it just in case.

PR #10169, opened 2024-01-22
suggest Tokaido as a good learning game

benjaminxscott:
tweaks some copy to make it more clear which games are good for learning the ropes

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -100,7 +100,8 @@ module View
           h(:a, { attrs: { href: '/login' } }, 'login'), ' to play multiplayer.'
         ]
       end
-      description << h(:p, 'If you are new to 18xx games then Shikoku 1889, 18Chesapeake or 18MS are good games to begin with.')
+      description << h(:p, 'New? Shikoku 1889 and 18Tokaido have Beginner variants '\
+                           'while 18Chesapeake and 18MS are good learning games')
       render_form('Create New Game', inputs, description)
     end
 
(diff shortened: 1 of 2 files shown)


tobymao:
doesn’t seem necessary. i don’t believe beginner variants aren’t particularly useful

lib/engine/game/g_18_tokaido/meta.rb
```diff
@@ -26,12 +26,12 @@ module Meta
...
             sym: :newbie_rules,
             short_name: 'Local 普通',
-            desc: '[newbie variant] slightly expanded tileset, removes yellow zone from stock market',
+            desc: '[Beginner variant] expanded tileset and removes yellow zone from stock market',
           },
           {
             sym: :limited_express,
             short_name: 'Limited Express 特急',
-            desc: '[brutal variant] one less 6 train, 50% to float, 1882-like stock market',
+            desc: '[Experienced variant] one fewer 6 train, only 50% to float, 1882-like stock market',
```
doubt72, line 34: I'd quibble.  This isn't really an "experienced" variant, it's just brutal.  There's a difference.

assets/app/view/create_game.rb
```diff
@@ -100,7 +100,8 @@ def render_content
           h(:a, { attrs: { href: '/login' } }, 'login'), ' to play multiplayer.'
         ]
       end
-      description << h(:p, 'If you are new to 18xx games then Shikoku 1889, 18Chesapeake or 18MS are good games to begin with.')
+      description << h(:p, 'New? Shikoku 1889 and 18Tokaido have Beginner variants '\
+                           'while 18Chesapeake and 18MS are good learning games')
```
doubt72, line 104: This seems to imply 1889 and 18Tokaido are not otherwise good learning games (I think they are).  If I'm making this change I'd just list out those four and leave it at that, if there's a beginner variant they can select it (or not).

doubt72:
I guess it's not happening anyway, but I had quibbles.

PR #10175, opened 2024-01-22
[1894] Improve logs in setup, fix a city name

Galatolol:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

What teleport company was randomly selected in setup is now printed.

I suppose that doesn't require pinning?


lib/engine/game/g_1894/game.rb
```diff
@@ -308,16 +308,19 @@ def setup
           paris_tiles = @all_tiles.select { |t| PARIS_TILES.include?(t.name) }
           paris_tiles.each { |t| t.add_reservation!(plm, 0) }
 
+          remove_random_teleport_company
```
michaeljb, line 311: With this function call moved to before the `sort_by { rand }` on line 315, this breaks 41 of 42 games in my local copy of the database. We can pin all the games if you want, but this seems like a large unintended side effect for just improving log output.

Galatolol: Thanks, for letting me know, I changed that.

PR #10188, opened 2024-01-26
[1888] Change bonus icon to be circular to hide circle below.

### LINKED ISSUE #6604: [1888] Mountain Tile L41 Renders Improperly

andrewzwicky:
Fixes #6604 

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

    In general, when the code draws a large icon, it always places a white circle with it.  This changes makes the +40 icon circular so it covers up that circle.

    The initial issue mentioned that it was hiding the revenue, but that isn't actually the case, it's just a weird quirk of how the icons are drawn.  If you delete that via the browser inspector, it's blank underneath.
    
    https://github.com/tobymao/18xx/blob/493e286b98aed2c240bb7ec5d8e75f8ad60428bc/assets/app/view/game/part/large_icons.rb#L32

* **Screenshots**

**OLD**
image

**NEW**
image

* **Any Assumptions / Hacks**

    It may be worth seeing if any of the large icons actually rely on that behavior, and only draw that circle if we need to.


michaeljb:
This does make the tile look better, so I'm happy to merge it, but it doesn't actually fix #6604, as the $10 revenue for the town on tile L41 is still not visible

andrewzwicky:
L41's revenue is 0.  If it did have revenue, it would render in a different part.

image


https://github.com/tobymao/18xx/blob/493e286b98aed2c240bb7ec5d8e75f8ad60428bc/lib/engine/game/g_1888/map.rb#L78

michaeljb:
Thanks for clarifying! I was just going off what was in the issue description, didn’t dig deeper myself 🙂

PR #10208, opened 2024-01-28
[18Uruguay] Adding basic support for loans

patrikolesen:
[18Uruguay] Adding basic support for loans
    All corperations will take a loan from RTPLA when floating
    It is then optional for the corperation to take one addtional loan per OR
    The interest is payed end of OR
    It is not possible to repay until Nationalization started
    
    This commit will only add the basic support additional possibilities to take loan and the Nationalization will be handled later
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**


patrikolesen:
Rebased to master instead of to of tree

PR #10212, opened 2024-01-28
[18Uruguay] Adding support for RPTLA dividens

patrikolesen:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**


patrikolesen:
Rebased on top of loans for easier review

PR #10213, opened 2024-01-28
[18Uruguay] Update Discard train. Only trains shall end up in the bankpool

patrikolesen:
Update Discard train. No ships shall end up in bankpool

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

* **Screenshots**

* **Any Assumptions / Hacks**


patrikolesen:
Rebased commit on top of master instead of top of tree

PR #10216, opened 2024-01-28
[rollingstock] market can have more than 5 shares

### LINKED ISSUE #10215: Rolling Stock [Stars] should not have a market share limit
**What's your Game ID?**

**How to Reproduce the Bug?**

Sounds like RS and RSS don't have a market share limit

I don't play either so not sure if RS would also have this rule, the issue was mentioned in the context of RSS
https://18xxgames.slack.com/archives/C68J3MK2A/p1706464717304489

benjaminxscott:
Fixes #10215

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Override the default 50% limit

Should also cover RSS since it imports the RS config

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_rolling_stock/game.rb
+++ b/lib/engine/game/g_rolling_stock/game.rb
@@ -39,6 +39,7 @@ module Engine
         BANK_CASH = 10_000
         STARTING_CASH = { 2 => 30, 3 => 30, 4 => 30, 5 => 30, 6 => 25 }.freeze
 
+        MARKET_SHARE_LIMIT = 100
         MARKET = [
           %w[0c
              5


roseundy:
Hey, as a courtesy, since I'm still around, can you give me a heads up before you fix bugs on Rolling Stock?

roseundy:
So, please talk to me before you make changes. Rolling Stock has very different stock rules than 18xx and isn't affected by MARKET_SHARE_LIMIT.

This change is completely unneeded.

tobymao:
@benjaminxscott did you test this change?

benjaminxscott:
I think wires were crossed, I was waiting for confirmation on whether it
also covered RS, and to find some test games prior to merging (I had kept
the test checkboxes unchecked in an attempt to make that apparent, but that
isn't really enough I realize)


-_Ben

roseundy:
I have a case that doesn't seem to show the bug. This change does fix a minor UI issue that shows an "L" next to the market when it gets to 5 shares.

michaeljb:
Sorry about that.

> (I had kept the test checkboxes unchecked in an attempt to make that apparent, but that isn't really enough I realize)

The unchecked boxes are the same thing covered by the github checks so when that passed I didn't think anything of the actual checkboxes. PRs can be marked as drafts, I've found that can be useful for cases like this.

Still I should have looped in @roseundy, sorry again.

PR #10247, opened 2024-02-02
[18Uruguay] Secondary capitalization

patrikolesen:
Secondary capitalization

The corporation get the rest of the capitalization when the reach there destination tile

### Before clicking "Create"


lib/engine/game/g_18_uruguay/step/track.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+
+          def lay_tile(action, extra_cost: 0, entity: nil, spender: nil)
+            ret = super
+            check_and_apply_destination_bonus
+            ret
+          end
+
+          def pass!
+            check_and_apply_destination_bonus
```
michaeljb, line 50: I think this check should go to a new step after the track step, and it should be handled in the `auto_actions`, like [1822's destination token](https://github.com/tobymao/18xx/blob/042efbf8305cc1c757f5ce984887e02d6e70eb16/lib/engine/game/g_1822/step/destination_token.rb#L20).

The reason for this is that graph checks are computationally expensive, so we try to avoid doing too much of that when possible. If it's in the `auto_actions` for a new step right after the `Track` step, then when a browser user finishes laying track, the graph computation will happen once in the browser, and it won't need to be re-computed every time the game loads--and every time an action is submitted to the server, it needs to recompute the game state to validate it.

[Here's an 1822 example](https://18xx.games/game/150318?action=472), where SECR lays one tile and then passes their track step; their home is connected to their destination, so they automatically get the destination token. The action JSON for this is:

```
    {
      "type": "pass",
      "entity": "SECR",
      "entity_type": "corporation",
      "id": 472,
      "created_at": 1706668669,
      "auto_actions": [
        {
          "type": "hex_token",
          "entity": "SECR",
          "entity_type": "corporation",
          "created_at": 1706668669,
          "hex": "P41",
          "token_type": "destination"
        }
      ]
    }
```

Here's an example of an earlier company's turn where their destination step was passed:

```
    {
      "type": "lay_tile",
      "entity": "CR",
      "entity_type": "corporation",
      "id": 385,
      "created_at": 1706627645,
      "auto_actions": [
        {
          "type": "pass",
          "entity": "CR",
          "entity_type": "corporation",
          "created_at": 1706627644
        }
      ],
      "hex": "F9",
      "tile": "9-5",
      "rotation": 2
    }
```

michaeljb (changes requested):

lib/engine/game/g_18_uruguay/step/track.rb
patrikolesen: Rewritten to use auto action instead

lib/engine/game/g_18_uruguay/step/destination_bonus.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G18Uruguay
+      module Step
+        class DestinationToken < Engine::Step::Base
```
michaeljb, line 7: Renaming to match name of the file and the actual function of the step

```suggestion
        class DestinationBonus< Engine::Step::Base
```

lib/engine/game/g_18_uruguay/game.rb
```diff
@@ -267,6 +267,7 @@ def operating_round(round_num)
             G18Uruguay::Step::TakeLoanBuyCompany,
             Engine::Step::HomeToken,
             Engine::Step::Track,
+            G18Uruguay::Step::DestinationToken,
```
michaeljb, line 270:
```suggestion
            G18Uruguay::Step::DestinationBonus,
```

michaeljb (approved):

lib/engine/game/g_18_uruguay/step/destination_bonus.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G18Uruguay
+      module Step
+        class DestinationBonus< Engine::Step::Base
```
michaeljb, line 7:
```suggestion
        class DestinationBonus < Engine::Step::Base
```

michaeljb (approved):

PR #10257, opened 2024-02-03
[1830][1880] Add new bottom center location to pointy tile labels

### LINKED ISSUE #8178: 1830 : overlapping labels on OO tile 59

andrewzwicky:
Fixes #8178

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

    Add a new location that labels can use.  This will give the OO in the 1830 tile a better place to land.

* **Screenshots**
    Here are the two tiles I could find that were affected by this, and the changes look better in both cases (imo).

**1830**
(old on left, new on right)
image image

**1880**
(old on left, new on right)
(ignore the color change here)
image image

* **Any Assumptions / Hacks**

### DIFF
--- a/assets/app/view/game/part/label.rb
+++ b/assets/app/view/game/part/label.rb
@@ -174,6 +174,12 @@ module View
             x: 50,
             y: 37,
           },
+          # bottom center
+          {
+            region_weights: { [21] => 1.0, [22] => 0.5 },
+            x: 0,
+            y: 60,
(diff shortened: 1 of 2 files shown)


michaeljb (approved):
Awesome, thanks!

PR #10263, opened 2024-02-03
Add deep_freeze method, apply to 1825 map constants

michaeljb:
While running a full db validation I found some failing 1825 games, and upon further investigation, found there were certain cases where I could validate a game and it would pass, I'd validate another game and it would pass, and then validating the first one again would fail.

It turned out that 1825's hex set up was actually mutating some of the constants, so when the first game was validated again, it didn't have the same starting data.

The constants have a `freeze` call applied to them, but for arrays and hashes that only shallowly freezes references. With 3c80bf7 checked out (includes `deep_freeze` changes but excludes the fix in `append_game_hexes`), the tests for 1825 fail when the code attempts to modify the constants.

It would be useful to apply `deep_freeze` to other constants across all the games, just starting small here.

### Before clicking "Create"

### DIFF
--- a/lib/engine.rb
+++ b/lib/engine.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative 'engine/deep_freeze'
 require_relative 'engine/jaro_winkler'
 
 if RUBY_ENGINE == 'opal'
--- a/lib/engine/deep_freeze.rb
+++ b/lib/engine/deep_freeze.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class Object
+  def deep_freeze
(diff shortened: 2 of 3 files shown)


roseundy:
Does this actually fix the bug, or just cause 1825 to fail immediately?

roseundy:
The intent was for 1825 to not mutate the starting constants at all, but modify a copy that is returned to game_hexes and game_tiles based on optional rules.

michaeljb:
The change in `append_game_hexes` fixes the bug - 5294033

roseundy:
OK. Sorry, I missed the edit at the end. I just saw freeze replaced with deep_freeze.

roseundy (approved):

PR #10270, opened 2024-02-04
scripts for dumping/loading and scrubbing the database

michaeljb:


scripts/db_scrub.rb
```diff
@@ -0,0 +1,32 @@
...
+
+raise "You probably don't want to scrub the prod db" unless ENV['RACK_ENV'] == 'development'
+
+require_relative '../db'
+require_relative '../models'
+require_relative '../models/action'
+require_relative '../models/game'
+require_relative '../models/game_user'
+require_relative '../models/user'
+Sequel.extension :pg_json_ops
```
michaeljb, line 11: This will just use `scripts_helper.rb` like the other scripts in #10264, but I needed to change it to this for this branch to work independently from that PR.

tobymao (approved):

PR #10287, opened 2024-02-06
[18Chesapeake: Off the Rails] Update links in meta.rb

### LINKED ISSUE #10286: Update 18Chesapeake: Off the Rails meta.rb with more relevant links
The more info link in the 18Chesapeake: Off the Rails game should link directly to it's respective wiki entry on Github.  Should users need or want to access the wiki for the base game they may do so from the link on the Off the Rails page.

The current rules link is only to the expansion rules which don't have much info.  A link to the base game rules would be helpful

wongrm:
This PR links Off the Rails to it's respective wiki page rather than the base game wiki.  From there users can navigate to the base game wiki entry if need be.

It also adds a link to the base 18Chesapeake rules as the current link to the expansion rules does not have a lot of info.  Follows the same pattern used in the 18Mag meta.rb: https://github.com/tobymao/18xx/blob/bc9b3a270f21aab9dd7f8b508da2117061e60116/lib/engine/game/g_18_mag/meta.rb#L21

Fixes #10286 

### DIFF
--- a/lib/engine/game/g_18_chesapeake_off_the_rails/meta.rb
+++ b/lib/engine/game/g_18_chesapeake_off_the_rails/meta.rb
@@ -13,9 +13,14 @@ module Engine
         DEPENDS_ON = '18Chesapeake'
 
         GAME_DESIGNER = 'Scott Petersen'
-        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Chesapeake'
+        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Chesapeake%3A-Off-the-Rails'
         GAME_PUBLISHER = :all_aboard_games
-        GAME_RULES_URL = 'https://www.dropbox.com/s/ivm4jsopnzabhru/18ChesOTR_Rules.png?dl=0'
+        GAME_RULES_URL = {
+          'Base 18Chesapeake Rules' =>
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_chesapeake_off_the_rails/meta.rb
```diff
@@ -13,9 +13,14 @@ module Meta
         DEPENDS_ON = '18Chesapeake'
 
         GAME_DESIGNER = 'Scott Petersen'
-        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Chesapeake'
+        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Chesapeake%3A-Off-the-Rails'
         GAME_PUBLISHER = :all_aboard_games
-        GAME_RULES_URL = 'https://www.dropbox.com/s/ivm4jsopnzabhru/18ChesOTR_Rules.png?dl=0'
+        GAME_RULES_URL = {
+          'Base 18Chesapeake Rules' =>
+          'https://cdn.shopify.com/s/files/1/0252/9371/7588/files/18Chesapeake.pdf?v=1597256917',
```
michaeljb, line 20: This URL is different from the one currently in `g_18_chesapeake/meta.rb`, and it looks like that one still works to me. If that URL ever should change, it would be best if it didn't need to update it in both places:

```suggestion
          G18Chesapeake::Meta::GAME_RULES_URL,
```

michaeljb:
Thanks!

Issue #10110, opened 2024-01-09
18RoyalGorge checklist


michaeljb:
## Known Bugs

- [x] auto-routing after adding engineer/coal cubes doesn't work, it finds a route for the unmodified train

## UI Enhancements needed

Functionally there, but not so much visually

- [x] VGC gold cube slots / phase limitations
- [x] CF&I and VGC descriptions on card (probably best through abilities)
- [x] during track and dividend steps (when `choices` are available), don't render irrelevant private companies as if they have usable abilities
- [x] "used" cities during routing should show towns after the plus, except for 6-trains

## Config

These should be taken care of with `map.rb` and `entities.rb`, no special code needed

- [x] incremental cap, float on 20%
- [x] towns don't count as stops
- [x] 6x2 trains do count towns as stops, but double total
- [x] phase changes on different trains than usual

## Initial Auction
- [x] 5 privates - 2 yellow, 2 green, 1 brown
- [x] auctioned one at a time, yellow to brown
- [x] priority deal to most cash left (don't reorder players)

## Yellow Privates
- [x] Y1 St Cloud Hotel - +$20 revenue in Silvercliffe; moves to Canon City in brown
- [x] Y2 Ghost Town Tour Co - after last Gold is shipped from a hex, add a +$10 ghost town
- [x] Y3 Coal Creek Mines - owning corp gets a coal cube from the coal mine card (supply of 12) when anyone runs through Coal Creek
- [x] Y4 William Palmer - 10% share of Rio Grande and 10% share of Steel Mill
- [x] Y5 Territorial Prison - 1/2 price to build over terrain
- [x] Y6 Local Jeweler - $5 from bank when Gold is shipped, moves to owner's cash at start of SR

## Green Privates
- [x] G1 Doc Holliday - see treaty of boston; closes when debt is paid
- [x] G2 13 lb Gold Nugget - 1x ship 1 Gold for +$130 revenue instead of +$50
- [x] G3 Hanging Bridge Lease - may run route through royal gorge if 10% dividend paid to rio grande, from the corporation's cash or its president's cash
- [x] G4 Florence Oil Fields - no revenue in yellow
- [x] G5 Metals Investor - discount on steel/gold shares; if used, cannot sell the bought shares till next SR
- [x] G6 Coal Depot - 1 coal cube per $10 paid (round down) during auction

## Brown Privates
- [x] B1 Silver Mines - revenue only in brown
[...]

PR #10294, opened 2024-02-07
[18RoyalGorge] implement paying steel company for track tiles

michaeljb:
#10110

## Before clicking "Create"

### Explanation of Change

* up to 6 track actions possible
* for each color laid, pick one column of that color from the steel table
    * a company can only take cubes from the chosen column 
    * a company can use columns that were used by previous companies 
    * cubes are refilled at the end of the OR set 
    * the cheapest option need not be used; if no explicit choice is made, the cheapest will be used 
    * cube/column choices are made by clicking directly on the steel table 
    * the currently chosen cube is highlighted in white 
    * when a column is committed, its header is higlighted in the column color
* re-uses the "map legend" from 21Moon to render the table, and allows clicking the table to process actions, so that the user can click directly on the cube they want to use

### Screenshots

By default, the first, cheapest column is selected:

steel before

Any cube for an available color can be selected before a tile has been laid; clicking on a cube will show a log message:

steel chose 20 C

Once a tile has been laid, a different column cannot be chosen; the next cube chosen by default is the one above the one used--my assumption there is if a company is not using the cheapest, they want to save the cheapest for a company operating later; the  money spent on the previously selected cube is visible in the log:

steel laid with C

### Any Assumptions / Hacks

An `action_processor` argument is added to pass a lambda to a game's `map_legend` method so that code from `View::Game::Actionable` can be used for processing the action when clicking on the table cell defined in the 18RoyalGorge game class.

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -4,6 +4,7 @@ require '../lib/storage'
 require '../lib/settings'
 require 'view/game/axis'
 require 'view/game/hex'
+require 'view/game/map_legend'
 require 'view/game/tile_confirmation'
 require 'view/game/tile_selector'
 require 'view/game/token_selector'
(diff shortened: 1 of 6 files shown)


tobymao (approved):

michaeljb:
cc @roseundy since I moved around the map legend code a bit, and added the `action_processor` arg

PR #10297, opened 2024-02-07
rename after_ipo > after_float symbol

### LINKED ISSUE #10255: 1817 rename "after_ipo" option for SELL_AFTER
The "after_ipo" option for SELL_AFTER really should be named "after_float". Currently only affects 1817. Is expected to be used more for System18 microgames that use the incremental cap rules. This change will make it easier to understand how this rule works.

MrSkinny:
Fixes #10255

### Before clicking "Create"

### Implementation Notes

FWIW @scottredracecar, the 1817 rules have no mention of the word "float". :)

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -177,13 +177,13 @@ module Engine
       TURN_SELL_LIMIT = nil
 
       # when can a share holder sell shares
-      # first           -- after first stock round
-      # after_float     -- after stock round in which company floated
-      # operate         -- after operation
-      # full_or_turn    -- after corp completes a full OR turn
-      # p_any_operate   -- pres any time, share holders after operation
-      # any_time        -- at any time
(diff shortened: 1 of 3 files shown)


crericha:
@roseundy and I discussed the naming on slack. We agreed upon `:after_sr_floated` since that's what the behavior really is.

lib/engine/game/base.rb
```diff
@@ -178,7 +178,7 @@ class Base
 
       # when can a share holder sell shares
       # first           -- after first stock round
-      # after_ipo       -- after stock round in which company is opened
+      # after_float     -- after stock round in which company floated
```
crericha, line 181: after_sr_floated throughout

crericha (changes requested):
Please change after_float to after_sr_floated throughout

crericha (approved):

PR #10300, opened 2024-02-08
[SteamOverHolland] implements price shelves

philcampeau:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Implements functional price shelves for Steam Over Holland.
President sales and share issuing aren't impacted by these rules
non-presidential sales do not drop price below market ledges.

### DIFF
--- a/lib/engine/game/g_steam_over_holland/game.rb
+++ b/lib/engine/game/g_steam_over_holland/game.rb
@@ -28,6 +28,7 @@ module Engine
         CURRENCY_FORMAT_STR = 'fl. %s'
         MUST_SELL_IN_BLOCKS = true
         SELL_MOVEMENT = :left_share
+        MUST_EMERGENCY_ISSUE_BEFORE_EBUY = true
 
         BANK_CASH = 99_999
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -225,20 +226,26 @@ def price_movement_chart
...
-              num_shares = 0
+              shares_sold = 0
             when :max_one_drop_unless_president
-              num_shares = 1
+              shares_sold = 1
             when :max_two_drops_unless_president
-              num_shares = 2 unless num_shares == 1
+              shares_sold = 2 unless shares_sold == 1
             end
+            shares_sold.times { @stock_market.move_left(corporation) }
```
michaeljb, line 246: Let's rename this variable here, it's confusing to reassign `shares_sold`, and its purpose here is stock market movement, so something referring to the market "moves" or "steps" would be better.

michaeljb, line 246: Let's rename this variable here, it's confusing to reassign shares_sold since it doesn't affect how many shares are sold, and its purpose here is stock market movement, so something referring to the market "moves" or "steps" would be better.

michaeljb (changes requested):

PR #10318, opened 2024-02-11
[18MT] fix train discount ability timing

### LINKED ISSUE #10317: [18MT] Yellowstone Park does not allow buying a train that the company can't afford for full price
**What's your Game ID?**
https://18xx.games/game/151520
**How to Reproduce the Bug?**
Have Yellowstone Park Railroad and in between 50% and 100% of the cost of the next train from the depot available. Try to activate the ability to buy the train for half price. The game then tells you that the company can't afford the train.
image
**Which Browser / Platform?**
Mozilla Firefox 1.22.0

benjaminxscott:
Fixes #10317 (and other games w/ train discount)

ref: https://github.com/tobymao/18xx/pull/10098#issuecomment-1880205008

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

`s/buying_trains/buy_trains`

May need pins, would want to run against the DB to check

* **Screenshots**

image
image

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_18_mt/companies.rb
+++ b/lib/engine/game/g_18_mt/companies.rb
@@ -79,7 +79,7 @@ module Engine
                           trains: %w[3 4 5],
                           count: 1,
                           closed_when_used_up: true,
-                          when: 'buying_train',
+                          when: 'buy_train',
                         }],
             color: nil,
           },


crericha:
I'm not sure this is the root issue. The `special_buy_train` step includes `'buying_train'` as a valid ability timing. If that isn't working, we should understand why.

benjaminxscott:
Ah could be, this patch fixes the 1840 bug so idk how buy_train works if
the step is looking for buying_train

-_Ben

michaeljb (changes requested):
Following up on Chris's comment, let's only change 18MT for now, we can apply this change to the other games when we're sure they have a problem that's fixed by this.

benjaminxscott:
Updated 

@crericha Is the issue in this LoC ?

https://github.com/benjaminxscott/18xx/blob/40159619f5bee3ce2bbfedeb3276a70d08e1b7d2/assets/app/view/game/buy_trains.rb#L321

michaeljb (approved):

benjaminxscott:
@michaeljb @crericha 
Seeing another (fixed) issue for this, I wonder if this issue is more widespread to games w/ this type of ability and whether our tests are catching it 

https://github.com/tobymao/18xx/pull/10391

Perhaps we individually check those games? I can do that, but wanted to also address the idea that it could be a more core bug in how the timing is working

michaeljb:
Yeah the other games should probably be individually checked, so we can get to the bottom of this:

> The special_buy_train step includes 'buying_train' as a valid ability timing. If that isn't working, we should understand why.

PR #10319, opened 2024-02-12
[18 India] - Implement some of Stock Round (new game)

EdPovi:
This PR implements part of the Stock Round differences for Stock Round
NOTE: implementation is not completed. 

### Implementation Notes
- Many shares are represented as "company proxies" that are bought from IPO Rows or Player Hands
- Only "Market Shares" are bought / sold as shares under the respective corp
- First SR turn is a "Selling Round", selling doesn't restrict purchases later in SR

* **Any Assumptions / Hacks**
- Temporary use of "Unsold Companies" under player class to use existing view, this will be removed eventually as some information needs to be hidden from other players.
- Will need to modify SharePool class as currently the president's share is given to a player when the company floats.  Need allow company to float and operate without Pres Share in play.


lib/engine/game/g_18_india/corporation.rb
```diff
@@ -6,6 +6,20 @@ module G18India
       class Corporation < Engine::Corporation
         def initialize(sym:, name:, **opts)
           super
+          return unless sym == 'GIPR' # Modify for GIPR differences
+
+          # Create replacement first share such that president: == false (allow sale to market / prevent receivorship)
```
michaeljb, line 11:
```suggestion
          # Create replacement first share such that president: == false (allow sale to market / prevent receivership)
```

lib/engine/game/g_18_india/game.rb
```diff
@@ -30,21 +30,40 @@ class Game < Game::Base
...
+
+        # Modified for corporation share limit
+        def cert_limit(entity = nil)
+          return @cert_limit if entity.nil?
+          return 3 if entity.corporation?
+
+          @cert_limit
+        end
+
+        # Modify to only count :private type companies (exclude Warrents and Bonds)
```
michaeljb, line 60:
```suggestion
        # Modify to only count :private type companies (exclude Warrants and Bonds)
```

lib/engine/game/g_18_india/step/sell_once_then_buy_certs.rb
```diff
@@ -0,0 +1,317 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/buy_sell_par_shares'
+require_relative '../../../step/share_buying'
+
+# For 18 India, first action is only sell any, then only buy
```
michaeljb, line 6:
```suggestion
# For 18 India, first turn is only sell any, subsequent turns are only buy
```

michaeljb (approved):

PR #10323, opened 2024-02-13
[21Moon] sets default buying slot to Space Port

### LINKED ISSUE #10302: [21Moon] Make default "destination for purchased train" be Space Port
**What's your idea?**
Generally, the 1st train you buy, every 'forced' buy, and probably the _last_ train you buy is for the SP, so that happens most often.
So that would save having to boink the selection menu so much.

also, if like me, you just boink to **Buy** a train, it's annoying to find later that your new super train is in the LB,
 and your decrepit train is in the SP, and now it's too late to reverse them...

(if the selectors were on the same line of text, then it would be more salient at the time of 'buy',
 but I expect that is a larger effort to redesign...)

Screenshot 2024-02-08 at 5 04 29 PM

**Are there related ideas?**

**Did you join the Slack?**

philcampeau:
Fixes #10302 

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

Swaps the position of Space Port and Local Base in the drop down menu when buying trains. 

I don't believe this will require pins

### DIFF
--- a/lib/engine/game/g_21_moon/step/buy_train.rb
+++ b/lib/engine/game/g_21_moon/step/buy_train.rb
@@ -133,8 +133,8 @@ module Engine
 
           def slot_dropdown_options(corp)
             options = []
-            options << { slot: 0, text: 'Local Base' } if room_for_lb?(corp)
             options << { slot: 1, text: 'Space Port Base' } if room_for_sp?(corp)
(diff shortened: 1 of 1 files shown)


roseundy (approved):
Assuming you tested this, it looks fine.

PR #10330, opened 2024-02-14
Add in region weight debugging information when 'grid' URL parameter is used

andrewzwicky:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**

    I found this helpful when debugging, so I thought it might be good to add.  It uses the same `grid` parameter as the region lines for the tiles.  Text needs to be a little small so it can all fit.

* **Screenshots**

image

* **Any Assumptions / Hacks**


michaeljb (approved):
Looks good, thanks for adding this!

PR #10343, opened 2024-02-17
[18Uruguay] Added some more logic to nationalization

patrikolesen:
[18Uruguay] Added some more logic to nationalization

### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
- Decrease stock value on nationalization added
- Remove surplus goods from map when nationalized
- The interest is 0 after nationalization
- Add final payout from RPTLA

* **Screenshots**

* **Any Assumptions / Hacks**

### DIFF
--- a/lib/engine/game/g_18_uruguay/nationalization.rb
+++ b/lib/engine/game/g_18_uruguay/nationalization.rb
@@ -158,7 +158,22 @@ module Engine
         end
 
         def nationalization_final_export!
-          @log << '  Nationalization: Final Export to be implemented'
+          return if number_of_goods_at_harbor.zero?
+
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_uruguay/nationalization.rb
```diff
@@ -206,6 +221,17 @@ def event_nationalization!
           buy_train(@fce, train, :free)
         end
 
+        # Move stock price one step left for each stock more than limit
```
michaeljb, line 224: Is this comment correct?

```suggestion
        # Move stock price one step left for each share more than limit
```

patrikolesen: No It should be loans

PR #10347, opened 2024-02-17
Suggestion for adding support for multirow revenue

patrikolesen:
### Before clicking "Create"

### Implementation Notes

* **Explanation of Change**
In 18 Uruguay the offboard revenue box gets to big and hiding the rails.

Inspired by https://www.18xx-maker.com/elements/ "Offboard Revenues"
I suggest here to add an extra parameter "rows" to the offboard tile which will allow the revenue to be splitted over multiple rows.

In addition to split the row we also need to move the tile name so it does not collide with the revenue when multiple rows is configured

Example from the 18 Uruguay map
`            ['G1'] => 'offboard=revenue:yellow_30|green_30|brown_40|gray_60,format:+%d,visit_cost:0,route:optional'\
                      ',groups:England,rows:2;path=a:3,b:_0;path=a:4,b:_0',
`

* **Screenshots**
Before
Single line revenue

After
Multirow revenue

* **Any Assumptions / Hacks**

### DIFF
--- a/assets/app/view/game/part/multi_revenue.rb
+++ b/assets/app/view/game/part/multi_revenue.rb
@@ -27,13 +27,11 @@ module View
           end
 
           # Compute total width of rectangles so we can center
-          total_width = []
           step_size = (computed_revenues.size / @rows.to_f).ceil
-          @rows.times do |row|
-            total_width_row = computed_revenues[(row - 1) * step_size, step_size].sum do |revenue|
+          total_width = @rows.times.map do |row|
+            computed_revenues[(row - 1) * step_size, step_size].sum do |revenue|
               revenue[:width]
             end
-            total_width.append(total_width_row)
           end
 
           row = 0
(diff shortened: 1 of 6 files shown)


assets/app/view/game/part/multi_revenue.rb
```diff
@@ -25,34 +27,48 @@ def render
...
           # Compute total width of rectangles so we can center
-          total_width = computed_revenues.sum do |revenue|
-            revenue[:width]
+          total_width = []
+          step_size = (computed_revenues.size / @rows.to_f).ceil
+          @rows.times do |row|
+            total_width_row = computed_revenues[(row - 1) * step_size, step_size].sum do |revenue|
+              revenue[:width]
+            end
+            total_width.append(total_width_row)
```
michaeljb, line 36: Does this work?

```suggestion
          step_size = (computed_revenues.size / @rows.to_f).ceil
          total_width = @rows.times.map do |row|
            computed_revenues[(row - 1) * step_size, step_size].sum do |revenue|
              revenue[:width]
            end
```

michaeljb (commented):
Love it, thanks for adding this! Just one minor suggestion to use `map` when possible.

michaeljb (approved):

PR #10350, opened 2024-02-17
[core] add global LOGGER

michaeljb:
By default, log level is `INFO` in production and `DEBUG` in development.

In the browser, the URL param `l` can be used to set the level. On the server, the environment variable `ENGINE_LOG_LEVEL` can be used. The level is set to an integer; 0/1/2/3/4 for `DEBUG`/`INFO`/`WARN`/`ERROR`/`FATAL`.

When the URL param or environment variable are not set, the log level will default to `INFO` in production and `DEBUG` in development.

Added logging:

* prealpha games - replace `puts` with `LOGGER.debug`

* auto_router - replace `puts` with `LOGGER.info`

* actionable view - log the rescued error with `LOGGER.error`

* add `LOGGER.debug` for some things I've made temporary local edits for countless times:

    * show each action ID as it is processed

    * start and end of:

        * processing game actions

        * rendering game

        * computing graph

Other changes:

* rename `LOGGER` in api.rb and queue.rb just to play it safe with possible name collisions

* add `rubocop-rails` for the `Output` cop, which flags uses of `puts`; only apply to `assets/` and `lib/engine/`

### Before clicking "Create"

### DIFF
--- a/api.rb
+++ b/api.rb
@@ -35,9 +35,9 @@ class Api < Roda
     csp.frame_ancestors :none
   end
 
-  LOGGER = Logger.new($stdout)
+  API_LOGGER = Logger.new($stdout)
 
-  plugin :common_logger, LOGGER
+  plugin :common_logger, API_LOGGER
 
   plugin :not_found do
     halt(404, 'Page not found')
@@ -46,8 +46,8 @@ class Api < Roda
   plugin :error_handler
 
   error do |e|
-    LOGGER.error e.backtrace
-    LOGGER.error e.message
+    API_LOGGER.error e.backtrace
+    API_LOGGER.error e.message
     { error: e.message }
   end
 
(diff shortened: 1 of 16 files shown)


michaeljb:
hmm this breaks on the main page, up till now I'd only tested with directly loading a game, will reopen after that's fixed.

michaeljb:
Converting to draft.

Unit tests and scripts should change the log level to FATAL (or a higher integer)

tobymao:
will this only happen in dev and not in prod?

michaeljb:
This won't add logging on the server in prod.

In the prod browser, auto-routing output will appear in the console with `INFO` output. I could change that to `DEBUG`, it's currently sent to the console with `puts`.

Adding the URL param `?l=0` turns on `DEBUG` output in the browser console.

michaeljb:
Made a few updates

* production's log level is now `FATAL` to ensure no server prod log messages appear
* I got rid of the `rubocop-rails` plugin, it's overkill for just checking for `puts`
* changed `auto_router.rb`'s output from `INFO` to `DEBUG`; my initial rationale for `INFO` was that it was already appearing in the console, but if a user knew about that and is really interested in seeing it, they're probably also the kind of user that can handle adding `?l=0` to their URL

tobymao (approved):

PR #10354, opened 2024-02-18
[1893] Fix exclusion of merge voter for share-only owner

### LINKED ISSUE #10311: [1893] shareholder unable to vote during merger (blocks step w no progression option)
https://18xx.games/game/149773

https://gist.github.com/benjaminxscott/3de5e4298cf7e6bc05b1686e57d35555

Attempting to float AGV via merge, player `wormjitsu` has 10 votes but is unable to vote on the merger (no UI elements)

perwestling:
Fixes #10311

I believe no games will break due to this fix. The fix now makes it possible to avoid getting stuck. Presumable those that got stuck, did instead do Undo, and then vote No, to get passed this lock. And as they now have worked around this situation nothing will happen.

### Implementation Notes

* **Explanation of Change**

In case a player owned shares of the corporation to be created by merge, the merge step did not regard this as a voter, so the GUI got stuck. Players had to decline merge to get passed this.

Now merge step considers share owners as well.

Also refactored and added a couple of simple methods to make code more readable.
And updated information to point out that shares in bank pool is regarded as No votes.

* **Screenshots**

Now vote buttons appear for the player that only owned share in AGV. The other players owned minors that were reserving shares in AGV so for them merge vote functionality worked.
image

* **Any Assumptions / Hacks**

None.

### DIFF
--- a/lib/engine/game/g_1893/step/merger.rb
+++ b/lib/engine/game/g_1893/step/merger.rb
@@ -18,9 +18,17 @@ module Engine
           def actions(entity)
             return [] if entity.company?
             return [] unless choice_available?
-            return [] if @game.round.merger_candidates_for(@game.round.current_entity).empty?
+            return [] if mergable_owned.empty? && !merge_target_shares_owned?
 
             ACTIONS
+          end
+
+          def mergable_owned
+            @game.round.merger_candidates_for(@game.round.current_entity)
+          end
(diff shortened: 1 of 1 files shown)


perwestling:
Gemfile.lock did manage to be included in first push, so I have now removed that from the commit.

Issue #10122, opened 2024-01-10
1832 implementation checklist


philcampeau:
Setup
- [x] Currency format - $xxx
- [x] Bank - $12,000
- [x] Starting cash -  2 => 1050, 3 => 700, 4 => 525, 5 => 420, 6 => 350, 7 => 300 
- [x] Certificate limit 
- [x] Meta
- [x] Map
  - [x] Show future tile upgrades for Charleston and Jacksonville (Y)
  - [x] Indicate Boomtown hexes
- [x] Tiles
  - [x] Label tile 611 as Y to simplify its reservation for Charleston and Jacksonville.
- [x] Corporation logos and colors

Private Companies
- [x] Sold for half to 2x value in phases 3-4
- [x] West Virginia Coal fields can be bought for half to face in Phase 2.
- [x] P1 - Carolina Stagecoach Company
  - [x] Value - $20, Revenue $5
- [ ] P2 - Cotton Warehouse
  - [x] Value - $40, Revenue $10
  - [x] Adds $10 to the value of one non-costal city for the owning corp
  - [ ] Must be able to assign it to a specific city in the case of Atlanta, which has 3 separate cities on it.
- [ ] P3 - Atlantic Shipping
  - [x] Value - $50,  Revenue 10
  - [x] Can be placed on any revenue hex with a Port symbol
  - [x] Increases value of stop by $20 for the owning corp
  - [x] increases value of stop by $10 for all other corps
- [ ]  P4 - London Investment
  - [x] Value - $70, Revenue $10
  - [x] cannot be bought in
  - [ ] Can be traded for a 10% IPO share of any corp
  - [x] Can only be exchanged during an SR
  - [ ] Corp must have been parred in the current SR (Including Central of Georgia Railway in SR1)
  - [ ] This counts as your purchase action
  - [ ] Continues to pay revenue until the corp pays its first dividend.
- [ ] P5 - West Virginia Coalfields
  - [x] Value - Value - $80, Revenue - $15
  - [x] Blocks hexes around WVCF hex until bought in by corp or closed. 
  - [x] owning corp receives a free coal token
  - [x] once bought in, other corps can buy a coal token from this corp, $40 to corp, $40 to bank
  - [ ] this token allows a corp to run to and through the WVCF hex
  - [ ] Corp must have a route to WVCF to buy token
  - [ ] buying WVCF token counts as a yellow tile lay
- [ ] P6 - Southern Bank
  - [ ] Variant rule only. May not initially implement. 
[...]

PR #10384, opened 2024-02-21
1832 initial dev work by jemiahlee

jemiahlee:
Works on #10122 

### Before clicking "Create"

- [x] Add the `pins` label if this change will break existing games (N/A)

### Implementation Notes

* **Explanation of Change**

@philcampeau did a ton of work here already to get the basic game setup done. This PR mostly just rearranges the files a little bit so that trains, phases, market are in their own files. There was also an issue in laying track that this fixes (I did not examine whether the behavior is truly correct, just that it was breaking tile laying and now it isn't).

***Screenshots**
N/A
* **Any Assumptions / Hacks**
N/A

### DIFF
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -62,61 +62,72 @@ module Engine
         STOCKMARKET_COLORS = Base::STOCKMARKET_COLORS.merge(unlimited: :green, par: :white,
                                                             ignore_one_sale: :red).freeze
 
-        PHASES = [{ name: '2', train_limit: 4, tiles: [:yellow], operating_rounds: 1 },
-                  {
-                    name: '3',
-                    on: '3',
-                    train_limit: 4,
-                    tiles: %i[yellow green],
-                    operating_rounds: 2,
-                    status: ['can_buy_companies'],
-                  },
-                  {
-                    name: '4',
-                    on: '4',
-                    train_limit: 3,
-                    tiles: %i[yellow green],
-                    operating_rounds: 2,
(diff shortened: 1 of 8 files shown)


perwestling (commented):
Do the diesel variant handle this event?

One way of doing it could be to add diesel to phases and trains, and maybe to delete what is not used during setup?  Have not checked code how/if this is handled.

jemiahlee:
On the first question, there is a longer road to go here with the game, I will consider what you stated; I'm also not against an in-place modification before the freeze occurs on the data during the initial game setup. The diesel variant is not yet fully wired up, but those two variants need the least code, so I'm certain I will get to them as part of the initial game creation.

On the squash, I figured I would let the squash/merge policy of the repo dictate on whether this ended up as one commit or two. They are functionally different changes, since the first commit does let you play the game (I was trying to see what was working and what wasn't and got stuck).

philcampeau:
The changes all look good to me, as the person who was working on this before!

michaeljb:
> On the squash, I figured I would let the squash/merge policy of the repo dictate on whether this ended up as one commit or two. They are functionally different changes, since the first commit does let you play the game (I was trying to see what was working and what wasn't and got stuck). The second commit is me just trying to slim down the main game file to segregate data from functionality.

There isn't a real policy in place for this, two commits is fine here. When merging a PR, I default to the "create a merge commit" option but If the PR branch is messy but the changes are good I'll just use the squash option. If the changes in the two commits are different enough then having them in separate PRs is encouraged, but with something like this where it's mostly config changes and not too large it's no big deal.

michaeljb (approved):

PR #10397, opened 2024-02-25
[1894] Fix 5+1 revenue bug

Galatolol:
## Before clicking "Create"

### Explanation of Change

If a 5+1 train was hitting exactly 5 any locations and 1 town, it would subtract the value of the town from the total revenue.

I looked around at other N+1 train definitions and it seems that town nodes should be declared before the general nodes. This indeed fixed the issue.

If I remember correctly, changes like this don't require pinning?


michaeljb:
> If I remember correctly, changes like this don't require pinning?

Right, when the game is being processed the value of routes aren't recomputed, the value that's stored in the action is trusted and used. This means that games with wrong values won't be retroactively fixed, but they also don't need to be pinned.

PR #10400, opened 2024-02-25
Update pry-byebug version for new ruby version

andrewzwicky:
## Before clicking "Create"

### Explanation of Change

After the latest round of gem updates, if you tried to `` require `pry-byebug` `` in a file, you would get this error:
```
Failure/Error: require 'pry-byebug'

NameError:
  undefined method `=~' for class `Pry::Code'
# /usr/local/bundle/gems/pry-0.13.1/lib/pry/code.rb:342:in `<class:Code>'
# /usr/local/bundle/gems/pry-0.13.1/lib/pry/code.rb:32:in `<class:Pry>'
# /usr/local/bundle/gems/pry-0.13.1/lib/pry/code.rb:5:in `<top (required)>'
# /usr/local/bundle/gems/pry-0.13.1/lib/pry.rb:70:in `<top (required)>'
# /usr/local/bundle/gems/pry-byebug-3.9.0/lib/pry-byebug.rb:3:in `<top (required)>'
# ./spec/lib/engine/games/g_1836jr56_spec.rb:6:in `<top (required)>'
```
I ran `bundle update pry-byebug` and this is the resulting gemfile.lock.  I found this other issues which recommended updating as the solution and it did solve the issue.  https://github.com/gushmazuko/metasploit_in_termux/issues/223

### Any Assumptions / Hacks

I don't know if there's any special flags that are used when updating gems.


michaeljb:
> I don't know if there's any special flags that are used when updating gems.

Nope, nothing special needed here. The next time anyone runs the usual commands to start their server they'll get the new pry-byebug.

Thanks!

PR #10410, opened 2024-02-28
[1854] Adding all of the hex stock market changes that affect shared code

andrewzwicky:
## Before clicking "Create"

## Implementation Notes

I tried to implement things in the least invasive way for the existing stock markets.

### Explanation of Change

This is all of the shared code changes for the hex stock market for 1854 (so far).  Because they're hexes, I couldn't use the same rendering solution, so ended up using something very similar to how the maps are constructed.

The markets are initialized the same way as the normal stock markets, but you have to be careful about how the first row is padded, because it determines how the hexes are laid out.  Under the hood, the share prices are still stored in a 2D array, but are offset, so when you want to move up and to the right, you move up one and right one.  This means not every possible index in the 2D array is actually populated.  Once 1854 is more stable, I plan to go add more comments around this to explain.

I left out the game file for 1854 because it's huge and I want to keep the PRs small.  Here's what the market initialization looks like:

```ruby
MARKET = [
  ['',    '', '', '86',  '89', '93p', '97S', '101S', '106S', '112S', '119S', '127S', '136S', '146S', '158S',
    '170S', '185S', '210S', '230S', '250S', '275S', '300e'],
  ['',    '', '', '81',  '84', '87p', '91', '95S', '99S', '104S', '110S', '117S', '125S', '134S', '143S',
    '153S', '165S', '180S', '200S'],
  ['',    '',    '76',  '79',  '82p',  '85', '89',  '93S', '97S', '102S', '108S', '115S', '123S'],
  ['',    '',    '71',  '74',  '77p',  '80', '83',  '87', '91S', '95S', '100S'],
  ['',    '66',  '69',  '72p', '75', '78', '81', '85', '89S'],
  ['',    '61y', '64',  '67p', '70', '73', '76y', '79y'],
  %w[56y 59y 62 65 68y 71y],
  ['', '54y', '57y', '60y', '63y'],
```

### Screenshots

image


michaeljb (approved):
LGTM, just need to resolve merge conflicts

PR #10416, opened 2024-02-29
[18 India] May Sell Trains during Train Buying Step

EdPovi:

## Implementation Notes
- Add new `sell_train` action.
- Derived new `SellBuyTrain` STEP class from `BuyTain`.
- Minimized change to view>operating.
- Added new render section to View > BuyTrains for selling trains.
- Used `train.salvage` as reference to the resale value of trains.

### Explanation of Change
Needed to customize the view for buying trains as there was no current support for the action.

### Screenshots
image

### Any Assumptions / Hacks
Need to verify timing of stock price movement of trainless corporations.


lib/engine/game/g_18_india/game.rb
```diff
@@ -594,6 +594,19 @@ def legal_tile_rotation?(_entity, _hex, _tile)
           true
         end
 
+        # Sell Train to the Depot
+        def sell_train(operator, train, price)
+          @bank.spend(price, operator) if price.positive?
+          @depot.reclaim_train(train)
+        end
+
+        def corp_is_trainless(corporation)
```
michaeljb, line 603: `corp_is_trainless` sounds like it would just return a boolean, could this be renamed something like `handle_trainless_corp`? Something more descriptive along the lines of `drop_price_for_trainless_corp` is longer but fine too, up to you.

lib/engine/game/g_18_india/step/sell_buy_train.rb
```diff
@@ -0,0 +1,88 @@
...
+            super
+          end
+
+          def pass!
+            # if company doesn't have a train, stock price moves left
+            @game.corp_is_trainless(current_entity) if current_entity.trains.empty?
+            super
+          end
+
+          # modified from step::train to prevent automatic passing after buying
```
michaeljb, line 80:
```suggestion
          # modified from Step::Train to prevent automatic passing after buying
```

michaeljb (changes requested):

lib/engine/game/g_18_india/game.rb
EdPovi: Thank you for the suggestion.  I also need to move the timing of the drop in price.  I have confirmed with Ken that this happens at the end of the company's operating turn (after sell/buy certs, not after sell/buy trains).  I think I need to hook into `def after_end_of_turn(operator); end` from Engine::Round::Operating

EdPovi: I made the suggested change, and moved the timing of the price drop for trainless corp.

EdPovi:
I made the modifications to the method names.  To move the timing, I expanded a hook in Round::Operating so that it can connect to a method in Engine::Game::Base.

michaeljb (approved):

PR #10417, opened 2024-02-29
[1850] Fixes Western Land Grant "Paths must be connected" error

### LINKED ISSUE #10033: [1850] - game #143563 - The game is rejecting a Western Land Grant tile lay on F9 with the message "Paths must be connected".  SOO is connected and it is allowing me to select the tile, then rejecting it incorrectly.
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

### LINKED ISSUE #10334: [1850] - game #151575 - The game is rejecting a Western Land Grant tile lay on H11 (Des Moines) with the message "Paths must be connected". RI is connected and the game is allowing RI to select the tile, then rejecting it incorrectly.  The game will be paused until the bug is fixed.
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

philcampeau:
Tiles don't need to be connected.

Fixes #10033 
Fixes #10334

## Before clicking "Create"

### Explanation of Change

Western land grant doesn't require its tiles to be connected, and this defaults to true on tile_lay abilities with more than one tile.

### DIFF
--- a/lib/engine/ability/README.md
+++ b/lib/engine/ability/README.md
@@ -260,7 +260,7 @@ remains open but the discount can no longer be used. Default false.
 - `special`: If true, do not check that the tile upgrade preserves
   labels and city count. Default true.
 - `connect`: If true, and `count` is greater than 1, tiles laid must
-  connect to each other. Default true.
+  connect to each other. Default true unless `count_per_or` is used, in which case `connect` isn't checked by default. 
 - `blocks`: If true and `when` is `sold`, then the step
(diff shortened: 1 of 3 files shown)


philcampeau:
Pardon the multiple commits. I initially fixed this in the 1850 code, but then decided that there's really no reason why `connect` should be checked if the ability is limited to one use per OR, since no game would expect users to track that from OR to OR (I think).

philcampeau:
will clean this up and resubmit

PR #10421, opened 2024-03-01
[1854] Adding second half of step overrides

andrewzwicky:
## Before clicking "Create"


lib/engine/game/g_1854/step/track_and_token.rb
```diff
@@ -0,0 +1,116 @@
...
+
+module Engine
+  module Game
+    module G1854
+      module Step
+        class TrackAndToken < Engine::Step::TrackAndToken
+          def actions(entity)
+            actions = []
+            return actions unless entity == pending_entity
+
```
ollybh, line 13:
```suggestion
            return [] unless entity == pending_entity

            actions = []
```

ollybh, line 13:
```suggestion
            return [] unless entity == pending_entity

            actions = []
```

lib/engine/game/g_1854/step/track_and_token.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/track_and_token'
+
+module Engine
+  module Game
+    module G1854
+      module Step
+        class TrackAndToken < Engine::Step::TrackAndToken
```
ollybh, line 9: The `TrackAndToken` is for when a corporation can lay track and place station tokens in any order. In 1854 you have to place track before you place any station tokens, so I think that you will need separate track and token steps. Or am I not be understanding what you're doing here? It's a bit difficult to fully review this without seeing the round definitions in the G1854::Game class.

lib/engine/game/g_1854/step/track_and_token.rb
```diff
@@ -0,0 +1,116 @@
...
+          def actions(entity)
+            actions = []
+            return actions unless entity == pending_entity
+
+            actions << 'assign' unless @selected_hex
+            actions << 'lay_tile' if can_lay_tile?(entity)
+            actions << 'place_token' if can_place_token?(entity)
+            actions
```
ollybh, line 17: May this step be passed?

lib/engine/game/g_1854/step/route.rb
```diff
@@ -0,0 +1,23 @@
...
+      module Step
+        class Route < Engine::Step::Route
+          def process_run_routes(action)
+            super
+            return if @round.routes.empty?
+
+            # Close when a train is actually run, as opposed to just reaching the dividends step.
+            # This event is not in the generic abilities, so create one that is the same as in the entities
+            # and rely on the rest of the code to match them up.
+            @game.close_companies_on_event!(action.entity, 'ran_train')
```
ollybh, line 17: I don't see how this differs from the default behaviour of triggering this event in the dividends step. That has the same check for `@round.routes.empty?`.

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G1854
+      module Step
+        class MergeMinors < Engine::Step::Base
+          def actions(entity)
+            return [] if entity.corporation? && !entity.minor?
```
ollybh, line 11: You don't need to also check for `!entity.minor?` that will always be false if `entity.corporation?` is true.
```suggestion
            return [] if entity.corporation?
```

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+            return [] if !entity.minor? || entity != current_entity #  || @game.done_this_round[entity]
```
ollybh, line 12:
```suggestion
            return [] unless entity.minor?
            return [] unless entity == current_entity
```

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+            return [] unless @game.mergeable?(entity)
+            return [] if target_corporations.empty?
+            return [] if @merge_target
+
+            actions = []
+            actions << 'merge' if @game.minor_mergers_allowed?
+            actions << 'pass' if @game.minor_mergers_allowed? && !@game.minor_mergers_required?
```
ollybh, line 19:
```suggestion
            return [] unless @game.minor_mergers_allowed?

            actions = ['merge']
            actions << 'pass' unless @game.minor_mergers_required?
```

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+
+          def merge_name(entity = nil)
+            return "Form #{entity&.name}" if @merging
+
+            "Merge with #{entity&.name}"
+          end
+
+          # corps available as a merge target
+          def target_corporations
+            @game.corporations.select { |c| @game.merge_target?(c) }
```
ollybh, line 55: What does the `@game.merge_target?` method do?

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+            @merge_target = nil
+          end
+
+          def auto_actions(entity)
+            return super if @merging
+
+            return [Engine::Action::Pass.new(entity)] if mergeable_candidates(entity).empty?
+
+            super
+          end
```
ollybh, line 45: I'm fairly certain that you don't need auto_actions here. Is this copied from another game? You'd need this where you have to check the merge candidates are connected by track, but want to avoid recalculating the graph when a game is being loaded. In 1854 there is no requirement for a track connection between the local railways, so this is not an issue.

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+            "Minors that can merge with #{corporation.name}"
+          end
+
+          def mergeable_candidates(corporation)
+            # # Mergeable candidates must be connected by track and not through a regional border
+            # # They must be the same type (major/minor)
+            # parts = @game.token_graph_for_entity(corporation).connected_nodes(corporation).keys
+            # mergeable = parts.select { |n| n.city? && !n.pass? }.flat_map { |c| c.tokens.compact.map(&:corporation) }
+            @mergeable_[corporation] ||=
+              @game.minors.uniq.select { |c| c != corporation && @game.mergeable?(c) }
```
ollybh, line 94: What does `@game.mergeable?` do?

ollybh (changes requested):

lib/engine/game/g_1854/step/merge_minors.rb
andrewzwicky: https://github.com/andrewzwicky/18xx/blob/80f2537d5a4c495dc9f4612e7f5b6cb913236bbb/lib/engine/game/g_1854/game.rb#L279-L281

lib/engine/game/g_1854/step/merge_minors.rb
andrewzwicky: https://github.com/andrewzwicky/18xx/blob/80f2537d5a4c495dc9f4612e7f5b6cb913236bbb/lib/engine/game/g_1854/game.rb#L283-L285

lib/engine/game/g_1854/step/route.rb
andrewzwicky: The first time the corp operates, you get to the dividend step (but can't pay out), have no trains, and the company was closing early.  This closes only when the corp has actually run a train.

lib/engine/game/g_1854/step/track_and_token.rb
andrewzwicky: No, because you must pick where the ABC home token must be.  In the case where there is no tile there, you can lay a tile, but they have to pick a home token.

lib/engine/game/g_1854/step/track_and_token.rb
andrewzwicky: This is for the part of the game where you merge two number minors into one of the ABC corporations.  It's only used to keep track of the need to pick a hex -> possibly lay tile -> then pick a city for home token.

ollybh: OK.
I think this might be more complicated than it needs to be. This could be split into two steps:

1. A subclass of `Step::HomeToken` where the president of the Lokahlbahn chooses the city for the home token (you will need to do something with `home_token_can_be_cheater` for the case where there is no available slot on the yellow tile in Amstetten). This would remove the need for the assign action, and all the code to do with pending tokens.
2. A subclass of `Step::Track` to allow the placement of the green tile in the home hex.

lib/engine/game/g_1854/step/route.rb
ollybh: The line in `Step::Dividend.process_dividend` is:
`@game.close_companies_on_event!(entity, 'ran_train') unless @round.routes.empty?`
So the company shouldn't be closing unless there is a route. How do you have a route if the corporation doesn't yet have any trains?

lib/engine/game/g_1854/step/merge_minors.rb
ollybh, line 11: You don't need to check `corporation?` at all, as you're just about to check `minor?`.
```suggestion
```

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,114 @@
...
+  module Game
+    module G1854
+      module Step
+        class MergeMinors < Engine::Step::Base
+          def actions(entity)
+            return [] if entity.corporation? && !entity.minor?
+            return [] if !entity.minor? || entity != current_entity #  || @game.done_this_round[entity]
+            return [] unless @game.mergeable?(entity)
+            return [] if target_corporations.empty?
+            return [] if @merge_target
```
ollybh, line 15: [Sorry, far too many suggestions close together, let's do this all at once].
```suggestion
            return [] unless entity == current_entity
            return [] unless entity.minor?
```
You don't need several of these checks.

- The `corporation?` check is redundant as you're checking `minor?`.
- AFAICT, `mergeable?` will always return true for a minor.
- `target_corporations.empty?` will only be true once there are no minors left to merge.
- `@merge_target` is for where you can merge more than two entities together.

lib/engine/game/g_1854/step/merge_minors.rb
ollybh: I don't know this bit of the codebase, but do you need the `@mergeable_` hash for this game? You might just need to return all non-closed minors.

lib/engine/game/g_1854/step/route.rb
andrewzwicky: Not sure what was happening before, but that was my recollection.  I tested it with this step deleted and it does work as advertised.  I'll delete this file.

lib/engine/game/g_1854/step/merge_minors.rb
andrewzwicky: This is the general flow for this step which i copied from one of 1841, 1862, 1867, 18ESP, 18NewEngland.  They all use basically the same logic for merging, and all of those have this same auto-action.  I didn't fully understand the intricacies re: track connection and stuff.  I will delete this, and there's still a long way to do and testing to do.

lib/engine/game/g_1854/step/merge_minors.rb
andrewzwicky: Probably not, it looks like a method to save time recomputing expensive graph calculations in other features.

lib/engine/game/g_1854/step/track_and_token.rb
andrewzwicky: I think I'd prefer to leave that for a future pr if possible so I can keep track of it more easily. Seems like a relatively big change.

andrewzwicky:
@ollybh I think this set is ready to check as well

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,99 @@
...
+  module Game
+    module G1854
+      module Step
+        class MergeMinors < Engine::Step::Base
+          def actions(entity)
+            return [] unless entity == current_entity
+            return [] unless entity.minor?
+            return [] unless @game.minor_mergers_allowed?
+
+            actions << ['merge']
```
ollybh, line 15: You haven't created the actions array yet, so can't append anything to it.
```suggestion
            actions = ['merge']
```

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,99 @@
...
+            # TODO: pass is not working, keeps same minor as current entity
+            # TODO: also need to elimiate pass when the two merging minors are selected.  You can't pass on picking A/B/C
+            actions << 'pass' unless @game.minor_mergers_required?
```
ollybh, line 18: Would this work?
```suggestion
            actions << 'pass' if @merging.nil? && !@game.minor_mergers_required?
```

lib/engine/game/g_1854/step/merge_minors.rb
```diff
@@ -0,0 +1,99 @@
...
+    module G1854
+      module Step
+        class MergeMinors < Engine::Step::Base
+          def actions(entity)
+            return [] unless entity == current_entity
+            return [] unless entity.minor?
+            return [] unless @game.minor_mergers_allowed?
+
+            actions << ['merge']
+            # TODO: pass is not working, keeps same minor as current entity
```
ollybh, line 16: What's the problem here? Nothing changes when you attempt to pass, you're stuck on this step?
This _might_ be something to do with your definition of `active?` below. Try adding a check against `@passed` in that method.

lib/engine/game/g_1854/step/merge_minors.rb
andrewzwicky: Done

lib/engine/game/g_1854/step/merge_minors.rb
andrewzwicky: Yeah, I was ready to ask to defer this until a future PR since I had never seen it working, but you're right it is the definition of `active?`  I'll add the @ passed  check in there.

andrewzwicky: Testing shows pass is working and pass button not present when selecting A/B/C

lib/engine/game/g_1854/step/track_and_token.rb
ollybh: I expect that you'd end up with less code if you did it this way, as you've got a lot of code in this step copied from `Step::HomeToken` that would no longer be needed.
But this can be left for later if you want. I don't think that there's anything wrong with this code, it just could be made simpler.

ollybh (approved):

PR #10425, opened 2024-03-01
[18ESP] Fixes interest calculation to round up.

### LINKED ISSUE #10404: 18ESP interest of the debt wrong calculated.
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

philcampeau:
Fixes #10404

## Before clicking "Create"

### Explanation of Change

Currently loan interest rounds down to nearest whole integer, but according to the rules it should round **up** to the nearest whole integer. 

Changed loan calculation from `interest = loan_amount * 0.5` to `interest = (loan_amount * 0.5).ceil`

This change will break any games where interest calculation was wrong, applied `archive_alpha_games` label.

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -1017,7 +1017,7 @@ module Engine
           @bank.spend(loan, player)
 
           loan_amount = loan
-          interest = loan_amount * 0.5
+          interest = (loan_amount * 0.5).ceil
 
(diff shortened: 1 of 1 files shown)


michaeljb:
@bentziaxl consider adding a section to https://github.com/tobymao/18xx/blob/master/.github/CODEOWNERS for games you implemented so that you'll automatically be added as a reviewer on PRs

bentziaxl (approved):
looks good

PR #10428, opened 2024-03-01
[18RoyalGorge] implement train-lengthening privates (Y3, G6, B5)

michaeljb:
https://github.com/tobymao/18xx/issues/10110

## Before clicking "Create"

### Explanation of Change

* B5 Track Engineer increase one train's length by one each OR

* Y3 Coal Creek Mines and G6 Coal Depot get coal cubes in different ways, but they are spent in the same way: the owning corporation can spend up to 2 coal cubes to increase one train's length by one per cube

* no train can be lengthened by more than one of these companies per OR; the `choose_state` logic in the route step ensures only one of these companies will be chosen at a time

### Any Assumptions / Hacks

* Depends on #10435 and #10436, but while in prealpha it doesn't matter if this one is merged first.

### DIFF
--- a/lib/engine/game/g_18_royal_gorge/entities.rb
+++ b/lib/engine/game/g_18_royal_gorge/entities.rb
@@ -36,15 +36,17 @@ module Engine
           {
             sym: 'Y3',
             name: 'Coal Creek Mines (Y3)',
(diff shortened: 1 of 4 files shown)


michaeljb:
Will split out the core changes to separate PRs.

PR #10448, opened 2024-03-05
[core] Allow companies to provide a discount when buying shares

michaeljb:
18RoyalGorge's share-buying step will implement `company_discounted_bundles`, which will be used for multiple private companies; see #10449 for a screenshot.

## Before clicking "Create"


roseundy (commented):
LGTM

PR #10463, opened 2024-03-08
18zoo - Implements fixes asked inside https://github.com/tobymao/18xx…

### LINKED ISSUE #10462: [18ZOO] selling not recorded - work in progress doesn't work properly
**What's your Game ID?**
`https://18xx.games/game/153535`

**How to Reproduce the Bug?**
*selling not recorded*
https://18xx.games/game/153535?action=96
- Previous step Serpentium sold PE
- Now PE is not recorded as sold, so Serpentium can buy again 
image

The expected behaviour is that item sold should be blocked
image

*work in progress doesn't work properly*
https://18xx.games/game/153535?action=198
image
- in the action 199 Serpentium used "Work in progress" and instead of going to second slot, it overrides the first city slot
image

The expected behaviour is that item goes to second slot
image

**Which Browser / Platform?**

carlorusso1984:
Fixes #10462

## Before clicking "Create"


lib/engine/player.rb
```diff
@@ -41,7 +41,7 @@ def corporation
     end
 
     def ==(other)
-      other&.player? && (@name == other&.name)
+      other&.player? && !other.player.nil? && (@name == other.name)
```
tobymao, line 44: why was this needed?

tobymao: you shouldn't call nil? other.playre is enough

carlorusso1984: As this discussion: https://18xxgames.slack.com/archives/C012K0CNY5C/p1708883331307309?thread_ts=1708871941.895459&cid=C012K0CNY5C
the issue is that in Javascript this code `other&.player?`  is translated into a check that player is not null, but in that case the Player is NilClass, so it's returning true also if the player is null
With this fix, the problem is solved also on JS

tobymao: can you do 

`!!other&.player? && name == other.name`

carlorusso1984: done in d293b0f

ollybh (approved):

PR #10471, opened 2024-03-12
[18NY] makes terminal offboards actually terminal

### LINKED ISSUE #10223: [18NY] NYC Formation Triggered Incorrectly
GameID: 150162

Turn link: https://18xx.games/game/150162?action=265

NYC formation is triggered, despite no Buffalo-Albany connection nor the 12H being purchased.

My guess is the bug is because both Buffalo and Albany are connected to the same red off-board area (Scranton) - however, Scranton is not pass through. I verified with Pierre (designer) that this should not trigger the formation.

philcampeau:
Fixes #10223

## Before clicking "Create"

### Explanation of Change

Terminal offboard track didn't have the terminal:1 tag on it, so the game was reading them as pass-throughs for this purpose. 

Tested on my end, game works as expected now. Please let me know if I'm neglecting anything with this fix.

### DIFF
--- a/lib/engine/game/g_18_ny/map.rb
+++ b/lib/engine/game/g_18_ny/map.rb
@@ -180,26 +180,27 @@ module Engine
 
         HEXES = {
           red: {
-            %w[A11] => 'offboard=revenue:yellow_10|green_20|brown_40|gray_80,groups:Ottawa;path=a:5,b:_0;border=edge:4',
+            %w[A11] => 'offboard=revenue:yellow_10|green_20|brown_40|gray_80,groups:Ottawa;path=a:5,b:_0,terminal:1;'\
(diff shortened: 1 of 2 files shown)


crericha (approved):
Nice find. I'm surprised this isn't default behavior for offboards that are rendered with the terminus track. Also glad you found this before I dug into graph to figure out why it wasn't working.

Pins will be needed.

benjaminxscott:
> surprised this isn't default behavior for offboards that are rendered with the terminus track

Should we change the default behavior?

philcampeau:
> > surprised this isn't default behavior for offboards that are rendered with the terminus track
> 
> Should we change the default behavior?

It would make sense, really. I can't see any reason why they shouldn't be.

PR #10532, opened 2024-03-19
[18 India] Implement Private Company Abilities

EdPovi:
- (n/a) Add the `pins` or `archive_alpha_games` label if this change will break existing games

## Implementation Notes

- Implement private company abilities
- Expanded on information on Phases and status
- Companies pay their owner when they close [updated core code and steps where companies close].
- Added "III'" phase to allow for multiple trains to be available at same time. 

### Explanation of Change
There a couple of changes to shared Core code

1.  Set up a hook in the base game class to 'do something' when a Company closes.
2.  Add a `'owning_player_token'` timing for ability, similar to 'owning_player_track'. 

### Any Assumptions / Hacks
NOTE:  Many of the changes are to use Company Abilities while owned by the player (president).  
NOTE:  `'owning_player_token'`  will require that `token_owner(entity)` accounts for a player owned company
NOTE:  Jewelry concession icon is just a placeholder, will be updated when icons available from designer.


lib/engine/game/g_18_india/game.rb
```diff
@@ -761,25 +802,25 @@ def commodity_bonus(route, _stops)
               revenue += bonus
             when 'SPICES'
               # SPICES => KOCHI [G36] + 70
+              if visited_names.include?('KOCHI')
+                bonus = 70
               # SPICES => COLOMBO [K40] + 50
               # SPICES => CHENNAI [K30] + 50
+              elsif visited_names.include?('COLOMBO') ||
+                    visited_names.include?('CHENNAI')
```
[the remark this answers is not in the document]
ollybh: This all could be clearer if you declare a hash for the various spice bonuses. Something like:
```
SPICE_BONUSES = {
    'KOCHI' => 70,
    'COLOMBO' => 50,
    'CHENNAI' => 50,
    'LAHORE' => 40,
    # …
}.freeze
```
And then, if you're looking for the maximum value visited location, you'd have something like this:
`visited_names.map { |loc| SPICE_BONUSES[loc] || 0 }.max`
[I feel that there should be a more elegant way of writing this].

lib/engine/game/g_18_india/game.rb
```diff
@@ -804,6 +845,10 @@ def commodity_bonus(route, _stops)
               bonus = 30 if visited_names.include?('CHINA') || visited_names.include?('NEPAL')
               claim_concession(['RICE'], corporation) if bonus.positive?
               revenue += bonus
+            when 'JEWELRY'
+              # Jewelry concession pays 20 if delivered to any commodity destination
+              bonus = 20 unless (visited_names & COMMODITY_DESTINATIONS).empty?
```
ollybh, line 850:
```suggestion
              bonus = 20 if visited_names.intersect?(COMMODITY_DESTINATIONS)
```

lib/engine/game/g_18_india/step/assign.rb
```diff
@@ -0,0 +1,33 @@
...
+      module Step
+        class Assign < Engine::Step::Assign
+          # Allow company 'P6' to assign to any hex that isn't a town or city
+          def available_hex(entity, hex)
+            return unless entity.company?
+            return unless entity.id == 'P6'
+            return if hex.assigned?(entity.id)
+            return unless hex.tile.city_towns.empty? # hex doesn't have a town or city
+
+            @game.hex_by_id(hex.id).neighbors.keys
```
ollybh, line 17: This method should just be returning a boolean value. Can you just do this?
```suggestion
            hex.tile.city_towns.empty?
```

lib/engine/game/g_18_india/step/special_track.rb
```diff
@@ -0,0 +1,44 @@
...
+            return false if tile.color != :yellow && ability.upgrade_count.zero?
+            return false if tile.color == :yellow && ability.lay_count.zero?
+
+            old_tile = hex.tile
+            if @game.yellow_town_to_city_upgrade?(old_tile, tile)
+              all_new_exits_valid = tile.exits.all? { |edge| hex.neighbors[edge] }
+              return false unless all_new_exits_valid
+
+              return old_tile.paths.all? { |old| tile.paths.any? { |new| old.exits == new.exits } } &&
+                     !(tile.exits & hex_neighbors(entity, hex)).empty?
```
ollybh, line 23: What are you testing here? I can't follow the nested iterations.

ollybh (changes requested):

lib/engine/game/base.rb
```diff
@@ -1932,17 +1932,20 @@ def after_par(corporation)
         place_home_token(corporation) if self.class::HOME_TOKEN_TIMING == :par
       end
 
+      # This is a hook to do something before company closes
+      def company_is_closing(company, silent = false)
```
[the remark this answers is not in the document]
EdPovi: I can change the name of the method.   I pulled in the log message so that it can be modified, in the case of 18 India, the message includes the amount of money received for closing the company.

lib/engine/game/g_18_india/game.rb
```diff
@@ -79,15 +79,52 @@ def num_certs(entity)
...
+            tiles: %i[yellow green brown gray],
+            operating_rounds: 2,
+            status: %w[warrants_expire no_gauge_change gauge_change_removal],
+          },
         ].freeze
 
+        STATUS_TEXT = Base::STATUS_TEXT.merge(
+          'gipr_may_operate' => ['GIPR may operate', 'GIPR may operate if it has been floated (three 10% shares in play).'],
+          'phase_four_trains' => ['Phase IV trains available', 'All Phase IV trains are simultaneously available.'],
+          'warrants_expire' => ['Guaranty Warrants expire', 'Guaranty Warrants immediately expire.'],
```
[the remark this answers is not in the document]
EdPovi: I happened to take this spelling from the rulebook, (can reference section 2.3 or 3.4 of rulebook).  It is an less common variant of the word in my experience.

From the internet (so take it with several grains of salt...): 

> Which is correct guaranty or guarantee?
> Guarantee is both a verb and a noun. Guaranty is a spelling variant for the noun, used in certain legal contexts. I can guarantee that Vicky will be back here within the week. What guarantee (or guaranty) can you offer to the other parties?

> What is difference between guaranty and guarantee?
> In contrast, the noun forms of guaranty and guarantee are similar, but guaranty has a narrower meaning. Companies make written or verbal guarantees all the time, but guaranty refers specifically to a written agreement that one party will pay the money required if another party fails to do so.Nov 18, 2021

lib/engine/game/g_18_india/game.rb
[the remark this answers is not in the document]
EdPovi: I like your ideas here.  Intersect looks cleaner. 

I will need to refactor some of this code later, as the Jewelry Concession can add a wrinkle to my current implementation.
I will verify with the designers that Jewelry could be on the same hex as other commodities, in which case I think it may be difficult for hex//tile to have more than one name.   Once I get the graphics from them, I'll add icons to the hexes, which I can use for the comparison, or simply use the hex ID.   I did like using the words / names as the code was easier to read and verify.

lib/engine/game/g_18_india/step/assign.rb
EdPovi: I modified the base code from Engine::Step::Assign. 
I added to the guard clauses. 

https://github.com/tobymao/18xx/blob/02f8ebe84a373e5768337512c2de0a45f1fede76/lib/engine/step/assign.rb#L67

EdPovi: Tested the modification.  It seems to work.  I will wonder if the base code I inherited this from should be updated.

lib/engine/game/g_18_india/step/special_track.rb
EdPovi: This section is for a yellow town to green single city upgrade.  This test is to ensure the old exits are preserved on the new tile.  

I have copied this from another game:
https://github.com/tobymao/18xx/blob/02f8ebe84a373e5768337512c2de0a45f1fede76/lib/engine/game/g_18_ny/step/track.rb#L23

lib/engine/game/base.rb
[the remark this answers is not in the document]
EdPovi: Modified name to indicate after ability and reverted those cases I used this call for closing via event / phase.

lib/engine/game/g_18_india/step/assign.rb
ollybh: The base code might be written this way to check that there are some neighbouring hexes, to make sure you can't select a dead hex that can never be connected to anything. Could this be a problem here?

lib/engine/game/g_18_india/step/special_track.rb
ollybh: That seems a very complicated way to compare whether the exits match.
Would `(old_tile.exits - tile.exits).empty?` work?

ollybh: Ah, I see what the original test is doing. It is for track paths that run from edge to edge, the test makes sure that these are preserved. These tiles only have paths between edges and the central town/city, so the simpler test should be enough.

EdPovi: I have verified that the simpler test works.  Thanks!

lib/engine/game/g_18_india/step/assign.rb
EdPovi: Your suggested code work when tested.  The only hex I saw that needed to be excluded was the blue water hex, added a guard clause.

PR #10539, opened 2024-03-22
pluralize spreadsheet column for corps

benjaminxscott:
I'm pretty sure this was a typo? The other headers are either plural or implicitely a plural-ish thing like a bank


ollybh:
I think singular corporation is right in this context. These columns (cash/trains/tokens/…) are the stuff owned by each corporation. The 'players' header is pluralised because there are columns for multiple players, and 'prices' because there are (usually) more than one price column.

scottredracecar:
Agree with Oliver.

Issue #10484, opened 2024-03-17
Multiple games broken with error `uninitialized constant Engine::SharePool::Set`


lavoieba:
Below is the message seen at the top of the screen after entering the game. The game has also reset itself to the SR1 from MR4.1.

We're sorry this game cannot be continued due to Set: uninitialized constant Engine::SharePool::Set
Broken action: { "type": "pass", "entity": 776, "entity_type": "player", "id": 121, "created_at": 1708890243 }
Please [raise a bug report](https://github.com/tobymao/18xx/issues?q=is%3Aissue+is%3Aopen+label%3A18USA,%22needs%20triage%22) and include this link (https://18xx.games/game/153253?action=120)

richpri:
I have the same type of message for 2 of my games. 
https://18xx.games/game/156063?action=39
https://18xx.games/game/151868?action=21
Will these games have to be abandoned?

ollybh:
> I have the same type of message for 2 of my games. https://18xx.games/game/156063?action=39 https://18xx.games/game/151868?action=21 Will these games have to be abandoned?

No, you just need to wait for a fix to be pushed to the server code. Everything should then start working again.

PR #10541, opened 2024-03-22
[18RoyalGorge] allow president sales to market for only specified corporations

michaeljb:
#10110 

This recreates #10447 (and effectively reverts #10506) with the following changes:

* in the switch statements, `when ::Set` is used instead of `when Set` to ensure that the root-level `Set` class is used, instead of `Engine::SharePool::Set`
* the logic added to `Engine::SharePool` is moved into a new `SharePool` subclass specific to 18RoyalGorge, so that if the above measure does not work, only 18RoyalGorge should be affected; this does duplicate a lot of code so if it looks like this works I would like to follow up with another PR moving the logic into the main `SharePool` class

I was unable to locally reproduce #10484 and the other issues listed in #10506, so like that PR, this is a "hopeful" fix.

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -145,7 +145,6 @@ module Engine
       # many remain?
       CERT_LIMIT_COUNTS_BANKRUPTED = false
 
-      # boolean or a Set<String> of corporation IDs
       PRESIDENT_SALES_TO_MARKET = false
 
       MULTIPLE_BUY_TYPES = %i[multiple_buy].freeze
@@ -1122,19 +1121,8 @@ module Engine
         max_bundle&.price || 0
       end
 
-      def president_sales_to_market?(corporation)
-        case self.class::PRESIDENT_SALES_TO_MARKET
-        when true
-          true
-        when ::Set
-          self.class::PRESIDENT_SALES_TO_MARKET.include?(corporation.id)
-        else
-          false
-        end
-      end
-
       def value_for_dumpable(player, corporation)
-        return value_for_sellable(player, corporation) if president_sales_to_market?(corporation)
+        return value_for_sellable(player, corporation) if self.class::PRESIDENT_SALES_TO_MARKET
 
         max_bundle = bundles_for_corporation(player, corporation)
           .select { |bundle| bundle.can_dump?(player) && @share_pool&.fit_in_bank?(bundle) }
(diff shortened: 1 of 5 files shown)


lib/engine/game/base.rb
```diff
@@ -1121,8 +1122,19 @@ def value_for_sellable(player, corporation)
...
+        when ::Set
+          self.class::PRESIDENT_SALES_TO_MARKET.include?(corporation.id)
+        else
+          false
+        end
+      end
+
       def value_for_dumpable(player, corporation)
-        return value_for_sellable(player, corporation) if self.class::PRESIDENT_SALES_TO_MARKET
+        return value_for_sellable(player, corporation) if president_sales_to_market?(corporation)
```
ollybh, line 1137: If there is still a problem with the `when ::Set` statement, then this call is going to affect games other than 18RoyalGeorge.

ollybh (commented):
I believe that, if we still get errors from the Set reference, this will affect more games than just 18RoyalGeorge.
Could you still do the refactoring from commit b54ae960, but just have the base class `Engine::Game.allow_president_sale?` method just return the `self.class::PRESIDENT_SALES_TO_MARKET` constant? And you could just override this method in the 18RoyalGeorge code until it's been shown that this isn't going to fail again.

ollybh:
@michaeljb Do you have any idea why neither of us were able to reproduce #10484 locally? I'd hope that a test could be added that would fail on commit b54ae960, any maybe catch any similar issues going forward.

michaeljb:
> @michaeljb Do you have any idea why neither of us were able to reproduce #10484 locally? I'd hope that a test could be added that would fail on commit [b54ae96](https://github.com/tobymao/18xx/commit/b54ae96044379fdc86c02bfe20c8e57b01034810), any maybe catch any similar issues going forward.

Not sure. I was thinking there might be some difference in the compilation depending on the environment, but nothing in [lib/assets.rb](https://github.com/tobymao/18xx/blob/master/lib/assets.rb) or the code search results for `RACK_ENV` look like that would be the case. Maybe something with the `precompiled` flag? A local server runs without that set.

lib/engine/game/base.rb
michaeljb: Moved all code on this PR into 18RoyalGorge's game class only.

michaeljb:
@ollybh I still haven't found the root cause or how to reproduce the issue, but the code is entirely contained in 18RoyalGorge's files, and is at least functioning correctly on my local server.

ollybh (approved):
Thanks, this does now look fully contained within the 18RoyalGeorge code, so I can't foresee any problems with other games, even if the Set reference still doesn't work.

PR #10549, opened 2024-03-24
[21Moon] Terminal private revenue s.b. 15 space bucks

benjaminxscott:
The BGG-hosted rules indicate Terminal revenue is 15, Scott confirmed
https://boardgamegeek.com/filepage/263361/21moon-rules

## Before clicking "Create"

Definitely needs pins


roseundy (commented):
Looks OK to me. That value changed several times.

I would have preferred a bug to have been filed first.

PR #10554, opened 2024-03-26
Adds hotkeys for speeding up operations, especially end of game runs.

philcampeau:
## Before clicking "Create"

### Explanation of Change

I've added a few hotkeys that will allow players to quickly speed through the final ORs of a game

[1] - handles all passing (Skip track/skip token/skip trains/skip companies)
[2] - clears all saved runs
[3] - hits the auto route button
[Enter] - Selects 'Pay Out' option, and also 'Submit' for the run

I find this really helpful with everyone agrees the game is over but still want to see a final score. One player enables master mode, goes through one OR where they maximize each route (if desired) and then it's just 1 and Enter for the rest of the game

### DIFF
--- a/assets/app/view/game/dividend.rb
+++ b/assets/app/view/game/dividend.rb
@@ -71,7 +71,11 @@ module View
             process_action(Engine::Action::Dividend.new(@step.current_entity, kind: type))
             cleanup
           end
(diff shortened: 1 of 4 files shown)


bentziaxl:
looks good. Should the documentation on the different hot keys be added to the wiki somwhere?

bentziaxl (approved):

PR #10575, opened 2024-03-29
[18Uruguay] Only auto withhold when nationalized

patrikolesen:
[18Uruguay] Only auto withhold when nationalized

Corrected a bug and only withhold when nationalized

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_uruguay/step/dividend.rb
+++ b/lib/engine/game/g_18_uruguay/step/dividend.rb
@@ -39,9 +39,10 @@ module Engine
           end
 
           def auto_actions(entity)
-            return [Engine::Action::Dividend.new(current_entity, kind: 'withhold')] if entity.loans.size.positive?
+            return [] unless @game.nationalized?
+            return [] if entity.loans.empty?
 
-            []
+            [Engine::Action::Dividend.new(current_entity, kind: 'withhold')]
           end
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_uruguay/step/dividend.rb
```diff
@@ -39,7 +39,10 @@ def description
           end
 
           def auto_actions(entity)
-            return [Engine::Action::Dividend.new(current_entity, kind: 'withhold')] if entity.loans.size.positive?
+            if entity.loans.size.positive? && @game.nationalized?
+              return [Engine::Action::Dividend.new(current_entity,
+                                                   kind: 'withhold')]
+            end
 
             []
```
ollybh, line 47: Testing `Array.size.positive?` isn't idiomatic Ruby – the `Array.empty?` method is the way to do this test. I'd also suggest reordering this method, I think it looks clearer like this:
```suggestion
            return [] unless @game.nationalized?
            return [] if entity.loans.empty?

            [Engine::Action::Dividend.new(current_entity,kind: 'withhold')]
```

patrikolesen: Agree it looks cleaner, updated according to suggestion

ollybh (approved):

PR #10577, opened 2024-03-29
[18Neb] t'aint no train discounts

### LINKED ISSUE #10576: [18NEB] Info Tab incorrectly shows train upgrade discount
Game ID is 158026, link to action is: https://18xx.games/game/158026?action=381 but I don't think this is just a text display issue

The info tab in 18NEB shows a $300 trade-in discount to upgrade a 4, 5, or 6 to a 4D. However, this is not implemented. When I went to look for the discount in the rules, I couldn't find it, and didn't see it on game images on BGG.

My assumption is that this is not actually something in the game, and the Info tab should be corrected. If it is something that is in the game, then it's not currently implemented (or, at least, was not present in the above linked action).

benjaminxscott:
Fixes #10576

Rules don't indicate there's a discount

## Before clicking "Create"

- [no, since alpha] Add the `pins` or `archive_alpha_games` label if this change will break existing games

### DIFF
--- a/lib/engine/game/g_18_neb/game.rb
+++ b/lib/engine/game/g_18_neb/game.rb
@@ -148,7 +148,6 @@ module Engine
             price: 900,
             num: 20,
             available_on: '6',
-            discount: { '4' => 300, '5' => 300, '6' => 300 },
           },
         ].freeze
 


ollybh (approved):

crericha (commented):
Approved (there is no button for some reason).

PR #10580, opened 2024-03-30
[18Uruguay] No RLPTA stock before phase 3

patrikolesen:
[18Uruguay] No RLPTA stock before phase 3
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_uruguay/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_uruguay/step/buy_sell_par_shares.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/buy_sell_par_shares'
+
+module Engine
+  module Game
+    module G18Uruguay
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParShares
+          def can_buy?(entity, bundle)
+            return false if bundle.corporation == @game.rptla && @game.phase.name == '2'
+
+            super(entity, bundle)
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_uruguay/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module Engine
+  module Game
+    module G18Uruguay
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParShares
+          def can_buy?(entity, bundle)
+            return false if bundle.corporation == @game.rptla && @game.phase.current[:name] == '2'
+
+            super(entity, bundle)
```
ollybh, line 13: A couple of very minor points:

- As you're checking the current phase, you can just write `@game.phase.name` instead of `@game.phase.current[:name]`.
- Calling `super` with no arguments results in a call to the superclass's method of the same name, with the same arguments passed. It's more normal for this codebase to use the bare `super` call, instead of explicitly passing on the arguments.

Have you considered using a phase status for when these RPTLA shares are available? It would make this visible to players on the 'info' tab.

patrikolesen: Good suggestions

Updated @game.phase.name and updated phase status

lib/engine/game/g_18_uruguay/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require_relative '../../../step/buy_sell_par_shares'
+
+module Engine
+  module Game
+    module G18Uruguay
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParShares
+          def can_buy?(entity, bundle)
+            return false if bundle.corporation == @game.rptla && @game.phase.name == '2'
```
ollybh, line 11: As you've added the availability of RPTLA shares to the phase's status, you should use it here:
```suggestion
            return false if bundle.corporation == @game.rptla && !@game.phase.status.include?('rptla_available')
```

ollybh (changes requested):

PR #10581, opened 2024-03-30
  [18Uruguay] Change to Alpha

patrikolesen:
[18Uruguay] Change to Alpha

## Before clicking "Create"


ollybh (changes requested):
This pull request includes the changes in #10574, #10575 and #10580. Can you please remove those commits from this branch?

patrikolesen:
Will wait for the others to be merge before transition into alpha

PR #10585, opened 2024-03-30
21Moon: fix space port upgrade broken by tile renumbering

### LINKED ISSUE #10579: [21Moon] Error on Space Port upgrade
In Game 158089 corporate round 3, I need to pass with VH but the error
no implicit conversion of NilClass into Integer
Pops up in the banner and I cannot progress this game.
https://18xx.games/game/158089

**Which Browser / Platform?**
Chrome Version 123.0.6312.86 (Official Build) (64-bit)

roseundy:
Fixes #10579

### DIFF
--- a/lib/engine/game/g_21_moon/game.rb
+++ b/lib/engine/game/g_21_moon/game.rb
@@ -189,7 +189,7 @@ module Engine
         MAX_OWNERSHIP_2P = 60
         LAST_OR = 11
         SP_HEX = 'E9'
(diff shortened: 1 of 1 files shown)


benjaminxscott:
likely needs pins? should we deploy the ~4 PRs at the same time?

roseundy:
Yes, it should be pinned along with the other tile numbering fix PR.

roseundy:
Actually, I don't _think_ this one needs a pin. As long as you validate before you pin it shouldn't matter.

PR #10587, opened 2024-03-31
[18esp] more 18esp fixes

bentziaxl:
fix logic when a combined train is blocked and go across the mountain passes
fix trains not showing when it should 
fix scenario where corp ipoed but hasn't floated


lib/engine/game/g_18_esp/step/acquire.rb
```diff
@@ -100,7 +100,7 @@ def can_swap?
           end
 
           def mz?(entity)
-            entity.name == 'MZ'
+            entity.name == 'MZ' && @game.corporations.any? { |c| c.name == 'MZA' }
```
ollybh, line 103: Should this be `c.id == 'MZA'`?

lib/engine/game/g_18_esp/game.rb
```diff
@@ -860,6 +860,7 @@ def must_buy_train?(entity)
             end
           end
           trains = trains.dup.reject { |t| t.track_type == :broad } if north_corp?(entity) && !entity.interchange?
+          trains = trains.dup.reject { |t| t.track_type == :all } if combined_train_blocked?(entity)
```
ollybh, line 863: I'm not keen on all these calls to `trains.dup`. Could the entire method be rewritten to something like this?
```
        def must_buy_train?(entity)
          return false if depot.depot_trains.empty?
          
          entity.trains.none? do |train|
            next false if extra_train?(train)
            
            case train.track_type
            when :narrow
              north_corp?(entity) || (entity.interchange? && entity.type != :minor)
            when :broad
              !north_corp?(entity) || entity.interchange?
            when :all
              !combined_train_blocked?(entity)
          end
        end
```
I've probably got the logic wrong in there, but I think it would be much clearer if it looked something like this.

ollybh (changes requested):

bentziaxl:
@ollybh i addressed your comments

ollybh (approved):

PR #10590, opened 2024-03-31
Adding support for stackable assignments

patrikolesen:
In 18 Uruguay there is a need to assigne multiple assignments to one tile. 
In current framework the tile quickly gets overloaded with the assignments

This is a proposal for a possibility to stack assignments

In this proposal you can tag an assignment with stackable like this

`          target.assign!(good, Engine::Assignable::STACKABLE)
`

When drawing the framework will then stack all stackable assignments and add a counter if the number of elements are 2 or larger

image

## Before clicking "Create"

### DIFF
--- a/assets/app/view/game/part/assignments.rb
+++ b/assets/app/view/game/part/assignments.rb
@@ -37,7 +37,7 @@ module View
           @assignments_to_show = []
           stack_group_count = {}
           stack_group_img = {}
-          @assignments.keys.each do |assignment|
+          @assignments.each_key do |assignment|
             stack_group = @game.assignment_stack_group(assignment)
             img = @game.assignment_tokens(assignment, setting_for(:simple_logos, @game))
 
--- a/assets/app/view/game/part/assignments.rb
+++ b/assets/app/view/game/part/assignments.rb
@@ -17,7 +17,7 @@ module View
 
         def preferred_render_locations
           if layout == :pointy
-            case @assignments&.size
+            case @assignments_to_show&.size
             when 1
               POINTY_SMALL_ITEM_LOCATIONS
             when 2
@@ -25,7 +25,7 @@ module View
             else
               POINTY_WIDE_ITEM_LOCATIONS
             end
-          elsif @assignments.one?
+          elsif @assignments_to_show&.one?
             SMALL_ITEM_LOCATIONS
           else
             WIDE_ITEM_LOCATIONS
@@ -34,6 +34,55 @@ module View
 
         def load_from_tile
           @assignments = @tile.hex&.assignments || {}
+          @stack_groups = @tile.hex&.assignment_stack_groups || {}
+          @assignments_to_show = []
+          @stack_groups.each do |key, assignemnts|
+            if key == 'NONE'
+              assignemnts.each do |assignemnt, _value|
+                @assignments_to_show.append({
+                                              'img' => @game.assignment_tokens(assignemnt, setting_for(:simple_logos, @game)),
+                                              'count' => 1,
+                                            })
+              end
+            else
+              @assignments_to_show.append({
+                                            'img' => @game.assignment_tokens(assignemnts.keys.last,
+                                                                             setting_for(:simple_logos, @game)),
+                                            'count' => assignemnts.size,
+                                          })
+            end
+          end
+        end
+
+        def stack_count_background_props(axis, axis_value)
+          props = {
+            attrs: {
+              href: '/icons/stack_background.svg',
+              width: "#{ICON_RADIUS * 2}px",
+              height: "#{ICON_RADIUS * 2}px",
+            },
+          }
+
+          props[:attrs][axis] = axis_value + 10
+          axis2 = (POINTY_TALL_ITEM_LOCATIONS.any?(render_location) ? :x : :y)
+          props[:attrs][axis2] = -5
+          props
+        end
+
+        def stack_count_props(axis, axis_value)
+          props = {
+            attrs: {
+              'dominant-baseline': 'central',
+              fill: 'black',
+            },
+            style: {
+              fontSize: '20px',
+            },
+          }
+          props[:attrs][axis] = axis_value + 35
+          axis2 = (POINTY_TALL_ITEM_LOCATIONS.any?(render_location) ? :x : :y)
+          props[:attrs][axis2] = 8
+          props
         end
 
         def render_part
@@ -41,22 +90,31 @@ module View
           multiplyer = (POINTY_WIDER_ITEM_LOCATIONS.any?(render_location) ? 3 : 1)
 
           if @game
-            children = @assignments.keys.map.with_index do |assignment, index|
-              img = @game.assignment_tokens(assignment, setting_for(:simple_logos, @game))
+            children = []
+            @assignments_to_show.each.with_index do |assignment, index|
+              count = assignment[:count]
 
               props = {
                 attrs: {
-                  href: img,
+                  href: assignment[:img],
                   width: "#{ICON_RADIUS * 2}px",
                   height: "#{ICON_RADIUS * 2}px",
                 },
               }
-              props[:attrs][axis] = ((index - ((@assignments.size - 1) / 2)) * multiplyer * DELTA).round(2)
 
-              h(:image, props)
+              axis_value = ((index - ((@assignments_to_show.size - 1) / 2)) * multiplyer * DELTA).round(2)
+              props[:attrs][axis] = axis_value
+
+              children.append(h(:image, props))
+              if count > 1
+                props = stack_count_background_props(axis, axis_value)
+                children.append(h(:image, props))
+
+                props = stack_count_props(axis, axis_value)
+                children.append(h(:text, props, count))
+              end
             end
           end
-
           h(:g, { attrs: { transform: "#{rotation_for_layout} translate(#{-ICON_RADIUS} #{-ICON_RADIUS})" } },
             [h(:g, { attrs: { transform: translate } }, children)])
         end
--- a/assets/app/view/game/part/assignments.rb
+++ b/assets/app/view/game/part/assignments.rb
@@ -46,8 +46,8 @@ module View
             @assignments_to_show.append({ 'img' => img, 'count' => 1 }) if stack_group.nil?
           end
 
-          stack_group_count.keys.each do |group|
-            @assignments_to_show.append({ 'img' => stack_group_img[group], 'count' => stack_group_count[group] })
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_18_uruguay/farm.rb
```diff
@@ -79,7 +79,7 @@ def process_assign(action)
           target = action.target
           good = retreive_goods!(@farm_id)
 
-          target.assign!(good)
+          target.assign!(good, Engine::Assignable::STACKABLE)
```
crericha, line 82: `target.assign!(good, stackable = true)` would better match how we've designed the interfaces in the framework.

lib/engine/assignable.rb
```diff
@@ -2,6 +2,8 @@
 
 module Engine
   module Assignable
+    ASSIGNED = 1
+    STACKABLE = 2
```
crericha, line 6: In Ruby, you generally use symbols instead of defining constants. `assignments[key] = :stackable`

crericha:
Can you experiment with other ways of visualizing the number? As it is now, I can't make out what the image is below the number because the number covers too much of the image. Maybe a small circle with the number that overlaps the icons circle. Oo , but slightly overlapped.

lib/engine/assignable.rb
```diff
@@ -10,7 +12,7 @@ def assigned?(key)
       assignments.key?(key)
     end
 
-    def assign!(key, value = true)
```
crericha, line 13: 1870 is using the original signature of this method. You'll have to figure out what its doing with it and make sure not to break it.

assets/app/view/game/part/assignments.rb
```diff
@@ -16,16 +16,24 @@ class Assignments < Base
         DELTA = (ICON_RADIUS * 2) + 2
 
         def preferred_render_locations
+          unless @assignments.nil?
+            assignments_to_show = @assignments.reject { |_key, value| value == Engine::Assignable::STACKABLE }
+            stack_count = @assignments.size - assignments_to_show.size
+            if stack_count.positive?
+              assignments_to_show[@assignments.key(Engine::Assignable::STACKABLE)] =
+                Engine::Assignable::STACKABLE
+            end
```
crericha, line 25: If I'm reading correctly, this code seems to assume there is only one stackable type on a hex. If so, this should be more generic and each type of icon should be separately stackable.

ollybh (changes requested):
I agree with @crericha that the suggested display of stacked assignments is not clear.
I'm also not too keen on the way that this has been implemented. Instead of each individual assignment having a stackable value/attribute, would it make more sense for this to a value defined at the hex or game level, where identical assignments are automatically stacked if there are more than this value?

assets/app/view/game/part/assignments.rb
patrikolesen: Updated to be able to support mulitple stacks

lib/engine/assignable.rb
patrikolesen: Adding stack_group: to not mess with the value

lib/engine/assignable.rb
patrikolesen: Change for a generic stack_group: so that we can support multiple stack_groups

lib/engine/game/g_18_uruguay/farm.rb
patrikolesen: Changed to a generic stack_group: so that we can support multiple stack_groups

patrikolesen:
Testing with circle instead of rect
image

lib/engine/assignable.rb
```diff
@@ -6,16 +6,23 @@ def assignments
...
+    def assignment_stack_groups
+      @assignment_stack_groups ||= {}
+    end
+
     def assigned?(key)
       assignments.key?(key)
     end
 
-    def assign!(key, value = true)
+    def assign!(key, value = true, stack_group: 'NONE')
```
ollybh, line 17: If this is the approach taken* then the `stack_group` parameter should be a label, not a string:
```suggestion
    def assign!(key, value = true, stack_group: :none)
```
[* I'm still not keen on this. The presentation of the assignments is an interface decision, so I believe the code for managing stacks of assignments should be done in the game view, not the game logic for assigning stacks to hexes.]

ollybh:
@patrikolesen Can you try and avoid forced pushes to the branch after you've submitted a pull request? It makes it harder to see what has been changed in the code in response to review comments.

lib/engine/assignable.rb
patrikolesen: I will try to rewrite on game level instead.
How should I then add the number should I have a unique icon for each number or are there a better solution?

ollybh:
> Testing with circle instead of rect image

That looks better. Maybe add a small amount of opacity (10–20%) to the circle fill so that you can see a little bit of the icon where it overlaps.

lib/engine/assignable.rb
ollybh: Do you mean the number of assignments shown in the circle? I like the current approach of drawing a circle and having the number as a text node in the SVG.

patrikolesen: Re-read your comment, the suggestion is rather to add it to game view not necesserly to the game?
Your main objection is that this proposal mixes the game logic for assignments with the graphical presentation, will see if I can find a better way to do this.

ollybh: I was thinking that you could leave the current code for `assign!` unchanged, and then modify `View::Game::Part::Assignments` to automatically stack assignments. In that class you could make a call to something like `@game.stack_assignments(tile, assignments)` and that method would be able to determine if there are multiple assignments of the same type and decide if they should be stacked. The default implementation should not change how existing games render this, so maybe don't stack assignments if there are less than six of the same type. The stacking limit could be defined as a constant, so this could be easily overridden in the 18Uruguay code.

patrikolesen: Let us see if I understood you correctly. Have updated with a new proposal that I think matches your suggestion

assets/app/view/game/part/assignments.rb
```diff
@@ -34,29 +34,84 @@ def preferred_render_locations
...
+            img = @game.assignment_tokens(assignment, setting_for(:simple_logos, @game))
+
+            stack_group_count[stack_group] = (stack_group_count[stack_group] || 0) + 1 unless stack_group.nil?
+            stack_group_img[stack_group] = img unless stack_group.nil?
+            @assignments_to_show.append({ 'img' => img, 'count' => 1 }) if stack_group.nil?
+          end
+
+          stack_group_count.keys.each do |group|
+            @assignments_to_show.append({ 'img' => stack_group_img[group], 'count' => stack_group_count[group] })
+          end
```
ollybh, line 51: You can use the `Hash.each` method to iterate through key/value pairs, instead of extracting an array of keys and then looking up the values in the hash.
```suggestion
          stack_group_count.each do |group, count|
            @assignments_to_show.append({ 'img' => stack_group_img[group], 'count' => count })
          end
```

assets/app/view/game/part/assignments.rb
```diff
@@ -34,29 +34,84 @@ def preferred_render_locations
 
         def load_from_tile
           @assignments = @tile.hex&.assignments || {}
+          @assignments_to_show = []
+          stack_group_count = {}
+          stack_group_img = {}
+          @assignments.keys.each do |assignment|
```
ollybh, line 40:
```suggestion
          @assignments.each_key do |assignment|
```

assets/app/view/game/part/assignments.rb
```diff
@@ -34,29 +34,84 @@ def preferred_render_locations
 
         def load_from_tile
           @assignments = @tile.hex&.assignments || {}
+          @assignments_to_show = []
+          stack_group_count = {}
```
ollybh, line 38: Setting a default value of zero for the hash would simplify your code later on.
```suggestion
          stack_group_count = Hash.new(0)
```

assets/app/view/game/part/assignments.rb
```diff
@@ -34,29 +34,84 @@ def preferred_render_locations
...
+          stack_group_img = {}
+          @assignments.keys.each do |assignment|
+            stack_group = @game.assignment_stack_group(assignment)
+            img = @game.assignment_tokens(assignment, setting_for(:simple_logos, @game))
+
+            stack_group_count[stack_group] = (stack_group_count[stack_group] || 0) + 1 unless stack_group.nil?
+            stack_group_img[stack_group] = img unless stack_group.nil?
+            @assignments_to_show.append({ 'img' => img, 'count' => 1 }) if stack_group.nil?
```
ollybh, line 46: Having a single if/else statement is clearer than the repeated unless/if modifiers all testing the same thing. And this is where specifying the default value of zero for the `stack_group_count` hash means that you don't need the slightly awkward test for the existence of the hash key.
```suggestion
            if stack_group
              stack_group_count[stack_group] += 1
              stack_group_img[stack_group] = img
            else
              @assignments_to_show.append({ 'img' => img, 'count' => 1 })
            end
```

assets/app/view/game/part/assignments.rb
```diff
@@ -34,29 +34,84 @@ def preferred_render_locations
...
-              props[:attrs][axis] = ((index - ((@assignments.size - 1) / 2)) * multiplyer * DELTA).round(2)
 
-              h(:image, props)
+              axis_value = ((index - ((@assignments_to_show.size - 1) / 2)) * multiplyer * DELTA).round(2)
+              props[:attrs][axis] = axis_value
+
+              children.append(h(:image, props))
+              if count > 1
+                props = stack_count_background_props(axis, axis_value)
+                children.append(h(:image, props))
```
ollybh, line 108: This looks like you're adding the assignment image twice if `count > 1`. Is that right?

ollybh (changes requested):
Thanks for redoing this, I much prefer the less invasive changes to the game code. I've made a few suggestions where adding a sprinkling of Ruby syntactic sugar would improve the code.

lib/engine/game/g_18_uruguay/farm.rb
patrikolesen: Removed

lib/engine/assignable.rb
patrikolesen: Removed

assets/app/view/game/part/assignments.rb
patrikolesen: No, the first image is the assignment.
The second is the circle that is only appended in the case of a stacked assignment

ollybh (approved):

PR #10597, opened 2024-04-03
[18Neb] Bug fixes

crericha:
@michaeljb or @roseundy can you please take a quick look at the core changes? The `blocks_hexes` abilities will now properly remove the block from the map when the ability goes away.


rjaey:
Somewhat linked to p4 bug fix: Would it be possible to also mention the blocking of C7 by P4 on the description of p4 under entities? It wasn't clear to me why I couldn't build on C7 until I went back to the rules and checked p4 there.

PR #10607, opened 2024-04-05
18hiawatha prealpha

philcampeau:
## Before clicking "Create"

### Explanation of Change

Initial commit (nearly complete implementation) of 18hiawatha


benjaminxscott:
seems cool to merge / deploy since the core changes seem like a no-op for other games (i.e. won't break stuff)

if it's not urgent I wouldn't personally deploy quick just since I'm on vaca but if you have access to deploy then its cool

PR #10613, opened 2024-04-08
[1847 AE] Fix share count bug

### LINKED ISSUE #10473: [1847] Saar share count wrong
**What's your game ID?

-->https://18xx.games/game/155636

**How to Reproduce the Bug?**

-->During Stock round 7, players Zerkan and Trainsinvestite each owned 50% of Saar with player Zerkan owning the president share and the 20% share. If player Zerkan sells more than 1 share without selling the double share, the share count goes up to 12. If Zerkan sells 2 single shares than Trainsinvestite will own 50% and presidency, Zerkan will own 50%, and 2 shares in the market. If Zerkan sells 3 shares Trainsinvestite will own 50%, Zerkan will own 40%, and 3 shares in the market. 

**Which Browser / Platform?**

Chrome, Mobile

Screenshot_20240313_120022_Chrome

Screenshot_20240313_120012_Chrome

Screenshot_20240313_115955_Chrome

Galatolol:
Fixes #10473 

## Before clicking "Create"

### Explanation of Change

Shares exchanged for president's certificate were incorrectly going to the previous president instead of the Market after the sale of the president's certificate.

**This requires pinning.**


ollybh:
Does this fix the part of issue #10473 where the share counts for corporations go above 10? That's generally caused by moving a share without updating the `percent_of` value for that corporation. `change_president` shouldn't result in anyone's percentage holding changing, so this might be happening somewhere else.

Galatolol:
Yes, it does fix that.

image

If I change back to `swap_to`:

image

ollybh (approved):

PR #10617, opened 2024-04-09
Adds instructions for scrubbing passwords to import_game info.

philcampeau:
## Before clicking "Create"

### Explanation of Change

~~Added details about how to reset passwords in the local dev environment to the section on importing games, since that's most likely when you're going to want that information.~~

Apparently I misunderstood something about the import script, and this is already done automatically. I modified the DEVELOPMENT.rb file instead to just explicitly say that the passwords are scrubbed and reset to "password".


roseundy:
Doesn't import_game do this already? It used to.

philcampeau:
> Doesn't import_game do this already? It used to.

PR #10688, opened 2024-04-29
[18 Hiawatha] signal_end_game text and fixes Liquidation section of the market

### LINKED ISSUE #10687: [18 Hiawatha] "Signal End Game" typo
In-game, the Info page says:

Signal End Game       Game Ends 3 ORs after purchase/export of first 8 train

in the events section.  It should presumably say something different, since there are no 8-trains.

### LINKED ISSUE #10689: [18Hiawatha] Corps in Liquidation can decline to be sold
Game hasn't reached the bug yet, but it's on its way there: 
https://18xx.games/game/160858

Gist link to HotSeat copy, which is 160858 but played through until the moment of bug: 
https://gist.github.com/yoyoson/715cf347f8559d9e1a05ccc9fe98befd

If a corp is in liquidation, then when it comes up in the Acquisition Round, the options are "Offer for Sale" or "Pass (Offer for Sale)", and Passing works and turn order proceeds to the next company. 

I don't see any rule difference for 18Hiawatha w.r.t. liquidation, but I could be just tired and missed it multiple times.

Note also that in other 1817-family titles, the market's Liquidation space is colored Red and separated from the rest of the market, whereas here it's normal colored and connected.

philcampeau:
Fixes #10687 
Fixes #10689

## Before clicking "Create"

### Explanation of Change
First commit:
forgot to add the changed signal_end_game to EVENTS_TEXT

Second commit:
improved the text from the first commit.

Third commit:
The far left market space wasn't coded as Liquidation as it should be. This will likely break some games. since we're in alpha, they should just be archived.

### DIFF
--- a/lib/engine/game/g_18_hiawatha/game.rb
+++ b/lib/engine/game/g_18_hiawatha/game.rb
@@ -30,7 +30,7 @@ module Engine
         ALWAYS_BUY_TRAINS_AT_FACE_VALUE = true
 
         MARKET = [
-          %w[0
+          %w[0l
              0a
              0a
              0a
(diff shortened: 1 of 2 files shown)


ollybh (approved):

phil-campeau:
@ollybh I just added another commit to this that removes a stray 'print' command. Should still be ok but I figured I'd let you know :)

PR #10703, opened 2024-05-03
[18Uruguay] Change to Alpha

patrikolesen:
{18Ururguay] Change to Alpha

## Before clicking "Create"


patrikolesen:
Found a blocking error with the farms

PR #10705, opened 2024-05-03
[18Norway] Update dividend rules

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/steps/dividend.rb
+++ b/lib/engine/game/g_18_norway/steps/dividend.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/dividend'
+require_relative '../../../step/half_pay'
+require_relative '../../../step/minor_half_pay'
+
+module Engine
+  module Game
+    module G18Norway
+      module Step
+        class Dividend < Engine::Step::Dividend
+          DIVIDEND_TYPES = %i[payout half withhold].freeze
+          include Engine::Step::HalfPay
+
+          def share_price_change(entity, revenue = 0)
+            return {} if entity.minor?
+
+            price = entity.share_price.price
+            return { share_direction: :left, share_times: 1 } if revenue < price / 2
+
+            times = 0
+            times = 1 if revenue >= price
+            times = 2 if revenue >= price * 2
+            times = 3 if revenue >= price * 3 && price >= 165
+            if times.positive?
+              { share_direction: :right, share_times: times }
+            else
+              {}
+            end
+          end
+
+          def skip!
+            super
+
+            return unless current_entity.receivership?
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_norway/steps/dividend.rb
```diff
@@ -0,0 +1,46 @@
...
+      module Step
+        class Dividend < Engine::Step::Dividend
+          DIVIDEND_TYPES = %i[payout half withhold].freeze
+          include Engine::Step::HalfPay
+
+          def share_price_change(entity, revenue = 0)
+            return {} if entity.minor?
+
+            price = entity.share_price.price
+            return { share_direction: :left, share_times: 1 } if revenue < price / 2
```
ollybh, line 19: You need to be careful with division: Opal translates the code to Javascript which only has floating point numbers. It's unlikely that this would be a problem (I assume the revenue will be a multiple of 10, so you're not going to be dealing with fractions), but multiplication is a bit safer than division.
```suggestion
            return { share_direction: :left, share_times: 1 } if revenue * 2 < price
```

lib/engine/game/g_18_norway/steps/dividend.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/dividend'
+require_relative '../../../step/half_pay'
+require_relative '../../../step/minor_half_pay'
```
ollybh, line 5: Are there minor companies in 18Norway?

lib/engine/game/g_18_norway/steps/dividend.rb
```diff
@@ -0,0 +1,46 @@
...
+module Engine
+  module Game
+    module G18Norway
+      module Step
+        class Dividend < Engine::Step::Dividend
+          DIVIDEND_TYPES = %i[payout half withhold].freeze
+          include Engine::Step::HalfPay
+
+          def share_price_change(entity, revenue = 0)
+            return {} if entity.minor?
```
ollybh, line 16: There's another reference to minors here, do these exist?

lib/engine/game/g_18_norway/steps/dividend.rb
patrikolesen: No there is not any minors in 18Norway, did copy a bit to much from 1846
Now removed

lib/engine/game/g_18_norway/steps/dividend.rb
patrikolesen: No there is not any minors in 18Norway, did copy a bit to much from 1846
Now removed

Also removed receivership

ollybh (approved):

PR #10712, opened 2024-05-04
[1844] allow P4 to be used at any time

### LINKED ISSUE #10649: [1844] Player unable to use P4 Furka-Oberalpbahn during a Stock Round
**Link**
https://18xx.games/game/160540?action=299

**Scenario**
It is Phase 3. It is Alex's turn in the Stock Round. Alex owns the P4 Furka-Oberalpbahn
Expected: Alex may activate the P4 for 80f at any time.
Observed: There is no way to activate the P4 during a Stock Round.

**Relevant Rules Text**
http://www.ohley.de/english/1844_Rules_14-01-04.pdf
1st Edition rules. Page 5: Properties
> The owner of the P4 (FO - Furka Oberalpbahn) can at any time after the sale of the first 3/3H locomotive -- including during a share dealing round -- build all 4 special track tiles of the FO at once on the hexes reserved for them. Building them gives a one-off payment of 80 sfr to the owner (player or company) and closes the private railway company. The route must be built the moment the first 5/5H locomotive is bought; the owner still receives the 80 sfr as payment.

https://lookout-spiele.de/upload/en_1844_1854.html_Rules_1844_EN.pdf
Lookout Rules. Page 19: P4
2nd Edition is less explicit, but consistent.
> Special ability: The owner of this Private Railway company may at any time place the Furka-Oberalp
special tile. This is an additional tile lay and free of cost.
Doing this closes the P4; its owner receives 80 SFR as a compensation. This happens at the latest by the
sale of the first 5/5H train (begin of Phase 5). Its owner can’t waive that build.

benjaminxscott:
Fixes #10649

## Before clicking "Create"

### Explanation of Change

Adds the existing step that has special handling for P4 to the SR . 

Changes P4 to be used at 'any' time

This doesn't affect P7 as it has a 'when' limiter for specific steps

I don't believe this requires pins as it enables an action that was previously not possible

### Screenshots
in an SR on a player's turn who doesn't own the P4:

image

image

### DIFF
--- a/lib/engine/game/g_1844/entities.rb
+++ b/lib/engine/game/g_1844/entities.rb
@@ -60,7 +60,7 @@ module Engine
               {
                 type: 'choose_ability',
                 after_phase: '2',
-                when: %w[track owning_player_track],
+                when: 'any',
                 choices: ['Place tile'],
               },
             ],
@@ -99,7 +99,7 @@ module Engine
               {
                 type: 'choose_ability',
                 after_phase: '4',
-                when: 'owning_player_or_turn',
+                when: 'any',
                 choices: [], # Defined in special_choose step
               },
             ],
--- a/lib/engine/game/g_1844/game.rb
+++ b/lib/engine/game/g_1844/game.rb
@@ -616,6 +616,7 @@ module Engine
         def stock_round
           Engine::Round::Stock.new(self, [
             G1844::Step::MountainRailwayTrack,
+            G1844::Step::SpecialChoose,
             G1844::Step::BuySellParShares,
           ])
         end


ollybh:
@benjaminxscott Have you tested and confirmed that this isn't giving the player the option of using P7’s ability during the stock round. As far as I can see, the code in the SpecialChoose step isn't testing the choose ability’s `when` value.

ollybh:
I think this change will require running the migrate_game script, to add some pass actions. There are probably going to be times where a player's stock round turn was skipped as they did not have any valid stock actions, but are now going to be asked whether they want to use P4’s ability.

benjaminxscott:
> require migration
Ah you're right

> ensure P7 is working as intended

Good point, I had implicitly tested this by making sure P7 didn't show up as an option in the above screenshots, but those were from phase 3 so P7 wouldn't show up anyway. P4 and P7 never show up at the same time fwiw since P4 is immediately forced to be laid on the 5 and P7 isn't usable until phase 5

Interestingly, my reading is that P7 *should* show up on an `when: any` since its desc says "They may assign it immediately or later" instead of the current `president_or_turn` . I imagine the latter was set just to avoid it showing up when it wasn't actually relevant

> when is `when:` checked?

I believe `when` gets checked here: (in P7's case, `on_phase` is also a relevant check)
https://github.com/tobymao/18xx/blob/master/lib/engine/game/base.rb#L2929

That gets called from the generic special step
https://github.com/tobymao/18xx/blob/master/lib/engine/step/special_choose.rb#L29

Which is inherited by the 1844 special step

lib/engine/game/g_1844/entities.rb
```diff
@@ -99,7 +99,7 @@ module Entities
               {
                 type: 'choose_ability',
                 after_phase: '4',
-                when: 'owning_player_or_turn',
+                when: 'any',
```
crericha, line 102: This change is unnecessary. Since you can only assign it to corporations you are a director of, limiting to their OR turns is sufficient. It reduces how much the ability has to show up without any play impact.

lib/engine/game/g_1844/entities.rb
```diff
@@ -60,7 +60,7 @@ module Entities
               {
                 type: 'choose_ability',
                 after_phase: '2',
-                when: %w[track owning_player_track],
+                when: 'any',
```
crericha, line 63: I'd like to still limit this ability as I feel `any` is obnoxious to deal with and we can limit to a short list that does not affect game play.

crericha:
Thanks for taking a stab at this one @benjaminxscott. After thinking about it a little, there are some additional, related changes I want to make and test. Rather than trying to work through you, I'm going to open my own PR.

lib/engine/game/g_1844/entities.rb
benjaminxscott: Agree, I was mulling whether it was better to implement it 'as written' or for better playability

PR #10717, opened 2024-05-05
[18 India] [Core] Create Multiple Selection Action and Refactor Certificate Selection

EdPovi:

## Implementation Notes
- Created a `SelectMultipleCompanies` action that stores an array of companies as part of a single action.
- Modified `View::Company` and `View::Round::Auction` behavior to support selection of multiple companies.
- Refactor Certificate Selection step to use multiple selection action. 

### Explanation of Change
The old version of the certificate selection step was very "clicky", often needing 20 - 30 clicks to complete and felt "clunky" as a user experience.  This refactor allows the completion of the step with 1/3 of the clicks, and without hiding the screen after individual certificate selection.
 
### Screenshots
image

### Any Assumptions / Hacks
- Uses the `click` -> `select_company` event on company card as the "button" to select and unselect the company.

### DIFF
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -22,6 +22,7 @@ module View
 
       def selected?
         return @step.company_selected?(@company) if @step.respond_to?(:company_selected?)
+
         @company == @selected_company
       end
 
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -36,7 +36,7 @@ module View
           return process_action(Engine::Action::Assign.new(entity, target: selected_company))
         end
 
-        if @game.round.actions_for(entity).include?('select_multiple_companies') && @step.respond_to?(:select_company)
+        if @game.round.actions_for(entity).include?('select_multiple_companies')
           @step.select_company(entity, @company)
           return store(:selected_company, nil)
         end
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -21,6 +21,7 @@ module View
       needs :interactive, default: true
 
       def selected?
+        return @step.company_selected?(@company) if @step.respond_to?(:company_selected?)
         @company == @selected_company
       end
 
@@ -32,6 +33,11 @@ module View
         if selected_company && @game.round.actions_for(entity).include?('assign') &&
           (@game.class::ALL_COMPANIES_ASSIGNABLE || entity.respond_to?(:assign!))
           return process_action(Engine::Action::Assign.new(entity, target: selected_company))
+        end
+
+        if @game.round.actions_for(entity).include?('select_multiple_companies') && @step.respond_to?(:select_company)
+          @step.select_company(entity, @company)
+          return store(:selected_company, nil)
         end
 
         store(:tile_selector, nil, skip: true)
@@ -81,6 +87,7 @@ module View
       end
 
       def render
+        @step = @game.round.active_step
         # use alternate view of corporation if needed
         if @game.respond_to?(:company_view) && (view = @game.company_view(@company))
           return send("render_#{view}")
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -124,7 +124,7 @@ module View
           select_button = [h(:button, { on: { click: mutiple_companies } }, 'Confirm Selections')]
 
           children = []
-          children << h(:div, @step.selection_note.map { |text_block| h(:p, text_block) } )
+          children << h(:div, @step.selection_note.map { |text_block| h(:p, text_block) })
           children << h(:div, select_button) if @step.selections_completed?
           h(:div, children)
         end
--- a/assets/app/view/game/round/auction.rb
+++ b/assets/app/view/game/round/auction.rb
@@ -42,6 +42,7 @@ module View
             h(:div, [
               render_turn_bid,
               render_show_button,
+              render_confirm_selections,
               *render_companies,
               *render_minors,
               *render_corporations,
@@ -109,6 +110,28 @@ module View
           h(:button, props, "#{hidden? ? 'Show' : 'Hide'} #{@step.visible? ? 'Player' : 'Companies'}")
         end
 
+        def render_confirm_selections
+          return nil if hidden? && !@step.visible?
+          return nil unless @current_actions.include?('select_multiple_companies')
+
+          mutiple_companies = lambda do
+            hide!
+            process_action(Engine::Action::SelectMultipleCompanies.new(
+              @current_entity,
+              companies: @step.selected_companies
+            ))
+          end
+          select_button = [h(:button, { on: { click: mutiple_companies } }, 'Confirm Selections')]
+
+          children = []
+          if @step.respond_to?(:selection_note)
+            note = @step.selection_note.map { |text_block| h(:p, text_block) }
+            children << h(:div, note)
+          end
+          children << h(:div, select_button) if @step.selections_completed?
+          h(:div, children)
+        end
+
         def render_companies
           return [] if hidden? && !@step.visible?
           return [] if !@current_actions.include?('bid') &&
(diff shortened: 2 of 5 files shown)


assets/app/view/game/company.rb
```diff
@@ -34,6 +36,11 @@ def select_company(event)
           return process_action(Engine::Action::Assign.new(entity, target: selected_company))
         end
 
+        if @game.round.actions_for(entity).include?('select_multiple_companies') && @step.respond_to?(:select_company)
```
ollybh, line 39: You don't need the `respond_to?` check here, any game that wants to have a `select_multiple_companies` action will have to implement this method.
```suggestion
        if @game.round.actions_for(entity).include?('select_multiple_companies')
```

assets/app/view/game/round/auction.rb
```diff
@@ -109,6 +110,28 @@ def render_show_button
...
+              companies: @step.selected_companies
+            ))
+          end
+          select_button = [h(:button, { on: { click: mutiple_companies } }, 'Confirm Selections')]
+
+          children = []
+          if @step.respond_to?(:selection_note)
+            note = @step.selection_note.map { |text_block| h(:p, text_block) }
+            children << h(:div, note)
+          end
```
ollybh, line 130: Again, you don't need to include the `respond_to?` check.

lib/engine/game/g_18_india/step/certificate_selection.rb
```diff
@@ -11,13 +11,18 @@ module Game
...
+          def actions(entity)
+            return [] unless entity == current_entity
+            return [] if finished?
+
+            ACTIONS
+          end
 
           def setup
-            @companies = @game.companies
             @choices = Hash.new { |h, k| h[k] = [] }
```
ollybh, line 26: You're setting up a hash whose keys are players and values are arrays of companies. Will this ever have any more than a single player in the hash? Or will it just be the current player?
If it's just the current player, then this can be changed to be a simple array.
Or you could have it as a hash whose values are companies and values are booleans.

lib/engine/game/g_18_india/step/certificate_selection.rb
```diff
@@ -35,10 +40,8 @@ def may_purchase?(_company)
             false
           end
 
-          def may_choose?(company)
-            return false if selections_completed? && company.owner.nil?
-
-            true
+          def may_choose?(_company)
+            false
           end
```
ollybh, line 45: Is this method still needed?

lib/engine/game/g_18_india/step/certificate_selection.rb
```diff
@@ -55,57 +58,75 @@ def players_visible?
...
+          def select_company(player, company)
+            return if selections_completed? && company.owner.nil? # prevent selecting more than allowed
+
+            # add or remove from choices
+            if company_selected?(company)
+              @choices[player].delete(company)
+              company.owner = nil
+            else
+              @choices[player] << company
+              @choices[player].sort!
```
ollybh, line 93: Does it matter whether this array is sorted?

lib/engine/game/g_18_india/step/certificate_selection.rb
```diff
@@ -55,57 +58,75 @@ def players_visible?
...
 
-          def selections_completed?
-            number_of_selections == @cards_to_keep
+          def select_company(player, company)
+            return if selections_completed? && company.owner.nil? # prevent selecting more than allowed
+
+            # add or remove from choices
+            if company_selected?(company)
+              @choices[player].delete(company)
+              company.owner = nil
```
ollybh, line 90: You're setting company ownership in `process_select_multiple_companies`. Does this need to be done here as well?

lib/engine/game/g_18_india/step/certificate_selection.rb
```diff
@@ -55,57 +58,75 @@ def players_visible?
...
+            [
+              "Made #{number_of_selections} of #{@cards_to_keep} selections.",
+              'Click on card to select or unselect it.',
+            ]
           end
 
-          def selections_completed?
-            number_of_selections == @cards_to_keep
+          def select_company(player, company)
+            return if selections_completed? && company.owner.nil? # prevent selecting more than allowed
```
ollybh, line 85: It might be a better user interface to allow a player to select more items than they are allowed to keep, but not allow them to finish their selection until they have unselected other items. Otherwise it might not be obvious why nothing is happening when they are clicking on a share.

lib/engine/game/g_18_india/step/certificate_selection.rb
```diff
@@ -55,57 +58,75 @@ def players_visible?
...
+          def selections_completed?
+            number_of_selections == @cards_to_keep
           end
 
-          def process_bid(action)
-            choose_company(action.entity, action.company)
-            @game.next_turn!
+          def process_select_multiple_companies(action)
+            player = action.entity
+            selected_companies = action.companies
```
ollybh, line 116: You could be paranoid here and check that all of `selected_companies` are in `player.hand` and throw an error if not.

ollybh (commented):
This looks like a good idea, much better than selecting a single card at a time.

lib/engine/game/g_18_india/step/certificate_selection.rb
EdPovi: It's a bit nicer to see the companies in order on the player card.  You can see this in the screenshot above for player 5, which has the selections also listed on the player card...

assets/app/view/game/round/auction.rb
EdPovi: The `respond_to?` check is useful here that a `selection_note` can be an optional feature.

lib/engine/game/g_18_india/step/certificate_selection.rb
EdPovi: This was true before to create a "Choose" button.  I set it to false to prevent the button.  But with the other changes to company view, this works without the method.  I have removed it.

lib/engine/game/g_18_india/step/certificate_selection.rb
EdPovi: It doesn't NEED to be done here as well. Originally, changing ownership was the only visual cue that the player has selected that company card.  With this updates, there are three indications: the blue highlight, the owner change, and the list of companies on the player card.  My take was the extra visual cue didn't hurt, so I left it there.  The "real" change in ownership happens in `process_select_multiple_companies`, as that is the only method run when the game is loading.

lib/engine/game/g_18_india/step/certificate_selection.rb
EdPovi: Thanks for the suggestion, I will re-work it to allow selecting too many, but also change the NOTE to warn that too many are selected.

lib/engine/game/g_18_india/step/certificate_selection.rb
EdPovi: Added a check as a little paranoia can be healthy...

lib/engine/game/g_18_india/step/certificate_selection.rb
EdPovi: removed check to allow selecting past limit.

assets/app/view/game/round/auction.rb
ollybh: It's not an optional feature at the moment, as 18India will be the only game that has the multiple selection and it uses the selection note. If another game reuses multiple selection and doesn't want a note shown it could be made optional at that time. (And if it were to be optional I'd prefer that this would be done by `selection_note` returning nil or an empty array).

EdPovi: That makes sense.  I do have a tendency to overthink future possibilities at times.  

I have removed the `respond_to?` check and simplified the note section.

ollybh (approved):
Looks good to me.

PR #10722, opened 2024-05-06
[18RG] fix ability keyword

### LINKED ISSUE #10695: [18RoyalGorge] unknown keyword :closed_when_used_up
**How to Reproduce the Bug?**

**Which Browser / Platform?**

benjaminxscott:
Fixes #10695

## Before clicking "Create"

### Explanation of Change

`remove_when_used_up` is used instead of `closed_when_used_up` in the context of a `choose_ability`

`choose_ability` isn't documented in the Abilities readme, so I assume it's a newer feature that seems used for adding boutique abilities that aren't tied to a private
https://github.com/tobymao/18xx/blob/master/lib/engine/ability/README.md

I don't think this needs pins, as games were stuck on that step anyway (also it's alpha)

adds fixture

### Screenshots

Screenshot 2024-05-06 9 04 01 AM

fwiw I also tested that the doc holiday debt ability worked as expected, and it seemed fine

https://gist.github.com/benjaminxscott/549b21902bf4da3da52707a4596ca0de

### DIFF
--- a/lib/engine/game/g_18_royal_gorge/game.rb
+++ b/lib/engine/game/g_18_royal_gorge/game.rb
@@ -390,7 +390,7 @@ module Engine
             choices: { 'pay_debt' => 'Pay Doc Holliday for one DEBT token' },
             count: 2,
             count_per_or: 1,
-            closed_when_used_up: true,
+            remove_when_used_up: true,
           )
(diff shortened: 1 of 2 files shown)


michaeljb:
Thanks!

michaeljb (approved):

PR #10737, opened 2024-05-10
[1849] Implements the official variants from the rulebook

philcampeau:
## Before clicking "Create"

## Implementation Notes

Implements the 3 official variant rules for 1849: Acquiring Station Tokens, Bonds, and Electric Dreams, plus the reduced corporation for 4 players mentioned at the bottom of page 2. The three main variants are found on page 16.

### Explanation of Change

I've numerous test games, hopefully weeded out all the bugs already. 

I tested the code against well over 30 completed games on the site to ensure that the changes here don't negatively impact existing games. 

I made minor changes to /assets/app/view/game/corporation.rb and /assets/app/view/game/loans.rb for display purposes to this game. It uses a loan system, but refers to the loans as Bonds. I incorporated some code to the two files mentioned to allow for buttons to display differently.

I had to modify the game.rb page for 1849 - Kingdom of Two Sicilies. Since I changed the `TRAINS` constant in 1849 to a `def game_trains` method, it was superseding the `TRAINS` constant in 1849K2S, so I modified K2S to `def game_trains` as well.

I listed all the variants as (alpha) on the meta page to set expectations. 
### Any Assumptions / Hacks
EDIT: The text below is no longer relevant, but I wanted to leave it visible. The code referenced below has been removed through an improvement in my approach.

I want to change the `check_for_cheapest_train(train)` method in lib/engine/game/step/buy_trains.rb to also accept an `entity` argument, but since this will require changing code in a few games, I will wait until this has been merged in, and then create a separate request for all that. As such, there's some commented placeholder code in this game's buy_trains.rb doing the work for now.


assets/app/view/game/corporation.rb
```diff
@@ -638,9 +638,10 @@ def render_loans
           loan_props[:style][:color] = contrast_on(color)
         end
 
+        corp_loans_text = @corporation.respond_to?(:corp_loans_text) ? @corporation.corp_loans_text : 'Loans'
```
crericha, line 641: Why is the text in the corporation? Shouldn't it be derived from the step?

assets/app/view/game/loans.rb
```diff
@@ -12,8 +12,10 @@ class Loans < Snabberb::Component
       def render
         actions = @game.round.actions_for(@corporation)
         children = []
+        step = @game.round.active_step
 
         if actions.include?('payoff_loan')
+          payoff_loan_button_text = step.respond_to?(:payoff_loan_button_text) ? step.payoff_loan_button_text : 'Payoff Loan'
```
crericha, line 18: I think we should remove the `button` from the attribute/method. The step doesn't care how its presented..only what the text should be.

assets/app/view/game/loans.rb
```diff
@@ -22,18 +24,19 @@ def render
                 end,
               },
             },
-            'Payoff Loan',)
+            payoff_loan_button_text,)
         end
 
         if actions.include?('take_loan')
+          take_loan_button_text = step.respond_to?(:take_loan_button_text) ? step.take_loan_button_text : 'Take Loan'
```
crericha, line 31: Same comment about `button`.

lib/engine/game/g_1849/corporation.rb
```diff
@@ -5,7 +5,7 @@ module Game
     module G1849
       class Corporation < Engine::Corporation
         attr_reader :token_fee
-        attr_accessor :next_to_par, :closed_recently, :slot_open, :reached_max_value, :sms_hexes
+        attr_accessor :next_to_par, :closed_recently, :slot_open, :reached_max_value, :sms_hexes, :e_token
```
crericha, line 8: Should e_token be an ability? That way other players can see the corporation has it.

lib/engine/game/g_1849/depot.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative '../../depot'
+
+module Engine
+  module Game
+    module G1849
+      class Depot < Engine::Depot
+        def min_depot_train_no_e_token
```
crericha, line 9: This should be done in `buy_train` step. Both 18Neb and 1844 has examples of filtering trains by corporation type.

lib/engine/game/g_1849/game.rb
```diff
@@ -29,6 +32,11 @@ class Game < Game::Base
         BANK_CASH = 7760
 
         CERT_LIMIT = { 3 => 12, 4 => 11, 5 => 9 }.freeze
+        REDUCED_CERT_LIMIT = { 3 => 12, 4 => 9, 5 => 9 }.freeze
+
+        def game_cert_limit
+          reduced_4p_corps? ? self.class::REDUCED_CERT_LIMIT : self.class::CERT_LIMIT
```
crericha, line 38: In this case, I'd avoid the duplication of 3 and 5 player values and instead do something like
```
limit = super.dup
limit[4] = 9 if reduced_4p_corps?
limit
```

lib/engine/game/g_1849/game.rb
```diff
@@ -83,43 +91,53 @@ class Game < Game::Base
...
-                    name: '12H',
-                    num: 1,
-                    distance: 12,
-                    price: 800,
-                    events: [{ 'type' => 'close_companies' }, { 'type' => 'earthquake' }],
-                  },
-                  { name: '16H', distance: 16, price: 1100 },
-                  { name: 'R6H', num: 2, available_on: '16H', distance: 6, price: 350 }].freeze
+        def game_trains
+          train_list = [
```
crericha, line 107: leave the trains as they were and call `super` here to get the trains before modifying.

lib/engine/game/g_1849/game.rb
```diff
@@ -83,43 +91,53 @@ class Game < Game::Base
...
+              price: 800,
+              events: [{ 'type' => 'close_companies' }, { 'type' => 'earthquake' }],
+            },
+            { name: '16H', distance: 16, price: 1100 },
+            { name: 'E', num: 6, available_on: '12H', distance: 99, price: 550 },
+            { name: 'R6H', num: 2, available_on: '16H', distance: 6, price: 350 },
+          ]
+          train_list.reject! { |t| t[:name] == 'E' } unless electric_dreams?
+          train_list.find { |t| t[:name] == '6H' }[:events] << { 'type' => 'buy_tokens' } if acquiring_station_tokens?
+          train_list.find { |t| t[:name] == '8H' }[:events] << { 'type' => 'bonds' } if bonds?
```
crericha, line 137: If doesn't make sense to make bonds both an event and a status. Pick whichever you feel fits best.

lib/engine/game/g_1849/game.rb
```diff
@@ -436,6 +466,12 @@ def close_corporation(corporation, quiet: false)
           end
           corporation.next_to_par = true if @corporations[index - 1].floated?
           update_garibaldi
+          # code below is for variant rules
+          @log << 'E-token returned to bank' if corporation.e_token
+          corporation.e_token = false if electric_dreams?
```
crericha, line 471: Doesn't the bank's e-token count need to be increased?

crericha: Also don't need to gate by `if electric_dreams?` here because the attribute always exists on the corporation.

lib/engine/game/g_1849/game.rb
```diff
@@ -436,6 +466,12 @@ def close_corporation(corporation, quiet: false)
...
+
+          repaid_bond = corporation.loans.pop if bonds? && corporation.loans.any?
```
crericha, line 473: Here also, don't need to check bonds. If the corporation has loans, that tells us bonds are enabled. `corporation.loans.clear()`.

lib/engine/game/g_1849/game.rb
```diff
@@ -436,6 +466,12 @@ def close_corporation(corporation, quiet: false)
...
+          @loans << repaid_bond if repaid_bond
```
crericha, line 474: Why keep track of loans in the game class? Each corp gets one bond, so they don't need to be tracked except within the corporation.

lib/engine/game/g_1849/game.rb
```diff
@@ -480,9 +517,13 @@ def operating_round(round_num)
...
             G1849::Step::Dividend,
+            G1849::Step::BuyToken,
+            G1849::Step::BuyEToken,
             Engine::Step::DiscardTrain,
             G1849::Step::BuyTrain,
             G1849::Step::IssueShares,
+            G1849::Step::BondInterestPayment,
+            G1849::Step::Bond,
             [Engine::Step::BuyCompany, { blocks: true }],
           ], round_num: round_num)
```
crericha, line 528: Should we remove the unused steps here? At a minimum we shouldn't show in the log "skips bond" etc. if not enabled.

lib/engine/game/g_1849/game.rb
```diff
@@ -532,13 +573,25 @@ def check_distance(route, _visits)
...
           return unless (route.stops.map(&:hex).map(&:id) & PORT_HEXES).any?
 
           raise GameError, 'Route must include two non-port stops.' unless route.stops.size > 2
         end
 
+        def check_track_type(route)
+          train = route.train
+          paths = route.paths
+
+          raise GameError, 'E-Trains cannot use narrow gauge' if paths.any? { |p| p.track == :narrow } && train.name == 'E'
```
crericha, line 586: For performance, check the train name first, so we can stop checking if not an E train.

lib/engine/game/g_1849/game.rb
```diff
@@ -532,13 +573,25 @@ def check_distance(route, _visits)
         end
 
         def check_other(route)
+          check_track_type(route)
           return unless (route.stops.map(&:hex).map(&:id) & PORT_HEXES).any?
 
           raise GameError, 'Route must include two non-port stops.' unless route.stops.size > 2
         end
 
+        def check_track_type(route)
```
crericha, line 582: Let's come up with a more descriptive method name. Something like `check_e_train_track`.

lib/engine/game/g_1849/game.rb
```diff
@@ -547,9 +600,9 @@ def cnm_bonus(corp, stops)
         end
 
         def stop_revenue(stop, phase, train)
-          return gray_revenue(stop) if GRAY_REVENUE_CENTERS.key?(stop.hex.id)
-
-          stop.route_revenue(phase, train)
+          revenue = gray_revenue(stop) if GRAY_REVENUE_CENTERS.key?(stop.hex.id)
+          revenue ||= stop.route_revenue(phase, train)
+          revenue
```
crericha, line 605: Why was this method altered?

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+        def event_bonds!
+          @log << "-- Event: #{EVENTS_TEXT[:bonds][1]} --"
+        end
+
+        def event_e_tokens!
+          @log << "-- Event: #{EVENTS_TEXT[:e_tokens][1]} --"
+        end
+
+        # code below is for Bonds variant
+        NUM_LOANS = 6
```
crericha, line 867: not needed?

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+        MAXIMUM_LOANS = 1
+        LOAN_VALUE = 500
+        INTEREST_RATE = 50
+
+        def init_loans
+          Array.new(num_loans) { |id| Loan.new(id, loan_value) }
+        end
+
+        def num_loans
+          bonds? ? NUM_LOANS : 0
```
crericha, line 877: Always return 1 if you really need this.

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+        end
+
+        def maximum_loans(_entity)
+          MAXIMUM_LOANS
```
crericha, line 881: self.class::MAXIUM_LOANS

Applies to other constants too.

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+        def loan_value(_entity = nil)
+          LOAN_VALUE
+        end
+
+        def interest_rate
+          INTEREST_RATE
+        end
+
+        def interest_owed_for_loans(loans)
+          interest_rate * loans
```
crericha, line 893: Its just `INTEREST_RATE`. Do you need the calculation?

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+          @log << "#{entity.name} pays #{amount_fmt} interest for its issued bond"
+        end
+
+        def take_loan(entity, loan)
+          raise GameError, 'Cannot issue bond' unless can_take_loan?(entity)
+
+          name = entity.name
+          @log << "#{name} issues its bond and receives #{format_currency(loan_value)}"
+          @bank.spend(loan_value, entity)
+          entity.loans << loan
```
crericha, line 915: Just create a loan here. No need to have an array of them.

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+
+        # code below is for the Electric Dreams variant
+
+        def status_array(corporation)
+          return unless corporation.e_token
+
+          ['E-Token Purchased']
+        end
+
+        def set_e_token_sold
```
crericha, line 945: Make this an attribute instead of a method.

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+        def can_pay_interest?(entity, extra_cash = 0)
+          entity.cash + extra_cash >= interest_owed(entity)
+        end
+
+        # code below is for the Electric Dreams variant
+
+        def status_array(corporation)
+          return unless corporation.e_token
+
+          ['E-Token Purchased']
```
crericha, line 942: Maybe just `E-Token`.

crericha: I traditionally use abilities for tracking things like this. Is the status array clearer?

lib/engine/game/g_1849/game.rb
```diff
@@ -771,12 +824,137 @@ def bank_sort(corporations)
...
+        end
+
+        def init_train_handler
+          trains = game_trains.flat_map do |train|
+            Array.new((train[:num] || num_trains(train))) do |index|
+              Train.new(**train, index: index)
+            end
+          end
+
+          electric_dreams? ? G1849::Depot.new(trains, self) : self.class::DEPOT_CLASS.new(trains, self)
```
crericha, line 956: All the logic seems to be handled in 1849 depot for both cases, so why check here? Instead use the 49 depot always?

lib/engine/game/g_1849/game.rb
crericha: If you stick with statuses, make them only show up conditionally if that variant it works with is enabled.

lib/engine/game/g_1849/meta.rb
```diff
@@ -24,6 +24,27 @@ module Meta
             short_name: 'Delay IFT',
             desc: 'IFT may not be one of the first three corporations (recommended for newer players)',
           },
+          {
+            sym: :reduced_4p_corps,
+            short_name: '(Alpha) - Reduced corps at 4 players',
```
crericha, line 29: `()` and `-` are redundant. Suggest removing the `-` in all of these.

lib/engine/game/g_1849/round/operating.rb
```diff
@@ -7,11 +7,24 @@ module Game
...
+          attr_reader :issued_bond, :repaid_bond
+
           def next_entity!
             return @game.end_game! if @entities[@entity_index].reached_max_value
 
             super
           end
+
+          def after_setup
+            @repaid_bond = {}
```
crericha, line 19: why `after_setup` instead of in the constructor?

lib/engine/game/g_1849_boot/game.rb
```diff
@@ -86,31 +86,35 @@ class Game < G1849::Game
...
-                  {
-                    name: '12H',
-                    num: 1,
-                    distance: 12,
-                    price: 800,
-                    events: [{ 'type' => 'close_companies' }, { 'type' => 'earthquake' }],
-                  },
-                  { name: '16H', distance: 16, price: 1100 },
-                  { name: 'R6H', num: 2, available_on: '16H', distance: 6, price: 600 }].freeze
+        def game_trains
```
crericha, line 89: hopefully the changes to 1849 boot can be reverted by my other suggestions.

lib/engine/game/g_1849/corporation.rb
philcampeau: I enabled the display on corp charters with this code in game.rb: 

https://github.com/philcampeau/18xx/blob/b5487960d95197a18c153d5245c9d3ac97a47775/lib/engine/game/g_1849/game.rb#L939

```     
def status_array(corporation)
   return unless corporation.e_token
   ['E-Token Purchased']
end
```

lib/engine/game/g_1849/depot.rb
philcampeau: I can do it the way it's done in 1844 if that's preferred. This is the way it was done in 1848, which is where I borrowed the code from.

lib/engine/game/g_1849/game.rb
philcampeau: Not sure. Changed it back.

lib/engine/game/g_1849/game.rb
philcampeau: Changed to check_e_train_track_type. A little wordy, but clearer.

lib/engine/game/g_1849/game.rb
philcampeau: Good call, switched.

lib/engine/game/g_1849/game.rb
philcampeau: done.

lib/engine/game/g_1849/game.rb
philcampeau: Done, except for this one, which doesn't seem to allow that format without crashing my fixture games:

```
        def num_loans
          bonds? ? NUM_LOANS : 0
        end
```

is that expected?

assets/app/view/game/corporation.rb
philcampeau: I just thought it made sense to put it in the game's corporation file since this was also a corporation file.
I can change it.

philcampeau:
Hey just a heads up that I haven't abandoned this. My laptop had to be replaced and I haven't had the chance to reconfigure my dev environment. I'll get to these fixes soon!

lib/engine/game/g_1849/game.rb
philcampeau: Looks like I had that idea too, the attr was already there, just forgot to take this out and remove the call from setup. Done.

lib/engine/game/g_1849/meta.rb
philcampeau: Done.

assets/app/view/game/corporation.rb
philcampeau: It's possible I'm misunderstanding, but since this info was already in /assets/app/view/game/corporation.rb, doesn't it make more sense to do this in the Corporation?

lib/engine/game/g_1849/game.rb
philcampeau: commit `add self.class to all definitions`

lib/engine/game/g_1849/meta.rb
philcampeau: commit name: `meta fix`

lib/engine/game/g_1849/game.rb
philcampeau: commit name: `remove set_e_token_sold`

lib/engine/game/g_1849/game.rb
philcampeau: removed statuses, kept events. Was able to use existing statuses for checks.

Commit name: `eliminate unnecessary statuses`

lib/engine/game/g_1849/game.rb
philcampeau: removed the `if electric_dreams?` gate. 

The bank doesn't actually have e-tokens, it's just a true/false that I set up. The log is just using the terminology used by the rules.

Commit name: `remove unnecessary electric_dreams? gate`

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+
+            actions
+          end
+
+          def description
+            can_payoff_loan?(current_entity) ? 'Repay Bond' : 'Issue Bond'
+          end
+
+          def log_skip(entity)
+            super if @game.bonds? && !@game.phase.status.include?('gray_uses_white')
```
crericha, line 26: Why did you decide to gate by status instead of something like `@game.phase.previous.include? { |p| p.name == "6H" }` or an event or status that says bonds are available?

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+          end
+
+          def take_loan_button_text
```
crericha, line 29: Do the other steps that override text include the word button? Generally, its best to not assume how the view will display the info and use a name like `take_loan_text`. Same applys to payoff loan.

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+            "Issue Bond (#{@game.format_currency(@game.loan_value)})"
+          end
+
+          def payoff_loan_button_text
+            "Repay Bond (#{@game.format_currency(@game.loan_value)})"
+          end
+
+          def can_payoff_loan?(entity)
+            !@round.issued_bond[entity] &&
+              entity.loans.any? &&
```
crericha, line 39: Never use `.any?` unless you need to pass in a block with it. Use `!empty?`. It is significantly more efficient.

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+              entity.cash >= entity.loans.first.amount
+          end
+
+          def process_take_loan(action)
+            raise GameError, 'Cannot issue bond' unless @game.can_take_loan?(action.entity)
+
+            @game.take_loan(action.entity, action.loan)
```
crericha, line 46: This is okay, but imo would have been better to ignore the loan in the action and dynamically create the loan during `take_loan`.

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+          end
+
+          def process_payoff_loan(action)
+            entity = action.entity
+            loan = action.loan
+            amount = loan.amount
+            raise GameError, "Bond doesn't belong to that entity" unless entity.loans.include?(loan)
```
crericha, line 53: Probably better to check `can_payoff_loan?` than this as the cod ein the method is assuming the entity has the cash to pay off the loan etc.

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+
+            @log << "#{entity.name} repays its outstanding bond for #{@game.format_currency(amount)}"
+            entity.spend(amount, @game.bank)
+
+            entity.loans.delete(loan)
+            @game.loans << loan
```
crericha, line 59: I recommend getting rid of having a loans container in game. Just create them on the fly as needed. There isn't a loan count in 49 like in 17.

lib/engine/game/g_1849/round/operating.rb
```diff
@@ -7,11 +7,24 @@ module Game
     module G1849
       module Round
         class Operating < Engine::Round::Operating
+          attr_reader :issued_bond, :repaid_bond
```
crericha, line 10: Don't explicitly add this. Use `def round_state` in the bond step instead.

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G1849
+      module Step
+        class BuyEToken < Engine::Step::Base
```
crericha, line 9: I expected this to be part of buy_train, not a separate step. Why is a separate step needed?

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
...
+          def actions(entity)
+            actions = super.dup
```
crericha, line 11: What actions are you duping? You aren't inheriting from a step, just the base class.

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+
+          def auto_actions(entity)
+            return [Engine::Action::Pass.new(entity)] unless can_buy_e_token?(entity)
+
+            super
+          end
+
+          def e_token_cost
+            if @game.e_token_sold == true
```
crericha, line 46: Don't need to compare to true. Its a boolean already.

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
...
+            choices << ["Buy E-Token for #{@game.format_currency(e_token_cost)}"] if can_buy_e_token?(current_entity)
+            choices
+          end
+
+          def choice_name
+            'E-Token allows the purchase of E-Trains'
+          end
+
+          def log_skip(entity)
+            super if @game.electric_dreams? && @game.phase.status.include?('gray_uses_black')
```
crericha, line 36: Use some other check that is connected to electric dreams, not another status that just happens to line up.

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
...
+              @game.phase.name == '12H' ? 800 : 550
+            else
+              @game.phase.name == '12H' ? 1100 : 800
+            end
+          end
+
+          def process_choose(action)
+            raise GameError, 'Cannot buy E-token' unless can_buy_e_token?(action.entity)
+
+            buy_e_token(action.entity) if action.choice == "Buy E-Token for #{@game.format_currency(e_token_cost)}"
```
crericha, line 56: There is only one choice, so you don't have to compare the output. And if you wanted to compare the output, you should generate that output in one place and use that in both `def choices` and here.

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+
+          def buy_e_token(corporation)
+            total_cost = e_token_cost
+            corporation.spend(total_cost, @game.bank)
+            @log << "#{corporation.name} buys an E-Token for #{@game.format_currency(total_cost)}."
+            @log << "#{corporation.name} can now buy E-Trains."
+            corporation.e_token = true
+            @game.e_token_sold = true
+            # @acted = false
```
crericha, line 66: Remove commented out code

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+
+          def can_buy_e_token?(entity)
+            @game.electric_dreams? &&
+              entity.corporation? &&
+              @game.phase.status.include?('gray_uses_black') &&
```
crericha, line 72: Use something related instead of the status

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+  module Game
+    module G1849
+      module Step
+        class BuyToken < Engine::Step::Base
+          ACTIONS = %w[buy_token pass].freeze
+          MIN_PRICE = 1
+
+          def actions(entity)
+            return [] unless entity == current_entity
+            return [] unless can_buy_token?(entity)
```
crericha, line 15: Can be one line `return [] if entity != current_entity || !can_buy_token?(entity)`

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+
+            ACTIONS
+          end
+
+          def round_state
+            super.merge(
+              {
+                bought_token: false,
```
crericha, line 23: You don't need this in round state. It is only used within this step. Just do something like `@bought_token = false` in `def setup`.

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+              !available_tokens(entity).empty? &&
+              MIN_PRICE <= buying_power(entity)
+          end
+
+          def available_tokens(entity)
+            entity.tokens_by_type
+          end
+
+          def can_sell_token?(token, entity)
+            token.corporation.placed_tokens.size > 1 &&
```
crericha, line 47: This check in unnecessary if you can't sell your first token. By definition you have more than one placed token (your home + this one).

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+            token != token.corporation.tokens.first &&
+            !token.city.tokened_by?(entity)
+          end
+
+          # look for any cities reachable from entity that are tokened by another corporation that
+          # has at least 2 tokens placed
+          #
+          # this is expensive - move to auto_actions
```
crericha, line 55: Looks like this is an auto action? Is the comment out of date?

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+            false
+          end
+
+          def auto_actions(entity)
+            return [Engine::Action::Pass.new(entity)] unless any_buyable_tokens_placed?(entity)
+
+            super
+          end
+
+          # 1849 doesn't allow more than one token per corporation per tile
```
crericha, line 76: 1849 doesn't have double O tiles, so why is this relevant?

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+          def can_token_city?(entity, city)
+            city.tile.nodes.select(&:city?).none? { |c| c.tokened_by?(entity) }
+          end
+
+          def can_replace_token?(entity, token)
```
crericha, line 81: How is this different from `can_sell_token?`

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+              (city.pass? || other_corporation.placed_tokens.count { |t| !t.city.pass? } > 1) &&
+              @game.token_graph_for_entity(entity).connected_nodes(entity)[city]
+          end
+
+          def description
+            'Buy a Token From Another Corporation'
+          end
+
+          def pass_description
+            'Skip (Buy a Token)'
```
crericha, line 101: Pass instead of Skip? Skip means you have no action in our vernacular.

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+          end
+
+          def log_skip(entity)
+            return if !@game.acquiring_station_tokens? || !@game.phase.tiles.include?(:green)
+
+            @log << "#{entity.name} can't buy any tokens from other corporations"
+          end
+
+          def log_pass(entity)
```
crericha, line 110: Pass doesn't mean you can't buy a token. You may have just passed. Need a way to discern.

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+            @log << "#{entity.name} can't buy any tokens from other corporations"
+          end
+
+          def max_price(entity)
+            buying_power(entity)
+          end
+
+          def available_hex(entity, hex)
+            @game.token_graph_for_entity(entity).reachable_hexes(entity)[hex]
```
crericha, line 119: Does `connected_nodes` align with `reachable_hexes`? I was thinking this would be `connected_hexes`.

lib/engine/game/g_1849/step/buy_token.rb
crericha: And actually, this doesn't make sense in conjunction with the auto action.

lib/engine/game/g_1849/step/buy_train.rb
```diff
@@ -1,12 +1,15 @@
...
 
+require_relative '../../../step/base'
 require_relative '../../../step/buy_train'
 
 module Engine
   module Game
     module G1849
       module Step
         class BuyTrain < Engine::Step::BuyTrain
+          attr_accessor :e_token
```
crericha, line 11: I didn't see step.e_token used anywhere. Is that correct? If so, then you don't need to expose it external to the step. @e_token suffices to use throughout the step.

crericha: And actually seems to not be used at all, in which case can be removed.

lib/engine/game/g_1849/step/buy_train.rb
```diff
@@ -34,6 +37,34 @@ def can_sell?(entity, bundle)
 
             super
           end
+
+          def buyable_trains(entity)
+            # Cannot buy E-train without E-token
+            trains_to_buy = super
+
+            trains_to_buy = trains_to_buy.reject { |t| t.name == 'E' }
+            trains_to_buy << e_train if @game.electric_dreams? && can_buy_e?(entity)
```
crericha, line 46: It doesn't make sense to reject and then add back in the e train. instead do something like
```
if !@game.electric_dreams? || !can_buy_e?(entity)
  trains_to_buy.reject! { |t| t.name == "E" }
```
trains_to_buy

lib/engine/game/g_1849/step/buy_train.rb
```diff
@@ -34,6 +37,34 @@ def can_sell?(entity, bundle)
...
+            trains_to_buy.uniq
+          end
+
+          def e_train
+            @depot.depot_trains.find { |t| t.name == 'E' }
+          end
```
crericha, line 52: Not needed. Remove.

lib/engine/game/g_1849/step/buy_train.rb
```diff
@@ -34,6 +37,34 @@ def can_sell?(entity, bundle)
...
+
+          def can_buy_e?(entity)
+            entity.e_token &&
+              e_train.price <= entity.cash
```
crericha, line 56: Can you not e_buy an E train if you have the e token? If you can, then `e_train.price <= entity.cash` isn't needed. `buyable_trains` method will auto handle it. You only need to know if the corp has an e token.

lib/engine/game/g_1849/step/buy_train.rb
```diff
@@ -34,6 +37,34 @@ def can_sell?(entity, bundle)
...
+          end
+
+          def check_for_cheapest_train(train)
+            entity = @game.round.current_operator
+
+            cheapest = @depot.send(@game.electric_dreams? && !entity.e_token ? :min_depot_train_no_e_token : :min_depot_train)
```
crericha, line 62: Why do you need to check `electric_dreams?` Isn't checking for `e_token` sufficient? Or does corporation not have `e_token` attr unless electric dreams is enabled?

crericha: `send` is a weird way to do this. The more standard way is
```
entity.e_token ? @depot.min_depot_train_no_e_token : @depot.min_depot_train
```

crericha, line 62: `send` is a weird way to do this. The more standard way is
```
entity.e_token ? @depot.min_depot_train_no_e_token : @depot.min_depot_train
```

lib/engine/game/g_1849/step/emergency_money_raising.rb
```diff
@@ -0,0 +1,93 @@
...
+          def description
+            'Emergency Money Raising'
+          end
+
+          def cash_crisis?
+            true
+          end
+
+          def active?
+            active_entities.any?
```
crericha, line 40: !empty?

lib/engine/game/g_1849/step/special_buy.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/special_buy'
+require_relative 'buy_e_token'
+
+module Engine
+  module Game
+    module G1849
+      module Step
+        class SpecialBuy < Engine::Step::SpecialBuy
```
crericha, line 10: Why is there this step and a buy_e_token step? That's redundant.

lib/engine/game/g_1849_boot/game.rb
```diff
@@ -86,31 +86,33 @@ class Game < G1849::Game
           },
         ].freeze
 
-        TRAINS = [{ name: '4H', num: 4, distance: 4, price: 100, rusts_on: '8H' },
```
crericha, line 89: Why this change? Is there anything different than formatting?

assets/app/view/game/corporation.rb
crericha: My issue is that the corporation class is being queried. It shouldn't be used. Get the text from game. Set the default in the base game class to `Loans`, so you don't have to use `respond_to?`

assets/app/view/game/corporation.rb
```diff
@@ -638,9 +638,10 @@ def render_loans
...
         [
           h('tr.ipo', loan_props, [
-            h('td.right', 'Loans'),
+            h('td.right', corp_loans_text),
```
crericha, line 644: Only used once, so you don't need a local var. You can just put the logic here. Local vars are useful if used more than once.

assets/app/view/game/loans.rb
crericha: Button still in the method name.

assets/app/view/game/loans.rb
crericha: Not resolved.

lib/engine/game/g_1849/corporation.rb
crericha: Last I remember from slack, you decided to make `e_token` an ability, so this accessor shouldn't be needed.

lib/engine/game/g_1849/corporation.rb
```diff
@@ -15,6 +15,10 @@ def initialize(sym:, name:, **opts)
           shares.last.last_cert = true
           shares.last.double_cert = true
         end
+
+        def corp_loans_text
+          'Bonds Issued'
+        end
```
crericha, line 21: Move to base game.

lib/engine/game/g_1849/depot.rb
crericha: This is fine

lib/engine/game/g_1849/game.rb
crericha: `no_bonds` still shows up in status.

lib/engine/game/g_1849/game.rb
```diff
@@ -83,43 +92,55 @@ class Game < Game::Base
             train_limit: 2,
             tiles: %i[yellow green brown],
             operating_rounds: 3,
-            status: ['gray_uses_black'],
+            status: %w[gray_uses_black e_tokens_available],
```
crericha, line 95: `e_tokens_available` needs to be conditional on that optional rule being enabled.

lib/engine/game/g_1849/game.rb
```diff
@@ -51,15 +60,15 @@ class Game < Game::Base
             train_limit: 4,
             tiles: [:yellow],
             operating_rounds: 1,
-            status: ['gray_uses_white'],
+            status: %w[gray_uses_white no_bonds no_buy_token],
```
crericha, line 63: `no_buy_tokens` needs to be conditional on that optional rule being enabled.

lib/engine/game/g_1849/game.rb
```diff
@@ -83,43 +92,55 @@ class Game < Game::Base
...
+          { name: 'E', num: 6, available_on: '12H', distance: 99, price: 550 },
+          { name: 'R6H', num: 2, available_on: '16H', distance: 6, price: 350 },
+        ].freeze
+
+        def game_trains
+          train_list = super.dup
+          train_list.reject! { |t| t[:name] == 'E' } unless electric_dreams?
+          train_list.find { |t| t[:name] == '6H' }[:events] << { 'type' => 'buy_tokens' } if acquiring_station_tokens?
+          train_list.find { |t| t[:name] == '8H' }[:events] << { 'type' => 'bonds' } if bonds?
+          train_list.find { |t| t[:name] == '12H' }[:events] << { 'type' => 'e_tokens' } if electric_dreams?
```
crericha, line 141: e_tokens shouldn't be an event and status. Pick one.

lib/engine/game/g_1849/game.rb
```diff
@@ -83,43 +92,55 @@ class Game < Game::Base
...
+          },
+          { name: '16H', distance: 16, price: 1100 },
+          { name: 'E', num: 6, available_on: '12H', distance: 99, price: 550 },
+          { name: 'R6H', num: 2, available_on: '16H', distance: 6, price: 350 },
+        ].freeze
+
+        def game_trains
+          train_list = super.dup
+          train_list.reject! { |t| t[:name] == 'E' } unless electric_dreams?
+          train_list.find { |t| t[:name] == '6H' }[:events] << { 'type' => 'buy_tokens' } if acquiring_station_tokens?
```
crericha, line 139: buy_tokens shouldn't be an event and status. Pick one.

lib/engine/game/g_1849/game.rb
```diff
@@ -284,6 +313,7 @@ def setup
           setup_companies
           afg # init afg helper
           remove_corp if @players.size == 3
+          remove_corp if @players.size == 4 && reduced_4p_corps?
```
crericha, line 316: Make this an `||` with the previous line.

lib/engine/game/g_1849/game.rb
crericha: Steps still here.

lib/engine/game/g_1849/game.rb
```diff
@@ -167,7 +188,13 @@ class Game < Game::Base
...
           earthquake: ['Messina Earthquake',
                        'Messina (B14) downgraded to yellow, tokens removed from game.
-                        Cannot be upgraded until after next stock round']
+                       Cannot be upgraded until after next stock round'],
+          buy_tokens: ['Cross-buy Station Tokens',
+                       'Corporations may now buy station tokens from other corporations'],
+          bonds: ['Bonds Available',
+                  'Corporations can issue a single L.500 bond, with L.50 interest per OR'],
+          e_tokens: ['E-Tokens Available',
+                     'Corporations can buy E-tokens to allow the purchase of E-Trains'],
```
crericha, line 197: Do these optional events only show up on the info tab when the rule is enabled?

assets/app/view/game/corporation.rb
philcampeau: done

lib/engine/game/g_1849/game.rb
philcampeau: I spoke with Chris; now I understand he meant to remove them _when_ they're unused (variant not enabled). Done with commit `variant steps are only included if respective variant is enabled.`

assets/app/view/game/corporation.rb
```diff
@@ -638,9 +638,11 @@ def render_loans
           loan_props[:style][:color] = contrast_on(color)
         end
 
+        # the 'Loans' text is now defined in lib/engine/game/base.rb and linked here as @game.corp_loans_text.
+        # This way, games can customize the text if desired (initially used in 1849's Bonds variant)
```
crericha, line 642: Remove the comments. They get out of date as the code evolves. If someone wants to see why the code was added, they can use blame and it'll link them back to this PR.

assets/app/view/game/loans.rb
crericha: `scrap_button_text` was poor naming. The idea is the "view" layer (where the html is) is separate from the "model" layer (where the logic is). The model layer shouldn't know anything about how the html is structured. In this case, the knowledge that the view layer implements paying off loans via a button isn't something it should know. Additionally, the model layer shouldn't have to change if the button was changed to something something else. All the model layer has to do is tell the view layer what phrase to use for paying off a loan. This is why `payoff_loan_text` is a better name because now the model layer isn't implying anything about the "view" layer. Reference to the model - view architecture used by the site: https://www.geeksforgeeks.org/mvc-design-pattern/.

crericha: And ideally, you could use something like `"Payoff #{@game.corp_loans_text}"`, so you don't need to provide a name twice. `corp_loans_text` is plural, so some work would have to be done there like make it `corp_loan_name` and then pluralize where/when needed.

lib/engine/game/g_1849/game.rb
crericha: My comment in regards to why do you bother subtracting and adding loans. All you need to do is define total_loans = 1 to accomplish what the linked logic requires. Once the rest of this is ready, I'll add a commit that changes the view code to not require total_loans at all.

lib/engine/game/g_1849/game.rb
crericha: If you don't add/subtract loans when they are taken or paid off, then it doesn't matter. Let's remove.

lib/engine/game/g_1849/game.rb
crericha: You can still give the entity a loan. It just doesn't have to come from a common loan pool. It can be a new loan each time. Similar to how infinite tile logic works.

lib/engine/game/g_1849/step/bond.rb
crericha: Discussed previously wy button shouldn't be in the name. See model view framework discussion.

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -0,0 +1,71 @@
...
+  module Game
+    module G1849
+      module Step
+        class Bond < Engine::Step::Base
+          def actions(entity)
+            return [] if !entity.corporation? || entity != current_entity
+
+            actions = []
+            actions << 'payoff_loan' if can_payoff_loan?(entity)
+            actions << 'take_loan' if @game.can_take_loan?(entity)
```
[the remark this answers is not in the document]
crericha: Okay. No action needed.

lib/engine/game/g_1849/step/bond.rb
crericha: 1817 loans have no factor on internals to a specific game (e.g. `take_loan` is not in the base class, its game specific).

lib/engine/game/g_1849/step/bond.rb
crericha: Again, not needed for this specific game. Its unnecessary cruft.

lib/engine/game/g_1849/step/buy_token.rb
crericha: steps that utilize auto actions don't have an early return in `actions`. They always make the possible actions available and the auto action logic chooses pass for you if you can't act.

lib/engine/game/g_1849/step/buy_token.rb
crericha: Looks like there is some confusion. I'm saying that `bought_token` should not be added to round at all. It is only used here, so it doesn't need to be available outside of this step.

lib/engine/game/g_1849/step/buy_token.rb
crericha: If this is related to buying a token, why is it called `can_replace_token?` instead of `can_buy_token?`?

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,154 @@
...
+
+          def description
+            'Buy a Token From Another Corporation'
+          end
+
+          def pass_description
+            'Skip (Buy a Token)'
+          end
+
+          def log_skip(entity)
```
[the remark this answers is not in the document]
crericha: What do you mean by the comment above?  What I'm saying is that this method can be deleted. It is dead code.

lib/engine/game/g_1849/step/buy_token.rb
crericha: How does that resolve the issue? Its either, the company can buy a token and chooses not to or cannot buy a token and must pass. The current log message only covers the first case.

lib/engine/game/g_1849/step/buy_token.rb
crericha: Okay, I know you've tested it, so if it works... Please verify you aren't able to buy a token in a non-connected hex that is upgradeable to make it connected.

lib/engine/game/g_1849/step/bond.rb
philcampeau: changed name to take_loan_text

assets/app/view/game/corporation.rb
philcampeau: done

assets/app/view/game/loans.rb
philcampeau: This was actually pretty simple, since 1849 only allows one bond to be issued, no plural was needed. Done.

philcampeau: For what it's worth though, if you search button_text, you'll find that used in a lot of instances on the site.

lib/engine/game/g_1849/game.rb
philcampeau: Alright, I've set NUM_LOANS = 1 and removed all instances of subtracting and adding loans to @loans.

lib/engine/game/g_1849/step/bond.rb
philcampeau: I think I misunderstood what you were saying initially. I removed action.loan from the method, and it still works fine, so consider this done.

lib/engine/game/g_1849/step/bond.rb
philcampeau: removed `@game.loans << loan`

assets/app/view/game/loans.rb
philcampeau: done

lib/engine/game/g_1849/game.rb
philcampeau: done

lib/engine/game/g_1849/game.rb
philcampeau: done

lib/engine/game/g_1849/step/buy_token.rb
philcampeau: got it. Fixed now.

lib/engine/game/g_1849/step/buy_token.rb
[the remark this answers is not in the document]
philcampeau: Gotcha. Gone. done.

lib/engine/game/g_1849/game.rb
```diff
@@ -531,13 +578,25 @@ def check_distance(route, _visits)
...
           return unless (route.stops.map(&:hex).map(&:id) & PORT_HEXES).any?
 
           raise GameError, 'Route must include two non-port stops.' unless route.stops.size > 2
         end
 
+        def check_e_train_track_type(route)
+          train = route.train
+          paths = route.paths
+
+          raise GameError, 'E-Trains cannot use narrow gauge' if train.name == 'E' && paths.any? { |p| p.track == :narrow }
```
crericha, line 591: Is this message too long for the route display? If so, shorten to `cannot use narrow gauge`

lib/engine/game/g_1849/game.rb
```diff
@@ -282,13 +319,18 @@ def price_movement_chart
...
 
           @available_par_groups = %i[par]
 
           @player_debts = Hash.new { |h, k| h[k] = 0 }
           @moved_this_turn = []
+
+          # below is for variant rules
+          @e_tokens_enabled = false
+          @buy_tokens_enabled = false
+          @issue_bonds_enabled = false
```
crericha, line 333: These variables do not need to be initialized. Can delete from here.

lib/engine/game/g_1849/game.rb
```diff
@@ -770,12 +829,141 @@ def bank_sort(corporations)
...
+
+        def event_e_tokens!
+          @e_tokens_enabled = true
+          @log << "-- Event: #{EVENTS_TEXT[:e_tokens][1]} --"
+        end
+
+        # code below is for Bonds variant
+        MAXIMUM_LOANS = 1
+        LOAN_VALUE = 500
+        INTEREST_RATE = 50
```
crericha, line 877: Should this go at the top with the other constants?

lib/engine/game/g_1849/round/operating.rb
```diff
@@ -7,11 +7,24 @@ module Game
...
+
           def next_entity!
             return @game.end_game! if @entities[@entity_index].reached_max_value
 
             super
           end
+
+          def cash_crisis_entity
+            @game.corporations.find { |corp| corp.cash.negative? }
+          end
```
crericha, line 27: None of the changes in this file should be needed. The @issue_bond and @redeemed_bound should be done via `round_state` and `cash_crisis_entity` can be done directly in the emr step.

lib/engine/game/g_1849/step/buy_e_token.rb
```diff
@@ -0,0 +1,84 @@
...
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G1849
+      module Step
+        class BuyEToken < Engine::Step::Base
+          def actions(entity)
+            actions = []
+            actions += %w[choose pass] if can_buy_e_token?(entity) && entity == current_entity
```
crericha, line 12: The step shouldn't use auto actions and conditional actions here. Since `can_buy_e_token?` is relatively cheap, remove the auto action logic.

lib/engine/game/g_1849/step/buy_token.rb
```diff
@@ -0,0 +1,127 @@
...
+              (city.pass? || other_corporation.placed_tokens.count { |t| !t.city.pass? } > 1) &&
+              @game.token_graph_for_entity(entity).connected_nodes(entity)[city]
+          end
+
+          def description
+            'Buy a Token From Another Corporation'
+          end
+
+          def pass_description
+            'Skip (Buy Token)'
```
crericha, line 84: Its pass if the player hits pass and skip if it automatically passes for them. I guess for now, to not complicate things, just use the default pass behavior and don't override it. We can enhance later if it becomes an issue for players.

lib/engine/game/g_1849/game.rb
philcampeau: done

lib/engine/game/g_1849/step/buy_e_token.rb
philcampeau: Done

lib/engine/game/g_1849/game.rb
philcampeau: done

lib/engine/game/g_1849/round/operating.rb
philcampeau: done

lib/engine/game/g_1849/step/buy_e_token.rb
philcampeau: done

lib/engine/game/g_1849/step/buy_token.rb
philcampeau: done

lib/engine/game/g_1849/game.rb
philcampeau: Looks ok to me: 
image

crericha (approved):

PR #10747, opened 2024-05-11
[1817] fix check for exporting trains (relevant when other games inherit)

### LINKED ISSUE #10746: [18Hiawatha] inherited 1817 logic for or_round_finished bails when upcoming train is nil
18Hiawatha "Pass" Button Not Working 

[GameID 162717](https://18xx.games/game/162717)
"Pass" button isn't working.

Loans and other features already to still be working but the player can't end their turn. 

All platforms.

benjaminxscott:
Fixes #10746

## Before clicking "Create"

### Explanation of Change

Fixes an issue where the 1817 export logic would bail if there were no upcoming trains 

Only relevant in cases where games inherit from 1817 code since 1817 has infinite trains, but I wanted to fix it at the source rather than expecting those games to handle this case

Specifically 18Hiawatha

### Screenshots

Screenshot 2024-05-11 4 24 21 PM

### DIFF
--- a/lib/engine/game/g_1817/game.rb
+++ b/lib/engine/game/g_1817/game.rb
@@ -978,6 +978,8 @@ module Engine
         end
 
         def or_round_finished
+          return if @depot.upcoming.empty?
+
           if @depot.upcoming.first.name == '2'
             depot.export_all!('2')
           else


lib/engine/game/g_1817/game.rb
```diff
@@ -978,6 +978,8 @@ def operating_round(round_num)
         end
 
         def or_round_finished
+          return unless @depot.upcoming.any?
```
ollybh, line 981: Don't call `Enumerable.any?` (or `Enumerable.none?`) without a block, this method will iterate over each element in the array testing if it is nil. Use `Array.empty?` instead 
```suggestion
          return if @depot.upcoming.empty?
```

ollybh (changes requested):

lib/engine/game/g_1817/game.rb
crericha: Agree with this change. `any?` is significantly more expensive than `.empty?`.

benjaminxscott:
Thanks I knew that one of any? / empty? were expensive but not which one, so I figured rubocop would fix it

crericha:
It's expensive after being transpiled to JavaScript, so RoboCop would have
no idea.

PR #10790, opened 2024-05-23
[base] removes version info from docker-compose files

philcampeau:
## Before clicking "Create"

### Explanation of Change

Removed the `version: 3.5` entry from the docker-compose files in the root folder of the site. [Version info is deprecated in docker-compose](https://nickjanetakis.com/blog/docker-tip-51-which-docker-compose-api-version-should-you-use)

### Screenshots

image

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,5 +1,3 @@
-version: '3.5'
-
 x-rack: &rack
   image: 18xx_rack:dev
   environment:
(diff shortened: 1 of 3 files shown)


benjaminxscott:
fwiw the docker-compose on prod is really old so I'm not sure this will work as expected 

> deploy@18xxgames:~$ docker-compose --version
docker-compose version 1.25.5, build 8a1c60f6

philcampeau:
I had a feeling there was probably a reason this wouldn't work!

PR #10791, opened 2024-05-23
[18 India] [Core] Implement Hidden Hand for Players View and IPO Rows View

EdPovi:

## Implementation Notes
Modify view for player card to allow for hidden hand and implement a show/hide button for use during stock round. 
Create new view for IPO Rows, to replace a long list of companies on Entities page and during stock round.

### Explanation of Change
- Add ability to show "hidden hand" on player card (a new view that checks current user to show a button)
- Add button to show/hide player's hand of options during stock round
- Create a new view to improve the display of companies in the IPO Rows
- During Stock Round, the IPO rows show a :company card" for the first item on the row, which can be used to buy it. 

### Screenshots
image
image
image
image

### Any Assumptions / Hacks
Added two methods to the base game code to manage the views: `hand_companies_for_stock_round` and `show_ipo_rows?`.


assets/app/view/game/player.rb
```diff
@@ -38,6 +39,7 @@ def render
         if @show_companies
           divs << h(Companies, owner: @player, game: @game, show_hidden: @show_hidden) if @player.companies.any? || @show_hidden
           divs << h(UnsoldCompanies, owner: @player, game: @game) unless @player.unsold_companies.empty?
+          divs << h(HiddenHand, player: @player, game: @game, user: @user) if @hidden_hand
```
michaeljb, line 42: This currently renders `HiddenHand` on player cards for every game. You need some way to gate this for only 18 India. Maybe instead of `hidden_hand` being a need for this view, this line checks a new method on the game, and it works basically like the method `show_ipo_rows?` you're adding here.

michaeljb (changes requested):
Looking good once you keep `HiddenHand` from rendering for other games, and it looks like there's a merge conflict after the gauge change stuff got merged.

assets/app/view/game/player.rb
EdPovi: I think my original idea was to check if `hand_companies_for_stock_round` was empty, but a boolean flag in the game/base file seems more straightforward, as you suggested.

EdPovi:
I think I have resolved your comment and resolved the merge conflict.  

Thanks,
Ed

michaeljb (approved):
Looks good. Nice job with this!

This and the gauge changes were definitely features that stood out to me as things that would be challenging/annoying to implement when I first played 18 India. Exciting to see it progressing!

PR #10796, opened 2024-05-25
[TOP1871] fixes liquidity calculation

### LINKED ISSUE #8991: 1871 The Old Prince incorrect liquidity
https://18xx.games/game/115076?action=238#entities

Screenshot 2023-03-07 at 15 29 06

Girrafic Violence is shown with 700 liquidity, which is 94 + (101 x 6). However, because he can only sell them three at a time, his maximum liquidity is actually 94 + (3 x 100) + (3 x 94) = 679.

philcampeau:
Fixes #8991 

## Before clicking "Create"

### Explanation of Change

I was thinking about this, and I agree with the initial report that the liquidity is wrong, but I disagree with the calculation. Liquidity is how much money the player potentially has access to _now_. So actually, liquidity is player cash + the value of maximum 3 sellable shares per corporation.

Additionally, the calculation would need to ignore the value of any shares in corporations that the player has already sold that turn. 

To do this, I took the `def value_for_dumpable` method, which is referenced in `def liquidity`, from base.rb, which is:

```
      def value_for_dumpable(player, corporation)
        return value_for_sellable(player, corporation) if self.class::PRESIDENT_SALES_TO_MARKET

        max_bundle = bundles_for_corporation(player, corporation)
          .select { |bundle| bundle.can_dump?(player) && @share_pool&.fit_in_bank?(bundle) }
          .max_by(&:price)
        max_bundle&.price || 0
      end
```

I added this method to the game.rb file and added two .select conditions ahead of the conditions already in place: `@round.active_step&.can_sell?(player, bundle)` and `bundle.num_shares <= 3`.

### Any Assumptions / Hacks

I don't see why this would impact any games in place, since it only affects the displayed liquidity value, but I still tested this with a few games from the site and they all loaded correctly.

### DIFF
--- a/lib/engine/game/g_1871/game.rb
+++ b/lib/engine/game/g_1871/game.rb
@@ -554,6 +554,15 @@ module Engine
           max_bundle&.price || 0
         end
 
+        # Forces Union Bank's liquidity to only every display its cash holding unless emergency funding,
+        # since the Union Bank cannot sell shares outside of emergency money raising.
+
+        def liquidity(player, emergency: false)
+          return player.cash if player == @union_bank && !emergency
+
+          super
+        end
+
         # Need to redefine this in order to add in union bank to the mix after
         # each round. The base function removes the bank since it's not active
         # during the stock round.
--- a/lib/engine/game/g_1871/game.rb
+++ b/lib/engine/game/g_1871/game.rb
@@ -539,6 +539,22 @@ module Engine
           player.value
         end
 
+        # This code limits the liquidity calculation to only include a maximum of
+        # 3 shares per corporation, since that is the most a player can sell in
+        # a single turn. It also doesn't count the value of shares for corporations
+        # that the player has already sold this turn.
+        def value_for_dumpable(player, corporation)
+          max_bundle = bundles_for_corporation(player, corporation)
+            .select do |bundle|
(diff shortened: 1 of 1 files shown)


philcampeau:
After submitting, I noticed that the code threw an error when trying to access the Entities tab during Operating rounds. 

I modified the method, so that now it reads as follows: 

```
        def value_for_dumpable(player, corporation)
          max_bundle = bundles_for_corporation(player, corporation)
            .select do |bundle|
            @round.active_step&.can_sell?(player, bundle) if @round.is_a?(G1871::Round::Stock)
            bundle.num_shares <= 3 &&
            bundle.can_dump?(player, bundle)
          end
            .max_by(&:price)
          max_bundle&.price || 0
        end
```

Now the @round.active_step&.can_sell? is only evaluated during stock rounds. I also removed the condition to check if there was room in the bank pool. Since 1871 has an 80% bankpool limit and doesn't allow the sale of presidencies, this was an unnecessary check.

philcampeau:
Unrelated to the first issue, but since I was working on liquidity, I added a new calculation for liquidity specifically for the Union bank. Since the Union Bank can only ever sell shares while emergency funding, I added this block of code: 

```
        def liquidity(player, emergency: false)
          return player.cash if player == @union_bank && !emergency

          super
        end
```

PR #10800, opened 2024-05-26
[1822CA] fix Prescott-New York river

### LINKED ISSUE #10799: [1822CA] River not removed during track lay
The track lay between Prescott and New York does not remove the river barrier and should since track spans the barrier.

https://18xx.games/game/164921

michaeljb:
Fixes #10799


michaeljb:
Changing the border at setup breaks the consistent-borders check, will change it to remove the border when the tile is laid

PR #10797, opened 2024-05-25
Dev mode: compile only game files for requested games

michaeljb:
Stop waiting for the files for all titles to build in local dev mode; only build for titles as needed.

## Before clicking "Create"

- [x] Branch is derived from the latest `master`
- [x] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

PR #10810, opened 2024-05-27
 Dev mode: compile only game files for requested games

michaeljb:
I sorted the caching issues I was having in #10797 and organized the changes here into more descriptive commits for each piece.

This changes functions in `lib/assets.rb` to use a `titles` param so that only the given titles are built, and change cached functions to cache things per-title when possible instead of caching the whole function call. In non-production mode, the API will trigger builds for game files that aren't already built.

It's best to view this diff with the "Hide whitespace" setting enabled.

## Before clicking "Create"

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -102,7 +102,7 @@ desc 'Precompile assets for production'
 task :precompile do
   require_relative 'lib/assets'
   assets = Assets.new(cache: false, compress: true, gzip: true)
-  assets.combine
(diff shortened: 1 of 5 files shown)


michaeljb:
This works for multiplayer games but not for hotseat games. In api.rb, looks like the title for hotseat games isn't being passed in to `js_tags()` so the specific game title doesn't get requested.

PR #10817, opened 2024-05-27
[1846] during auction, show private companies under bank in Entities tab

### LINKED ISSUE #10535: [1846] Show privates in the game under Entities during Draft
Currently during Draft we don't display any privates under Entities in a >2 player game, likely due to the 'hidden hand' code

It would be useful to show the list of privates under Entities, since the alternative is memorizing all the privates and then subtracting the ones that were removed from the game. (which isn't secret information)

michaeljb:
Fixes #10535

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1846/game.rb
+++ b/lib/engine/game/g_1846/game.rb
@@ -989,6 +989,10 @@ module Engine
 
           super
         end
+
+        def unowned_purchasable_companies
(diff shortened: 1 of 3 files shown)


benjaminxscott (approved):
Sweet I don't have my laptop or a build env but it looks like that block would be rendered under entities

PR #10828, opened 2024-05-28
[1860] implements simplified insolvency variant

### LINKED ISSUE #9484: [1860] Simplified insolvency calculation
**What's your idea?**
The simplified revenue calculation for insolvent companies from the rules: the company's route doesn't actually matter, just the max theoretical route for the next available train, as printed on the insolvency cards: 
4=120, 5=140, 6=160, 7=180, 8=200, 9=220

Can also combine with the half-revenue optional rule

**Are there related ideas?**
No

philcampeau:
Fixes #9484

## Before clicking "Create"

### Explanation of Change

Implements the simplified insolvency rules from the game rules.


roseundy (approved):
Looks good.

PR #10834, opened 2024-05-31
[18Mag] restricts expansion content to 3+ players

philcampeau:
## Before clicking "Create"

### Explanation of Change

The expansion content is restricted to 3-6 players, but currently the site allows you to create 2-player games using expansion content. Some of these do not work and throw out errors. 

This update will prevent 2 player games from being started with those options enabled.


roseundy (approved):
Looks good

PR #10836, opened 2024-05-31
[core] end game if SR 100 is reached

### LINKED ISSUE #10743: 1846 (and possibly other games): Game temporarily soft-locks if all companies close
I'm not sure how to copy the hotseat JSON data but it's easy to replicate. 

Create a game of 1846. Open all the companies, get to the Brown phase which eliminates all private companies, then keep holding until all companies are closed. Make sure the bank does not break (when I tested this I had a bank at over $7,000 in a 5P game when I was able to make the bug happen). 

When the last company is operating at share price $10, if you withhold, the game soft-locks and you cannot access 18xx.games on your device for a few minutes. 

What I suspect is happening is an endless loop of "Player has no valid actions and passes" where nobody can respond with game input. 

It's a rather ridiculous set of circumstances, but it can technically happen. 

Suggestion: If "no valid actions" loops for all players 10 times without a change from SR to OR, OR to next OR, or OR to SR, the game should be ended immediately.

Soft lock ends after about 10-20 minutes and goes back to the previous action.

michaeljb:
End the game at SR 100 to catch infinite loops where no valid actions can be found, and possibly no defined ending has been reached.

I spent some time trying to catch proper end games in such technically-legal-but-not-going-to-happen-for-real scenarios, like all corporations closing in 1846 while private companies eventually drain the bank, but I just kept discovering more possible edge cases and the best simple solution I did have still broke another test.

Fixes #10743

## Before clicking "Create"


philcampeau:
Well there goes my idea to create an ultra-epic 300-set game. :P

PR #10840, opened 2024-06-01
[18RoyalGorge] bug fixes for DEBT and US Mint Worker (B6)

### LINKED ISSUE #10838: [18royalgorge] Mint Worker private doesn't discount 2-share bundle price
https://18xx.games/game/165364

 private B6 doesn’t work as intended.  The 50% discount is applied if purchasing one share of VGC.   However, if two shares are purchased, they are at full price rather than the 50% discount.

michaeljb:
* only increase DEBT after Treaty of Boston

* fix US Mint Worker (B6)'s discount when buying 2 shares

    * the `modify_purchase_price` method actually affects the per-share price
      of the bundle, not the overall bundle price, so it didn't work with B6's
      two-share bundle

    * change the `buy_share` button to display the `modified_price` times the
      number of shares to given the total modified bundle price

Fixes #10838 

## Before clicking "Create"

### DIFF
--- a/assets/app/view/game/button/buy_share.rb
+++ b/assets/app/view/game/button/buy_share.rb
@@ -28,7 +28,7 @@ module View
           reduced_price = @game.format_currency(bundle.price - @swap_share.price) if @swap_share
           if step.respond_to?(:modify_purchase_price)
             modified_price = step.modify_purchase_price(bundle)
-            modified_price = nil if bundle.price == modified_price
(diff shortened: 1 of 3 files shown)


michaeljb:
@roseundy I think Rolling Stock is the only other game that uses `modify_purchase_price`. I think in the view before my change here, it expected the modified bundles to always be only one share bundles. Will this change work with Rolling Stock?

roseundy (commented):
I think this is fine for RS, since you can only buy 1 share at a time.

PR #10841, opened 2024-06-01
[core] fix being stuck on entity without valid actions after closing corporations

### LINKED ISSUE #10821: [1846] corporation closure can cause a softlock where active player has no available actions in the UI
Copied from discussion on #10743:

-----

While setting up my own repro case, I stumbled upon another bug; Player 1 is president of the IC, sells a share so that its price drops and it closes. Before the time of the sale, Player 2 had two shares, so it logs that Player 2 becomes president, which doesn't really make sense at that point, but the real bug is that on Player 2's turn, they have no valid actions, but the game blocks for them to take their turn.

Since pass is not valid, the pass button is not visible and the game cannot proceed.

I think this is because company closures are handled in action_processed(), at the end of processing an action, so there's probably some weird interactions where Player 2's valid actions are checked before the corporation is closed, and it's determined that Player 2 could sell a share, but after the closure is executed, they no longer could sell that share.

https://gist.github.com/michaeljb/bcd0437bba665def8052b5080a3b9efc

michaeljb:
Fixes #10821, where Player 2's only valid actions were to sell shares in the IC, but the IC was closed on Player 1's turn. Since the corporation closure check happens after the `skip_steps`/etc, the game thinks it's Player 2's turn, but now they don't have any valid actions for the current step, and no pass or share buttons in the UI.

This adds another check for valid actions after processing corporation closures.

## Before clicking "Create"


lib/engine/game/base.rb
```diff
@@ -2798,13 +2798,24 @@ def additional_ending_after_text
...
+        return unless close_corporations_in_close_cell!
+
+        # Closing corporations can end up removing the current entity's only
+        # available actions, for example if all they could do was sell shares in
+        # the now closed corporation. If that is the case, skip to the next
+        # entity.
+        return unless active_step&.actions(current_entity)&.empty?
+
+        @round.skip_steps
+        @round.next_entity!
```
crericha, line 2810: `next_entity!` is not defined in `round/base.rb`. Can it be added there? The auction based classes are the only ones that don't implement it.

michaeljb: Good catch, thanks. Done.

PR #10854, opened 2024-06-02
[18Norway] Issue/redeem shares

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -487,7 +487,7 @@ module Engine
         end
 
         def emergency_issuable_bundles(corp)
-          return [] if corp.trains.any?
+          return [] unless corp.trains.empty?
 
           available = @depot.available_upcoming_trains.reject { |train| ship?(train) }
           return [] unless (train = available.min_by(&:price))
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -252,6 +252,7 @@ module Engine
             Engine::Step::SpecialTrack,
             Engine::Step::SpecialToken,
             Engine::Step::BuyCompany,
+            G18Norway::Step::IssueShares,
             Engine::Step::HomeToken,
             G18Norway::Step::Track,
             G18Norway::Step::BuildTunnel,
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -366,6 +367,67 @@ def check_route_token(route, token)
...
+
+          share_price = stock_market.find_share_price(entity, :right).price
+
+          bundles_for_corporation(share_pool, entity)
+            .each { |bundle| bundle.share_price = share_price }
+            .reject { |bundle| entity.cash < bundle.price }
+        end
+
+        def emergency_issuable_bundles(corp)
+          return [] if corp.trains.any?
```
michaeljb, line 396: We use `!empty?` instead of `any?` as it compiles to more efficient JavaScript. `any?` gives a callback function that is called iteratively for the collection, whereas `empty?` can just look at the size of the array.

```suggestion
          return [] unless corp.trains.empty?
```

lib/engine/game/g_18_norway/game.rb
```diff
@@ -366,6 +367,67 @@ def check_route_token(route, token)
...
+            directions = [:left] * (1 + bundle.num_shares)
+            bundle.share_price = stock_market.find_share_price(corp, directions).price
+          end
+
+          bundles.reject! { |b| b.price.zero? }
+
+          bundles.sort_by!(&:price)
+
+          train_buying_bundles = bundles.select { |b| (corp.cash + b.price) >= train.price }
+          if train_buying_bundles.any?
```
michaeljb, line 417:
```suggestion
          unless train_buying_bundles.empty?
```

michaeljb (changes requested):
A couple lines need the same small change, and you'll need to resolve merge conflicts in `g_18_norway/game.rb` (and same thing for your other open PRs I didn't merge yet). Otherwise, looks good!

michaeljb (approved):

michaeljb:
Looks like after merging #10852 and #10853 this one has conflicts again.

Git branches will always conflict if they're adding code to the same spot in one file, so one strategy I used a lot when implementing my last few games was to put new functions in different parts of the main game file, instead of always at the end. Then I would test for conflicts by merging all branches at once, and if the merge was successful, I knew I could open several PRs at once and they wouldn't produce conflicts. e.g.:

```
# checkout new branch just for checking for merge conflicts
git checkout -b 18Norway origin/master

# merge all active feature branches at once; if no conflicts, then PRs for all four feature
# requests can be opened at the same time
git merge 18Norway-feature1 18Norway-feature2 18Norway-feature3 18Norway-feature4
 
# the 18Norway branch can be reset to before the merge
git reset --hard origin/master
```

patrikolesen:
Yes it is a bit of a problem that I always update game for each new feature

PR #10868, opened 2024-06-03
[18FL] restricts 2p variant to 2 players

philcampeau:
## Before clicking "Create"

### Explanation of Change

Adds a `players` attribute to the 2 player share limit variant so this can only be used in 2p games.


philcampeau:
In retrospect, this isn't really needed, since the variant only enables if it's selected _and_ there are only two players. Leaving it out also allows players to create a game with an open play count and set it  to 70% if and only if they only get two.

closing.

PR #10871, opened 2024-06-04
[1841] color-coding the concessions

philcampeau:
## Before clicking "Create"

### Explanation of Change

Pretty much everyone I know who plays 1841 refers to the different private companies by their color, not the random jumble of 4-5 letters in its abbreviation or even its name. We have the ability to color code the privates, I think we should do it. 

### Screenshots

Here's how it would look: 

image


roseundy (approved):
Great idea. Looks good.

PR #10872, opened 2024-06-04
[18CZ] must buy usable train in last train slot

### LINKED ISSUE #9570: [18CZ] company is not forced to buy a train it can run
In this game: https://18xx.games/game/133678?action=272 BTE buys two small trains which it can't run. It should then be forced to buy the available 3+3e, but it isn't, the move is ended and it's BN's turn. Same thing happens here: If BN buys two small trains for 250 each, putting it at 0 money, it's not forced to buy a train, even though the rules say, that it must buy another train, even if that means going over the train limit and then discard one of the other trains.

philcampeau:
Fixes #9570 

## Before clicking "Create"

## Implementation Notes

_Note: this will likely require pins, for any games where players have maxed out on trains the corp couldn't run. Seems like an unlikely edge case of rules lawyering though._ 

Currently, corporations can buy trains they can't run up to the train limit, and this ends the `must_buy_train` check.

### Explanation of Change

I've added in a `needs_usable_train?` method that checks if the corp is currently in 'must_buy_train' status and if it only has one space for trains left. If this returns true, the only trains available for purchase will be trains of the size this corp can run. 

I also modified the render code so that if this is the case, the player sees a message that they need to buy a train the corp can run (top of screenshot below)

### Screenshots

image

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -218,6 +218,8 @@ module View
             children << if @step.must_issue_before_ebuy?(@corporation)
                           h(:div, "#{@corporation.name} must buy a train from another corporation, "\
                                   'or issue shares and then buy an available train ')
+                        elsif @step.respond_to?(:needs_usable_train?) && @step.needs_usable_train?(@corporation)
+                          h(:div, "#{@corporation.name} must buy a train that it can run")
                         else
                           h(:div, "#{@corporation.name} must buy an available train")
                         end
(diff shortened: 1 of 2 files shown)


philcampeau:
wait. After re-reading the bug report, this isn't fixing the problem. Will try something different.

kepakiano:
@philcampeau why does this PR not address the issue? Isn't what you changed here exactly what Lonny said? https://github.com/tobymao/18xx/issues/9570#issuecomment-2150780032

PR #10873, opened 2024-06-04
Rotate map

patrikolesen:
## Before clicking "Create"

### Screenshots
1822
image
image

18Norway
image
image

### DIFF
--- a/assets/app/view/game/axis.rb
+++ b/assets/app/view/game/axis.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 require 'view/game/hex'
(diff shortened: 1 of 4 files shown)


michaeljb (changes requested):
I don't want to add this rotate feature where the map gets presented with all the elements being difficult to read without tilting your head. After rotating it should look the same way it would look if the game had just been implemented in the new orientation in the first place.

patrikolesen:
I also agree the text and icons all becomes wrong with this implementation.
Will need to find another way to make that happen
Withdraw this proposal and will look into doing another one at a later point in time

PR #10875, opened 2024-06-05
[1822PNW] all minors get a turn in merger rounds, not just associated

### LINKED ISSUE #10662: [1822PNW] Not all companies offered for merger
https://18xx.games/game/159496?action=528

18xx.games only offers the opportunity to merge on the turn of the associated minor company. The rules are clear that the merger can be initiated on the turn of the association or unassociated minor company.

_6.1.2 The merger round takes place in descending share price order, which means that the first eligible minor company with the highest stock value has the first opportunity to merge. If an eligible minor passes, it may still be merged with another minor further down in operating order when the lower valued minor company is given the option to merge._

_6.2.1 To be eligible to merge, the minor companies must meet all requirements listed below:
• Must be Phase 2 or later.
• Both minor companies must be owned by the same player.
• One minor company must be an associated Minor (M1, M5, M7, M8, M17, M18, M20).
• The other minor company must be an unassociated Minor (all other minor companies not named above).
• A regional railway may not be substituted for an unassociated minor company.
• The first minor company must be able to trace a route of any length from its home station token to the home station token of the other minor company_

In the game below, the operating order is: M6, M1, M9, M14, M21, M10, M8, M2, M17, M16, M18. The order of operations should be:

- Unassociated M6 should be skipped because it is not connected to an eligible associated minor
- Associated M1 should be skipped because it is not connected to an eligible unassociated minor
- Unassociated M9 is skipped because it is not connected to an eligible associated minor
- Unassociated M14 should be offered a merger with associated M8
- Unassociated M21 should be offered a merger with associated M17
- Unassociated M10 should be offered a merger with associated M16
- Associated M8 should be offered a merger with unassociated M14
- Unassociated M2 is skipped because it is not connected to an eligible associated minor
- Associated M17 should be offered a merger with unassociated M21
- Associated M16 should be offered a merger with unassociated M10
- Associated M18 should be skipped because it is not connected to an eligible unassociated minor

Screenshot 2024-04-22 at 18 00 07

michaeljb:
Fixes #10662

## Before clicking "Create"

### Explanation of Change

* a merger is still between an associated minor and an unassociated one connected to it, but can be initialized on either minor's turn

     * change `next_entity!` to work when the final minor in turn order for the round was already merged by the time the turn gets to them; with the old logic, the merger round would repeat instead of ending and moving on to the next round

        * the new 1822PNW fixture added here was affected by this problem

* a migration is needed for all 1822PNW games which have made it to a merger round where an unassociated minor should have had a turn; the migration will simply insert a `pass` action for the minor

     * add helper function to `migrate_game.rb` to improve entity comparison; false positive could occur with player ID of a small integer, and minors in 1822 games whose IDs are also small integers 

    * affected 1822PNW fixtures are migrated

### Any Assumptions / Hacks

* create `migrate_1822_pnw_merger.rb`; copied from `migrate_game.rb`, and modified to delete auto actions more aggressively

    * there's probably a smarter way to handle auto actions generally but I didn't want to focus on that right now

    * this may be applicable to the main migrate but I'm keeping it in the PNW script for now


scripts/migrate_1822_pnw_merger.rb
```diff
@@ -0,0 +1,389 @@
...
+  if pry_db
+    require 'pry-byebug'
+    binding.pry
+  end
+
+  # Generic handling for when a change just needs pass actions to be
+  # inserted/deleted
+
+  # try deleting auto_action pass
+  if game.round.actions_for(game.current_entity).include?(broken_action['type']) &&
```
michaeljb, line 103: This is the only change in this file from the normal `migrate_game.rb`

The check for the current action entities matching is removed, maybe it would be fine to remove in the base on as well.

michaeljb:
Running the migration locally worked fine against all active games (about 5 minutes, worked for all ~30 games), but I want to preserve finished games too. I kicked it off to run while I went to bed, and evidently things went to a crawl, it was only 335 games through the ~450, and it looked like pins hadn't been working. Not sure if it's my machine or the migration, going to close this while I investigate.

Wondering if I may just need to run the migration in smaller batches.

michaeljb:
My local migration worked for all 401 affected games out of 486 total 1822PNW games, and then all games pass validation.
It took 2.5 hours to run, but running migrations in smaller batches seems to run match faster--when migrating all at once, output for the first few games flew by, and then when I checked back later the pace had slowed down.

I'm not sure where in the stack the issue is--that is, if the problem is in docker/ruby/etc that would affect the 18xx.games server, or if it's something to do with Ubuntu running in WSL on my Windows machine.

When migrating the prod DB, I'll start with only active games, and then can migrate finished games in batches.

lib/engine/game/g_1822_pnw/step/merge.rb
```diff
@@ -46,7 +46,14 @@ def buyer
           def mergeable(corporation)
             return [] unless @merge_state == :none
 
-            @game.unassociated_minors.select do |m|
+            mergeable_minors =
+              if @game.associated_minor?(corporation)
+                @game.unassociated_minors
+              else
+                @game.associated_minors
+              end
```
crericha, line 54: Nit: this could be condensed to `@game.associated_minor?(corporation) ? @game.unassociated_minors : @game.associated_minors`

crericha (approved):

michaeljb:
Planning to merge this weekend when I have time to keep an eye on the migration.

PR #10883, opened 2024-06-06
[1870] implements Station Wars rulebook variant

philcampeau:
## Before clicking "Create"

## Implementation Notes

This variant makes the destination tokens on non-offboard spaces `cheater` tokens instead of `extra_slot` tokens. 

### Screenshots

from the rulebook:

image

### DIFF
--- a/lib/engine/game/g_1870/step/connection_token.rb
+++ b/lib/engine/game/g_1870/step/connection_token.rb
@@ -10,6 +10,8 @@ module Engine
         class ConnectionToken < Engine::Step::Token
           include Connection
 
+          STATION_WARS_CORPS = %w[GMO MP SP SSW TP].freeze
+
           def actions(_entity)
             %w[choose]
           end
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1870/step/connection_token.rb
```diff
@@ -39,7 +41,11 @@ def process_choose(action)
             destination.remove_assignment!(entity)
 
             if action.choice == 'Map'
-              destination.tile.cities.first.place_token(entity, token, free: true, extra_slot: true)
+              if @game.station_wars? && STATION_WARS_CORPS.include?(entity.id)
+                destination.tile.cities.first.place_token(entity, token, free: true, cheater: true)
+              else
+                destination.tile.cities.first.place_token(entity, token, free: true, extra_slot: true)
+              end
```
michaeljb, line 48: I find it more readable to have only one call with a lot of arguments, with the code inside of the conditional only saying what's different with the variant or not, something like this (I don't think `token_args` is a great name, please change it if you have a better idea):

```suggestion
              token_args =
                if @game.station_wars? && STATION_WARS_CORPS.include?(entity.id)
                  { cheater: true }
                else
                  { extra_slot: true }
                end
              destination.tile.cities.first.place_token(entity, token, free: true, **token_args)
```

philcampeau: cool suggestion! I went with token_placement_arg. submitted.

PR #10889, opened 2024-06-08
Enforce a train's availability when buying it

### LINKED ISSUE #9598: [1830] I bought 3T while 2Ts are still available
https://18xx.games/game/133915?action=103

I don't know how I was able to do this, but I bought the 3T while there are two 2Ts still available.

DO NOT PASTE THE GAME JSON! Please use [gist](https://gist.github.com/) instead.

If your game is "Hotseat", provide a link to your raw game data by using the `Tools` tab and `Copy Game Data`.

Paste that data [into a github gist](https://gist.github.com/) and share the resulting link which will look something like `https://gist.github.com/[your_name]/[some_letters]`.

-->

**How to Reproduce the Bug?**

**Which Browser / Platform?**

michaeljb:
Fixes #9598

Make sure a train is available when a player buys it. If not, throw an error.

In some cases, trains that were not at the top of the train depot were bought; sometimes this was the same rank of train, just a later than ID than expected, and sometimes it was the next rank of train before it should have been available, as described in the linked issue.

I manually migrated several fixtures, and had to replace a couple others. This isn't limited or correlated to any particular title, and I'm not sure the root cause, but if we have people hit this exception and start reporting that, hopefully we can figure out how to repro and get to the bottom of it.

I validated against all active games and got 9 failures from seven different titles. While I manually migrated some fixtures, I don't think it's worth the effort to get a migration script for this working. For finished games, 629 out of 23449 (2.7%) failed. I think pinning affected games would be fine.

## Before clicking "Create"


michaeljb:
A migration for this would be difficult because it's different from the normal migration script--it's not just adding/removing actions, but changing potentially many actions in a game.

If the trains of one rank are bought out of order, all actions in the game need to be loaded to tell what the correct order is, and then actions need to be carefully mutated. For instance, in one of the fixtures the last 3-train, `3-5`, was bought first, and the rest were bought in order; this meant that all actions with `"train": "3-5"` needed to be updated to `"train": "3-0"`, `3-0` became `3-1`, and so on, and collisions needed to be avoided while migrating. One annoyance with migrating these is that right now the only way to see the final set of actions for a game after undos/redos are cleaned up is from the game class, but when an error is hit due to an illegal train being bought the migration script would stop at the broken point. Logic to identify the undos/redos would need to be extracted from or duplicated outside of the game class.

Some games, like the one reported in #9598 actually had trains bought from a rank that should not have been available yet, so those would need to be pinned anyway.

michaeljb:
Going to pin these games - https://gist.github.com/michaeljb/6b80a20f2b0b164d9f84d399941c3784

PR #10896, opened 2024-06-09
[1822MX] include NdeM shares in liquidity calculation

### LINKED ISSUE #9821: [1822mx] Ndem share not being included in liquidity
https://18xx.games/game/138251?action=242

Not sure if it's because of the variant or if 22mx always does that.

michaeljb:
Fixes #9821

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1128,9 +1128,17 @@ module Engine
         return value_for_sellable(player, corporation) if self.class::PRESIDENT_SALES_TO_MARKET
 
         max_bundle = bundles_for_corporation(player, corporation)
-          .select { |bundle| bundle.can_dump?(player) && @share_pool&.fit_in_bank?(bundle) }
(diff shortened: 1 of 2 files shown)


lib/engine/game/base.rb
```diff
@@ -1128,11 +1128,19 @@ def value_for_dumpable(player, corporation)
...
       end
 
+      def can_dump?(entity, bundle)
+        if active_step.respond_to?(:can_dump?)
+          active_step.can_dump?(entity, bundle)
+        else
+          bundle.can_dump?(entity)
+        end
+      end
+
```
roseundy, line 1143: Yeah, this breaks 1841 because I already have a game method named can_dump? Either the 1841 or this has to change the name of the method.

roseundy: Probably the 1841 game method should be renamed since it uses different parameters than the step versions.

PR #10902, opened 2024-06-11
[1822PNW] stop requiring cash on hand to convert an associated minor to a major

### LINKED ISSUE #10899: [1822PNW] Cannot convert minor to major without cash on hand
GameID: 165500
Link: https://18xx.games/game/165500
Current Action

We are in Phase 6. Current player owns M1 and wishes to convert it to the CPR. M1 has a current share value of $100. Per rule 4.7.2, the owner of M1 should be able to just convert M1 into the presidency of the CPR. However, clicking on the presidency of the CPR does not display the buttons to choose a par price.

Current player has $9. If player sells a $100 share, then the par buttons will appear. If player sells a single $75 share the buttons will not appear, but they will if player sells two $75 shares.

michaeljb:
Fixes #10899

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1822_pnw/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1822_pnw/step/buy_sell_par_shares.rb
@@ -139,6 +139,12 @@ module Engine
           def selected_corporation
             @converting_major
           end
+
+          def can_buy?(entity, bundle, available_cash: nil)
+            return true if convertable_by?(bundle.corporation, entity)
+
+            super
+          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


djanyreason:
The issue identified in #10899 is still present in the linked game. I am proceeding assuming that the deployment can't fix that game - if it should have fixed the issue in that game, the link to the move is: https://18xx.games/game/165500?action=1026

michaeljb:
At https://18xx.games/game/165500?action=1026 I can click on the major and par buttons appear, if I clone the game there I can take the action as normal.

With the way that GitHub handles issues they get marked as fixed when the code is merged, but deploying to the site is a separate step, so I’m assuming you tried after the code was merged but before I finished the deployment last night.

Thanks for reporting the issue!

PR #10904, opened 2024-06-11
[1822 family] add private company colors

michaeljb:
match the physical games, where companies which can be acquired by minors are green, and companies which cannot are red; this is already done for 1822CA

## Before clicking "Create"


scottredracecar:
Can this be done globally for all 1822 games? I don’t see this feature
changing for potential future ones either.

michaeljb:
Good call, thanks Scott! #10908

PR #10907, opened 2024-06-11
[1822MX] include NdeM shares in liquidity calculation

### LINKED ISSUE #9821: [1822mx] Ndem share not being included in liquidity

michaeljb:
Fixes #9821 without causing #10900, which #10896 did

* adds an 1841 test that would have caught #10900 

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1125,9 +1125,17 @@ module Engine
         return value_for_sellable(player, corporation) if self.class::PRESIDENT_SALES_TO_MARKET
 
         max_bundle = bundles_for_corporation(player, corporation)
-          .select { |bundle| bundle.can_dump?(player) && @share_pool&.fit_in_bank?(bundle) }
+          .select { |bundle| can_dump?(player, bundle) && @share_pool&.fit_in_bank?(bundle) }
           .max_by(&:price)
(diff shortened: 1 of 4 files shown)


roseundy (approved):
LGTM

ollybh:
@michaeljb This broke the (pre-alpha) implementation of 18Ardennes. It caused a 'maximum stack size exceeded' error. This is because there is a step in 18Ardenees where `actions` calls `@game.liquidity`, and so the call to `active_step` in this new `can_dump?` function is causing an infinite loop.

This is easy enough to work around in 18Ardennes (I've just overridden `Game.can_dump?` so that it unconditionally calls `Bundle.can_dump?`), but is this implementation leaving a trap for the future? Might it be better to have a simpler implementation of `can_dump?` in the base class that could be overridden by 1822MX and any other games that need it?

From a quick look through the code base, it's possible that this would affect 18OE too, as that can call `liquidity` from an `actions` method.

michaeljb:
@ollybh yes that’s a good call, the base method should be the simpler version and 22MX can override to look in the step.

I can make the change (or review a PR for it if you beat me to it) tonight

PR #10928, opened 2024-06-16
[Autorouting] create groups of different trains for autorouting

### LINKED ISSUE #7982: [1822 family] Autorouting does not use E train properly
In a game of 1822MX I clicked on autorouting when I was going to run E train for first time.
6(+) train had a route, the P3/2 had a route, and autorouting filled in the E train.

image
https://18xx.games/game/87015?action=1254

But the chosen route (G22-F21-F23) is not a legal E route as F23 has no token!

And auto routing for E train does not seem to run on the same track as the other trains - which is allowed.

So this is an example where auto routing really misses the target.

Of course getting the right route when adding an E train is not difficult so I will manage.

michaeljb:
Trains within a group cannot have overlapping routes, but they can overlap with routes of trains in different groups. Trains not listed in `TRAIN_AUTOROUTE_GROUPS` end up in a default group.

* in 1822, E-trains can overlap with other trains; all others are automatically placed in another group
* in 18 Los Angeles, freight  (e.g., 3/5) trains can have routes overlapping with passenger trains

Fixes #7982 

## Before clicking "Create"

### DIFF
--- a/lib/engine/auto_router.rb
+++ b/lib/engine/auto_router.rb
@@ -14,13 +14,27 @@ module Engine
     end
 
     def compute(corporation, **opts)
+      trains = @game.route_trains(corporation).sort_by(&:price).reverse
+
+      train_groups =
+        if (groups = @game.class::TRAIN_AUTOROUTE_GROUPS)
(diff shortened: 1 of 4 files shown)


michaeljb:
needs more testing

PR #10930, opened 2024-06-16
[18 India] Add Commodity Icons and Bug Fixes

EdPovi:

## Implementation Notes
- Added commodity icons to hexes, refactor Jewelry to use assignment. 
- Edit IPO Row View to handle empty Rows and Corporate Stock Round
- Update Player Value to include corporation Book Value
- Add ability for corporations to convert Railroad Bonds in Phase IV
- May not visit MUMBAI or NEPAL twice 

### Explanation of Change
- Jewelry is allowed to be on the same hex as other commodities, therefore prior method of adding location name could have issues.  
- Added IPO Shares to cache method for shares, it wasn't storing all shares for action hash. 
- IPO Row needed to also create `CorporateBuyCompany` action
- Gauge Change border "Click" action referenced `hex` from border, but this reference can be `nil` after tile upgrade.  Using `@tile` for reference is more reliable. 
- Fixed bug if there were no bonds in Bank Pool. (nil vs empty list)
- Converting Bonds will also check if Manager has changed. 
- Fix bug if all players pass in selling round
- prevent GC border being deleted on tile upgrade.
- Fixed only allowed one phase IV train.
- Allow sale of Director share but may not sell Manager (0%) share


lib/engine/game/g_18_india/game.rb
```diff
@@ -233,9 +231,15 @@ def init_corporations(stock_market)
           end
         end
 
-        # Include bond shares for cache_objects
+        # Include IPO Pool shares and bond shares for cache_objects
         def shares
-          @corporations.flat_map(&:shares) + @players.flat_map(&:shares) + @share_pool.shares + gipr.bond_shares
+          LOGGER.debug "@corporations.flat_map(&:shares) => #{@corporations.flat_map(&:shares)}"
```
michaeljb, line 236: One note on these `LOGGER` calls, and I think I missed this in a prior PR, but the interpolated string should be passed in a block. This defers evaluating the string, meaning in environments where the string is not going to actually get logged, it never gets evaluated. I put some more info about this in #10946, and if we pull in my custom rubocop plugin you'd be able to have rubocop autocorrect these for you. So consider it optional to change them for now.

```suggestion
          LOGGER.debug { "@corporations.flat_map(&:shares) => #{@corporations.flat_map(&:shares)}" }
```

lib/engine/game/g_18_india/game.rb
```diff
@@ -641,17 +646,19 @@ def buyable_bank_owned_companies
         end
 
         def first_bond_in_bank
-          [] << bank.companies.find { |c| c.type == :bond }
+          [bank.companies.find { |c| c.type == :bond }].compact
```
michaeljb, line 649: This looks a little odd, maybe this instead?

```suggestion
          Array(bank.companies.find { |c| c.type == :bond })
```

lib/engine/game/g_18_india/game.rb
```diff
@@ -765,6 +772,29 @@ def gipr_may_operate?
...
+          # Check if if there is new Manager for GIPR
+          check_manager_change(entity, gipr) if entity.player?
+        end
+
+        def check_manager_change(entity, corporation)
+          return unless entity.percent_of(corporation) > gipr.owner.percent_of(corporation)
+
+          @share_pool.change_president(corporation.presidents_share, corporation.owner, entity)
+          corporation.owner = entity
+          @log << "#{entity.name} is the new Mangager of #{corporation.name}"
```
michaeljb, line 795:
```suggestion
          @log << "#{entity.name} is the new Manager of #{corporation.name}"
```

michaeljb (changes requested):
A few small things to change. Exciting that it's close to alpha ready!

lib/engine/game/g_18_india/game.rb
EdPovi: Many of these LOGGER statements are just for me to get some state information while I was tracking down a bug.  Or simply verify result of some code. 

How do I pull in your plugin?  I'll gladly auto correct them.

lib/engine/game/g_18_india/game.rb
EdPovi: Will the Array call remove the `nil` result if there isn't a bond?

lib/engine/game/g_18_india/game.rb
EdPovi: I have updated the LOGGER calls in this file.  I can also do the other, or wait to try using the RuboCop to auto fix them.

lib/engine/game/g_18_india/game.rb
michaeljb: Yep, `Array(nil)` returns `[]`:

```
> docker-compose exec rack bundle exec rake irb
irb -I lib/ -I assets/app/
irb(main):001:0> Array('nil')
=> ["nil"]
irb(main):002:0> Array(nil)
=> []
```

lib/engine/game/g_18_india/game.rb
michaeljb: To pull in the plugin you'd add it to your `Gemfile`, as done here - https://github.com/tobymao/18xx/pull/10946/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55f

Then run `bundle install` in your container (or simply kill the container and run `make dev_up_b` to rebuild it)

I think it's fine to wait until that's merged in. Thanks for updating the calls here!

michaeljb (approved):

PR #10936, opened 2024-06-18
[Core/1822MX] Simplify base implementation of Game.can_dump?

ollybh:
The implementation of `Game.can_dump?` introduced by tobymao#10907 broke the (pre-alpha) implementation of 18Ardennes. It caused a 'maximum stack size exceeded' error. This is because there is a step in 18Ardenees where `actions` calls `@game.liquidity`, and so the call to `active_step` in this new `can_dump?` function is causing an infinite loop.

To fix this, and avoid the same error being caused in future, this commit makes the base class implementation of `can_dump?` as simple as possible, just calling `Bundle.can_dump?`. The version of the method which checks `active_step` is moved into the 1822MX game code.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1131,11 +1131,7 @@ module Engine
       end
 
       def can_dump?(entity, bundle)
-        if active_step.respond_to?(:can_dump?)
-          active_step.can_dump?(entity, bundle)
-        else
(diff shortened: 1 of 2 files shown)


michaeljb (approved):

michaeljb:
Thanks!

PR #10943, opened 2024-06-19
Dev mode: compile only game files for requested games 

michaeljb:
After the issues with #10797 and #10810, I've used this branch for a few weeks to make sure it was stable before opening another PR for this.

Stop waiting for the files for all titles to build in local dev mode; only build for titles as needed. This reduces time for initial development setup, and reduces the pain for every `make clean`.

This changes functions in `lib/assets.rb` to use a `titles` param so that only the given titles are built, and change cached functions to cache things per-title when possible instead of caching the whole function call. In non-production mode, the API will trigger builds for game files that aren't already built.

It's best to view this diff with the "Hide whitespace" setting enabled. If the changes as a whole are confusing / difficult to review, looking at the individual commits one at a time might be easier.

## Before clicking "Create"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ group :development do
   gem 'rubocop-performance', require: 'false'
   gem 'sequel-annotate'
   gem 'stackprof'
+  gem 'tilt'
 end
 
 group :test do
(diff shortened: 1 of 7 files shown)


tobymao (approved):
nice work!

michaeljb:
Thanks!

Issue #10331, opened 2024-02-14
[18GB] cannot lay a tile, slow processing and/or rendering


michaeljb:
Discussed in the site chat this morning:

> Can anyone please tell me what is wrong with the following. Trying to place tile 24 on B25 in game https://18xx.games/game/151565
>
> it does let me when I clone that game
though the game is running *super* slowly on my browser, takes forever to load
>
>Thanks it is running slow here too and keeps refreshing and losing the tile

Haven't had time to profile or do a cursory investigation of whether the slowness is the game processing or the rendering.

Copied the game data to [this gist](https://gist.github.com/michaeljb/3d738a10ccf9965a42f91b6134253b0c)

Maybe similar? #10123

michaeljb:
Firefox profiling shows the slowness coming from computing the graph during rendering, starting from the `connected_hexes` call in `tracker_available_hex`:

https://github.com/tobymao/18xx/blob/3794ce0dd/lib/engine/step/tracker.rb#L494

From there it's a pretty deep stack of `walk`ing the graph

![Screenshot 2024-02-17 092523](https://github.com/tobymao/18xx/assets/1045173/ab3652bc-7093-4345-84f9-0b1a175f91de)

michaeljb:
https://github.com/tobymao/18xx/blob/3794ce0ddcfbb174d74b1d3ef609e0a0ead633aa/lib/engine/step/tracker.rb#L494

https://github.com/tobymao/18xx/blob/3794ce0ddcfbb174d74b1d3ef609e0a0ead633aa/lib/engine/step/tracker.rb#L468

https://github.com/tobymao/18xx/blob/3794ce0ddcfbb174d74b1d3ef609e0a0ead633aa/lib/engine/graph.rb#L91

https://github.com/tobymao/18xx/blob/3794ce0ddcfbb174d74b1d3ef609e0a0ead633aa/lib/engine/graph.rb#L234-L252

https://github.com/tobymao/18xx/blob/3794ce0ddcfbb174d74b1d3ef609e0a0ead633aa/lib/engine/part/node.rb#L39

michaeljb:
The problem is something to do with the deletion on the converging check from #7788, removing these lines lets my browser compute the graph in about 0.02 seconds instead of 55 seconds:

https://github.com/tobymao/18xx/blob/5df8d6f3e747faa236ae829063270a2359ea2e46/lib/engine/part/node.rb#L86

[...]

PR #10957, opened 2024-06-24
[1822CA] don't recompute `destination_node_check?`

### LINKED ISSUE #10956: [1822CA] Game Won't Let Major Lay Second Yellow Track
Game ID  164866

https://18xx.games/game/164866

ICR was floated in SR 10.  In its operation in OR 10.1 "ICR lays tile #8 with rotation 0 on AP2" just fine, but for each of the several attempts (more than 5) to lay a second yellow tile, the server takes a long time to "accept" the move to advance to the token placement phase.  The wait time between clicking the green check mark to commit the second tile lay to advancing to token placement was so long that two page unresponsive messages offering to close the application or wait for it to respond appeared.  After each of the several times after each of the times the game did move beyond this point to letting the ICR place a token, buy a train, redeem a share, and appear to start the turn of the next RR, red messages were given that updates were not accepted due to refreshing and something like " can't process move 1590".

Here's a screen shot of the attempted tile lay:

image

I didn't seem to experience any trouble in other games taking turns immediately before or after this trouble.  I had taken turns in at least two games immediately before this problem.  I made several attempts in this game as described above.  I started this bug report and copied it into a document.  I closed and reopened my browser.  I made more attempts as described above.  I took turns in other games with no apparent problems.  I rebooted my computer, and I continued to have this problem.

michaeljb:
Closes #10956, but a lot of that issue is caused by #10331.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1822_ca/step/destination_token.rb
+++ b/lib/engine/game/g_1822_ca/step/destination_token.rb
@@ -11,6 +11,7 @@ module Engine
 
           def setup
             @connected_destination_nodes = []
+            @dest_checked = false
           end
 
           def actions(entity)
@@ -80,6 +81,8 @@ module Engine
           end
 
           def destination_node_check?(entity)
+            return @connected_destination_nodes if @dest_checked
+
             destination_hex = @game.hex_by_id(entity.destination_coordinates)
             home_node = entity.tokens.first.city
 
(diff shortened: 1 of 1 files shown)


michaeljb:
Reverted this in https://github.com/tobymao/18xx/commit/29f45c5fd8615d63ac301b05d7f7957e70b5a09b to fix 10960

lib/engine/game/g_1822_ca/step/destination_token.rb
```diff
@@ -80,6 +81,8 @@ def process_place_token(action)
           end
 
           def destination_node_check?(entity)
+            return @connected_destination_nodes if @dest_checked
```
michaeljb, line 84: This isn't the same thing normally returned by this function.

Fixed in #10964

PR #10958, opened 2024-06-24
[System18] dividend step: `puts` -> `LOGGER.debug`

michaeljb:


roseundy (approved):
This is fine. Or delete the puts. Either way.

PR #10966, opened 2024-06-25
[18Norway] Ø may pass tokened out city

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -479,6 +479,40 @@ module Engine
           mjosa.lay(tile_by_id('LM2-0'))
         end
 
+        def compute_stops(route)
+          stops = super
+          if @round.train_upgrade_assignments[route.train]
+            skipped = skipped_stop(route, stops)
+            stops.reject! { |stop| stop == skipped } if skipped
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -479,6 +479,40 @@ def event_lm_brown!
...
+
+          # Skipping is optional - if we are using STRICTLY fewer stops than distance (jumping adds 1) we don't need to skip
+          return nil if counted_stops.size < route.train.distance
+
+          # Count how many of our tokens are on the route; if only one we cannot skip that one.
+          tokened_stops = counted_stops.select { |stop| stop.tokened_by?(route.train.owner) }
+          counted_stops.delete(tokened_stops.first) if tokened_stops.one?
+
+          # Find the lowest revenue stop that can be skipped
+          counted_stops.max_by { |stop| revenue_for(route, stops.reject { |s| s == stop }) }
```
michaeljb, line 513: The comment says lowest revenue, should these be `min_by`?

```suggestion
          counted_stops.min_by { |stop| revenue_for(route, stops.reject { |s| s == stop }) }
```

PR #10967, opened 2024-06-25
[18Norway] No sales during first SR, then p_any_operational

patrikolesen:
## Before clicking "Create"


lib/engine/game/g_18_norway/game.rb
```diff
@@ -539,6 +539,19 @@ def emergency_issuable_bundles(corp)
 
           []
         end
+
+        def sell_movement(corporation = nil)
+          return self.class::SELL_MOVEMENT if corporation.nil?
```
michaeljb, line 544: can avoid `nil?` call:

```suggestion
          return self.class::SELL_MOVEMENT unless corporation
```

patrikolesen: Looks better

michaeljb (approved):

PR #10970, opened 2024-06-25
[18Norway] allow for owner sell_shares

patrikolesen:
## Before clicking "Create"


patrikolesen:
Updated the implementation

PR #10971, opened 2024-06-25
[18Norway] Handle when P3 is closed

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/steps/build_tunnel.rb
+++ b/lib/engine/game/g_18_norway/steps/build_tunnel.rb
@@ -50,8 +50,8 @@ module Engine
             entity.spend(cost, @game.bank) if cost.positive?
 
             factory_owner = @game.company_by_id('P3').owner
-            @game.bank.spend(10, factory_owner)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_norway/steps/build_tunnel.rb
```diff
@@ -50,8 +50,8 @@ def build_tunnel(entity, cost)
             entity.spend(cost, @game.bank) if cost.positive?
 
             factory_owner = @game.company_by_id('P3').owner
-            @game.bank.spend(10, factory_owner)
-            @log << "#{factory_owner.name} receives #{@game.format_currency(10)} for building a tunnel"
+            @game.bank.spend(10, factory_owner) unless factory_owner.nil?
+            @log << "#{factory_owner.name} receives #{@game.format_currency(10)} for building a tunnel" unless factory_owner.nil?
```
michaeljb, line 54: `nil?` calls can be avoided:

```suggestion
            @game.bank.spend(10, factory_owner) if factory_owner
            @log << "#{factory_owner.name} receives #{@game.format_currency(10)} for building a tunnel" if factory_owner
```

PR #10985, opened 2024-06-29
[Core/18Ardennes] Allow notes to be shown on corporation cards

ollybh:
## Implementation notes / explanation of change

This allows arbitrary text to be added to a corporation card.

This is useful in 18Ardennes, where there are auction rounds for the rights to start public companies (corporations). When placing a bid the player must specify which minor company they will be using to start the public company. In the stock round after the auction round these notes are used to show which minor has been pledged to start each new public company. Without this you had to scroll through the game log to see which bids had been placed.

#### The BY and N public companies with added notes:
image

### Any Assumptions / Hacks

This uses a `@game.respond_to?(:corporation_notes)` test. It would also be possible to add a `corporation_notes` method to the base Corporation class that returns nil.

## Before clicking "Create"

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -96,6 +96,7 @@ module View
         end
         extras << render_capitalization_type if @corporation.corporation? && @corporation.respond_to?(:capitalization_type_desc)
         extras << render_escrow_account if @corporation.corporation? && @corporation.respond_to?(:escrow) && @corporation.escrow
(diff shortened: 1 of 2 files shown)


michaeljb:
Is this something you could just use the base method `status_array(corporation)` for?

ollybh:
> Is this something you could just use the base method `status_array(corporation)` for?

Yes, that's exactly the method I need. I had missed this one when poking around `View::Game::Corporation`.

PR #10987, opened 2024-06-30
18Rhl: fix typo in name of company No. 3

bsnqrk:
## Before clicking "Create"

- ~~[ ] Add the `pins` or `archive_alpha_games` label if this change will break existing games~~

### Explanation of Change
just a typo.


perwestling (approved):
Good catch

PR #10990, opened 2024-06-30
[18Norway] Updated harbor

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -104,7 +104,7 @@ module Engine
           @nsb ||= corporation_by_id('NSB')
         end
 
-        def harbor_city_id_by_harbor_id(hex_id)
+        def harbor_city_coordinates(hex_id)
           CITY_HARBOR_MAP.key(hex_id)
         end
 
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -104,6 +104,14 @@ module Engine
           @nsb ||= corporation_by_id('NSB')
         end
 
+        def harbor_city_id_by_harbor_id(hex_id)
+          CITY_HARBOR_MAP.key(hex_id)
+        end
+
+        def harbor_hex?(hex)
+          HARBOR_HEXES.include?(hex.id)
+        end
+
         def price_movement_chart
           [
             ['Action', 'Share Price Change'],
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -104,7 +104,7 @@ module Engine
           @nsb ||= corporation_by_id('NSB')
         end
 
-        def harbor_city_coordinates(hex_id)
+        def harbor_city_id_by_harbor_id(hex_id)
           CITY_HARBOR_MAP.key(hex_id)
         end
 
--- a/lib/engine/game/g_18_norway/steps/token.rb
+++ b/lib/engine/game/g_18_norway/steps/token.rb
@@ -8,10 +8,25 @@ module Engine
       module Step
         class Token < Engine::Step::Token
           def available_hex(entity, hex)
-            return true if super(entity, hex)
-            return true if @game.ferry_graph.reachable_hexes(entity)[hex]
+            return true if super
 
-            false
+            check_available_harbour(entity, hex)
+          end
+
+          def tokener_available_hex(entity, hex)
+            return true if super
+
+            check_available_harbour(entity, hex)
+          end
+
+          def check_available_harbour(entity, hex)
+            city = @game.harbor_city_id_by_harbor_id(hex.id)
+            return false if city.nil?
+
+            other_hex = @game.hex_by_id(city)
+            return true if @game.graph.reachable_hexes(entity)[other_hex]
+
+            @game.ferry_graph.reachable_hexes(entity)[other_hex]
           end
 
           def can_place_token?(_entity)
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_18_norway/steps/token.rb
```diff
@@ -21,24 +36,48 @@ def can_place_token?(_entity)
...
+              raise GameError, "Cannot place token on #{hex.name}#{city_string} because it is not connected"
             end
             @game.clear_graph
             pass!
           end
 
-          def check_connected(entity, city, hex)
-            return if @game.loading || @game.token_graph_for_entity(entity).connected_nodes(entity)[city]
-            return if @game.loading || @game.ferry_graph.connected_nodes(entity)[city]
+          def place_token(entity, city, token, connected: true, extra_action: false,
```
ollybh, line 58: I don't like all this code cut'n'pasted into this method. I'm not sure I've understood the rules correctly, but is the code to allow placing port tokens in the sea hexes with the port symbols? It might be possible to have this as hex tokens, so instead of having to reimplement this entire method just to set `extra_slot`, you could do something like:
```
          def place_token(entity, city, token, connected: true, extra_action: false,
                          special_ability: nil, check_tokenable: true, spender: nil, same_hex_allowed: false)
            if @game.harbor_hex?(hex)
              # will need to check connection and pay cost
              hex.place_token(hex)
            else
              super
            end
          end
```

patrikolesen: I have tried to find another way to do this but have not found something that seems to be better so if you find another way then I am all for it.

One problem that might occur with hex_token is that it might be very crowded on the the hex.
Since all companies are allowed to put a harbor token in the harbor as long as they can trace to it.

The rules are that you are using the harbor tokens to "open" up for ships to trace a route through the port. The harbor token does not have any effect on the trains.

You can trace to a harbor either by train to the port or via the ships.

lib/engine/game/g_18_norway/game.rb
```diff
@@ -104,6 +104,14 @@ def nsb
           @nsb ||= corporation_by_id('NSB')
         end
 
+        def harbor_city_id_by_harbor_id(hex_id)
```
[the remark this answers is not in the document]
patrikolesen: Updated

lib/engine/game/g_18_norway/steps/token.rb
```diff
@@ -8,10 +8,25 @@ module G18Norway
...
+            check_available_harbour(entity, hex)
+          end
+
+          def tokener_available_hex(entity, hex)
+            return true if super
+
+            check_available_harbour(entity, hex)
+          end
+
+          def check_available_harbour(entity, hex)
```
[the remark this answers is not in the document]
patrikolesen: Updated

lib/engine/game/g_18_norway/steps/token.rb
```diff
@@ -8,10 +8,25 @@ module G18Norway
...
+            city = @game.harbor_city_id_by_harbor_id(hex.id)
+            return false if city.nil?
+
+            other_hex = @game.hex_by_id(city)
+            return true if @game.graph.reachable_hexes(entity)[other_hex]
+
+            @game.ferry_graph.reachable_hexes(entity)[other_hex]
```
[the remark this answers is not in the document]
patrikolesen: Updated

lib/engine/game/g_18_norway/steps/token.rb
patrikolesen: image

ollybh: OK, that makes sense.
Instead of copying most of the `place_token` code into the 18Norway code, it might be better to add an `extra_slot` named parameter to `Engine::Step::Tokener.place_token` with a default value of `false`.

ollybh: @patrikolesen Have you had a chance to look at making this change?

patrikolesen: Sorry got caught in day work
But I finally got around to it and I agree that was a good improvment

Should I rebase to latest master?

patrikolesen:
The harbor logic where broken, it was possible to place a token in any open city

lib/engine/game/g_18_norway/steps/token.rb
patrikolesen: It is still broken

patrikolesen: Sorry for the mess, but now it should be ok to review again

patrikolesen:
Sorry for the mess, but now it should be ok to review again

lib/engine/game/g_18_norway/steps/token.rb
```diff
@@ -9,37 +9,44 @@ module Step
...
+            return true if harbor_available?(entity, hex)
 
             false
           end
 
+          def harbor_available?(entity, hex)
+            city = @game.harbor_city_coordinates(hex.id)
+            return false if city.nil?
+
+            other_hex = @game.hex_by_id(city)
```
ollybh, line 22: I'd avoid calling a variable `city` unless you're referring to a City object. Maybe `city_hex` would be better?
Or (because the various `*_by_id` functions are implemented by hashes, you can safely pass nil or garbage values to them and expect to get nil back) you could do:
```suggestion
            other_hex = @game.hex_by_id(@game.harbor_city_coordinates(hex.id))
            return false unless other_hex

```

lib/engine/game/g_18_norway/steps/token.rb
```diff
@@ -9,37 +9,44 @@ module Step
...
-
-            unless connected_city
+            connected_city = @game.token_graph_for_entity(entity).connected_nodes(entity)[city]
+            if connected_city
+              place_token(entity, action.city, action.token)
+            elsif harbor_available?(entity, city.hex)
               @game.abilities(entity, :token) do |ability|
                 place_token(entity, action.city, action.token, connected: false, special_ability: ability)
+                entity.add_ability(Ability::Token.new(type: 'token', hexes: Engine::Game::G18Norway::Game::HARBOR_HEXES,
+                                                      from_owner: true, discount: 0, connected: true))
```
ollybh, line 42: Instead of adding a new token ability, could you give the original one a `count` attribute that's large enough for it to be reused throughout the game?

lib/engine/game/g_18_norway/steps/token.rb
```diff
@@ -9,37 +9,44 @@ module Step
         class Token < Engine::Step::Token
           def available_hex(entity, hex)
             return true if super(entity, hex)
-            return true if @game.ferry_graph.reachable_hexes(entity)[hex]
+
+            return true if harbor_available?(entity, hex)
 
             false
```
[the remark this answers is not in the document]
patrikolesen: Updated

lib/engine/game/g_18_norway/steps/token.rb
patrikolesen: This is not trivial since tokener.rb does not use the ability but remove it instead.

I have tried to change it but am afraid on how that will affect other games that uses ability for token
I have seen a problem with 1822CA and P10 which then would need an update to special_token
Would it be ok if I did this in another commit, want to make sure that I do not affect any other game to solve this

ollybh: Yes, putting this in a separate commit would be sensible.

ollybh (approved):

PR #10991, opened 2024-06-30
[18Norway] Update cert limit

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -197,6 +197,7 @@ module Engine
           @all_tiles.each { |t| t.ignore_gauge_walk = true }
           @_tiles.values.each { |t| t.ignore_gauge_walk = true }
           @hexes.each { |h| h.tile.ignore_gauge_walk = true }
+          update_cert_limit
 
           # Allow to build against Mjosa
           hex_by_id('H26').neighbors[1] = hex_by_id('G27')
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -483,6 +484,17 @@ def check_route_token(route, token)
           raise NoToken, 'Route must contain token' unless token
         end
 
+        def update_cert_limit_to(new_cert_limit)
+          @cert_limit = new_cert_limit
+          @log << "Certificate limit is now #{@cert_limit}"
+        end
+
+        def update_cert_limit
+          nr = @corporations.sum(0) { |corporation| nationalized?(corporation) ? 1 : 0 }
```
ollybh, line 493:
```suggestion
          nr = @corporations.count(&:nationalized?)
```

ollybh (changes requested):

michaeljb (approved):

lib/engine/game/g_18_norway/game.rb
michaeljb: Whoops, I was too hasty in adding this. I agree that `count` is the right call, but this syntax only works if `nationalized?` is a method of `corporation`.

Fixed in 5f61f40cb864eaefd877492b5838661c8af5a7f6

ollybh: Oops, I must have been half asleep when reviewing this. The version in 5f61f40 looks good to me.

PR #10998, opened 2024-07-02
[1822PNW] fix backroom negotiations targets

### LINKED ISSUE #10996: [1822pnw] Invalid companies available for Backroom Negotations
https://18xx.games/game/165580?action=923

When attempting to use Backroom Negotations, this game offers 7 and 5 as options. These companies were floated directly and their associated minors discarded from the bid box. When attempting to choose them, it gives an error about nil not being an invalid option.

I suspect this error came in when fixing a recent bug with backroom negotations that didn't allow you to choose associated minors that had fallen out of the bid boxes and were removed; it looks like it needs to make sure those companies weren't floated.

michaeljb:
Fixes #10996

Pins not needed as invalid targets were presented to the user, but selecting one caused an error instead of a rules-breaking game state, as reported in the issue.

## Before clicking "Create"


lib/engine/game/g_1822_pnw/step/acquire_company.rb
```diff
@@ -95,21 +95,21 @@ def process_choose(action)
...
               major = @game.corporation_by_id(major_id)
 
-              targets << minor if !company.owner && company.closed? &&
-                                  !minor.owner && minor.closed? &&
-                                  !major.owner && !major.floated?
-            end
+              already_discarded = (!company.owner && company.closed? && minor.closed?)
+
+              minor_available = !minor.owner &&
+                                (already_discarded || !bidbox_corporations.include?(minor))
```
ollybh, line 108: Would this give the same result? It would avoid the need to create the `bidbox_corporations` set.
```suggestion
              minor_available = !minor.owner &&
                                (already_discarded || !@game.bidbox_minors.include?(company))
```

michaeljb: Done in 61e8975a51d6b5f708c47991cf3f830ca140482a.

`@game.bidbox_minors` recomputes the minors every time, so I kept the call to it outside of the loop here.

PR #11002, opened 2024-07-02
[1826] entities tweaks

philcampeau:
## Before clicking "Create"

### Explanation of Change

mostly formatting changes, plus explicitly setting the share distribution for 5-share corps


philcampeau:
I thought I had created two separate branches for these changes, but it looks like both commits are here. That's fine too. 

The first commit was removing the 'can_buy_trains' status, since I had also implemented it as an event, and having both was redundant.

philcampeau:
Added third commit which modifies the 6 corps in Paris to include an ability that reserves the hexes surrounding Paris by borrowing the code from 1822 which does the same thing for london.

ollybh (approved):

PR #11008, opened 2024-07-04
S18: UK map updates

### LINKED ISSUE #10727: System18 UK Limited Playtesting Changes
Some rule changes post-TraXX playtesting.

**Mechanism for changing cities**
Add a rule that the cities that are currently single dot-towns are changed to an outlined circle with no fill (I am planning to make them about the same size and line weight as a city, but with no white fill). When tiles are built in these hexes, the player laying the tile may choose to place it as either a city or plain tile and the hex retains that type for the rest of the game, upgrades are as usual.

**Change rusting/round schedule**
Remove the rule about automatic phase changes on a schedule. 

Trains (still) do not rust based on new train purchases.

There is a new tracker for the rusting schedule. At the end of each OR, increment the rusting round tracker. First and second ORs have no effects. 
- At the end of the third OR, the 2-trains rust.
- At the end of the fourth OR, the 3-trains rust.
- At the end of the fifth OR, the 4-trains rust.

The phases and number of ORs per set is done as usual (one OR until Phase 3, then two ORs for the rest of the game)

**Game End**
The game end condition is changed to now be as in the base rules (after set when D is bought, then run one additional set).

roseundy:
Fixes #10727 

Major changes for the UK Limited map. Any existing games with that map will need to be archived.

## Implementation Notes

This adds a new option to city rendering: `outline`. This draws an unfilled circle for cities. Needed for the UK Limited map.

This changes:
lib/engine/tile.rb
lib/engine/part/city.rb
assets/app/view/game/part/city_slot.rb

### Screenshots
city_outline


ollybh:
There's no code included in the pull request that does anything different for the outlined cities, apart from their rendering. Has a file been missed? Or is this going in a different PR?

roseundy:
No, it's all there. I added a new keyword: "outline" and the rendering for it. There is map-specific code that uses that keyword for modifying upgrades.

ollybh:
My mistake, I hadn't realised that `upgrade_ignore_num_cities` was an existing method that's called from the core code.
LGTM.

ollybh (approved):

PR #11024, opened 2024-07-08
[1826] implements logic for e-trains

philcampeau:
## Before clicking "Create"

### Explanation of Change

Game has E-trains which run a variable distance based on the number of that train currently sold. I'm tracking this in an attr that is incremented through the 'buy_train_action(action)' method in the buy_trains.rb file of the game.

### DIFF
--- a/lib/engine/game/g_1826/game.rb
+++ b/lib/engine/game/g_1826/game.rb
@@ -9,7 +9,7 @@ module Engine
   module Game
     module G1826
       class Game < Game::Base
-        attr_reader :recently_floated, :can_buy_trains
+        attr_reader :recently_floated, :can_buy_trains, :e_trains_purchased, :e_train_range, :tgz_purchased
 
         include_meta(G1826::Meta)
         include Entities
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1826/game.rb
```diff
@@ -64,68 +64,73 @@ class Game < Game::Base
...
+            },
+            {
+              name: 'E',
+              # distance is equal to the number of E and TGV trains in play. The run is doubled until a TVG is purchased.
+              distance: [
+                {
+                  'nodes' => %w[city offboard],
+                  'pay' => @e_train_range,
+                  'visit' => @e_train_range,
+                  'multiplier' => @tgv_purchased ? 1 : 2,
```
ollybh, line 116: Have you tested this? I would have thought that these variables would be evaluated when the train object is first created, not when the train is being used.

lib/engine/game/g_1826/game.rb
```diff
@@ -203,6 +215,16 @@ def event_remove_abilities!
...
           end
         end
+
+        def event_can_buy_trains!
+          @can_buy_trains = true
+          @log << 'Corporations can buy trains from other corporations'
+        end
+
+        def event_tgz_purchased!
+          @tgz_purchased = true
```
ollybh, line 225: Do you need to set `@e_train_range` here?

michaeljb:
After merging #11002, this branch has conflicts.

PR #11028, opened 2024-07-09
[1849 K2S] fix Societa Marittima description

### LINKED ISSUE #10775: [1849: Kingdom of the Two Sicilies] P4 Private text problem
The P4 private (SMS) allows a tile lay/upgrade and token lay anywhere on the map, for the owning corp.  It says this only works on coastal hexes, mentioning that's all hexes **except Foggia, Campobasso, and Potenza**. But L'Aquila (Hex A9) should also be mentioned in the list. 

L'Aquila isn't on the coast in the real world. And accordingly, in the current implementation, L'Aquila isn't a possible target for P4. It just so happens to be on the edge of the 1849: KotS map, so one might think it was coastal at a quick glance. 

**The text of P4:**
> 
> Società Marittima Siciliana
> During its operating turn, the owning corporation may close this private company in lieu of performing both its tile and token placement steps. Performing this action allows the corporation to select any coastal city hex (all cities except Foggia, Campobasso, and Potenza), optionally lay or upgrade a tile there, and optionally place a station token there. This power may be used even if the corporation is unable to trace a route to that city, but all other normal tile placement and station token placement rules apply.

**A place to verify that P4 is working correctly:** 

In this game, M&C owns SMS (P4). L'Aquila is green and has a spot for a token (even without a tile action), but M&C can't token there. There's no error or bug here, it's just an example highlighting the situation. The only issue is with the list of excluded cities in the Private's textbox. 
https://18xx.games/game/160856?action=286

philcampeau:
Fixes #10775 

## Before clicking "Create"

### Explanation of Change

Adds "L'Aquila" to the description of this private, since even though this city is not a coastal city, it is on the edge of the map and it's the only city that is neither a valid target for the private ability nor mentioned in the exceptions. 

This makes it clearer.

### DIFF
--- a/lib/engine/game/g_1849_boot/entities.rb
+++ b/lib/engine/game/g_1849_boot/entities.rb
@@ -66,7 +66,7 @@ module Engine
                   'this private company in lieu of performing both its tile '\
                   'and token placement steps. Performing this action allows '\
                   'the corporation to select any coastal city hex (all cities '\
-                  'except Foggia, Campobasso, and Potenza), optionally lay or '\
+                  "except Foggia, Campobasso, L'Aquila, and Potenza), optionally lay or "\
                   'upgrade a tile there, and optionally place a station token '\
                   'there. This power may be used even if the corporation is '\
                   'unable to trace a route to that city, but all other normal '\


ollybh (approved):

yoyoson:
🎉 👍

PR #11029, opened 2024-07-10
[1830] implements private company cross-buys

### LINKED ISSUE #9735: 1830 sales of private companies between players
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

philcampeau:
Fixes #9735 

## Before clicking "Create"

### Explanation of Change

Both the Avalon Hill and Mayfair rules for 1830 say that players can buy private companies from each other in any stock round after SR1. This implements the ability. 

I tested with several live games, but it would be a good idea to check for pins, in the case where players who had no possible actions in subsequent SRs were automatically skipped, which wouldn't happen with this code.

### DIFF
--- a/lib/engine/game/g_1830/game.rb
+++ b/lib/engine/game/g_1830/game.rb
@@ -109,14 +109,20 @@ module Engine
           ['', '', '', '10b', '20b', '30b', '40o'],
         ].freeze
 
-        PHASES = [{ name: '2', train_limit: 4, tiles: [:yellow], operating_rounds: 1 },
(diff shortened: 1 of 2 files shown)


philcampeau:
solid point! Thanks for that. 

I don't think the game needs any additional checks in the code for this though, correct? I think the `def purchaseable_companies` check covers all that. 

I'll add the pins tag. Let me know if you have any other comments.

michaeljb (approved):

PR #11035, opened 2024-07-15
[18 India]  Bug Fixes and Prepare for Alpha

EdPovi:

## Implementation Notes
- Bug fixes found during testing in hot seat and server mode.
- Added draft history to replace player hand information in Timeline Section
- Set development stage to Alpha

### Explanation of Change
- Modified timeline information to be a static string created during the draft to show history
- refactored `commodity_bonus` to have separate return when commodities used is needed to ensure correct commodies for route during processing.
- Added additional checks for cert limits in corporate stock round. 
- Changed `show_hand` to `show_sr_hand` to prevent entanglement of similar button on player card.  It was showing a hidden hand of other players when it wasn't your turn.


lib/engine/game/g_18_india/game.rb
```diff
@@ -1104,27 +1110,27 @@ def visit_jewelery(route)
...
 
           revenue = 0
-          @round.commodities_used = []
+          commodities_used = []
           commodity_sources.each do |source|
             bonus = COMMODITY_BONUSES[source]
             LOGGER.debug { "GAME.commodity_bonus >> bonus: #{bonus}" }
-            if visited_names.intersect?(bonus['locations'])
+            if visited_names.intersect?(bonus[:locations])
               revenue += bonus['value'] || visited_names.map { |loc| SPICE_BONUSES[loc] || 0 }.max
```
ollybh, line 1125: The `bonus` hash key needs to be changed to a label here as well.
```suggestion
              revenue += bonus[:value] || visited_names.map { |loc| SPICE_BONUSES[loc] || 0 }.max
```

lib/engine/game/g_18_india/game.rb
```diff
@@ -1104,27 +1110,27 @@ def visit_jewelery(route)
           route.all_hexes.include?(@jewlery_hex) ? ['JEWELRY'] : []
         end
 
-        def commodity_bonus(route, _stops = nil)
+        def commodity_bonus(route, _stops = nil, report_commodities: false)
```
ollybh, line 1113: This feels a bit unclean having the method have two different types of information returned, determined by the `report_commodities` parameter.

Could this be rewritten as two separate methods?

- `commodities_visited` which contains most of this method, apart from the revenue calculation, and returns the array currently collected in `commodities_used`.
- `commodity_bonus` which iterates over the array returned by `commodities_visited` and calculates the bonus.

Or, if you're better at Ruby than I am, you could write a single method that takes a `block` argument and `yield`s it it to either calculate the revenue or return the array of commodities.

lib/engine/game/g_18_india/step/draft.rb
```diff
@@ -75,6 +75,8 @@ def process_bid(action)
             player.hand << company
             player.hand.sort_by! { |item| [item.name, -item.value] }
 
+            player.draft_history = player.draft_history + company.name + (company.type == :president ? '(Dir), ' : ', ')
```
ollybh, line 78: This is going to give a trailing comma (and space) on the draft history string. Instead you could define `draft_history` as an empty array in the `Player.initialize` method, then here add an item to the array:
```suggestion
            player.draft_history << company.name + (company.type == :president ? '(Dir)' : '')
```
And in the `Game.timeline` method join the array items together:
```
timeline << "#{p.name}: #{p.draft_history.join(', ')}"
```

ollybh (changes requested):

lib/engine/game/g_18_india/game.rb
EdPovi: I'm not an expert in Ruby as this was the only way I considered to have a single method yield both results.  Unless I have the result return an array, hash, or tuple.  I.e, I could have it return `[commodities_used, revenue] `and would need to make sure the correct element is used by the caller.

EdPovi: Would it be cleaner to have the method return both values in an array?  Or have separate methods for each return value, but the contents would be very similar.

EdPovi: The Commodity bonus method would also need the array of visited locations due to the Spice payout depends on where it goes.  Also the source list isn't the same as the commodity list as there are multiple sources for the same commodity.

ollybh: Ruby lets you pass a block to a method, which can then be invoked inside the method using the `yield` keyword.

This would let you write the core iteration that calculates which commodities are on the route once, and then pass a block that either calculates the commodity bonuses, or claims the concessions.

Extremely unreliable code follows:
```
# This is the outer bit of the existing `commodity_bonus` method, but instead of
# calculating the revenue and saving the commodity name when we find a commodity
# location, we instead `yield` to the block passed to the method, passing the block a couple
# of parameters (the `bonus` hash and the `visited_names` array).
def route_commodities(route)
  visited_names = (route.all_hexes.map(&:location_name) + visit_jewelery(route)).compact
  corporation = route.train.owner
  commodity_sources = visited_names & available_commodities(corporation)

  commodity_sources.each do |source|
    bonus = COMMODITY_BONUSES[source]
    yield bonus, visited_names if visited_names.intersect?(bonus[:locations])
  end
end

# This method returns the bonuses.
# I've removed the `_stops` parameter as it didn't seem to be used anywhere.
def commodity_bonus(route)
  revenue = 0
  route_commodities(route) do |bonus, visited|
    revenue += bonus[:value] || visited.map { |loc| SPICE_BONUSES[loc] || 0 }.max
  end
  revenue
end

# The only place you set the `report_commodities` parameter was in `process_run_routes`.
# You could call this method from inside the `routes.each` iterator to claim concessions.
# `routes.each { |route| claim_concessions(entity, route, ability) }`
def claim_concessions(entity, route, ability)
  route_commodities(route) do |bonus, _|
    commodity = bonus[:commodity]
    @log << "#{entity.name} delivered #{commodity}"
    @game.claim_concession(entity, commodity) unless ability.description.include?(commodity)
  end
end
```

ollybh: I'll be amazed if that code works (`yield` makes my brain hurt), but I hope it's close enough to give you an idea as to how this could be done.

EdPovi: That is very interesting stuff.  And I may try it out (sometime later) as an educational exercise.  But help with the idea of cleaning up the current method, I'm looking into just two separate methods at the moment....

EdPovi: I tested your code using `yield` and it seems to work well!  I will run some more tests just to confirm...  :)

EdPovi: Thank you for providing this suggestion!  I have added your code and tested it against an existing json I have for a full game in the IRB.

EdPovi:
@ollybh I have adopted your recommendations into the code.  Thanks!

ollybh (approved):
Thanks for making these changes. This looks good to me.

PR #11042, opened 2024-07-19
assets: Emit source maps

PistonPercy:
## Before clicking "Create"

## Implementation Notes

I had to rename deps because there was a name collision where the file was written to twice once in compile_lib and another time in combine and thus the corresponding source maps would get mangled. 

### Explanation of Change

This made it much easier to debug some issues. I was able to get ruby source code in the web inspector and see where exceptions where being thrown from.

### Any Assumptions / Hacks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ CONTAINER_COMPOSE ?= $(CONTAINER_ENGINE)-compose
 CONTAINER_ENGINE ?= docker
 
 clean:
-	sudo rm -rfv build/ public/assets/*.js public/assets/*.js.gz public/assets/version.json
+	sudo rm -rfv build/ public/assets/*.js public/assets/*.js.map public/assets/*.js.gz public/assets/version.json
 
 cleandeps:
 	sudo rm -rfv public/assets/deps.js
--- a/api.rb
+++ b/api.rb
@@ -77,7 +77,7 @@ class Api < Roda
     }
   end
 
-  ASSETS = Assets.new(precompiled: PRODUCTION)
+  ASSETS = Assets.new(precompiled: PRODUCTION, source_maps: !PRODUCTION)
 
   Bus.configure
 
--- a/lib/assets.rb
+++ b/lib/assets.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'json'
 require 'opal'
 require 'snabberb'
 require 'tempfile'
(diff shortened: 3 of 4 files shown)


vandamm:
I think source maps are only relevant in development, so you shouldn't add them if `@precompiled` is true

PistonPercy:
It's mostly for development, but it doesn't hurt if it is in production. I guess we do need to have to change the deploy script to actual make it into production.

My ruby is pretty weak, but I'm attempting to refactor out all the sourcemap building logic into a helper (which is probably a good idea anyways) and then it'll easier to wrap it with conditionals.

vandamm:
I can at least confirm it's working as intended and is quite useful
<details>
<summary>Screenshots</summary>
Screenshot 2024-07-23 at 12 54 59
Screenshot 2024-07-23 at 12 55 18


</details>

PistonPercy:
This isn't very well tested yet, but I'm not sure if I'll be able to get back to this for a few days. Just dropping in my most recent revision which does fix the double reading of builds and I think properly gates source map generation. My ruby is pretty weak and I'm not sure about the project conventions, so feel free to ask for fixes.

PistonPercy:
I ran `make clean && docker-compose exec rack rake precompile` and saw no source maps.

Then I ran `make clean` and loaded the minimal repro from https://github.com/tobymao/18xx/pull/11020 and saw a backtrace.



When I looking at timings of individual files, it looks like it's very little overhead.

ollybh:
I'd not come across source maps before. I've had a play around with the browser debugger with these maps added, and this does look like this is going to be very helpful. Trying to debug using the Opal-generated Javascript was borderline impossible, being able to set breakpoints and inspect values in the Ruby code is so much better.

@PistonPercy you've tagged me to review this pull request, but I don't understand the build system well enough to be able to do that. I'm sure that when @michaeljb has some time he'll be able to assess the changes that you've made.

One thing I don't follow is the different methods you've used for generating the source maps in `compile_libs` and `combine`. In the former you use `Opal::Builder::source_map.map.to_json` but in the latter you've taken a different approach with the `SourceMap` class. What's the difference between these? The Opal docs are not great, so I don't really follow what's happening in these two methods.

michaeljb (approved):
> I'm sure that when @michaeljb has some time he'll be able to assess the changes that you've made.

@PistonPercy Sorry it took a while to find some time, but this looks good to me! Did a bit of local testing and it's really nice to debug Ruby directly the browser.

Thanks for getting this to work!

PR #11050, opened 2024-07-22
[steamOverHolland] fixes share issuing, EMR, move to alpha

philcampeau:
## Before clicking "Create"

### Explanation of Change
Modifies the code for issuing shares. This game doesn't handle it quite like 1846. you can only issue as many shares as held by players, but you can do so again in subsequent ORs so long as the 50% total cap isn't exceeded. Therefore, a player can buy just the presidency, then issue 2 shares in ORx.1 and another 2 shares in ORx.2


philcampeau:
undoing move to alpha, just noticed something that's still not functioning properly.

philcampeau:
I ran a few more tests and noticed an issue with selling shares during EMR. I was able to resolve this with the most recent commit.

philcampeau:
Well that's weird. The code passes rake on my end. I'll have to look at this tomorrow.

philcampeau:
Sigh just noticed a few more issures. I'll close this PR and submit a new one once everything is worked out.

PR #11065, opened 2024-07-26
[SteamOverHolland] final fixes - EMR, share issuing. Move to Alpha.

philcampeau:
## Before clicking "Create"

### Explanation of Change

Corps can only issue as many shares as are currently held by players, but unlike 1846, the number of shares in the market has no bearing on this. 

Steam Over Holland has a somewhat unique EMR rule where the corp first must issue shares subject to the normal rules, and then if the corp cash + president cash is still insufficient, the president can either sell shares to raise the money, or they can voluntarily close their corp without compensation.


philcampeau:
Digging through BGG rules threads, I noticed that priority deal doesn't work as I thought. In this game priority deal rotates one player to the left after each SR. I implemented this in the base.rb code and applied it. 

After the initial auction, priority goes to the player with the most cash.

philcampeau:
Holy cow. Alright. I found quite a few more issues after reading through the rules questions on BGG, double-checking the rulebook, and playing through a couple of full games. But after 16 additional commits, I think we're ready to go!

philcampeau:
Wow. Shortly after submitting my "move to alpha" commit, I noticed some details I had missed while reading through rules questions on bgg, and then some stuff I'd missed in the rulebook as well. 18 more commits and several playthroughs later, I'm happy to report this is now _actually_ ready for alpha!

I guess I handled some of the alpha testing myself.

lib/engine/game/base.rb
```diff
@@ -2918,6 +2923,12 @@ def reorder_players(order = nil, log_player_order: false, silent: false)
         when :least_cash
           current_order = @players.dup
           @players.sort_by! { |p| [p.cash, current_order.index(p)] }
+        when :next_clockwise
+          current_order = @players.first
```
ollybh, line 2927:
```suggestion
```

lib/engine/game/base.rb
```diff
@@ -221,6 +221,7 @@ class Base
 
       # :after_last_to_act -- player after the last to act goes first. Order remains the same.
       # :first_to_pass -- players ordered by when they first started passing.
+      # :next_clockwise -- priority passses to the next player in clockwise order.
```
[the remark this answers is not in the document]
ollybh: And can you comment the missing possibilities too whilst you're doing this? (`:most_cash` and `:least_cash`).

ollybh (changes requested):
I haven't finished looking through this pull request, but here a some initial thoughts.

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -193,6 +196,10 @@ def setup
           @or = 0
         end
 
+        def next_sr_player_order
+          @round.is_a?(Engine::Round::Auction) ? :most_cash_keep_order : :next_clockwise
```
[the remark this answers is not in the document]
philcampeau: done

assets/app/view/welcome.rb
```diff
@@ -16,7 +16,7 @@ def render
 
     def render_notification
       message = <<~MESSAGE
-        <p>18 India and 18Uruguay are in alpha.</p>
+        <p>Steam Over Holland, 18 India, and 18Uruguay are in alpha.</p>
```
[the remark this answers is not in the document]
philcampeau: Done

lib/engine/game/base.rb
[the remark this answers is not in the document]
philcampeau: done

lib/engine/game/base.rb
philcampeau: done

lib/engine/game/base.rb
```diff
@@ -2891,6 +2895,10 @@ def next_sr_position(entity)
                            @players.sort_by { |p| [p.cash, @players.index(p)] }.reverse
                          when :least_cash
                            @players.sort_by { |p| [p.cash, @players.index(p)] }
+                         when :next_clockwise
+                           @round.next_entity_index
+                         when :most_cash_keep_order
+                           @players.max_by(&:cash)
```
ollybh, line 2901: This method returns an array of player objects. You're returning either a number or a single player object. This should probably be something like:
```suggestion
                         when :next_clockwise
                           @players.rotate
                         when :most_cash_keep_order
                           @players.rotate(@players.index(@players.max_by(&:cash)))
```
However, I don't think that this code will get called unless you add `:next_clockwise` and `:max_cash_keep_order` here:
https://github.com/tobymao/18xx/blob/f6a04a92260bb3c0af5aeb2022457f1d417d0166/assets/app/view/game/player.rb#L193-L194

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -356,25 +363,42 @@ def non_floated_corporations
...
           raise GameError, 'Route cannot begin/end in a town' if visits.first.town? || visits.last.town?
+
+          super
         end
 
         def revenue_for(route, stops)
-          super
+          revenue = super
+
+          if (ability = abilities(route.corporation, :hex_bonus))
```
ollybh, line 374: Looking at the company definitions, there are two companies that have a hex bonus ability. If the same corporation owns both of these then `abilities` will return an array instead of a single ability object.

ollybh:
```suggestion
          abilities = Array(abilities(route.corporation, :hex_bonus))
```
This would ensure you always get an array that you could work with.

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -403,14 +427,52 @@ def issuable_shares(entity)
...
+        def emergency_issuable_bundles(entity)
+          return [] if @round.issued_shares[entity] || entity.cash >= @depot.min_depot_price
+
+          num_shares = [entity.num_player_shares, 5 - entity.num_market_shares].min
+          bundles = bundles_for_corporation(entity, entity)
+          share_price = stock_market.find_share_price(entity, :current).price
+
+          bundles
+            .each { |bundle| bundle.share_price = share_price }
+            .reject { |bundle| bundle.num_shares > num_shares }
```
ollybh, line 448: I don't think you need to do anything with the bundles' share prices, they should have the current market value by default. Other games need to tweak the price of the bundle if you don't receive the full market value for issued shares.
```suggestion
          bundles_for_corporation(entity, entity)
            .reject { |bundle| bundle.num_shares > num_shares }
```

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -403,14 +427,52 @@ def issuable_shares(entity)
...
+        end
+
+        def upgrades_to?(from, to, _special = false, selected_company: nil)
+          return %w[SOH2 SOH3 SOH4 SOH5].include?(to.name) if from.hex.tile.name == '3'
+          return %w[SOH2 SOH3 981 991].include?(to.name) if from.hex.tile.name == '4'
+          return %w[SOH4 SOH5 981 991].include?(to.name) if from.hex.tile.name == '58'
+
+          super
+        end
```
ollybh, line 457: Instead of hardcoding all these upgrades, you could override `upgrades_to_correct_city_town?` and allow upgrades where `from.towns.size == 1 && to.towns.size == 2`.

lib/engine/game/g_steam_over_holland/step/buy_train.rb
```diff
@@ -0,0 +1,69 @@
...
+            return false if corporation.cash >= @game.depot.min_depot_price || !corporation.trains.empty?
+
+            can_issue?(corporation)
+          end
+
+          def can_issue?(corporation)
+            return false unless corporation.corporation?
+            return false if @round.issued_shares[corporation]
+
+            @game.issuable_shares(corporation).any?
```
ollybh, line 46: Don't call `any?` or `none?` without a block. Use `empty?` instead.
```suggestion
            !@game.issuable_shares(corporation).empty?
```

lib/engine/game/g_steam_over_holland/step/buy_train.rb
```diff
@@ -0,0 +1,69 @@
...
+          end
+
+          def should_buy_train?(entity)
+            :close_corp if can_close_corp?(entity)
```
ollybh, line 50: A method whose name ends in a question mark should only return boolean values, not a label.

lib/engine/game/g_steam_over_holland/step/buy_train.rb
```diff
@@ -0,0 +1,69 @@
...
+            entity.trains.empty? &&
+              !can_issue?(entity) &&
+              entity.cash + entity.owner.cash < @game.depot.min_depot_price &&
+              @game.graph.route_info(entity)&.dig(:route_train_purchase)
+          end
+
+          def can_afford_needed_train?(entity)
+            entity.trains.empty? &&
+              entity.cash + entity.owner.cash >= @game.depot.min_depot_price &&
+              @game.graph.route_info(entity)&.dig(:route_train_purchase)
```
ollybh, line 63: This method is called from `actions` which means it will get called many times when a game is being loaded. Because of this you shouldn't do anything that involves checking the game graph, as that will slow down the game load massively, as it will have to be recalculated multiple times.

You've copied this line from `Engine::Step::Train.must_buy_train?` but that method is called from the game UI, not from the `actions` method. You'll need to work out a way of doing something similar here.

Maybe always give the player the option of passing, but then throw an error in `process_pass` if they still don't have a train and the corporation has a route. There should be something in 1830-type games you can copy.

ollybh (changes requested):

lib/engine/game/base.rb
philcampeau: Hey Olly, I have a couple of questions. First, I'm not sure what you mean with regards to the /assets/app/view/game/player.rb part. 

For the first part above, your suggested code works on it's own. Mine did too, but I'm curious why yours is an improvement.

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: Yes, true! I think I had done it this way because initially I thought each player could only win one private, but that isn't the case. Fixed.

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: Done

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: very cool, didn't know about that method. Thanks! Done.

lib/engine/game/g_steam_over_holland/step/buy_train.rb
philcampeau: done

lib/engine/game/g_steam_over_holland/step/buy_train.rb
philcampeau: I don't disagree, but I took this from 1817 which also does this with the same method. I wanted to work within that framework, since it's very similar to how 1817's liquidation works.

lib/engine/game/base.rb
ollybh: The only place that `Game.next_sr_position` is called is in `View::Game::Player`, and the call to `next_sr_position` is guarded by a check that `Game.next_sr_player_order` is one of `:first_to_pass`, `:most_cash` or `:least_cash`.

Here, `next_sr_player_order` will return either `:most_cash_keep_order` or `:next_clockwise`, so this version of `next_sr_position` is never being called, which is why you're not seeing any errors.

lib/engine/game/g_steam_over_holland/step/buy_train.rb
ollybh: OK. This isn't ideal, but it's matching existing code, so it will do.

lib/engine/game/base.rb
philcampeau: Gotcha! Done.

lib/engine/game/g_steam_over_holland/step/buy_train.rb
philcampeau: very good point. I think I've done it correctly now. Have a look and let me know.

lib/engine/game/base.rb
```diff
@@ -2891,7 +2895,10 @@ def next_sr_position(entity)
                            @players.sort_by { |p| [p.cash, @players.index(p)] }.reverse
                          when :least_cash
                            @players.sort_by { |p| [p.cash, @players.index(p)] }
-                         else
+                         when :next_clockwise
+                           @players.rotate
+                         when :most_cash_keep_order
+                           @players.rotate(@players.index(@players.max_by(&:cash)))
                            []
```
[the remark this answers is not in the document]
philcampeau: Oops! Good catch

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -356,25 +363,43 @@ def non_floated_corporations
...
+          revenue = super
 
-          raise GameError, 'Route visits same hex twice' if route.hexes.size != route.hexes.uniq.size
+          abilities(route.corporation, :hex_bonus).each do |ability|
+            stops.each do |stop|
+              next unless ability.hexes.include?(stop.hex.name)
+
+              revenue += 20
+            end
+          end
```
ollybh, line 380: Your code will work, but it would also be possible to write this without an iterator. In theory this might be more efficient code. Something like this (would need to be tested):
```suggestion
          abilities = abilities(route.corporation, :hex_bonus)
          return revenue if abilities.empty?
          
          bonus_hexes = abilities.flat_map(&:hexes)
          bonus = 20 * stops.map(&:hex).map(&:coordinates).intersect(bonus_hexes).size
          revenue + bonus
```

philcampeau: sure, yes this works too. Done

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -48,6 +49,8 @@ class Game < Game::Base
           { lay: true, upgrade: :not_if_upgraded, cannot_reuse_same_hex: true },
         ].freeze
 
+        attr_accessor :close_corp
```
ollybh, line 52: Is this used by anything?

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -356,25 +363,40 @@ def non_floated_corporations
...
+          bonus = 20 * stops.map(&:hex).map(&:coordinates).intersect(bonus_hexes).size
+          revenue + bonus
         end
 
         def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil, movement: nil)
           price_drop = bundle.num_shares
           corporation = bundle.corporation
           old_price = corporation.share_price
+          old_owner = bundle.owner
+          was_president = corporation.president?(bundle.owner)
```
ollybh, line 387: You've just saved this value in a variable, so you might as well not look it up again.
```suggestion
          was_president = corporation.president?(old_owner)
```

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: nope, it was but got changed. Removing.

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: very good point! done

ollybh:
BTW, I've just noticed a typo in the location names: it's Groningen, not Broningen. The board does have a truly horrific font.

philcampeau:
Hmm maybe I'll check an online map. I don't even have a copy of the game, I was relying on bgg images!

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -356,25 +361,40 @@ def non_floated_corporations
...
-          raise GameError, 'Route visits same hex twice' if route.hexes.size != route.hexes.uniq.size
+          abilities = abilities(route.corporation, :hex_bonus)
+          return revenue if abilities.empty?
+
+          bonus_hexes = abilities.flat_map(&:hexes)
+          bonus = 20 * stops.map(&:hex).map(&:coordinates).intersect(bonus_hexes).size
+          revenue + bonus
         end
 
         def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil, movement: nil)
```
ollybh, line 380: This method could be tidied up.

All three branches of the `if` statement call `@stock_market.move_left` but only one branch uses the `price_drop` variable you set at the start of the method.

Could this be rewritten to have the `if` statement modify `price_drop` as required, moving the call to `move_left` out of the branches. And move the `@round.issued_shares[corporation] = true` to `GSteamOverHollad::Step::BuyTrain.process_sell_shares` to match where it's set in the `IssueShares` step.

philcampeau: Alright, I cleaned up that section quite a bit, thanks for the suggestion. I also made a change to revenue_for, because I realized that with the suggested code, abilities is only an array if the corp has a hex_bonus ability. So now the gate on that is `return revenue unless abilities`. Tested, works.

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -356,44 +361,48 @@ def non_floated_corporations
...
+
+          super
         end
 
         def revenue_for(route, stops)
-          super
+          revenue = super
+
+          abilities = abilities(route.corporation, :hex_bonus)
+          return revenue unless abilities
```
ollybh, line 373: I'd missed this last time, sorry.
```suggestion
          abilities = Array(abilities(route.corporation, :hex_bonus))
          return revenue if abilities.empty?
```
`Game.abilities` is, IMHO, a total pain. It can either return nil, a single ability object, or an array of ability objects. My approach is, unless you can be absolutely certain about what type will be returned, is to wrap the call inside `Array(…)`. That way you are guaranteed to get an array under all circumstances (`Array(nil) == []`, `Array(Object) = [Object]` and `Array([1, 2, 3]) = [1, 2, 3]`).

The code as currently written will work if zero or 2+ abilities are found. It will fail if there's a single matching ability, as `flat_map` isn't a method implemented by `Ability`.

philcampeau: Ha, good point. Thanks. Fixed.

lib/engine/game/g_steam_over_holland/step/buy_train.rb
```diff
@@ -0,0 +1,84 @@
...
+              "Close #{current_entity.name}"
+            else
+              @acted ? 'Done (Trains)' : 'Skip (Trains)'
+            end
+          end
+
+          def process_pass(action)
+            entity = action.entity
+
+            raise GameError, 'Corporation can afford train, a train must be purchased.' if can_afford_needed_train?(entity)
```
ollybh, line 31: This is where you need to check if the corporation has a route, and only throw an error if there is one.
To avoid the graph being calculated during game load, structure the test along the lines of:
```raise GameError '…' if !@game.loading && has_route && can_afford```
That means it will get checked when a player is making a move, but when the game is being loaded it is assumed that this has already been checked and is OK.

File: lib/engine/step/buy_train.rb [hunk 19]
```diff
@@ -71,6 +71,8 @@ def process_buy_train(action)
       end
 
       def swap_sell(_player, _corporation, _bundle, _pool_share); end
+
+      def should_buy_train?(entity); end
```
ollybh, line 75: This isn't needed. Engine::Step::BuyTrain includes Engine::Step::Train which already defines this method.
```suggestion

      def should_buy_train?(entity); end
```

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: done

lib/engine/game/g_steam_over_holland/step/buy_train.rb
philcampeau: Yep, had to flip it around into if... raise GameError, but it's done.

lib/engine/step/buy_train.rb
philcampeau: done

lib/engine/game/g_steam_over_holland/step/buy_train.rb
philcampeau: Whoops, that should definitely be !@game.loading 

Fixing now...

ollybh (approved):

PR #11069, opened 2024-07-26
Fix issue checking playercount 

### LINKED ISSUE #11064: Unable to import hotseat game
It is no longer possible to import a JSON game file. Attempting to do this gives an error:
```Uncaught title: undefined method `title' for nil```

The error is occurring here:
https://github.com/tobymao/18xx/blob/edc5f9e1b9ad202232e4e7d896fcbd74cc7d0050/assets/app/view/create_game.rb#L411

Reverting commit a0ced7ac74 makes it possible to import hotseat games again.

[CC @benjaminxscott]

### PR DESCRIPTION
**Pull request by benjaminxscott**
Fixes #11064

After this change, JSON import and game creation all work

In the case of imported games, `selected_game_or_variant` is Nil so the code was bailing when `title` was set.

This was due to where I was checking playercount, as part of a codepath that handled all 3 game types (multi, hotseat, json)

Now it only checks playercount for multiplayer 

I refactored this function to be (hopefully) more readable. 

## Before clicking "Create"

### Screenshots

To confirm that game creation and import works:

[import game.webm](https://github.com/user-attachments/assets/9ee0e72c-4961-4c59-8d4b-eae3a65ee0fd)

[create_multi_and_hotseat.webm](https://github.com/user-attachments/assets/095cf2d8-aad1-46b1-b21e-9977cc94f31c)

### DIFF
--- a/assets/app/view/create_game.rb
+++ b/assets/app/view/create_game.rb
@@ -412,26 +412,20 @@ module View
 
       game_params = params
 
-      if @mode == :multi
+      title = selected_game_or_variant&.title
+
+      case @mode
+      when :multi
+        if game_params[:min_players].to_i < @min_p[title] || game_params[:max_players].to_i > @max_p[title]
+          return store(:flash_opts,
+                       'Invalid playercount')
+        end
+
         game_params[:seed] = game_params[:seed].to_i
         game_params[:seed] = nil if (game_params[:seed]).zero?
         return create_game(game_params)
-      end
 
-      players = game_params
-                  .select { |k, _| k.start_with?('player_') }
-                  .values
-                  .map { |name| name.gsub(/\s+/, ' ').strip }
-
-      return store(:flash_opts, 'Cannot have duplicate player names') if players.uniq.size != players.size
-
-      if @mode == :json
-        begin
-          game_data = JSON.parse(game_params['game_data'])
-        rescue JSON::ParserError => e
-          return store(:flash_opts, e.message)
-        end
-      else
+      when :hotseat
         game_data = {
           settings: {
             optional_rules: game_params[:optional_rules] || [],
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: assets/app/view/create_game.rb [hunk 1]
```diff
@@ -408,44 +408,47 @@ def render_game_type
...
+          game_data = JSON.parse(game_params['game_data'])
+        rescue JSON::ParserError => e
+          return store(:flash_opts, e.message)
+        end
       end
 
+      players = game_params
+      .select { |k, _| k.start_with?('player_') }
+      .values
+      .map { |name| name.gsub(/\s+/, ' ').strip }
```
ollybh, line 448:
```suggestion
      .map { |_, name| name.gsub(/\s+/, ' ').strip }
```

assets/app/view/create_game.rb
```diff
@@ -408,44 +408,47 @@ def render_game_type
...
-      if game_params[:min_players].to_i < @min_p[title] || game_params[:max_players].to_i > @max_p[title]
-        return store(:flash_opts,
-                     'Invalid playercount')
-      end
 
-      if @mode == :multi
+      title = selected_game_or_variant&.title
+
+      case @mode
+      when :multi
```
ollybh, line 418:
```suggestion

      case @mode
      when :multi
        title = selected_game_or_variant&.title
```

ollybh (commented):
I've tested this locally and confirm it fixes #11064 for me. Thanks!
The rest of the changes look mostly OK, though I'm not familiar with this bit of the codebase.

assets/app/view/create_game.rb
benjaminxscott: Sweet yeah I didn't tweak this LoC but I grok this being easier to read while we're refactoring stuff

PR #11070, opened 2024-07-27
[1846] optional rules are XOR

benjaminxscott:
Currently one can create a '46 in which only 1E privates are allowed and also 2E privates are guaranteed.

This implements the 'check options' feature and prevents that at the frontend level. 

Also will throw an error if a game is imported with such options

## Before clicking "Create"

### Screenshots

image

Screenshot 2024-07-26 5 16 07 PM

### DIFF
--- a/lib/engine/game/g_1846/meta.rb
+++ b/lib/engine/game/g_1846/meta.rb
@@ -40,6 +40,17 @@ module Engine
             players: [2, 3, 4, 5],
           },
         ].freeze
+
+        MUTEX_RULES = [
+          %i[first_ed second_ed_co],
+].freeze
+
+        def self.check_options(options, _min_players, _max_players)
+          optional_rules = (options || []).map(&:to_sym)
+          return unless (optional_rules & %i[first_ed second_ed_co]).length == 2
+
+          { error: 'Cannot guarantee 2nd Edition companies if using 1st Edition' }
+        end
       end
     end
   end


lib/engine/game/g_1846/meta.rb
```diff
@@ -40,6 +40,13 @@ module Meta
             players: [2, 3, 4, 5],
           },
         ].freeze
+
+        def self.check_options(options, _min_players, _max_players)
+          optional_rules = (options || []).map(&:to_sym)
+          return unless (optional_rules & %i[first_ed second_ed_co]).length == 2
```
benjaminxscott, line 46: This syntax was is due to rubocop not liking an if statement ( Style/GuardClause) and also not liking AND in a guard clause
(Style/UnlessLogicalOperators)

lib/engine/game/g_1846/meta.rb
philcampeau, line 1: You could also add 

```
MUTEX_RULES = [
          %i[first_ed second_ed_co],
        ].freeze
```

which would make the site uncheck one option when the other is chosen.

benjaminxscott: added

lib/engine/game/g_1846/meta.rb
```diff
@@ -40,6 +40,17 @@ module Meta
             players: [2, 3, 4, 5],
           },
         ].freeze
+
+        MUTEX_RULES = [
+          %i[first_ed second_ed_co],
+].freeze
+
+        def self.check_options(options, _min_players, _max_players)
```
ollybh, line 48: This method isn't needed any more, as the MUTEX_RULES mean it will be impossible to select both of these options. (Though of course it doesn't do any harm either).

PR #11084, opened 2024-07-30
[18SJ] Set in production

perwestling:

## Implementation Notes
Updated Welcome text.
Added two fixtures for 18SJ. One full standard game, and one full gane for Oscarian variant. Do cover a good selection of all privates. 18SJ only had a previous fixture that tested the fix of an error case for the 2 player variant.
So it does cover a good bit of options, including incremental and full cap.

### Explanation of Change
18SJ has been stable for over a year, with only few issues. No open isses at the moment.

### Screenshots
None

### Any Assumptions / Hacks
None

### DIFF
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -17,6 +17,7 @@ module View
     def render_notification
       message = <<~MESSAGE
         <p>18 India and 18Uruguay are in alpha.</p>
+        <p>18 SJ is now in production.</p>
 
         <p>Report bugs and make feature requests <a href='https://github.com/tobymao/18xx/issues'>on GitHub</a>.</p>
       MESSAGE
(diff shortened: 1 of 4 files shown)


assets/app/view/welcome.rb
```diff
@@ -16,7 +16,8 @@ def render
 
     def render_notification
       message = <<~MESSAGE
-        <p>18 India and 18Uruguay are in alpha.</p>
+        <p>18 India and 18Uruguay are in alpha.</br>
+           18 SJ is now in production.</p>
```
ollybh, line 20: Those HTML tags are mismatched. A line break on its own should be `<br/>`, or it might be better to put these in separate paragraphs.
```suggestion
        <p>18 India and 18Uruguay are in alpha.</p>
        <p>18 SJ is now in production.</p>
```

ollybh (changes requested):
Need to fix broken HTML on the start page.

assets/app/view/welcome.rb
perwestling: Made the proposed change.

ollybh (approved):

PR #11085, opened 2024-07-30
[18 India] Fix OO reservation relocate bug 

### LINKED ISSUE #11058: [18 India] Can't lay home token on OO hexes
https://18xx.games/game/171575?action=215

When EBR first operates, it lays its home OO tile on P17. Once the tile is laid, a prompt to select the home city from the 2 OO options comes up. When attempting to click on either city, the page gives an error and force refreshes and rolls back the action. Both myself and the owner in master mode had the same behavior. 

126.0.6478.128 (Official Build) (64-bit)

EdPovi:
Fixes #11058 

## Implementation Notes
This is "unlikely" to break existing games that haven't gotten past this point or didn't have two corps in a OO tile. 

### Explanation of Change
Rewrote method that moved a reservation from a hex back to a city after it was moved by laying the first yellow OO tile. 

### Screenshots
image

### Any Assumptions / Hacks
The default code doesn't handle the case when  yellow OO tile is placed where there is a home token and a reservation to track.   This case is handled by first moving the reservation to the hex and then after the tile and token are placed, moving the reservation back to the other un-tokened city.

### DIFF
--- a/lib/engine/game/g_18_india/step/home_track.rb
+++ b/lib/engine/game/g_18_india/step/home_track.rb
@@ -58,12 +58,11 @@ module Engine
           # Base code doesn't handle one token and one reservation on a OO tile
           # Moves a reservation from hex to untoken city
           def replace_oo_reservations(tile)
-            return if !tile.label == 'OO'
+            return unless tile.label == 'OO'
 
             cities = tile.cities
             reservations = tile.reservations.dup
             LOGGER.debug { "replace_oo_reservations > reservations: #{reservations}" }
-            LOGGER.debug { "replace_oo_reservations > label: #{tile.label}" }
             cities.each do |city|
               next if city.tokened?
 
--- a/lib/engine/game/g_18_india/step/home_track.rb
+++ b/lib/engine/game/g_18_india/step/home_track.rb
@@ -58,19 +58,12 @@ module Engine
           # Base code doesn't handle one token and one reservation on a OO tile
           # Moves a reservation from hex to untoken city
           def replace_oo_reservations(tile)
-            return unless tile.label == 'OO'
+            return unless tile.name == '235'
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -48,29 +48,29 @@ def process_lay_tile(action)
...
-              end
+            LOGGER.debug { "replace_oo_reservations > reservations: #{reservations}" }
+            cities.each do |city|
+              next if city.tokened?
+
+              corp = reservations.pop
+              tile.remove_reservation!(corp)
+              city.add_reservation!(corp)
+              LOGGER.debug { "replace_oo_reservations > city reservation: #{city.reservations}" }
             end
```
ollybh, line 73: If I've understood this right, this will only get called when a yellow OO tile is being placed, and there will be one tile reservation, and two cities – one with a token and one empty. The reservation needs to be moved from the tile to the open city. Is that correct?

What was wrong with the older version of the code? The `cities.count == 1` test?

I don't see the need for a loop here. I guess that you're trying to generalise the code here, but this would fail if it got called when there were more open cities than tile reservations. You either need to add some more checks, or simplify it based on what we know.

```
corp = tile.reservations.first
city = cities.reject(&:tokened?).first
city.add_reservation!(corp)
tile.reservations.clear
```

ollybh:
BTW, if there's any chance that any games could be broken by this change you should add an archive/pin tag to the pull request. What would have happened if someone ran into this situation before?

lib/engine/game/g_18_india/step/home_track.rb
EdPovi: Modified method to remove loop as suggested. Thanks!

ollybh (approved):

PR #11090, opened 2024-07-31
[18 India] Fix Gauge Change with Mumbai/Nepal bug and Director's Share Price Paid bug

### LINKED ISSUE #11058: [18 India] Can't lay home token on OO hexes

### LINKED ISSUE #11082: [18 India] Train Routes cannot end in Nepal or Mumbai if they go over Gauge Change
In a game I was playing with others I noticed that with an express train, if I ended my route on Mumbai or Nepal, I was getting an error message that "Route may not visit MUMBAI more than once" (or Nepal respectively). I created a test game to see if I could recreate that condition. From testing, it appears that if a train tries to end on one of these two cities, and the route passes over a gauge change, you get this error message. The test game is [172240](https://18xx.games/game/172240).

2 Routes, one to Nepal, one to Mumbai, both considered invalid because they are going over gauge change, and end in one of these respective cities:
2 Routes, one to Nepal, one to Mumbai, both considered invalid because they are going over gauge change, and end in one of these respective cities

Same 3x2 route to Mumbai, with gauge change removed, now valid:
Same 3x2 route to Mumbai, with gauge change removed, now valid

Same 4E route to Nepal, with gauge change removed, now valid:
Same 4E route to Nepal, with gauge change removed, now valid

Note you must end your route on one of these two cities to encounter this error. If you were to try either of these two routes in reverse (starting with the cities in question) you do not get this error. Just containing the city is not enough to get the error either, it must be the last city.

3x2 route, drawn in reverse, starting with Mumbai, and crosses gauge change is valid:
3x2 route, drawn in reverse, starting with Mumbai, and crosses gauge change is valid

I encountered the issue on Firefox 128.0, Edge 126.0.2592.113, and mobile Chrome 126.0.6478.188.

### LINKED ISSUE #11087: [18India]  Buying the Director's share of a previously operated Managed company charges market price not face value
Hotseat game, Firefox

As per title.  BNR operated in the 3rd set of operating rounds as a managed company, and the stock price rose to $90 going into SR4.  Player purchased the director's share.  The game log indicates it was purchased at face value for $164 but their actual cash stockpile decreased from $587 to $407, a drop of $180.

Purchasing single shares that are not the director's share only cost $82.  

https://gist.github.com/ScottDrone/8c55178c02b7c76349df6458fc7517ba

image

image

image

EdPovi:
Fixes #11082 
Fixes #11087 

## Implementation Notes
This PR can break existing games as price difference may affect money on hand for actions. 

### Explanation of Change
When the route includes Gauge Changes, additional zero revenue city stops are added when 'visited_stops' is called. 
The stops need a dummy tile to reference.   The bug was caused by just referencing the first tile in the route, which gave the impression of visiting it too many times.  A new method was added to provide a tile without a location name that should avoid any side effects. 

Modified Method that converts a company from managed to directed (purchase of Director's share) to use Par Price, instead of Share Price. 

### Any Assumptions / Hacks
The tile picking method assumes that a tile without a location name will exist within the tiles next to gauge changes.


lib/engine/game/g_18_india/game.rb
```diff
@@ -963,27 +963,41 @@ def companies_to_payout(ignore: [])
...
+          gc_hexes = @gauge_change_markers.flatten
+          hexes = route.hexes & gc_hexes
+          hexes.each do |hex|
+            return hex.tile unless hex.tile.location_name
+          end
+          # if there isn't an unnamed intersecting hex, use the other GC hexes
+          gc_hexes.each do |hex|
+            return hex.tile unless hex.tile.location_name
+          end
+          raise GameError, 'Tile not found for gauge change'
```
ollybh, line 998: This feels like you will have made encountering a bug harder, but not impossible.

Does the gauge change stop's tile actually get used for anything? Could you pass it a dummy Tile object [`Engine::Tile.new(…)`]?

BTW, from seeing this I've realised that you've rolled your own code for checking whether two cities on the same tile are visited in a route. There is already code for this, if you add a 'groups' element to the tile's definition. That would mean you didn't need to worry about tiles with named locations (but it wouldn't have prevented this bug).

ollybh: Actually, using groups to check the route instead of tile names would avoid this bug. Groups are an attribute of the revenue centre, not the tile, so you wouldn't be introducing new stops that had this attribute.

EdPovi: I agree with the idea to use a dummy "new" tile. I have figured out what to pass to the method to initialize a new one.

EdPovi: OK, I see the 'Groups' definition for revenue centers.  I will test puting those cities into tile groups.

EdPovi: Tested using groups, and removed test in `check_other`.

EdPovi: Thanks for the suggestions, I think it is better with these adjustments.

ollybh (approved):

PR #11095, opened 2024-08-02
[18TS] Map and companies

douglasfir:
Map and companies for 18TS, an 1825 inspired design by site user Martin2D. My implementation rips heavily from the 1825 implementation. Unlike 1825, there are only the 3 "units", no additional variants, so the setup scripts are significantly simpler.


lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+
+        def adjust_certs(certs, chash)
+          chash.keys.each do |nplayers|
+            if certs[nplayers]
+              certs[nplayers] += chash[nplayers]
+            else
+              certs[nplayers] = chash[nplayers]
+            end
+          end
+        end
```
ollybh, line 474: This method isn't called anywhere in this pull request, so is it needed?

If it is called elsewhere, it would be better to use a hash with a default value. `certs = Hash.new(0)` gives you a hash where you can later write `certs[5] += 2` and it will either add two to the existing value, or create a new entry with a value of zero and then add two to it.

lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+
+        def tile_valid_for_phase?(tile, hex: nil, phase_color_cache: nil)
+          phase_color_cache ||= @phase.tiles
+
+          # 119 upgrades from yellow in phase 3
+          return false if tile.name == '119' && !phase_color_cache.include?(:brown)
+
+          # 166 upgrades from green in phase 4
+          return false if tile.name == '166' && !phase_color_cache.include?(:gray)
```
ollybh, line 623: Are these needed? 119 is a green tile with brown stripes, so should automatically be available when green tiles are available. Likewise 166 is a brown tile with grey stripes, so should become available at the same time as brown tiles.

lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+        end
+
+        def location_name(coord)
+          @location_names ||= game_location_names
+          @location_names[coord]
+        end
+
+        def game_location_names
+          locations = LOCATION_NAMES.dup
+          if optional_rules.include?(:unit_4)
```
ollybh, line 635: Obsolete code? There isn't a unit 4 mentioned elsewhere.

lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+          return false if illegal_upgrade?(from, to) # only really needed for upgrades shown on tile manifest
+
+          # deal with striped tiles
+          # 119 upgrades from yellow in phase 3, and upgrades to gray
+          return false if from.name == '119' && to.color == :brown
+          # 166 upgrades from green in phase 4, but doesn't upgrade
+          return false if from.name == '166'
+          # 200 upgrades from pre-printed brown tiles on Crewe, Wolverton or Swindon, doesn't upgrade
+          return false if from.name == '200'
+          return false if to.name == '200' && from.color != :sepia
```
ollybh, line 660: Are the pre-printed hexes brown or sepia?

lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+        # -1 if a has a higher par price than b
+        # 1 if a has a lower par price than b
+        # if they are the same, then use the order of formation (generally flotation)
+        def par_compare(a, b)
+          if a.par_price.price > b.par_price.price
+            -1
+          elsif a.par_price.price < b.par_price.price
+            1
+          else
+            @formed.find_index(a) < @formed.find_index(b) ? -1 : 1
```
ollybh, line 842: You can use the spaceship operator here:
```suggestion
            @formed.find_index(a) <=> @formed.find_index(b)
```

lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+        end
+
+        def can_run_route?(entity)
+          super && !silent_receivership?(entity)
+        end
+
+        def after_buying_train(train, source)
+          return unless new_rank_available?(train, source)
+          if @depot.upcoming.first.name == '5'
+          end
```
ollybh, line 881: ?

ollybh (changes requested):
It's going to take a while to review a pull request that's this big. Here are some comments and questions on the first part of this PR.

lib/engine/game/g_18_ts/game.rb
```diff
@@ -0,0 +1,1166 @@
...
+require_relative '../base'
+require_relative '../../option_error'
+require_relative '../../distance_graph'
+require_relative 'entities'
+require_relative 'map'
+
+module Engine
+  module Game
+    module G18TS
+      class Game < Game::Base
```
ollybh, line 13: As this is an 1825 derivative which looks like it duplicates a lot of code from the 1825 implementation, it would be better if this was a subclass of `G1825::Game` instead of `Game::Base`. That would give you most of 1825’s rules without all this copied code. Though you might have to refactor the 1825 code in places to allow you to override behaviour here.

ollybh:
@douglasfir Can you check that you haven't included more code in this pull request than you had meant to?
The PR cover note says that this game has three units and no other variants, but there is code for handling kits and regional variants. And the PR title says this is the map and companies, but there is far more than this included.

douglasfir:
> @douglasfir Can you check that you haven't included more code in this pull request than you had meant to? The PR cover note says that this game has three units and no other variants, but there is code for handling kits and regional variants. And the PR title says this is the map and companies, but there is far more than this included.

Sorry for the delayed response, been away from home for the past week. My original intention with this pull request was to just post my static files (map + companies + meta information). I only realized afterwards that the rakefile requires a functioning game to accept static files, so I then included my other files since they are all necessary dependencies without removing all of the 1825 vestiges. 

Thanks for the advice of using a subclass; you're right, that makes much more sense. I will refactor my code accordingly, but I'm wondering if this can be releasse as an alpha quality game now to allow the designer to playtest in parallel.

ollybh:
> My original intention with this pull request was to just post my static files (map + companies + meta information). I only realized afterwards that the rakefile requires a functioning game to accept static files, so I then included my other files since they are all necessary dependencies without removing all of the 1825 vestiges.

It doesn't require a functioning game, it just needs files that can be loaded without errors. If you need to include files other than ones for the all the definitions you can just include stub files for the moment – these shouldn't cause any errors and you can add the code later on:
```
# frozen_string_literal: true

require_relative '../../../step/dividend'

module Engine
  module Game
    module G18TS
      module Step
        class Dividend < Engine::Step::Dividend
        end
      end
    end
  end
end
```

ollybh:
> Thanks for the advice of using a subclass; you're right, that makes much more sense. I will refactor my code accordingly, but I'm wondering if this can be releasse as an alpha quality game now to allow the designer to playtest in parallel.

We only have a production branch and a single production server, so code needs to be in a reasonable state before it is merged.

If you need to be able to work with the game developer to test your code before it is merged then you've got a couple of options:

1. If they've got enough computer skills they could a test instance themselves (either on their own computer or a Github codespace), clone your git repository and test the code themselves.
2. You could give them access to a development instance you are running yourself. This is what I do, I have a port forwarding rule set up on my home router to pass on requests to a computer running the server code. It might also be possible to share access to a Github codespace, I haven't done anything with these myself.

ollybh:
@douglasfir Has there been any progress with this branch? It's been a couple of months since you said you were going to look at redoing this as a subclass of G1825.

douglasfir:
> @douglasfir Has there been any progress with this branch? It's been a couple of months since you said you were going to look at redoing this as a subclass of G1825.

Hi @ollybh, 
Thanks for checking in. I haven't made much progress since we last spoke. I am going to devote more time to it in the coming weeks. If anyone else wants to take a look as well, they are of course welcome.

ollybh:
@douglasfir I'm going to close this pull request for the moment, just to get it off the list of PRs for review.

When you've reworked your code you can either reopen this pull request, or submit a new one.

PR #11112, opened 2024-08-10
[18uruguay] various fixes

### LINKED ISSUE #11047: [18Uruguay] - names of sheep/corn/cattle privates
The abbreviations for the three agricultural privates are inconsistent.

LA_CORN
LO_SHEEP
LO_CATTLE

Shouldn't they all be either "LA" or "LO"?

### LINKED ISSUE #11051: [18Uruguay] - RPTLA charter mis-identifies the float percentage
In the initial auction, the RPTLA charter states that it takes 20% to float.  However, the RPTLA president's certificate is a 30% share.  It appears that the charter statement is an error.

### LINKED ISSUE #11080: [18Uruguay] 4D trains counted incorrectly
**What's your Game ID?**

https://18xx.games/game/172187?action=732

**How to Reproduce the Bug?**

Run a 4D train including dots and 4 cities.

**Which Browser / Platform?**

Chrome Version 126.0.6478.127

The rules v.0.98.4 section 8.2.4 on page 9 says "The 4D train may include four stops on its run but may skip towns. It counts the value of the included towns and cities double."  The run seems to count 4 cities and as many towns as you like then double the total, but I believe the intent of the rules is that dots count towards the 4 stops if you want to count them, otherwise they score 0.

philcampeau:
Fixes #11047
Fixes #11051 
Fixes #11080

## Before clicking "Create"

### Explanation of Change

- Changes sym for Latifundio Agrícola to LO_CORN (and all references to it) to match the other Latifundio private syms.
- Sets float percentage for RPTLA to 30, since it has a 30% presidency.
- Corrects 4D train. It currently runs 4 cities/offboards and infinite towns, it should be max 4 stops with the ability to skip towns.

### DIFF
--- a/lib/engine/game/g_18_uruguay/companies.rb
+++ b/lib/engine/game/g_18_uruguay/companies.rb
@@ -24,7 +24,7 @@ module Engine
           revenue: 10,
           desc: 'Owner (or controlling president) moves one crop cube from a countryside'\
                 'hex to an adjacent town or city hex each OR. Closes when the last crop cube has been moved.',
-          sym: 'LA_CORN',
+          sym: 'LO_CORN',
           abilities: [
                 {
                   type: 'assign_hexes',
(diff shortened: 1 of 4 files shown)


ollybh (commented):
Looks good to me (although I can't see this fixing #11051), but this should be reviewed by @patrikolesen.

bentziaxl (approved):

PR #11117, opened 2024-08-11
[1858] Refactoring in preparation for '1858 Switzerland' variant

ollybh:
## Implementation Notes

Ian D Wilson has a prototype in development for an 1858 variant set in Switzerland.

This pull request contains the changes needed to the 1858 codebase to make it possible to implement this new game on 18xx.games.

The code has been made more generic in a few places, with some methods extracted or renamed, and constants moved and renamed.

There is nothing here that affects anything outside the 1858 codebase, and there should not be anything that will affect 1858 games in progress.

## Before clicking "Create"


scottredracecar:
Olly, kudos for taking this on!

PR #11119, opened 2024-08-11
[1858Switzerland] Add entities, map, tiles, icons and logos

ollybh:
## Implementation Notes

1858 Switzerland is a new prototype from Ian D Wilson. As the name suggests, it is a game based on 1858 set in Switzerland.

This adds definitions of:
 - The public companies (majors, 5-/10-share corporations).
 - The private railway companies (implemented as a pair of a private and a minor).
 - The trains and phases.
 - The map.
 - The tiles.

There's almost no actual code in this, just enough to override some of the definitions for trains and phases inherited from the base 1858 code.

It also adds SVG icons and logos for the companies, plus a few icons for the map.

### Screenshots

image

## Before clicking "Create"


lib/engine/game/g_1858_switzerland/game.rb
```diff
@@ -0,0 +1,128 @@
...
+          trains = super
+          _train_2h, _train_4h, train_6h, _train_5e, train_6e, train_5d = trains
+          train_6h.delete(:obsolete_on) # Wounded on second grey train, handled in code
+          train_6h[:events] = [{ 'type' => 'blue_privates_available' }]
+          train_6e[:events] = [{ 'type' => 'privates_close2' }]
+          train_6e[:price] = 700
+          train_6e[:variants][0][:price] = 600
+          train_5d[:available_on] = '6'
+          trains
+        end
```
bentziaxl, line 111: since you're only modifying 2 trains from the list of trains, would it make sense to refactor this as:
trains_6h = trains.find{ |t| t.name == '6'}
or 
trains[5]

instead of the multiple assignments you're not using?

bentziaxl (approved):

lib/engine/game/g_1858_switzerland/game.rb
ollybh: This code is currently modifying three of the six types of train. And there's a variant (not included in this pull request) that modifies a fourth. Both of your suggestions would work, but I thought that this was as clear as any way. And the interpreter/compiler could optimise away the unused assignments (in theory anyway, I'm not sure if this happens).

bentziaxl: i read 6h and 6e as the same train. ok let's keep it as is

PR #11120, opened 2024-08-12
[1849] Forces games with Electric Dreams variant to also use Bonds and Acquiring Tokens

### LINKED ISSUE #11116: [1849] Electric dream variant should require other variants to be in play
Example game:
https://18xx.games/game/173634

Running with only E-train variant but no bond or token purchase.

Section 14.3 of the AAGames rulebook states

> This variant, which can be used only with the other variant rules, allows you to explore the possibility of an earlier electrification of the Sicilian railway network.

This implies that the other variant rules should also be in play if Electric Dreams is selected.

It's possible to choose to play with all the variants as suggested by the rules or to select only Electric Dreams, which is not recommended by the designer.

philcampeau:
Fixes #11116

## Before clicking "Create"

### Explanation of Change

Rules say that games with Electric Dreams also must use the Acquiring Station Tokens and Bonds variants. This automatically includes the other two variants if Electric Dreams is chosen.

Note: This will require pins for any games that are using the Electric Dreams variant without both Bonds and Acquiring Station Tokens.

### DIFF
--- a/lib/engine/game/g_1849/game.rb
+++ b/lib/engine/game/g_1849/game.rb
@@ -829,7 +829,7 @@ module Engine
         # code below is for variant rules
 
         def acquiring_station_tokens?
-          @acquiring_station_tokens ||= @optional_rules&.include?(:acquiring_station_tokens)
+          @optional_rules&.include?(:electric_dreams) || @optional_rules&.include?(:acquiring_station_tokens)
         end
 
         def electric_dreams?
@@ -837,7 +837,7 @@ module Engine
         end
 
         def bonds?
-          @bonds ||= @optional_rules&.include?(:bonds)
+          @optional_rules&.include?(:electric_dreams) || @optional_rules&.include?(:bonds)
         end
 
         def reduced_4p_corps?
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1849/game.rb
```diff
@@ -829,15 +829,15 @@ def par_prices
         # code below is for variant rules
 
         def acquiring_station_tokens?
-          @acquiring_station_tokens ||= @optional_rules&.include?(:acquiring_station_tokens)
+          @optional_rules&.include?(:electric_dreams) || @optional_rules&.include?(:acquiring_station_tokens)
```
crericha, line 832: I don't think this is the right approach. There is a way to select groups of optional rules. See 1825 for inspiration.

philcampeau: I feel as though the way 1825 does it is done out of necessity because of the complexity of all the possible combinations. This also works, and is much simpler in my opinion.

philcampeau: alright, I modified things after a discussion with crericha over Slack

crericha (approved):

philcampeau:
As of right now, this would kill games 173567 & 173634. Pinning them would be ideal since 1849 isn't in Alpha, but I'll leave that to the discretion of the person who handles the deployment.

PR #11123, opened 2024-08-12
[22Mars] Initial Permit distribution round

vandamm:
## Before clicking "Create"

### Explanation of Change
It's like ISR, but permits (privates) are randomly distributed to players (1 each) and then players can choose whether to buy the assigned permit for listed price or to remove it from the game. And you can view which permits are in player hands on Entities tab.

Also game round tracker on Info tab now has interactive Revolt markers on rounds where revolt can start.

### Screenshots
Screenshot 2024-08-12 at 16 17 48

Screenshot 2024-08-12 at 16 23 58
Screenshot 2024-08-12 at 16 32 35
Screenshot 2024-08-12 at 16 32 45

### DIFF
--- a/lib/engine/game/g_22_mars/game.rb
+++ b/lib/engine/game/g_22_mars/game.rb
@@ -168,6 +168,7 @@ module Engine
           },
         ].freeze
 
+        # Game specific constants
         LAST_OR = 11
 
         @or = 0
@@ -179,7 +180,21 @@ module Engine
         def new_operating_round(round_num = 1)
           @or += 1
 
+          if @or == 9
+            @operating_rounds = 3
+            @three_or_round = true
+          end
+
           super
+        end
+
+        def or_round_finished
+          # In case we get phase change during the last OR set we ensure we have 3 ORs
+          @operating_rounds = 3 if @three_or_round
+        end
+
+        def ipo_name(_entity = nil)
+          'Treasury'
         end
 
         def timeline
--- a/lib/engine/game/g_22_mars/game.rb
+++ b/lib/engine/game/g_22_mars/game.rb
@@ -170,8 +170,14 @@ module Engine
 
         # Game specific constants
         LAST_OR = 11
+        FIRST_REVOLT_CHECK_OR = 7
+        LAST_REVOLT_CHECK_OR = 10
 
-        @or = 0
+        def setup
+          @or = 0
+          @revolt_round = -1
+          @three_or_round = false
+        end
 
         def end_now?(_after)
           @or == LAST_OR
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_22_mars/game.rb
```diff
@@ -168,18 +168,103 @@ class Game < Game::Base
...
         def end_now?(_after)
           @or == LAST_OR
         end
 
+        def setup_preround
+          revolts, permits = @companies.partition(&:revolt?)
+          sorted_permits = permits.sort_by { rand }
+          sorted_revolts = revolts.sort_by { rand }
+
+          @companies = sorted_permits + sorted_revolts
```
ollybh, line 191: You're deleting most of the elements from `@companies` before you leave this method. You could just add the companies that will be given to the players:
```suggestion
          @companies = sorted_permits.take(@players.size)
          assigned_companies = sorted_permits.drop(@players.size) + sorted_revolts
```

lib/engine/game/g_22_mars/game.rb
```diff
@@ -168,18 +168,103 @@ class Game < Game::Base
...
+
+          @companies.drop(@players.size).each_with_index do |company, index|
```
ollybh, line 193: `Array.zip` steps through two arrays in sequence.
```
assigned_companies.zip(@corporations).each do |company, corporation|
  if corporation
    corporation.companies << company
    ...
  else
    company.close!
    ...
  end
end
```

lib/engine/game/g_22_mars/game.rb
```diff
@@ -168,18 +168,103 @@ class Game < Game::Base
...
+
+        def revolt_happened?
+          @revolt_happened
+        end
+
+        def check_revolt!
+          return if @or < FIRST_REVOLT_CHECK_OR || revolt_happened?
+
+          # Generate a random number 0 to 3 (2, 1) to simulate 1 in 4 (in OR7, 1 in 3 in OR8 etc) cube draw chance
+          revolt_check = (rand % (LAST_REVOLT_CHECK_OR - @or + 1))
```
ollybh, line 244: Is the timing of the revolt going to be determined by the game seed? If there are other things that call `rand` between operating rounds then this might be non-deterministic, if `rand` could be called a different number of times. An alternative would be to determine the order of cube draws at the start of the game.

lib/engine/game/g_22_mars/step/select_or_discard.rb
```diff
@@ -0,0 +1,52 @@
...
+          def setup
+            @companies = @game.companies
+          end
+
+          def available
+            [@companies.first]
+          end
+
+          def finished?
+            @companies.none?
```
ollybh, line 19: `Array.none?` shouldn't be used unless you are passing it a block. Use `Array.empty?` instead.
```suggestion
            @companies.empty?
```

vandamm:
@ollybh all comments addressed now, thanks for the review!

lib/engine/game/g_22_mars/game.rb
```diff
@@ -168,18 +168,105 @@ class Game < Game::Base
...
+
+        def revolt_happened?
+          @revolt_happened
+        end
+
+        def determine_revolt_round
+          (FIRST_REVOLT_CHECK_OR..LAST_REVOLT_CHECK_OR).each do |round|
+            probability = LAST_REVOLT_CHECK_OR - round + 1
+
+            return round if (rand % probability).zero?
```
ollybh, line 242: This will work, but a simpler version would be:
```suggestion
          (FIRST_REVOLT_CHECK_OR..LAST_REVOLT_CHECK_OR).to_a.sort_by { rand }.first
```

ollybh (approved):

lib/engine/game/g_22_mars/game.rb
vandamm: Thanks for the tip, it should be identical. I'll try it out in the next part

Issue #11128, opened 2024-08-12
18India - commodity destinations and connection runs missing from the map


ddaybell:
It would be nice to have the commodity destinations identified on the map.  It would also be nice to have the connection runs identified on the map.  Alternatively, this information could be published on the Info tab.  I now understand that it's on the 18India wiki page, but that's a resource that many folks are not aware of.

ddaybell:
This can be closed.  PR 11134 fixed this one.

Issue #11068, opened 2024-07-26
[18India] - show commodity values


charliewilsonstaindrop:
The commodity values are not listed anywhere.

Game ID https://18xx.games/game/171527

- Browser Version
  - On Edge
- Platform (Mobile, Tablet, Desktop)
 - Desktop

-->

ddaybell:
This issue was addressed by PR #11134.  This can be closed.

PR #11134, opened 2024-08-13
[18 India] Add Commodity Destination Table Below Map

### LINKED ISSUE #11066: [18 India] Commodity destinations not shown
Not a bug, per se, and only the places where commodities can be picked up are on the map, not the delivery locations.  It would be nice to have this added.

EdPovi:
Fixes #11066, #11068, #11128 

## Before clicking "Create"

## Implementation Notes
NOTE: This PR should not break existing game. 

### Explanation of Change
Added table for Commodity Destination Bonus and Connection Bonus using table like in RoyalGorge.
Also modified the look of Large Icons as the circle "decoration" looks a bit like a city.  

### Screenshots
image
image
image

### Any Assumptions / Hacks
Added commodity price icons, but the hexes are too crowded to add 2-3 icons to the city hexes.  Also the `LargeIcons` feature doesn't work well with more than one Icon.  Plan to customize or create new class to display larger icons.


lib/engine/game/g_18_india/game.rb
```diff
@@ -1230,6 +1230,222 @@ def price_movement_chart
...
+            [
+              { text: 'Karachi (A16) <=> Chennai (K30)', props: { style: cell_style } },
+              { text: format_currency(80), props: { style: cell_style } },
+            ],
+            [
+              { text: 'Lahore (D3) <=> Kolkata (P17)', props: { style: cell_style } },
+              { text: format_currency(80), props: { style: cell_style } },
+            ],
+            [
+              { text: 'Nepal (M10) <=> Mumbai (D23)', props: { style: cell_style } },
```
ollybh, line 1288: Unicode support should be universal these days, so we don't have to use Ascii compositions for arrows:
```suggestion
              { text: 'Delhi (G8) ⟷ Kocchi (G36)', props: { style: cell_style } },
              { text: format_currency(100), props: { style: cell_style } },
            ],
            [
              { text: 'Karachi (A16) ⟷ Chennai (K30)', props: { style: cell_style } },
              { text: format_currency(80), props: { style: cell_style } },
            ],
            [
              { text: 'Lahore (D3) ⟷ Kolkata (P17)', props: { style: cell_style } },
              { text: format_currency(80), props: { style: cell_style } },
            ],
            [
              { text: 'Nepal (M10) ⟷ Mumbai (D23)', props: { style: cell_style } },
```

ollybh:
This looks good, but I've got a couple of suggestions:

1. The commodity icons would look better with no surrounding shape. Return a `{ shape: :none }` from `decorate_markers` and update the case statement in `View::Game::Part::LargeIcons.render_part` to handle this.
2. Add an extra column to the commodities table to show the hex coordinates of each city.

EdPovi:
> 1. The commodity icons would look better with no surrounding shape. Return a `{ shape: :none }` from `decorate_markers` and update the case statement in `View::Game::Part::LargeIcons.render_part` to handle this.

I actually did almost exactly that at first, and then reverted to minimize Core changes.
This gives me the support to make this change and maybe even extend it further to support multiple large icons on a single hex.  This requires looking at setting alternate locations based on number of large icons.  But I think I can use Icon information and the decorate info to control what happens. 

> 2. Add an extra column to the commodities table to show the hex coordinates of each city.

This should be easy enough to do.

EdPovi:
Updated Table and added hex icons using by adapting `large_icons` to support up to 3 large icons in an empty hex. 

image
image

ollybh:
This has improved the appearance of the icons a lot.

The one thing I'm not sure is right is the `loc:hex` attribute on the large icon tile part. If someone created a tile which had some large icons with this set and some without, then only the ones with `loc:hex` would be rendered and the others would be not shown. This could cause confusion.

Does this render at all sensibly if you just add the large icons without changing the code for their positions? It might be worth getting the rest of this pull request merged without the `render_hex_icons` bit, and separating that into a separate pull request.

EdPovi:
> Does this render at all sensibly if you just add the large icons without changing the code for their positions? It might be worth getting the rest of this pull request merged without the `render_hex_icons` bit, and separating that into a separate pull request.

Unfortunately, it doesn't the original code for `large_icons` doesn't support more than one icon, and they noted that limitation in the code: `# multiple large icons not supported`

Also, using `loc:hex` is intended to be used only for icons that will be rendered that way.  Using both types would not work, which is why set the code path to use one or the other render path.  Also, this is intended only for **empty** hexes as the position make no consideration for any other items on that hex.

ollybh:
> Unfortunately, the original code for `large_icons` doesn't support more than one icon, and they noted that limitation in the code: `# multiple large icons not supported`. Therefore this alternate path is necessary if I place more than one on hex.

The code does handle multiple large icons (there's an iterator in `render_part`, it just doesn't handle them well:
image

ollybh:
Could these icons be added with this sub-optimal rendering? I'm happy to approve the rest of this pull request, but would need one of the more experienced developers to approve the changes to the large icon positioning, so it might be better split into its own pull request.

EdPovi:
> Could these icons be added with this sub-optimal rendering? I'm happy to approve the rest of this pull request, but would need one of the more experienced developers to approve the changes to the large icon positioning, so it might be better split into its own pull request.

I understand and I can split the alternate rendering to a separate PR.

ollybh (approved):

PR #11169, opened 2024-08-22
[SteamOverHolland] fixes train discount bug

### LINKED ISSUE #11163: [Steam over Holland] can only use p6 if full amount in treasury
https://18xx.games/game/174408?action=235

The company should be able to buy a 4T for 270 but can only do so it it has 300.

philcampeau:
Fixes #11163

## Before clicking "Create"

### Explanation of Change

Changes the emergency_issuable_bundles method to read @depot.min_price(ability: nil) instead of @depot.min_depot_price

I had to remove a fixture file that broke as a result. Also, this will break any alpha games with this private, and possibly all of them.


lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -422,7 +422,7 @@ def redeemable_shares(entity)
         end
 
         def emergency_issuable_bundles(entity)
-          return [] if @round.issued_shares[entity] || entity.cash >= @depot.min_depot_price
+          return [] if @round.issued_shares[entity] || entity.cash >= @depot.min_price(ability: nil)
```
ollybh, line 425: The `min_price` method has a `corporation` parameter. Omitting this parameter is probably going to cause an exception when run in server mode (and doesn't make any sense, if you're not trying to find the cheapest train for a specific corporation, then `depot_min_price` is the right method to use).
https://github.com/tobymao/18xx/blob/38a729965ef394d5cd3ace16e45952672fa64aac/lib/engine/depot.rb#L54-L56

ollybh (changes requested):

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: Yep, 100% right. I meant to include that. Thanks!

ollybh: Does the train discount from the private still apply if you're buying a train during emergency money raising? I think you will need to change the call to `min_price` for this to work: `@depot.min_price(entity, abilities(entity, :train_discount)`.

philcampeau: When I change it to this, I get the following error in the console: 

```
ArgumentError: expected kwargs
    at Opal.send2 (deps.js:1958:17)
    at Opal.send (deps.js:1946:17)
    at Function.$$exception (deps.js:6052:14)
    at Function.$$raise (deps.js:5653:31)
    at $raise (deps.js:146:16)
    at Opal.ensure_kwargs (deps.js:2942:7)
    at klass.$$min_price (main.js:6790:19)
    at klass.$$emergency_issuable_bundles (g_steam_over_holland.js:414:116)
    at klass.$$issuable_shares (main.js:30539:28)
    at klass.$$render (main.js:49656:66)
    at klass.$$h (deps.js:27934:28)
    at klass.$$render (main.js:56432:34)
    at klass.$$h (deps.js:27934:28)
    at klass.$$render_action (main.js:61784:25)
    at klass.$$render_game (main.js:61831:30)
    at klass.$$render (main.js:61460:74)
    at klass.$$h (deps.js:27934:28)
    at klass.$$render_game (main.js:38001:19)
    at klass.$$render_content (main.js:37969:275)
    at klass.$$render (main.js:37957:154)
    at klass.$Component_update$excl$2 (deps.js:27962:26)
    at deps.js:27951:109
```

ollybh: Yes, my mistake. It should be `@depot.min_price(entity, ability: abilities(entity, :train_discount)`.

The "expected kwargs" error means we tried passing a parameter when a named parameter was needed.

philcampeau: hmmm something's wrong now. I did a print of @depot.min_price(entity,ability: abilities(entity,:train_discount)), and it returns 1, rather than the price of the cheapest depot train minus 10%.

I'm gonna have to take a look at this in a couple of days. I'm going to be out all weekend.

philcampeau: taking a quick look, even my code before was returning 1. So I'm still not pulling from the right spot. Will investigate more in a couple days.

lib/engine/game/g_steam_over_holland/entities.rb
```diff
@@ -116,6 +116,8 @@ module Entities
                 type: 'train_discount',
                 discount: 0.1,
                 trains: %w[3 4 5],
+                when: 'buy_train',
```
ollybh, line 119:
```suggestion
                when: 'buying_train',
```

lib/engine/game/g_steam_over_holland/game.rb
ollybh: 1 will be the price for buying a train from another company.

lib/engine/game/g_steam_over_holland/entities.rb
philcampeau: I'm pretty sure we've established repeatedly that 'buying_train' leads to an issue where the associated ability doesn't work when it's the last remaining ability, or something of the sort.

Examples: 
https://github.com/tobymao/18xx/pull/10098
https://github.com/tobymao/18xx/pull/10318

Then again, other abilities in this game don't ever close, so this may be a non-issue for this game. What do you think?

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: yep. I submitted a new update, but it's still causing issues. the battle continues.

lib/engine/game/g_steam_over_holland/entities.rb
ollybh: I was just going off the `when` values listed in ability/README.md. This has `buying_train` but not `buy_train`.

Looking at those issues you've linked this doesn't appear to have been fully resolved.

lib/engine/game/g_steam_over_holland/game.rb
```diff
@@ -423,8 +423,11 @@ def redeemable_shares(entity)
 
         def emergency_issuable_bundles(entity)
           return [] if @round.issued_shares[entity]
+          return [] if entity.all_abilities.find { |a| a.type == :train_discount } && entity.cash >= @depot.min_depot_price * 0.9
```
ollybh, line 426: You're splitting the game logic up: for most cases you're checking whether the corporation can afford a train in `Step::BuyTrain.can_afford_neeed_train?`. But if the corporation has the discount ability you test against the reduced price here.

It would make the logic clearer if you instead modified `can_afford_needed_train?` to test against the discounted price.

lib/engine/game/g_steam_over_holland/entities.rb
philcampeau: well, annoyingly, this does seem to fix the issue entirely. So for the time being, I'm using this.

lib/engine/game/g_steam_over_holland/game.rb
philcampeau: resolved

philcampeau:
I've just submitted a new commit, which seems to resolve the issue, and shouldn't break any games.

ollybh (commented):
I think that there might still be a potential bug lurking in here: part of the test at the start of `Step::BuyTrain.must_sell_shares?` is whether the corporation has more cash than the face value of the cheapest train. What happens if the corporation that owns the Werkspoor private has no train and between 90% and 100% the price of the cheapest train? Will they be able to sell shares, even though this isn't needed to buy the train.
You might need to change this test to use the `Depot.min_price(corporation, ability)` function.

bentziaxl:
no activity since sep. Closing. Reopen if needed

PR #11186, opened 2024-09-01
[1858Switzerland] Add game code for 1858 Switzerland variant

ollybh:
### Explanation of Change

This adds code handling the differences between the original 1858 game, and the prototype of the variant set in Switzerland.

These differences include:

- Bonuses for running north-south or east-west routes.
- Three sets of private railway companies instead of two.
- Two private closure rounds instead of one.
- Different timings for wounding and rusting mid-game trains.
- Mountain railway tiles which can be laid on hexes with a 120SF terrain cost. A public company that has laid one of these will then get a 40SF bonus to its revenue whenever it runs trains.
- Special rules for the private railway companies which have Gotthard as a home hex. Normally a public company is allowed to acquire a private railway if there is a connection to anywhere in the private's home hex; for GB you need a connection to the broad gauge track in the Gotthard hex, and for FOB you need a connection to the narrow (metre) gauge track.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1858_switzerland/game.rb
+++ b/lib/engine/game/g_1858_switzerland/game.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'entities'
+require_relative 'graph'
 require_relative 'map'
 require_relative 'meta'
 require_relative 'tiles'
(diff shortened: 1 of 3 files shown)


bentziaxl (commented):
Code looks good.
Only minor concern, and it might be nitpicky, it's not very readable to filter companies based on color.
I know that base 1858 does that, but the coloring there matches yellow/green/brown/gray scheme, so it's clear that a green private comes out on green phase.
lightblue is not clear in this scheme, my concern is that if the company color changes, it will break the game completely.

Maybe it makes more sense to handle it like 1846, where privates are stored in an array, and the filtering happens that way.
 ORANGE_GROUP = [....], etc.

PR #11202, opened 2024-09-10
[Core] Add subsidies to dividend step

crericha:
Add generic implementation of subsidies in the dividend step.

Previously, games that used subsidies, copy and pasted the subsidy logic into their dividend step. This PR moves that copy and pasted code into the base dividend step. If a game does not have a subsidy, the added logic has no effect.

### DIFF
--- a/lib/engine/game/g_1822/step/dividend.rb
+++ b/lib/engine/game/g_1822/step/dividend.rb
@@ -41,15 +41,13 @@ module Engine
             if extra_train
               train = find_extra_train(entity)
               extra_train_revenue = routes.find { |r| r.train == train }.revenue
-              extra_train_payout = send(@extra_train_choice, entity, extra_train_revenue, 0)
+              extra_train_payout = send(@extra_train_choice, entity, extra_train_revenue)
               revenue = total_revenue - extra_train_revenue
             else
               revenue = total_revenue
             end
-            subsidy = @game.routes_subsidy(routes)
-            total_revenue += subsidy
             dividend_types.to_h do |type|
-              payout = send(type, entity, revenue, subsidy)
+              payout = send(type, entity, revenue)
               if extra_train
                 payout[:corporation] += extra_train_payout[:corporation]
                 payout[:per_share] += extra_train_payout[:per_share]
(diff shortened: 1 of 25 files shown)


crericha:
This PR is not complete yet. Posting main code for feedback before I update affected game code.

roseundy:
If you want to update 60/62 you will probably want to take to a look at app/view/game/run_routes as well.

crericha:
> If you want to update 60/62 you will probably want to take to a look at app/view/game/run_routes as well.

run_routes seemed to work out of the box when `routes_subsidy` returns a positive number, but I'll double check.

lib/engine/game/g_1822/step/dividend.rb
```diff
@@ -92,8 +90,8 @@ def log_run_payout(entity, kind, revenue, subsidy, _action, payout)
             @log << "#{entity.name} earns subsidy of #{@game.format_currency(subsidy)}" if subsidy.positive?
```
perwestling, line 90: Subsidy is a parameter in this method but has no effect more than appearing in the log message if positive.

lib/engine/game/g_1856/step/dividend.rb
```diff
@@ -75,7 +43,11 @@ def holder_for_corporation(_entity)
             @game.share_pool
           end
 
-          def log_run_payout(entity, kind, revenue, action, payout)
+          def total_revenue
+            super - (@round.interest_penalty[@round.current_operator] || 0)
```
perwestling, line 47: I think the previous construct might be a bit more clearer

            penalty = @round.interest_penalty[entity] || 0
            revenue = @game.routes_revenue(routes) - penalty
            
            i.e. to put the @round.in.... into penalty.

lib/engine/game/g_1822/step/dividend.rb
crericha: That's correct. The purpose of the method is to log and there is a log for the subsidy if present.

lib/engine/game/g_1856/step/dividend.rb
crericha: penalty is actually a bad term as its not a penalty, but more so garnishing of revenue.

crericha:
I ran this branch against the entire db of games and no failures were found.

PR #11205, opened 2024-09-11
[18SJ] Corrects edge case nationalization calculation

philcampeau:
## Before clicking "Create"

### Explanation of Change

Currently, the nationalization method uses this code to determine which corp of the eligible candidates will be nationalized:

```
          merged = candidates.max_by { |c| [c.share_price.price, -@round.entities.find_index(c)] }
```

The problem is that the second sorting criteria uses the initial round order, meaning that if one corp runs and withholds, and moves into the same space as another corp, and they then are both eligible for nationalization, the corp that withheld will be nationalized, even though its token is under the other corp. This issue was reported on boardgamegeek here: https://boardgamegeek.com/thread/3366312/nationalisation-the-first-corporation-in-descendin

I've replaced this with the sorting used by `def sort_order_key` in lib/engine/corporation.rb:

```
          merged = candidates.max_by do |c|
            [c.share_price.price, -c.share_price.coordinates.last, -c.share_price.coordinates.first]
          end
```

which produces the desired result. 

This will require pins for any games that have done the wrong nationalization, of course.


perwestling (approved):
Looks good.

lib/engine/game/g_18_sj/game.rb
```diff
@@ -788,8 +788,10 @@ def perform_nationalization
             return
           end
 
-          # Merge the corporation with highest share price, and use the first operated as tie break
-          merged = candidates.max_by { |c| [c.share_price.price, -@round.entities.find_index(c)] }
+          # Merge the corporation with highest share price, and use the first to operate as tie break
+          merged = candidates.max_by do |c|
+            [c.share_price.price, -c.share_price.coordinates.last, -c.share_price.coordinates.first]
```
ollybh, line 793: I do not believe that this is sufficient (and the comment is out-of-date). There is no test for position when you have multiple tokens on the same space on the stock market. I also don't think you need to define custom sort order, the default specified in `Corporation.<=>` should be correct, so you could just do:
```suggestion
          # Merge the corporation first in share price order.
          candidates.max
```

philcampeau: This doesn't work. I tried implementing `merged = candidates.max` (or alternately `nationalize_major(candidates_max)`, and it selects the wrong corp. Worse, it selects the lowest priced corp instead of the highest. 

Look at the game for the player who reported the issue: https://18xx.games/game/176266?action=405 This code broke the game because of the previous merger at the 4 train. At the time of the merge, there are two corps left trainless, SNJ at 67 and KFJ at 30. When buying the 4 train with this code, KFJ closes. I even added some code to inspect what's happening: 

```
          print candidates

          print "candidates.max: #{candidates.max.id} : #{candidates.max.share_price}"
          print "candidates.min: #{candidates.min.id} : #{candidates.min.share_price}"
```

Here's the output in Console: 

image

lib/engine/game/g_18_sj/game.rb
```diff
@@ -788,8 +788,10 @@ def perform_nationalization
             return
           end
 
-          # Merge the corporation with highest share price, and use the first operated as tie break
-          merged = candidates.max_by { |c| [c.share_price.price, -@round.entities.find_index(c)] }
+          # Merge the corporation with highest share price, and use highest token on the pile as tie break
+          merged = candidates.max_by do |c|
+            [c.share_price.price, c.share_price.coordinates.last, c.share_price.corporations.find_index(c)]
```
ollybh, line 793: The share_price corporations array is ordered such that the token on the top of the stack is first, with the index increasing as you move down the pile of tokens. So you'd need to use `-share_price.corporations.index(c)` here, otherwise ties will be broken in favour of the tokens lower down the stack.

And I've checked the code for the default sorting of corporations, and it's the opposite of what I thought. So `candiates.min` should work for you.
https://github.com/tobymao/18xx/blob/011ed24d0e169c1f9a203105fe9078561591a0e1/lib/engine/corporation.rb#L93-L106

ollybh (changes requested):

lib/engine/game/g_18_sj/game.rb
philcampeau: yep that works too. Done.

ollybh (approved):

PR #11207, opened 2024-09-11
[18GA] Implements Cotton Port variant

philcampeau:
## Before clicking "Create"

### Explanation of Change

Implements the Cotton Port variant, designed by the game designer and described here: https://boardgamegeek.com/image/603446/18ga?commentid=1289013#comment1289013


lib/engine/game/g_18_ga/game.rb
```diff
@@ -40,6 +40,16 @@ class Game < Game::Base
 
         GAME_END_CHECK = { bankrupt: :immediate, stock_market: :current_or, bank: :current_or }.freeze
 
+        OPTION_REMOVE_HEXES = ['G13'].freeze
```
perwestling, line 43: Minor, but there are other options in 18GA that removes tiles (rule 7d) so consider including COTTON or COTTON_PORT in the game. Same of OPTION_ADD_HEXES.

I suppose a future implementation can do the renaming if needed, but maybe then a comment before line 43 that these two belong to the Cotton Port variant?

philcampeau: For what it's worth, rule 7d removes tiles. These are map hexes that are being removed. No other variant does that.

philcampeau: Done anyway.

perwestling (approved):

lib/engine/game/g_18_ga/game.rb
```diff
@@ -193,6 +203,47 @@ def event_close_companies!
...
+            end
+            COTTON_ADD_HEXES[color]&.each { |coords, tile_str| new_map[coords] = tile_str }
+            new_hexes[color] = new_map
+          end
+
+          new_hexes
+        end
+
+        def place_home_token(corporation)
+          return if corporation.tokens.first&.used == true
```
ollybh, line 235: Don't compare a boolean to true.
Do you need the safe navigation operator? The only way you could get a nil is if the corporation doesn't have any tokens.
```suggestion
          return if corporation.tokens.first.used
```

lib/engine/game/g_18_ga/game.rb
```diff
@@ -193,6 +203,47 @@ def event_close_companies!
...
+
+        def charleston
+          @charleston ||= hex_by_id('E15')
+        end
+
+        def optional_hexes
+          return self.class::HEXES unless @optional_rules&.include?(:cotton_port)
+
+          new_hexes = {}
+          HEXES.keys.each do |color|
```
ollybh, line 223:
```suggestion
          HEXES.each_key do |color|
```

lib/engine/game/g_18_ga/game.rb
philcampeau: yep, completely right. I think I copied this code from 1847AE. Thanks for that. Fixed.

lib/engine/game/g_18_ga/game.rb
```diff
@@ -193,6 +203,47 @@ def event_close_companies!
         def remove_icon_from_waycross
           waycross_hex.tile.icons = []
         end
+
+        def georgia_railroad
+          corporation_by_id('GA')
```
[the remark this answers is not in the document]
philcampeau: done

lib/engine/game/g_18_ga/game.rb
philcampeau: done, though I'll note that every other instance of this type of code on the site uses HEXES.keys.each.

this works too, so I'll go with it.

ollybh: > done, though I'll note that every other instance of this type of code on the site uses HEXES.keys.each.

All the other code is wrong ;-) Using Hash.keys.each (at least in Opal) creates a new array containing a copy of the hash's keys, whereas Hash.each_key directly uses the hash to iterate, without needing to create a new array.

lib/engine/game/g_18_ga/game.rb
```diff
@@ -193,6 +203,51 @@ def event_close_companies!
...
+          georgia_railroad.coordinates.each do |coordinate|
+            hex = hex_by_id(coordinate)
+            tile = hex&.tile
+            tile.cities.first.place_token(georgia_railroad, georgia_railroad.next_token)
+          end
+          georgia_railroad.coordinates = [georgia_railroad.coordinates.first]
+        end
+
+        def cotton_port?
+          @cotton_port || @optional_rules.include?(:cotton_port)
```
ollybh, line 249:
```suggestion
          @cotton_port ||= @optional_rules.include?(:cotton_port)
```

ollybh (changes requested):
Just need the fix in the `cotton_port?` method, then this will be good to go.

ollybh (approved):

ollybh:
I think the build failures are coming from #11275, not this pull request. No idea why the build checks weren't failing for that pull request. I think it's just a game JSON file that needs removing from the tests.

PR #11208, opened 2024-09-12
[18Uruguay] Company should get full cap after nationalization and no destination bonus

### LINKED ISSUE #11060: [18Uruguay] company not receiving full capital after nationalization
https://18xx.games/game/170866?action=726

company should have 10x par value in treasury according to 7.6.3 after nationalization

patrikolesen:
Company should get full cap but no destination bonus after nationalization 

Fixes #11060

## Before clicking "Create"

This can break existing games


ollybh (approved):
This looks fine.
I assume that the extra money in the companies might break some existing games, so these will need archiving or pinning. Right?

patrikolesen:
Yes exactly, the extra money might break some existing game.

PR #11210, opened 2024-09-12
[18Uruguay] Do not ship more than ship capacity for rptla

### LINKED ISSUE #11140: 18Uruguay - RTPLA liquidation shipping phase does not follow the printed rules
https://18xx.games/game/172678?action=711

When the RPTLA does its final shipping phase as part of being liquidated, it is ignoring the shipping capacity of the ships it owns.  As I understand it, a 4-ship can carry 2 goods, and a 5-ship can carry 3 goods.  But here, the RPTLA was allowed to ship 8 goods (instead of 5) in its final shipping phase.  The rules provide as follows:

10.2.1 RPTLA - Final Export and Payout
Move all cubes in Ports to the Frigorificos Anglo-box.
Then export cubes from the Frigorificos Anglo-box
to the European Market equalling the Total Shipping Capacity and calculate the Export Revenue, 50
per cube exported. 

In the final export phase, the RPTLA still is limited to the total shipping capacity.  This bug has been encountered in all of the 18Uruguay games I have played that have reached liquidation phase.

patrikolesen:
Fixes #11140

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_uruguay/game.rb
+++ b/lib/engine/game/g_18_uruguay/game.rb
@@ -32,6 +32,16 @@ module Engine
         include Goods
         include Nationalization
 
+        SHIP_CAPACITY =
+          {
+            'Ship 1' => 1,
+            'Ship 2' => 1,
+            'Ship 3' => 2,
+            'Ship 4' => 2,
+            'Ship 5' => 3,
+            'Ship 6' => 3,
+          }.freeze
+
         EBUY_SELL_MORE_THAN_NEEDED = true
         EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false
 
@@ -534,6 +544,10 @@ module Engine
           return super if nationalized?
 
           super.reject { |c| c == @rptla }.append(@rptla)
+        end
+
+        def ship_capacity(train)
+          SHIP_CAPACITY[train.name.partition('+')[0]]
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_uruguay/game.rb
```diff
@@ -535,6 +545,10 @@ def operating_order
 
           super.reject { |c| c == @rptla }.append(@rptla)
         end
+
+        def ship_capacity(train)
+          SHIP_CAPACITY[train.name.partition('+')[0]]
```
ollybh, line 550: What's the call to partition doing here? Isn't `train.name` going to be something like `Ship 3` with no plus sign in it?

lib/engine/game/g_18_uruguay/step/route_rptla.rb
```diff
@@ -55,13 +55,9 @@ def choices
             choices
           end
 
-          def ship_capacity(train)
-            SHIP_CAPACITY[train.name.partition('+')[0]]
```
ollybh, line 59: As you're removing this method from this class, the `SHIP_CAPACITY` constant at the start of this file is no longer needed.

ollybh (changes requested):

lib/engine/game/g_18_uruguay/game.rb
patrikolesen: You are correct this was legacy code, removed

Also added nil check in ship_capacity

lib/engine/game/g_18_uruguay/step/route_rptla.rb
patrikolesen: Removed

ollybh (approved):

PR #11214, opened 2024-09-12
[18Uruguay] Company shall only retreive destination_bonus in their own turn

### LINKED ISSUE #11073: [18uruguay] - Company received destination cash on another company's turn
-->](https://18xx.games/game/171349?action=771)

The FCNE received its destination cash when the EER laid the connecting tile.  The rules say that a company gets its cash on its own turn.

8.2.4: "The first time, during its own turn, that a com- pany can trace a run with an infinite train from home token to destination city, the company gets an additional 5 times its par value from the bank (If floated before nationalization)."

patrikolesen:
[18Uruguay] Company shall only retreive destination_bonus in their own turn

Fixes #11073

Fixes #[PUT_ISSUE_NUMBER_HERE]

## Before clicking "Create"

Will break existing games

### DIFF
--- a/lib/engine/game/g_18_uruguay/step/destination_bonus.rb
+++ b/lib/engine/game/g_18_uruguay/step/destination_bonus.rb
@@ -35,11 +35,11 @@ module Engine
           end
 
           def destination_node_check?(corporation)
-            return if corporation.closed?
-            return if corporation.destination_coordinates.nil?
+            return false if corporation.closed?
+            return false if corporation.destination_coordinates.nil?
 
             ability = @game.abilities(corporation, :destination_bonus)
-            return if ability.nil?
+            return false if ability.nil?
 
             destination_hex = @game.hex_by_id(corporation.destination_coordinates)
             home_node = corporation.tokens.first.city
--- a/lib/engine/game/g_18_uruguay/step/destination_bonus.rb
+++ b/lib/engine/game/g_18_uruguay/step/destination_bonus.rb
@@ -12,8 +12,10 @@ module Engine
           end
 
           def log_skip(entity)
-            return if entity.minor?
+            return unless entity.corporation?
             return if entity == @game.rptla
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_uruguay/step/destination_bonus.rb
```diff
@@ -12,24 +12,27 @@ def description
           end
 
           def log_skip(entity)
-            return if entity.minor?
+            return unless entity.corporation?
             return if entity == @game.rptla
+            return if @game.abilities(entity, :destination_bonus).nil?
+            return if destination_node_check?(entity).nil?
```
ollybh, line 18:
```suggestion
            return unless destination_node_check?(entity)
```

1. You're checking for a destination bonus ability in `destination_node_check?` so you don't need to repeat that test here.
2. A method that has a question mark in its name should only return `true` or `false`, so testing `nil?` looks wrong.
3. It would be better to change the guard clauses in `destination_node_check?` from `return if …` to `return false if …`.

lib/engine/game/g_18_uruguay/step/destination_bonus.rb
```diff
@@ -12,24 +12,27 @@ def description
...
+            return [] if @game.abilities(entity, :destination_bonus).nil?
 
             self.class::ACTIONS
           end
 
           def auto_actions(entity)
-            corporations = @round.entities.select { |c| destination_node_check?(c) }
-            return [Engine::Action::Pass.new(entity)] if corporations.empty?
+            return [] unless entity.corporation?
+            return [Engine::Action::Pass.new(entity)] if destination_node_check?(entity).nil?
```
ollybh, line 33:
```suggestion
            return [Engine::Action::Pass.new(entity)] unless destination_node_check?(entity)
```

ollybh (changes requested):

ollybh (approved):

PR #11226, opened 2024-09-16
[auto-routing] Clears previous routes before running

### LINKED ISSUE #10136: [autoroute] clear routes if any connectivity changed
Prior train routes are maintained across ORs until cleared, to make it easier for manual routing

However for auto-routing, it adds the necessary step to clear existing routes, then hit Auto to re-calculate - since otherwise the existing routes might result in lower payouts

Perhaps we can call the route cache to see if anything new was connected, and if so then clear all routes. Possibly only in the case where Auto-routing is enabled

I grepped the issue history but didn't see a similar issue

philcampeau:
Fixes #10136

## Before clicking "Create"

### Explanation of Change

Now, when a player clicks the Auto button, it doesn't update the routes. This change causes the Auto button to clear the existing routes before running its script. 

I think this makes more sense than the current application. 

In addition to saving everyone clicks, I had a new player tell me well into our second game on the site that he had just then realized that auto-routing wasn't updating his previous routes, and he'd been running under for the majority of the games we had played.

Open to feedback.

### DIFF
--- a/assets/app/view/game/route_selector.rb
+++ b/assets/app/view/game/route_selector.rb
@@ -264,7 +264,7 @@ module View
         end
 
         auto = lambda do
-        @routes.each(&:reset!)
+          @routes.each(&:reset!)
           router = Engine::AutoRouter.new(@game, flash)
           @routes = router.compute(
             @game.current_entity,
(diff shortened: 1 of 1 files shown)


roseundy:
I disagree with the intent of this PR.

philcampeau:
Yeah, having read your argument against it on the original request, I agree. Closing.

PR #11259, opened 2024-10-06
Move the auto-router path timeout check to an inner loop

### LINKED ISSUE #11252: Autorouter taking too long to time out
Issue #11239 was a report of being unable to select a route. This was probably caused by the autorouter being started and locking up the interface.

This occurred at [action 792 of game 177538](https://18xx.games/game/177538?action=792). This game exhibits horrendously bad autorouter performance: it is a game of 1858 which has Lawson track and a lot of tiles have been laid with many open cities. This means that the autorouter is going to struggle, but the problem here is that it is not timing out correctly: with the default setting of 30 seconds for path timeout, it took more than 10 *hours* to timeout on my desktop computer (which is far from being new or high-spec, but 😱).

```
D, [2024-09-30T22:26:36.021000] DEBUG -- : Rendering game view...
D, [2024-09-30T22:26:36.231000] DEBUG -- : Done rendering game view
D, [2024-09-30T22:26:39.094000] DEBUG -- : Path search: 0 / 57 - paths starting from L13
D, [2024-10-01T08:44:27.660000] DEBUG -- : Path timeout reached
D, [2024-10-01T08:44:27.661000] DEBUG -- : Evaluated 3791 paths, found 232 unique hexsides, and found valid routes 7E:3054 in: 37068.567
D, [2024-10-01T08:44:27.666000] DEBUG -- : Finding route combos of best 7E:3054 routes with depth 3054
D, [2024-10-01T08:44:27.892000] DEBUG -- : Found 741 possible combos (5 best) and rejected 0 conflicting combos in: 0.225
D, [2024-10-01T08:44:27.899000] DEBUG -- : Rendering game view...
D, [2024-10-01T08:44:50.770000] DEBUG -- : Done rendering game view
D, [2024-10-01T08:44:51.202000] DEBUG -- : Rendering game view...
D, [2024-10-01T08:45:14.141000] DEBUG -- : Done rendering game view
```

The path timeout is checked once in the autorouter's outer loop, iterating over the corporation's reachable nodes and looking at the potential routes from each. In this case it took multiple hours to check the nodes from the first node, and then only timed out when moving onto the second node.

Moving the timeout check into an inner loop won't fix the performance of the autorouter, but will at least means it times out much more quickly.

ollybh:
### Explanation of Change

The auto-router's path timeout is supposed to stop too much time being spent on building up the set of possible routes for trains. It was being checked once in each iteration of possible starting nodes for routes.

However, in particularly degenerate cases this meant that far too long was elapsing between timeout checks: issue tobymao#11252 had more than ten hours pass before the first and second timeout checks.

This moves the timeout check to an inner loop, so it is checked more frequently.

Fixes #11252.

### Any Assumptions / Hacks

This uses a custom error class to break out of the loop if too long has elapsed. There are other ways that this could be done, I'm not sure if this is the best.

## Before clicking "Create"


lib/engine/auto_router.rb
```diff
@@ -64,16 +64,12 @@ def compute_for_train_group(trains, corporation, **opts)
...
-          path_walk_timed_out = true
-          break
-        else
-          LOGGER.debug { "Path search: #{nodes.index(node)} / #{nodes.size} - paths starting from #{node.hex.name}" }
-        end
+        LOGGER.debug { "Path search: #{nodes.index(node)} / #{nodes.size} - paths starting from #{node.hex.name}" }
 
         walk_corporation = graph.no_blocking? ? nil : corporation
         node.walk(corporation: walk_corporation, skip_paths: skip_paths) do |_, vp|
+          raise PathTimeout if Time.now - now > path_timeout
```
tobymao, line 71: can you only do the time.now check every N loops?

ollybh:
I’ve not had time to lock into the performance implications of this change. I've changed this PR’s status to draft until I’ve done this.

bentziaxl:
closing the PR, test the performance locally, then reopen when ready to merge

PR #11270, opened 2024-10-11
[18ESP] sets minimum buy-in price for P6 and P7 to 1

### LINKED ISSUE #11265: [18ESP] Minimum for buying privates is $1
https://18xx.games/game/181479

For buying the private company "Ferrocarril Vasco-Navarro" from me the program suggests a minimum amount of $80 (i.e. 50% of the initial value), but it should be $1.

philcampeau:
Fixes #11265

## Before clicking "Create"

### Explanation of Change

Privates 6 and 7 didn't have the min_price set to 1. 

I also corrected the description of the P6 variant that comes with a random share.


lib/engine/game/g_18_esp/entities.rb
```diff
@@ -119,7 +120,8 @@ module Entities
             name: 'Ferrocarril Vasco-Navarro',
             value: 160,
             revenue: 20,
-            desc: 'It provides a 10% certificate from the Southern company CRB.',
+            min_price: 1,
```
bentziaxl, line 123: This isn't needed. There are 2 P6 companies, one picks at random, one picks the CRB. Depends on which companies exist in the game.

see game.setup_companies for the setup logic

philcampeau: Ah yeah I see it now. Interesting approach. Fixed.

ollybh (approved):

PR #11275, opened 2024-10-12
System18: Remove discarded trains from the game

### LINKED ISSUE #11264: [System18: Poland] Discarded train available for purchase
https://18xx.games/game/181169

OR5.1
SPX is operating for the first time, and gets to the train buying step. There is a 3T in the discard and it is being offered to buy it. The rules are:
"A company that owns more trains than currently allowed must discard excess trains. Any trains that are discarded are _removed from the game_."
The company cannot buy the 6T because there is a cheaper train available i.e. the 3T - this is correct (if the 3T was really available).

This will presumably also affect other maps.

vandamm:
Fixes #11264

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_system18/game.rb
+++ b/lib/engine/game/g_system18/game.rb
@@ -460,7 +460,7 @@ module Engine
               GSystem18::Step::Token,
               Engine::Step::Route,
               GSystem18::Step::Dividend,
-              Engine::Step::DiscardTrain,
+              GSystem18::Step::DiscardTrain,
               GSystem18::Step::BuyTrain,
             ]
           else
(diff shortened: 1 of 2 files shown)


roseundy:
Sorry, this is not the correct way to fix this.

Take a look at the variable DISCARDED_TRAINS

roseundy (changes requested):
Please look at the variable defined in the Game base class: DISCARDED_TRAINS
This just needs to be changed in the map method map_*_constants

vandamm:
@roseundy sorry did not find this and found 1822 implementation instead. 
Fixed to using `DISCARDED_TRAINS = :remove` in the base System18 because it's a base rule. Checked all implemented maps and none of them need to override it per the rules

roseundy (approved):

roseundy:
We are still archiving System18 games, not pinning them.

PR #11302, opened 2024-10-24
[System18] Remove broken fixture file

ollybh:
The game 182070 fixture file for System18 includes buying a previously discarded train at action 258. PR tobymao#11275 has changed it so that discarded trains are removed from the game, so there is no longer a train to buy, breaking this fixture.

[CC @roseundy]


roseundy:
Thanks

PR #11312, opened 2024-11-01
[1856] must sell in blocks when forced to sell

### LINKED ISSUE #11311: 1856 Id: 183019 - allows shares to be sold individually when doing forced stock sales
This is incorrectly implemented to allow shares to be sold one at a time, lowering the stock price for each sale.  Per the game rules they are sold as one block

philcampeau:
Fixes #11311

## Before clicking "Create"

### Explanation of Change

in 1856 the rules state that forced sales must be done in blocks. While we could write more complicated code that only enforces this during forced sales, there's really zero reason anyone would not sell in blocks during an SR, since share price drops by share sold, not by block sold. 

I've set MUST_SELL_IN_BLOCKS to true for the game, which would also impact 1836jr56. As a result, I had to remove one fixture game for 36jr56 which failed Rake. 

Let me know if you disagree with this approach. Some games may need pinning.


ollybh:
Did the rules interpretation about selling in blocks get confirmed? I recall seeing some messages about this on the 18xx mailing list.

philcampeau:
The rule wasn't clarified on the list, but like I pointed both there and above, there really is no functional reason why you wouldn't sell in blocks during the Stock Round in a game without price protection.

crericha (approved):

PR #11314, opened 2024-11-03
[18Uruguay] New updated tile sheet from designer

patrikolesen:
New updated tile sheet from the Designer

## Before clicking "Create"

I cannot add pins but this change might break existing games

### DIFF
--- a/lib/engine/game/g_18_uruguay/map.rb
+++ b/lib/engine/game/g_18_uruguay/map.rb
@@ -53,7 +53,6 @@ module Engine
           '172' => 1,
           '887' => 1,
           '888' => 1,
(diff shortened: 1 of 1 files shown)


ollybh (approved):

ollybh:
@patrikolesen Four 18Uruguay games are broken after deploying this pull request: game IDs 176676, 177067, 177447 and 178966. They fail with errors `Blocking step Deliver goods cannot process action` or `Blocking step Destination Bonus cannot process action`. Pinning them to the previously deployed version doesn't fix them.

Do you think these might be fixable, or shall I just archive these games?

PR #11321, opened 2024-11-05
[1871] Prevents sale of more than 30% during EMR

### LINKED ISSUE #8260: [The Old Prince 1871] In a forced train purchase, president should not be allowed to sell more than 30% of company
Please see game action where president of MS was allowed to sell 40% of Gt. 

https://18xx.games/game/95023?action=458

Per section 7.6.4 in the prototype rules "May not sell more than 30% of a single company."

philcampeau:
Fixes #8260 

## Before clicking "Create"

This will likely require pins. 

### Explanation of Change

Players should not be able to sell more than 30% of a corp during funding, as per rule:
image

Corporations may issue above 30%, so I put in an added guard clause to ensure only players are blocked.

### DIFF
--- a/lib/engine/game/g_1871/step/buy_train.rb
+++ b/lib/engine/game/g_1871/step/buy_train.rb
@@ -59,9 +59,8 @@ module Engine
             return false unless sellable_bundle?(bundle)
             return false if @game.class::MUST_SELL_IN_BLOCKS && @corporations_sold.include?(bundle.corporation)
 
-            # This is our new clause for 1871, if this is the corporation
-            # selling, we can sell all of them
-            return true if bundle.corporation == bundle.owner
+            # Corporations can sell all of their treasury shares during EMR, players may not sell more than 30% of a corporation
+            return false if bundle.corporation != bundle.owner && (bundle.percent > @game.class::TURN_SELL_LIMIT)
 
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1871/step/buy_train.rb
```diff
@@ -59,9 +59,8 @@ def can_sell?(entity, bundle)
             return false unless sellable_bundle?(bundle)
             return false if @game.class::MUST_SELL_IN_BLOCKS && @corporations_sold.include?(bundle.corporation)
 
-            # This is our new clause for 1871, if this is the corporation
-            # selling, we can sell all of them
-            return true if bundle.corporation == bundle.owner
```
ollybh, line 64: Does removing this test change anything? Are there any bundles of treasury shares that would be rejected by the `selling_miniumum_shares?` test which should be issuable?

philcampeau: good call. Update incoming. 

I'm going to simplify my clause and restore this one above it, only returning false if the bundle.percent > @game.class::TURN_SELL_LIMIT, since the corp sale will already have returned `true` before that check.

ollybh (approved):

PR #11322, opened 2024-11-05
[site] Discard Trains step to display page description instead of default 'Discard Trains'

### LINKED ISSUE #9128: [18NY] Change Text for Forced Salvage of Trains Over Limit on Phase Change
In 18NY, when a phase change triggers a drop in train limit the UI prompts the player to choose a train to discard. However, the player is actually choosing which train to salvage, and gets the salvage value for the train. I would suggest changing the text to prompt to choose the train to 'salvage', not to 'discard'.

philcampeau:
Fixes #9128 

## Before clicking "Create"

### Explanation of Change

This change, initially done for 18NY, will now allow all games to define what their Discard Trains step should display when prompting the discard of excess trains. 

I also updated the description in lib/engine/step/discard_train.rb from 'Discard Train' to 'Discard Trains', since that's what was being displayed anyway. 

### Any Assumptions / Hacks

I know a previous fix (#9136) broke some things on the site. From the tests I've run on my end, these changes seem to work, but I'm happy to hear it if I'm wrong!

### DIFF
--- a/lib/engine/game/g_18_ny/step/discard_train.rb
+++ b/lib/engine/game/g_18_ny/step/discard_train.rb
@@ -10,6 +10,10 @@ module Engine
           def process_discard_train(action)
             @game.salvage_train(action.train)
           end
+
+          def description
+            'Salvage Excess Trains'
+          end
         end
       end
     end
--- a/assets/app/view/game/discard_trains.rb
+++ b/assets/app/view/game/discard_trains.rb
@@ -39,8 +39,9 @@ module View
             &.include?('scrap_train')
         overflow << h(Map, game: @game) if @game.round.is_a?(Engine::Round::Operating)
 
+        discard_description = step.description
         h(:div, [
-          h(:h3, 'Discard Trains'),
+          h(:h3, discard_description),
           *overflow,
         ])
       end
--- a/lib/engine/step/discard_train.rb
+++ b/lib/engine/step/discard_train.rb
@@ -23,7 +23,7 @@ module Engine
       end
 
       def description
-        'Discard Train'
+        'Discard Trains'
       end
 
       def process_discard_train(action)


lib/engine/game/g_18_ny/step/discard_train.rb
```diff
@@ -10,6 +10,10 @@ class DiscardTrain < Engine::Step::DiscardTrain
           def process_discard_train(action)
             @game.salvage_train(action.train)
           end
+
+          def description
+            'Salvage Excess Trains'
```
crericha, line 15: The rulebook explicitly uses `discard` here.

```
Whenever trains are salvaged or discarded, the owning company receives the salvage value
```

```
If a phase change includes a reduction in the Train Limit [3], all companies, including the one
operating, must immediately discard any trains in excess of the Train Limit to the Bank Pool 
for salvage value.
```

philcampeau: It does, but is that the optimum phrasing?

philcampeau: Since the site doesn't add the text "for salvage value", I feel this is an improvement.

ollybh: Salvage, as a verb, means to save something from destruction. So salvaging a train would be rescuing one from the scrap yard and putting it back into use, the opposite of what is happening here.

philcampeau:
Throughout the rules of 18NY, and everywhere else in this game, salvaging is used to mean removing a train from a corp and receiving compensation.

lib/engine/game/g_18_ny/step/discard_train.rb
crericha: How about `discard for salvage value`?

crericha:
There hasn't been any responses/new commits since my last comment over a month ago, so I'm going to close for now. Feel free to reopen if the PR is actively going to be worked on.

PR #11323, opened 2024-11-07
[18USA] Fix P16 description

### LINKED ISSUE #11025: [18USA] Add tile information to P16 - Regional Headquarters
> Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. May be placed on any existing normal city. Three of the track segments are optional and can be placed pointing toward water or other off map areas.

There are no cities on the map that require three optional track segments. Also, it would help if this mentioned that there is a Regional Headquarters tile, thus:

﻿﻿﻿Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. _Tile #X23 RHQ_ may be placed on any existing normal city. _Two_ of the track segments are optional and can be placed pointing toward water or other off map areas.

Screenshot 2024-07-09 at 00 32 40

Screenshot 2024-07-09 at 00 32 17

philcampeau:
Fixes #11025

## Before clicking "Create"

### Explanation of Change

Description for P16 wasn't clear about tile being used - fixed
Clarified description as well.

### DIFF
--- a/lib/engine/game/g_18_usa/entities.rb
+++ b/lib/engine/game/g_18_usa/entities.rb
@@ -295,9 +295,9 @@ module Engine
             name: 'P16 - Regional Headquarters',
             value: 60,
             revenue: 0,
-            desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. ' \
-                  'May be placed on any existing normal city. Three of the track segments are optional and can be placed ' \
-                  'pointing toward water or other off map areas.',
+            desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. '\
+                  'The X23 tile (RHQ label) may be placed on any existing normal city. Two of the track segments are optional '\
+                  'and can be placed pointing toward water or other off map areas.',
             sym: 'P16',
             abilities: [
               {
--- a/lib/engine/game/g_18_usa/entities.rb
+++ b/lib/engine/game/g_18_usa/entities.rb
@@ -296,7 +296,7 @@ module Engine
             value: 60,
             revenue: 0,
             desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. '\
-                  'The X23 tile (RHQ label) may be placed on any existing normal city. Two of the track segments are optional '\
+                  'The X23 tile may be placed on any existing normal city. Two of the track segments are optional '\
                   'and can be placed pointing toward water or other off map areas.',
             sym: 'P16',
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_usa/entities.rb
```diff
@@ -295,9 +295,9 @@ module Entities
             name: 'P16 - Regional Headquarters',
             value: 60,
             revenue: 0,
-            desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. ' \
-                  'May be placed on any existing normal city. Three of the track segments are optional and can be placed ' \
-                  'pointing toward water or other off map areas.',
+            desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. '\
+                  'The X23 tile (RHQ label) may be placed on any existing normal city. Two of the track segments are optional '\
```
crericha, line 299: The tile has 3 optional connections that are contiguous.

Recommend removing `(RHQ label)`. The tile number is enough and the description is already long.

philcampeau: It's true that the private text says 3 of the segments are optional in the physical version, but there are no spaces on the board that would require more than 2 to be ignored. Should we still have it say 3?

I'll remove the (RHQ label) text.

lib/engine/game/g_18_usa/entities.rb
```diff
@@ -295,9 +295,9 @@ module Entities
             name: 'P16 - Regional Headquarters',
             value: 60,
             revenue: 0,
-            desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. ' \
-                  'May be placed on any existing normal city. Three of the track segments are optional and can be placed ' \
-                  'pointing toward water or other off map areas.',
+            desc: 'Regional Headquarters may be used to upgrade a green or brown non-metropolis city after phase 5 begins. '\
+                  'The X23 tile may be placed on any existing normal city. Two of the track segments are optional '\
+                  'and can be placed pointing toward water or other off map areas.',
```
crericha, line 300: Why did you change the optional track segments from three to two? Here are the rules (All-aboard edition):

> Comes with the gray Regional Headquarters tile. #e may be used to upgrade a green or
brown non-metropolis city a$er Phase 5 begins.
Three of the track segments are optional and can be placed pointing toward water or other o!
map areas.

crericha:
This doesn't look correct and I haven't gotten a response, so I'm going to close for now. Feel to reopen with the fixes when you have time.

PR #11340, opened 2024-11-11
[1860] fix player value in game variant

### LINKED ISSUE #11332: 1860 - player values and liquidity values glitch
**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

philcampeau:
Fixes #11332

## Before clicking "Create"

### Explanation of Change

With the non_operated_full_value variant, when I modified the `def player_value` code, I forgot to add in the value of shares of corps with trains. Admittedly a big oversight!

I suspect we'll need to pin any finished games on that variant, since their recorded final scores will be incorrect.


roseundy:
@philcampeau - did the original change affect liquidity? If so, I'm not sure where that happened - but I'm at work and can't poke around the code very well.

PR #11374, opened 2024-11-22
[1856] eliminates the use of 2', 3', etc trains.

philcampeau:
## Before clicking "Create"

### Explanation of Change

Currently, 1856 (and its cousin 1836Jr56) use a prime symbol to identify the last train of each rank, effectively creating several new ranks of trains. The reason is that 1856 has a floating mechanism based on the currently available train, rather than the current phase. 

There's a better way of doing this, which I've implemented here, where the event is triggered by specific trains in the rank. 

The only downside to this update is it will require pinning all current games of 1856. I tried for quite a while to get a script working that would replace all the relevant trains in every active game, but I couldn't get it to work. If someone wants to give me a hand with that, I could add that to this pull request and then eliminate the need for pinning.


ollybh (commented):
This looks reasonable. A couple of things to think about:

- Would it be better to add events like `3_train_available` on the last 2-train and changing the float percentage from those events, instead of overriding the `buy_train_action` method?
- Would it be clearer to remove the `facing_X` statuses and put this information in the game timeline?

I would say that this definitely needs a migration script before it is deployed, this is going to affect 300+ games. It should be fairly simple to do, it just needs to change some train names in the actions. I can help with this.

philcampeau:
There are still more changes to do, after some testing. Will close for now.

PR #11375, opened 2024-11-22
[1832] update to P4 code

philcampeau:
## Before clicking "Create"

### Explanation of Change

Configured which shares can be received in exchange for P4 - London Investment

### DIFF
--- a/lib/engine/game/g_1832/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1832/step/buy_sell_par_shares.rb
@@ -10,6 +10,25 @@ module Engine
           def visible_corporations
             @game.sorted_corporations.reject { |item| item.type == :system unless item.floated? }
           end
+
+          def can_exchange_any?(entity)
+            return false unless entity.player?
+
+            entity.companies.each do |company|
+              next unless company.id == 'P4'
+
+              # Only true if there are shares available from an ipoed corp that has not yet operated
+              corps = @game.corporations.select(&:ipoed).reject do |c|
+                c.operating_history.size.positive?
+              end
+
+              corps.each do |c|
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1832/step/buy_sell_par_shares.rb
```diff
@@ -10,6 +10,25 @@ class BuySellParShares < G1870::Step::BuySellParShares
...
+                c.operating_history.size.positive?
+              end
+
+              corps.each do |c|
+                return true unless c.num_ipo_shares.empty
+              end
+            end
+
+            false
+          end
```
ollybh, line 31: `num_shares_of` returns a number, so testing `empty?` (which needs a question mark) isn't appropriate. But you should use `empty?` test on the `operating_history` hash, instead of testing `size.positive?`.

You could also combine the two iterators over corporations to make the code clearer. And I think it would be best if the corporation iterator was moved out of the companies iterator, as they are not related.
```suggestion
            return false if entity.companies.none? { |company| company.id == 'P4' }
            
            # Can only exchange if there are shares available from an ipoed corp that has not yet operated
            @game.corporations.any? do |corp|
              corp.ipoed && !corp.operating_history.empty? && corp.num_ipo_shares.positive?
            end
          end
```

ollybh (changes requested):
The code will cause errors as `c.num_ipo_shares.empty` is incorrect. The logic could be simplified too.

ollybh (approved):

ollybh:
@philcampeau Rubocop is complaining about trailing whitespace. Can you find and remove the guilty character?

philcampeau:
Ugh, every time I try to use the function to automatically use the suggested changes, this type of error comes up.

PR #11376, opened 2024-11-22
[1849] fix for the Bonds variant

### LINKED ISSUE #11138: [1849 variants] server caught in a refreshing loop
Game https://18xx.games/game/173565

Error message 
>The server did not accept this action due to: Illegal action: Blocking step Lay/Upgrade Track cannot process action 246... refreshing

action 246 is
  {
      "type": "lay_tile",
      "entity": "CTL",
      "entity_type": "corporation",
      "id": 246,
      "created_at": 1723662227,
      "hex": "E3",
      "tile": "8-1",
      "rotation": 4
    },

For what it's worth Ive seen this twice now, the first time resolved itself by repeatedly inputting my actions.

Edit: Running with the all the new variants (except for delay IFT)

philcampeau:
Fixes #11138

(hopefully)

## Before clicking "Create"

### Explanation of Change

We keep getting blocking step errors when using the Bonds variant, and it's always happening after the game enters phase 8, which is when the bonds are enabled. 

It's difficult to test this fix, as the error doesn't happen in my local test environment, but I think the fix here may help.  I changed the round_state definition, and modified the related code.

### DIFF
--- a/lib/engine/game/g_1849/step/bond.rb
+++ b/lib/engine/game/g_1849/step/bond.rb
@@ -20,8 +20,8 @@ module Engine
 
           def round_state
             {
-              issued_bond: nil,
-              redeemed_bond: nil,
+              issued_bond: false,
+              redeemed_bond: false,
             }
           end
 
--- a/lib/engine/game/g_1849/step/bond.rb
+++ b/lib/engine/game/g_1849/step/bond.rb
@@ -7,8 +7,6 @@ module Engine
     module G1849
       module Step
         class Bond < Engine::Step::Base
-          attr_reader :issued_bond, :redeemed_bond
-
           def actions(entity)
             return [] if !entity.corporation? || entity != current_entity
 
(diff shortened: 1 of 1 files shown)


ollybh (commented):
@philcampeau Can you explain what you're changing here? There doesn't look to be anything wrong with this change (apart from my minor quibble about using boolean values), but how will this fix the bug?

philcampeau:
> @philcampeau Can you explain what you're changing here? There doesn't look to be anything wrong with this change (apart from my minor quibble about using boolean values), but how will this fix the bug?

To be honest, I don't know if it will fix the issue or not, but I can't figure out what the issue is, and I noticed I did this differently than every other instance on the site, so I figured I might as well try this to see if it helps.

ollybh (approved):
I can't see how this is going to fix the reported issue, but I can't see any problems with this patch.

PR #11386, opened 2024-12-01
[18RoyalGorge] Require graph connection for cross-buying trains

### LINKED ISSUE #11364: [18RoyalGorge] Prohibit train purchase when companies aren't connected
https://18xx.games/game/186716?action=256 

At this point, 18xx.games is allowing SF to buy a train from SPP.  This is not allowed under the rules because a company can only buy a train from another company if the two companies are connected.

michaeljb:
From page 13 of 24 in the [current rulebook PDF][0].

Check for the connection when processing the action--instead of when finding available trains to buy--to minimize the necessary graph checks.

Fixes #11364

[0]: https://boardgamegeek.com/filepage/284918/graphic-rulebook-preliminary

-----

Pins needed: any games where an illegal (unconnected) cross-buy occurred.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_royal_gorge/step/buy_train.rb
+++ b/lib/engine/game/g_18_royal_gorge/step/buy_train.rb
@@ -16,6 +16,27 @@ module Engine
             # even when train tight, there's room for a self-rust
             return true if entity.trains.any? { |t| t.rusts_on == upcoming_train.name }
           end
+
+          def check_connected!(buyer, train)
+            return if @game.loading
+            return if train.owner == @game.depot
+
+            seller = train.corporation
+            seller_nodes = @game.graph.connected_nodes(seller).keys
+            buyer_nodes = @game.graph.connected_nodes(buyer).keys
+
+            common_nodes = seller_nodes & buyer_nodes
+            return if common_nodes.any? do |n|
+                        n.town? || n.available_slots.positive? || n.tokened_by?(seller) || n.tokened_by?(buyer)
+                      end
+
+            raise GameError, "#{buyer.name} must be connected to #{seller.name} to buy their train."
+          end
+
+          def process_buy_train(action)
+            check_connected!(action.entity, action.train)
+            super
+          end
         end
       end
     end


lib/engine/game/g_18_royal_gorge/step/buy_train.rb
```diff
@@ -16,6 +16,21 @@ def room?(entity)
             # even when train tight, there's room for a self-rust
             return true if entity.trains.any? { |t| t.rusts_on == upcoming_train.name }
           end
+
+          def check_connected!(buyer, train)
+            return if train.owner == @game.depot
```
ollybh, line 21: Adding a `return if @game.loading` guard cause here would remove the need for the game graph to be recalculated when a game is being loaded, at the cost of not rechecking the legality of the move.

crericha: I agree that the check should be omitted if loading.

lib/engine/game/g_18_royal_gorge/step/buy_train.rb
```diff
@@ -16,6 +16,21 @@ def room?(entity)
...
+
+            seller = train.corporation
+            seller_nodes = @game.graph.connected_nodes(seller)
+            return if @game.graph.connected_nodes(buyer).any? { |node, _| seller_nodes.include?(node) }
```
crericha, line 25: Would it be more efficient to do `return unless (@game.graph.connected_nodes(train.corporation) & @game.graph.connected_nodes(buyer)).empty?`

ollybh: `graph.connected_nodes` is a hash, and so doesn't have a `&` method. You could do it as:
```
@game.graph.connected_nodes(train.corporation).keys.intersect?(@game.graph.connected_nodes(buyer).keys)
```
But I'm not sure if that's clearer.

michaeljb: Since the return of `connected_nodes` is a hash and we're looking at the keys, `seller_nodes.include?(node)` inside the loop is O(1) while the outer loop is O(buyer's nodes), so linear time overall.

When I first read these comments I was thinking the same as @crericha, that what I originally wrote was actually O(buyer's * seller's), which it would be if they were arrays instead of hashes. I didn't realize that with hashes it was fine until I just pushed my latest version, which is what @ollybh suggested, only split up across multiple lines.

I think this version should do the job in a more readable fashion.

michaeljb:
> @michaeljb This pull request has been open for three months now. Have you got any updates for this?

Hoping to contribute more regularly again 🙂

ollybh (approved):
The code looks fine, but I'm now wondering about the rules. These say that a train can be bought from another company "if two companies are connected by a route". As far as I can see, there isn't a definition as to what counts as being connected by a route. Normally I would expect this to mean that there is a route that includes tokens of both companies.

This is going to allow the purchase if there is any node on the map that both companies can trace a route to, this could be a tokened out city that both companies can trace a route to, but from different directions and not sharing any track on their routes.

michaeljb:
> The code looks fine, but I'm now wondering about the rules. These say that a train can be bought from another company "if two companies are connected by a route". As far as I can see, there isn't a definition as to what counts as being connected by a route. Normally I would expect this to mean that there is a route that includes tokens of both companies.
> 
> This is going to allow the purchase if there is any node on the map that both companies can trace a route to, this could be a tokened out city that both companies can trace a route to, but from different directions and not sharing any track on their routes.

Oh good call, a tokened out city should block the train purchase.

michaeljb:
OK, now when they are mutually tokened out of a city, the cross buy is blocked. Thanks @ollybh!

lib/engine/game/g_18_royal_gorge/step/buy_train.rb
```diff
@@ -16,6 +16,25 @@ def room?(entity)
...
+          def check_connected!(buyer, train)
+            return if @game.loading
+            return if train.owner == @game.depot
+
+            seller = train.corporation
+            seller_nodes = @game.graph.connected_nodes(seller).keys
+            buyer_nodes = @game.graph.connected_nodes(buyer).keys
+
+            common_nodes = seller_nodes & buyer_nodes
+            return if common_nodes.any? { |n| n.available_slots.positive? || n.tokened_by?(seller) || n.tokened_by?(buyer) }
```
ollybh: `available_slots` is only a method for city nodes. Maybe
```suggestion
            return if common_nodes.any? { |n| n.town? || n.available_slots.positive? || n.tokened_by?(seller) || n.tokened_by?(buyer) }
```

michaeljb: Done.

PR #11393, opened 2024-12-02
[18ESP] fix a few minor bugs

### LINKED ISSUE #11337: 18ESP allows two tokens in the same city.
Hello,

Game https://18xx.games/game/184913

N (Norte) company takes over MH minor company in E21. Firstly, N company laid a token there. Then, it took over MH. When a major company takes over a minor company, the major company may exchange the minor company token by a major token, BUT it cannot do this if there is already a token in the same hex, because no company can own two station tokens or more in any hex.

In this scenario, the program shouldn't allow to exchange the token, but leave it in Norte company charter.

Thanks,

Enrique Trigueros, designer of 18ESP.

bentziaxl:
transfer all but one tender to the bank when buying p5.
fix broken game when buying p5
fix having two tokens in the same city when merging major with minor

Fixes #11337

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -880,10 +880,10 @@ module Engine
             token = corporation.tokens.first
             hex = hex_by_id(corporation.coordinates)
             city = if corporation_by_id('MZ')&.ipoed
-                     # mza special case if mz already exists on the map
+                     # MZA special case if MZ already exists on the map
                      hex.tile.cities.size > 1 ? city_by_id("#{hex.tile.id}-#{corporation.city}") : hex.tile.cities.first
                    else
-                     # mza exists, but no mz. Place on original location
+                     # MZA exists, but no MZ. Place on original location
                      city_by_id("#{hex.tile.id}-#{corporation.city}")
                    end
             @log << "#{corporation.name} places a token on #{hex.id}"
@@ -1232,7 +1232,7 @@ module Engine
           if !yellow_green
             delete_slot = city.slots > 4 ? city.slots : false
             city.delete_token!(token, remove_slot: delete_slot)
-            # add mza reservation if mza not tokened in madrid yet
+            # add MZA reservation if MZA not tokened in madrid yet
             mza_token = city.tokens.compact.find { |t| t.corporation == mza }
             city.add_reservation!(mza) unless mza_token
             token.destroy!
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_esp/game.rb
```diff
@@ -876,11 +876,16 @@ def num_corp_trains(entity)
...
 
         def place_home_token(corporation)
-          if corporation == mza && corporation_by_id('MZ')&.ipoed && !corporation.tokens.first.used
-            # mza special case if mz already exists on the map
+          if corporation == mza && !corporation.tokens.first.used
             token = corporation.tokens.first
             hex = hex_by_id(corporation.coordinates)
-            city = hex.tile.cities.size > 1 ? city_by_id("#{hex.tile.id}-#{corporation.city}") : hex.tile.cities.first
+            city = if corporation_by_id('MZ')&.ipoed
+                     # mza special case if mz already exists on the map
```
perwestling, line 883: In general I think the corporation abbreviations in texts and comments should use upper case, so MZA and MZ in this case. Same comment for line 886.

lib/engine/game/g_18_esp/step/acquire.rb
```diff
@@ -90,10 +90,14 @@ def choices
           end
 
           def can_swap?
```
perwestling, line 92: An alternative implementation could be to move line 95-97 plus the unless into a method that returns true/false.
And then
return true if is_some_method

followed by merged_token_in_shared_city? content.
Not sure which is most readable?

lib/engine/game/g_18_esp/game.rb
```diff
@@ -917,6 +923,16 @@ def transfer_luxury_ability(company, entity)
           company.close!
         end
 
+        def transfer_tenders_to_bank
+          @luxury_carriages.dup.each do |owner, amount|
+            next if owner == 'bank'
+
+            transfer_amount = amount - 1
+            @luxury_carriages['bank'] += transfer_amount
+            @luxury_carriages[owner] = 1
```
perwestling, line 932: Is this correct? I do not know the code so I might be wrong. Feels strange to set this to one here. Just checking.

lib/engine/game/g_18_esp/step/acquire.rb
bentziaxl: i think this is readable as it is, i don't think refactoring will make it more readable

lib/engine/game/g_18_esp/game.rb
bentziaxl: Yes, there is a private that comes with 5 tenders, but tenders can end up in the "bank"
other corps can buy tenders either from player owner of the private or the bank.
If private is bought in, all the tenders are transferred the the bank, except one which is owned by the corp buying the private in.

PR #11408, opened 2024-12-17
[1894] Fix a bug allowing for upgradin a green city that you don't have access to

Galatolol:
## Before clicking "Create"

### Explanation of Change

Currently a corporation can upgrade a green city tile it is next to but doesn't have access to the city (like Ouest on the first screenshot). It shouldn't be possible, it first has to connect to the city.

**⚠️ This requires pinning ⚠️**


ollybh (changes requested):
This isn't the correct way to do this. The player shouldn't be offered the option to place those tiles if the corporation does not have access to any paths on the placed tile.

1894’s implementation of `Step::Tracker::legal_tile_rotation?` is, for brown city tiles, only checking whether the replacement tile has track connected to the same exits. It should also be checking that one of the paths on the tile is connected to an exit on the corporation’s graph.

This is the test in `Engine::Step::Tracker::legal_tile_rotation?` that you need to keep:
https://github.com/tobymao/18xx/blob/4d661ea1075be7c6ec2c679933b2519008faac32/lib/engine/step/tracker.rb#L424-L425

Galatolol:
@ollybh Thank you for the suggestion, I agree that it's better to not hightlight the hex at all.

lib/engine/game/g_1894/step/track.rb
```diff
@@ -15,7 +15,12 @@ def actions(entity)
             super
           end
 
-          def legal_tile_rotation?(_entity, hex, tile)
+          def legal_tile_rotation?(entity, hex, tile)
+            if @game.class::GREEN_CITY_TILES.include?(hex.tile.name)
+              entity_reaches_a_new_exit = !(tile.exits & hex_neighbors(entity, hex)).empty?
+              return false unless entity_reaches_a_new_exit
+            end
```
ollybh, line 22: The logic here looks correct. The only thing I would change is to use a slightly more efficient way for testing for array intersection: `arr1.intersect?(arr2)` instead of `(arr1 & arr2).empty?`. You could also stick it as a single liner, but Rubocop might object to the line length.
```suggestion
            return false if @game.class::GREEN_CITY_TILES.include?(hex.tile.name) && !tile.exits.intersect?(hex_neighbors(entity, hex))
```

ollybh (approved):

PR #11411, opened 2024-12-21
[1858 Switzerland] Avoid modifying 1858's train definitions

### LINKED ISSUE #11372: [1858] Random refresh with error message about blue privates
https://18xx.games/game/183151

OR4.1 MZA bought the first 6H, and discarded a 2H. The A was also over train limit, but when I tried to discard a 2H, an error message flashed up briefly (something about blue privates?) and the program refreshed.
After a couple more tries, it accepted the discard.
Looks like something from Switzerland has interfered with the base game.

ollybh:
## Implementation Notes

Issue #11372 was caused by modifying 1858's train definitions without a making a deep copy. If a game of 1858 is loaded on the server after a game of 1858 Switzerland then it is was seeing the version of the definitions after these modifications, causing errors as 1858 does not have a method for handling the `blue_privates_available` event.

Making a deep copy of the train definitions prevents this error.

This commit also makes a deep copy of the phase definitions. No problems have been observed with this, but making the copy is safer.

Fixes #11372.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1858_switzerland/game.rb
+++ b/lib/engine/game/g_1858_switzerland/game.rb
@@ -65,12 +65,14 @@ module Engine
         ROBOT_MAJOR_TILE_LAYS = [{ lay: true, upgrade: true }].freeze
 
         def game_phases
-          phases = super
-          _phase2, _phase3, phase4, phase5, phase6 = phases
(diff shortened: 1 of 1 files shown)


ianwilson156:
Hurrah! You found the cause of the problem - well done!

PR #11415, opened 2024-12-22
[18FR] Initial commit: map, tiles, info

Galatolol:
## Before clicking "Create"

### Explanation of Change

Initial commit for the game 18FR based on 1817. Map, tiles, phase and train table.

### Screenshots

image

### Hacks and assumptions

Private auction is skipped for now.


lib/engine/game/g_18_fr/game.rb
```diff
@@ -0,0 +1,225 @@
...
+                    events: [{ 'type' => 'free_ports' }],
+                  },
+                  { name: '2P', distance: 2, price: 300, num: 5 },
+                  { name: '5', distance: 5, price: 600, num: 6 },
+                  { name: '6*D', distance: 6, price: 800, num: 30 },
+                  { name: '2P*', distance: 2, price: 200, num: 1 }].freeze
+
+        EVENTS_TEXT = Base::EVENTS_TEXT.merge(
+          'two_tile_lays' => ['Corporations may lay two tiles for 20 F. One may be upgrade'],
+          'free_ports' => ['Ports no longer count towards train length']
```
ollybh, line 144: As these are ongoing effects, these might be better implemented as phases statuses. If they are events they will disappear from the game information page once they have happened, but statuses are visible for the whole game.

lib/engine/game/g_18_fr/game.rb
```diff
@@ -0,0 +1,225 @@
...
+
+          G1817::Round::Stock.new(self, [
+            Engine::Step::DiscardTrain,
+            Engine::Step::HomeToken,
+            G18FR::Step::BuySellParShares,
+          ])
+        end
+
+        def operating_round(round_num)
+          @interest_fixed = nil
```
ollybh, line 180: This line is pointless, this variable is set on the next line.
```suggestion
```

lib/engine/game/g_18_fr/game.rb
```diff
@@ -0,0 +1,225 @@
...
+          @interest_fixed = interest_rate
+          # Revaluate if private companies are owned by corps with trains
+          @companies.each do |company|
+            next unless company.owner
+
+            abilities(company, :revenue_change, time: 'has_train') do |ability|
+              company.revenue = company.owner.trains.any? ? ability.revenue : 0
```
ollybh, line 187: Don't use `any?` (or `none?`) when you are not passing a code block. Use `empty?` instead, it is more efficient.
```suggestion
              company.revenue = company.owner.trains.empty? ? 0 : ability.revenue
```

lib/engine/game/g_18_fr/game.rb
```diff
@@ -0,0 +1,225 @@
...
+        end
+
+        def operating_round(round_num)
+          @interest_fixed = nil
+          @interest_fixed = interest_rate
+          # Revaluate if private companies are owned by corps with trains
+          @companies.each do |company|
+            next unless company.owner
+
+            abilities(company, :revenue_change, time: 'has_train') do |ability|
```
ollybh, line 186: `Game.abilities` can return nil, a single ability object, or an array of ability objects, depending on how many matching abilities are found. This code is going to fail unless the company has multiple revenue change abilities. And if a company has multiple revenue change abilities, should you be summing the revenues?

lib/engine/game/g_18_fr/game.rb
ollybh: Can companies be owned by players as well as corporations?

lib/engine/game/g_18_fr/game.rb
```diff
@@ -0,0 +1,225 @@
...
+          'free_ports' => ['Ports no longer count towards train length']
+        ).freeze
+
+        ONE_YELLOW_TILE_LAY = [{ lay: true, upgrade: false }].freeze
+        TWO_TILE_LAYS = [
+          { lay: true, upgrade: true },
+          { lay: true, upgrade: :not_if_upgraded, cost: 20, cannot_reuse_same_hex: true },
+        ].freeze
+
+        B_HEX_NAMES = %w[F3].freeze
```
ollybh, line 153: Is this just a placeholder for something different in the future? If this is just going to be a single hex then having this as an array is more complicated than you need.

ollybh (changes requested):

lib/engine/game/g_18_fr/game.rb
Galatolol: I forgot to update the array, thanks.

lib/engine/game/g_18_fr/game.rb
Galatolol: I removed that piece of code - it was directly copied from 1817 and it's not needed here.

lib/engine/game/g_18_fr/game.rb
```diff
@@ -0,0 +1,213 @@
...
+            @log << '-- From now on, corporations may lay two for 20 F. Only one may be upgrade --'
+            @extra_tile_lay = true
+          when 'Blue'
+            @log << '-- From now on, corporations may lay two for 20 F. Only one may be upgrade --'
+            @free_ports = true
+          end
+        end
+
+        def tile_lays(_entity)
+          @extra_tile_lay ? TWO_TILE_LAYS : ONE_YELLOW_TILE_LAY
```
ollybh, line 200: If youŕe using the phase statuses you do not need to have a variable tracking the status of the phase changes. Instead you can query the current phase's status like this:
```suggestion
          @phase.status.include?(´two_tile_lays) ? TWO_TILE_LAYS : ONE_YELLOW_TILE_LAY
```
If you do this you will need neither the `@extra_tile_lay`/`@free_ports` variables, nor the `setup`/`after_phase_change` methods.

You can also create a methods (`two_tile_lays?`/`free_ports?`) that wrap these checks to the phase status array. That might make the code easier to read.

Galatolol: Ok. If it gets used in more than one place, I'll wrap it.

ollybh (approved):

PR #11417, opened 2024-12-23
[1849] modification to code for variants

philcampeau:
Hopefully fixes issues with Variants rules. I'll manually close the bug reports after confirming. (see below)

## Before clicking "Create"

### Explanation of Change

Players trying the variants have all encountered variations of this error at various points: 

image

Often, refreshing and trying again and again seems to resolve the issue temporarily, but that's obviously not a solution. This issue has proven difficult to troubleshoot, since it doesn't happen in Hotseat mode, or even when playing multiplayer on my local environment.

I tried changing the code for def operating_round, removing the logic that gates the inclusion of the variant steps, and instead adding `log_pass` and `log_skip` methods to those steps to mask those steps unless using the variants. 

Once I did this, rake would fail on the fixture files, which I found interesting since the fixture doesn't use the variants. I went through the logic of the variants themselves, and added a gate to the `def actions` to return [] if that variant isn't active. 

That resolved the Rake error, and I still seem to be able to play through local  games, so at worst this is no worse than the current state, and I think the code is cleaner and closer to the standard code seen throughout the site.

### Any Assumptions / Hacks

This may well require archiving/pinning any games using the Bonds/Electric Dreams/Buy Tokens variants, but I don't believe it would affect other games of 1849.


lib/engine/game/g_1849/game.rb
```diff
@@ -517,13 +517,13 @@ def operating_round(round_num)
...
+            G1849::Step::BuyEToken,
             Engine::Step::DiscardTrain,
             G1849::Step::BuyTrain,
             G1849::Step::IssueShares,
-            bonds? ? G1849::Step::BondInterestPayment : nil,
-            bonds? ? G1849::Step::Bond : nil,
+            G1849::Step::BondInterestPayment,
+            G1849::Step::Bond,
             [Engine::Step::BuyCompany, { blocks: true }],
         ].compact, round_num: round_num)
```
ollybh, line 528: You don´t need to `compact` the array any more, as it will not contain any nils.
```suggestion
        ], round_num: round_num)
```

lib/engine/game/g_1849/step/bond.rb
```diff
@@ -30,7 +30,12 @@ def description
           end
 
           def log_skip(entity)
-            super if @game.bonds? && @game.issue_bonds_enabled
+            print "hi #{can_take_loan?(entity)}"
```
ollybh, line 33: A stray debugging statement has made it through to the pull request.
```suggestion
```

ollybh (approved):
This change looks fine to me. I don't think that this should break any existing games as the steps will be skipped if the variants aren't selected. But leave the pins tag in place to make sure that this gets tested before the code is deployed.

ollybh (changes requested):
A couple of small fixes are needed before this gets merged.

lib/engine/game/g_1849/step/bond.rb
philcampeau: removed

ollybh (approved):

PR #11426, opened 2024-12-29
[18Norway] Setup privates for 1.5X

patrikolesen:
It shall be allowed to buy privets to up to 1.5X

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -154,7 +154,7 @@ module Engine
         EXTRA_TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false }].freeze
 
         def setup
-          setup_company_price_up_to_face
+          setup_company_price_up_to_one_and_half_face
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -154,6 +154,7 @@ def switcher?(corporation)
         EXTRA_TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false }].freeze
 
         def setup
+          setup_company_price_up_to_face
```
perwestling, line 157: This title might cause confusion as company price is not set up to face, or is it?

ollybh: Or, instead of this code to change the companies' min/max prices, you could just add these prices to the definitions in COMPANIES.

patrikolesen: Good catch I agree that the naming could be confusing I hope this is better
@ollybh It is also a possibility I kind of like that the function is only done once but it requires an extra function which is sad.

PR #11427, opened 2024-12-29
[18Norway] Only ships are allowed to use harbors

patrikolesen:
Only ships are allowed to use harbors

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -453,20 +453,19 @@ module Engine
           harbor_token?(city, entity)
         end
 
-        def connected?(a, b, corporation)
+        def connected?(a, b, corporation, train)
           return true if a.connects_to?(b, corporation)
 
           [a.a, a.b].each do |part|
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -453,20 +453,19 @@ def for_graph_city_tokened_by?(city, entity, graph)
           harbor_token?(city, entity)
         end
 
-        def connected?(a, b, corporation)
+        def connected?(a, b, corporation, train)
```
perwestling, line 456: "train" might not be a good name of this parameter, as it can be a ship.
How about train_or_ship? Or some other name that includes both? (moveable?)

ollybh: I think that naming is fine. The ships are being implemented as Train objects, so it will be a train that is passed as a parameter.

ollybh (approved):

bentziaxl (approved):

PR #11429, opened 2024-12-29
[18Norway] Updated nationalization sequence

patrikolesen:
New nationalization sequence so that we do not corrupt the spreadsheet

## Before clicking "Create"


perwestling (approved):

ollybh (commented):
@patrikolesen Can you explain why you are changing this? The previous implementation with a separate round for nationalisation seems to make more sense to me.

In this version the non-interactive TriggerNationalization step is unusual. Does this need to be implemented as a separate step? And I think that there might be a bug in here when you are setting `@nationalization_candidates` to `nil` in `reset_nationalization_candidates` but not checking for nils in `NationalizeCorporation.actions`.

patrikolesen:
@ollybh The problem I got with the other implementation was that the spreadsheet looked strange where I got a few extra rounds

ollybh:
> @ollybh The problem I got with the other implementation was that the spreadsheet looked strange where I got a few extra rounds

What was the problem with the spreadsheet? There are other games that have rounds that are neither stock rounds nor operating rounds (merger rounds in 1861/1867, merger and acquisition rounds in 1817), and those don't break the spreadsheet.

patrikolesen:
@ollybh I have probably done some thing wrong in my implementation then, will need to look closer
But here is an example as you can see here the OR1.1 is reported twice
image
The game also got out of sync when calculating the end game
Will keep you posted when I have investigated further

ollybh:
> But here is an example as you can see here the OR1.1 is reported twice

You need to implement a `self.short_name` method in `Engine::Game::G18Norway::Round::Nationalization`. At the moment this is inheriting the default implementation from `Engine::Round::Operating` which returns `OR`. You want to change this to `NR` (or something like that) and that will change nationalisation round's entry in that list to `OR1.1`.

> The game also got out of sync when calculating the end game Will keep you posted when I have investigated further

That sounds like a different issue.

patrikolesen:
@ollybh I think you are correct let me correct this in another way

PR #11441, opened 2025-01-02
[1807] Change terminology: public companies, not majors

ollybh:
In the 1807/18BF rules the 10-share companies are called 'public companies', so use that terminology in the user interface.

### DIFF
--- a/lib/engine/game/g_1807/entities.rb
+++ b/lib/engine/game/g_1807/entities.rb
@@ -746,7 +746,7 @@ module Engine
             float_percent: 20,
             always_market_price: true,
             tokens: [0, 20, 20, 20],
-            type: 'major',
+            type: 'public',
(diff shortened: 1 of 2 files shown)


bentziaxl:
every game uses major for 10 share in code for type and variable names. Is there a reason you're changing the code as well as the descriptions visible to users? why not just descriptions?

ollybh:
> every game uses major for 10 share in code for type and variable names. Is there a reason you're changing the code as well as the descriptions visible to users? why not just descriptions?

I believe that I've only changed the bits that are visible to users. Are you thinking of the corporation type attribute? That gets shown on the train limit section on the game info page:
image

Or is there something else that you think that should be left unchanged? I've not renamed variables called `major`/`majors`.

bentziaxl (approved):

PR #11442, opened 2025-01-02
[1807] Minor fixes and improvements

ollybh:
## Implementation Notes

One fix: the underground railway privates were not being made available due to misuse of the `Array.delete` method (it does not take a block, and does not return the deleted item).

Also, use the hex cache and find hexes using the `hex_by_id` method instead of `@hexes.find`.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1807/game.rb
+++ b/lib/engine/game/g_1807/game.rb
@@ -105,12 +105,14 @@ module Engine
         end
 
         def event_u1_available!
-          u1 = @future_companies.delete { |company| company.sym == 'U1' }
+          u1 = @future_companies.find { |company| company.sym == 'U1' }
+          @future_companies.delete(u1)
           @companies << u1
         end
 
         def event_u2_available!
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1807/game.rb
```diff
@@ -105,12 +105,14 @@ def event_minors_batch3!
         end
 
         def event_u1_available!
-          u1 = @future_companies.delete { |company| company.sym == 'U1' }
+          u1 = @future_companies.find { |company| company.sym == 'U1' }
+          @future_companies.delete(u1)
           @companies << u1
```
bentziaxl, line 110: you can use reject! to do this in one line

ollybh: > you can use reject! to do this in one line

But I need to grab hold of the U1 company object, to add it to the `@companies` array after it has been removed from `@future_companies`. `reject!` doesn't return the item(s) removed from the array, so I'd still need to use something like `find`.

bentziaxl (approved):

PR #11449, opened 2025-01-11
[18FR] Share Redemption Round

Galatolol:
## Before clicking "Create"

### Explanation of Change

In Stock Round, corporations can't redeem shares (contrarily to 1817). Instead, after SR, there is Share Redemption Round in which all companies with shares in market, going in decreasing price order, may take loans and redeem shares. After SRR share price is updated for sold out companies and shares in market.

After each SR in which a player owns any short share, they have their certificate limit increased by 1 for the rest of the game.

### Screenshots, Problems

In SRR there is an unnecessary 'Buy Market Share' button which works exactly as 'Redeem 1' (and then you can click it again if you want to redeem more). 

To not show this button, I think the `def render_market_shares` method should be modified.


lib/engine/game/g_18_fr/step/buy_sell_par_shares.rb
```diff
@@ -8,6 +8,18 @@ module G18FR
...
           MIN_BID = 90
+
+          def corporate_actions(entity)
+            # In Stock Round corporations can only take loans
+            return [] if @winning_bid
+
+            return [] if @corporate_action && @corporate_action.entity != entity
+
+            actions = []
+            actions << 'take_loan' if @round.current_actions.none? && @game.can_take_loan?(entity)
```
ollybh, line 19: Don't call `none?` without a block, use `empty?` instead.
```suggestion
            actions << 'take_loan' if @round.current_actions.empty? && @game.can_take_loan?(entity)
```

ollybh (changes requested):
One minor fix needed, you have a call to `Array.none?` without passing a code block. This is more expensive than calling `Array.empty?`.

To remove the unnecessary 'Buy market share' button in the share redemption round, try adding a `pool_shares` method to `G18FR::Step::RedeemShares`. If that method returns an empty array then I think that should get rid of the button. If you're also getting buttons for any IPO shares then you might need to look at adding a `can_buy?` method to the step.

(Also, the variable `@player_sr_with_short_count` is not the most clear name. Could it be something simpler like `@extra_certificates`?).

lib/engine/game/g_18_fr/round/share_redemption.rb
```diff
@@ -0,0 +1,71 @@
...
+            if corporations.empty?
+              @log << 'No corporation can redeem shares'
+              finish_round
+            end
+
+            corporations.sort
+          end
+
+          def next_entity!
+            raise GameError, 'Must redeem share if loan taken' if @loan_taken && !@share_redeemed
```
crericha, line 39: Where is @loan_taken and @share_redeemed reset between entities?

Galatolol: Good catch!

Galatolol:
> To remove the unnecessary 'Buy market share' button in the share redemption round, try adding a `pool_shares` method to `G18FR::Step::RedeemShares`. If that method returns an empty array then I think that should get rid of the button.

It did the trick!

lib/engine/game/g_18_fr/round/stock.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Game
+    module G18FR
+      module Round
+        class Stock < G1817::Round::Stock
+          def finish_round
+            # Do not move stock prices after Stock Round, they'll move after Share Redemption Round
+
+            @game.players.each do |player|
+              if player.shares.any? { |s| s.percent.negative? }
+                @game.player_sr_with_short_count[player] += 1
```
ollybh, line 15: You missed renaming the variable here:
```suggestion
                @game.extra_cert_limit[player] += 1
```

ollybh (changes requested):

ollybh (approved):

PR #11455, opened 2025-01-15
[18Uruguay] Interest should not be paid for penalty loans

### LINKED ISSUE #11447: [18Uruguay] Interest should not be paid for penalty loans
Interest should not be paid for loans that are enforce to pay interest.

patrikolesen:
Fixes #11447
 
Can potentially break existing games

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_uruguay/loans.rb
+++ b/lib/engine/game/g_18_uruguay/loans.rb
@@ -65,12 +65,23 @@ module Engine
           end
         end
 
-        def take_loan_if_needed_for_interest!(entity)
+        def pay_interest!(entity)
           owed = interest_owed(entity)
-          return if owed.zero?
+          interest_paid[entity] = owed
 
-          remaining = owed - entity.cash
-          perform_ebuy_loans(entity, remaining + 10) if remaining.positive?
+          while owed > entity.cash &&
+              (loan = loans[0])
+            take_loan(entity, loan, ebuy: true)
+          end
+
+          if owed <= entity.cash
+            if owed.positive?
+              log_interest_payment(entity, owed)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_uruguay/loans.rb
```diff
@@ -65,20 +65,30 @@ def adjust_stock_market_loan_penalty(entity)
...
           owed = interest_owed(entity)
-          return if owed.zero?
+          interest_paid[entity] = owed
 
-          remaining = owed - entity.cash
-          perform_ebuy_loans(entity, remaining + 10) if remaining.positive?
+          while owed > entity.cash &&
+              (loan = loans[0])
+            take_loan(entity, loan, ebuy: true)
+          end
```
bentziaxl, line 75: did you test this code? in the original implementation (interest_on_loans.rb),
owed = interest_owed(entity) is inside the loop, to update the owed value.
What's updating the owed value to prevent an infinite loop?

bentziaxl: If you do need to add that line, it looks like this code is identical to the one in interest_on_loans.rb, except 
entity.loans.size < maximum_loans(entity)

could you make the max high enough and then prevent this duplication?

patrikolesen: There are two changes here from the "original"
1) That you in this case are allowed to take more than what is normally allowed
2) It is by intension that owed is not updated. We do not want to pay interest on loans that we are forced to take due to interest payments. The entity.cash is increased for every loop so it will drop out when the entity has enough cash to pay the interest

bentziaxl (approved):

PR #11459, opened 2025-01-19
[1894] Change the G2 hex

Galatolol:
## Before clicking "Create"

### Explanation of Change

The G2 hex is now a gray 20 town instead of an empty hex.

:warning: **This requires pinning**

### Screenshots

image


Galatolol:
Hmm, can I run the tests again?

ollybh (approved):

PR #11513, opened 2025-02-04
[Core] Check for existence of `can_finance?` method

ollybh:
### Explanation of Change

The `can_finance?` method was added for 1837 in commit d08cc8435. It's used to check whether the bank will step in to pay the cost of a train.

This has broken 1877 Venezuela. There is a default implementation of `can_finance?` added to the Step::BuyTrain class, but 1877 Venezuela allows trains to be bought in the BuySellParShares step, which did not have this method.

This commit tests for the existence of this method before calling it.

Fixes tobymao#11508.

### Before clicking "Create"

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -300,7 +300,7 @@ module View
           children << h(:div, issue_str)
         end
 
-        if @step.can_finance?(@corporation)
+        if @step.respond_to?(:can_finance?) && @step.can_finance?(@corporation)
           text = "#{@game.bank.name} will provide financing for the amount the corporation cannot pay."
           children << h(:div, text)
         end


assets/app/view/game/buy_trains.rb
```diff
@@ -300,7 +300,7 @@ def render
           children << h(:div, issue_str)
         end
 
-        if @step.can_finance?(@corporation)
+        if @step.respond_to?(:can_finance?) && @step.can_finance?(@corporation)
```
crericha, line 303: I purposely avoided doing needing to do this by adding the method to the buy_train step. Looking at 1877's buy_sell_par_shares step, it copies over all the methods from the buy_train step. I think it should copy over this new method too.

ollybh: That would work too. I'd done the `respond_to?` check as we've already got a zillion tests like this in the UI code.

I've had a quick grep and can't see any other games that have a BuyTrain action in a step that's not inheriting from Step::BuyTrain.

crericha: Yeah, I'd prefer to follow the pattern of what 1877 already did, rather than in this one case use `respond_to?`. I don't see a reason to make this one a special case that behaves differently. Do you want to make that change or should I?

ollybh: I probably won't have time to do this for a couple of days, so go for it if you can do it sooner.

crericha:
Fix in #11524

PR #11531, opened 2025-02-09
[1871] Fix dividend payout when there are three PEIR shares

ollybh:
`entity.total_shares` doesn't always work correctly for PEIR.  If there are three shares then each is a 33% share, and `total_shares` returns 3.030303...

This can lead to rounding errors in dividend payout, with a revenue of 300 being calculated as 99 per share.

Fixes tobymao#11525.

### DIFF
--- a/lib/engine/game/g_1871/step/dividend.rb
+++ b/lib/engine/game/g_1871/step/dividend.rb
@@ -31,17 +31,10 @@ module Engine
           end
 
           def payout_per_share(entity, revenue)
-            share_count =
-              if entity == @game.peir
-                # entity.total_shares doesn't always work correctly for PEIR.
-                # If there are three shares then each is a 33% share, and
-                # total_shares returns 3.030303...
-                @game.peir_shares.size
-              else
-                entity.total_shares
-              end
+            return super unless entity == @game.peir
+
             # PEIR rounds up
-            (revenue.to_f / share_count).ceil
+            (revenue.to_f / @game.peir_shares.size).ceil
           end
 
           def half_pay_withhold_amount(_entity, revenue)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1871/step/dividend.rb
```diff
@@ -31,8 +31,17 @@ def share_price_change(_entity, revenue)
...
           def payout_per_share(entity, revenue)
+            share_count =
+              if entity == @game.peir
+                # entity.total_shares doesn't always work correctly for PEIR.
+                # If there are three shares then each is a 33% share, and
+                # total_shares returns 3.030303...
+                @game.peir_shares.size
+              else
+                entity.total_shares
+              end
```
crericha, line 42: Consider a ternary operator here.

lib/engine/game/g_1871/step/dividend.rb
```diff
@@ -31,8 +31,17 @@ def share_price_change(_entity, revenue)
           end
 
           def payout_per_share(entity, revenue)
+            share_count =
+              if entity == @game.peir
+                # entity.total_shares doesn't always work correctly for PEIR.
+                # If there are three shares then each is a 33% share, and
+                # total_shares returns 3.030303...
```
crericha, line 38: imho, the comment isn't necessary. If someone is curious, they can look at the commit that introduced it and know why.

crericha (approved):

lib/engine/game/g_1871/step/dividend.rb
ollybh: I had considered that, but the ternary operator made the line was too long for my tastes. But I've thought of a much cleaner implementation since…

PR #11534, opened 2025-02-10
[1844] Allow destination before home token laid

crericha:


lib/engine/game/g_1844/game.rb
```diff
@@ -771,7 +771,7 @@ def all_potential_upgrades(tile, tile_manifest: false, selected_company: nil)
         end
 
         def destinated?(entity)
-          home_node = entity.tokens.first&.city
+          home_node = hex_by_id(corporation.coordinates).cities[corporation.city || 0]
```
ollybh, line 774: This will work, but only because both JS and STB have their home locations in the first city on their tiles. If either were based in the second city then you'd get bugs if the two cities merged before the company started operating.

ollybh (approved):

lib/engine/game/g_1844/game.rb
crericha: Good point. I'll changed to use `reserved_by?`.

kepakiano: Hey, could this lead to a regression? When the current SR is ended in a clone of https://18xx.games/game/196670 it says that the method "cities" is not available

PR #11540, opened 2025-02-12
[1812] multiple pre-alpha updates

philcampeau:
## Before clicking "Create"

### Explanation of Change

- Sets game to depend on 1867 code
- add numbers to names of private companies for convenience
- set minors to 100% float
- moved PHASES and TRAINS back into base.rb
- implemented loans, and penalty for not paying interest
- implemented 2-train export 
- corrected train counts & limits
- corrected GAME_END_CHECK
- made p11 and p12 never close


lib/engine/game/g_1812/entities.rb
```diff
@@ -38,7 +38,7 @@ module Entities
             abilities: [], # G-train bonus is defined in game.rb revenue_for method
           },
           {
-            name: 'Sheffeild & Rotherham Railway',
+            name: '4. Sheffeild & Rotherham Railway',
```
ollybh, line 41:
```suggestion
            name: '4. Sheffield & Rotherham Railway',
```

philcampeau: fixed

lib/engine/game/g_1812/step/buy_train.rb
```diff
@@ -0,0 +1,87 @@
...
+
+require_relative '../../../step/base'
+require_relative '../../../step/buy_train'
+require_relative '../../../step/automatic_loan'
+
+module Engine
+  module Game
+    module G1812
+      module Step
+        class BuyTrain < Engine::Step::BuyTrain
```
ollybh, line 11: Could this class inherit from `G1867::Step::BuyTrain`? You've got several methods that are cut'n'pasted from that class.

lib/engine/game/g_1812/step/loan_operations.rb
```diff
@@ -0,0 +1,28 @@
...
+    module G1812
+      module Step
+        class LoanOperations < G1867::Step::LoanOperations
+          def skip!
+            pass!
+            entity = current_entity
+
+            owed = @game.pay_interest!(entity)
+            if owed
+              @game.unpaid_loan(entity, owed)
```
ollybh, line 17: Refactoring `G1867::Step::LoanOperations` to extract a (private) `unpaid_interest` method called at this point would mean that only that method would need to be redone in this class, instead of copying all of the code in this method.

lib/engine/game/g_1812/game.rb
```diff
@@ -1,23 +1,23 @@
...
     module G1812
-      class Game < Game::Base
+      class Game < G1867::Game
         include_meta(G1812::Meta)
-        include Entities
-        include Map
-        include Phases
-        include Trains
+        include G1812::Entities
+        include G1812::Map
```
ollybh, line 18: These don't need to be qualified, do they?
```suggestion
        include Entities
        include Map
```

lib/engine/game/g_1812/game.rb
```diff
@@ -42,69 +42,341 @@ class Game < Game::Base
...
+        end
+
+        def replace_three_with_three_plus_one(remaining_trains_count)
+          @log << '--All remaining 3/2G trains in the supply replaced with with 3+1/2G+1 trains--'
+
+          remaining_trains_count.times do
+            train = @three_plus_one.shift
+            train.reserved = false
+            @depot.unshift_train(train)
+          end
```
ollybh, line 297: You're stepping through two arrays the same number of times. Using `Array.zip` lets you do this in a single loop. Something like this, maybe?
```suggestion
          replace_three_with_three_plus_one
        end

        def replace_three_with_three_plus_one
          trains_3t = depot.upcoming.select { |t| t.name == '3' }
          return unless trains_3t

          @log << '--All remaining 3/2G trains in the supply replaced with with 3+1/2G+1 trains--'
          trains_3t.zip(@three_plus_one) do |t3, t3plus|
            depot.forget_train(t3)
            t3plus.reserved = false
            @depot.unshift_train(t3plus)
          end
          @convert_3s = false
        end
```

lib/engine/game/g_1812/game.rb
```diff
@@ -42,69 +42,341 @@ class Game < Game::Base
...
+
+          # store the 3+1 trains in reserve for now
+
+          @three_plus_one = []
+          5.times do
+            train = depot.upcoming.find { |t| t.name == '3+1' }
+            @three_plus_one << train
+            depot.remove_train(train)
+            train.reserved = true
+          end
```
ollybh, line 311: You are stashing all of the 3+1 trains, aren't you? So you don't need to hardcode the '5' quantity. And I think this could be a little simpler too.
```suggestion
          @three_plus_one = depot.upcoming.select { |t| t.name == '3+1' }
          @three_plus_one.each do |train|
            depot.remove_train(train)
            train.reserved = true
          end
```

lib/engine/game/g_1812/game.rb
```diff
@@ -42,69 +42,341 @@ class Game < Game::Base
...
+
+          case @players.size
+          when 2
+            rejected = minors.sort_by { rand }.take(4)
+          when 3
+            rejected = minors.sort_by { rand }.take(2)
+          end
+
+          rejected.sort_by { |c| @corporations.index(c) }.each do |corp|
+            hex = @hexes.find { |h| h.id == corp.coordinates }
```
ollybh, line 354:
```suggestion
            hex = hex_by_id(corp.coordinates)
```

lib/engine/game/g_1812/game.rb
```diff
@@ -139,58 +396,118 @@ def can_par?(corporation, parrer)
...
+
+        def g_train?(train)
+          self.class::G_TRAINS.include?(train.name)
         end
 
-        def ns_bonus
-          @hexes.find { |hex| hex.coordinates == 'I1' }.tile.offboards.first
+        def check_other(route)
+          return if g_train?(route.train)
+          return unless (route.stops.map(&:hex).map(&:id) & PORT_HEXES).any?
```
ollybh, line 439: Don't use `any?` without a block. And, as you don't need to do anything with the port hexes on the route, `intersect?` would be a better method to use.
```suggestion
          return unless route.stops.map(&:hex).map(&:id).intersect?(PORT_HEXES)
```
I also suggest creating a `port_hexes` method, which caches the Hex objects for the port hexes. That would let you can remove the `map(&:id)` from that test.
```
def port_hexes
  @port_hexes ||= PORT_HEXES.map { |coord| hex_by_id(coord) }
end
```

lib/engine/game/g_1812/game.rb
```diff
@@ -139,58 +396,118 @@ def can_par?(corporation, parrer)
...
         end
 
         def revenue_for(route, stops)
           revenue = super
-          hex = route.hexes
-          gtrain = route.train.variant.name?(GTRAINS)
+          train = route.train
+          corp = route.corporation
+
+          revenue += 10 if corp.assigned?(p4_company) && stops.find { |s| C18_HEX.include?(s.hex.id) }
```
ollybh, line 449: The hex coordinate test is backwards, `C18_HEX` is a string, so `include?` is looking at substring matches. And `find` is the wrong test to use as you're not using the found stop.
```suggestion
          revenue += 10 if corp.assigned?(p4_company) && stops.any? { |s| s.hex.id == C18_HEX) }
```
You've made the first of these mistakes in the next few lines (but correctly use `any?` instead of `find`.

lib/engine/game/g_1812/game.rb
```diff
@@ -139,58 +396,118 @@ def can_par?(corporation, parrer)
...
+
+          revenue += 10 if g_train?(train) && corp.assigned?(p3_company) && stops.any? { |s| F3_PORT.include?(s.hex.id) }
+          revenue += 10 if g_train?(train) && corp.assigned?(p8_company) && stops.any? { |s| F3_PORT.include?(s.hex.id) }
+          revenue += 10 if g_train?(train) && corp.assigned?(p9_company) && stops.any? { |s| H9_PORT.include?(s.hex.id) }
+          revenue += 20 if g_train?(train) && corp.assigned?(p12_company) && stops.any? { |s| G6_PORT.include?(s.hex.id) }
```
ollybh, line 454: This very repetitive. Could you make a method that encapsulates these tests?

lib/engine/game/g_1812/game.rb
```diff
@@ -139,58 +396,118 @@ def can_par?(corporation, parrer)
...
+          revenue += port_mine_bonus(stops)[:revenue]
 
           revenue
         end
 
+        def north_south_bonus(stops)
+          bonus = { revenue: 0 }
+
+          north = stops.find { |stop| stop.tile.label&.to_s == 'N' }
+          south = stops.find { |stop| stop.tile.label&.to_s == 'S' }
```
ollybh, line 466:
```suggestion
          north = stops.any? { |stop| stop.tile.label&.to_s == 'N' }
          south = stops.any? { |stop| stop.tile.label&.to_s == 'S' }
```

lib/engine/game/g_1812/game.rb
```diff
@@ -139,58 +396,118 @@ def can_par?(corporation, parrer)
...
+          end
+
+          bonus
+        end
+
+        def port_mine_bonus(stops)
+          bonus = { revenue: 0 }
+
+          port = stops.find { |stop| stop.tile.icons.any? { |i| i.name == 'port' } }
+          mine = stops.find { |stop| stop.tile.icons.any? { |i| i.name == 'mine' } }
```
ollybh, line 480:
```suggestion
          port = stops.any? { |stop| stop.tile.icons.any? { |i| i.name == 'port' } }
          mine = stops.any? { |stop| stop.tile.icons.any? { |i| i.name == 'mine' } }
```

ollybh, line 480: Or would this work?
```suggestion
          port = stops.map(&:tile).map(&:icons).any? { |i| i.name == 'port' } }
          mine = stops.map(&:tile).map(&:icons).any? { |i| i.name == 'mine' } }
```

ollybh (changes requested):
This is much better now that it is based on the 1867 code. But I've found a few things that I think could be improved.

lib/engine/game/g_1812/step/buy_train.rb
philcampeau: 100% correct. I wasn't sure how to inherit a couple of methods directly from Engine::Step::BuyTrain, but I think I got it now. I needed to inherit from Engine because the 1867 methods included code I didn't want, which prevented me from calling `super`. Let me know if I can improve this, it's in update `fixes buy_train.rb inheritance`

lib/engine/game/g_1812/game.rb
philcampeau: done

lib/engine/game/g_1812/game.rb
philcampeau: done

lib/engine/game/g_1812/game.rb
philcampeau: done

lib/engine/game/g_1812/game.rb
philcampeau: done

lib/engine/game/g_1812/game.rb
philcampeau: For some reason I always get that backwards. Thanks!

lib/engine/game/g_1812/game.rb
philcampeau: Actually, I think I got something. Let me know what you think. Tbh, I feel like it looks more cluttered. The reason I had it as separate lines was they were all clean and easily read/understood.

lib/engine/game/g_1812/game.rb
philcampeau: done

lib/engine/game/g_1812/game.rb
```diff
@@ -446,12 +448,13 @@ def revenue_for(route, stops)
...
+          revenue += 10 if corp.assigned?(p4_company) && stops.any? { |s| s.hex.id == C18_HEX }
 
-          revenue += 10 if g_train?(train) && corp.assigned?(p3_company) && stops.any? { |s| F3_PORT.include?(s.hex.id) }
-          revenue += 10 if g_train?(train) && corp.assigned?(p8_company) && stops.any? { |s| F3_PORT.include?(s.hex.id) }
-          revenue += 10 if g_train?(train) && corp.assigned?(p9_company) && stops.any? { |s| H9_PORT.include?(s.hex.id) }
-          revenue += 20 if g_train?(train) && corp.assigned?(p12_company) && stops.any? { |s| G6_PORT.include?(s.hex.id) }
+          { p3_company => F3_PORT, p8_company => F3_PORT, p9_company => H9_PORT, p12_company => G6_PORT }.each do |company, port|
+            revenue += (company == p12_company ? 20 : 10) if g_train?(train) && corp.assigned?(company) && stops.any? do |s|
+                                                               s.hex.id == port
+                                                             end
```
ollybh, line 449: Yeah, this isn't clearer. The `HexBonus` ability _almost_ matches what you need here (and does exactly for P4), you just need some way of testing that the train is a goods train. Maybe add a `trains` array to that ability, then in `revenue_for` test both that you have a matching hex and a matching train.

lib/engine/game/g_1812/step/buy_train.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module Engine
+  module Game
+    module G1812
+      module Step
+        class BuyTrain < G1867::Step::BuyTrain
+          include Engine::Step::AutomaticLoan
+
+          def pass!
+            Engine::Step::BuyTrain.instance_method(:pass!).bind_call(self)
```
ollybh, line 15: This is a bit clunky. Could you just call `super`, and override `nationalize!` in the game class so that it does nothing?

lib/engine/game/g_1812/step/buy_train.rb
```diff
@@ -0,0 +1,33 @@
...
+
+          def discountable_trains_allowed?(_entity)
+            @game.phase.name.to_i >= 5
+          end
+
+          def buy_train_action(action, _entity = nil)
+            @depot_train = action.train.from_depot?
+
+            Engine::Step::BuyTrain.instance_method(:buy_train_action).bind_call(self)
+          end
```
ollybh, line 28: You could get rid of this method entirely, by having a `post_train_buy` method in the game class that does nothing (but with a comment explaining why you've done this).

lib/engine/game/g_1812/step/buy_train.rb
philcampeau: that is much simpler, yes! done.

lib/engine/game/g_1812/step/buy_train.rb
philcampeau: done

lib/engine/game/g_1812/game.rb
philcampeau: Ok, I did some clean up and improved some methods I think. 

commit: `cleans up revenue_for and associated methods`

lib/engine/game/g_1812/step/loan_operations.rb
philcampeau: @ollybh Thanks I must have missed/forgotten about this one. 

I'm pretty sure I get what you're saying. I submitted a new commit, have a look.

lib/engine/game/g_1867/step/loan_operations.rb
```diff
@@ -23,17 +23,17 @@ def skip!
...
-              # @todo: will this skip the rest of the entities turn?
-              return
-            end
+            return unpaid_interest(entity, owed) if owed
 
             @game.repay_loan(entity, entity.loans.first) while can_payoff?(entity)
             @game.calculate_corporation_interest(entity)
           end
+
+          def unpaid_interest(entity, owed)
```
ollybh, line 32: This is better, but I've changed my mind about the name of the method I suggested. `unpaid_interest` sounds like it should calculate the amount of interest that is unpaid. `unpaid_interest!` or `interest_unpaid!` would be better as they suggest it is an action that occurs when the interest is unpaid (or `credit_default!`, or `interest_default!`, or anything else you can think of.

lib/engine/game/g_1812/step/loan_operations.rb
```diff
@@ -0,0 +1,21 @@
...
+require_relative '../../g_1867/step/loan_operations'
+
+module Engine
+  module Game
+    module G1812
+      module Step
+        class LoanOperations < G1867::Step::LoanOperations
+          def unpaid_interest(entity, owed)
+            @log << "#{entity.name} owes #{format_currency(owed)} in loan interest but has #{format_currency(entity.cash)}"
+            @log << "#{entity.name} will pay all remaining treasury cash (#{format_currency(entity.cash)}) to the "\
```
ollybh, line 13: It needs to be `@game.format_currency` here, doesn't it?

lib/engine/game/g_1812/game.rb
```diff
@@ -420,68 +420,102 @@ def trainless_penalty(entity)
...
-          revenue += north_south_bonus(stops)[:revenue]
-          revenue += port_mine_bonus(stops)[:revenue]
-
           revenue
         end
 
-        def north_south_bonus(stops)
-          bonus = { revenue: 0 }
+        def bonus_for_p4_company(corp, stops)
+          return 10 if corp.assigned?(p4_company) && stops.any? { |s| s.hex.id == C18_HEX }
```
ollybh, line 486: The `assigned?` isn't right, that is for AssignCorporation abilities, which are generally ones where the ability can be used when the private company is owned by a player, who chooses which of their corporation gets to use the ability.

Can this be changed to give P4 (and the other privates with revenue bonuses) a HexBonus ability? You'd then have something like this code to work out the bonuses:
https://github.com/tobymao/18xx/blob/1c6e10f63f6889c61c6fe61493d5f1572d042e83/lib/engine/game/g_1867/game.rb#L681-L685
I don't think you'd need to add any extra tests to the loop for goods trains for the non-P4 privates, as (if I understand this correctly) they give bonuses for ports, and it isn't possible for non-goods trains to run to ports.

ollybh: Or `revenue += ability.amount if ability.hexes.intersect?(stops.map(&:hex).map(&:id))`

lib/engine/game/g_1812/step/loan_operations.rb
philcampeau: fixed

lib/engine/game/g_1867/step/loan_operations.rb
philcampeau: fixed

lib/engine/game/g_1812/game.rb
philcampeau: done

ollybh (approved):

PR #11541, opened 2025-02-13
[1867] Remove "medium" from corp sizes

philcampeau:
## Before clicking "Create"

### Explanation of Change

Looks like this bit was copied over from 1817, but 1861 and 1867 don't have medium sized 5-share corps.


ollybh:
It might be possible to completely remove `CORPORATIONS_SIZES` and `corporation_size`. At first glance only `corporation_size_name` is needed, and that doesn't require either of these. I'll try and have a proper look at this when I get a bit of time.

ollybh (approved):

PR #11568, opened 2025-02-20
Improve usability of create game form

### LINKED ISSUE #11071: [iOS] setting playercount gets overriden by helper logic
In a browser on iOS, values for min/max players are reset to defaults when focusing the field and entering a new value
e.g. if max_players is 7 then tapping min_players and entering 5 will reset to 7

Workaround is to long press on the field and 'Select All', then enter a value

Likely due to how we reset the element value to max/min based in the validation logic

https://github.com/tobymao/18xx/blob/master/assets/app/view/create_game.rb#L532

Since we have more frontend validation and messaging now, I'd propose:
1. don't reset those values (`elm.value`)
1. add helper text to the field label  e.g. `Max Players (7)`
1. dogfood to make sure it's usable

garrettngarcia:
Fixes #11071

## Before clicking "Create"

### Explanation of Change
Now selects number input field value on focus.  Also added min and max player info to the form labels.

Tested on Safari and Chrome on iOS and Mac.


ollybh (changes requested):
@garrettngarcia Thanks for looking into this highly annoying bug. However, I don't think that this is the correct way to fix this one.

The problem, as far as I understand it, is that we've got some over-zealous validation code. If you delete the player count from one of the boxes it sees the blank as zero, which is less than the minimum player could. So it helpfully re-populates the box with the minimum player count.

This needs to be redone so that the blank doesn't trigger adding the minimum player count. And we'd probably need to make sure that the form isn't submitted with an invalid player count – I think this validation code was added to prevent games being created with invalid counts.

garrettngarcia:
@ollybh I see.  I had it that way in an earlier version, but didn't want games to be created with invalid min and max so came up with the selection solution.  I can change it back.

Where should this form validation be done?  In JS or in the back end?  Can you point me to some current validation logic to get me started?

ollybh:
> Where should this form validation be done? In JS or in the back end? Can you point me to some current validation logic to get me started?

Ideally both. There does appear to already be some front end validation:
https://github.com/tobymao/18xx/blob/4fa14061688529794c8ebd288799cdb3137a13a7/assets/app/view/create_game.rb#L418-L421

I'm not sure how much back end validation there is. It might go in `Lib::Engine::Game::Base.initialize`. There is sort of a check for player count in `check_optional_rules!`…
https://github.com/tobymao/18xx/blob/4fa14061688529794c8ebd288799cdb3137a13a7/lib/engine/game/base.rb#L437-L442

…but the default implementation of `meta.check_options` does nothing:
https://github.com/tobymao/18xx/blob/4fa14061688529794c8ebd288799cdb3137a13a7/lib/engine/game/meta.rb#L130

michaeljb:
Closing this for #12037

Issue #11383, opened 2024-11-30
[18ESP] Token placement wrong


queroscia:
Hello,

Game Id: 179387.

MZA (yellow) company has changed its home base in Madrid (hex F24). MZA home base is the slot in the right upper corner, and this slot is the only that can share the slot with another company, but only with the minor MZ (gray).
The game is in phase 5, so there are not more minor companies, so MZA cannot share the slot with no companies. 
Having MZA and MCP sharing the same slot (MCP home base) is wrong. MCP company (blue) location is correct, but not MZA.
Thanks and regards,

Enrique (designer of 18ESP).

Issue #10729, opened 2024-05-07
18ESP - Private #2 – player must choose train


queroscia:
Hello,

Private #2 provides a 2 or 2+1 train, but are the players who must decide what type of train they want. In the game https://18xx.games/game/160684 company CFLG has purchased the P#2, but I have not had the opportunity to select the train, but the program directly has assigned a 1+2-train. (Maybe because the company is a Northern one and they run "usually" +trains?)
I am agree with the decision of the program for this game, but the bug is that the player should choose what type of train wants, and the program shouldn't assign plus train for Northern and conventional trains for Southern.

Thanks,

Enrique Trigueros (aka queroscia): designer of the game.

bentziaxl:
fixed a while ago

PR #11575, opened 2025-02-23
[18ES] fix minor 18esp errors

bentziaxl:
two minor rule changes/bug fixes:

1- when merging, corp doesn't unlock a new token. It either replaces an existing free token with one on map, or nothing (token of non survivor gets removed as usual)
2- minor companies can't own + trains (aka track_type narrow), when buying private that gets train, there's no choice, it gets a regular 2 train. Majors get the regular choice

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -1128,7 +1128,7 @@ module Engine
           move_assets(corporation, minor)
 
           # handle token
-          keep_token ? swap_token(corporation, minor) : gain_token(corporation, minor)
+          swap_token(corporation, minor) if keep_token
 
           # get share
           get_reserved_share(minor.owner, corporation) if minor.ipoed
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_esp/game.rb
```diff
@@ -1248,14 +1242,13 @@ def delete_token_mz(minor)
         end
 
         def swap_token(survivor, nonsurvivor)
-          new_token = survivor.tokens.last
+          new_token = survivor.next_token
```
ollybh, line 1245: Is it possible that `survivor` will not have any remaining tokens?

ollybh:
@bentziaxl Will this fix any of the open 18ESP issues? #11383? #10729? Could you please add "Fixes #[issue number]" to the PR description?

lib/engine/game/g_18_esp/game.rb
bentziaxl: No, as the method calling this ( start_merge(corporation, minor, keep_token)), is called by the acquire step, where the ability to swap is checked in can_swap? there, which checks for next_token of survivor.

PR #11576, opened 2025-02-23
[1867] Fix active entity for home token step

ollybh:
Commit d8abe5a (from pull request tobymao#11503) had changed the calculation of the active entities in a home token step. This had broken 1867, where the home token is placed by a player (the one who just started an auction), not a corporation.

This commit restores the previous behaviour.

Fixes tobymao#11565.

### DIFF
--- a/lib/engine/game/g_1867/game.rb
+++ b/lib/engine/game/g_1867/game.rb
@@ -830,7 +830,7 @@ module Engine
           G1867::Round::Stock.new(self, [
             G1867::Step::MajorTrainless,
             Engine::Step::DiscardTrain,
-            Engine::Step::HomeToken,
+            G1867::Step::HomeToken,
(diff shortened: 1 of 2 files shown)


crericha:
Is this an edge case the general solution should support or a special case that truly needs overriding?

ollybh:
> Is this an edge case the general solution should support or a special case that truly needs overriding?

¯\_(ツ)_/¯

I'd lean towards 1867 being a bit too unusual to need to be covered by the general solution. This fix is needed for when a new minor company is auctioned, the home token location is chosen by the person starting the auction. So it is done before the company has a president and before it has floated. I can't think of any other 18xx games that have a case like this, not even in the 1861/1812 family of games.

crericha (approved):

PR #11585, opened 2025-02-27
[18Uruguay] Corrected aquire fce share sequence

### LINKED ISSUE #11578: [18Uruguay] Player unable to payoff loans or pass
Game ID 197633

https://18xx.games/game/197633

Game is in state with the error. You can back up, but are unable to move passed the current point in the game. Not even in a cloned game.

The error occurred during the nationalization phase. The error given is "all shares must come from the same company"

Opera (Mobile)
Firefox (PC)

patrikolesen:
Fixes #11578

Also corrected half share payout so there is a small risk that games that has done the nationalization will brake 

## Before clicking "Create"


ollybh (approved):
I always forget that zero is truthy in Ruby. I've added the pins label to remind that we'll need to check for breakage when this is deployed.

PR #11586, opened 2025-02-27
[Core] Move marking trains as not operated to start of operating round

crericha:
This PR moves marking trains as not operated from before the entity runs to the start of the operating round. This allows for in operating round mergers to correctly not allow merged in trains to run.


roseundy (commented):
I think this is OK for 1841, since I don't think the rules care about whether a train has operated before.

ollybh:
@crericha Why are the 18NL fixture files deleted in this PR? Were they failing after this change?

PR #11595, opened 2025-03-02
[18Norway] You are only allowed to buy the cheapest train during emergency

patrikolesen:
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -34,15 +34,15 @@ module Engine
         SELL_BUY_ORDER = :sell_buy
         TILE_RESERVATION_BLOCKS_OTHERS = :always
         CURRENCY_FORMAT_STR = '%skr'
-        EBUY_SELL_MORE_THAN_NEEDED = true
+        EBUY_SELL_MORE_THAN_NEEDED = false
         CAPITALIZATION = :incremental
         MUST_BUY_TRAIN = :always
         POOL_SHARE_DROP = :none
         SELL_AFTER = :p_any_operate
         SELL_MOVEMENT = :left_block
         HOME_TOKEN_TIMING = :float
-
-        EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false
+        EBUY_PRES_SWAP = false
+        EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = true
         MUST_EMERGENCY_ISSUE_BEFORE_EBUY = true
 
         BANKRUPTCY_ENDS_GAME_AFTER = :one
@@ -217,6 +217,16 @@ module Engine
           hex_by_id('I27').neighbors[2] = hex_by_id('H26')
         end
 
+        def init_train_handler
+          trains = game_trains.flat_map do |train|
+            Array.new((train[:num] || num_trains(train))) do |index|
+              Train.new(**train, index: index)
+            end
+          end
+
+          G18Norway::Depot.new(trains, self)
+        end
+
         def p4
           @p4 ||= company_by_id('P4')
         end
@@ -312,10 +322,13 @@ module Engine
           train.track_type == :narrow
         end
 
-        def cheapest_train_price
+        def cheapest_train
           depot_trains = depot.depot_trains.reject { |train| ship?(train) }
-          train = depot_trains.min_by(&:price)
-          train.price
+          depot_trains.min_by(&:price)
+        end
+
+        def cheapest_train_price
+          cheapest_train.price
         end
 
         def can_go_bankrupt?(player, corporation)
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -217,6 +217,16 @@ def setup
...
 
+        def init_train_handler
+          trains = game_trains.flat_map do |train|
+            Array.new((train[:num] || num_trains(train))) do |index|
+              Train.new(**train, index: index)
+            end
+          end
+
+          G18Norway::Depot.new(trains, self)
+        end
```
ollybh, line 228: Overriding this method shouldn't be needed. Just set `DEPOT_CLASS = G18Norway::Depot` and the default implementation of `init_train_handler` will work for you.

lib/engine/game/g_18_norway/steps/buy_train.rb
```diff
@@ -75,6 +84,19 @@ def spend_minmax(entity, _train)
...
           end
 
+          def check_for_cheapest_train(train)
+            cheapest = @game.cheapest_train
+            cheapest_names = names_of_cheapest_variants(cheapest)
+            raise GameError, "Cannot purchase #{train.name} train: cheaper train available (#{cheapest_names.first})" if
+              !cheapest_names.include?(train.name) &&
+              @game.class::EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST &&
+              (@game.class::EBUY_FROM_OTHERS == :never || train.from_depot?)
+          end
```
ollybh, line 94: Is this right? The rules say that the company must by the cheapest available train, but the rules don't count ships as being trains. I think that this is going to force the company to buy a ship instead of a train. The base implementation of `check_for_cheapest_train` uses `@depot.min_depot_train` which would make sure that it is a train being bought, not a ship.

ollybh (changes requested):
Use `DEPOT_CLASS` instead of overriding `init_train_handler`. Also check that the emergency train buying rules are correct.

lib/engine/game/g_18_norway/game.rb
patrikolesen: Updated

lib/engine/game/g_18_norway/steps/buy_train.rb
patrikolesen: You are correct about the rules ant that was why did a override of check_for_cheapest_train.
@game.cheapest_train will give you the train and not a ship as it was prior to this change

ollybh: Apologies, I had misread `cheapest_train`. But the `cheapest_train` method is now exactly the same as `Depot.min_depot_train`. Right? Won't the default version of `check_for_cheapest_train` work correctly?

ollybh: @patrikolesen Do you need this version of `check_for_cheapest_train`?

patrikolesen: Probably not anymore let me remove it and validate.
I think your other comment is valid as well but in a slightly different flavor.
With this new depot solution we will probably not be able to by a ship in the case the company already have a train, let me look into that as well

patrikolesen: Now updated

lib/engine/game/g_18_norway/steps/buy_train.rb
```diff
@@ -41,11 +41,25 @@ def must_buy_train?(entity)
...
           def buyable_trains(entity)
-            super.reject do |train|
-              next false unless @game.ship?(train)
+            # We need to add the ships in the case that the company have a train but have less money then the next available train
+            ships = if must_buy_train?(entity) || entity.cash >= @depot.min_depot_price
+                      []
+                    else
+                      @depot.depot_trains.reject do |train|
+                        true unless @game.ship?(train)
+                        true if train.price > entity.cash
```
ollybh, line 50: Does this work? I would have thought that the first line in the block would do nothing, shouldn't it be `next true unless @game.ship?(train)`? Or you could rewrite the block as:
```suggestion
                        @game.ship?(train) || train.price > entity.cash
```

patrikolesen: It looked like it worked but I have compacted the implementation a bit

lib/engine/game/g_18_norway/game.rb
```diff
@@ -312,10 +314,13 @@ def ship?(train)
...
 
-        def cheapest_train_price
+        def cheapest_train
           depot_trains = depot.depot_trains.reject { |train| ship?(train) }
-          train = depot_trains.min_by(&:price)
-          train.price
+          depot_trains.min_by(&:price)
+        end
+
+        def cheapest_train_price
```
ollybh, line 322: This gets called with a parameter from the buy train step. It's not used here, but should be included in the method signature.
```suggestion
        def cheapest_train_price(_corporation)
```

ollybh (approved):

PR #11602, opened 2025-03-03
[1837] Use backtracking graph in the token step

ollybh:
### Explanation of Change

1837 has two graphs: a standard one and one that allows track to backtrack at junctions. The latter is needed for placing station tokens, where this type of route is allowed.

The token step was not using this graph, it was using the standard one. This meant that not all possible token locations were being included, and an auto-pass was being generated for the token step if the only possible token locations involved backtracking.

Fixes tobymao#11600.

This shouldn't break any games. There will be games where the BH should have been able to lay a token on its first turn but wasn't allowed to, but a pass action will have been generated for those turns and that will be in the game actions.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1837/step/token.rb
+++ b/lib/engine/game/g_1837/step/token.rb
@@ -27,7 +27,7 @@ module Engine
             # Cheaper to do the graph first, then check affordability
             current_entity == entity &&
               !(token = entity.next_token).nil? &&
-              token_graph.can_token?(entity) &&
+              @game.token_graph.can_token?(entity) &&
               can_afford_token?(token, buying_power(entity))
           end
 
@@ -40,7 +40,7 @@ module Engine
           end
 
           def can_afford_token?(token, cash)
-            token_graph.tokenable_cities(token.corporation).any? do |city|
+            @game.token_graph.tokenable_cities(token.corporation).any? do |city|
               token_price(token, city.tile.hex) <= cash
             end
           end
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1837/step/token.rb
```diff
@@ -27,7 +27,7 @@ def can_place_token?(entity)
             # Cheaper to do the graph first, then check affordability
             current_entity == entity &&
               !(token = entity.next_token).nil? &&
-              @game.graph.can_token?(entity) &&
+              @game.token_graph.can_token?(entity) &&
```
crericha, line 30: `@game.token_graph_for_entity(entity).can_token?(entity)[hex]`

lib/engine/game/g_1837/step/token.rb
```diff
@@ -40,7 +40,7 @@ def multiple_tokens?(entity)
           end
 
           def can_afford_token?(token, cash)
-            @game.graph.tokenable_cities(token.corporation).any? do |city|
+            @game.token_graph.tokenable_cities(token.corporation).any? do |city|
```
crericha, line 43: `@game.token_graph_for_entity(entity).tokenable_cities(token.corporation).any?`

lib/engine/game/g_1837/step/token.rb
ollybh: Updated (even though it does the same thing).

lib/engine/game/g_1837/step/token.rb
ollybh: Updated.

lib/engine/game/g_1837/step/token.rb
crericha: It does today, but the implementations can diverge later and I'd rather have the code use the one its supposed to.

crericha (approved):

PR #11605, opened 2025-03-05
[18ESP] minor 18esp rand fix and move to beta

bentziaxl:
the rand generator doesn't work well when doing modulo even numbers, it is impacting the code in 18esp, as the random corp selected for random_share and random_president is not truly random.
A quick and easy fix is to divide the rand number by a 100, to get a more random distribution.

also moving 18esp to beta. Removed 1837 from welcome.rb as it's been there for over a month.

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -472,8 +472,10 @@ module Engine
 
         def init_company_abilities
           northern_corps = @corporations.select { |c| north_corp?(c) }
-          random_corporation = northern_corps[rand % northern_corps.size]
-          another_random_corporation = northern_corps[rand % northern_corps.size]
+          #  the PRNG doesn't work well when doing mod even numbers, as
+          # the lower digits aren't truly random. Right shifting by 12 fixes the issue.
+          random_corporation = northern_corps[(rand >> 12) % northern_corps.size]
+          another_random_corporation = northern_corps[(rand >> 12) % northern_corps.size]
           @companies.each do |company|
             next unless (ability = abilities(company, :shares))
 
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -18,7 +18,7 @@ module View
       message = <<~MESSAGE
         <p><a href='https://www.kickstarter.com/projects/1840vienna/1807-the-big-four'><em>1807: The Big Four</em> is now on Kickstarter</a>.</p>
 
-        <p><a href="https://github.com/tobymao/18xx/wiki/1837">1837</a> is now in alpha.</p>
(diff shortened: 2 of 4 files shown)


lib/engine/game/g_18_esp/game.rb
```diff
@@ -472,8 +472,8 @@ def find_and_remove_train_for_minor(train_id, buyable = true)
 
         def init_company_abilities
           northern_corps = @corporations.select { |c| north_corp?(c) }
-          random_corporation = northern_corps[rand % northern_corps.size]
-          another_random_corporation = northern_corps[rand % northern_corps.size]
+          random_corporation = northern_corps[(rand / 100) % northern_corps.size]
+          another_random_corporation = northern_corps[(rand / 100) % northern_corps.size]
```
ollybh, line 476: This might well have a problem with differences in behaviour between the Ruby and Opal versions of the code. In Ruby `rand / 100` will be integer division and return an integer, but in Opal you will get a floating point number.

If you've got problems using the least significant bits of `rand` then use a bit shift operator, as suggested on Slack.

ollybh (changes requested):

lib/engine/game/g_18_esp/game.rb
bentziaxl: i can't get bit shifting to work. I always end up with an even number. 

Would something like this work in Opal?
rand.to_s[0..-3].to_i

I just want to remove the last 3 digits of the rand number, this seems to make it rand enough for this applicaiton that i don't get a pattern

bentziaxl: nvm, i was shifting left and not right. Looks good. fixing it to >> 12

ollybh: That looks better.
I suggest you add a comment explaining why the bit shift is necessary. There's nothing in the code itself that makes this obvious.

ollybh:
@bentziaxl I assume that this is going to break a bunch of games when this is deployed, as they'll get different companies excluded from the game. Right?

bentziaxl:
the rand isn't used to exclude corps, but to determine which shares you get with two of the starting privates. 
But it will break pretty much every game. Which is fine, 18esp is in alpha, breakages are expected. no need to pin any of them.

bentziaxl:
merging, comments have been addressed. The only other change is tile numbering in manifest, which is trivial.

ollybh:
@bentziaxl With all these changes to the 18ESP code and needing to archive most/all existing games, I think  it might be better to back out the move to beta status for now, to allow a little time to see if everything is working as expected.

bentziaxl:
makes sense, let me fix

PR #11615, opened 2025-03-12
[1844] sets min buy-in price to 1 SFR

### LINKED ISSUE #11551: [1844] Unable to purchase private below half price
[Game ID 194981](https://18xx.games/game/194981)

`https://18xx.games/game/31337`

**How to Reproduce the Bug?**
Go back to 
1844: Phase 4 - Operating Round 6.1 (of 2) - Buy Trains
[07:15] RhB buys a 3H train for 80 SFR from AB

Try to buy P4 for 1 SFR
Note shows 55-220 required for purchase price, but rules allow 1 to 220

Image

Chrome Win 11
Version 132.0.6834.160 (Official Build) (64-bit)
-

philcampeau:
Fixes #11551 

## Before clicking "Create"

### Explanation of Change

Currently, buy-in price was using the default half to double. Rules state it's 1 to double.

I created a new company_price_one_to_double.rb file in /lib/engine/game (similar to the other current files there) and made the necessary changes to 1844's game.rb code.

I needed to delete the fixture file, and likely many (probably all) existing games will need to be pinned, since any game that skipped over the buy companies blocking step will now re-evaluate that decision if a corp has at least 1 SFR in its treasury.


ollybh:
The default game migration script should be able to fix any games broken by this. It will add additional pass actions where needed.

ollybh (approved):

crericha:
I think this would work, but I think the extra modules and hoops to call it are unnecessary. I prefer the simple approach of being explicit in the entities.rb file.

philcampeau:
I was proposing adding that module since there are other games that also use a 1 to double buy-in price, and they also code it explicitly in the game code. Since we already have modules for company_price_50_to_150 and company_price_up_to_face, why not follow that standard?

philcampeau:
@crericha I was proposing adding that module since there are other games that also use a 1 to double buy-in price, and they also code it explicitly in the game code. Since we already have modules for company_price_50_to_150 and company_price_up_to_face, why not follow that standard?

PR #11617, opened 2025-03-13
[18Uruguay] Payoff loans after nationalizations

### LINKED ISSUE #11616: [18Uruguay] Forced repay of loans
https://18xx.games/game/193721

If a corporation is not closed and has loan left after it shall withhold the revenue and pay off the loans.
In current version it will only withhold the revenue and never pay of the loans

patrikolesen:
Fixes #11616

Payoff loans after nationalizations
Might break existing games

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_uruguay/step/dividend.rb
+++ b/lib/engine/game/g_18_uruguay/step/dividend.rb
@@ -41,6 +41,16 @@ module Engine
             [Engine::Action::Dividend.new(current_entity, kind: 'withhold')]
           end
 
+          def payoff_loans(entity)
+            loans_to_pay_off = [(entity.cash / 100).floor, entity&.loans&.size].min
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_uruguay/step/dividend.rb
```diff
@@ -41,6 +41,16 @@ def auto_actions(entity)
             [Engine::Action::Dividend.new(current_entity, kind: 'withhold')]
           end
 
+          def payoff_loans(entity)
+            loans_to_pay_off = [(entity.cash / 100).floor, entity&.loans&.size].min
```
ollybh, line 45: You don't need the safe navigation operator here. It doesn't do any harm, but you're testing `entity.loans.size` before invoking this method, so you're never going to have nil values.
```suggestion
            loans_to_pay_off = [(entity.cash / 100).floor, entity.loans.size].min
```

ollybh (approved):

PR #11619, opened 2025-03-14
[1858] Clear pass flag when bidding for a private

ollybh:
## Implementation Notes

A player's pass status was not being cleared when a bid for a private railway company was placed. This meant that it was possible for the initial stock round to end prematurely – if a player had passed on an auction earlier in the round but bid on a later one they were still marked as having passed, meaning the round could end if all other players had passed.

Fixes tobymao#11446.

This breaks a few games (seven of 1858, two of 1858 Switzerland, according to my testing) where the stock round had been ended too early. These will require pinning.

## Before clicking "Create"


ianwilson156:
Great!

PR #11621, opened 2025-03-14
[1858] Disable autopass on private railway company bids

### LINKED ISSUE #11620: [1858] Auto-pass causing confusion in initial stock round
> This might be caused by auto-pass being enabled. Auto-pass doesn't make much sense in the 1858 private company auction round, it should not be allowed. 

 _Originally posted by @ollybh in [#10684](https://github.com/tobymao/18xx/issues/10684#issuecomment-2081554354)_

When a private railway auction is underway, a player will be presented with options to bid, pass, or auto-pass. The last of these will make them pass the current auction, but also then pass when any other companies are put up for auction in the same stock round. This is causing confusion, as some players have been clicking that button and not intending to pass on subsequent auctions.

ollybh:
## Implementation Notes

Players will sometimes click the autopass button to pass during an auction, then get confused when they are not allowed to bid on future auctions in the stock round.

This PR clears these programmed passes when a bid on a private railway is made, unless the player can't afford to bid on any available private railway companies. This exception is for later stock rounds where there are green privates available, but a player has spent all their money and doesn't want to be prompted to sell shares.

Fixes #11620.

This will not require any pins. There are games where players should have been given the option to participate in auctions, but these will not be broken by this change, as the pass actions have been inserted in the game action list.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1858/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1858/step/buy_sell_par_shares.rb
@@ -183,6 +183,31 @@ module Engine
             super
           end
 
+          def process_bid(action)
+            # Players will sometimes click the autopass button to pass during an
+            # auction, then get confused when they are not allowed to bid on
+            # future auctions in the stock round. Clear these programmed passes,
+            # unless the player can't afford to bid on any available private
+            # railway companies (this check is for later stock rounds where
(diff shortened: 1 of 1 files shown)


ollybh:
FFS. Messed up the merging. Will redo.

PR #11629, opened 2025-03-19
[18FR] Add train discounts when exporting

Galatolol:
## Before clicking "Create"

### Explanation of Change

When a train is exported, all trains of that type are discounted by 10% of the base price. If they already were at the minimum price, all are exported. If 6*D hits the minimum price, the end game is triggered.

Apart from that, players may buy trains from companies for $0.


lib/engine/game/g_18_fr/train.rb
```diff
@@ -0,0 +1,29 @@
...
+        attr_reader :base_price, :final_price
+
+        def initialize(name:, distance:, price:, index: 0, **opts)
+          super
+
+          @base_price = price
+          @final_price = opts[:final_price] || price
+        end
+
+        def new_price(price)
```
ollybh, line 16: You could do this as a setter method:
```suggestion
        def price=(price)
```
That would mean you would be able to say `train.price = xxx` instead of `train.new_price(xxx)` which feels a bit more natural to me.

ollybh (approved):

PR #11630, opened 2025-03-20
[1848] Adds lowest bid price info to auction

### LINKED ISSUE #11041: [1848] Add minimum price information to the private certificates
In the 1848  auction, it would be helpful to see the minimum price for each private.  
I think this would be more useful than the supposed face value, which is of no real consequence.

philcampeau:
Fixes #11041 

## Before clicking "Create"

### Explanation of Change

Currently, when you reduce the price of a private in 1848's dutch auction, it looks like the private on the right here: 

image

I added a new property to private companies called lowest_bid_level, and then called it in the rendering page for companies in order to display more information . The result looks like this:

image

I toyed with just placing Value with the Current Price info, but I thought it would still be useful to know what the original price was.

I double-checked with several other games and didn't see any of them affected by this change.

Please let me know if you have better ideas for what to call each of the fields, I've never been great at naming things.

### Any Assumptions / Hacks

Since the information is now stored in a property on the company, I removed a constant in the dutch_auction.rb file and called the information directly where it was needed instead.

### DIFF
--- a/assets/app/view/game/company.rb
+++ b/assets/app/view/game/company.rb
@@ -128,7 +128,7 @@ module View
             float: 'left',
           }
 
-          lowest_bid_price_style = {
+          lowest_bid_style = {
             float: 'right',
           }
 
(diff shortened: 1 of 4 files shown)


bentziaxl (approved):
looks good, thanks for the fix

Issue #10459, opened 2024-03-08
System18 Rules


scottredracecar:
System18 rules are here.
https://www.dropbox.com/scl/fi/or38t89eczc5eki8e21y7/System18_Rules.pdf?rlkey=n9kblkl7kvnz96oobs6hsatk2&dl=0

I think it will still be best to link to the wiki for map-specific rules.

philcampeau:
I think the Rules link going to the Wiki makes the most sense, since there are a lot of things that are only there. That said, I added the link to the rulebook to the wikipage.

PR #11631, opened 2025-03-20
[System18] Changes requested by designer

### LINKED ISSUE #10456: System18 Incremental Cap Rules - Price moves down one space
Steve, I am changing the stock price change rules for the **incremental cap** game.

For both president sales and EMR issuing shares, the price moves down one space per block of shares sold/issued.

Also, share issuing is done automatically in one block. It should not be allowed to sell shares in multiple blocks. In practice, this could be an auto action, but I am not sure if that is worthwhile to code.

Ask me if you have questions--I just did another draft of these rules.

philcampeau:
Fixes #10456

## Before clicking "Create"

### Explanation of Change

Requested changes from designer: 

- For both president sales and EMR issuing shares, the price moves down one space per block of shares sold/issued.
- Also, share issuing is done automatically in one block. It should not be allowed to sell shares in multiple blocks.

I made these changes. Additionally, I noted that there were two separate variables tracking if shared had been emergency issued: @emergency_issued and @emr_issue. I consolidated these together as @emergency_issued, as they didn't appear to be tracking different information. 

I also removed this code below, since @emergency_issued is already being reset  in the `def setup` method of that same step.

https://github.com/tobymao/18xx/blob/29e21ad40ea90842d980341ffb00980629e52741/lib/engine/game/g_system18/step/buy_train.rb#L37-L40

### DIFF
--- a/lib/engine/game/g_system18/game.rb
+++ b/lib/engine/game/g_system18/game.rb
@@ -427,6 +427,7 @@ module Engine
             redef_const(:SELL_MOVEMENT, :left_block_pres)
             redef_const(:SOLD_OUT_INCREASE, true)
             redef_const(:MUST_EMERGENCY_ISSUE_BEFORE_EBUY, true)
+            redef_const(:MUST_SELL_IN_BLOCKS, true)
             redef_const(:BANKRUPTCY_ENDS_GAME_AFTER, :all_but_one)
           else
             ######################################################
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_system18/meta.rb
```diff
@@ -15,7 +15,7 @@ module Meta
         GAME_LOCATION = 'Various'
         GAME_PUBLISHER = :all_aboard_games
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/System18'
-        GAME_RULES_URL = 'https://github.com/tobymao/18xx/wiki/System18'
+        GAME_RULES_URL = 'https://www.dropbox.com/scl/fi/or38t89eczc5eki8e21y7/System18_Rules.pdf'
```
ollybh, line 18: The link needs to include the `rkley` parameter (from #10459). That lets you access the file without being signed into Dropbox.
```suggestion
        GAME_RULES_URL = 'https://www.dropbox.com/scl/fi/or38t89eczc5eki8e21y7/System18_Rules.pdf?rlkey=n9kblkl7kvnz96oobs6hsatk2&dl=0'
```

lib/engine/game/g_system18/step/buy_train.rb
```diff
@@ -14,7 +14,7 @@ def actions(entity)
           end
 
           def setup
-            @emr_issue = false
+            @emergency_issued = false
```
ollybh, line 17: You're referencing both `@emergency_issued` and `@round.emergency_issued` in this file. These will be two different variables.

lib/engine/game/g_system18/game.rb
```diff
@@ -524,22 +525,28 @@ def emergency_issuable_bundles(entity)
...
+          issuable_bundles.sort_by!(&:price)
+
+          # Find the smallest bundle that raises enough cash to buy the train
+          issuable_bundles.each do |bundle|
+            total_cash = bundle.price + entity.cash
+            return [bundle] if total_cash >= min_train_price
+          end
+
+          # If no single bundle is enough, return the largest bundle
+          [issuable_bundles.last]
```
ollybh, line 549: `issuable_bundles` _could_ contain share bundles for more than one corporation. I guess this is for an 1841-style game where corporations can own shares of other corporations, and these could be issued in emergency money raising.

This is going to return a single bundle for one corporation. It should return a single bundle for each corporation. So this bit needs to be moved into the `@corporations.flat_map` block.

I don't know if any currently implemented rules for System18 allow corporations to own other corporations' shares, but this has been written to allow this, so should be preserved.

ollybh (changes requested):

philcampeau:
Whoops, totally missed those notes. Will correct this soon!

PR #11634, opened 2025-03-21
[System18] no corp started game end tweak

### LINKED ISSUE #11104: [System18: Poland] Game ended due to no player starting a company in SR1, but game continued to OR1.1 (minor)
https://18xx.games/game/172620

Two out of three players bought concessions in PR1, but neither of them chose to start a company in SR1, so the game ended - this is correct.
But the game progressed to OR1.1 - this shouldn't happen, the game has ended!

philcampeau:
Fixes #11104 

## Before clicking "Create"

### Explanation of Change
Because of the way it was implemented, if the game ended due to no one starting a corp in SR1, the game would end but the log would then indicate it was OR 1.1:

image

I moved around the sequence of events to prevent this.

### DIFF
--- a/lib/engine/game/g_system18/map_britain_customization.rb
+++ b/lib/engine/game/g_system18/map_britain_customization.rb
@@ -320,8 +320,8 @@ module Engine
             case @round
             when Engine::Round::Stock
               map_britain_stock_round_finished
-              @operating_rounds = @phase.operating_rounds
               reorder_players
+              @operating_rounds = @phase.operating_rounds
               new_operating_round
             when Round::Operating
               if @round.round_num < @operating_rounds
--- a/lib/engine/game/g_system18/map_poland_customization.rb
+++ b/lib/engine/game/g_system18/map_poland_customization.rb
@@ -267,8 +267,10 @@ module Engine
             case @round
             when Engine::Round::Stock
               map_poland_stock_round_finished
-              @operating_rounds = @phase.operating_rounds
               reorder_players
+              return if corporations.none?(&:floated)
+
+              @operating_rounds = @phase.operating_rounds
               new_operating_round
             when Round::Operating
               if @round.round_num < @operating_rounds
--- a/lib/engine/game/g_system18/game.rb
+++ b/lib/engine/game/g_system18/game.rb
@@ -508,7 +508,7 @@ module Engine
         def reorder_players
           if corporations.none?(&:floated)
             @log << '-- Stock round ended with no floated corporations. Ending game.'
-            end_game!
+            return end_game!
           end
 
           super
--- a/lib/engine/game/g_system18/map_northern_italy_customization.rb
+++ b/lib/engine/game/g_system18/map_northern_italy_customization.rb
@@ -291,8 +291,8 @@ module Engine
             case @round
             when Engine::Round::Stock
               map_northern_italy_stock_round_finished
-              @operating_rounds = @phase.operating_rounds
               reorder_players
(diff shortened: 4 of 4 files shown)


lib/engine/game/g_system18/map_britain_customization.rb
```diff
@@ -320,8 +320,8 @@ def map_britain_next_round!
             case @round
             when Engine::Round::Stock
               map_britain_stock_round_finished
-              @operating_rounds = @phase.operating_rounds
               reorder_players
+              @operating_rounds = @phase.operating_rounds
```
ollybh, line 324: What difference does reordering these lines make?

philcampeau: If `reorder_players` happens after `@operating_rounds = @phase.operating_rounds`, the log shows the switch to the OR before the game ends. It doesn't have a functional difference, just esthetic.  It's because I'm sending the `return end_game!` command in the `reorder_players` method.

ollybh: Does this suppress the log message? It comes from `new_operating_round`, which is still being called.
Instead, would sticking an `unless @finished` check on the `new_operating_round` line work?

philcampeau: @ollybh I tested this, it functionally makes no difference. In both cases the log output is the same.

ollybh: @philcampeau Can you please check this again, just to make sure? I can't see how the code in this PR doesn't output the log message, the first line of `new_operating_round` has an unconditional log call.

michaeljb: Could this be a case for some more tests in [`game_state_spec.rb`](https://github.com/tobymao/18xx/blob/master/spec/game_state_spec.rb)?

philcampeau: Hmmm @ollybh I'm not sure if I was missing it before or what, but yeah now I'm still getting the operating round log message. 

Adding `unless @finished` results in a console error of `Uncaught round_num: undefined method 'round_num' for nil` but this works: 

```
        def map_britain_next_round!
          @round =
            case @round
            when Engine::Round::Stock
              map_britain_stock_round_finished
              @operating_rounds = @phase.operating_rounds
              reorder_players
              return if @finished

              new_operating_round
```

What do you think about that? I'd make the same changes for the other games. Here's the log output now: 

image

roseundy:
So having implemented a new map - Mississippi - that uses reorder_players after an initial auction round, I'm not a fan of putting the check there.  Meaning: it caused my game to end before the first stock round.

lib/engine/game/g_system18/map_poland_customization.rb
```diff
@@ -269,6 +269,8 @@ def map_poland_next_round!
               map_poland_stock_round_finished
               @operating_rounds = @phase.operating_rounds
               reorder_players
+              return if corporations.none?(&:floated)
+
               new_operating_round
             when Round::Operating
               if @round.round_num < @operating_rounds
```
roseundy, line 276: Why is the Poland map handled differently than Britain and Northern Italy?

roseundy:
I think a better way to solve this is to create a custom Stock round for System18 and put the end-game check in the finish_round method.

In fact, I'm going to do this in my current PR (for the Mississippi map), so this PR can be closed.

philcampeau:
Sounds good!

PR #11640, opened 2025-03-23
[18Norway] Reimplement that you only are allowed to have one ship per  type

patrikolesen:
Removed a bit to much in latest commit re-adding the filter for only one ship per type

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/depot.rb
+++ b/lib/engine/game/g_18_norway/depot.rb
@@ -8,7 +8,7 @@ module Engine
       class Depot < Engine::Depot
         def min_depot_train
           current_entity = @game.round.current_entity
-          return super unless current_entity.trains.empty?
+          return super if current_entity.trains.any? { |train| !@game.ship?(train) }
 
           # Ships doesn't count towards needing a train, should be ignored when checking for min
           depot_trains.reject { |train| @game.ship?(train) }.min_by(&:price)
--- a/lib/engine/game/g_18_norway/depot.rb
+++ b/lib/engine/game/g_18_norway/depot.rb
@@ -7,6 +7,9 @@ module Engine
     module G18Norway
       class Depot < Engine::Depot
         def min_depot_train
+          current_entity = @game.round.current_entity
+          return super unless current_entity.trains.empty?
+
           # Ships doesn't count towards needing a train, should be ignored when checking for min
           depot_trains.reject { |train| @game.ship?(train) }.min_by(&:price)
         end
--- a/lib/engine/game/g_18_norway/steps/buy_train.rb
+++ b/lib/engine/game/g_18_norway/steps/buy_train.rb
@@ -41,11 +41,14 @@ module Engine
           end
 
           def buyable_trains(entity)
-            return super if must_buy_train?(entity) || entity.cash >= @depot.min_depot_price
+            trains = super.reject do |train|
+              next false unless @game.ship?(train)
+              next true if must_buy_train?(entity)
 
-            # We need to add the ships in the case that the company have a train but have less money then the next available train
-            super + @depot.depot_trains.select do |train|
-              @game.ship?(train) && entity.cash >= train.price
+              entity.trains.any? { |ship| ship.name == train.name }
+            end
+            trains + @depot.depot_trains.select do |train|
+              @game.ship?(train) && entity.cash >= train.price && entity.trains.none? { |ship| ship.name == train.name }
             end
           end
 
--- a/lib/engine/game/g_18_norway/steps/buy_train.rb
+++ b/lib/engine/game/g_18_norway/steps/buy_train.rb
@@ -42,17 +42,6 @@ module Engine
 
           def buyable_trains(entity)
             trains = super
-            # Due to that the min_depot only looks at trains we need to add a ship in the case where
-            # we already have a train, can afford the next ship but not the next train
-            if entity.cash < cheapest_train_price(entity)
-              trains += @depot.depot_trains.select do |train|
-                next unless @game.ship?(train)                                # Must be a ship
-                next if entity.trains.any? { |ship| ship.name == train.name } # Must not already own the ship
-                next if trains.any? { |ship| ship.name == train.name }        # Must not already be in the list
-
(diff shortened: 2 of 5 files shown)


lib/engine/game/g_18_norway/steps/buy_train.rb
```diff
@@ -41,11 +41,14 @@ def must_buy_train?(entity)
           end
 
           def buyable_trains(entity)
-            return super if must_buy_train?(entity) || entity.cash >= @depot.min_depot_price
+            trains = super.reject do |train|
+              next false unless @game.ship?(train)
```
crericha, line 45: Nit: `next` will return nil (which is false) if no value provided. That's the standard way of doing this.

lib/engine/game/g_18_norway/steps/buy_train.rb
```diff
@@ -41,11 +41,14 @@ def must_buy_train?(entity)
...
+              next true if must_buy_train?(entity)
```
crericha, line 46: I don't think this will improve performance...maybe even slow it down. I recommend removing it.

lib/engine/game/g_18_norway/steps/buy_train.rb
```diff
@@ -41,11 +41,14 @@ def must_buy_train?(entity)
...
 
-            # We need to add the ships in the case that the company have a train but have less money then the next available train
-            super + @depot.depot_trains.select do |train|
-              @game.ship?(train) && entity.cash >= train.price
+              entity.trains.any? { |ship| ship.name == train.name }
+            end
+            trains + @depot.depot_trains.select do |train|
```
crericha, line 50: In the `Train` step's implementation of `buyable_trains`, depot trains are included. Since you are calling super and getting those, why do you ned to explicitly get depot_trains here?

lib/engine/game/g_18_norway/steps/buy_train.rb
patrikolesen: The would not work since we only want to allow one ship per type

lib/engine/game/g_18_norway/steps/buy_train.rb
patrikolesen: Since we have updated the depot to ignore ships we get a fault when the corporation have a train and enough money to by a ship but not the next train.
E.g. the corp has a 2 train and 150 then the corp should be able to buy a S3 ship

lib/engine/game/g_18_norway/steps/buy_train.rb
patrikolesen: Corrected

lib/engine/game/g_18_norway/steps/buy_train.rb
patrikolesen: I also rearranged a bit so the addition is only done in this case

ollybh:
@patrikolesen What's the problem that you're trying to fix with this? I'm struggling to follow the logic of the changes in `buyable_trains`…

patrikolesen:
here are multiple things that I try to achieve in the buy train step

1) Ships does not count towards needing a train so Depot::min_depot_train in depot.rb will only return a train
```
        def min_depot_train
          # Ships doesn't count towards needing a train, should be ignored when checking for min
          depot_trains.reject { |train| @game.ship?(train) }.min_by(&:price)
        end
```

2) The framework will in the case that the entity have less money than the cheapest train only return the min_depot_train.
However there exist a coner case where for example when the entity has 150kr and next available train is a 3 (180kr)

  - Then entity already has 2 so it is not forced to buy a train
  - The entitry have less money than the next available 150kr < 180kr the frame work will return only a 3
  - There is still S3 ships available for 150kr

In this specific corner case we need to add the S3 ship to the available trains

3) We are only allowed to own one ship of each type

lib/engine/game/g_18_norway/steps/buy_train.rb
```diff
@@ -41,11 +41,25 @@ def must_buy_train?(entity)
...
+                next if trains.any? { |ship| ship.name == train.name }        # Must not already be in the list
+
+                entity.cash >= train.price                                    # Can we afford it?
+              end
+            end
+
+            # Remove any ships that we already have
+            trains = trains.reject do |train|
+              next unless @game.ship?(train)                            # We want to keep all trains
+              next true if must_buy_train?(entity)                      # Remove all ships if we must buy a train
```
ollybh, line 60: Is this correct? The rules say that if a company does not own a train it must buy one before the end of its operating turn, but does that stop a company from buying a ship before it buys a train? 8.11 says that the company must buy the cheapest train if it is in emergency money raising.

This might be relevant if a company has no train and enough money to buy either the cheapest train or the cheapest ship, but not both. Could the company buy the ship, and then go into emergency money raising and buy the train?

ollybh (commented):
As you've already got a custom depot class, you might be able to simplify this code by moving some of the logic there. You could add methods for finding the cheapest ship, or all ships buyable by a corporation.

lib/engine/game/g_18_norway/steps/buy_train.rb
patrikolesen: No a company my not first buy a ship and then do an emergency by train

patrikolesen: Moved to depot instead

lib/engine/game/g_18_norway/depot.rb
```diff
@@ -7,6 +7,9 @@ module Game
     module G18Norway
       class Depot < Engine::Depot
         def min_depot_train
+          current_entity = @game.round.current_entity
+          return super unless current_entity.trains.empty?
```
ollybh, line 11: If this is called when the current entity owns a ship but not a train, then this is going to return both ships and trains. Is that right?

patrikolesen: Well spotted no that is not correct, will update shortly

patrikolesen: Updated

ollybh (approved):

PR #11645, opened 2025-03-25
[1841] prevent new token buy for frozen secession corps

### LINKED ISSUE #11310: [1841] Frozen corporation offered opportunity to buy token
https://18xx.games/game/182442?action=341

Screenshot 2024-10-30 at 10 26 35

https://18xx.games/game/182442?action=341#entities

Screenshot 2024-10-30 at 10 26 03

SFL is frozen because nobody had two shares after the exchange in the Ferdinandea Secession. I should not be offered the opportunity to buy extra tokens.

Note that choosing zero tokens works, while choosing one token fails.

philcampeau:
Fixes #11310 

## Before clicking "Create"

### Explanation of Change

I noticed in TheCat's report, their screenshot showed that while the text indicated that he was buying tokens for SFL, the entity shown below was actually his player box. 

There might be a "better" fix for this, but I added a gate to the buy_new_tokens.rb actions to return [] if entity isn't a corporation, and this resolves the issue. As far as I can tell, there's no reason why a non-corporation entity would be buying new tokens, so this seems logical.

### DIFF
--- a/lib/engine/game/g_1841/step/buy_new_tokens.rb
+++ b/lib/engine/game/g_1841/step/buy_new_tokens.rb
@@ -12,6 +12,7 @@ module Engine
 
           def actions(entity)
             return [] unless entity == pending_entity
+            return [] unless entity.corporation?
 
             %w[choose]
           end


crericha:
Is the frozen corp supposed to get a certain amount of tokens? Is that logic handled somewhere?

lib/engine/game/g_1841/step/buy_new_tokens.rb
```diff
@@ -12,6 +12,7 @@ class BuyNewTokens < Engine::Step::Base
 
           def actions(entity)
             return [] unless entity == pending_entity
+            return [] unless entity.corporation?
```
crericha, line 15: This isn't the root of the issue, so probably not the best fix. The root is that the corporation gets set to `pending_entity`.

PR #11646, opened 2025-03-25
[core] fixes Restart Last Turn after waterfall auction

### LINKED ISSUE #6605: [18MEX] Restart Last Turn during auction restarts the whole waterfall
Game ID: 63597 (link: https://18xx.games/game/63597)

During the auction, all privates except Mexico City-Acapulco Railroad had bidders. I purchased Mexico City-Acapulco Railroad, which triggered the cascade of individual-private auctions. We completed all auctions and, after winning Mexican National Railroad, I set NdM's par. I decided I wanted to undo and continue bidding on Mexican International Railroad, so I clicked "Restart Last Turn", which reverted the game to, apparently, the previous action, which was my purchase of Mexico City-Acapulco Railroad.

I'm not sure that it matters, since I think this is server-side, but I'm on Chrome 96.0.4664.93 (Official Build) (64-bit) (cohort: Stable)

philcampeau:
Fixes #6605

## Before clicking "Create"

### Explanation of Change

In the comments, Per pointed out that @crericha had made this comment in the Slack chat: 

image

Thankfully that advice was documented, because sure enough, adding `@game.next_turn!` to the end of the `def resolve_bids_for_company(company)` method in the WaterfallAuction step resolved the issue.

### DIFF
--- a/lib/engine/step/waterfall_auction.rb
+++ b/lib/engine/step/waterfall_auction.rb
@@ -121,6 +121,7 @@ module Engine
           @log << "#{@auctioning.name} goes up for auction" if is_new_auction
         end
 
+        @game.next_turn!
         resolved
       end
 
--- a/lib/engine/step/waterfall_auction.rb
+++ b/lib/engine/step/waterfall_auction.rb
@@ -32,6 +32,7 @@ module Engine
           all_passed! if entities.all?(&:passed?)
           @round.next_entity_index!
         end
+        @game.next_turn!
       end
 
       def process_bid(action)
@@ -43,6 +44,7 @@ module Engine
           placement_bid(action)
           @round.next_entity_index!
         end
+        @game.next_turn!
       end
 
       def active_entities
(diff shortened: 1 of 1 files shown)


lib/engine/step/waterfall_auction.rb
```diff
@@ -121,6 +121,7 @@ def resolve_bids_for_company(company)
           @log << "#{@auctioning.name} goes up for auction" if is_new_auction
         end
 
+        @game.next_turn!
```
crericha, line 124: I'm not sure this catches all the edge cases. For example, I don't think it catches a player who passes when there is nothing actively being auctioned. Would adding `@game.next_turn!` to `process_pass` and `process_bid` bid more robust?

ollybh:
Might this end up breaking a bunch of existing games? If this changes how much is rewound with an undo action then that means games with undos in these auctions might need pinning. It would be worth testing how much this will affect before this is merged/deployed.

philcampeau:
@ollybh you're definitely not wrong. is it worth the effort, or just leave it as is and close the bug report?

PR #11667, opened 2025-04-06
[18Norway] Correcte minor map issue in Kristiansand

patrikolesen:
Correcte minor map issue in Kristiansand

## Before clicking "Create"


ollybh (approved):
FYI: if you want to draw track which doesn't affect upgrades, you can use `track:future` (see 1858, Harzbahn 1873 and 1840 for examples of this being used).

patrikolesen:
Good to know, in this case it was seen as the better solution by the designers.
Note that it did not work to just add the track segment to the next level, will look at that when time allows

PR #11673, opened 2025-04-07
[1862] Fix autopathing issues

PistonPercy:
In 1862, the autopather can't find paths which don't include the home token. This is because the check for the home token is in `check_overlap` when it should be in `check_other`. `check_overlap` is intended to be used to make sure trains don't reuse tracks whereas `check_other` is for checks that require the full set of routes to check the validity of. The key is that the autopather doesn't call `check_other` when calculating the set of all possible routes per train.

## Before clicking "Create"

### Explanation of Change

The main key is the autorouter suppresses `check_other` when looking for valid routes. If we move this code from `check_overlap` into `check_other`, then the autorouter can find routes that don't include the home token. I also moved the other two over because there isn't a reason to check them for single routes so we get a little bit of autorouter performance to skip the work.

This is part of a stack of changes with more information here:
https://github.com/PistonPercy/18xx/pull/1

### DIFF
--- a/lib/engine/auto_router.rb
+++ b/lib/engine/auto_router.rb
@@ -151,7 +151,9 @@ module Engine
               bitfield: bitfield_from_connection(connection, hexside_bits),
             )
             route.routes = [route]
-            route.revenue(suppress_check_other: true) # defer route-collection checks til later
+
+            # defer route combination checks til later
+            route.revenue(suppress_check_other: true, suppress_check_route_combination: true)
             train_routes[train] << route
           rescue RouteTooLong
             # ignore for this train, and abort walking this path if ignored for all trains
@@ -431,6 +433,7 @@ module Engine
         try {
           for (let rt of cb.routes) {
             rt.route['$check_other!']() // throws if bad combo
+            rt.route['$check_route_combination!']() // throws if bad combo
           }
           return true
         }
--- a/lib/engine/auto_router.rb
+++ b/lib/engine/auto_router.rb
@@ -151,9 +151,8 @@ module Engine
               bitfield: bitfield_from_connection(connection, hexside_bits),
             )
             route.routes = [route]
-
-            # defer route combination checks til later
-            route.revenue(suppress_check_other: true, suppress_check_route_combination: true)
+            # defer route combination checks until we have the full combination of routes to check
+            route.revenue(suppress_check_route_combination: true)
             train_routes[train] << route
           rescue RouteTooLong
             # ignore for this train, and abort walking this path if ignored for all trains
@@ -419,20 +418,19 @@ module Engine
 
     %x{
       // do final combo validation using game-specific checks driven by
-      // route.check_other! that was skipped when building routes
+      // route.check_route_combination!
       function is_valid_combo(cb) {
-        // temporarily marshall back to opal since we need to call the opal route.check_other!
+        // temporarily marshall back to opal since we need to call the opal route.check_route_combination!
         let rb_rts = []
         for (let rt of cb.routes) {
-          rt.route['$routes='](rb_rts) // allows route.check_other! to process all routes
+          rt.route['$routes='](rb_rts) // allows route.check_route_combination! to process all routes
           rb_rts['$<<'](rt.route)
         }
 
-        // Run route.check_other! for the full combo, to see if game- and action-specific rules are followed.
+        // Run check_route_combination
         // Eg. 1870 destination runs should reject combos that don't have a route from home to destination city
         try {
           for (let rt of cb.routes) {
-            rt.route['$check_other!']() // throws if bad combo
             rt.route['$check_route_combination!']() // throws if bad combo
           }
           return true
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -1482,7 +1482,11 @@ module Engine
         end
       end
 
+      # Intended for game specific rules that only depend on one route
       def check_other(_route); end
+
+      # Intended for game specific rules that depend on multiple routes
+      def check_route_combination(_routes); end
 
       def compute_stops(route, train = nil)
         train ||= route.train
--- a/lib/engine/game/g_1860/game.rb
+++ b/lib/engine/game/g_1860/game.rb
@@ -996,8 +996,11 @@ module Engine
 
         def check_other(route)
           check_hex_reentry(route) unless @optional_rules&.include?(:re_enter_hexes)
-          check_home_token(current_entity, route.routes) unless route.routes.empty?
-          check_intersection(route.routes) unless route.routes.empty?
+        end
+
+        def check_route_combination(routes)
+          check_home_token(current_entity, routes) unless routes.empty?
+          check_intersection(routes) unless routes.empty?
         end
 
         # must stop at all towns on route or must maximize revenue
--- a/lib/engine/game/g_1862/game.rb
+++ b/lib/engine/game/g_1862/game.rb
@@ -1506,19 +1506,9 @@ module Engine
         end
 
         # Can reuse track between routes, but not within a route, so this method
-        # doesn't check track reuse, but instead checks:
-        # - home token requirement
-        # - route intersection requirement
-        # - freight track end-to-end requirements
-        def check_overlap(routes)
-          return if routes.empty?
+        # doesn't check track reuse
+        def check_overlap(routes); end
 
-          check_home_token(current_entity, routes)
-          check_intersection(routes)
-          check_freight_intersections(routes)
-        end
-
-        # This checks track reuse within a route
         def check_overlap_single(route)
           tracks = []
 
@@ -1592,6 +1582,16 @@ module Engine
 
         def check_other(route)
           check_overlap_single(route)
+        end
+
+        # check for:
+        # - home token requirement
+        # - route intersection requirement
+        # - freight track end-to-end requirements
+        def check_route_combination(routes)
+          check_home_token(current_entity, routes)
+          check_intersection(routes)
+          check_freight_intersections(routes)
         end
 
         def stop_on_other_route?(this_route, stop)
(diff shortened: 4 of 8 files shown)


roseundy (approved):
Looks OK to me. I'm assuming you have tested this beyond just running rake.

PistonPercy:
Yeah, I've played a couple 1862 games end to end with it and plenty of hotseat experiments.

ollybh:
I'd thought that `check_other` is *not* for checks that require all train routes. It is only passed a single route as a parameter, and it should be where title-specific rules for a valid route are made. So this is where I would expect that 1862 checks the local trains don't visit off-board areas ­– having that in `check_distance` doesn't make sense. Most other games seem to be using `check_other` like this.

But several games seem to be abusing this, and are looking at `route.routes` to check the combination of routes. And the auto-router is suppressing the calls to `check_other` for these multi-route checks. From the git history it looks like this was done for 1870 destination run checks.

If you're looking to improve the auto-router, it would be good to have these separated out: leave `check_other(route)` as the checks for a single route, and add something else (`check_route_combination(routes)` maybe) where any checks that need to examine more than a single route should go. By suppressing `check_other` the auto-router is going to be considering routes that are never going to be legal and so should be dropped.

ollybh:
If you want a fun game to test the auto-router with, have a look at tobymao#11252. That has a link to a game where the auto-router takes more than ten hours to reach the first timeout check. (I drafted a fix that made the auto-router time out more quickly in tobymao#11259 but never got round to properly testing and finishing that off).

roseundy:
I used check_overlap() specifically _because_ it was passed all routes. And because the typical purpose of overlap checking doesn't apply to 1862.

Since the home token check _does_ need to be looked at within the context of all routes, I put it in check_overlap(). Essentially, I treated check_overlap() like the check_route_combination(routes) that ollybh suggested since it has no purpose otherwise for 1862. Perhaps I was a bit too timid at the time to make yet another change to the Route code.

I'm not sure why I put those other checks in check_distance, other than the fact it is being passed visits and check_other is not.

Now that ollybh has mentioned it, I agree that check_other shouldn't be used to check that the set of routes is valid.

PistonPercy:
I'm happy to create a `check_route_combination` method, since it makes sense to me. But it'd be a bit of work to refactor to and test, should I do it? 

While we are talking about route's design, how unreasonable (although probably not in this stack of diffs) would it be to remove routes as a member? It was super counterintuitive to me and since y'all are referring to it as abuse to read the routes member, it seems like it shouldn't be there.

roseundy:
You're proposing to remove the routes member from the route class? That won't work. The routes member is absolutely needed.

ollybh:
> I'm happy to create a `check_route_combination` method, since it makes sense to me. But it'd be a bit of work to refactor to and test, should I do it?

Yes, please do this. Looking through the various implementations of `check_other`, I don't think this would be too big of a change.
 
> While we are talking about route's design, how unreasonable (although probably not in this stack of diffs) would it be to remove routes as a member? It was super counterintuitive to me and since y'all are referring to it as abuse to read the routes member, it seems like it shouldn't be there.

I agree that this is a bit odd, but this would be a much more invasive change. There are quite a few games where `route.routes` is used in calculation of bonuses, for games like 1822 where it needs to make sure that a destination bonus is not given to multiple trains.

roseundy:
I'm at a loss as to how games like 1860 and 1862 - which need to check legality across all routes - could validate a route without access to all routes. Please don't refactor this out.

PistonPercy:
So I think the stack is mostly valid, but wanted to upload and get some eyes + opinions on some weirdness on it before doing more manual testing on each game I tweaked. I've done an autoroute of 1862 + a Harzbahn 1873 (although with the full stack since 1862 autoroute is mostly broken until you add in the other fixes from https://github.com/PistonPercy/18xx/pull/1).

The main weirdness that I don't like after doing the change is that check_route_combination is called on each route in the combination. So in the case of lib/engine/game/g_1873/game.rb, before it would only check the relevant checks for that route, now it checks the full combination creating a bit of O(n^2) behavior. So I think either accepting that bit of weirdness (which I do think could be fixed with some refactoring so we only do full combo checking once per combo) or we pass the route being checked into check_route_combination (which is a little bummer to me). 

The other weird quirk shows up in [1822](https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1822/game.rb#L653) that it accesses the full route combo, but the rule can be checked even with single routes so it is useful for the autorouter to still check it. So perhaps `check_route_combination` is only for rules that require the full combination to validate?

Also this stack has the annoying detail worth making explicit that if I mess up and don't move a check out of check_other into check_route_combination, then the autorouter will no longer see a valid route.

roseundy:
All I ask is that if you must make changes for the auto-router, you keep the changes to the minimum required. Please don't start refactoring my code just because you don't like how I organized it - especially if it's game-specific.

If there are new site-wide changes that require refactoring, so be it. But if you start changing my code because you just don't like how I wrote it, I'm going to become pretty frustrated.

Don't fix what ain't broke. There's plenty of other work that can be done on this site.

PistonPercy:
I'm a little confused on what I'm doing wrong here. The autorouter is broken for 1862. I'm have fixed that on my private server and I'm trying to upstream my changes so everyone can benefit. I uploaded [a very minimal patch](https://github.com/PistonPercy/18xx/pull/1/commits/cf49e63d9ee0414ae31696a3d24b2d80af0b0daa). ollybh requested and you seemed to agree that I should create check_route_combination which I did.

I'm attempting to do the minimal changes that allow the autorouter to work. I'm also attempting to help leave things better. I'm happy to do it however it is desired.

roseundy:
@PistonPercy, I don't think you are doing anything wrong. I was just concerned about some of the comments being made about how 1862 didn't implement some of the route methods "correctly". The auto-router changes you are making are fine with me.

PistonPercy:
Would you please give me guidance on how I can get this change in? The entire goal is that this line

https://github.com/tobymao/18xx/blob/ea449a2647b7cc36f9cdafdbe01da138a0c68b0e/lib/engine/game/g_1862/game.rb#L1516

can not be run by the autorouter's pathing phase. I don't care how we do this, but that line has to be suppressed otherwise the autorouter can not find any routes that don't have the home token.

The easiest way to do that is to only change 1862 to move that line to check_other which the autorouter already suppresses. (this was v1 of my diff)

The medium way to do that is add check_route_combination and also suppress that in the autorouter's pathing phase.

The hardest way to do that is change the autorouter and create check_route_combination and get rid of check_other suppression. (this is the current version). I'm not a fan of how hard it is to test and since it's been so hard for me to get commits into the repo, I'd rather go with a commit I'm sure is correct.

ollybh:
Sorry for the delay in reviewing these changes.

This looks OK to me. There are some bits that I think could be done better if you were starting from scratch, but should be left as is to keep the refactoring changes as small as possible.

> The main weirdness that I don't like after doing the change is that check_route_combination is called on each route in the combination. So in the case of lib/engine/game/g_1873/game.rb, before it would only check the relevant checks for that route, now it checks the full combination creating a bit of O(n^2) behavior. So I think either accepting that bit of weirdness (which I do think could be fixed with some refactoring so we only do full combo checking once per combo) or we pass the route being checked into check_route_combination (which is a little bummer to me).

Yes, you could make an argument for moving the call to `check_route_combination`. But I don't think this will be too big an issue, we're not dealing with huge numbers of routes. That can be done later if needed.

> The other weird quirk shows up in [1822](https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1822/game.rb#L653) that it accesses the full route combo, but the rule can be checked even with single routes so it is useful for the autorouter to still check it. So perhaps `check_route_combination` is only for rules that require the full combination to validate?

Does this check in 1822 not need to be able to see all of the routes? It is checking that you aren't running two local trains in the same city.

> Also this stack has the annoying detail worth making explicit that if I mess up and don't move a check out of check_other into check_route_combination, then the autorouter will no longer see a valid route.

I can't see anything that you have missed.

ollybh (approved):

PistonPercy:
> Does this check in 1822 not need to be able to see all of the routes?

The crux of the autorouter's pathing phase is it needs to disable rules which will incorrectly throw an exception for potentially valid routes. In the 1822 case, it won't throw for potentially valid routes, so it is fine from the autorouter pov to be in either check_other or check_route_combination and it might be beneficial for it to be in one or the other depending on whether it can never invalidate a single route as a performance optimization.


In other news, I did some manual testing of the different games I tested with just this set of changes. I didn't see any thing wrong, so I think we are good for this to go in now?

1860 - Tested a few autoroutes, the autorouter was same before and after this change. (It has the bug the autorouter can’t see routes that go through cities)

1862 - Tested a few autoroutes, the autorouter is better or same as before. As an example, in this game, https://18xx.games/game/206941?action=748 I see this route on my server:
Screenshot 2025-05-02 at 7 30 57 AM
(note that the red route doesn't go through the home token which is what this whole stack is about)


1870 - Tested a few autoroutes, the same results as before.
1873 - Tested a few autoroutes, the same results as before.
21Moon - Tested a few autoroutes, the same results as before.

ollybh:
> > Does this check in 1822 not need to be able to see all of the routes?
> 
> The crux of the autorouter's pathing phase is it needs to disable rules which will incorrectly throw an exception for potentially valid routes. In the 1822 case, it won't throw for potentially valid routes, so it is fine from the autorouter pov to be in either check_other or check_route_combination and it might be beneficial for it to be in one or the other depending on whether it can never invalidate a single route as a performance optimization.

OK, I see. In 1822 you have routes that look legal that might need to be excluded depending on the other routes that are being run, whereas in 1862 (and the other games) you have routes that look illegal but might become legal depending on the other routes.

> In other news, I did some manual testing of the different games I tested with just this set of changes. I didn't see any thing wrong, so I think we are good for this to go in now?

Thanks for doing that testing. I'll merge this PR now.

PR #11676, opened 2025-04-09
[18ESP] more bug fixes + add bank card custom info

bentziaxl:
fix mza bug when brown
add bank info on tenders, fix bug where there one less tender in bank


ollybh (approved):
A more flexible version of `other_bank_info` would be to return a hash rather than an array. This could be used if a game needs to add several rows to the summary table, on each the hash key could be printed on the left and the value on the right.

PR #11681, opened 2025-04-14
[1849] disable variants for now

philcampeau:
I haven't been able to identify the issue yet, and I don't want players starting new games that are inevitably going to hit the Bonds-related Blocking Step error.

For now I'm going to leave the code intact on the site and just disable the options to enable the variants in the meta page, but if we're unable to resolve the issue, I'll submit a pull request to remove all the code related to these variants. 

I imagine any games using those variants will need to be pinned or possibly archived.


philcampeau:
Note, I left the Buy tokens variant, since it doesn't cause issues.

philcampeau:
Ollybh may have found the solution to the issue. I'll close this request for now.

PR #11682, opened 2025-04-15
[1849] fixes blocking step error in variants

philcampeau:
Fixes #[11138]

## Before clicking "Create"

### Explanation of Change

@ollybh ran into a similar issue with 1858 Switzerland, which he resolved with this pull request: #11411. After discussion with him, I took the same approach here, and this passes validate_all on a mix of games experiencing the blocking step error and base game files that aren't facing the issue.

I also took the opportunity to improve a method in the game's buy_train step, adding a `.dup` to another method.

I don't believe this will require any pins.

### DIFF
--- a/lib/engine/game/g_1849/step/buy_train.rb
+++ b/lib/engine/game/g_1849/step/buy_train.rb
@@ -38,7 +38,7 @@ module Engine
 
           def buyable_trains(entity)
             # Cannot buy E-train without E-token
-            trains_to_buy = super.dup
+            trains_to_buy = super
 
             trains_to_buy = trains_to_buy.reject { |t| t.name == 'E' } unless @game.e_token?(entity)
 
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1849/step/buy_train.rb
```diff
@@ -38,7 +38,7 @@ def can_sell?(entity, bundle)
 
           def buyable_trains(entity)
             # Cannot buy E-train without E-token
-            trains_to_buy = super
+            trains_to_buy = super.dup
```
ollybh, line 41: There's no point doing this. Both `reject` and `uniq` will be making copies of the array, nothing in the original array is altered.

ollybh (approved):

lib/engine/game/g_1849/step/buy_train.rb
philcampeau: ah, of course. Fixed!

PR #11691, opened 2025-04-17
[1880] Add log message after SR when corp can't buy train

### LINKED ISSUE #8808: [1880] Add log message if a company can't afford a train after the SR.
https://18xx.games/game/111675?action=233

SR2 was triggered when no companies had bought a 2 train for an entire round.  The remaining 2 trains were removed and the SR started.   At the end of the SR, play presumably returned to the train buying phase of the currently acting company (the BCR).   As the BCR couldn't afford a train, play moved to the next company.  It looked to me like a bug, because play seemed not to have returned to the active company.  A log message would explain the flow of play.
E.g. "BCR passes buying train (not enough cash)".

philcampeau:
Fixes #8808

## Before clicking "Create"

### Explanation of Change

Initially I was going to add this log message only if the corp was coming out of an SR, but I wasn't sure that I was really limiting it to that scenario. I realised though, these added log messages works fine even if the corp isn't coming out of an SR, so I simplified the code. 

I don't even think `super` is needed at this point, but I figured I'd leave it as a catch-all

### DIFF
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -33,6 +33,10 @@ module Engine
 
           def log_skip(entity)
             return if entity.minor?
+            return @log << "#{entity.name} is at train limit and cannot buy a train" if !room?(entity) &&
+                                                                  @game.saved_or_round&.round_num == @round.round_num
+            return @log << "#{entity.name} cannot afford a train" if !can_buy_train?(entity) &&
+                                                                      @game.saved_or_round&.round_num == @round.round_num
 
             super
           end
--- a/lib/engine/game/g_1880/step/buy_train.rb
+++ b/lib/engine/game/g_1880/step/buy_train.rb
@@ -33,8 +33,21 @@ module Engine
 
           def log_skip(entity)
             return if entity.minor?
+
             return @log << "#{entity.name} is at train limit and cannot buy a train" unless room?(entity)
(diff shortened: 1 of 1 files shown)


ollybh (commented):
Does this work if a company had bought the last train of a type, triggering the stock round? I suspect that `log_skip` might not be called in that case.

ollybh:
This is going to produce inconsistent logging. For example if a company buys a train from a friend and hits train limit then you may or may not see the log message about being at train limit, depending on which is the next train in the depot. And, IMHO, we don't generally need to be adding more to the log, it's noisy enough as is.

I would have thought that a better place to try and add the log message would be before the `start_operating` line in `G1880::Round::Operating.setup`. That block of code is only being executed when an operating round is being resumed after a stock round, you could check if `current` is at train limit.
https://github.com/tobymao/18xx/blob/ea449a2647b7cc36f9cdafdbe01da138a0c68b0e/lib/engine/game/g_1880/round/operating.rb#L10-L19

philcampeau:
Wow. Not sure how I missed that! I'll agree that's definitely the right place for it, and significantly easier too.

ollybh (changes requested):
Setting this PR to changes requested, so it doesn't get merged until it has been updated.

ollybh:
@philcampeau Any updates for this PR?

ollybh:
ping @philcampeau

philcampeau:
Sorry @ollybh, I looked at this a couple of times and wasn't able to get it working right. I'll close for now and resubmit when i get it right. Thanks for the reminder.

PR #11711, opened 2025-04-28
1837: fix for multiple train discards #11705

### LINKED ISSUE #11705: [1837] Companies are not given full choice of what train to discard on phase change
In this game: https://18xx.games/game/199400
Action: https://18xx.games/game/199400?action=827

TR has just bought the 5 train, dropping the train limit to 2/3.  Both TR and BK are over the train limit and need to discard.  The interface only offers them the option of discarding the 2G trains, but BK has 4 trains and only 1 2G, it needs to discard another train but that option is not provided.

Even if BK discards the 2G, the button stays there and you can click it infinitely to keep discarding a non-existent 2G train.  This is blocking the game from progressing

Browser: Chrome: 135.0.7049.96

timpelican:
Fixes #11705 

## Before clicking "Create"

## Implementation Notes

`@round.merged_trains[action.entity].delete(action.train)` is a silent no-op (returns `nil`, but we're not using the return value) if either:
- `action.train` (the discarded train) is not a member of `@round.merged_trains[action.entity]` (the list of merged trains)
-  The list is empty.
 
It can be safely called every time we discard a train.

### Explanation of Change

Phase change which both lowers the train limit and merges coal companies can leave a corporation needing to discard multiple trains to meet the new train limit.  This could be trains from both the merged coal companies, which must be discarded first, and also from the corporation's trains that were already owned before the merge.

The existing code chooses either the list of 'merged' trains, if non-empty, or the list of already-owned trains, and tries to satisfy ALL of the required discards from that list.  If more trains must be discarded than are in the 'merged' list, the game stalls.  It does not empty the list of 'merged' trains until after the corporation is back down to the train limit.

By removing each train from the list of 'merged' trains as it is discarded, we can get to an empty list of 'merged' trains during the discard step, which triggers the logic to show the list of already-owned trains to complete the discard process.

### DIFF
--- a/lib/engine/game/g_1837/step/discard_train.rb
+++ b/lib/engine/game/g_1837/step/discard_train.rb
@@ -9,6 +9,7 @@ module Engine
         class DiscardTrain < Engine::Step::DiscardTrain
           def process_discard_train(action)
             super
+            @round.merged_trains[action.entity].delete(action.train)
             return if crowded_corps.include?(action.entity)
 
             @round.merged_trains[action.entity].clear


crericha:
Thanks for this; however, I want to go in a little more involved direction. 

I want to see if the corp is train tight or over the limit without merged merged trains and if so discard the train during the merger and include a log message saying why.

If you'd like to give that a try, let me know. Otherwise, I'll try to get to it next week.

ollybh (changes requested):
Setting this PR to changes requested, so it doesn't get merged until it has been updated.

ollybh:
I'm closing this, @crericha has created PR #11720 to fix this issue.

PR #11713, opened 2025-05-03
[1848] Fix lowest bid price for P6

### LINKED ISSUE #11712: [1848] North Australian Railway lowest price not correct
**What's your Game ID?**

https://18xx.games/game/206586?action=7

**How to Reproduce the Bug?**

Since the recent addition of Lowest Price, P6 can no longer go down to $200 like it is supposed to. Currently can only go down to $220. All privates should be able to be discounted 6 times for $5 each for a maximum of $30.

**Which Browser / Platform?**

ollybh:
All privates should have a lowest bid price that is £30 less than their face value.

There's a very small chance this might require pinning or migrating some games.

Fixes #11712.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1848/entities.rb
+++ b/lib/engine/game/g_1848/entities.rb
@@ -119,7 +119,7 @@ module Engine
             sym: 'P6',
             name: 'North Australian Railway',
             value: 230,
-            lowest_bid_price: 220,
+            lowest_bid_price: 200,
             revenue: 30,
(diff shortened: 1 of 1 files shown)


ollybh:
This is fixing a mistake in #11630, but that PR was reverted by #11704 so this no longer applies.

@philcampeau when you redo this change, can you make sure you update the lowest price for P6?

PR #11720, opened 2025-05-06
[1837] Handle edge case where merger happens when corporation is at or over the train limit

crericha:


lib/engine/game/g_1837/game.rb
```diff
@@ -512,14 +512,20 @@ def merge_minor!(minor, corporation, allow_president_change: true)
...
+            @log << "#{corporation.name} receives #{format_currency(minor.cash)} from #{minor.name}'s treasury"
             minor.spend(minor.cash, corporation)
           end
 
           unless minor.trains.empty?
-            @log << "#{corporation.name} receives #{minor.trains.map(&:name).join(', ')} train#{minor.trains.size > 1 ? 's' : ''}"
-            @round.merged_trains[corporation].concat(minor.trains)
-            minor.trains.dup.each { |t| buy_train(corporation, t, :free) }
+            trains_str = "#{minor.trains.map(&:name).join(', ')} train#{minor.trains.size > 1 ? 's' : ''}"
+            if @round.merged_trains[corporation].empty? && corporation.trains.size >= train_limit(corporation)
```
ollybh, line 521: This is highly unlikely, but it would be possible for Ug to form in phase 4, when each of UG1, UG2 and UG3 have two trains. Ug is going to receive four trains from the first two minors, and then the two trains from the last minor will be discarded as its train limit is four. It should receive all six trains and the president gets to decide which to keep.

ollybh: And I think you could get the same problem at the start of phase 5 if two coal companies with trains fold into the same major company.

crericha:
I didn't see this PR. Will look this weekend.

DestrinStorm:
Anyone able to review/merge this? It’s been a few weeks

lib/engine/game/g_1837/game.rb
crericha: @ollybh when UG1 merges, won't the else case be hit and add UG1's trains to merged_trains, which would then for UG2 and UG3 have the if case evaluate to false? All 6 trains should end up in merged_trains and the player will get to pick which to choose.

crericha:
> I believe that this is going to introduce other potential bugs, when a major or national is receiving trains from multiple minors.
> 
> Could #11705 be fixed by having an extra discard train step? When a phase 5 starts any companies over the limit must discard trains, and then you have the closure of coal companies and minors, and the formation of the Ug, and then the second round of discards if companies have received too many trains.

@ollybh I responded to your comment about a potential bug. I believe there is no bug; however, please point out if I'm incorrect in my thinking.

Why have an extra discard step if you have to discard everything? I only want to prompt the player for a choice when they actually have a choice to make.

lib/engine/game/g_1837/game.rb
ollybh: Got it. I hadn't picked up on the `merged_trains[corporation].empty?` part of the test.

PR #11721, opened 2025-05-07
[18esp] fix mz mza cheater token

bentziaxl:
MZA and MZ share a single slot city in madrid (f24).
When mz is taken over, and MZA was placed as a cheater token, the closing of MZ doesn't remove the extra slot.
added minor code to fix this behaviour.

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -777,11 +777,9 @@ module Engine
           hexes = route.hexes.reject { |h| self.class::DOUBLE_HEX.include?(h.name) }
           raise GameError, 'Route visits same hex twice' if hexes.size != hexes.uniq.size
 
-          if route.train.id == '2P-0' && !@perm2_ran_aranjuez && route.hexes.none? do |hex|
-               hex.id == ARANJUEZ_HEX
-             end
+          if route.train.id == '2P-0' && !@perm2_ran_aranjuez && !p2_route_runs_aranjuez_to_mardid?(route)
             raise GameError,
-                  '2P first run must include Aranjuez'
+                  '2P first run must include Aranjuez and Madrid'
           end
           wrong_track = skip_route_track_type(route.train)
           raise GameError, 'Routes must use correct gauage' if wrong_track && route.paths.any? { |p| p.track == wrong_track }
@@ -792,11 +790,14 @@ module Engine
         def check_p2_aranjuez(routes)
           return if @perm2_ran_aranjuez
 
-          @perm2_ran_aranjuez = true if routes.any? do |route|
-                                          route.train.id == '2P-0' && route.hexes.any? do |hex|
-                                            hex.id == ARANJUEZ_HEX
-                                          end
-                                        end
+          @perm2_ran_aranjuez = routes.any? do |route|
+            p2_route_runs_aranjuez_to_mardid?(route)
+          end
+        end
+
+        def p2_route_runs_aranjuez_to_mardid?(route)
+          route.train.id == '2P-0' &&
+          ([ARANJUEZ_HEX, MADRID_HEX] - route.hexes.map(&:id)).empty?
         end
 
         def valid_interchange?(tile, entity)
@@ -885,7 +886,7 @@ module Engine
             hex = hex_by_id(corporation.coordinates)
             cities = hex.tile.cities
             city =
-              if cities.one? || !corporation_by_id('MZ')&.ipoed
+              if cities.one?
                 cities.first
               else
                 cities[corporation.city]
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_esp/game.rb
```diff
@@ -1143,6 +1143,20 @@ def start_merge(corporation, minor, keep_token)
 
           # close corp
           close_corporation(minor)
+
+          # delete mz slot if mz
+          fix_cheater_slot_madrid(minor.city) if minor.id == 'MZ'
+        end
+
+        def fix_cheater_slot_madrid(slot)
```
ollybh, line 1151: Calling this variable `slot` is confusing. It is the index of a city that is being passed to this method.

But I don't think that passing the city index will be reliable. MZ has a city index of 2, which is correct when the Madrid tile is yellow or green, but could MZ close after Madrid has been upgraded to brown, and there is a single city on the tile?

lib/engine/game/g_18_esp/game.rb
```diff
@@ -1143,6 +1143,20 @@ def start_merge(corporation, minor, keep_token)
...
+          cities = hex_by_id('F24')&.tile&.cities
+          return unless cities
```
ollybh, line 1153: You don't need to be so defensive. Something is terminally wrong if `hex_by_id('F24').tile.cities` doesn't return an array.

lib/engine/game/g_18_esp/game.rb
```diff
@@ -1143,6 +1143,20 @@ def start_merge(corporation, minor, keep_token)
...
+
+          city = cities[slot]
+          return unless city
+
+          tokens = city.tokens
+          tokens.compact! if tokens.size > city.normal_slots
```
ollybh, line 1159: Would it be possible for MZ to close when the Madrid cities have merged and there are slots with reservations? This could cause the ordering of tokens and reservations to go wrong.

ollybh (changes requested):
I think that this might not work in all cases.

Would it be possible to do something like this, overriding `close_corporation` instead of adding the `fix_cheater_slot_madrid` method?
```
def close_corporation(corporation, quiet: false)
  if corporation.type == :minor
    corporation.placed_tokens.each do |token|
      city = token.city
      remove_slot = city.tokens.any? { |t| t.cheater? }
      city.delete_token!(token, remove_slot: remove_slot)
    end
  end
  super
end
```

ollybh:
> there is no cheater? method, the best i could come up with is city.tokens.size > city.normal_slots as cheater adds an extra token, but not another normal_slots.

My mistake, it should have been `token.cheater` (without a question mark), not `token.cheater?`.

bentziaxl:
found 2 new minor bugs that i'm adding to this:
1- @game.city_by_id('F24-0-2') breaks when the city has been upgraded to green (so obvious!)
added code to determine the city programatically.

2- the rules say that the first run on 2 permanent needs to connect both aranjuez and madrid, not just aranzuez. fixed the logic to reflect that.

lib/engine/game/g_18_esp/game.rb
```diff
@@ -777,11 +777,9 @@ def check_distance(route, visits)
           hexes = route.hexes.reject { |h| self.class::DOUBLE_HEX.include?(h.name) }
           raise GameError, 'Route visits same hex twice' if hexes.size != hexes.uniq.size
 
-          if route.train.id == '2P-0' && !@perm2_ran_aranjuez && route.hexes.none? do |hex|
-               hex.id == ARANJUEZ_HEX
-             end
+          if route.train.id == '2P-0' && !@perm2_ran_aranjuez && !([ARANJUEZ_HEX, MADRID_HEX] - route.hexes.map(&:id)).empty?
```
ollybh, line 780: Most of this test is repeated in `check_p2_aranjuez`, so you could wrap it in a method to avoid the repetition.

lib/engine/game/g_18_esp/step/home_token.rb
```diff
@@ -11,10 +11,15 @@ def auto_actions(entity)
...
 
             [Engine::Action::PlaceToken.new(entity,
-                                            city: @game.city_by_id('F24-0-2'),
+                                            city: get_mz_city(entity),
                                             slot: 0)]
           end
 
+          def get_mz_city(entity)
+            cities = @game.hex_by_id(entity.coordinates).tile.cities
+            entity.city < cities.length ? cities[entity.city + 1] : cities.first
```
ollybh, line 20: Why `cities[entity.city + 1]`? Shouldn't this be `cities[entity.city]`?

lib/engine/game/g_18_esp/game.rb
bentziaxl: fixed, i still need to check for train 2p-0, otherwise i get this error for every train not running this route. unless there is a better way of doing this

PR #11743, opened 2025-05-13
[1847AE] should LFK only pay on on first new train

### LINKED ISSUE #11735: [1847AE] LFK pays dividend for train purchase between companies
Game ID 203876 OR 9.1 Saar https://18xx.games/game/203876?action=452

I understand the rules so, that the LFK pays dividend only when a train is bought from the bank. In the linked turn Saar buys a 6E not from the bank but from M. Nevertheless LFK pays 55 dividend.

philcampeau:
Fixes #11735

## Before clicking "Create"

### Explanation of Change

The rules say that a dividend is paid to the owner of LFK every time the first train has been bought from the Bank, but as implemented, the code was giving the dividend every time the first train was bought from anywhere other than the discards, which means it was also paying out if the first train was a cross buy.

This will require pins. 

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
+++ b/lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
@@ -17,11 +17,9 @@ module Engine
 
           def process_buy_train(action)
             new_train = @depot.upcoming.include?(action.train)
-            print "new train: #{new_train} "
             super
 
             lfk = @game.lfk
-            print "train bought this round: #{@game.train_bought_this_round} "
             return if @game.train_bought_this_round || !lfk.floated? || !new_train
 
             lfk_owner = lfk.owner
--- a/lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
+++ b/lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
@@ -16,11 +16,13 @@ module Engine
           end
 
           def process_buy_train(action)
-            from_discard = @depot.discarded.include?(action.train)
+            new_train = @depot.upcoming.include?(action.train)
+            print "new train: #{new_train} "
             super
 
             lfk = @game.lfk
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
```diff
@@ -16,11 +16,13 @@ def buyable_trains(entity)
           end
 
           def process_buy_train(action)
-            from_discard = @depot.discarded.include?(action.train)
+            new_train = @depot.upcoming.include?(action.train)
+            print "new train: #{new_train} "
```
ollybh, line 20: Remove debugging statement.
```suggestion
```

lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
```diff
@@ -16,11 +16,13 @@ def buyable_trains(entity)
...
             super
 
             lfk = @game.lfk
-            return if @game.train_bought_this_round || !lfk.floated? || from_discard
+            print "train bought this round: #{@game.train_bought_this_round} "
```
ollybh, line 24: Remove debugging statement.
```suggestion
```

ollybh (changes requested):
The code changes in this look good, but you've left some debugging statements in place.

lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
philcampeau: thanks for catching that. fixed!

lib/engine/game/g_1847_ae/step/buy_single_train_of_type.rb
philcampeau: removed

philcampeau: @ollybh done!

Issue #11036, opened 2024-07-15
[1822 family] - Double-credit for destination city when using autorouter


ddaybell:
https://18xx.games/game/168857?action=1240

I didn't notice this until I looked at the same trains running the same route the next OR, and it was for less money.  

But I think when the CMPS' run was computed at the above-linked action (00:29 on 7/15/2024 (PST)), the destination was doubled for both the E and the 5 train.  I'm not sure how that happened.

Don

michaeljb:
Thanks for reporting this.

I was able to reproduce this only by using autoroutin. Each autorouted train getting the destination bonus probably affects all 1822 titles.

I also noticed the E train is getting the Ski Haus bonus, and it shouldn’t.

ddaybell:
This also occurred in 1822MX.  

https://18xx.games/game/171948

I initially used the autorouter, but then I "clear all" the routes and recreate them all manually and it still is double-counting the destination token, once for the E and once for the 5 train.

Curiously, if I totally undo the run trains step (go back to the assign Pullman step) and then do everything manually, it doesn't double-count the destination.  So there's something in the autorouter that is setting some value somewhere, that once set stays set even if you undo the actual train runs (but don't undo the entire run trains step).

fanman80:
This also happened in game 176532 in action https://18xx.games/game/176532?action=1264 and in action https://18xx.games/game/176532?action=1393.  It may be happening elsewhere in the game.

In action 1264 the destination bonus is being applied to two trains and the Paper Mill bonus is also being applied to two trains.
In action 1393 the destination bonus is being applied to two trains, the Paper Mill bonus is being applied to two trains, the ski-haus bonus is being applied to the E train, and it seems that an extra $10 is being applied to the 5P train.

However, in action https://18xx.games/game/176532?action=1332 the revenue appears to be calculated correctly.

fanman80:
This looks related to https://github.com/tobymao/18xx/issues/11036 and https://github.com/tobymao/18xx/issues/11078

michaeljb:
Fixed by [#11745](https://github.com/tobymao/18xx/pull/11745), but reopened #7982

PR #11745, opened 2025-05-13
Reimplement the routing phase of the autorouter

### LINKED ISSUE #11707: 1822+ Autorouter provides lower run value than I can produce manaully
Game ID:  202678

https://18xx.games/game/202678

LYR operating in OR 9.1, at 9:20, with Pullman attached to 4 train, autoran for only 430.  I successfully created and manually a run for 460. 

https://18xx.games/game/202678?action=1116

Image

Used Microsoft Edge.  Version 135.0.3179.85 (Official build) (64-bit)

PistonPercy:
Fixes #11707 and #11036 and probably more. 

## Before clicking "Create"

## Implementation Notes

I wrote up substantially more information in https://github.com/PistonPercy/18xx/pull/1 and in the commit messages, so the short version is replace the routing code with a simple brute force version that checks every combination (this is find_best_combo in the second commit in the pr). Then various little fixes to support it and generally improve little bugs I found while testing it. I also made it give update the UI as it finds valid routes so you can choose to stop the autorouting because in really big maps I've noticed that pretty quickly it gets a "good enough" route combo and something is better than nothing.

### Explanation of Change

The main reason to do this is the old autorouting code had several issues:
- It didn't allow running routes which overlapped (unless the game used TRAIN_AUTOROUTE_GROUPS which introduced #11036 because of how it works). This means 1862 can't effectively autoroute
- It assumed that the revenue of a combo of routes was the simple addition of the three routes. This is often not true such as bonuses often can only apply to one route.
- It assumes an invalid route combo can't become a valid route if another train gets added (e.g. a home token rule).


PistonPercy:
@ollybh / @tobymao Since I've chatted with each of you about these diffs, any advice on next steps here?

tobymao:
@roseundy could you take a look?

tobymao:
@JoeB989

JoeB989:
I'm too far removed from the code to give meaningful detailed feedback, but here are some overall thoughts.

I imagine the performance takes a big hit, but I understand why the change to brute force consider all routes;; too many games allow overlapping. routes.  My original rewrite was based on prior work optimizing an 1856/1870 style game, and I was completely blind to the range of shared-route games in 18xx.games.

Afterward, I had considered injecting a game-specific optimization helper so each game could contribute custom auto-routing rules, but the size of that task was too daunting.  How to optimize deeply but generically...

FYI the original brute force approach was about 1000 times as slow as this optimized algorithm you're replacing.  A factor of 10 was just from writing javascript instead of opal interpreted ruby for the core loops.  I'm not clear how much of the js you kept, but I suggest avoiding ruby/opal for the exponential loops

I'm glad someone's taking this on; I was burnt out on generalizing it.  Good luck!

PistonPercy:
> Afterward, I had considered injecting a game-specific optimization helper so each game could contribute custom auto-routing rules, but the size of that task was too daunting. How to optimize deeply but generically...

Yeah I considered this or exposing better routing rule hooks and somehow migrating games to them (I wish we had this just so we could show the routing rules in the UI). Either path though feels very hard for how hard it has been for me to get changes in. I'm unclear if I can be doing something to make this easier, I have a WIP patch for the pathing part being slow, but I'm concerned if this patch will even be accepted. 

> FYI the original brute force approach was about 1000 times as slow as this optimized algorithm you're replacing. A factor of 10 was just from writing javascript instead of opal interpreted ruby for the core loops. I'm not clear how much of the js you kept, but I suggest avoiding ruby/opal for the exponential loops

Yeah, ruby/opal is filled with performance footguns which really shows up with how hot this loop is. I think I kept most of the core strategies to avoid having to call into ruby (eg the connection_data bitmap) since everything is free compared to calling ruby. 

I think in hindsight, it's a mistake to call it "a brute force" because I have two hooks `is_worth_adding_trains` (can this subset of trains "grow" into a valid set of trains that is better than the current set of trains) and `estimate_revenue` (is this train set possibly valid and possibly better than the current best route combo?). These allow the autoroute to avoid calling into ruby for the vast majority of routes and hopefully one day we can add more things to it as we find performance problems. I called it brute force because it's more clear that in javascript we are checking every single set of trains and passing it to `estimate_revenue` (unless it gets culled by `is_worth_adding_trains`).

roseundy (commented):
I can only comment on the impact to 1862 (not the autorouter itself). Looks fine to me.

PistonPercy:
tl;dr: The performance is pretty much the same (both ~1 second in the routing portion). 1844 + 18USA have the same final routes. 1822 is the one with the interesting asterisk, but performance wise is fine imo. As a note, the performance issues on these maps is purely pathing (finding valid routes per train) and not routing (what this is changing). The ickiest is 1862 (which the old autorouter doesn't support) because it allows overlapping train routes so it requires calling into opal/ruby for revenue more often and performance in this portion of the good is 100% correlated with # of revenue calls. That game is actually what got me to do the live updating in progress of best route found because usually you get good enough in 1 second and can just hit submit. 


1844: https://18xx.games/game/205895?action=940&l=info

Pathing times out (this is finding possible routes per train) for both, but they both get `Evaluated 9718 paths, found 220 unique hexsides, and found valid routes 5H:47, 8E:8313 in: 37.034`.

Old: Found 141080 possible combos (8 best) and rejected 257944 conflicting combos in: 0.052
New: AutoRouter took 26ms
They find the same best combo.

18USA: https://18xx.games/game/210281?action=2065&l=info
Pathing also times out at `Evaluated 21843 paths, found 182 unique hexsides, and found valid routes 8:11088, 5:1720 in: 34.427`.  (note that the pathing only keeps the best 10k combos).

Old: Found 2339420 possible combos (8 best) and rejected 14870580 conflicting combos in: 1.321
New: AutoRouter took 404ms
They find the same best combo.

1822: https://18xx.games/game/210556?action=1144&l=info
So this is where it's slightly tricky because the old one has a slightly different structure so it gives the shorter trains more time to search for paths so it found a better path for them. `TRAIN_AUTOROUTE_GROUPS` splits the E trains and the normal trains into two different groups giving the shorter trains a separate pathing timeout so they were able to path completely by themselves and find better routes. So the old autorouter found a better combo because it had more paths for the shorter trains. IMO, the right thing is to speed up pathing (I have some notes+diffs I can follow this up with that gets over a magnitude win there). It also has better performance because it picks the best E train and then picks the best combo for the short trains. This isn't actually safe to do (it causes this bug: https://github.com/tobymao/18xx/issues/11036), but it does help performance :)

But focused on performance:
Old: 0.004 + 0.006 (seconds)
New: AutoRouter took 634ms

crericha:
Thanks for the testing info. Looks great.

PistonPercy:
Just a tiny update to the stack to change out the logging statement to be a little clearer and use the LOGGER class.

PistonPercy:
Fixed an off by one error that I'm not sure how I never hit before.

PR #11758, opened 2025-05-22
[18Norway] Don't allow selling president's share to NSB

### LINKED ISSUE #11730: [18Norway] Hovedbanen's president may sell out when NSB owns two stocks
When Hovedbanen uses its special power and sells two stock to NSB, it is possible for the president to sell out in the next stock round - see images. 

Before: the player Kelly is the president of Hovedbanen:
Image

After: the president gets to sell their shares even if that shouldn't be possible:

Image

patrikolesen:
Fixes #11730

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/steps/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_norway/steps/buy_sell_par_shares.rb
@@ -12,9 +12,11 @@ module Engine
             # Check if selling would make NSB president
             if bundle.shares.any?(&:president) &&
                   corporation.share_holders[@game.nsb] >= bundle.presidents_share.percent &&
-                  corporation.player_share_holders.reject do |p, _|
-                    p == bundle.owner || p == @game.nsb
-                  end.values.max.to_i < bundle.presidents_share.percent
+                  corporation.player_share_holders.none? do |sh, pct|
+                    next false if sh == bundle.owner
+                    
+                    pct >= bundle.presidents_share.percent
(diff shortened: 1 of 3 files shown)


patrikolesen:
Updated

lib/engine/game/g_18_norway/steps/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,27 @@
...
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParShares
+          def can_dump?(entity, bundle)
+            corporation = bundle.corporation
+            # Check if selling would make NSB president
+            if bundle.shares.any?(&:president) &&
+                  corporation.share_holders[@game.nsb] >= bundle.presidents_share.percent &&
+                  corporation.player_share_holders.reject do |p, _|
+                    p == bundle.owner || p == @game.nsb
+                  end.values.max.to_i < bundle.presidents_share.percent
```
ollybh: I found this test a bit hard to read. And you've got a couple of things that aren't needed: `player_share_holders` won't (or at least shouldn't) include NSB, so you don't need to test for that, and the `to_i` conversion seems redundant.

Is this any clearer?
```suggestion
                  corporation.player_share_holders.none? do |sh, pct|
                    next false if sh == bundle.owner
                    
                    pct >= bundle.presidents_share.percent
                  end
```

ollybh (approved):

PR #11759, opened 2025-05-23
[18Norway] Nationalization rounds should not pay out companies

### LINKED ISSUE #11732: [18Norway] privates pay i Nationalization round
**What's your Game ID?**
207515

**How to Reproduce the Bug?**

**Which Browser / Platform?**

patrikolesen:
Fixes #11732

Will break existing games

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/round/nationalization.rb
+++ b/lib/engine/game/g_18_norway/round/nationalization.rb
@@ -7,6 +7,18 @@ module Engine
     module G18Norway
       module Round
         class Nationalization < Engine::Round::Operating
+          def initialize(game, steps, **opts)
+            super
+            @nationalization_complete = false
+          end
+
+          def process_action(action)
+            super
+            @nationalization_complete = true if action.is_a?(Engine::Action::Choose) && action.choice != 'decline'
+          end
+
+          attr_reader :nationalization_complete
+
           def name
             'Nationalization round'
           end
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_norway/steps/nationalize_corporation.rb
```diff
@@ -12,6 +12,7 @@ def description
           def actions(entity)
             return [] if !entity.corporation? || entity != current_entity
             return [] if @game.nationalized?(entity.corporation)
+            return [] if @game.round.is_a?(G18Norway::Round::Nationalization) && @game.round.nationalization_complete
```
ollybh: This step only exists in nationalisation rounds, so the test for the round type is pointless. And there's a `@round` variable you can use instead of `@game.round`.
```suggestion
            return [] if @round.nationalization_complete
```

lib/engine/game/g_18_norway/round/nationalization.rb
```diff
@@ -7,6 +7,18 @@ module Game
...
       module Round
         class Nationalization < Engine::Round::Operating
+          def initialize(game, steps, **opts)
+            super
+            @nationalization_complete = false
+          end
+
+          def process_action(action)
+            super
+            @nationalization_complete = true if action.is_a?(Engine::Action::Choose) && action.choice != 'decline'
```
ollybh: It would be better to put this in `Step::NationalizeCorporation.process_choose`.

patrikolesen: Moved

ollybh (approved):

PR #11757, opened 2025-05-22
[18Norway] Update nationalization rounds

patrikolesen:
Fixes #11728

Adding nationalization rounds end of phase 5

Will break all games that have triggered end game

## Before clicking "Create"

- [x] Branch is derived from the latest `master`
- [ ] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

PR #11760, opened 2025-05-23
[18Norway] Stock movement should be left block if in bank pool and right block if sold out

### LINKED ISSUE #11731: [18Norway] Stock price doesn't get adjusted at the end of SR
https://18xx.games/game/207600?action=98

Stock prices don't move at the end of stock round as rule section 7.6 describes:

> Then, in stock price order, adjust each company’s stock price one column left if it has any shares in the Bank Pool, and one column right if no shares are in the Bank Pool and no shares are in the company treasury. If a company token is already in the column, put the new token below the one already there, at the bottom of the column.

BB's price did not move up (every share owned), SB's price did not move down (has share in market).

On Chrome Version 136.0.7103.93

patrikolesen:
Fixes #11731

Can impact existing games
## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -37,13 +37,14 @@ module Engine
         EBUY_SELL_MORE_THAN_NEEDED = false
         CAPITALIZATION = :incremental
         MUST_BUY_TRAIN = :always
-        POOL_SHARE_DROP = :none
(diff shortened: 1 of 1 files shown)


ollybh:
I've added the hold-merge label to this PR. There are several other 18Norway changes that will also require archiving games (#11757, #11758, #11759, #11763), this one should be merged at the same time as the others.

PR #11763, opened 2025-05-24
[18Norway] P5 nationalization

### LINKED ISSUE #11742: [18Norway] p5 not implemented properly?
https://18xx.games/game/207586?action=510

V should have the option to nationalize before anyone else but it doesn’t seem to get that option?

patrikolesen:
Fixes #11742

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_norway/round/nationalization.rb
+++ b/lib/engine/game/g_18_norway/round/nationalization.rb
@@ -12,7 +12,11 @@ module Engine
           end
 
           def select_entities
-            @game.operating_order.reverse
+            entities = @game.operating_order.reverse
+            p5_owner = @game.company_by_id('P5')&.owner
+            return entities unless p5_owner&.corporation?
+
+            [p5_owner] + entities.reject { |e| e == p5_owner }
           end
 
           def show_in_history?


patrikolesen:
You are correct from a strict rule perspective. However there is no benefit for the corporation to go in normal operating order.
So to not need to ask the owner twice for each NR which would slow down the online version the proposal is to only offer the nationalization first

ollybh:
OK. But I wouldn't be entirely surprised if someone wants the option to decide later in the round.

ollybh (approved):

patrikolesen:
What would be the easiest way to implement that? E.g. how can I get the same company execute twice but with different order?
Will it not just skip the second instance since the company is already tagged for that round?

ollybh:
> What would be the easiest way to implement that?

This could be done in a similar way to how 1822 handles the private companies that allow their owner to count their cash as doubled when calculating priority order. It has a Choices round that follows the stock round, the player owning the private gets to click a button that decides whether their cash is double-counted or not.

You could have a choices round before the nationalisation round, then calculate the operating order for the nationalisation round based on the choice made.

patrikolesen:
Checked the designers and the company should only get one request as implemented here

PR #11769, opened 2025-05-26
[1848] must sell in blocks -> confirmed by designer

### LINKED ISSUE #11764: [1848] Selling shares bug
**What's your Game ID?**

https://18xx.games/game/207865?action=468

**How to Reproduce the Bug?**

When selling shares, gives option to sell shares from the same company one after the other to cause share price to drop once per share. Rules explicitly state selling shares from the same company causes the stock price to fall once. 

Image

As there is no provision for selling shares from the same company one after the other on the same turn, it should be infered this doesnt cause a price change greater than the one the written rule establishes. 

**Which Browser / Platform?**

philcampeau:
Fixes #11764

## Before clicking "Create"

### Explanation of Change

Currently, players can sell one share of a corp, then another, then another, etc to cause multiple price drops in a single turn.  While the rules as written don't have a specific ruling against this, I know we encountered the same issue with 1844 and Helmut's ruling was that [you must sell all shares as a single block.](https://boardgamegeek.com/thread/3308703/can-players-sell-multiple-shares-in-a-company-one) Given that he's also the designer here, I would assume that holds true. 

I can try reaching out to him if we think a confirmation would be helpful.

### DIFF
--- a/lib/engine/game/g_1848/game.rb
+++ b/lib/engine/game/g_1848/game.rb
@@ -46,7 +46,7 @@ module Engine
 
         CAPITALIZATION = :full
 
-        MUST_SELL_IN_BLOCKS = false
+        MUST_SELL_IN_BLOCKS = true
 
         EBUY_PRES_SWAP = false
 


philcampeau:
Heard back from Helmut, confirmed, must sell in blocks: 

image

ollybh (approved):

PR #11792, opened 2025-06-12
[18esp] usability fixes

bentziaxl:
1- In this game, a subset of major corps are only available in phase 3.
previously the code hid the corps until phase 3, added a change to include them, make them not buyable (via corporation_available?) and add a status to the crop.

2- add end game log event, when custom end game is triggered. (in this case purchase of second 8 train)

3- fix the end game status text at the top. When end game is triggered current OR is done, the another full set of SR and ORs. previously it would say 7.2 (of 3), then game would move to SR 8 as needed. 
rewrote total_rounds to return current round number if game end has triggered, but we're not in the last OR set.

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -576,15 +576,10 @@ module Engine
           if end_game_trigger && !@game_ending
             @log << '-- Event: Endgame triggered --'
             @log << "Finish this OR; Game ends at the end of #{@turn + 1}.#{@operating_rounds}"
+            @operating_rounds = @round.round_num
             @game_ending = true
           end
           end_game_trigger
-        end
-
-        def total_rounds(name)
-          return super if !@game_ending || (@turn == @final_turn)
-
-          @round.round_num if name == self.class::OPERATING_ROUND_NAME
         end
 
         def event_close_companies!


ollybh:
I don't understand the issue you are trying to fix with the end game status. Can you give an example of the incorrect status text, and what it will look like after these changes?

bentziaxl:
Game rules:
if game end is triggered during 7.2, you finish the current OR, you do not go to 7.3, instead go into an SR and than 3 final OR sets.

Before the fix:
Operating Round 7.2 (of 3) - Game ends at the end of 8.3

after the fix:
Operating Round 7.2 (of 2) - Game ends at the end of 8.3

ollybh:
> Game rules:
> if game end is triggered during 7.2, you finish the current OR, you do not go to 7.3, instead go into an SR and than 3 final OR sets.

Could you set `@operating_rounds = 2` and `@final_operating_rounds = 3` when this happens?

bentziaxl:
this could also happen at 7.1 for example, then the game will go to SR8, then 3 ORS.
similarly if it happens at 7.3, then no changes

bentziaxl:
you mean just set operating round to current round number, when the game end is triggered?

ollybh:
Yes. That should mean the set of operating rounds ends after the one in which the game end is triggered, but you still have three operating rounds in the final set.

If this works, it will avoid the need for any custom code for the end game status.

bentziaxl:
added @operating_rounds = @round.round_num to address comment. 
@final_operating_rounds is set in base.rb as part of one_more_full_or_set game end

ollybh (approved):

PR #11802, opened 2025-06-15
[1858 India] Mine, oil and port tokens

ollybh:
This adds mine, oil and port tokens to the 1858 India implementation.

These are similar to the fort tokens in 18Ardennes. They start off on the map in plain hexes. Once track has been laid to that hex a public company can (for a fee) collect the token as part of its operating turn. These tokens then provide a revenue bonus whenever the owning company runs its trains.

This also includes a couple of changes from further playtesting of this game: the 2M train price has been increased, and there are a couple of extra brown city tiles added.

## Before clicking "Create"

- ~Add the `pins` or `archive_alpha_games` label if this change will break existing games


lib/engine/game/g_1858_india/step/collect_tokens.rb
```diff
@@ -0,0 +1,139 @@
...
+
+          # Returns all mine/oil/port tokens on the map which have not yet
+          # been collected, where there has been track laid in the hex.
+          # For port tokens the corporation must be able to afford the token
+          # and the current phase must allow them to be taken.
+          def available_tokens(corporation)
+            tokens = []
+            tokens += mines if corporation.cash >= TOKEN_COST['mine']
+            tokens += oil if corporation.cash >= TOKEN_COST['oil']
+            tokens += ports if corporation.cash >= TOKEN_COST['port']
```
michaeljb: A minor point, `+=` creates a new array, whereas `tokens.concat(oil)` will add the elements of `oil` to the existing `tokens` array. It won't make much difference for very small arrays but is generally more efficient to use `concat()`.

ollybh: OK. I had thought that `tokens.concat(oil)` was the same as `tokens << oil` which doesn't do what I need.

ollybh: I've changed these to use `concat`.

PR #11804, opened 2025-06-16
autorouter: Fix invalid array check

PistonPercy:
## Before clicking "Create"

## Implementation Notes

This was intended to be an array/non-nil check but unfortunately https://github.com/tobymao/18xx/commit/75be0ddd65e101037ac5f1d62b529a67fbbeec5d changed the behavior of how ruby's nil appeared in this code path (before it used to be undefined and now it is NilClass.. which seems to also have an length property :/ ). Switch from a truthy check to an explicit array check.

I tested 1858 (this has no train autogroup, so falls into the else),  1822 with a combo with no E train and a combo with both an E train and no Etrain (this falls into the code path for this if statement) as well as 1862 and ~dozen other games for good measure. Hopefully, this finally fixes this finicky codepath.

### DIFF
--- a/lib/engine/auto_router.rb
+++ b/lib/engine/auto_router.rb
@@ -376,7 +376,7 @@ module Engine
             if (game_group_rules == "each_train_separate") {
               train_group = number_train_groups;
               ++number_train_groups;
-            } else if (game_group_rules && routes.length > 0) {
+            } else if (Array.isArray(game_group_rules) && routes.length > 0) {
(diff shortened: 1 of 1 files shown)


PistonPercy:
Btw if there are any additional issues when this code gets deployed, feel free to ping me on slack, I'm happy to help fix them. Sorry for these little fixups.

ollybh (approved):

PR #11806, opened 2025-06-17
Create a git tag when deploying

michaeljb:
This adds `make tag` and executes it during `make prod_deploy`. It creates a git tag on the deployed commit. The tag name is a human-readable UTC timestamp that uses the same source as the timestamp shown on the `/about` page.

## Before clicking "Create"


ollybh:
@michaeljb Thanks for doing this. Tagging releases is something I've had on my mental to-do list.

michaeljb:
@ollybh I still ran into a couple hiccups on my last deployment, #11812 should sort them out

PR #11810, opened 2025-06-19
fix illegal india ferry tile lay

### LINKED ISSUE #11484: [18 India] Orientation restriction on the ferry tile
**What's your Game ID?** This occurs in every game of 18India at the moment due to being a new rule.

**How to Reproduce the Bug?**

Per errata/FAQ:

Q: Can a Ferry tile be placed in J37 in such a way to block access to Sri Lanka (K38)?
A: No. The tile laid in J37 must be placed so that it connects hex K38 to J35, I36, or I38.

**Which Browser / Platform?**

MatthewDeKoning:
Fixes #11484

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_india/step/track.rb
+++ b/lib/engine/game/g_18_india/step/track.rb
@@ -104,6 +104,12 @@ module Engine
           # Added multple yellow tile check and Yellow OO reservation check
           def process_lay_tile(action)
             if action.tile.color == :yellow
+              #check for invalid ferry placement
+              if action.tile.id == "IF2-0"
+                if action.tile.rotation == 1 && action.hex.id == "J37"
+                  raise GameError, "Invalid Ferry Tile Placement" 
+                end
+              end
               raise GameError, 'New yellow tiles must extend path from railhead and previously laid tiles' \
                unless connected_to_track_laying_path?(action.hex)
 


michaeljb (changes requested):
Style checks failed in https://github.com/tobymao/18xx/actions/runs/15766795192/job/44444699785?pr=11810

Run the rubocop command in the PR description to get the styling changes (the problems look auto-correctable, so the `-a` flag should take care of them for you)

MatthewDeKoning:
fixed those errors, now to add a label to this...

michaeljb:
affected games as of today:

```
181923, 195990, 181202, 182385, 188782, 200143, 210978
```

ollybh:
@MatthewDeKoning This will stop the tile being laid in the incorrect orientation by throwing an error if the player attempts to do this. It would be better if, as well as this check, the player was not given the option of laying the tile the wrong way round. There is a `legal_tile_rotation?` method that is called by the tile selector UI, you can add a check to that which will limit the orientations allowed.

This code in 1837 is an example of how this can be done:
https://github.com/tobymao/18xx/blob/ca0e7b3234052ecd23c23cdb6eed1d74693471ad/lib/engine/game/g_1837/game.rb#L783-L788

PR #11811, opened 2025-06-19
[core] Getting started changes: explicitly set mode for `db/data/`, use `docker compose`

michaeljb:
`umask` should have covered the mode for `db/data/` but apparently it did not work for @MatthewDeKoning when he was getting set up for #11810

Docker Compose v2 uses `docker compose`, and now the compose plugin has been
installed on the server, so local dev environments don't need to use
`docker-compose` just to be compatible with the server.

If a dev is still using the older version, `CONTAINER_COMPOSE=docker-compose`
should still work for most uses.

When merging I will also update the commands on these wiki pages:
- [ ] https://github.com/tobymao/18xx/wiki/Developing-for-18xx.games
- [ ] https://github.com/tobymao/18xx/wiki/Developing-FAQ


michaeljb:
Closing this, combining with some other infra-related changes. See #11812

Issue #11199, opened 2024-09-09
[18RoyalGorge] wrong game end (OR) calculation


rjaey:
[https://18xx.games/game/177497](https://18xx.games/game/177497)

At the end of OR 3.2 the first 6x2 train was exported. The game status message at the top then changed to  "Game Ends at conclusion of OR 4.2". 

![image](https://github.com/user-attachments/assets/aa482f15-22d4-43f0-b7ca-f85ce88c2b6f)

After playing through the stock round it changed to  "Game Ends at conclusion of OR 5.2".

![image](https://github.com/user-attachments/assets/1a8f797c-9fa1-4ffa-bfb7-28e4ee6c3d18)

Actual game end was after 5.2.

michaeljb:
https://18xx.games/game/177497?action=387 - last 5 train is bought, silver phase is triggered

https://18xx.games/game/177497?action=388 - pass action ends the turn, exports the 6x2, "Game Ends at conclusion of OR 4.2" is visible

https://18xx.games/game/177497?action=412 - last action in SR 4, still showing end will be 4.2

https://18xx.games/game/177497?action=413 - start of OR 4.1, now it says the end will be 5.2

The [rulebook (page 16 of 24 of the PDF)](https://boardgamegeek.com/filepage/284918/graphic-rulebook-preliminary) says this:

> Ending 3: Silver Phase is Reached
> 
> If a company buys a train and reveals the silver phase event card, the players finish the current set of operating rounds, complete one more stock round, followed by a final set of operating rounds, and then the game ends. Everyone adds up their scores.

I think the correct behavior for this game would have been ending at the end of OR 4.2

michaeljb:
I've requested clarification from the designer about which round should be the final round.

Regardless of that, these changes are needed:
* the message about which round is final should not change
* the endgame trigger is now the last 5-train rather than the first 6-train

michaeljb:
game data copied to [gist](https://gist.github.com/michaeljb/b36426b7d09f500c9ba83eea6be5fd87)

PR #11818, opened 2025-06-21
[18RoyalGorge] fix endgame trigger, add optional rule for alternate endgame

michaeljb:
* In earlier prototypes, the endgame was triggered on the first 6x2 train, now it
is on the last 5+ train.

    * Also remove `GAME_END_CHECK_STOCK` which was overcomplicating things and not
working correctly anyway, making the "advertised" end of game change values

* Implement variant "Shorter Game End"

* add note that Y4 cannot be sold to a corporation - [BGG thread](https://boardgamegeek.com/thread/3416338/article/45279634#45279634)

Fixes https://github.com/tobymao/18xx/issues/11199

## Before clicking "Create"


ollybh (approved):

michaeljb:
I got an email from the designer that the rules I have are correct and the linked BGG rules are actually what’s out of date. So I’ll need to revert the first 6 vs last 5 change and keep the other parts of this PR. Closing for now until I can address this.

PR #11852, opened 2025-07-05
[18MEX] Correct bug for upgrading F5 for non-470 tile #11462

### LINKED ISSUE #11462: [18MEX] Error when attempting to upgrade non-# 470 yellow tile on hex F5
https://18xx.games/game/188730

---

**Steps to Reproduce**

1) Have a non-# 470 yellow tile on hex F5:

Image

.
.
.
.
.

2) Attempt to upgrade hex F5 to green:

Image

.
.
.
.
.

3) Confirm upgrading hex F5 to green and receive "undefined method 'teardown' for nil" error:

Image

.
.
.
.
.

---

**Browsers Reproduced With**

- Brave 1.74.48 / Chromium 132.0.6834.83 64-bit desktop
- Chrome 131.0.6778.266 64-bit desktop
- Firefox 134.0.1 64-bit desktop
- Vivaldi 7.0.3495.29 (Stable channel) 64-bit desktop

perwestling:
Fixes #11462

- [ ] Add the `pins` or `archive_alpha_games` label if this change will break existing games: N/A

## Implementation Notes

Simple bug fix.

### Explanation of Change

Bug was caused by an attempt to remove the P2 ability when upgrading F5 to green. This had already been done when yellow non-470 tile was put there, so the ability did not remain.

Fix was just to not remove if ability already lost.

### Screenshots

None. Tested manually to do the upgrade of F5, and after bug fix it works.

### Any Assumptions / Hacks

None

### DIFF
--- a/lib/engine/game/g_18_mex/game.rb
+++ b/lib/engine/game/g_18_mex/game.rb
@@ -724,8 +724,13 @@ module Engine
             return unless action.hex.id == 'F5'
             return if p2_company.closed? || action.entity == p2_company
 
-            p2_company.remove_ability(p2_company.all_abilities.first)
-            @log << "#{p2_company.name} loses the ability to lay F5"
+            # Remove p2 ability to lay F5, but if someone else already
+            # have built there the ability has already been lost
+            p2_ability = p2_company.all_abilities.first
+            return if p2_ability.nil?
+
+            p2_company.remove_ability(p2_ability)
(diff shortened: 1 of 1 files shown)


perwestling:
Should not affect any existing games, as it fixes a bug that made it impossible to upgrade F5 from yellow to green. So no game would have such an action.

ollybh (approved):

Issue #11699, opened 2025-04-21
[18India] bug with SPD & NWR home tile placement


jtamplin:
**What's your Game ID?**

https://18xx.games/game/205760?action=100

**How to Reproduce the Bug?**

Start a game with both SPD and NWR floated, try and place the home tile.

After placing the yellow OO tile, selecting the home token slot, you should be able to continue playing yellow tiles that connect.  Instead, the only option is to upgrade the home hex to green (which you shouldn't be able to do normally), but it only shows single-slot green tiles.  After skipping the tile lay with the NWR to let the game continue, SPD has the same problem but even shows NWR placing the token instead of SPD.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/9da46c6b-d54c-4db2-b79f-e09c9c0ef747" />

**Which Browser / Platform?**

Windows 11, Chrome 135.0.7049.43

MatthewDeKoning:
Cloning the game it seems once the yellow is placed on NWR's turn, illegal green options will show up if you click the tile again (should be fixed), or if "the opposite" OO token is clicked, an SPD token appears, and you can finish your track operation as expected. I suspect the fix here would be to automatically assign SPD the non-NWR token. I will see what I can do.

PR #11860, opened 2025-07-08
[18 India] fix OO SPD / NWR both floated #11699

MatthewDeKoning:
Fixes #[#11699]

Potentially clumsy but seems to fix most of the issue. 

"After skipping the tile lay with the NWR to let the game continue, SPD has the same problem but even shows NWR placing the token instead of SPD." - this issue remains, i will keep looking into it. 

I also added a check to avoid displaying invalid green tiles should the player click the tile not the city while the game waits for a token. 

## Before clicking "Create"

This needs an "archive_alpha_games" tag but i am not able to add it. 

## Implementation Notes
Basically just force lays the opposite company's token in the other spot. 

### Explanation of Change
Current implementation that needs double clicks without prompt has confused many users. 

### Any Assumptions / Hacks
If NWR is floated by another company between turns (or NWR passes track actions when both are floated which seems unlikely) SPD places NWR token still. I will make a new issue for this.


MatthewDeKoning:
If this is left open a few days i will try to fix the NWR token on SPD turn issue and add it here

MatthewDeKoning:
fixed!

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,12 +43,39 @@ def process_lay_tile(action)
...
+
             @round.pending_tokens.shift
           end
 
           def process_place_token(action)
             super
             tile = action.city.tile
+            # apply "other" OO token once first to operate is applied if floated
+            if action.entity.name == 'NWR' && tile.name == '235'
+              corporation = @game.corporations.find { |c| c.name == 'SPD' }
```
ollybh: There's a method to find corporations easily:
```suggestion
              corporation = @game.corporation_by_id('SPD')
```

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,12 +43,39 @@ def process_lay_tile(action)
...
+              if corporation&.floated
+                token = corporation.find_token_by_type
```
ollybh:
```suggestion
                token = corporation.next_token
```

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,12 +43,39 @@ def process_lay_tile(action)
...
+                city = tile.cities.reject(&:tokened?).first
+                city.place_token(corporation, token) if city.tokenable?(corporation, tokens: token)
+              end
+              @round.pending_tokens.shift
+            elsif action.entity.name == 'SPD' && tile.name == '235'
+              # remove and swap the NWR that appeared... hacky...
+              city = action.city
+              old_token = city.tokens[0]
```
ollybh: You're using `.first` instead of `[0]` in other places, it would be better to be consistent.
```suggestion
              old_token = city.tokens.first
```

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,12 +43,39 @@ def process_lay_tile(action)
...
+              old_token.remove!
+              city.exchange_token(action.entity.find_token_by_type)
+
+              # apply "other" OO token once first to operate is applied if floated
+              corporation = @game.corporations.find { |c| c.name == 'NWR' }
+              if corporation&.floated
+                token = corporation.find_token_by_type
+                city = tile.cities.reject(&:tokened?).first
+                city.place_token(corporation, token) if city.tokenable?(corporation, tokens: token)
+              end
```
ollybh: This is mostly repeating the code in the first branch of the `if` statement. Can this be extracted into a method?

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -79,7 +106,9 @@ def legal_tile_rotation?(_entity, hex, tile)
             (old_tile.exits - tile.exits).empty?
           end
 
-          def potential_tiles(_entity_or_entities, _hex)
+          def potential_tiles(_entity_or_entities, hex)
+            return if hex.name == 'G8'
```
ollybh: Do you need a check to see if there is a pending home token, so you don't block laying track in the home hex?

ollybh (changes requested):

lib/engine/game/g_18_india/step/home_track.rb
MatthewDeKoning: thanks for bearing with me as i get used to the code base / development cycle!
i believe this function is primarily used when upgrading a yellow town to a green city. this was called incorrectly if the player misclicked on the OO tile vs on the token when laying the home OO token. this change does not break the ability to lay normal home tile in yellow on G8.

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,12 +43,38 @@ def process_lay_tile(action)
...
+              # remove and swap the NWR that appeared... hacky...
+              city = action.city
+              old_token = city.tokens.first
+              old_token.remove!
+              city.exchange_token(action.entity.find_token_by_type)
+
+              # apply "other" OO token once first to operate is applied if floated
+              place_second_oo_token(tile, 'NWR')
+              @round.pending_tokens.shift
+            end
```
michaeljb: You mentioned elsewhere some other companies/tiles that could also be impacted by this. It got me thinking, and you might be able to use [`case/when`](https://docs.ruby-lang.org/en/3.2/syntax/control_expressions_rdoc.html#label-case+Expression) (Ruby's `switch`) to DRY it up a little bit and make it simpler to add those other companies and tiles, for example:

```suggestion
            other_corp =
              case [action.entity.name, tile.name]
              when ['NWR', '235']
                'SPD'
              when ['SPD', '235']
                # remove and swap the NWR that appeared... hacky...
                city = action.city
                old_token = city.tokens.first
                old_token.remove!
                city.exchange_token(action.entity.find_token_by_type)
                'NWR'
              end
            if other_corp
              # apply "other" OO token once first to operate is applied if floated
              place_second_oo_token(tile, other_corp)
              @round.pending_tokens.shift
            end
```

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -79,7 +105,9 @@ def legal_tile_rotation?(_entity, hex, tile)
             (old_tile.exits - tile.exits).empty?
           end
 
-          def potential_tiles(_entity_or_entities, _hex)
+          def potential_tiles(_entity_or_entities, hex)
+            return if hex.name == 'G8'
```
michaeljb: @MatthewDeKoning maybe something like this? Then it should apply for other companies and gets rid of the mysterious magic string `"G8"`

```suggestion
          def potential_tiles(entity_or_entities, _hex)
            # entity_or_entities is an array when combining private company abilities
            entity, *_combo_entities = Array(entity_or_entities)
            return if actions(entity).include?('place_token')
```

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,12 +43,38 @@ def process_lay_tile(action)
...
+            # apply "other" OO token once first to operate is applied if floated
+            if action.entity.name == 'NWR' && tile.name == '235'
+              place_second_oo_token(tile, 'SPD')
+              @round.pending_tokens.shift
+            elsif action.entity.name == 'SPD' && tile.name == '235'
+              # remove and swap the NWR that appeared... hacky...
+              city = action.city
+              old_token = city.tokens.first
+              old_token.remove!
+              city.exchange_token(action.entity.find_token_by_type)
```
michaeljb: Just to check if I'm understanding the issue correctly: are these lines only necessary here on the SPD case and not on the NWR case, because this problem can only happen when they both float in the same round? And since they have fixed par values, the operating order would always be the same (NWR first)?

MatthewDeKoning: Yeah so for whatever reason, the main actual bug this began from is that if SPD somehow operates first it places an NWR token if NWR is also floated. The OO tile token placement has confused players since India was implemented, but it is not technically broken, just not clear they need to click to place "the second company" token in the other spot. This fix auto-places that token.

lib/engine/game/g_18_india/step/home_track.rb
MatthewDeKoning: this function basically returns the yellow town -> green city tile set. I am concerned if we exclude entities trying to place a token (i.e. GIPR upgrading a town to a city on float and tokening, somewhat common) we break that functionality.

lib/engine/game/g_18_india/step/home_track.rb
MatthewDeKoning: I have confirmed this breaks the GIPR float case.

michaeljb:
@MatthewDeKoning would you mind adding a fixture? Just save the JSON data for a hotseat game under `public/fixtures/18 India/`, and then that game will be automatically added to the test suite. The game does need to be ended in the 18xx.games engine but doesn't have to reach an actual endgame according to the rules, i.e., you could hit "End Game" in the Tools tab right after the token resolution happens.

MatthewDeKoning:
I am not certain this is working, but the example is in the repo at least.
docker-compose exec rack bundle exec rspec -e '18 India'/OO-Fix.json
does not fail but also claims to not run any examples. what is the proper command to run this test?
 i am also perturbed by the space in the directory name but this has been a well documented issue on this title for a long time :(

MatthewDeKoning:
hmmm. the build failure maybe suggests the test fixtures are not setup for India game parsing (or something, honestly no idea, seems to fail in an India specific setup portion of the game log: # ./lib/engine/game/g_18_india/game.rb:277:in `block (2 levels) in setup_corporations_by_region!').

michaeljb:
> I am not certain this is working, but the example is in the repo at least. docker-compose exec rack bundle exec rspec -e '18 India'/OO-Fix.json does not fail but also claims to not run any examples. what is the proper command to run this test? i am also perturbed by the space in the directory name but this has been a well documented issue on this title for a long time :(

`docker-compose exec rack bundle exec rspec -e "OO-Fix"` should work. `-e` gives rspec a string to match for the test name, and the test name is constructed from the strings given to `describe` or `context`, the string given to the actual test, at `it`. So for this case, it would be these three lines here:

https://github.com/tobymao/18xx/blob/3921ac42a6f30720a220a24e1adb547bc198f55e/spec/lib/engine/games/game_spec.rb#L15-L17

So the full test name would be `"18 India OO-Fix matches result exactly"` (no `/` or `.json`). When you use `-e`, rspec looks for all its tests and runs any tests where your given string is a substring of the test name, so you can always use `-e India` to run all 18 India-specific tests, or `-e OO-Fix` for tests just for the new fixture.

michaeljb:
> hmmm. the build failure maybe suggests the test fixtures are not setup for India game parsing (or something, honestly no idea, seems to fail in an India specific setup portion of the game log: # ./lib/engine/game/g_18_india/game.rb:277:in `block (2 levels) in setup_corporations_by_region!').

> 
```
18 India OO-Fix matches result exactly
     Failure/Error: ((RAND_A * @rand) + RAND_C) % RAND_M
```

The randomization `seed` in your fixture file is a string, but needs to be a number--just take the quotes out so that bit is `"seed":1751744766` instead of `"seed":"1751744766"`. I'm surprised this is a problem, did you add in the seed by hand? If it was put in there when you copied the data out via the Tools tab that's definitely something we'll want to fix.

MatthewDeKoning:
I did add the seed by hand (needed a setup that included SPD and NWR, you would be surprised now many games you can create before that shows up haha). i moved the json "string" to a number and it is fine now.

MatthewDeKoning:
i would fix how this is parsed if i new more about the test fixture aspect of this, still lots to learn!

lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -43,13 +43,46 @@ def process_lay_tile(action)
...
-            replace_oo_reservations(tile) unless tile.reservations.empty? # move hex reservation
+            other_corp =
+              case [action.entity.name, tile.name]
+              when %w[NWR 235]
+                'SPD'
+              when %w[SPD 235]
+                city = action.city
+                old_token = city.tokens.first
+                old_token.remove!
+                city.exchange_token(action.entity.find_token_by_type)
```
michaeljb: This is repeated in the next `when` block, so it could be extracted to a function as well. Something like this?

```
def place_first_oo_token(city, entity)
  old_token = city.tokens.first
  old_token.remove!
  city.exchange_token(entity.find_token_by_type)
end
```

```suggestion
                place_first_oo_token(city, action.entity)
```

MatthewDeKoning: sorry for the delay, i made those changes, should be good to go.

PR #11865, opened 2025-07-13
[1837] Exchanges shouldn't allow presidency change before corporation parred

crericha:
Also update presidency logic for the situation where multiple shares have already exchanged before the corporation has parred. The parring player should maintain presidency if a majority share holder.

### DIFF
--- a/lib/engine/game/g_1837/game.rb
+++ b/lib/engine/game/g_1837/game.rb
@@ -471,7 +471,7 @@ module Engine
         def exchange_coal_minor(minor)
           target = exchange_target(minor)
           @log << "#{minor.id} exchanged for a share of #{target.id}"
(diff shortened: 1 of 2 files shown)


michaeljb (approved):

perwestling:
@crericha I noticed a strange behavior when trying 1824. It was a National (SD) which only had reserved shares. When SD formed the exchange took place (giving president's share to SD1 owner, and a 10% to SD3 owner, SD2 was not selected so its reserved had already been unreserved) where a player got 30% in the exchange. And then the code changed presidency and swapped 2 shares from SD to SD... This resulted that the player lost 10%.

I made a fix for that as I figured it was an 1824 issue but I will try to see if this fix solves the issue instead.

PR #11866, opened 2025-07-14
[1825] 3T and 4T trains should not visit same hex twice.

### LINKED ISSUE #11864: 1825 - train routing rule no longer enforced
Section 4.5.3 of the 1825 rules state that a train cannot run to two cities on the same hex or tile. The game used to enforce the rule but no longer does so trains are able to run out of London and other hexes then back in. Auto routing also runs trains this way.

**What's your Game ID?**
212819

Image

Image

-->

philcampeau:
Fixes #11864

## Before clicking "Create"

### Explanation of Change

Currently, the check_distance code sent a `return` for 3T and 4T trains prior to the check against visiting multiple stops on the same hex, but as far as I can see in the rules, they are still bound by that restriction as well.

I moved their return call to where the U3 and 2+2 trains' is, which still permits them to end their run on a town.

This will definitely require pins, as currently there could be games where these trains broke that rule.

### DIFF
--- a/lib/engine/game/g_1825/game.rb
+++ b/lib/engine/game/g_1825/game.rb
@@ -1109,22 +1109,11 @@ module Engine
           else
             super(route, visits, build_dummy_train(route, num))
           end
-          return if %w[3T 4T].include?(route.train.name)
 
-          node_hexes = {}
-          visits.each do |node|
-            # For 3T and 4T trains, allow multiple towns in same hex but not cities
-            if %w[3T 4T].include?(route.train.name)
-              raise GameError, 'Cannot visit multiple cities in same hex' if node.city? && node_hexes[node.hex]
-
-              node_hexes[node.hex] = true if node.city?
-              # Towns are allowed multiple times for 3T/4T, so no check needed
-            else
-              # For all other trains, the existing rule applies
-              raise GameError, 'Cannot visit multiple towns/cities in same hex' if node_hexes[node.hex]
-
-              node_hexes[node.hex] = true
-            end
+          # A train route cannot visit more than one stop on the same hex,
+          # unless it is a 3T or a 4T and all the stops are towns.
+          raise GameError, 'Cannot visit multiple towns/cities in same hex' if visits.group_by(&:hex).any? do |_, stops|
+            stops.size > 1 && !(%w[3T 4T].include?(route.train.name) && stops.all?(&:town?))
           end
 
           return if %w[3T 4T U3 2+2].include?(route.train.name)


lib/engine/game/g_1825/game.rb
```diff
@@ -1105,15 +1105,14 @@ def check_distance(route, visits)
           else
             super(route, visits, build_dummy_train(route, num))
           end
-          return if %w[3T 4T].include?(route.train.name)
```
roseundy: See my update to the referenced bug. This needs to be fixed so that we allow multiple dits to be visisted on the same tile by 3T and 4T trains, but not cities.

philcampeau:
Hey there @roseundy. Looks like I neglected to fix this before. Let me know what you think of my new fix. I think this does the job right.

lib/engine/game/g_1825/game.rb
roseundy: Won't a 3T or 4T skip the check you added below? Seems like you need to remove this line.

roseundy (changes requested):

lib/engine/game/g_1825/game.rb
```diff
@@ -1109,11 +1109,21 @@ def check_distance(route, visits)
 
           node_hexes = {}
           visits.each do |node|
-            raise GameError, 'Cannot visit multiple towns/cities in same hex' if node_hexes[node.hex]
+            # For 3T and 4T trains, allow multiple towns in same hex but not cities
+            if %w[3T 4T].include?(route.train.name)
+              raise GameError, 'Cannot visit multiple cities in same hex' if node.city? && node_hexes[node.hex]
 
-            node_hexes[node.hex] = true
+              node_hexes[node.hex] = true if node.city?
```
ollybh: There is at least one tile with both a town and a city on it. This test is going to allow the route to visit both of these – is that right? Or is the route only allowed to visit two stops on the same tile if both are towns?

roseundy: @ollybh quite right. It's illegal unless it's a "T" train and both are dits.

ollybh: Maybe something like this would work:
```
raise GameError, 'Cannot visit multiple towns/cities in the same hex' if visits.group_by(&:hex).any? do |_, stops|
  stops.size > 1 && !(%w[3T 4T].include?(route.train.name) && stops.all?(&:town?))
end
```

ollybh:
@philcampeau It's been a couple of months since any updates to this PR. Have you looked at this any more?

philcampeau:
I haven't I've been a little busier than usual and haven't had much time. I don't mind if someone wants to close this and do the change themselves.

ollybh:
@roseundy I've hijacked @philcampeau's PR, and suggested an alternative implementation for this.

I've carried out a minimal amount of testing which showed that:
- 4T trains can visit two towns on the same hex.
- 3T and 4T trains can't visit two cities on the same hex.
- Other trains can't visit two towns on the same hex.

I believe that this is the correct behaviour, do you  agree? And does the code look good?

roseundy (approved):
Looks good

ollybh:
I've removed the pins label. Old train routes aren't revalidated, so this won't break any existing games even if they had 3T/4T trains that had run to two cities on the same hex.

PR #11882, opened 2025-07-30
[System18] New map - Mississippi

roseundy:
This adds a new map - Mississippi - to System18.

Not pins/archives are needed.


roseundy:
Added a commit to address the "no corps started" game end rule. I created a new System18-specific stock round to do this.

PR #11898, opened 2025-08-05
[18Ardennes] adds desc text to minors saying which corp they can form.

philcampeau:
## Before clicking "Create"

### Explanation of Change

A complaint I heard was difficulty in remembering which minor could be used to form which corps. So I added a description to each of them with that information. By default it's very short, but if you click on the ability it gives you a more verbose explanation.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_18_ardennes/entities.rb
+++ b/lib/engine/game/g_18_ardennes/entities.rb
@@ -40,7 +40,16 @@ module Engine
             capitalization: :none,
             coordinates: 'B8',
             city: 0,
-            abilities: [{ type: 'exchange', corporations: 'ipoed', from: %w[ipo market] }],
+            abilities: [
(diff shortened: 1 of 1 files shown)


ollybh (changes requested):
No, this isn't the right approach. The public companies that can be formed using a minor are not fixed. The ones you have listed are for the minors' initial tokens in the first public company auction. If a minor places its second token then there might be other public companies they could form. And in the second (or later) public company auctions there are no geographical restrictions.

ollybh:
A better way to do this would be to add something to the `status_array` method, returning a description of which public companies the minor could be used to start. This shouldn't be too difficult, there is existing code that checks the association between minors and public companies.

philcampeau:
I'll take a look at that in the next few days. Thanks!

ollybh:
PR #11987 is an alternate implementation for this, which calculates which public companies can be started by each minor based on both its tokens, not just the home token.

PR #11902, opened 2025-08-08
A major rewrite of 1824, heavily based on 1837

perwestling:
Have fixed some bugs, and implemented missing pieces, for the 1824 title. Have also played through a complete game until bank break.

The game is ready for some initial testing but has not been set in alpha yet.

Yesterday's master

Not applicable - no games yet.

Yes

Yes, but this commit affects no general code, nor 1837

## Implementation Notes

Some issues still remain:
1. Coal Railway closed, but reserv of mine still remained
2. Did notice that presidency change did not work as it should but that might be related to a later fixed reservation issue.
3. Two player variant not implemented yet

There is potential refactorings of 1837 code so that some 1824 code can be removed, but I wait with that until
there are some tests available for 1824, to avoid breaking anything.

### Explanation of Change

Have been working on 1824 for almost 3 years. In the meantime 1837 was implemented. As those two games are similar in some ways I changed major parts of the 1824 implementation to use 1837 as a base.

But major parts from my initial work still remains.

### Screenshots

No applicable.

### Any Assumptions / Hacks

Not really


lib/engine/game/g_1824/corporation.rb
```diff
@@ -6,16 +6,89 @@ module Engine
...
+
+        # Used when a pre-staatsbahn is unsold during initial SR.
+        # We need to unreserve one of the shares of the national.
+        def unreserve_one_share!
+          shares.each do |share|
+            next if share.buyable
+
+            share.buyable = true
+            break
+          end
```
michaeljb: nitpick: IMO `find` is more readable than `each` with a `break`, maybe something like this?

```suggestion
          (shares.find { |s| !s.buyable })&.buyable = true
```

Or if you don't like the `&` syntax

```suggestion
          share = shares.find { |s| !s.buyable }
          share.buyable = true if share
```

lib/engine/game/g_1824/round/first_stock.rb
```diff
@@ -40,70 +40,80 @@ def next_entity_index!
...
-              # Private is a control of a Coal Railway
-              regional = @game.associated_regional_railway(minor)
-              @game.log << "Coal Railway #{minor.name} closes; #{regional.name}'s presidency share "\
-                           'is no longer reserved'
-
-              close_minor(minor)
-              c.close!
+              # There is no implementation for close of a major Pre-Staatbahn - this is a weird corner case
+              if company.sym.end_with?('1')
+                raise GameError, 'The weird case of unsold SD1, KK1, UG1 is not supported. Please reconsider. They are good!'
```
michaeljb, line 81: 😄

lib/engine/game/g_1824/step/buy_train.rb
```diff
@@ -11,11 +11,34 @@ def can_entity_buy_train?(_entity)
             true
           end
 
+          def must_buy_train?(entity)
+            # Rule VII.11 all entities must own a train, unless coal company and there are no g-trains in the depot
+            return false unless entity.trains.empty?
+            return true unless @game.coal_railway?(entity)
+
+            depot_g_trains = @depot.depot_trains.select { |t| @game.goods_train?(t.name) }
+            depot_g_trains.any?
```
michaeljb:
```suggestion
            !depot_g_trains.empty?
```

lib/engine/game/g_1824/game.rb
```diff
@@ -1,192 +1,87 @@
...
+        CORPORATION_CLASS = G1824::Corporation
+        DEPOT_CLASS = G1824::Depot
 
+        attr_accessor :two_train_bought, :forced_mountain_railway_exchange, :coal_company_initial_cash, :player_debts
+
+        # TODO: Can these be removed? They do not seem to be used in 1824?
         register_colors(
           gray70: '#B3B3B3',
           gray50: '#7F7F7F'
         )
```
michaeljb: Yeah it looks like when a game defines these, they can use them as tile/hex colors, e.g., see 21Moon with hexes that are `gray50` and `gray60`. AFAICT these can be removed.

```suggestion
```

lib/engine/game/g_1824/game.rb
```diff
@@ -368,18 +274,50 @@ def optional_hexes
...
+          return :corporation unless two_player?
+
+          # Rule X.4, bullet 2 - 2 player 1824 has a bank pool
+          :bank
+        end
+
+        # 1824 use normal tile upgrades so use Base version instead
+        def upgrades_to?(from, to, special = false, selected_company: nil)
+          Base.instance_method(:upgrades_to?).bind_call(self, from, to, special, selected_company)
+        end
```
michaeljb: Hmmm I don't like `bind_call` here.

Maybe in the 1837 code you could change `upgrades_to?` to something like this:

```
        def upgrades_to?(from, to, special = false, selected_company: nil)
          return yellow_town_tile_upgrades_to?(from, to) if use_yellow_town_tile_upgrades_to?(from)

          super
        end

        def use_yellow_town_tile_upgrades_to?(from)
          from.color == :yellow && !from.towns.empty?
        end
```

And then here in 1824 you could just have this:

```suggestion
        # 1824 use normal tile upgrades so use Base version instead
        def use_yellow_town_tile_upgrades_to?(_from)
          false
        end
```

This reminds me of an idea I had ages ago but never got around to trying, where instead of having one game class inherit directly from another there would be an abstract base class for that "family" of games, i.e., `G1837` and `G1824` would both inherit from `G1837Abstract`. Then the override logic in cases like this could be cleaner. Could be worth revisiting that idea...

lib/engine/game/g_1824/game.rb
```diff
@@ -392,87 +330,165 @@ def init_round
...
+          @two_train_bought = false
+
+          # When 1st 4-train is bought any remaining MRs will be exchanged
+          @forced_mountain_railway_exchange = []
+
+          # Used to remember what coal companies was bought for
+          # TODO: Improve this solution
+          @coal_company_initial_cash = Hash.new { |h, k| h[k] = [] }
+
+          # Initialize the player depts, if player have to take an emergency loan
```
michaeljb:
```suggestion
          # Initialize the player debts, if player have to take an emergency loan
```

lib/engine/game/g_1824/game.rb
```diff
@@ -557,251 +579,297 @@ def sorted_corporations
...
-          @corporations.reject(&:removed)
-        end
-
-        def event_close_mountain_railways!
-          @log << '-- Any remaining Mountain Railways are either exchanged or discarded'
-          # If this list contains any companies it will trigger an interrupt exchange/pass step
-          @forced_mountain_railway_exchange = @companies.select { |c| mountain_railway?(c) && !c.closed? }
+        # 1824 uses standard sold out stock movement. Rule VIII.3, bullet 4: move up 1 if not at top
+        def sold_out_stock_movement(corporation)
+          Base.instance_method(:sold_out_stock_movement).bind_call(self, corporation)
```
michaeljb: See my other `bind_call` comment; we should add a non-base method in G1837 to check if its special implementation should be used, and G1824 can override that to false so that it doesn't have to use `bind_call`.

michaeljb (changes requested):
I want to get rid of the `bind_call` usage, otherwise my comments are pretty minor and the rest looks good.

(disclaimer: I don't have any experience with 1837 or 1824 so I can't really say anything about details of implementing their special rules)

lib/engine/game/g_1824/round/first_stock.rb
perwestling: When I come across this in the rules I opened a thread on BGG (https://boardgamegeek.com/thread/3546208/what-if-main-preestatsbahn-not-sold-during-startin). DO not think it is that hard to implement so I will put up a TODO here and possibly fix it in a later commit.

lib/engine/game/g_1824/step/buy_train.rb
perwestling: Fixed

lib/engine/game/g_1824/game.rb
perwestling: Removed it.

lib/engine/game/g_1824/game.rb
perwestling: I have done my changes without making any changes to 1837. I think your suggestion is good, and have applied it.

lib/engine/game/g_1824/game.rb
perwestling: Done

lib/engine/game/g_1824/game.rb
perwestling: @crericha I have made this, and the one below, in 1837 - in this commit.

lib/engine/game/g_1824/game.rb
perwestling: Have introduced a boolean method to do this and added to 1837.

lib/engine/game/g_1824/game.rb
perwestling: I put this as resolved as I think it is.

lib/engine/game/g_1837/game.rb
```diff
@@ -770,11 +770,16 @@ def token_graph_for_entity(_entity)
...
 
         def upgrades_to?(from, to, special = false, selected_company: nil)
-          return yellow_town_tile_upgrades_to?(from, to) if from.color == :yellow && !from.towns.empty?
+          return yellow_town_tile_upgrades_to?(from, to) if use_yellow_town_tile_upgrades_to?(from)
 
           super
         end
 
+        # Need this separate method due to 1824
+        def use_yellow_town_tile_upgrades_to?(from)
```
crericha: Why did yo need this method instead of overwriting `yellow_town_tile_upgrades_to?` to the proper implementation for 24?

lib/engine/game/g_1837/game.rb
```diff
@@ -798,13 +803,20 @@ def sold_out?(corporation)
         end
 
         def sold_out_stock_movement(corporation)
+          return super unless use_diagonal_stock_movement?
```
crericha: Can we base this off of the hex stock market instead?

perwestling: Yes, that is a better idea.

lib/engine/game/g_1837/game.rb
perwestling: It was a simpler solution, as 1824 upgrade of town tile worked with base. This was a simple refactoring which had little impact on 1837.

perwestling: But you are right. Implementing yellow_town_tile_upgrades_to? was simple. Just like 1837 that method do not need to check for color progression, but just that existing paths are respected.
So I have changed this to remove the change in 1837, and do it in 1824 only.

perwestling:
@crericha I have now made changes after your comments.

perwestling:
@crericha - is it OK after the changes you suggested?

perwestling:
@michaeljb I have not gotten any feedback from @crericha since my update after his comments. 
But maybe this is good enough to approve?

I do have at least two PRs I need to make based on this, before 1824 can be put in Alpha:

- A PR with come corrections due to bugs found in playtesting
- A PR with support for 2 player variant

Playtesting of these have been going for some time, and 1824 works quite well.
Have a short list of things to fix.
But it simplifies things if master does contain this PR.

crericha:
@perwestling changes to the 1837 game class look good. Thanks for making them. 👍

PR #11935, opened 2025-08-25
[18FL] Corrected price for POSC private company

ollybh:
The Peninsular and Occidental Steamship Company costs $40, not $30.

This change is likely to need some games pinning. Pin cf50d53ee can be used for these games.

Fixes tobymao#11923.

## Before clicking "Create"


philcampeau:
Wow, that looks like it's been there since 18FL was first coded. 

We'll likely need to scrap the fixture files too.

ollybh:
> Wow, that looks like it's been there since 18FL was first coded.
> 
> We'll likely need to scrap the fixture files too.

Yup. Fixture `hotseat04` could be salvaged by changing Player 2’s final score. The other ones probably aren't fixable.

michaeljb:
@philcampeau should new 18FL fixtures be created so that the functionality covered by these fixtures is still tested? For reference the removed fixtures were originally added in 751ee298eb869b04f5b6fc0d17c4e37daf51f014 and 8bd8f11f196b17cc9834bd3a39c829f9b0f84a6c.

Issue #11929, opened 2025-08-23
Receivership entities appear twice in Entity view


perwestling:
A corporation in receivership appears twice under Bank.

**What's your Game ID?**

I believe this is a general issue, but I did encounter it during play test of a new title.

**How to Reproduce the Bug?**

Set a corporation into receivership.
Go to Entity tab.

**Which Browser / Platform?**

N/A

**Likely source of problem**

In entity.rb (view) the following code renders the corporations:

<pre>
extra_bank = []
if @game.respond_to?(:unstarted_corporation_summary)
  others = @game.unstarted_corporation_summary.last
  extra_bank.concat(others.map { |c| h(Corporation, corporation: c) })
else
  extra_bank.concat(bank_owned.map { |c| h(Corporation, corporation: c) })
end
children << h(:div, [
  h(Bank, game: @game),
  h(GameInfo, game: @game, layout: 'upcoming_trains'),
  *@game.unowned_purchasable_companies(@current_entity).map { |company| h(Company, company: company) },
  *@game.receivership_corporations.map { |c| h(Corporation, corporation: c) },
  *extra_bank,
].compact)
</pre>

The base implementation of receivership_corporations is just to select all corporations that is in receivership.
At the same time all corporations are split into player_owned and bank_owned. Therefor the corporation appear in two places, giving the duplication.

perwestling:
As a workaround (in 1824 2-player - bond railway) I implemented **receivership_corporations** to just return an empty list, as that method is only used in this view.
This method is only overridden in 1841, but after checking a few games of 1841, I did not see any duplicates.

So I put this here, and in case it does appear in some other title it might need to be fixed properly.

PR #11937, opened 2025-08-25
[1824] Add support for 2 players, and do and some corrections

perwestling:
2-player 1824 uses Cisleithania map. It also use a special drafting, as well as construction and bond companies. Also uses reduced bank, minors/corporations, mountain railways, trains.

This version has been play tested with both 4 players and 2 players with good result. There still remains some things to fix before
it will be put as pre-alpha according to me, but it is good enough to be a pre-alpha game already.

## Before clicking "Create"

## Implementation Notes

This commit contains both bug fixes as well as adding support for 2 players. I believe future commits will be smaller as this has been a catch up game.

### Explanation of Change

Trying to get 1824 up to alpha status. It should be before end of year but hopefully during September 2025.

### Screenshots

This is a snapshot from ongoing play test, that uses this baseline of this commit.
image

And a completed 2 player game.
image

### Any Assumptions / Hacks

[Work around for issue 11929.](https://github.com/tobymao/18xx/issues/11929)


lib/engine/game/g_1837/game.rb
```diff
@@ -800,7 +800,7 @@ def sold_out?(corporation)
 
         def sold_out_stock_movement(corporation)
           # Needed for 1824, which use base behavior for sold out (and do not use diagonal stock movements)
-          return super unless @stock_market.hex_market
+          return super unless @stock_market.hex_market?
```
michaeljb, line 803: This change has been made in another PR, so you might need to merge master into this branch so that GitHub doesn't think there is a conflict.

michaeljb (approved):

lib/engine/game/g_1837/game.rb
michaeljb: Apparently this isn't actually a problem, GitHub is reporting no conflict with #11939 merged

Issue #11932, opened 2025-08-24
System18 - NE US Playtesting Changes


scottredracecar:
Add support for four players (starting capital 430, cert limit 8)

Add a single private exactly like this private in 1889 (for all player counts). Private is sold in an ISR in an upwards auction (like in S18 Mississippi, for example), PD goes to the player after the winning bid.

<img width="330" height="183" alt="Image" src="https://github.com/user-attachments/assets/1be12941-6f4e-48ea-9b4a-c394d68b667a" />

No games to pin, I don't mind losing the completed games. I think this will make for a bit more variation in the gameplay.

roseundy:
What happens if no-one bids on the private?

Also, since this is the first S18 game with an actual private, what are the rules for buying privates across to a corp? What phases, what range of values is legal?

scottredracecar:
Hmm, good question about if no one bids. I think the game ends in a tie. I think this is priced such that people SHOULD want to bid, but the rules should also give the players the idea that that is the case (or else they should play something else).

The rules already cover the privates, buy for half to double, starting Phase 3. These can be hard coded into the S18 base rules. Maps that have private companies that cannot be bought (or with different rules about this) in will have to provide override rules.

scottredracecar:
I would like to change the base rules for the upwards auction to be like 1861. I'll work on a new rulebook revision now.

```
If all players pass during the first Stock Round, and at
least one of the Private Companies has not yet been
purchased, the minimum bid for the next Private 
Company to be auctioned is reduced by R5, and the Stock
Round resumes with the player who initiated the round of
passes. If the minimum bid for a certificate reaches R0 in
this way, the player who last initiated a full round of
passes receives the Private Company for no cost.
```

roseundy:
OK. This will be the first "true" upwards auction in S18, since the Britain map isn't auctioning privates, but charters instead (nothing especially interesting happens if no-one auctions a charter) and the Mississippi map isn't auctioning privates but player order - which can't drop in price if no-one bids (so I require a bid of 0).

scottredracecar:
I just went in to update this rule about all passing and I see that it's already covered so I am not going to touch the rules now.

Sometime soon, I will update a couple clarifications Ian has made to the pdf rules and also update the default train limit to three.

PR #11938, opened 2025-08-25
[System18] Implement changed rules for NEUS and Northern Italy

### LINKED ISSUE #11922: [System18 Italy] Train limit needs to return to the original 4 (phases 2 & 3)
We played a game of Italy recently, and realised that - unlike most maps - this one needs the train limit to be 4 during phases 2 & 3.
In other words, this needs to be an exception to the recent change to drop the limit to 3 across all Sys18 maps.

roseundy:
Fixes #11922 and #11932 

No pins needed.

## Before clicking "Create"

### Explanation of Change

Added new private and auction round to S18: Northeastern US
Increased train limit in phases 2 and 3 for S18: Northern Italy

I revamped UpwardAuction to handle dropping prices when all privates aren't sold when every player passes. I modified CharterAuction to not do this and modifed OrderAuction to take advantage of the new UpwardAuction. As a result, the Mississippi map no longer requires a player to start an auction.

### DIFF
--- a/lib/engine/game/g_system18/map_neus_customization.rb
+++ b/lib/engine/game/g_system18/map_neus_customization.rb
@@ -98,7 +98,25 @@ module Engine
         end
 
         def map_neus_game_companies
-          []
+          [
+            {
+              name: 'Locomotive Works',
+              value: 150,
+              revenue: 30,
+              desc: 'Does not close while owned by a player. If owned by a player '\
+                    'when the first 5-train is purchased it may no longer be sold '\
+                    'to a public company and the revenue is increased to 50.',
+              sym: 'LW',
+              abilities: [{ type: 'close', on_phase: 'never', owner_type: 'player' },
+                          {
+                            type: 'revenue_change',
+                            revenue: 50,
+                            on_phase: '5',
+                            owner_type: 'player',
+                          }],
+              color: nil,
+            },
+          ]
         end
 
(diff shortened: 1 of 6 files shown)


lib/engine/game/g_system18/map_neus_customization.rb
```diff
@@ -137,16 +155,31 @@ def map_neus_game_trains(trains)
...
           find_train(trains, '6')[:num] = 1
           find_train(trains, 'D')[:num] = 10
           trains
         end
 
         def map_neus_game_phases
-          self.class::S18_FULLCAP_PHASES
+          phases = self.class::S18_FULLCAP_PHASES
+          phases[1][:status] = %w[can_buy_companies]
+          phases[2][:status] = %w[can_buy_companies]
```
michaeljb: Won't this mutate `S18_FULLCAP_PHASES`, which affects multiple titles? That could cause weird things when running tests across different titles, IIRC that's been a problem before. Unfortunately `.freeze` only freezes the top-level hash

roseundy: Good point.

Would `phases = self.class::S18_FULLCAP_PHASES.dup` work?

roseundy: Nevermind. I have a better way to fix it.

michaeljb (approved):

PR #11943, opened 2025-08-28
[1824] Refactor to put all Cisleithania specifics in g_1824_cisleithania

perwestling:
Refactor 1824 to split out Cisleithania variants (2 and 3 players, on a special map) in a subgame, deriving from 1824 base.

## Before clicking "Create"

## Implementation Notes

Discovered that some code was not needed.

Some files have been split into a 1824 and an 1824Cisleithania version. In most cases it was because variant code was mixed in.

Some of the files that was in 1824 has been moved to 1824Cisleithania, as they only contained variant exceptions.

A similar split has been done with the wiki page for 1824.

The commit also contains a small fix - when KK is formed, and one of the two tokens on Wien are returned, the cost of this is now 40G. This was after an answer by Lonny.

### Explanation of Change

By doing this the base game implementation became more clean, as a lot of exceptions could be removed / moved.

Also, it should make it possible to get different maps and tile sets in the wiki.

### Screenshots

No changes on looks

### Any Assumptions / Hacks

None


lib/engine/game/g_1824_cisleithania/step/buy_sell_par_shares_first_sr.rb
```diff
@@ -0,0 +1,78 @@
...
+  module Game
+    module G1824Cisleithania
+      module Step
+        class BuySellParSharesFirstSr < G1824::Step::BuySellParSharesFirstSr
+          def actions(entity)
+            actions = super
+
+            return actions unless @game.two_player?
+
+            actions.delete(:pass) if @game.companies.find { |c| c.stack && c.stack < 4 }
```
michaeljb: I expect this doesn't work on the server; the `actions` arrays use strings everywhere, not symbols, e.g.:

https://github.com/tobymao/18xx/blob/97a283921e47ddebadb6974aa693078a322c45fc/lib/engine/step/buy_sell_par_shares.rb#L28

```suggestion
            actions.delete('pass') if @game.companies.find { |c| c.stack && c.stack < 4 }
```

lib/engine/game/g_1824_cisleithania/game.rb
```diff
@@ -0,0 +1,532 @@
...
+        def init_companies(players)
+          companies = COMPANIES.dup
+
+          mountain_railway_count =
+            case players.size
+            when 2
+              2
+            when 3
+              3
+            end
```
michaeljb:
```suggestion
          mountain_railway_count = players.size
```

perwestling:
@michaeljb Thanks for the review. I have now applied your suggestions.

michaeljb (approved):

PR #11945, opened 2025-08-29
[dev] set `"loaded": true` when importing fixtures; require for all fixtures

michaeljb:
without this, `/fixture/<title>/<id>` ends up redirecting to `/game/<id>` and loading the game from the local DB instead of the fixture file

when copying/downloading hotseat data, `loaded` is already `true` so I only encountered this when trying `rake fixture_import`

## Before clicking "Create"


michaeljb:
closing this in favor of #11948

PR #11948, opened 2025-08-29
[dev] enforce formatted fixtures

michaeljb:
* **add specs to enforce whitespace removal, anonymization, and removal of chat messages
  from game fixtures** via `rake fixture_format`
    * no more silly-large diffs for giant JSON fixture files ([after this one, anyway](https://github.com/user-attachments/assets/c945d3f1-7855-4ce7-9c4b-787413cdc515))
* add `.gitattributes` and docs in `public/fixtures/README.md` for readable
  diffs on single-line JSON
* split auto-actions into a separate file from the standard fixture spec, makes
  it easier to run the other fixture tests in isolation
* update function `format_fixture_json` (called by `rake fixture_format`) to not
  anonymize if the player name is already something like "Player 1" or a single
  letter

* set `"loaded": true` when importing fixtures; require for all fixtures
    * without this, `/fixture/<title>/<id>` ends up redirecting to `/game/<id>`
      and loading the game from the local DB instead of the fixture file

## Before clicking "Create"


michaeljb:
got a bunch of test failures I didn't see when I ran tests locally, will reopen when fixed

PR #11964, opened 2025-09-06
[System18] Adding Gotland

patrikolesen:
Adding a new System 18 map called Gotland made by Pontus

## Before clicking "Create"

### DIFF
--- a/lib/engine/ability/float_cost.rb
+++ b/lib/engine/ability/float_cost.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Engine
+  module Ability
+    class FloatCost < Base
+      attr_accessor :float_cost
+
+      def setup(float_cost:)
+        @float_cost = float_cost
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_system18/game.rb
+++ b/lib/engine/game/g_system18/game.rb
@@ -16,6 +16,7 @@ require_relative 'map_britain_customization'
 require_relative 'map_northern_italy_customization'
 require_relative 'map_ms_customization'
 require_relative 'map_scotland_customization'
+require_relative 'map_gotland_customization'
 
 module Engine
   module Game
@@ -36,6 +37,7 @@ module Engine
         include MapNorthernItalyCustomization
         include MapMsCustomization
         include MapScotlandCustomization
+        include MapGotlandCustomization
 
         register_colors(red: '#d1232a',
                         orange: '#f58121',
@@ -485,15 +487,6 @@ module Engine
           stock_round
         end
 
-        def stock_round
-          GSystem18::Round::Stock.new(self, [
-            Engine::Step::DiscardTrain,
-            Engine::Step::Exchange,
-            Engine::Step::SpecialTrack,
-            Engine::Step::BuySellParShares,
-          ])
-        end
-
         def operating_steps
           if respond_to?("map_#{cmap_name}_operating_steps")
             send("map_#{cmap_name}_operating_steps")
@@ -547,6 +540,25 @@ module Engine
           send("map_#{cmap_name}_next_round!")
         end
 
+        def stock_steps
+          unless respond_to?("map_#{cmap_name}_stock_steps")
+            return [
+              Engine::Step::DiscardTrain,
+              Engine::Step::Exchange,
+              Engine::Step::SpecialTrack,
+              Engine::Step::BuySellParShares,
+            ]
+          end
+
+          send("map_#{cmap_name}_stock_steps")
+        end
+
+        def stock_round
+          return super unless respond_to?("map_#{cmap_name}_stock_round")
+
+          send("map_#{cmap_name}_stock_round")
+        end
+
         def emergency_issuable_bundles(entity)
           return [] if game_capitalization != :incremental
           return [] if entity.trains.any?
@@ -592,6 +604,8 @@ module Engine
         end
 
         def close_corporation(corporation, quiet: false)
+          return send("map_#{cmap_name}_close_corporation", corporation) if respond_to?("map_#{cmap_name}_close_corporation")
+
           super
 
           corporation = reset_corporation(corporation)
@@ -774,6 +788,17 @@ module Engine
           return super unless respond_to?("map_#{cmap_name}_game_end_check_values")
 
           send("map_#{cmap_name}_game_end_check_values")
+        end
+
+        def bankruptcy_limit_reached?
+          case self.class::BANKRUPTCY_ENDS_GAME_AFTER
+          when :one
+            @players.any?(&:bankrupt)
+          when :all_but_one
+            @players.count { |p| !p.bankrupt } == 1
+          when :all
+            @players.all?(&:bankrupt)
+          end
         end
       end
     end
--- a/lib/engine/game/g_system18/map_base_customization.rb
+++ b/lib/engine/game/g_system18/map_base_customization.rb
@@ -37,11 +37,11 @@ module Engine
         end
 
         def map_base_game_cash
-          { 2 => 800, 3 => 500, 4 => 400, 5 => 300 }
+          { 1 => 1000, 2 => 800, 3 => 500, 4 => 400, 5 => 300 }
         end
 
         def map_base_game_cert_limit
-          { 2 => 20, 3 => 15, 4 => 10, 5 => 8 }
+          { 1 => 20, 2 => 20, 3 => 15, 4 => 10, 5 => 8 }
         end
 
         def map_base_game_capitalization
--- a/lib/engine/game/g_system18/map_gotland_customization.rb
+++ b/lib/engine/game/g_system18/map_gotland_customization.rb
@@ -1,0 +1,535 @@
+# frozen_string_literal: true
+
+require_relative 'step/gotland_difficulty_selection'
+require_relative 'round/gotland_difficulty_selection'
+require_relative 'step/gotland_nationalize_corporation'
+require_relative 'round/gotland_nationalization'
+
+module Engine
+  module Game
+    module GSystem18
+      module MapGotlandCustomization
+        D_TILE_MAP = {
+          'B9' => [3, 4, 4, 5, 5, 6],
+          'D3' => [4, 4, 5, 0, 0, 1],
+          'D7' => [3, 4, 5, 0, 1, 2],
+        }.freeze
+
+        DIFFICULTY_VALUES = {
+          'easy' => 10,
+          'normal' => 20,
+          'hard' => 30,
+          'very_hard' => 40,
+        }.freeze
+
+        CHOICES = {
+          1 => 'Sell All Shares',
+          2 => 'Sell 1 Share',
+          3 => 'Status quo',
+          4 => 'Buy 1 Share',
+          5 => 'Buy 2 Shares',
+          6 => 'Buy 3 Shares',
+        }.freeze
+
+        MARKET_1D_WITH_30B = [
+          %w[30c
+             40
+             45
+             50p
+             55p
+             60p
+             65p
+             70p
+             80p
+             90p
+             100p
+             110p
+             120p
+             135p
+             150p
+             165
+             180
+             200
+             220
+             245
+             270
+             300
+             330
+             360
+             400
+             440
+             490
+             540],
+        ].freeze
+
+        def map_gotland_setup
+          rival.owner = rival
+        end
+
+        def map_gotland_constants
+          redef_const(:CURRENCY_FORMAT_STR, '%s SEK')
+          redef_const(:BANKRUPTCY_ENDS_GAME_AFTER, :all)
+        end
+
+        def map_gotland_game_companies
+          []
+        end
+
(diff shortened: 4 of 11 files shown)


lib/engine/game/g_system18/game.rb
```diff
@@ -547,6 +540,25 @@ def next_round!
...
+              Engine::Step::SpecialTrack,
+              Engine::Step::BuySellParShares,
+            ]
+          end
+
+          send("map_#{cmap_name}_stock_steps")
+        end
+
+        def stock_round
+          return super unless respond_to?("map_#{cmap_name}_stock_round")
```
roseundy: This is not correct. I understand you want a map to be able to override the System18 default stock round, but this will use the 18xx base stock round if a map doesn't create a map_*_stock_round.

In addition, stock_steps isn't being called in this file. If a map wanted to override just that, it wouldn't have any effect.

Specifically, the "end game if no floated corporations" test will be gone.

roseundy: Please make sure the stock_round I created for System18 is used if a map isn't overriding it.

lib/engine/game/g_system18/game.rb
```diff
@@ -775,6 +789,17 @@ def game_end_check_values
...
         end
+
+        def bankruptcy_limit_reached?
+          case self.class::BANKRUPTCY_ENDS_GAME_AFTER
+          when :one
+            @players.any?(&:bankrupt)
+          when :all_but_one
+            @players.count { |p| !p.bankrupt } == 1
+          when :all
+            @players.all?(&:bankrupt)
```
roseundy: Why is this needed? Why doesn't `:all` work for Gotland?

lib/engine/game/g_system18/map_base_customization.rb
```diff
@@ -37,11 +37,11 @@ def map_base_game_corporations(corps)
         end
 
         def map_base_game_cash
-          { 2 => 800, 3 => 500, 4 => 400, 5 => 300 }
+          { 1 => 1000, 2 => 800, 3 => 500, 4 => 400, 5 => 300 }
         end
```
roseundy, line 41: Is this needed? These values are only used when no map is selected.

lib/engine/game/g_system18/map_base_customization.rb
```diff
@@ -37,11 +37,11 @@ def map_base_game_corporations(corps)
...
 
         def map_base_game_cert_limit
-          { 2 => 20, 3 => 15, 4 => 10, 5 => 8 }
+          { 1 => 20, 2 => 20, 3 => 15, 4 => 10, 5 => 8 }
```
roseundy, line 44: Same as above.

lib/engine/ability/float_cost.rb
```diff
@@ -0,0 +1,15 @@
...
+  module Ability
+    class FloatCost < Base
+      attr_accessor :float_cost
+
+      def setup(float_cost:)
+        @float_cost = float_cost
+      end
+    end
+  end
+end
```
roseundy: I believe there is a way to add a game-specific ability. I try to avoid adding to the common code base unless I have no choice or I believe the code may be shared across games.

lib/engine/game/g_system18/map_base_customization.rb
roseundy: Nevermind. You changed the meta.

lib/engine/game/g_system18/round/gotland_difficulty_selection.rb
```diff
@@ -0,0 +1,30 @@
...
+          end
+
+          def show_in_history?
+            false
+          end
+        end
+      end
+    end
+  end
+end
```
roseundy, line 30: It feels like this should be a game option rather than an in-game choice.

lib/engine/game/g_system18/round/gotland_stock.rb
```diff
@@ -0,0 +1,42 @@
...
+          end
+
+          def track_newly_floated(corporation)
+            @newly_floated_corporations << corporation
+          end
+        end
+      end
+    end
+  end
+end
```
roseundy: I think I might prefer this to be a generic Stock round for System18 that calls `@game` methods that can be overridden by maps.

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
+# frozen_string_literal: true
+
+require_relative 'step/gotland_difficulty_selection'
+require_relative 'round/gotland_difficulty_selection'
+require_relative 'step/gotland_nationalize_corporation'
+require_relative 'round/gotland_nationalization'
```
roseundy: Are these requires actually required? I haven't needed similar lines for the maps I've done.

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
...
+
+        def assign_difficulty_level(level)
+          @difficulty_level = level
+          redef_const(:TILE_LAYS, [{ lay: true, upgrade: true, cost: difficulty_level_value }].freeze)
+          reduce_float_costs(0)
+
+          @green_value = 50 - difficulty_level_value
+          @brown_value = 70 - difficulty_level_value
+          # Update the B3 hex with the new values
+          hex = hex_by_id('B3')
```
roseundy, line 222: Should use a constant rather than hard-coding the hex name directly.

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
...
+          return GSystem18::Round::DifficultySelection.new(self, [GSystem18::Step::DifficultySelection]) unless difficulty_level
+
+          @log << "-- #{round_description('Stock', 1)} --"
+          @round_counter = 1
+          stock_round
+        end
+
+        def reduce_float_costs(floated_corporation_index)
+          order = @corporation_float_order
+          if floated_corporation_index&.positive? && floated_corporation_index&.<(order.size)
```
roseundy, line 308: Why is the second `&.` needed here? If the first fails, it won't get to the 2nd.

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
...
+            end
+          end
+        end
+
+        def can_par?(corporation, player)
+          player_president_shares = corporations.count { |corp| corp.president?(player) }
+          return false if player_president_shares >= 4
+
+          super
+        end
```
roseundy, line 340: Never override game methods here!  Create a map_gotland_can_par? method. This may break other maps!

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
...
+            player.spend(float_cost, @bank)
+            @log << "#{player.name} pays #{format_currency(float_cost)} to float #{corporation.name}"
+            corporation.remove_ability(ability)
+          end
+          floated_index = @corporation_float_order.index { |c| c[:sym] == corporation.id }
+          reduce_float_costs(floated_index)
+          super
+          # Track newly floated corporations for the current stock round
+          @round.track_newly_floated(corporation) if @round.is_a?(GSystem18::Round::GotlandStock)
+        end
```
roseundy, line 370: map_gotland_float_corporation instead. You will have to add the framework to the game.rb file. That way future maps can provide their own float_corporation.

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
...
+            share = corp.treasury_shares&.first || share_pool.shares_by_corporation[corp]&.first
+            bundles << share.to_bundle if share
+          end
+          bundles
+        end
+
+        def process_buy_shares(bundle)
+          @log << "Rival buys one share from #{bundle.corporation.name}"
+          share_pool.buy_shares(rival, bundle, exchange: true)
+        end
```
roseundy, line 410: process_* methods should be in step/ files, not at this level. You need to create your own step/buy_sell_shares.rb that is gotland-specific.

lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -0,0 +1,535 @@
...
+
+        def process_sell_shares(shares)
+          shares = [shares] unless shares.is_a?(Array)
+          shares.group_by(&:corporation).each do |corp, corp_shares|
+            @log << "Rival sells #{corp_shares.sum(&:percent)}% of #{corp.name}"
+            corp_shares.each do |share|
+              share_pool.transfer_shares(share.to_bundle, share_pool, allow_president_change: false)
+            end
+          end
+        end
```
roseundy, line 428: See above.

roseundy:
You have merge conflicts. You need to merge with the latest code.

roseundy (changes requested):
Please make the changes I commented on above.

lib/engine/game/g_system18/round/gotland_stock.rb
roseundy: I've already created a System18 stock round that you can modify. It already overrides finish_round.

lib/engine/game/g_system18/step/token.rb
```diff
@@ -52,6 +52,8 @@ def process_place_token(action)
 
             place_token(entity, action.city, action.token,
                         same_hex_allowed: @game.token_same_hex?(entity, action.city.hex, action.token))
+
+            @game.bank.spend(-action.token.price, entity) if action.token.price.negative?
```
roseundy, line 56: Out of curiosity, what's this about? Why would the token price be negative?

lib/engine/game/g_system18/game.rb
patrikolesen: Updated

lib/engine/game/g_system18/map_base_customization.rb
patrikolesen: Yes it is needed since we now also have single player

lib/engine/game/g_system18/map_base_customization.rb
patrikolesen: Yes it is needed since we now also have single player

lib/engine/game/g_system18/round/gotland_difficulty_selection.rb
patrikolesen: We tried that first but it got a bit strange since we then had either multiple Gotland or extra options that only were applicable for Gotland so we ended up with this solution instead it was more intuitive for the player to choose this way

Or do you have any other suggestion?

lib/engine/game/g_system18/game.rb
patrikolesen: :all works for Gotland, that was what was added but also :one does seem to work that is part of the base. I do not remember why I added it. Now removed

lib/engine/ability/float_cost.rb
patrikolesen: Moved to system instead

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Removed

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Updated

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Updated

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Updated

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Updated

lib/engine/game/g_system18/step/token.rb
patrikolesen: It is a way to encourage the player to place tokens in a single player game. Instead of paying the corporation will receive money. Else it is very little that encourage you to token the cities and block the others

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Bad naming of the function, updated

lib/engine/game/g_system18/map_gotland_customization.rb
patrikolesen: Bad naming of the function, updated

lib/engine/game/g_system18/round/gotland_difficulty_selection.rb
roseundy: I think it's possible to create options that only apply when other options (i.e. maps for System18) are selected.

Not a big deal either way.

PR #11966, opened 2025-09-06
[18Norway] new privates

patrikolesen:
Added a few new privates from the authors

## Before clicking "Create"


patrikolesen:
Note this is a breaking change

PR #11969, opened 2025-09-07
[1824] Set game (and Cisleithania) in Alpha

perwestling:
This PR is depending on #11968

## Before clicking "Create"

## Implementation Notes

Depending on #11968 

Removed 1804 alpha notice. Or should this just add 1824 and let 1804 remain?

### Explanation of Change

With #11968 the 1824 implementation is good enough to play as Alpha.

### Any Assumptions / Hacks

I removed GAME:_IS_VARIANT_OF meta info, as that ment 1824 Cisleithania did not appear in the create game part. As it does not appear as a variant when creating 1824 I thought it was best. Or should it remains and something be changed in 1824 to make it work?

### DIFF
--- a/lib/engine/game/g_1824_cisleithania/meta.rb
+++ b/lib/engine/game/g_1824_cisleithania/meta.rb
@@ -12,9 +12,8 @@ module Engine
 
         DEPENDS_ON = '1824'
 
-        DEV_STAGE = :prealpha
+        DEV_STAGE = :alpha
 
-        GAME_IS_VARIANT_OF = G1824::Meta
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/1824-Cisleithania-map'.freeze
         GAME_TITLE = '1824 Cisleithania'.freeze
 
--- a/assets/app/view/welcome.rb
+++ b/assets/app/view/welcome.rb
@@ -16,7 +16,8 @@ module View
 
     def render_notification
       message = <<~MESSAGE
-        <p><a href="https://github.com/tobymao/18xx/wiki/1804">1804</a> is now in alpha.</p>
+        <p><a href="https://github.com/tobymao/18xx/wiki/1824">1824</a> is now in alpha, together with the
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_1824_cisleithania/meta.rb
```diff
@@ -12,9 +12,8 @@ module Meta
 
         DEPENDS_ON = '1824'
 
-        DEV_STAGE = :prealpha
+        DEV_STAGE = :alpha
```
michaeljb, line 16: > I removed GAME:_IS_VARIANT_OF meta info, as that ment 1824 Cisleithania did not appear in the create game part. As it does not appear as a variant when creating 1824 I thought it was best. Or should it remains and something be changed in 1824 to make it work?

The variant needs `GAME_IS_VARIANT_OF`, and the "parent" game needs the variant added under `GAME_VARIANTS`. So, change this back...

```suggestion

        GAME_IS_VARIANT_OF = G1824::Meta
```

...and in `g_1824/meta.rb` add something like this (most important bit here is that `title` is set correctly, the other fields can be changed how you like):

```ruby
        GAME_VARIANTS = [
          {
            sym: :cis,
            name: 'Cisleithania',
            title: '1824 Cisleithania',
            desc: 'Alternate map for 2-3 players',
          },
        ].freeze
```

michaeljb (changes requested):
> Removed 1804 alpha notice. Or should this just add 1824 and let 1804 remain?

1804's announcement has been up for two weeks, that's probably enough.

lib/engine/game/g_1824_cisleithania/meta.rb
perwestling: Have now applied your suggestions. And also tried various combination of optional rule (Goods Time 3-6 players for 1824) with base game / Cisleithania variant, and it worked fine.

New commit pushed.

perwestling:
@michaeljb Fixed after comments

PR #11996, opened 2025-09-12
[1824] Disallow upgrade of yellow one town to yellow two town #11993

### LINKED ISSUE #11993: [1824] double-dit tiles allowed to be played on single-dit cities
https://18xx.games/game/222944#map

EOD was allowed to place a double dit tile on C10, it should only be allowed to play a single dit one

### LINKED ISSUE #12002: [1824] Incorrect upgrade of double town tiles
**What's your Game ID?**

https://18xx.games/game/223118

SD1 gets upgrade options for the double town hex in F11.

Image

**How to Reproduce the Bug?**

During OR, build phase, click on an accessible yellow double town.

**Which Browser / Platform?**

Repeated in different versions

perwestling:
Corrected this bug by adding a check that upgrade of yellow town tile is only allowed it is a single town (both from and to) and to a green tile.

Fixes #11993 
Fixes #12002 

**Note! The bug led to an illegal upgrade in one game. That game will break and can be archived. Games without illegal upgrades will not be affected.**

## Implementation Notes

Code is a change from 1837 as they have different rules. But the check in 1824 did not check that
it was a single town upgrade to green. Have now added that to ensure only single towns can be upgraded.

### Explanation of Change

The bug made it possible to upgrade yellow single town to yellow double town, which is illegal.

### Screenshots

Double towns has now disappeared from chices

### Any Assumptions / Hacks

None

### DIFF
--- a/lib/engine/game/g_1824/game.rb
+++ b/lib/engine/game/g_1824/game.rb
@@ -196,10 +196,11 @@ module Engine
           :corporation
         end
 
-        # 1824 differ from 1837 as it allows any legal single town upgrade to green (duble towns have no green tiles)
+        # 1824 differ from 1837 as it allows any legal single town upgrade to green (double towns have no green tiles)
+        # TODO: Add a unit test for upgrade yellow single town to green, and no upgrade for double town
         def yellow_town_tile_upgrades_to?(from, to)
           # honors pre-existing track?
-          from.paths_are_subset_of?(to.paths)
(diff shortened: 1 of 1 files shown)


michaeljb (approved):

perwestling:
This correction also fixes the issue reported in #12002 

The fix is to check that upgrade is from yellow to green, and that the number of towns is the same both before and after upgrade.
1824 does not allow upgrade/downgrade of town hexes that change their number. And only single town hexes can be upgraded to green.

PR #11997, opened 2025-09-13
[System18] Add Russia map; move to beta; change SELL_AFTER

roseundy:
I don't think this will require additional pinning

## Before clicking "Create"

### Explanation of Change

Add Russia map and rules to System18
Move System18 to Beta
Move incremental cap games to `SELL_AFTER = :first`
Add more general support for game-specific stock rounds and/or steps
Add support for minors
Simplify maps w/o privates

### DIFF
--- a/lib/engine/game/g_system18/step/buy_sell_par_bid_merge.rb
+++ b/lib/engine/game/g_system18/step/buy_sell_par_bid_merge.rb
@@ -1,0 +1,193 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/buy_sell_par_shares_via_bid'
+
+module Engine
+  module Game
+    module GSystem18
+      module Step
+        class BuySellParBidMerge < Engine::Step::BuySellParSharesViaBid
+          # currently this is very specific to the Russia map (and cohorts)
+          MIN_BID = 100
+          MERGE_PHASE = 3
+          MAJOR_PHASE = 4
+          PURCHASE_ACTIONS = [Action::BuyCompany, Action::BuyShares, Action::Par, Action::Choose].freeze
+
+          def actions(entity)
+            return corporation_actions(entity) if entity.corporation?
+            return [] unless entity == current_entity
+            return %w[bid pass] if @auctioning
+
+            actions = []
+            actions << 'choose' if can_convert_any?(entity) || can_merge_any?(entity)
+            actions.concat(super)
+            actions << 'pass' if !actions.empty? && !actions.include?('pass')
+
+            actions
+          end
+
+          def setup
+            super
+            @chosen_corporation = nil
+          end
+
+          def win_bid(winner, _company)
+            entity = winner.entity
+            corporation = winner.corporation
+            price = winner.price
+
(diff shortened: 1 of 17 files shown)


lib/engine/game/g_system18/step/buy_sell_par_bid_merge.rb
```diff
@@ -0,0 +1,193 @@
...
+            return false if @game.phase.name.to_i < MERGE_PHASE
+            return false if bought?
+
+            @game.corporations.any? { |corporation| can_convert?(corporation) }
+          end
+
+          def connected?(minor_a, minor_b)
+            # Mergeable candidates must be connected by track, minors only have one token which simplifies it
+            # (partially borrowed from 1867)
+            # FIXME: avoid graph by using destination_connection
```
michaeljb: Any chance this could be addressed before merging?

roseundy: I've been struggling to figure how to do it at all. There are two difficulties with this. First is that merging is just one action of many in a stock round turn. This means that I can't just auto-insert a pass if a minor isn't connected to another minor. The second is that destination_connection just supplies a list of entities that have reached their destination. I really need a list of minors with each entry containing a list of minors that minor reaches. I would have to create a new action - and I'm really trying to avoid touching common code with all of these System18 maps.

In the end, I decided that the maps in System18 were small enough that graph wasn't nearly expensive as in a "big" 18xx.  Maybe I should just pull out the FIXME.

roseundy: I just realized this is even harder: when a major is started in a stock round, it can drop its home token in such a way that it disconnects formerly connected minors. This means any auto-action I generate in an OR is pointless, so any "connection" action I create would have to be a stock action.

roseundy: Hold on, I just realized something. I don't need to know whether a minor is connected in the buy_sell_par_bid_merge::actions method! The same action - "choose" is used for both converting and merging minors. Any minor that can be merged is always guaranteed to be able to convert. As long as I don't validate merge candidates calling graph will only occur as a result of rendering. So depending on how I coded this (I'll take a look in the morning), I don't think there is a performance issue at all here with graph.

lib/engine/game/g_system18/step/buy_sell_par_bid_merge.rb
```diff
@@ -0,0 +1,193 @@
...
+
+          def can_merge_any?(_entity)
+            return false if @game.phase.name.to_i < MERGE_PHASE
+            return false if bought?
+
+            !possible_mergers(current_entity).empty?
+          end
+
+          def corporation_actions(corporation)
+            return [] unless choice_available?(corporation)
```
roseundy, line 138: I just need to rewrite this to avoid calling graph.

roseundy:
Fixed the performance issue with calling graph unnecessarily.

Issue #11973, opened 2025-09-07
Fixture improvements


michaeljb:
rspecs for fixtures are inconsistently located and formatted, and repeat code for loading files

This is an old bugbear of mine. Some specs for fixtures are at https://github.com/tobymao/18xx/blob/master/spec/game_state_spec.rb, and a bunch more are under https://github.com/tobymao/18xx/tree/master/spec/lib/engine/games

I want them to be located in a consistent directory structure mirroring the main engine structure, e.g., #11949 

With #11970, after some restructuring of the `describe` blocks, a test can call `fixture_at_action()` to load and process a game fixture to the given point, and if the tests don't mutate the game state then it will cache the game and the next test can carry on from there, making them a little more performant.

Some other fixture-related PRs that have been finished:

* #11814 - rake task to format fixtures
  * #11873 - more required formatting
* #11948 - enforce using that rake task
* #11848 - rake task to make it easier to import a fixture file

Goals:
- [x] all specs that load fixtures and examine game state have consistent file structure
- [ ] add rubocop-rspec
- [ ] for local dev environment, add functionality in the browser UI to convert a hotseat game to a fixture
- [ ] for local dev environment, when at the `/fixture/<title>/<id>` route, add button(s) to update the saved fixture file
- [ ] `rake fixture_format` cleans up `undo` and `redo` actions, add enforcement of this
- [x] enforce that all beta+production titles have at least one fixture that reaches a normal endgame (i.e., is not ended by the `end_game` action)
- [x] enforce that production titles have a fixture reaching the endgame for each endgame reason? this would require additional code for the game to expose the reason it ended

PR #11998, opened 2025-09-13
[dev] More fixture formatting

michaeljb:
In #11948 the player regex was not strict enough, so not all usernames were removed from the fixtures.

This fixes the regex problem and also scrubs the fixture's `description` and `user`, and adds an
optional `fixture_format` key that can be used to keep the `user`, `description`, and `chat` as they
are, or keep the chat actions but replace their contents.

https://github.com/tobymao/18xx/issues/11973

The diff for the fixtures can actually be read seen locally with a CLI tool like [jq](https://jqlang.org/) installed and
this (or equivalent command with a different tool) added to your `.git/config` or `~/.git/config`:
```
[diff "json"]
	textconv = bash -c 'jq . \"$0\" -'
```

## Before clicking "Create"

### Any Assumptions / Hacks

Since this interacts with a large number of fixtures, I'll wait to merge this until there are few other open PRs,
and will avoid opening any new ones myself until this one is merged.


michaeljb:
I think some of the test failures here might be due to #12010

PR #12001, opened 2025-09-13
[1862] Initial commit to add solo variant to 1862 (EA)

perwestling:
This is a commit that adds the solo variant to 1862 as prealpha. It should not affect existing implementation of 1862.

## Before clicking "Create"

## Implementation Notes

This is an initial commit that just do some setup, and display IPO rows in SR. It is based on 1862 base game.

It is far from playable, so it is in prealpha.

Have not added permits a la 1862 solo yet (6 permits - 2 of each - are used, and corporation gets the first one when it is parred). Instead the base 1862 mechanism is used.

My idea is to continue with SR work to get that to work, before I make the next PR. But it requires some architectural decisions how to handle the SR actions specific to 1862 Solo before I can continue.

### Explanation of Change

Initial commit.

### Screenshots

Stock Market:
image

Timeline about IPO rows:
image

Entities view:
image

### Any Assumptions / Hacks

None

### DIFF
--- a/lib/engine/game/g_1862_solo/game.rb
+++ b/lib/engine/game/g_1862_solo/game.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative '../g_1862/game'
+require_relative 'meta'
+
+module Engine
+  module Game
+    module G1862Solo
+      class Game < G1862::Game
+        include_meta(G1862Solo::Meta)
+        CERT_LIMIT = {
+          1 => 25,
+        }.freeze
+
+        STARTING_CASH = {
+          1 => 1200,
+        }.freeze
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


roseundy (approved):
I think this looks fine.

lib/engine/game/g_1862_solo/game.rb
```diff
@@ -0,0 +1,215 @@
...
+          false
+        end
+
+        def setup_preround
+          @original_corporations = @corporations.dup
+
+          super
+
+          @original_corporations.reject { |c| @corporations.include?(c) }.each do |c|
+            hex = @hexes.find { |h| h.id == c.coordinates } # hex_by_id doesn't work here
```
roseundy, line 35: Out of curiosity, why doesn't this work here?

perwestling: @roseundy No idea - this code was copied from g_1862/game.rb and the comment is there in the code.

roseundy: That's funny. I must have written it :-)

PR #12014, opened 2025-09-15
[core] don't render prealpha game variants in production

michaeljb:
This will hide the 1862 "Solo" checkbox until it is alpha

## Before clicking "Create"


roseundy (approved):
LGTM

perwestling (approved):

PR #12037, opened 2025-09-20
[core] Create Game page: stop enforcing values in the input element

### LINKED ISSUE #11071: [iOS] setting playercount gets overriden by helper logic

michaeljb:
Instead, only validate when the game is actually created.

* simplify view logic to not have to enforce the inputs fitting the player range
* add `min_players` to base `Meta` module so the `respond_to?` check in `View::CreatGame` can be removed
* throw error in `Engine::Game::Base` if trying to create with invalid player count, so that the engine is not dependent on the view to catch problems with the request

Tested on iOS 18.6.2 in Firefox and Safari.

Fixes #11071

## Before clicking "Create"


lib/engine/game/base.rb
```diff
@@ -3419,6 +3421,16 @@ def second_icon(corporation); end
       def corp_loans_text
         'Loans'
       end
+
+      def check_player_range!
+        min_players = self.class.min_players(@optional_rules, @players.size)
+        max_players = self.class::PLAYER_RANGE[1]
```
ollybh: There are variants that limit the maximum number of players, so it would make sense to have a call to a `max_players` method here.

ollybh (approved):
This looks good. The next step could be to add a `max_players` method to the base Meta class, and having the base `min_players` and `max_players` method check the player ranges in any selected options.

Issue #12000, opened 2025-09-13
[various] Cash is occasionally misplaced, producing incorrect Bank totals


michaeljb:
While chipping away at #11973, I discovered that some fixtures have floating point totals in their `result`, and while tracking down the particular actions that caused cash totals to go from `Integer` to `Float`, I went another step further and added tests to assert that the total cash in a game remained constant.

I found some rogue cases where instead of `.spend()`, an entity's `.cash=` method was used, as in: 

https://github.com/tobymao/18xx/blob/987df2d6fc21282baf64eb0b78afa41c1ce99cd8/lib/engine/game/g_18_al/step/token.rb#L18

At https://18xx.games/game/220004?action=181, The Bank has $5591. The next action has L&N spend $40 to place a token, and then receive a $100 bonus, so The Bank's new total should be $5531, but it is $5631; `.cash +=` was used without updating The Bank's total, as if the president of the L&N slipped some cash from up their sleeve.

Another case is cash from bankrupted entities not being returned to The Bank, and instead being effectively dropped on the floor.

At https://18xx.games/game/219169?action=629, CNR has $372. After the next action, CNR has been nationalized but The Bank's cash does not increase by $372. Unfortunately, because the endgame is already triggered in that sample game, the Bank cash is hidden from the UI; #11999 includes a fixture derived from this game, with a test that fails by $372 if the `nationalization_transfer_assets()` change is removed.

 #11999 adds a test for one fixture checking the overall cash totals. After my current open PRs are merged, I'll have another one ready that will add a test for each fixture to verify the in-game cash totals, and will also make `cash=` in the [`Spender`](https://github.com/tobymao/18xx/blob/master/lib/engine/spender.rb) module `protected` method to prevent it from being called outside of `spend()`, guaranteeing that when an entity's cash total is adjusted the total cash in the game remains unchanged.

PR #12052, opened 2025-09-23
[1824/core] fix player cash when taking and paying off loans

### LINKED ISSUE #12051: [1824] player receives extra cash with loan, and does not actually pay back loan on pay off action
I found this while working on #12000. There are two problems:

1) when receiving a loan, the player receives the interest amount as additional cash
2) when paying off a loan, the player doesn't actually lose cash to pay off the loan

With the stack running locally, these links can be used to see the problems in the 1824 fixture `receivership_and_emergency_buy`:

* [action 394][394]: Player 3 has just taken a loan of 293G to buy a train, their debt has increased by 440G, and their cash total is 147G, but should be 0G. 147G is the amount of interest they owe on their loan.

* [action 538][538]: Player 3 has "paid off" 100G of their loan. Their loan value has decreased by 100G, but their cash total has not changed, and their value has increased by 100G.

* [action 539][539]: Player 3 has "paid off" the rest of their loan, 340G. Their loan value has gone to 0, but their cash total has not changed, and their value has increased by 340G.

[394]: http://localhost:9292/fixture/1824/receivership_and_emergency_buy?action=394
[538]: http://localhost:9292/fixture/1824/receivership_and_emergency_buy?action=538
[539]: http://localhost:9292/fixture/1824/receivership_and_emergency_buy?action=539

michaeljb:
Fixes #12051

[core] Also add a guard function `check_receiver()` in `spend()` to prevent the paying off
issue (where `entity.spend(cash, entity)` is called: `entity` paying itself for something)
from recurring anywhere else.

Fixture updates:

* `1824/receivership_and_emergency_buy`: `result.2` (the final score for `Player 3`)
  is changed from `7093` to `6506`, a difference of 587. The loan player 3
  took (action 394) was for 440, and they didn't actually lose cash when paying
  it back (actions 538 and 539). They also incorrectly received 147 in cash when
  taking the loan, further boosting their score. 440 + 147 = 587 ✅

* `18CO/22032`: action 646 has `ATSF` buy the share `CM_1`, which it had already bought
  at action 645, so ATSF pays itself and it's effectively a no-op. This was
  caugh by the added guard, `check_receiver()`. In the UI, I am unable to
  reproduce buying the share it already owns from itself, so no other bug fix is
  needed for this.

* `18JPT`: added two fixtures from another in-progress branch (see #11973,
  ensuring production games have fixtures) so that the change in its `setup`
  code from `@share_pool.buy_shares` to `@share_pool.transfer_shares` is
  verified by tests.

See also #12000

## Before clicking "Create"


michaeljb:
Going to split this into 2 PRs, one to just fix 1824's bug, and one to fix the other issues I found thanks to adding the `check_receiver()` guard.

PR #12054, opened 2025-09-23
[core] prevent `receiver` from being the same entity as `spender`

michaeljb:
On #12052 I added a guard to `spend()`, `check_receiver()`, to prevent the issue of the spender and receiver being the same entity from recurring. This ended up affecting more places than I expected, so I spun that off into this new branch and PR, and opened #12053 for just the 1824 change.

* 18CZ: in 2p, Vaclav's (the "third player"/bot entity) tile lays are free;
  make Vaclav the spender instead of the bank, and make Vaclav its own class to
  override `check_cash` allowing it to have negative cash

* 18JPT: use `transfer_shares()` instead of `buy_shares()` to move shares from IPO to bank pool during setup

Fixture updates:

* `18CO/22032`: action 646 has `ATSF` buy the share `CM_1`, which it had already bought at action 645, so ATSF pays itself and it's effectively a no-op. This was caught by the added guard, `check_receiver()`. In the UI, I am unable to reproduce buying the share it already owns from itself, so no other bug fix is needed for this.

* `18JPT`: added two fixtures from another in-progress branch (see #11973, ensuring production games have fixtures) so that the change in its `setup` code from `@share_pool.buy_shares` to `@share_pool.transfer_shares` is verified by tests.

See #12000 and #12051

## Before clicking "Create"

**Note:** this PR is currently pointing at the `1824-fix-loans` branch, because without the changes on that branch, `check_receiver()` will throw an error when paying off loans in 1824. Once #12053 is merged, this PR's branch can be udpated to `master`.


michaeljb:
@Zwergenpunk sorry, after I initially pushed this I found the Vaclav fix with the `free` kwarg on the tile laying stuff was breaking a lot of other tests, then I put this into "Draft" status.

I've redone that bit now with a change contained to 18CZ's code.

Zwergenpunk:
Infinite money Vaclav is also fine :)

> @Zwergenpunk sorry, after I initially pushed this I found the Vaclav fix with the `free` kwarg on the tile laying stuff was breaking a lot of other tests, then I put this into "Draft" status.
> 
> I've redone that bit now with a change contained to 18CZ's code.

PR #12057, opened 2025-09-23
[1824] Fix error with tokens, fixes #12017

### LINKED ISSUE #12017: [1824] Cisleithania: unable to build with UG1 from red offboard.
Game ID: 223273.

Trying to build out of UG1 during OR1.1. It is not possible to build on F13 or H11 - an error is raised:

Image

Note that SD2 has already built from Graz.

Link to game: [https://18xx.games/game/223273](https://18xx.games/game/223273)
Last action before this was 52.
Created a Hotseat game: [https://gist.github.com/alexarnon/987d3650844a5bd5bf996e1edb4e28fb](https://gist.github.com/alexarnon/987d3650844a5bd5bf996e1edb4e28fb)

**How to Reproduce the Bug?**

Build from UG1. Presence of adjacent track might be a factor.
Here is the state in the above game:

Image

Both F13 and H11 are not buildable (above error raised, then reset).

**Which Browser / Platform?**

Found on Firefox.

perwestling:
Fixes #12017

## Implementation Notes

N/A

### Explanation of Change

1824 place home token when corporation operates the first time. For some reason this caused an issue in 1824.

### Screenshots

N/A

### Any Assumptions / Hacks

Have added pending_tokens to 1824 round, to avoid getting the error.

### DIFF
--- a/lib/engine/game/g_1824/game.rb
+++ b/lib/engine/game/g_1824/game.rb
@@ -280,9 +280,8 @@ module Engine
         end
 
         def operating_round(round_num)
-          Engine::Round::Operating.new(self, [
+          G1824::Round::Operating.new(self, [
             G1824::Step::KkTokenChoice,
-            Engine::Step::HomeToken,
             G1837::Step::Bankrupt,
             G1824::Step::DiscardTrain,
             Engine::Step::SpecialTrack,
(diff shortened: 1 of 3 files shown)


perwestling:
I tested to apply this PR on the #12065 baseline and it still show just the same 2 games failing. So I think this change is OK now.
Am removing the archive_alpha_games label as I do not think this will cause any issues with existing games.

PR #12058, opened 2025-09-23
[18USA] Fix Western Land Grant bug.

MasN2:
Fix the Western Land Grant bug.

Fixes #[11992]

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_18_usa/step/acquire.rb
+++ b/lib/engine/game/g_18_usa/step/acquire.rb
@@ -31,6 +31,7 @@ module Engine
             end
           end
 
+
           def process_acquire(buyer)
             @passed_scrap_trains = false
             super


lib/engine/game/g_18_usa/step/acquire.rb
```diff
@@ -23,6 +23,14 @@ def pass_description
...
           end
 
+          def max_bid_for_corporation(corporation, acquired_corp)
+            if @game.owns_p15?(acquired_corp)
+              super + 100
+            else
+              super
+            end
+          end
+
```
michaeljb: rubocop found trailing whitespace on this line

```suggestion

```

lib/engine/game/g_18_usa/step/acquire.rb
```diff
@@ -23,6 +23,13 @@ def pass_description
             super
           end
 
+          def max_bid_for_corporation(corporation, acquired_corp)
+            if @game.owns_p15?(acquired_corp)
+              super + 100
+            else
+              super
+            end
+          end
```
michaeljb, line 32: Sorry, didn't realize my previous suggestion would delete the whole line. Rubocop's going to complain about no blank line between this function definition and the next one.

```suggestion
          end

```

Are you making these edits via the GitHub UI? If you want to make more robust contributions in the future, you should take a look at [DEVELOPMENT.md](https://github.com/tobymao/18xx/blob/master/DEVELOPMENT.md) and the [wiki](https://github.com/tobymao/18xx/wiki#developers), so that you can run `rake` and tick off those `Before clicking "Create"` checkboxes.

MasN2: I've just been using the Github UI, yes. I'm not a Ruby developer, so I don't think I'll be able to make major contributions to this site.

lib/engine/game/g_18_usa/step/acquire.rb
```diff
@@ -23,6 +23,15 @@ def pass_description
             super
           end
 
+          def max_bid_for_corporation(corporation, acquired_corp)
+            if @game.owns_p15?(acquired_corp)
+              super + 100
```
crericha, line 28: I believe this fix is too narrow in scope. The core issue is that the general buying power calculation is using the wrong number of available loans. The available loans method needs to be overridden in 18USA to include the western land grant.

Here is where the method is defined for 1817 (the parent class to 18USA) -
https://github.com/tobymao/18xx/blob/80732d5bdd3815ba39c1cd39e4fb2d0f1bb8cf82/lib/engine/game/g_1817/game.rb#L737

crericha: There may be some other specialty western land grant code that then needs to be updated based on available loans now including it.

crericha: Looked into this more and I didn't realize it was because the company will acquire the western land grant, but doesn't currently have it, so that's why it isn't included. This IS the right place to fix it.

crericha:
I wanted to clean up the PR and it was easier to open a new one, so closing this one. Thanks for the report and finding the right spot to fix it.

PR #12061, opened 2025-09-24
[System18] make all optional rules automatically mutex

michaeljb:
Just a small thing I thought of when I came across System18's `meta.rb` while looking at #11973 and #12000. Seems easier to deal with long-term.


roseundy (approved):
Yes, makes sense.

PR #12077, opened 2025-09-27
[System18] Update Gotland to not allow dumping to rival

patrikolesen:
It should not be allowed to dump to rivals

Might break existing games

## Before clicking "Create"


ollybh (approved):

roseundy (approved):
In the future, can you make sure I'm asked to review System18 changes?

ollybh:
> In the future, can you make sure I'm asked to review System18 changes?

@roseundy You can update [.github/CODEOWNERS](https://github.com/tobymao/18xx/blob/master/.github/CODEOWNERS) to automatically have any PR that touches the System18 files sent to you for review.

PR #12079, opened 2025-09-27
[1824] Add some shenanigans texts, fixes #12027

### LINKED ISSUE #12027: [1824] Strange message when disabling auto-actions after mining company exchanged
Auto-actions in a stock round are disabled when another player exchanges a mine company for the presidency of a regional railway.

It is right that this exchange is classed as a shenanigan and auto-actions are disabled. However the message isn't very informative:

> Disabled programmed action 'Buy BK from IPO until floated' due to 'Unknown action special_buy disabling for safety'

It would be better to implement a custom `action_is_shenanigan?` method and catch this type of action and output an improved log message.

perwestling:
Added shenanigans texts for all special actions that exists in 1824 stock round.

Fixes #12027

## Before clicking "Create"

## Implementation Notes

Have added shenanigan texts for the 3 extra actions available in 1824 SRs.

### Explanation of Change

See issue - strange log texts.

### Screenshots

N/A

### Any Assumptions / Hacks

N/A

### DIFF
--- a/lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
+++ b/lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
@@ -151,6 +151,14 @@ module Engine
             reserved.none? { |s| s.percent == 20 }
           end
 
+          def action_is_shenanigan?(_entity, _other_entity, _action, _corporation, _corp_buying)
+            return 'Exchange of Coal Minor' if action.is_a?(Action::SpecialBuy)
+            return 'Payoff of player debt' if action.is_a?(Action::PayoffPlayerDebt)
+            return 'Partial payoff of player debt' if action.is_a?(Action::PayoffPlayerDebtPartial)
+
+            super
+          end
+
           private
 
           def exchangable_ability(entity)


lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
```diff
@@ -151,6 +151,14 @@ def allow_president_change?(corporation)
             reserved.none? { |s| s.percent == 20 }
           end
 
+          def action_is_shenanigan?(_entity, _other_entity, _action, _corporation, _corp_buying)
```
ollybh:
```suggestion
          def action_is_shenanigan?(_entity, _other_entity, action, _corporation, _corp_buying)
```

ollybh: You're not testing any parameters passed to this method at the moment.

lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
```diff
@@ -151,6 +151,14 @@ def allow_president_change?(corporation)
...
+            return 'Exchange of Coal Minor' if action.is_a?(Action::SpecialBuy)
+            return 'Payoff of player debt' if action.is_a?(Action::PayoffPlayerDebt)
+            return 'Partial payoff of player debt' if action.is_a?(Action::PayoffPlayerDebtPartial)
+
+            super
```
ollybh: As you're repeatedly testing the class of the `action`, you could rewrite this as a `case` statement. I'm not sure if this is clearer or not.
```suggestion
            case action
            when Action::SpecialBuy
              'Exchange of Coal Minor'
            when Action::PayoffPlayerDebt
              'Payoff of player debt'
            when Action::PayoffPlayerDebtPartial
              'Partial payoff of player debt'
            else
              super
            end
```

ollybh: You could even use the compact form of the `case` statement.
```suggestion
            case action
            when Action::SpecialBuy then 'Exchange of Coal Minor'
            when Action::PayoffPlayerDebt then 'Payoff of player debt'
            when Action::PayoffPlayerDebtPartial then 'Partial payoff of player debt'
            else super
            end
```

ollybh (changes requested):
I don't think this will work at the moment, you're not testing any parameters passed to `action_is_shenanigan?`.

lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
perwestling: I do not think there is anything to test. What would you suggest?

ollybh: I meant that you've got a mistake in the parameter names: the third parameter is called `_action`. The code in the method is testing `action`, which is not the parameter passed to the method.

Actually, since you are calling `super`, none of the parameters should start with an underscore. The declaration should be:
```suggestion
          def action_is_shenanigan?(entity, other_entity, action, corporation, corp_buying)
```

ollybh:
@perwestling Please do not force-push once a pull request is under review, unless it is absolutely necessary. I think you have lost an earlier commit when doing this.

perwestling:
@ollybh In this case it was a commit that came from review comments, which was not a complete change, so I forced push to get it right.

ollybh:
@perwestling I thought you had changed the `action_is_shenanigan?` method to use a `case` statement. That commit has been lost when you did the force push.

perwestling:
I have now fixed that case, @ollybh.

PR #12080, opened 2025-09-27
[1824] Correct capitalization for floating, fixes #12306

perwestling:
A better algorithm is now used to calculate capitalization for float. This breaks quite a few games.

Have removed the remaining reference games, as they are broken by this fix. May add new reference games in a later commit.

Fixes #12306

## Before clicking "Create"

This fix might affect most games of 1824 and 1824 Cisleithania in the database, so archive it is.
Therefor I will set this as draft as I would prefer to make some more corrections before this correction is applied, to decrease disruption.

## Implementation Notes

Have a counter for how many shares (1 per 10%) should be used during float. This is set at setup, and modified when minors are unsold during SR1.

### Explanation of Change

By keeping track on how many reserved IPO shares (1 per 10%) it is possible to get a correct floating capitalization amount.
This is set to 6 SD, 7 KK & UG, 8 majors with associated coal minor, 10 the rest, and when floating is done this number is multiplied by par to get the capitalization.

If minors are unsold during SR1, the above number is reduced by two if the the exchangable share is a president share, and by one if a regular share.

### Screenshots

N/A

### Any Assumptions / Hacks

N/A


lib/engine/game/g_1824/game.rb
```diff
@@ -728,10 +728,11 @@ def float_minor!(minor, price)
         end
 
         # Modifed 1837 version as it will log incorrect received amount after
-        # formation + float.
+        # formation + float. Also use special float amount factor to handle unreserve of
+        # IPO shares, which caused a bug previously.
         def float_corporation(corporation)
           @log << "#{corporation.name} floats"
-          capitilization = corporation.par_price.price * corporation.total_ipo_shares
+          capitilization = corporation.par_price.price * corporation.float_amount_factor
```
ollybh: As the reserved shares are marked as not buyable, could you do something like this?
```suggestion
          unreserved_shares = corporation.shares.reject(:buyable).sum(&:percent) / 10
          capitilization = corporation.par_price.price * unreserved_shares
```
You then wouldn't need to track the `float_amount_factor`.

lib/engine/game/g_1824/corporation.rb
```diff
@@ -109,6 +121,19 @@ def receivership?
         def unpresidentable?
           player_share_holders.reject { |_, p| p < 20 }.empty?
         end
+
+        private
+
+        def do_unreserve_one_share!(president_share = false)
+          share_percent = president_share ? 20 : 10
+          a_share = shares.find { |s| !s.buyable && s.percent == share_percent }
```
ollybh:
```suggestion
          a_share = shares.find { |s| !s.buyable && s.president == president_share }
```

lib/engine/game/g_1824/corporation.rb
```diff
@@ -109,6 +121,19 @@ def receivership?
...
+          return unless a_share
```
ollybh: I would suggest throwing an error here if a reserved share hasn't been found. Something has gone badly wrong if that happens.

lib/engine/game/g_1824/game.rb
perwestling: The number of non-buyable cannot be used because shares are made buyable during exchange/merge. The code here does handle all cases (including unsold minors during SR1 and float after merge).

lib/engine/game/g_1824/corporation.rb
perwestling: Good idea

lib/engine/game/g_1824/game.rb
ollybh: Do you need to make the share buyable at that point?

perwestling: It need to be buyable.

In the first case the reserved share becomes a normal IPO share and can be bought from IPO during SR.
In the second case the reserved share becomes a normal share owned by a player and can be sold during SR.

There is a tricky detail in 1824 concerning SD1, UG1, KK1, the 20% in SD/UG/KK should be not be buyable. But Staatsbahn is formed, as soon as a player owns 20% or more, they will get the president's share.

In the current implementation this is handled by setting buyable to true, but then some special cases to handle so that it cannot be bought, and presidency handling. I have not looked into if the current implementation for handling this case can be simplified.

perwestling:
@ollybh I thought my solution was the simplest, as one at end of SR1 knows the amount to use for float. (In case of BH it can float during first SR so therefor 10 times par is default.)
If using some other methods one needs to handle the special cases:

- Unsold minors in SR1
- Non-floated regionals after coal minor exchange

Using remaining reserved works in most cases but not in the non-floated regionals as the float happens some time after exchange has been done.

E.g. CL exchange is done and no other share is sold - then CL will be at 20% sold and not floated. Later in the game 30% more is sold, which floats. At this point there is no information about reserved shares left, as presidency share has already been exchanged for.
And using a special exceptions for regionals does not help as a regional can become an unassociated regional in case **associated coal mine is not sold during SR1**.
For pre-staatbahn we have the case that the **primary minor is not sold**. This means the presidency share should not be counted when setting capitalization. In this case presidency share need to be not buyable as presidency share is decided after formation or at the latest when someone buys up to 20%.

As can be seen a couple of special cases. Which makes my solution much simpler - just set the future capitalization at the end of SR1, as we then know what this will be regardless of what happens later.

And I use factor as the factor is known, but the amount is not decided yet (for reginal without association) - for all others the amount (par times factor) is known.  But at float having factor and par makes it easy to calculate capitalization.

ollybh:
OK, you have persuaded me. One request, could you rename the variable `@float_amount_factor`? ‘Factor’ suggests that this is a ratio or percentage. Maybe call it `@capitalization_share_count`?

perwestling:
@ollybh Good point. I used your suggestion. Also review related comments and updated some.

perwestling:
It looks like this fix will break the 1824 bank reference game, so I will have to see if I can fix that issue.
Might need to rebase this one to latest master.

michaeljb:
> It looks like this fix will break the 1824 bank reference game, so I will have to see if I can fix that issue.
> Might need to rebase this one to latest master.

You could use [`find_game_end_fixtures.rb](https://github.com/tobymao/18xx/blob/master/scripts/find_game_end_fixtures.rb) to get a new one from your local DB:

```
docker compose exec rack rake irb
> load 'find_game_end_fixtures.rb'
> find_and_import_fixtures!('1824')
```

crericha (commented):
This basically looks like a dup of 1837/corporation.rb. The variable just has a different name.

crericha:
Also, the architecture is designed thus that the corporation class doesn't process actions, like unreserve share. It is basically a data class. The game class is supposed to do that.

perwestling:
> This basically looks like a dup of 1837/corporation.rb. The variable just has a different name.

Are you thinking about capitalization_share_count? Functionality differ quite a bit from 1837 as the number of reserved shares do not change dynamically in that title, which it does it 1824.

perwestling:
> Also, the architecture is designed thus that the corporation class doesn't process actions, like unreserve share. It is basically a data class. The game class is supposed to do that.

You've got a point, although it was little functionality added with this PR. It is not completely data class as floated are handled in the class and can change state with just a querry.

I thought it was handy to have some of the methods that change values in the class in the actual class. 
I might revisit corporation.rb in the future to refactor this a bit and move some to game.rb.

ollybh:
@perwestling The code in this PR now looks good to me, but you need to fix or replace the failing fixture before this can be merged.

lib/engine/game/g_1824/corporation.rb
```diff
@@ -109,6 +121,18 @@ def receivership?
         def unpresidentable?
           player_share_holders.reject { |_, p| p < 20 }.empty?
         end
+
+        private
+
+        def do_unreserve_one_share!(president_share = false)
+          a_share = shares.find { |s| !s.buyable && s.percent == (president_share ? 20 : 10)}
```
ollybh: Rubocop needs a space added.
```suggestion
          a_share = shares.find { |s| !s.buyable && s.percent == (president_share ? 20 : 10) }
```

ollybh:
@perwestling Can you add some new fixtures for 1824? The checks are failing at the moment because these have all been removed. Either manually choose some games, or use @michaeljb's `find_game_end_fixtures.rb` script.

perwestling:
@ollybh I lacked time to create a new fixture; all fixtures I tried was broken by this commit.

Also, I wanted to fix some other bugs that have occured. Having existing games for those would help to fix the bugs. But if these are archived due to this commit it will make them harder to fix.

But it looks like it was not as bad as I feared as quite a few 1824 games seem to have survied the fix.

So I guess it might have been good to take the plunge after all.

ollybh:
@perwestling This has not been deployed yet. I think all existing games will be broken by this change.

You could download a bunch of fixtures now. Or do you want to revert this PR before it is deployed?

Issue #8862, opened 2023-02-18
[1861] Game incorrectly ending when bank broke


OedipussyRex:
Not a game I was in, but https://18xx.games/game/112652 ended at the end of the OR when the bank broke before the first 8T was purchased, even though the RSR had the funds to buy the first 8T.

Relevant rule from the A-A edition rules:
"The game may also end if the Bank runs out of money
before the purchase of the first 8-train. If the Bank runs
out of money in a Stock Round, complete the Stock
Round and one Operating Round. If the Bank runs out of
money in an Operating Round, complete that Operating
Round. __In either case, if the 8-train is purchased before
the end of the Operating Round, continue with the extra
Rounds as described above.__"
(emphasis mine)

ianwilson156:
Actually, the GRR bought the first 8T in the same round (OR7.2).
There was a correct message about the game ending after OR8.3, but it ended immediately instead.

This is a repeat of issue #8541

ollybh:
This is slightly different from #8541, that was the bank breaking after the first 8T has been bought, this is the 8T being bought after the bank has broken.

@ianwilson156 From my reading of the AAG rules, the same behaviour should apply to 1867 as well as 1861, with the purchase of the 8T overriding the bank breaking. Right?

ianwilson156:
Yes, 1867 should be the same.

ollybh:
I've got a fix for this (ollybh/18xx@5705f27), but I need to test it more before I push this upstream, to make sure I'm not breaking any other game end conditions. I'll try and get this done in the next couple of days.

PR #12082, opened 2025-09-27
[dropped cash 2/5] Refactor game_end_check

michaeljb:

----

* add constant GAME_END_TIMING_PRIORITY to define priority of different timing values

* add constant GAME_END_LOCK_FIRST_REASON for 1822 and any other games where once the endgame is triggered in one way, other triggers going off can't change it

* add method `game_end_timing()` so a particular game end reason can have different timing conditions based on different circumstances

* add new methods for each game end reason, only call relevant ones instead of the whole suite; now the whole `game_end_check` method doesn't need to be overridden and reimplemented to change one small part of it, e.g., see the 1862 changes

* move logic for `:one_more_full_or_set` impacting the final turn into a new method `game_end_modify_final_turns!` which can be overridden for other endgame rules that mess with the number of rounds left

* refactor `end_now?` overrides that ignored the `after` value and were really checking if a `reason` had been met

### DIFF
--- a/lib/engine/game/g_18_usa/step/acquire.rb
+++ b/lib/engine/game/g_18_usa/step/acquire.rb
@@ -54,6 +54,10 @@ module Engine
           def can_scrap_train?(entity)
             return true if entity.corporation? && !@passed_scrap_trains && entity.trains.find { |t| @game.pullman_train?(t) }
           end
+
+          def max_bid_for_corporation(corporation, acquired_corp)
+            super + (@game.owns_p15?(acquired_corp) ? @game.loan_value : 0)
+          end
         end
       end
     end


lib/engine/game/base.rb
```diff
@@ -104,18 +104,36 @@ class Base
       # with reason => after
       #   reason: What kind of game end check to do
       #   after: When game should end if check triggered
-      # Leave out a reason if game does not support that.
-      # Allowed reasons:
-      #  bankrupt, stock_market, bank, final_train, final_phase, all_closed, custom
+      # Reasons implemented here:
+      #   - all_closed: all corporations and companies have closed
+      #   - bank: bank is broken
+      #   - bankrupt
```
crericha: Nit: Ad a description since all the others have one.

lib/engine/game/base.rb
```diff
@@ -104,18 +104,36 @@ class Base
...
-      #  immediate - ends in current turn
+      #  immediate - ends after current action
       #  current_round - ends at the end of the current round
       #  current_or - ends at the next end of an OR
       #  full_or - ends at the next end of a complete OR set
       #  one_more_full_or_set - finish the current OR set, then
       #                         end after the next complete OR set
       GAME_END_CHECK = { bankrupt: :immediate, bank: :full_or }.freeze
 
+      GAME_END_TIMING_PRIORITY = {
```
crericha: Why hash instead of an array where index is the priority?

lib/engine/game/base.rb
```diff
@@ -104,18 +104,36 @@ class Base
...
+        immediate: 0,
+        current_round: 1,
+        current_or: 2,
+        full_or: 3,
+        one_more_full_or_set: 4,
+      }.freeze
+
+      # Once a game end condition has been met, stop checking the other possible
+      # game end conditions
+      GAME_END_LOCK_FIRST_REASON = false
```
crericha: Nit: _CONDITION instead of _REASON?

lib/engine/game/base.rb
michaeljb: I wanted O(1) time instead of the O(N) time of `.index`. Admittedly I probably don't need to be worrying about that when the size of the array is usually going to be 2 or 3, and never more than like 5.

lib/engine/game/base.rb
michaeljb: The existing code tends to use `reason`, e.g.,

https://github.com/tobymao/18xx/blob/d018e7baf792369a6a17fc341261d051585e65e9/lib/engine/game/base.rb#L2826

So probably I should change my comments from using `condition` to `reason`?

lib/engine/game/base.rb
```diff
@@ -2807,37 +2825,55 @@ def game_end_check_values
...
-        %i[immediate current_round current_or full_or one_more_full_or_set].each do |after|
-          triggers.keys.each do |reason|
-            if game_end_check_values[reason] == after
-              @final_turn ||= @turn + 1 if after == :one_more_full_or_set
-              return [reason, after]
-            end
-          end
+        return @game_end_trigger if self.class::GAME_END_LOCK_FIRST_REASON && @game_end_trigger
+
+        triggers = game_end_check_values.each_with_object([]) do |(reason, _after), t|
```
crericha: Is it not possible to use `filter_map` here?

lib/engine/game/base.rb
```diff
@@ -2807,37 +2825,55 @@ def game_end_check_values
...
+          t << [reason, game_end_timing(reason)] if send("game_end_check_#{reason}?")
         end
+        return if triggers.empty?
 
-        nil
+        trigger = triggers.min_by { |_, after| self.class::GAME_END_TIMING_PRIORITY[after] }
+        game_end_modify_final_turns!(*trigger)
```
crericha: Nit: modify to me seems like it changes the content of the turns, where this is more of a set operation, no?

lib/engine/game/base.rb
michaeljb: Changing my comments to say `condition` instead.

lib/engine/game/base.rb
```diff
@@ -2807,37 +2825,55 @@ def game_end_check_values
...
+
+      def game_end_check_final_train?
+        @depot.empty?
+      end
+
+      def game_end_check_final_phase?
+        @phase&.phases&.last == @phase&.current
+      end
+
+      def game_end_check_custom?
```
crericha: Since you use `send` to call the game end checks, it'd be nice to get rid of custom (or discourage in the future) and instead have games define a specific check.

crericha (approved):
Looks good. I left some suggestions that I'll leave up to you if you want to act on them or not.

lib/engine/game/base.rb
michaeljb: The games that override it don't just set `@final_turn`, it's more open-ended depending on the game rules:
https://github.com/tobymao/18xx/blob/319aefbf76669605191010971823ed2c5d412491/lib/engine/game/g_1822_ca/game.rb#L535-L537

https://github.com/tobymao/18xx/blob/319aefbf76669605191010971823ed2c5d412491/lib/engine/game/g_1866/game.rb#L1012-L1018

https://github.com/tobymao/18xx/blob/319aefbf76669605191010971823ed2c5d412491/lib/engine/game/g_18_royal_gorge/game.rb#L1037-L1039

I don't really like the name I landed on though. Maybe `game_end_set_final_turn!` is still fine even if the overrides aren't expected to literally set `@final_turn`?

michaeljb: Going with `game_end_set_final_turn!`

lib/engine/game/base.rb
michaeljb: Getting rid of `custom`

lib/engine/game/base.rb
crericha: I personally tend to err on the side of maintainability when the performance gains are minimal. In this case, if I wanted the O(1), I'd construct the hash from the array at init time, giving the best of both worlds at the cost of a meager amount of memory.

lib/engine/game/base.rb
crericha: Yeah, using `reason` everywhere is fine too 👍

lib/engine/game/base.rb
crericha: I think its good enough. 👍

lib/engine/game/base.rb
michaeljb: Changed to an array using `.index`

roseundy (approved):

Zwergenpunk (approved):

lib/engine/game/g_18_ms/game.rb
```diff
@@ -157,11 +157,7 @@ class Game < Game::Base
         ].freeze
 
         # Game will end after 10 ORs (or 11 in case of optional rule) - checked in end_now? below
```
ollybh, line 159: This comment is now out-of-date, there is no longer an `end_now?` method.
```suggestion
        # Game will end after 10 ORs (or 11 in case of optional rule)
```

ollybh (commented):
You haven't touched the 1861/1867 game end code, but it looks like this could be simplified after these changes. These currently need some custom code because the game will end at the end of the current OR if the bank breaks – except when the final phase is reached after the bank is broken, then you get another set of ORs. Would setting `GAME_END_TIMING_PRIORITY = %i[one_more_full_or_set current_or]` work for these games?

michaeljb:
> You haven't touched the 1861/1867 game end code, but it looks like this could be simplified after these changes. These currently need some custom code because the game will end at the end of the current OR if the bank breaks – except when the final phase is reached after the bank is broken, then you get another set of ORs. Would setting `GAME_END_TIMING_PRIORITY = %i[one_more_full_or_set current_or]` work for these games?

Oh yeah, forgot to mention this. The problem I ran into here was I couldn't find in the DB (including an older copy of the DB I had on my computer) a single game of 1861 or 1867 that ended with the bank running out instead of ending on the final phase. Didn't want to mess with that code without having test coverage for it.

I guess I could make a script to look for 1861/1867 games that have the bank running out before the final phase, even if the game does ultimately end due to the final phase.

lib/engine/game/base.rb
michaeljb: I refactored this some more, with validating against the full DB I discovered some more bugs. Specifically there was a problem in 1822 games if the bank runs out and a corp reaches the end of the stock market on the same action.

ollybh:
> Oh yeah, forgot to mention this. The problem I ran into here was I couldn't find in the DB (including an older copy of the DB I had on my computer) a single game of 1861 or 1867 that ended with the bank running out instead of ending on the final phase. Didn't want to mess with that code without having test coverage for it.

Game [112652](https://18xx.games/game/112652) was the game which reported the end conditions not working when the final phase is reached after the bank breaks (#8862). Have you got a DB copy with that game in? It's archived now.

michaeljb:
> Game [112652](https://18xx.games/game/112652) was the game which reported the end conditions not working when the final phase is reached after the bank breaks (#8862). Have you got a DB copy with that game in? It's archived now.

No, the oldest one I have on my computer is from June 2024, and that game is from February 2023 so it was already archived 😞

michaeljb:
> > Game [112652](https://18xx.games/game/112652) was the game which reported the end conditions not working when the final phase is reached after the bank breaks (#8862). Have you got a DB copy with that game in? It's archived now.
> 
> No, the oldest one I have on my computer is from June 2024, and that game is from February 2023 so it was already archived 😞

Created #12105 to not lose track of this.

ollybh (changes requested):
This is currently breaking 1861/1867, so I'm setting this to changes required until #12105 is fixed.

PR #12083, opened 2025-09-27
[dropped cash 3/5] add `game_end_reason` to Game object, add tests requiring it for fixtures

michaeljb:

----

* requires `end_game!()` to be called with the reason for ending it
* expose the reason `.game_end_reason`
* require fixtures to have a `game_end_reason` key
    * require alpha games to have at least one fixture with a `game_end_reason` that is not `manually_ended`
    * require beta/production games to have at least one fixture for each `game_end_reason`
    * exceptions do currently exist in constants at the top of `spec/lib/engine/fixtures_required_spec.rb`; no examples for those title/game end reason combinations were found in the real DB

### DIFF
--- a/lib/engine/game/g_18_usa/game.rb
+++ b/lib/engine/game/g_18_usa/game.rb
@@ -924,6 +924,7 @@ module Engine
         def add_subsidy(corporation, hex)
           return unless (subsidy = @subsidies_by_hex.delete(hex.coordinates))
 
+          @log << "#{corporation.name} receives the #{subsidy[:name]}" unless subsidy[:name] == 'No Subsidy'
           hex.tile.icons.reject! { |icon| icon.name.include?('subsidy') }
           subsidy_company = create_company_from_subsidy(subsidy)
           assign_boomtown_subsidy(hex, corporation) if subsidy_company.id == 'S8'


lib/engine/game/base.rb
```diff
@@ -1367,11 +1369,12 @@ def discard_discount(train, price)
...
       end
 
-      def end_game!(player_initiated: false)
+      def end_game!(game_end_reason)
         return if @finished
 
         @finished = true
-        @manually_ended = player_initiated
+        @game_end_reason = game_end_reason
+        @manually_ended = @game_end_reason == :manually_ended
```
crericha: Do we need attr if the reason now states this? I don't think there are many places where we check manually_ended. ELO ratings is one...probably others.

crericha (approved):

lib/engine/game/base.rb
michaeljb: Yeah I wanted to leave it alone for now for a few reasons
* storing `game_end_reason` in the DB would require adding a new column to the DB
* some kind of migration would be needed to add `game_end_reason` to already-completed in the DB
* we'd need a value for already-archived games in the DB, their actual `game_end_reason` cannot be computed now

I do want to do that hopefully soon after this is merged. I added a `dnf` reason for 1822 games where nothing is bid on in the first SR (game ends by RAW, but is everyone saying `pass` once *really* playing a game of 1822?), and a game of 1846 actually ending on `all_closed` or `turn_100` is a little sus, so I'd like to exclude those from ELO as well.

crericha: Okay, I'm good with migrating the db as a follow-up PR. I'd like to exclude those from ELO as well.

perwestling: Will be exclusion of game end results be defined per title?

roseundy (approved):

Zwergenpunk (approved):

ollybh (approved):

lib/engine/game/base.rb
michaeljb: > Okay, I'm good with migrating the db as a follow-up PR. I'd like to exclude those from ELO as well.

I've created #12094 to track this.

> Will be exclusion of game end results be defined per title?

I wouldn't think this is necessary, going by the `game_end_reason` should be enough.

michaeljb:
While github has marked this PR as merged, `game_end_reason` code changes have not been merged into `master`. More in https://github.com/tobymao/18xx/pull/12086#issuecomment-3369241826

PR #12084, opened 2025-09-27
[dropped cash 4/5] fix player loans, refactor player loans into base engine objects

michaeljb:

----

I found some weird behavior and comments--plus a lot of duplicated code--around
player loans while I was working on other parts of #12000. It seemed best to
handle the loan-specific changes separately from the rest.

* `G1844:Player` was added only for tracking `@debt`; move that onto
  `Engine::Player`, and add `@penalty` for debt which counts against the
  player's score but can never be repaid

* add debt and loan methods to `Spender`

* add `@debt` to the Bank, so that all debt given to players can be tracked; add
  test for each fixture to ensure the net debt is 0

* add methods and constants to `Game::Base` to manage player loans

Notable assumptions:

1) loans are not really outside money

Loans weren't always taken from the bank; there was a comment copy-pasted
numerous times that "the money from loans is outside money, doesnt [sic] count
toards the normal bank money." I think this stemmed from a misinterpretation of
1822's rule 4.3.5, which states "The sum of any loans in play does not count
towards the bank limit of £12,000 (i.e. treat the loan as coming from a separate
source, not the bank)."

In 1822, when a player receives a loan, they get cash from the bank, and then
immediately spend that cash (back to the bank) on a train. When playing at a
physical table, an extra stack of chips is often used to track the remaining
balance of the player's debt. The important point is that the stack does not
count as coming from the bank, even though the initial cash in the loan did come
from the bank.

Now the amount owed by the player is tracked on their `@debt`, and the `Bank`
uses its own `@debt` to also track the total amount owed it; the sum of each
player's `@debt` and the `Bank`'s `@debt` should always be zero

2) 18ESP doesn't need to track interest separately from the debt

18ESP had code to track debt and interest separately, and code for some strange
math to apply some of a payment to the principal and some to the interest, but
that code never apparently ran; the UI does not present an optional to partially
pay off the debt if the player's cash is less than the total debt amount.

https://github.com/tobymao/18xx/blob/d96bdc6ddd61aff5f7ea5793b3a7d18b58ec21c7/lib/engine/game/g_18_esp/game.rb#L1058-L1077

### DIFF
--- a/lib/engine/game/g_18_ny/step/buy_train.rb
+++ b/lib/engine/game/g_18_ny/step/buy_train.rb
@@ -119,6 +119,13 @@ module Engine
             @train_salvaged = true
             @game.salvage_train(action.train)
           end
+
+          def help
+            return if !@round.active_train_loan && !can_take_loan?(current_entity)
+
+            'Once a loan is taken, a train must be purchased. Use the undo button to return the loans' \
+              ' in order to pass without purchasing a train.'
+          end
         end
       end
     end


assets/app/view/game/player.rb
```diff
@@ -142,17 +142,17 @@ def render_info
...
+        if @player.debt.positive?
           trs << h(:tr, [
             h(:td, 'Loan'),
-            h('td.right', @game.format_currency(@game.player_debt(@player))),
+            h('td.right', @game.format_currency(@player.debt)),
           ])
         end
 
-        if @game.respond_to?(:player_loans)
+        if @player.permadebt.positive?
```
crericha: Nit: perma_debt or permanent_debt?

lib/engine/game/base.rb
```diff
@@ -287,6 +288,20 @@ class Base
       # loans taken during ebuy can lead to receviership
       EBUY_CORP_LOANS_RECEIVERSHIP = false
 
+      # true - no restrictions
+      # :after_sell - must sell shares before taking loan
+      # :no_sell - cannot take loan after selling shares
```
crericha: should false be in the comments of allowed values?

lib/engine/game/base.rb
```diff
@@ -287,6 +288,20 @@ class Base
...
+      EBUY_CAN_TAKE_PLAYER_LOAN = false
+
+      # When a player takes a loan, this amount (as a percentage of the loan) is
+      # immediately added to the player's total debt. It can be repaid along
```
crericha: Its also added in stock rounds if not paid off. Maybe make the comment less detailed...I'm not sure the detail is needed.

lib/engine/player.rb
```diff
@@ -25,7 +25,7 @@ def initialize(id, name)
     end
 
     def value
-      @cash + shares.select { |s| s.corporation.ipoed }.sum(&:price) + @companies.sum(&:value)
+      @cash + shares.select { |s| s.corporation.ipoed }.sum(&:price) + @companies.sum(&:value) - debt - permadebt
```
crericha: I'm confused on the difference between debt and permadebt if they both subtract from value.

lib/engine/spender.rb
```diff
@@ -37,5 +45,58 @@ def spend(cash, receiver, check_cash: true, check_positive: true, borrow_from: n
 
       receiver.cash += cash
     end
+
+    def take_cash_loan(cash, bank, lender, interest: 0, permadebt: 0)
```
crericha: Should bank be part of init? Feels weird to pass it in unless some game changes the lender source

lib/engine/spender.rb
```diff
@@ -37,5 +45,58 @@ def spend(cash, receiver, check_cash: true, check_positive: true, borrow_from: n
...
+
+      total_debt = cash + interest_amount(cash, interest)
+      self.debt += total_debt
+      lender.debt -= total_debt
+
+      total_permadebt = interest_amount(cash, permadebt)
+      self.permadebt += total_permadebt
+      lender.permadebt -= total_permadebt
+
+      [cash, total_debt, total_permadebt]
```
crericha: I'm not a fan of returning an array of values. You always have to look back at this method to remember the order. I feel like hashes are easier to manage as return values.

assets/app/view/game/player.rb
michaeljb: changed to `permanent_debt`

lib/engine/step/train.rb
```diff
@@ -262,6 +262,22 @@ def can_finance?(_entity)
         false
       end
 
+      def try_take_player_loan(entity, cost)
```
crericha: Is it truly only the train step right now? I'm not sure if there are any games with company loans and player loans, where the player may be responsible for paying the company loan interest if the company is broke.

lib/engine/player.rb
michaeljb: `permanent_debt` cannot be repaid. I think most games with player loans have either debt that can either be repaid or debt that can never be repaid.

18CZ (and maybe others? idk) has both: when you take a loan of, say, $100, you owe $100 back before you can buy shares, but instead of interest that applies right away or accrues in stock rounds, there is a one-time penalty interest equal to the amount of the loan that counts against your endgame score, so you also lose $100 at the end of the game.

Maybe `game_end_debt` instead of `permanent_debt`?

lib/engine/step/train.rb
michaeljb: Right now on master it is only in the train step, with a stub in the base step - https://github.com/search?q=repo%3Atobymao%2F18xx%20try_take_player_loan&type=code

perwestling (commented):
No comments by me

lib/engine/player.rb
crericha: Hmm...for CZ, the end game value doesn't feel like debt (since you pay off the debt). Its more like a penalty. I wonder if you make it like that, you can simplify to one debt value and use config to define if that debt is payable or not.

crericha: Basically, the code is much more complex to support that one game and I wonder if its better handled within that game rather than generically as a one-off.

roseundy (approved):

lib/engine/player.rb
michaeljb: I've refactored `permanent_debt` into a simpler `penalty`.

lib/engine/spender.rb
michaeljb: I tried doing this, but ran into headaches with the variety of ways and places that the various corporation/company types can be initialized. Ultimately it's a lot simpler to just pass the `Game`'s `@bank` into this function.

I did do another refactor to just get rid of the `Lender` and just use the `Bank`'s `@debt`.

Zwergenpunk (approved):

crericha (approved):
Overall, looks good! 

Thinking of the loan penalty as interest is a little weird imho as its only applied once. I wonder if there is better terminology that is easier to reason about.

michaeljb:
> Thinking of the loan penalty as interest is a little weird imho as its only applied once. I wonder if there is better terminology that is easier to reason about.

Renamed `PLAYER_LOAN_PENALTY_INTEREST` to `PLAYER_LOAN_ENDGAME_PENALTY`

michaeljb:
While github has marked this PR as merged, the loan refactoring has not been merged into `master`. More in https://github.com/tobymao/18xx/pull/12086#issuecomment-3369241826

PR #12085, opened 2025-09-27
[dropped cash 5/5] ensure all money can be tracked correctly

michaeljb:
See full PR

----

* add `Game` method `spenders`; this defines the complete list of entities which have included `Spender` and hold all of the cash in the game

   * all transactions should be zero-sum; if the sum total of their `cash` is ever not equal to `bank_starting_cash`, that is analogous to either dropping some cash on the floor, or someone pulling extra cash out of their sleeve

    * make `cash=()` protected in `Spender` module; this prevents other parts of the code from calling things like `.cash += 100`, so they must call `spender()` which will guarantee that both entities have their cash mutated appropriately

    * allow an entity to define something else as their `spender`, e.g., the train Depot's spender is The Bank

    * add test for each fixture to make sure endgame cash totals match the starting cash

* add new methods `bank_starting_cash` and `init_bank_kwargs` for fewer direct calls to `Bank.new`, and to expose the starting cash so it can be verified in tests

* override new `Game` and `Spender` methods where needed

    * `spenders`

    * `init_bank_kwargs`

    * `bank_starting_cash`

    * 1868 Wyoming - set `spender` instead of overriding `cash` on coal/oil companies

    * 18 India - set `@ipo_pool`'s `spender`

* use `Spender` methods to mutate cash in tests

* add `fixture_every_action_debug.rb` for debugging

    * this test is useful for identifying exactly which action caused a broken game state, but is far too costly to have as regular part of the tests

    * it is not included when rspec runs because its name does not end in `_spec.rb`


lib/engine/spender.rb
```diff
@@ -4,32 +4,35 @@
...
-      available = @cash + (borrow_from ? borrow_from.cash : 0)
-      raise GameError, "#{name} has #{@cash} and cannot spend #{amount}" if (available - amount).negative?
+    def cash
+      spender == self ? @cash : spender.cash
     end
 
-    def check_positive(amount)
-      raise GameError, "#{amount} is not valid to spend" unless amount.positive?
+    def debt
+      spender == self ? (@debt || 0) : spender.debt
```
crericha: Is there a reason not to have @debt protected too?

lib/engine/spender.rb
```diff
@@ -38,12 +41,16 @@ def spend(cash, receiver, check_cash: true, check_positive: true, borrow_from: n
       if borrow_from && (cash > @cash)
         amount_borrowed = cash - @cash
         @cash = 0
-        borrow_from.cash -= amount_borrowed
+        borrow_from.spender.cash -= amount_borrowed
```
crericha: Should this be accessing the `borrow_from` cash directly? I'm not super familiar with this code, but this felt off relative to the rest of the code.

lib/engine/spender.rb
```diff
@@ -38,12 +41,16 @@ def spend(cash, receiver, check_cash: true, check_positive: true, borrow_from: n
...
       else
         @cash -= cash
       end
 
-      receiver.cash += cash
+      receiver.spender.cash += cash
+    end
+
+    def set_cash(cash, other_spender)
+      other_spender.spend(cash - self.cash, self, check_cash: false, check_positive: false)
```
crericha: Are we sure negative cash is always zero sum? I think its probably right.

lib/engine/spender.rb
```diff
@@ -77,8 +84,23 @@ def repay_cash_loan(bank, lender, payoff_amount: nil)
...
     protected
 
+    # This is protected so that only `spend()` can call this directly, to ensure
+    # that no money is ever "dropped on the floor", or conjured from nothing
+    # when The Bank should be used.
+    def cash=(cash)
+      if spender == self
+        @cash = cash
+      else
+        spender.cash = cash
```
crericha: Shouldn't spender handle their cash? Maybe I'm misunderstanding the spender class vs the spender entity.

lib/engine/spender.rb
```diff
@@ -95,6 +117,19 @@ def lender?
 
     private
 
+    def check_cash(amount, borrow_from: nil)
+      available = @cash + (borrow_from ? borrow_from.cash : 0)
+      raise GameError, "#{name} has #{@cash} and cannot spend #{amount}" if (available - amount).negative?
+    end
+
+    def check_positive(amount)
+      raise GameError, "#{amount} is not valid to spend" unless amount.positive?
```
crericha: I suggest a more specific error message here, so its easier for devs to track down.

lib/engine/spender.rb
```diff
@@ -95,6 +117,19 @@ def lender?
...
+    end
+
+    def check_receiver(cash, receiver)
+      raise GameError, "Cash receiver must be a different entity: #{name}.spend(#{cash}, #{receiver.name})" if receiver == self
```
crericha: Cash spender and receiver must be different entities

lib/engine/spender.rb
michaeljb: Yeah it's safe to do the `borrow_from.cash` directly. This `spender` method is mainly for things like letting the train Depot set `spender = @bank` and getting rid of this code:

https://github.com/tobymao/18xx/blob/d018e7baf792369a6a17fc341261d051585e65e9/lib/engine/depot.rb#L138-L144

It wouldn't make sense to ever be borrowing from the Depot or a similar entity.

michaeljb: Also the `cash` and `cash=` methods take care of the `spender` call, so it's really not needed.

lib/engine/spender.rb
michaeljb: I do have `debt=` protected like `cash=`. The public `debt` and `cash` methods here are wrappers similar to `attr_reader`

roseundy (approved):

Zwergenpunk (approved):

lib/engine/spender.rb
michaeljb: The latest version of `spender.rb` is hopefully a little simpler now, referencing `@spender` directly instead of going through the `spender` method, and fewer references to `self`. Hopefully this is more clear now.

If an entity defines its own `spender` method, it can now be the target of spending:

* the [train Depot](https://github.com/tobymao/18xx/pull/12085/files#diff-f88b61fa025bda544ab13ab727d060b84028d641d7a493a038c4574d5d801ba1) now uses a `spender` method so that its cash transactions are actually handled by the `Bank` instead of having cash methods that just access the `Bank`'s cash; the `Depot` is passed in to `spend` in `buy_train()` as the `train.owner` https://github.com/tobymao/18xx/blob/d018e7baf792369a6a17fc341261d051585e65e9/lib/engine/game/base.rb#L2107
 
In a couple cases, there are entities that are built with classes that include `Spender`, but they are not supposed to be responsible for their own cash:

* in 1868 Wyoming, coal and oil companies have operating steps to place tokens and are implemented as Minors; however they have no cash of their own and just use the player cash
* in 18 India, there is a special share holder entity used to help manage shares in the IPO pool; cash paid for shares ends up there but should go the `Bank`, and now with its `@spender` sent it will correctly go to the `Bank`

crericha: Okay, I follow now. 

I wonder though why we need to do things like `@spender&.cash || @cash`. Couldn't the mixin do something like:

```
def spender
  self
 end
 ```
 
Then wherever you need to access spender, you can do `spender.cash`. It'll either be self or overwritten to be a different `spender`. imho, that's much simpler to work with because you don't care if another spender is being used or not.

crericha (approved):

michaeljb:
While github has marked this PR as merged, the money tracking code changes have not been merged into `master`. More in https://github.com/tobymao/18xx/pull/12086#issuecomment-3369241826

PR #12092, opened 2025-09-28
[1824] Corrections for 2-player variant, fixes #12030

### LINKED ISSUE #12030: [1824] Bank pool in Cisleithania
This might be more of feature request than a bug.

According to Lonny in Cisleithania there should be a bank pool, and it may contain 100% of a corporation.

I have asked Lonny if this applies to both 2 player and 3 player.

Also, does this mean that you can sell president's share? If so, how to handle Receivership for regular corporations?

perwestling:
Some corrections for 2-player variant:
- Do include market percent when selling shares
- Update finish text for SR1
- Make bond railway shares exchangable (by MR exchange)
- Update stack titles during drafting

Fixes #12030 

Unlikely this will break any 2 player game, but there is a small possibility for it to affect SR in some existing 2p game.

## Implementation Notes
N/A

### Explanation of Change
See commit

### Screenshots
N/A

### Any Assumptions / Hacks
N/A


lib/engine/game/g_1824_cisleithania/game.rb
```diff
@@ -453,25 +459,37 @@ def select_randomly(collection)
...
         def closed_construction
           @close_construction_company_when_first_5_sold ? '5' : '4'
         end
 
-        def create_construction_railway_from_bought_pre_staatsbahn(company, minor)
-          make_minor_construction_railway(minor)
+        def make_minor_construction_railway(minor)
+          @log << "#{minor.name} becomes a construction railway. "\
+                  'It returns its cash to the bank as it does not use any money.'
+          minor.spend(minor.cash, @bank)
```
ollybh: Is it possible that the minor will have no cash? You will get an error.

lib/engine/game/g_1824/game.rb
```diff
@@ -498,7 +498,11 @@ def kk
         end
 
         def exchangable_for_mountain_railway?(player, corporation)
-          corporation.type == :major && @companies.find { |c| mountain_railway?(c) && c.owned_by?(player) }
+          shares_exchangable?(corporation) && @companies.find { |c| mountain_railway?(c) && c.owned_by?(player) }
```
ollybh: You're not using the return value from `@companies.find` except to see if anything was found. `any?` is the correct way to test this.
```suggestion
          shares_exchangable?(corporation) && @companies.any? { |c| mountain_railway?(c) && c.owned_by?(player) }
```

lib/engine/game/g_1824_cisleithania/game.rb
ollybh: There's a typo in this method name, it should be `buy` not `by`.
As this method returns a boolean, it would be better to have a name that ends in a question mark.

lib/engine/game/g_1824_cisleithania/game.rb
```diff
@@ -376,10 +381,11 @@ def special_handling_after_by_company(company, minor)
 
           # Need to handle construction railways when two player variant
           if pre_staatsbahn?(minor)
-            create_construction_railway_from_bought_pre_staatsbahn(company, minor)
+            handle_unreserve_of_pre_staatsbahn(company)
           else
-            create_construction_railways_from_coal_mine(company, minor)
+            create_bond_railway(company, minor)
```
ollybh, line 390: I don't like the structure of this code. This method is returning a boolean value to determine whether the last section of code in `after_buy_company` is executed, but you are also making changes to companies/corporations in the method. Can this block be moved to a separate method, called from `after_buy_company`?

lib/engine/game/g_1824_cisleithania/game.rb
```diff
@@ -453,25 +459,37 @@ def select_randomly(collection)
...
+          case company.stack
+          when 1
+            'Construction Railway (Stack 1)'
+          when 2
+            'Large Pre-staatsbahn (Stack 2)'
+          when 3
+            'Small Pre-staatsbahn (Stack 3)'
+          when 4
+            'Coal Railway (Stack 4)'
+          end
```
ollybh, line 469: Style suggestion: very simple `case` statements like this can be more readable if you use the single line format for the `when` tests.
```suggestion
          case company.stack
          when 1 then 'Construction Railway (Stack 1)'
          when 2 then 'Large Pre-staatsbahn (Stack 2)'
          when 3 then 'Small Pre-staatsbahn (Stack 3)'
          when 4 then 'Coal Railway (Stack 4)'
          end
```

lib/engine/game/g_1824_cisleithania/game.rb
```diff
@@ -453,25 +459,37 @@ def select_randomly(collection)
...
 
-        def create_construction_railway_from_bought_pre_staatsbahn(company, minor)
-          make_minor_construction_railway(minor)
+        def make_minor_construction_railway(minor)
+          @log << "#{minor.name} becomes a construction railway. "\
+                  'It returns its cash to the bank as it does not use any money.'
+          minor.spend(minor.cash, @bank)
+          minor.add_ability(free_tile_lay_ability)
+          minor.make_construction_railway!
+        end
```
ollybh: Could this code be moved into `G1824::Corporation.make_construction_railway!`?

lib/engine/game/g_1824_cisleithania/game.rb
perwestling: The pre-staatsbahn gets the price paid as starting money. This is the pre-staatsbahn drafted from stack 1 during the 4 first turns in SR1.

lib/engine/game/g_1824_cisleithania/game.rb
perwestling: Have now refactored this a bit.

lib/engine/game/g_1824_cisleithania/game.rb
perwestling: Have now moved it to G1824Cisleithania::Corporation.make_construction_railway!

lib/engine/game/g_1824_cisleithania/game.rb
perwestling: Indeed it would be. But have removed the method after refactoring (see below)

lib/engine/game/g_1824_cisleithania/game.rb
perwestling: Have now refactored it a bit. Goal is to avoid any Cisleithania knowledge in the 1824 game.rb. Have introduced a new method which is implemented in both 1824 and Cisleithania, replacing the previous construction.

perwestling:
@ollybh Ready for new review.

perwestling:
I created a baseline of all 1824 fixes up to 2025-10-12, and then looked into #12127, and that issue does not seem to exist any longer. But not sure which fix that fixed that. So I think #12127 can be closed when this one is closed.

PR #12099, opened 2025-10-01
[1856] eliminate prime trains

philcampeau:
## Before clicking "Create"

### Explanation of Change

Long ago when 1856 was coded, rather than programming in a way for events to happen on specific trains other than the first of each rank, it was decided to create extra "prime" ranks for the final train of each type in order to trigger the float percentage change that happens in 1856 when the last train of a rank is purchased. 1836Jr56, which is built on 1856, also used this method.

I had to make a few changes for this to work. 

- eliminated the prime ranks
- added an additional train to each rank that previously had a prime rank
- added events to each rank triggering the change on the final train
- These new events cluttered the game info page, ~so I added a new HIDDEN_EVENTS constant to /game/base.rb and tweaked the game_info.rb code to exclude any events listed in HIDDEN_EVENTS.~
  - This is changed to a check in `View::Game::GameInfo` for a `hidden` key in the event hash definition. If this key is present and the value is truthy then the event will not be shown.
- I took the opportunity to clean up some code in the game_trains and game_phases methods.

~Unfortunately, every attempt I made at writing a migration script resulted in some drift of the loans. I'm not sure why, the scripts I wrote were only changing the train ids, but every time the corporations would end up with a different amount of money at some point which would result in a failure when they couldn't afford a train that they had purchased in the game.~

~I've done test runs through several games and haven't witnessed this in my own plays, so I suspect the issue is more to do with the way loans are handled in the 1856 code itself, which is pretty wonky if I'm being honest.~

~**All that to say, we'll need to pin all existing games of 1856 and 1836Jr56**, and I've had to delete the fixture files. Once some new games are completed on the site, we can use them to create new fixtures, of course.~

**🏴‍☠️ This pull request has been hijacked by @ollybh 🏴‍☠️**

A migration has been added that can migrate all games of 1836Jr56, and all 1856 games except one (that has a train bought from the bank pool whose identifier is changed). The fixture files have been updated to work with these changes.

This PR does not touch any 18KA code. That game is derived from 1856 so will also need updating if it is resurrected – it is still in pre-alpha status and hasn't had any significant development committed in a few years.


michaeljb:
I don't like needing to pin all of the instances of a "production" game, and I don't want to remove all the fixtures to replace them later--in fact in #12083 I'm working on adding spec tests that require fixtures are present for all released games.

I was able to edit `51745.json` by hand to "migrate" it:
* `"train": "2'-0"` -> `"train": "2-5"` and `"variant": "2'"` -> `"variant": "2"`
* `"train": "3'-0"` -> `"train": "3-4"` and so on

So I think a migration script ought to be possible.

michaeljb (changes requested):
I don't like needing to pin all of the instances of a prod game, and I don't want to remove all the fixtures to replace them later--in fact in #12083 I'm working on adding spec tests that require fixtures are present for all released games.

I was able to edit `51745.json` by hand to "migrate" it:
* `"train": "2'-0"` -> `"train": "2-5"` and `"variant": "2'"` -> `"variant": "2"`
* `"train": "3'-0"` -> `"train": "3-4"` and so on

So I think a migration script ought to be possible.

Screenshot 2025-10-02 at 21 36 34
Screenshot 2025-10-02 at 21 36 36

philcampeau:
The issue comes up after the merger, which happens on the 6-train. Looking at all the fixture files, not a single one goes to the 6-train. 

I'll keep tinkering and see what I can find.

philcampeau:
Everything I try, the fix fails like above.

It's a shame, the prime trains was such a bad hack to solve the issue, but whatever is causing the migration script to fail on the prod games is a deeper issue in the code that I don't think I'll be able to fix.

lib/engine/game/base.rb
```diff
@@ -332,6 +332,8 @@ class Base
           ['Companies Close', 'All companies unless otherwise noted are discarded from the game'],
       }.freeze
 
+      HIDDEN_EVENTS = [].freeze
+
```
ollybh: Instead of defining yet another constant, you could just add `'hidden' => true` to the event definitions and test for this hash key.

ollybh (changes requested):
Like @michaeljb said, this needs a migration script to avoid pinning almost every single game. This shouldn't be complicated and I tried putting one together.

The first issue I ran into is that this does not change to having full capitalisation once the first 6-train is available. Capitalisation is determined from the phase status array, and phase 5 has an `incremental` status. `event_no_more_incremental_corps!` doesn't do anything other than printing a log message.

ollybh:
> The first issue I ran into is that this does not change to having full capitalisation once the first 6-train is available. 

And capitalisation is also wrong when the first 5-train is available. It should change from escrow to incremental.

This incorrect capitalisation was changing the amount of money that corporations have, causing the migration failures.

ollybh:
After fixing the capitalisation, I can migrate all the 1856 games – apart from one. I'll dig into that one, tidy up my migration script and then push an update to this PR.

ollybh:
The one 1856 game still failing is game 222637, at [action 711](https://18xx.games/game/222637?action=711). A 4-train is being bought from the pool. Earlier in the game, four 4-trains were discarded, including the 4'-train. The order of trains in the pool is different causing the failure.

This game can be pinned, it doesn't seem worth the effort to fix the migration script for a single completed game.

ollybh:
I've updated this pull request with:

- The migration script.
- Fixes for capitalisation.
- Fixes for 1886Jr.
- Updated fixture files that work with the new code.
- Check for event `hidden` key instead of using a global `PRIVATE_EVENTS` variable.

philcampeau:
Damn dude! This is great. I was away on vacation the last couple of weeks. I'll take a loot at this tomorrow. Thanks!

philcampeau:
Hey @ollybh I just want to say that I really appreciate the work you put into this. I missed some obvious stuff which you caught, and you also fixed some issues I couldn't figure out for the life of me. 

Cheers.

ollybh (approved):
Getting rid of my request for changes flag on this PR.

philcampeau:
@michaeljb unless I'm mistaken, @ollybh fixed the issues with this pull request.

crericha (approved):
@ollybh migration script looks good to me; however, if you have any concerns, maybe import a few games locally and try the script on them before the entire prod db?

michaeljb (approved):

michaeljb:
Running the migration locally worked for 361/363 games. Will pin those two and migrate the rest.

[222637](https://18xx.games/game/222637) failed during the migration:

```
Game 222637 [1856]: failed validation Not a buyable train: 4-0
```

[245040](https://18xx.games/game/245040) appeared to pass the migration but failed during subsequent validation:

```
#<Engine::GameError: Blocking step Buy Trains cannot process action pass at 411>
```

PR #12102, opened 2025-10-03
[1840] add hexes with Stadtbahnen tokens to available_hexes

### LINKED ISSUE #12097: 1840: removing and laying tokens
Hello,

I'm playing a 1840 game (Id: 218809) and it's turn of line 16. The program doesn't allow to remove Stadtbahnen tokens (rules allow to remove as many as you want and connection is not required). The only Stadtbahn token I can remove is which is in D22. Rules say it doesn't matter whether you prefer lay first tracks or tokens. However, if I upgrade D22 then I am unable to lay my own token there or any other legal slot or remove the Stadtbahn token there, considering before upgrading I could remove it.
The oddest about this bug is 1840 worked fine for months and months so I cannot understand what has happen with.
The game is stopped for the moment while we wait a solution or an answer about.

Thank you so much,

Enrique Trigueros.

michaeljb:
Fixes #12097

## Before clicking "Create"


Zwergenpunk (approved):

Zwergenpunk:
Ty for the fix

Issue #12104, opened 2025-10-03
[1862 Solo] Checklist for implementation


perwestling:
Implementation of 1862 Solo has started.

This is an attempt to create a task list of things to do. Add more if needed.

### Stock Round
- [ ] Remove corporations when dropped
- [ ] Improve UX for "stacks" to make it easier to know which corporation is which
TBD

### Operating Round
TBD

### Misc
- [ ] Wiki page
- [ ] Pre alpha test
- [ ] Alpha test
- [ ] Beta test
- [ ] Add unit tests
TBD

waaayback:
Stoked to see that this is in implementation!  Will be happy to playtest when the alpha is ready!

fanman80:
Also Stoked to see that this is in implementation! Will be happy to playtest when the alpha is ready!

PR #12117, opened 2025-10-05
Add functionality for 1862 Solo

perwestling:
Add functionality for 1862 Solo Stock Round. Still not a playable game.

There are no games in the database so it should not affect anything.
It should also not affect 1862 base game - have not made any changes of that code, and use dup during setup to avoiding changing base classes.

#### Change of general functionality
- For IPO rows it is now possible to add general choices to a company in an IPO rows, and not just buy_company. This is done by adding the method general_input_rendering_ipo_row to a game. This method should return a list with [a, b] where a = a string containing the choice name b = a string with description It is up to the step of the game to handle this by adding a process_choose which handles the action which has action.choice with matching choice name.

1. Add Permits which are assigned when corporation floats. Cannot float when no permits remain unassigned.
2. Add actions to IPO rows (using general_input_rendering_ipo_rows) a) Buy, which buys a(n already IPOed) top share in an IPO row. This floats the corporation when 30% sold. b) Move, which moves a top share in an IPO row to another IPO row (which has the same corporation on top). c) Remove, which removes a top share, and close the corporation, removing all shares. d) Par, which pars the corporation, either unchartered or chartered, with suitable functions. e) Deal, which adds 6 to an empty IPO row, and removes a corporation.
3. Add IPO row and permit information to Info page.
4. Sell actions is possible, but need to be revisited. It correctly stops you from selling shares you have bought the same SR.

#### Remaining issues
1. Chartered corporation does not do anythin during ORs
2. Unchartered corporations get to buy 3 tokens, but it should be 2-7
3. Removing a corporation does "bomb" starting hex. If there are things in the starting hex it should be "freezed" instead.
4. Only lowest available par value is shown as an option. Should we show all using similar choices or are there some other way we can do this?
5. Having president's share as reserved 30% might not be good. First attempt did just use 70% but then first share was president's share and it caused issues. This way at least it was only regular shares available.
6. We need to check if we handle unfloated corporations properly:
   - skipped during OR without affecting share price
   - sold shares get half price, as train less. Also need to check they end up in bank pool, and cannot be bought directly by player.
7. Mulligan is not supported. Probably not needed, as you can always create a new solo game.

### Explanation of Change

Attempt to get Stock Round to get full functionality in 1862 Solo. Remaining issues are described above.

### Screenshots

image

### Any Assumptions / Hacks

Have added some TODOs to the code.

### DIFF
--- a/assets/app/view/game/ipo_rows.rb
+++ b/assets/app/view/game/ipo_rows.rb
@@ -40,7 +40,7 @@ module View
         return h(:div) if companies.empty?
 
         divs = [render_title(number)]
-        divs << render_first_ipo(companies) if @show_first
+        divs << render_first_ipo(companies, number) if @show_first
         divs << h(CompaniesTable, game: @game, companies: companies) unless companies.empty?
 
         h('div.player.card', { style: card_style }, divs)
@@ -59,7 +59,7 @@ module View
         h('div.player.title.nowrap', props, ["IPO Row #{number}"])
       end
 
-      def render_first_ipo(ipo_row)
+      def render_first_ipo(ipo_row, number)
         button_props = {
           style: {
             display: 'grid',
@@ -70,6 +70,15 @@ module View
         first_company = ipo_row.shift
         inputs = []
         inputs.concat(render_buy_input(first_company)) if @current_actions.intersect?(%w[buy_company corporate_buy_company])
+
+        # Make it possible for a title to inject more choices than just buy
+        if @step.respond_to?(:general_input_renderings_ipo_row)
+          renderings = @step.general_input_renderings_ipo_row(@current_entity, first_company, number - 1)
+          renderings.each do |(choice, description)|
+            inputs.concat(render_general_input(choice, description))
+          end
+        end
+
         children = []
         children << h(Company, company: first_company, interactive: !inputs.empty?)
         children << h('div.margined_bottom', button_props, inputs) if !inputs.empty? && @selected_company == first_company
@@ -91,6 +100,19 @@ module View
         [h(:button,
            { on: { click: buy } },
            "Buy #{company.name} from IPO Row for #{@game.format_currency(company.value)}")]
+      end
+
+      def render_general_input(choice, description)
+        action = lambda do
+          process_action(Engine::Action::Choose.new(
+            @current_entity,
+            choice: choice,
+          ))
+          store(:selected_company, nil, skip: true)
+        end
+        [h(:button,
+           { on: { click: action } },
+           description)]
       end
     end
   end
--- a/lib/engine/game/g_1862_solo/company.rb
+++ b/lib/engine/game/g_1862_solo/company.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require_relative '../../company'
+
+module Engine
+  module Game
+    module G1862Solo
+      class Company < Engine::Company
+        attr_accessor :ipo_row_index
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_1862_solo/game.rb
+++ b/lib/engine/game/g_1862_solo/game.rb
@@ -27,20 +27,49 @@ module Engine
         end
 
         def setup_preround
-          @original_corporations = @corporations.dup
+          @base_tiles = []
+          @skip_round = {}
+          @lner_triggered = nil
+          @lner = nil
 
-          super
+          # remove reservations (nice to have in starting map)
+          @corporations.each { |corp| remove_reservation(corp) }
 
-          @original_corporations.reject { |c| @corporations.include?(c) }.each do |c|
-            hex = @hexes.find { |h| h.id == c.coordinates } # hex_by_id doesn't work here
-            old_tile = hex.tile
-            tile_string = ''
-            hex.tile = Tile.from_code(old_tile.name, 'brown', tile_string)
+          @double_parliament = false
+
+          # randomize order of corporations, then remove some based on player count
+          @offer_order = @corporations.sort_by { rand }
+          removed = @offer_order.take(4)
+          removed.each do |corp|
+            @offer_order.delete(corp)
+            @corporations.delete(corp)
+
+            remove_corporation_hex(corp)
+
+            @log << "Removing #{corp.name} from game"
           end
+
+          # add markers for remaining companies
+          @corporations.each { |corp| add_marker(corp) }
+
+          @chartered = {}
+
+          # randomize and distribute train permits
+          permit_list = 2.times.flat_map { %i[freight express local] }
+          permit_list.sort_by! { rand }
+          @all_permits = permit_list.dup
+          @permits = Hash.new { |h, k| h[k] = [] }
+          @original_permits = Hash.new { |h, k| h[k] = [] }
+
+          @corporations.each { |c| convert_to_full!(c) }
 
           # Build draw and draft decks for player hand and IPO rows
           @ipo_rows = [[], [], [], [], [], [], [], [], []]
           create_decks(@corporations)
+        end
+
+        def ready_corporations
+          @corporations.reject(:closed?)
         end
 
         def create_decks(corporations)
@@ -48,9 +77,12 @@ module Engine
 
           corporations.each do |corporation|
             corporation.ipo_shares.each do |share|
+              next if share.percent > 10
+
               company = convert_share_to_company(share)
               company.owner = bank
               draw_deck << company
+              @companies << company
             end
           end
 
@@ -61,7 +93,7 @@ module Engine
         # create a placeholder 'company' for shares in IPO
         def convert_share_to_company(share)
           description = "Certificate for #{share.percent}\% of #{share.corporation.full_name}."
-          Company.new(
+          G1862Solo::Company.new(
             sym: share.id,
             name: share.corporation.name,
             value: 0,
@@ -78,6 +110,9 @@ module Engine
         def deal_deck_to_ipo(deck)
           all_rows_indexes.each do |row|
             @ipo_rows[row] = deck.pop(6) # 6 shares per row
+            @ipo_rows[row].each do |company|
+              company.ipo_row_index = row
+            end
           end
         end
 
@@ -138,6 +173,10 @@ module Engine
           true
         end
 
+        def can_par_corporations?
+          @all_permits.any?
+        end
+
         def in_ipo?(company)
           @ipo_rows.flatten.include?(company)
         end
@@ -158,24 +197,46 @@ module Engine
           []
         end
 
+        def companies_to_payout(_ignore)
+          []
+        end
+
         def init_corporations(_stock_market)
           self.class::CORPORATIONS.map do |corp|
             corporation = corp.dup
+            corp.deep_freeze
             corporation[:float_percent] = 30
-            corporation[:shares] = [10, 10, 10, 10, 10, 10, 10]
+            corporation[:shares] = [30, 10, 10, 10, 10, 10, 10, 10]
             corporation[:max_ownership_percent] = 70
             corporation[:min_price] = 1
             initiated_corp = G1862Solo::Corporation.new(
               **corporation.merge(corporation_opts),
             )
-            initiated_corp.forced_share_percent = 10
+            initiated_corp.presidents_share.buyable = false
             initiated_corp
           end
         end
 
+        def stock_prices
+          par_prices
+        end
+
+        def par_prices
+          @par_prices ||= stock_market.market.first.select { |p| p.type == :par }
+        end
+
+        def repar_prices
+          @repar_prices ||= stock_market.market.first.select { |p| p.type == :repar }
+        end
+
         # So that value is not shown on company cards representing shares
-        def company_value(_company)
-          0
+        def company_value(company)
+          corporation = company.treasury.corporation
+          corporation.share_price ? corporation.share_price.price : 0
+        end
+
+        def company_header(_company)
+          '10% SHARE'
         end
 
         def show_value_of_companies?(_owner)
@@ -189,6 +250,12 @@ module Engine
         # Timeline information for INFO tab
         def timeline
           timeline = []
+
+          timeline << if @all_permits.empty?
+                        'Permits left to assign: None'
+                      else
+                        "Permits left to assign: #{@all_permits.map(&:to_s).join(', ')}"
+                      end
 
           all_rows_indexes.each do |i|
             ipo_row_i = ipo_timeline(i)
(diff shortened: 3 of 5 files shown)


lib/engine/game/g_1862_solo/company.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require_relative '../../company'
+
+module Engine
+  module Game
+    module G1862Solo
+      class Company < Engine::Company
+        attr_accessor :ipo_row_index
```
roseundy: Are you subclassing Company here only to add this attribute? If so, it might be better to just add a hash to the 62_solo game class that is indexed by the company.

perwestling:
Have updated https://github.com/tobymao/18xx/issues/12104 a bit.

lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb
```diff
@@ -7,7 +7,7 @@ module Game
     module G1862Solo
       module Step
         class BuySellParShares < Engine::Step::BuySellParShares
```
roseundy: Is there a reason you are not subclassing 1862's BuySellParShares?

perwestling: Good question. Not sure there was a reason, but I can try to do that. A quick browse did not see any reason why it should not work.

BTW, UNCHARTERED_TOKEN_COST exist in both 1862 buy_sell_par_shares and game.

lib/engine/game/g_1862_solo/company.rb
perwestling: I will change that.

perwestling: Have now done that in a new commit

lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb
perwestling: Have now done that in a new commit

roseundy (approved):
Look good

lib/engine/game/g_18_royal_gorge/step/dividend.rb
perwestling, line 1: Maybe this can be removed from the commit, as it can be handled by a 18RoyalGeorge commit instead?

perwestling:
@roseundy I did some more work after your review. I think this is good enough to go into main branch so that those that want to work on it can have access to it.  And to avoid gargantuan commits.

@michaeljb Does the IPO stuff looks OK? That is the only core part.

assets/app/view/game/ipo_rows.rb
```diff
@@ -59,7 +59,7 @@ def render_title(number)
         h('div.player.title.nowrap', props, ["IPO Row #{number}"])
       end
 
-      def render_first_ipo(ipo_row)
+      def render_first_ipo(ipo_row, number)
```
ollybh: The parameter names are not at all descriptive. I would have guessed that `ipo_row` was a row number and `number` a position within the rown.

`render_first_ipo(companies, row_number)` might be better.

lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb
```diff
@@ -17,18 +17,212 @@ def actions(entity)
...
+              action_move(choice)
+            elsif choice.start_with?('par_chartered')
+              action_par_chartered(choice)
+            elsif choice.start_with?('par_unchartered')
+              action_par_unchartered(choice)
+            elsif choice.start_with?('remove')
+              action_remove(choice)
+            else
+              raise GameError, "Unknown choice #{choice}"
+            end
```
ollybh: This would be clearer with a `case` statement. You can use a regexp to match the start of a string.
```suggestion
            case choice
            when /^buy/ then action_buy(choice)
            when /^deal/ then action_deal(choice)
            when /^move/ then action_move(choice)
            when /^par_chartered/ then action_par_chartered(choice)
            when /^par_unchartered/ then action_par_unchartered(choice)
            when /^remove/ then action_remove(choice)
            else raise GameError, "Unknown choice #{choice}"
            end
```

ollybh: Or you could remove all the calls to `String.split` in the `action_*` methods by doing it here instead:
```
action, company, row = action.choice.split('#')
case choice
when 'buy' then action_buy(company)
…
when 'move' then action_move(company, row)
…
end

Comment (ollybh) [reply lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb]: Or, instead of constructing and splitting strings, you could use a hash as the key to the choices array. 1858 does this: https://github.com/tobymao/18xx/blob/f66389367d0e05aeb45b45c88c4a0e985bf69dda/lib/engine/game/g_1858/step/private_closure.rb#L71-L103

This gives an action that looks like this:
```
{
  "choice": {
    "corporation": "W",
    "from": "treasury"
  },
  "created_at": 1738190630,
  "entity": "V&J",
  "entity_type": "minor",
  "id": 573,
  "type": "choose"
}
```

The values are then extracted using `action.choice['corporation']` and `action.choice['from']`.

File: assets/app/view/game/ipo_rows.rb (hunk 3 above)
Comment (perwestling) [reply assets/app/view/game/ipo_rows.rb]: Have now refactored to use better names. ipo_row_number is commonly used.

File: lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb (hunk 4 above)
Comment (perwestling) [reply lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb]: Have refactored code to use case in one commit, and then to use hash in another commit, so code should be fine now.

**Comment by perwestling** [comment]
@ollybh Have now changed the code after your review comments, so please check that it looks OK.

Have set this as mergable again.

File: lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb [hunk 5]
```diff
@@ -17,18 +17,235 @@ def actions(entity)
...
           def get_par_prices(entity, _corp)
             @game.repar_prices.select { |p| p.price * 3 <= entity.cash }
           end
+
+          def general_input_renderings_ipo_row(_entity, company, ipo_row_number)
+            renderings = []
+
+            if @game.ipo_rows[ipo_row_number - 1].empty?
+              # No company in this IPO row so 2nd part of action is nil
+              renderings << ["deal###{ipo_row_number}", "Deal to #{ipo_row_title(ipo_row_number)}"]
```
ollybh: This one hasn't been updated to use the `create_choice` method.

lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb
```diff
@@ -17,18 +17,235 @@ def actions(entity)
...
+          end
+
+          def get_company(id)
+            company = @game.company_by_id(id)
+            raise GameError, "Company with ID #{id} not found in IPO rows" unless company
+
+            company
+          end
+
+          def get_ip_row_index(ipo_row_number)
```
ollybh, line 204: Minor: you use `ipo_row` elsewhere but this is `ip_row`. Is there a reason for the change?

assets/app/view/game/ipo_rows.rb
```diff
@@ -39,14 +39,14 @@ def render_ipo_row(ipo_row, number)
         companies = ipo_row.dup
         return h(:div) if companies.empty?
 
-        divs = [render_title(number)]
-        divs << render_first_ipo(companies) if @show_first
-        divs << h(CompaniesTable, game: @game, companies: companies) unless companies.empty?
+        divs = [render_title(ipo_row_number)]
+        divs << render_first_ipo(companies, ipo_row_number) if @show_first
+        divs << h(CompaniesTable, game: @game, companies: companies)
```
ollybh, line 44: You've removed the `companies.empty?` test here. Could this affect 18India?

assets/app/view/game/ipo_rows.rb
```diff
@@ -56,20 +56,30 @@ def render_title(number)
...
+      def render_first_ipo(ipo_row, ipo_row_number)
         button_props = {
           style: {
             display: 'grid',
             gridColumn: '1/4',
             width: 'max-content',
           },
         }
-        first_company = ipo_row.shift
+        first_company = ipo_row.first
```
ollybh: I don't think this change is right. This means that the first card is shown twice in the row, once in the detailed view and then at the head of the queue of cards in the row.

assets/app/view/game/ipo_rows.rb
```diff
@@ -56,20 +56,30 @@ def render_title(number)
...
         inputs = []
-        inputs.concat(render_buy_input(first_company)) if @current_actions.intersect?(%w[buy_company corporate_buy_company])
+
+        # Make it possible for a title to inject more choices than just buy
+        if @step.respond_to?(:general_input_renderings_ipo_row)
+          renderings = @step.general_input_renderings_ipo_row(@current_entity, first_company, ipo_row_number)
+          renderings.each do |(choice, description)|
```
ollybh: `general_input_renderings_ipo_row` is very long. And 'renderings' isn't an English word that is used much. We have choice and choices elsewhere in this code, would `ipo_row_choices` be a shorter and simpler name?

ollybh (changes requested):
This looks better with the removal of the string concatenation and splitting. But it is changing the IPO display for 18India. Can that part of the PR be reverted?

Before this change:
Image

After this change – note the repeated NWR and SPD shares at the top of the queues:
Image

I've got a few more suggestions, all minor.

assets/app/view/game/ipo_rows.rb
perwestling: It is probably this line that causes the issue with 18 India. I will check it.

assets/app/view/game/ipo_rows.rb
perwestling: No, on line 40 there is a return if companies.empty? so the unless is not needed in line 44.

lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb
perwestling: IP row number is the name of the IP row.
IP row number is index of IP row plus one.

assets/app/view/game/ipo_rows.rb
perwestling: Renderings might have a slightly different meanins than the intention is here.
But it is not really renderings, so yes I should rename this, and probably use something with choice in.

lib/engine/game/g_1862_solo/step/buy_sell_par_shares.rb
perwestling: Corrected in next commit.

assets/app/view/game/ipo_rows.rb
perwestling: It seems this was the issue, so after changing that back it now looks correct.

assets/app/view/game/ipo_rows.rb
perwestling: Renamed things a bit in the following commit.

lib/engine/game/g_18_royal_gorge/step/dividend.rb
perwestling: Removed this change in the following commit.

perwestling:
@ollybh Ready for review again.

assets/app/view/game/ipo_rows.rb
```diff
@@ -69,7 +69,17 @@ def render_first_ipo(ipo_row)
         }
         first_company = ipo_row.shift
         inputs = []
-        inputs.concat(render_buy_input(first_company)) if @current_actions.intersect?(%w[buy_company corporate_buy_company])
+
+        # Make it possible for a title to inject any choices
+        if @step.respond_to?(:choices_for_ipo_row)
+          choices_with_description = @step.choices_for_ipo_row(@current_entity, first_company, ipo_row_number)
+          choices_with_description.each do |(choice, description)|
```
ollybh:
```suggestion
          choices_with_description.each do |choice, description|
```

lib/engine/game/g_1862_solo/game.rb
```diff
@@ -27,41 +27,81 @@ def show_game_cert_limit?
...
 
+          # add markers for remaining companies
+          @corporations.each { |corp| add_marker(corp) }
+
+          @chartered = {}
+
+          # randomize and distribute train permits
+          permit_list = 2.times.flat_map { %i[freight express local] }
+          permit_list.sort_by! { rand }
+          @all_permits = permit_list.dup
```
ollybh:
```suggestion
          @all_permits = (%i[freight express local] * 2).sort_by { rand }
```

lib/engine/game/g_1862_solo/game.rb
```diff
@@ -128,14 +183,19 @@ def init_round
...
           ])
         end
 
         def show_ipo_rows?
-          true
+          @round.buy_tokens.nil?
+        end
+
+        def can_par_corporations?
+          @all_permits.any?
```
ollybh:
```suggestion
          !@all_permits.empty?
```

lib/engine/game/g_1862_solo/game.rb
```diff
@@ -158,24 +218,42 @@ def buyable_bank_owned_companies
...
         end
 
+        def companies_to_payout(_ignore)
+          []
+        end
+
         def init_corporations(_stock_market)
           self.class::CORPORATIONS.map do |corp|
             corporation = corp.dup
+            corp.deep_freeze
```
ollybh: The `deep_freeze` would be better applied to the definition of the `CORPORATIONS` array.

lib/engine/game/g_1862_solo/game.rb
```diff
@@ -205,11 +289,86 @@ def ipo_timeline(index)
...
+          ipoed.sort + others
+        end
+
+        def random_corporation
+          # Get a random corporation, which player does not own any shares of. TODO: Is this correct?
+          @corporations.reject { |c| c.closed? || c.ipoed || player_owns_any_shares_of?(c) }.min_by { rand }
+        end
+
+        def player_owns_any_shares_of?(corp)
+          @players.first.shares_by_corporation[corp].any?
```
ollybh:
```suggestion
          !@players.first.shares_by_corporation[corp].empty?
```

lib/engine/game/g_1862_solo/game.rb
```diff
@@ -205,11 +289,86 @@ def ipo_timeline(index)
...
 
         def all_rows_indexes
           (0..8)
         end
+
+        def remove_corporation_hex(corp)
+          hex = @hexes.find { |h| h.id == corp.coordinates } # hex_by_id doesn't work here
+          old_tile = hex.tile
+          tile_string = ''
+          hex.tile = Tile.from_code(old_tile.name, 'brown', tile_string)
```
ollybh: Would grey be better than brown? It's a standard colour for non-upgradeable tiles. Or would it cause problems for laying track in adjacent hexes?

perwestling: I tried black but that was too intrusive.

So will go with gray, and then we will handle it in some way if it does cause hex upgrade issues.

lib/engine/game/g_1862_solo/game.rb
perwestling: Definition is in g_1826/entities.rb. I have now added it there. Is that OK?

ollybh (approved):
Thanks for making all these changes.

PR #12129, opened 2025-10-12
[dev] Update DEVELOPMENT.md to clarify port 9292 needed.

deeseja:

## Implementation Notes
Simple text update in DEVELOPMENT file.

### Explanation of Change
From Safari after doing a build, the default url opened in the browser is not using port 9292 which is needed.  Adding some clarification.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -17,7 +17,7 @@ This will create a VM hosted by Github just for you with the repo already cloned
 
 You are able to run `make` from the Terminal tab to build and start the 18xx server (based on the current branch checked out)
 
-After the server is running, a pop-up in the bottom right should appear informing you that the server can be opened in a browser.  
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -17,9 +17,9 @@ This will create a VM hosted by Github just for you with the repo already cloned
 
 You are able to run `make` from the Terminal tab to build and start the 18xx server (based on the current branch checked out)
 
-After the server is running, a pop-up in the bottom right should appear informing you that the server can be opened in a browser.
+After the server is running, a pop-up in the bottom right should appear informing you that the server can be opened in a browser.
```
perwestling: I do not see any change on this line? Is it added whitespace?

deeseja: removed whitespace

PR #12143, opened 2025-10-16
[18SJ] Correct private text, fixes #12136

### LINKED ISSUE #12136: 18SJ - Text description for Motala Verkstad private is inconsistent with the rules.
The text description for the Motala Verkstad private company say that "at any time" the owning company may buy a train using this private.  The rules, however, state that this ability can only be used immediately before trains are run.  The private text needs to be changed to conform to the rules as written.

perwestling:
Fixes #12136

## Before clicking "Create"

Only textual change of private text, no other effect.

## Implementation Notes

Description changed for Motala Verkstad.

### Explanation of Change

Functionality is correct. Have more or less copied the text from AAG rulebook.

### Screenshots

image

### Any Assumptions / Hacks

N/A

### DIFF
--- a/lib/engine/game/g_18_sj/entities.rb
+++ b/lib/engine/game/g_18_sj/entities.rb
@@ -99,7 +99,7 @@ module Engine
             name: 'Motala Verkstad',
             value: 90,
             revenue: 20,
-            desc: 'Once during the game, immediately before the run trains action, owning corporation may buy one or '\
+            desc: 'Once during the game, immediately before the run trains action, the owning corporation may buy one or '\
                   'more new trains from the bank. The regular buy train action must be skipped by this corporation '\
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_sj/entities.rb
```diff
@@ -99,9 +99,9 @@ module Entities
             name: 'Motala Verkstad',
             value: 90,
             revenue: 20,
-            desc: "During the owning corporation's operating turn, that corporation may at any time buy one or more "\
-                  'train/s from the bank. This power may only be used one time during the game and must be used '\
-                  'before this private company closes.',
+            desc: 'Once during the game, immediately before the run trains action, owning corporation may buy one or '\
```
ollybh:
```suggestion
            desc: 'Once during the game, immediately before the run trains action, the owning corporation may buy one or '\
```

Issue #12095, opened 2025-09-28
scripts/validate.rb should provide a multithreaded function


michaeljb:
Validating many games can go a lot faster with some process forking

PR #12086, opened 2025-09-27
[core] Fix dropped cash - full PR

michaeljb:
**tl;dr**
* make `Spender#cash=` protected to prevent changing cash totals outside of `Spender#spend()`
* increase fixture coverage for better confidence in that change and related refactorings

----

Fixes #12000
Fixes #12060
Progress on #11973

This is the full PR to fix the dropped cash issue. I've created smaller PRs to make reviewing this more manageable:

- `<link>`
- `<link>`
- `<link>`
- `<link>`
- `<link>`

Once the smaller PRs are approved, my plan is to merge this one, closing all of them at once and creating a single merge commit.

----

In #11975 I noticed that some fixtures had floating point values in their `result` key. I started adding tests to ensure that results were always integers, and then thought to add tests checking that the total cash in the game matched the bank's total cash. This led to the discovery that not all cash transactions were going through the `Spender#spend()` method, and that player loans were handling cash in incorrect and unexpected ways.

In order to feel confident in my fixes for those issues, I wanted test fixtures for more games, and so I started  adding tests that would ensure titles have fixtures that reach real endgame conclusions (instead of just being `manually_ended`), but that required exposing `game_end_reason`, and some issues with my first pass at that led to refactoring the base `game_end_check`.

----

## Before clicking "Create"

- [x] Branch is derived from the latest `master`
- [x] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

## TODO:

- [ ] get approval on the broken down PRs
- [ ] validate full local DB to identify affected games for pins/archives so that a full validate does not need to be run on prod

PR #12156, opened 2025-10-22
[core] replace `undefined method _ for nil` with new `ActionError`

michaeljb:
note to reviewers: look at the first commit to more easily find the files with the key changes

----

raise `ActionError` when `REQUIRED_ARGS` are missing for action creation

While validating locally for https://github.com/tobymao/18xx/pull/12086 and https://github.com/tobymao/18xx/issues/12095, I happened to have local
games broken by https://github.com/tobymao/18xx/pull/12073 (they are pinned in prod), and the error was very
unhelpful:

`#<NoMethodError: undefined method 'id' for nil:NilClass>"`

Additionally, `broken_action.to_h` caused its own error in the validate
script, as the error was coming from the Action -> hash conversion.

With these changes, the error is much more helpful:

`#<Engine::ActionError: Cannot create Engine::Action::LayTile, h_to_args() returned nil :tile from action 745>`

----

In Ruby, [`Method#parameters`][0] can be used to inspect the arguments of a
and identify the required kwargs, so that could be used instead of defining
`REQUIRED_ARGS` for every action.

However, for this to work in Opal, the `Opal::Compiler` must have `arity_check:
true`.

This setting enforces that methods/lambdas/procs/functions are called with the
correct number of arguments in the JavaScript; after a few minutes of testing,
this appears to break most of our views, and fixing an error quickly leads to
the discovery of the next one.

JS event handlers, for example, are called with more args by the browser than we
care about or define. To make this work with Opal's arity check, dummy args need
to be added to every such function. It doesn't seem feasible to fix everything
in this way--which functions are affected and how many arguments are expected is
not clear until Opal throws the error, so fully implementing this check at this
point would be quite painful, and essentially impossible to roll out with
confidence that users wouldn't encounter the error.

[0]: https://docs.ruby-lang.org/en/master/Method.html#method-i-parameters

## Before clicking "Create"

### DIFF
--- a/spec/lib/engine/action_spec.rb
+++ b/spec/lib/engine/action_spec.rb
@@ -17,7 +17,7 @@ describe Engine::Action do
         args = action_class.instance_method(:initialize).parameters.filter_map do |kind, name|
           (kind == :keyreq) && name
         end
-        expect(action_class::REQUIRED_ARGS).to eq(args)
+        expect(action_class::REQUIRED_ARGS).to match_array(args)
       end
     end
   end
--- a/lib/engine/action/lay_tile.rb
+++ b/lib/engine/action/lay_tile.rb
@@ -7,7 +7,7 @@ module Engine
     class LayTile < Base
       attr_reader :hex, :tile, :rotation, :combo_entities
 
-      REQUIRED_ARGS = %i[tile hex rotation].freeze
+      REQUIRED_ARGS = %i[rotation tile hex].freeze
 
       def initialize(entity, tile:, hex:, rotation:, combo_entities: [])
         super(entity)


spec/lib/engine/action_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      it "REQUIRED_ARGS match #initialize()'s default-less keyword arguments" do
+        # instance_method(:initialize).parameters doesn't work in the compiled
+        # JS, so instead of simply using this code in
+        # Engine::Action::Base.from_h to check the appropriate args, this test
+        # is used to ensure the action classes have those args listed in their
+        # REQUIRED_ARGS
+        args = action_class.instance_method(:initialize).parameters.filter_map do |kind, name|
+          (kind == :keyreq) && name
+        end
+        expect(action_class::REQUIRED_ARGS).to eq(args)
```
ollybh: Very minor quibble, would it be more appropriate to use `match_array(args)` instead of `eq(args)` here? We don't really care about the order.

michaeljb: SG

PR #12157, opened 2025-10-22
[1822PNW] don't add Major reservation when associated minor closes

### LINKED ISSUE #11346: [1822PNW] Incorrect placement of home token resulting in remove of another player's home token
[Before the start of the SPS company](https://18xx.games/game/179700?action=1301)

- - During Operating Round 7.1, when SPS Company starts, it automatically places its home token in Portland, which resulted in the removal of ORNC сompany's home token at turn [1302](https://18xx.games/game/179700?action=1302)

- - This is contrary to paragraph 5.10.3 of the rules: "Destination station tokens and exchange station tokens do not count against this limit."

Before SPS starts:
image
After SPS starts:
image

**Any Browser / Platform.**

michaeljb:
* add error message when a reservation and token clash -- would have been thrown instead when this bug was encountered
* although the the reservation disappearing is a prerequisite for an extra slot to used for a home token, 1822PNW is already configured to allow the home token to go in an extra slot, so no other functional changes were needed; a test has been aded spelling out the extra slot use

Fixes #11346

## Before clicking "Create"


michaeljb:
I guess the core change wasn’t as innocuous as I thought, will reopen once I can fix the tests

PR #12168, opened 2025-10-23
[dev] add helper method for more convenient debugging

michaeljb:
## Before clicking "Create"

### Explanation of Change

* adds globally available function, `debug!` for more convenient setting of breakpoints
* `debug!` works for local testing in Ruby as well as in the browser
    * if `binding.pry` is hit when loading a game in the browser, it throws an error since `pry-byebug` cannot be imported and `binding.pry` is undefined; with `debug!` the standard JS `debugger` call is hit instead
* more convenient breakpoints in unit tests, no `require` statement needed
* the added rubocop config means any usages of `debug!` will cause a violation, which should be enough to keep any `debug!` calls out of prod

### Any Assumptions / Hacks

* the `debugger` JS statement needs something after it in the function to work, which is why `debug!` explicitly returns `nil`
* `engine/debug` is imported in `engine.rb`, but since `pry-byebug` is only imported inside the function call, it won't cause an error in prod (if it is somehow added despite RuboCop)
* if `debug!` is somehow called in prod on the server, a warning will be logged but the breakpoint will not be reached, and no attempt to import `pry-byebug` will be made
    * tested: locally modified `Gemfile` and `Gemfile.lock` so that `pry-byebug` was not installed, and confirmed that there is no problem with importing `lib/engine/debug.rb`; an error is only thrown when `debug!` is actually called

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -34,6 +34,10 @@ Layout/MultilineOperationIndentation:
 Layout/SpaceBeforeBrackets:
   Enabled: false
 
+Lint/Debugger:
+  DebuggerMethods:
+    Engine: debug!
+
 Lint/MissingSuper: # TODO enable & check offenses
   Enabled: false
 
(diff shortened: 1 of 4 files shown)


ollybh (approved):
I've finally got round to testing this, albeit just the Javascript interface. This looks like a useful addition.

PR #12171, opened 2025-10-24
[1840] add player name to logs in `Step::AcquisitionAuction`

### LINKED ISSUE #11557: 1840, add (PlayerName) to charter name in the gamelog?
Hello. Hopefully this is non-controversial, and super easy to implement.

By rule in 1840 the player charters are hopelessly tied to the players, a charter cannot be transferred. Yet, in the game log---and particularly in bidding---only the boilerplate charter name is shown, which forces the other players (some without good memory skills) to cross-reference. For example

[18:17] BBG bids 45 for 4

It might be wholly beneficial to add the player name? Like

[18:17] BBG (TexasTex) bids 45 for 4

It would make gameplay simpler!

To whoever may be reading this or acting on this, thank you!

"TexasTex" a/k/a Kevin

michaeljb:
Fixes #11557

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1840/step/acquisition_auction.rb
+++ b/lib/engine/game/g_1840/step/acquisition_auction.rb
@@ -22,6 +22,10 @@ module Engine
             actions
           end
 
+          def name_with_player(entity)
+            "#{entity.name} (#{entity.player.name})"
+          end
+
           def description
             'Acquire Corporations'
           end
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1840/step/acquisition_auction.rb
```diff
@@ -22,6 +22,10 @@ def actions(_entity)
             actions
           end
 
+          def l(entity)
```
Zwergenpunk: Why is the function called l? 😅

michaeljb: l for log, just wanted something short so it didn’t get in the way of reading the strings elsewhere in the file  🙃

I could rename it `log_with_player`

michaeljb: Similarly while working on the loan/bank bugs, I imagined renaming `format_currency` to `fc` everywhere 😅

Zwergenpunk: I'm fine with both, I just personally don't like suuuper short function names

lib/engine/game/g_1840/step/acquisition_auction.rb
```diff
@@ -22,6 +22,10 @@ def actions(_entity)
             actions
           end
 
+          def log_with_player(entity)
```
ollybh: This method doesn't log anything, so its name is misleading. How about `name_with_player`?

PR #12172, opened 2025-10-24
[core/1822PNW] stop Rockport coal acting as token for E-train revenue or graph connectivity

### LINKED ISSUE #11139: [1822PNW] The Rockport Coal Mine is included as an E train station.
The Rockport Coal Mine adds $60 revenue to an E train. I'm assuming it shouldn't be counted as a station for E train purposes?

https://18xx.games/game/166751?action=1325

michaeljb:
* core change: add parameter to `City#tokened_by?` to specify acceptable types of tokens; by default, all are valid
* add `type: coal` to the Rockport coal token
* in base 1822 code, pass `[:normal, :destination]` as acceptable tokens so that the coal token is excluded

Fixes #11139

Pins won't be needed since routes aren't strictly validated.

## Before clicking "Create"

### DIFF
--- a/lib/engine/part/city.rb
+++ b/lib/engine/part/city.rb
@@ -48,11 +48,11 @@ module Engine
       end
 
       def tokened_by?(corporation, types: [])
-        token = (@tokens + @extra_tokens).find { |t| t&.corporation == corporation }
-        return false unless token
+        corp_tokens = (@tokens + @extra_tokens).select { |t| t&.corporation == corporation }
+        return false if corp_tokens.empty?
         return true if types == []
 
-        types.include?(token.type)
+        corp_tokens.map(&:type).intersect?(types)
       end
 
       def find_reservation(corporation)


lib/engine/part/city.rb
```diff
@@ -47,8 +47,12 @@ def tokened?
...
       end
 
-      def tokened_by?(corporation)
-        @tokens.any? { |t| t&.corporation == corporation } || @extra_tokens.any? { |t| t&.corporation == corporation }
+      def tokened_by?(corporation, types: [])
+        token = (@tokens + @extra_tokens).find { |t| t&.corporation == corporation }
+        return false unless token
+        return true if types == []
+
+        types.include?(token.type)
```
ollybh: Do we need to consider a game where a corporation could have both a normal station token and another type of token in the same city?

lib/engine/part/city.rb
```diff
@@ -47,8 +47,12 @@ def tokened?
         !@tokens.compact.empty?
       end
 
-      def tokened_by?(corporation)
-        @tokens.any? { |t| t&.corporation == corporation } || @extra_tokens.any? { |t| t&.corporation == corporation }
+      def tokened_by?(corporation, types: [])
+        token = (@tokens + @extra_tokens).find { |t| t&.corporation == corporation }
+        return false unless token
+        return true if types == []
```
ollybh, line 53: Style
```suggestion
        return true if types.empty?
```

lib/engine/part/city.rb
michaeljb: Good call. 46e7f5702 should account for that possibility

PR #12178, opened 2025-10-26
[core] transform auto_actions attached to messages in `#to_h` and when scrubbing

### LINKED ISSUE #12131: auto_actions should never attach to messages
https://18xx.games/game/222776?action=361 appears broken, I presumed from #12130 but I found a different problem.

I was having trouble debugging this and discovered that `run_game()` in `scripts/validate.rb` gave a different result on my machine vs in prod, with the same git commit checked out.

Eventually I found the source of the difference, a `pass` action attached as an `auto_action` to a message:

```
{"type"=>"message",
 "entity"=>605,
 "entity_type"=>"player",
 "id"=>356,
 "created_at"=>1760368079,
 "auto_actions"=>
  [{"type"=>"pass", "entity"=>3739, "entity_type"=>"player", "created_at"=>1757206476}],
 "message"=>"[redacted]"}
```

The changes in `models/game.rb` in #8171 keep messages from being returned when I, as a non-player, look at 222776, so the `pass` is not included, and the game appears broken in my browser; but running validation on the prod server where the message is included leads to the game appearing unbroken.

----

Additionally maybe migrations should preserve `created_at` when possible? Or update the `auto_actions` to match? Since the migration updated this action's `created_at` but not the auto_action, it looks like the chat is from `2025-10-13 15:07:59.769154 UTC` while the auto pass following it is from `2025-09-07 00:54:36 +0000`

michaeljb:
* model change: when removing messages, if a message had auto actions attached, then the first auto action instead takes the place of the message

* scrub change: if a chat message has auto actions attached, replace the contents of the message with `"[redacted]"` so that the auto actions appear in the scrubbed DB as they do in the prod DB, but the user's chat content is still kept out of the scrubbed DB

Closes #12131

In current prod, accessing https://18xx.games/game/222776 appears broken if not logged in as one of the players in that game. The API response keeps messages hidden from you, but somehow a `pass` auto action got attached to a chat message; with that `pass` action missing, the game cannot be processed successfully by your browser.

This change doesn't prevent auto actions from attaching to chats again, but does prevent games with such occurrences from appearing broken, and fixes consistency when the game is opened in the scrubbed DB.

## Before clicking "Create"

### DIFF
--- a/scripts/db_scrub.rb
+++ b/scripts/db_scrub.rb
@@ -38,6 +38,6 @@ end
 def chat_actions(auto_actions:)
   Action.where(**{
                  Sequel.pg_jsonb_op(:action).get_text('type') => 'message',
-                 Sequel.pg_jsonb_op(:action).key?('auto_actions') => auto_actions,
+                 Sequel.pg_jsonb_op(:action).has_key?('auto_actions') => auto_actions, # rubocop:disable Style/PreferredHashMethods
                })
 end


models/game.rb
```diff
@@ -137,10 +137,25 @@ def archive!
   end
 
   def to_h(include_actions: false, logged_in_user_id: nil)
-    actions_h = include_actions ? actions.map(&:to_h) : []
-    if !players.find { |p| p.id == logged_in_user_id } && user_id != logged_in_user_id
-      actions_h.reject! { |a| a['type'] == 'message' }
-    end
+    remove_messages = !players.find { |p| p.id == logged_in_user_id } && user_id != logged_in_user_id
```
ollybh:
```suggestion
    remove_messages = players.none? { |p| p.id == logged_in_user_id } && user_id != logged_in_user_id
```

PR #12186, opened 2025-10-27
[1871 TOP] fixes corp display order in SRs

### LINKED ISSUE #11103: [1871] Companies are not organized according to turn order during SR
**What's your Game ID?**

For most games (at least that I have played), the companies are listed in turn order during the SR, which is really helpful when deciding what stock top buy. For _The Old Prince 1871_ games, the order in which they are listed seems arbitrary, and does not (necessarily) follow turn order.

E.g. in https://18xx.games/game/171967?action=530 they are listed in this order (turn order between brackets): S(1), Gt(3), So(2), A(5), C(4).

Not sure if this is considered a bug or a feature request, but it would be nice if they are listed in the correct order, as is true for other games.

philcampeau:
Fixes #11103

## Before clicking "Create"

### Explanation of Change

I changed the sorting method for visible_corporations. 

Before, the way it was sorted, it showed them in this order:
- Mainline
- Shortline
- ipoed PEIR corps in Tranche 1
- ipoed PEIR corps in Tranche 2
- ipoed PEIR corps in Tranche 3
- non-ipoed PEIR corps
- IPOed non-PEIR corps

Which is not at all how users of the site expect them to be sorted. 

Now, it shows the corporations in this order:

- IPOed corps in descending operating order
- non-IPOed PEIR corps if there are tranche spaces available
- PEIR

In the end, the Mainline and Shortline are labelled. They don't need to also be the first two to appear in the list by default. 

### Screenshots

Before: 
image

After change: 
image

### DIFF
--- a/lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
+++ b/lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
@@ -124,16 +124,19 @@ module Engine
           end
 
           def visible_corporations
-            corps = Array(@game.corporations).compact.reject(&:closed?)
+            corps = @game.corporations
 
-            # Split IPOed vs not
-            ipoed, _others = corps.partition(&:ipoed)
+            # First, display ipoed corporations in descending operating order
+            ipoed = corps.select(&:ipoed)
 
-            # Grab PEIR corps that aren’t IPOed
-            peir_not_ipoed = Array(corps[0..6]).compact.reject(&:ipoed)
+            # Next, select non-ipoed PEIR corporations
+            peir_not_ipoed = corps[0..6].reject(&:ipoed)
 
+            # Only display non-ipoed PEIR corps if an open tranch space is available
             ordered = ipoed.sort
             ordered.concat(peir_not_ipoed) if @game.tranch_available?
+
+            # Finally, display the PEIR itself
             ordered << @game.peir
 
             ordered


lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
```diff
@@ -124,18 +124,19 @@ def can_gain?(entity, bundle, exchange: false)
           end
 
           def visible_corporations
-            # Always show ML and SL
-            @game.corporations[0..1].reject(&:closed?) +
+            corps = Array(@game.corporations).compact.reject(&:closed?)
```
ollybh: Why have you got `Array(…).compact`? Isn't `@game.corporations` always an array? And can it ever hold `nil` values?

lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
```diff
@@ -124,18 +124,19 @@ def can_gain?(entity, bundle, exchange: false)
...
 
-              # Show ipoed
-              @game.corporations[2..6].reject(&:closed?).select(&:ipoed) +
-              (@game.tranch_available? ? @game.corporations[2..6].reject(&:closed?).reject(&:ipoed) : []) +
+            # Split IPOed vs not
+            ipoed, _others = corps.partition(&:ipoed)
```
ollybh: You're not doing anything with the non-ipoed corporations, so `partition` is overkill.
```suggestion
            ipoed = corps.select(&:ipoed)
```

lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
```diff
@@ -124,18 +124,19 @@ def can_gain?(entity, bundle, exchange: false)
...
 
-              # Only show Branches when ipoed
-              @game.corporations[8..13].select(&:ipoed).reject(&:closed?) +
+            # Grab PEIR corps that aren’t IPOed
+            peir_not_ipoed = Array(corps[0..6]).compact.reject(&:ipoed)
```
ollybh: Again, why `Array(…).compact`?

ollybh (changes requested):

lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
philcampeau: fixed

lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
philcampeau: fixed

lib/engine/game/g_1871/step/buy_sell_par_split_shares.rb
philcampeau: fixed

philcampeau:
Solid points Olly. I've been doing online programming courses (finally), and I picked up some overly prudent habits. 

I trimmed it down and added some clarifying comments.

ollybh (commented):
This looks much better.

One question – the code being replaced is making sure that closed corporations are excluded from the returned array. The new code isn't doing this, should it? You could just change the last line of the method to `ordered.reject(&:closed)`.

philcampeau:
> One question – the code being replaced is making sure that closed corporations are excluded from the returned array. The new code isn't doing this, should it? You could just change the last line of the method to `ordered.reject(&:closed)`.

Corporations don't close in TOP1871, there's no reason for it. 

I think the original code was using this to hide corporations that weren't available to start (or it was just overly cautious), but my code already does this with the line `ordered.concat(peir_not_ipoed) if @game.tranch_available?`. This means it won't show un-parred corps unless there is a tranche open to start a new corp.

PR #12196, opened 2025-10-29
[Core] disable Restart Turn button during auctions

### LINKED ISSUE #6605: [18MEX] Restart Last Turn during auction restarts the whole waterfall

philcampeau:
Fixes #6605

## Before clicking "Create"

### Explanation of Change

Hitting the Restart Turn button during auctions can cause problems, notably in waterfall auctions where it can rewind the entire waterfall.

I added a simple exception to disable the Restart Turn button if @game.round.is_a?(Engine::Round::Auction)

I did consider the fact that this could potentially affect games that use Auction rounds in unexpected ways, but in the end, disabling the Restart Turn button doesn't impact the gameplay in any way, at worst a player might need to click undo 2-3 times.

### DIFF
--- a/assets/app/view/game/restart_turn_button.rb
+++ b/assets/app/view/game/restart_turn_button.rb
@@ -25,7 +25,7 @@ module View
       end
 
       def button_disabled?
-        !@game.undo_possible || !@game.turn_start_action_id
+        !@game.undo_possible || !@game.turn_start_action_id || @game.round.is_a?(Engine::Round::Auction)
       end
     end
   end


michaeljb (approved):
LGTM

The Restart Turn button could probably be made to work correctly with auctions with the right `next_turn!` call somewhere, though in a lot of games you only have one action during your turn in an auction round anyway.

PR #12197, opened 2025-10-29
[front page] Reset filters button clears selection

### LINKED ISSUE #7998: "Reset filters" doesn't reset input field
On the main page, in the Filters pulldown, select 1830. Only 1830 games are displayed (correctly)

Click Reset filters. All games are displayed again (correctly). But the pulldown still shows 1830 (incorrectly).

philcampeau:
Fixes #7998

## Before clicking "Create"

### Explanation of Change

Currently, if you filter for a game on the homepage of the site, and then click the `Reset filters` button, it clears the filter, but  leaves the previously selected filter game in the drop-down menu. 

This corrects that so that clicking `Reset filters` also clears the drop-down selection and resets it to (All titles)

### DIFF
--- a/assets/app/view/game_row_filters.rb
+++ b/assets/app/view/game_row_filters.rb
@@ -11,10 +11,12 @@ module View
       url_search_params = Lib::Params::URLSearchParams.new
       return '' if url_search_params.unsupported
 
+      title_elm = render_title(url_search_params)
+
       h('div#game_filters.game_row', { key: @header }, [
         h(:h2, 'Filters'),
-        render_title(url_search_params),
-        render_reset,
+        title_elm,
+        render_reset(title_elm),
       ])
     end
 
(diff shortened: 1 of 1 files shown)


assets/app/view/game_row_filters.rb
```diff
@@ -41,7 +41,12 @@ def render_title(url_search_params)
 
     def render_reset
       attrs = {
-        on: { click: -> { update_filters('') } },
+        on: {
+          click: lambda do |_|
+            `document.querySelector('#game_filters select').value = ''`
```
michaeljb: We should avoid writing raw JS like this when possible. Up in `render` you could capture the return value for `render_title`, pass that element as a new argument to `render_reset`, and then use the `Native` API to get to the value directly (there are a few examples of using `Native` and `elm` like this throughout the view codebase):

```suggestion
          click: lambda do
            Native(title_elm)&.elm&.value = ''`
```

philcampeau: Thanks, done!

michaeljb (approved):

Issue #7614, opened 2022-04-13
[1862]Some maybe unnecessary fix


NoodleGuy2:
So implementation in online game are against the rule,but these situations barely occur and it might be unnecessary to fix them,I will list them here for you to decide:
1.In a company's first OT,if its home base tile does not contain a track tile yet,it has to build a track,according to rulebook page 12,top right corner paragraph.But in online games,you can skip the building action.
2.In a company's first OT,if the company needs to upgrade track to make room for its home station marker,the upgrade action will happen AFTER the option to merge,the marker will be put aside by the station and treat as it has full connection to all tracks on the tile,see more explanation here:[https://boardgamegeek.com/thread/2837311/get-confused-first-operating-round-rules](url),there's a FAQ and a official ruling by Mike in it.
In online games,it seems that the company is allow to build track before merge.
3. When a player under obligation fails to float that chartered company,he cannot sell shares he bought in the current SR to pay the fine(I asked Mike and that's what he said).But in online games,he can.

roseundy:
#1 this was not my intention. Do you have a link to a game where this doesn't happen?

#2 will be difficult to implement, but I'll take a look. This had not been clarified when I implemented this on 18xx.games.

As far as #3 goes, I'm pretty frustrated. I asked Mike about this on BGG and never saw any clarification. He still hasn't clarified what happens to a player that loses all their money.

NoodleGuy2:
1. [https://18xx.games/game/79750](url)  it's my turn,and if you click "skip(track)",you can skip the track building.
3 .I send Mike a Geekmail yesterday and this is a screenshot of his reply.I think by saying this he means the fine is pay during SR and you cannot sell shares you bought during this SR to pay it,consistent with his clarification on solo rule here,when he said "1. you're not allowed to sell it until the next SR"in the reply to Wilbert.[https://boardgamegeek.com/thread/2377083/article/34159073#34159073](url)
![QQ图片20220413223037](https://user-images.githubusercontent.com/102156539/163205811-a1f5773e-d222-4e7a-80b6-fb9a45619655.png)

roseundy:
OK, thanks and thanks for posting this info on BGG too.

NoodleGuy2:
ya I did.[https://boardgamegeek.com/thread/2840603/article/39866329#39866329](url)

philcampeau:
The PR linked above should fix problem 1 in your list.

PR #12200, opened 2025-10-29
[1862] require corp to lay home tile

philcampeau:
## Before clicking "Create"

### Explanation of Change

Per rule 7.2, a corporation _must_ lay its home tile on its first operating turn. Since all non-offboards in this game start white, I've added a check in the process_path method to raise a GameError if the hex_by_id(entity.coordinates).tile.color == :white.

This was discussed in Issue #7614 . I didn't label this PR as fixing that issue, because there are other issues raised there. 

Any games where they passed the Track step without laying the home tile will need to be pinned, and there's no way around that.

### DIFF
--- a/lib/engine/game/g_1862/step/track.rb
+++ b/lib/engine/game/g_1862/step/track.rb
@@ -25,6 +25,19 @@ module Engine
             super
           end
 
+          def process_pass(action)
+            raise GameError, "Must lay home track tile for #{action.entity.name}" if must_lay_track?(action.entity)
+
+            log_pass(action.entity)
+            pass!
+          end
+
+          def must_lay_track?(entity)
+            return false unless entity == current_entity
+
+            @game.hex_by_id(entity.coordinates).tile.color == :white
+          end
+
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1862/step/track.rb
```diff
@@ -25,6 +25,19 @@ def lay_tile_action(action, entity: nil, spender: nil)
...
+            raise GameError, "Must lay home track tile for #{action.entity.name}" if must_lay_track?(action.entity)
+
+            log_pass(action.entity)
+            pass!
+          end
+
+          def must_lay_track?(entity)
+            return false unless entity == current_entity
+
+            @game.hex_by_id(entity.coordinates).tile.color == :white
```
roseundy, line 38: Technically, this should really be checking to see if the home token was laid, since the company is required to _upgrade_ a tile if there isn't room for the home token.

But considering I didn't code any checks for this, this is fine.

roseundy (approved):

PR #12201, opened 2025-10-30
[18Mag] Corrects assignment of priority deal after draft

### LINKED ISSUE #6524: 18Mag PD after the draft is given to the wrong player
I think the PD is given to the wrong player after the draft. Apparently it should stay with the player that acted first in the last round of drafting. Currently, it passes to the player after the one that started the last round of drafting.

Details including ruling are explained here:
https://boardgamegeek.com/thread/2760909/article/38886169#38886169

philcampeau:
Fixes #6524 

## Before clicking "Create"

- [x] Tests pass cleanly with `docker compose exec rack rake`  **SEE BELOW**

### Explanation of Change

Currently, priority passes to the left at the end of the auction, but in a [BGG thread](https://boardgamegeek.com/thread/2760909/article/46813561#46813561), designer Lonny confirmed that the intention was for the priority to remain in place after the draft. This fix corrects that. 

Unfortunately, yet again there's really nothing that can be done to preserve existing games beyond pinning them. Changing the assignment of priority has a significant impact on the way the game plays out. Even if everyone passes in the subsequent stock round, the priority would be in the wrong place for the one after that, and so on.


philcampeau:
Yes I should have mentioned, this will clearly cause the fixture files to fail. I've been told that we don't want to remove fixtures anymore. That said, the entire turn order for the game will be changed.  The only way of scripting a fix to these fixtures would be to insert passes where it doesn't line up, but I'm not sure even that would work.

roseundy:
I think you are going to have to remove the fixtures. This is too big a change.

lib/engine/game/g_18_mag/step/simple_draft.rb
```diff
@@ -177,7 +177,7 @@ def process_bid(action)
               raise GameError, 'Logic error: must specify minor or corporation on bid action'
             end
 
-            @round.next_entity_index!
+            @round.next_entity_index! unless finished?
```
roseundy, line 180: Looks reasonable. You've tested this?

philcampeau: yep, this works because the draft is always for a fixed amount of items. I tested and the priority always landed on the same player who started the last round of drafting.

philcampeau:
Removed the broken fixture files and added a new one.

PR #12202, opened 2025-10-30
[Core/1862] Confirm pass (when under obligation in 1862)

### LINKED ISSUE #10322: 1862 - warning when under obligation in stock round
We recently had a game of 1862 where a player forgot he was under obligation in a stock round and was penalised for not floating the company.

Is it possible to put a red warning bar or something to remind a player in a stock round if they are under obligation?

Also when he won the aution in the parliament round he forgot he could but extra shares - the game does say you can do it but he said it was not obvious how - you click on the charter as usual to buy a share but he said this wasn't immediately obvious.

Thanks

philcampeau:
Fixes #10322

## Before clicking "Create"

### Explanation of Change

In 1862, if you leave a stock round without floating certain corporations, they are closed and the owner of the presidency is penalised for it. A user requested some kind of warning to users when they pass in the stock round while still under obligation to float a corporation. 

As I was working on this, I found the most logical solution here would be to have a yellow `Confirm?` bar like the one used for consenting train buys from other players, among other things. This was a bit outside of my normal wheelhouse, but I believe I've accomplished what I was aiming to do.

This PR adds a `confirm_pass?` check to /lib/engine/step/base.rb that defaults to false, as well as a `confirm_pass_text` default. Then it modifies the relevant elements of /assets/app/view (pass, pass_button, and pass_auto_button) to implement this check. 

I imported more that three dozen completed games (all different titles) from the site, and all loaded fine. From what I can tell, there's no reason why this would impact any existing games, since they'll all have the `confirm_pass?` method defined as false by default. 

I recognise that I'm implementing changes to core functions of the site, so please scrutinize every character of this PR!

### Screenshots

image

### DIFF
--- a/assets/app/view/game/pass.rb
+++ b/assets/app/view/game/pass.rb
@@ -10,11 +10,43 @@ module View
       needs :actions, default: []
 
       def render
+        # This lambda runs *before* the PassButton executes its action
+        confirm_before = lambda do
+          # Recompute step on every click
+          step =
+            if @game.round.respond_to?(:active_step)
+              @game.round.active_step
+            else
+              @game.round
+            end
+
+          needs_confirm = step.respond_to?(:confirm_pass?) && step&.confirm_pass?
+          message =
+            if step.respond_to?(:confirm_pass_message) && (m = step.confirm_pass_message).to_s.strip != ''
+              m
+            else
+              'Are you sure you want to pass?'
+            end
+
+          do_pass = -> { process_action(Engine::Action::Pass.new(@game.pass_entity(@user))) }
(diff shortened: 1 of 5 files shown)


ollybh:
An alternative could be to add `help` text that lists the player's unfulfilled obligation(s) and the terrible penalties that will be incurred if the company is not floated before the end of the stock round.

philcampeau:
@ollybh ... well that is a much simpler solution, yes! tbh I don't think I'd ever noticed a help module before. 

I'll delete this PR and submit a much simpler one :D

PR #12207, opened 2025-10-31
[1862] Adds help text for unfloated chartered companies

### LINKED ISSUE #10322: 1862 - warning when under obligation in stock round

philcampeau:
Fixes #10322 

## Before clicking "Create"

### Explanation of Change

Adds a `help` message above the Pass button in the Stock Round if the player is still under obligation to float one or more Chartered Companies.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1862/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_1862/step/buy_sell_par_shares.rb
@@ -10,7 +10,7 @@ module Engine
           UNCHARTERED_TOKEN_COST = 40
 
           def help
-            return if current_entity.companies.none?
+            return if current_entity.companies.empty?
 
             company_plural = current_entity.companies.size > 1 ? 'Companies' : 'Company'
 


lib/engine/game/g_1862/step/buy_sell_par_shares.rb
```diff
@@ -9,6 +9,14 @@ module Step
         class BuySellParShares < Engine::Step::BuySellParShares
           UNCHARTERED_TOKEN_COST = 40
 
+          def help
+            return if current_entity.companies.none?
```
ollybh: Don't use `none?` or `all?` without a block. `empty?` is more efficient.
```suggestion
            return if current_entity.companies.empty?
```

PR #12211, opened 2025-11-01
[1824 Family] Only show single MR when doing forced MR exchange, fixes #12205

### LINKED ISSUE #12205: [1824] exchanging the 'wrong' mountain railway hangs the game
Game: https://18xx.games/game/229838

Like a good little QA tester I am, I exchanged B6 mountain railway during my turn instead of B2. Cue hanging the game and a little JSON bug in F12:

Uncaught corporation?: undefined method `corporation?' for nil

perwestling:
Fixes #12205 

## Before clicking "Create"

This should not break any games as it is an UI change.

## Implementation Notes

When in forced MR exchange, ensure only one exchange button appear for a player, even if owning more than one. Need to do exchange one at a time.

### Explanation of Change

If player select another MR than the one with lowest number, the game crashes and is broken.

But even more so, allowing exchange of non-lowest MR is breaking the rules, as a player might be doing an exchange in front of a player with a lower MR number. 

The alternative would be to show some error message if exchange is attempted with wrong MR, but I think not showing later ones is a better UX.

### Screenshots

N/A

### Any Assumptions / Hacks

N/A

### DIFF
--- a/lib/engine/game/g_1824/game.rb
+++ b/lib/engine/game/g_1824/game.rb
@@ -471,7 +471,7 @@ module Engine
           # In case of a forced MR exchange, we only want to show one MR for a player
           # as the player can only exchange the MR with the lowest number. We do not
           # do this always as a player can do an unforced exchange with any MR.
-          candidates.group_by(&:owner).transform_values(&:first).values
+          candidates.group_by(&:owner).map { |_o, c| c.first }
         end
 
         def mountain_railway?(entity)


lib/engine/game/g_1824/game.rb
```diff
@@ -465,7 +465,17 @@ def coal_railway_exchangable?
...
+          return candidates if forced_mountain_railway_exchange.empty?
+
+          # In case of a forced MR exchange, we only want to show one MR for a player
+          # as the player can only exchange the MR with the lowest number. We do not
+          # do this always as a player can do an unforced exchange with any MR.
+          one_per_owner = []
+          candidates.each do |c|
+            one_per_owner << c unless one_per_owner.find { |o| o.owner == c.owner }
+          end
+          one_per_owner
```
michaeljb: Would you mind testing out this code change? This would be more idiomatic in Ruby.

```suggestion
          candidates.group_by(&:owner).transform_values(&:first).values
```

michaeljb: Or I think it could be done in a `map` instead of `transform_values` and `values`:

```suggestion
          candidates.group_by(&:owner).map { |_o, c| c.first }
```

perwestling:
Managed to push wrong commit here, so did a force push back to first commit.

perwestling:
It seems both suggestions work, so I have chosen the map one.

michaeljb (approved):

PR #12214, opened 2025-11-01
[1824 Family] Consider remaining MR for valid actions left, fixes #12195

### LINKED ISSUE #12195: [1824] Auto pass "no valid actions" ignores Mountain railway
TL;DR:  I don't think the valid actions check is taking into account the Mountain Railway exchange.

https://18xx.games/game/228392

I had two Mountain railways in the SR before the SD formed, I started the BH as well as the CL. I exchanged one Mountain Railway for a CL, bought one more CL, then set the auto buy to float the BH, expecting to be able to exchange the other Mountain for the fifth CL share to float it as well. 

But once I bought the last BH I had no money left and got "tomhoward87 has no valid actions and passes" (as did everyone else) and the next OR started. As it happened someone bought the 4 almost immediately so I was able to exchange the MR and float the CL anyway, but I don't think the valid actions check is taking into account the MR exchange.

perwestling:
If there is a possible MR exchange available, ensure that pass is available actions, so that game do not skip player when no other possible action left.

Fixes #12195

## Before clicking "Create"

This should break some games (as there should be out of order stock actions) but for some reason all games validate OK.

## Implementation Notes

Should only add Pass as an possible action, if there is no Pass among actions, and there is a legal possibility to exchange an MR.

### Explanation of Change

Now the user will get a Pass, and then they can click on one of the regionals and get a button choice to Exchange for each MR they have. (Exchange is a buy action, so only one per turn.)

### Screenshots

N/A

### Any Assumptions / Hacks

N/A

### DIFF
--- a/lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
+++ b/lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
@@ -24,6 +24,8 @@ module Engine
               actions << 'pass' if actions.one?
             end
 
+            actions << 'pass' if add_pass_to_allow_mr_change?(entity, actions)
+
             # If debt exists, add actions to pay off
             if player_debt.positive? && entity.cash.positive?
               actions << 'payoff_player_debt'
@@ -160,6 +162,15 @@ module Engine
             return false unless exchange
 
             !@game.mountain_railway_exchangable? || !@game.exchangable_for_mountain_railway?(entity, corporation)
+          end
+
+          def add_pass_to_allow_mr_change?(entity, actions)
+            # There might be that we have no other possible actions, but still have an MR we might want to exchange
+            return false unless entity.player?
+            return false unless @game.mountain_railway_exchangable?
+            return false if bought? || actions.include?('pass')
+
+            @game.companies.find { |c| @game.mountain_railway?(c) && c.owner == entity }
           end
         end
       end


perwestling:
Then doing Validate for all 1824 games in the latest backup no game fail.

But if I try https://18xx.games/game/228392 in the browser it does fail at action 226.

And I believe it is a correct fail as:
- tomdidiot did not have any valid actions previously at action 226, but it does now have an MR exchange (owning B3)
- When happened in the earlier game is that:
  -- steak had no valid (but owned a MR as well)
  -- Tekopo buys another BK
  -- tomhoward87 had an autobuy for BH
So the next action in the game was for Tekopo to buy a share, but as tomdidiot is now the active player, the game breaks.

Question is why this does not fail when doing Validate?
Is there anything in the changes that differ between Ruby and JS?

So I need to look more into this before putting this as non-draft.

lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
```diff
@@ -161,6 +163,15 @@ def invalid_mountain_railway_exchange?(entity, corporation, exchange)
...
             !@game.mountain_railway_exchangable? || !@game.exchangable_for_mountain_railway?(entity, corporation)
           end
+
+          def add_pass_to_allow_mr_change?(entity, actions)
+            # There might be that we have no other possible actions, but still have an MR we might want to exchange
+            return false unless entity.player?
+            return false unless @game.mountain_railway_exchangable?
+            return false if bought? || actions.include?('pass')
+
+            @game.companies.find { |c| @game.mountain_railway?(c) && c.owner == entity }
```
perwestling, line 173: Should we check for closed as well, just to be sure that closed MRs does not affect this?

perwestling:
[#12262](https://github.com/tobymao/18xx/pull/12262) contains a much cleaner solution. So I am closing this PR as it is no longer needed.

PR #12218, opened 2025-11-03
[1852 MO] fix East St Louis as optional stop

### LINKED ISSUE #12199: [1852 Missouri] East St Louis not correctly being skipped in revenue calculation
[1852 Missouri] East St Louis not correctly being skipped in revenue calculation
https://18xx.games/game/227581

SSW running a 6 train from Topeka to Joliet going via Chapin J4 earns $400 
$400 is correct, but the route shows J4 is being skipped. I don't think the city Chapin can be skipped. It should be K7 the skippable East St Louis.
A7-C7-G9-J8-K7-(J4)-K3 + E/W

Running the train from Topeka to Joliet via going via Springfield K5 earns $300
But it shows Topeka A7 is skipped, so the E/W bonus is lost. Again it gets the wrong city skipped, only East St Louis is skippable.
So the revenue on this route should be $410
(A7)-C7-G9-J8-K7-K5-K3

Chrome Version 141.0.7390.123 (desktop)

Image

michaeljb:
the existing code removes the lowest-value city instead of East St Louis, and contrary to the comment, doesn't even make sure the value of the removed city is lower than $30

Fixes #12199

Screenshot 2025-11-02 at 21 57 48

### DIFF
--- a/lib/engine/game/g_18_mo/game.rb
+++ b/lib/engine/game/g_18_mo/game.rb
@@ -378,9 +378,8 @@ module Engine
           # return if not enough normal stops
           return visits if normal_stops.size < paying_distance['pay']
 
-          # remove smallest normal stop that has less revenue than dit
-          worst_stop = normal_stops.min_by { |stop| stop.route_revenue(route.phase, route.train) }
-          visits.reject { |stop| stop == worst_stop }
+          # ignore the optional dit stop
+          normal_stops
         end
 
         def setup_turn


lib/engine/game/g_18_mo/game.rb
```diff
@@ -378,9 +378,8 @@ def compute_stops(route)
           # return if not enough normal stops
           return visits if normal_stops.size < paying_distance['pay']
 
-          # remove smallest normal stop that has less revenue than dit
-          worst_stop = normal_stops.min_by { |stop| stop.route_revenue(route.phase, route.train) }
-          visits.reject { |stop| stop == worst_stop }
+          # ignore the optional dit stop
+          normal_stops
```
roseundy, line 382: Not sure what the hell I was thinking when I wrote this.

roseundy (approved):
Looks OK.

PR #12221, opened 2025-11-08
[System18] Update Gotland to not nationalize in yellow phase

### LINKED ISSUE #12071: [System18] Gotland - Nationalization offer being made in yellow
https://gist.github.com/roseundy/ca8d972476f7e0615bd673f24b8e892b

At the end of OR 1.2, DGN is being offered to nationalize even though we are still in phase 2. According to the rules, this should only happen starting in phase 3.

patrikolesen:
Fixes #12071

Will break existing games

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_system18/map_gotland_customization.rb
+++ b/lib/engine/game/g_system18/map_gotland_customization.rb
@@ -67,6 +67,9 @@ module Engine
         def map_gotland_constants
           redef_const(:CURRENCY_FORMAT_STR, '%s SEK')
           redef_const(:BANKRUPTCY_ENDS_GAME_AFTER, :one)
+          redef_const(:STATUS_TEXT, {
+                        'nr_or' => ['NR', 'Nationalization Round after Operating Rounds'],
+                      })
         end
 
         def map_gotland_game_companies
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_system18/map_gotland_customization.rb
```diff
@@ -137,7 +140,54 @@ def map_gotland_game_trains(trains)
...
+            },
+            {
+              name: '8',
+              on: '8',
+              train_limit: 2,
+              tiles: %i[yellow green brown gray],
+              operating_rounds: 2,
+              status: ['nr_or'],
+            },
+          ].deep_freeze
```
roseundy, line 190: As an alternative to essentially redefining S18_INCCAP_PHASES just to add status lines, you could just add the status to post_game_phases. See Poland for an example.

roseundy (approved):

lib/engine/game/g_system18/map_gotland_customization.rb
michaeljb: Since you approved this I'm gonna merge, and this could be changed in a follow-up if y'all want to.

PR #12227, opened 2025-11-09
[18Norway] Updated dev stage to beta

patrikolesen:
## Before clicking "Create"


michaeljb:
Made a couple related changes in https://github.com/tobymao/18xx/commit/791100fe0f415fe96522145a86b35bbe8f2c5427

PR #12229, opened 2025-11-10
[SteamOverHolland] Fixes train discount bug, move to beta

### LINKED ISSUE #11163: [Steam over Holland] can only use p6 if full amount in treasury

philcampeau:
Fixes #11163

## Before clicking "Create"

### Explanation of Change

**NOTE**: No archives or pins necessary. The only active game doesn't have the affected private in the game.

The reason this wasn't working was that I had defined emergency_issuable_bundles(entity) to only check against the @depot.min_depot_price. I created a new method, min_depot_price(entity), to apply a 10% discount to that price if the corp owns the discount private. 

Then for good measure I updated every method that checked @depot.min_depot_price to use the new method instead.

As this is the only bug that has been opened for the game in over a year, I moved it to Beta as well.

### DIFF
--- a/spec/lib/engine/fixtures_required_spec.rb
+++ b/spec/lib/engine/fixtures_required_spec.rb
@@ -45,6 +45,7 @@ SKIP_BETA_PROD = {
   '18ZOO - Map E' => %i[stock_market fixed_round],
   '18ZOO - Map F' => %i[stock_market fixed_round],
   '18ZOO' => %i[stock_market fixed_round],
+  'Steam Over Holland' => [:stock_market],
 }.freeze
 
 def fixtures(meta)


philcampeau:
I'll need to put some effort into including a fixture for the other end game trigger. Actually hitting the top of the market in this game is one heck of a challenge!

ollybh:
> I'll need to put some effort into including a fixture for the other end game trigger. Actually hitting the top of the market in this game is one heck of a challenge!

If you can't manufacture a fixture with this game end condition then you can add an exception to the `SKIP_BETA_PROD` hash in `spec/lib/engine/fixture_required_spec.rb`.

philcampeau:
haha that's funny ollybh, I finally gave up trying today, and I was just adding the exception and trying to figure out why there was another commit I hadn't seen yet. Turns out you did it an hour ago!

Issue #12193, opened 2025-10-29
Lazy Engine


michaeljb:
**tl;dr**: eagerly computing the graph, actions for all entities, the current step, etc can be expensive, maybe we can get some performance gains with a [lazier](https://en.wikipedia.org/wiki/Lazy_evaluation) approach

## High-level Goals

* stop computing graph whenever possible, so that when #10331 hits the impact is lower
* step-based ability timing is kind of a circular headache that can be simplified if steps are never skipped, but instead must be explicitly `pass`ed
* improve performance 
    * this [stackprof output](https://gist.github.com/michaeljb/5f29c1670b3892328ce4d06d6513b8f5) shows nearly 20% of the time of processing a game is spent in `skip_steps!`, and over 20% of the time spent in computing `actions` for various entities (together they don't take over 40% of the time, there is some overlap between the two)

## Implementation Ideas

* validation can happen the first time an action is processed, after that the `Engine` can trust the action is valid and just process it

    * `process_<action_name>` methods should just process the action, not validate
    * `Step` classes should have new `prevalidate_<action_name>` (and `postvalidate_`?) methods
    * `strict` and `game.loading` should have consistent naming, and modes reworked to:
        * fully strict - every action is strictly validated (validation script, optionally in browser)
        * new action - when a new action is added, it is strictly validated, but all prior actions are not (when the server receives a new action and processes the game; [current source](https://github.com/tobymao/18xx/blob/10ec022d557f58ac62cbae0cf981d39a466870ab/routes/game.rb#L107-L112))
        * zero strict - just process the actions (loading game in browser)

* after processing an action, the `step` updates a new `@step` field on the `Action`
    * when the game is being loaded in the future, [`step_for`](https://github.com/tobymao/18xx/blob/10ec022d557f58ac62cbae0cf981d39a466870ab/lib/engine/round/base.rb#L119-L130) doesn't need to run to find the processing step
    * infer `pass` actions? If the next action to process has a step that does not match the currently active step, infer the necessary `pass` and move on

* [`skip_steps`](https://github.com/tobymao/18xx/blob/10ec022d557f58ac62cbae0cf981d39a466870ab/lib/engine/round/base.rb#L209-L216) should not be used
    * players won't want to manually pass steps where they have no valid actions, so `auto_actions` will need to be more heavily used for `pass`es
    * better mechanism for explaining why a step was auto-skipped, e.g., #11824

[...]

PR #12234, opened 2025-11-11
[Engine V2] `step` field on actions

michaeljb:
#12193

This PR depends on #12217.

## Before clicking "Create"

### Explanation of Change

* `Action::Base`
    * add a `step` attr to `Action` objects
* `Round::Base`
    * after processing an action, set the `Action`'s `step`
    * before processing an action, if it has a `step`, use it to quickly select which `Step` to use; otherwise, use the old/non-lazy method of checking all `Step`s in order for one that can process the given action, and throwing an error if a blocking `Step` that cannot process the action is found
* update 1889 fixtures
    * add `step` to actions
    * some `undo` and `redo` actions have been removed
* if the `Game`'s `use_engine_v2` is not `true`, `step` is not set on any actions, and the way `Round` processes actions is unchanged

### Screenshots

1889 fixture showing faster processing with the `step` field changes. Legacy on top (via `v2=f`) processed the actions in 0.379 seconds, and Engine V2 (`v2=t`) on the bottom took 0.263 seconds:

Screenshot 2025-11-10 at 22 35 55

The above screenshot is one of many trial results:

Screenshot 2025-11-10 at 22 54 24

### Any Assumptions / Hacks

* the high-level assumption is that once a step has processed an action, it is safe to rely on that same step to process the action when the game is processed in the future
    * if there is a bug fix that _should_ break a game by introducing a blocking step where there wasn't one before, that game will be processed by the engine without error, even though we want an error
        * a `TODO` comment has been added to address this, and a checklist item has been added to [#12193](https://github.com/tobymao/18xx/issues/12193).


lib/engine/round/base.rb
```diff
@@ -39,6 +39,33 @@ def initialize(game, steps, **opts)
...
+              grouped_steps.each.with_index do |grouped_step, index|
+                key =
+                  if index.zero?
+                    step_type
+                  else
+                    "#{step_type}#{index + 1}"
+                  end
+                steps_h[key] = grouped_step
+                steps_h[grouped_step] = key
+              end
```
ollybh, line 64: This seems over-optimised for a block that's going to get called just once a round. Most of the code on both branches of the `if` statement is repeated.

```suggestion
            grouped_steps.each_with_index do |grouped_step, index|
              key = "#{step_type}#{index.zero? ? '' : index + 1}"
              steps_h[key] = grouped_step
              steps_h[grouped_step] = key
            end
```

I've tested this and it's a few milliseconds faster than your code, mostly because `each_with_index` is more efficient that `each.with_index` (it doesn't need to create an Iterator). The change in the inner block is basically the same speed.

ollybh (commented):
This looks like a good change to me, though I've got a suggestion for a tweaked algorithm for calculating the lookup hash for steps.

Issue #12233, opened 2025-11-11
[18RoyalGorge] 6x2 trains quantity should be unlimited


crericha:
The implementation only has 5 6x2 trains; however, in the physical copy, these trains show an infinity sign for quantity. I believe the number of 6x2 trains should be raised to something like 10 (5 companies x 2 trains each). In the most extreme case, you could have all the 4+ and 5+ exported.

michaeljb:
Fixed by #12236

PR #12236, opened 2025-11-12
[core/18RoyalGorge] support `unlimited` trains

michaeljb:
Ever since #1846 added unlimited tiles, I've wanted the same thing for trains, just never wanted it badly enough till #12233 was reported I guess. 🐌 

----

Use `unlimited` trains in 18RoyalGorge (#12233) and 1889. There are many more games that can use this, but starting with just these two to minimize potential breakages.

* for `unlimited` trains, initialize just one `Train`, or enough so that the latest one with a defined `event` is created
* when an `unlimited` train is removed from the `Depot`'s `@upcoming`, a new `Train` with `@index + 1` is created
* display '∞' upcoming trains for `unlimited` trains
* `Engine::Game::Base#update_trains_cache` so that `train_by_id` works after creating the new train

Other changes:

* dedupe some code in `Depot`
    * delete `unshift_train` and replace calls to it with `insert_train` with `index = 0`
    * `forget_train` calls `remove_train` instead of just repeating the code
    * `add_train` calls `insert_train` instead of just repeating the code
* change `Game::Base#num_trains` to use `train[:num]` by default instead of being not implemented, this simplifies the code in `init_train_handler`, enables the changes needed for `unlimited` trains, and doesn't require subclasses that override it to change anything
* implement `Train#to_s`

----

- [x] Existing games shouldn't break, I'll do a local validation to check for sure.

## Before clicking "Create"


assets/app/view/game/borrow_train.rb
```diff
@@ -78,9 +78,10 @@ def remaining_trains
 
         rows = @depot.upcoming.group_by(&:name).flat_map do |_, trains|
           names_to_prices = trains.first.names_to_prices
+          train_count = trains.first.unlimited ? '∞' : trains.size
           [h(:div, names_to_prices.keys.join(', ')),
            h(:div, names_to_prices.values.map { |p| @game.format_currency(p) }.join(', ')),
-           h(:div, trains.size)]
+           h(:div, train_count)]
```
crericha: If the count is only used here, do we need the var?

assets/app/view/game/buy_trains.rb
```diff
@@ -628,9 +628,10 @@ def remaining_trains
         }
 
         rows = @depot.upcoming.reject(&:reserved).group_by(&:name).flat_map do |_, trains|
+          train_count = trains.first.unlimited ? '∞' : trains.size
           [h(:div, @game.info_train_name(trains.first)),
            h(:div, @game.info_train_price(trains.first)),
-           h(:div, trains.size)]
+           h(:div, train_count)]
```
crericha: same

lib/engine/game/base.rb
```diff
@@ -2647,7 +2651,7 @@ def game_companies
 
       def init_train_handler
         trains = game_trains.flat_map do |train|
-          Array.new((train[:num] || num_trains(train))) do |index|
+          Array.new(num_trains(train) || train[:num]) do |index|
```
crericha: I don't think `|| train[:num]` is needed now.

crericha (approved):

crericha:
Nice refactor!

lib/engine/game/base.rb
michaeljb: Removing it breaks 1812 and 1840 because of how they override `num_trains`. (I only found out from the GitHub workflows failing)

We could just make sure anything overriding it always returns the right value.

michaeljb: 70709acfd

assets/app/view/game/borrow_train.rb
michaeljb: 0a1c592fb

assets/app/view/game/buy_trains.rb
michaeljb: 0a1c592fb

lib/engine/game/base.rb
crericha: Yeah, let's do that. One source of truth.

lib/engine/depot.rb
```diff
@@ -70,30 +70,28 @@ def max_depot_price
...
-      @depot_trains = nil
+      remove_train(train)
+    end
+
+    def delete_from_upcoming(train)
+      index = @upcoming.index(train)
+      deleted = @upcoming.delete(train)
+      return if !index || !deleted&.unlimited
+
+      insert_train(train.clone_unlimited, index)
```
crericha, line 89: Is index correct here? If you have 3 unlimited trains (due to an event on 3rd train), when the first unlimited train is removed, will this insert a new train at that position and thus we'll never get to the 3rd train?

michaeljb: That’s a good call. I added this to fix an 1889 game where a D was bought before the second 6, wasn’t thinking about events at that moment.

So the answer I think is to get the index of the last train with the same name as the removed one before it is actually removed.

crericha: Something like that would work. It may be simpler to check if index + 1 has the same name.

`return if !index || !deleted&.unlimited || @upcoming[index + 1]&.name == train.name`

michaeljb: Oh yeah, of course if there's already another copy of the train in `@upcoming` you don't need to insert one at all 😅

crericha: Looks good

crericha (approved):

lib/engine/depot.rb
michaeljb: 8ced8fba8 - didn't need `index + 1` since it's after the `delete` call

crericha: Ahh, right. Good catch.

michaeljb: Hmmm we'll need some way to disable the `clone_unlimited` call so that all the 2-trains can be exported in 1817 variants

https://github.com/tobymao/18xx/blob/0d0332895d1e4a6357c6348b498fb43e46bce213/lib/engine/depot.rb#L27-L33

Setting those to `unlimited` right now hits an infinite loop when the `export_all!` call is reached, as they are continually spawned and removed.

crericha: Right, that logic will need to be refactored. Something like `remove_train(train, export = true)`.

lib/engine/game/base.rb
```diff
@@ -2659,8 +2663,12 @@ def game_trains
...
       end
 
-      def num_trains(_train)
-        raise NotImplementedError
+      def num_trains(train)
+        if train[:num] == 'unlimited'
+          (train[:events] || []).filter_map { |e| e['when'] }.max || 1
+        else
+          train[:num]
+        end
```
ollybh, line 2671: @michaeljb Several games have already implemented their own version of `num_trains` and won't be inheriting this code. What should they do if they want to add unlimited trains?

For example, the 1858 family have versions that look like:
```
def num_trains(train)
  SOME_CONSTANT[train[:name]]
end
```

Setting the train count to `unlimited` in the constant won't work.

philcampeau:
You're going to run into issues with games like 1817 which have an unlimited number of an early rank of trains and use export_all to sweep them away at some point:

https://github.com/tobymao/18xx/blob/d1faff307d046d31e2f8832ff31d66938477a141/lib/engine/depot.rb#L27-L33

lib/engine/game/base.rb
philcampeau: Looks like for 1858 you can do this:

```
def num_trains(train)
  counts = two_player? ? TRAIN_COUNTS_2P : TRAIN_COUNTS_NORMAL
  counts[train[:name]] || super
end
```

and put the `num: 'unlimited',` value directly in the trains constant.

Should be simple enough to apply that to other games too.

ollybh: @philcampeau That wouldn't call the code that calculates how many unlimited trains need to be created, so events still occur at the right point.

philcampeau: hmmmm

Ok, I'll have to look more closely at it.

Issue #12275, opened 2025-11-29
[1824] Impossible to buy a BH share and setup a par price.


barattie:
Game ID: 232142

Link: https://18xx.games/game/232142

Four shares of BH are in the hands of players, 3 for me and 1 for Tirerndil.
All these shares were received in exchange for Mountain Railways certificates.
Apparently, I received the presidency, although the rules say:
_"The share price of a Regional Company without an associated Coal Railway will
be determined later by the player who buys the 20% Director‘s Share Certificate
(see section VI.5)."_
I think that when I attempt to buy a BH share, I should be presented with the choice of par price, but this does not happen (screenshot 1).
When I click on "Buy Bank Share", an error pops out in the top bar: "undefined method 'price' for nil" (screenshot 2).

I'm almost sure that the same error happened my previous stock turn: I attempted to exchange my last Mountain Railway for the 5th BH share (which would have floated the company - but without a par price!) and the action was not accepted. At the moment I thought I was going against some rule, and I made another move; but now I think it was the same bug.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/f0c99ea4-a15d-4eca-b100-f1cec2e47784" />
<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/c0494853-440c-4694-838a-51cd53875ece" />

Please include screenshots (you can drag and drop or upload them into this issue)

-->

perwestling:
Yes, this was a bug.

I have now pushed a PR that fixes that for any unassociated regional (like BH and any other regional where no one buys the associated Coal Railway in SR1).

It might take some time before the correction is deployed.

PR #12317, opened 2025-12-26
[1824] Correct floating for non-parred regional, fixes #12228, #12275

### LINKED ISSUE #12228: 1824 Mountain Exchange error to BH    Id: 229952
In anticipation of exchanging Mountains I cloned the game 229952.  If a player exchanges Mountains for 2+ of the unfloated BH, it doesn't allow you to choose a par price and thereafter it is unpar-able giving an error that no one can buy a share for a value of 0.
-->

**Which Browser / Platform?**

perwestling:
Fixes #12228 and #12275

This correction is not expected to affect any ongoing games. If it does, it might that this bug existed in that game.

## Implementation Notes
- If a corporation (regional) is not parred, doing MR exchange should not give presidency to anyone
- If a corporation is not parred, it cannot be floated by an MR exchange
- Text for needed for floating has been updated with "Par to float" for the not-parred-but-50%-exchanged case

### Explanation of Change
This is to follow the 1824 rules for exchange.

To recreate the issue clone the game mentioned in the issue at start of SR3, and then using exchange on BH.

### Screenshots
Here is a clone of the game mentioned in the Issue, where MR only have been used to get BH shares.
image

Yet another BH exchanged reaching 50%, but not floated as not parred.
image

And if later BH is parred it now looks like.
image

### Any Assumptions / Hacks
N/A

### DIFF
--- a/lib/engine/game/g_1824/game.rb
+++ b/lib/engine/game/g_1824/game.rb
@@ -459,6 +459,15 @@ module Engine
           corporation_by_id(exchange_ability.corporations.first)
         end
 
+        def regional_railway_with_active_association?(regional)
+          @corporations.any? do |corp|
+            next false unless coal_railway?(corp)
+            next false if corp.closed?
+
+            get_associated_regional_railway(corp) == regional
+          end
+        end
+
         def coal_railway_exchangable?
           @phase.status.include?('may_exchange_coal_railways')
         end
@@ -714,20 +723,18 @@ module Engine
           return super unless entity.corporation
 
           case entity.id
-          when 'BK', 'MS', 'CL', 'SB'
+          when 'BH', 'BK', 'MS', 'CL', 'SB' # Need all regionals here although some might not be associated with a coal railway
             needed = entity.percent_to_float
             if needed.positive?
-              need_exchange = entity.floatable ? '' : ' + exchange '
+              need_exchange = entity.floatable || !regional_railway_with_active_association?(entity) ? '' : ' + exchange '
               "#{entity.percent_to_float}%#{need_exchange} to float"
-            else
+            elsif regional_railway_with_active_association?(entity)
               'Exchange to float'
+            else # An unassociated regional that has reached 50%+ by MR exchanges
+              'Par to float'
             end
-          when 'UG'
-            'UG formation to float'
-          when 'KK'
-            'KK formation to float'
-          when 'SD'
-            'SD formation to float'
+          when 'UG', 'KK', 'SD'
+            "#{entity.id} formation to float"
           else
             super
           end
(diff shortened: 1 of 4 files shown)


perwestling:
Also fixes #12275

lib/engine/game/g_1824/game.rb
```diff
@@ -460,6 +460,17 @@ def get_associated_regional_railway(minor)
...
 
+        def regional_railway_with_active_association?(regional)
+          coal_railways = @corporations.select do |c|
+            coal_railway?(c) && !c.closed?
+          end
+          coal_railways.each do |coal|
+            target = get_associated_regional_railway(coal)
+            return true if target == regional
+          end
+          false
```
ollybh: The structure in your code:
```
Array.each do |e|
  return true if TEST
end
false
```
Can be written as `Array.any? { |e| TEST }`. It can also be combined with the previous loop:

```suggestion
          @corporations.any? do |corp|
            next false unless coal_railway?(corp)
            next false if corp.closed?
            
            get_associated_regional_railway(corp) == regional
          end
```

lib/engine/game/g_1824/step/buy_sell_par_exchange_shares.rb
```diff
@@ -110,7 +110,9 @@ def process_buy_shares(action)
             company = action.entity
             bundle = action.bundle
             bundle.share_price = 0
-            buy_shares(company.owner, bundle, exchange: company)
+
+            # In case of exchange the company is passed, and also no president change allowed (according to rule IV.4.1 bullet 6)
```
ollybh: If I understand the rules correctly, this comment isn't correct: you can acquire the presidency from another player when exchanging a mountain railway, you just can't become president if the president's certificate is still in the IPO.

Your code looks correct.

perwestling: Updated comment in next commit.

perwestling:
> Is the _Par to float_ status for the BH company (and potentially other regional railways) clear enough? Would _Buy president’s certificate to float_ be better?

"Buy president's certificate to float" is quite long so it would hardly fit. 
Today we use "Exchange to float" so "Par to float" is short and contains enough info, I think. 
(This is a very edge case - 50% of a regional "bought" via MR exchange. Unlikely to ever happen.)

But if I were to change text I think it should be changed to from "20% to float" and "10% to float" to "Par to float", as par is required in both cases. (When corporation is not parred.)

Do not think it is worth the effort, though.

perwestling:
I was trying to add a unit test for this issue, and encountered an issue with the capitalization. 
So I will look into that, and add a unit test to verify this, before making it current again.

Need the PR which adds unit tests and fixtures to complete this one.

perwestling:
The first commit is the fix #12325 which is used as base (needed to get unit tests).

The second commit is the previously approved commit(s) that has been squashed.

The third commit fixes a remaining issue with floating when parring a corporation that reaches or has reached 50% sold with the parring. And commit also adds unit test case to verify #12275.

This is still in draft, and will be rebased + rereviewed when #12325 has been merged.

perwestling:
Not really any changes of previous version, but have now rebased it to latest master.

PR #12328, opened 2025-12-31
[TOP1871] disable Done button in token selection until at least one is chosen

### LINKED ISSUE #12327: [The old prince 1871] allowed to pass without choosing a token for branch
**What's your Game ID?**

`https://18xx.games/game/235530`

**How to Reproduce the Bug?**

-In SR7, Alberton was to be split to produce Hunter River Branch.
Action 499 is where the prompt to choose tokens, D2 or D14 appeared. It seems neither was chosen and perhaps instead the player performing the split just clicked done. 
Stepping to action 500 then shows that no HRB token is on the map--in neither D2 nor D14, instead both still have Alberton tokens.  And the HRB charter has no map locations for the tokens, just 0 80 80 80

-Later in the game, for example in SR8, the following text appears in a red banner at the top of the screen when trying to buy a market share (the last 10% share to float) from HRB, hunter river branch:
Undefined method 'pending tokens' for<Engine::Game::G1871::Round::Stock Round 8>

**Which Browser / Platform?**

- Browser Version
  - Chrome 143.0.7499.146
- Mobile

philcampeau:
Fixes #12327

## Before clicking "Create"

### Explanation of Change

User reports that they split a corporation and were able to select no home token for the branch corporation, which caused the game to error out when they attempted to float that corp in a subsequent SR.

Currently, when selecting tokens while splitting a corporation, the code was set to only include a Done button if the parent corporation has fewer than 4 tokens. The problem with this is that a corporation can be split more than once, and after the first time, it already has fewer than 4 tokens before any are given to the new branch corp. I modified the code to only include the Done button once the new branch corp's first token is used.

### DIFF
--- a/lib/engine/game/g_1871/round/stock.rb
+++ b/lib/engine/game/g_1871/round/stock.rb
@@ -200,6 +200,8 @@ module Engine
               split_pick_branch(branch)
             when SPLIT_PICK_TOKENS
               if choose.choice == 'done'
+                raise GameError, "#{@split_branch.full_name} must swap at least one token" unless @split_branch.tokens.first&.used
+
                 @log << "#{current_entity.name} is done swapping tokens"
                 split_next
               else


philcampeau:
Good call. Done.

PR #12325, opened 2025-12-30
[1824] Fix capitalization for unsold, fixes #12324

perwestling:
Fixes a bug when adjusting capitalization count for unsold during SR1.

Also adds a fixture for 1824, which is an ongoing game which I have cloned and ended prematurely.
This fixes #12290

Adds some unit tests for 1824. More will be added in the future. This fixes #12024

Fixes #12324
Fixes #12290 
Fixes #12024

## Before clicking "Create"

- [X] Branch is derived from the latest `master`
- [X] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [X] Code passes linter with `docker compose exec rack rubocop -a`
- [X] Tests pass cleanly with `docker compose exec rack rake`

There was a bug in which SD, KK, UG could get too low capitalization, in case any pre-staatsbahn of theirs was unsold during SR1. These games might now be broken - in case change of money do change possible actions.
So in case any game breaks, pin or archive it - 1824 is still in Alpha.

## Implementation Notes
The added fixture is a not yet completed game of 1824 which will have an UG formation with shortness of money, as the game continues. I have taken that and made a manual end, and is using that fixture to do some unit tests. I cannot fully test the capitalization for UG with this fixture as game has not reach that yet but is testing the internal state prior to this.

I might change fixture, possibly with the completed game, in the future.

Also added some test cases for first to operate in SR1, priority deal, SD formation. More might be added in the future.

### Explanation of Change
Found an error in the code when adding a unit test. This has now been corrected. No other code changes.

### Screenshots
N/A

### Any Assumptions / Hacks
N/A

Issue #12243, opened 2025-11-13
[1824] wrong assignation of UG presidency


barattie:
Game 228967

At the moment the UG was activated, two players had 50% of shares. Presidency should have been awarded to the player holding SD1, but it was given to the other player.
See description by player Tacitus below.

[14:54] Tacitus: looks like we have a problem here. I have simulated the game until the situation when bulb holds 40% of UG and I have 30% and the 5 train comes into play.
[14:55] Tacitus: According to the rules UG1 is changed into the director certificate and the UG 2 into a 10% share. Which brings us both to 50%. But since I hold the director certrificate and Bulb does not have MORE percent than I should direct the company
[14:57] Tacitus: But that does not happen. The director goes to Bulb which seems to be wrong for me. Any explainations why this isn't wrong and where I made a mistake in my thinking?

Log:
[23:14] UG forms
[23:14] UG receives 840G
[23:14] UG1 merges into UG
[23:14] Tacitus receives 2 shares of UG
[23:14] Tacitus becomes the president of UG
[23:14] UG receives 105G
[23:14] UG receives 4 train
[23:14] UG receives token (F17)
[23:14] UG2 merges into UG
[23:14] Bulb receives 1 share of UG
[23:14] UG receives 146G
[23:14] UG receives 3 train
[23:14] UG receives token (H15)
[23:14] Bulb becomes the president of UG

ollybh:
Link to the action before the UG formation:
https://18xx.games/game/228967?action=390

discoking7:
Same thing happened with KK in 233270:

[09:05] KK forms
[09:05] KK receives 840G
[09:05] KK1 merges into KK
[09:05] discoking7 receives 2 shares of KK
[09:05] discoking7 becomes the president of KK
[09:05] KK receives 280G
[09:05] KK receives 4 train
[09:05] KK receives token (E12)
[09:05] KK2 merges into KK
[09:05] Shaw80 receives 1 share of KK
[09:05] KK receives 280G
[09:05] KK receives token (E12)
[09:05] Shaw80 becomes the president of KK

tom-blight:
@perwestling i'm afraid we've had the same issue in this game: https://18xx.games/game/238315 directorship has gone to wrong player.

perwestling:
@tom-blight Yes, the bug has not been fixed in the production code yet, but there is a fix on its way.

Issue #12063, opened 2025-09-24
ID: 223899 1824


RandallNlsn:
Go to 16:19 on timeline. 

Upon formation, it appears the kk President certificate didn't change hands to the shareholder with the most shares. 

Caveman67 had 60 percent after formation,, Wyobadger 30 percent. However, it appears Wyobadger retained the double-share President certificate, which should have gone to Caveman67. 

Thank you

**What's your Game ID?**

**How to Reproduce the Bug?**

**Which Browser / Platform?**

perwestling:
To me it looks correct.

https://18xx.games/game/223899?action=453#entities

Caveman67 does have 6 shares and is president, after the KK formation.

Log says:
[23:19] KK forms
[23:19] KK receives 840G
[23:19] KK1 merges into KK
[23:19] WyoBadger receives 2 shares of KK
[23:19] caveman67 becomes the president of KK
[23:19] KK receives 216G
[23:19] KK receives 4 train
[23:19] KK receives token (E12)
[23:19] KK2 merges into KK
[23:19] caveman67 receives 1 share of KK
[23:19] KK receives 455G
[23:19] KK receives token (E12)
[23:19] Both KK tokens are in different cities in E12, so caveman67 to select one to return to charter

But I think there might be a slight confusion if KK1 merge causes one player to become president, and then KK2 merge changes this. Anyway, the result seems to be OK.

(Timestamp might probably be different because of time zones.)

RandallNlsn:
So, to be clear, it wasn't the share count that was off... It was the
certificate count.

Caveman67 had five 10% shares of kk prior to kk formation (five
certificates)

Wyobadger had one 10% share of kk prior to formation (one certificate).

Upon kk formation, Wyobadger exchanges kk1 Minor for the 20% kk Major
president share 20% (on one certificate). Wyobadger now has 30% of kk Major
on two certs (20+10).

Caveman67 exchanges kk2 Minor for a 10% share, giving him 60% total (all on
six certificates).

I believe Wyobadger would have to, by rule, give up the 20% president
certificate to Caveman67, in exchange for two 10% certificates.

And I don't believe that took place because Wyobadger finished with only
15/16 certificates despite running two companies... And Caveman67 ended up
with 16/16 certificates, while running two companies.

Hope this makes sense....
Randall

perwestling:
You are correct - certificates is off by one. I investigate it and see if I can fix the bug.

Thanks for reporting this issue.

[...]

PR #12331, opened 2026-01-02
[1824] Correct president share assignment and ownership of national

### LINKED ISSUE #12226: [1824] Incorrect cert counts
Game :https://18xx.games/game/228694

The four players all have 16 certs, and two presidencies. We should all have 18 shares. But I have 17 shares and player TheFlow has 19 shares. 

Somewhere along the way the cert count got derailed! 

I'm not sure how I can check where exactly it went wrong.

Thanks!

perwestling:
Fixes #12226, #12063, #12243

## Before clicking "Create"

This does affect at least some games, even completed ones. Do pin / archive as appropriate.

This PR is derived from #12325 which has the fixtures and unit tests, and is therefor in Draft.
So when #12325 is merged I will rebase and publish this for review.

## Implementation Notes
Have revisited merge method to make it more robust.

Changed assignment of president share. This caused an issue during formation as president share was given to the owner of primary pre-staatsbahn, and then transfer was done, which might mess up ownership.

Have created 4 unit tests for formation; one for each national, and then one where presidency goes to a player that do not own primary pre-staatsbahn.

### Explanation of Change
Some differences in how minors are handled in 1824 and 1837 requires the code to be split. 

### Screenshots
N/A

### Any Assumptions / Hacks
N/A

### DIFF
--- a/lib/engine/game/g_1824/game.rb
+++ b/lib/engine/game/g_1824/game.rb
@@ -215,15 +215,13 @@ module Engine
             return
           end
 
-          @players.each do |p|
-            tie_breaker << p unless tie_breaker.include?(p)
-          end
+          tie_breaker |= @players
 
           president_factors = president_candidates.to_h do |player, percent|
-            [[percent, tie_breaker.index(player) || -1, player == current_president ? 1 : 0], player]
+            [[percent, tie_breaker.index(player), player == current_president ? 1 : 0], player]
           end
           president = president_factors[president_factors.keys.max]
-          return unless current_president != president
+          return if current_president == president
 
           @log << "#{president.name} becomes the president of #{national.name}"
           national.owner = president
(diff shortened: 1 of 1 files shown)


perwestling:
Have now rebased this to latest version of master. Ready for review.

lib/engine/game/g_1824/game.rb
ollybh: This isn't part of the current PR, but I noticed this block: these three lines can be replaced with a single statement: `tie_breaker |= @players`.

ollybh: The `|| -1` isn't needed, is it? You've just added any missing players to `tie_breaker`.

lib/engine/game/g_1824/game.rb
```diff
@@ -226,8 +226,10 @@ def set_national_president!(national, tie_breaker)
           return unless current_president != president
```
ollybh: I'm surprised Rubocop doesn't object to the double negative:
```suggestion
          return if current_president == president
```

lib/engine/game/g_1824/game.rb
```diff
@@ -617,14 +619,12 @@ def merge_minor!(minor, corporation, allow_president_change: true)
...
           @log << "#{owner.name} receives #{num_shares} share#{num_shares > 1 ? 's' : ''} of #{corporation.name}"
           share.buyable = true
-
-          # 1824 fix. We explicitly set allow_president_change to true here as we otherwise get a strange
-          # behavior when presidency decided for nationals. Might need revisiting.
-          @share_pool.transfer_shares(share.to_bundle, owner, allow_president_change: true)
-
+          @share_pool.transfer_shares(ShareBundle.new([share]), owner, allow_president_change: allow_president_change)
+          minor_share = owner.shares_by_corporation[minor].first
+          @share_pool.transfer_shares(ShareBundle.new([minor_share]), minor, allow_president_change: false)
```
ollybh: Is this needed? You're about to close the minor and they can't be restarted.

PR #12333, opened 2026-01-02
[18Norway] clarifies route fees in Select Routes step

### LINKED ISSUE #12291: [18Norway] The Mjøsa hex should be free from Phase 5 onwards
**What's your Game ID?**
https://18xx.games/game/229012?action=926

**How to Reproduce the Bug?**
At the link above, Ø has only 9kr and cannot run over the Mjøsa hex. But it should be able to, since it's after phase 5 and it should now be free to run there.

philcampeau:
Fixes #12291

## Before clicking "Create"

### Explanation of Change

User reported an error with fees. There wasn't one, he was being charged a fee for a different thing. But this raised the point that there are two different route fees in the game, so I modified the code to be able to display each of the fees separately in the breakdown, while still charging and referencing the combined fees where appropriate.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_18_norway/game.rb
+++ b/lib/engine/game/g_18_norway/game.rb
@@ -274,16 +274,11 @@ module Engine
         end
 
         def mjosa_fee(route)
-          cost = 0
-          mult = 2
-          mult = 1 if @phase.tiles.include?(:green)
-          mult = 0 if @phase.tiles.include?(:brown)
-          cost += 5 * mult if !owns_hvite_svan?(route.train.owner) && route.all_hexes.include?(mjosa)
+          return 0 if owns_hvite_svan?(route.train.owner)
+          return 0 if @phase.tiles.include?(:brown)
+          return 0 unless route.all_hexes.include?(mjosa)
 
-          # P2 Thunes mekaniske verksted do not need to pay maintainance
-          return cost if owns_thunes_mekaniske?(route.train.owner)
-
-          cost
+          @phase.tiles.include?(:green) ? 5 : 10
         end
 
         def mountain_fee(route)


lib/engine/game/g_18_norway/game.rb
```diff
@@ -283,7 +283,15 @@ def route_cost(route)
           # P2 Thunes mekaniske verksted do not need to pay maintainance
           return cost if owns_thunes_mekaniske?(route.train.owner)
 
-          cost + (route.all_hexes.count { |hex| mountain?(hex) } * 10)
+          cost
```
perwestling: This looks a bit strange after code change. It returns cost regardless if "Thunes mekaniska verskstad" is owned or not.

lib/engine/game/g_18_norway/game.rb
```diff
@@ -283,7 +283,15 @@ def route_cost(route)
...
+        end
```
ollybh, line 282: This could be simplified a lot. Something like:
```
def mjosa_fee(route)
  return 0 if owns_hvite_svan(route.train.owner)
  return 0 if @phase.tiles.include?(:brown)
  return 0 unless route.all_hexes.include?(mjosa)

  @phase.tiles.include?(:green) ? 5 : 10
end

File: lib/engine/game/g_18_norway/game.rb [hunk 3]
```diff
@@ -292,15 +300,15 @@ def check_other(route)
...
           raise GameError, 'Cannot afford the fees for this route' if route.train.owner.cash < cost
         end
 
         def revenue_str(route)
           stop_hexes = route.stops.map(&:hex)
           str = route.hexes.map { |h| stop_hexes.include?(h) ? h&.name : "(#{h&.name})" }.join('-')
-          cost = route_cost(route)
-          str += " -Fee(#{cost})" if cost.positive?
+          str += " + Mjøsa fee (#{format_currency(mjosa_fee(route))})" if mjosa_fee(route).positive?
+          str += " + Mountain fee (#{format_currency(mountain_fee(route))})" if mountain_fee(route).positive?
```
ollybh, line 306: Does this render OK? Or is the text too long for the route selector?

ollybh (changes requested):
The `mjosa_fee` method needs to be tidied up. And can you add a screenshot showing how the proposed change looks?

lib/engine/game/g_18_norway/game.rb
philcampeau: good point. I was clearly working on autopilot when I edited that!

lib/engine/game/g_18_norway/game.rb
philcampeau: I added a screenshot to my original post above. To me it looks fine, let me know if you disagree.

perwestling (approved):

PR #12334, opened 2026-01-04
[18USA] Do not allow multi-resource type abilities to use both resource types at once

crericha:
The helper method wasn't properly deduping ability usage due to an opal bug around array.shift. This introduces a workaround to fix that and simplifies the code a bit.

### DIFF
--- a/lib/engine/game/g_18_usa/game.rb
+++ b/lib/engine/game/g_18_usa/game.rb
@@ -439,15 +439,8 @@ module Engine
           # Filter out duplicates
           resource_ability_values = resource_abilities.values
           dups = resource_ability_values[0].intersection(resource_ability_values[1])
-          dup_pos = 0
-          resource_abilities.transform_values do |abilities|
-            unless (ability = (abilities - dups)&.first)
-              # Workaround for opal bug with array.shift
-              ability = dups[dup_pos]
-              dup_pos += 1
-            end
-            ability
-          end
+          resource_abilities.transform_values! { |v| (v - dups)&.first }
+          resource_abilities.transform_values { |v| v || dups.shift }
         end
 
         def consume_abilities_to_lay_resource_tile(hex, tile, selected_companies)


lib/engine/game/g_18_usa/game.rb
```diff
@@ -428,18 +428,26 @@ def resource_abilities_for_hex(hex, resource, selected_companies)
...
           # Filter out duplicates
-          dups = resources.values[0].intersection(resources.values[1])
-          resources.transform_values! { |abilities| (abilities - dups)&.first || dups.shift }
-          resources
+          resource_ability_values = resource_abilities.values
+          dups = resource_ability_values[0].intersection(resource_ability_values[1])
+          dup_pos = 0
+          resource_abilities.transform_values do |abilities|
+            unless (ability = (abilities - dups)&.first)
+              # Workaround for opal bug with array.shift
```
ollybh: I don't think this is an Opal bug.

If you've got a single company with an ability `a1` that provides both resources, and no other companies with resource abilities then you have `resources = { coal: [a1], oil: [a1] }` and `dups = [a1]`. 

On the first loop around `transform_values!` (in the original code) `abilities - dups` is an empty array, so `dups.shift` returns `a1` and clears `dups`. In the second loop `abilities - dups` is giving `[a1]`. So the duplication is still there.

lib/engine/game/g_18_usa/game.rb
```diff
@@ -428,18 +428,26 @@ def resource_abilities_for_hex(hex, resource, selected_companies)
...
+              ability = dups[dup_pos]
+              dup_pos += 1
+            end
+            ability
```
ollybh: This updated algorithm looks like it would work.

If you want a short, but slightly evil, alternative, this might also give the same result:
```resource_abilities.transform_values(&:first).invert.invert```
The first invert call returns an array with abilities as keys, so will remove the duplicates.

File: lib/engine/game/g_18_usa/game.rb (hunk 1 above)
Comment (crericha) [reply lib/engine/game/g_18_usa/game.rb]: Ahh, right! Great catch. I updated the original algorithm to do it in two steps.

File: lib/engine/game/g_18_usa/game.rb (hunk 2 above)
Comment (crericha) [reply lib/engine/game/g_18_usa/game.rb]: I love my one-liners, but I worry I won't remember what's going on here, so going to do it in two steps instead.

**Comment by ollybh** [comment]
@crericha This PR has broken a few games in the database. Here's the result of the validation script run against a month-old copy of the database:

```
[195763,"18USA","finished",[],"#<Engine::GameError: B10 is not upgradeable to 9coalore10>"]
[197669,"18USA","finished",[],"#<NoMethodError: undefined method `consume_tile_lay' for nil:NilClass>"]
[199588,"18USA","finished",[],"#<NoMethodError: undefined method `consume_tile_lay' for nil:NilClass>"]
[199665,"18USA","finished",[],"#<NoMethodError: undefined method `consume_tile_lay' for nil:NilClass>"]
[226657,"18USA","finished",[],"#<Engine::GameError: B12 is not upgradeable to 9coaloil>"]
[227212,"18USA","finished",[],"#<Engine::GameError: B12 is not upgradeable to 9coaloil>"]
[227480,"18USA","finished",[],"#<Engine::GameError: B12 is not upgradeable to 9coaloil>"]
[227709,"18USA","finished",[],"#<Engine::GameError: B12 is not upgradeable to 9coaloil>"]
[228704,"18USA","finished",[],"#<Engine::GameError: B12 is not upgradeable to 9coaloil>"]
[230028,"18USA","finished",[],"#<Engine::GameError: B10 is not upgradeable to 9coalore10>"]
```

Is this a problem? Or do we just pin these games when we next deploy?

PR #12340, opened 2026-01-06
[site] Improves game selection page

### LINKED ISSUE #12310: 18 Christmas Eve Player Counter
When setting up the game the player numbers can exceed the maximum and minimum values

philcampeau:
Fixes #12310

## Before clicking "Create"

### Explanation of Change

These changes will do a few things. 

- restores the upper and lower limit hard stop
- dynamically changes the upper and lower limits as different options are selected
- this allows all checkboxes to be enabled, rather than greying them out and forcing the user to adjust player count first. 

As usual, please let me know if there are any changes that need to be done!


assets/app/view/create_game.rb
```diff
@@ -711,5 +730,37 @@ def update_player_range(meta)
...
     end
+
+    def allowed_players_for(meta)
+      title = meta.title
+      base = (@min_p[title]..@max_p[title]).to_a
+
+      rules = meta::OPTIONAL_RULES
+      selected = rules.select { |r| @optional_rules.include?(r[:sym]) && r[:players] }
+
+      selected.reduce(base) { |acc, r| acc & r[:players] }
```
ollybh, line 754: Is this finding the intersection of ranges? According to the docs, [Range](https://ruby-doc.org/3.4.1/Range.html) does not have an `&` method, and I get an error when I try this in `irb`.

ollybh:
@philcampeau Can you add some screenshots ­– or, even better, a screencast – showing how this works?

assets/app/view/create_game.rb
philcampeau: Range doesn't have an & method, but Array does. That's why I have  `base = (@min_p[title]..@max_p[title]).to_a` in the code just above what you quoted here. `r[:players]` is already an array.

I could add in a comment saying something like `# base and r[:players] are Arrays, the & method is acting on those Arrays`. or something to that effect.

philcampeau:
Here's a video of the changes in action: 

https://github.com/user-attachments/assets/cbdf0e94-75e3-42dc-b34b-5a417c24869f

While I was trying to make the video, I noticed the create game page for 1822 didn't load properly, because of the game_variants constant it uses. I managed to fix that, and I'm submitting a commit for that now. All other games appear to be loading correctly.

philcampeau:
And I'm going to make one more tweak now. The way I initially had it coded was that it would only reset the player count range if the current count was outside of the range of that option. So for example, if you had your range set as 3-4 and you chose an option with 3-5, it would stay the same, but if you had set it to 2-4 and the new option had a range of 3-4, it would reset. I think it would be more consistent to always reset the range if an option with a player count range is selected.

assets/app/view/create_game.rb
ollybh: Sorry, my memory was faulty. I would have sworn that these were ranges, not arrays…

ollybh:
This is the type of change that gives a great opportunity for bikeshedding. Here's my suggested colours…

My preference would be for not resetting the player counts when the variant selection is changed. I can see someone setting up the desired player count first and then choosing variants as that is further down the page, and finding that they have to redo the player count – or not noticing it has changed and creating a game with a different player range than they had intended.

I like how the updating of the min/max player counts are shown. But I say it's more user-friendly to allow illegal values or combinations to be entered whilst the user is selecting options, as long as they are not able to create a game with these (both by having the submit button disabled and by backend validation) with some kind of visual highlighting of problematic selections.

ollybh (changes requested):
I've just tested this on mobile browsers (Firefox and Chrome on Android and Safari on iOS) and this breaks selecting player ranges on those. Once an optional rule has been chosen I can't change either the minimum or maximum number of players.

philcampeau:
hmmmmm I'll have a look

philcampeau:
I submitted what should be a fix for that mobile issue.

ollybh:
I've tried the updated code on both desktop and mobile browsers. I tried starting a 1846 game, if I select the first edition privates variant I am unable to change either the minimum or maximum number of players.

ollybh:
@philcampeau Have you looked at this any more? It was still broken for me after the latest changes.

philcampeau:
Busy week at work. I'll have a look this week.

ollybh:
@philcampeau Sorry, I haven't yet had a chance to look at the latest changes.

Have you tested these on different systems and browsers?

philcampeau:
Tested on my computer in Edge, Firefox, and Chrome
Tested on my Samsung phone in Firefox, DuckDuckGo, and Chrome. 

All appeared to work fine. I tested with multiple games including checking the 1846 issue, and various different 1825 modules.

ollybh:
I've finally got round to testing this again (sorry for the delay, I had missed your last update). The good news is that it appears to be working now, I didn't run into any issues with the interface getting stuck this time.

However, I'm still to be convinced that this is the right approach, see my [earlier comment](https://github.com/tobymao/18xx/pull/12340#issuecomment-3720806602).

I'd want some consensus whether this is an improvement before this is merged.

philcampeau:
I updated the code a while back so that it only resets the values if the selected player count is outside the range of the newly-selected variant, so that should address your concern, Olly.

philcampeau:
@ollybh Should I open a discussion on the slack about this, see what folks think would be best?

philcampeau:
Well, the discussion didn't really go anywhere: https://18xxgames.slack.com/archives/C012K0CNY5C/p1778815641807259

I'm not sure where to go from here..

philcampeau:
@ollybh was there anything else you wanted me to change here?

PR #12342, opened 2026-01-06
[Profile page] implements Change Password function

### LINKED ISSUE #2586: Change password
Currently there is no option to change your password from the profile page. It would be nice to be able to do it

philcampeau:
Fixes #2586

## Before clicking "Create"

### Explanation of Change

Allows users to change their password from their profile page without having to use the Forgot My Password function. 

### Screenshots

The feature in the profile page: 
image

Error if the passwords don't match: 
image

Error if they enter the incorrect current password:
image

Confirmation when it is successful: 
image

### DIFF
--- a/assets/app/user_manager.rb
+++ b/assets/app/user_manager.rb
@@ -20,7 +20,8 @@ module UserManager
 
   def edit_user(params)
     @connection.safe_post('/user/edit', params) do |data|
-      store(:user, data, skip: false)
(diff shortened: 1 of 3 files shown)


michaeljb (approved):

michaeljb (approved):
Tested locally and it all looks good, thanks! Kind of wild for this to be a new feature for a [site just turning six years old today](https://boardgamegeek.com/thread/2415089/article/34645881#34645881). 😂

PR #12345, opened 2026-01-07
[core] adds BidToBuyAuction step into lib/engine/step

philcampeau:
## Before clicking "Create"

### Explanation of Change

I used the Core tag, but _currently_ this would only touch 1860.

Currently, 1860 has its own BuyCert step. This step could be useful for other games (18Scan comes to mind). This pull is the first step in centralizing the code.

I created a new step in lib/engine/step with a more descriptive name (BidToBuyAuction). It copies most of the code from 1860's BuyCert step, with a couple of modifications:

- Defaults to a minimum bid of zero
- Modifies how minimum bid raise is calculated, to accommodate an opening bid of zero
- simplifies the setup (1860 will still use the setup it used before when I submit the modification for 1860)

I also added `return nil if in_auction?` to the beginning of the `def get_par_prices(entity, corp)` method, in order to prevent par prices from displaying if you click on a corp before winning an auction.

Once this is merged in, I'll submit the pull request to modify 1860. If you'd prefer, I can add a commit to this pull request to do the 1860 modifications as well.

### DIFF
--- a/lib/engine/step/bid_to_buy_auction.rb
+++ b/lib/engine/step/bid_to_buy_auction.rb
@@ -1,0 +1,278 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+require_relative '../action/par'
+
+module Engine
+  module Step
+    class BidToBuyAuction < Engine::Step::Base
+      attr_reader :companies
+
+      AUCTION_ACTIONS = %w[bid pass].freeze
+      BUY_ACTION = %w[bid par].freeze
+      PASS_ACTION = %w[pass].freeze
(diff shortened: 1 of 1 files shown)


ollybh:
This is a reasonable suggestion, but you seem to be leaving a lot of 1860 specific code in this shared file. Have you looked at what will be done to make this usable for 18Scan?

A couple of things I'd like:

- Some documentation explaining this type of auction.
- More methods moved to the `private` section of the class.

philcampeau:
I missed this comment. I'll take a look this week.

PR #12353, opened 2026-01-10
[18Cuba] Adding narrow gauge trains

chrstas:
Picking up the 18Cuba implementation. Adding narrow gauge trains.

## Before clicking "Create"

## Implementation Notes

Add train definitions and play count restrictions.


lib/engine/game/g_18_cuba/game.rb
```diff
@@ -42,11 +42,11 @@ class Game < Game::Base
         ].freeze
 
         TRAIN_FOR_PLAYER_COUNT = {
-          2 => { '2': 5, '3': 4, '4': 2, '5': 3, '6': 3, '8': 4 },
-          3 => { '2': 7, '3': 5, '4': 3, '5': 3, '6': 3, '8': 6 },
-          4 => { '2': 9, '3': 7, '4': 4, '5': 3, '6': 3, '8': 8 },
-          5 => { '2': 10, '3': 8, '4': 5, '5': 3, '6': 3, '8': 10 },
-          6 => { '2': 10, '3': 9, '4': 5, '5': 3, '6': 3, '8': 12 },
+          2 => { '2': 5, '3': 4, '4': 2, '5': 3, '6': 3, '8': 4, '2n': 7, '3n': 5, '4n': 5, '5n': 4 },
```
ollybh: This doesn't match the table in my copy of the rules.
```suggestion
          2 => { '2': 5, '3': 4, '4': 2, '5': 3, '6': 3, '8': 4, '2n': 7, '3n': 5, '4n': 4, '5n': 5 },
```

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -86,7 +86,14 @@ class Game < Game::Base
...
                   }].freeze
 
-        TRAINS = [{ name: '2', distance: 2, price: 100, rusts_on: '4' },
+        TRAINS = [
+                  # Regular Trains
+                  {
+                    name: '2',
+                    distance: 2,
+                    price: 100,
+                    rusts_on: '4',
```
ollybh, line 96: This isn't strictly necessary as `:broad` is the default, but I'd suggest explicitly setting the train gauge, as you have both broad-gauge and narrow-gauge trains in this game
```suggestion
                    price: 100,
                    gauge: :broad,
                    rusts_on: '4',
```

chrstas:
@ollybh I think it would be then standard gauge (not broad). I have made the requested adjustment... thanks for your work

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -141,14 +157,43 @@ class Game < Game::Base
...
                       },
                     ],
-                  }].freeze
+                  },
+                  # Narrow Gauge Trains
+                  {
+                    name: '2n',
+                    distance: 2,
+                    price: 80,
+                    gauge: :narrow,
```
ollybh:
```suggestion
                    track_type: :narrow,
```

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -86,16 +86,26 @@ class Game < Game::Base
...
                   }].freeze
 
-        TRAINS = [{ name: '2', distance: 2, price: 100, rusts_on: '4' },
+        TRAINS = [
+                  # Regular Trains
+                  {
+                    name: '2',
+                    distance: 2,
+                    price: 100,
+                    gauge: :standard,
```
ollybh:
```suggestion
                    track_type: :broad,
```

ollybh:
> @ollybh I think it would be then standard gauge (not broad). I have made the requested adjustment... thanks for your work

No, the types of track gauge we have are `:broad`, `:narrow` and `:dual`. This doesn't always match the terminology used in the game rules.

I should have spotted in my initial review that you had used the wrong attribute for the train's gauge, it is `track_type` not `gauge`.

chrstas:
my fault - sorry. next time I will look deeper... thanks a lot

.bundle/config
ollybh, line 1: Why are you deleting this file? If it does need to be removed, can this be done in a separate pull request as this has nothing to do with the 18Cuba changes.

PR #12363, opened 2026-01-22
[18Cuba] Sugar cubes production logic

chrstas:
According to the rules adding now the sugar cubes production logic. Depending on the total revenue of the minor corporation sugar cubes needs to be generated and stored.

At the end of the OR the sugar cubes are emptied. The FC handling is not yet part of this PR but will be added afterwards.

## Before clicking "Create"

## Implementation Notes

I have choosen the process_dividend to trigger sugar cube production because I need the total revenue not only a single route.

### DIFF
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -244,23 +244,18 @@ module Engine
         end
 
         def sugar_production(corporation, total_revenue)
-          return if total_revenue.zero? || !corporation.minor?
+          return if total_revenue.zero? || corporation.type != :minor
 
           sugar_cubes = case total_revenue
-                        when 0..29
-                          0
-                        when 30..79
-                          1
-                        when 80..150
-                          2
-                        else
-                          3
+                        when 0..29 then 0
+                        when 30..79 then 1
+                        when 80..150 then 2
+                        else 3
                         end
 
           @sugar_cubes[corporation] += sugar_cubes
           @log << "#{corporation.name} produces #{sugar_cubes} sugar cube(s) "\
-                  "from #{format_currency(total_revenue)} revenue "\
-                  "(total sugar cubes: #{@sugar_cubes[corporation]})."
+                  "from #{format_currency(total_revenue)} revenue."
         end
 
         def transition_to_next_round!
@@ -277,10 +272,10 @@ module Engine
 
         def or_round_finished
           # For the moment reset sugar cubes, handling for FC to be implemented later
-          return if @sugar_cubes.empty?
+          return if @sugar_cubes.values.none?(&:positive?)
 
           @sugar_cubes.clear
-          @log << 'All sugar cubes are removed at the end of the Operating Round.'
+          @log << 'All remainingsugar cubes are removed at the end of the Operating Round.'
         end
       end
     end


lib/engine/game/g_18_cuba/game.rb
```diff
@@ -236,11 +236,52 @@ def setup
...
           @unused_tiles = []
+          @sugar_cubes = Hash.new(0)
         end
 
         def init_tile_groups
           self.class::TILE_GROUPS
         end
+
+        def sugar_production(corporation, total_revenue)
+          return if total_revenue.zero? || !corporation.minor?
```
ollybh: This won't work.

Minors can be a little bit confusing: `entity.minor?` will only be true for objects of type `Minor`, which are usually defined in `MINORS`. These are companies that own trains and have a treasury but do not have a share price – the minors in 1846 work like this.

In most games with minors these are companies that own trains and have a treasury but also have a share price. 1861 and 1822 minors are examples of these. These are represented by Corporation objects, and `entity.minor?` will be false. They are generally (including here) created with a type attribute to easily identify them.

```suggestion
          return if total_revenue.zero? || corporation.type != :minor
```

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -236,11 +236,52 @@ def setup
...
+                        when 80..150
+                          2
+                        else
+                          3
+                        end
+
+          @sugar_cubes[corporation] += sugar_cubes
+          @log << "#{corporation.name} produces #{sugar_cubes} sugar cube(s) "\
+                  "from #{format_currency(total_revenue)} revenue "\
+                  "(total sugar cubes: #{@sugar_cubes[corporation]})."
```
ollybh: I've not played 18Cuba, but from flicking through the rules it looks to me as though a minor company cannot start its operating turn with any sugar cubes. Or have I understood that wrong? Any cubes that are earned during a round will either be collected by a major company or taken by the Ferrocarril Central, so the minor company will start its next operating turn with zero cubes.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -236,11 +236,52 @@ def setup
...
+          sugar_cubes = case total_revenue
+                        when 0..29
+                          0
+                        when 30..79
+                          1
+                        when 80..150
+                          2
+                        else
+                          3
+                        end
```
ollybh, line 254: Feel free to ignore this suggestion, but for simple case statements like this I prefer the condensed syntax, I find it easier to read with less white space.
```suggestion
          sugar_cubes = case total_revenue
                        when 0..29 then 0
                        when 30..79 then 1
                        when 80..150 then 2
                        else 3
                        end
```

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -236,11 +236,52 @@ def setup
...
+        def finish_round!
+          case @round
+          when Round::Operating
+            or_round_finished
+          end
+        end
+
+        def or_round_finished
+          # For the moment reset sugar cubes, handling for FC to be implemented later
+          return if @sugar_cubes.empty?
```
ollybh: I'm going to quibble over temporary code, so this suggestion can also be ignored.

The hash will only be empty if no minor companies have operated in the operating round. You could have a situation where there are minor companies which have operated but there are no sugar cubes left, either because none were earned or they were all taken by major companies.

To avoid printing the cube removal message in this case:
```suggestion
          return if @sugar_cubes.values.none?(&:positive?)
```

chrstas:
closing it to create a new clean PR (https://github.com/tobymao/18xx/pull/12364), 
the rebase let into commits which are not part of the PR. Assume sequence of my rebase was the problem and it does not look clean.

PR #12371, opened 2026-01-30
[18Cuba] Auction and Draft mechanism for commssions and concessions

chrstas:
Auction mechanism for commissions and simple draft for concession. 

Round Logic: Init -> Auction commissions -> Draft concessions -> SR -> OR

Exchanging concession for minor will follow in subsequent PR. Not yet considered in this PR.

## Before clicking "Create"

## Implementation Notes

Prefix correction: Commission = 'C', Concession 'M'

Based on work from @bentziaxl - thanks

### DIFF
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -27,9 +27,6 @@ module Engine
         TRACK_RESTRICTION = :permissive
         CURRENCY_FORMAT_STR = '$%s'
 
-        COMPANY_CONCESSION_PREFIX = 'M'
-        COMPANY_COMMISSIONER_PREFIX = 'C'
-
         BANK_CASH = 10_000
 
         CERT_LIMIT = { 2 => 35, 3 => 30, 4 => 20, 5 => 17, 6 => 15 }.freeze
@@ -41,6 +38,7 @@ module Engine
              151 158 172 180 188 196 204 013 222 231 240 250 260 275 290 300],
         ].freeze
 
+        # Currently variants 2p medium, 3p short setup. Further variant support to be added later.
         TRAIN_FOR_PLAYER_COUNT = {
           2 => { '2': 5, '3': 4, '4': 2, '5': 3, '6': 3, '8': 4, '2n': 7, '3n': 5, '4n': 4, '5n': 5 },
           3 => { '2': 7, '3': 5, '4': 3, '5': 3, '6': 3, '8': 6, '2n': 5, '3n': 5, '4n': 3, '5n': 4 },
@@ -228,15 +226,22 @@ module Engine
         end
 
         def company_header(company)
-          company.id[0] == self.class::COMPANY_CONCESSION_PREFIX ? 'CONCESSION' : 'COMMISSIONER'
+          case company.type
+          when :concession
+            'CONCESSION'
+          when :commissioner
+            'COMMISSIONER'
+          else
+            raise "Unknown company type: #{company.type}"
+          end
         end
 
         def commissioners
-          @commissioners ||= @companies.select { |c| c.id.start_with?(COMPANY_COMMISSIONER_PREFIX) }
+          @commissioners ||= @companies.select { |c| c.type == :commission }
         end
 
         def concessions
-          @concessions ||= @companies.select { |c| c.id.start_with?(COMPANY_CONCESSION_PREFIX) }
+          @concessions ||= @companies.select { |c| c.type == :concession }
         end
 
         def setup
--- a/lib/engine/game/g_18_cuba/step/simple_draft.rb
+++ b/lib/engine/game/g_18_cuba/step/simple_draft.rb
@@ -10,12 +10,22 @@ module Engine
           def setup
             super
             @companies = @game.concessions.sort
-            @completed_players = {}
+            @completed_players = Hash.new(0)
+          end
+
+          def actions_per_player
+            # currently only supports 2p medium and 3p short variants,  more variants in the future
+            # 2p 2 concessions max, 3-6p 1 concession max
+            case @game.players.size
+            when 2 then 2
+            else 1
+            end
           end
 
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_18_cuba/game.rb
```diff
@@ -27,8 +27,8 @@ class Game < Game::Base
         TRACK_RESTRICTION = :permissive
         CURRENCY_FORMAT_STR = '$%s'
 
-        COMPANY_CONCESSION_PREFIX = 'C'
-        COMPANY_COMMISIONER_PREFIX = 'M'
+        COMPANY_CONCESSION_PREFIX = 'M'
+        COMPANY_COMMISIONER_PREFIX = 'C'
```
ollybh: Spelling fix:
```suggestion
        COMPANY_COMMISSIONER_PREFIX = 'C'
```

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -231,6 +231,14 @@ def company_header(company)
           company.id[0] == self.class::COMPANY_CONCESSION_PREFIX ? 'CONCESSION' : 'COMMISSIONER'
         end
 
+        def commissioners
+          @commissioners ||= @companies.select { |c| c.id[0] == self.class::COMPANY_COMMISIONER_PREFIX }
```
ollybh: This will work correctly, but ruby has a `String.starts_with?` method that is, IMHO, a bit clearer:
```suggestion
          @commissioners ||= @companies.select { |c| c.id.starts_with?(COMPANY_COMMISIONER_PREFIX) }
```
You also don't need to qualify constants with the `self.class::` prefix, unless you're planning on these being overridden in subclasses.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -243,6 +251,47 @@ def init_tile_groups
...
+        end
+
+        def next_round!
+          # After Init -> Auction Commissions -> Draft Concessions -> Stock Round -> Operating Rounds
+          @round =
+            case @round
+            when Round::Draft
+              new_stock_round
+            when Round::Stock
+              close_unopened_minors if @turn == 1 && @round.round_num == 1
```
ollybh: Do you need the `@round.round_num` test? That only gets changed in operating rounds.

lib/engine/game/g_18_cuba/step/simple_draft.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G18Cuba
+      module Step
+        class SimpleDraft < Engine::Step::Base
```
ollybh: Could this be a subclass of `Engine::Step::SimpleDraft` instead of `Engine::Step::Base`? That would give you some of the methods below for free.

lib/engine/game/g_18_cuba/step/simple_draft.rb
```diff
@@ -0,0 +1,120 @@
...
+          def setup
+            @companies = @game.concessions.sort
+            @finished = false
+          end
+
+          def round_state
+            {
+              companies_pending_par: [],
+            }
+          end
```
ollybh: Is this for future use? AFAICT, there's nothing in the code being added that uses this round state.

lib/engine/game/g_18_cuba/step/selection_auction.rb
```diff
@@ -0,0 +1,45 @@
...
+            @cheapest = @companies.first
+            @auction_triggerer = current_entity
+            auction_entity(@cheapest)
+          end
+
+          def round_state
+            {
+              companies_pending_par: [],
+            }
+          end
```
ollybh: What is this for?

ollybh (changes requested):
Can you look at using `Engine::Step::SimpleDraft` as the parent class for your `SimpleDraft` step?

There are a few other minor suggestions too.

lib/engine/game/g_18_cuba/step/simple_draft.rb
chrstas: yes you are right, thanks for the hint

chrstas:
@ollybh, thanks again. I have incorporate the changes and also used the simple_draft from the engine. Make sense and reduces code. 

Possibility for player to pass draft added with subsequent commit.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -231,6 +231,14 @@ def company_header(company)
           company.id[0] == self.class::COMPANY_CONCESSION_PREFIX ? 'CONCESSION' : 'COMMISSIONER'
         end
 
+        def commissioners
+          @commissioners ||= @companies.select { |c| c.id.start_with?(COMPANY_COMMISSIONER_PREFIX) }
+        end
+
+        def concessions
+          @concessions ||= @companies.select { |c| c.id.start_with?(COMPANY_CONCESSION_PREFIX) }
+        end
```
ollybh, line 245: An alternative way to do this would be to add a 'type' attribute to the company definitions and test that here.

chrstas:
> Thanks for making these changes, this looks much better.
> 
> The concession draft is going to end after all players have either bought a concession or passed. This is correct when there are four or more players, but if there are three players then each person can take up to two concessions. And in the two-player game each player can have three concessions.
> 
> You could replace `@completed_players` with a hash that counts how many actions each player has taken, incrementing it after each bid or pass action, and finish the step when each player has taken 1/2/3 actions, depending on the player count.

@ollybh  you are absolutely right. I have completely overlooked this part of the rules and we always only took 1 concession.  I have made the changes. thanks again.

There are multiple 2p/3p variants. I have currently only 2p medium, 3p short in. Will extend it later for other variants if the other stuff is running.

chrstas:
will change the variable name with the next PR... thanks

PR #12372, opened 2026-01-30
[1880 Romania] Pre-alpha initial commit

philcampeau:
## Before clicking "Create"

### Explanation of Change

This is the initial commit for 1880 Romania. I have permission from Lonny to implement (see screenshot below)

I also took the liberty of adding several tiles used in the game to lib/engine/config/tile.rb, as they're also used in several other games of Lonny's, and are coded individually into each of those games. Once this commit gets merged in, I'll submit new pull requests to update the other games' code. 

This game is still in pre-alpha. I still need to implement the private company abilities, as well as setting up the border restriction in the game.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1880_romania/game.rb
+++ b/lib/engine/game/g_1880_romania/game.rb
@@ -4,14 +4,18 @@ require_relative 'meta'
 require_relative '../g_1880/game'
 require_relative 'map'
 require_relative 'entities'
+require_relative 'phases'
+require_relative 'trains'
 
 module Engine
   module Game
     module G1880Romania
       class Game < G1880::Game
         include_meta(G1880Romania::Meta)
-        include Map
-        include Entities
+        include G1880Romania::Map
+        include G1880Romania::Entities
+        include G1880Romania::Phases
+        include G1880Romania::Trains
 
         CURRENCY_FORMAT_STR = 'L%s'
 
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_1880_romania/entities.rb
```diff
@@ -0,0 +1,303 @@
...
+    module G1880Romania
+      module Entities
+        COMPANIES = [
+          {
+            name: 'Banater Montanbahn',
+            sym: 'P1',
+            value: 20,
+            revenue: 5,
+            desc: 'May be closed any time during the game to give the owner 20 (if player) or 40 (if company) out of the bank.',
+            color: nil,
```
crericha: I don't think you need to include values you aren't using. They'll default to the proper value.

lib/engine/game/g_1880_romania/meta.rb
```diff
@@ -0,0 +1,25 @@
...
+        include Game::Meta
+
+        DEV_STAGE = :prealpha
+        DEPENDS_ON = '1880'
+
+        GAME_DESIGNER = 'Lonny Orgler'
+        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/1880-Romania' # todo
+        GAME_LOCATION = 'Romania'
+        GAME_PUBLISHER = :lonny_games
+        GAME_RULES_URL = # todo
```
crericha: this won't compile correctly afaik. Set to an empty string.

lib/engine/game/g_1880_romania/meta.rb
```diff
@@ -0,0 +1,25 @@
...
+          GAME_TITLE = '1880 Romania'
```
crericha: wrong indent

crericha (approved):

perwestling:
Have you sent the permission to Toby? I do not think this PR need to mention it.

lib/engine/game/g_1880_romania/entities.rb
```diff
@@ -0,0 +1,303 @@
...
+          },
+        ].freeze
+
+        CORPORATIONS = [
+          {
+            float_percent: 20,
+            sym: 'BR',
+            name: 'Banat Railways',
+            logo: '1880_romania/BR',
+            # simple_logo: '1880_romania/BR.alt',
```
perwestling: Are these intended to be added later?

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -0,0 +1,155 @@
...
+        include Map
+        include Entities
+
+        CURRENCY_FORMAT_STR = 'L%s'
+
+        CERT_LIMIT = { 3 => 20, 4 => 16, 5 => 14, 6 => 12 }.freeze
+
+        STARTING_CASH = { 3 => 600, 4 => 480, 5 => 400, 6 => 340 }.freeze
+
+        PHASES = [{ name: 'A1', train_limit: 4, tiles: [:yellow] },
```
perwestling: When doing 1824 I put many things in separate files, like phases, trains.

lib/engine/game/g_1880_romania/map.rb
```diff
@@ -0,0 +1,310 @@
...
+          '401' => 2,
+          '447' => 2,
+          '8850' => 3,
+          '8851' => 7,
+          '8852' => 7,
+          '8854' => 1,
+          '8855' => 1,
+          '8856' => 1,
+          '8857' => 1,
+          '8858a' =>
```
perwestling, line 27: Just curious why this one was put in this file, and not in the general tile.rb?

lib/engine/game/g_1880_romania/meta.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Game
+    module G1880Romania
+      module Meta
+        include Game::Meta
+
+        DEV_STAGE = :prealpha
+        DEPENDS_ON = '1880'
+
+        GAME_DESIGNER = 'Lonny Orgler'
+        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/1880-Romania' # todo
```
perwestling: Todo might be removed.

lib/engine/game/g_1880_romania/meta.rb
perwestling: Maybe "TODO"?

lib/engine/game/g_1880_romania/entities.rb
philcampeau: Yeah I just haven't done it yet.

lib/engine/game/g_1880_romania/game.rb
philcampeau: That's not how most games on this site are coded, though. I prefer to do things the way other games have it done, 1880 China for example.

lib/engine/game/g_1880_romania/map.rb
philcampeau: That one was the only example of that tile on the site.

lib/engine/config/tile.rb
```diff
@@ -69,6 +71,13 @@ module Tile
         '790' => 'city=revenue:40,slots:2;path=a:0,b:_0;path=a:_0,b:3;path=a:1,b:_0;path=a:_0,b:4;label=N',
         '956' => 'city=revenue:20;path=a:0,b:_0,track:narrow;path=a:3,b:_0,track:narrow',
         '441a' => 'city=revenue:10;path=a:0,b:_0;label=B',
+        '8850' => 'town=revenue:20;path=a:0,b:_0;path=a:1,b:_0',
+        '8851' => 'town=revenue:20;path=a:0,b:_0;path=a:2,b:_0',
```
ollybh: These tiles are equivalent to the ones that are defined in 1880 but they have different orientations. This means that moving 1880 to use these tile definitions would require a migration script to change the tiles' rotation when they have been played in existing games. It would be far simpler to change these definitions to match the orientations in the 1880 code.

### 1880 tiles
Image

#### 1880 Romania tiles
Image

lib/engine/config/tile.rb
```diff
@@ -46,10 +46,12 @@ module Tile
         '201' => 'city=revenue:30;path=a:0,b:_0;path=a:1,b:_0;label=Y',
         '202' => 'city=revenue:30;path=a:0,b:_0;path=a:2,b:_0;label=Y',
         '235' => 'city=revenue:30;city=revenue:30;path=a:0,b:_0;label=OO',
+        '401' => 'city=revenue:30;path=a:0,b:_0;path=a:1,b:_0;label=T',
         # '403' => 'city=revenue:30;town=revenue:30;path=a:0,b:_0;label=B;upgrade=cost:40',
         '437' => 'town=revenue:30;path=a:0,b:_0;path=a:_0,b:2;icon=image:port,blocks_lay:1',
         '438' => 'city=revenue:40;path=a:0,b:_0;path=a:2,b:_0;label=H;upgrade=cost:80',
         '445' => 'town=revenue:20;path=a:0,b:_0;path=a:_0,b:2;icon=image:18_al/tree,blocks_lay:1',
+        '447' => 'city=revenue:30;path=a:0,b:_0;path=a:2,b:_0;label=T',
```
ollybh: This tile has a different orientation to the definition in the 1824 code.

lib/engine/game/g_1880_romania/map.rb
ollybh: It looks identical to tile 8858 in 1880. Or am I missing something?

lib/engine/game/g_1880_romania/game.rb
perwestling: I do think there are several advantages of splitting it up. E.g. it makes it clearer when you have variants in other directories.

But I guess it is a personal preference (I like collapsible regions as well...)

lib/engine/game/g_1880_romania/map.rb
philcampeau: You're right, this wasn't the only example. This one has a different issue. 

The problem here is, 1880 uses 8858 for that tile. 1840, 18Mag, 1888, and 18Mag use '8858' for a green OO. This game used 8858 for both, until I pointed out the issue to Lonny. 

FWTWR lists both tiles, but assigned the 8858a label to the green OO. I don't mind flipping this around to match FWTWR, if that makes more sense?

lib/engine/config/tile.rb
philcampeau: That's a good point, I hadn't even considered that. I was trying to orient them the same way that the (7, 8, 9), (3, 4, 58), and 5, 6, and 57 tiles are oriented. It's too bad this wasn't done originally, but yes, I'll change that now.

lib/engine/game/g_1880_romania/entities.rb
philcampeau: Done.

lib/engine/game/g_1880_romania/map.rb
philcampeau: I think, for now, I'll leave the 8858 tiles as they are in this commit. We can always come back to this later when I submit a different PR to clean up all the games that use these tiles.

lib/engine/config/tile.rb
philcampeau: fixed

lib/engine/game/g_1880_romania/meta.rb
philcampeau: fixed

lib/engine/game/g_1880_romania/meta.rb
philcampeau: fixed

lib/engine/game/g_1880_romania/entities.rb
philcampeau: removed

lib/engine/game/g_1880_romania/map.rb
ollybh: It would be best to fix any tile numbers before this game gets released to alpha testing, as changing them afterwards is going to break games.

For the reuse of tile number 8858, I'd vote for copying the designation used on FWTWR. That is the nearest thing to a official reference for tile numbers.

PR #12377, opened 2026-02-02
Add kickstarter link 1880 romania

perwestling:
Add link for Kickstarter of 1880 Romania, to first page.

### Screenshots

image


perwestling:
Not sure what happened. Should have been just one commit.

perwestling:
OK, I'll redo it.

PR #12383, opened 2026-02-08
[18Cuba] Par minors with concession exchange

chrstas:
Implements 18Cuba-specific logic for parring minor corporations via concessions. Overrides process_par to handle minors differently: sets the corporation’s par, exchanges a concession for the minor’s starting shares, applies any exchange discounts, updates player ownership, closes the concession if needed, and tracks the action. Also adjusts share-buying rules to prevent minors from selling additional shares after floating.

## Before clicking "Create"

## Implementation Notes

- Variable name changed to meaningful name in simple_draft (pending reveiwer comment from last PR)
- Entities consistently completed with types

### DIFF
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -278,7 +278,7 @@ module Engine
         def can_par?(corporation, entity)
           # FC cannot be parred
           # Minors can only be parred by players with a concession to exchange
-          return false if corporation.name == 'FC'
+          return false if corporation.type == :state
           return super unless corporation.type == :minor
 
           entity.companies.any? { |c| abilities(c, :exchange) }
--- a/lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
@@ -88,11 +88,6 @@ module Engine
             end
           end
 
-          def available_cash(entity)
-            # Extends available cash with concession discount for step action availability checks.
-            entity.cash + concession_discount(entity)
-          end
-
           def minor_starting_bundle(corporation)
             # Builds the starting share bundle required to par a minor.
             shares = corporation.ipo_shares.take(2)
@@ -120,6 +115,22 @@ module Engine
 
           def special_abilities?(company)
             company.abilities.any? { |a| a.type != :exchange }
+          end
+
+          def can_buy?(entity, bundle)
+            corporation = bundle&.corporation
+
+            # Only apply custom logic to minors
+            return super unless corporation&.type == :minor
+
+            minor_not_parred = !corporation.ipoed
+            discount = exchange_concession(entity)&.discount.to_i
+
+            # Allow par if the minor is not parred and a concession discount is available
+            return true if minor_not_parred && discount.positive?
+
+            # Fallback to default engine rules
+            super
           end
         end
       end


lib/engine/game/g_18_cuba/entities.rb
```diff
@@ -207,66 +277,73 @@ module Entities
...
-            type: 'minor',
             tokens: [0],
             shares: [40, 20, 20, 20],
             color: :lightgreen,
+            type: :minor,
+            capitalization: :incremental,
           },
           {
             float_percent: 60,
             sym: 'SB',
```
ollybh: My copy of the game uses CS for this company, not SB. It would be best to be consistent with this.

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,92 @@
...
+            exchange_price = (share_price.price * bundle.num_shares) - exchange_discount
+
+            @round.players_bought[entity][corporation] += bundle.percent
+
+            buy_shares(
+                entity,
+                bundle,
+                exchange: concession,
+                exchange_price: exchange_price,
+              )
```
ollybh, line 47: Could the player choose a par price that is too high for them to afford three shares, even with the concession discount? What happens if they do?

You might want to override `available_par_cash`, that method is called from `View::Game::Par` and is used to calculate how many shares the player can afford for each par price.

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,92 @@
...
+
+            track_action(action, corporation)
+          end
+
+          def minor_starting_bundle(corporation)
+            shares = corporation.ipo_shares.take(2)
+            ShareBundle.new(shares)
+          end
+
+          def exchange_concession!(entity)
```
ollybh: The exclamation mark on the end of this method implies that it is making changes to objects, but you're just finding something here.

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,92 @@
...
+          def close_concession_if_applicable(concession)
+            concession.close! if special_abilities?(concession)
+          end
+
+          def special_abilities?(company)
+            company.abilities.any? { |a| a.type != :exchange }
+          end
+
+          def can_buy?(entity, bundle)
+            super # &&
```
ollybh: Should this line be removed?

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,92 @@
...
+            # In G18Cuba, minor corporations may not sell additional shares to other players
+            !(bundle.owner.corporation? && bundle.corporation.type == :minor && bundle.corporation.floated?)
```
ollybh: An alternative approach might be to mark the last two shares of each minor as non-buyable, using their `buyable` attribute. I'm not sure if that's better than this.

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,92 @@
...
+          def remove_exchange_ability(concession)
+            ability = concession.abilities.find { |a| a.type == :exchange }
+            return unless ability
+
+            concession.remove_ability(ability)
+            concession.value = 0
+          end
+
+          def close_concession_if_applicable(concession)
+            concession.close! if special_abilities?(concession)
```
ollybh: Don't you want to leave the company open if it has any other abilities, and only close concession 1 after it is exchanged for a minor?

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
chrstas: You are right... have added the suggestion and also secured it with GameError

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
chrstas: I needed to review the rules again and think about it. But I have changed it now and it makes sense because UI then also shows it as reserved IPO shares. 

thanks for the hint

chrstas:
Did additional tests for cases with players with low cash but concession. To my understanding it should be covered now.

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,128 @@
...
+            available = available_par_cash(entity, corporation)
+
+            @game.stock_market.par_prices.select do |price|
+              (price.price * bundle_size) <= available
+            end
+          end
+
+          def available_cash(entity)
+            # Extends available cash with concession discount for step action availability checks.
+            entity.cash + concession_discount(entity)
```
ollybh: I think that this is going to cause problems, this method isn't only called when parring a minor company.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -264,11 +264,26 @@ def new_draft_round
...
+
         def close_unopened_minors
           @corporations.each { |c| c.close! if c.type == :minor && !c.floated? }
           @log << 'Unopened minors close'
         end
 
+        def can_par?(corporation, entity)
+          # FC cannot be parred
+          # Minors can only be parred by players with a concession to exchange
+          return false if corporation.name == 'FC'
```
ollybh: This would be more consistent with the following line.
```suggestion
          return false if corporation.type == :state
```

lib/engine/game/g_18_cuba/step/buy_sell_par_shares.rb
chrstas: Yes - I spend more thoughts on it and I have removed the available_cash overwrite. Instead I use now "can_buy" to add explicitly for unparred minors the concession discount (if available). this should have less side effects and hopefully be less invasive.

chrstas:
Adjustments made based on review. Switch to "can_buy" instead of "available_cash" to reduce sideeffects.

chrstas:
Thanks to you... very much appreciate your reviews

PR #12386, opened 2026-02-12
[1832] London Investment private implementation

JudeH1:
### Explanation of Change

- Implements missing features of London Investment Private:
- Counts as purchasing action
- Pays revenue until corp exchanged for pays first dividend

- Changed WV Coal fields private to not count as a track lay (despite being mentioned in the checklist, I didn't see this requirement anywhere in the rules, if I'm wrong let me know and I'll undo this change

I'd like to get 1832 working online, so more changes will be coming soon!

### DIFF
--- a/lib/engine/game/g_1832/entities.rb
+++ b/lib/engine/game/g_1832/entities.rb
@@ -287,10 +287,12 @@ module Engine
             abilities: [{
               type: 'exchange',
               owner_type: 'player',
-              corporations: 'any',
+              corporations: 'london',
+              count: 1,
+              remove_when_used_up: false,
               from: 'ipo',
             },
-                        { type: 'no_buy' }],
+            { type: 'no_buy' }],
             color: nil,
           },
           {
--- a/lib/engine/game/g_1832/entities.rb
+++ b/lib/engine/game/g_1832/entities.rb
@@ -287,7 +287,7 @@ module Engine
             abilities: [{
               type: 'exchange',
               owner_type: 'player',
-              corporations: 'london',
+              corporations: 'ipoed',
               when: 'owning_player_sr_turn',
               count: 1,
               remove_when_used_up: false,
--- a/lib/engine/game/g_1832/entities.rb
+++ b/lib/engine/game/g_1832/entities.rb
@@ -288,6 +288,8 @@ module Engine
               type: 'exchange',
               owner_type: 'player',
               corporations: 'london',
+              owner_type: 'player',
+              when: 'owning_player_sr_turn',
               count: 1,
               remove_when_used_up: false,
               from: 'ipo',
--- a/lib/engine/game/g_1832/entities.rb
+++ b/lib/engine/game/g_1832/entities.rb
@@ -293,7 +293,7 @@ module Engine
               remove_when_used_up: false,
               from: 'ipo',
             },
-            { type: 'no_buy' }],
+                        { type: 'no_buy' }],
             color: nil,
           },
           {
--- a/lib/engine/game/g_1832/entities.rb
+++ b/lib/engine/game/g_1832/entities.rb
@@ -288,7 +288,6 @@ module Engine
               type: 'exchange',
               owner_type: 'player',
               corporations: 'london',
-              owner_type: 'player',
               when: 'owning_player_sr_turn',
               count: 1,
               remove_when_used_up: false,
--- a/lib/engine/game/g_1832/entities.rb
+++ b/lib/engine/game/g_1832/entities.rb
@@ -290,7 +290,7 @@ module Engine
               corporations: 'ipoed',
               when: 'owning_player_sr_turn',
               count: 1,
-              remove_when_used_up: false,
+              remove_when_used_up: true,
               from: 'ipo',
             },
                         { type: 'no_buy' }],
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -246,34 +246,34 @@ module Engine
           @london_company = corporation
         end
 
-
         def exchange_corporations(exchange_ability)
-        candidates = case exchange_ability.corporations
-                     when 'any'
-                       corporations
-                     when 'ipoed'
-                       corporations.select(&:ipoed)
-                    when 'london'
-                      corporations.select { |c| (!@round.current_actions.any? { |x| x.instance_of?(Action::BuyShares) }) && !c.player_share_holders.none? && !c.operated? && exchange_ability.count > 0}
-                    else
-                       exchange_ability.corporations.map { |c| corporation_by_id(c) }
-                    end
-        candidates.reject(&:closed?)
-      end
+          candidates = case exchange_ability.corporations
+                       when 'any'
+                         corporations
+                       when 'ipoed'
+                         corporations.select(&:ipoed)
+                       when 'london'
+                         corporations.select do |c|
+                           !@round.current_actions.any? do |x|
+                             x.instance_of?(Action::BuyShares)
+                           end && !c.player_share_holders.none? && !c.operated? && exchange_ability.count > 0
+                         end
+                       else
+                         exchange_ability.corporations.map { |c| corporation_by_id(c) }
+                       end
+          candidates.reject(&:closed?)
+        end
 
+        def use!(**_kwargs)
+          @used = true
 
-      def use!(**_kwargs)
-        @used = true
+          @count_this_or += 1 if @count_per_or
 
-        @count_this_or += 1 if @count_per_or
+          return unless @count
 
-        return unless @count
-
-        @count -= 1
-        owner.remove_ability(self) if !@count.positive? && @remove_when_used_up
-      end
-
-      
+          @count -= 1
+          owner.remove_ability(self) if !@count.positive? && @remove_when_used_up
+        end
 
         def revenue_for(route, stops)
           revenue = super
@@ -370,9 +370,11 @@ module Engine
           company = action.entity
           bundle = action.bundle
 
-          return unless action.entity.name == "London Investment"
-          log << "Processing buy shares for London Investment"
-          raise GameError, "Cannot exchange #{action.entity.id} for #{bundle.corporation.id}" unless can_exchange?(company, bundle)
+          return unless action.entity.name == 'London Investment'
+
+          log << 'Processing buy shares for London Investment'
+          raise GameError, "Cannot exchange #{action.entity.id} for #{bundle.corporation.id}" unless can_exchange?(company,
+                                                                                                                   bundle)
 
           buy_shares(company.owner, bundle, exchange: company)
           @round.players_history[company.owner][bundle.corporation] << action if @round.respond_to?(:players_history)
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -8,6 +8,7 @@ require_relative 'meta'
 require_relative 'phases'
 require_relative 'trains'
 require_relative '../base'
+require_relative 'step/exchange'
 
 module Engine
   module Game
@@ -98,6 +99,7 @@ module Engine
         def stock_round
           G1870::Round::Stock.new(self, [
             Engine::Step::DiscardTrain,
+            G1832::Step::Exchange,
             G1832::Step::BuySellParShares,
             G1850::Step::PriceProtection,
           ])
@@ -106,7 +108,7 @@ module Engine
         def operating_round(round_num)
           Engine::Round::Operating.new(self, [
             Engine::Step::Bankrupt,
-            Engine::Step::Exchange,
+            G1832::Step::Exchange,
             G1832::Step::BuyCompany,
             G1870::Step::Assign,
             G1870::Step::SpecialTrack,
@@ -236,6 +238,43 @@ module Engine
           @coal_hex ||= hex_by_id('B14')
         end
 
+        def london_company
+          @london_company || nil
+        end
+
+        def assign_london_company(corporation)
+          @london_company = corporation
+        end
+
+
+        def exchange_corporations(exchange_ability)
+        candidates = case exchange_ability.corporations
+                     when 'any'
+                       corporations
+                     when 'ipoed'
+                       corporations.select(&:ipoed)
+                    when 'london'
+                      corporations.select { |c| c.floated? && !c.operated? && exchange_ability.count > 0}
+                    else
+                       exchange_ability.corporations.map { |c| corporation_by_id(c) }
+                    end
+        candidates.reject(&:closed?)
+      end
+
+
+      def use!(**_kwargs)
+        @used = true
+
+        @count_this_or += 1 if @count_per_or
+
+        return unless @count
+
+        @count -= 1
+        owner.remove_ability(self) if !@count.positive? && @remove_when_used_up
+      end
+
+      
+
         def revenue_for(route, stops)
           revenue = super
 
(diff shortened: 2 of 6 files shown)


lib/engine/game/g_1832/step/dividend.rb
```diff
@@ -0,0 +1,49 @@
...
+            entity
+          end
+
+          def share_price_change(_entity, revenue)
+            return { share_direction: :left, share_times: 1 } if revenue.zero?
+
+            return { share_direction: :right, share_times: 1 } if revenue == @game.routes_revenue(routes)
+
+            {}
+          end
```
ollybh: 1832 is an 1870/1850 derivative so, where possible, you should be inheriting from those games' classes to reduce the amount of new code needed.
```suggestion
require_relative '../../g_1870/step/dividend'

module Engine
  module Game
    module G1870
      module Step
        class Dividend < G1870::Step::Dividend
```

lib/engine/game/g_1832/step/dividend.rb
```diff
@@ -0,0 +1,49 @@
...
+            receivers = payouts
+                        .sort_by { |_r, c| -c }
+                        .map { |receiver, cash| "#{@game.format_currency(cash)} to #{receiver.name}" }.join(', ')
+
+            if entity == @game.london_company
+              @log << 'London Investment closes'
+              li = @game.companies.find { |c| c.sym == 'P4' }
+              li.close! if li
+            end
+            log_payout_shares(entity, revenue, per_share, receivers)
```
ollybh: Most of this code is cut-n-pasted from `Engine::Step::Dividend.payout_shares`. It can be replaced with a call to `super`.

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          def actions(entity)
+            return ACTIONS if can_exchange?(entity)
+
+            []
+          end
+
+          def blocks?
+            false
+          end
```
ollybh: These methods are identical to the ones in the base class, so do not need to be repeated here.
```suggestion
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/exchange'
+require_relative '../../../step/share_buying'
```
ollybh: I don't think this is needed. This class will get pulled in by Step::Exchange.
```suggestion
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,67 @@
...
+
+            shares = []
+            @game.exchange_corporations(ability).each do |corporation|
+              shares << corporation.reserved_shares.first if ability.from.include?(:reserved)
+              shares << corporation.available_share if ability.from.include?(:ipo)
+              shares << @game.share_pool.shares_by_corporation[corporation]&.first if ability.from.include?(:market)
+            end
+
+            shares.compact.any? { |s| can_gain?(entity.owner, s&.to_bundle, exchange: true) }
+          end
```
ollybh: These are also unneeded duplicates of base class methods.
```suggestion
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,67 @@
...
+            false
+          end
+
+          def process_buy_shares(action)
+            company = action.entity
+            bundle = action.bundle
+            unless (ability = @game.abilities(company, :exchange))
+              raise GameError,
+                    "Could not assign #{company.name} to #{target.name}; :exchange ability not found"
+            end
```
ollybh, line 16: This isn't needed, it's duplicating a check that's already being made in `can_exchange?`.

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,67 @@
...
+                                                                                                                     bundle)
+
+            buy_shares(company.owner, bundle, exchange: company)
+            @round.players_history[company.owner][bundle.corporation] << action if @round.respond_to?(:players_history)
+            if company.name == 'London Investment'
+              @game.assign_london_company(bundle.corporation)
+              ability.use!
+              @round.current_actions << action
+            end
+            company.close! if company.name != 'London Investment'
```
ollybh: Use an else branch, instead of repeating the previous test.
```suggestion
            else
              company.close!
            end
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,67 @@
...
+            end
+            raise GameError, "Cannot exchange #{action.entity.id} for #{bundle.corporation.id}" unless can_exchange?(company,
+                                                                                                                     bundle)
+
+            buy_shares(company.owner, bundle, exchange: company)
+            @round.players_history[company.owner][bundle.corporation] << action if @round.respond_to?(:players_history)
+            if company.name == 'London Investment'
+              @game.assign_london_company(bundle.corporation)
+              ability.use!
+              @round.current_actions << action
```
ollybh: Why is this only in the London Investment branch? Doesn't the action need to be recorded for other companies?

ollybh: Are there other companies with exchange abilities? This can be simplified by removing the test if we know this will only be called for this company.

lib/engine/game/g_1832/step/track.rb
```diff
@@ -29,7 +29,7 @@ def actions(entity)
 
           def choices
             choices = []
-            choices << ['Buy Coal Token'] if can_buy_coal_token?(current_entity)
+            choices << ['Buy Coal Token ($80)'] if can_buy_coal_token?(current_entity)
             choices
```
ollybh: `choices` can either be an array or a hash. I much, much prefer using the hash. The problem with the array form is, as you have discovered, it stores the button text as the choice in the game JSON. That means that any future changes to the button text will break existing games.

If you use a hash then the hash value is used for the button label, but the hash key is stored as the choice. So changing the button label doesn't break anything, as long as the key is unchanged.

Here you would have something like:
```
def choices
  can_buy_coal_token?(current_entity) ? { 'buy_coal_token' => 'Buy Coal Token } : {}
end
```

And then in `process_choose` you would need to change the test to `… if action.choice == 'buy_coal_token'`.

lib/engine/game/g_1832/entities.rb
```diff
@@ -287,7 +287,10 @@ module Entities
             abilities: [{
               type: 'exchange',
               owner_type: 'player',
-              corporations: 'any',
+              corporations: 'london',
```
ollybh: This isn't one of the defined options for `corporations` for an exchange ability. I think it would be better to use `ipoed` here, and then add some additional tests in `exchange_corporations` to further limit the options to corporations that have only just been opened.

lib/engine/game/g_1832/entities.rb
```diff
@@ -287,7 +287,10 @@ module Entities
...
+              when: 'owning_player_sr_turn',
+              count: 1,
+              remove_when_used_up: false,
```
ollybh: Why is this false?

lib/engine/game/g_1832/game.rb
```diff
@@ -236,6 +238,43 @@ def coal_hex
           @coal_hex ||= hex_by_id('B14')
         end
 
+        def london_company
```
ollybh: This method name confused me. In this code base 'company' refers to a private company, and 'corporation' to a major/public company. So I would expect `london_company` to be P4, but this is the public company whose share has been bought by the London investors, right?

lib/engine/game/g_1832/game.rb
```diff
@@ -236,6 +238,43 @@ def coal_hex
...
+          @london_company || nil
```
ollybh: The or operator isn't doing anything here. You're either returning the value of `@london_company` or, if it is nil, returning nil, which is also the value of `@london_company`.

lib/engine/game/g_1832/game.rb
```diff
@@ -236,6 +238,43 @@ def coal_hex
...
+        end
+
+        def assign_london_company(corporation)
+          @london_company = corporation
+        end
```
ollybh: You've written custom setter/getter methods, which are just assigning/returning a simple value. You could instead just add `attr_accessor :london_company` (ideally with a better name) to let the step object read and set this variable.

lib/engine/game/g_1832/game.rb
```diff
@@ -236,6 +238,43 @@ def coal_hex
...
+                       when 'london'
+                         corporations.select do |c|
+                           !@round.current_actions.any? do |x|
+                             x.instance_of?(Action::BuyShares)
+                           end && !c.player_share_holders.none? && !c.operated? && exchange_ability.count > 0
+                         end
+                       else
+                         exchange_ability.corporations.map { |c| corporation_by_id(c) }
+                       end
+          candidates.reject(&:closed?)
```
ollybh: If I've understood the rules correctly, they London Investment private can be used to acquire a single share of a public company that has been started in the current operating round. If you change the exchange ability's `corporations` attribute to `ipoded` then I think that this method can be simplified to just this:
```suggestion
          super.reject(&:operated)
```

lib/engine/game/g_1832/game.rb
```diff
@@ -236,6 +238,43 @@ def coal_hex
...
+        def use!(**_kwargs)
+          @used = true
+
+          @count_this_or += 1 if @count_per_or
+
+          return unless @count
+
+          @count -= 1
+          owner.remove_ability(self) if !@count.positive? && @remove_when_used_up
+        end
```
ollybh: What is this? It looks like a direct copy of `use!` from `Engine::Ability::Base`, but I don't think it will get called from anywhere in this class.

ollybh (changes requested):
Thanks for taking up the development of this game.

I've got a few suggestions for changes, most of which are removing duplicated code.

lib/engine/game/g_1832/entities.rb
JudeH1: I just thought it would be confusing for the player if the private disappeared when exchanged, as it still continues to pay out until the share exchanged for is operated.

lib/engine/game/g_1832/game.rb
JudeH1: When testing this change I realized both it and my original code have an issue. The rulebook states "You may use this private to purchase one share in any company that has had its president's share purchased during this stock round (including the Central of Georgia Railway in the first stock round)." With this logic, a company could have its presidents share purchased in the first SR, not get fully floated, and then the next sr it would be an exchange target for the LI, when it shouldn't be anymore.

JudeH1: Just fixed this issue. Thank you for all the detail in your comments, I'm a pretty novice programmer so it helps a lot.

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,28 @@
...
+            bundle = action.bundle
+            raise GameError, "Cannot exchange #{action.entity.id} for #{bundle.corporation.id}" unless can_exchange?(company,
+                                                                                                                     bundle)
+
+            buy_shares(company.owner, bundle, exchange: company)
+            @round.players_history[company.owner][bundle.corporation] << action if @round.respond_to?(:players_history)
+            @game.london_corporation = bundle.corporation
+            ability.use!
+            @round.current_actions << action
+            company.close!
```
ollybh: Is this correct? Doesn't the private company want to stay open, paying its revenue out as normal, until the associated corporation pays its first dividend?

lib/engine/game/g_1832/entities.rb
ollybh: @JudeH1?

lib/engine/game/g_1832/game.rb
```diff
@@ -236,6 +240,13 @@ def coal_hex
           @coal_hex ||= hex_by_id('B14')
         end
 
+
+        def exchange_corporations(exchange_ability)
+          candidates = super
+          candidates.reject(&:closed?).reject(&:operated?).select { |c| @parred_this_sr.include?(c.id) }
```
ollybh, line 247: I've realised why `super.reject(&:operated)` will not work. That would include corporations that were parred in a previous stock round but less than 60% of its shares were sold, so had not operated.

So tracking which corporations have been parred in the current stock round is correct. However you've haven't picked the right way to do this. Look at `round_state` in other games. You add a `round_state` method to a step which adds a hash key, something like this:
```
def round_state
  super.merge({ parred_this_sr: [] })
end
```
Then, in the `process_par` method you can say `@round.parred_this_sr << action.corporation`. And in this method could become `super.intersect?(@round.parred_this_sr)`.

ollybh: Actually, the body of this method could just be `@round.parred_this_sr`.

lib/engine/game/g_1832/game.rb
```diff
@@ -326,6 +337,8 @@ def graph_skip_paths(entity)
 
           @skip_paths
         end
+
+
```
ollybh: Blank lines to delete.

lib/engine/game/g_1832/step/dividend.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Game
+    module G1832
+      module Step
+        class Dividend < G1870::Step::Dividend
+
+          def payout_shares(entity, revenue)
+            super
+            if entity == @game.london_corporation
+              @log << 'London Investment closes'
+              li = @game.companies.find { |c| c.sym == 'P4' }
```
ollybh: There is a more efficient way of finding a company by its symbol from a cache, rather than searching through an array:
```suggestion
              li = @game.company_by_id('P4')
```

lib/engine/game/g_1832/step/dividend.rb
```diff
@@ -0,0 +1,26 @@
...
+              li.close! if li
```
ollybh: Do you have to worry about the company not being found?

ollybh (changes requested):
Thanks for the changes. There are a few more things to do before this can be committed.

lib/engine/game/g_1832/entities.rb
JudeH1: Whoops, I left my answer to this question (and my other comment) as a draft :/

JudeH1: On second look though, you're totally right. I think I looked at remove_when_used_up sometime after testing and misunderstood what it actually does. True is the way to go.

lib/engine/game/g_1832/game.rb
ollybh: @JudeH1 You still need to move the `parred_this_sr` from the game class to round state.

lib/engine/game/g_1832/step/exchange.rb
ollybh: @JudeH1 I don't believe this check is needed. Is it?

ollybh (changes requested):

ollybh:
@JudeH1 Can you remove the 'fixes' line from the PR description? That would close the associated issue when this PR is merged, but this only completes one item on the checklist, not the whole issue.

JudeH1:
@ollybh when I remove the logic on 26-29 of exchange.rb, the game breaks when the london private is exchanged

ollybh:
> @ollybh when I remove the logic on 26-29 of exchange.rb, the game breaks when the london private is exchanged

I'd missed that you'd removed the call to `can_exchange?`, this was checking the same thing twice.

philcampeau:
Hey there, I had been tinkering with an idea for a while and finally got it working. [This commit](https://github.com/tobymao/18xx/pull/12475) implements assigning the Cotton Warehouse to a specific city in Atlanta. 

I'm not trying to take over 1832 or anything, just thought I'd submit it since it was working :)

JudeH1:
> Hey there, I had been tinkering with an idea for a while and finally got it working. [This commit](https://github.com/tobymao/18xx/pull/12475) implements assigning the Cotton Warehouse to a specific city in Atlanta.
> 
> I'm not trying to take over 1832 or anything, just thought I'd submit it since it was working :)

That’s great, no need to worry about taking it over, the more fixes the better. Plus I’ve been pretty busy this month and haven’t made a lot of progress, but I should be able to get more done in April. :)

ollybh:
This PR had got buried at the bottom of the pile and I had forgotten about it.

Is this still needed? Or has it been superseded by other 1832 PRs?

If so, have all the review issues been fixed?

PR #12391, opened 2026-02-15
Minor visual improvements in 1824 and 1824 Cisleithania

### LINKED ISSUE #12330: [1824] Suggestions for some visual enhancements.
**What's your idea?**

In 1824 when a coal railway is closed the "home" is replaced with a black token. I think this makes it stand out a bit and would prefer to use a gray token, same color as the hex.

Image

Another potential improvement is to move the coal symbol to the east side of the hex, at least for the two northernmost ones, to avoid it colliding with the revenue symbol.

And a third thing would be to move Dresden revenue slightly (up?) to avoid it covering the track going into the hex.

**Are there related ideas?**

None.

**Did you join the Slack?**

Oh yes.

perwestling:
Fixes #12330 

Still no changes for the coal car issue.

## Before clicking "Create"

Only visual changes so should not break any games.

## Implementation Notes

Change any off board with double values to just show the first.

Block mining centers when mining company is closed, but putting a gray icon there.

### Explanation of Change

Visually pleasing.

### Screenshots

Before change
image

After change
image

### Any Assumptions / Hacks

None

### DIFF
--- a/lib/engine/game/g_1824/map.rb
+++ b/lib/engine/game/g_1824/map.rb
@@ -208,15 +208,15 @@ module Engine
         SARAJEVO = 'city=revenue:yellow_10|brown_50;'\
                    'path=a:1,b:_0,lanes:2,terminal:1;path=a:2,b:_0,terminal:1;border=edge:1;'\
                    'path=a:3,b:_0,terminal:1;path=a:4,b:_0,lanes:2,terminal:1;border=edge:4'
-        SARAJEVO_W = 'path=a:2,b:4,b_lane:2.0;path=a:3,b:4,b_lane:2.1;border=edge:4;border=edge:5'
-        SARAJEVO_E = 'path=a:2,b:1,b_lane:2.0;path=a:3,b:1,b_lane:2.1;border=edge:0;border=edge:1'
+        SARAJEVO_W = 'path=a:2,b:4,b_lane:2.0;path=a:3,b:4,b_lane:2.1;border=edge:4'
+        SARAJEVO_E = 'path=a:2,b:1,b_lane:2.0;path=a:3,b:1,b_lane:2.1;border=edge:1'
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1824/map.rb
```diff
@@ -205,18 +205,18 @@ module Map
                      'path=a:1,b:_0,terminal:1;border=edge:0'
         BUKAREST_2 = 'offboard=revenue:yellow_10|green_30|brown_40|gray_50,groups:Bukarest;path=a:2,b:_0,terminal:1;'\
                      'border=edge:3'
-        SARAJEVO = 'city=revenue:yellow_10|green_10|brown_50|gray_50;'\
+        SARAJEVO = 'city=revenue:yellow_10|brown_50;'\
                    'path=a:1,b:_0,lanes:2,terminal:1;path=a:2,b:_0,terminal:1;border=edge:1;'\
                    'path=a:3,b:_0,terminal:1;path=a:4,b:_0,lanes:2,terminal:1;border=edge:4'
         SARAJEVO_W = 'path=a:2,b:4,b_lane:2.0;path=a:3,b:4,b_lane:2.1;border=edge:4;border=edge:5'
```
perwestling: Looks like there is an extra border statement on SARAJEVO_W and SARAJEVO_E. Remove the one which border off-map.

ollybh: Yes, it should just be edge 4 on the western hex, and edge 1 on the eastern hex.

perwestling:
Sarajevo fixed.

I made a small adjustment on the mines, so that now all symbols are separated.

Does this look OK: @ollybh ?

image

perwestling:
Correction: mine 2, 3, and 4. Mine 1 is unchanged. (Cannot move it as the paths are to close.)

ollybh:
Yes, looks good.

PR #12393, opened 2026-02-16
[Core] Reduce graph redundant walks from backtracking

crericha:
Graph walks with backtracking can exponentially increase the amount of paths that need to walk when converging track is encountered. To explain, here is an example:

image

The south east edge is 0. Each path is given a color: blue, red, green.

Consider a walk coming from edge 3. It first will walk the red path to its full depth and then walk the blue path to its full depth. After completing the red path walk, all further paths down that network are forgotten, so that they can be run by the blue path (see converging logic in the path.rb walk code). Why? Well the blue path may be able to walk that same path in the opposite direction, opening up new areas of the network.

Backtracking makes it worse. The red path will backtrack to the green path, which will build a network out beyond edge 0. Then after the red path is fully explored, the converging logic will have us forget that network, and the blue path will build the same network again out from edge 0 AND backtrack across green to build the same network that the red path build out from edge 5. In the current code with backtracking enabled, this means we've built networks out through edge 0 and edge 5 twice, doubling the compute. Now imagine a series of tiles like this, the compute grows fast. In late game 1837, the token graph compute can time out and lock-up the game.

This PR introduces a simple optimization, that I believe will always be safe. That optimization is to first find all the edges that are reachable from the originating edge. For edge 3, that would be 0 and 5. Then when deciding whether or not to backtrack, if the backtrack would reach an edge that is already reachable, don't backtrack. Since the green path opens up edge 0 for the red path and we see that the blue path already has edge 0 covered, we won't backtrack. The same will be true for the blue path when it evaluates backtracking over green to edge 5. Since its covered by the red path, it won't walk through the green backtrack.

In my testing of late game 1837, this optimization reduced the token graph walk compute time from ~5-6 seconds to 0.7 seconds. 

Before moving forward, I'd like to get some feedback on the approach and if there are any edge cases I'm not thinking of.

Also, there are more complex brown track that my optimization doesn't account for yet.

image

A walk that originates at edge 3 has two different ways of backtracking to end up at edge 1; however, since edge 3 doesn't directly connect to edge 1, it isn't considered. The algorithm would have to go two levels deep to find this redundant walk. Is that going too far? I'd have to figure out which backtrack gets to run and which doesn't since the two path walks from edge 3 are separate walks orchestrated by the path on H26.

This was actually my second approach. My first approach was to globally track directional path walks (ignoring converging). The ideas was that if one path was walked from edge 0 to 3, we already computed whatever network is beyond that and don't need to do it again. However, in testing, I found that it messed up distance calculations. If the first route to go down a path wasn't the shortest between points A and B, a second shorter route, wouldn't walk the redundant directional path, and thus the shorter route would be omitted. I discovered the issue testing 18NY where tokens cost per hex distance across the railroad network (not as the crow flies). It could also be an issue for hex trains or 1870 destinations. I liked that it was a more general solution, but alas, I don't think there is a path forward here. I'm including in my write-up in case someone else sees a way to make it work.

### DIFF
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -173,6 +173,14 @@ module Engine
           end
         end
 
+        # Backtracking optimization to not backtrack to edges already reachable from this edge, avoiding redundant walks
+        reachable_edges_from_edge = []
+        if backtracking && skip && @tile.converging_exit?(skip)
+          reachable_edges_from_edge = hex.paths[skip].reject do |p|
+                                        p == self
+                                      end.flat_map(&:edges).map(&:num).reject { |e| e == skip }
+        end
+
         edges.each do |edge|
           edge_id = edge.id
           edge = edge.num
@@ -181,6 +189,7 @@ module Engine
           if backtracking
             hex.paths[edge].reject { |p| visited[p] || skip_paths&.key?(p) }.each do |p|
               next unless lane_match?(@exit_lanes[edge], p.exit_lanes[edge])
+              next unless (p.edges.map(&:num).reject { |e| e == edge } & reachable_edges_from_edge).empty?
 
               p.walk(skip: edge, visited: visited, skip_paths: skip_paths, counter: counter, skip_track: skip_track,
                      converging: converging || @tile.converging_exit?(edge), walk_calls: walk_calls,
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -176,9 +176,8 @@ module Engine
         # Backtracking optimization to not backtrack to edges already reachable from this edge, avoiding redundant walks
         reachable_edges_from_edge = []
         if backtracking && skip && @tile.converging_exit?(skip)
-          reachable_edges_from_edge = hex.paths[skip].reject do |p|
-                                        p == self
-                                      end.flat_map(&:edges).map(&:num).reject { |e| e == skip }
+          reachable_edges_from_edge =
+            hex.paths[skip].reject { |p| p == self }.flat_map(&:edges).map(&:num).reject { |e| e == skip }
         end
 
         edges.each do |edge|


michaeljb:
> In my testing of late game 1837, this optimization reduced the token graph walk compute time from ~5-6 seconds to 0.7 seconds.

I'm curious to see the reduction in `walk` calls from the debug logs, e.g.,

```
D, [2026-02-21T12:30:16.362000] DEBUG -- :     Graph computed with 976 completed walk calls (skipped 228)
```

> The algorithm would have to go two levels deep to find this redundant walk. Is that going too far? 

This probably depends on how much more going that extra level will help. If your test case already includes a lot of those "complex brown track" tiles, maybe that would be enough to tell if it really needs more improvement? You could see what order of magnitude your `walk` calls are in with the debug logs.

> This was actually my second approach. My first approach was to globally track directional path walks (ignoring converging). The ideas was that if one path was walked from edge 0 to 3, we already computed whatever network is beyond that and don't need to do it again. However, in testing, I found that it messed up distance calculations. If the first route to go down a path wasn't the shortest between points A and B, a second shorter route, wouldn't walk the redundant directional path, and thus the shorter route would be omitted. I discovered the issue testing 18NY where tokens cost per hex distance across the railroad network (not as the crow flies). It could also be an issue for hex trains or 1870 destinations. I liked that it was a more general solution, but alas, I don't think there is a path forward here. I'm including in my write-up in case someone else sees a way to make it work.

Maybe shortest distances from one hex to another could be stored/computed à la Dijkstra's algorithm? This sounds interesting, I might think about this / tinker with it...

> My first approach was to globally track directional path walks

I think I did also try something like that (or at the very least did some pen-and-paper consideration of it) for #10331 but the [ultimate problem there](https://github.com/tobymao/18xx/issues/10331#issuecomment-2198452131) is different than this backtracking feature for 1837--I didn't realize when first looking at this PR that this is unique to 1837.


----

> Before moving forward, I'd like to get some feedback on the approach and if there are any edge cases I'm not thinking of.

Overall your changes and explanations here make sense to me, no obvious edge cases coming to mind for me.

michaeljb (approved):

lib/engine/part/path.rb
```diff
@@ -181,6 +188,7 @@ def walk(
           if backtracking
             hex.paths[edge].reject { |p| visited[p] || skip_paths&.key?(p) }.each do |p|
               next unless lane_match?(@exit_lanes[edge], p.exit_lanes[edge])
+              next unless (p.edges.map(&:num).reject { |e| e == edge } & reachable_edges_from_edge).empty?
```
ollybh, line 191:
```suggestion
              next if p.edges.map(&:num).reject { |e| e == edge }.intersect?(reachable_edges_from_edge)
```

ollybh: Or modify the `reachable_edges_from_edge` assignment above to exclude the current edge, and you can do:
```suggestion
              next if p.edges.map(&:num).intersect?(reachable_edges_from_edge)
```

lib/engine/part/path.rb
```diff
@@ -173,6 +173,13 @@ def walk(
           end
         end
 
+        # Backtracking optimization to not backtrack to edges already reachable from this edge, avoiding redundant walks
+        reachable_edges_from_edge = []
```
ollybh, line 177: Maybe move this assignment inside an `if backtracking` test. The creation of any empty array isn't expensive, but potentially we have this method being called hundreds of thousands of times, so this overhead might be worth avoiding where possible.

ollybh (commented):
I can't see any problems with this idea. I have suggested a couple of micro-optimisations.

ollybh:
@crericha I suggested a couple of optimisations to the code a while back. Do these make sense to you?

PR #12394, opened 2026-02-16
1880 Romania pre-alpha updates

philcampeau:
## Before clicking "Create"

### Explanation of Change

First commit was changing the map coordinates, as well as some tile and hex definitions. 

Second commit includes changes to track.rb, tracker.rb, buy_trains.rb, and multiple updates to the game.rb, including creating dummy companies to nullify some actions taken by the base 1880 code that this game references extensively. 

Feb 23 Update: 

I eliminated the dummy companies, since I needed to rewrite the methods that referenced these companies anyway. At this point, all that remains of the main game is the implementation of private company abilities:

- [x] P1. Banater Montanbahn
- [x] P2. Consorţiu Bethel Henry Strousberg
- [x] P3. Valea Jiului
- [x] P4. The Danube and Black Sea Railway and Free Port of Küstendjie
- [x] P5. REMAR – Reparatii Material Ruland
- [x] P6. Malaxa Manufacturing Company
- [x] P7. Electroputere S.A.
- [x] P8. Orient Express


lib/engine/game/g_1880_romania/step/track.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/track'
+require_relative 'tracker'
+
+module Engine
+  module Game
+    module G1880Romania
+      module Step
+        class Track < Engine::Step::Track
```
crericha, line 10: Do you need to inherit from 1880 here? It looks like 1880 has a couple functions it defines, but I didn't look closely to see if they were applicable.

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -35,6 +33,113 @@ def stock_round
...
+            G1880Romania::Step::Track,
+            G1880::Step::Token,
+            G1880::Step::Route,
+            G1880::Step::Dividend,
+            G1880Romania::Step::BuyTrain,
+            G1880::Step::CheckFIConnection,
+          ], round_num: round_num)
+        end
+
+        def event_modify_borders!
```
crericha: nit: can we call this event open borders?

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -35,6 +33,113 @@ def stock_round
...
+          @log << '-- Event: Borders removed --'
+
+          self.class::BORDERS.each do |coord, edges|
+            hex = hex_by_id(coord)
+            next unless hex
+
+            hex.tile.modify_borders(edges, type: nil)
+          end
+
+          clear_graph
```
crericha, line 235: Is graph actually affected here, since the borders were open?

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -35,6 +33,113 @@ def stock_round
...
+          income = 20 * crossings.size
+          bank.spend(income, p2.owner)
+
+          @log << "#{p2.owner.name} receives #{format_currency(income)} for province crossing"
+        end
+
+        def remove_border_calculate_cost!(tile, entity_or_entities, spender)
+          total_cost, border_types = super
+
+          handle_province_crossing_income(tile.hex, entity_or_entities)
```
crericha: Is it safe to give out income when calculating cost? Is there a better place we can trigger this when the cost is being paid?

lib/engine/game/g_1880_romania/step/buy_train.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Game
+    module G1880Romania
+      module Step
+        class BuyTrain < G1880::Step::BuyTrain
+          def pass!
+            train = @game.depot.upcoming.first
+            train_name = train.name
+            train_index = train.index
+
+            return super if (train_name == '8' && train_index == 1) || %w[8E 2P].include?(train_name) || !discard_trains?
```
perwestling: In Transilvania PR I made another change, that changed 1880 China to introduce a method avoid_discarding_all_trains?.

Advantage is to reduce code duplication a slight bit.

One can still have this file, but then implementing the avoid_discarding_all_trains? is all that is needed.

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -35,6 +33,113 @@ def stock_round
             G1880::Step::BuySellParShares,
           ])
         end
+
+        def operating_round(round_num)
+          G1880::Round::Operating.new(self, [
+            Engine::Step::HomeToken,
+            G1880::Step::RocketPurchaseTrain,
```
perwestling: Rocket private is not used in Romania, so this step is not needed, is it?

philcampeau:
I had to revert the changes to tiles 8858 and 8858a in the /engine/config/tile.rb, because of course that would impact all other games that use those tiles.

Once everything is completed and this game is moved to Alpha, I'll submit a PR to update all games to the FWTWR definitions for those two tiles.

lib/engine/game/g_1880_romania/step/track.rb
philcampeau: I don't need the tile_lay_abilities_should_block method. The max_exits method might be needed, I'll check and include that method alone if needed.

lib/engine/game/g_1880_romania/game.rb
philcampeau: good idea. I'm terrible at naming methods.

lib/engine/game/g_1880_romania/game.rb
philcampeau: you're right. I copied the method from the first one and left that in. Fixed.

lib/engine/game/g_1880_romania/game.rb
philcampeau: correct. Thanks!

lib/engine/game/g_1880_romania/step/buy_train.rb
philcampeau: Yep, I've got it in my new commit.

lib/engine/game/g_1880_romania/game.rb
philcampeau: To be frank, I copied this straight out of 1882's code, which does a similar thing, but this was more of a place holder. 

I've got a better implementation included in my new commit, tell me what you think.

philcampeau:
Everything seems to be working. I'll test more tomorrow and if everything looks good, create a couple of fixture files and we'll be good to go!

lib/engine/game/g_1880_romania/game.rb
philcampeau: this is outdated at this point.

lib/engine/game/g_1880/step/buy_train.rb
```diff
@@ -38,11 +38,15 @@ def round_state
           end
 
           def pass!
-            train_name = @game.depot.upcoming.first.name
-            train_index = @game.depot.upcoming.first.index
-            return super if (train_name == '8E' && train_index == 1) || %w[10 2P].include?(train_name) || !discard_trains?
+            train = @game.depot.upcoming.first
+            return super unless train
+            return super if avoid_discarding_final_trains?(train)
```
crericha: Nit: Its not great to have return super twice. Can we refactor to `return super if !train or avoid_discard_final_trains?(train)`?

lib/engine/game/g_1880_romania/round/stock.rb
crericha, line 1: Nothing is overridden in this file. Remove?

lib/engine/game/g_1880_romania/step/assign.rb
crericha, line 1: There is a non-trivial amount of interweaving between this step and the 1880 version. It could be hard to maintain or reason through issues.

I think the assign logic could be generalized between remar and the 1880 building permit private.

Maybe something can be done with danube port too. At the surface, maybe `@game.abilities(entity, :assign_hexes)` could be or'd with `@game.abilities(entity, :assign_corporation)` in 1880 code?

lib/engine/game/g_1880_romania/step/buy_sell_par_shares.rb
crericha, line 1: Nothing is overridden in this file, that isn't already in 1880. Do you need it?

lib/engine/game/g_1880_romania/step/company_pending_par.rb
crericha, line 1: Nothing is overridden here that isn't in 1880. Do you need it?

lib/engine/game/g_1880_romania/step/parrer.rb
crericha, line 1: Ahh, I see. This is why you overrode those other steps. Its hard to see how this differs from the 1880 version other than TR versus BCR. Can you provide some context of where the rules are different? Maybe there is an opportunity to make a small change to 1880 and not have to duplicate a bunch of code.

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+
+require_relative '../../g_1880/step/route'
+
+module Engine
+  module Game
+    module G1880Romania
+      module Step
+        class Route < G1880::Step::Route
+          def actions(entity)
+            regular_actions = super
```
crericha: I believe this needs to be super.dup in order to not mutate the parent classes actions array.

crericha: actually you aren't mutating the array. This only needs to change if you take my recommendation below.

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+            return regular_actions if regular_actions.empty?
+            return regular_actions + ['choose'] if choosing_tender?(entity)
```
crericha: Small suggestion on how to simplify this code

actions.append('choose') if !actions.empty? && choosing_tender?(entity)
actions

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+
+            regular_actions
+          end
+
+          def choosing?(entity)
```
crericha: This method isn't used here. Is it required by the view logic?

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+            @tender_train = tender_train_choices(entity)[action.choice.to_i]
+            @log << "#{entity.name} attaches the REMAR tender to a #{@tender_train.name} train"
+            attach_tender
+          end
+
+          def process_run_routes(action)
+            super
+            detach_tender if @tender_train
+
+            bonus = @game.stock_market_bonus(action.entity)
```
crericha, line 46: Is the bonus rules different? Or is this aimed at improving the logging?

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+            return unless bonus.positive?
+
+            @round.extra_revenue = (@round.extra_revenue || 0) + bonus
+            @log << "#{action.entity.name} receives #{@game.format_currency(bonus)} stock market bonus"
+          end
+
+          private
+
+          def choosing_tender?(entity)
+            @tender_train ||= nil
```
crericha: should this be in the round state?

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+            else
+              town_part_index = @tender_train.distance.index { |n| n['nodes'] == ['town'] }
+              if town_part_index
+                new_distance = @tender_train.distance.map(&:dup)
+                new_distance[town_part_index]['pay'] += 1
+                new_distance[town_part_index]['visit'] += 1
+                @tender_train.distance = new_distance
+              else
+                @tender_train.distance = @tender_train.distance + [{ 'nodes' => ['town'], 'pay' => 1, 'visit' => 1 }]
+              end
```
crericha: There is opportunity to simplify this code, so you don't have to add the town in multiple places.

if distance is numeric, add nodes for city/offboard/town option
then if no 'town' node, add an empty town node of pay 0 and visit 0.
then always add in the pay 1 and visit 1

up to you whatever you prefer.

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+            end
+
+            @tender_train.name = tender_name(@tender_train.name)
+            @game.clear_graph
```
crericha, line 85: Why does the graph need to be cleared?

lib/engine/game/g_1880_romania/step/route.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+
+          def detach_tender
+            @tender_train.name = @tender_original_train.name
+            @tender_train.distance = @tender_original_train.distance
+            @tender_original_train = nil
+            @tender_train = nil
+            @game.clear_graph
```
crericha, line 93: Why does the graph need to be cleared?

lib/engine/game/g_1880_romania/step/simple_draft.rb
```diff
@@ -0,0 +1,33 @@
...
+
+            minor.owner = player
+            minor.float!
+
+            @minors.delete(minor)
+
+            @log << "#{player.name} chooses #{minor.full_name} (#{minor.name})"
+
+            # this line hijacks the BCR share assignment from 1880 China to assign to the TR instead
+            assign_bcr_share_to_fi(action.entity, minor) if player.shares.find { |s| s.corporation.id == 'TR' }
```
crericha, line 24: Maybe add a bcr_corporation instance var to game, like with the rocket and then perhaps code can be reused?

lib/engine/game/g_1880_romania/step/special_choose.rb
```diff
@@ -0,0 +1,103 @@
...
+
+require_relative '../../../step/special_choose'
+
+module Engine
+  module Game
+    module G1880Romania
+      module Step
+        class SpecialChoose < Engine::Step::SpecialChoose
+          def actions(entity)
+            return [] unless entity.company?
```
crericha: seems unnecessary given the next check for two explicit entitites

lib/engine/game/g_1880_romania/step/tracker.rb
```diff
@@ -0,0 +1,66 @@
...
+          [total_cost, border_types]
+        end
+
+        def can_ignore_borders?(entity)
+          return false unless entity&.corporation?
+
+          entity.owner == @game.consortiu&.owner
+        end
+
+        def legal_tile_rotation?(entity_or_entities, hex, tile)
```
crericha, line 25: What needed to be added to this method?

lib/engine/game/g_1880_romania/corporation.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require_relative '../g_1880/corporation'
+
+module Engine
+  module Game
+    module G1880Romania
+      class Corporation < G1880::Corporation
+        def all_abilities
```
crericha, line 9: I thought the game class already had this? Are we sure this isn't redundant? I would have expected 1880 to need logic like this.

lib/engine/game/g_1880_romania/minor.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require_relative '../g_1880/minor'
+
+module Engine
+  module Game
+    module G1880Romania
+      class Minor < G1880::Minor
+        def all_abilities
```
crericha, line 9: Hoping we don't need this. tbd.

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -14,27 +14,358 @@ class Game < G1880::Game
...
+          'signal_end_game' => ['Signal End Game', 'Game ends 3 ORs after purchase/export of last 6E train']
         ).freeze
 
+        def init_minors
+          game_minors.map { |minor| G1880Romania::Minor.new(**minor) }
+        end
+
+        def upgrades_to_correct_label?(from, to)
+          # B-labeled tiles are only for Bucuresti (I17)
+          return false if to.label.to_s == 'B' && from.label.to_s != 'B'
```
crericha, line 171: I thought label matching was a standard check in upgrades_to_correct_label?

lib/engine/game/g_1880_romania/game.rb
```diff
@@ -14,27 +14,358 @@ class Game < G1880::Game
...
+          end
+
+          clear_graph if removed
+        end
+
+        def routes_revenue(routes)
+          # Override to exclude stock_market_bonus from routes_revenue so the autorouter
+          # can compare route combinations fairly. The bonus is added in extra_revenue
+          # via the route step.
+          routes.sum(&:revenue)
```
crericha, line 302: Is this a bug in 1880?

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: old code I forgot to remove. Done

lib/engine/game/g_1880/step/buy_train.rb
philcampeau: true. done.

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: Improves the logging, it also improves the autorouting. I explained it in a comment on the `def routes_revenue` method in /1880Romania/game.rb

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: done

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: done

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: done

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: done

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: done

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: updated in `changes to route step` commit [447a2c5](https://github.com/tobymao/18xx/pull/12394/commits/447a2c56f048ebe7a86699e25a9d70184e8180b0)

lib/engine/game/g_1880_romania/step/special_choose.rb
philcampeau: true. fixed.

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: done

lib/engine/game/g_1880_romania/step/special_choose.rb
philcampeau: true. removed.

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: I incorporated this into 1880's code, per slack discussion with crericha

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: You're right, it didn't need to be cleared. Fixed.

lib/engine/game/g_1880_romania/step/route.rb
philcampeau: You're right, didn't need to be cleared. Fixed.

lib/engine/game/g_1880_romania/corporation.rb
philcampeau: Changed REMAR's ability to an assignment, that fixes this.

lib/engine/game/g_1880_romania/step/assign.rb
philcampeau: Ok I was able to drastically simplify this.

lib/engine/game/g_1880_romania/step/buy_sell_par_shares.rb
philcampeau: Yes, needed for parrer changes

lib/engine/game/g_1880_romania/step/parrer.rb
philcampeau: drastically simplified

lib/engine/game/g_1880_romania/step/simple_draft.rb
philcampeau: I tried this. the problem is that TR works like BCR half the time, but other times it works very differently.

lib/engine/game/g_1880_romania/game.rb
philcampeau: yep fixed

lib/engine/game/g_1880_romania/game.rb
philcampeau: this has been discussed in slack and corrected across both games.

lib/engine/game/g_1880_romania/round/stock.rb
philcampeau: you're right. I think I created this for something, and then it ended up getting moved out but I kept the file.

lib/engine/game/g_1880_romania/step/tracker.rb
philcampeau: I modified this per our discussion in slack. Now there's a small change to the `legal_tile_rotation?` method in the base Tracker step, and added a little modifier in romania's code

lib/engine/game/g_1880/step/parrer.rb
```diff
@@ -39,10 +39,14 @@ def choices
 
           def president_percent_choices
             max_shares = (@parring[:entity].cash.to_f / @parring[:share_price].price).floor
-            max_percent = @parring[:corporation] == @game.bcr ? 20 : max_shares * @parring[:corporation].share_percent
+            max_percent = private_company_president_corp? ? 20 : max_shares * @parring[:corporation].share_percent
             { 20 => '20%', 30 => '30%', 40 => '40%' }.select { |k, _v| k <= max_percent }
           end
 
+          def private_company_president_corp?
```
crericha: Can we pass the corp into this function, so it doesn't need to know about the @parring aspect?

lib/engine/step/tracker.rb
```diff
@@ -449,6 +449,10 @@ def legal_tile_rotation?(entity_or_entities, hex, tile)
         true
       end
 
+      def hex_neighbor_valid?(_entity, hex, edge)
```
crericha: valid or exists? I think its checking the latter?

crericha (approved):

crericha (approved):

PR #12398, opened 2026-02-19
[18 Rhineland] New subgames: 18 Rhineland (2024) and 18 Lower Rhineland Area

perwestling:
Both are from the 18 RHL 2023 kickstarter.

## Before clicking "Create"

Should not affect existing 18Rhl games.

## Implementation Notes

Just some basic files for new 18 Rhineland and 18 LRA games.

### Explanation of Change

New version (2024) of 18 Rhineland.
It also included the 18 LRA variant map.

### Screenshots

3 changes of 18 Rhl map compared to 18 Rhineland.

### Any Assumptions / Hacks

In the first attempt I tried to change base game but now instead I just add the two new and let 18 Rhl still be base game.


public/logos/18_rhl/MKB.svg
ollybh, line 1: The text doesn't fit inside the token circle. Can the font size be reduced?

lib/engine/game/g_18_rhineland/map.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G18Rhineland
+      module Map
+        def game_tiles
+          tiles = super
```
ollybh, line 8: You've got to be careful about modifying array/hash definitions in subclasses, these can end up affecting the base games.
```suggestion
          tiles = super.dup
```

ollybh: To be safe, you can change the TILES definition in the base class from using `.freeze` to using `.deep_freeze`. This should cause an error if anything is modified, which is much better than the hard-to-reproduce bugs you get if these constants are altered.

lib/engine/game/g_18_rhineland/map.rb
```diff
@@ -0,0 +1,45 @@
...
+          tiles['7'] = 4
+          tiles['25'] = 1
+          tiles['58'] = 8
+          tiles['88'] = 8
+
+          # TODO: Add tile 1990, 1991 if RS tiles are implemented
+          tiles
+        end
+
+        def modify_map(map)
```
ollybh, line 21: You will also need to make sure you don't modify the base class's definition. You haven't yet added the call to `modify_map`, but you'll need to do something like this:
```
map = self.class::HEXES.to_h { |k1, v1| [k1, v1.to_h { |k2, v2| [k2.dup, v2] }] }
```

lib/engine/game/g_18_lra/trains.rb
```diff
@@ -0,0 +1,35 @@
...
+            2
+          when '6'
+            6
+          when '8'
+            0
+          end
+        end
+
+        def game_trains
+          trains = super
```
ollybh, line 23:
```suggestion
          trains = super.map(&:dup)
```

ollybh (changes requested):
Is there a simple summary of the differences between these three games? Would it make more sense for these to be implemented as optional rules instead of game variants?

However this is done, you need to make sure you don't modify the contents of array/hash constants. I suggest you change the base class definitions to use `.deep_freeze` instead of `.freeze`.

lib/engine/game/g_18_rhineland/map.rb
perwestling: True. I need to modify the tiles definition in the base game as well, due to optional features. E.g. Ratingen variant adds tiles, which is not present in the base game.

public/logos/18_rhl/MKB.svg
perwestling: I'll try. And possibly this should be black, not gray, as MKB has black company charter.

MKB does not exist in 18Rhl/18Rhineland, only in 18LRA.
So I could possibly create an 18_lra directory and put it there? (But I put it under 18_rhl to keep all logos together.)

lib/engine/game/g_18_rhl/map.rb
```diff
@@ -8,278 +8,280 @@ module Map
...
+              'color' => 'gray',
+              'code' => 'city=revenue:60,slots:4;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;'\
+                        'path=a:5,b:_0;label=D/DU/K',
+            },
+            '932V' =>
+            {
+              'count' => 2,
+              'color' => 'gray',
+              'code' => 'city=revenue:60,slots:4;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;'\
+                        'path=a:5,b:_0;label=D/K',
```
perwestling, line 191: Need to revisist this.

1. I am not sure K and S are used in 18Lra; has asked BGG forum if they are.
2. Also, only a few K and S icon should be shown on the board, and none on tiles. In reality the few K/S that are should be clearly presented. And the remaining could be pentagons on the board, to show potential placement for K/S. The pentagons can be removed at setup - only shown when map is viewed from the outside.

Maybe the K/S on the off-board areas could be put in separate city slots to make them appear better?

lib/engine/game/g_18_rhl/map.rb
```diff
@@ -8,278 +8,280 @@ module Map
...
+            },
+            '931' =>
+            {
+              'count' => 1,
+              'color' => 'brown',
+              'code' => 'city=revenue:40,slots:3;path=a:0,b:_0;path=a:2,b:_0;path=a:3,b:_0;'\
+                        'path=a:4,b:_0;label=AC',
+            },
+            '932' =>
+            {
```
perwestling, line 180: Have received errata on the tiles from Marflow, so this file need to be updated.

public/logos/18_rhl/MKB.svg
perwestling: Reduced font size from 4 to 3, so it fits.

And changed MKB background to black instead of gray. Looks fine.

lib/engine/game/g_18_rhl/map.rb
perwestling: Got an answer: https://boardgamegeek.com/thread/3666574/18lra-coal-and-steel

Rule 6.2.4.5 specify Harbour bonuses: 10 M per symbol included in routes passing harbours.
This bonus can be added to the bonus legend.

I still think making icons for K/S (and maybe also ERh) can be made more prominent.  But it is enough for 3 x K, 1 x S and maybe 3 ERh.

perwestling: Might need to be explicit that ERh need to be from West and East.

perwestling:
> Is there a simple summary of the differences between these three games? Would it make more sense for these to be implemented as optional rules instead of game variants?
> 
> However this is done, you need to make sure you don't modify the contents of array/hash constants. I suggest you change the base class definitions to use `.deep_freeze` instead of `.freeze`.

From a User perspective it looks like it is optional rules. You select 18Rhl and then click in 18Lra, 18Rhineland (2024), or none. The optional variants are different for all three games.

I think 18Rhl 2007 and 18Rhineland 2024 are pretty close, but I hope 2024 Rhineland should be the main 18Rhl version when it is time for beta.

18Lra does have some differences that might make it more clear as a game of its own.

I have started to put up a list of differences between 18Rhl (2007) and 18Rhineland (2024) on the wiki page.

So far there are very little code in the g_18_rhineland.

lib/engine/game/g_18_lra/map.rb
```diff
@@ -0,0 +1,546 @@
...
+                        'path=a:3,b:_0,track:narrow;path=a:0,b:_0;path=a:1,b:_0;'\
+                        'border=edge:4,type:impassable,color:blue;border=edge:5,type:impassable,color:blue;'\
+                        'icon=image:18_rhl/trajekt,sticky:0;icon=image:../logos/18_rhl/S,sticky:0',
+            },
+            'L30' =>
+            {
+              'count' => 1,
+              'color' => 'green',
+              'code' => 'city=revenue:20,slots:2;label=RU;upgrade=cost:30,terrain:river;'\
+                        'path=a:0,b:_0,track:narrow;path=a:3,b:_0;path=a:5,b:_0;'\
```
perwestling, line 267: This should probably have a Trajekt icon as it has a river crossing with a ferry.

lib/engine/game/g_18_lra/entities.rb
```diff
@@ -0,0 +1,296 @@
...
+module Engine
+  module Game
+    module G18Lra
+      module Entities
+        CORPORATIONS = [
+          {
+            float_percent: 50,
+            name: 'Aachen-Düsseldorf-Ruhrorter Eisenbahn',
+            sym: 'ADR',
+            tokens: [0, 60, 80, 100],
```
perwestling: I think Harbour tokens need to be redisgned.

Instead of specifying them this way, they could be added for each corporation (excluding those without).
And the harbour token could be created dynamically:
- cost for it can be 20 more than rightmost one
- logo could be <CORPORATION>_harbour.svg (and same for alt), compare how it is done in 1822 for destination (it might be easy to just add a rectangle - like how it is done in 1822, having a more advanced visual marker can be saved for the future)
- type could be harbour, maybe?
- showing both cost might not be possible, but compare B&O in 1846, where this is shown as an ability for the corporation

perwestling: Have done most of this now, using Destination style a la 1822 for the tokens.

perwestling:
I will make a new attempt of this, maybe wait with LRA until later?  

Anyway, close this for now.

PR #12399, opened 2026-02-20
[1837] Stock market tokens are placed when floating

### LINKED ISSUE #11698: [1837] Share price marker appears on stock chart when major is parred [minor]
The rules say:
"As soon as 50% of a corporation’s shares have been bought from the initial offering, it floats ....
A marker is placed on the stock market..."

The program is placing the stock token immediately the par value is set (i.e. when the director's certificate is bought), just like it would in most titles, but 1837 is different.

Potentially this could affect the order of operation if two majors were floated in the same SR at the same price, but were floated in the opposite order from when they were parred.

crericha:


Zwergenpunk (approved):

andrewzwicky (approved):

perwestling (approved):
I check changes in the below games and they look OK.

1824 (and Cisleithania)
1862 Solo
1893
18SJ

ollybh (commented):
This is a lot of code churn. An alternative approach which doesn't require as much would be to change the code in Engine::StockMarket to something like this:
```
    def place_token!(corporation, share_price)
      share_price.corporations << corporation
      corporation.share_price = share_price
    end

    def set_par(corporation, share_price, place_token: true)
      place_token! if place_token
      corporation.par_price = share_price
      corporation.original_par_price = share_price
    end
```

Then you could just change the call in the 1837 code to `@stock_market.set_par(corporation, share_price, place_token: false)` and call `@stock_market.place_token!` from `float_corporation`.

This does mean you end up with a call to `StockMarket.set_par` that isn't affecting the state of the stock market object, which is a little odd. But it would be a much smaller overall change.

michaeljb (approved):

ollybh:
@crericha This PR now has merge conflicts.

Do you want to fix these and have it merged in the current form, or look at my suggestion of a more limited change?

PR #12401, opened 2026-02-21
[Core] Selection auction next_entity! shouldn't autopass entities when no privates are left

### LINKED ISSUE #11794: [1837] Wrong assignment of priority at end of initial auction
https://18xx.games/game/209839

I bought the last item in the initial action (LRB) but I was then given the priority.
The rules state: "Now give the priority deal card to the player to the left of the last
player to purchase something."
So the priority should have gone to daveb.

crericha:


ollybh (commented):
Is the wrong assignment of priority occurring when all the minors have been auctioned, but there are still players left that can afford the second UG1/UG3 shares? (The games linked in the issue are both archived, so I can't check them).

If it is, do you need to unpass players once the minor auction has finished? Otherwise could you have a player passing when they could have bought an additional share, ending the round before other players get the chance to buy these shares?

ollybh:
@crericha ?

PR #12408, opened 2026-02-27
[18Cuba] Home token logic and map update

chrstas:
This PR updates the map, separates the trains and includes the home token logic for minors 

- Update of map:
   - [x] FC hexes and icons
   - [x] Sugar cane fields can have revenue (hidden towns)
   - [x] tiles with "sugar mills" for minor/major tokens 
   - [x] make restrictions sticky

- Moved trains out to separate module, not letting game.rb becoming too big
- Allow minors (and FEC major) to place home token with token timing :par according to rules

## Before clicking "Create"


lib/engine/game/g_18_cuba/game.rb
```diff
@@ -26,6 +28,7 @@ class Game < Game::Base
                         brightGreen: '#6ec037')
         TRACK_RESTRICTION = :permissive
         CURRENCY_FORMAT_STR = '$%s'
+        HOME_TOKEN_TIMING = :par
```
ollybh: From my reading of the rules, minor companies place their home tokens when they are parred, but major companies place theirs on the first operating turns.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -312,6 +197,22 @@ def next_round!
             end
         end
 
+        def home_token_locations(corporation)
+          if corporation.type == :minor || corporation.sym == 'FEC'
+            hexes.select do |hex|
+              hex.tile.cities.any? { |city| city.tokenable?(corporation, free: true) }
+            end
```
ollybh, line 213: This isn't fully implementing the rules – can you either this, or add a TODO note to say that it isn't finished?

- Minors can't start in Y or H cities, or in cities with other sugar mills.
- There is a corner case for FEC where it can add its token to a fully tokened city if there are no empty city slots. This should be implemented by using a 'cheater' token.

lib/engine/game/g_18_cuba/map.rb
```diff
@@ -421,6 +427,8 @@ module Map
           },
         }.freeze
 
+        FC_HEX = %w[B9 B11 C14 D15 D19 E22 E24 G24].freeze
```
ollybh: E24 isn't marked with an FC logo on the map.
```suggestion
        FC_HEX = %w[B9 B11 C14 D15 D19 E22 G24].freeze
```

lib/engine/game/g_18_cuba/game.rb
chrstas: thanks - I have added now restrictions on minor side (Y, H and no other sugar mill). The FEC logic and corner case I have put into a TODO note

lib/engine/game/g_18_cuba/game.rb
chrstas: you  right - have adjusted it

chrstas:
I have made the adjustments, the FEC logic I left as TODO note. 

The Tile 512 topic I have fixed it with the last commit in Cuba locally, but if I understand you right, the better way would be to correct it in the config/tiles. I will take care of this in a separate PR.

Also corrected the slot numbers of L46, L76 and L77

Thanks as always...

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -312,6 +197,29 @@ def next_round!
             end
         end
 
+        def home_token_locations(corporation)
+          # TODO: FEC corner case to be added with no available token -> use cheater token
+          return super if corporation.type != :minor && corporation.sym != 'FEC'
+
+          hexes.select do |hex|
+            # no token allowed on Y and H cities
+            next false if hex.tile.labels.any? { |l| %w[Y H].include?(l.to_s) }
```
ollybh, line 206: This test is only needed for minors. FEC can start in these cities.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -312,6 +197,29 @@ def next_round!
...
+
+            hex.tile.cities.any? do |city|
+              next false unless city.tokenable?(corporation, free: true)
+
+              # no other minor may already be here
+              city.tokens.none? { |t| t&.corporation&.type == :minor }
```
ollybh, line 212: This is also for minors only.

lib/engine/game/g_18_cuba/todo.md
ollybh, line 1: This to do list would be better done as a Github issue, to avoid having to submit a pull request and wait for it to be merged every time you think of something to be added/changed.

chrstas:
Home token restrictions changed to be only applied for minors - you are right there. FEC logic will follow in a separate PR - TODO comment is in.

todo.md removed will follow your suggestion regading issues in github

thanks again

PR #12411, opened 2026-03-01
Hide market if not used, fixes #12406

### LINKED ISSUE #12406: [1824] Possibly incorrect liquidity
It looks like liquidity is wrong for Tacitus, 115 wrong, which is the value of BH share. Is it that it might be more than 50% in the stock market?

Same for the other player that owns 10% of BH.

Tested to clone game and the players cannot sell BH (as 50% is in bank pool) so it is just the liquidity calculation that is wrong.

**What's your Game ID?**

https://18xx.games/game/243666?action=214

**How to Reproduce the Bug?**

Not sure. Check game.

**Which Browser / Platform?**

Chrome, but probably not browser related.

perwestling:
Add possibility to hide Market in Corporation and Spreadsheet if not used. Do use this for 1824.

## Before clicking "Create"

Only visual changes - does not affect game play.

## Implementation Notes

Added a display_bank_pool? method to base which is set to true as default
In case this returns false, the Market column in Spreadsheet, and Market row in Corporation, is hidden.
This should not affect any other titles at the moment.

### Explanation of Change

Market is not used in 1824 (except 2 player). Therefor it is just as well to hide it.

### Screenshots

These screen shots is obsolete, and do contain other changes.

Entities view for 1824 before change:
image

Entities view for 1824 after change:
image

Spread sheet view for 1824 before change:
image

Spread sheet view for 1824 after change:
image

### Any Assumptions / Hacks

Assumption is that this should not break any games.

It is probably possible to add this change (hide market) to some other titles, e.g. 1893 Cologne, but I have not done that.

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -511,12 +511,15 @@ module View
           ])
         end
 
-        separator_props = { style: { height: '1px', borderBottom: '1px solid currentColor', }, }
-        separator = [
-          h('td.left', separator_props, ''),
-          h('td.right', separator_props, ''),
-          h('td.passed_number', separator_props, '')
-        ]
+        separator = []
+        if !pool_rows.empty? && (!player_rows.empty? || !other_corp_rows.empty? || !other_minor_rows.empty?)
+          separator_props = { style: { height: '1px', borderBottom: '1px solid currentColor', }, }
+          separator << h('tr.market', separator_props, [
+            h('td.left', separator_props, ''),
+            h('td.right', separator_props, ''),
+            h('td.passed_number', separator_props, '')
+          ])
+        end
 
         rows = [
           *pool_rows,


lib/engine/game/g_1824/game.rb
```diff
@@ -791,6 +795,11 @@ def can_go_bankrupt?(_entity, _corporation)
           false
         end
 
+        def can_dump?(entity, bundle)
+          # If market is not used, then Bank cannot exceed 50%
+          super && (market? || (bundle.corporation.num_ipo_shares * 10) + bundle.percent <= 50)
```
perwestling: Code could possible be changed to work for even 2p 1824, by calculating shares in Bank and Market.
Consider adding unit test for 1824 and Cisleithania.

perwestling:
@ollybh, some responses.

> You have changed the spreadsheet display to have a column of zero-width spaces instead of the pool share counts. This column should be removed if it is not needed. Wrap each `h(:th, …)` or `h(:td, …)` call in an `if` test.

I did this first but that destroyed the separator in each line. OK, I will make another try.
 
> Could the new `market?` method be called something else? It's not clear from the name what this does. Every game has a stock market, setting this to false might suggest a game doesn't. How about `display_bank_pool?`? And can you add a comment to this method, explaining what it does?

Tried "has_market" and "has_market?" but style check complained. 

I suppose "use_bank_pool?" could work? The method returns true if the game title has a bank pool, but used "market" as that is the term used in the UI.

1824 (and some more titles, 1893, 1880) does not use the separate IPO / bank pool (except 2 player Cisleithania variant of 1824), and instead just use one place for the shares.

perwestling:
@ollybh Maybe I should split this PR into 2? The liquidity issue is separate. (And maybe that should be put together with the EMR fix I am working on for 1824?)

The visual improvement is not needed for this issue. And could probably be fixed for 1893 Cologne and 1880 China (and Romania) as well? Probably more titles as well.

The visual fix could then possibly also try to fix the separator which do disappear with my change.

What do you think?

lib/engine/game/g_1824/game.rb
perwestling: 2p rules are different, see [Lonny's posting](https://boardgamegeek.com/thread/3577945/cisleithania-variant-the-forgotten-rule). So the rules for the 2p variant need to be reviewed.

ollybh:
> @ollybh Maybe I should split this PR into 2? The liquidity issue is separate. (And maybe that should be put together with the EMR fix I am working on for 1824?)

Yes, that would better.

ollybh:
> > You have changed the spreadsheet display to have a column of zero-width spaces instead of the pool share counts. This column should be removed if it is not needed. Wrap each `h(:th, …)` or `h(:td, …)` call in an `if` test.
> 
> I did this first but that destroyed the separator in each line. OK, I will make another try.

I hadn't thought of that. You will need to add a `borderRight` to the style definition for the bank shares if there isn't going to be a column for the pool shares.

https://github.com/tobymao/18xx/blob/dc8c2aaab804ec781edbfc78f8b89493094958aa/assets/app/view/game/spreadsheet.rb#L542-L548

> > Could the new `market?` method be called something else? It's not clear from the name what this does. Every game has a stock market, setting this to false might suggest a game doesn't. How about `display_bank_pool?`? And can you add a comment to this method, explaining what it does?
> 
> Tried "has_market" and "has_market?" but style check complained.
> 
> I suppose "use_bank_pool?" could work? The method returns true if the game title has a bank pool, but used "market" as that is the term used in the UI.

As this method is only changing how things are shown, not any of the game logic, I would suggest that it is called `show_<something>?` or `display_<something>?`. Maybe `share_pool`? We have a SharePool class that manages shares that have been sold, so that might be the right name.

> 1824 (and some more titles, 1893, 1880) does not use the separate IPO / bank pool (except 2 player Cisleithania variant of 1824), and instead just use one place for the shares.

And 18CZ, I think.

lib/engine/game/g_1824/game.rb
perwestling: Removed from this PR.

perwestling:
Have removed the bank restriction code; moved to other PR.

Fixed review comments and spreadsheet. Still remains to fix entity view.

Also need to implement hide in a couple of more titles.

Finally, the PR description need to be updated to reflect the changing PR.

perwestling:
I think this is a nice idea, and I almost got it to work, but....

If there are no shares in Bank/IPO the row containing the stock price disappear.
Some games also use Market line to show different colors, based on stock price.

So implementing this would require a bigger redesign of corporation.rb view file.

I will close this for now.

PR #12420, opened 2026-03-08
[1824] Correct emergency financing

### LINKED ISSUE #12412: [1824] Incorrect Emergency Amount
Incorrect needed amount for emergency buy, see below image.

Text states:

"KK2 has 159G. Azathul must contribute 21G for KK2 to afford a train from the Depot."

Cheapest train in depot is 240G, so it should be 60 more ie 81 if director is helping.

"Azathul has 27G in cash.  Azathul does not have enough liquidity to contribute towards KK2 buying a train from the Depot. KK2 must buy a train from another corporation, or Azathul must take a loan of at least -6G"

Take a loan of at least -6G? 27G in cash should be enough if 21G had been what was required.

Image

Also, Azathul should be able to sell shares if they want to, instead of taking a loan. But taking a loan is an option.

Here is the situation:

Image

(SD and MS are sellable, BH is not.)

**What's your Game ID?**

https://18xx.games/game/243666?action=311

[Gist](https://gist.github.com/perwestling/c9f6a928a7d5586f8a1c4f3bd0a54909)

**How to Reproduce the Bug?**

This happens during Emergency Buy.

**Which Browser / Platform?**

Chrome, on Windows, but browser probably does not matter.

perwestling:
Fixes #12412

Emergency financing (EF) was broken for 1824.

Now the correct amount needed to assist with should be shown.
And you are not allowed to sell more shares than needed (to buy most expensive train)

Also added a guard against selling shares, and then cross-buy.

Finally, added a proper behavior for selling shares to bank - at most 50% after sale.
 
## Before clicking "Create"

Tested validate with latest database. 
2 games did not validate correctly, but those are already broken by commits on master.
So it does not seem any game is broken by this change (as it gives possibility to do things
that was not previously possible to do.

## Implementation Notes

Two constants have changed value:
- EBAY_FROM_OTHERS changed from :always to :value, as a player cannot pay more than value for a cross-buy
- EBUY_SELL_MORE_THAN_NEEDED changed from :true to :false, to stop from selling more shares than needed (but needed is most expensive depot train for 1824)

### Explanation of Change

Emergency Financing worked badly. So I have rewritten it a bit to work better.

### Screenshots

N/A

### Any Assumptions / Hacks

Possibly a unit test need to be written.

### DIFF
--- a/lib/engine/game/g_1824/step/buy_train.rb
+++ b/lib/engine/game/g_1824/step/buy_train.rb
@@ -11,7 +11,6 @@ module Engine
             super.merge(
               {
                 train_exchanged_by_entity: [],
-                shares_sold_by_owner: false,
               }
             )
           end
@@ -99,24 +98,9 @@ module Engine
               raise GameError, 'Coal railways can only own g-trains'
             end
 
-            ensure_no_cross_buy_train_after_selling_share(action)
-
             super
 
             @game.two_train_bought = true if train.name == '2'
-          end
-
-          def process_sell_shares(action)
-            super
-            @round.shares_sold_by_owner = true
-          end
-
-          def ensure_no_cross_buy_train_after_selling_share(action)
-            return unless @round.shares_sold_by_owner
-            return if action.train.from_depot?
-
-            # During emegency financing we may not buy cross-over train if we have sold shares
-            raise GameError, 'Cannot buy train from other corporation if sold shares during emergency financing'
           end
 
           def buyable_trains(entity)


lib/engine/game/g_1824/game.rb
```diff
@@ -52,12 +52,17 @@ class Game < G1837::Game
         MUST_SELL_IN_BLOCKS = false
 
         EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false
-        EBUY_FROM_OTHERS = :always
         EBUY_SELL_MORE_THAN_NEEDED = true
-        EBUY_SELL_MORE_THAN_NEEDED_SETS_PURCHASE_MIN = true
         EBUY_CAN_TAKE_PLAYER_LOAN = true
         MUST_BUY_TRAIN = :always
```
perwestling, line 60: Probably incorrect, keep as always.

But there is code in train.rb - that check for this, and might be incorrect. Need to test if it works to restore the value => can we still buy a more expensive train?

lib/engine/game/g_1824/step/buy_train.rb
```diff
@@ -30,9 +30,7 @@ def actions(entity)
           end
 
           def can_entity_buy_train?(entity)
-            return false if entity.receivership?
-
-            entity.operator?
+            !entity.receivership?
```
perwestling, line 32: Maybe keep the original? And the change at line 24 might not be needed, in that case, as can_buy_train? will return false if not operator.

ollybh:
I think this is reinventing things that are already present in the code base. If you set these constants do you get the correct behaviour?
```
EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false # can buy any train from the bank
EBUY_FROM_OTHERS = :value # can cross-buy trains, but not for more than face value
EBUY_SELL_MORE_THAN_NEEDED_SETS_PURCHASE_MIN = true # cannot issue more shares than needed to buy a train
EBUY_CAN_TAKE_PLAYER_LOAN = true # can take a loan, do not have to sell shares first
EBUY_PRES_SWAP = false # cannot sell shares if it would cause a change in director
```
If not, what isn't working?

perwestling:
> I think this is reinventing things that are already present in the code base. If you set these constants do you get the correct behaviour?
> 
> ```
> EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false # can buy any train from the bank
> EBUY_FROM_OTHERS = :value # can cross-buy trains, but not for more than face value
> EBUY_SELL_MORE_THAN_NEEDED_SETS_PURCHASE_MIN = true # cannot issue more shares than needed to buy a train
> EBUY_CAN_TAKE_PLAYER_LOAN = true # can take a loan, do not have to sell shares first
> EBUY_PRES_SWAP = false # cannot sell shares if it would cause a change in director
> ```
> 
> If not, what isn't working?

I had a rule wrong, so EBUY_FROM_OTHERS = :value should be fine.
I studied the code in buy_trains.rb again, and I think my question mark has been removed.

EBUY_SELL_MORE_THAN_NEEDED_SETS_PURCHASE_MIN should be false, though, as in 1824 you can sell shares for more than min to buy a more expensive train, and all available trains in depot should be available as choices.

I do not think the code support this case fully, so therefor this PR implements checks to stop illegal selling.

lib/engine/game/g_1824/game.rb
perwestling: Should be :value, as max price for cross buy as train value, during Emergency Financing.

lib/engine/game/g_1824/step/buy_train.rb
perwestling: No, this is needed, as otherwise action sell_shares is not available.

ollybh:
> EBUY_SELL_MORE_THAN_NEEDED_SETS_PURCHASE_MIN should be false, though, as in 1824 you can sell shares for more than min to buy a more expensive train, and all available trains in depot should be available as choices.
> 
> I do not think the code support this case fully, so therefor this PR implements checks to stop illegal selling.

Have you tried this? My understanding is that this constant is for when there are different trains available. Setting it to true stops the company from issuing enough shares to be able to afford an expensive train, but then buying a train it could have afforded by issuing fewer shares.

perwestling:
> > EBUY_SELL_MORE_THAN_NEEDED_SETS_PURCHASE_MIN should be false, though, as in 1824 you can sell shares for more than min to buy a more expensive train, and all available trains in depot should be available as choices.
> > I do not think the code support this case fully, so therefor this PR implements checks to stop illegal selling.
> 
> Have you tried this? My understanding is that this constant is for when there are different trains available. Setting it to true stops the company from issuing enough shares to be able to afford an expensive train, but then buying a train it could have afforded by issuing fewer shares.

You are right. I mistook behavior for some constants based on their names. With corrected values it now behaves properly.

But the behavior to not allow cross train purchase when having sold shares - that I could not find in the code.
So I am keeping that check.

So is this a bug in other titles?

ollybh:
The eighth bullet point in section VII.12 says that trains can be bought from another company, and the ninth bullet point says that they cannot be bought for more than face value. I do not see anything that says a train cannot be bought across if shares have been sold.

ollybh:
If you do need to check whether shares have been sold, you can test the `@last_share_sold_price` variable (from Step::Train). You do not need to track this yourself.

perwestling:
> The eighth bullet point in section VII.12 says that trains can be bought from another company, and the ninth bullet point says that they cannot be bought for more than face value. I do not see anything that says a train cannot be bought across if shares have been sold.

It is an interesting case that maybe the rules do not forbid explicitly, but knowing 18xx in general it is hard to see that it would actually be intentional to allow it.

For example, the rules states (last bullet VII.12) that if director assist with their own funds, the corporation may not have any funds left. So that means it would break the rule from VII.11 that minimum price is 1.

Allowing this means that you can shuffle money from director to another corporation.

I think I will investigate how it is in other 18xx titles.

ollybh:
There are definitely 18xx titles where you can sell shares in EMR and then buy a train across. 1830 is one. It wouldn't violate the rule about now having any money in the company after EMR, any left over money would be with the player, not the company.

perwestling:
You are right - it seems that is allowed. So I removed that check from the implementation.

But there is one minor fault in the implementation.

You can sell shares up to max depot train, and then buy a cheap one from another corporation.
So I suppose the GUI should do like it does for depot train - remove trains that break the rules of not selling more than needed.

Possibly a minor thing is also cross-buy between your own corporations, then all available cash will be used (up to face), although the rules probably allow that you charge a little bit less.

Anyway, these are probably not worth the effort, as the first is not of interest for a player to sell extra shares and then send money to another player. At least not in games where you cannot dump a corporation buy selling.
And the 2nd is a very minor corner case.

OK, implementation is complete, and I don't think this is worth unit testing (as most is about the GUI).

ollybh:
You might be able to handle this case by overriding `spend_minmax`.

perwestling:
> You might be able to handle this case by overriding `spend_minmax`.

It is a "fault" in the general code, I would say.

I think this is good enough.

perwestling:
> If this doesn't fully implement the cross-buying rules, can you make a note of this on the [wiki](https://github.com/tobymao/18xx/wiki/1824#implementation-notes)? And maybe add a TODO comment in the code too.

I have now updated the wiki page. I think that is enough.

@ollybh I think you can improve this now.

PR #12430, opened 2026-03-15
[18Cuba] Tile laying yellow

chrstas:
This PR implements the first part (yellow) of the track-laying rules for minor and major companies in 18Cuba, focusing on which tiles can be laid in which hexes. It enforces restrictions on tile types, track gauge, and hex locations, aligning with the game rules.

**Minors**
May lay only yellow city tiles with narrow gauge track (:narrow) except for their home hex, where they may lay the sugar mill tiles with broad and narrow gauge tracks

**Majors**
May lay only yellow city tiles with standard gauge track (:broad). City tiles in white-circle cities must use only one city tiles.

Open todos for next PRs:
- "none left" tiles which are illegal should not be shown in the UI. Struggling a bit with this - any hint welcome...
- Upgrade logic to be checked
- FC logic with fees

## Before clicking "Create"

## Implementation Notes

`tracker_available_hex` prevents minors from laying tiles in non-home hex cities.
`potential_tiles filters` out illegal tiles for both minor and major companies:
`minor_tile_blocked?` prevents minors from laying disallowed track types and non-home city tiles.
`major_tile_blocked?` prevents majors from laying yellow tiles with incorrect track type.
Helper `tile_has_only_track_type?` checks if a tile contains only a specific track type.

### DIFF
--- a/lib/engine/game/g_18_cuba/step/track.rb
+++ b/lib/engine/game/g_18_cuba/step/track.rb
@@ -10,6 +10,7 @@ module Engine
         class Track < Engine::Step::Track
           def tracker_available_hex(entity, hex)
             # TODO: FC logic with fee payments
+            # TODO: upgrade logic ensure narrow track connect to narrow track and broad to broad
             corp = entity.corporation? ? entity : @game.current_entity
 
             # 18Cuba: minors cannot build cities except on their home hex
@@ -21,11 +22,10 @@ module Engine
           end
 
           def potential_tiles(entity, hex)
-            # TODO: upgrade logic to be checked
             corp = entity.corporation? ? entity : @game.current_entity
 
             super.reject do |tile|
-              @game.tile_blocked_for_corp?(tile, corp)
+              @game.tile_blocked_for_corp?(tile, corp, hex)
             end
           end
         end
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -247,15 +247,16 @@ module Engine
           corp = selected_company || @round&.current_entity&.corporation
 
           super.reject do |t|
-            tile_blocked_for_corp?(t, corp, tile.hex)
+            # Hex not available for selector, therefore passing nil and ignoring home hex check for minors
+            tile_blocked_for_corp?(t, corp, nil, for_selector: true)
           end
         end
 
-        def tile_blocked_for_corp?(tile, corp)
+        def tile_blocked_for_corp?(tile, corp, hex, for_selector: false)
           return false unless corp
 
           if corp.type == :minor
-            minor_tile_blocked?(tile, corp.tokens.first.hex)
+            minor_tile_blocked?(tile, corp.tokens.first.hex, hex, for_selector: for_selector)
           else
             major_tile_blocked?(tile)
           end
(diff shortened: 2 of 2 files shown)


lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -8,7 +8,58 @@ module Game
...
       module Step
         class Track < Engine::Step::Track
-          # Implementation in next PR, will work in parallel on that
+          def tracker_available_hex(entity, hex)
+            # TODO: FC logic with fee payments
+            corp = entity.corporation? ? entity : @game.current_entity
+
+            # 18Cuba: minors cannot build cities except on their home hex
+            return nil if corp.type == :minor &&
+                          hex.tile.cities.any? &&
```
ollybh: For efficiency you should (almost) never call `Array.any?` or `Array.none?` unless you are passing a block to the method – those methods create an iterator to examine the contents of the array. To test whether there are any members in an array use `Array.empty?` instead.
```suggestion
                          !hex.tile.cities.empty? &&
```

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -8,7 +8,58 @@ module Game
...
+
+            tiles.reject! do |tile|
+              if corp.type == :minor
+                minor_tile_blocked?(tile, hex, home_hex, track_to_reject)
+              else
+                major_tile_blocked?(tile)
+              end
+            end
+
+            tiles
```
ollybh: The code could be made much more concise by having fewer intermediate variables.
```suggestion

            super.reject do |tile|
              if corp.type == :minor
                minor_tile_blocked?(tile, corp.tokens.first.hex)
              else
                major_tile_blocked?(tile)
              end
            end
```
You've got the broad track type hard coded within `major_tile_blocked?`, I think it would be better to do the same for narrow gauge track and `minor_tile_blocked?`.

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -8,7 +8,58 @@ module Game
...
+          end
+
```
ollybh: It would be better to have internal methods declared as private.
```suggestion

      private
      
```
Note that the indentation on this suggestion is probably wrong, the Github interface makes it difficult to manage whitespace.

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -8,7 +8,58 @@ module Game
...
+            tile.paths.all? { |path| path.track == track_type }
+          end
+
+          def minor_tile_blocked?(tile, hex, home_hex, track_to_reject)
+            # Returns true if a tile is illegal for a minor:
+            # home hex forbids pure rejected track (broad for minors),
+            # other hexes also forbid city tiles
+            return tile_has_only_track_type?(tile, track_to_reject) if hex == home_hex
+
+            tile.cities.any? ||
```
ollybh:
```suggestion
            !tile.cities.empty? ||
```

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -8,7 +8,58 @@ module Game
...
+            return nil if corp.type == :minor &&
+                          hex.tile.cities.any? &&
+                          hex != corp.tokens.first.hex
+
+            super
+          end
+
+          def potential_tiles(entity, hex)
+            # TODO: upgrade logic to be checked
+            # TODO: "none left" illegal tiles still showed in the UI, but should be hidden
```
ollybh: The tiles shown in the tile selector come from `Game.all_potential_upgrades`. You'll need to override that method to suppress these tiles.

ollybh (changes requested):
This is broadly OK, but I've got a few suggestions for code style changes.

chrstas:
thanks @ollybh... I have made the adjustments and included all_potential_tiles for the selector in the game class. hope I did not overlook anything.

PR #12435, opened 2026-03-18
[1849] corrects corp order calculation when new corp is started

### LINKED ISSUE #12421: [1849] Unexpected change in stock market order
Newly parred companies are placed on top of other companies in the same position on the market, at least those that entered that position earlier in the same turn due to selling stock.

E,g, I sell stock in company A, which ends in the L.100 position of the stock market; then sell stock in company B, which also ends in the L.100 position of the stock market; then par company C at the L.100 position.

The game positions the companies in the L.100 position in the order A->B->C, with C on top.

The game rearranges the order of the companies based on the previous positions of companies A and B - which is correct, but also puts company C on top, which is incorrect as far as I can tell.

See this behavior at this point in the game: https://18xx.games/game/244224?action=320

This seems to be a copy of an earlier issue that is closed https://github.com/tobymao/18xx/issues/7937

philcampeau:
Fixes #12421

## Before clicking "Create"

### Explanation of Change

There was an edge case not accounted for in the initial code for 1849. I've included a screenshot of the 1849 rules for selling stock, as they have a specific rule in section 3 and the paragraph below, which is slightly unusual for 18xx.

Currently, the code in place accommodates those rules perfectly when it comes to selling shares. Where it doesn't work is when shares are sold, a new corp is started, and the price of the sold shares lands in the same price box as that of a newly started corp. My understanding of the rule as written would be that you perform all the sales, adjust the order of the corporations that were just sold which ended up in the same price box, and only _then_ would you add the newly started corp's token to the bottom of that stack of tokens. As it's coded now, the newly started corp falls into the "unmoved" triage and ends up on top of the moved price markers. 

This new code corrects that, to put the newly started corporation at the bottom of the stack.

***NOTE*** some games will need to be pinned, including the one in the linked report. This will affect both 1849 and 1849 Kingdom Of Two Sicilies.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1849/game.rb
+++ b/lib/engine/game/g_1849/game.rb
@@ -605,9 +605,13 @@ module Engine
 
           same_spot.each do |sp, corps|
             current_order = corps.sort
-            moved, unmoved = current_order.partition { |c| just_moved.include?(c) }
+
+            newly_floated, others = current_order.partition { |c| !c.operated? }
+            moved, unmoved = others.partition { |c| just_moved.include?(c) }
+
             moved_ordered = moved.sort_by { |c| old_operating_order.index(c) }
-            new_order = unmoved + moved_ordered
+
+            new_order = unmoved + moved_ordered + newly_floated
             next if current_order == new_order
 
             @log << 'Updating operating order for sold (and moved) corporations now


lib/engine/game/g_1849/game.rb
```diff
@@ -605,9 +605,13 @@ def reorder_corps
 
           same_spot.each do |sp, corps|
             current_order = corps.sort
-            moved, unmoved = current_order.partition { |c| just_moved.include?(c) }
+
+            newly_floated, others = current_order.partition { |c| !c.operated? }
+            moved, unmoved = others.partition { |c| just_moved.include?(c) }
+
             moved_ordered = moved.sort_by { |c| old_operating_order.index(c) }
```
ollybh, line 612: Is the line that is causing the problem? The companies whose shares have been sold and the newly floated company will all be in `just_moved` but the newly floated company wasn't operating when `old_operating_order` was calculated. Would this work?
```
moved_ordered = moved.sort_by { |c| [c.operated? 0 : 1, old_operating_order.index(c)] }
```

michaeljb:
Can the `reorder_corps` check just be moved to happen at the end of `process_sell_shares`?

Took me a while to parse out why floating a corp after selling should have anything to do with what happens to the sold companies in a sell-buy game. It's weird to have the movement fixed at the end of the SR turn instead of after the sale, and it's what caused this bug I think.

michaeljb (changes requested):
> Can the `reorder_corps` check just be moved to happen at the end of `process_sell_shares`?
> 
> Took me a while to parse out why floating a corp after selling should have anything to do with what happens to the sold companies in a sell-buy game.

I tried the approach of moving `reorder_corps` and it looks good, see PR #12537. Do you know of any reasons that would not work?

ollybh:
Superseded by #12537.

PR #12436, opened 2026-03-18
[core] add "webhook user" checks to Create Account page

### LINKED ISSUE #12299: [Profile] Misleading error message if `webhook_user_id` is empty
**What's your Game ID?**

This doesn't apply to any game.

**How to Reproduce the Bug?**

- register for a new account
- select option for a webhook notification, select custom
- https://18xx.games/ opens and agreeing on privacy policy is not possible
- this error is visible:

Image

**Which Browser / Platform?**

Firefox and Brave (app version 5d29cd3a4).

philcampeau:
Fixes #12299

## Before clicking "Create"

### Explanation of Change

Currently, if a user selects the webhook option during account creation and either leaves the Webhook User ID field blank or puts a space in the user name, they're allowed to click Create Account, but when trying to click the "I agree to the privacy policy" button on the next page, they get the error shown in the screenshot below.

The problem is, they have no way of going back to the previous page to correct the issue, and the account is already created, so they're forced to delete the account and restart from scratch.

I updated the user creation flow to use `User.new`,  `user.update_settings(r.params)` ,  and `user.save` instead of `User.create(params)` to ensure the webhook-related fields are populated before validation runs. Then I added webhook user ID validation to User#validate

This checks that webhook_user_id is not blank and does not contain spaces when webhook is enabled.

I also made sure update_settings was resilient by initializing settings when nil using `self.settings ||= {}`

Now, Webhook validation now runs during Create Account instead of only after accepting the privacy policy, and has separate errors depending if the field is blank or has spaces in it.

As a final change, I took out the mention to "look the wiki" from the error message, because there's nothing in the wiki about spaces in Webhook names.

### Screenshots

current error prior to fix: 

image

### DIFF
--- a/routes/user.rb
+++ b/routes/user.rb
@@ -17,7 +17,7 @@ class Api
         # POST '/api/user/'
         r.is do
           user = User.new
-          user.password = r.params['password'] if r.params['password'].present?
+          user.password = r.params['password'] unless r.params['password']&.strip&.empty?
           user.update_settings(r.params)
           user.save
 
--- a/routes/user.rb
+++ b/routes/user.rb
@@ -17,7 +17,7 @@ class Api
         # POST '/api/user/'
         r.is do
           user = User.new
-          user.password = r.params['password']
+          user.password = r.params['password'] if r.params['password'].present?
           user.update_settings(r.params)
           user.save
 
--- a/routes/user.rb
+++ b/routes/user.rb
@@ -16,19 +16,12 @@ class Api
 
         # POST '/api/user/'
         r.is do
-          params = {
-            name: r.params['name']&.strip,
-            email: r.params['email'],
-            password: r.params['password'],
-            settings: {
-              notifications: r.params['notifications'],
-              webhook: r.params['webhook'],
-              webhook_url: r.params['webhook_url'],
-              webhook_user_id: r.params['webhook_user_id'],
-            },
-          }.reject { |_, v| v.empty? }
+          user = User.new
+          user.password = r.params['password']
+          user.update_settings(r.params)
+          user.save
 
-          login_user(User.create(params))
+          login_user(user)
         end
 
         # POST '/api/user/forgot'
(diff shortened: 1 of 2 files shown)


routes/user.rb
```diff
@@ -16,19 +16,12 @@ class Api
...
-              webhook: r.params['webhook'],
-              webhook_url: r.params['webhook_url'],
-              webhook_user_id: r.params['webhook_user_id'],
-            },
-          }.reject { |_, v| v.empty? }
-
-          login_user(User.create(params))
+          user = User.new
+          user.password = r.params['password']
+          user.update_settings(r.params)
```
ollybh, line 21: This has removed the `.reject { |_, v| v.empty? }` call on the `params` hash. Was that important? Could we end up with blank usernames or passwords?

philcampeau: Thanks for catching that. I had meant to use `user.password = r.params['password'] if r.params['password']`, which is the same format that update_settings already uses to guard the username and email:

https://github.com/tobymao/18xx/blob/master/models/user.rb#L26-L28

I pushed a commit to add that guard here

ollybh: That will guard against missing parameters, but what about zero-length strings? That's what the `params.reject { |_, v| v.empty? }` call does.

routes/user.rb
```diff
@@ -17,7 +17,7 @@ class Api
         # POST '/api/user/'
         r.is do
           user = User.new
-          user.password = r.params['password']
+          user.password = r.params['password'] if r.params['password'].present?
```
ollybh: What is `.present?`?

philcampeau: Oops that's a ruby on rails thing. Fixed.  Changed to `user.password = r.params['password'] unless r.params['password']&.strip&.empty?`

routes/user.rb
michaeljb: Just tested with empty input fields and it all behaved correctly

michaeljb (approved):

PR #12437, opened 2026-03-18
[System18] Charters shouldn't count agaist cert limit

### LINKED ISSUE #12122: [System18: Britain] Charter counts towards certificate limit
In a game where I was on cert limit (12) during the previous op rounds, and then won the auction for the SPX charter. I now find myself one over limit - but charters aren't supposed to count! (They are not shares.)

https://18xx.games/game/224598?action=467

Never showed up before, because one usually acquire charters when below cert limit, but here the game has stalled with most players on (or near) limit.

philcampeau:
Fixes #12122

## Before clicking "Create"

### Explanation of Change

I just added the CERT_LIMIT_INCLUDES_PRIVATES = false constant to the relevant games, and a `true` one to the game.rb. 

I thought about using a redef_const in `def setup` checking for games with the CharterAuction step, but decided this might not be ideal since it's possible future games might want a charter auction where the charters do count. 

I included the `pins` tag, but the only situation I can think of where it might come up is when a player is forced to sell down due to being at cert limit, and I'm not sure that would actually break the game if their cert count were reduced.

### DIFF
--- a/lib/engine/game/g_system18/game.rb
+++ b/lib/engine/game/g_system18/game.rb
@@ -295,6 +295,7 @@ module Engine
         TILE_UPGRADES_MUST_USE_MAX_EXITS = [].freeze
         DISCARDED_TRAINS = :remove
         REMOVE_UNUSED_RESERVATIONS = false
+        CERT_LIMIT_INCLUDES_PRIVATES = true
 
         def find_map_name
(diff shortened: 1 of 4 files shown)


roseundy (approved):
Looks good.

PR #12440, opened 2026-03-19
[core] Deprecate can_act

### LINKED ISSUE #7900: Deprecate can_act?
On line 66 of `lib/engine/round/base.rb`, it says that `can_act?` is deprecated. But it is being called a few times elsewhere.

Should these be replaced by something else?

Line 44 of `assets/app/view/game/entity_list.rb` is underlining `@acting_entity == entity || @round.can_act?(entity)`. I want to underline a specific corp on `context_entities` that isn't currently active (it's just a context), so I'm passing it as the `@acting_entity`. But `can_act?` is also underlining the one that is currently acting and that doesn't make any sense.

Can I remove it from there? `EntityList` should be more general than that (maybe `@acting_entity` needs to be a list)

philcampeau:
Fixes #7900

## Before clicking "Create"

### Explanation of Change

`can_act?` has been listed as deprecated on the site for about 5 years. The only place in the code it was still used was in Rolling Stock. 

I removed the method, and adjusted the relevant View pages to perform the same action. In alternate_corporations.rb, I removed it entirely from render_independent_mine and render_public_mine, since those entries appear to have been put in for 1873, but that game no longer referenced can_act? anywhere, and it didn't seem to actually use it anywhere from what I could see in a few games on the site.

### DIFF
--- a/assets/app/view/game/alternate_corporations.rb
+++ b/assets/app/view/game/alternate_corporations.rb
@@ -23,8 +23,6 @@ module View
           card_style[:border] = '1px dashed'
         end
 
-        card_style[:border] = '4px solid' if @game.round.can_act?(@corporation)
-
         if selected?
           card_style[:backgroundColor] = 'lightblue'
           card_style[:color] = 'black'
@@ -60,8 +58,6 @@ module View
           card_style[:backgroundColor] = convert_hex_to_rgba(color_for(:bg2), factor)
           card_style[:border] = '1px dashed'
         end
-
-        card_style[:border] = '4px solid' if @game.round.can_act?(@corporation)
 
         if selected?
           card_style[:backgroundColor] = 'lightblue'
@@ -316,7 +312,7 @@ module View
           card_style[:border] = '1px dashed'
         end
 
-        card_style[:border] = '4px solid' if @game.round.can_act?(@corporation)
+        card_style[:border] = '4px solid' if @game.round.active_step&.current_entity == @corporation
 
         if selected?
           card_style[:backgroundColor] = 'lightblue'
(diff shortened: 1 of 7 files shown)


roseundy:
Please make sure this doesn't break games where multiple entities are allowed to make a move. Rolling Stock for sure, and I believe in the 1841 variant that allows player-to-player stock sales (first edition rules).

philcampeau:
@roseundy I think the only part of 1841 that would be simultaneous would be the first edition auction rules, and it looks like the way you implemented it was to have each player in turn order submit their bids for every private blindly, but still in turn order.

roseundy:
For 1841, I was specifically refering to the buy-stock-directly-from-a-player variation. I could be wrong, but I remember thinking when I implemented that it was a good thing I had already implemented the Rolling Stock Acquisition Phase (which allows multiple players to make a move).

philcampeau:
@roseundy Why would that work simultaneously? a bunch of players all buying shares from each other like a Black Friday free-for-all at a Walmart?

roseundy:
Like I said, I'm not sure it applies to 1841, but yes, the aquisition phase in Rolling Stock really is a free-for-all. Any player can propose a company purchase at any point in the phase. Multiple players can be "on turn".

philcampeau:
I've triple checked now, 1841 does not have a simultaneous action phase.

roseundy:
OK, but does this break Rolling Stock?

philcampeau:
@roseundy I'm not familiar with RS or RSS directly, but I imported a few completed RSS games and they loaded fine locally. I also imported an active RS game, rolled back to a simultaneous step (Acquisition), logged into 3 different accounts, and all 3 appeared active: 

image

Issue #12446, opened 2026-03-23
[Meta] Regularize GAME_LOCATION for various games


sunfall:
This is a metanit for making the values of GAME_LOCATION more consistent across the various games, in an attempt to make it easier to do this piecemeal.

Why?  Per the original pull request:

The use of "US" and "USA" was not particularly consistent; the same goes for "East X" versus "Eastern X" and similar derivations. When a location was a city or region, this generally now includes the country.

Below are some specific notes (updated from the original PR):

- At the time of The Old Prince 1871, PEI was not yet actually part of Canada, so it's not changing.
- "North East England" has a special meaning as one of the nine regions of England, and 1812 covers more area than that.  In the original I suggested "Northeastern", but apparently that's an Americanism, so I'll be going with "North Eastern England."
- "Netherlands" vs. "The Netherlands" is pretty arbitrary and different games use one or the other.  The official name of the country in English is just "Netherlands," but they use "the Netherlands" just about everywhere too, so I honestly don't know which would be better.  I'll, uh, submit those changes last in the hopes that someone pipes up.

sunfall:
Stopping at four for the time being; if breaking this out into a metanit and filing separate PRs ends up being the wrong thing to do, I don't want to spam the database (or the developers).

sunfall:
That's all of the ones that don't deal with [the] Netherlands.  Guidance on preferences there would be appreciated.

My argument for eliding the 'The' is not just brevity, but to match both 'United States [of America]' and 'United Kingdom' as well; those all colloquially have 'the' added to the front of them when discussing them in conversation, but not as a standalone concept-of-a-nation, and both of those already exist in the current corpus of GAME_LOCATIONs.

sunfall:
I pulled the trigger on 'The Netherlands' -> 'Netherlands' because I'm pretty convinced it's the right choice, and with that I believe I've done all the regularization I personally saw needed doing.  If anyone has any further requests, though, I'd be happy to hear them and shepherd them as well.

sunfall:
Looks like this is done!  Thanks for the discussion and the work in getting this committed.

PR #12457, opened 2026-03-25
[18Dixie] Regularize GAME_LOCATION

sunfall:
Part of the work on #12446.

### DIFF
--- a/lib/engine/game/g_18_dixie/meta.rb
+++ b/lib/engine/game/g_18_dixie/meta.rb
@@ -13,7 +13,7 @@ module Engine
         GAME_SUBTITLE = 'The Railroads Come to the Deep South'
         GAME_DESIGNER = 'Mark Derrick'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Dixie'
-        GAME_LOCATION = 'USA South'
+        GAME_LOCATION = 'Southern USA'
         GAME_PUBLISHER = :golden_spike
         GAME_RULES_URL = 'https://18xx.games' # TODO: Make a stable rules link
 


lib/engine/game/g_18_dixie/meta.rb
```diff
@@ -13,7 +13,7 @@ module Meta
         GAME_SUBTITLE = 'The Railroads Come to the Deep South'
         GAME_DESIGNER = 'Mark Derrick'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Dixie'
-        GAME_LOCATION = 'USA South'
+        GAME_LOCATION = 'Southern USA'
```
perwestling, line 16: This is interesting as I had a discussion on Facebook if California was part of southern USA. My view was that The South and Southern are different things. The South is a concept with a meaning, while Southern is a geographical concept - ie the states that are in the southernmost part of USA.

But if native speakers of English do want to use Southern USA for this region I am fine with that.

perwestling: And of course, it is located it Southern USA regardless, as not all parts of Southern USA need to be included in a game to have it as a location.

perwestling (approved):

lib/engine/game/g_18_dixie/meta.rb
sunfall: Yes, this sort of thing is dang-near philosophical.  I would never conceptualize California as either "southern" or "the South," and it is _definitely_ not the latter, but it certainly actually is the former.  Definitionally.

This game is both set in "the South" and in the "southern USA."  I will note that "the South" has a bunch of political meanings and nuance (and the name of the game, 18Dixie, brings up even more of those political meanings) that one could compose a treatise on.  Because of that, I personally think the boring-but-factual "Southern USA" is better both for consistency with all the other regions and neutrality.

PR #12458, opened 2026-03-25
[18Hiawatha] GAME_LOCATION to 'Midwestern USA'

sunfall:
Part of the work on #12446.

### DIFF
--- a/lib/engine/game/g_18_hiawatha/meta.rb
+++ b/lib/engine/game/g_18_hiawatha/meta.rb
@@ -13,7 +13,7 @@ module Engine
 
         GAME_DESIGNER = 'Michael Carter, Anthony Fryer, & Nick Neylon'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Hiawatha'
-        GAME_LOCATION = 'Midwest USA'
+        GAME_LOCATION = 'Midwestern USA'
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_hiawatha/meta.rb
```diff
@@ -13,7 +13,7 @@ module Meta
 
         GAME_DESIGNER = 'Michael Carter, Anthony Fryer, & Nick Neylon'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18Hiawatha'
-        GAME_LOCATION = 'Midwest USA'
+        GAME_LOCATION = 'Midwestern USA'
```
perwestling, line 16: Should it be United States of America here as well? Just wondering as that was the change in another PR.

sunfall: I don't (personally) think so.  I would actually be fine collapsing 'United States of America' to just 'USA' for 18USA, honestly, but I think that might look both odd amongst all the longer GAME_LOCATIONs and at odds with the extant 'United Kingdom'.  For regions of the US, I feel like the abbreviation is fine.  (And, yes, I noticed that my natural use of American English there wants 'US' and not 'USA.'  Language is hard.)

perwestling (approved):

PR #12461, opened 2026-03-25
[18USA] Change GAME_LOCATION to 'Continental USA'

sunfall:
Part of the work on #12446.  It's a bit weird to actually lengthen one of these, but given that part of the regularization is making them all say "USA," not having this match would be even more weird, I believe.

### DIFF
--- a/lib/engine/game/g_18_usa/meta.rb
+++ b/lib/engine/game/g_18_usa/meta.rb
@@ -13,7 +13,7 @@ module Engine
 
         GAME_DESIGNER = 'Edward Reece, Mark Hendrickson, and Shawn Fox'
         GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18USA'
-        GAME_LOCATION = 'United States'
+        GAME_LOCATION = 'United States of America'
         GAME_PUBLISHER = :all_aboard_games
         GAME_RULES_URL = {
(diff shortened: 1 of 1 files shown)


perwestling (approved):

sunfall:
I'm going to do a possibly dumb thing here and argue not-so-briefly against my own PR.

I would say that part of GAME_LOCATION's purpose is to provide additional location-specific context when you're picking a game.  The you-who-is-picking sees a long list of numbered games that mean essentially nothing: 1846? 1850? 1860?  But next to 1846 you can see that it's set in the Midwest, so that makes the abstract at least a little more concrete.

Changing 18USA to just 'USA' is the sort of thing that should prompt a wry look at the camera, _The Office_-style.  And, sure, this expands it to "United States of America" for consistency's sake--it's weird to have a game literally called USA say that it's just set in the US--but pretty much literally anyone who can read English knows that 'USA' and 'United States of America' mean the same thing.  We're just using more letters to spell it.

So: should this possibly be something like 'Continental USA'?  Because even though it's called 18USA, you rail-ify neither Alaska nor Hawai'i (let's... ignore for the moment that neither of these were states at the presumed time this game is taking place), and at least that gives _some_ context as to where the heck the game is taking place.

I'm almost certainly overthinking this.

perwestling:
I think Continental USA sounds like a good idea.

sunfall:
Agreed, obviously, and done.

PR #12467, opened 2026-03-27
Roling Stock: fix multiple receivership acquisitions

### LINKED ISSUE #12466: [Rolling Stock Stars] Receivership corps only buy 1 company from FI
Game ID: 243592
Action ID: 202
https://18xx.games/game/243592?action=202

DA, a corp in receivership, enters the Turn 6 Acquisition phase with $47 in cash. The Foreign Investor has two companies: OL (FV: $15, high price $20) and HE (FV: $14, high price $18).

DA correctly attempts to buy the highest-value FI company first (OL), which OS declines to preempt. The purchase goes through, and DA now owns OL with $27 in cash remaining.

The problem is that DA still has enough cash ($27) to *also* buy HE from the FI at the high price ($18). Per the [game rules](https://cdn.shopify.com/s/files/1/0252/9371/7588/files/RSS.pdf?v=1597256869) for receivership corp behavior (emphasis added):

> Phase 3: Corporations in receivership are very eager to buy from the foreign investor. However, in most cases, they don’t have enough cash. In practical terms, you should check right at the beginning of phase 3 if any corporation in receivership is able to buy a company from the foreign investor at all. If that’s the case, start with the most expensive eligible company and pretend that the receiver has just announced the intention to buy the company. Then follow the usual rules of intervention and priorities (including the special ability of the Overseas Trading corporation). **Repeat with the less expensive companies until no companies are eligible anymore.**

Basically - instead of a single trigger to try and buy an FI company when entering the Acquisition phase, receivership corps should loop and continue making offers as long as they're eligible.

Image

roseundy:
Fixes #12466 

It's possible that it will require pins. There are a couple dozen active Rolling Stock and Rolling Stock Starts games. Not sure any of them have a receivership corporation that is in a position to buy multiple companies from the Private Investor.

## Before clicking "Create"

### Explanation of Change

Latent bug. Receivership  step `receiver_propose_and_purchase.rb`  overrides the standard `propose_and_purchase.rb` and has an overriding method called `process_response`, but the base method is called `process_respond`. This fixes the typo.

### DIFF
--- a/lib/engine/game/g_rolling_stock/step/receiver_propose_and_purchase.rb
+++ b/lib/engine/game/g_rolling_stock/step/receiver_propose_and_purchase.rb
@@ -44,7 +44,7 @@ module Engine
             add_next_receiver_offer
           end
 
-          def process_response(action)
+          def process_respond(action)
             super
 
             add_next_receiver_offer if receiver_offers.empty?


roseundy:
Apparently rubocop doesn't check for a `super` in a method that doesn't have an overidden method.

ollybh (approved):

PR #12471, opened 2026-03-30
[18ESP] skip token swap check if merging corp is out of tokens

### LINKED ISSUE #11690: [18ESP] player is asked whether to replace minor token or leave token on the charter even though company has no token
In this game https://18xx.games/game/204098?action=518 GSSR does not have a token anymore, so the minor token will be removed, no matter what. The "choose" step could be skipped. If "map" is click, the engine throws an exception "undefined method 'place' for nil".

philcampeau:
Fixes #11690 

## Before clicking "Create"

### Explanation of Change

I cleaned up the can_swap? method and now it skips the manual check if the acquiring corp doesn't have a token left to swap.

### DIFF
--- a/lib/engine/game/g_18_esp/step/acquire.rb
+++ b/lib/engine/game/g_18_esp/step/acquire.rb
@@ -90,10 +90,9 @@ module Engine
           end
 
           def can_swap?
-            return merged_token_in_shared_city? unless mz?(@merging.last)
+            return false unless @merging.first.next_token
+            return false if mz?(@merging.last)
 
-            @merging.first.next_token &&
-            !mz?(@merging.last) &&
             merged_token_in_shared_city?
           end
 
--- a/lib/engine/game/g_18_esp/step/acquire.rb
+++ b/lib/engine/game/g_18_esp/step/acquire.rb
@@ -16,7 +16,11 @@ module Engine
           end
 
           def auto_actions(entity)
-            return super if @merging
+            if @merging
+              return [Engine::Action::Choose.new(entity, choice: 'charter')] unless can_swap?
+
+              return super
+            end
             return [Engine::Action::Pass.new(entity)] unless can_merge?(entity)
 
             super
--- a/lib/engine/game/g_18_esp/step/acquire.rb
+++ b/lib/engine/game/g_18_esp/step/acquire.rb
@@ -16,14 +16,13 @@ module Engine
           end
 
           def auto_actions(entity)
-            if @merging
-              return [Engine::Action::Choose.new(entity, choice: 'charter')] unless can_swap?
-
-              return super
+            if @merging && !can_swap?
+              [Engine::Action::Choose.new(entity, choice: 'charter')]
+            elsif !@merging && !can_merge?(entity)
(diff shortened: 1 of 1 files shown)


ollybh (commented):
I can't see how this is going to fix the issue. `can_swap?` is clearer following your rewrite, but this isn't going to stop `choose` being returned as the action during the merger process. Is it?

ollybh:
@philcampeau ??

philcampeau:
@ollybh thanks for the nudge! 

Yes you're right. I had written a fix for this when you first commented, I must have stashed it for one reason or another and forgot to commit it. 

I've pushed the commit now, can you take a look?

lib/engine/game/g_18_esp/step/acquire.rb
```diff
@@ -16,7 +16,11 @@ def actions(entity)
           end
 
           def auto_actions(entity)
-            return super if @merging
+            if @merging
+              return [Engine::Action::Choose.new(entity, choice: 'charter')] unless can_swap?
+
+              return super
+            end
             return [Engine::Action::Pass.new(entity)] unless can_merge?(entity)
```
ollybh: I found the code structure with branches and a postfix conditional a little hard to read. Would this work the same?
```suggestion
            return [Engine::Action::Choose.new(entity, choice: 'charter')] if @merging && !can_swap?
            return [Engine::Action::Pass.new(entity)] unless can_merge?(entity)
```

ollybh: No, it wouldn't. Hmmm.
```suggestion
            return [Engine::Action::Choose.new(entity, choice: 'charter')] if @merging && !can_swap?
            return [Engine::Action::Pass.new(entity)] if !@merging && can_merge?(entity)
```

ollybh: That should work, but is equally unreadable. Third time lucky:
```suggestion
            if @merging && !can_swap?
              return [Engine::Action::Choose.new(entity, choice: 'charter')]
            elsif !@merging && can_merge?(entity)
              return [Engine::Action::Pass.new(entity)]
            else
              super
            end
```

philcampeau: Shouldn't the elsif be `elsif !@merging && !can_merge?(entity)`, or am I missing something?

ollybh: Yes, it should. And the branches shouldn't have had `return` statements.
V4:
```suggestion
            if @merging && !can_swap?
              [Engine::Action::Choose.new(entity, choice: 'charter')]
            elsif !@merging && !can_merge?(entity)
              [Engine::Action::Pass.new(entity)]
            else
              super
            end
```

ollybh:
ping @philcampeau

lib/engine/game/g_18_esp/step/acquire.rb
philcampeau: Ack my bad! Submitting the fix now.

ollybh (approved):

PR #12473, opened 2026-03-31
[1850] fixes Western Land Grant ability

### LINKED ISSUE #9867: [1850] Western Land Grant allows all 3 tile lays to be used anywhere.
Currently the code for the Western Land Grand private allows all three tile lays to be used anywhere on the map. This is incorrect. One tile lay can be anywhere, the other two must be west of the Mississippi. 

There's already a class WEST_RIVER_HEXES defined at the top of the entities.rb file, which is also used for the Credit Mobilier private.

philcampeau:
Fixes #9867

## Before clicking "Create"

### Explanation of Change

The Western Land Grant private is supposed to grant 3 additional tile lays, one per OR, with one being allowed anywhere on the map, and the other two being west of the Misssissippi. Currently, the private doesn't restrict where the 3 lays can be done. 

This fixes the issue by allowing the private to ignore the hexes restriction until it has laid a tile outside of the ones listed in WEST_RIVER_HEXES (a constant that was already defined for another private). 

This will likely require pinning, since some players may have used illegal tile lays. 1850 is still in beta, so this is acceptable.


ollybh (changes requested):
This is a bit of a hack.

Here's a suggestion for an alternative implementation: leave the WLG ability hexes definition as an empty array, then set it to point to the western hexes if it is used to lay a tile in the east of the map:
```
def process_lay_tile(action)
  super
  return unless action.entity == @game.wlg_company
  return if @game.western_hex?(action.hex)

  abilities(action.entity).hexes = @game.class.WEST_RIVER_HEXES
end

philcampeau:
Why yes.... that was quite a bit simpler. Thanks @ollybh , pushing now with a couple of minor tweaks.

ollybh:
Hmm… it looks like this was the original algorithm before this PR. The only thing that has changed is there is a check for a non-nil `ability`. But could that happen? This is the special track step.

Does this work correctly? Did the original code work?

ollybh:
@philcampeau ??

philcampeau:
Bizarre. I just tested, and it works in both the master code and this change. Unsure what happened here, but I'll close this PR in that case.

PR #12474, opened 2026-03-31
[1828] Fix Erie home station lay

### LINKED ISSUE #11805: [1828] Erie should be given option to token as soon as tile is laid
**What's your Game ID?**

[Gist link](https://gist.github.com/philcampeau/eb2313c03f13e22b0f94f9f84816f01e)

Per rule 13.3.2:

Image

Erie should immediately.be given the option to place a token in hex E15 if it has floated but not yet operated and another corporation lays a green tile in that hex. Currently this doesn't happen.

philcampeau:
Fixes #11805

## Before clicking "Create"

### Explanation of Change

Per the rules (quoted below) if Erie is floated but not yet operated, Erie should be given the opportunity to lay its home station as soon as E15 is upgraded. This update seems to fix all possible issues, based on my testing.

Also, I tested against 20 completed games on the site, and all loaded fine, so I think this won't require pins.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1828/step/track.rb
+++ b/lib/engine/game/g_1828/step/track.rb
@@ -11,12 +11,28 @@ module Engine
           include AcquireVaTunnelCoalMarker
 
           def update_token!(action, entity, tile, old_tile)
-            if action.hex.id == 'E15' && (token = tile.cities.flat_map(&:tokens).find(&:itself))
-              # If there are blocking tokens in both cities, no decisions to be made
-              return if @game.blocking_token?(token)
+            if action.hex.id == 'E15'
+              # When track is first laid on E15, trigger Erie to place its home token if already floated
+              if old_tile.paths.empty? && !tile.paths.empty?
+                erie = @game.corporation_by_id('ERIE')
+                if erie && !erie.closed? && erie.floated? && !erie.tokens.first&.used &&
+                    @round.pending_tokens.none? { |p| p[:entity] == erie }
+                  @round.pending_tokens << {
+                    entity: erie,
+                    hexes: [action.hex],
+                    token: erie.find_token_by_type,
+                  }
+                  @log << "#{erie.name} must choose city for home token"
+                end
+              end
 
-              # Otherwise, the token owner gets to decide the token location
-              entity = token.corporation
+              if (token = tile.cities.flat_map(&:tokens).find(&:itself))
+                # If there are blocking tokens in both cities, no decisions to be made
+                return if @game.blocking_token?(token)
+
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1828/step/track.rb
```diff
@@ -11,12 +11,28 @@ class Track < Engine::Step::Track
...
+                    hexes: [action.hex],
+                    token: erie.find_token_by_type,
+                  }
+                  @log << "#{erie.name} must choose city for home token"
+                end
+              end
 
-              # Otherwise, the token owner gets to decide the token location
-              entity = token.corporation
+              if (token = tile.cities.flat_map(&:tokens).find(&:itself))
```
ollybh, line 29: This line confused me at first. I think this would be clearer:
```suggestion
              if (token = tile.cities.flat_map(&:tokens).compact.first)
```

lib/engine/game/g_1828/step/track.rb
```diff
@@ -11,12 +11,28 @@ class Track < Engine::Step::Track
...
+                # If there are blocking tokens in both cities, no decisions to be made
+                return if @game.blocking_token?(token)
```
ollybh, line 31: This code line doesn't match the comment. Does it? It's just testing the first token found, not tokens in both cities.

lib/engine/game/g_1828/step/track.rb
```diff
@@ -11,12 +11,28 @@ class Track < Engine::Step::Track
           include AcquireVaTunnelCoalMarker
 
           def update_token!(action, entity, tile, old_tile)
-            if action.hex.id == 'E15' && (token = tile.cities.flat_map(&:tokens).find(&:itself))
-              # If there are blocking tokens in both cities, no decisions to be made
-              return if @game.blocking_token?(token)
+            if action.hex.id == 'E15'
+              # When track is first laid on E15, trigger Erie to place its home token if already floated
+              if old_tile.paths.empty? && !tile.paths.empty?
```
ollybh, line 16: Or maybe:
```suggestion
              if old_tile.color == :yellow
```

ollybh (changes requested):

lib/engine/game/g_1828/step/track.rb
philcampeau: done

lib/engine/game/g_1828/step/track.rb
philcampeau: done

philcampeau:
Hmmm now it's failing Rake. I'm gonna have to review this and resubmit another time.

PR #12480, opened 2026-04-03
[Rolling Stock] Fix RSS receivership order for OS

### LINKED ISSUE #12472: [Rolling Stock Stars] OS in receivership can be FI preempted
Game ID: 210560
Action: 282
https://18xx.games/game/210560?action=282

The corporation OS has two special abilities, both of which only apply in the "Acquisition" phase:
1. Buys from the Foreign Investor (FI) at company face value rather than high price.
2. If buying from the FI is possible, is always treated as being first in priority for buying from the FI.

Corps in receivership always try and buy from the FI if possible. From my analysis, it looks like OS in receivership correctly gets the first power applied (see [this game/action](https://18xx.games/game/210896?action=312), for instance), but not the second - on action 282, SI (in receivership) has a share price of $50 (higher than OS's share price of $16) and purchases SJ from the FI for the high price of $41, even though OS (in receivership) has enough cash ($34) to purchase SJ from the FI at its face value of $31.

To be fair, the rules are a little ambiguous on corp special powers and receivership rules, so I asked on [BGG](https://boardgamegeek.com/thread/3684095/do-os-special-abilities-apply-when-in-receivership) and the game designer confirmed the following:

> Special abilities always apply, including during receivership and even if you don't want to apply them.

Note that the action played above is pretty old (2025-06-29), but I verified that the same bug occurs by replaying the game against the current version of the RSS engine (commit `796bd6a` on branch `master`), and confirmed that the game in question is not pinned to an earlier version. I also created another receivership bug previously at #12466 - I verified that the proposed fix does not resolve this, so I'm filing it as a separate issue. Let me know if I should close this one and just put the context over there.

Finally, just wanted to say that I super appreciate all the work that's been done for the 18xx.games RSS implementation - sorry if it feels like I'm just coming in here and demanding fixes, that's not my intent. :)

Image

iwooden:
Fixes #12472

Like #12467 , this may require pins, although it seems unlikely - I've only found one RSS game (ID 210560) that was affected by this bug.

## Before clicking "Create"

### Explanation of Change
Latent bug. In the "Acquisition" phase, we calculated Foreign Investor priority differently between corps in receivership and active corps - they should be calculated in the same way. The code change mirrors [the existing method](https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_rolling_stock/game.rb#L659) for calculating FI priority of active corps.

### DIFF
--- a/lib/engine/game/g_rolling_stock/step/receiver_propose_and_purchase.rb
+++ b/lib/engine/game/g_rolling_stock/step/receiver_propose_and_purchase.rb
@@ -92,7 +92,9 @@ module Engine
           end
 
           def elegible_receiver_and_company
-            @game.operating_order.select(&:receivership?).each do |candidate|
+            receivers = @game.operating_order.select(&:receivership?)
+            oversea_corp, others = receivers.partition { |corp| @game.abilities(corp, :overseas) }
+            (oversea_corp + others).each do |candidate|
               @game.foreign_investor.companies.sort_by(&:value).reverse_each do |company|
                 return [candidate, company] if candidate.cash >= foreign_price(candidate, company)
               end


iwooden:
Don't think I have the permissions to add labels (?), but it should have `RollingStock` and `pins`.

roseundy:
No offense, but I marked myself as owning this bug. I was waiting for the previous fix to be merged before making a 2nd fix.

iwooden:
None taken, sorry about that - feel free to close this!

roseundy (approved):

roseundy:
This looks good.

lib/engine/game/g_rolling_stock/step/receiver_propose_and_purchase.rb
```diff
@@ -92,7 +92,9 @@ def create_receiver_offer
           end
 
           def elegible_receiver_and_company
-            @game.operating_order.select(&:receivership?).each do |candidate|
+            receivers = @game.operating_order.select(&:receivership?)
+            oversea_corp, others = receivers.partition { |corp| @game.abilities(corp, :overseas) }
+            (oversea_corp + others).each do |candidate|
```
ollybh, line 97: Does this method get called much? It probably doesn't matter if this isn't a performance critical path, but this will create four intermediate arrays. Something like this might be more efficient:

```suggestion
            receivers = @game.operating_order.select(&:receivership?).sort_by { |corp| !!@game.abilities(corp, :overseas) }.each do |candidate|
```

PR #12482, opened 2026-04-04
Allows front page "Filter" to select multiple titles

lmx2020-18xx:
Titles are added as small chips below Filter section and can be individually deleted by clicking (x) on the chip. URL schema list titles separated by "." Players can bookmark preferred games as https://18xx.games/?title=1846.1889.1817 

**Claude code** output summary:
 models/game.rb (backend):
  - SQL queries now use ANY(:titles) instead of :title = title (lines 20, 36)
  - Title param is split on dot  into a Sequel.pg_array (line 94-95)
  - Single title still works (array of one), empty/nil means no filter

  assets/app/view/game_row_filters.rb (frontend):
  - Dropdown acts as a picker — selecting a title adds it to the filter and resets the dropdown
  - Selected titles appear as dark chips with × buttons to remove individually
  - Already-selected titles are hidden from the dropdown
  - URL uses dot separated format: ?title=1846.1889
  - "Reset filters" clears all selected titles
  
  
Screenshot 2026-04-04 022905

Screenshot 2026-04-04 022923

Screenshot 2026-04-04 024253

### DIFF
--- a/assets/app/lib/params.rb
+++ b/assets/app/lib/params.rb
@@ -30,8 +30,7 @@ module Lib
       def get_all(key)
         return [] if @unsupported
 
-        result = Native(`Array.from(#{@native.to_n}.getAll(#{key}))`)
-        result.to_a
+        @native.getAll(key)
       end
 
       def []=(key, value)
@@ -43,12 +42,12 @@ module Lib
       end
 
       def set_array(key, values)
-        @native&.delete(key)
-        Array(values).each { |v| `#{@native.to_n}.append(#{key}, #{v})` }
+        @native.delete(key)
+        values.each { |v| @native.append(key, v) }
       end
 
       def to_query_string
-        (@native&.toString() || '').gsub('%5B%5D', '[]')
+        @native&.toString() || ''
       end
     end
   end
--- a/assets/app/lib/params.rb
+++ b/assets/app/lib/params.rb
@@ -27,6 +27,13 @@ module Lib
         @native&.get(key)
       end
 
+      def get_all(key)
+        return [] if @unsupported
+
+        result = Native(`Array.from(#{@native.to_n}.getAll(#{key}))`)
+        result.to_a
+      end
+
       def []=(key, value)
         if !value || value.to_s.empty?
           @native&.delete(key)
@@ -35,8 +42,13 @@ module Lib
         end
       end
 
+      def set_array(key, values)
+        @native&.delete(key)
+        Array(values).each { |v| `#{@native.to_n}.append(#{key}, #{v})` }
+      end
+
       def to_query_string
-        @native&.toString() || ''
+        (@native&.toString() || '').gsub('%5B%5D', '[]')
       end
     end
   end
--- a/assets/app/view/game_row_filters.rb
+++ b/assets/app/view/game_row_filters.rb
@@ -16,15 +16,14 @@ module View
       h('div#game_filters.game_row', { key: @header }, [
         h(:h2, 'Filters'),
         title_elm,
-        render_reset(title_elm),
+        render_reset,
       ])
     end
 
     private
 
     def render_title(url_search_params)
-      raw = url_search_params['title'] || ''
-      selected_titles = raw.split('.').map { |t| t.strip.gsub('~', '.') }.reject(&:empty?)
+      selected_titles = url_search_params.get_all('title[]')
 
       # Build lookup for display names
       title_display = {}
@@ -46,7 +45,7 @@ module View
         return if title.empty?
 
         new_titles = (selected_titles + [title]).uniq
-        url_search_params['title'] = new_titles.map { |t| t.gsub('.', '~') }.join('.')
+        url_search_params.set_array('title[]', new_titles)
         update_filters(url_search_params.to_query_string)
         target.JS['value'] = ''
       end
@@ -59,7 +58,7 @@ module View
           display = title_display[t] || t
           on_remove = lambda do
             new_titles = selected_titles.reject { |x| x == t }
-            url_search_params['title'] = new_titles.empty? ? '' : new_titles.map { |t| t.gsub('.', '~') }.join('.')
+            url_search_params.set_array('title[]', new_titles)
             update_filters(url_search_params.to_query_string)
           end
 
@@ -91,7 +90,7 @@ module View
       h(:div, children)
     end
 
-    def render_reset(_title_elm)
+    def render_reset
       attrs = {
         on: {
           click: lambda do
--- a/assets/app/view/game_row_filters.rb
+++ b/assets/app/view/game_row_filters.rb
@@ -24,7 +24,7 @@ module View
 
     def render_title(url_search_params)
       raw = url_search_params['title'] || ''
-      selected_titles = raw.split('.').map(&:strip).reject(&:empty?)
+      selected_titles = raw.split('.').map { |t| t.strip.gsub('~', '.') }.reject(&:empty?)
 
       # Build lookup for display names
       title_display = {}
@@ -46,7 +46,7 @@ module View
         return if title.empty?
 
         new_titles = (selected_titles + [title]).uniq
-        url_search_params['title'] = new_titles.join('.')
+        url_search_params['title'] = new_titles.map { |t| t.gsub('.', '~') }.join('.')
         update_filters(url_search_params.to_query_string)
         target.JS['value'] = ''
       end
@@ -59,7 +59,7 @@ module View
           display = title_display[t] || t
           on_remove = lambda do
             new_titles = selected_titles.reject { |x| x == t }
-            url_search_params['title'] = new_titles.empty? ? '' : new_titles.join('.')
+            url_search_params['title'] = new_titles.empty? ? '' : new_titles.map { |t| t.gsub('.', '~') }.join('.')
             update_filters(url_search_params.to_query_string)
           end
 
--- a/assets/app/view/game_row_filters.rb
+++ b/assets/app/view/game_row_filters.rb
@@ -25,7 +25,7 @@ module View
     private
 
     def render_title(url_search_params)
-      selected_titles = url_search_params.get_all('title[]')
+      selected_titles = url_search_params.get_all('title')
 
       # Build lookup for display names
       title_display = {}
@@ -47,7 +47,7 @@ module View
         return if title.empty?
 
         new_titles = (selected_titles + [title]).uniq
-        url_search_params.set_array('title[]', new_titles)
+        url_search_params.set_array('title', new_titles)
         update_filters(url_search_params.to_query_string)
         target.JS['value'] = ''
       end
@@ -60,7 +60,7 @@ module View
           display = title_display[t] || t
           on_remove = lambda do
             new_titles = selected_titles.reject { |x| x == t }
-            url_search_params.set_array('title[]', new_titles)
+            url_search_params.set_array('title', new_titles)
             update_filters(url_search_params.to_query_string)
           end
 
@@ -93,19 +93,14 @@ module View
     end
 
     def render_mode(url_search_params)
-      current = url_search_params['mode'] || 'all'
-
-      buttons = [
-        { label: 'All', value: 'all' },
-        { label: 'Live', value: 'live' },
-        { label: 'Async', value: 'async' },
-      ].map do |opt|
-        active = current == opt[:value]
+      current = url_search_params['mode'] || ''
+      options = { 'All' => '', 'Live' => 'live', 'Async' => 'async' }
+      buttons = options.map do |label, value|
         on_click = lambda do
-          url_search_params['mode'] = opt[:value] == 'all' ? '' : opt[:value]
+          url_search_params['mode'] = value
           update_filters(url_search_params.to_query_string)
         end
-
+        active = current == value
         h(:button, {
             style: {
               padding: '4px 12px',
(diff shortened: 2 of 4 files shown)


ollybh (changes requested):
Thanks for working on this, it looks like a good improvement to the game list page.

However splitting the `title` parameter on dots isn't going to work – we have a game that includes a full stop in its name: 1828.Games. Attempting to filter by this game fails:

Image

My recollection of front-end development is probably both twenty years out-of-date and incorrect, but would it be better to repeat the `title` parameter here, something like `https://18xx.games/?title=game1&title=game2&title=game3`?

lmx2020-18xx:
I totally missed "1828.Games" on my first pass 🙈  I tried a few URL separators; main motivation is to keep URL for bookmarks short, simple and human readable. avoid showing non-technical players something like %2C in URLs. 

For small number of exceptions like 1828 just committed a fix where specific games with "." will be presented in URL as 1828~Games and converted back by param parser. 

Screenshot 2026-04-04 154236

models/game.rb
```diff
@@ -91,7 +91,12 @@ def self.home_games(user, **opts)
       }.transform_values { |v| v&.to_i || 0 }
 
       kwargs[:user_id] = user.id if user
-      kwargs[:title] = opts['title'] != '' ? opts['title'] : nil
+      kwargs[:titles] = opts['title']&.then do |t|
+        t = t.strip
+        next nil if t.empty?
+
+        t.split('.').map { |s| s.gsub('~', '.') }.join('|')
+      end
```
ollybh, line 103: Don't do this array to string to array conversion, it's not necessary. Leave it as an array and in the queries say `… AND (:titles IS NULL OR title IN :titles)`

assets/app/view/game_row_filters.rb
```diff
@@ -23,29 +23,78 @@ def render
     private
 
     def render_title(url_search_params)
-      selected_value = url_search_params['title']
-      game_options = [h(:option, { attrs: { value: '' } }, '(All titles)')]
+      raw = url_search_params['title'] || ''
+      selected_titles = raw.split('.').map { |t| t.strip.gsub('~', '.') }.reject(&:empty?)
```
ollybh: Substituting characters in the game titles is making this code more complex than it needs to be. Either choose a delimiter that won't clash with any titles (a comma?) or add a `get_all` method to UrlSeachParameters and repeat the `title` parameter for each selected title and rely on URL encoding to handle everything automatically. My preference would be for the latter, it would be simpler overall and less fragile.

assets/app/view/game_row_filters.rb
```diff
@@ -23,29 +23,78 @@ def render
...
+        children << h(:div, { style: { marginBottom: '4px' } }, chips)
+      end
+
+      children << h('select', { on: { input: on_select }, style: { maxWidth: '90vw' } }, game_options)
+
+      h(:div, children)
     end
 
-    def render_reset(title_elm)
+    def render_reset(_title_elm)
```
ollybh: This method is only called from one place. If the parameter isn't needed any more then remove it completely, here and where the method is called.

ollybh (changes requested):
This now works for 1828.Games, but I like the code less. I still think that passing multiple title parameters would be better. You would need to add some methods to `Lib::Params::URLSearchParams` to handle this, but then you can leave it to URL encoding to handle any game title translations needed. If you do stick with concatenating titles then choose a delimiter that doesn't clash with any titles, rather than this character substitution.

lmx2020-18xx:
> This now works for 1828.Games, but I like the code less. I still think that passing multiple title parameters would be better. You would need to add some methods to `Lib::Params::URLSearchParams` to handle this, but then you can leave it to URL encoding to handle any game title translations needed. If you do stick with concatenating titles then choose a delimiter that doesn't clash with any titles, rather than this character substitution.

Let me rework this along your suggestions. canonical ruby way would be `title[]=1817&title[]=1846` which makes it far less fun for human editing. my personal pref was to be able to just quickly bang out `title=1817.1846` to start the evening, but if we going to lose "human editable" aspect might as well do it the right ruby way. 

core problem is that dropdown with 100+ entries and no autocomplete is just terrible selector and it makes it url edit by hand look easier. real solution is probably do it in "ruby cannon" as you suggesting and then rewrite that combo box later into modern autocomplete convenient box for multi-title selection. 

I also forgot one flag I was planning to add, so I will add edit to PR rewrite. filter will also support `&live` to filter out only live games. that the final schema for my concept of "18xx evening bookmark": filter 2-3 titles, live only and see what is on tap that evening.

ollybh:
Adding a filter for live games is a great idea, but can you make the UI also work for people who are looking for async games? Something like radio buttons with three options: live, async or all.

lmx2020-18xx:
New URL schema, live/async/all filter. 

Screenshot 2026-04-11 233352

Screenshot 2026-04-11 233314

assets/app/lib/params.rb
```diff
@@ -27,6 +27,13 @@ def [](key)
         @native&.get(key)
       end
 
+      def get_all(key)
+        return [] if @unsupported
+
+        result = Native(`Array.from(#{@native.to_n}.getAll(#{key}))`)
+        result.to_a
```
ollybh: Does this work?
```suggestion
        @native.getAll(key)
```

assets/app/lib/params.rb
```diff
@@ -35,8 +42,13 @@ def []=(key, value)
         end
       end
 
+      def set_array(key, values)
+        @native&.delete(key)
```
ollybh: I wouldn't bother with the safe navigation operator here, we're going to get an error on the next line if `@native` isn't working.
```suggestion
        @native.delete(key)
```

assets/app/lib/params.rb
```diff
@@ -35,8 +42,13 @@ def []=(key, value)
...
+        Array(values).each { |v| `#{@native.to_n}.append(#{key}, #{v})` }
```
ollybh:
```suggestion
        values.each { |v| @native.append(key, v) }
```

assets/app/view/game_row_filters.rb
```diff
@@ -12,40 +12,120 @@ def render
...
+              padding: '4px 12px',
+              cursor: 'pointer',
+              background: active ? '#4a4a4a' : '#ddd',
+              color: active ? '#fff' : '#333',
+              border: '1px solid #888',
+              fontWeight: active ? 'bold' : 'normal',
+            },
+            on: { click: on_click },
+          }, opt[:label])
+      end
```
ollybh: The array of hashes is probably more complex that is necessary. Is this any clearer? It's no problem if you prefer your suggestion, it's not a significant change.
```suggestion
      current = url_search_params['mode'] || ''

      options = { 'All' => '', 'Live' => 'live', 'Async' => 'async' }
      buttons = options.map do |label, value|
        on_click = lambda do
          url_search_params['mode'] = value
          update_filters(url_search_params.to_query_string)
        end

        active = current == value
        h(:button, {
            style: {
              padding: '4px 12px',
              cursor: 'pointer',
              background: active ? '#4a4a4a' : '#ddd',
              color: active ? '#fff' : '#333',
              border: '1px solid #888',
              fontWeight: active ? 'bold' : 'normal',
            },
            on: { click: on_click },
          }, label)
      end
```

models/game.rb
```diff
@@ -17,7 +17,10 @@ class Game < Base
       SELECT *
       FROM games
       WHERE status = '%<status>s'
-        AND (:title IS NULL OR :title = title)
+        AND (:titles IS NULL OR title = ANY(:titles))
+        AND (:mode IS NULL
+             OR (:mode = 'async' AND COALESCE((settings->>'is_async')::boolean, false))
+             OR (:mode = 'live' AND NOT COALESCE((settings->>'is_async')::boolean, false)))
```
ollybh:
```suggestion
        AND (:mode IS NULL OR COALESCE((settings->>'is_async')::boolean, :mode == 'async'))
```

ollybh (changes requested):
Thanks for making these changes, this looks much better without the string substitution and manual encoding/decoding.

Could you stick with the existing `title` parameter instead of changing it to `title[]`? That would mean any existing bookmarks would still work and (IMHO) the square brackets look odd.

I've suggested a few changes. I don't think you need to call `Native` as much in the URLSearchParams code. And there are a few other tweaks.

lmx2020-18xx:
> Could you stick with the existing `title` parameter instead of changing it to `title[]`? That would mean any existing bookmarks would still work and (IMHO) the square brackets look odd.

As far as I understand that is Rails default we can not change if we want to use native Rails processing of array params?

> In Ruby on Rails, URL array parameters work by appending empty square brackets [] to the key name in the query string. When the Rails server (specifically the Rack middleware) sees multiple parameters with the same name followed by [], it automatically parses them into a single Ruby array within the params hash. 

> Access in the Controller: Rails automatically handles the incoming string and converts it into an array that you can access via the params object.

I can research if we can without conflict support 2 different params, like `title` as non-array for legacy URLs with old code, and `titles[]` for new filter, yet the price will be more complex code and we effectively will be carrying code only to support legacy bookmarks with no place in UI to trigger generation of legacy "title=<name>". Perhaps asking people to press ctrl-D one more time is not too big of a price for new schema?

ollybh:
> > Could you stick with the existing `title` parameter instead of changing it to `title[]`? That would mean any existing bookmarks would still work and (IMHO) the square brackets look odd.
> 
> As far as I understand that is Rails default we can not change if we want to use native Rails processing of array params?

We're not using Rails.

The URLSearchParams class is using Opal's [Native](https://opalrb.com/docs/api/v1.8.2/stdlib/Native.html) module to wrap a call to the Javascript [URLSearchParams ](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) object, so we can use any parameters syntax supported by that.

models/game.rb
lmx2020-18xx: **ccode** response: one liner as written would have a bug and recommends keeping it as is for readability:

> I don't think the logic is equivalent.

> The suggestion: COALESCE((settings->>'is_async')::boolean, :mode = 'async')

> The COALESCE default changes based on the mode, which means games with NULL is_async would always pass the filter regardless of
  which mode is selected:
  - mode = 'async' → COALESCE(NULL, true) → true → passes
  - mode = 'live' → COALESCE(NULL, false) → false... wait, that's just a bare boolean with nothing to compare against

>  Actually the whole expression is just a boolean with no comparison — it would filter based on the value of is_async being truthy, not whether it matches the selected mode. So mode = 'live' would require is_async = true, which is the opposite of what we want.

>  I'd keep it as-is for consistency with the rest of the file. The reviewer's suggestion was clever but doesn't quite work, and the "correct" single-line version is harder to read at a glance.

lmx2020-18xx:
> The URLSearchParams class is using Opal's [Native](https://opalrb.com/docs/api/v1.8.2/stdlib/Native.html) module to wrap a call to the Javascript [URLSearchParams ](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) object, so we can use any parameters syntax supported by that.

That makes it far easier to change. Comited and fully compatible with old URLs (they are treated as single item arrays)

assets/app/view/game_row_filters.rb
lmx2020-18xx: accepted in last commit

assets/app/lib/params.rb
lmx2020-18xx: accepted in last commit

assets/app/lib/params.rb
lmx2020-18xx: accepted in last commit

assets/app/lib/params.rb
lmx2020-18xx: accepted in last commit

models/game.rb
ollybh: Yeah, I messed up. What I meant was:
```AND (:mode IS NULL OR COALESCE((settings->>'is_async')::boolean, true) = (:mode = 'async'))```

File: routes/game.rb [hunk 9]
```diff
@@ -185,7 +185,8 @@ class Api
       end
 
       r.get do
-        { games: Game.home_games(user, **r.params) }
+        params = Rack::Utils.parse_query(r.query_string)
+        { games: Game.home_games(user, **params) }
```
ollybh, line 189: Can you explain what this change does?

lmx2020-18xx: Rack converts array params in query to an array. without this line it only picks up last "&title=" in the chain. previous "title[]=" syntax is processed automatically, without bracket syntax we now have to manually parse query to pack all "title=" into array.

models/game.rb
lmx2020-18xx: **ccode** is unhappy again:

>  No — the true default breaks it. Games with NULL is_async (older games, should be treated as live) would show up under "Async" and
  be hidden under "Live".

>  With COALESCE default true:
>  - is_async is NULL, mode='async': COALESCE(NULL, true) = true, (:mode='async') = true → true = true → included (wrong, should be
  excluded)
>  - is_async is NULL, mode='live': COALESCE(NULL, true) = true, (:mode='async') = false → true = false → excluded (wrong, should be
  included)

ollybh: Oh Claude, you understand so much and yet understand so little.

The UI treats games with null `is_async` setting the same as if `is_async` is false. So they should appear when the filter is for async games and not when the filter is for live games.

lmx2020-18xx: I doubt Claude understands anything at all, just that this was the answer is highest probability 📈

ollybh (commented):
Can you change the mode test in the USER_STATUS_QUERY to match the updated STATUS_QUERY?

Once that is done I think this will be all good.

ollybh (approved):

lmx2020-18xx:
@ollybh Is there anything else I should do on this PR?

PR #12484, opened 2026-04-06
[18Cuba] Narrow Gauge Graph

chrstas:
This PR introduces entity-specific track graphs to enforce the gauge restrictions between minor and major companies in 18Cuba.

Minors can only build and trace routes using narrow gauge
Majors can only build and trace routes using standard (broad) gauge

This is implemented by assigning separate graph instances per entity type, ensuring that illegal connectivity is prevented at the graph level rather than during route validation.

## Before clicking "Create"

### Explanation of Change

**Minor Companies**

“Once a tile has been placed in a minor company's sugar mill hex, it may lay tiles [...] reachable [...] using a route of narrow gauge track only.”
“A minor company may lay only yellow tiles with narrow gauge track (except the home city tile).”

Enforced via:
- Graph.new(... skip_track: :broad)
- Tile filtering (existing logic)

**Major Companies**

“Once a tile has been placed in a major company's home hex, it may lay tiles [...] reachable [...] using a route of standard gauge track only.”
“A major company may only lay yellow tiles with standard gauge track.”

Enforced via:
- init_graph skipping :narrow

**Connectivity Rules**

“A track tile may be built in such a way that it fails to connect to tracks on a neighboring tile.”
“A track tile may be laid in such a way that narrow gauge links with standard gauge across a hexside. Then both tracks have a dead end.”

Supported via:
- Mixed tiles are still allowed
- Dead-ends naturally emerge without special handling

### DIFF
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -134,6 +134,22 @@ module Engine
           initialize_tile_opposites!
           @unused_tiles = []
           @sugar_cubes = {}
+          @minor_graph = Graph.new(self, skip_track: :broad)
+        end
+
+        def init_graph
+          Graph.new(self, skip_track: :narrow)
+        end
+
+        def graph_for_entity(entity)
+          return @graph unless entity&.type == :minor
+
+          @minor_graph ||= Graph.new(self, skip_track: :broad)
+        end
+
+        def clear_graph
+          @minor_graph = nil
+          super
         end
 
         def init_tile_groups
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -148,8 +148,16 @@ module Engine
         end
 
         def clear_graph
-          @minor_graph = nil
+          @minor_graph.clear
           super
+        end
+
+        def clear_graph_for_entity(entity)
+          if entity&.type == :minor
+            @minor_graph.clear
+          else
+            super
+          end
         end
 
         def init_tile_groups


lib/engine/game/g_18_cuba/game.rb
```diff
@@ -134,6 +134,22 @@ def setup
...
+          Graph.new(self, skip_track: :narrow)
+        end
+
+        def graph_for_entity(entity)
+          return @graph unless entity&.type == :minor
+
+          @minor_graph ||= Graph.new(self, skip_track: :broad)
+        end
+
+        def clear_graph
```
ollybh, line 150: You'll need to provide an implementation of `clear_graph_for_entity` as well as this.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -134,6 +134,22 @@ def setup
...
+          @minor_graph = nil
```
ollybh: Setting the graph to `nil` is going to cause failures. Use `Graph.clear` to indicate the graph needs to be recalculated before it can be used again.
```suggestion
          @minor_graph.clear
```

ollybh (changes requested):
This is the right approach to take for handling the different type of gauge, but there are a couple of things that need to be fixed.

chrstas:
thanks for the review and hints. have changed to .clear and added the entity clearance for the graph with logic to clear the minor graph. 

Alternative that I saw in 18Norway:

        def clear_graph_for_entity(_entity)
          clear_graph
        end
  
would be maybe even better to clear both graphs?

ollybh (approved):

PR #12498, opened 2026-04-11
[18Cuba] Train buy logic and rusting

chrstas:
Implements the narrow gauge train buy and exchange rules for minors, including the 4n→4-1n downgrade event triggered by the first 8/4D train purchase.

Includes also the emergency buy rule to only buy the cheapest train / variant from bank or bank pool

**Rules reference**

A minor railroad can exchange its n-train for the next higher type once per operating round. It receives half the value of the old train as a credit toward the new one. It cannot skip a type.

When the first 8 or 4D train is purchased, all 4n trains owned by corporations downgrade to 4-1n trains (distance 3, worth $130). A 4-1n can be exchanged for a 5n ($380) with a $65 credit, for a net cost of $315.

Emergency Financing: The train may be bought only from the bank or the bank pool, but not from another company. If different trains are available, then the director must buy the cheapest train available. The director may add only the amount of money needed in order to buy the train; thus, following the train purchase, the company will have no money le in its treasury.

## Before clicking "Create"

#### `step/buy_train.rb` (new)

Custom BuyTrain step for 18Cuba:

- Minors may only buy narrow gauge trains (2n, 3n, 4n, 5n); majors may only buy broad gauge trains.
- Narrow gauge trains are now visible in the depot once the corresponding phase is active (`available_on`).
- Implements the once-per-OR exchange mechanic: a minor may trade in its current n-train for the next type, receiving half the original purchase price as a credit. Attempting to exchange a second time in the same OR raises an error.
- Exchange tracking uses `round_state` (resets automatically each OR), consistent with the pattern used in 1824.

#### `trains.rb`

- Added `available_on` 
- Added `discount` for n-train exchanges:
- Added `events: [{ 'type' => 'downgrade_4n_trains' }]` to the `8` train.
- Added event_downgrade_4n_trains!: when the first 8 or 4D train is purchased, all 4n trains (corporation-owned, depot, and discard) downgrade to the 4-1n variant. The 4-1n can be exchanged toward a 5n at a $65 credit — half the credit of a full 4n ($130).

### DIFF
--- a/lib/engine/game/g_18_cuba/step/buy_train.rb
+++ b/lib/engine/game/g_18_cuba/step/buy_train.rb
@@ -13,13 +13,19 @@ module Engine
           end
 
           def buyable_trains(entity)
-            trains = super
+            track_type = entity.type == :minor ? :narrow : :broad
 
-            if entity.type == :minor
-              trains.select { |t| t.track_type == :narrow }
-            else
-              trains.select { |t| t.track_type == :broad }
+            if must_buy_train?(entity)
+              # Emergency buy: only depot/discard trains allowed (not from other companies),
+              # and only the cheapest of the correct track type.
+              available = @depot.depot_trains.select { |t| t.track_type == track_type }
+              cheapest_price = available.map { |t| t.price }.min
+              result = cheapest_price ? available.select { |t| t.price == cheapest_price } : []
+              return result
             end
+
+            # Normal buy: filter by track type.
+            return super.select { |t| t.track_type == track_type }
           end
 
           def buyable_train_variants(train, entity)
@@ -27,6 +33,12 @@ module Engine
             # 4n trains can downgrade to 4-1n.
             # Only the currently active variant (based on train.name) is buyable.
             return variants.select { |variant| variant[:name] == train.name } if train.variants.key?('4-1n')
+
+            # During emergency buy, only the cheapest variant is allowed.
+            if must_buy_train?(entity)
+              min_price = variants.map { |v| v[:price] }.min
+              return variants.select { |v| v[:price] == min_price }
+            end
 
             variants
           end
--- a/lib/engine/game/g_18_cuba/step/buy_train.rb
+++ b/lib/engine/game/g_18_cuba/step/buy_train.rb
@@ -19,13 +19,13 @@ module Engine
               # Emergency buy: only depot/discard trains allowed (not from other companies),
               # and only the cheapest of the correct track type.
               available = @depot.depot_trains.select { |t| t.track_type == track_type }
-              cheapest_price = available.map { |t| t.price }.min
+              cheapest_price = available.map(&:price).min
               result = cheapest_price ? available.select { |t| t.price == cheapest_price } : []
               return result
             end
 
             # Normal buy: filter by track type.
-            return super.select { |t| t.track_type == track_type }
+            super.select { |t| t.track_type == track_type }
           end
 
           def buyable_train_variants(train, entity)
--- a/lib/engine/game/g_18_cuba/step/buy_train.rb
+++ b/lib/engine/game/g_18_cuba/step/buy_train.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/buy_train'
+
+module Engine
+  module Game
+    module G18Cuba
+      module Step
+        class BuyTrain < Engine::Step::BuyTrain
+          def round_state
+            # Track which minor corporations have exchanged for narrow gauge trains this OR
+            super.merge({ narrow_gauge_exchanged_by: [] })
+          end
+
+          def buyable_trains(entity)
+            trains = super
+
+            if entity.type == :minor
+              trains.select { |t| t.track_type == :narrow }
+            else
+              trains.select { |t| t.track_type == :broad }
+            end
+          end
+
+          def discountable_trains_allowed?(entity)
+            # Minors can only exchange for narrow gauge trains, and can only do so once per OR.
+            entity.type == :minor && !@round.narrow_gauge_exchanged_by.include?(entity.id)
+          end
+
+          def process_buy_train(action)
+            # If the player is exchanging a narrow gauge train, ensure they haven't already done so this OR,
+            # and track that they have.
+            if action.exchange
+              raise GameError, "#{action.entity.name} has already exchanged a narrow gauge train this OR" \
+                unless discountable_trains_allowed?(action.entity)
+
+              @round.narrow_gauge_exchanged_by << action.entity.id
+            end
+            super
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_18_cuba/step/buy_train.rb
+++ b/lib/engine/game/g_18_cuba/step/buy_train.rb
@@ -32,9 +32,9 @@ module Engine
             variants = super
             # Downgrade train variants (e.g. 4n-1) are not player-chooseable; only the current one is buyable.
             if train.variants.values.any? { |v| v[:event_downgrade_variant] }
-              return variants.select do |v|
-                       v[:name] == train.name
-                     end
+              result = variants.select { |v| v == train.variant }
+
+              return result
             end
 
             # During emergency buy, only the cheapest variant is allowed.
--- a/lib/engine/game/g_18_cuba/step/buy_train.rb
+++ b/lib/engine/game/g_18_cuba/step/buy_train.rb
@@ -30,9 +30,12 @@ module Engine
 
           def buyable_train_variants(train, entity)
             variants = super
-            # 4n trains can downgrade to 4-1n.
-            # Only the currently active variant (based on train.name) is buyable.
-            return variants.select { |variant| variant[:name] == train.name } if train.variants.key?('4-1n')
+            # Downgrade train variants (e.g. 4n-1) are not player-chooseable; only the current one is buyable.
+            if train.variants.values.any? { |v| v[:event_downgrade_variant] }
+              return variants.select do |v|
+                       v[:name] == train.name
+                     end
+            end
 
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_18_cuba/trains.rb
```diff
@@ -100,28 +101,47 @@ module Trains
...
+            available_on: '5',
+            discount: { '4n' => 130, '4-1n' => 65 },
           },
           ].freeze
+
+        def event_downgrade_4n_trains!
+          @log << '-- Event: 4n trains downgrade to 4-1n trains --'
+          trains.each do |train|
+            next unless train.name == '4n'
+            next unless train.owned_by_corporation?
```
ollybh: Could there be unsold 4n trains when this happens? The rulebook says that the price of aged 4n trains is halved, so I'm thinking that these might still be available to purchase.

lib/engine/game/g_18_cuba/trains.rb
```diff
@@ -100,28 +101,47 @@ module Trains
             distance: 2,
             price: 80,
             track_type: :narrow,
+            available_on: '2',
             rusts_on: '4',
           },
           {
             name: '3n',
             distance: 3,
             price: 160,
```
ollybh: I hadn't noticed this before: the rule book shows a 3n train costing $160 but the train cards show $180. According to [Lonny](https://boardgamegeek.com/thread/3232920/article/43676587#43676587) the correct cost is $180.

lib/engine/game/g_18_cuba/trains.rb
```diff
@@ -100,28 +101,47 @@ module Trains
...
             rusts_on: '6',
+            discount: { '2n' => 40 },
           },
           {
             name: '4n',
             distance: 4,
             price: 260,
             track_type: :narrow,
+            available_on: '4',
+            discount: { '3n' => 80 },
```
ollybh, line 122: If the 3n costs $180 then I assume this discount should be $90. Maybe. The train cards show a discount of $80. :shrug:

ollybh: I couldn't find clarification of this on BGG.
@ViennaTramways should a minor company get a $80 or $90 discount when trading a 3n train in for a 4n?

ollybh (commented):
The code looks fine, but I've got a couple of rules questions.

lib/engine/game/g_18_cuba/trains.rb
chrstas: I will follow for the moment the cards with 80 then it is consistent. Will place a marker in my checklist.

chrstas:
@ollybh thanks for the review and I am impressed by your rules findings... I have changed:

- 3n to 180$
- Discount 4n kept with 80$ (to be consistent with cards). But I will make a note on my checklist. If we get another input from Lonny I will adapt accordingly
- Downgrade also trains in depot/discard

lib/engine/game/g_18_cuba/trains.rb
```diff
@@ -100,28 +101,53 @@ module Trains
...
+            @log << if train.owned_by_corporation?
+                      "#{train.owner.name}'s 4n train downgrades to 4-1n"
+                    else
+                      'Depot/Discard 4n train downgrades to 4-1n'
+                    end
+            train.name = '4-1n'
+            train.distance = 3
+            train.variant[:name] = '4-1n'
+            train.variant[:distance] = 3
+            train.variants['4-1n'] = train.variants.delete('4n') if train.variants.key?('4n')
```
ollybh:
```suggestion
            train.variants.transform_keys!( { '4n' => '4n-1' })
```

ollybh (commented):
Are the 4n-1 trains shown correctly on the game info tab? Or do you still see 4n trains on there?

An alternative approach would be to have the 4n-1 trains defined as 4n train variants in the TRAINS constant. Then have a `buyable_train_variants` method that looks at the phase and either excludes the 4n or the 4n-1 train variant according to the current phase. And the downgrade event then switches owned trains between the variants.

chrstas:
> Are the 4n-1 trains shown correctly on the game info tab? Or do you still see 4n trains on there?
> 
> An alternative approach would be to have the 4n-1 trains defined as 4n train variants in the TRAINS constant. Then have a `buyable_train_variants` method that looks at the phase and either excludes the 4n or the 4n-1 train variant according to the current phase. And the downgrade event then switches owned trains between the variants.

hi @ollybh, I have checked the info tab and it looks fine. I have attached the screenshot...

and thanks for the `.transform_keys`, learned something new again

image

ollybh:
> I have checked the info tab and it looks fine. I have attached the screenshot...

Actually, that isn't quite right – 4n-1 trains should cost $130.

ollybh:
Including the 4n-1 trains as variants in the TRAINS definition would mean that these get shown in the train info table from the start of the game, along with the 4n train. I don't know if that's needed or not.

chrstas:
@ollybh: yes - if info should be shown from the beginning then an own variant is the only way I see. I have adjusted it and implemented it with variant and `buyable_trains`

With the last commits I have considered also the emergency buy rule to ensure only buying the cheapest train variant from bank / pool and considering track type. PR description is updated as well.

including also the screenshot... hope I did not overlook anything:

image

lib/engine/game/g_18_cuba/step/buy_train.rb
```diff
@@ -0,0 +1,66 @@
...
+
+            # Normal buy: filter by track type.
+            super.select { |t| t.track_type == track_type }
+          end
+
+          def buyable_train_variants(train, entity)
+            variants = super
+            # 4n trains can downgrade to 4-1n.
+            # Only the currently active variant (based on train.name) is buyable.
+            return variants.select { |variant| variant[:name] == train.name } if train.variants.key?('4-1n')
```
ollybh: I don't understand this line. Can you explain what it does?

chrstas: hi @ollybh, yes for sure... and maybe there is another way to do it more elegant. The reasoning behind this line from my side that the player should not be allowed to switch variants freely, he / she only can buy the currently active variant (-> set in the downgrade event)

The filter should only kick in if for a train, for which a 4n-1 variant exists.

chrstas: your question triggered some thoughts on my side and the hard-coding is probably not that ideal. I would push another commit and try to go with a flag in the variant definition (e.g. event_downgrade_variant: true) and use this for filtering

ollybh: I still don't follow how this works. The 4n train object will be set up something like this:
```
name: '4n',
variants: { '4n' => { 'name' => '4n', …. }, '4n-1' => { 'name' => '4n-1', … },
…
```
So won't `variant[:name] == train.name` only ever be true for the 4n variant?

chrstas: @ollybh, thank you again for digging into this — I really appreciate your efforts.

I've added some debug logging to clarify exactly what's happening, and I think it answers your question:

```
[downgrade_4n] BEFORE: train.name=4n  train.variant={"name"=>"4n",  ...}
[downgrade_4n] AFTER:  train.name=4-1n train.variant={"name"=>"4-1n", ...}
```
based on:

```
        def event_downgrade_4n_trains!
          @log << '-- Event: 4n trains downgrade to 4-1n trains --'
          trains.each do |train|
            next unless train.name == '4n'

            @log << if train.owned_by_corporation?
                      "#{train.owner.name}'s 4n train downgrades to 4-1n"
                    else
                      'Depot/Discard 4n train downgrades to 4-1n'
                    end
            LOGGER.debug { "[downgrade_4n] BEFORE: train.name=#{train.name} train.variant=#{train.variant}" }
            train.variant = '4-1n'
            LOGGER.debug { "[downgrade_4n] AFTER:  train.name=#{train.name} train.variant=#{train.variant}" }
          end
```

And then when buyable_train_variants is called after the event, it correctly shows only the 4-1n variant as buyable:

`[buyable_variants] train.name=4-1n train.variant=4-1n all=4n,4-1n buyable=4-1n`

based on:

```
          def buyable_train_variants(train, entity)
            variants = super
            # Downgrade train variants (e.g. 4n-1) are not player-chooseable; only the current one is buyable.
            if train.variants.values.any? { |v| v[:event_downgrade_variant] }
              result = variants.select { |v| v == train.variant }
              LOGGER.debug do
                "[buyable_variants] train.name=#{train.name} train.variant=#{train.variant[:name]} " \
                  "all=#{variants.map { |v| v[:name] }} buyable=#{result.map { |v| v[:name] }}"
              end
              return result
            end
```

Your review made me realise the original v[:name] == train.name was unnecessarily indirect. I've updated the code to use v == train.variant instead, which directly compares the variant object rather than going via the name. I think this makes the intent a bit clearer.

chrstas: also wanted to understand it a bit better how train.name behaves in terms of variants (with AI help for finding the root cause, to be honest). I assume it is this part of the engine in trains.rb which updates train.name:

`variant[:name] == train.name` would always match only the 4n variant if `train.name` never changed. When `event_downgrade_4n_trains!` calls `train.variant = '4-1n'`, the `instance_variable_set` call updates all instance variables from the new variant's data — including `@name`

```
   def variant=(new_variant)
      return unless new_variant

      @variant = @variants[new_variant]
      @variant.each { |k, v| instance_variable_set("@#{k}", v) }

      # Remove the @local variable, this to get the local? method evaluate the new variant
      remove_instance_variable(:@local) if defined?(@local)
    end
```

ollybh: Got it, thanks for explaining.

ollybh (approved):

PR #12500, opened 2026-04-12
[UI] Game data - Show additional details

### LINKED ISSUE #12344: [Various] Add specific game data to Info or Tools tab
**What's your idea?**

In the Tools tab, one can see all JSON data related to a game by clicking "Show" under the **Game Data** header. The specific data points I was thinking could be the _description_ (which one doesn't see after clicking into the game, otherwise), _the user that started the game_ (`user.name`, maybe linked to their account?) and, perhaps, the start date/`created_at` time.

**Are there related ideas?**

I found [this similar issue](https://github.com/tobymao/18xx/issues/11092), but it didn't seem to go anywhere. I'm happy to implement this myself if other people would like this info as well.

**Did you join the Slack?**

chrstas:
Fixes #12344

Add game metadata to the Game Data section in the Tools tab

Displays additional information about the game in the section "Game Details": description, host (user name), creation time, and last updated time. Each field is shown, also if data is null.

## Before clicking "Create"

## Implementation Notes

- I decided to stick with the format like in the other sections of the tools page.
- I decided to go for an own section "Game Details" (first approach with Game Data had side effects and details were shown also in the box for "broken games")
- Time/Date format in the same way as on the start page.
- User name links to profile page

### Screenshots

image

### DIFF
--- a/assets/app/view/game/tools.rb
+++ b/assets/app/view/game/tools.rb
@@ -25,8 +25,8 @@ module View
           h(RenameHotseat),
           *render_tools,
           h(AutoRouterSettings),
-          h(GameDetails),
           h(GameData, actions: @game.raw_actions.map(&:to_h)),
+          h(GameDetails),
           *help_links,
         ])
       end
--- a/assets/app/view/game/tools.rb
+++ b/assets/app/view/game/tools.rb
@@ -3,6 +3,7 @@
 require 'game_manager'
 require 'lib/storage'
 require 'view/game/game_data'
+require 'view/game/game_details'
 require 'view/game/notepad'
 require 'view/game/actionable'
 require 'view/game/auto_router_settings'
@@ -24,6 +25,7 @@ module View
           h(RenameHotseat),
           *render_tools,
           h(AutoRouterSettings),
+          h(GameDetails),
           h(GameData, actions: @game.raw_actions.map(&:to_h)),
           *help_links,
         ])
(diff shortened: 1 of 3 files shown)


assets/app/view/game/tools.rb
```diff
@@ -24,6 +25,7 @@ def render
           h(RenameHotseat),
           *render_tools,
           h(AutoRouterSettings),
+          h(GameDetails),
           h(GameData, actions: @game.raw_actions.map(&:to_h)),
```
michaeljb:
```suggestion
          h(GameData, actions: @game.raw_actions.map(&:to_h)),
          h(GameDetails),
```

Could you move Details to after the Data?

For development/debugging I use the the Game Data buttons quite a lot and don't want to see them moved for this feature.

michaeljb (changes requested):

assets/app/view/game/tools.rb
chrstas: yes, of course. changed the sequence accordingly. thanks.

ollybh (changes requested):
I get this for a hotseat game:
Image

There are several things to change:
- The 'You' links to a broken page.
- I created the game slightly more recently than is shown.
- (also applies to non-hotseat games) It would be better to suppress the description line if there isn't one.

chrstas:
> I get this for a hotseat game: Image
> 
> There are several things to change:
> 
> * The 'You' links to a broken page.
> * I created the game slightly more recently than is shown.
> * (also applies to non-hotseat games) It would be better to suppress the description line if there isn't one.

valid points - have included the logic to deal with the specifics as suggested. thanks...

ollybh (approved):

PR #12502, opened 2026-04-13
[1880 Romania] Foundation

philcampeau:
## Summary
- Pre-alpha implementation of 1880 Romania
- Updates to trains, track, and tile configuration
- Reintegrates TRAINS and PHASES constants directly into game.rb (removes separate phases.rb and trains.rb files)

I also shifted the map over one hex to match with changes Lonny made on his end.


philcampeau:
Just compared with my older PR #12394 and this isn't the same at all. Claude Code fail. Closing this.

PR #12505, opened 2026-04-14
[1880 Romania] PR3 - Game Mechanics

philcampeau:
## Summary
Implements core game mechanics for 1880 Romania:

- Restrict B-labeled tile to Bucuresti hex
- Fix routing for + trains
- Mask company abilities on corps unless assigned
- Tile lay rules (1 lay, 2 as of green phase)
- Fix FI assignment for TR corp
- TR hijacks BCR role from 1880 China; receives ABC building permit automatically
- Province border income and border removal at communism phase
- Signal end game event text
- Prevent last 8 train from triggering SR

Depends on #12502 and #12503

PR #12503, opened 2026-04-13
[1880 Romania] PR2 - private companies

philcampeau:
Depends on #12502

## Before clicking "Create"

- [x] Branch is derived from the latest `master`
- [x] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

### Explanation of Change

  Implements the private company abilities for 1880 Romania (P1–P7):

  - P1 (Banca Română): building permit token ability
  - P2 (Consorțiul CFR): province/border crossing income
  - P3 (Regia Autonomă CFR): train discount
  - P4 (Portul Dunării): Danube bonus assignment
  - P5 (Remar Pașcani): train cost reduction ability
  - P6 (Electroputere S.A.): forced train exchange
  - P7 (Malaxa Manufacturing): track-building restriction

PR #12506, opened 2026-04-14
[1880 Romania] PR4 - Fixture file

philcampeau:
## Summary
- Adds a game fixture for the final phase / game end condition

Depends on #12502 ,#12503, and #12505.


philcampeau:
Just compared with my older PR #12394 and this isn't the same at all. Claude Code fail. Closing this.

PR #12507, opened 2026-04-15
[1880] Show capitalization status on corporation charter card

### LINKED ISSUE #12416: [1880]: indicate company received 2nd capitalization round
**What's your idea?**

In 1880 China, companies get half their capital when they float, and the other half when 50% is out of the IPO any point after start of phase B1. In the tabletop game, there is a card that indicates whether a company has received this money already or not  (see rulebook p20: "Capitalisation of a major comany", 4th bullet), which is a useful tracking aid if a companies stocks get sold later putting them back over 50% in IPO. 
I believe that the implentation on .games does not have any indicator on whether or not a company has received their full capitalisation already. Instead, one needs to search the log, which is kind of cumbersome.

It would be great if there could be some indication on the company charter to show whether it has received capitalization already.

ddaybell:
Corporations in 1880 receive capital in two tranches: 50% on float and the remaining 50% once 50% of shares have been sold. The charter card now displays "Capitalization: Half" or "Capitalization: Full" to make this status visible at a glance.

Fixes #12416 

image

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1880/game.rb
+++ b/lib/engine/game/g_1880/game.rb
@@ -536,11 +536,7 @@ module Engine
           status = ["Building Permits: #{corporation.building_permits}"]
           status << ["Presidency: #{corporation.presidents_percent}%"]
           if corporation.floated?
-            status << if corporation.fully_funded
-                        ['Capitalization: Full', 'bold']
-                      else
-                        ['Capitalization: Half', nil]
-                      end
+          status << ["Capitalization: #{corporation.fully_funded ? 'Full' : 'Half'}"] if corporation.floated?
           end
           status
         end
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_1880/game.rb
```diff
@@ -535,6 +535,13 @@ def status_array(corporation)
 
           status = ["Building Permits: #{corporation.building_permits}"]
           status << ["Presidency: #{corporation.presidents_percent}%"]
+          if corporation.floated?
+            status << if corporation.fully_funded
+                        ['Capitalization: Full', 'bold']
+                      else
+                        ['Capitalization: Half', nil]
+                      end
+          end
```
ollybh: Having the font change from normal to bold isn't necessary. And using a ternary rather than an if statement might be clearer by removing repetition.
```suggestion
          if corporation.floated?
          status << ["Capitalization: #{corporation.fully_funded ? 'Full' : 'Half'}"] if corporation.floated?
          end
```

crericha (approved):

PR #12508, opened 2026-04-15
[18OE] Consolidation round, major phase, train-buy rules & map fixes

Witzman:
Consolidation round
- Add Round::G18OE::Consolidation and Step::Consolidate (pass-only stub for now)
- Inject consolidation round between OR set end and next SR once Phase 5+
  triggers @consolidation_triggered; runs exactly once (@consolidation_done)
- Consolidation entities: players holding at least one minor or regional corp

Major phase detection to prevent selling shares in SRs before first major forms
- Replace ad-hoc corporations.all?(&:ipoed) checks with new major_phase?
  helper (true once MAX_FLOATED_REGIONALS have been floated)
- Gate IPO share purchases and regional conversions behind major_phase?

Train-buy rules
- must_buy_train?: minors exempt; obligation waived from Phase 4 onward
- can_buy_train_from_others?: inter-corp sales only in Phase 4+
- BuyTrain#buyable_trains: enforce 2+2-first obligation (Phase < 4, no
  trains) and depot-level gating (cheapest tier must sell out first)
- BuyTrain#can_entity_buy_train?: exclude minors explicitly

Operating/stock round hooks
- or_set_finished: set @consolidation_triggered on Phase 5+
- sold_out_increase?: share-price UP only for :major and :national types
- next_round!: route through consolidation round before next SR

Map fix
- Hex AA82: upgrade from bare city to dual-city (revenue 30/30) with pre-
  printed paths and label=C (Constantinople)

Temporary workarounds (guarded behind NATIONAL_REGION_HEXES_COMPLETE=false) to enable testing midgame
- hex_within_national_region?: bypass zone filter until all 8 hex lists done
- home_token_locations: allow any non-metropolis tokenable city on the map
- home_token step: guard delete_at against nil region index for out-of-zone
  placements
- TODO comment: missing PHS/AH/IT/SP/RU/SC hex lists pending map data

FYI my tracking list:
## 1. Nationals 
- [ ] **1.1** Define `National` company type — type `:national` wired into corporation; inherits major's stock market position on conversion
- [ ] **1.2** National formation trigger — `event_nationals_can_form!` fires when first level-4/6/8 train purchased; players queued in purchaser-first order
- [ ] **1.3** National formation steps implemented in `convert_to_national`:
  - [ ] Major's cash → bank
  - [ ] All major treasury certs → Open Market (50% limit temporarily exceeded as per rules)
  - [ ] Remove all major tokens from map
  - [ ] Abandon any merged minors *(stub — deferred)*
  - [ ] Remove track rights, OE markers, private markers *(stub — not yet tracked in engine)*
  - [ ] Nationals operate by share value (included in majors+nationals sort bucket)
  - [ ] National inherits all trains including rusted (`rust?` override protects nationals' trains)
- [ ] **1.4** National revenue calculation (`national_revenue` in game.rb):
  - [ ] Virtual tokens via `Graph.new(home_as_token: true, no_blocking: true)`
  - [ ] Linked cities/towns counted at face value (best-first)
  - [ ] Unlinked consume capacity at £0; remaining capacity fills at £60/city, £10/town
  - [ ] D trains doubling linked cities *(deferred — D trains not yet in UK-FR train set)*
  - [ ] Red zones in zone via port arrow *(deferred)*
  - [ ] Must pay ALL revenue as dividends — `dividend_types` returns `[:payout]` only for nationals
- [ ] **1.5** Inherent Pullman bonus: +£10 × level of highest non-rusted train added in `national_revenue`
- [ ] **1.6** Nationals cannot place tokens — `Token#actions` returns `[]` for nationals
- [ ] **1.7** Nationals exempt from terrain costs — `tile_cost` returns 0 for nationals
- [ ] **1.8** National train limits: Phase 4=4, Phase 5–6=3, Phase 7–8=4 — already in `PHASES`; `enforce_national_train_limit` discards excess on formation
- [ ] **1.9** Nationals can claim rusted trains for free from the bank *(deferred to next PR)*
- [ ] **1.10** `NATIONAL_REGION_HEXES` in `game.rb` only defines UK and FR zones — the remaining 6 zones (PHS, AH, IT, SP, RU, SC) need their hex lists; `convert_to_national` and `national_revenue` reference this for all 8 zones *(blocked on map implementation — openpoints.md 2.5)*
- [ ] **1.11** `@minor_available_regions` in base game `setup` now derived dynamically from all regional corporations via `CORPORATIONS_TRACK_RIGHTS` — no hardcoded zone list

---

## 2. Base Game Entities + Map ~~(2.1–2.5 done; 2.6 in progress)~~

`g_18_oe/entities.rb` complete. `g_18_oe/map.rb` created and loadable; terrain,
blue coverage, and location names done; city types/revenues/red revenues still TBD.

- [x] **2.1** Define all 24 regionals in `g_18_oe/entities.rb` — home cities and coordinates from physical map; all 8 track rights zones covered in `CORPORATIONS_TRACK_RIGHTS` and `TRACK_RIGHTS_COST`
- [x] **2.2** Define all 12 minor corporations (A–M, excl. I) — type `:minor`, tokens, shares, no fixed coordinates, full `desc:` ability text from Rulebook §15 for all 12
- [x] **2.3** Define all 10 privates with face values, revenues, and `desc:` ability text from Rulebook §14 (RSC, PeC, WS, BBBT, SHTC, CCTC, WCF, HMLC, BBE, SML); auction rows renumbered into 6 price-tier rows (£20–£120)
- [x] **2.4** Minor-exchange auction cards for all 12 minors — rows 7–12, two per row; `par_via_exchange` auto-wired in `setup` via sym match
- [x] **2.5** Define 10 concession cards (CON1–CON10; £100 face value; TODO: verify against physical packet) *(concession phase deferred — see §15)*
- [ ] **2.6** Logo artwork — all current logos are grey circle stubs with abbreviated text; proper SVG artwork needed for all 25 corporations that lack real logos:
  - Minors (8): A, B, D, E, F, G, J, L
  - Regionals (17): BHB, POB, KSS, KBS, SB (PHS); MAV, SFAI, SFR (AH); CHN (IT); MZA, RCP (SP); MSP, MKV, LRZD (RU); WW, DSJ, BJV (SC)
  - Already have real logos: LNWR, GWR, GSWR (UK); PLM, MIDI, OU, BEL (FR); all minors C, H, K, M
- [~] **2.5** Base-game map `g_18_oe/map.rb` created and game-startable:
  - [x] Full grid coverage — all 666 blue (sea) hexes filled; no undefined gaps
  - [x] All 22 red off-board hexes with coordinates and names
  - [x] Terrain costs: UK/FR/Spain/Portugal/Scandinavia/Alps/Italy/Adriatic done
  - [x] LOCATION_NAMES: 258 entries covering all named cities, towns, and off-board labels
  - [x] Pre-printed yellows: Liverpool J25, Manchester J27, London M28 (revenue placeholders)
  - [x] Station markers: all 258 named locations now have `city=revenue:0` or `town=revenue:0` geometry — 131 hexes corrected (96 blank→city, 35 terrain+city); 5 border hexes (Odense, Larvik, Karlstad, Örebro, Vyborg) also fixed
  - [x] City vs town distinction applied from `csv/cities.csv`: town/double-town hexes reclassified; labeled cities (Y/A/B/N/S) added individually; ~60 cities in two `%w[]` groups; ~30 double-towns in group; terrain defs updated to town where appropriate
  - [ ] City revenues: all hexes have placeholder `revenue:0`; actual starting revenues needed from physical map
  - [x] Constantinople AA82: promoted to pre-printed yellow metropolis (OE5-equivalent, label=C, revenue:30 placeholder, edges 0/1/3/5 placeholder pending physical map)
  - [ ] Off-board revenues: all 22 red hexes have placeholder revenue:0
  - [ ] Terrain: Carpathians, Balkans, Caucasus, river crossings not yet filled
  - [ ] Sea zone borders, ferry paths, distance numbers not encoded
  - [~] NATIONAL_REGION_HEXES: only UK and FR defined (6 zones missing) — **temporary workaround active**: `NATIONAL_REGION_HEXES_COMPLETE = false` in game.rb bypasses zone filter for both home token placement and track laying, allowing unrestricted placement anywhere on the map. To remove the workaround when the map is complete: flip `NATIONAL_REGION_HEXES_COMPLETE` to `true`, delete the `unless` branch in `home_token_locations`, remove the `return true unless` branch in `hex_within_national_region?`, and remove the `if idx` guard in `step/home_token.rb#process_place_token`. The underlying zone-filtering logic is already there and will just work.

---

## 3. Train Mechanics (`step/buy_train.rb`)

Partially implemented (2026-04-12): 2+2 obligation, depot level gating, and inter-corp train gate are in. President shortfall path relies on base engine. Insolvency and nationals-claim-rusted deferred.

**Confirmed train data (Tables chart):**

| Level | Local/Express | Qty | Face Value | Rusts at |
|-------|--------------|-----|-----------|---------|
| 2     | 2+2 / —      | 30  | £100      | Level 4 |
| 3     | 3+3 / 3      | 20  | £225 / £200 | Level 6 |
| 4     | 4+4 / 4      | 10  | £350 / £300 | —       |
| 5     | 5+5 / 5      | 8   | £475 / £400 | —       |
| 6     | 6+6 / 6      | 6   | £600 / £525 | —       |
| 7     | 7+7 / 4D     | 14  | £750 / £850 | —       |
| 8     | 8+8 / 5D     | 8   | £900 / £1000 | —      |

Level 8 trains become available after the 4th Level 7 train is purchased.

- [x] **3.1** Reserved 2+2 train — `must_buy_train?` in game.rb returns true while corp has no trains and Phase < 4; `buyable_trains` in step restricts to 2+2 only during that window
- [ ] **3.2** Forced purchase — president covers shortfall from personal cash; if still insufficient, RR converts to national (majors) or enters insolvency (minors/regionals) *(president contribution relies on base engine `president_may_contribute?`; national/insolvency conversion deferred)*
- [ ] **3.3** First-round insolvency (minors/regionals): president's cash → treasury; RR gets reserved 2+2; presidential cert → Open Market; president receives face value payment; RR doesn't operate until new owner acquires it
- [x] **3.4** Reserved 2+2 obligation waived if Phase 4 starts before the RR's first OR — `must_buy_train?` checks `@phase.name.to_i >= 4`
- [x] **3.5** Depot level gating — `buyable_trains` filters to cheapest available depot level only; higher levels blocked until current level sold out
- [x] **3.6** Train purchase from other RRs gated to Phase 4+ — `can_buy_train_from_others?` in game.rb; price negotiation (min £1 / max treasury) handled by base engine
- [ ] **3.7** Nationals claim rusted trains for free *(trigger is implemented — free claim action in BuyTrain step is deferred)*

---

## 4. Stock Market Movement

- [x] **4.1** LEFT movement — zero dividend triggers `share_direction: :left` in `Dividend#share_price_change`
- [x] **4.2** No movement — dividend > 0 but < share value returns `{}` (explicit no-op now correct for majors and nationals)
- [x] **4.3** UP movement — `sold_out_increase?` in game.rb gates UP to `:major`/`:national` only; base engine `finish_round` calls `sold_out_stock_movement` → `move_up` at end of each SR
- [ ] **4.4** +3 RIGHT — on first Orient Express run by a major (in addition to normal dividend movement)

---

## 5. Orient Express Mechanic

- [ ] **5.1** Detect valid OE route: must include Constantinople + one of Paris/London/Berlin/Madrid/Sankt-Peterburg; must include some land track
- [ ] **5.2** First-time bonus: £30 (Phase 2–4), £60 (Phase 5–6), £100 (Phase 7–8); bank pays to treasury; place OE marker on major
- [ ] **5.3** Stock marker moves RIGHT×3 on first OE run (in addition to dividend movement)
- [ ] **5.4** Train combining for OE: trains of level ≤4 can combine (combined level = sum); level 5+ cannot combine; 4D/5D cannot combine
- [ ] **5.5** Subsequent OE runs: no bonus, no extra stock movement
- [ ] **5.6** Mandatory OE: if OE route is part of best possible routes, president must run it (unless tied)
- [x] **5.7** OE runs blocked for nationals — nationals have no route step; `routes_revenue` bypasses route calculation entirely for nationals
- [ ] **5.8** D-train bonus does NOT apply to OE first-time bonus

---

## 6. Pullman Cars

- [ ] **6.1** Pullman car asset type — does not count against train limit; max 1 per non-national RR
- [ ] **6.2** Revenue: +£10 × train level added to one designated train's revenue per OR
- [ ] **6.3** Acquisition from Minor M: £150 (+ £15 royalty to Minor M or its major); J-minor discount applies to purchase price but NOT to royalty
- [ ] **6.4** Acquisition from Open Market: £150 (no royalty); available starting Phase 4
- [ ] **6.5** Acquisition from another RR: negotiated price
- [x] **6.6** Nationals: inherent Pullman — +£10 × highest non-rusted train level computed in `national_revenue`
- [ ] **6.7** Minor M free Pullman: at Phase 4 start, if M hasn't been closed, place free Pullman in M's or its major's possession
- [ ] **6.8** Discard: when RR exceeds train limit, rusted trains go first; remaining Pullman may need discarding; can voluntarily discard to Open Market during buy trains step
- [ ] **6.9** If RR has zero trains + Pullman: Pullman stays on charter until next train acquired

---

## 7. Minor Special Abilities

**Status: descriptions only.** All 12 minors have rulebook `desc:` text in `entities.rb`
but zero functional `abilities:` entries — the engine will not trigger or enforce any
minor ability during play. Neither the ability logic nor the transfer-on-merge mechanism
is implemented.

- [ ] **7.1** Ability transfer mechanism — when minor merges with major, minor's charter placed under major's (ability inherited); nationals cannot inherit abilities (except stock-type)
- [ ] **7.2** Minor A (Silver Banner): bank pays major treasury = major's current share value (one-time on merge)
- [ ] **7.3** Minor B (Orange Scroll): all track upgrades cost 1 tile point (except cities, grand cities, metropolises)
- [ ] **7.4** Minor C (Golden Bell): president chooses operating position (first/last/normal) each OR; uses extra marker on Operating Order track
- [ ] **7.5** Minor D (Green Junction): places token in any non-metropolis/non-red-zone city; adds £20 bonus (Phase 2–4) or £40 bonus (Phase 5+) to trains running to/through it; removed at Phase 5 start
- [ ] **7.6** Minor E (Blue Coast): 33% discount on blue terrain costs; +1 tile point per OR for yellow tiles in blue terrain hexes
- [ ] **7.7** Minor F (White Peak): 33% discount on green terrain costs; +1 tile point per OR for yellow tiles in green terrain hexes
- [ ] **7.8** Minor G (Indigo Foundry): +2 tile points per OR
- [ ] **7.9** Minor H (Great Western Steamship): reduces sea zones counted toward city limit by 1 (Phase 1–6) or 2 (Phase 7–8)
- [ ] **7.10** Minor J (Grey Locomotive Works): 10% discount on all train purchases
- [ ] **7.11** Minor K (Vermilion Seal): mail contract pays revenue to treasury at start of each OR (not counted as dividend)
- [ ] **7.12** Minor L (Krasnaya Strela): +1+1 marker adds 1 city limit and 1 town count to assigned train each OR; reassigned each OR
- [ ] **7.13** Minor M (CIWL): comes with 10 Pullman cars; see Pullman section

---

## 8. Private Special Abilities

**Status: descriptions only.** All 10 privates have rulebook `desc:` text in `entities.rb`
but zero functional `abilities:` entries — the engine will not trigger or enforce any
private ability during play.

- [ ] **8.1** Wien Südbahnhof: places RR station token on map for free during any RR's Place Token step (token must be reachable; can pay White Cliffs Ferry token cost)
- [ ] **8.2** Barclay, Bevan, Barclay & Tritton (choose one exercise): re-set par value of owned RR with par marker; OR reserve one RR's share; OR prevent one RR's stock marker from moving DOWN for rest of SR
- [ ] **8.3** Star Harbor Trading Company: places token in port city (doesn't consume position; token = public/private port for owning RR; reduces sea zone city limit by 1/2; doesn't block city)
- [ ] **8.4** Central Circle Transport Corporation: places token in non-port city (counts as town; revenue £10/£20/£40/£60 by phase; doesn't consume token position)
- [ ] **8.5** White Cliffs Ferry: at Phase 5 start, place RR token on Lille (prohibited position); enables ferry crossing; ferry open to all once White Cliffs token removed (national formation)
- [ ] **8.6** Hochberg Mining & Lumber: places token in rough terrain hex; only owner's RRs use tracks with token; removal = original terrain cost + 1 tile point (nationals skip terrain cost)
- [ ] **8.7** Brandt & Brandau, Engineers: 4 tokens; up to 2 per OR; yellow tile placed at no cost (counts toward tile points); only owner's RRs use tracks with token; last token → private closes
- [ ] **8.8** Swift Metropolitan Line: protects one 2+2 train; owner designates RR; protected train doesn't count against limit; cannot run same track as other RR trains; cannot be sold; transfers with merged minor

---

## 9. Tile Definitions — Status and Gaps

**Confirmed tile point costs (Tables chart):** Yellow lay = 1pt (normal hex) / 2pt (metropolis); Upgrade = 2pt (normal) / 4pt (metropolis).

**Confirmed double-town tile quantities (Tile Manifest chart):**
- Yellow double towns: 3 orientations — qty 4, 6, 2 → **implemented as OE1, OE2, OE3** ✓
- Green double towns: 3 orientations — qty 3, 3, 3 → OE9–OE11, commented out (path orientations unknown)
- Brown double towns: 3 orientations — qty 3, 2, 6 → OE20–OE22, commented out (path orientations unknown)

**OE tile implementation state:**
- OE1–OE8:   yellow — ✓ defined (double-towns OE1–3; special city tiles OE4–8)
- OE9–OE11:  green double-towns — commented out; quantities known (3/3/3), path edges unknown
- OE12–OE18: green special city tiles — ✓ defined
- OE19:      **missing entirely** — not defined and not commented out; tile type unknown
- OE20–OE22: brown double-towns — commented out; quantities known (3/2/6), path edges unknown
- OE23–OE33: brown special city tiles — ✓ defined
- OE34–OE44: gray special city tiles — ✓ defined

**Standard tile quantities needing verification:**
- Tiles `'8'` and `'9'` currently set to qty 99 (placeholder) — need actual manifest values
- All other standard tile quantities (3, 4, 5, 6, 7, 12, 13, 57, 58, 80–83, 141–147, 201–208, 544–546, 621–622) need verification against physical manifest

- [ ] **9.1** OE9–OE11: green double-town tiles — provide path edge orientations to uncomment (qty 3/3/3 already known)
- [ ] **9.2** OE20–OE22: brown double-town tiles — provide path edge orientations to uncomment (qty 3/2/6 already known)
- [ ] **9.3** OE19: identify tile type and define (gap in the OE numbering between OE18 green and OE20 brown)
- [ ] **9.4** Verify standard tile quantities against physical manifest; fix placeholders for tiles 8 and 9
- [ ] **9.5** Audit all OE-specific tile codes and upgrade paths against manifest once full manifest is available

---

## 10. Route & Revenue Rules

- [ ] **10.1** Cross-water costs (confirmed from Tables chart):
  - Track: Ferry = +£5 × distance; Sea = +£10 × #sea zones
  - Tokens: Ferry = +£20 × distance; Sea = +£40 × #sea zones
- [ ] **10.2** Ferry mechanics: distance (number in oval) counts against city limit; ferry track is public (any RR); enemy tokens block public port city
- [ ] **10.3** Port authority markers: each reduces sea zone + ferry distance count by 2 (for both track and tokens); 16 total (8 North Sea, 8 Mediterranean)
- [ ] **10.4** Port types: public (any RR, anchor in light blue circle), private (anchor in red circle, restricted to token-owning RR)
- [ ] **10.5** Offshore port mechanics: train connects directly to city regardless of intervening hexes
- [ ] **10.6** Channel passages: adjacent sea zone indicators (Copenhagen, Constantinople areas)
- [ ] **10.7** Local train town counting: local trains count towns up to train level number (additional towns = level; max total = level)
- [ ] **10.8** Combined train OE runs: levels ≤4 combine; combined level = sum; city limit = combined level; towns = combined level

---

## 11. Consolidation Phase

- [x] **11.1** Trigger — `or_set_finished` sets `@consolidation_triggered` when `@phase.name.to_i >= 5`; `next_round!` routes to `Round::G18OE::Consolidation` before the next SR (once only; `@consolidation_done` prevents re-entry)
- [~] **11.2** All remaining minors/regionals must merge or be abandoned — `Round::G18OE::Consolidation` identifies affected players and runs `Step::Consolidate`; step currently accepts only `pass` (merge/abandon actions TBD)
- [ ] **11.3** Cannot pass if owning unfloated minors/regionals *(step scaffold; enforcement deferred)*
- [ ] **11.4** Conditional merger: solicit offers from eligible majors/nationals; player chooses; abandon if no offers *(deferred)*

---

## 12. Token Transfer Between Majors

- [ ] **12.1** During major's Transfer Tokens step in OR: same player may transfer token between their two majors
- [ ] **12.2** Cost: token cost from charter (of paying major) + token cost from receiving major's same zone + transfer fee (Normal City £20, Grand City £40, Metropolis £60)
- [ ] **12.3** Selling token: returns to charter at highest-cost open position

---

## 13. Minor Merger Rules (Refinements)

- [ ] **13.1** Merge minor into national: national cannot inherit minor's special ability (except stock-type); minor is abandoned after assets transferred
- [ ] **13.2** Cash transfer: minor's cash goes to major only if major has treasury stock available; otherwise forfeited to bank
- [ ] **13.3** Stock exchange: minor owner gets 1 share from major's treasury; if none, from Open Market; if neither, can only merge if hypothetical train connects minor's token to major's token (majors) or to city in national's home zone
- [ ] **13.4** One merger per major/national per SR

---

## 14. End Game Rules

- [ ] **14.1** Bank break before Level 8: finish current OR only (if during SR: play one more OR)
- [ ] **14.2** Level 8 available path: add remainder cash to bank; finish current OR; one SR; two final ORs
- [ ] **14.3** Final two ORs — second OR: each RR pays same revenue as first OR (no recalculation); no track/token/train actions; exception if new train acquired in first OR
- [ ] **14.4** Win condition: highest (cash + stock at share value + face-up privates at face value)

---

## 15. Concession Railroad Phase (DEFERRED — Out of Scope)

The Concession Railroad Phase (Game Phase 2 per the rulebook) is explicitly deferred and will NOT be implemented in the current development pass.

**What it involves:** After the Auction Phase ends, 10 "Concession Railroad" float actions occur in numbered order (CON1 first, CON10 last). Each concession holder pays 2× par to float their regional or major corporation. This phase gates the transition into the Regional/Minor Phase.

**Reason for deferring:** Implementing the concession phase requires a distinct round type, queue management, and round sequencing that significantly increases complexity without enabling playable testing of other mechanics. The current implementation skips directly from the Auction Phase to the Regional/Minor Phase (or equivalent).

- [ ] **15.1** Define Concession round type with ordered float actions (CON1–CON10)
- [ ] **15.2** Concession cards (CON1–CON10): £100 face value; wire to specific regional/major home tokens and par values
- [ ] **15.3** Float obligation: concession holder must pay 2× par; if unable, president's obligation transfers to next bidder (or concession becomes open)
- [ ] **15.4** Round sequencing: Auction → Concession → Regional/Minor Phase (update `next_round!` in game.rb)
- [ ] **15.5** 2-player without-concessions variant: skip concession phase entirely (already reflected in STARTING_CASH)

---

## 16. Tests

- [ ] **16.1** Basic game flow test (auction → regional/minor phase → major phase)
- [ ] **16.2** Train phase transition tests (rusting, limit changes, national formation trigger)
- [ ] **16.3** Stock market movement tests (right/left/up/down triggers, edge cases)
- [ ] **16.4** National revenue calculation test
- [ ] **16.5** Orient Express bonus test (first run, subsequent runs)
- [ ] **16.6** Minor ability transfer tests
- [ ] **16.7** Pullman car revenue test (nationals inherent Pullman)

---

## 17. Scenario / Variant Specifics

- [ ] **17.1** Confirm UK-FR variant entities match rulebook (4 minors: C, H, K, M; 7 regionals: BEL, GSWR, GWR, LNWR, MIDI, OU, PLM)
- [ ] **17.2** UK-FR variant train rusting rules (variant-specific, partially referenced in code)
- [ ] **17.3** Validate UK-FR map hex definitions against rulebook map
- [ ] **17.4** Other scenarios (medium/short) — reduced RR counts, modified OE destination cities, different bank sizes

# 18OE — Map Implementation Questions (mapquest.txt)

Everything needed to implement `g_18_oe/map.rb` — the full European base game map.
Items marked ✓ are already known / implemented; items marked ? require information
from the physical map (zoomed images needed).

---

## Status: map.rb created and game-startable (2026-04-11); station markers complete (2026-04-11)

`g_18_oe/map.rb` now exists and the full game initialises cleanly (Phase 2,
all 36 corporations: 12 minors + 24 regionals). Sections 1, 4, and partial
terrain are fully resolved. Remaining open questions are in sections 2, 3, 5–12 below.

## What Is Already Known / Implemented

- Map orientation: `LAYOUT = :pointy`, `AXES = { x: :number, y: :letter }`
- Coordinate system: rows A–Z, then AA, BB, CC, DD, EE, FF, GG, HH (34 rows total)
- All 24 regional home city coordinates (from 18oe_regionals.csv + interactive session)
- UK and France/Belgium national region hex lists (NATIONAL_REGION_HEXES in game.rb)
- All 258 LOCATION_NAMES entries — UK/FR/SP complete; all eastern cities and off-board names imported
- Sea zone label hexes: E34=North Sea, G60=Baltic Sea, Z55=Adriatic Sea, Y84=Black Sea
- All custom OE tile codes OE1–OE8, OE12–OE19, OE23–OE44 (defined in game.rb)
- All standard track tile quantities (TILES in game.rb)
- Constantinople: **on-map**, pre-printed yellow hex at **AA82**; Orient Express eastern terminus
- Metropolis hex list confirmed: A56, B41, C74, F87, K26, M28, M50, Q30, R55, Y14, AA82, BB51
- All six sea zones confirmed: North Sea, Baltic, Mediterranean, Adriatic, Aegean, Black Sea
- Map shape: irregular; columns 0–88, rows A–HH
- Blue hex coverage: 666 hexes — full grid covered, no undefined gaps
- Terrain: UK/Ireland/France/Spain/Portugal/Scandinavia/Alps/Italy/Adriatic all done
- **Station markers**: all 258 named locations now have `city=revenue:0` geometry (131 hexes
  corrected — 96 blank-white→city, 35 terrain-only→city+terrain, 5 border hexes fixed);
  all markers are `city=revenue:0` placeholders pending physical map confirmation of which
  should be towns, double-towns, or cities with non-zero starting revenues

### ✓ Section 4 — Red (Off-Board) Hex Coordinates (fully resolved — implemented in map.rb)

All 22 red hex coordinates confirmed and implemented in map.rb red: section:

Revenue values and path edge numbers still TBD (all set to placeholder 0 in map.rb).

### ✓ Terrain — UK/Ireland/France/Spain/Portugal/Scandinavia/Alps/Italy/Adria (implemented in map.rb)

Terrain costs applied to map.rb for:
- UK, Ireland, France, Spain, Portugal (terrain.csv lines 3–79)
- Scandinavia (terrain.csv lines 82–106)
- Alps / Italy / Adriatic region (terrain.csv lines 108–153)
- Highest Alpine passes confirmed at £120 (T39–T47, U44; rows T/U in central Alps)

---

## Section 2 — City and Town Types

The engine distinguishes: town (dot), normal city (single circle), grand city (double
circle), and metropolis (special large hex or multi-city tile).

**Status (2026-04-11)**: All 258 named locations now have `city=revenue:0` geometry in
map.rb — station circles are visible on the map. The questions below remain open for
final values. Needed for each:
- Correct type: town / normal city / grand city / metropolis (currently all `city`)
- Starting revenue (currently all `revenue:0`)
- Label letter if any (A/Y/L/P/etc.)
- Any that should be `town=revenue:0` or double-town instead of city

---

## Section 3 — Pre-printed Tiles (Yellow / Green / Gray)

Known pre-printed tiles (all pending edge number confirmation from zoomed map):

    J25  Liverpool    — yellow city=revenue:30, label=Y
    J27  Manchester   — yellow city=revenue:20, mountain:30 terrain
    M28  London       — yellow metropolis city=revenue:30, water:30 terrain
    AA82 Constantinople — yellow metropolis; revenue and edges TBD

? **3.1** Confirm edge numbers for all pre-printed yellow tiles above.
       Are there any additional pre-printed yellow hexes in central/eastern Europe?
       (Frankfurt? Warsaw? Wien area?)

? **3.2** List all hexes with pre-printed green tiles. For each:
       coordinate, revenue value, track connection edges, label.

? **3.3** List all hexes with pre-printed gray tiles. Same fields.

? **3.4** Are there any hexes with pre-printed brown tiles?

---

## Section 5 — Terrain Costs

Terrain implemented so far (see terrain.csv for full list):
- UK / Ireland / France / Spain / Portugal: ✓ done
- Scandinavia: ✓ done
- Alps / Italy / Adria: ✓ done (S38–DD55, EE52, FF51, GG50)
- Cost range confirmed: £30 / £45 / £60 / £120

? **5.1** Remaining terrain regions — fill in terrain.csv for each:
       - **Carpathians**: mountain hexes east of Wien through Slovakia/Romania
         (approx rows R–X, columns 55–75)
       - **Balkans**: Yugoslavia/Greece mountain terrain
         (approx rows X–GG, columns 55–80)
       - **Caucasus**: east of Black Sea
         (approx rows AA–EE, columns 80–88)
       - **River crossings**: Rhine, Elbe, Danube, Vistula, Dnieper water hexes
         (water terrain £30–£45 along major rivers in Germany / Eastern Europe)

? **5.2** Are there any hexes with combined terrain (e.g. mountain + water at same hex)?
       These need a combined tile definition: `city=revenue:0;upgrade=cost:N,terrain:type`

---

## Section 6 — Sea Zones (Blue Hexes)

All six confirmed present. For each zone, needed:
- All hex coordinates
- Which edges (0-5) have water borders
- Any ferry/junction path connections
- Distance oval number (for crossing costs: Track +£10/zone, Token +£40/zone)

? **6.1** North Sea — hex list, borders, ferry connections, distance number
? **6.2** Baltic Sea — hex list, borders, ferry connections, distance number
? **6.3** Western Mediterranean — eastern extension beyond UK-FR area
? **6.4** Adriatic Sea — hex list, borders, ferry connections, distance number
? **6.5** Aegean Sea — hex list, borders, ferry connections, distance number
? **6.6** Black Sea — hex list, borders, ferry connections, distance number

---

## Section 7 — National Zone Boundaries (NATIONAL_REGION_HEXES)

UK and FR already defined in game.rb. Six zones still needed:

? **7.1** PHS zone (Prussia/Holland/Switzerland) — complete hex list
? **7.2** AH zone (Austria-Hungary) — complete hex list
? **7.3** IT zone (Italy) — complete hex list
? **7.4** SP zone (Spain/Portugal) — complete hex list
? **7.5** RU zone (Russia) — complete hex list
? **7.6** SC zone (Scandinavia) — complete hex list

Note: Belgium appears to overlap PHS/FR boundary — confirm which zone.

---

## Section 8 — Ports and Ferry Routes

? **8.1** List all port city hexes (land hexes adjacent to sea zones with port symbol):
       coordinate, port type (public light-blue anchor / private red anchor)

? **8.2** List all ferry route paths through sea zones:
       start hex+edge, end hex+edge, distance number

? **8.3** North Sea port authority positions (8 markers) — which hexes?
       Mediterranean port authority positions (8 markers) — which hexes?

? **8.4** White Cliffs Ferry token slot near Lille (N31) — confirm position and
       how it is encoded in the full-game map vs UK-FR map.

---

## Section 9 — Additional Location Names

All currently known names are implemented in map.rb LOCATION_NAMES. Still needed:

? **9.1** Confirm Venezia/Venice coordinate (expected ~V47 — not yet in map.rb).
       Any other named locations visible on zoomed eastern map images not yet listed?

---

## Section 10 — Missing Custom Tile Codes (OE9–OE11, OE20–OE22)

Double-town tiles commented out in game.rb — orientations unknown.

? **10.1** Path orientations for yellow double-town tiles:
       OE9 = edges ? and ?
       OE10 = edges ? and ?
       OE11 = edges ? and ?

? **10.2** Path orientations for brown double-town tiles:
       OE20 = edges ? and ?
       OE21 = edges ? and ?
       OE22 = edges ? and ?

? **10.3** Do OE9–OE11 upgrade directly to OE20–OE22 (same orientation)?

---

## Section 11 — Patronage Tiles

? **11.1** Which cities receive patronage tiles? Fixed list or any city?

? **11.2** Are patronage tiles represented as hex icons in map.rb, or pure game logic?

---

## Section 12 — Off-Board Revenue Values and Path Edges

All 22 red hex coordinates are implemented. Still needed for each:
- Revenue by phase: yellow / green / brown / gray values
- Which edge numbers the track stubs connect to

Highest priority (affect routing immediately):
- Scottish Highlands D25 (partial: yellow_20|green_40|brown_50 guessed; verify)
- Moskva E88/F87/G88 (major destination; all three hexes share same off-board)
- New York N1 (major western destination)
- Lisboa Z1 (2-slot special; RCP home)
- Levant BB87, Alexandria GG88/HH87 (Orient Express area)

---

## Open Issues Summary

| Priority | Item | Needed |
|---|---|---|
| High | §2 City types | Revenue values for all hexes (placeholder 0); type distinction applied from CSV |
| High | §3.1 Pre-printed edges | Edge numbers for J25, J27, M28, AA82 |
| High | §5.1 Terrain remaining | Carpathians, Balkans, Caucasus, rivers |
| High | §12 Off-board revenues | Phase revenues + path edges for all 22 red hexes |
| Medium | §6 Sea zones | All blue hex coordinates, borders, distances |
| Medium | §7 National zones | PHS, AH, IT, SP, RU, SC hex lists |
| Low | §8 Ports/ferries | Port markers, ferry routes, distances |
| Low | §10 Tile orientations | OE9–11, OE20–22 edge pairs |
| Low | §11 Patronage | Fixed list or game logic only |

---

_Created: 2026-04-07 | 
Updated: 2026-04-11 — station markers added to all 258 named locations; city/town/double-town distinction applied from cities.csv; labeled cities (Y/A/B/N/S) added; revenue values still placeholder_


lib/engine/game/g_18_oe/round/consolidation.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        def name
+          'Consolidation Round'
+        end
+
+        def select_entities
+          # Players who own at least one minor or regional corporation
+          @game.players.select do |p|
+            p.shares.map(&:corporation).any? { |c| %i[minor regional].include?(c.type) }
```
crericha: Is it any player who owns a share or the player who owns the president share?

lib/engine/game/g_18_oe/round/consolidation.rb
```diff
@@ -0,0 +1,36 @@
...
+          end
+        end
+
+        def next_entity!
```
crericha: `next_entity!` and `finished?` methods look off to me. If I remember the game correctly, the round keeps happening until all remaining regionals grow up? If so, check out round/stock.rb for an example of how to do that. What's here now will end after everyone acts once.

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -7,13 +7,34 @@ module Game
     module G18OE
       module Step
         class BuyTrain < Engine::Step::BuyTrain
-          def can_entity_buy_train?(_entity)
+          def can_entity_buy_train?(entity)
```
crericha, line 10: This method returns a bool, so you can simplify to one logic statement.

`entity.corporation? && entity.type != :minor`

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -7,13 +7,34 @@ module Game
...
-          # And nationals getting rusted trains
+          # Override buyable_trains to enforce:
+          # (a) 2+2 obligation: only 2+2 available while corp has no trains and Phase < 4
+          # (b) depot level gating: all trains at the cheapest level must sell out
+          #     before the next level becomes available
+          def buyable_trains(entity)
+            trains = super
+
+            # (a) 2+2 obligation window: only 2+2 purchasable
+            return trains.select { |t| t.name == '2+2' } if entity.trains.empty? && @game.phase.name.to_i < 4
```
crericha: This doesn't look right per my understanding of the rules. Each entity is reserved one 2+2 train and must purchase it on their first operation, even if they converted to a major before operating (prior to train phase 4). 

Additionally, when the major phase starts, which is usually (always?) before phase 4, an entity can buy across, leaving another one trainless. Under this logic, it'd force a 2+2 buy for the trainless company, which is incorrect, since it already bought its 2+2.

crericha: I think you'll need a way to track if an entity fulfilled its 2+2 obligation. And only offer them the 2+2 if they haven't.

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -7,13 +7,34 @@ module Game
...
+
+            # (b) depot level gating: filter to cheapest depot level only
+            depot_trains = trains.select(&:from_depot?)
+            unless depot_trains.empty?
+              min_price = depot_trains.map(&:price).min
+              trains = trains.reject { |t| t.from_depot? && t.price > min_price }
+            end
```
crericha: I'm pretty sure this logic already exists

lib/engine/game/g_18_oe/step/consolidate.rb
```diff
@@ -0,0 +1,48 @@
...
+            return [] if pending_corps(entity).empty?
+
+            ACTIONS
+          end
+
+          def description
+            'Consolidate or abandon minors/regionals'
+          end
+
+          def pass_description
```
crericha: I don't think you'll have a pass. Either you have a minor/regional and you must act or your skipped because you have nothing to do.

lib/engine/game/g_18_oe/step/home_token.rb
```diff
@@ -12,7 +12,14 @@ def process_place_token(action)
...
 
             token.price = G18OE::Game::TRACK_RIGHTS_COST[region] || 0
-            @game.minor_available_regions.delete_at(@game.minor_available_regions.index(region))
+
+            # TEMPORARY WORKAROUND: region is nil when hex falls outside all defined zones
+            # (NATIONAL_REGION_HEXES_COMPLETE = false). Guard delete_at against nil index
+            # so placement in undefined zones doesn't raise. Remove guard once all 8 zone
+            # hex lists are complete and NATIONAL_REGION_HEXES_COMPLETE is flipped to true.
+            idx = @game.minor_available_regions.index(region)
+            @game.minor_available_regions.delete_at(idx) if idx
```
crericha: If the region isn't available, don't you want to raise? This seems to prevent it.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +617,57 @@ def metropolis_tile?(tile)
         end
 
         def must_buy_train?(entity)
-          # must buy the reserved 2+2, otherwise only majors must buy trains
-          # return unless entity.type == 'major'
-          return false if depot.depot_trains.first&.name != '2+2' && entity.type != :major
+          return false unless entity.corporation?
```
crericha: Do you have any non-corporation entities? Note: corporation is just a class, corporation type var determines if its a minor, regional, major.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +617,57 @@ def metropolis_tile?(tile)
...
+          return false if entity.trains.any?
+          # 2+2 obligation waived once Phase 4 starts
+          return false if @phase.name.to_i >= 4
```
crericha: I think the obligation is have you bought a 2+2 before or has phase 4 happened

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +617,57 @@ def metropolis_tile?(tile)
...
+
+          entity.floated?
+        end
+
+        # Inter-corp train sales only allowed in Major Phase (Phase 4+)
+        def can_buy_train_from_others?
+          @phase.name.to_i >= 4
```
crericha: Can't major phase start during train phase 3?

crericha: Also, most games implement buy across via the @phases, so it appears in the info tab.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +617,57 @@ def metropolis_tile?(tile)
...
+
+        # UP movement at end of SR: only for majors and nationals that are fully player-held
+        def sold_out_increase?(corporation)
+          %i[major national].include?(corporation.type)
+        end
 
+        # Set consolidation trigger at the end of the first OR set played under Phase 5+
+        def or_set_finished
           super
+          @consolidation_triggered ||= (@phase.name.to_i >= 5)
```
crericha: There is a way to set events on train purchases, so will give a hook to set the consolidation trigger. And it'll show up in the info tab then.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +617,57 @@ def metropolis_tile?(tile)
...
+        end
+
+        def next_round!
```
crericha, line 645: Its much easier to implement this method as a case statement. See other games.

Witzman:
## A note on this PR

I apologise for any confusion caused by the state of this pull request. I'm relatively new to collaborative Git workflows and made some structural mistakes with how I organised my branches. Thank you for your patience.

---

### What happened

When I originally set up this branch (`18oe_features`), I had already branched `18oe_mapupdates` off it — and later also `18oe-nationals`. During that time I kept making changes across all three branches without a clear separation of concerns, which caused logic that belongs here to end up in the downstream branches instead.

### What was fixed

The following changes have now been moved **into this branch** where they belong:

- **`round/consolidation.rb`**: Changed base class from `Engine::Round::Base` to `Engine::Round::Stock`, which is the correct parent for the Consolidation round.
- **`game.rb` — phase status flags**: Added `status: ['train_obligation']` to phases 2–3 and `status: ['can_buy_trains_from_others']` to phases 4–8, replacing hardcoded phase-number comparisons.
- **`game.rb` — Level 5 train event**: Added `events: [{ 'type' => 'consolidation_triggered' }]` to the Level 5 train entry, so the consolidation round is triggered via the standard event system.
- **`game.rb` — `must_buy_train?`**: Refactored to use `@phase.status.include?('train_obligation')` instead of comparing `@phase.name.to_i`.
- **`game.rb` — `can_buy_train_from_others?`**: Refactored to use `@phase.status.include?('can_buy_trains_from_others')` instead of comparing `@phase.name.to_i`.
- **`game.rb` — `event_consolidation_triggered!`**: Replaced the `or_set_finished` hook with a proper event handler method.

### Effect on this PR

This PR now contains all features-related logic cleanly. The downstream PRs (`18oe_mapupdates` and `18oe-nationals`) have been rebased on top of this branch, so their reviewers will no longer see these changes duplicated in their diffs.

lib/engine/game/g_18_oe/round/consolidation.rb
Witzman: It should be the player who owns the president's share (i.e., controls the corporation), not just any shareholder.

  - Minors are 100% owned by one player via the charter, so they're unambiguous.
  - Regionals have a 50% president cert + two 25% secondaries. Only the president controls and can initiate a merge.

 The current code using `p.shares.map(&:corporation).any?` incorrectly gives a turn to any player holding even a 25% secondary regional share.
 It should `use c.president?(p)` instead.

Consolidation Phase rules (§10.6):
"If a player owns a major that has not yet floated, the player may not select the pass option during the SR until the major has floated."

  This implies players with unfloated majors also need to act during the Consolidation Round — they must float their major before they can
  pass.

  So select_entities should include players who:
  1. Are president of at least one minor or regional (to merge/abandon them), OR
  2. Are president of an unfloated major (forced to float it)

  The fix would be:
```
  @game.players.select do |p|
    @game.corporations.any? do |c|
      (%i[minor regional].include?(c.type) && c.president?(p)) ||
      (c.type == :major && !c.floated? && c.president?(p))
    end
  end
```

lib/engine/game/g_18_oe/round/consolidation.rb
Witzman: correct. The current overrides are broken:

  - next_entity! stops at the last entity index — one pass through players, done.
  - finished? ends on !active_step — could terminate mid-round.

  Per §10.6, the Consolidation Round must continue until all remaining minors and regionals are merged or abandoned — players cycle repeatedly
   until that's resolved.

  The base Stock#finished? already handles cycling correctly: it loops until @entities.all?(&:passed?). The consolidation step should
  auto-pass a player once they have no more unresolved minors/regionals to act on.

  The fix: remove both overrides and let the base Stock round cycle naturally. The consolidation step handles when each player is done
  (passes). The only thing the Consolidation round needs to add is a custom finished? that checks both that all players have passed AND all
  minors/regionals are resolved — or just trust the step to block passing while work remains.

Fix:
Remove both next_entity! and finished? overrides entirely from consolidation.rb. The class becomes:

```
class Consolidation < Engine::Round::Stock
    def self.short_name
      'C'
    end

    def name
      'Consolidation Round'
    end

    def select_entities
      @game.players.select do |p|
        @game.corporations.any? do |c|
          c.president?(p) &&
            (%i[minor regional].include?(c.type) ||
              (c.type == :major && !c.floated?))
        end
      end
    end
  end
```

  The inherited Stock#next_entity! cycles through @entities repeatedly. The inherited Stock#finished? returns true only when
  @entities.all?(&:passed?). The consolidation step's pass_description/blocks? logic controls when each player is allowed to pass — blocking
  them while they still have unresolved minors/regionals or an unfloated major. Once a player has nothing left to do, the step auto-passes
  them. When all players are passed, the round ends.

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: require_relative '../../../step/buy_train'

  module Engine
    module Game
      module G18OE
        module Step
              entity.corporation? && entity.type != :minor
            end
...

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: this was overtaken from oeukfr - havent worked on this yet. added to openpoints list 

`87 -- [x] **3.1** Reserved 2+2 train — `must_buy_train?` in game.rb returns true while corp has no trains and Phase < 4; `buyable_trains` in
    -step restricts to 2+2 only during that window
 87 +- [~] **3.1** Reserved 2+2 train — `buyable_trains` restricts to 2+2 only during obligation window, but implementation is buggy:
 88 +  - **Bug**: uses `entity.trains.empty?` to detect unfulfilled obligation; a company that bought its 2+2 then sold it across (buy-across)
    + is incorrectly restricted to 2+2 again
 89 +  - **Bug**: uses `@game.phase.name.to_i < 4` (hard-coded phase integer — coding guidelines Issue 1); must use `@game.phase.status.includ
    +e?('train_obligation')` instead
 90 +  - **Fix needed**: add `@fulfilled_train_obligation = Set.new` in `game.rb#setup`; record entity in set in `process_buy_train` when firs
    +t train purchased (`entity.trains.one?`); replace `entity.trains.empty?` check with `!@game.fulfilled_train_obligation.include?(entity.id
    +)`
`

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: Depot#available_upcoming_trains already handles this:

  def available_upcoming_trains
    [@upcoming.first,
     *@upcoming.select { |t| @game.phase.available?(t.available_on) }].compact.uniq(&:name)
  end

  @upcoming.first naturally returns only the cheapest current level — higher levels stay locked until the current level sells out. The depot
  already gates by level before super even returns its train list.

  The custom block in buyable_trains:
  depot_trains = trains.select(&:from_depot?)
  unless depot_trains.empty?
    min_price = depot_trains.map(&:price).min
    trains = trains.reject { |t| t.from_depot? && t.price > min_price }
  end

  ...is dead weight — it re-filters something the depot already filtered.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: right. The step's logic contradicts itself:

  - blocks? returns true when the entity has pending corps — meaning they must act
  - But actions offers pass — letting them skip out anyway
  - pass_description and process_pass shouldn't exist at all

  The correct model:
  - Pending corps exist → blocks? = true, player must act (merge or abandon); no pass available
  - No pending corps → actions returns [], step doesn't block, player is auto-skipped by the round

  Also pending_corps has the same bug discussed earlier — uses entity.shares.map(&:corporation) instead of checking presidency. Should be:

  def pending_corps(entity)
    @game.corporations.select { |c| %i[minor regional].include?(c.type) && c.president?(entity) }
  end

  The cleaned-up step removes ACTIONS, pass_description, and process_pass. actions returns the real merge/abandon actions (empty array for now since those are deferred), and blocks?
  enforces that the player can't move on while pending corps remain:
```

  ACTIONS = [].freeze  # merge/abandon actions to be added

  def actions(entity)
    return [] if pending_corps(entity).empty?
    ACTIONS
  end

  def blocks?
    !pending_corps(current_entity).empty?
  end

```

lib/engine/game/g_18_oe/step/home_token.rb
Witzman: .delete returns nil silently if the region isn't found. That masks two
  real error cases:

  1. region is nil — hex isn't in any defined zone (bad coordinate or map bug)
  2. region is valid but already fully allocated — shouldn't be possible but would silently corrupt state

  Both should raise, not silently no-op. The fix adds guard clauses before the delete:

    hex = action.city.hex
    region = @game.class::CITY_NATIONAL_ZONE[hex.coordinates] ||
             @game.class::NATIONAL_REGION_HEXES.find { |_, hexes| hexes.include?(hex.coordinates) }&.first

    raise GameError, "#{hex.coordinates} is not in any track rights zone" unless region
    raise GameError, "Region #{region} is not available" unless @game.minor_available_regions.include?(region)

    token.price = @game.class::TRACK_RIGHTS_COST[region] || 0
    @game.minor_available_regions.delete(region)
    @game.minor_floated_regions[action.entity.id] = region

    super
  end

  Also note: the old version of the file had the NATIONAL_REGION_HEXES_COMPLETE feature-flag comment (guideline Issue 5) — already gone in the next PR version, which is correct.

lib/engine/game/g_18_oe/game.rb
Witzman: - entity.corporation? is a class check — is this object a Corporation instance?
  - entity.type (:minor, :regional, :major, :national) is the type discriminator

  In must_buy_train?, the entity is always a corporation — only corporations operate in an OR. Players, the bank, and the depot never reach this method. So return false unless
  entity.corporation? is always true and the guard is redundant noise.

  The meaningful check is entity.type — e.g. :minor vs :major. The entity.corporation? line should be removed entirely. The method becomes:

```
  def must_buy_train?(entity)
    return false unless entity.trains.empty?
    return false unless @phase.status.include?('train_obligation')

    entity.floated?
  end

```

lib/engine/game/g_18_oe/game.rb
Witzman: 2+2 not touched in detail , added to openpoint list

lib/engine/game/g_18_oe/game.rb
Witzman: 1. Logic bug:
  The Major RR Phase and Train Phase are independent progressions — the Major Phase can begin during Train Phase 3. @phase.name.to_i >= 4 ties inter-corp train sales to the Train Phase
  number, which is wrong. 

need some testgaming later

for now:

  def can_buy_train_from_others?
    @phase.status.include?('can_buy_trains_from_others')
  end

lib/engine/game/g_18_oe/game.rb
Witzman: # In TRAINS:
  { name: '5', ..., events: [{ 'type' => 'consolidation_triggered' }] }

  # In game.rb:
  def event_consolidation_triggered!
    @consolidation_triggered = true
    @log << '-- Event: Consolidation phase triggered --'
  end

The or_set_finished override can then be removed entirely if consolidation triggering is its only job.

lib/engine/game/g_18_oe/game.rb
Witzman: The current implementation uses chained if/return guards  — the codebase convention is a case @round statement, as seen in g_1867. It's cleaner and easier to extend.

  The rewrite:

```
  def next_round!
    @round =
      case @round
      when Engine::Round::Operating
        if @consolidation_triggered && !@consolidation_done
          @log << '-- Consolidation Phase --'
          new_consolidation_round
        else
          super
          return
        end
      when Round::G18OE::Consolidation
        @consolidation_done = true
        @turn += 1
        new_stock_round
      else
        super
        return
      end
  end

```

lib/engine/game/g_18_oe/round/consolidation.rb
```diff
@@ -0,0 +1,29 @@
...
+        def self.short_name
+          'C'
+        end
+
+        def name
+          'Consolidation Round'
+        end
+
+        def select_entities
+          @game.players.select do |p|
```
crericha: Like @ollybh said, I think you can include all players and the ones without actions will simply be skipped.

lib/engine/game/g_18_oe/step/buy_train.rb
crericha: Can we remove the function for now then? The implementation isn't correct, so I don't see value in keeping it.

lib/engine/game/g_18_oe/step/consolidate.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G18OE
+      module Step
+        class Consolidate < Engine::Step::Base
```
crericha: Is the consolidation round the only time you can grow a regional to a major or can you do it during a normal stock round too?

lib/engine/game/g_18_oe/step/home_token.rb
```diff
@@ -9,10 +9,14 @@ module Step
         class HomeToken < Engine::Step::HomeToken
           def process_place_token(action)
             hex = action.city.hex
-            region = G18OE::Game::NATIONAL_REGION_HEXES.select { |_key, value| value.include?(hex.name.to_s) }.keys.first
+            region = @game.class::CITY_NATIONAL_ZONE[hex.name] ||
+                     @game.class::NATIONAL_REGION_HEXES.find { |_, hexes| hexes.include?(hex.name) }&.first
 
-            token.price = G18OE::Game::TRACK_RIGHTS_COST[region] || 0
-            @game.minor_available_regions.delete_at(@game.minor_available_regions.index(region))
+            raise GameError, "#{hex.name} is not in any track rights zone" unless region
```
crericha: I'm not sure you need this GameError. imho the one below satisfies both cases.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -550,11 +573,25 @@ def ipo_name(_entity = nil)
...
+        # True once MAX_FLOATED_REGIONALS have been floated and the 6 remaining
+        # unfloated regionals have been closed. This is the correct trigger for
+        # "Major Railroad Phase" entry: conversions and secondary-share purchases
+        # become available from this point on.
+        def major_phase?
+          @regional_corps_floated >= self.class::MAX_FLOATED_REGIONALS
+        end
+
         def operating_order
           @minor_regional_order + (@corporations.select(&:floated?) - @minor_regional_order).sort
```
crericha: Maybe `@corporations.select { |c| c.type == :major }` instead of subtracting

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +629,51 @@ def metropolis_tile?(tile)
         end
 
         def must_buy_train?(entity)
```
crericha: I'm not sure you need this function. the buy_train step should handle must buying a train.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +629,51 @@ def metropolis_tile?(tile)
...
+        def next_round!
+          @round =
+            case @round
+            when Engine::Round::Operating
+              if @consolidation_triggered && !@consolidation_done
+                @log << '-- Consolidation Phase --'
+                new_consolidation_round
+              else
+                super
+                return
```
crericha: You don't need the return. The output of the last line is always automatically returned.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -583,11 +629,51 @@ def metropolis_tile?(tile)
...
+              end
+            when Round::G18OE::Consolidation
+              @consolidation_done = true
+              @turn += 1
+              new_stock_round
+            else
+              super
+              return
```
crericha: return not needed

lib/engine/game/g_18_oe/round/consolidation.rb
Witzman: The base Stock round's select_entities already returns all non-bankrupt players. The step's blocks? method controls whether a player has anything to do — if blocks?
   is false, the round auto-skips that player.

  Our custom select_entities pre-filters to only presidents of minors/regionals/unfloated-majors, which is redundant and potentially wrong — if a player's last regional gets merged
  mid-round by someone else, they'd still be in @entities and need to be handled gracefully anyway.

  The fix is simply to remove select_entities entirely from consolidation.rb:

```
  class Consolidation < Engine::Round::Stock
    def self.short_name
      'C'
    end

    def name
      'Consolidation Round'
    end
  end
```
  The inherited select_entities returns @game.players.reject(&:bankrupt) — all players. Players with nothing to act on (pending_corps empty, blocks? false) are auto-skipped by the round
  machinery. No custom filtering needed.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: Per §8.3 and §9.3, regional → major expansion is a normal SR stock action available from the start of the Major RR Phase (Phase 4) — not exclusive to the Consolidation Round.
  Any player who is president of a regional, or who owns both secondary shares, may expand it during a standard SR.

  The Consolidation Round (§10.6) is a separate mechanism: it forces resolution of whatever minors/regionals remain unmerged at Phase 5. It's the deadline, not the only opportunity.

  So the Consolidate step is correctly scoped to forced end-game resolution, but the expansion mechanic itself is currently missing from BuySellParShares — players have no way to
  voluntarily expand a regional into a major during a normal SR. That's a gap that needs to be filled in the SR step, not here.

Witzman: in BuySellParShares the implementation is already there and quite complete. The convert action handles exactly this:

  - can_convert?(entity) — checks Major RR phase, entity is regional, no prior sells, ownership requirements (president OR 50% holder who can buy in)
  - process_convert — calls float_major which restructures shares, changes type to :major, moves stock market, adds tokens
  - corporation_actions — offers convert/pass when it's a regional's turn
  - actions — adds 'convert' to player actions when can_convert_any? is true

  So regional → major expansion during normal SRs is already implemented. The Consolidate step is correctly a separate, forced-resolution-only step.

lib/engine/game/g_18_oe/step/home_token.rb
Witzman: Agree and remove the first raise.

lib/engine/game/g_18_oe/game.rb
Witzman: The replacement is cleaner and more explicit:

```
  def operating_order
    @minor_regional_order + @corporations.select { |c| c.type == :major }.sort
  end
```

 Nationals still will operate in the OR (they run trains and pay dividends). So they should also belong in operating_order i think. The fix should include them:

```
  def operating_order
    @minor_regional_order + @corporations.select { |c| %i[major national].include?(c.type) }.sort
  end
```

lib/engine/game/g_18_oe/game.rb
Witzman: Looking at step/train.rb:

```
  def must_buy_train?(entity)
    @game.must_buy_train?(entity)
  end
```

  The step already delegates to @game.must_buy_train? — so the game method is just a relay. The base Game::Base#must_buy_train? checks MUST_BUY_TRAIN == :always or :route. Our override
  replaces that with a phase-status check, which is step-level logic.

  You are right that this belongs in the G18OE BuyTrain step by overriding must_buy_train? there directly:

  def must_buy_train?(entity)
    return false unless entity.trains.empty?
    return false unless @game.phase.status.include?('train_obligation')

    entity.floated?
  end

  Then must_buy_train? can be removed from game.rb entirely, and the base game's implementation (which checks the MUST_BUY_TRAIN constant) is never called. The obligation logic lives with
  the step that enforces it, which is the right place.

lib/engine/game/g_18_oe/game.rb
Witzman: correct, thanks

lib/engine/game/g_18_oe/game.rb
Witzman: correct, thanks

Witzman:
Thanks for the detailed review — all points addressed. Summary of changes:

---

### `round/consolidation.rb` — `select_entities` removed

The custom `select_entities` pre-filtered to presidents of minors/regionals/unfloated majors, which is both redundant and fragile (a player whose last regional is merged mid-round would still be in `@entities`). The inherited `Stock#select_entities` returns all non-bankrupt players, which is correct — the step's `blocks?` method handles auto-skipping players who have nothing to do.

---

### `step/consolidate.rb` — presidency check + pass action removed

`pending_corps` was using `entity.shares.map(&:corporation)` which incorrectly included any player holding even a 25% secondary regional share. Fixed to `c.president?(entity)` — only the president controls and can initiate a merge.

The `pass` action and `process_pass` have been removed. The correct model: if pending corps exist, `blocks?` returns `true` and the player must act — offering a pass alongside that was contradictory. When a player has nothing left to do, `blocks?` is false and the round auto-skips them.

---

### `step/buy_train.rb` — simplification + `must_buy_train?` moved here

`can_entity_buy_train?` simplified from three lines to a single boolean expression:
```ruby
entity.corporation? && entity.type != :minor
```

The depot-level-gating block in `buyable_trains` has been removed — `Depot#available_upcoming_trains` already handles this; the block was re-filtering something the depot had already filtered.

`must_buy_train?` has been moved from `game.rb` into this step, where it belongs. The base step delegates to `@game.must_buy_train?` but our obligation logic is phase-status-based, which is step-level concern:
```ruby
def must_buy_train?(entity)
  return false unless entity.trains.empty?
  return false unless @game.phase.status.include?('train_obligation')
  entity.floated?
end
```
The 2+2 obligation filter in `buyable_trains` has been removed. The implementation was buggy in two ways: it used `entity.trains.empty?` to detect an unfulfilled obligation (incorrectly re-restricting a company that bought its 2+2 then had it bought across), and used a hard-coded phase integer instead of `phase.status`. Removed until properly implemented with obligation tracking (openpoints §3.1).



---

### `step/home_token.rb` — guard clauses + redundant raise removed

Replaced the nil-silent workaround (`delete_at(idx) if idx`) with proper guard clauses that raise on invalid state. The `NATIONAL_REGION_HEXES_COMPLETE` feature-flag comment has been removed.

The first `GameError` (`"#{hex.name} is not in any track rights zone"`) has been removed as redundant — the single `include?` check covers both cases: a nil region (hex not in any zone) and a valid region that is no longer available.

Added `CITY_NATIONAL_ZONE = {}.freeze` as a stub constant for future per-city zone overrides, used as the fast-path lookup before falling back to `NATIONAL_REGION_HEXES`.

---

### `game.rb` — several cleanups

- **`operating_order`**: replaced set subtraction with an explicit type filter. `@corporations.select { |c| %i[major national].include?(c.type) }.sort` is clearer and correctly includes nationals (which do operate in the OR).
- **`must_buy_train?`**: removed entirely — logic moved to the `BuyTrain` step.
- **`event_consolidation_triggered!`**: added a log message so the event appears in the info tab.
- **`next_round!`**: rewritten as a `case` statement (matching codebase convention e.g. `g_1867`). Redundant `return` statements after `super` in `else` branches removed — Ruby returns the last expression automatically.
- **`CITY_NATIONAL_ZONE`**: added stub constant (see home_token above).

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: will remove

lib/engine/game/g_18_oe/map.rb
```diff
@@ -406,7 +406,7 @@ module Map
             ['AD9'] => 'city=revenue:0;upgrade=cost:30,terrain:mountain',
             ['D57'] => 'city=revenue:0;upgrade=cost:30,terrain:water',
             ['I50'] => 'city=revenue:0;label=Y;upgrade=cost:45,terrain:water',
-            ['AA82'] => 'city=revenue:0;upgrade=cost:45,terrain:water',
+            ['AA82'] => 'city=revenue:30;city=revenue:30;upgrade=cost:45,terrain:water;label=C',
```
Kalnoch: Constantinople is 20 revenue before any tile lays

Witzman: fixed https://github.com/tobymao/18xx/pull/12508/commits/a2084a43727c1178180283ba1d1443be23ebadd4

lib/engine/game/g_18_oe/step/consolidate.rb
crericha: Okay, then I recommend adding the action to an 18oe version of buy_sell_par_shares and then inherit from that for the consolidation round, removing the buy and sell actions. That way you won't have duplicate code.

crericha (approved):

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: Good call. Implemented as suggested — `Consolidate` now inherits from `G18OE::Step::BuySellParShares` instead of `Engine::Step::Base`.

The convert logic (`float_major`, `can_convert_any?`, `process_convert`) lives in one place in `G18OE::Step::BuySellParShares` and is shared by both the normal SR and the Consolidation Round. No duplication.

`Consolidate` overrides only what differs in the Consolidation Round context:

- **`actions`**: returns only `['convert']` — no buy, sell, par, or buy_company
- **`can_convert?`**: simplified to president-only (§10.6 does not allow a non-president 50% holder to convert during Consolidation), with the sell/buy guards removed as irrelevant
- **`process_convert`**: calls `super` then auto-passes — one conversion per turn, the round cycles back if the player has further regionals to convert
- **`pass!`** from `BuySellParShares` is inherited unchanged — its president check is harmless since Consolidation is president-only, so `@converted.president?(current_entity)` is always true here

Witzman: changed regionals to majors with https://github.com/tobymao/18xx/pull/12508/commits/8d3c09cbb2f93ae2a1b8f80d6c8e0582bb4fd522

Kalnoch:
`must_buy_train?: minors exempt; obligation waived from Phase 4 onward`

Majors must always have a train at the end of the OR, even after phase 4. Might need to keep an eye on this for edge case handling

PR #12515, opened 2026-04-17
[core] [UI]  Add bankrupt player column

### LINKED ISSUE #12059: [18USA] Display eliminated players in a single column
**What's your idea?**

In this game, https://18xx.games/game/224778#entities, you can see that there are two eliminated players, and they are displayed in separate columns. I think it would be a better usage of screen space if they were displayed in a single column.

**Are there related ideas?**

**Did you join the Slack?**

ddaybell:
Fixes #12059 

Title: Display bankrupt players in a column to save horizontal space

## Summary

- On the **Entities tab**, eliminated players are grouped into a single column on the right side of the layout rather than each occupying a separate column in the main grid. Cards are shown at 50% scale to minimize screen footprint while still showing relevant info.
- On the **Game tab**, bankrupt players appear in a 50%-scaled column to the right of the active players row, aligned to the same vertical position.
- When no players are bankrupt, both layouts are unchanged.

## Test plan

- [ ] Load a finished 1817 game with multiple eliminations and verify the Entities tab shows a compact column of scaled-down bankrupt player cards on the right, with active players and their corporations filling the main grid normally
- [ ] Verify the Game tab shows bankrupt players in a column to the right of the active player row at 50% scale
- [ ] Verify a game with no bankrupt players looks unchanged on both tabs
- [ ] Also ran rubocop and test suites.

## Implementation Notes

This seems like a decent way to handle those games that can generate multiple bankruptcies.  Won't matter for those games that end on bankruptcy of course, nor for low player-count games where there won't be more than 1 or 2 bankruptcies.  But for the 12-player 1817s that show up from time to time they'll probably like it.  I don't think it's a significant change to the UI, but it is a change so it would be worth a bit of thought as to whether we want to add this kind of feature or not.  I'll leave that to those at a higher pay grade than me! :-)

### Screenshots

This is the Entities tab of a 12-player 1817 game, with many bankrupts
image

This is the Entities tab of that same test game, early in the game with no bankrupts.
image

This is the Game tab of that same test game, showing the players listed at the bottom of the tab.
image

This is the Game tab from that same game, early in the game with no bankrupts.
image

### DIFF
--- a/assets/app/view/game/entities.rb
+++ b/assets/app/view/game/entities.rb
@@ -10,13 +10,6 @@ module View
       needs :user
 
       def render
-        div_props = {
-          style: {
-            display: 'grid',
-            grid: 'auto / repeat(auto-fill, minmax(17rem, 1fr))',
-            gap: '3rem 1.2rem',
-          },
-        }
         players = @game.player_entities
         if (i = players.map(&:name).rindex(@user&.dig(:name)))
           players = players.rotate(i)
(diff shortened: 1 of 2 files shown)


crericha:
I'm personally not a fan of smaller entities and special view logic to maintain to support that size.

Why not show the bankrupt players at the end of player cards, so you don't have to scroll to them unless you want to see them?

Also, is something needed for the spreadsheet tab? I can't remember if bankrupt players show there or not.

ddaybell:
This issue isn't really relevant to the Spreadsheet.  The players cards aren't used there.  Instead the player info is presented in a table, which is already small enough that the bankrupt players aren't a problem.  

To your other question, the current behavior puts all the bankrupt players at the end of what is implemented as a long row of players (for high-count games).  That row wraps on most displays, so the bankrupts will sort to the bottom.  I don't believe our UI supports horizontal scrolling, just vertical.  So your comment is how the UI works currently.  I think the OP saw some blank space in his game and was thinking it could be more effectively used.  The more I think about this one, I'm not convinced we need it.  I think this is a solution in search of a problem... :-)

crericha:
I agree with your assessment. Let's keep it simple and close the PR. I'll close the issue.

PR #12517, opened 2026-04-17
[1822 family] skip choice for doubling cash if inconsequential

### LINKED ISSUE #10333: [1822][1822CA][1822MX] Feature Request - Automated Money Doubler Private function
Each of these games has a private that doubles player cash for turn order purposes.  In two cases, doubling player cash would not affect turn order:
1. If the player who owns the private already has the most cash of all players in the game
2. If the player who owns the private would not, even with double their current cash, overtake the next player

Perhaps it is worth implementing a script that checks
1. If the Private is in player hands
2. If the use of the private would make a difference to turn order
3. Skip the prompt asking the player if they would like to use the private if 2 is true.
This would help speed up gameplay loops because sometimes you are waiting hours for another player to sign back in just to press this one button.

kepakiano:
Fixes #10333

### DIFF
--- a/lib/engine/game/g_1822/step/choose.rb
+++ b/lib/engine/game/g_1822/step/choose.rb
@@ -46,6 +46,11 @@ module Engine
             @company = @game.company_by_id(@game.class::COMPANY_DOUBLE_CASH)
             return nil if !@company || @company&.owner != entity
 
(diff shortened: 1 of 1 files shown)


michaeljb (approved):
With the server running locally, cloning from http://localhost:9292/fixture/1822NRS/1822NRS_game_end_bank?action=315 and then passing `Player 1`'s final SR turn demonstrates the new functionality working correctly.

Looks good!

PR #12520, opened 2026-04-19
[18Royal Gorge] [pins] Game end stock reaches trigger area - finish only current OR 

### LINKED ISSUE #12518: [18RoyalGorge] Red zone end game trigger not overriding event card trigger
**What's your Game ID?**

https://18xx.games/game/248831

Action link: 

https://18xx.games/game/248831?action=535

**How to Reproduce the Bug?**

- Trigger end game by purchasing or exporting the first 6x2 train, without any company stocks in the red zone. This results in finishing current ORs, then one more SR and two more ORs, then game ends. 

- Pay out dividends on a company to increase its stock into the red zone. According to the rules [1], this should override the above trigger, ending the game immediately following the current ORs.

- Game still allows another SR followed by two more ORs.

[1] Page 18, Triggering end of game:
> **Ending 2: Stock value Cap**
> If a company’s stock marker reaches the red spaces of the stock market, players will finish the current set of operating rounds and then total their scores. This applies even if the game end was previously triggered by the purchase or export of a train.

**Which Browser / Platform?**

Any

chrstas:
Fixes #12518

If a company's stock reaches the red spaces, the game ends after the current OR set — even if the final train trigger was already active. This was broken because `GAME_END_TIMING_PRIORITY` had `one_more_full_or_set` at a higher priority than `full_or`, preventing the stock market trigger from overriding the train trigger.

Breaking change: Any saved game where the final train was already triggered and a stock subsequently reached the red zone will now end one full OR set earlier than before.

## Before clicking "Create"

## Implementation Notes

Swap the order to `%i[immediate full_or one_more_full_or_set]`, so that earlier endings have higher priority.

The new spec test mocks game_end_check_stock_market? on an existing fixture (endgame_triggered_on_buy, after the 6x2 train is bought) to simulate a stock hitting the red zone. Would you need a dedicated fixture or is this way acceptable?

Two fixtures required an update on the game end reason (`endgame_triggered_on_export.json` and `endgame_shorter_triggered_on_export.json`). In both cases corporation CM hit the stock area in the red area.

AI Usage Note: I have used AI to cross-check my finding and to support creating the game spec test.

### Screenshots

image


spec/lib/engine/game/g_18_royal_gorge/game_spec.rb
```diff
@@ -48,6 +48,19 @@
       expect(game.round.stock?).to eq(false)
       expect(game.finished).to eq(true)
     end
+
+    it 'stock_market red zone overrides final_train end trigger' do
+      game = fixture_at_action(431, clear_cache: true)
+
+      expect(game.instance_variable_get(:@game_end_trigger)).to eq(%i[final_train one_more_full_or_set])
```
michaeljb, line 55: `Game::Base` has `:game_end_trigger` in its `attr_reader` call, so this should work:
```suggestion
      expect(game.game_end_trigger).to eq(%i[final_train one_more_full_or_set])
```

spec/lib/engine/game/g_18_royal_gorge/game_spec.rb
```diff
@@ -48,6 +48,19 @@
...
+      expect(game.instance_variable_get(:@final_turn)).to eq(5)
+
+      allow(game).to receive(:game_end_check_stock_market?).and_return(true)
+      game.send(:game_end_check)
+
+      expect(game.instance_variable_get(:@game_end_trigger)).to eq(%i[stock_market full_or])
```
michaeljb, line 61:
```suggestion
      expect(game.game_end_trigger).to eq(%i[stock_market full_or])
```

spec/lib/engine/game/g_18_royal_gorge/game_spec.rb
```diff
@@ -48,6 +48,19 @@
...
+      expect(game.instance_variable_get(:@final_turn)).to eq(game.turn)
+    end
   end
 
   describe 'endgame_triggered_on_export' do
```
michaeljb, line 66: (optional) I'd like to see `endgame_triggered_on_export` (and its `shorter` fixture) renamed since the final `reason` does change in the same way, but in that fixture the game still ends at the same point so maybe it's not too important.

michaeljb (changes requested):
> The new spec test mocks game_end_check_stock_market? on an existing fixture (endgame_triggered_on_buy, after the 6x2 train is bought) to simulate a stock hitting the red zone. Would you need a dedicated fixture or is this way acceptable?

Since the existing fixtures still end at the same time, but with a different reason, I would prefer to see a new fixture added where reaching the end of the market actually does make the game shorter, as in the reported issue. My usual approach there is to just take the game from the issue and make it a new fixture, I find that a lot easier than hand crafting a new one 😄 . 

Could you name the new fixture something that references the export being the initial trigger, with the market ending cutting it short? e.g., `endgame_market_trigger_overrides_export_trigger`

Then run `make fixture_format` to make it consistent with the other fixtures in the repo.

chrstas:
> > The new spec test mocks game_end_check_stock_market? on an existing fixture (endgame_triggered_on_buy, after the 6x2 train is bought) to simulate a stock hitting the red zone. Would you need a dedicated fixture or is this way acceptable?
> 
> Since the existing fixtures still end at the same time, but with a different reason, I would prefer to see a new fixture added where reaching the end of the market actually does make the game shorter, as in the reported issue. My usual approach there is to just take the game from the issue and make it a new fixture, I find that a lot easier than hand crafting a new one 😄 .
> 
> Could you name the new fixture something that references the export being the initial trigger, with the market ending cutting it short? e.g., `endgame_market_trigger_overrides_export_trigger`
> 
> Then run `make fixture_format` to make it consistent with the other fixtures in the repo.

thanks a lot @michaeljb - first time fixtures fix for me ;-). I have created a new one based on the hotseat game and renamed the other two fixtures.

michaeljb (approved):
looks great, thanks!

PR #12523, opened 2026-04-19
[1835] implement vanilla starting package sale

kepakiano:
Refers to  #3059

This PR is my attempt at implementing the vanilla starting package sale of 1835. The change is cut out of a bigger chunk of changes I have locally in which more of 1835 is already implemented, but in order to keep the PRs digestible, I'll limit them to one or just a few features.

## Implementation Notes

The sale of the starting package is implemented as a Draft. This is only because that's what the code was before and I didn't know any better at first. It's not really a draft though the way e.g. 18CZ's start is a draft and it forces me to implement all the "bid" methods (max_bid, min_bid, process_bid etc), which seems unnecessary. Any objection to making the draft just a step in the SR which is skipped if the draft is done (and the actual SR step is skipped while the draft is not done)? 

I left the configuration of the certificates (see the setting of `double_cert` in `Game#setup`) in, because I wanted to have the Entities file complete, but the Entities file in its final form only makes sense with the double certs properly set.

### Any Assumptions / Hacks

As noted in the code, I found it weird that I had to implement `SpecialToken#available_tokens` this way, but without bending the tokening entity to `current_entity`, the game crashed saying, `tokens_by_type` was not implemented for Player X. I'm pretty sure it has to do with the Nürnberg-Fürth and Pfalzbahnen private companies which allow a corporation to lay an extra token when its owner has one of these private companies.
Since this is not material for the sale of the starting package, we can also ignore it for now. It just needed a fix or you wouldn't be able to go draft - OR - draft while testing.


lib/engine/game/g_1835/step/draft.rb
```diff
@@ -12,15 +12,41 @@ class Draft < Engine::Step::Base
...
 
           def setup
-            @companies = @game.companies.sort_by { |item| [item.revenue, item.value] }
+            @companies = @game.companies.select { |c| c.owner.nil? && !c.closed? }
+
+            # set up the tiered companies as 2d array that might contain empty arrays if the starting package was not
+            # fully sold in a previous draft round. These empty arrays are important for may_purchase: companies in rows
+            # after an empty row can only be purchased if all rows before are empty. If we were to only group_by
+            # auction_row, we might loose these empty rows.
+            @tiered_companies = Array.new(4) { [] }
```
ollybh, line 21: If you're not doing anything in the block being passed to the array constructor you can instead use the two-argument version of the new method:
```suggestion
            @tiered_companies = Array.new(4, [])
```

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -12,15 +12,41 @@ class Draft < Engine::Step::Base
...
-          def may_purchase?(_company)
-            true
+          def tiered_auction_companies
+            @tiered_companies
+          end
+
+          def may_purchase?(company)
+            # in the vanilla draft a company can only be purchased if it is either in the top-most row or furthest to
+            # the left in the second top-most row if the top-most row only has one company
+            false unless company
```
ollybh:
```suggestion
            return false unless company
```

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -12,15 +12,41 @@ class Draft < Engine::Step::Base
...
+            company_row = company.auction_row
+            return true if company_row.zero?
+            return true if @tiered_companies[0...company_row].all?(&:empty?)
+            return false unless @tiered_companies[0...(company_row - 1)].all?(&:empty?)
+            return false unless @tiered_companies[company_row - 1].size == 1
+
+            @tiered_companies[company_row][0].sym == _company.sym
```
ollybh: Generally `Array.first` is preferred to `Array[0]`. And you can compare the company objects directly, you don't need to look up their symbols.
```suggestion
            @tiered_companies[company_row].first == company
```

ollybh: Might it be clearer to do something like this:
```
first_row = @tiered_companies.index { |row| !row.empty? }
return true if company_row == first_row

(company_row == first_row + 1) && @tiered_companies[first_row].one? && (@tiered_companies[company_row].first == company)

File: lib/engine/game/g_1835/step/draft.rb
Comment (ollybh) [inline lib/engine/game/g_1835/step/draft.rb]: Are these methods needed? They seem to have been cut-n-pasted from Engine::Step::SimpleDraft. If they are needed, should this class be a subclass of SimpleDraft?

File: lib/engine/game/g_1835/step/draft.rb [hunk 4]
```diff
@@ -68,11 +105,50 @@ def process_bid(action)
             company.owner = player
             player.companies << company
             player.spend(price, @game.bank)
+            @tiered_companies.each do |row|
+              next unless row.index(company)
```
ollybh: `Array.index` returns the index of the first match. As we don't care about the index, just whether there is a match, the correct method is `Array.includes?`
```suggestion
              next unless row.includes?(company)
```

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -68,11 +105,50 @@ def process_bid(action)
...
+            company.close! if company.sym == 'BY_D'
+
+            corporation = @game.corporation_by_id(company.id)
+
+            if corporation && corporation.type == :minor
+              share = corporation.shares.first
+              @game.share_pool.transfer_shares(ShareBundle.new(share), player)
+              @game.bank.spend(price, corporation)
+              company.close!
+              @game.place_home_token(corporation)
```
ollybh: Maybe you could give each of the privates 1­–6 a shares ability that gives the share in the associated minor corporation.

lib/engine/game/g_1835/step/special_token.rb
```diff
@@ -0,0 +1,28 @@
...
+
+          # def place_token(entity, city, token, connected: nil, extra_action: nil, special_ability: nil,
+          # check_tokenable: nil, spender: nil, same_hex_allowed: nil)
+          #   entity = current_entity
+          #   super
+          # end
+          # def adjust_token_price_ability!(_entity, token, hex, _city, special_ability: nil)
+          #   _entity = current_entity
+          #   super
+          # end
```
ollybh: Dead code shouldn't be included in pull requests.

lib/engine/game/g_1835/step/special_token.rb
```diff
@@ -0,0 +1,28 @@
...
+require_relative '../../../step/special_track'
+
+module Engine
+  module Game
+    module G1835
+      module Step
+        class SpecialToken < Engine::Step::SpecialToken
+          def available_tokens(_entity)
+            # this looks a little hacky, but without this, BY cannot have its first turn
+            super(current_entity)
```
ollybh: Yes, this doesn't look right. What's going wrong with BY?

lib/engine/game/g_1835/entities.rb
```diff
@@ -102,243 +209,202 @@ module Entities
...
+            floatable: false,
+            max_ownership_percent: 100,
             shares: [10, 10, 10, 10, 10, 10, 10, 10, 5, 5, 5, 5],
-            # shares for minors and Privates should be reserved
-            coordinates: 'E19',
             color: '#37383a',
           },
-        ].freeze
-
-        MINORS = [
```
ollybh, line 229: I think that MINORS is the right place for the minor companies, not CORPORATIONS. The critical difference is that an entity represented by a Corporation object has a share price, but a Minor object does not. 1835 minors don't have share prices.

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -68,11 +105,50 @@ def process_bid(action)
...
+              next unless row.index(company)
+
+              row.delete(company)
+            end
             @companies.delete(company)
 
             @log << "#{player.name} buys #{company.name} for #{@game.format_currency(price)}"
 
+            @game.abilities(company, :shares) do |ability|
+              ability.shares.each do |share|
```
ollybh, line 99: When the BY and SX presidencies are given to players, is the share price being set?

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -60,6 +87,16 @@ def actions(entity)
...
+            current_entity.pass!
+            @round.next_entity_index!
+            action_finalized
+          end
+
+          def log_skip(entity)
+            @log << "#{entity.name} cannot afford any company and passes"
+          end
+
           def process_bid(action)
```
ollybh, line 87: This method is very long. Can it be broken up?

lib/engine/game/g_1835/game.rb
```diff
@@ -209,25 +213,68 @@ class Game < Game::Base
...
+          # once PR floats.
+          corporation_by_id('PR').shares.last(8).each { |s| s.buyable = false }
+
+          corporation_by_id('BA').shares.last.double_cert = true
+          corporation_by_id('WT').shares.last.double_cert = true
+          corporation_by_id('HE').shares.last.double_cert = true
+          corporation_by_id('MS').shares[1].double_cert = true
+          corporation_by_id('MS').shares[2].double_cert = true
+          corporation_by_id('OL').shares[1].double_cert = true
+          corporation_by_id('OL').shares[2].double_cert = true
```
ollybh: That's a lot of embedded logic. Maybe:
```
@corporations.each do |corp|
  corp.shares.reject(&:president).each { |share| share.double_cert = (share.percent == 20) }
end
```

lib/engine/game/g_1835/step/draft.rb
ollybh: OK, I see that the par prices are fixed, not chosen by players.

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -68,11 +105,50 @@ def process_bid(action)
...
+                # check whether there is someone else with more shares
+                if share.president && share.corporation.name == 'BY'
+                  share.corporation.owner = player
+                  @log << "#{player.name} becomes the president of #{share.corporation.name}"
+                end
+                @game.bank.spend(share.price, share.corporation)
+
+                # allow president change for SX when LD is being sold, but for BY only if the BY director has been sold,
+                # which might be during the current or any earlier action
+                allow_president_change = @companies.find { |c| c.sym == 'BY_D' }.nil? || share.corporation.id == 'SX'
```
ollybh: Would this work? Allow the presidency to move if this is the president's share, or if the company is already floated. That should stop the presidency of BY being assigned too early if all the single shares are exchanged.
```
@game.bank.spend(share.price, share.corporation)
allow_president_change == share.president || share.corporation.floated?
@game.share_pool_transfer_shares(share.to_bundle, player, allow_president_change: allow_president_change)
```

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -68,11 +105,50 @@ def process_bid(action)
...
+                @game.share_pool.transfer_shares(ShareBundle.new(share), player, allow_president_change: allow_president_change)
+
+                @game.place_home_token(share.corporation) if share.corporation.floated?
+              end
+            end
+
+            company.close! if company.sym == 'BY_D'
```
ollybh: To avoid hardcoding a reference to the company abbreviation, you could do this, closing the company if it's only ability was the share exchange.
```suggestion
              company.close! if company.abilities.one?

            end

```

lib/engine/game/g_1835/round/operating.rb
```diff
@@ -0,0 +1,17 @@
...
+require_relative '../../../round/operating'
+
+module Engine
+  module Game
+    module G1835
+      module Round
+        class Operating < Engine::Round::Operating
+          def description
+            "Operating Round #{@game.turn}"
+          end
```
ollybh: I don't think you want to be doing this. The round description is actually the current step's description. I think you'll find that this will result in the game description looking something like `1835: Phase 2 - Operating Round 1.1 (of 1) - Operating Round 1`

ollybh (changes requested):
From glancing at the 1835 rules, it looks like the minor companies don't have share prices. That means they should be represented by Minor objects, not Corporation objects. So I think that moving their definition from the MINORS constant to the CORPORATIONS looks like a mistake. Or have I misunderstood something?

Other than this, I've got a few suggestions for changes.

lib/engine/game/g_1835/round/operating.rb
kepakiano: I kept getting an error stating that `round_num` wasn't defined for `nil` after ending an unfinished draft round. Adding this implementation seemed to help. I only just now realized that it only helped, because whenever I got that error I had only replace one of the two occurences of "OperatingRound" in game.rb (one being in new_operating_round and one in the switch in next_round). Now that I use the generic OperatingRound in both, I don't need this subclass anymore :)

lib/engine/game/g_1835/step/draft.rb
kepakiano: that makes sense

lib/engine/game/g_1835/step/draft.rb
kepakiano: whoops, yes

lib/engine/game/g_1835/step/draft.rb
kepakiano: Actually, it doesn't work. Which is weird. 
With your suggestion the draft looks like this:

Image

With the old code it's tiered as it's supposed to be, even though irb tells me both should yield the same:

Image

lib/engine/game/g_1835/step/draft.rb
kepakiano: see below

kepakiano: beautiful, so much more concise :)

lib/engine/game/g_1835/step/draft.rb
kepakiano: The need for some of these comes from auction#render_company_actions: If the step says a company cannot be purchased (not all company can be purchased in the vanilla sale), the step gets asked about `may_choose` and if that's false too, then the step need to answer `min_bid`, `min_increment`, `max_bid` and `may_bid`. This may be a short-coming of the logic in `render_company_actions`. I'm thinking that `min_bid`, `min_increment`, and `max_bid`  should only be called, if `may_bid` returns true. But since the `input` variable is used in several spots, I'm guessing someone pulled the code up above the call to `may_bid`. I would suggest moving the formation of `input` into a function and only calling it after `if @step.may_bid?(company)`, but I don't dare make a change in the core library yet :D

Also, as stated in the opening post of this PR, I'm not against changing this from a draft to a regular sale. But maybe if the `auction` class were changed, the overhead here would be small enough to leave it as a draft.

lib/engine/game/g_1835/step/draft.rb
kepakiano: yes, makes sense

lib/engine/game/g_1835/step/draft.rb
kepakiano: certainly :)

lib/engine/game/g_1835/step/draft.rb
kepakiano: yeah, all par prices in 1835 are fixed

lib/engine/game/g_1835/step/draft.rb
kepakiano: I need the reference to BY for the edge case a few lines above. I'll move closing the company that gives the BY director there

lib/engine/game/g_1835/step/draft.rb
kepakiano: I don't think so, since BY doesn't float before 50% are sold. But, I can allow the president change if either this is the president share or the company does not own the president share anymore :)

lib/engine/game/g_1835/step/draft.rb
kepakiano: I use `@game.minor_by_id` now like some other games. Is that OK?

lib/engine/game/g_1835/step/special_token.rb
kepakiano: true, yoinked

lib/engine/game/g_1835/entities.rb
kepakiano: you're right. This was a change from way back then when I began. They're back in their own array.

lib/engine/game/g_1835/game.rb
kepakiano: yeah, that's neat :) couldn't this even be the standard? Is there any game with a cert that's twice the regular cert percentage that does not want to be marked as a double_cert?

lib/engine/game/g_1835/step/special_token.rb
kepakiano: Wasn't actually a BY problem. `available_tokens` gets called for all kinds of entities, also players, which I didn't realize. I now just return an empty array if it's not called with a corp

lib/engine/game/g_1835/entities.rb
kepakiano: Moving the minors into their own array somehow broke the home token location of 5 and 6, because they are not in city slot 0, but the defined city in `entities.rb` got ignored.
To fix this instead of asking the game class to place the home token, I do it by hand like 18Mag does:
```
hex = @game.hex_by_id(minor.coordinates)
hex.tile.cities[minor.city || 0].place_token(minor, minor.next_token)
```

lib/engine/game/g_1835/step/draft.rb
kepakiano: although rubocop complained about the line being too long, so I had to break the three conditions up again :D

lib/engine/game/g_1835/step/draft.rb
ollybh: Yeah, that was a stupid suggestion. `Array.new(4, [])` creates an array of arrays, but each item is a reference to the same array.
```
irb(main):001> a = Array.new(4, [])
=> [[], [], [], []]
irb(main):002> a.first.push(:a)
=> [:a]
irb(main):003> a
=> [[:a], [:a], [:a], [:a]]
irb(main):004> a.map(&:object_id)
=> [46860, 46860, 46860, 46860]
```

ollybh:
@kepakiano In future can you not force push a branch once a review is under way? It makes it impossible to see what has been changed in response to review comments.

kepakiano:
> @kepakiano In future can you not force push a branch once a review is under way? It makes it impossible to see what has been changed in response to review comments.

Hey, yeah, sure, sorry. It was just a few lines of the last commit and I didn't see you online in Slack, so I figured it wouldn't be a problem. But I'll stop :)

lib/engine/game/g_1835/step/draft.rb
kepakiano: That explains why it looked weird. Thanks!

ollybh:
> Hey, yeah, sure, sorry. It was just a few lines of the last commit and I didn't see you online in Slack, so I figured it wouldn't be a problem. But I'll stop :)

Adding new commits to the branch is fine. It's the forced push that's a problem, it breaks the branch history.

lib/engine/game/g_1835/game.rb
ollybh: > couldn't this even be the standard? Is there any game with a cert that's twice the regular cert percentage that does not want to be marked as a double_cert?

Maybe. But I wouldn't be surprised if there is some game that does something weird with certificate percentages and doesn't follow this rule.

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -87,17 +134,55 @@ def process_pass(action)
...
+              share.corporation.owner = player
+              @log << "#{player.name} becomes the president of #{share.corporation.name}"
+              company.close!
+            end
+
+            @game.bank.spend(share.price, share.corporation)
+
+            # allow president change for SX when LD is being sold, but for BY only if the BY director has been sold,
+            # which might be during the current or any earlier action
+            allow_president_change = share.president || !share.corporation.shares.first.president
```
ollybh, line 173: `share.corporation.shares.first.president` is always going to be true, isn't it? That's testing whether the first share is the president's certificate.

If you're attempting to test whether the first share has been bought you want `share.corporation.shares.first.owner.player?`.

lib/engine/game/g_1835/game.rb
```diff
@@ -27,7 +29,7 @@ class Game < Game::Base
                         red: '#d81e3e',
                         turquoise: '#00a993',
                         blue: '#0189d1',
-                        brown: '#7b352a')
+                        brown: '#7b352a',)
```
ollybh: I'm surprised that rubocop doesn't object to this. This would get a trailing comma if the closing bracket was on the next line, but not here where it is immediately followed by the bracket.
```suggestion
                        brown: '#7b352a')
```

lib/engine/game/g_1835/round/operating.rb
ollybh: So can this file be removed?

lib/engine/game/g_1835/step/draft.rb
```diff
@@ -6,27 +6,45 @@ module Engine
...
           ACTIONS = %w[bid pass].freeze
 
           def setup
-            @companies = @game.companies.sort_by { |item| [item.revenue, item.value] }
+            @companies = @game.companies.select { |c| c.owner.nil? && !c.closed? }
+
+            # set up the tiered companies as 2d array that might contain empty arrays if the starting package was not
+            # fully sold in a previous draft round. These empty arrays are important for may_purchase: companies in rows
+            # after an empty row can only be purchased if all rows before are empty. If we were to only group_by
+            # auction_row, we might loose these empty rows.
```
ollybh: Spelling.
```suggestion
            # auction_row, we might lose these empty rows.
```

lib/engine/game/g_1835/step/draft.rb
kepakiano: I think the `shares` array in the corporation loses its shares once they're transferred. The logic works and if I put debug output here, it states that "BY_0" is the first share when buying the first Bayern private and after BY_0 has been transferred after buying the director share and the next Bayern private is bought, it says "BY_1" or one of the other two.

lib/engine/game/g_1835/game.rb
kepakiano: leftover from when I was playing around. Where can I read up on what `register_colors` actually does? I tried to look at the code and couldn't figure it out.

lib/engine/game/g_1835/step/draft.rb
kepakiano: whoops

lib/engine/game/g_1835/round/operating.rb
kepakiano: oh, thought I did, yeah

lib/engine/game/g_1835/game.rb
ollybh: `register_colors` sets up a `COLORS` constant using  some slightly funky metaprogramming techniques…
https://github.com/tobymao/18xx/blob/c43364bf1579885a65e411a48cca3d6d559995a8/lib/engine/game/base.rb#L536-L542

…and this constant does not appear to be used anywhere. This was set up as a method of allowing corporation colour names to be defined, but it looks like the code that used these colours was deleted five years ago :roll_eyes:

ollybh (approved):

PR #12531, opened 2026-04-24
[1835] implement stock rounds

kepakiano:
Refers to  #3059

### Explanation of Change
Adds the following features:
- In 1835 there are three blocks of shares: [BY, SX], [BA, WT, HE, PR] and [MS, OL]. Shares of a corporation are only available if the corporation before it in the same block was floated. The first company in a block is only available if all non-reserved shares of the previous block have been sold. The Preußen is an edge-case: Its four IPO shares can be bought once the BA president cert has been bought.
- If a player holds more than 50% of a corporation, they can buy certs from other players for 150% of the current share price
- Every corporation of which a player holds 80% or more increases the cert limit by one


ollybh:
@kepakiano This PR has merge conflicts. Can you please resolve these and push an update to this branch?

kepakiano:
> @kepakiano This PR has merge conflicts. Can you please resolve these and push an update to this branch?

Thanks, done :)

lib/engine/game/g_1835/game.rb
```diff
@@ -168,6 +169,8 @@ def setup
 
           corporation_by_id('BY').ipoed = true
           corporation_by_id('SX').ipoed = true
+
+          @corporation_blocks = [%w[BY SX], %w[BA WT HE PR], %w[MS OL]]
```
ollybh: There are several places where you are converting between corporations and their IDs. Doing this just once at the start would make your code simpler. Something like this maybe:
```
CORPORATION_BLOCKS = [%w[BY SX], %w[BA WT HE PR], %w[MS OL]].freeze
…
@corporation_blocks = CORPORATION_BLOCKS.map { |block| block.map { |c| corporation_by_id(c) } }
```

lib/engine/game/g_1835/game.rb
```diff
@@ -216,6 +219,51 @@ def operating_round(round_num)
...
+          return if block == @corporation_blocks.last
+
+          next_block = @corporation_blocks[@corporation_blocks.index(block) + 1]
+          @log << 'All shares of the current block have been sold.'\
+                  " The next block is now available, starting with #{next_block.first}"
+          next_block.map { |c| corporation_by_id(c) }.each { |corp_to_ipo| corp_to_ipo.ipoed = true }
+        end
+
+        def cert_limit(player = nil)
+          return 0 unless player
```
ollybh: I'm not sure if this will make any difference anywhere, but returning the global certificate limit would make more sense than zero.
```suggestion
          return @cert_limit unless player
```

lib/engine/game/g_1835/game.rb
```diff
@@ -216,6 +219,51 @@ def operating_round(round_num)
...
+
+          @cert_limit + @corporations.count { |corporation| corporation.type == :major && player.percent_of(corporation) >= 80 }
+        end
+
+        def corporation_available?(corp)
+          return !corporation_by_id('BA').shares.first&.president? if corp == corporation_by_id('PR')
```
ollybh: The method to test whether a share is a president's certificate _should_ be called `president?` but it is actually called `president`.
```suggestion
          return !corporation_by_id('BA').shares.first&.president if corp == corporation_by_id('PR')
```

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1835
+      module Step
+        class BuySellParShares < Engine::Step::BuySellParShares
+          def process_buy_shares(action)
+            if action.bundle.owner.player?
+              action.bundle.share_price = nationalization_price(action.bundle.corporation.share_price.price)
```
ollybh, line 12: Maybe check where whether the corporation is being nationalised, and throw an error if not.

ollybh (commented):
This looks good, but I've added a couple of suggestions.

lib/engine/game/g_1835/game.rb
kepakiano: yeah, that simplifies a lot of code, thanks!

lib/engine/game/g_1835/game.rb
kepakiano: I guess it looks better :)

lib/engine/game/g_1835/game.rb
kepakiano: Interesting, I must've had the `president?` method from `corporation` in my head that takes an argument. Curious that the code still worked. Thanks for catching that!

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
kepakiano: I'll check whether the player can actually nationalize here

ollybh (approved):

PR #12532, opened 2026-04-25
[1840] Fix crash when all Y/O trains exhausted during CR1

### LINKED ISSUE #12257: [1840] Game broken after buying all Y and O trains in the first tram-buying moment
Image

Game ID: 232291 (https://18xx.games/game/232291)

**How to Reproduce the Bug?**

Three-player 1840. 
First player buys 2 Y1 trains, scraps one, buys one, scraps one, buys O1. Bought 4 Y1 and 1 O1; ends up with 2 Y1, 1 O1. Second player buys 3 O3 trains. Game crashes when buying the third. 
According to the rules, the red trains ought to be available for 500 each (as per VIII.5.3)

**Which Browser / Platform?**

Server-side. Breaks for all players. Try opening the URL; won't work.

ddaybell:
When all Y and O trams are exhausted during the first CR (CR1), the game previously crashed because no trains were available to buy while corporations were still required to purchase one. Per rules VIII.5.3, R trains should become available at 500 in this case.

Also defensively fix cheapest_train_price to handle empty buyable_trains and empty variant lists, preventing a NoMethodError on nil.

Changes:
- `available_trains` now appends R1 to the available train list when
  in CR1 and all Y1/O1 trains are exhausted (per rules VIII.5.3)
- `cheapest_train_price` guards against nil when no buyable trains exist
- 
Fixes #12257

### DIFF
--- a/lib/engine/game/g_1840/step/buy_train.rb
+++ b/lib/engine/game/g_1840/step/buy_train.rb
@@ -58,8 +58,10 @@ module Engine
 
           def cheapest_train_price(corporation)
             cheapest_train = buyable_trains(corporation).min_by(&:price)
-            cheapest_variant = buyable_train_variants(cheapest_train, corporation).first
-            cheapest_variant[:price]
+            return 0 unless cheapest_train
+
+            cheapest_variant = buyable_train_variants(cheapest_train, corporation).min_by { |v| v[:price] }
+            cheapest_variant ? cheapest_variant[:price] : 0
           end
 
           def process_buy_train(action)
--- a/lib/engine/game/g_1840/step/buy_train.rb
+++ b/lib/engine/game/g_1840/step/buy_train.rb
@@ -49,13 +49,13 @@ module Engine
 
           def can_buy_train?(entity = nil, _shell = nil)
             entity ||= current_entity
-            return false if @depot.depot_trains.none? { |t| buyable_train_variants(t, entity).any? }
+            return false unless buyable_depot_trains?(entity)
 
             super
           end
 
           def must_buy_train?(entity)
-            return false if @depot.depot_trains.none? { |t| buyable_train_variants(t, entity).any? }
+            return false unless buyable_depot_trains?(entity)
 
             scrappable_trains(entity).size.zero?
           end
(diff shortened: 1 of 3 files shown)


ollybh (commented):
The bug report was for when the first rank of trams are sold out earlier than expected, and this PR addresses this. Could the same problem occur with later ranks of trams? (I have played 1840 before, but I've only got a vague recollection of the rules).

ddaybell:
I'm not sure; I will investigate.

ddaybell:
In investigating this issue further, to look into olly's question above, I located Lonny's FAQ for this title.  He has a different solution to this issue.  Per the  FAQ (https://boardgamegeek.com/filepage/226146/1840-faq), when no trams are in any buying box and a corporation is required to own one, the obligation is temporarily excused until trams become available again. 

I implemented this by returning false from can_buy_train? and must_buy_train? when no depot train has a purchasable variant in the current CR, and logging a  descriptive message via log_skip when the step auto-skips for that reason.

Removes the previous approach of making R1 trains available early.

Also adds the official FAQ URL to the Info tab alongside the rules link.

lib/engine/game/g_1840/step/buy_train.rb
```diff
@@ -47,10 +47,30 @@ def process_scrap_train(action)
             @game.scrap_train(action.train, action.entity)
           end
 
+          def can_buy_train?(entity = nil, _shell = nil)
+            entity ||= current_entity
+            return false if @depot.depot_trains.none? { |t| buyable_train_variants(t, entity).any? }
```
ollybh: Methods like `any?` and `none?` are relatively computationally expensive as they create an enumerator object to step through the collection. If you're just interested whether there is anything in the collection you should be calling `empty?` instead.
```suggestion
            return false if @depot.depot_trains.none? { |t| !buyable_train_variants(t, entity).empty? }
```

ollybh: This check for buyable train variants is being made three times (here, in `must_buy_train?` and in `log_skip`) so it would be better to extract a method (`buyable_depot_train?` or something like that) to encapsulate this check.

ollybh (changes requested):
Good catch on finding the rule clarification.

I've suggested a minor tweak to your code to improve clarity and performance. It should then be good to go.

ddaybell:
Comments addressed.

ollybh (approved):

PR #12537, opened 2026-04-26
[1849] reorder corps after sell action, instead of at end of SR turn

### LINKED ISSUE #12421: [1849] Unexpected change in stock market order

michaeljb:
## Before clicking "Create"

### Explanation of Change

* `@moved_this_turn` is set to `[]` at the start of each player's SR turn, so it does not need to be set in `reorder_corps`
* if 3 or more corporations have shares sold and end on the same spot, `reorder_corps` will be called and possibly show up in the game log multiple times, keeping the order correct after each action
* Fixes #12421 - with tokens in the right order at the end of the sale action, floating a new corporation naturally puts its token in the right spot

### DIFF
--- a/lib/engine/game/g_1849/game.rb
+++ b/lib/engine/game/g_1849/game.rb
@@ -595,7 +595,8 @@ module Engine
 
         def reorder_corps
           just_moved = @moved_this_turn.uniq
-          @moved_this_turn = []
+          return unless just_moved.size > 1
+
           same_spot =
             @corporations
               .select(&:floated?)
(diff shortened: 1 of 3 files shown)


philcampeau (approved):
Looks good!

PR #12543, opened 2026-04-26
[18OE] Reserved 2+2 train obligation

Witzman:
## 18OE: Reserved 2+2 train obligation                                                                                                              
                                                                                                                                                      
  Implements the rule that every floated railroad must purchase a 2+2                                                                                 
  train before it may buy any other train level (§3.1 of the rulebook).                                                                               
  The obligation is tracked as a one-time flag per corporation — it is                                                                                
  not re-triggered if the 2+2 later rusts.                                                                                                            
                                                                                                                                                      
  ### Rules implemented                                                                                                                               
                                                                                                                                                      
  - **Regional/Minor Phase**: no corporation may buy a level-3 or higher                                                                              
    train from the depot while the game is still in the Regional/Minor
    Phase. Only 2+2 trains are available from the depot during this                                                                                   
    period.                                                                                                                                           
  - **Major Phase obligation window**: once the Major Phase begins, any                                                                               
    corporation that has not yet fulfilled its obligation is restricted to                                                                            
    the cheapest available depot train (the 2+2, while stock lasts) until                                                                             
    it has made that purchase. After fulfilling the obligation the full                                                                               
    depot becomes available.                                                                                                                          
  - **Obligation survives rusting**: the flag is stored in a `Set` keyed                                                                              
    by corporation id rather than checking `entity.trains.empty?`. A                                                                                  
    corporation whose 2+2 rusts is not forced to buy another one.                                                                                     
  - **Phase-status driven**: the obligation window is gated on                                                                                        
    `phase.status.include?('train_obligation')`, not a hard-coded phase                                                                               
    number, so the phase configuration in `PHASES` remains the single                                                                                 
    source of truth.                                                                                                                                  
  - **Emergency buy**: the corporation president (a player entity) is                                                                                 
    now allowed to call `can_entity_buy_train?` — required for the                                                                                    
    base engine's emergency-buy flow to work correctly. The stale                                                                                     
    `:minor` exclusion is also removed (minors are now gated by the                                                                                   
    obligation logic instead).                                                                                                                        
                                                                                                                                                      
  ### Changes                                                                                                                                         
                                                            
  #### `game.rb`                                                                                                                                      
  
  | Item | Change |                                                                                                                                   
  |---|---|                                                 
  | `attr_reader :fulfilled_train_obligation` | Exposes the Set to step classes |
  | `setup` | Initialises `@fulfilled_train_obligation = Set.new` |                                                                                   
                                                                                                                                                      
  #### `step/buy_train.rb`                                                                                                                            
                                                                                                                                                      
  | Method | Change |                                       
  |---|---|
  | `can_entity_buy_train?` | Returns `true` for a player entity that is the current corporation's president (emergency buy path); falls back to
  `entity.corporation?` for all other callers; removes the old `entity.type != :minor` exclusion |                                                    
  | `must_buy_train?` | Uses `fulfilled_train_obligation.include?(entity.id)` instead of `entity.trains.empty?`; still gated on
  `phase.status.include?('train_obligation')` and `entity.floated?` |                                                                                 
  | `buyable_trains` *(new)* | During Regional/Minor Phase: returns only `2+2` trains. During the Major Phase obligation window (entity not yet
  fulfilled): returns only the cheapest depot train. Otherwise delegates to `super` |                                                                 
  | `process_buy_train` *(new)* | Snapshots whether the purchase is inside the obligation window before calling `super`, then adds the entity to
  `fulfilled_train_obligation` if so |                                                                                                                
                                                            
  ### What is not in this PR                                                                                                                          
                                                            
  - Forced purchase / insolvency when a corporation cannot afford a train (openpoints §3.2–§3.3) — deferred                                           
  - Nationals claiming rusted trains for free (openpoints §1.9, §3.7) — deferred


lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,41 @@ module G18OE
...
+            trains = super
+
+            # Regional/Minor Phase: level 3+ trains blocked for all entities
+            return trains.select { |t| t.name == '2+2' } unless @game.major_phase?
+
+            # Obligation window in Major Phase: unfulfilled entity restricted to 2+2
+            if @game.phase.status.include?('train_obligation') &&
+               !@game.fulfilled_train_obligation.include?(entity.id)
+              min = @game.depot.min_depot_train
+              return min ? trains.select { |t| t.price == min.price } : []
```
michaeljb: If the restriction is to only buy `2+2` trains here, why can't this use the same `select` as line 27?


```suggestion
              return trains.select { |t| t.name == '2+2' }
```

Witzman: buyable_trains enforces two independent restrictions: (1) during the Regional/Minor Phase all entities are restricted to level 2 trains only (per §8.2 "No RR may purchase a level 3 train"); (2) during Major Phase, an entity with an outstanding obligation is restricted to the cheapest depot train (the 2+2, while any remain), even if 3 is also available.

if your suggestion matches this, its fine

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,41 @@ module G18OE
...
           def can_entity_buy_train?(entity)
-            entity.corporation? && entity.type != :minor
+            return true if entity.player? && entity == current_entity&.owner
+
+            entity.corporation?
           end
 
           def must_buy_train?(entity)
-            return false unless entity.trains.empty?
+            return false if @game.fulfilled_train_obligation.include?(entity.id)
```
crericha: Similar to the previous PR, its best to use methods `fulfilled_train_objligation?(entity)`

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,41 @@ module G18OE
...
             return false unless @game.phase.status.include?('train_obligation')
 
             entity.floated?
           end
 
+          def buyable_trains(entity)
+            trains = super
+
+            # Regional/Minor Phase: level 3+ trains blocked for all entities
+            return trains.select { |t| t.name == '2+2' } unless @game.major_phase?
```
crericha: Is this correct? I thought 3+ trains were available once an entity has bought a 2+2?

lib/engine/game/g_18_oe/step/buy_train.rb
crericha: Also, an entity cannot buy any train until they buy a 2+2, I believe. AND they cannot buy a second 2+2 from the depot.

If I understand correctly (big IF), then the logic to me should be

```
if @game.phase.status.include?('train_obligation')
  if !fulfilled_obligation(entity)
    trains.reject! {|t| t.name != '2+2'}
  else
    trains.reject! {t| t.name == '2+2'}
  end
end
```

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,41 @@ module G18OE
...
+              return min ? trains.select { |t| t.price == min.price } : []
+            end
+
+            trains
+          end
+
+          def process_buy_train(action)
+            in_obligation_window = @game.phase.status.include?('train_obligation')
+            super
+            @game.fulfilled_train_obligation.add(action.entity.id) if in_obligation_window
```
crericha: Another good place to have a method

`@game.fulfilled_train_obligation(action.entity) if action.train.name == '2+2' && train_from_depot`

lib/engine/game/g_18_oe/game.rb
```diff
@@ -16,6 +16,7 @@ class Game < Game::Base
         include G18OE::Map
         attr_accessor :minor_regional_order, :minor_available_regions, :minor_floated_regions, :regional_corps_floated,
                       :consolidation_triggered, :consolidation_done
+        attr_reader :fulfilled_train_obligation
```
crericha: Probably don't need an attr for this, if you move to methods per my suggestion.

Witzman: fulfilled_train_obligation? / fulfill_train_obligation in game.rb; attr_reader removed

Witzman:
Changes:

- price-based filter gone, t.name == '2+2' throughout                                                                    
- major_phase? dropped; obligation phase → unfulfilled gets only 2+2, fulfilled gets no 2+2 (can't buy a second depot copy)
- fulfilled_train_obligation? / fulfill_train_obligation in game.rb; attr_reader removed                          
- process_buy_train: fulfillment triggered on 2+2 depot purchase only, not on phase window Rule: Reserved 2+2 Obligation (§3.1)                                                                                                

Rules:                                                                                                                                  
  Each floated regional and minor must purchase a 2+2 train as its very first train purchase. This one-time obligation is tracked per entity and survives rusting.                                                                                                                                                                           
  Obligation window — active while train_obligation is in the phase status (Phases 2 and 3):                                          
  - Unfulfilled entity (has not yet bought its first 2+2): may only buy a 2+2 from the depot. It cannot skip ahead to a level-3 train even after Phase 3 opens.
  - Fulfilled entity (has already bought its 2+2): may buy any available train except a depot 2+2. It cannot acquire a second one.    
  - An entity is marked fulfilled the moment it buys a 2+2 directly from the depot; buying one from another corporation does not count.                                                                                                                              
                                                                                                                                      
  Outside the obligation window (Phase 4+): no restriction — normal train purchasing applies.                                         
                  
  Must-buy rule: a floated entity that has not yet fulfilled its obligation is forced to buy a train (the mandatory-purchase prompt   
  fires). Once fulfilled, the obligation never re-triggers — even if the entity's 2+2 later rusts and it is left trainless.

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: updated

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,35 @@ module G18OE
...
           def can_entity_buy_train?(entity)
-            entity.corporation? && entity.type != :minor
+            return true if entity.player? && entity == current_entity&.owner
+
+            entity.corporation?
           end
 
           def must_buy_train?(entity)
-            return false unless entity.trains.empty?
+            return false if @game.fulfilled_train_obligation?(entity)
```
crericha: Is this true for all entities? Or only certain types?

crericha: Like aren't majors required to own a train?

Witzman: The original override only handled the 2+2 reservation window, so must_buy_train? returned false for all entities once the phase moved past train_obligation — including trainless majors.                                                                                                 
                                                                                                                                                      
Fixed by splitting the method into two blocks:                                                                                                    
                                                                                                                                                      
  1. Obligation window (train_obligation phase status active): uses the fulfilled-Set check, not trains.empty? — a one-time flag per entity, as before.                                                                                                                                             
  2. General rule (all other phases): entity.floated? && entity.trains.empty? && entity.type == :major — only majors are required to own a train (§11.6: "Every major must own at least one train at the end of its OR. Minors and regionals are not required to purchase a train.").                
   
Minors and regionals are only ever forced to buy during the 2+2 obligation window. §11.6.5 confirms this explicitly: "When a minor or regional must force buy a train (which can only happen during the purchase of its reserved 2+2 train)."

Nationals are excluded from the general check for now — their train acquisition works differently (§11.6.6, §3.7 deferred).                         
Forced-purchase/insolvency mechanics (president shortfall, national conversion) remain open at §3.2.

Witzman:
## Level-3 train purchase restrictions (Rules 8.3 & 11.6)                                                                                                     
                                                                                                                                                                
  Two missing restrictions on level-3 train purchases:                                                                                                          
                                                                                                                                                                
  - **OR1 block**: level-3 trains may never be bought in the first OR, even                                                                                     
    if the Major Phase has started and the 2+2 obligation is fulfilled.
    (Rule 11.6; confirmed by designer Mark Frazier:                                                                                                             
    https://boardgamegeek.com/thread/1500166)                                                                                                                   
  - **Major Phase gate**: level-3 trains also require 18 regionals to have                                                                                      
    floated before they become available. (Rule 8.3)                                                                                                            
                                                                                                                                                                
  Implementation: `@first_or_done` latch in `next_round!`; `buyable_trains`                                                                                     
  returns `[]` for obligation-fulfilled companies until both conditions are met.                                                                                
                                                                                                                                                                
  2+2 count (30 = 12 minors + 18 regionals) confirmed - there have been spare cards in the box:                                                                                                          
  https://boardgamegeek.com/thread/1156774

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,43 @@ module G18OE
       module Step
         class BuyTrain < Engine::Step::BuyTrain
           def can_entity_buy_train?(entity)
-            entity.corporation? && entity.type != :minor
+            return true if entity.player? && entity == current_entity&.owner
```
crericha: A player is buying a train?

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,43 @@ module G18OE
...
+
+            entity.corporation?
           end
 
           def must_buy_train?(entity)
-            return false unless entity.trains.empty?
-            return false unless @game.phase.status.include?('train_obligation')
+            # During 2+2 reservation window: unfulfilled floated entities must buy
+            if @game.phase.status.include?('train_obligation')
+              return false if @game.fulfilled_train_obligation?(entity)
```
crericha: This logic doesn't seem right. If its a regional and its trainless, it still must buy a train.

crericha: I think the logic should be

```
entity.floated?? && entity.trains.empty? &&  (!@game.fulfilled_train_obligation?(entity) || entity.type == :major)
```

crericha: The ` @game.phase.status.include?('train_obligation')` bit should be included in the  `@game.fulfilled_train_obligation?(entity)` implementation.

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,43 @@ module G18OE
...
+              return [] if !@game.major_phase? || !@game.first_or_done
+
+              trains.reject { |t| t.name == '2+2' }
+            end
+          end
 
-            entity.floated?
+          def process_buy_train(action)
+            super
+            @game.fulfill_train_obligation(action.entity) if action.train.name == '2+2' && action.train.from_depot?
```
crericha: Add an `!` at the end of the method. That indicates the method mutates data.

crericha: `fulfill_train_obligation!(entity)`

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,43 @@ module G18OE
...
+
+          def buyable_trains(entity)
+            trains = super
+            return trains unless @game.phase.status.include?('train_obligation')
+
+            if !@game.fulfilled_train_obligation?(entity)
+              trains.select { |t| t.name == '2+2' }
+            else
+              # Rules 8.3 & 11.6: level 3+ requires Major Phase AND OR1 to have ended
+              return [] if !@game.major_phase? || !@game.first_or_done
```
crericha: Recommendation: Keep the details in the game class and expose simple methods to the steps. Something like, `@game.non_starter_trains_available?`.

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: misunderstanding on my side:
The return true if entity.player? check was added so players could pass the can_entity_buy_train? gate in the base actions method — which leads to the sell_shares e-buy path. But §11.6.4 explicitly states: "The president cannot sell stock to raise cash."    
                                                                                                                                            
Learning:  President personal cash contribution is handled automatically inside buy_train_action when president_may_contribute? returns true — no separate player action is needed.

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: moved the phase check into fulfilled_train_obligation? so it returns true when outside the obligation window, and collapsed must_buy_train? to the single expression you suggested. 
§11.6 confirms only majors must own a train obligation outside the window, so the logic should be correct for all entity types.

§11.6 Every major must own at least one train at the
end of its OR. Minors and regionals are not
required to purchase a train (exception: see
11.6.1).

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,32 @@ module G18OE
       module Step
         class BuyTrain < Engine::Step::BuyTrain
           def can_entity_buy_train?(entity)
-            entity.corporation? && entity.type != :minor
+            entity.corporation?
```
crericha: I don't think this method needs to be overridden anymore. The base class logic should suffice.

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,32 @@ module G18OE
...
           end
 
           def must_buy_train?(entity)
-            return false unless entity.trains.empty?
-            return false unless @game.phase.status.include?('train_obligation')
+            entity.floated? && entity.trains.empty? && (!@game.fulfilled_train_obligation?(entity) || entity.type == :major)
```
crericha, line 11: Is there an edge case where a regional/minor has a train (bought across because in the major phase), but still has an obligation to buy a 2+2?

crericha: As the code is written, `must_buy_train?` will return false if they have a train.

lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -8,18 +8,32 @@ module G18OE
...
+          end
+
+          def buyable_trains(entity)
+            trains = super
+            return trains unless @game.train_obligation_active?
+
+            if @game.fulfilled_train_obligation?(entity)
+              return [] unless @game.non_starter_trains_available?
```
crericha, line 19: Is this needed? Won't the next line accomplish the same thing?

lib/engine/game/g_18_oe/game.rb
```diff
@@ -15,7 +15,7 @@ class Game < Game::Base
         include G18OE::Entities
         include G18OE::Map
         attr_accessor :minor_regional_order, :minor_available_regions, :minor_floated_regions, :regional_corps_floated,
-                      :consolidation_triggered, :consolidation_done
+                      :consolidation_triggered, :consolidation_done, :first_or_done
```
crericha: I think you can remove the first_or_done accessor now.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -652,6 +654,18 @@ def major_phase?
...
+        def fulfilled_train_obligation?(entity)
+          !phase.status.include?('train_obligation') || @fulfilled_train_obligation.include?(entity.id)
+        end
+
+        def fulfill_train_obligation!(entity)
+          @fulfilled_train_obligation.add(entity.id)
+        end
+
+        def non_starter_trains_available?
+          major_phase? && first_or_done
```
crericha: If accessor removed, this is now @first_or_done

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: The guard is needed. In phase 3, available_upcoming_trains surfaces level-3 trains (their available_on phase is satisfied), so super returns both 2+2 and 3-trains. Without the guard, an entity that has fulfilled its obligation could purchase a level-3 train before non_starter_trains_available? is true. The next line alone doesn't accomplish the same thing.

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: Right, and that's intentional. During the obligation window (phases 2–3), buyable_trains restricts unfulfilled entities to 2+2 only, and can_buy_train_from_others? doesn't activate until phase 4. So the only train a minor/regional can hold while the obligation is active is a 2+2 —  meaning the obligation is already fulfilled. If they have a train, must_buy_train? = false is correct.

crericha: If its phase 4 when the minor/regional operates for the first time (assuming that's possible), isn't it legit to buy across first and then buy your 2+2?

Witzman: §11.6.3 covers this explicitly: "If Train Phase 4 begins prior to a RR's first OR after floating, all reserved trains are rusted and this rule no longer applies."
  
  So if a minor/regional's first OR falls in phase 4, the obligation is cancelled by rule. The code is correct — train_obligation only appears in phases 2 and 3; once phase 4 starts the status is gone and must_buy_train? returns false at the second guard. The 2+2s are rusted at that point too, so there's nothing to force-buy.

  The ordering requirement in §11.6.1 ("before any other train purchases") only applies when the obligation is still active — i.e., the entity's first  OR falls in phase 2 or 3.

crericha (approved):

PR #12545, opened 2026-04-26
[18Cuba] Tile upgrading and sugar cane fields

chrstas:
This PR enfores rules for tile upgrading. Additionally it adds sugar_cane_open_for_majors event triggered by the first 3-train

## Before clicking "Create"

Thanks for reviewing this PR. 

### Explanation of Change

The rules say "It is not necessary that any of the new track is usable by the company" (VII.7), so `:permissive` is correct for city and sugar cane upgrades. The plain track restriction ("may upgrade only if gauge track is extended") is enforced explicitly in `check_track_restrictions!`.

**Tile blocking**

- Minors: pure broad tiles blocked everywhere; on sugar cane hexes only tiles with towns are legal (plain track has no towns); city tiles only allowed on home hex
- Majors: pure narrow tiles blocked; yellow mixed-gauge city tiles (sugar mills) blocked — green/brown only allowed as upgrades from an existing sugar mill on that hex; other yellow non-broad tiles blocked

**Sugar Cane**

Minors build on sugar cane hexes for free; majors only from phase 3 (`sugar_cane_open_for_majors`)
`upgrades_to_correct_city_town?` override allows upgrading a sugar cane hex (hidden town) to plain track

### DIFF
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -18,14 +18,6 @@ module Engine
 
         include DoubleSidedTiles
 
-        register_colors(red: '#d1232a',
-                        orange: '#f58121',
-                        black: '#110a0c',
-                        blue: '#025aaa',
-                        lightBlue: '#8dd7f6',
-                        yellow: '#ffe600',
-                        green: '#32763f',
-                        brightGreen: '#6ec037')
         def sugar_cane_open_for_majors?
           @sugar_cane_open_for_majors
         end
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -301,7 +301,7 @@ module Engine
           if corp.type == :minor
             minor_tile_blocked?(tile, corp.tokens.first.hex, hex, for_selector: for_selector)
           else
-            major_tile_blocked?(tile)
+            major_tile_blocked?(tile, hex, for_selector: for_selector)
           end
         end
 
@@ -312,30 +312,40 @@ module Engine
         end
 
         def tile_has_only_track_type?(tile, track_type)
-          # Returns true if all paths on the tile are of the given track type
           tile.paths.all? { |path| path.track == track_type }
+        end
+
+        def mixed_gauge_city_tile?(tile)
+          tile && !tile.cities.empty? && tile.paths.any? { |p| p.track == :narrow }
         end
 
         def minor_tile_blocked?(tile, home_hex, current_hex, for_selector: false)
           # Determines if a tile is illegal for a minor:
           # - Tiles with only broad tracks are always illegal
           # - City tiles are illegal except on the minor's home hex
-          # - When `for_selector` is true, the home/city rule is ignored because the hex is unknown
+          # - On sugar cane hexes, only tiles with hidden towns are legal (no plain track)
+          # - When `for_selector` is true, the rules which require current_hex is ignored because the hex is unknown
           pure_broad = tile_has_only_track_type?(tile, :broad)
 
           return pure_broad if for_selector
-
           return pure_broad if current_hex == home_hex
+          return true if sugar_cane_hex?(current_hex) && tile.towns.empty?
 
           !tile.cities.empty? || pure_broad
         end
 
-        def major_tile_blocked?(tile)
+        def major_tile_blocked?(tile, hex = nil, for_selector: false)
           # Pure narrow tiles cannot be part of a major's route
           return true if tile_has_only_track_type?(tile, :narrow)
 
-          # Mixed gauge city tiles (sugar mill, e.g. L66, L77) are only for minor home cities
-          return true if !tile.cities.empty? && tile.paths.any? { |p| p.track == :narrow }
+          # Mixed gauge city tiles (sugar mill) are minor-only in yellow.
+          # In green/brown they are only allowed as upgrades from an existing sugar mill
+          # (e.g. L53 → L67, L67 → brown sugar mill); majors cannot place them on plain hexes.
+          if mixed_gauge_city_tile?(tile)
+            return true if tile.color == :yellow
+            return false if for_selector
+            return true unless mixed_gauge_city_tile?(hex&.tile)
+          end
 
           # Yellow tiles must be pure broad for majors
           tile.color == :yellow && !tile_has_only_track_type?(tile, :broad)
--- a/lib/engine/game/g_18_cuba/game.rb
+++ b/lib/engine/game/g_18_cuba/game.rb
@@ -18,6 +18,16 @@ module Engine
 
         include DoubleSidedTiles
 
+        register_colors(red: '#d1232a',
+                        orange: '#f58121',
+                        black: '#110a0c',
+                        blue: '#025aaa',
+                        lightBlue: '#8dd7f6',
+                        yellow: '#ffe600',
+                        green: '#32763f',
+                        brightGreen: '#6ec037')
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -28,6 +32,19 @@ def potential_tiles(entity, hex)
               @game.tile_blocked_for_corp?(tile, corp, hex)
             end
           end
+
+          def check_track_restrictions!(entity, old_tile, new_tile)
+            # City tiles: "It is not necessary that any of the new track is usable by the company."
```
michaeljb, line 39: This sentence means that `TRACK_RESTRICTION` needs to be `:permissive`.

"Semi-restrictive" is the other most commonly-used track restriction type, which is where the company must be able to reach either *some new* track on the new tile, or a revenue center whose value is upgraded by the new tile.

"Restrictive" is used in only a couple of titles on the site, IIRC it requires the company to reach *all* track segments on the new tile.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -26,7 +26,9 @@ class Game < Game::Base
                         yellow: '#ffe600',
                         green: '#32763f',
                         brightGreen: '#6ec037')
-        TRACK_RESTRICTION = :permissive
+        attr_reader :sugar_cane_open_for_majors
+
+        TRACK_RESTRICTION = :restrictive
```
michaeljb:
```suggestion
        TRACK_RESTRICTION = :permissive
```

michaeljb (changes requested):

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -28,6 +32,19 @@ def potential_tiles(entity, hex)
...
+            return unless old_tile.cities.empty?
+
+            # Sugar cane (town) tiles: "A minor company may upgrade any sugar cane field tile."
+            # Majors are already blocked from pure narrow upgrades via major_tile_blocked?.
+            return unless old_tile.towns.empty?
+
+            # Plain track tiles: must extend the entity's route.
+            # graph_for_entity filters by gauge (narrow for minors, broad for majors).
+            super
```
ollybh: I don't think this comment is right. The seventh bullet point in section VII.7 says "It is not necessary that any of the new track on the tile is usable by the company".

The special rules for minor and major companies later in VII.7 add restrictions to the upgrades which mean (if I've understood correctly) that the company must have a route to new track on a mixed gauge green tile. But if a major company upgrades a broad gauge yellow tile to a broad gauge green tile (for example a 9 to a 23) then I don't think that they need to be able to reach the new track.

Like @michaeljb says, this means that you should have `:permissive` for `TRACK_RESTRICTION`.

chrstas:
thanks a lot to you both - appreciate your help. I have changed it back to :permissive and included the upgrade rules. I have updated the PR description so that it fits again to the change.

ollybh:
@chrstas This PR has a merge conflict (caused by my PR #12551, removing the `register_colors` method). Can you please resolve this and push an update?

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -28,6 +32,30 @@ def potential_tiles(entity, hex)
...
+            old_paths = old_tile.paths
+            used_new_track = old_paths.empty?
+
+            new_tile.paths.each do |np|
+              next unless graph.connected_paths(entity)[np]
+
+              used_new_track = true unless old_paths.any? { |path| np <= path }
+            end
+
+            raise GameError, 'Must use new track' unless used_new_track
```
ollybh: From my reading of the rules, there has to be new track of the correct gauge laid, but it does not have to be reachable by the company laying the track. So I think that this is too restrictive, it is checking whether the company has a route to the new track.

ollybh: I think the only time this will make a difference is when you have a major company upgrading a yellow plain track hex to a green plain track hex. It could upgrade an 8 tile to a 19 or a 23, even if it only has a route to the existing track.

chrstas:
Merge conflict resolved and reviewer feedback considered. @ollybh - you are right, I was too restrictive. I have checked the rules and your example is right.

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -28,6 +32,24 @@ def potential_tiles(entity, hex)
...
+            # Sugar cane (town) tiles: "A minor company may upgrade any sugar cane field tile."
+            # Majors are already blocked from pure narrow upgrades via major_tile_blocked?.
+            return unless old_tile.towns.empty?
+
+            # Plain track: the gauge track must be extended, but new track need not be reachable
+            # ("it is not necessary that any of the new track is usable by the company").
+            # Hex reachability is already enforced by tracker_available_hex.
+            old_paths = old_tile.paths
+            raise GameError, 'Must use new track' if !old_paths.empty? &&
+                                                      new_tile.paths.all? { |np| old_paths.any? { |op| np <= op } }
```
ollybh: What is this check doing? If I've understood it correctly (which I probably haven't), this is checking that all the paths on the new tile are copies of ones on the old tile. Which isn't much of an upgrade.

lib/engine/game/g_18_cuba/step/track.rb
chrstas: You're right, the check was incomplete — I missed the gauge dimension entirely. The original condition only verified that some new path was added, but didn't check whether it was the right gauge. A major could have "upgraded" a plain broad tile by only adding narrow track, leaving the standard gauge unchanged. Fixed by making the check gauge-aware: majors require at least one new non-narrow path, minors at least one new non-broad path.

lib/engine/game/g_18_cuba/step/track.rb
chrstas: You're right, the check was incomplete — I missed the gauge dimension entirely. The original condition only verified that some new path was added, but didn't check whether it was the right gauge. Fixed by making the check gauge-aware: majors require at least one new non-narrow path, minors at least one new non-broad path.

Example Screenshot: the green update (done by major MS) is for my understanding against the rule. 

image

After the bug fix now this upgrade is rejected:

image

lib/engine/game/g_18_cuba/step/track.rb
```diff
@@ -28,6 +32,32 @@ def potential_tiles(entity, hex)
...
+            case corp.type
+            when :major
+              raise GameError, 'Standard gauge track must be extended' unless new_tile.paths.any? do |np|
+                np.track != :narrow && old_paths.none? { |op| np <= op }
+              end
+            else
+              raise GameError, 'Narrow gauge track must be extended' unless new_tile.paths.any? do |np|
+                np.track != :broad && old_paths.none? { |op| np <= op }
+              end
+            end
```
ollybh: You could possibly extract a bit of common code by writing this something like:
```
added_paths = new_tile.paths.reject { |path| old_paths.any? { |op| path <= op }
raise GameError, '…' unless added_paths.map(&:track) == [corp.type == :major ? :broad : :narrow]
```
But I'm not sure if that's needed. The suggested code is probably fine.

ollybh (approved):
You cou

lib/engine/game/g_18_cuba/step/track.rb
chrstas: @ollybh I have included this as well - makes sense and looks cleaner. Could you then merge it? thanks again for your efforts...

ollybh (approved):

PR #12547, opened 2026-04-26
[1835] correct city slots on brown Hamburg tile and calculate ferry fee

kepakiano:
Refers to  #3059

### Explanation of Change

The previous tile 221 was not correct, because it had two cities. Tokening one of them out would block trains going from north to south and vice versa. That's not how it's supposed to be in 1835: As long as any slot is open, trains can freely traverse Hamburg. The little bridge on the brown tile of the physical copy is just there to indicate that trains crossing from any of the three edges south of the Elbe to any edge north of the Elbe have to pay the 10M ferry fee. This PR therefore merges the cities again into a regular brown tile with one city with three slots.
To indicate that the tile can have two revenues it shows these in a strip. The colors used for 50 and 60 are both slightly off brown so that the automatic revenue calculation in `revenue_center#route_base_revenue` returns 0. The numbers stated in `map.rb` are therefore purely there for the UI. The actually calculation happens in `game.rb`.

### Screenshots

All these screenshots where taken from a modified `game.rb` with increased train limit and cash for the minors so that I could quickly simulate this.

image
The revenue is correctly calculated as 60 + 10

image

Including Magdeburg in the run only increases revenue by 10 because of the now applied ferry fee (see adjusted revenue str)

image

Excluding Schwerin leaves the revenue at 80 (fee dropped, but Schwerin only brought 10, so total remains unchanged)

image

again with hint and Hamburg only counting 50

image

and no hint and Hamburg counting 60


lib/engine/game/g_1835/game.rb
```diff
@@ -279,6 +279,39 @@ def operating_round(round_num)
...
+        end
+
+        def brown_hamburg_tile_revenue(route)
+          return 0 unless uses_brown_hamburg_tile(route)
+          return 60 unless uses_hamburg_ferry(route)
+
+          50
+        end
+
+        def uses_hamburg_ferry(route)
```
ollybh: The Ruby style is to give methods which return a boolean a name ending with a question mark. So this would be better names `uses_hamburg_ferry?` or just `hamburg_ferry?`.

lib/engine/config/tile.rb
```diff
@@ -358,7 +358,7 @@ module Tile
         '218' => 'city=revenue:40,slots:2;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0;path=a:4,b:_0;label=X',
         '219' => 'city=revenue:40,slots:2;path=a:0,b:_0;path=a:1,b:_0;path=a:3,b:_0;path=a:5,b:_0;label=X',
         '220' => 'city=revenue:60,slots:3;path=a:0,b:_0;path=a:1,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:5,b:_0;label=B',
-        '221' => 'city=revenue:60,slots:2,loc:3;city=revenue:60,loc:0;path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0;path=a:1,b:_1;path=a:0,b:_1;path=a:5,b:_1;path=a:_0,b:_1;label=H',
+        '221' => 'city=revenue:#cb7743_50|#cb7744_60,slots:3;path=a:_0,b:0;path=a:_0,b:1;path=a:_0,b:2;path=a:_0,b:3;path=a:_0,b:4;path=a:_0,b:5;label=HH',
```
ollybh: This is a bit underhand. If you're going to use this technique for displaying the revenue, can you please add a comment that explains why you're doing this?

lib/engine/game/g_1835/game.rb
kepakiano: true, changed, thanks

lib/engine/config/tile.rb
kepakiano: You're right, I'd expect a comment, too, if I were to read this.

ollybh:
Does this work OK with the auto-router? That calls into `revenue_for` which will be returning zero for the brown Hamburg tile.

You could have the second revenue for the Hamburg tile be brown, keeping the off-brown colour for the first. That would mean that `revenue_for` would return 60. Then you could adjust the route revenue calculation to deduct 10 for the Hamburg ferry if needed.

kepakiano:
Yeah, I guess putting the proper value in the tile manifest makes sense. This also allows the code to be a little shorter. There is now only one method that answers the entire question of "has the Hamburg ferry been used for a route" and includes the check whether the tile is brown or not.

Alright, in green everything works as expected (if you want to test this yourself, make sure to either merge the +-train PR or correct the distance on +-trains yourself):
image


auto-routing came up with this route and it is of course optimal. The Hamburg ferry is not deducted since the tile is not brown.

After laying the brown tile, 3 can run this 3+3 route (also returned by auto-router):

image

This is also optimal: 20 from Braunschweig, 20 from Magdeburg, 60 minus 10 from Hamburg and three towns = 120

After upgrading the tile south-east of Hamburg auto-router correctly realizes it can avoid the ferry:

image

Since going into Hamburg on the south-west edge and existing north-east would mean using the ferry again:

image

ollybh (approved):

Issue #10369, opened 2024-02-19
Add a "Review since my last action" button to the history navigator


tdbs:
When playing async one of the most used things (at least by me) is to review what happened since your last play. Right now the easiest way to do this is going back through the log and find your last action, click the action and select: "Review from here".

Since this is something you want to do many times I believe it should be easier to do. Why isn't it super easy?
* Because there may have been a lot of actions since you last acted.
* Your own actions aren't highlighted by a different color (which could be a suggestion on its own).

I believe adding a button to go back in history to your last action shouldn't be that hard and could be a quality of life improvement.
Alternatively, or better yet in conjunction with, having your own actions on the log having a different color could also help (you could use the color for player 1 defined on the profile).

PR #12549, opened 2026-04-27
[Core] Implement game log indentation groups

### LINKED ISSUE #12400: Improve the log readability and some features
I basically only play async games so reading the log is important and I feel we could make it a lot easie to see what is going on (: 

We could use one or many of the ideas below, and maybe they could be an option on our profile, to check the ones we want, if any

Currently:

[20:48] -- Operating Round 5.1 (of 2) --
[20:48] PedroS operates PRR
[20:50] PRR spends $20 and lays tile #15 with rotation 4 on G15 (Columbus)
[20:50] PRR spends $40 and lays tile #8 with rotation 0 on H16
[20:51] PRR runs a 5 train for $250: B18-B16-C15-D14-D6
[20:51] PRR runs a 4/6 train for $310: C21-(D20)-E17-(E11)-D6-C5 + E/W
[20:51] PRR pays out $560 = $56 per share ($280 to PedroS, $168 to bad company, $56 to Silberluchs, $56 to your_average_euro)
[20:51] PRR's share price moves right from $230 to $270 (2 steps)
[20:51] PRR skips buy/use companies
[20:51] bad company operates IC
[21:15] IC spends $20 and lays tile #300 with rotation 0 on D6 (Chicago)

But it could be

[20:48] -- Operating Round 5.1 (of 2) --
**[20:48] PedroS operates PRR**
(besides being bold it could also use the color of that player, either the text, the background for their whole actions, or maybe just on the time stamp?)

- [20:50] PRR spends $20 and lays tile 15 with rotation 4 on G15 (Columbus)
- [20:50] PRR spends $40 and lays tile 8 with rotation 0 on H16
- [20:51] PRR runs a 5 train for $250: B18-B16-C15-D14-D6
- [20:51] PRR runs a 4/6 train for $310: C21-(D20)-E17-(E11)-D6-C5 + E/W
- [20:51] PRR pays out $560 = $56 per share ($280 to PedroS, $168 to bad company, $56 to Silberluchs, $56 to your_average_euro)
- [20:51] PRR's share price moves right from $230 to $270 (2 steps)
- [20:51] PRR skips buy/use companies

**[20:51] bad company operates IC**

- [21:15] IC spends $20 and lays tile 300 with rotation 0 on D6 (Chicago)

I used bullets as that was what I had available here, but it could just be a small indentation to help group actions from a single player/company

Additionally, clicking on the log could (besides the option to review from here) make that action pop up on the map, just as a visual, while leaving all the menus the same for us to do our next action. So clicking on a:

"PRR spends $20 and lays tile 15 with rotation 4 on G15 (Columbus)"

line would just make the G15 hex blink

to see this one in action you can check the page below where people play splotters, click on history and then click on a "acquired company" line, and you will see it pop up on the map, e.g. here

https://www.onlineboardgamers.com/IND/18766/show/

Thanks! :D

ddaybell:
Implements an indentation function for the game logs.  Log entries are now indented to visually indicate a group of related log entries (e.g. multiple actions in an OR by the same company, or multiple bids in an auction on the same entity, or multiple actions related to the same player's actions in an SR (buy/sell/float).

Two types of grouping are used:

    Subject-centric (SR/OR): lines group under the acting player or the
  operating corporation; player lines during an OR stay grouped under
  the corp.

Object-centric (auction): all bid/pass/win lines indent under the lot
  being auctioned, regardless of which player acts.

Infrastructure (See attached summary file for more fulsome description of each change):
- lib/engine/game_log.rb: Entry now carries auctioning_lot; GameLog#<< captures it from active_step at write time.
- lib/engine/step/auctioner.rb: adds auctioning_lot returning @auctioning as the base implementation for all auctioning steps.
- lib/engine/round/operating.rb and game-specific overrides (1825, 1860, 18Hiawatha): operation banner changed to "-- X operates Y --" format so the view can detect it as a banner and set @in_or_operation.

Game log view (assets/app/view/game/game_log.rb):
- entity_for_line now accepts auctioning_lot and applies object-centric grouping when a lot is active.
- Corp-at-start guard requires next char to be a space, preventing single-char corp IDs from matching mid-word (fixes 1822CA SR1).
- @in_or_operation gates the "player line stays under corp" rule so it only applies during OR operations, not SR turns.
- Operates banner sets @last_entity and @in_or_operation; Round banner resets both.

Per-game fixes:
- 1867 / 18RoyalGorge single_item_auction: set @auctioning before logging "is up for auction" so the line is tagged to the lot.
- 1866 single_item_auction / game: moved first auction announcement out of setup and into game#setup so it follows stock-turn-token messages.
- buy_sell_par_shares_via_bid: @lot_for_log pattern tags "auctions X for" and home-token lines to the lot without disturbing @auctioning.
- waterfall_auction: defers "goes up for auction" announcement until after pending par actions complete (fixes 1870 MKT/SLSF ordering).
- 1858 private_exchange / exchange: acquire_private silent: option lets exchange_for_presidency log in correct order (exchange → par → acquires).

Fixes #12400, #10369

## Before clicking "Create"

### Explanation of Change

[game-log-indenter-changes.txt](https://github.com/user-attachments/files/27111312/game-log-indenter-changes.txt)

### Screenshots

image

image

image

### DIFF
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -22,9 +22,9 @@ module View
       needs :show_log, default: true, store: true
 
       def render
-        children = [render_log_choices, render_log]
-
         @player = @game.player_by_id(@user['id']) if @user
+
+        children = [render_log_choices, render_log]
 
         key_event = lambda do |event|
           event = Native(event)
@@ -102,6 +102,10 @@ module View
         blank_action.created_at = @game.actions[0]&.created_at || Time.now
 
         last_action = nil
+        @last_entity = nil
+        @in_or_operation = false
+        @eft_corps = nil
+        @eft_companies = nil
 
         actions = @game.actions.to_h { |a| [a.id, a] }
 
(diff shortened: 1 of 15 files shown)


ddaybell:
As best I can tell, the rspec failures were pre-existing and not caused by my code.  Happy to investigate further if desired.

perwestling:
> As best I can tell, the rspec failures were pre-existing and not caused by my code. Happy to investigate further if desired.

Do your local code pass all tests?
> * Code passes linter with `docker compose exec rack rubocop -a`
> * Tests pass cleanly with `docker compose exec rack rake`

ollybh (changes requested):
I get the rake errors when testing this branch locally. This seems to be breaking something important.

ollybh:
Each item in the game log is an Entry object. This includes a reference to the ID of the action that created the log line. Could you use this to work out the state of the current round instead of the fragile-looking text parsing?

ollybh:
Item number 900 on my mental TODO list is to enhance the logging to record a lot more information with the log entries. This would be to allow easy navigation between player turns, and show images of tiles that have been laid.

ddaybell:
I'm inclined to agree that text parsing to derive intent is a bit fragile, as it relies on subsequent coders to not muck with the text.  I will put some thought into how I can use the action ID references here, as those should be a bit more stable than the text would be.  

I also really like your ideas about recording more information with log entries.  I think that would open a lot of doors to building a number of other enhancements such as you indicate.  As I was coding this, I was wishing that I had more such information available. :-)  

I propose to take both of those ideas back to my workshop for a bit, and circle back with another rev of this.  I'll also dig further into the rake messages and see what's going on there.

ddaybell:
I have some initial thoughts that I'd appreciate your views on.  The code currently captures only two items for each log entry, the message string and the action ID.  My indenter code adds a third item, auctioning_lot, which stores the ID of the entity being auctioned off in an auction (or nil if there is no auction ocurring).  The indenter uses this to decide how to group entries during auctions, because grouping is different there than in non-auction settings).  I added this item to the shovel definition.  At shovel time, we also have access to quite a bit more information that we could add to the log entry data structure.  This includes everything in @game (turn, round number, phase, etc.) as well as information about the active entity (ID, sym, name, controlling player).  If we extended the log entry data structure to include this information, that wouldn't add a lot to the size, and it would open doors to folks who wanted to use that information for other improvements.  

We could also add other fields that aren't always present but sometimes are, such as the ID number and location for a tile when one is laid, or the name of a train if one is involved in the log message (e.g. bought, sold, rusted, etc.).  These would let folks base other features off of this information.  The shovel definition already has a rescue that takes care of nil values, so we wouldn't have to worry about the code breaking if we tried to capture a non-existent value.  

Something like:

def <<(message)
  entry = message.is_a?(Entry) ? message : Entry.new(message, @game.current_action_id)
  begin
    step = @game.active_step
    entry.auctioning_lot = step.auctioning_lot if step.respond_to?(:auctioning_lot)
    entity = @game.current_entity
    entry.turn           = @game.turn
    entry.round_num      = @game.round.round_num
    entry.round_type     = @game.round.name
    entry.phase_name     = @game.phase.name
    entry.round_counter  = @game.round_counter
    entry.turn_start_action_id = @game.turn_start_action_id
    entry.entity_id      = entity&.id
    entry.entity_sym     = entity&.sym
    entry.entity_type    = entity&.class&.name
    entry.player_id      = @game.acting_for_player(entity&.player)&.id
  rescue StandardError
  end
  super(entry)
end

Could be a good start to this information expansion.

ddaybell:
Also, I figured out the test issues.  It was a subtle recursion issue created by the change to the shovel def to catch auctioning situations.  Will be good to keep in mind that changes to the shovel def can have unusual effects in unexpected places...

ollybh:
@ddaybell Can this PR be closed as we've agreed the log text parsing will be unreliable, and a better approach would be to add to the information stored in the game log?

When you're looking at how the log can be extended, keep the amount of information added to the minimum you need for the moment. More can be added later.

ddaybell:
Yes, that's fine.

ollybh:
Superseded by #12564.

PR #12551, opened 2026-04-27
[Core/various] Remove `register_colors` method

ollybh:
The `register_colors` method was added in PR tobymao#471 to give a way of assigning custom colours to corporations. This was done by creating a `COLORS` hash. The code that used this hash to assign the colours to the corporations was removed in PR tobymao#4167 but the `register_colors` method was left in place and it is still called from lots of games, even though it now has no effects.

This commit purges this method from the code base.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -531,14 +531,6 @@ module Engine
       # use to modify tiles based on optional rules
       def optional_tiles; end
 
-      def self.register_colors(colors)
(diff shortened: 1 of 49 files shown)


philcampeau (approved):

perwestling (approved):

crericha (approved):

roseundy (approved):

philcampeau:
I had always wondered what that was about!

PR #12558, opened 2026-04-29
[18OE] Full Map for alpha functionality -- Privates/Minors charter&station Markers colors

Witzman:
### Public Ports added to map.

Notice - this Pull Request also contains https://github.com/tobymao/18xx/pull/12578 - as this change is essential and i want to avoid merge conflicts.
If this PR gets merged before 112578 - 12578 can be closed

in THIS PR.
-intra-hex province border lines via partition type :province (PR 12578)  
- Red Hexed fixed 
- Terrain visibility changed
- charter Colors for changes to more make the 30+ Companies to make them more distinguishable - to few colors if using defined colors

Not Included:
- Private Ports 
- Port / Ferry logic
- Inland Ports

grafik

grafik


lib/engine/game/g_18_oe/entities.rb
```diff
@@ -581,7 +603,7 @@ module Entities
             float_percent: 50,
             max_ownership_percent: 100,
             coordinates: 'J27',
-            color: :black,
+            color: '#333333',
```
crericha, line 606: Do you need all these specific colors? Its okay to be "close enough".

Witzman: in my test games i was totally confused before, as most of the minors have been gray and also station markers and charters 
i did not replicate the indentical colors of the base game - having all "red comapies" exactly the same color was hard identifying the station markers on the map in browser, when companies operate in the same area.

As the engine has to less colors defined, i decided to simply use hexcodes

assets/app/view/game/part/partitions.rb
crericha, line 1: Looks like your branches converged. This code is a duplicate of another PR.

lib/engine/ability/blocks_partition.rb
```diff
@@ -8,7 +8,7 @@ class BlocksPartition < Base
       attr_reader :partition_type
 
       def setup(partition_type:)
-        @partition_type = partition_type
+        @partition_type = partition_type&.to_sym
```
crericha: I don't think to_sym is needed.

lib/engine/part/partition.rb
crericha, line 1: Duplicate code from another PR

crericha (approved):

PR #12561, opened 2026-05-01
[18OE] Adjustment of Dividend Options for Minor, Major, and National

Witzman:
Switched from old 18oeUKFR custom code to engine defaults based on 18OE rules:

Dividend Options:

    Minors: half pay only
    Nationals: payout only
    Majors: payout, half pay, withhold

Stock Movement: G18OE adaptation as described.

    majors & nationals move_right if Revenue >= Share price

Files:

    lib/engine/game/g_18_oe/step/dividend.rb
    lib/engine/game/g_18_oe/game.rb

### DIFF
--- a/lib/engine/game/g_18_oe/game.rb
+++ b/lib/engine/game/g_18_oe/game.rb
@@ -761,6 +761,23 @@ module Engine
           corporation.spend(cost, @bank) if cost&.positive?
         end
 
+        # Override stock price movement according to 18OE rules
+        # - Minors & Regionals: no movement
+        # - Majors & Nationals:
+        #   * revenue >= share price -> move right
+        #   * revenue between 0 and share price -> no move
+        #   * revenue = 0 -> move left
+        def change_share_price(entity, revenue)
+          return if entity.type == :minor || entity.type == :regional
+
+          share_price = entity.share_price.price
+          if revenue >= share_price
+            @stock_market.move_right(entity)
+          elsif revenue.zero?
+            @stock_market.move_left(entity)
+          end
+        end
+
         def issuable_shares(entity)
           return [] if !entity.corporation? || entity.type != :major
 
--- a/lib/engine/game/g_18_oe/step/dividend.rb
+++ b/lib/engine/game/g_18_oe/step/dividend.rb
@@ -1,37 +1,23 @@
 # frozen_string_literal: true
 
 require_relative '../../../step/dividend'
-require_relative '../../../step/minor_half_pay'
+require_relative '../../../step/half_pay'
 
 module Engine
   module Game
     module G18OE
       module Step
         class Dividend < Engine::Step::Dividend
-          DIVIDEND_TYPES = %i[payout half withhold].freeze
-          include Engine::Step::MinorHalfPay
+          include Engine::Step::HalfPay
 
-          def half(entity, revenue)
-            withheld = half_pay_withhold_amount(entity, revenue)
-            { corporation: withheld, per_share: payout_per_share(entity, revenue - withheld) }
-          end
-
-          def half_pay_withhold_amount(entity, revenue)
-            (revenue / 2 / entity.total_shares) * entity.total_shares
-          end
-
-          def share_price_change(entity, revenue = 0)
-            return {} if @game.minor_regional_order.include?(entity)
-
-            price = entity.share_price.price
-            return { share_direction: :left, share_times: 1 } if revenue < 1
-
-            times = 0
-            times = 1 if revenue >= price
-            if times.positive?
-              { share_direction: :right, share_times: times }
+          def dividend_types(entity = current_entity)
+            case entity&.type
+            when :minor
+              [:half]
+            when :national
+              [:payout]
             else
-              {}
+              %i[withhold half payout]
             end
           end
         end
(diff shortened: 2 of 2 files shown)


lib/engine/game/g_18_oe/step/dividend.rb
```diff
@@ -1,37 +1,23 @@
...
-
-            price = entity.share_price.price
-            return { share_direction: :left, share_times: 1 } if revenue < 1
-
-            times = 0
-            times = 1 if revenue >= price
-            if times.positive?
-              { share_direction: :right, share_times: times }
+          def dividend_types(entity = current_entity)
+            case entity&.type
```
ollybh: This method doesn't take a parameter. And is the `&` operator needed here? If `current_entity` is returning nil then something bad has gone wrong.
```suggestion
          def dividend_types
            case current_entity.type
```

lib/engine/game/g_18_oe/game.rb
```diff
@@ -761,6 +761,23 @@ def after_par(corporation)
...
         end
 
+        # Override stock price movement according to 18OE rules
+        # - Minors & Regionals: no movement
+        # - Majors & Nationals:
+        #   * revenue >= share price -> move right
+        #   * revenue between 0 and share price -> no move
+        #   * revenue = 0 -> move left
+        def change_share_price(entity, revenue)
+          return if entity.type == :minor || entity.type == :regional
```
ollybh, line 771: Hmm. If minors don't have a share price then these would be best represented by Minor objects instead of Corporations. Maybe regionals would too, but they have multiple shares so that is less straightforward.

ollybh (changes requested):

lib/engine/game/g_18_oe/game.rb
Witzman: I can refacor the whole company stuff which was did in 18oeukfr prealpha before, but then i would appreciate getting my prs done first. 
I can plan a feature freeze for rewriting old code, but i dont want to do this on 5 open branches simultaneously with no real testing possibility:)

Already started refactoring of old UKFR Code in the regional to major conversion PR - also this PR replaces old legacy code.

The old 18OE UKFR was broken anyways, i just grabbed whats there and extended everything to the full game and make it startable for further Development/Testing. Currently testing is nearly impossible as all the required smaller changes live in diffrent PRS as my initial PR was to big.


I would suggest approving my 5 PRs for now. They bring the game to a state, where everything can be floated and further tested in ORs & SRs.

I would plan the refactoring of the minors with the upcoming "merging a minor into a major" PR, wich would be the next big step. Regional i would continue to handle as coorperations, as they have 4 shares, presidents - this makes the conversion to major more straightforward i currently think.

lib/engine/game/g_18_oe/step/dividend.rb
Witzman: Changed

lib/engine/game/g_18_oe/game.rb
crericha: I've found that I prefer the corporation class, even if the corporation doesn't have a share price. It works great in 1837.

https://github.com/tobymao/18xx/blob/3ccbbb2878b81ddfba1e84a1f7d42b70c95807ec/lib/engine/game/g_1837/game.rb#L675 -- no par price
https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1837/step/dividend.rb

I believe the game code is smart enough to know not to change share price if there is no share price.

lib/engine/game/g_18_oe/step/dividend.rb
```diff
@@ -1,37 +1,23 @@
...
-            times = 0
-            times = 1 if revenue >= price
-            if times.positive?
-              { share_direction: :right, share_times: times }
+          def dividend_types
+            case current_entity.type
+            when :minor
+              [:half]
+            when :national
+              [:payout]
```
crericha: Minors and nationals shouldn't have to pick a dividend since there is only one. See here for how to do it automatically https://github.com/tobymao/18xx/blob/3ccbbb2878b81ddfba1e84a1f7d42b70c95807ec/lib/engine/game/g_1837/step/minor_half_pay.rb#L14.

lib/engine/game/g_18_oe/game.rb
crericha: TL;DR; I don't think the early return is required.

I think you can do this

```
          def share_price_change(_entity, revenue)
            if revenue.zero?
              { share_direction: :left, share_times: 1 }
            elsif revenue == total_revenue
              { share_direction: :right, share_times: 1 }
            else
              { share_direction: :diagonally_down_right, share_times: 1 }
            end
          end
```

lib/engine/game/g_18_oe/step/dividend.rb
Witzman: not sure, if i did this right:
- actions returns [] for minors/nationals so the step is automatically skipped (no user prompt)                                           
- skip! auto-selects half for minors and payout for nationals when revenue is positive, falling back to withhold for zero revenue; 
- dividend_types now includes :withhold alongside the primary type for minors/nationals — this is required because process_dividend calls dividend_options(entity)[kind] and would crash if :withhold wasn't in the hash for the zero-revenue case

crericha (approved):

PR #12533, opened 2026-04-25
[18 India] Fix second OO corp crash

ddaybell:
Fixes #12413, #12329, #11719, #11308

## Before clicking "Create"

- [x] Branch is derived from the latest `master`
- [x] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

## Implementation Notes

Note, this comment has been rewritten to document the current state of the code for this fix.  As reflected in the list of comments below, the original code (which was supplemented by an attempt to fix bugs) was needlessly complex and failed to efficiently use existing engine-level processes.  After several attempts to further patch the prior code, I decided to chuck it all and rewrite the OO token placement feature the way it probably should have been written in the beginning.  As explained in detail below, the current code is more efficient and relies more heavily on existing engine-level processes where appropriate, with less purpose-built code.

### Explanation of Change

**Root cause:** The four OO corporations (NWR and SPD at G8/Delhi; EIR and EBR at P17/Kolkata) share a home hex. With `HOME_TOKEN_TIMING = :float`, each corp places its home token when it first operates. If one OO partner had already floated and the shared hex was since upgraded to yellow by the time the second partner floated, the second placement crashed.

**Design decision - tile-level reservations:** An early approach assigned `city: 0` / `city: 1` to each OO corp (NWR/EIR ? city 0, SPD/EBR ? city 1) so each would reserve a specific city slot, shown in the circle. This turned out to cause problems: since tokens aren't placed until float time, the forced city pre-assignment created conflicts in the placement logic, and required complex reservation-shuffling code in `home_track.rb` and `track.rb` to manage the city-vs-hex reservation state as tiles upgraded. Reverting to **tile-level reservations** (removing `city:` from the corp definitions) is cleaner - the corp's initials appear as text labels on the tile until placement, and city assignment is deferred to the moment the token is actually placed. This let us delete `place_second_oo_token`, `swap_higher_value_oo_token`, `replace_oo_reservations`, and `move_oo_reservations` entirely.

**Fix - `game.rb`:**
- `place_home_token` now has an OO-specific branch. On a white tile: places the token directly into the first untokened city (no pending queue needed). On a yellow tile (already upgraded before this corp floated): adds to `pending_tokens` for the tile-laying player to choose a city for each token already on the tile (this behavior conforms to the rules as clarified by FAQ 8.4.1.1, q:2.
[...]

PR #12562, opened 2026-05-03
[18India] Display hand size on Info tab and Entities player card

### LINKED ISSUE #12519: [18India] - Display number of cards in each player's hand
In 18India, every player has a secret hand of share cards, that they can buy from.  While the contents of a player's starting hand is secret, the number of cards is known.  It's fixed at the start of the game, and every buy action from a player's hand is public, so the size of the player's hand is always a publicly determinable fact.

The UI, however, does not display the player's hand size anywhere that I can find.  I would suggest putting hand size either on the Info tab, to the left of the Player Draft History rows, or perhaps on the Entity tab, below each player card.

ddaybell:
Fixes #12519

## Before clicking "Create"

- [x ] Tests pass cleanly with `docker compose exec rack rake`

### Explanation of Change

This is a split off of PR #12533 to isolate the hand size issue from the OO city fix.

Adds the number of shares a player currently holds in their hand to two places in the UI:

Player card (Entities tab): A new "Shares in hand" row appears on each player's card, below the share count row. This is implemented via a new player_card_rows hook in View::Game::Player — any game can implement player_card_rows(player) to inject additional labeled rows into the player card table.

Info tab (Timeline): Each player's draft history entry now includes their current hand size, e.g. Alice (3 cards in hand): CompanyA, CompanyB, CompanyC.

### Screenshots

image

image

### DIFF
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -177,6 +177,11 @@ module View
           h(:td, 'Shares'),
           h('td.right', td_cert_props, (@game.all_corporations.sum { |c| c.minor? ? 0 : num_shares_of(@player, c) }).to_s),
         ])
+        if @game.respond_to?(:player_card_rows)
+          @game.player_card_rows(@player).each do |label, value|
+            trs << h(:tr, [h(:td, label), h('td.right', value)])
+          end
+        end
 
         priority_props = {
           attrs: { colspan: '2' },
--- a/assets/app/view/game/player.rb
+++ b/assets/app/view/game/player.rb
@@ -178,9 +178,8 @@ module View
           h('td.right', td_cert_props, (@game.all_corporations.sum { |c| c.minor? ? 0 : num_shares_of(@player, c) }).to_s),
         ])
         if @game.respond_to?(:player_card_rows)
-          @game.player_card_rows(@player).each do |label, value|
-            trs << h(:tr, [h(:td, label), h('td.right', value)])
-          end
+          label, value = @game.player_card_rows(@player)
+          trs << h(:tr, [h(:td, label), h('td.right', value)])
         end
 
         priority_props = {
--- a/lib/engine/game/g_18_india/game.rb
+++ b/lib/engine/game/g_18_india/game.rb
@@ -679,7 +679,7 @@ module Engine
         end
 
         def player_card_rows(player)
-          [['Shares in hand', player.hand.size.to_s]]
+          { 'Hand size' => player.hand.size.to_s }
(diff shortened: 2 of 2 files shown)


ollybh (changes requested):
Is there a reason that `player_card_rows` returns an array of arrays? The code would be slightly simpler if it just returned a single label/value pair.

Also, this doesn't render well on the browser I tested with (Chromium on Linux):
Image

Could the label be made shorter to remove the line break? `Hand shares` maybe?

ddaybell:
Both concerns addressed.  code returns a hash now instead of the nested array.  Text says "hand size" now, which should avoid the line wrap in most instances.

lib/engine/game/g_18_india/game.rb
```diff
@@ -625,7 +625,7 @@ def timeline
 
           timeline << 'Player Draft History'
           @players.each do |p|
-            timeline << "#{p.name}: #{p.draft_history.join(', ')}"
+            timeline << "#{p.name} (#{p.hand.size} cards in hand): #{p.draft_history.join(', ')}"
```
ollybh: If a player has a single share in hand this will show as `1 cards` which fails basic grammar.
```suggestion
            hand_size = "#{p.hand.size} card#{p.hand.one? ? '' : 's'} in hand"
            timeline << "#{p.name} (#{hand_size}): #{p.draft_history.join(', ')}"
```

ollybh: Or:
```suggestion
            hand_size =
              case p.hand.size
                when 0 then 'no cards'
                when 1 then '1 card'
                else "#{p.hand.size} cards"
              end
            timeline << "#{p.name} (#{hand_size} in hand): #{p.draft_history.join(', ')}"
```

assets/app/view/game/player.rb
```diff
@@ -177,6 +177,11 @@ def render_info
           h(:td, 'Shares'),
           h('td.right', td_cert_props, (@game.all_corporations.sum { |c| c.minor? ? 0 : num_shares_of(@player, c) }).to_s),
         ])
+        if @game.respond_to?(:player_card_rows)
+          @game.player_card_rows(@player).each do |label, value|
+            trs << h(:tr, [h(:td, label), h('td.right', value)])
+          end
```
ollybh: This is still iterating over `player_card_rows`. There will only ever be a single row for this, so an iterator isn't needed.

ddaybell:
Comments addressed.

ollybh (approved):

PR #12563, opened 2026-05-03
18India: use theme font color in map legend tables

ddaybell:
In 18 India, the commodity tables use hardcoded colors in certain places.  Specifically, commodity_legend and connection_legend both accepted a font_color parameter but ignored it, hardcoding 'black' instead. This makes the legend text invisible for users with dark background themes.

## Before clicking "Create"

### Explanation of Change

Updated both sections of code to use the main font color for these legends, instead of hardcoded black.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_18_india/game.rb
+++ b/lib/engine/game/g_18_india/game.rb
@@ -1222,10 +1222,10 @@ module Engine
           %i[commodity_legend connection_legend]
         end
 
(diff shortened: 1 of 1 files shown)


perwestling:
I find Destination and Hex a bit hard to read on the yellow background. Is it possible to have black foreground for thos?

ddaybell:
Changed to black for the two banners (yellow and green)

image

image

ollybh (approved):

PR #12564, opened 2026-05-04
[Core] Implement game log indentation

### LINKED ISSUE #12400: Improve the log readability and some features

ddaybell:
Fixes #12400

## Before clicking "Create"

- [x ] Code passes linter with `docker compose exec rack rubocop -a`
- [x ] Tests pass cleanly with `docker compose exec rack rake`

Summary
This PR creates an indenter for the game log, to visually group all related actions for the currently operating entity (corporation, player or in auctions the entity being auctioned).  The standard behavior is to write the first action for a given group without indent, and to indent all subsequent actions related to the operating entity.  

Provides an explicit indent_group field stamped on each GameLog::Entry at write time. The round machinery maintains @log.indent_group and sets it at well-defined moments, making indentation deterministic and game-rule-driven.

Changes to the GameLog#<< (shovel) operator
Before: The Entry stored only message and action_id. Indentation was not recorded at write time.

Added in this PR: Each Entry now also stores indent_group — a string (typically an entity id) stamped at write time from @log.indent_group as maintained by the round machinery. One special case is handled directly in <<: if an auction step with an auctioning_lot is active, the lot's id is used regardless of what the round machinery has set. This allows for grouping by the entity (private, minor or corporation) being auctioned when an auction is underway.  

This is expected to be the first module which extends the information saved with log entries.  Future development can add additional information to the log shovel def, to allow for retention of additional information in the log entries.  

Available at write time for future PRs: The full @game object is accessible in <<, including @game.round, @game.turn, @game.phase, @game.current_entity, and @game.players. Any of these could be stamped onto entries to support features like per-phase coloring, turn-number display in the log, or player-color highlighting without requiring a second pass over the log.  Additional information could be provided as needed, with the addition of further fields in the shovel, and updating of those fields at write time.  For example, recording the track objects laid in tile-lays.

How indentation works:

Round::Base#initialize clears @log.indent_group = nil at round creation
Round::Base#process_action sets indent_group to the acting entity's id via the overridable indent_group_for_action hook, then resets to the current entity after the action completes
Round::Operating overrides indent_group_for_action to always use @current_operator.id, so private-company actions taken during a corporation's OR turn are correctly grouped under that corporation
Round::Stock#start_entity sets the indent_group to the current player; next_entity! clears it at round end
game/base.rb#transition_to_next_round! clears indent_group before starting the next round so round-opening banners are never indented

Test plan
 Verified indentation behavior visually across 18NY (auction round, OR), 1871 (SR with private conversion + share buy, OR with PEIR), 1861 (merger round), 18 India (OR with private company track action), and 18Ireland.
 bundle exec rspec spec/lib/engine/game/fixtures_spec.rb — 7429 examples, 0 failures
 rubocop — no offenses on all 7 modified files

### Screenshots

image

image

image

### DIFF
--- a/lib/engine/game_log.rb
+++ b/lib/engine/game_log.rb
@@ -2,18 +2,31 @@
 
 module Engine
   class GameLog < Array
+    attr_writer :indent_group
+
     def initialize(game)
       super()
       @game = game
+      @indent_group = nil
     end
 
     def <<(message)
-      message = Entry.new(message, @game.current_action_id) unless message.is_a?(Entry)
-      super(message)
+      entry = message.is_a?(Entry) ? message : Entry.new(message, @game.current_action_id)
+      begin
+        step = @game.round&.steps&.find { |s| s.respond_to?(:auctioning_lot) && s.auctioning_lot }
+        entry.indent_group = if step
+                               step.auctioning_lot.id.to_s
+                             else
+                               @indent_group
+                             end
+      rescue StandardError
+        # may raise during game initialization before a round exists
+      end
+      super(entry)
     end
 
     class Entry
-      attr_accessor :message, :action_id
+      attr_accessor :message, :action_id, :indent_group
 
       def initialize(message, action_id)
         @message = message
--- a/lib/engine/round/base.rb
+++ b/lib/engine/round/base.rb
@@ -22,6 +22,7 @@ module Engine
       def initialize(game, steps, **opts)
         @game = game
         @log = game.log
+        @game.log.indent_group = nil
         @entity_index = 0
         @round_num = opts[:round_num] || 1
         @entities = select_entities
@@ -92,11 +93,13 @@ module Engine
         raise GameError, "No step found for action #{type} at #{action.id}: #{action.to_h}" unless step
 
         step.acted = true
+        @game.log.indent_group = indent_group_for_action(action)
         step.send("process_#{action.type}", action)
 
         @at_start = false
 
         after_process_before_skip(action)
+        @game.log.indent_group = @entities[@entity_index]&.id&.to_s
         skip_steps
         clear_cache!
         after_process(action)
@@ -216,6 +219,11 @@ module Engine
       end
 
       def highest_bid(_entity); end
+
+      def indent_group_for_action(action)
+        act = action.entity
+        (act&.company? ? act.owner : act)&.id&.to_s
+      end
 
       private
 
--- a/lib/engine/round/operating.rb
+++ b/lib/engine/round/operating.rb
@@ -85,7 +85,8 @@ module Engine
 
         @current_operator = entity
         @current_operator_acted = false
-        @log << "#{@game.acting_for_entity(entity).name} operates #{entity.name}" unless finished?
+        @game.log.indent_group = entity&.id&.to_s
+        @log << "-- #{@game.acting_for_entity(entity).name} operates #{entity.name} --" unless finished?
         @game.place_home_token(entity) if @home_token_timing == :operate
         skip_steps
         return unless finished?
@@ -97,6 +98,10 @@ module Engine
       def recalculate_order
         unsorted_corps = @entities.pop(@entities.size - @entity_index - 1)
         @entities.concat(@game.operating_order.select { |e| unsorted_corps.include?(e) })
+      end
+
+      def indent_group_for_action(_action)
+        @current_operator&.id&.to_s || super
       end
 
       def operating?
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -102,6 +102,7 @@ module View
         blank_action.created_at = @game.actions[0]&.created_at || Time.now
 
         last_action = nil
+        @last_indent_group = nil
 
         actions = @game.actions.to_h { |a| [a.id, a] }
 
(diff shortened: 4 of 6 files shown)


lib/engine/round/base.rb
crericha, line 1: base round should know nothing about log formatting. We should look for another way to implement indenting. Originally, I thought you were going to include the entity in the log action and group by that.

lib/engine/round/operating.rb
crericha, line 1: Ditto, none of the round classes should know about indenting.

lib/engine/game_log.rb
```diff
@@ -2,18 +2,31 @@
...
       @game = game
+      @indent_group = nil
     end
 
     def <<(message)
-      message = Entry.new(message, @game.current_action_id) unless message.is_a?(Entry)
-      super(message)
+      entry = message.is_a?(Entry) ? message : Entry.new(message, @game.current_action_id)
+      begin
+        step = @game.round&.steps&.find { |s| s.respond_to?(:auctioning_lot) && s.auctioning_lot }
```
crericha, line 16: This is too specific for the log class. It should only care about logging. It shouldn't know anything about auctions.

ddaybell:
Thanks for the feedback —  I see your point about keeping the log formatting out of the round classes.  

I'd like to use just the entity information that is available at log write time, but there's an issue with a couple of use cases.  

First, with private company auctions the log is more readable if the company being auctioned is not indented, and all of the bids on that company are.  That's the natural way an auction should be grouped.  But, during an auction, active_step.current_entity returns the bidding player, and that changes with each bid as the rotation advances. If we stamp current_entity on each entry, bids from Player1, Player2, and Player3 on the same private all get different group keys — so nothing gets indented together. What we actually want is for all bids on a given company to be grouped under that company's ID.

That's what the auctioning_lot check in GameLog#<< is doing: when a lot is active, it uses the lot's ID as the group key rather than the acting player's ID. If we don't want the log class knowing specifically about auctioning_lot, we could use a generic log_group_id method on steps that steps can optionally implement — auction steps return the lot's id, everything else returns nil. The log just calls that if available and falls back to current_entity. No auction-specific knowledge in the log class.

Second, during an OR when a private company takes a special action (e.g. a tile lay), current_entity returns the company ID rather than the operating corporation's ID. This causes the private company to be seen as a new entity, and it is not indented (nor is the following line when the corporation starts back up as the active entity).  But the action is best seen as part of the acting corporation's turn, and thus should be indented along with all of the other corporation actions.  The current implementation handles this via current_operator in the operating round. I could instead read @game.round.current_operator directly in << via respond_to? — this would avoid changes to the round class.

Happy to rework the implementation however you'd like. Do you have a specific approach in mind?

ollybh (changes requested):
Setting this to changes required, until @crericha's comments have been addressed.

PR #12565, opened 2026-05-04
[Core] Add My Last Move button to game log

### LINKED ISSUE #10369: Add a "Review since my last action" button to the history navigator

ddaybell:
Adds a button to the log toolbar that scrolls to and highlights the current player's most recent action, making it easy to find where you left off in a long game.

Fixes #10369

## Before clicking "Create"

- [x ] Branch is derived from the latest `master`

### Screenshots

Before:
image

After:
image

### DIFF
--- a/assets/app/view/game/game_log.rb
+++ b/assets/app/view/game/game_log.rb
@@ -22,9 +22,9 @@ module View
       needs :show_log, default: true, store: true
 
       def render
-        children = [render_log_choices, render_log]
-
         @player = @game.player_by_id(@user['id']) if @user
+
+        children = [render_log_choices, render_log]
 
         key_event = lambda do |event|
           event = Native(event)
@@ -164,6 +164,24 @@ module View
         h('div#chatlog', props, the_log)
       end
 
+      def player_for(entity)
+        return nil unless entity
+        return entity if @game.players.include?(entity)
+
+        if entity.respond_to?(:player)
+          found = entity.player
+          return found if @game.players.include?(found)
+        end
+
+        if entity.respond_to?(:owner)
+          owner = entity.owner
+          return owner if @game.players.include?(owner)
+          return player_for(owner) if owner && owner != entity
+        end
+
+        nil
+      end
+
       def render_log_for_action(log, action)
         timestamp_props = {
           style: {
@@ -214,7 +232,16 @@ module View
           action_log << render_action_buttons(action.id)
         end
 
-        h(:div, action_log)
+        h(:div, { attrs: { id: "action-#{action.id}" } }, action_log)
+      end
+
+      def last_player_action_id
+        return nil unless @player
+
+        @game.actions.reverse_each do |action|
+          return action.id if player_for(action.entity) == @player
+        end
+        nil
       end
 
       def render_action_buttons(action_id)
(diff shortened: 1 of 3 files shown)


assets/app/view/game/game_log.rb
```diff
@@ -164,6 +164,24 @@ def render_log
         h('div#chatlog', props, the_log)
       end
 
+      def player_for(entity)
+        return nil unless entity
+        return entity if @game.players.include?(entity)
```
ollybh: Simpler:
```suggestion
        return entity if entity.player?
```

assets/app/view/game/game_log.rb
```diff
@@ -164,6 +164,24 @@ def render_log
...
+
+        if entity.respond_to?(:player)
+          found = entity.player
+          return found if @game.players.include?(found)
+        end
```
ollybh: Could you do this?
```suggestion
        return entity.player if entity.player&.player?
```

assets/app/view/game/game_log.rb
```diff
@@ -164,6 +164,24 @@ def render_log
...
+
+        if entity.respond_to?(:owner)
+          owner = entity.owner
+          return owner if @game.players.include?(owner)
+          return player_for(owner) if owner && owner != entity
+        end
```
ollybh:
```suggestion
        if (owner = entity.owner&)
          return owner if owner.player?
          return player_for(owner)
        end
```

assets/app/view/game/game_log.rb
```diff
@@ -164,6 +164,24 @@ def render_log
...
+
+        nil
```
ollybh: Rolling this all together, you could have something like this:
```suggestion
        if entity&.player?
          entity
        elsif (owner = entity&.owner)
          owner.player? ? owner : player_for(entity.owner)
        else
          nil
        end
```

ollybh (commented):
For the UI, I'd think that it would make more sense to have the new button next to the existing history navigation buttons below the game log.

At the moment nothing happens if you click the button again. Should it take you to your previous move? A keybinding would be useful to.

I'm not keen on all the `respond_to` checks in `player_for`. You can use the save navigation operator (`&`) to avoid these checks which gives you simpler code. And you can check `foo.player?` instead of `@game.players.include?(foo)`.

ddaybell:
Thanks for the review! Here's what changed in response:

**Button placement** — The button has been replaced with two navigation links, `My <` and `My >`, placed in the history bar between `<` (Previous Action) and `>` (Next Action). They jump to the previous/next action taken by the current player (or their corporations in OR), matching the existing navigation style.

**Repeated clicking / keybinding** — Clicking `My <` repeatedly steps back through the player's actions one at a time. `My >` steps forward. Shift+← and Shift+→ are the keyboard shortcuts.

**`respond_to` / player detection** — Removed entirely. The player is now found via `hotseat?` (which uses `acting_for_entity(current_entity)` for hotseat games, or `@game.player_by_id(@user['id'])` for online games). Corporation ownership uses `player.presidencies` from `ShareHolder`. No `respond_to` calls remain.

**Hotseat games** — The earlier approach keyed off `@user`, which is the logged-in user and meaningless in a hotseat game. The fix detects the active player at the time of the first render (when there's no history cursor) and pins their id in a `history_player_id` store so it doesn't drift while navigating history.

One known limitation: automatic end-of-turn actions (e.g. transitioning to the next company's turn) are attributed to the corporation in the action log, so `My <` may land on the last of those automatic actions rather than the last voluntary one. There's no reliable way to distinguish voluntary from automatic actions in the data.

ddaybell:
image

PR #12567, opened 2026-05-05
18USA: log subsidies removed from pool at game start

### LINKED ISSUE #12536: [18USA] Add removed subsidies to the log
https://18xx.games/game/249866?action=0

Image

> [11:26] Removing +10/20 and $50 subsidies

Feature request: add a third entry indicating which two subsidies were not selected for the map.

ddaybell:
During setup, 18USA randomly selects 14 of 16 subsidies to place on the map. Previously the 2 excluded subsidies were silently discarded with no log entry.

This change logs those excluded subsidies at game start so all players can see what didn't make it onto the map.

Example log entry:

Removing: No Subsidy, $50 Subsidy

Closes #12536

## Before clicking "Create"

### Screenshots
image

### DIFF
--- a/lib/engine/game/g_18_usa/game.rb
+++ b/lib/engine/game/g_18_usa/game.rb
@@ -372,13 +372,15 @@ module Engine
             'Placing a track and the resource token from the Resource Subsidy is a free extra ' \
             'track lay in addition to the normal track placements.'
 
-          @log << "Resource subsidy includes #{resources.join(', ')}"
+          @resource_subsidy_resources = resources.join(', ')
         end
 
         def randomize_subsidies
-          randomized_subsidies = @subsidies.sort_by { rand }.take(SUBSIDIZED_HEXES.size)
-          excluded = @subsidies - randomized_subsidies
-          @log << "Removing: #{excluded.map { |s| s['name'] }.join(', ')}" if excluded.any?
+          shuffled = @subsidies.sort_by { rand }
+          randomized_subsidies = shuffled.take(SUBSIDIZED_HEXES.size)
+          excluded = shuffled.drop(SUBSIDIZED_HEXES.size)
+          @log << "Resource subsidy includes #{@resource_subsidy_resources}" unless excluded.any? { |s| s[:id] == 'S16' }
+          @log << "Removing: #{excluded.map { |s| s[:name] }.join(', ')}" unless excluded.empty?
           @subsidies_by_hex = {}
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_usa/game.rb
```diff
@@ -377,6 +377,8 @@ def setup_resource_subsidy
 
         def randomize_subsidies
           randomized_subsidies = @subsidies.sort_by { rand }.take(SUBSIDIZED_HEXES.size)
+          excluded = @subsidies - randomized_subsidies
+          @log << "Removing: #{excluded.map { |s| s['name'] }.join(', ')}" if excluded.any?
```
crericha: `excluded.any?` is an expensive method. Please change to `!excluded.empty?`, which is a simple mathematical check.

lib/engine/game/g_18_usa/game.rb
```diff
@@ -377,6 +377,8 @@ def setup_resource_subsidy
 
         def randomize_subsidies
           randomized_subsidies = @subsidies.sort_by { rand }.take(SUBSIDIZED_HEXES.size)
+          excluded = @subsidies - randomized_subsidies
```
crericha: If we keep `@subsidies.sort_by { rand }` in a var, you can avoid doing a full array to array comparison.

```
randomized_subsidies = @subsidies.sort_by { rand }
used_subsidies = randomized_subsidies.take(SUBSIDIZED_HEXES.size)
excluded_subsidies = randomized_subsidies.last(max(randomized_subsidies.size() - SUBSIDIZED_HEXES.size), 0))
```

crericha:
This PR points out another minor log issue. The resource subsidy may not be in the game and in that situation, its contents shouldn't be logged. If you feel up to it, it'd be nice to fix that one too.

ddaybell:
Original: setup_resource_subsidy always logged "Resource subsidy includes oil, coal" at game start, even if S16 was one of the two subsidies randomly excluded from the map.

Changes made to randomize_subsidies:

Store the shuffled array in a variable (shuffled) so the excluded subsidies can be derived via .drop instead of a full array subtraction (@subsidies - randomized_subsidies)
Replace if excluded.any? with unless excluded.empty? (O(1) size check vs O(n) iteration)
Fix key access from s['name'] to s[:name] to match the symbol keys used in the SUBSIDIES constant
Changes made to setup_resource_subsidy:

Remove the log line; instead save the resource string to @resource_subsidy_resources
The resource subsidy log is now emitted in randomize_subsidies, but only if S16 actually made it onto the map

crericha (approved):

PR #12572, opened 2026-05-07
[1832] Update constants for use with variants, fix other minor mistakes.

jemiahlee:
Updates a lot of constants in the code to allow for the variants, and then adds the methods within game.rb to optionally choose those variants.

This is the first of several PRs I will be creating for this game implementation, which is mostly done now.

## Before clicking "Create"

- [x] (N/A) Add the `pins` or `archive_alpha_games` label if this change will break existing games

## Implementation Notes

using the `game_` method for train roster, phase roster, private companies, and the market, all for the variants.

Atl was too long a string for the Atlanta tile in brown, so made them all just "A"


lib/engine/game/g_1832/market.rb
```diff
@@ -18,6 +18,20 @@ module Market
           %w[0c 0c 0c 10b 20b 30b 40o],
         ].freeze
 
+        GAME_END_AT_400_MARKET = [
```
perwestling: As the only change (as far as I can see) is the "e" added for 400 - would it not be possible to implement it so that same base is used in both cases, only with "e" added if game end at 400 option is used?

jemiahlee: I'm not necessarily opposed to adding logic for this instead, but I also think that it's nice for the constants to be true constants. It also reduces the amount of judgment by an implementer on "how much logic is too much logic for this thing that is a constant in other games?"

Having said that, I can change it. I was just trying to follow the ethos of the repo.

jemiahlee:
Did the same for the rest of the constants. Thanks @perwestling for the feedback.

perwestling (approved):

lib/engine/game/g_1832/game.rb
```diff
@@ -53,14 +68,13 @@ class Game < Game::Base
...
 
           self.class::TILE_LAYS
         end
 
-        def system?(corporation)
-          return false unless corporation
-
-          corporation.type == :system
+        def system?(entity)
+          entity.is_a?(Corporation) && entity.type == :system
```
ollybh: Why is this change needed?

If it is necessary, then checking `entity.corporation?` is in line with the code base, not `entity.is_a?(Corporation)`.

lib/engine/game/g_1832/game.rb
```diff
@@ -147,6 +162,30 @@ def setup
...
+          companies
+        end
+
+        def game_market
+          market = self.class::PROTO_MARKET.dup
+          first_row = market[0].dup
+          first_row.push(@optional_rules&.include?(:finish_on_400) ? '400e' : '400')
+          market[0] = first_row
+          market
+        end
```
ollybh, line 204: Not having the 400 value space included in the market definition constant makes the code harder to understand.

How about something like:
```
def game_market
  return self.class::MARKET unless @optional_rules&.include?(:finish_on_400)

  market = self.class::MARKET.map(&:dup)
  market[0][-1] = '400e'
  market
end
```

ollybh: Or it might be better to override `init_stock_market` and swap in a SharePrice with the `endgame` flag set.

lib/engine/game/g_1832/meta.rb
```diff
@@ -8,27 +8,47 @@ module G1832
...
+            short_name: 'Diesels',
+            desc: '8- and 10-trains are removed. 12-trains become 1830-style Diesels available '\
+                  'after the first 6-train. A train may be traded in on a Diesel for a $300 credit. '\
+                  '5-trains are permanent.',
+          },
+          {
+            sym: :southern_bank,
+            short_name: 'Southern Bank',
+            desc: 'Adds P6, the Southern Bank private company. Starting capital increases to '\
+                  '$2400 divided by the number of players. This private currently does not do '\
```
ollybh, line 39: Why does it 'currently' not do anything?

ollybh (commented):
Several game have adopted a pattern of adding methods to their Game class for checking which optional rules have been set.

Something like:
```
def option_diesel?
  @optional_rules&.include(:diesels)
end
```

This makes the code that tests for the rules more concise.

lib/engine/game/g_1832/game.rb
jemiahlee: I have the implementation basically fully working, so I was just selecting changes, and this was just something that got included. It seemed boilerplate/innocuous enough. I will update to the idiom though.

lib/engine/game/g_1832/game.rb
jemiahlee: I think my preference would be to set two vars to be standard and variant first market lines and combine them similar to what I did with the trains and phases. I'll update the PR for that, I think it's cleaner and clearer to someone reading the code.

lib/engine/game/g_1832/meta.rb
jemiahlee: Well, the behaviors of the Southern Bank are weird and not simple. If it gets implemented, it will be the last thing to implement, because I'm not sure that it will be used. JCL loves it, Bill Dixon told me he hasn't used it in many years.

The current implementation treats it as a sellable private and does increase the starting capital. My preference is to leave it in aspirationally for right now and remove it when moving to alpha if not fully implemented. That also preserves it within git in case it doesn't get fully implemented and someone later comes along fully motivated about its weirdness (I will make an attempt at making it work).

lib/engine/game/g_1832/market.rb
```diff
@@ -4,15 +4,17 @@ module Engine
...
-          %w[64y 68 72 76 82 90 100p 110 120 140 160 180 200 225 250 275 300 325 350 375 400],
+        STANDARD_TOP_LINE = %w[64y 68 72 76 82 90 100p 110 120 140 160 180 200 225 250 275 300 325 350 375 400].freeze
+        FINISH_ON_400_TOP_LINE = %w[64y 68 72 76 82 90 100p 110 120 140 160 180 200 225 250 275 300 325 350 375 400e].freeze
+
+        REST_OF_MARKET = [
           %w[60y 64y 68 72 76 82 90p 100 110 120 140 160 180 200 225 250 275 300 325 350 375],
           %w[55y 60y 64y 68 72 76 82p 90 100 110 120 140 160 180 200 225 250i 275i 300i 325i 350i],
           %w[50o 55y 60y 64y 68 72 76p 82 90 100 110 120 140 160i 180i 200i 225i 250i 275i 300i 325i],
-          %w[40b 50o 55y 60y 64 68 72p 76 82 90 100 110i 120i 140i 160i 180i],
+          %w[40o 50o 55y 60y 64 68 72p 76 82 90 100 110i 120i 140i 160i 180i],
```
jemiahlee, line 14: These had been in this codebase, I believe, when I first started on this a few years ago. I cross-indexed with my copy's stock market and discovered these discrepancies.

ollybh (approved):

PR #12578, opened 2026-05-10
[Core] intra-hex province border lines via partition type :province - Required for 18OE

Witzman:
intra-hex province border lines via partition type :province
## Why this is needed for 18OE

18OE draws national-zone province borders as orange dashed lines along
hex edges using border=edge:N,type:province. Where a province boundary
crosses through a hex diagonally rather than along an edge, the border
must be drawn as a line from a hex corner to the hex centre and onward
to another corner. Hex O38 is the first such case: the boundary enters
at corner 2 (upper-left) and exits at corner 5 (lower-right), passing
through the centre.

grafik

The existing partition DSL handles corner-to-corner lines but had two
gaps that made this impossible without engine changes:

1. No centre-vertex support — the DSL parser reads only the first
   character of the 'a' string (a[0].to_i), so '-1' was silently
   misread as 0. lib/engine/part/partition.rb is fixed to detect
   negative inputs via start_with?('-') and parse them with to_i.
   @inner/@outer are set to [] for centre partitions because edge-
   direction restriction has no meaning when one endpoint is the
   centre.

2. No :province render type in Partitions — View::Game::Part::Partitions
   only knew nil/mountain/water/impassable/divider. :province is added
   with orange colour (matching View::Game::Part::Borders) and
   stroke-dasharray '20 20' (identical to the dashed style used for
   province edge borders). The render guard is widened from
   `type != :divider` to `!%i[divider province].include?(type)` so
   province partitions always render without requiring a blocker company,
   consistent with their purely decorative role.

## Why no other titles are affected

Engine file changes:

lib/engine/part/partition.rb:
  The start_with?('-') branch is only entered when the 'a' parameter
  begins with a minus sign. A grep of all HEXES definitions across the
  entire game library confirms that no existing title passes a negative
  vertex value to partition=. Every current DSL value is in the range
  0–5, optionally suffixed with + or -, none of which start with '-'.
  The inner/outer change is equally guarded by @A == -1.

assets/app/view/game/part/partitions.rb:
  Titles using typed partitions fall into three groups:
  - type:divider (g_1862, g_18_ny, g_18_ny_1e, g_18_gb): none of these
    use blocks_partition abilities. In Opal 1.8.2 (the project's runtime)
    string/symbol equality holds in === comparisons, so 'divider' already
    matched :divider in the old guard and these partitions already always
    rendered. The new guard !%i[divider province].include?(:divider) is
    false for the same reason — no change.
  - type:water (g_1870, g_18_mag, g_18_rhl): these games DO use
    blocks_partition abilities. 'water'.to_sym == :water, and :water is
    absent from %i[divider province], so the guard result is identical to
    the old `'water' != :divider` (true in both cases). Rendering remains
    gated on the blocker company being active — unchanged.
  - stroke-dasharray: 'none' is now set explicitly on non-province
    partitions. 'none' is the SVG default; the visual output is identical.
  Adding :province to the color case and the guard affects only partitions
  whose type is 'province'. No title other than 18OE declares such a
  partition.


assets/app/view/game/part/partitions.rb
```diff
@@ -29,7 +29,7 @@ class Partitions < Base
 
         def color(partition)
           color =
-            case partition.type
+            case partition.type&.to_sym
```
crericha: I believe partition types are always symbols. Is this change necessary?

assets/app/view/game/part/partitions.rb
```diff
@@ -53,12 +55,16 @@ def render_part
           children = []
 
           @tile.partitions.each do |partition|
-            next if partition.type != :divider && partition.blockers.none? do |blocker|
+            next if !%i[divider province].include?(partition.type&.to_sym) && partition.blockers.none? do |blocker|
```
crericha: Same comment about `to_sym`

assets/app/view/game/part/partitions.rb
```diff
@@ -53,12 +55,16 @@ def render_part
...
               !@game || @game&.abilities(blocker, :blocks_partition)&.blocks?(partition.type)
             end
 
             a_control = VERTICES[(partition.a + partition.a_sign) % 6]
-            vertex_a = convex_combination(VERTICES[partition.a], a_control)
+            vertex_a = if partition.a == -1
+                         VERTICES[-1]
```
crericha: -1 means last value in the array. Do we not know the size of the array, so that we can set partition.a to a positive value?

assets/app/view/game/part/partitions.rb
```diff
@@ -90,6 +96,7 @@ def render_part
                             d: d,
                             stroke: color(partition),
                             'stroke-width': '8',
+                            'stroke-dasharray': partition.type&.to_sym == :province ? '20 20' : 'none',
```
crericha: Remove to_sym, if we can

lib/engine/part/partition.rb
```diff
@@ -17,9 +17,16 @@ def initialize(a, b, type, restrict)
...
         a, b = [a, b].minmax
-        @a = a[0].to_i
-        @a_sign = SIGN[a[1]]
+        if a.start_with?('-')
+          @a = a.to_i
+          @a_sign = 0
+        else
+          @a = a[0].to_i
+          @a_sign = SIGN[a[1]]
+        end
```
crericha: This feels kind of hacky for adding a -1 special case. Is there a way to simplify?

lib/engine/part/partition.rb
```diff
@@ -28,8 +35,13 @@ def initialize(a, b, type, restrict)
         @restrict = restrict
         @blockers = []
 
-        @inner = (restrict == 'outer' ? [] : (@a..(@b - 1)).to_a)
-        @outer = (restrict == 'inner' ? [] : (0..5).to_a - (@a..(@b - 1)).to_a)
+        if @a == -1
+          @inner = []
+          @outer = []
```
crericha: Does this special case make sense? Couldn't we do something like `(@a.positive? && restrict == 'outer' ? ...)`.

Also, can you help me understand why going to the center needs special inner and outer handling?

assets/app/view/game/part/partitions.rb
Witzman: VERTICES is a Hash, not an Array — VERTICES[-1] is a direct key lookup (=> [0, 0]), not a Ruby array last access.

Why we can't use a positive index (e.g., 6): The DSL parser does a, b = [a, b].minmax on strings before parsing. '-1' has '-' (ASCII 45) which sorts before '0'–'5' (ASCII 48–53), so '-1' always lands in a. If we used '6', it sorts after '5', landing in b — then @b = b[0].to_i would set @b = 6, and we'd need an identical special-case block for b == 6. The −1 sentinel is load-bearing in the string-sort semantics of minmax.

Optionally, i can add a brief comment to partition.rb:22 explaining why −1:
like # '-1' sentinel: as a string it sorts before '0'-'5' so it always lands in a after minmax

lib/engine/part/partition.rb
Witzman: ` @inner / @outer define which hex edges lie on each side of the partition for blocking logic (e.g. g_1870, g_18_mag). `
Was finding from my research
So i wanted a way that wont mess up with anything i dont know about :)

crericha: My suggestion keeps the logic you want, but is less verbose.

lib/engine/part/partition.rb
Witzman: Done — replaced with `a.to_i` which handles negative strings natively (Ruby parses `'-1'` as `-1` directly), then `@a.negative?` to suppress the sign for centre partitions. No branch needed.

lib/engine/part/partition.rb
Witzman: Used `@a.negative?` rather than `@a.positive?` — `positive?` is false for both `-1` and `0`, which would break normal partitions whose lower vertex is 0 (bottom of hex). `negative?` is true only for `-1`, so it's the precise guard. Result is the two-line ternary form you suggested:

```ruby
@inner = (@a.negative? || restrict == 'outer') ? [] : (@a..(@b - 1)).to_a
@outer = (@a.negative? || restrict == 'inner') ? [] : (0..5).to_a - (@a..(@b - 1)).to_a
```

Witzman:
After checking all other titles that use `partition=`, the `0`=centre convention would require bumping corner vertex numbers in g_1862, g_1870, g_18_ny, g_18_ny_1e, g_18_rhl, g_18_gb, and g_18_mag — too wide a blast radius for a single-game feature PR.

Instead the PR now uses **`a.to_i`** to parse the endpoint string, which handles `'-1'` naturally without any branch (`'-1'.to_i == -1`), and **`@a.negative?`** as the centre guard. Corner vertices remain 0–5 for all existing games; 18OE writes `a:-1` in its map for the centre endpoint. No other title is touched.

crericha (approved):

PR #12580, opened 2026-05-11
[1880Romania] train count

### LINKED ISSUE #12575: [1880 Romania] Incorrect number of 6/6E trains, and also game end trigger
Are the train info and game end triggers correct?

Here is how it looks in the game:

Image

My copy of this from Lonny looks like:

Image

I am currently implementing code for this in Transilvania variant, but it does have even fewer trains.

Also notes that 8E and 2P is not present in Lonny's data.

**How to Reproduce the Bug?**

Create game, check game info

**Which Browser / Platform?**

Any

philcampeau:
Fixes #12575

## Before clicking "Create"

### Explanation of Change

_Accidentally left old 1880 train roster. Corrected to the new count._  - no longer accurate.

After the release of the official rules, this PR contains multiple changes including updates to train roster, tile roster, and link to game rules, as well as an updated fixture file to account for these changes.


philcampeau:
Annoyingly, Lonny has just posted the official rules, and it looks like the train count has changed since the draft he sent me. 

I'm confirming with him and I'll update here later with a new fixture.

philcampeau:
good grief. Ok Lonny confirmed that the restored 2 trains do exist in this game, even though they aren't mentioned in the train roster.

image

perwestling (approved):

ollybh (approved):

PR #12582, opened 2026-05-12
Add constant to allow games to always show par price on charter

### LINKED ISSUE #10406: [1850/all] initial par price not visible
In our game of 1850, https://18xx.games/game/151326, we noticed that the IPO price is not shown when the stock is all out of the IPO.  If shares are reissued, the market price may not be high enough to produce a reissue price over the initial IPO price, and then the price would be the initial IPO price.  This may also occur in 1870.

### LINKED ISSUE #10645: 18ESP PAR value
Hello,

18ESP is a game that have the capitalization and the stations according to achieve some goals. While the companies have shares on the IPO we can see the PAR value (as IPO value). But, when the last IPO share is sold, the IPO row disappear from the charter.

It is similar coded as other games. What is then the problem? Each company has 3 goals. When they achieve the first goal, they receive the PAR (or IPO) value x1; With the second goal, PAR value x2; With the third, PAR value x3.
What happens if a company has not more IPO shares and the row disappear, and still has some goals to achieve? Easy, nobody know how many money will enter into the treasury when a goal is achieved, because the PAR value (= IPO) has been removed.

For making the code as easier as possible, I suggest: on the bottom of the charter the goals are listed, for knowing what goals are achieved and what left. I would add simply a little note: PAR = x (the PAR value that the director chosen when bought the director certificate). In this way, all player can in any moment check the potential money a company could reach in an OR. It is very important for deciding selling shares or not during a SR.

Thank you so much,

Enrique Trigueros (queroscia): designer of 18ESP.

philcampeau:
Fixes #10406
Fixes #10645

## Before clicking "Create"

### Explanation of Change

Some games require players to know the par price of a corporation even after the initial IPO has sold out. For example, 1870, 1850, (and 1832) all allow corporations to re-par at new prices based on the initial par price. 

I've implemented a constant, ALWAYS_SHOW_PAR_PRICE, that is checked against the code that hides that row once the IPO is empty. By default it's set to false, and games that need it can be set to true. 

### Screenshots

Before change: 
image

After change: 
image

### DIFF
--- a/assets/app/view/game/corporation.rb
+++ b/assets/app/view/game/corporation.rb
@@ -462,7 +462,10 @@ module View
         num_treasury_shares = share_number_str(@corporation.num_treasury_shares)
 
         pool_rows = []
-        if !num_ipo_shares.empty? || num_dc_avail.positive? || !%i[full none].include?(@corporation.capitalization)
+        if !num_ipo_shares.empty? ||
+           num_dc_avail.positive? ||
+           !%i[full none].include?(@corporation.capitalization) ||
+           @game.class::ALWAYS_SHOW_PAR_PRICE
           pool_rows << h('tr.ipo', [
             h('td.left', @game.ipo_name(@corporation)),
             h('td.right', shares_props, ('d' * num_dc_avail) + num_ipo_shares),
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -188,6 +188,9 @@ module Engine
 
       CAPITALIZATION = :full
 
+      # In some games, it's relevant to know the par price of a corporation even after the IPO sells out
+      ALWAYS_SHOW_PAR_PRICE = false
+
       # Must sell all shares of a company in one action per turn
       MUST_SELL_IN_BLOCKS = false
 
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -188,7 +188,9 @@ module Engine
 
       CAPITALIZATION = :full
 
-      # In some games, it's relevant to know the par price of a corporation even after the IPO sells out
+      # In some games, it's relevant to know the par price of a corporation even after the IPO sells out.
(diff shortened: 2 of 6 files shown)


assets/app/view/game/corporation.rb
```diff
@@ -462,7 +462,10 @@ def render_shares
         num_treasury_shares = share_number_str(@corporation.num_treasury_shares)
 
         pool_rows = []
-        if !num_ipo_shares.empty? || num_dc_avail.positive? || !%i[full none].include?(@corporation.capitalization)
+        if !num_ipo_shares.empty? ||
```
perwestling, line 465: It looks a bit string to have empty? on num_ipo_shares and positive? for num_dc_avail.

I suppose num_ipo_shares is wrong variable name?

philcampeau: That's what was already there. I just added a fourth condition

ollybh: It's not the clearest name for a variable. `num_ipo_shares` is not a number, it is a string that describes the number of shares available. If there are none it will be a zero-length string, which is what is tested for here.

ollybh (approved):

assets/app/view/game/corporation.rb
philcampeau: I wouldn't recommend that we change the variable name at this point.

lib/engine/game/base.rb
```diff
@@ -188,6 +188,9 @@ class Base
 
       CAPITALIZATION = :full
 
+      # In some games, it's relevant to know the par price of a corporation even after the IPO sells out
```
ollybh:
```suggestion
      # In some games, it's relevant to know the par price of a corporation even after the IPO sells out.
      # Setting this to true will make the IPO row always appear on the corporation charter.
      # If false then this row will be omitted when there are no IPO shares remaining.
```

PR #12583, opened 2026-05-12
[1871] Add support for auto-floating from market

### LINKED ISSUE #10146: [1871] autobuy from treasury buys from market
https://18xx.games/game/147167?action=563

**How to Reproduce the Bug?**

When setting autobuy from treasury, the game bought shares from the market instead. Note that this would specifically affect corps that split into another corp, since those are the only corps with Treasury shares. 

Happened on Google chrome v120 and on my android chrome browser as well

https://github.com/tobymao/18xx/assets/26125362/fbc33909-8640-48fd-9f83-06dd4f4e3cf9

philcampeau:
Fixes #10146

## Before clicking "Create"

### Explanation of Change

As implemented, auto float from market shares wasn't working. This is because (I believe) the feature would buy a share from the market, then run super and try to buy a second share, which would fail. 

This corrects that. 

### Screenshots

Before: 

https://github.com/user-attachments/assets/866c3061-53f4-477f-9adc-9d607153a161

After:

https://github.com/user-attachments/assets/7018406c-72d4-4a57-96ea-2e10b1655fbb

### DIFF
--- a/assets/app/view/game/auto_action/buy_shares.rb
+++ b/assets/app/view/game/auto_action/buy_shares.rb
@@ -174,8 +174,12 @@ module View
           )
         end
 
+        def float_from_market
+          @game.class::AUTOFLOAT_FROM_MARKET_SHARES
+        end
+
         def conditions
-          return ['float', false] if @settings['float']
+          return ['float', float_from_market] if @settings['float']
           return [@settings['buy_ipo'].to_i, false] if @settings['ipo']
           return [@settings['buy_market'].to_i, true] if @settings['market']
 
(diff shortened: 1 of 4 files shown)


ollybh:
Can you add some screenshots showing the changes from this PR?

philcampeau:
As I was recording some screenshots, I noticed some unexpected behaviour, and after digging around and correcting things, I came up with a much simpler approach that only modifies the game's buy_sell_par_split_shares.rb file. 

Uploaded video captures in the main body.

PR #12589, opened 2026-05-12
[18FL] set last rank of trains to unlimited

philcampeau:
## Before clicking "Create"

### Explanation of Change

 sets last rank of trains to unlimited


perwestling:
I have not implemented 18FL. Just like 18GA I am wondering if the number of last trains are intentional? That is, that it should intentionally be a limited number of trains?

philcampeau:
image

perwestling (approved):
LGTM

ollybh (approved):

PR #12590, opened 2026-05-12
[18GA] set last rank of trains to unlimited

philcampeau:
## Before clicking "Create"

### Explanation of Change

 sets last rank of trains to unlimited


perwestling (commented):
Rules (http://www.18xx.net/18GA/18GAr.txt) does state 5 8 trains. In practice it might be the same, but....

philcampeau:
Ah you're right. I saw "8 trains unlimited" in the rules, but that's actually a variant rule. I'll just close this one.

Issue #12237, opened 2025-11-12
[various] many games should have unlimited trains


michaeljb:
#12233 is an actual bug for 18RoyalGorge. The fix in #12236 includes general support for unlimited trains.

Games can be checked off of this list when a relevant PR is open, or if they don't require any changes (i.e., it doesn't have any unlimited trains):

- [ ] g_18_al
- [x] g_18_ardennes
- [ ] g_18_carolinas
- [ ] g_18_chesapeake
- [ ] g_18_chesapeake_off_the_rails
- [ ] g_18_christmas_eve
- [ ] g_18_co
- [ ] g_18_cuba
- [ ] g_18_cz
- [ ] g_18_dixie
- [ ] g_18_esp
- [ ] g_18_eu
- [ ] g_18_fl
- [ ] g_18_fr
- [ ] g_18_ga
- [ ] g_18_gb
- [ ] g_18_hiawatha
- [ ] g_18_india
- [ ] g_18_ireland
- [ ] g_18_jpt
- [ ] g_18_ka
- [ ] g_18_los_angeles
- [ ] g_18_los_angeles1
- [ ] g_18_mag
- [ ] g_18_mex
- [ ] g_18_mo
- [ ] g_18_ms
- [ ] g_18_mt
- [ ] g_18_neb
- [ ] g_18_new_england
- [ ] g_18_new_england_north
- [ ] g_18_nl
- [ ] g_18_norway
- [ ] g_18_ny
- [ ] g_18_ny_1e
- [ ] g_18_oe
- [ ] g_18_oe_uk_fr
- [ ] g_18_pa
- [ ] g_18_rhl
- [x] g_18_royal_gorge
- [ ] g_18_scan
- [ ] g_18_sj
- [ ] g_18_texas
- [ ] g_18_tn
- [ ] g_18_tokaido
- [ ] g_18_uruguay
- [x] g_18_usa
- [ ] g_18_va
- [ ] g_18_we
- [ ] g_18_west
- [ ] g_18_zoo
- [ ] g_18_zoo_map_a
- [ ] g_18_zoo_map_b
- [ ] g_18_zoo_map_c
- [ ] g_18_zoo_map_d
- [ ] g_18_zoo_map_e
- [ ] g_18_zoo_map_f
- [ ] g_1804
- [x] g_1807
- [ ] g_1812
- [ ] g_1817
- [ ] g_1817_de
- [ ] g_1817_na
- [ ] g_1817_wo
- [ ] g_1822
[...]

PR #12593, opened 2026-05-12
[many games] replace placeholder train counts with unlimited value

philcampeau:
Not a fix for #12237 yet, but we're making some progress!

## Before clicking "Create"

### Explanation of Change

I'm in the process of setting train counts to unlimited. For the most part, it's the final rank of trains.

I was able to get around the 1858 issue. 

This is still a work in progress.


vandamm (approved):

ollybh (changes requested):
The changes to the games where you're setting a train's count to `unlimited` are fine. I'm assuming that you've checked the rules and these trains are truly unlimited.

But a hard NAK on the changes to 1858/1866/18Tokaido (and possibly others that I've missed) where you're not setting a train count and relying on this being interpreted as unlimited. That makes the train definitions in the game code unclear.

Also, what is the change the 1854 code, making the 80x a variant of the 8 train?

philcampeau:
@ollybh I think you've misunderstood the changes I made in those games.

In those games, the number of trains was being defined in `def num_trains` rather than in the TRAINS constant. If you try to include `'unlimited'` in `num_trains`, you get a console error saying `Uncaught TypeError: no implicit conversion of String into Integer`. What I've done in all those games is added `num: 'unlimited'` directly in the TRAINS constant for the relevant train. Then I call `super` in the num_trains method to handle the cases where the number of trains for a specific rank isn't specified in that method (ie: the ones that are now unlimited).

philcampeau:
As for 1854, the 8E and 8Ox trains become available at the same time, and both are unlimited, so making one a variant of the other makes sense.

ollybh:
> What I've done in all those games is added `num: 'unlimited'` directly in the TRAINS constant for the relevant train.

I'm not seeing that. For example, in 1866 this PR deletes the train count for the 10-trains. But I don't see anywhere that this is declared as unlimited.

ollybh:
> As for 1854, the 8E and 8Ox trains become available at the same time, and both are unlimited, so making one a variant of the other makes sense.

Have you tested this? Won't this change the train IDs and break existing games?

philcampeau:
> > What I've done in all those games is added `num: 'unlimited'` directly in the TRAINS constant for the relevant train.
> 
> I'm not seeing that. For example, in 1866 this PR deletes the train count for the 10-trains. But I don't see anywhere that this is declared as unlimited.

Look at line 365 of 1866's game file in my commit.

philcampeau:
> > As for 1854, the 8E and 8Ox trains become available at the same time, and both are unlimited, so making one a variant of the other makes sense.
> 
> Have you tested this? Won't this change the train IDs and break existing games?

Seems unlikely to break existing games. 

Screenshot_20260515_235940_Chrome

andrewzwicky (approved):

ollybh:
> > > What I've done in all those games is added `num: 'unlimited'` directly in the TRAINS constant for the relevant train.
> > 
> > I'm not seeing that. For example, in 1866 this PR deletes the train count for the 10-trains. But I don't see anywhere that this is declared as unlimited.
> 
> Look at line 365 of 1866's game file in my commit.

I think that has proved that having the train counts split up like this is confusing :dizzy_face: 

This should be done so that the 'unlimited' train counts are set in exactly the same place as the limited train counts.

@philcampeau Can you drop the changes to 1858/1866/18Tokaido from this PR and just keep the simple swaps from a number to `unlimited`? The more complex changes could be saved for a later PR which has a more complete fix that allows unlimited to be specified anywhere.

lib/engine/game/g_18_mag/game.rb
```diff
@@ -1562,13 +1562,13 @@ def game_trains
...
               price: 80,
-              num: 25,
+              num: 'unlimited',
             },
             {
               name: '3',
               distance: 3,
               price: 120,
-              num: 25,
+              num: 'unlimited',
```
roseundy, line 1571: While I think this is OK. I can't find anything in the rule book that talks about the number of trains in the game. There are 12 2's, 24 3's, 24 4's and 24 6's in the physical game. Since there is no rusting it does seem implied they are unlimited.

roseundy (approved):

lib/engine/game/g_18_mag/game.rb
philcampeau: image

philcampeau:
@ollybh I've removed the changes to 1866, 18Tokaido, and all versions of 1858 from this commit.

lib/engine/game/g_18_mag/game.rb
roseundy: Thanks. I even looked at this section and missed it.

ollybh (approved):
@philcampeau You've got a merge conflict to fix. Other than that, this now looks good to me.

crericha (approved):

PR #12595, opened 2026-05-13
[1824] Set 10-train to unlimited

philcampeau:
## Before clicking "Create"

### Explanation of Change

As per Issue #12237, this sets the 10-train to unlimited for 1824 and its variants. 

I did this one separately so that @perwestling could look at the changes. I've tested it and it works, but I figured I'd let Per have a look.


perwestling (approved):
LGTM

lib/engine/game/g_1824_cisleithania/game.rb
```diff
@@ -77,12 +77,9 @@ def game_trains
...
         end
 
-        def num_trains_map
-          if two_player?
-            self.class::TRAIN_COUNT_2P_CISLETHANIA.freeze
-          else
-            self.class::TRAIN_COUNT_3P_CISLETHANIA.freeze
-          end
+        def num_trains(train)
+          counts = two_player? ? TRAIN_COUNT_2P_CISLETHANIA : TRAIN_COUNT_3P_CISLETHANIA
```
crericha, line 81: Nit: This could be cached to avoid the lookup after the first call.

`@train_counts ||= ...`

lib/engine/game/g_1824/game.rb
```diff
@@ -122,20 +122,8 @@ def game_trains
...
-        # Modified base version as the number of trains vary between player count and variant
-        def init_train_handler
-          train_count_map = num_trains_map
-          trains = game_trains.flat_map do |train|
-            Array.new(train_count_map[train[:name]]) do |index|
-              Train.new(**train, index: index)
-            end
-          end
-
-          G1824::Depot.new(trains, self)
```
crericha, line 134: This code was using G1824::Depot. Is that still the case after your change?

ollybh (changes requested):
As well as the points raised by @crericha, this has got the same problem as 1858/1866/18Tokaido in #12593 with the train counts being split into two different locations, some defined in the `TRAIN_COUNT_*` constants, and some embedded in the `TRAINS` definition.

ollybh:
Closing this PR. A common fix for 1824, 1858, 1866 and 18Tokaido is needed.

Issue #12579, opened 2026-05-10
[18ESP] Excessive graph computations loading game


ollybh:
> These games are taking quite a while to load. There are lots of graph recalculations as the game is loaded, these should be skipped as much as possible. Here's a profile of loading the `18ESP_game_end_second_eight.json` fixture:
> ```
> ==================================
>   Mode: cpu(10)
>   Samples: 1301 (0.00% miss rate)
>   GC: 84 (6.46%)
> ==================================
>      TOTAL    (pct)     SAMPLES    (pct)     FRAME
>        698  (53.7%)         208  (16.0%)     Engine::Part::Path#walk
>        834  (64.1%)         118   (9.1%)     Engine::Graph#compute
>       1217  (93.5%)          86   (6.6%)     Array#each
>        133  (10.2%)          73   (5.6%)     Engine::Hex#paths
>         60   (4.6%)          60   (4.6%)     Kernel#hash
>        701  (53.9%)          52   (4.0%)     Engine::Part::Node#walk
>         44   (3.4%)          44   (3.4%)     (marking)
>         39   (3.0%)          39   (3.0%)     (sweeping)
>         22   (1.7%)          22   (1.7%)     Hash#initialize
>         30   (2.3%)          20   (1.5%)     Engine::Part::City#tokened_by?
>       1215  (93.4%)          19   (1.5%)     Class#new
>         22   (1.7%)          17   (1.3%)     Hash#delete
>         16   (1.2%)          16   (1.2%)     Integer#==
>         16   (1.2%)          16   (1.2%)     IO#write
>         33   (2.5%)          14   (1.1%)     Array#sum
>         20   (1.5%)          13   (1.0%)     Engine::Corporation#all_abilities
>         12   (0.9%)          12   (0.9%)     Engine::Part::Base#hex
>         12   (0.9%)          11   (0.8%)     Engine::Hex#invert
>         40   (3.1%)          11   (0.8%)     Array#select
>         10   (0.8%)          10   (0.8%)     Time#strftime
>       1088  (83.6%)          10   (0.8%)     Engine::Round::Base#process_action
>         10   (0.8%)          10   (0.8%)     Array#<<
>        263  (20.2%)          10   (0.8%)     Engine::Round::Base#skip_steps
>         18   (1.4%)           9   (0.7%)     Hash#to_h
>          9   (0.7%)           9   (0.7%)     String#split
>          9   (0.7%)           9   (0.7%)     String#gsub
[...]

Issue #12571, opened 2026-05-07
[18ESP] Cannot undo or restart turn. Game reverts back to current state


Rysiuvg:
**What's your Game ID?**
248680

**How to Reproduce the Bug?**

https://18xx.games/game/248680

Click "undo" or "restart turn" and the site will thrown an error and refresh and revert to the current game state. Same behaviour was observed when clicking "undo to here" in the game log. I have added a screenshot of the error message

<img width="500" height="82" alt="Image" src="https://github.com/user-attachments/assets/948d0bc8-f960-455f-bab0-8cc842e9e093" />

If I clone the game, it is possible to undo and then move forward.

**Which Browser / Platform?**
147.0.7727.138 (Official Build) (64-bit) (cohort: Stable)

chrstas:
same topic here 

https://18xx.games/game/241033

18ESP seems to lead to heavy load in the later phases

ollybh:
I'm not able to reproduce this error locally. @Rysiuvg is this still occurring, or was it a transient issue?

ollybh:
These games are taking quite a while to load. There are lots of graph recalculations as the game is loaded, these should be skipped as much as possible. Here's a profile of loading the `18ESP_game_end_second_eight.json` fixture:
```
==================================
  Mode: cpu(10)
  Samples: 1301 (0.00% miss rate)
  GC: 84 (6.46%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       698  (53.7%)         208  (16.0%)     Engine::Part::Path#walk
       834  (64.1%)         118   (9.1%)     Engine::Graph#compute
      1217  (93.5%)          86   (6.6%)     Array#each
       133  (10.2%)          73   (5.6%)     Engine::Hex#paths
        60   (4.6%)          60   (4.6%)     Kernel#hash
       701  (53.9%)          52   (4.0%)     Engine::Part::Node#walk
        44   (3.4%)          44   (3.4%)     (marking)
        39   (3.0%)          39   (3.0%)     (sweeping)
        22   (1.7%)          22   (1.7%)     Hash#initialize
        30   (2.3%)          20   (1.5%)     Engine::Part::City#tokened_by?
      1215  (93.4%)          19   (1.5%)     Class#new
        22   (1.7%)          17   (1.3%)     Hash#delete
        16   (1.2%)          16   (1.2%)     Integer#==
        16   (1.2%)          16   (1.2%)     IO#write
[...]

PR #12598, opened 2026-05-14
[18ESP] Graph processing

chrstas:
Relates to #12571 
Relates to #12579

Had the issue in my games and wanted to understand the topic a bit better based also on @ollybh's anaylsis. This PR is a **suggested fix** — I'm not an expert on the graph calculation system and I don't want to step on @bentziaxl's toes. Please treat this as a proposal rather than a definitive solution, happy to adjust or close in favour of a better approach.

## Before clicking "Create"

## Implementation Notes

- `check_for_destination_connection`: caches `@no_blocking_graph` instead of creating a new `Graph` instance on every call.
- `clear_graph_for_entity`: overridden to also call `@no_blocking_graph&.clear`, so the cache is invalidated automatically whenever tile lays or token placements occur — no explicit calls needed at the call sites.
- `opening_new_mountain_pass`: skips `@graph.clear` during loading — the graph is still computed lazily on first access, but is no longer force-cleared and recomputed on every step evaluation.
- `tracker.rb`: removed the now-redundant explicit graph clear (the base class `clear_graph_for_entity` already handles this).
- `track.rb`: removed `process_place_token` override which only called `super` + `@game.graph.clear` — both are now handled automatically.

`check_for_destination_connection` cannot simply be skipped during loading because reaching the destination goal has immediate side effects (cash and token grants) that subsequent actions may depend on — caching the graph object is used here instead of a `@game.loading` guard.

Here's a profile after the fix, loading the same `18ESP_game_end_second_eight.json` fixture:

```
==================================
  Mode: cpu(10)
  Samples: 1274 (0.00% miss rate)
  GC: 192 (15.07%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       583  (45.8%)         180  (14.1%)     Engine::Part::Path#walk
       127  (10.0%)         127  (10.0%)     (marking)
       658  (51.6%)          80   (6.3%)     Engine::Graph#compute
       577  (45.3%)          73   (5.7%)     Engine::Part::Node#walk
      1059  (83.1%)          69   (5.4%)     Array#each
        65   (5.1%)          65   (5.1%)     (sweeping)
       137  (10.8%)          61   (4.8%)     Engine::Hex#paths
        36   (2.8%)          36   (2.8%)     Hash#initialize
        21   (1.6%)          21   (1.6%)     Array#flatten
        26   (2.0%)          19   (1.5%)     Engine::Part::City#tokened_by?


```
`Engine::Graph#compute` dropped from **834 total samples (64.1%)** to **658 (51.6%)** — an ~18% reduction. `Path#walk` and `Node#walk` follow the same trend as both are called from within `compute`. `Kernel#hash` (60 samples / 4.6% before) is no longer in the top 30 at all, which reflects the reduction in `Graph.new` allocations.

The remaining `Graph#compute` cost is largely unavoidable: `check_for_destination_connection` must still run during loading because reaching a destination goal has immediate side effects (cash and token grants) that subsequent actions may depend on.

### DIFF
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -720,9 +720,12 @@ module Engine
           return false unless entity&.corporation?
           return true if entity.destination_connected?
 
-          graph = Graph.new(self, no_blocking: true)
-          graph.compute(entity)
-          graph.reachable_hexes(entity).include?(hex_by_id(entity.destination))
+          @no_blocking_graph ||= Graph.new(self, no_blocking: true)
+          @no_blocking_graph.reachable_hexes(entity).include?(hex_by_id(entity.destination))
+        end
+
+        def clear_no_blocking_graph
+          @no_blocking_graph = nil
         end
 
         def check_offboard_goal(entity, routes)
@@ -895,6 +898,7 @@ module Engine
             super
           end
           clear_graph_for_entity(corporation)
+          clear_no_blocking_graph
           corporation.goal_reached!(:destination) if check_for_destination_connection(corporation)
         end
 
@@ -1073,7 +1077,7 @@ module Engine
         def opening_new_mountain_pass(entity, p4_ability = false)
           return {} unless entity
 
-          @graph.clear
+          @graph.clear unless @loading
           openable_passes = @graph.connected_hexes(entity).keys.select do |hex|
             mountain_pass_token_hex?(hex)
           end
--- a/lib/engine/game/g_18_esp/game.rb
+++ b/lib/engine/game/g_18_esp/game.rb
@@ -724,8 +724,9 @@ module Engine
           @no_blocking_graph.reachable_hexes(entity).include?(hex_by_id(entity.destination))
         end
 
-        def clear_no_blocking_graph
-          @no_blocking_graph = nil
+        def clear_graph_for_entity(entity)
+          super
+          @no_blocking_graph&.clear
         end
 
         def check_offboard_goal(entity, routes)
@@ -898,7 +899,6 @@ module Engine
             super
           end
           clear_graph_for_entity(corporation)
-          clear_no_blocking_graph
           corporation.goal_reached!(:destination) if check_for_destination_connection(corporation)
         end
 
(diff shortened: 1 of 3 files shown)


bentziaxl:
I don't own the code :-)
and clearly i didn't put any thought into performance and graphing calculation cost when i implemented the game.
It seems that you're reducing somewhat the amount of compute called, and if it's not breaking anything i have not objection in merging this, and working on more improvements in another CR.

ollybh:
> Here's a profile after the fix, loading the same `18ESP_game_end_second_eight.json` fixture:
> […]
>      TOTAL    (pct)     SAMPLES    (pct)     FRAME
>        583  (45.8%)         180  (14.1%)     Engine::Part::Path#walk
>        127  (10.0%)         127  (10.0%)     (marking)
>        658  (51.6%)          80   (6.3%)     Engine::Graph#compute
>        577  (45.3%)          73   (5.7%)     Engine::Part::Node#walk
> […]
> `Engine::Graph#compute` dropped from **834 total samples (64.1%)** to **658 (51.6%)** — an ~18% reduction. `Path#walk` and `Node#walk` follow the same trend as both are called from within `compute`. `Kernel#hash` (60 samples / 4.6% before) is no longer in the top 30 at all, which reflects the reduction in `Graph.new` allocations.

That is an improvement, and it is worth making the suggested code changes. But it might not enough to fully fix the intermittent failures that have been reported.

> The remaining `Graph#compute` cost is largely unavoidable: `check_for_destination_connection` must still run during loading because reaching a destination goal has immediate side effects (cash and token grants) that subsequent actions may depend on.

More invasive surgery is needed to avoid the cost for the destination connect checks. The pattern used in other games is to have a step for the connection check.

If the game is being played interactively then this step would compute the graph and (if a new connection has been made) an auto action triggered to add a `destination_connection` action to the game action log. Processing this action would call `goal_reached!`.

If the game is being loaded (so `@game.loading` is true) we just assume that having this action in the action history means that the connection is valid and so don't recalculate the game graph, we just process the action.

ollybh (approved):

PR #12602, opened 2026-05-14
[18OE] implement §4.4 three-way OR share price movement

Witzman:
## Explanation of Change

Previously `share_price_change` called `super` for majors and nationals,
which returns `:right` for any positive revenue — ignoring whether the
dividend met the share-price threshold. A major at £150 paying £100
would incorrectly move RIGHT instead of staying put.

Replace the `super` call with the correct three-way rule (§4.4):
- revenue = 0          → move LEFT  (withhold)
- 0 < revenue < price  → no move    (partial dividend)
- revenue ≥ price      → move RIGHT (full dividend meets threshold)

Also remove `game.rb#change_share_price` 
this method contained the correct three-way logic but was never called — OR movement
is driven by `Step::Dividend#share_price_change` returning a direction
hash, not by an imperative game-level method. Keeping it was a
misleading dead code path.

Rule ref: §4.4 p.14

### DIFF
--- a/lib/engine/game/g_18_oe/step/dividend.rb
+++ b/lib/engine/game/g_18_oe/step/dividend.rb
@@ -28,6 +28,13 @@ module Engine
             process_dividend(Action::Dividend.new(current_entity, kind: kind))
           end
 
+          def share_price_change(entity, revenue)
+            return {} if entity.type == :minor || entity.type == :regional
+            return { share_direction: :left, share_times: 1 } if revenue.zero?
+            return {} if revenue < entity.share_price.price
+
+            { share_direction: :right, share_times: 1 }
+          end
           def dividend_types
             case current_entity.type
             when :minor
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_18_oe/step/dividend.rb
```diff
@@ -28,6 +28,14 @@ def skip!
             process_dividend(Action::Dividend.new(current_entity, kind: kind))
           end
 
+          def share_price_change(entity, revenue)
+            return {} if entity.type == :minor || entity.type == :regional
```
crericha: `%i[minor regional].include?(entity.type)`

Witzman: Done, updated to `%i[minor regional].include?(entity.type)`.

crericha (approved):

PR #12603, opened 2026-05-14
[18OE] president pool purchase above 60% at 2× price

Witzman:
## Explanation of Change

§10.2 — "The president of a major or national may purchase shares of that
RR from the Open Market (not from the treasury!), paying the bank double
the current share value per share." This is the only way to voluntarily
exceed 60% ownership.

Implementation in `G18OE::Step::BuySellParShares`:

- `president_pool_overcap_buy?`: returns true when bundle is from the
  pool, corp type is `:major` or `:national`, entity is the corp's
  president, and entity already holds ≥60%.
- `can_buy?` short-circuits before `super`, bypassing `holding_ok?`
  while still enforcing one-buy-per-turn, 2× cash requirement, and cert
  limit.
- `process_buy_shares` logs the 2× price and passes
  `exchange_price: bundle.price * 2` to `buy_shares`.

Non-presidents remain blocked above 60%; treasury share purchases also
blocked (bundle.owner is not the pool).

Rule ref: §10.2

### DIFF
--- a/lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
@@ -57,21 +57,23 @@ module Engine
               return false if bought_corporation == bundle.corporation
             end
 
-            # §10.2: president may buy own pool shares above 60% cap at 2× price.
-            # Bypass holding_ok? check but still enforce bought?, cash, and cert limit.
-            if president_pool_overcap_buy?(entity, bundle)
+            super
+          end
+
+          def can_gain?(entity, bundle, exchange: false)
+            # §10.2: president buying own pool shares is allowed above the holding cap.
+            # Bypass holding_ok?; all other checks (bought?, cash, cert limit) run via super.
+            if bundle&.owner&.share_pool? && bundle.corporation.president?(entity)
               corp = bundle.corporation
-              return !bought? &&
-                     available_cash(entity) >= modify_purchase_price(bundle) &&
-                     (!corp.counts_for_limit || @game.num_certs(entity) < @game.cert_limit(entity))
+              return !corp.counts_for_limit || exchange || @game.num_certs(entity) < @game.cert_limit(entity)
             end
 
             super
           end
 
           def modify_purchase_price(bundle)
-            # bundle.share_price is set when replaying an action that already carries the
-            # adjusted price — skip doubling to avoid charging 4×.
+            # On replay, bundle.share_price is set to the already-adjusted price from the
+            # stored action — returning bundle.price * 2 here would charge 4×.
             return bundle.price if bundle.share_price
             return bundle.price * 2 if president_pool_overcap_buy?(current_entity, bundle)
 
--- a/lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
@@ -62,23 +62,20 @@ module Engine
             if president_pool_overcap_buy?(entity, bundle)
               corp = bundle.corporation
               return !bought? &&
-                     available_cash(entity) >= bundle.price * 2 &&
+                     available_cash(entity) >= modify_purchase_price(bundle) &&
                      (!corp.counts_for_limit || @game.num_certs(entity) < @game.cert_limit(entity))
             end
 
             super
           end
 
-          def process_buy_shares(action)
-            if president_pool_overcap_buy?(action.entity, action.bundle)
-              price = action.bundle.price * 2
-              @log << "#{action.entity.name} pays #{@game.format_currency(price)} "\
-                      "(2× market) for #{action.bundle.corporation.name} pool share above 60%"
-              buy_shares(action.entity, action.bundle, exchange_price: price)
-              track_action(action, action.bundle.corporation)
-            else
-              super
-            end
+          def modify_purchase_price(bundle)
+            # bundle.share_price is set when replaying an action that already carries the
+            # adjusted price — skip doubling to avoid charging 4×.
+            return bundle.price if bundle.share_price
+            return bundle.price * 2 if president_pool_overcap_buy?(current_entity, bundle)
+
+            super
           end
 
           def can_sell?(entity, bundle)
--- a/lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
+++ b/lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
@@ -72,10 +72,7 @@ module Engine
           end
 
           def modify_purchase_price(bundle)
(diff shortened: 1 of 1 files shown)


lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -57,9 +57,30 @@ def can_buy?(entity, bundle)
...
+                     available_cash(entity) >= bundle.price * 2 &&
+                     (!corp.counts_for_limit || @game.num_certs(entity) < @game.cert_limit(entity))
+            end
+
             super
           end
 
+          def process_buy_shares(action)
+            if president_pool_overcap_buy?(action.entity, action.bundle)
+              price = action.bundle.price * 2
```
crericha: I don't think this is a good way to handle it because the price won't show correctly in the UI when the player clicks the button to buy it. Can we set the bundle price correct for the situation instead?

crericha: Probably something in or near `can_buy_share?` (or a similar name)

Witzman: Replaced the `exchange_price` approach with a `modify_purchase_price` override — the UI now reads the 2× price from there and passes it as `share_price` on the action, so `share_pool.buy_shares` charges the correct amount via `bundle.price`. One guard needed: `bundle.share_price` is already set on the action-replay path, so we skip doubling to avoid charging 4×.

Witzman: Addressed via `modify_purchase_price` as you suggested.

Witzman:
Addressed both review comments. The `exchange_price` approach has been replaced with a `modify_purchase_price` override:

- The UI button now reads the 2× price from `modify_purchase_price` and passes it as `share_price` on the action, so `share_pool.buy_shares` charges the correct amount via `bundle.price` — no custom `process_buy_shares` needed.
- One guard is required: when replaying an action, `bundle.share_price` is already set to the adjusted price by the engine's action deserialiser, so `modify_purchase_price` returns `bundle.price` as-is rather than doubling it again (which would produce 4×).

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -282,6 +300,16 @@ def bought_corporation
             @round.current_actions.find { |x| x.is_a?(Action::BuyShares) }&.bundle&.corporation
           end
 
+          def president_pool_overcap_buy?(entity, bundle)
```
crericha, line 307: This method is in the code twice and each is implemented differently.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -57,6 +57,24 @@ def can_buy?(entity, bundle)
...
+                     (!corp.counts_for_limit || @game.num_certs(entity) < @game.cert_limit(entity))
+            end
+
+            super
+          end
+
+          def modify_purchase_price(bundle)
+            # bundle.share_price is set when replaying an action that already carries the
+            # adjusted price — skip doubling to avoid charging 4×.
+            return bundle.price if bundle.share_price
```
crericha: How does this code 4x the price? In both cases bundle.price is used, not bundle.share_price. I think you can remove this line?

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
crericha: Also, given that this logic is handled at a lower level now (modify_purchase_price), you don't need all the gating. It can be something like
`bundle.corporation.president?(entity) && bundle&.owner&.share_pool?`

Witzman: Fixed in the latest commit — the  block in `can_buy?` is gone. Holding-cap bypass is now handled by a `can_gain?` override (the standard engine extension point), which uses the simpler `bundle&.owner&.share_pool? && bundle.corporation.president?(entity)` check you suggested. `president_pool_overcap_buy?` is now called only from `modify_purchase_price`.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Witzman: The guard is needed because of how action replay works. When an action is replayed, `BuyShares#h_to_args` sets `@bundle.share_price = h['share_price']` from the serialised JSON. Since `ShareBundle#price` returns `@share_price || @shares.first.price_per_share`, when `share_price` is already set (to the 2× adjusted price), `bundle.price` **already returns the doubled value**. Without the guard, `bundle.price * 2` would charge 4×. The guard is essential on the replay path.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Witzman: Done — the complex block in `can_buy?` is replaced with a `can_gain?` override. The parent `can_buy?` now handles cash (using the 2× from `modify_purchase_price`), `bought?`, and sell-before-buy automatically via `super`; `can_gain?` just bypasses `holding_ok?` and keeps the cert-limit check.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Witzman: You're right — fixed. The guard was the wrong approach. Changed to `bundle.corporation.share_price.price * 2`, reading directly from game state. This is deterministic on both fresh execution and replay (no dependence on `bundle.share_price`), and matches the pattern other games use.

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
```diff
@@ -60,6 +60,23 @@ def can_buy?(entity, bundle)
             super
           end
 
+          def can_gain?(entity, bundle, exchange: false)
+            # §10.2: president buying own pool shares is allowed above the holding cap.
+            # Bypass holding_ok?; all other checks (bought?, cash, cert limit) run via super.
+            if bundle&.owner&.share_pool? && bundle.corporation.president?(entity)
+              corp = bundle.corporation
```
crericha: Nit: corp is only used once in the logic, so the temporary variable isn't doing much.

crericha (approved):

lib/engine/game/g_18_oe/step/buy_sell_par_shares.rb
Witzman: Fixed — inlined `bundle.corporation` directly on the return line. 4abee4a

PR #12604, opened 2026-05-14
[18OE] stock round fixes + zone-based terrain discount

Witzman:
## Explanation of Change

Three commits grouped because BUG-008/009/010 modifies files introduced
by BUG-015 and BUG-022.

### BUG-015 — Sold-out UP moves in share-price order

§4.4 requires sold-out UP moves at SR end to process in descending
share-value order. The base engine uses alphabetical sort. Introduce
`G18OE::Round::Stock` with a `finish_round` override using
`.sort_by { |c| -c.share_price.price }`. Also fixes namespace:
`Engine::Round::G18OE::Stock` (not `Engine::Game::G18OE::Round::Stock`).

Rule ref: §4.4 p.14

### BUG-022 — Stock price DOWN when major issues treasury shares

§11.7/§10.1: share issuance triggers DOWN movement per share. Base
`Engine::Step::IssueShares` bypasses price adjustment. Introduce
`G18OE::Step::IssueShares` calling
`@game.sell_shares_and_change_price(bundle)` instead. Also fixes bad
require path (3 levels, not 4).

Rule ref: §11.7, §10.1

### BUG-008/009/010 — Zone-based terrain discount + E/F augmentation + rounding

**BUG-008**: Zone 20% discount not implemented. All entities with track
rights in SP/IT/SC/RU receive 20% off construction in that zone.
Implemented via `upgrade_cost` + `entity_track_rights_zone` helper.

**BUG-009**: E/F ability gives no standalone discount; it augments an
existing zone discount from 20% to 50% when terrain matches. Replaced
`TERRAIN_DISCOUNT_RATE` with `ZONE_DISCOUNT_ZONES`.

**BUG-010**: Rounding favoured bank. `(base * (1 - rate)).floor` now
rounds DOWN in favour of the RR. Example: £45 × 50% → £22 ✓.

Rule ref: §11.1.5 p.29

### Map fix — N31 Lille pre-printed revenue corrected to £20

Physical map shows £20 on the Lille yellow tile. `map.rb` had `revenue:10`.
One-line fix: `city=revenue:20`.

## Checklist


lib/engine/game/g_18_oe/entities.rb
```diff
@@ -448,6 +448,20 @@ module Entities
                              'Each OR may also spend 1 extra tile point to place a yellow tile '\
                              'in a hex with a blue terrain cost.',
               },
+              {
+                type: 'tile_discount',
+                terrain: 'water',
+                discount: 0, # augments zone discount to 50% when zone match; 0 prevents base engine applying its own discount
```
crericha: This seems like a hack that just happens to work right now. If the base code changes, who knows if this hack will still work. Is there not a method that can be overridden in the base class to accomplish this?

lib/engine/game/g_18_oe/entities.rb
```diff
@@ -448,6 +448,20 @@ module Entities
...
+                owner_type: 'corporation',
+              },
+              {
+                type: 'tile_lay',
+                tiles: [],
+                when: 'track',
+                count_per_or: 1,
+                owner_type: 'corporation',
+                consume_tile_lay: false,
```
crericha: This ability doesn't really match the 1 extra track point. The track point could be spent on a M city, which is still a yellow tile. And the ability doesn't even restrict to yellow tiles. I think this ability needs to be custom code within the game. Or create a new track_point ability and augment it with a terrain field.

lib/engine/game/g_18_oe/entities.rb
```diff
@@ -469,6 +483,20 @@ module Entities
                              'Each OR may also spend 1 extra tile point to place a yellow tile '\
                              'in a hex with a green terrain cost.',
               },
+              {
+                type: 'tile_discount',
+                terrain: 'mountain',
+                discount: 0, # augments zone discount to 50% when zone match; 0 prevents base engine applying its own discount
```
crericha: Same feedback as blue coast bridge private

lib/engine/game/g_18_oe/game.rb
```diff
@@ -224,6 +224,10 @@ class Game < Game::Base
         ASTERISKED_ZONES = %w[UK PHS FR].freeze
         ASTERISKED_ZONES_CAP = 4
 
+        ZONE_DISCOUNT_ZONES         = %w[SP IT SC RU].freeze
+        ZONE_DISCOUNT_RATE          = Rational(1, 5).freeze # 20% §11.1.5
+        ZONE_TERRAIN_DISCOUNT_RATE  = Rational(1, 2).freeze # 50% §11.1.5
```
crericha: I haven't seen the rational class used before. Does 0.2 and 0.5 not work?

lib/engine/game/g_18_oe/game.rb
```diff
@@ -887,9 +915,30 @@ def operating_round(round_num)
...
+        end
+
+        def terrain_discount_ability(entity, tile)
+          resolved = resolve_corporation(entity)
+          return nil unless resolved
+
+          resolved.all_abilities.find { |a| a.type == :tile_discount && a.terrain && a.discounts_tile?(tile) }
+        end
+
+        def resolve_corporation(entity)
```
crericha: Nit: Can we make the function now clearer to its intention? Right now if someone reads the method call, there is no context to what this method does. If the name is clearer, then it will make reading the code easier.

lib/engine/game/g_18_oe/entities.rb
Witzman: upgrade_cost in game/base.rb should be that method — we're already overriding it.
Replaced the sentinel with an `EF_TERRAIN_AUGMENT = { 'E' => :water, 'F' => :mountain }` constant. `terrain_augmented_by?` now does a direct corp-ID lookup

lib/engine/game/g_18_oe/entities.rb
Witzman: Agreed — it doesn't model the rule correctly, and the SpecialTrack step isn't wired in the OR so it was dead code. Removed. The extra-tile-point mechanic needs a proper Track step redesign; logged as BUG-034 to address separately.

lib/engine/game/g_18_oe/entities.rb
Witzman: Same fix applied to F — both the `tile_discount:0` and dead `tile_lay` removed.

lib/engine/game/g_18_oe/game.rb
Witzman: Switched to plain floats.

lib/engine/game/g_18_oe/game.rb
Witzman: Renamed to `owning_corporation` — it unwraps an entity to the corporation holding the relevant track right or ability

crericha (approved):

PR #12605, opened 2026-05-14
[18OE] level-8 train gate + game-end timing + remainder cash

Witzman:
## Explanation of Change

Four bugs grouped: BUG-027 sets the initial `GAME_END_CHECK` override;
BUG-024/025 modifies it further — coupled on the same constant, all
relate to game-end rules and level-8 trains.

### BUG-027 — Remove bankrupt: :immediate game-end trigger

18OE has no bankruptcy game-end (§11.6.4: force-buy → president
contribution or national conversion; §11.6.5: minor insolvency →
suspend + Open Market). Override `GAME_END_CHECK` to drop
`bankrupt: :immediate`, retaining only `bank: :full_or`.

Rule ref: §11.6.4, §11.6.5

### BUG-018 — Gate level-8 trains behind 4th level-7 purchase

§11.6: level-8 trains available only after the 4th level-7 purchase
(not after all 17). `G18OE::Step::BuyTrain#buyable_trains` injects the
first 8+8 from `depot.upcoming` once `game.level8_train_available?`
returns true (counts sold level-7s via depot diff).

Rule ref: §11.6

### BUG-024/025 — Bank-break timing + level-8 game-end + remainder cash

§13: bank-break → `:current_or` timing (not `:full_or`). First level-8
purchase → `:one_more_full_or_set` end, triggers injection of
`REMAINDER_CASH` (20×£5,000 = £100,000) into bank. Bank-break wins if
both fire (priority ordering).

Also refactors `level8_train_available?` from step to `game.rb` per
coding guidelines.

### DIFF
--- a/lib/engine/bank.rb
+++ b/lib/engine/bank.rb
@@ -30,6 +30,10 @@ module Engine
       break!
     end
 
+    def receive(cash)
+      @cash += cash
+    end
+
     def break!
       @log << '-- The bank has broken --' unless @broken
       @broken = true
--- a/lib/engine/bank.rb
+++ b/lib/engine/bank.rb
@@ -30,7 +30,7 @@ module Engine
       break!
     end
 
-    def receive(cash)
+    def add_cash(cash)
       @cash += cash
     end
 
--- a/lib/engine/game/g_18_oe/game.rb
+++ b/lib/engine/game/g_18_oe/game.rb
@@ -37,6 +37,8 @@ module Engine
         MUST_SELL_IN_BLOCKS = false
         HOME_TOKEN_TIMING = :float
         TILE_UPGRADES_MUST_USE_MAX_EXITS = [:cities].freeze
+        # §11.6.4/11.6.5: no player-bankruptcy game-end; force-buy converts major or suspends minor
+        GAME_END_CHECK = { bank: :full_or }.freeze
 
         STOCKMARKET_COLORS = {
           par: :blue,
--- a/lib/engine/game/g_18_oe/game.rb
+++ b/lib/engine/game/g_18_oe/game.rb
@@ -37,8 +37,10 @@ module Engine
         MUST_SELL_IN_BLOCKS = false
         HOME_TOKEN_TIMING = :float
         TILE_UPGRADES_MUST_USE_MAX_EXITS = [:cities].freeze
-        # §11.6.4/11.6.5: no player-bankruptcy game-end; force-buy converts major or suspends minor
-        GAME_END_CHECK = { bank: :full_or }.freeze
+        # §13: bank-break ends at current OR; level-8 purchase ends at one_more_full_or_set
+        GAME_END_CHECK = { bank: :current_or, final_phase: :one_more_full_or_set }.freeze
+        # Physical game includes 20×£5,000 notes set aside at setup; injected when first level-8 bought
+        REMAINDER_CASH = 100_000
 
         STOCKMARKET_COLORS = {
           par: :blue,
@@ -195,6 +197,7 @@ module Engine
             num: 17,
           },
           # Level 8 — gray double-sided (8+8 / 5D); permanent
+          # NOTE: purchase of the FIRST level-8 triggers game end (§13)
           {
             name: '8+8',
             distance: [{ 'nodes' => ['town'], 'pay' => 8, 'visit' => 99 },
@@ -207,6 +210,7 @@ module Engine
               price: 1000,
             }],
             num: 11,
+            events: [{ 'type' => 'level8_train_purchased' }],
           },
         ].freeze
 
@@ -775,6 +779,20 @@ module Engine
             @log << "#{spender.name} receives a #{label} discount of #{format_currency(discount)}"
           end
           cost
+        end
+
+        def game_end_check_final_phase?
+          @level8_train_purchased
+        end
+
+        def event_level8_train_purchased!
+          return if @level8_train_purchased
+
+          @level8_train_purchased = true
+          remainder = self.class::REMAINDER_CASH
+          @bank.instance_variable_set(:@cash, @bank.cash + remainder)
+          @log << "-- Event: First level 8 train purchased --"
+          @log << "#{format_currency(remainder)} remainder cash added to bank (§13)"
         end
 
         # UP movement at end of SR: only for majors and nationals that are fully player-held
--- a/lib/engine/game/g_18_oe/game.rb
+++ b/lib/engine/game/g_18_oe/game.rb
@@ -210,6 +210,7 @@ module Engine
               price: 1000,
             }],
             num: 11,
+            available_on: '7+7',
             events: [{ 'type' => 'level8_train_purchased' }],
           },
         ].freeze
@@ -785,10 +786,6 @@ module Engine
           level7_remaining = depot.upcoming.count { |t| t.name == '7+7' }
           level7_total = depot.trains.count { |t| %w[7+7 4D].include?(t.name) }
           level7_total - level7_remaining >= 4
-        end
-
-        def game_end_check_final_phase?
-          @level8_train_purchased
         end
 
         def event_level8_train_purchased!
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_18_oe/step/buy_train.rb
```diff
@@ -13,6 +13,17 @@ def must_buy_train?(entity)
 
           def buyable_trains(entity)
             trains = super
+
+            # Level 8 trains only available after 4th level-7 purchase (§11.6)
+            if @game.level8_train_available?
+              unless trains.any? { |t| t.name == '8+8' }
+                lvl8 = @game.depot.upcoming.find { |t| t.name == '8+8' }
+                trains += [lvl8] if lvl8
```
crericha: I think you can configure the train type with the  `available_on` field and set it to level-7 purchases. Then reject the 8+8s if not level8_train_available?. That way you don't need this whole if and only need the else logic.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -37,6 +37,10 @@ class Game < Game::Base
         MUST_SELL_IN_BLOCKS = false
         HOME_TOKEN_TIMING = :float
         TILE_UPGRADES_MUST_USE_MAX_EXITS = [:cities].freeze
+        # §13: bank-break ends at current OR; level-8 purchase ends at one_more_full_or_set
+        GAME_END_CHECK = { bank: :current_or, final_phase: :one_more_full_or_set }.freeze
+        # Physical game includes 20×£5,000 notes set aside at setup; injected when first level-8 bought
+        REMAINDER_CASH = 100_000
```
crericha, line 43: Is this needed for the digital implementation? If a level-8 train is bought, isn't the game end condition already set?

lib/engine/game/g_18_oe/game.rb
```diff
@@ -746,6 +752,25 @@ def train_obligation_active?
...
         end
 
+        def level8_train_available?
+          level7_remaining = depot.upcoming.count { |t| t.name == '7+7' }
+          level7_total = depot.trains.count { |t| %w[7+7 4D].include?(t.name) }
+          level7_total - level7_remaining >= 4
+        end
+
+        def game_end_check_final_phase?
+          @level8_train_purchased
```
crericha: Are you sure this needs to be custom? I thought it was supported out of the box.

lib/engine/game/g_18_oe/step/buy_train.rb
Witzman: Good call. Added `available_on: '7+7'` to the 8+8 train so the depot surfaces it automatically once the 7+7 phase is active. `buyable_trains` now just needs the single `reject` guard for the 4th-purchase condition.

lib/engine/game/g_18_oe/game.rb
Witzman: Yes, it is needed. Once the first 8+8 is bought, `final_phase: :one_more_full_or_set` schedules game end after a full OR set — but `bank: :current_or` is still active and can fire sooner if the bank breaks during those final ORs. The £100,000 remainder (20 × £5,000 notes set aside in the physical game) prevents that premature break. Without it the bank could run dry mid-endgame and cut the game short.

lib/engine/game/g_18_oe/game.rb
Witzman: You're right — the base `game_end_check_final_phase?` already returns true when the current phase is the last phase, and `on: '8+8'` is the last phase, so buying the first 8+8 naturally satisfies it. The custom override was redundant. Removed.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -205,6 +210,8 @@ class Game < Game::Base
               price: 1000,
             }],
             num: 11,
+            available_on: '7+7',
+            events: [{ 'type' => 'level8_train_purchased' }],
```
crericha: Nit: Isn't the event `add remainder cash`? When a player is looking at the info tab, an event called level8_train_purchased, doesn't tell them much.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -746,6 +753,21 @@ def train_obligation_active?
           phase.status.include?('train_obligation')
         end
 
+        def level8_train_available?
```
crericha, line 792: This function isn't very efficient. I suggest checking phase and using index. The first train of each level is always index 0 and it increments from there. So index 4 would be the 5th train.

```
return false if @game.phase.to_i < 7
return true if @game.phase.to_i == 8

next_train = depot.upcoming.first
next_train.index >= 4 || next_train.name != '7+7'
```

lib/engine/game/g_18_oe/game.rb
```diff
@@ -746,6 +753,21 @@ def train_obligation_active?
...
+          level7_total = depot.trains.count { |t| %w[7+7 4D].include?(t.name) }
+          level7_total - level7_remaining >= 4
+        end
+
+        def event_level8_train_purchased!
+          return if @level8_train_purchased
+
+          @level8_train_purchased = true
+          remainder = self.class::REMAINDER_CASH
+          @bank.instance_variable_set(:@cash, @bank.cash + remainder)
```
crericha: This is a hack. Add a method to the bank class to add cash and then use that.

lib/engine/game/g_18_oe/game.rb
crericha: Also, you'll want to add to `EVENT_TEXT` like this
https://github.com/tobymao/18xx/blob/e5d80f182ad4c5b9127084b5ffd5e3ba9e46fc7d/lib/engine/game/g_1846/game.rb#L201

Witzman: Renamed to `remainder_cash_added` — describes what players see rather than the trigger. Also added an `EVENTS_TEXT` entry (see next comment).

Witzman: Done — added `EVENTS_TEXT = Base::EVENTS_TEXT.merge('remainder_cash_added' => ['Remainder Cash Added', '£100,000 remainder cash injected into bank; game ends after one more full OR set (§13)'])`.

lib/engine/game/g_18_oe/game.rb
Witzman: Good call. Adopted your approach with adaptations for game.rb (`phase.name.to_i` instead of `@game.phase.to_i`, `@depot.upcoming.first` instead of `depot.upcoming.first`). Two early returns cover the clear cases; the index check handles the 4th-purchase boundary in O(1).

lib/engine/game/g_18_oe/game.rb
Witzman: Added `Bank#receive(cash)` to `lib/engine/bank.rb` — a minimal public method that increments `@cash` directly, giving games a clean way to inject externally-set-aside cash without abusing `instance_variable_set`.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -37,12 +37,23 @@ class Game < Game::Base
...
 
         STOCKMARKET_COLORS = {
           par: :blue,
           convert_range: :red,
         }.freeze
 
+        EVENTS_TEXT = Base::EVENTS_TEXT.merge(
+          'remainder_cash_added' => [
+            'Remainder Cash Added',
+            '£100,000 remainder cash injected into bank; game ends after one more full OR set (§13)',
```
crericha: Is 13 the rules section? Do you want that in the description? I don't think we do that for any other games.

crericha: Should it be `game ends after one more full OR set or bank breaking`?

lib/engine/bank.rb
```diff
@@ -30,6 +30,10 @@ def check_cash(amount, borrow_from: nil)
       break!
     end
 
+    def receive(cash)
```
crericha: I'm wondering if we should make this name more specific to not confuse people in the future. `add_cash`, `inject_cash`, ...

lib/engine/game/g_18_oe/game.rb
Witzman: done

lib/engine/bank.rb
Witzman: changed to add_cash

crericha (approved):

PR #12620, opened 2026-05-17
[18ESP] Reduce excessive graph computations - prework

chrstas:
Relates to #12579

**2-Step Approach to execute surgery**

This is the first of two PRs addressing #12579. PR 1 lays down a test safety net (unit tests for `Corporation#goal_reached!(:destination)` and fixture-replay tests) and adds a new `CheckDestinationConnection` step as dead code — the class is not yet wired into the operating round, so this PR ships zero behavior change. 

Subsequent PR 2 will wire the step into the step list, remove the five existing direct calls to `check_for_destination_connection`, add the `@game.loading` guard that delivers the actual performance win, and include profiling and replay verification as acceptance criteria. 

Splitting the work this way isolates the dead-code review from the behavior-change review and gives the safety net a chance to be merged before the refactor that depends on it.

## Before clicking "Create"

## Implementation Notes

**Step `check_destination_connection`**

`CheckDestinationConnection` is a single-entity auto-action step that fires after each tile lay or token placement to detect whether the active corporation has connected to its destination hex. Unlike the 1870/System18 reference implementations which store the connection result in round state for later evaluation, 18ESP awards the goal immediately via `corporation.goal_reached!(:destination)` since the connection itself is the trigger. `pass!` is overridden as a no-op to prevent `force_next_entity!` from skipping the step before its auto-action has fired.

**Tests added**

`corporation_spec.rb` exercises `Corporation#goal_reached!(:destination) `in isolation — verifying that the bonus is paid exactly once regardless of how many times the method is called, that a second call is a true no-op (no cash movement, no log entry, no token unblock), and that offboard and harbor goal flags are unaffected. 

`game_spec.rb` establishes a frozen snapshot of the full fixture replay: it pins `goals_reached_counter` and `destination_connected?` for all major corporations at game end, asserts that exactly 8 destination goal log entries are produced, and confirms that all destination hex icons are removed — giving a regression baseline for every subsequent refactor step. 

`check_destination_connection_spec.rb` exercises the new step in isolation against a real game instance, covering entity type guards, empty vs. populated corporations list in the auto-action,` goal_reached!` call delegation, `@passed` state after processing, and the `pass!` no-op contract.

### DIFF
--- a/lib/engine/game/g_18_esp/step/check_destination_connection.rb
+++ b/lib/engine/game/g_18_esp/step/check_destination_connection.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require_relative '../../../step/base'
+
+module Engine
+  module Game
+    module G18ESP
+      module Step
+        class CheckDestinationConnection < Engine::Step::Base
+          ACTIONS = %w[destination_connection].freeze
+
+          def actions(_entity)
+            ACTIONS
+          end
+
+          def auto_actions(entity)
+            return [] unless entity&.corporation?
+
+            connected = @game.check_for_destination_connection(entity) ? [entity] : []
+            [Engine::Action::DestinationConnection.new(entity, corporations: connected)]
+          end
+
+          def description
+            'Check destination connection'
+          end
+
+          # Overridden to be a no-op: prevents force_next_entity! from skipping
+          # this step before it has had a chance to fire its auto_action.
+          def pass!; end
+
+          def process_destination_connection(action)
+            action.corporations.each { |c| c.goal_reached!(:destination) }
+            @passed = true
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_18_esp/step/check_destination_connection.rb
+++ b/lib/engine/game/g_18_esp/step/check_destination_connection.rb
@@ -24,8 +24,11 @@ module Engine
             'Check destination connection'
           end
 
-          # Old fixtures have no destination_connection sub-action stored; the
-          # direct calls in tracker/home_token still handle loading correctly.
+          # During loading, auto_actions returns [] (performance win) — without this
+          # guard the round would hang because no destination_connection action is
+          # emitted at runtime. New saves carry the action in the log and run through
+          # process_destination_connection; pre-PR-2 saves are state-corrected by
+          # Game#backfill_destination_connections! after replay finishes.
           def blocking?
             return false if @game.loading
 
--- a/lib/engine/game/g_18_esp/step/check_destination_connection.rb
+++ b/lib/engine/game/g_18_esp/step/check_destination_connection.rb
@@ -24,6 +24,14 @@ module Engine
             'Check destination connection'
           end
 
+          # Old fixtures have no destination_connection sub-action stored; the
+          # direct calls in tracker/home_token still handle loading correctly.
+          def blocking?
+            return false if @game.loading
+
+            super
+          end
+
           # Overridden to be a no-op: prevents force_next_entity! from skipping
           # this step before it has had a chance to fire its auto_action.
           def pass!; end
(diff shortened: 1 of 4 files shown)


lib/engine/game/g_18_esp/step/check_destination_connection.rb
```diff
@@ -0,0 +1,50 @@
...
+
+          def actions(_entity)
+            ACTIONS
+          end
+
+          def auto_actions(entity)
+            return [] unless entity&.corporation?
+
+            connected = @game.check_for_destination_connection(entity) ? [entity] : []
+            [Engine::Action::DestinationConnection.new(entity, corporations: connected)]
```
ollybh: It would be better to not insert the action if there is not a new destination connection.
```suggestion
            return [] unless @game.check_for_destination_connection(entity)

            [Engine::Action::DestinationConnection.new(entity, corporations: [entity])]
```

lib/engine/game/g_18_esp/step/check_destination_connection.rb
```diff
@@ -0,0 +1,50 @@
...
+          end
+
+          # Overridden to be a no-op: prevents force_next_entity! from skipping
+          # this step before it has had a chance to fire its auto_action.
+          def pass!; end
+
+          def process_destination_connection(action)
+            action.corporations.each { |c| c.goal_reached!(:destination) }
+            @passed = true
+          end
```
ollybh, line 36: If I've understood the existing code (and I'm far from certain that I have), it is only checking if the active company is connected to its destination. So you probably don't need an iterator here, you could just pass a single one in the action.

chrstas:
setting to draft: discovered that l need to make 2-3 adaptions before finally submitting

spec/lib/engine/game/g_18_esp/game_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+# Values captured with Engine::Game.load(data, strict: false) at commit bc6657312.
+# AVT and TBF reach their destination not via goal_reached!(:destination) but through
+# the full-cap mechanism (game.rb ~line 589), hence goals=0.
+
+describe Engine::Game::G18ESP::Game do
+  describe '18ESP_game_end_second_eight' do
+    subject(:game) do
+      data = JSON.parse(File.read("#{FIXTURES_DIR}/18ESP/18ESP_game_end_second_eight.json"))
+      Engine::Game.load(data, strict: false).tap(&:maybe_raise!)
+    end
```
ollybh: You can take advantage of existing code to load fixture files.
```suggestion
    let(:game) { fixture_at_action(1366) }
```

lib/engine/game/g_18_esp/step/check_destination_connection.rb
ollybh: And you'd want to check for `!entity.destination_connected?` too.

ollybh (commented):
This isn't a game where you need to check for multiple companies becoming connected to their destinations at the same time, just the active one, so processing the action could expect to be passed a single entity.

As it's written, this will be adding an action each time the `auto_actions` step is called for a corporation, most times passing an empty array for the list of connected corporations. This would be better done as only adding the action when somebody is newly connected to their destination.

lib/engine/game/g_18_esp/step/check_destination_connection.rb
chrstas: Both addressed via a shared `new_destination_connection?(entity)` predicate that checks `entity&.corporation?`, `!entity.destination_connected?`, and `check_for_destination_connection` — `auto_actions` only emits when the predicate is true and always with `corporations: [entity]`.

lib/engine/game/g_18_esp/step/check_destination_connection.rb
chrstas: yes your are right... Switched to single-entity — auto_actions always emits with `corporations: [entity]` and `process_destination_connection` uses `.first` instead of iterating, with a dedicated spec documenting the single-entity invariant.

chrstas:
Thanks again for the review. Your points make absolutely sense... 

All three review threads addressed in the latest push: the step now emits a `DestinationConnection` only when a corporation is newly destination-connected, processes a single entity via `.first`, and the game spec uses `fixture_at_action` as suggested.

ollybh (approved):

PR #12624, opened 2026-05-17
[1862 USA & Canada] PR 02: corporation group unlock

neutronc:
Stacked on #12622 (PR 01 — data skeleton). Adds corporation group unlock logic.

## Before clicking "Create"

- [x] Branch is derived from `1862_usa_canada_p01_data`
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

### Explanation of Change

Implements `can_par?` with the three progressive group unlock conditions:

| Group | Corps | Can par when… |
|-------|-------|--------------|
| 1 | NYH, NYC, CP | All 8 private companies have been sold (player-owned or closed) |
| 2 | CPR, UP, ATS, SP | All Group 1 IPO shares are fully sold out (`num_ipo_shares == 0`) |
| 3 | NP, CN, TP, ORN, WP, GMO | All Group 2 corporations have floated |

Adds helpers `corp_group`, `all_privates_sold?`, `group_available?` in `game.rb`.
Spec covers all three lock/unlock gates.

### Screenshots

N/A — logic change, no UI change.

### Any Assumptions / Hacks

"All papers sold" for Groups 1→2 means the IPO pool is empty for every Group 1 corp, not merely that they have floated. A corp floats on a threshold (e.g. 60%) but may still have unsold IPO shares; group unlock waits for the full sellout.

PR #12622, opened 2026-05-17
[1862 USA & Canada] PR 01: data skeleton — map, entities, market, trains, logos

neutronc:
## Before clicking "Create"

- [x] Branch is derived from the latest `master`
- [ ] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

### Explanation of Change

This is PR 01 of ~13 planned PRs implementing **1862 USA & Canada** (Helmut Ohley, *The First Transcontinental Railroad*), a 3–7 player 18xx game set on the North American continent.

This first PR contains only the static data — no custom game logic:

- **`meta.rb`** — game metadata; `GAME_TITLE = '1862UsaCanada'`, `GAME_DISPLAY_TITLE` for the long name
- **`map.rb`** — full hex map with pre-printed tiles (Canada + USA, coast to coast)
- **`entities.rb`** — 8 private companies and 13 corporations across 3 unlock groups
- **`game.rb`** — constants only: `MARKET`, `TRAINS` (2–8 with E-variants as `variants:`), `PHASES`, `CERT_LIMIT`, `STARTING_CASH`, plus round definitions using engine defaults
- **`public/logos/1862_usa_canada/`** — placeholder SVG logos for all 13 corporations
- **`spec/`** — minimal spec covering initialisation, corp count, private count, phase, capitalisation, turn order

Subsequent PRs will layer in custom mechanics one at a time (corporation group unlock, SLC/Golden Spike bonus, bond system, E-train rusting, etc.).

### Corporation groups

| Group | Corps | Director share | Unlocks when |
|-------|-------|---------------|--------------|
| 1 | NYH, NYC, CP | 30% + 7×10% | game start |
| 2 | CPR, UP, ATS, SP | 20% + 8×10% | all Group 1 floated |
| 3 | NP, CN, TP, ORN, WP, GMO | mixed | all Group 2 floated |

### Screenshots
Screenshot 2026-05-15 222603

### Any Assumptions / Hacks

- Logos are placeholder SVGs (corp abbreviation + colour); final artwork TBD.
- Several constants marked `# FIXME: unconfirmed from rulebook` (Golden Spike bonus amount, TP El Paso bonus, operating rounds per phase) — to be resolved against the physical rulebook in a later PR.
[...]

PR #12625, opened 2026-05-17
[1862 USA & Canada] PR 03: fix graph cache cleared after home token placement

neutronc:
Stacked on #12624 (PR 02 — group unlock).

## Before clicking "Create"

- [x] Branch is derived from `1862_usa_canada_p02_group_unlock`

### Explanation of Change

`HOME_TOKEN_TIMING = :operate` means each corporation's home token is placed automatically at the start of its first OR turn via `place_home_token`. That method calls `city.place_token` directly — it bypasses the step layer — so `clear_graph_for_entity` is never triggered. As a result the graph cache holds stale connectivity data (no token → no connected hexes) and the corporation cannot lay track on that same turn.

Fix: override `place_home_token` in `game.rb` to call `super` then `clear_graph_for_entity(corporation)`, flushing the cache immediately after the token lands.

### Screenshots

N/A — logic fix, no visible UI change.

### Any Assumptions / Hacks

None. Pattern taken directly from other titles that use `HOME_TOKEN_TIMING = :operate`.

### DIFF
--- a/lib/engine/game/g_1862_usa_canada.rb
+++ b/lib/engine/game/g_1862_usa_canada.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1862UsaCanada
+    end
+  end
+end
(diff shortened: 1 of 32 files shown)


neutronc:

PR #12626, opened 2026-05-17
[1862 USA & Canada] PR 04: fix E-train distance hash key style

neutronc:
Stacked on #12625 (PR 03 — home token fix).

## Before clicking "Create"

- [x] Branch is derived from `1862_usa_canada_p03_home_token`

### Explanation of Change

E-trains (2E–7E) are defined as `variants:` on each base train. The distance array for each variant was using Ruby symbol keys (`:nodes`, `:pay`, `:visit`) but the engine reads these with string keys (`n['nodes']`, `n['pay']`, `n['visit']` — confirmed in `train.rb:145` and throughout `route.rb`). With symbol keys the engine silently ignored the distance, breaking route validation for all E-train variants.

Fix: changed all six E-train variant distance hashes to string keys.

Also adds spec examples that:
- verify every base train 2–7 has exactly one E-variant
- assert distance uses string keys and `visit: 999`
- verify rust triggers: 2/2E→4, 3/3E→6, 4/4E→8, 5+ never rust

### Screenshots

N/A.

### Any Assumptions / Hacks

None. String keys are the engine convention for all distance hash arrays (see g_1861, g_18_ny, etc.).

### DIFF
--- a/lib/engine/game/g_1862_usa_canada.rb
+++ b/lib/engine/game/g_1862_usa_canada.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1862UsaCanada
+    end
+  end
+end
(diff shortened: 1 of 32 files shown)


neutronc:

PR #12627, opened 2026-05-17
[1862 USA & Canada] PR 05: phase-gated tile-lay budget

neutronc:
Stacked on #12626 (PR 04 — E-train fix).

## Before clicking "Create"

- [x] Branch is derived from `1862_usa_canada_p04_etrains`
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

### Explanation of Change

Implements the phase-gated tile-lay budget:

| Phase | Trigger | Tile lays allowed |
|-------|---------|------------------|
| 2 | game start | 1 yellow tile only |
| 3–8 | first 3/3E purchase | 2 yellow tiles **OR** 1 upgrade (no mixing) |

"No mixing" means: if the first tile laid is an upgrade, the second slot is blocked (`:not_if_upgraded`). A yellow + upgrade combination is not permitted.

Implementation:
- `YELLOW_TILE_LAY` / `TWO_TILE_LAYS` constants define the two budgets
- `status: ['two_tile_lays']` added to phases 3–8 in `PHASES`
- `tile_lays` override keys on `@phase.status`
- `STATUS_TEXT` entry added so the phase legend in the UI shows the rule

### Screenshots

N/A.

### Any Assumptions / Hacks

None. Pattern is directly analogous to 18FR, corrected to use `lay: :not_if_upgraded` (not `lay: true`) on the second slot so yellow+upgrade mixing is blocked.

PR #12628, opened 2026-05-17
[1862 USA & Canada] PR 06: GHU token discount + NYH par price lock

neutronc:
Stacked on #12627 (PR 05 — tile-lay budget).

## Before clicking "Create"

- [x] Branch is derived from `1862_usa_canada_p05_tile_lays`

### Explanation of Change

**GHU (Bahnhoflizenz, $75 private)**
- The player who owns GHU gives their corporation an $80 discount on station token placement (minimum $0).
- Fix: changed `owner_type: 'corporation'` → `owner_type: 'player'` (corporations cannot buy privates in this game, so the ability was unreachable).
- New `step/token.rb` overrides `process_place_token` to auto-apply the discount when the operating corporation's director holds GHU.

**NHSC / NYH par price lock**
- NHSC gives its buyer NYH's 30% director certificate immediately. NYH must be parred at exactly $100 (rulebook restriction).
- New `step/buy_sell_par_shares.rb` overrides `get_par_prices` to return only the $100 par price cell when the corporation being parred is NYH.

Both steps wired into `stock_round` and `operating_round` in `game.rb`.

### Screenshots

N/A — logic change.

### Any Assumptions / Hacks

- GHU discount is auto-applied (always used when available). The discount always reduces cost, so there is no reason a director would decline it.
- The NYH $100 par restriction applies unconditionally — any player who holds NYH_0 must par at $100, which is correct since NHSC always delivers NYH_0 to its buyer.

### DIFF
--- a/lib/engine/game/g_1862_usa_canada.rb
+++ b/lib/engine/game/g_1862_usa_canada.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1862UsaCanada
+    end
+  end
+end
--- a/lib/engine/game/g_1862_usa_canada/entities.rb
+++ b/lib/engine/game/g_1862_usa_canada/entities.rb
@@ -52,7 +52,7 @@ module Engine
                 type: 'tile_discount',
                 discount: 80,
                 terrain: 'station',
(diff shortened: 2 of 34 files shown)


neutronc:

PR #12631, opened 2026-05-17
[18OE] fix partition Borders

Witzman:
## Summary

This PR addresses a bug in `Partition#initialize` introduced alongside the `length:` parameter (PR #12578), and completes the 18OE intra-hex province border data.

---

## Engine fix — `lib/engine/part/partition.rb`

**Problem**

`Partition#initialize` always applies `[a, b].minmax` before reading the vertex numbers:

```ruby
a, b = [a, b].minmax   # always re-orders
@a = a[0].to_i
```

When `length:` is set this silently breaks the intended draw anchor. For example:

```
partition=a:4,b:1,type:province,length:0.5
```

After `minmax`, `a` becomes `"1"` (the lower vertex), so `@a = 1`. The renderer draws a half-line from **corner 1 → center** instead of the intended **corner 4 → center**.

`VERTICES[-1] = [0, 0]` was added to `partitions.rb` in PR #12578 precisely to support a centre sentinel, and the `length:` truncation was designed so that `a:4,b:1,length:0.5` draws the line from vertex 4 toward vertex 1, stopping at the midpoint (= hex centre for opposite corners). The `minmax` reorder defeats this design.

**Fix**

```ruby
# Skip minmax when length is set so the DSL anchor is honoured exactly.
a, b = [a, b].minmax unless length
```

`inner` / `outer` are still computed from `[@a, @b].minmax`, preserving routing semantics for all existing games. **No existing game uses the `length` parameter**, so there is zero behavioural change for any title other than 18OE.

---

## Map data — `lib/engine/game/g_18_oe/map.rb`

- Reformats multi-line province partition strings to single lines (style only, no value change)
- Adds intra-hex partition lines for the Russian southern border hexes missing them: U78, U80, S76, R75 (province line crossing the hex face, same pattern as neighbouring hexes)
- Removes erroneous intra-hex partitions from P63 & U50 — thess hexes only has edge borders, no province line crosses its face

---

### DIFF
--- a/lib/engine/part/partition.rb
+++ b/lib/engine/part/partition.rb
@@ -17,7 +17,9 @@ module Engine
         # a and b are vertices of the hex. 0 represents the bottom one and then you go clockwise
         # The sign tells if the partition should be drawn a little bit before or after the vertex,
         # but doesn't have any impact on the game.
-        a, b = [a, b].minmax
+        # When length is set, 'a' is the explicit draw anchor — minmax is skipped so the
+        # DSL author controls which end the line starts from.
+        a, b = [a, b].minmax unless length
         @a = a[0].to_i
         @a_sign = SIGN[a[1]]
         @b = b[0].to_i
@@ -29,8 +31,9 @@ module Engine
         @length = length&.to_f
         @blockers = []
 
-        @inner = (restrict == 'outer' ? [] : (@a..(@b - 1)).to_a)
-        @outer = (restrict == 'inner' ? [] : (0..5).to_a - (@a..(@b - 1)).to_a)
+        a_lo, b_hi = [@a, @b].minmax
+        @inner = (restrict == 'outer' ? [] : (a_lo..(b_hi - 1)).to_a)
+        @outer = (restrict == 'inner' ? [] : (0..5).to_a - (a_lo..(b_hi - 1)).to_a)
       end
 
       def add_blocker!(private_company)
--- a/lib/engine/part/partition.rb
+++ b/lib/engine/part/partition.rb
@@ -17,7 +17,6 @@ module Engine
         # a and b are vertices of the hex. 0 represents the bottom one and then you go clockwise
         # The sign tells if the partition should be drawn a little bit before or after the vertex,
         # but doesn't have any impact on the game.
-        a, b = [a, b].minmax unless length
         @a = a[0].to_i
         @a_sign = SIGN[a[1]]
         @b = b[0].to_i
--- a/lib/engine/part/partition.rb
+++ b/lib/engine/part/partition.rb
@@ -17,8 +17,6 @@ module Engine
         # a and b are vertices of the hex. 0 represents the bottom one and then you go clockwise
         # The sign tells if the partition should be drawn a little bit before or after the vertex,
         # but doesn't have any impact on the game.
-        # When length is set, 'a' is the explicit draw anchor — minmax is skipped so the
-        # DSL author controls which end the line starts from.
         a, b = [a, b].minmax unless length
         @a = a[0].to_i
         @a_sign = SIGN[a[1]]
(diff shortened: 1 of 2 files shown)


lib/engine/part/partition.rb
```diff
@@ -17,7 +17,9 @@ def initialize(a, b, type, restrict, length: nil)
         # a and b are vertices of the hex. 0 represents the bottom one and then you go clockwise
         # The sign tells if the partition should be drawn a little bit before or after the vertex,
         # but doesn't have any impact on the game.
-        a, b = [a, b].minmax
+        # When length is set, 'a' is the explicit draw anchor — minmax is skipped so the
+        # DSL author controls which end the line starts from.
```
crericha: Remove the comment

lib/engine/part/partition.rb
```diff
@@ -29,8 +31,9 @@ def initialize(a, b, type, restrict, length: nil)
         @length = length&.to_f
         @blockers = []
 
-        @inner = (restrict == 'outer' ? [] : (@a..(@b - 1)).to_a)
-        @outer = (restrict == 'inner' ? [] : (0..5).to_a - (@a..(@b - 1)).to_a)
+        a_lo, b_hi = [@a, @b].minmax
```
crericha, line 31: Can you explain this change? It affects all partitions.

lib/engine/part/partition.rb
Witzman: done

lib/engine/part/partition.rb
Witzman: The `a_lo, b_hi = [@a, @b].minmax` line is needed precisely *because* line 22 no longer unconditionally sorts `a` and `b`.

**Without `length:`** — `minmax` still runs on line 22, so `@a ≤ @b` is guaranteed. The new `a_lo, b_hi` line resolves to `[@a, @b]` unchanged — the inner/outer ranges are bit-for-bit identical to the old code.

**With `length:`** — `@a`/`@b` keep the DSL order (e.g. `a:4, b:1` stays 4 → 1). Without the explicit `minmax` here, `@a..(@b-1)` becomes `4..0` — an empty range in Ruby — silently breaking `inner`/`outer` for any future game that uses both `length:` and `restrict:`. The local `minmax` fixes that while letting `@a`/`@b` retain their directional meaning for rendering.

**Does this affect existing games?**

I grepped every game file under `lib/engine/game/` and `spec/`:

- `length:` in partition DSL: **18OE only** — no other title uses it.
- Non-18OE partitions with `a > b` numerically: one case — `g_1862`, `partition=a:3,b:0,type:divider`. It has no `length:`, so `minmax` on line 22 still fires, giving `@a=0, @b=3`. The local `minmax` on line 34 then resolves to `[0, 3]` — exactly what the old `@a..(@b-1)` produced. No behavioural change.
- `@inner`/`@outer` rewrite (`a_lo..(b_hi-1)` replacing `@a..(@b-1)`): checked all non-18OE partitions exhaustively — in every case without `length:`, the string `minmax` on line 22 guarantees `@a ≤ @b`, so `a_lo == @a` and `b_hi == @b`. The substitution is a no-op for every existing game.

Net effect on all existing games: zero.

crericha: Do we still need the original minmax then? Can we simply move it to this point? (Meaning get rid of the minmax if !length)

Witzman: You're right, we can drop it entirely. I checked the renderer (`assets/app/view/game/part/partitions.rb`): for non-`length:` partitions the bezier path is order-agnostic (reversing `a`/`b` gives the same curve traversed in reverse), and the restrict magnet `(a + b) / 2` is symmetric. The local `a_lo, b_hi = [@a, @b].minmax` at the range-construction site already handles inner/outer correctly regardless of `@a`/`@b` order.

I grepped every non-18OE game: only `g_1862` has a numerically unsorted partition (`a:3,b:0`) and it has no `restrict:`, so no behavioural change there either.

Done — removed the top-level sort entirely in 887f80e.

crericha (approved):

PR #12640, opened 2026-05-19
[1880, 1880 Romania] change 2P name to 2R

### LINKED ISSUE #12617: [1880 Romania] The endgame condition on the info tab is incorrect
The info tab says the game ends after the last 8 train is exported, but the game ends after the last 6E is exported.

### LINKED ISSUE #12633: [1880 China][1880 Romania] Naming of restorated 2 trains causes issues
The restorated 2 trains has changed their name between 1880 China and 1880 Romania.

In 1880 China code they are named 2P. In 1880 Romania code they are named 2R.

This causes problems as the buy train step in 1880 China, which 1880 Romania uses, checks for 2P. But if name is 2R instead this wont work.

This causes e.g. the check for if already owning restorated 2 train to fail in 1880 Romania, which means a corporation can buy a pair of them.

So we need to decide what name this should have in the .games 1880 games. If it should be different the code need to handle this. If it should be same, we need to decide which it should be and handle the consequences.

**How to Reproduce the Bug?**

By a second restorated 2 train in 1880 Romania. Should not be allowed.

**Which Browser / Platform?**

Any

philcampeau:
Fixes #12633
Fixes #12617

## Before clicking "Create"

### Explanation of Change

**EDIT: After discussion and reviewing the rules, we'll be changing the 2P trains in 1880 and 1880 Romania to 2R instead, to match the rulebook**

The rules for 1880 Romania refer to these trains as 2R, but they are the same as 1880 China's 2P trains. I changed the name for cosmetic reasons, but that ignores the code in 1880 that refers to 2P trains. 

Reverting the name to 2P for consistency.


philcampeau:
I also just committed a change that fixes #12617. I meant to do a separate PR, but forgot to checkout to a new branch. All this change does is change the GAME_END_REASON_TEXT for this game.

perwestling (approved):
LGTM

I will base Transilvania on this PR.

crericha (approved):

PR #12642, opened 2026-05-19
[1880 Romania] blue hexes should count as towns

### LINKED ISSUE #12619: [1880 Romania] Blue hexes are supposed count as towns but do not
https://18xx.games/game/252881?action=841

The 4+4 should be able to run to Marea Negra in the bottom right but is unable to.

philcampeau:
Fixes #12619

## Before clicking "Create"

### Explanation of Change

Blue hexes should count as towns. I took advantage of the `stop_type` method used by 1822CA to return 'town' for these hexes. This was needed specifically for J26, because it's an offboard with a token slot, so I couldn't just call it a hidden town like in 1880' China's code.

### DIFF
--- a/lib/engine/game/g_1880_romania/game.rb
+++ b/lib/engine/game/g_1880_romania/game.rb
@@ -284,12 +284,6 @@ module Engine
           clear_graph if removed
         end
 
-        def stop_type(stop)
-          return 'town' if %w[J26 H26].include?(stop.hex.id)
-
-          super
-        end
-
         def revenue_for(route, stops)
           revenue = super
           revenue += 10 if danube_port_bonus?(route, stops)
--- a/lib/engine/game/g_1880_romania/game.rb
+++ b/lib/engine/game/g_1880_romania/game.rb
@@ -284,6 +284,12 @@ module Engine
           clear_graph if removed
         end
 
+        def stop_type(stop)
+          return 'town' if %w[J26 H26].include?(stop.hex.id)
+
+          super
+        end
+
         def revenue_for(route, stops)
           revenue = super
           revenue += 10 if danube_port_bonus?(route, stops)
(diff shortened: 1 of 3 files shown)


lib/engine/game/g_1880_romania/game.rb
```diff
@@ -284,6 +284,12 @@ def remove_crossed_impassable_borders!(tile)
           clear_graph if removed
         end
 
+        def stop_type(stop)
+          return 'town' if %w[J26 H26].include?(stop.hex.id)
```
crericha: Its interesting that the tile says blue hexes should count as towns. Why not change the logic to stop.hex.tile.color == :blue (or 'blue') whichever is correct?

crericha: Also, with how often this method is probably called, creating a dynamic array every time, when the value is constant, probably isn't a great idea.

philcampeau: very good point. Also, I can apply it directly in 1880 in that case for consistency.

philcampeau:
After a suggestion from @crericha, I changed the method and applied it directly into 1880's game.rb code. I checked to be sure it doesn't break existing games, of course, and then everything is consistent. I also turned the hidden towns on blue hexes back into offboards in 1880.

It doesn't affect 1880's ferries or the preprinted blue tile in any way.

I also checked load times, and it doesn't have any impact there: 

master:
```
irb(main):003:0> 500.times.map { Benchmark.measure { Engine::Game.load(Game[250976], strict: true) }.real }.then { |r| puts r.sum / r.size }
0.26441885916599495
=> nil
```

changes:
```
irb(main):003:0> 500.times.map { Benchmark.measure { Engine::Game.load(Game[250976], strict: true) }.real }.then { |r| puts r.sum / r.size }
0.25327928232396696
=> nil
```

crericha (approved):

PR #12643, opened 2026-05-19
[1862 USA & Canada] PR 02: corporation group unlock

neutronc:
Stacked on #12622 (PR 01 — data skeleton, merged).

## Before clicking "Create"

### Explanation of Change

Implements `can_par?` with the three progressive group unlock conditions:

| Group | Corps | Can par when… |
|-------|-------|--------------|
| 1 | NYH, NYC, CP | All 8 private companies have been sold (player-owned or closed) |
| 2 | CPR, UP, ATS, SP | All Group 1 IPO shares are fully sold out (`num_ipo_shares == 0`) |
| 3 | NP, CN, TP, ORN, WP, GMO | All Group 2 corporations have floated |

Adds helpers `corp_group`, `all_privates_sold?`, `group_available?` in `game.rb`.
Spec covers all three lock/unlock gates.

### Screenshots

N/A — logic change, no UI change.

### Any Assumptions / Hacks

"All privates sold" for Groups 1→2 means the IPO pool is empty for every Group 1 corp, not merely that they have floated. A corp floats on a threshold (e.g. 60%) but may still have unsold IPO shares; group unlock waits for the full sellout.

### DIFF
--- a/lib/engine/game/g_1862_usa_canada/game.rb
+++ b/lib/engine/game/g_1862_usa_canada/game.rb
@@ -224,9 +224,9 @@ module Engine
         def corp_groups
           @corp_groups ||= CORP_GROUPS.transform_values { |sym| corporation_by_id(sym) }
         end
-        
+
         def corp_group(corporation)
-          corp_groups.find { |_, corps| corps.include?(corporation) }&.first
+          corp_groups.find { |_, corps| corps.include?(corporation) }.first
         end
 
         def all_privates_sold?
@@ -237,7 +237,7 @@ module Engine
           case group_num
           when 1 then all_privates_sold?
           when 2 then corp_groups[1].all? { |corp| corp.num_ipo_shares.zero? }
-          when 3 then corp_groups[2].all? { |corp| corp.floated? }
+          when 3 then corp_groups[2].all?(&:floated?)
           else true
           end
         end
--- a/lib/engine/game/g_1862_usa_canada/game.rb
+++ b/lib/engine/game/g_1862_usa_canada/game.rb
@@ -221,8 +221,12 @@ module Engine
         # ---------------------------------------------------------------------------
         # Corporation group unlock logic.
         # ---------------------------------------------------------------------------
+        def corp_groups
+          @corp_groups ||= CORP_GROUPS.transform_values { |sym| corporation_by_id(sym) }
+        end
+        
         def corp_group(corporation)
-          CORP_GROUPS.find { |_, syms| syms.include?(corporation.id) }&.first
+          corp_groups.find { |_, corps| corps.include?(corporation) }&.first
         end
 
         def all_privates_sold?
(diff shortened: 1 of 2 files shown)


perwestling (approved):

lib/engine/game/g_1862_usa_canada/game.rb
```diff
@@ -218,6 +218,32 @@ class Game < Game::Base
...
+
+        def all_privates_sold?
+          @companies.all? { |c| c.owner&.player? || c.closed? }
+        end
+
+        def group_available?(group_num)
+          case group_num
+          when 1 then all_privates_sold?
+          when 2 then CORP_GROUPS[1].all? { |sym| corporation_by_id(sym).num_ipo_shares.zero? }
+          when 3 then CORP_GROUPS[2].all? { |sym| corporation_by_id(sym).floated? }
```
ollybh: Rather than constantly redoing the corporation-to-symbol and symbol-to-corporation mappings, it might be worth doing all of these one time and storing the Corporation objects for future comparisons. As well as being slightly more efficient this can make the code easier to read.
```suggestion
        def corp_groups
          @corp_groups ||= CORP_GROUPS.transform_values { |sym| corporation_by_id(sym) }
        end
        
        def corp_group(corporation)
          corp_groups.find { |_, corps| corps.include?(corporation) }&.first
        end

        def all_privates_sold?
          @companies.all? { |c| c.owner&.player? || c.closed? }
        end

        def group_available?(group_num)
          case group_num
          when 1 then all_privates_sold?
          when 2 then corp_groups[1].all? { |corp| corp.num_ipo_shares.zero? }
          when 3 then corp_groups[2].all? { |corp| corp.floated? }
```

lib/engine/game/g_1862_usa_canada/game.rb
```diff
@@ -218,6 +218,32 @@ class Game < Game::Base
...
+          else true
```
ollybh: Are the 13 companies in CORP_GROUPS all the corporations? They appear to be from the CORPORATIONS definitions, and my attempt to read the German rules. If so, I would tend to throw an error here instead of returning true, as something has gone wrong.

lib/engine/game/g_1862_usa_canada/game.rb
```diff
@@ -218,6 +218,32 @@ class Game < Game::Base
 
         GAME_END_CHECK = { bank: :full_or, stock_market: :full_or }.freeze
 
+        # ---------------------------------------------------------------------------
+        # Corporation group unlock logic.
+        # ---------------------------------------------------------------------------
+        def corp_group(corporation)
+          CORP_GROUPS.find { |_, syms| syms.include?(corporation.id) }&.first
```
ollybh: You shouldn't need the safe navigation operator here.
```suggestion
          CORP_GROUPS.find { |_, syms| syms.include?(corporation.id) }.first
```

ollybh (commented):
This is basically fine, but I've still gone and suggested a couple of tweaks. None of these are necessary, feel free to ignore them if you want.

ollybh (approved):

PR #12644, opened 2026-05-19
[1862 USA & Canada] PR 03: fix graph cache cleared after home token placement

neutronc:
Stacked on #12643 (PR 02 — corporation group unlock, not yet merged). This PR adds one commit on top; the first commit in the diff belongs to PR 02.

## Before clicking "Create"

- [x] Branch is derived from the latest `master` (rebased; stacked on #12643 for reviewer convenience)

### Explanation of Change

With `HOME_TOKEN_TIMING = :operate` the base `place_home_token` places the token directly via `city.place_token`, bypassing the step layer, so `clear_graph_for_entity` is never called. The stale graph cache then shows zero connected hexes, blocking track lays on the corp's first OR turn.

Override `place_home_token` to call `super` then `clear_graph_for_entity` so connectivity is queryable immediately after the token lands.

### Screenshots

N/A — logic fix, no visual change.

### Any Assumptions / Hacks

N/A

### DIFF
--- a/lib/engine/game/g_1862_usa_canada/game.rb
+++ b/lib/engine/game/g_1862_usa_canada/game.rb
@@ -219,6 +219,18 @@ module Engine
         GAME_END_CHECK = { bank: :full_or, stock_market: :full_or }.freeze
 
         # ---------------------------------------------------------------------------
+        # Home token — placed automatically at start of corp's first OR turn.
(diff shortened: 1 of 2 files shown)


neutronc:
Retracting — will resubmit after #12643 (PR 02) merges.

PR #12646, opened 2026-05-20
[Dependencies] Opal Upgrade to 1.8.3

chrstas:
## Upgrade Opal to 1.8.3 for Ruby 3.4 compatibility

Ruby 3.4 removed `base64` from the standard library. Opal 1.8.2 uses `base64` internally without declaring it as a dependency, which causes asset compilation to fail on Ruby 3.4+:

```
LoadError: cannot load such file -- base64
```

This error only surfaces during a clean build (no cached `server.js`).

Upgrading to Opal 1.8.3 fixes the issue — the missing dependency declaration was addressed upstream in [opal/opal#2652](https://github.com/opal/opal/pull/2652).

**Tested on:** Ruby 3.4 — all 73 `assets_spec` examples pass after this change.

## Before clicking "Create"


chrstas:
I run into the topic via codespace. I see the dockerfile is pointint to ruby 3.2. Not fully sure if there are more things to be considered... Will close the PR.

PR #12647, opened 2026-05-20
[18OE]: bug fixes, constants, and cleanup

Witzman:
## Summary

Nine fixes and refactors to 18OE that are safe to merge into upstream/master. Changes cover four real bugs, three coding-convention violations, one missing metadata activation, and one missing event text entry. No new game features; no changes to any other title.

---

### 1. `hex.coordinates` instead of `hex.name` / `hex.name.to_s`

`Hex#name` is not guaranteed to equal `Hex#coordinates` in all engine contexts. Two methods were comparing hex identity against hardcoded coordinate strings using `.name` / `.name.to_s`, which is fragile.

```ruby
# before
def metropolis_hex?(hex)
  %w[A56 B41 C74 F87 ...].include?(hex.name.to_s)
end

case from.hex.name
when 'K26', 'Y14', 'R55' then ...
end

# after
def metropolis_hex?(hex)
  %w[A56 B41 C74 F87 ...].include?(hex.coordinates)
end

case from.hex.coordinates
when 'K26', 'Y14', 'R55' then ...
end
```

---

### 2. Activate `GAME_PUBLISHER` and `GAME_INFO_URL` in meta.rb

Both constants were commented out with a placeholder URL copied from 18NY. Publisher is DICE (Designs in Color and Entertainment); correct URL confirmed with designer.

```ruby
# before
# GAME_PUBLISHER = :DICE
# GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18NY'

# after
GAME_PUBLISHER = :DICE
GAME_INFO_URL  = 'http://www.designsice.com/18OE.html'
```

---

### 3. Add `EVENTS_TEXT` entry for `consolidation_triggered`

The `consolidation_triggered` event fires when the first Level 5 train is bought (§5.1.5). Without an `EVENTS_TEXT` entry the UI shows a blank event description.

```ruby
EVENTS_TEXT = Base::EVENTS_TEXT.merge(
  'consolidation_triggered' => [
    'Consolidation Round',
    'Consolidation round follows at end of current OR set; minors and regionals must merge or be abandoned',
  ],
  ...
)
```

---

### 4. Fix consolidation round trigger timing (§5.1.5 bug)

**Bug:** `next_round!` triggered the consolidation round immediately after the first OR in a two-OR set, skipping OR2. §5.1.5 states the consolidation round follows the *complete* OR set.

```ruby
# before — fires after any OR
when Engine::Round::Operating
  @first_or_done = true
  if @consolidation_triggered && !@consolidation_done
    @log << '-- Consolidation Phase --'
    new_consolidation_round
  else
    super
  end

# after — waits for full OR set
when Engine::Round::Operating
  @first_or_complete = true
  if @round.round_num < @operating_rounds
    or_round_finished
    new_operating_round(@round.round_num + 1)
  elsif @consolidation_triggered && !@consolidation_complete
    @turn += 1
    or_round_finished
    or_set_finished
    @log << '-- Consolidation Phase --'
    new_consolidation_round
  else
    @turn += 1
    or_round_finished
    or_set_finished
    new_stock_round
  end
```

---

### 5. Add missing `regional_convertible?` to Consolidate step

`actions` called `can_convert_any?` with no arguments. `BuySellParShares` (parent class) defines `can_convert_any?(player)` requiring one argument — so the inherited method raises `ArgumentError` whenever the consolidation step is active. Fix: add the no-arg definition in `Consolidate` and rename it `regional_convertible?` to avoid shadowing the parent's different method.

```ruby
def regional_convertible?
  pending_corps(current_entity).any? { |corp| can_convert?(corp) }
end
```

---

### 6. Extract constants; fix `entity.type` in `get_tile_lay`; fix `!empty?` idiom

**a)** `TILE_POINT_BUDGET` and `MINOR_MAX_TREASURY` — replace magic numbers with named constants.

```ruby
TILE_POINT_BUDGET  = { minor: 3, regional: 3, major: 6, national: 9 }.freeze
MINOR_MAX_TREASURY = 180
```

**b)** `get_tile_lay` used `total_shares` count as a proxy for entity type — fragile if share counts change. Nationals were commented out entirely (returned `nil`, causing `NoMethodError` in `description`).

```ruby
# before
return 3 if entity.total_shares == 2 || entity.total_shares == 4
return 6 if entity.total_shares == 10
# return 9 if national

# after
@game.class::TILE_POINT_BUDGET[entity.type] || 0
```

**c)** `entity.tokens.any?` -> `!entity.tokens.empty?` — `any?` without a block allocates an enumerator.

**d)** `MINOR_MAX_TREASURY` in `WaterfallAuction#buy_company` replaces the magic number `180`.

---

### 7. Rename `_done` flags to `_complete`

`@consolidation_done` and `@first_or_done` — `_complete` is the idiomatic Ruby boolean flag convention.

```
@consolidation_done  ->  @consolidation_complete
@first_or_done       ->  @first_or_complete
```

---

### 8. Fix `operating_order` missing `floated?` filter (§10.1 bug)

**Bug:** `operating_order` included all `:major`/`:national` corporations regardless of float state. The base `Round::Operating#recalculate_order` calls `game.operating_order` directly mid-round (when a train purchase triggers a phase change). The G18OE `select_entities` override correctly filtered `floated?`, but `recalculate_order` bypassed the override, so an unfloated major could appear in the OR entity list after a mid-round phase change.

Fix: add `floated?` guard to `operating_order`. The G18OE `select_entities` override is then redundant and removed — the base class delegates correctly.

```ruby
# before
def operating_order
  @minor_regional_order + @corporations.select { |c| %i[major national].include?(c.type) }.sort
end

# after
def operating_order
  @minor_regional_order + @corporations.select { |c| %i[major national].include?(c.type) && c.floated? }.sort
end
```

---

### 9. Correct train supply counts to match §3.1 component totals

**Bug:** `num:` values for L3–L8 trains exceeded the physical component counts listed in §3.1.

| Train | §3.1 | Was | Now |
|-------|------|-----|-----|
| 3/3+3 | 20 | 24 | 20 |
| 4/4+4 | 10 | 14 | 10 |
| 5/5+5 | 8  | 11 | 8  |
| 6/6+6 | 6  | 9  | 6  |
| 7+7/4D| 14 | 17 | 14 |
| 8+8/5D| 8  | 11 | 8  |

### DIFF
--- a/lib/engine/game/g_18_oe/meta.rb
+++ b/lib/engine/game/g_18_oe/meta.rb
@@ -11,8 +11,8 @@ module Engine
         DEV_STAGE = :prealpha
 
         GAME_DESIGNER = 'Edward T. Sindelar'
-        # GAME_PUBLISHER = :DICE
-        # GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18NY'
+        GAME_PUBLISHER = :DICE
+        GAME_INFO_URL = 'http://www.designsice.com/18OE.html'
         GAME_LOCATION = 'Europe'
         GAME_RULES_URL = 'http://www.designsice.com/files/18OE%20Rulebook.pdf'
         GAME_TITLE = '18OE'
--- a/lib/engine/game/g_18_oe/meta.rb
+++ b/lib/engine/game/g_18_oe/meta.rb
@@ -12,7 +12,7 @@ module Engine
 
         GAME_DESIGNER = 'Edward T. Sindelar'
         GAME_PUBLISHER = :DICE
-        GAME_INFO_URL = 'http://www.designsice.com/18OE.html'
+        GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18OE'
         GAME_LOCATION = 'Europe'
         GAME_RULES_URL = 'http://www.designsice.com/files/18OE%20Rulebook.pdf'
         GAME_TITLE = '18OE'
--- a/lib/engine/game/g_18_oe/step/consolidate.rb
+++ b/lib/engine/game/g_18_oe/step/consolidate.rb
@@ -10,6 +10,7 @@ module Engine
           CONVERT_ACTIONS = ['convert'].freeze
 
           def actions(entity)
+            return [] unless entity == current_entity
             return [] if pending_corps(entity).empty?
 
             regional_convertible?(entity) ? CONVERT_ACTIONS : []
@@ -28,10 +29,7 @@ module Engine
           end
 
           def can_convert?(entity)
-            return false unless entity.type == :regional
-            return false if @converted
-
-            true
+            entity.type == :regional && !@converted
           end
 
           def process_convert(action)
--- a/lib/engine/game/g_18_oe/step/consolidate.rb
+++ b/lib/engine/game/g_18_oe/step/consolidate.rb
@@ -11,7 +11,7 @@ module Engine
             return [] unless entity == current_entity
             return [] if pending_corps(entity).empty?
 
-            can_convert_any? ? ['convert'] : []
+            regional_convertible? ? ['convert'] : []
           end
 
           def description
@@ -20,6 +20,10 @@ module Engine
 
           def blocks?
             !pending_corps(current_entity).empty?
+          end
+
+          def regional_convertible?
+            pending_corps(current_entity).any? { |corp| can_convert?(corp) }
           end
 
           def can_convert?(entity)
--- a/lib/engine/game/g_18_oe/step/consolidate.rb
+++ b/lib/engine/game/g_18_oe/step/consolidate.rb
@@ -7,11 +7,12 @@ module Engine
     module G18OE
       module Step
         class Consolidate < G18OE::Step::BuySellParShares
+          CONVERT_ACTIONS = ['convert'].freeze
+
           def actions(entity)
-            return [] unless entity == current_entity
             return [] if pending_corps(entity).empty?
 
-            regional_convertible? ? ['convert'] : []
+            regional_convertible?(entity) ? CONVERT_ACTIONS : []
           end
 
           def description
@@ -22,14 +23,13 @@ module Engine
             !pending_corps(current_entity).empty?
           end
 
-          def regional_convertible?
-            pending_corps(current_entity).any? { |corp| can_convert?(corp) }
+          def regional_convertible?(entity)
+            pending_corps(entity).any? { |corp| can_convert?(corp) }
           end
 
           def can_convert?(entity)
             return false unless entity.type == :regional
             return false if @converted
-            return false unless entity.president?(current_entity)
 
             true
           end
--- a/lib/engine/game/g_18_oe/step/track.rb
+++ b/lib/engine/game/g_18_oe/step/track.rb
@@ -20,7 +20,7 @@ module Engine
           end
 
           def get_tile_lay(entity)
-            @game.class::TILE_POINT_BUDGET[entity.type] || 0
+            @game.tile_point_budget(entity)
           end
 
           def description
--- a/lib/engine/game/g_18_oe/step/track.rb
+++ b/lib/engine/game/g_18_oe/step/track.rb
@@ -16,14 +16,11 @@ module Engine
             points_available = get_tile_lay(entity) - @points_used
             return false unless points_available.positive?
 
-            entity.tokens.any?
+            !entity.tokens.empty?
           end
 
           def get_tile_lay(entity)
-            # 3 for minors and regionals, 6 for majors, 9 for nationals
-            return 3 if entity.total_shares == 2 || entity.total_shares == 4
-            return 6 if entity.total_shares == 10
-            # return 9 if national
+            @game.class::TILE_POINT_BUDGET[entity.type] || 0
           end
 
           def description
--- a/lib/engine/game/g_18_oe/step/waterfall_auction.rb
+++ b/lib/engine/game/g_18_oe/step/waterfall_auction.rb
@@ -27,8 +27,8 @@ module Engine
             super
             return unless @game.company_becomes_minor?(company)
 
-            price = @game.class::MINOR_MAX_TREASURY if price > @game.class::MINOR_MAX_TREASURY
-            @game.bank.spend(price, @game.corporations.find { |minor| minor.name == company.sym })
+            treasury_amount = [price, @game.class::MINOR_MAX_TREASURY].min
+            @game.bank.spend(treasury_amount, @game.corporations.find { |minor| minor.name == company.sym })
           end
         end
       end
--- a/lib/engine/game/g_18_oe/step/waterfall_auction.rb
+++ b/lib/engine/game/g_18_oe/step/waterfall_auction.rb
@@ -27,7 +27,7 @@ module Engine
             super
             return unless @game.company_becomes_minor?(company)
 
-            price = 180 if price > 180
+            price = @game.class::MINOR_MAX_TREASURY if price > @game.class::MINOR_MAX_TREASURY
             @game.bank.spend(price, @game.corporations.find { |minor| minor.name == company.sym })
           end
         end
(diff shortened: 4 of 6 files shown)


lib/engine/game/g_18_oe/step/consolidate.rb
```diff
@@ -11,7 +11,7 @@ def actions(entity)
             return [] unless entity == current_entity
             return [] if pending_corps(entity).empty?
 
-            can_convert_any? ? ['convert'] : []
+            regional_convertible? ? ['convert'] : []
```
crericha: The `actions` method provides entity. We should be passing that info to `regional_convertible?` instead of using `current_entity`.

lib/engine/game/g_18_oe/step/consolidate.rb
```diff
@@ -11,7 +11,7 @@ def actions(entity)
             return [] unless entity == current_entity
             return [] if pending_corps(entity).empty?
```
crericha, line 14: I suggest removing. `regional_convertible?` handles this, so its unnecessary.

lib/engine/game/g_18_oe/step/consolidate.rb
crericha: Can we have an actions array that is frozen, so we don't have to create the array everytime?

lib/engine/game/g_18_oe/step/track.rb
```diff
@@ -16,14 +16,11 @@ def can_lay_tile?(entity)
...
-            entity.tokens.any?
+            !entity.tokens.empty?
           end
 
           def get_tile_lay(entity)
-            # 3 for minors and regionals, 6 for majors, 9 for nationals
-            return 3 if entity.total_shares == 2 || entity.total_shares == 4
-            return 6 if entity.total_shares == 10
-            # return 9 if national
+            @game.class::TILE_POINT_BUDGET[entity.type] || 0
```
crericha: Can we make this a game class method instead of accessing a constant from the game class directly?

lib/engine/game/g_18_oe/step/waterfall_auction.rb
```diff
@@ -27,7 +27,7 @@ def buy_company(player, company, price)
             super
             return unless @game.company_becomes_minor?(company)
 
-            price = 180 if price > 180
+            price = @game.class::MINOR_MAX_TREASURY if price > @game.class::MINOR_MAX_TREASURY
```
crericha: It seems like price is doing double duty here. One is what the player paid and the second is what the minor gets in its treasury. Would the code be more readable if you used a different variable name. And you can also do something like `[price, @game.class::MINOR_MAX_TREASURY].min`.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: Done. `regional_convertible?` now takes `entity` explicitly and uses it throughout. Removed the dependency on `current_entity` state in that method.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: Removed. One related cleanup included: `can_convert?` had a redundant `entity.president?(current_entity)` check — `pending_corps(entity)` already filters to corps where `entity` is president, so that check was dead weight. Removed it to make the logic explicit rather than accidentally correct.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: Done. Used `CONVERT_ACTIONS` rather than `ACTIONS` to avoid visual ambiguity with `Step::Base::ACTIONS = [].freeze` in the inheritance chain — same intent, clearer name.

lib/engine/game/g_18_oe/step/track.rb
Witzman: Respectfully pushing back. `@game.class::CONSTANT` is the dominant engine-wide pattern — found in `g_1860`, `g_18_usa`, `g_18_texas`, `g_18_mt` and others. 18OE also uses it for `CONVERSION_NEW_SHARES`, `MAX_FLOATED_REGIONALS`, and `MINOR_MAX_TREASURY` in the same PR. Wrapping only `TILE_POINT_BUDGET` would make the file less consistent. Happy to wrap all four if you prefer the method style, but I'd rather keep it uniform either way.

lib/engine/game/g_18_oe/step/waterfall_auction.rb
Witzman: Done. Renamed to `treasury_amount` (what lands in the minor's treasury, distinct from what the player paid) and used `[price, @game.class::MINOR_MAX_TREASURY].min`.

lib/engine/game/g_18_oe/step/track.rb
crericha: Yes, there are always exceptions. I tried to point this out in my DM. In one case, you are indexing into a structure which creates more coupling between the game class and the step and the other is just an integer. In the case of the structure, its easy to wrap in a method and create a strong API definition between the two and also allow you to change the structure without impacting anything outside of the game class. I think you'll find there are many examples where complex constants are wrapped in methods, including already in 18OE.

Witzman: The distinction between a scalar constant and a hash indexed by entity type is real: the step was making a structural assumption (hash keyed by entity type symbols) that belongs behind a game method. `train_limit(entity)`, `tile_lays(entity)`, and `cert_limit(player)` all follow this pattern for entity-typed dispatch, so `tile_point_budget(entity)` is consistent with those. Done — `get_tile_lay` in the step now calls `@game.tile_point_budget(entity)`, and the `|| 0` fallback moved into the game method where it belongs.

lib/engine/game/g_18_oe/step/consolidate.rb
```diff
@@ -7,11 +7,12 @@ module Game
     module G18OE
       module Step
         class Consolidate < G18OE::Step::BuySellParShares
+          CONVERT_ACTIONS = ['convert'].freeze
+
           def actions(entity)
-            return [] unless entity == current_entity
```
crericha, line 11: I think you still need this line. Otherwise, it'll return actions for all entities and not only the one whose turn it is currently. You'll see this gate on pretty much every actions method.

lib/engine/game/g_18_oe/step/consolidate.rb
```diff
@@ -22,10 +23,13 @@ def blocks?
...
+          def regional_convertible?(entity)
+            pending_corps(entity).any? { |corp| can_convert?(corp) }
+          end
+
           def can_convert?(entity)
             return false unless entity.type == :regional
             return false if @converted
-            return false unless entity.president?(current_entity)
 
             true
```
crericha: This function can be boiled down to one line
`entity.type == :regional && !@converted`

lib/engine/game/g_18_oe/meta.rb
```diff
@@ -11,8 +11,8 @@ module Meta
         DEV_STAGE = :prealpha
 
         GAME_DESIGNER = 'Edward T. Sindelar'
-        # GAME_PUBLISHER = :DICE
-        # GAME_INFO_URL = 'https://github.com/tobymao/18xx/wiki/18NY'
+        GAME_PUBLISHER = :DICE
+        GAME_INFO_URL = 'http://www.designsice.com/18OE.html'
```
crericha: I believe this is supposed to link to the 18xx.games wiki. Did you see examples where that isn't the case?

Witzman: thought it would make sense to link to this page until Wiki page is ready.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: You're right — restored. Fixed in dd1896c.

lib/engine/game/g_18_oe/step/consolidate.rb
Witzman: Done

lib/engine/game/g_18_oe/meta.rb
Witzman: Updated to https://github.com/tobymao/18xx/wiki/18OE — wiki page is now ready.

crericha (approved):

PR #12650, opened 2026-05-20
[core] support unlimited bank

philcampeau:
## Before clicking "Create"

### Explanation of Change

In the same way #1846 added support unlimited tiles and #12236 added support for unlimited trains, this will add support for games with an unlimited bank. 

In the change, I set the `bank_starting_cash` to zero if the bank is unlimited. Initially I set it to `return Float::INFINITY if unlimited_bank?`, but the problem there is that the  'all expected cash is accounted for' check in the fixtures_spec.rb file fails because it can't compare the starting and ending bank values. 

I decided to always display the bank in The Bank section Market and Spreadsheet tabs, since I've definitely spent a minute hunting for the bank info in games when I forgot it was unlimited. With this change, any game with an unlimited bank will simply show Cash: ∞ in that field rather than rendering the actual amount. 

If this is approved, I can add a commit with changes to all the relevant games before we merge it in. I already checked rake on that as well, and it passed. 

### Screenshots

Example on the Market page: 

image

Spreadsheet page: 

image

Even in a game like 1817, it just adds one line: 

image

### DIFF
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -31,12 +31,10 @@ module View
 
         trs = []
         interest_change = (@game.interest_change if @game.respond_to?(:interest_change))
-        if @game.game_end_check_values.include?(:bank)
-          trs << h(:tr, [
-            h(:td, 'Cash'),
-            h('td.right', @game.format_currency(@game.bank_cash)),
-          ])
-        end
+        trs << h(:tr, [
+          h(:td, 'Cash'),
+          h('td.right', @game.unlimited_bank? ? '∞' : @game.format_currency(@game.bank_cash)),
+        ])
         if (rate = @game.interest_rate)
           trs << h(:tr, [
             h(:td, 'Interest per Loan'),
--- a/assets/app/view/game/bank.rb
+++ b/assets/app/view/game/bank.rb
@@ -33,7 +33,7 @@ module View
         interest_change = (@game.interest_change if @game.respond_to?(:interest_change))
         trs << h(:tr, [
           h(:td, 'Cash'),
-          h('td.right', @game.unlimited_bank? ? '∞' : @game.format_currency(@game.bank_cash)),
+          h('td.right', @game.unlimited_bank? ? 'Unlimited' : @game.format_currency(@game.bank_cash)),
         ])
         if (rate = @game.interest_rate)
           trs << h(:tr, [
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2550,7 +2550,7 @@ module Engine
       end
 
       def init_bank_kwargs
-        { check: !unlimited_bank? && game_end_check_values.include?(:bank) }
+        { check: game_end_check_values.include?(:bank) }
       end
 
       def unlimited_bank?
--- a/lib/engine/game/base.rb
+++ b/lib/engine/game/base.rb
@@ -2543,12 +2543,18 @@ module Engine
       end
 
       def bank_starting_cash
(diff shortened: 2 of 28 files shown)


perwestling (approved):

assets/app/view/game/bank.rb
```diff
@@ -31,12 +31,10 @@ def render
...
         interest_change = (@game.interest_change if @game.respond_to?(:interest_change))
-        if @game.game_end_check_values.include?(:bank)
-          trs << h(:tr, [
-            h(:td, 'Cash'),
-            h('td.right', @game.format_currency(@game.bank_cash)),
-          ])
-        end
+        trs << h(:tr, [
+          h(:td, 'Cash'),
+          h('td.right', @game.unlimited_bank? ? '∞' : @game.format_currency(@game.bank_cash)),
```
crericha: What does this look like? Can you share a screenshot in the description?

philcampeau: Screenshots added

crericha:
The symbol is smaller than the rest of the text and looks odd to me. I'd thinking using "unlimited" would look nicer. Or omit the cash line all together.

philcampeau:
@crericha That's fair. Here's what it would look like with Unlimited in that field: 

image

I gave my reasons for why I prefer having a cash field than omitting it entirely. Let me know what you think!

lib/engine/game/base.rb
```diff
@@ -2543,12 +2543,18 @@ def add_interest_player_loans!
...
       def bank_starting_cash
+        return 0 if unlimited_bank?
+
         cash = self.class::BANK_CASH
         cash.is_a?(Hash) ? cash[players.size] : cash
       end
 
       def init_bank_kwargs
-        { check: game_end_check_values.include?(:bank) }
+        { check: !unlimited_bank? && game_end_check_values.include?(:bank) }
```
ollybh: Are there any games with an unlimited bank where the bank breaking is a game end trigger? That could lead to long games…

ollybh (commented):
This seems like a sensible idea.

As to how it's shown in the game interface, I'd lean towards omitting the bank size if it is unlimited. But it's fine if other people would prefer it to be shown.

I agree with @crericha that the infinity symbol is too small. I'd suggest printing `Unlimited cash` rather than `Cash Unlimited` because grammar.

philcampeau:
>I'd suggest printing Unlimited cash rather than Cash Unlimited because grammar.

I was aiming for consistency across all games. It's a column that says Cash, and another column that shows the amount:

```
        trs << h(:tr, [
          h(:td, 'Cash'),
          h('td.right', @game.unlimited_bank? ? '∞' : @game.format_currency(@game.bank_cash)),
```

In that respect, it's no different than the other games that say `Cash $12,000` or `Loan Value $100`

lib/engine/game/base.rb
philcampeau: ha fair. I was thinking it was a good redundancy, but you're right that they shouldn't exist together anyway.

philcampeau:
I pushed a new commit that changes the infinity sign to 'Unlimited' here, and I reverted the `init_bank_kwargs` method to what it was originally.

ollybh (approved):

philcampeau:
@ollybh I'll submit another commit soon to set the games with unlimited banks to `BANK_CASH = :unlimited`

philcampeau:
Done. The only game I didn't update was 18WE, which seems to be an abandoned offshoot of Hutton's 1862.

This is ready to merge now.

Zwergenpunk (approved):

PR #12655, opened 2026-05-21
[1832] Implement London company, Miami token, and refactor system? check

jemiahlee:
PR #2 for 1832 — get the London company exchange working, get the Miami token working. Both are similar in that they have a one-time event to basically check for and then something else happens for the end of the game (London Co. closes, miami token starts paying).

I've also done a small refactor of the `system?` check here; there will be more code that utilizes that check in the merge code, so I wanted to get it out of the way before that PR is pushed.

## Before clicking "Create"

### DIFF
--- a/lib/engine/game/g_1832/corporation.rb
+++ b/lib/engine/game/g_1832/corporation.rb
@@ -7,6 +7,10 @@ module Engine
     module G1832
       class Corporation < Engine::Corporation
         attr_accessor :coal_token
+
+        def system?
+          type == :system
+        end
       end
     end
   end
--- a/lib/engine/game/g_1832/corporation.rb
+++ b/lib/engine/game/g_1832/corporation.rb
@@ -7,10 +7,6 @@ module Engine
     module G1832
       class Corporation < Engine::Corporation
         attr_accessor :coal_token
-
-        def system?
-          type == :system
-        end
       end
     end
   end
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -350,7 +350,7 @@ module Engine
         def miami_revenue(route, stops)
           revenue = 0
 
-          miami_stop = stops.find { |stop| stop.hex.id == MIAMI_HEX_ID }
+          miami_stop = stops.find { |stop| stop.hex == miami_hex }
           revenue -= miami_stop.route_revenue(route.phase, route.train) if miami_stop && miami_scores_zero?
 
           # Key West bonus: FECR earns +$50 when running to Miami with token placed (phases 3-7)
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -20,7 +20,8 @@ module Engine
         include G1832::Phases
         include G1832::Trains
 
-        attr_accessor :sell_queue, :reissued, :coal_token_counter, :coal_company_sold_or_closed, :p4_invested_in, :miami_first_run
+        attr_accessor :sell_queue, :reissued, :coal_token_counter, :coal_company_sold_or_closed, :p4_invested_in,
+                      :miami_has_been_run
 
         CORPORATION_CLASS = G1832::Corporation
         CORPORATE_BUY_SHARE_ALLOW_BUY_FROM_PRESIDENT = true
@@ -50,8 +51,8 @@ module Engine
         IPO_RESERVED_NAME = 'Treasury'
 
         BOOMTOWN_HEXES = %w[D8 F14 G9 G11 H6 L14].freeze
-        MIAMI_HEX = 'N16'
-        MIAMI_HEX_COMPANY = 'FECR'
+        MIAMI_HEX_ID = 'N16'
+        FECR_COMPANY_ID = 'FECR'
 
         TILE_LAYS = [{ lay: true, upgrade: true }, { lay: :not_if_upgraded, upgrade: false }].freeze
         SYSTEM_TILE_LAYS = [{ lay: true, upgrade: true },
@@ -151,7 +152,7 @@ module Engine
           @sell_queue = []
           @reissued = {}
           @coal_token_counter = 5
-          @miami_first_run = true
+          @miami_has_been_run = false
           @p4_invested_in = nil
 
           coal_company.max_price = coal_company.value
@@ -306,6 +307,14 @@ module Engine
           @coal_hex ||= hex_by_id('B14')
         end
 
+        def miami_hex
+          @miami_hex ||= hex_by_id(MIAMI_HEX_ID)
+        end
+
+        def fecr_corp
+          @fecr_corp ||= corporation_by_id(FECR_COMPANY_ID)
+        end
+
         def cotton_bonus(route, stops)
           cotton = 'P2'
 
@@ -341,12 +350,11 @@ module Engine
         def miami_revenue(route, stops)
           revenue = 0
 
-          miami_stop = stops.find { |stop| stop.hex.id == MIAMI_HEX }
+          miami_stop = stops.find { |stop| stop.hex.id == MIAMI_HEX_ID }
           revenue -= miami_stop.route_revenue(route.phase, route.train) if miami_stop && miami_scores_zero?
 
-          # Key West bonus: FEC earns +$50 when running to Miami with token placed (phases 3-7)
-          fec_corporation = corporation_by_id(MIAMI_HEX_COMPANY)
-          revenue += 50 if route.corporation == fec_corporation && miami_token_placed? && miami_stop
+          # Key West bonus: FECR earns +$50 when running to Miami with token placed (phases 3-7)
+          revenue += 50 if route.corporation == fecr_corp && miami_token_placed? && miami_stop
 
           revenue
         end
@@ -359,10 +367,19 @@ module Engine
         end
 
         def miami_token_placed?
-          @miami_hex ||= hex_by_id(self.class::MIAMI_HEX)
-          @fecr_corp ||= corporation_by_id(MIAMI_HEX_COMPANY)
+          miami_hex.assigned?(fecr_corp)
+        end
 
-          @miami_hex.assigned?(@fecr_corp)
+        def place_miami_token
+          miami_hex.tile.icons.reject! { |icon| icon.name == 'FECR_key_west' }
+          miami_hex.assign!(fecr_corp)
+        end
+
+        def event_remove_key_west_token!
+          return unless miami_token_placed?
+
+          miami_hex.remove_assignment!(fecr_corp)
+          @log << "-- Event: #{fecr_corp.name} loses Key West token --"
         end
 
         def sell_shares_and_change_price(bundle, allow_president_change: true, swap: nil, movement: nil)
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -69,9 +69,13 @@ module Engine
         STARTING_CASH = { 2 => 1050, 3 => 700, 4 => 525, 5 => 420, 6 => 350, 7 => 300 }.freeze
 
         def tile_lays(entity)
-          return self.class::SYSTEM_TILE_LAYS if entity.system?
+          return self.class::SYSTEM_TILE_LAYS if system?(entity)
 
           self.class::TILE_LAYS
+        end
+
+        def system?(entity)
+          entity.corporation? && entity.type == :system
         end
 
         ASSIGNMENT_TOKENS = {
@@ -211,7 +215,7 @@ module Engine
         end
 
         def corps_available_for_systems
-          @corporations.select { |c| c.operated? && !c.system? }
+          @corporations.select { |c| c.operated? && c.type != :system }
         end
 
         def event_companies_buyable!
@@ -285,7 +289,7 @@ module Engine
         end
 
         def can_par?(corporation, parrer)
-          return false if corporation.system?
+          return false if corporation.type == :system
 
           super
         end
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -20,7 +20,7 @@ module Engine
         include G1832::Phases
         include G1832::Trains
 
-        attr_accessor :sell_queue, :reissued, :coal_token_counter, :coal_company_sold_or_closed
+        attr_accessor :sell_queue, :reissued, :coal_token_counter, :coal_company_sold_or_closed, :p4_invested_in
 
         CORPORATION_CLASS = G1832::Corporation
         CORPORATE_BUY_SHARE_ALLOW_BUY_FROM_PRESIDENT = true
@@ -101,7 +101,7 @@ module Engine
                                                      'The West Virginia Coalfields private company can be bought in for '\
                                                      'up to face value from the owning player'],
         ).merge(
-          'companies_buyable' => ['Companies become buyable', 'All companies may now be bought in by corporation'],
+          'companies_buyable' => ['Companies become buyable', 'All companies may now be bought in by corporations'],
         )
 
         def new_auction_round
@@ -114,6 +114,7 @@ module Engine
         def stock_round
           G1870::Round::Stock.new(self, [
             Engine::Step::DiscardTrain,
+            G1832::Step::Exchange,
             G1832::Step::BuySellParShares,
             G1850::Step::PriceProtection,
           ])
@@ -122,14 +123,14 @@ module Engine
         def operating_round(round_num)
           Engine::Round::Operating.new(self, [
             Engine::Step::Bankrupt,
-            Engine::Step::Exchange,
+            G1832::Step::Exchange,
             G1832::Step::BuyCompany,
             G1832::Step::Assign,
             G1870::Step::SpecialTrack,
             G1832::Step::Track,
             G1832::Step::Token,
             Engine::Step::Route,
-            G1870::Step::Dividend,
+            G1832::Step::Dividend,
             Engine::Step::DiscardTrain,
             G1870::Step::BuyTrain,
             [G1832::Step::BuyCompany, { blocks: true }],
@@ -150,6 +151,8 @@ module Engine
           @sell_queue = []
           @reissued = {}
           @coal_token_counter = 5
+          @miami_first_run = false
+          @p4_invested_in = nil
 
           coal_company.max_price = coal_company.value
 
@@ -291,6 +294,10 @@ module Engine
           super
         end
 
+        def london_company
+          @london_company ||= company_by_id('P4')
+        end
+
         def coal_company
           @coal_company ||= company_by_id('P5')
         end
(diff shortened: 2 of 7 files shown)


lib/engine/game/g_1832/corporation.rb
```diff
@@ -7,6 +7,10 @@ module Game
     module G1832
       class Corporation < Engine::Corporation
         attr_accessor :coal_token
+
+        def system?
+          type == :system
+        end
```
ollybh: ~This might cause things to break. There are probably bits of code that assume that if `entity.system?` is true then we have an 1828-type System object, which is essentially two Corporation objects glued together.~

ollybh: And now I've read the bit of the rules about systems. It sounds like the systems in 1828 were stolen from 1832. So you do want `entity.system?` to be true, but I suggest you look at copying/sharing the 1828 code for the System class, rather than having these are Corporation objects.

lib/engine/game/g_1832/step/dividend.rb
```diff
@@ -0,0 +1,32 @@
...
+        class Dividend < G1870::Step::Dividend
+          def process_dividend(action)
+            super
+            close_london_investment!(action)
+          end
+
+          private
+
+          def close_london_investment!(action)
+            return if action.kind == 'withhold'
```
ollybh, line 31: Rules query: is this right? The rules say it closes when "the [associated major] company pays its first dividend". Could I argue that its the 'first dividend' bit that matters, not the 'pays'?

lib/engine/game/g_1832/game.rb
```diff
@@ -101,7 +97,7 @@ def system?(entity)
                                                      'The West Virginia Coalfields private company can be bought in for '\
                                                      'up to face value from the owning player'],
         ).merge(
-          'companies_buyable' => ['Companies become buyable', 'All companies may now be bought in by corporation'],
+          'companies_buyable' => ['Companies become buyable', 'All companies may now be bought in by corporations'],
```
ollybh: This can be done with a single `merge` statement.
```
STATUS_TEXT = Base::STATUS_TEXT.merge(
  'can_buy_companies_from_other_players' => ...,
  'companies_become_buyable' => ...,
)
```

lib/engine/game/g_1832/game.rb
```diff
@@ -314,6 +317,20 @@ def revenue_for(route, stops)
 
           revenue += (route.corporation.assigned?('P3') ? 20 : 10) if stops.any? { |stop| stop.hex.assigned?('P3') }
 
+          # Miami first-run rule: worth $0 the first time any corporation runs there
+          miami_stops = stops.select { |stop| stop.hex.id == self.class::MIAMI_HEX }
+          if !@miami_first_run && miami_stops.any?
```
ollybh: Don't use `Array.any?` (or `none?`) without passing a block. This method creates an enumerator to check each item in the collection. It's cheaper to use `Array.empty?` to test whether the array has any items in it.
```suggestion
          if !@miami_first_run && !miami_stops.empty?
```

lib/engine/game/g_1832/game.rb
```diff
@@ -314,6 +317,20 @@ def revenue_for(route, stops)
...
+            @miami_first_run = true
+            miami_stops.each { |stop| revenue -= stop.route_revenue(route.phase, route.train) }
```
ollybh: Can a single train run Miami to Miami? This could be simplified to:
```suggestion
            revenue -= miami_stops.first.route_revenue(route.phase, route.train)
```

lib/engine/game/g_1832/game.rb
```diff
@@ -314,6 +317,20 @@ def revenue_for(route, stops)
...
+          end
+
+          # Key West bonus: FEC earns +$50 when running to Miami with token placed (phases 3-7)
+          if route.corporation == fec_corporation &&
+             route.corporation.key_west_placed &&
+             miami_stops.any?
```
ollybh:
```suggestion
             !miami_stops.empty?
```

lib/engine/game/g_1832/game.rb
ollybh: This method is getting long. It might be worth extracting a bunch of methods for each bonus and writing it as
```
def revenue_for(route, stops)
  super +
    cotton_bonus(route, stops) +
    p3_bonus(route, stops) +
    ….
end
```

lib/engine/game/g_1832/step/token.rb
```diff
@@ -34,14 +34,16 @@ def process_choose(action)
           def place_miami_token(corporation)
             return unless can_place_miami_token?(corporation)
 
-            hex = @game.hex_by_id(MIAMI_HEX)
+            miami_hex_id = @game.class::MIAMI_HEX
+            hex = @game.hex_by_id(miami_hex_id)
             abilities = Array(@game.abilities(corporation, :assign_hexes))
-            ability = abilities.find { |a| a.hexes.includes?(MIAMI_HEX) }
+            ability = abilities.find { |a| a.hexes.include?(miami_hex_id) }
```
ollybh: DRY: add a private method for finding this ability, and call that method here and in `can_place_miami_token?`.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -34,14 +34,16 @@ def process_choose(action)
...
 
             hex.tile.icons.reject! { |icon| icon.name == 'FECR_key_west' }
             hex.assign!(corporation)
             ability.use!
             corporation.spend(ability.cost, @game.bank)
+            corporation.key_west_placed = true
```
ollybh: Does this work? There isn't a `key_west_placed` property for Corporation objects. If it does work, don't do that. I'd prefer a `key_west_tokened?` method that tests whether the Miami hex has any assignments.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -50,11 +52,14 @@ def skip!
           end
 
           def can_place_miami_token?(entity)
+            return false unless entity.corporation?
+            return false if entity.key_west_placed
             return false if entity.cash < 100
```
ollybh: This should be checking against the ability cost instead of hardcoding the $100 cost.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -50,11 +52,14 @@ def skip!
...
             return false unless (3..6).cover?(@game.phase.name.to_i)
```
ollybh: I much prefer using phase statuses to checks like this. Add `key_west` to the status arrays in the phase definitions for phases 3, 4, 5 and 6, and here you test `@game.phase.status.include?('key_west')`.

I think this is better for two reasons: it gets rid of the hard coded phase numbers here, and it also means that the times when this token can be placed will be shown on the game info page.

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+require_relative '../../../action/par'
+require_relative '../../../action/buy_company'
+
+module Engine
+  module Game
+    module G1832
+      module Step
+        class Exchange < Engine::Step::Exchange
+          def can_gain?(entity, bundle, exchange: false)
+            if exchange && entity == @game.london_company&.owner
```
ollybh: You don't need the safe navigation operator here. Do you?
```suggestion
            if exchange && entity == @game.london_company.owner
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+              return false unless @game.p4_invested_in.nil?
```
ollybh:
```suggestion
              return false if @game.p4_invested_in
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+              @game.log << "#{company.name} invests in #{bundle.corporation.name}; closes after first dividend"
+            else
+              company.close!
+            end
+          end
+
+          private
+
+          def bought?
+            return false unless @round.respond_to?(:current_actions)
```
ollybh:
```suggestion
            return false unless @round.stock?
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+
+            @round.current_actions.any? do |x|
+              x.is_a?(Action::BuyShares) || x.is_a?(Action::Par) || x.is_a?(Action::BuyCompany)
+            end
```
ollybh:
```suggestion
            @round.current_actions.any? { |action| BuySellParShares::PURCHASE_ACTIONS.include?(action.class) }
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+            super
+          end
+
+          def can_exchange?(entity, bundle = nil)
+            return super unless entity == @game.london_company
+
+            return false unless entity.owner == @round.current_entity
+            return false unless @game.p4_invested_in.nil?
+            return false if bought?
+            return false if @game.sell_queue.any?
```
ollybh:
```suggestion
            return false unless @game.sell_queue.empty?
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+              return false if bundle.corporation.operated?
+            end
+            super
+          end
+
+          def can_exchange?(entity, bundle = nil)
+            return super unless entity == @game.london_company
+
+            return false unless entity.owner == @round.current_entity
+            return false unless @game.p4_invested_in.nil?
```
ollybh:
```suggestion
            return false if @game.p4_invested_in
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+              end
+            end
+          end
+
+          # P4 (London Investment) does not close immediately on exchange.
+          # It stays open until the invested corporation pays its first dividend.
+          def process_buy_shares(action)
+            company = action.entity
+            bundle = action.bundle
+            raise GameError, "Cannot exchange #{company.id} for #{bundle.corporation.id}" unless can_exchange?(company, bundle)
```
ollybh, line 55: Should you be passing `exchange: true` to `can_exchange?`?

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+
+            ability = @game.abilities(company, :exchange)
+            buy_shares(company.owner, bundle, exchange: company)
+            ability&.use!
+            @round.current_actions << action if @round.respond_to?(:current_actions)
+            @round.players_history[company.owner][bundle.corporation] << action if @round.respond_to?(:players_history)
```
ollybh:
```suggestion
            if @round.stock?
              @round.current_actions << action
              @round.players_history[company.owner][bundle.corporation] << action(:players_history)
            end
```

lib/engine/game/g_1832/step/exchange.rb
```diff
@@ -0,0 +1,84 @@
...
+
+            if company == @game.london_company
+              @game.p4_invested_in = bundle.corporation
+              @game.log << "#{company.name} invests in #{bundle.corporation.name}; closes after first dividend"
+            else
+              company.close!
```
ollybh: This code path cannot be reached, right? P4 has the only exchange ability, so you shouldn't need this branch.

lib/engine/game/g_1832/game.rb
ollybh: BTW, `can_buy_companies` is the label used in other games for when private companies can be bought in.

ollybh (changes requested):
At some point you're going to need to implement a class for systems. Don't include that in this PR, but drop the Corporation subclass that might return true for `system?`.

I've made a few other suggestions for changes too.

lib/engine/game/g_1832/step/dividend.rb
jemiahlee: I'm not 100% sure that I understand the subtlety that you are exposing. There is no requirement that the share be held for any length of time after the exchange, so the player who does the exchange need not receive the dividend.

If this is what you are asking about, I confirmed this interpretation with Bill Dixon.

jemiahlee:
Thanks so much for your thorough review @ollybh. Will make the changes you've suggested and re-push.

lib/engine/game/g_1832/corporation.rb
jemiahlee: Will do!

lib/engine/game/g_1832/step/dividend.rb
jemiahlee: And, FTR, I originally thought it was going to need to be held until it paid the dividend to the exchanging player. That complication led me to ask Bill about it to make sure. :-D

lib/engine/game/g_1832/step/token.rb
jemiahlee: I missed a change to Corporation for some properties when splitting up the diffs.

I'll revisit.

lib/engine/game/g_1832/step/dividend.rb
ollybh: No, I'm talking about when the private company closes. It seems a little bit strange to me that withholding would leave the private open and I wanted to make sure that's the correct interpretation.

lib/engine/game/g_1832/step/exchange.rb
jemiahlee: I think this code path is only called for P4, and `can_exchange` doesn't need it as an argument because the exchange is implicit in the codepath. `can_exchange` on the other hand, does pass `exchange: true` along to the `can_gain?` method so that it can differentiate from the P4 exchange and every other buy action. So I think this is correct, but I'm open to being wrong. :)

lib/engine/game/g_1832/step/dividend.rb
jemiahlee: I believe this to be correct. Section 9.1 defines the dividend types, and withhold is not considered to be a dividend type in 9.1.3. I think that with the "theme" of P4, it would make sense that the investors only know that you screwed them once the dividend had actually been distributed and they did not receive any of it.

jemiahlee: I confirmed this with Bill. If a company withheld repeatedly, the private could theoretically just close in phase 5 with the other privates. One possible weirdness he pointed out is that if the company withholds and then gets merged or taken over, the "first payout" property would transfer to the new company. But that is a problem for the next PR. :)

jemiahlee:
I cleaned up per the review, and I also realized that the miami behavior worked, but not if the auto-router was used. So I made some additional modifications to only set `@game.miami_first_run` once the dividend step is done.

I also removed the `system?` commit to not conflate the changes at all, esp since I will review the 1828 system code. I'm not sure how much of that will be necessary; the systems for this game are simpler than the systems for 1828.

ollybh:
@jemiahlee Please don't force push to a PR branch once a review is underway. It makes it difficult/impossible for the reviewer to see the changes that have been made in response to review comments. Any changes should be added as new commits and a normal (non-forced) push made to the branch.

jemiahlee:
👍

lib/engine/game/g_1832/step/token.rb
jemiahlee: this went away in favor of your suggestion.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -8,7 +8,7 @@ module G1832
       module Step
         class Token < Engine::Step::Token
           def actions(entity)
-            return super unless can_place_miami_token?(entity)
+            return super unless miami_token_ability(entity)
```
ollybh: You've taken out the check that the corporation can afford to pay for the token. That should probably be checked, otherwise the player might be offered the choice to place a token but will not be able to do this.

lib/engine/game/g_1832/step/token.rb
```diff
@@ -32,29 +32,27 @@ def process_choose(action)
...
 
           def place_miami_token(corporation)
-            return unless can_place_miami_token?(corporation)
-
-            hex = @game.hex_by_id(MIAMI_HEX)
-            abilities = Array(@game.abilities(corporation, :assign_hexes))
-            ability = abilities.find { |a| a.hexes.includes?(MIAMI_HEX) }
+            ability = miami_token_ability(corporation)
+            return unless ability
+            return unless corporation.cash >= ability.cost
```
ollybh: Raising an error would be better than silently failing to place the token.

lib/engine/game/g_1832/step/exchange.rb
ollybh: Yeah, I was reading this as `can_gain?` not `can_exchange?`.

ollybh (changes requested):
Thanks for making these changes. This is close to being ready to merge.

A couple of things that could still be improved:
- Add a `miami_hex` method to the Game class, this should store and return a reference to the Hex object for Miami. Call that from the various places you are currently looking at MIAMI_HEX.
- I found the `miami_first_run` attribute name confusing. It could be taken to mean that the first run has already taken place, not that the first run has yet to happen. Though I'm not sure what would be better. `miami_visited`?

I've added some other comments to the code.

jemiahlee:
@ollybh - ready for re-review, I think I've addressed everything.

lib/engine/game/g_1832/game.rb
```diff
@@ -359,10 +367,19 @@ def revenue_for(route, stops)
...
-          @fecr_corp ||= corporation_by_id(MIAMI_HEX_COMPANY)
+          miami_hex.assigned?(fecr_corp)
+        end
+
+        def place_miami_token
+          miami_hex.tile.icons.reject! { |icon| icon.name == 'FECR_key_west' }
+          miami_hex.assign!(fecr_corp)
+        end
+
+        def event_remove_key_west_token!
```
jemiahlee, line 378: This was something that I realized had not been put in and is related enough that I think it still made sense to include.

lib/engine/game/g_1832/game.rb
```diff
@@ -300,24 +307,81 @@ def coal_hex
...
+            stop.hex.assigned?('P3')
           end
 
-          revenue += (route.corporation.assigned?('P3') ? 20 : 10) if stops.any? { |stop| stop.hex.assigned?('P3') }
+          found ? revenue : 0
+        end
+
+        # Miami first-run rule: worth $0 the first time any corporation runs there prior to phase 5
+        def miami_scores_zero?
+          phase.status.include?('first_miami_run_is_zero') && @miami_first_run
```
ollybh: Variable name hasn't been updated here.
```suggestion
          phase.status.include?('first_miami_run_is_zero') && !@miami_has_been_run
```

lib/engine/game/g_1832/game.rb
```diff
@@ -300,24 +307,81 @@ def coal_hex
...
+        end
+
+        def miami_revenue(route, stops)
+          revenue = 0
+
+          miami_stop = stops.find { |stop| stop.hex.id == MIAMI_HEX_ID }
```
ollybh:
```suggestion
          miami_stop = stops.find { |stop| stop.hex == miami_hex }
```

ollybh (commented):
A couple of tiny changes to make.

ollybh (approved):

PR #12656, opened 2026-05-21
[Core] - Add :invisible track type to path traversal and renderer, for use by 2038.

ddaybell:
Adds a new :invisible track type that can be traversed by all train types when building routes. Invisible paths allow hex-to-hex movement without visible track on the map, needed for spaceship-based games where routes are not constrained to laid track.

- Add :invisible to MATCHES_BROAD so broad-gauge trains enter invisible paths
- Add when :invisible case to tracks_match? returning true (connects to all types)
- Add :invisible entry to TRACK and MULTI_PATH in the renderer so that TRACK[:invisible] returns a valid hash (color: none, width: 0) instead of nil, preventing a crash when invisible paths are present on a tile

### DIFF
--- a/assets/app/view/game/part/track.rb
+++ b/assets/app/view/game/part/track.rb
@@ -40,6 +40,11 @@ module View
             width: 6,
             dash: '9 3',
           },
+          invisible: {
+            color: 'none',
+            width: 0,
+            dash: '0',
+          },
         }.freeze
 
         # width here is added to track width
@@ -87,6 +92,9 @@ module View
           },
           dual: {
             width: 8,
+          },
+          invisible: {
+            width: 0,
           },
         }.freeze
 
--- a/lib/engine/part/path.rb
+++ b/lib/engine/part/path.rb
@@ -10,7 +10,7 @@ module Engine
       attr_accessor :ignore_gauge_walk, :ignore_gauge_compare
 
       LANES = [[1, 0].freeze, [1, 0].freeze].freeze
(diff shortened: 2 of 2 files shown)


ollybh:
Can you add some screenshots to show us this invisible track? :thinking: It would be useful to at least show some of the 2038 tiles before and after this change?

How is this going to work with the route selector? Will the track get shown when a route is selected?

Is all track in 2038 going to be hidden? An alternative might be to leave the track as broad gauge, but have an override for 2038 saying that it will be rendered as transparent (zero alpha channel).

ollybh (changes requested):
Flagging this PR as changes requested, so it is not merged until the last comment has been answered.

ddaybell:
I'm working on this as part of the next PR for 2038.  I don't think there's
a way to completely avoid modifying any core files, but it looks like I can
accomplish the same goal by making a small change to the renderer code
instead of this change, if that's a preferred way to do it.  Once I get it
tested over there, I'll close this PR and push the other one.

ollybh:
> I'm working on this as part of the next PR for 2038. I don't think there's a way to completely avoid modifying any core files, but it looks like I can accomplish the same goal by making a small change to the renderer code instead of this change, if that's a preferred way to do it. Once I get it tested over there, I'll close this PR and push the other one.

OK. I'm going to close this PR for the moment, to try and fight back against the ever-growing PR queue. Ping me if you want it reopened later.

PR #12657, opened 2026-05-21
[core] Move zoom controls above map to prevent obscuring map content

### LINKED ISSUE #12653: [core] - Map zoom buttons overwrite the map coordinates and tiles
**What's your idea?**

Re-locate or re-work the map zoom buttons, so that they do not sit on top of the other map features, such as the coordinates or even map tiles that are implemented in the top-left corner of the map.

1830: 

Image

1870: 

Image

2038 (development):

Image

ddaybell:
Fixes #12653

The zoom control (−, 0, +) was rendered with position: absolute; left: 0; top: 0 inside the scrollable map container, causing it to sit directly on top of the upper-left corner of the map. This obscured the first axis labels (column 1, 2) and any map hexes drawn near the upper-left corner.

This PR lifts the zoom widget out of the scrollable map container and renders it in a thin strip just above the map, flush left. The visual location is nearly identical — still top-left of the map area — but the buttons no longer overlap any map content.

Changes:

assets/app/view/game/map.rb: render MapZoom above the scrollable map div rather than inside it
assets/app/view/game/map_zoom.rb: replace position: absolute with display: inline-block and a 2px bottom margin

image

### DIFF
--- a/assets/app/view/game/map.rb
+++ b/assets/app/view/game/map.rb
@@ -82,7 +82,7 @@ module View
         end
         @hexes.compact!
 
-        children = [render_map, h(MapZoom, map_zoom: map_zoom)]
+        children = [render_map]
 
         if current_entity && @tile_selector
           left = (@tile_selector.x + map_x) * @scale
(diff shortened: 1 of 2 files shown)


ollybh (changes requested):
This is a good idea, but will need to be tested on multiple browsers to ensure it renders correctly.

I've done a little testing and saw this when the browser window is narrow (for example on mobile devices):
Image

Can you please adjust the padding/margins to keep these controls away from anything above them?

ddaybell:
I moved it a bit, and tested it on Chrome, Edge, Android Chrome and Android Samsung browsers.

Chrome:
image

Edge:
image

Android Chrome:
image

Samsung:
image

ollybh (approved):
That looks better.

I've tested this on:
- Chromium and Firefox on Linux.
- Brave, Safari and Edge on iOS.
- Opera, Firefox and Chrome on Android.
These all look OK now.

PR #12661, opened 2026-05-22
[18ESP] Reduce excessive graph calculations - wiring

chrstas:
Related to #12579
Related to #12571

Includes a migration script for existing active games (`scripts/migrate_18esp_destination_connection.rb`).

## Before clicking "Create"

### Explanation of Change

Sequel to PR #12620. Wires `destination_connection` into 18ESP so destination checks replay from the action log instead of recomputing the graph on every load.

**Two emission points** (unique to 18ESP — other games use one):

| Point | When | Why |
|---|---|---|
| CDC step | OR-start | Catches connections formed between ORs (e.g. token removal in SR) |
| `Track#auto_actions` | After tile lay | Destination bonus is immediate cash — must be available before token purchases in the same step |

Both use `new_destination_connection?` and only emit on actual new connections.

**Removed direct calls** to `check_for_destination_connection` from: `home_token.rb`, `operating.rb` (deleted empty subclass), `special_track.rb` (mine hexes can't complete destinations — documented), `tracker.rb` (legacy bridge removed).

**Performance guards** use `@loading` only — `@strict` (`validate_auto_actions`) intentionally runs full graph validation. CDC's `blocking?` checks `@loading || @strict` (flow control, not performance).

**MountainPass fix:** Moved `@graph.clear` from query method (`opening_new_mountain_pass`) to mutation method (`open_mountain_pass`) — queries shouldn't have side effects.

### Tests

Three-layer testing:
1. **`corporation_spec.rb`** — `goal_reached!` idempotency
2. **`game_spec.rb`** — Full fixture replay: no exceptions, exactly 8 destination logs (one per corp), frozen end-state per corporation
3. **`check_destination_connection_spec.rb`** + **`track_spec.rb`** — Guard logic across all contexts (`@loading`, `@strict`, live play)

### Performance

Profiled with `stackprof` (CPU mode, 10µs interval) on the full `18ESP_game_end_second_eight` fixture (1335 actions).

The engine now has two loading modes:
- **`strict: false`** (normal game load) — replays destination connections from the log, skips graph computations
- **`strict: true`** (`validate_auto_actions`) — fully validates all auto actions including graph computations

| Measurement | Mode | Samples | GC | GC % |
|---|---|---|---|---|
| **Current** | `strict: false` | 1286 | 187 | 14.54% |
| **Current** | `strict: true` | 1467 | 170 | 11.59% |

**Summary:** The `strict: true` path is intentionally ~14% slower because it fully validates all destination connections via graph computation — this is the correct trade-off (validation thoroughness over load speed). `strict: true` only runs during `validate_auto_actions`, not on every page load.

<details>
<summary>Full stackprof output — strict: false (1286 samples)</summary>

```
==================================
  Mode: cpu(10)
  Samples: 1286 (0.00% miss rate)
  GC: 187 (14.54%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       399  (31.0%)         139  (10.8%)     Engine::Part::Path#walk
       131  (10.2%)         131  (10.2%)     (marking)
       468  (36.4%)          56   (4.4%)     Engine::Graph#compute
       373  (29.0%)          52   (4.0%)     Engine::Part::Node#walk
        56   (4.4%)          56   (4.4%)     (sweeping)
       803  (62.4%)          43   (3.3%)     Array#each
        77   (6.0%)          42   (3.3%)     Engine::Hex#paths
        40   (3.1%)          40   (3.1%)     Kernel#hash
       803  (62.4%)          17   (1.3%)     Class#new
        12   (0.9%)          12   (0.9%)     IO#write
        12   (0.9%)          12   (0.9%)     Hash#initialize
        11   (0.9%)          11   (0.9%)     Array#flatten
        11   (0.9%)          11   (0.9%)     Engine::Part::Path#exits
        12   (0.9%)          10   (0.8%)     Engine::Part::City#tokened_by?
        10   (0.8%)          10   (0.8%)     Engine::Part::Base#hex
         9   (0.7%)           9   (0.7%)     String#split
        19   (1.5%)           9   (0.7%)     Array#select
        18   (1.4%)           9   (0.7%)     Array#sum
         8   (0.6%)           8   (0.6%)     Hash#delete
         9   (0.7%)           8   (0.6%)     Engine::Helper::Type#type_s
         8   (0.6%)           7   (0.5%)     Engine::Part::Path#ignore?
         7   (0.5%)           7   (0.5%)     Integer#==
         7   (0.5%)           7   (0.5%)     Time#strftime
       600  (46.7%)           7   (0.5%)     Engine::Round::Base#process_action
         7   (0.5%)           7   (0.5%)     Engine::Part::Path#tracks_match?
        13   (1.0%)           7   (0.5%)     Enumerable#flat_map
        17   (1.3%)           7   (0.5%)     Engine::ShareBundle#initialize
         6   (0.5%)           6   (0.5%)     String#gsub
         5   (0.4%)           5   (0.4%)     Array#==
         5   (0.4%)           5   (0.4%)     Numeric#positive?
```
</details>

<details>
<summary>Full stackprof output — strict: true (1467 samples)</summary>

```
==================================
  Mode: cpu(10)
  Samples: 1467 (0.00% miss rate)
  GC: 170 (11.59%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       551  (37.6%)         187  (12.7%)     Engine::Part::Path#walk
       637  (43.4%)          73   (5.0%)     Engine::Graph#compute
       113  (7.7%)          113   (7.7%)     (marking)
       512  (34.9%)          54   (3.7%)     Engine::Part::Node#walk
       960  (65.4%)          52   (3.5%)     Array#each
        57   (3.9%)          57   (3.9%)     (sweeping)
       105  (7.2%)          55   (3.7%)     Engine::Hex#paths
        42   (2.9%)          42   (2.9%)     Kernel#hash
       960  (65.4%)          20   (1.4%)     Class#new
        16   (1.1%)          16   (1.1%)     Hash#initialize
        14   (1.0%)          14   (1.0%)     Engine::Part::Path#exits
        14   (1.0%)          14   (1.0%)     Engine::Part::Base#hex
        15   (1.0%)          12   (0.8%)     Engine::Part::City#tokened_by?
        11   (0.7%)          11   (0.7%)     Array#flatten
        10   (0.7%)          10   (0.7%)     Integer#==
        10   (0.7%)          10   (0.7%)     IO#write
        27   (1.8%)          10   (0.7%)     Array#sum
        10   (0.7%)          10   (0.7%)     String#split
        21   (1.4%)           9   (0.6%)     Engine::Corporation#all_abilities
         9   (0.6%)           9   (0.6%)     Hash#delete
        26   (1.8%)           9   (0.6%)     Array#select
         8   (0.5%)           8   (0.5%)     Engine::Part::Path#tracks_match?
        14   (1.0%)           8   (0.5%)     Enumerable#flat_map
         8   (0.5%)           8   (0.5%)     Engine::Part::Path#ignore?
       740  (50.4%)           8   (0.5%)     Engine::Round::Base#process_action
         7   (0.5%)           7   (0.5%)     Numeric#positive?
        10   (0.7%)           7   (0.5%)     Engine::Part::Path#lane_match?
         7   (0.5%)           7   (0.5%)     Time#strftime
         7   (0.5%)           7   (0.5%)     String#gsub
       301  (20.5%)           7   (0.5%)     Engine::Round::Base#skip_steps
```
</details>

### Screenshots

N/A — backend-only change, no UI impact.

### Any Assumptions / Hacks

- `special_track.rb` intentionally does **not** emit `destination_connection` — mine hexes can't complete destinations. Code comment documents when this would need to change.
- `log_skip` overridden as empty in CDC to suppress per-corporation skip noise.


chrstas:
**Note:** This PR currently includes the commit from PR [#12620](https://github.com/tobymao/18xx/pull/12620) (the prework PR) because GitHub cannot set a fork branch as the base for a PR - as far as I understand it - but please correct me if I am wrong.

Once [#12620](https://github.com/tobymao/18xx/pull/12620) is merged, I'll rebase this branch onto master — the duplicate commits will drop off and only the 2 wiring-specific commits will remain ([d53cc69](https://github.com/tobymao/18xx/pull/12661/commits/d53cc69bc85ea1a3fd2f842df7656dbe0e847c0c) and [b5532a8](https://github.com/tobymao/18xx/pull/12661/commits/b5532a873a1746058798547eab3dcda2130f681b)).

ollybh:
@chrstas I've merged PR #12620. Can you update this PR to include those changes?

ollybh:
From a very quick look at this PR, there a couple of things that will need to be changed:

- Drop the `@game.replaying?` and just test `@game.loading`. It looks like you've done this to avoid calculating the game graph is `@strict` is true, but that's the point of `@strict`, it attempts to fully validate the game history.
- You've got code that will attempt to load games both with and without the destination connection actions. We don't attempt to parse older versions of games, instead we would either pin games to the older version of code or (the better option) attempt to write a migration script that inserts destination connection actions into the action log for existing games.

chrstas:
Thanks for the intput @ollybh... I have now both points addressed:

**`replaying?` removed** — guards now use `@game.loading` directly so `@strict` runs full graph validation. 

**Legacy bridge removed** — fixture replaced with migrated version. Added `scripts/migrate_18esp_destination_connection.rb` for active server games.

I have also updated the description of the PR accordingly.

ollybh:
> Full stackprof output — strict: false (1286 samples)
> ```
> ==================================
>   Mode: cpu(10)
>   Samples: 1286 (0.00% miss rate)
>   GC: 187 (14.54%)
> ==================================
>      TOTAL    (pct)     SAMPLES    (pct)     FRAME
>        399  (31.0%)         139  (10.8%)     Engine::Part::Path#walk
>        131  (10.2%)         131  (10.2%)     (marking)
>        468  (36.4%)          56   (4.4%)     Engine::Graph#compute
>        373  (29.0%)          52   (4.0%)     Engine::Part::Node#walk
> ```

This is better than before, but there's clearly still graph recalculation going on when games are being loaded.

@chrstas You next challenge is to eliminate the call to `@graph.connected_hexes` in `Game::opening_new_mountain_pass` when the game is being loaded. This one should be much more straightforward as there's already an action in the action log for when this is done.

lib/engine/game/g_18_esp/step/check_destination_connection.rb
```diff
@@ -27,21 +34,20 @@ def description
             'Check destination connection'
           end
 
-          # Step is not player-passable; @passed is set in process_destination_connection
+          def blocking?
+            return false if @game.loading || @game.strict
```
ollybh: Why do you need to check `@game.strict` here? Is it ever true when `@game.loading` is false?

scripts/migrate_18esp_destination_connection.rb
```diff
@@ -0,0 +1,212 @@
...
+
+    a['type'] == 'destination_connection' ||
+      Array(a['auto_actions']).any? { |sub| sub&.dig('type') == 'destination_connection' }
+  end
+end
+
+def inject_destination_connections(data)
+  return nil if already_migrated?(data)
+
+  game = Engine::Game.load(data, actions: [], strict: false)
```
ollybh, line 81: I'm attempting to test this migration script on a copy of the site database but it's failing here:
```
Game 251610
/18xx/lib/engine/game/base.rb:3607:in `check_player_range!': 1804 does not support 0 players (Engine::GameError)
	from /18xx/lib/engine/game/base.rb:603:in `initialize'
	from /18xx/lib/engine/game/base.rb:79:in `new'
	from /18xx/lib/engine/game/base.rb:79:in `load'
	from scripts/migrate_18esp_destination_connection.rb:55:in `inject_destination_connections'
	from scripts/migrate_18esp_destination_connection.rb:143:in `migrate_data'
	from scripts/migrate_18esp_destination_connection.rb:167:in `migrate_db_actions'
	from (irb):36:in `<main>'
	from /usr/local/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
	from /usr/local/bin/irb:25:in `load'
	from /usr/local/bin/irb:25:in `<main>'
```
It looks like this isn't being passed either the title or players that have been loaded. Not sure why, they appear to both be in the `data` hash that's passed to this method.

ollybh (approved):
The migration script is failing for me. Have you tested this against games in the database? This can be done by using the `import_game` script to load copy games off the site database and then running the migration script.

And can you explain the `strict` bits in the game code? I don't see what that is doing.

scripts/migrate_18esp_destination_connection.rb
chrstas: I will have a look and dig deeper - had tested it against the jsons from prod but the `import_game` is another way which I did not consider... first time with the migration script. thanks for the hint

lib/engine/game/g_18_esp/step/check_destination_connection.rb
chrstas: I had the same thought first when I implemented it but `@game.strict` can be true when `@game.loading` is false — they're mutually exclusive. `initialize_actions` sets `@loading = true unless @strict`, so in strict mode `@loading` stays false.

Without the check, strict mode would still trigger `connected_hexes` through the blocking path. Since strict mode replays actions similarly to loading, it seemed right to include it here as well. Happy to remove it if you see a reason it's not needed though...

scripts/migrate_18esp_destination_connection.rb
chrstas: Good catch, thanks! That was my mistake — `Game#to_h` returns symbol keys, but `Engine::Game.load` expects string keys. I missed that when writing the migration script. your hint regarding the` import_game` was key for me... thanks

chrstas:
The mountain pass optimization turned out to be a challenge — definitely not as straightforward as destination_connection, but satisfying to get right. And this whole performance topic brought some learnings... @ollybh thanks for the hints and patience.

The core problem is that `opening_new_mountain_pass` calls `@graph.connected_hexes(entity)` to determine which passes are reachable, and this affects Track's action set, which determines blocking behavior, which determines how `pass` actions are routed between steps. 

You can't just return all unopened passes during loading (like the naive approach for CDC) because that makes Track blocking for entities that never open a pass — and if those entities also can't lay tiles or place tokens, their `pass` action falls through to Route and crashes. The solution pre-scans the action log before replay, counting `choose`/`choose_ability` actions per entity that target mountain pass hexes. 

During loading, Track's `actions` method uses this pre-scanned data instead of the graph — but only when Track is already blocking for `lay_tile` or `place_token` (which is the case for ~99% of entities). When Track has no other actions, it falls back to the real graph computation to preserve correct action routing. 

This way the expensive graph call is eliminated for nearly all entities during loading, while the rare edge case (entity with mountain pass access but no tile/token actions) stays correct.

Strackprof now for` strict: false`:


```
==================================
  Mode: cpu(10)
  Samples: 901 (0.11% miss rate)
  GC: 165 (18.31%)
==================================

     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       114  (12.7%)         114  (12.7%)     (marking)
        51   (5.7%)          51   (5.7%)     (sweeping)
        35   (3.9%)          35   (3.9%)     Array#flatten
       682  (75.7%)          23   (2.6%)     Array#each
        93  (10.3%)          22   (2.4%)     Engine::Part::Path#walk
        21   (2.3%)          21   (2.3%)     String#split
       117  (13.0%)          19   (2.1%)     Engine::Graph#compute
        17   (1.9%)          17   (1.9%)     String#gsub
       736  (81.7%)          16   (1.8%)     Class#new
        41   (4.6%)          15   (1.7%)     Array#select
        29   (3.2%)          14   (1.6%)     Engine::Corporation#all_abilities
        13   (1.4%)          13   (1.4%)     Array#uniq
        98  (10.9%)          12   (1.3%)     Array#map
       225  (25.0%)          11   (1.2%)     Engine::Round::Base#skip_steps
        24   (2.7%)          11   (1.2%)     Engine::Action::Base#to_h
       576  (63.9%)          11   (1.2%)     Engine::Round::Base#process_action
        13   (1.4%)          10   (1.1%)     Enumerable#sort_by
       113  (12.5%)           9   (1.0%)     Engine::Round::Base#active_step
        17   (1.9%)           9   (1.0%)     Engine::Depot#other_trains
        18   (2.0%)           9   (1.0%)     Engine::Part::City#tokened_by?
        25   (2.8%)           8   (0.9%)     Engine::Hex#paths
        16   (1.8%)           8   (0.9%)     Array#to_h
       144  (16.0%)           8   (0.9%)     Array#any?
         8   (0.9%)           8   (0.9%)     Engine::Passer#active?
        10   (1.1%)           8   (0.9%)     Array#inspect
         7   (0.8%)           7   (0.8%)     Hash#initialize
         9   (1.0%)           7   (0.8%)     Engine::Tile#separate_parts
         8   (0.9%)           6   (0.7%)     Kernel#Array
        17   (1.9%)           6   (0.7%)     Engine::Step::BuySellParShares#can_sell?
        19   (2.1%)           6   (0.7%)     Array#reject

File: scripts/migrate_18esp_destination_connection.rb [hunk 3]
```diff
@@ -0,0 +1,212 @@
...
+  end
+
+  outfile = filename.sub(/\.json$/, '_migrated.json')
+  File.write(outfile, JSON.generate(migrated))
+  puts "  Written: #{outfile}"
+end
+
+def migrate_db_actions(game_record, dry_run: false)
+  puts "\nGame #{game_record.id}"
+  data     = JSON.parse(JSON.generate(game_record.to_h(include_actions: true)))
```
ollybh, line 239: It's not important for a script that will only be run once, but a more explicit way to ensure the hash has string keys would be:
```suggestion
  data     = game_record.to_h(include_actions: true)).transform_keys(&:to_s)
```

ollybh:
That's fixed the migration script when running with the games in the database ­– thanks.

I've not looked at the changes for the mountain passes yet, but they're causing problems when trying to migrate the games. When run against a copy of the site database from a fortnight ago, there are 41 migration failures when the mountain pass changes are included, compared to just five with just the destination connection changes.

Can you revert the last commit (the mountain passes) from this PR and put it through as a separate pull request?

Here are the outputs of the migration script:
- [migrate_18esp_destination_connection.txt](https://gist.github.com/ollybh/f46d609189529f8a388dae44188f7cbf#file-migrate_18esp_destination_connection)
- [migrate_18esp_destination_connection-mountain_pass.txt](https://gist.github.com/ollybh/f46d609189529f8a388dae44188f7cbf#file-migrate_18esp_destination_connection-mountain_pass-txt)

The five games that failed migration are IDs 215752, 219379, 224358, 234070 and 248071.

lib/engine/game/g_18_esp/game.rb
```diff
@@ -1356,6 +1378,18 @@ def other_bank_info
...
+        private
+
+        def precompute_mountain_pass_hints(actions)
+          @loading_mp_remaining = Hash.new(0)
+          actions.each do |action|
+            next unless %w[choose choose_ability].include?(action['type'])
+            next unless MOUNTAIN_PASS_TOKEN_HEXES.include?(action['choice'])
+
+            @loading_mp_remaining[action['entity']] += 1
+          end
```
ollybh: `Enumerable.tally` is one of the many obscure built-in methods that can be used for calculations like this.
```suggestion
          @loading_mp_remaining = actions
            .filter_map do |action|
              next false unless %w[choose choose_ability].include?(action['type'])
              next false MOUNTAIN_PASS_TOKEN_HEXES.include?(action['choice'])

              action['entity']
          end
            .tally
```

chrstas: will take this hint for the subsequent PR

scripts/migrate_18esp_destination_connection.rb
chrstas: habe adjusted it - thanks

chrstas:
Reverted the mountain pass loading optimization (last commit). I'll set it up as a separate PR — makes sense to keep it isolated. Will also dig deeper into the error messages provided in your comment... thanks

Also incorporated your script feedback. Thanks for the review!

scripts/migrate_18esp_destination_connection.rb
ollybh: Ugh, that was a stupid suggestion. The script is failing again now. There are probably nested keys that need their keys converting too.

lib/engine/game/g_18_esp/step/check_destination_connection.rb
ollybh: Yeah, of course. I was getting myself confused because strict is only set when a game is being loaded, but as you said, it causes `@game.loading` to be false.

You probably shouldn't be checking strict here. The idea behind strict is that we optimise things when loading is set, but if strict is set then we try and validate the actions (though this isn't done completely as there's a bunch of logic embedded in the view classes).

scripts/migrate_18esp_destination_connection.rb
chrstas: I reverted now the last commit with the suggestion

lib/engine/game/g_18_esp/step/check_destination_connection.rb
chrstas: Thanks @ollybh  — removed the `@strict `check from `blocking?`. Tests pass cleanly.

I'll admit I was confused here — I could have sworn this caused issues during earlier development, which is why I added the `@strict` guard in the first place. Looking at it now, I think the problem was with the legacy bridge: old saves had no `destination_connection` actions in the log, so in strict mode CDC would block without a matching action to process. Now that the legacy path is gone and the migration script ensures all saves have the actions in the log, the guard is no longer needed.

Your reasoning makes sense: `@loading` is the performance shortcut, `@strict` is meant to validate everything. Skipping the blocking check in strict mode undermines that purpose.

ollybh:
> The five games that failed migration are IDs 215752, 219379, 224358, 234070 and 248071.

@chrstas Have you been able to look into these games that failed migration?

chrstas:
> > The five games that failed migration are IDs 215752, 219379, 224358, 234070 and 248071.
> 
> @chrstas Have you been able to look into these games that failed migration?

this week has been a bit crazy... will try to do it tomorrow and report back. thanks for your patience

chrstas:
@ollybh  Update on the five failing migration games (215752, 219379, 224358, 234070, 248071):

**Fixes applied to the migration script:**

1. **Auto_action ordering** — DC auto_actions were appended after existing auto_actions (like `pass`). The `pass` was processed first, advancing step state before the DC could be handled. Fixed by inserting the DC at the beginning of the auto_actions array.

2. **Cross-corp detection** — Connections established by another corporation's tile lay are now tracked as "pending" and emitted as standalone actions at the start of the connected corporation's next operating turn.

**Status:**

- **4 of 5 games pass** both strict validation and local game load: 215752, 219379, 224358, 234070.
- **active games** current active games 255036, 251610, 253631, 245301, 245725,245079,254665 pass both as well
- **Game 248071** passes strict validation but fails on local game load. The issue: AVT's DC releases a blocked token via `goal_reached!`, but during loading the step cache evaluates token availability before the release is visible, causing Track to skip itself. This is specific to games where a DC releases a token immediately needed for the next `place_token` in the same turn.

First I thought about pinning as option for 248071, but I suspected an underlying issue which could also appear in new games. First tests confirm my suspicion...

Derived conclusion: the PR changes when `goal_reached!` is called — from inline (during `lay_tile_action` / `place_home_token`) to via auto_action (`process_destination_connection`). This shifts the timing relative to step evaluation. I added (locally, not committed) a test (`track_spec.rb`: "offers place_token after goal_reached! releases a blocked token") that shows Track returns `["lay_tile", "pass"]` instead of including `place_token` after a blocked token is released — with `@game.loading = false`, so this could affect live play too. Please give me 1-2 more days. I will chase the remaining issue.

chrstas:
@ollybh Update on game 248071 — found and fixed the root cause.

The bug: when `goal_reached!(:destination)` releases a blocked token, `Graph#can_token?` returns a stale cached `false`. During loading, `skip_steps` runs after the lay_tile but before CDC processes the standalone DC — at that point the token is still blocked, so Track has no actions and gets passed. When CDC then fires and releases the token, it's too late.

**Fix (3 lines of production code):**

1. `clear_graph_for_entity` after `goal_reached!` in both `Track#process_destination_connection` and `CDC#process_destination_connection` — cache invalidation so the released token is visible to `can_token?`
2. CDC resets Track's `@passed` flag after processing a standalone DC — undoes the premature skip. Guarded by `!@tokened` so it doesn't interfere when a token was already placed that turn.

Game 248071 turned out to be the perfect edge case for finding this — a corp that connects its destination mid-turn where the released token is the only one available and there's no second tile lay to keep Track blocking. I've added it as a fixture (`18ESP_destination_token_release`) with an integration test covering the full flow.

No performance impact — `clear_graph_for_entity` doesn't appear in StackProf top frames. Profiled on both fixtures, the `strict:false` → `strict:true` delta on `Graph#compute` (+30–35%) is unchanged.

All five originally failing games now pass strict validation and local game load.

ollybh (approved):
@chrstas Apologies for neglecting this PR. I've taken a look at the latest changes, and everything looks good.

Running the migration script against a more recent copy of the production database was successful for everything – thanks for putting this together.

I'll try get this merged, deployed and migrated in the next few days.

lib/engine/game/g_18_esp/game.rb
```diff
@@ -17,7 +17,7 @@ class Game < Game::Base
         include CitiesPlusTownsRouteDistanceStr
         include DoubleSidedTiles
 
-        attr_reader :can_acquire_minors
+        attr_reader :can_acquire_minors, :strict
```
ollybh: This accessor is no longer used.
```suggestion
        attr_reader :can_acquire_minors
```

chrstas: have removed it - thanks

chrstas:
> @chrstas Apologies for neglecting this PR. I've taken a look at the latest changes, and everything looks good.
> 
> Running the migration script against a more recent copy of the production database was successful for everything – thanks for putting this together.
> 
> I'll try get this merged, deployed and migrated in the next few days.

@ollybh thanks a lot and don't worry - I know there are a lot of PRs. Appreciate your efforts....

ollybh:
This fix has been rolled out and the games migrated. I've loaded a few on the server and everything looks good.

Thanks for this performance fix. Have you still got the code for the other half – the graph recalculation when checking for mountain passes that can be opened?

chrstas:
Hi @ollybh, have it on my list - wanted to do some additional testing. Will be traveling until 18th and then push the PR. 

Thanks to you for the efforts.

PR #12662, opened 2026-05-22
[Core] Fall-back handling of excessive discounts

perwestling:
During development of an 1880 Romania variant I encountered a case where a log appeared with negative payment.
This turned out to be a case where discounts exceeded cost.

## Before clicking "Create"

This should not break any game, but in case it does I believe games are broken. So pin or archive if any games do appear broken due to this bug fix.

## Implementation Notes

I chose to make the change in the pay method. Possibly one could have checked the callers but this catches any faulty cost.

### Explanation of Change

Negative costs should not exists. They should be a bonus if they are, and also should not be logged as negative pay.

### Screenshots

In the hot seat game I was trying the TR corporation laid two yellow tiles.
First on a hex with 40 terrain (mountain), and then on a hex with 10 terrain (hill).
TR do have 10 + 20 discount for mountain terrain.

Here is the log before the fix:

image

As can be seen discount is 30. When cost is 40 TR get to pay 10. But when cost is 10 TR
pays -20, in other words receives 20 money.

And here is the same log after the fix ((image below is after update after review comment)):

image

Discount is automatically adjusted to avoid negative costs.

### Any Assumptions / Hacks

Assumption is that no game uses negative cost to receive money. It looks strange in the log so that should have been reported.

### DIFF
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -240,6 +240,11 @@ module Engine
       end
 
       def pay_tile_cost!(entity_or_entities, tile, rotation, hex, spender, cost, _extra_cost)
+        if cost.negative?
+          cost = 0
+          @log << "Discounts exceed cost for tile lay, so cost reduced to zero for #{spender.name}."
+        end
+
         # entity_or_entities is an array when combining private company abilities
         entities = Array(entity_or_entities)
         entity, *_combo_entities = entities
--- a/lib/engine/step/tracker.rb
+++ b/lib/engine/step/tracker.rb
@@ -240,11 +240,6 @@ module Engine
       end
 
       def pay_tile_cost!(entity_or_entities, tile, rotation, hex, spender, cost, _extra_cost)
-        if cost.negative?
-          cost = 0
-          @log << "Discounts exceed cost for tile lay, so cost reduced to zero for #{spender.name}."
-        end
-
         # entity_or_entities is an array when combining private company abilities
         entities = Array(entity_or_entities)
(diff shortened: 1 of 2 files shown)


lib/engine/step/tracker.rb
```diff
@@ -240,6 +240,11 @@ def update_tile_lists(tile, old_tile)
       end
 
       def pay_tile_cost!(entity_or_entities, tile, rotation, hex, spender, cost, _extra_cost)
+        if cost.negative?
+          cost = 0
+          @log << "Discounts exceed cost for tile lay, so cost reduced to zero for #{spender.name}."
```
crericha: I'm not a fan of yet another log message. Can we detect when we calculate the total discount and use min on tile cost and discount?

perwestling: Have now improved the solution. The adjusted discounts is now shown directly in the log without any notice about adjustment.

perwestling: Have updated the screen shot to the situation after the latest correction.

perwestling: I thought about this - maybe the text in the screen shot above can be confusing? "Receive L10 discount" - why it should be L30?  An alternative text would be "TR receives full discount from TR, P3. Valea Jiului" or something similar. 

Or maybe this is good enough? After all, the player can see the terrain cost on the hex when they lay the tile.

ollybh (approved):
LGTM.

I looked for negative discounts and the only ones I could find were for companies, where the minimum bid/purchase price is higher than face value.

Issue #12403, opened 2026-02-22
add auto-pass on 1817 private bid (and similar games like 1817NA)


PedroRCSequeira:
in the simplest form this would be a tick box we could check and it would pass on the current bid

even better would be a box where we could type our max bid and when it comes to our turn, if there is a value there, it would either pass if the value was reached or min-increase if it had not

this could saveplenty of time in async games (:

ddaybell:
I believe that if we added a box for people to enter a max bid, that information would likely be discoverable by the other players.  Not sure that's what people would desire.  Extending autopass behavior to cover individual private auctions, and to reset once the private has been sold to someone, seems likely doable.  I'll look into this further.

ddaybell:
So 18EU already has a feature very similar to this.  It works on 18EU's minors, but it allows a player to set a max bid for a given minor.  If that minor is then put up for auction, it does minimum increment raises until the max is reached, and then passes.  All of this information is public, and the max bid is announced in the game log.  I question whether players would be willing to use this feature for private auctions, given that it publicly announces their maximum bid.  I could see it having some value in liquidations, particularly for the owner of the liquidated company who is otherwise precluded from raising by more than $10.  If there were a way to keep the max bid a secret, that would add a lot of value.  There's no way to keep this information absolutely secret, since Master Mode and Undo both exist.  But I think there are ways to keep it concealed in normal play. Continuing to investigate.

PedroRCSequeira:
It needs to be hidden, or else I might only be willing to bid 30$, but
since I know your max bid is 60 then i'll be sure to bid all the way to 50$

This is something that exists in a recent implementation of Indonesia, and
it helps move the game along

In that implementation there is also an undo button that any player can use
(it will be written in the log), but naturally if someone "checks" another
player max bid, then undo's, its a big no-no

Also, in that implementation we can choose one of the following options

A) Just bid X (then we will be prompted again if another player raises)

B) Bid X and then min bid up to Y, pass if we are outbid

C) Bid X and then min bid up to Y, ask us again if we are outbid


This last one is useful for 2 things:

1) we can set a safe min bid without thinking too much about the game
state, and then we can evaluate better if needed

2) we can make differentiate our max bid depending on which players are
still active

ddaybell:
[...]

PR #12663, opened 2026-05-23
[core] - Autobid engine updates

ddaybell:
Partially addresses #12403

## Summary

The engine already had a complete autobid mechanism for auction rounds (`ProgrammerAuctionBid`, `Action::ProgramAuctionBid`, `AuctionBid` view) — 18EU was the only game using it, via its `ModifiedDutchAuction` step. This PR extends that mechanism to `Engine::Step::SelectionAuction` so all games that use it inherit auto-bid support with no per-game code required.

- **`Engine::Step::SelectionAuction`** — add `require` + `include ProgrammerAuctionBid`, plus `auto_requires_auctioning?` (deactivates the program when a different company comes up) and `programmable_buy_price? false` (no dutch buy-price for selection auctions)
- **`ProgrammerAuctionBid`** — add `programmable_buy_price?` defaulting to `false`; only dutch-auction steps opt in with `true`
- **`G18EU::Step::ModifiedDutchAuction`** — add `programmable_buy_price? true` to opt in, since 18EU's price-drop mechanic is the only case where a buy price is meaningful
- **`Engine::Step::WaterfallAuction`** — add `auction_participant?(player)` returning false when an auction is active and the player has no existing bid; prevents the autobid panel from showing to ineligible players
- **`Engine::Round::Base#auto_actions`** — fix to call `step.current_entity` instead of `round.current_entity` so auto-bid chains correctly in passable auctions where `@active_bidders` rotates independently of `@entity_index`
- **`AuctionBid` view** — rewrite to use `step.auctioning` as the bid target (removes the dropdown); show contextual messages when no entity is up for auction or the player is not a participant; conditionally show buy price controls and adjust label text based on `programmable_buy_price?`
- **`Auto` view** — add a note that auction auto-actions are available when in an auction round

## Tested

| Game | Result |
|------|--------|
| 18EU | Max bid, buy price, auto-pass, and dutch price drop all verified. Regression check passed. |
| 18MT | Max bid and auto-pass verified. Buy price controls correctly hidden. Non-participating players correctly excluded from waterfall auction panel. |

## Notes

Per-game PRs to follow for games using a custom `SelectionAuction` subclass (~10 lines each). Games using `Engine::Step::SelectionAuction` directly only need `available_programmed_actions` wired up. Known waterfall auction display issues (panel appearing between auctions) are deferred to a follow-up PR.

Presently, only 18EU and for some reason 18MT had the hooks to use autobid.  18EU actually implemented it, 18MT did not but it was wired to do so once we updated the engine.  

Assuming this update is approved, I can start rolling this feature out to other games that would benefit from it.  I propose to start with games where users select assets to be auctioned (e.g. 1817, 18USA) and then move to games like 1861/67 and 1880 that run auctions in a set order, then move to the waterfall auctions.  This feature doesn't make much sense to implement in games that don't use immediate auctions (e.g. 1822, 1846, 18VA). 

## Test plan

- [x ] Start an 18EU game and verify max bid, buy price, and auto-pass all function correctly
- [x ] Start an 18MT game and verify: buy price controls are absent, only bidders see the panel during a waterfall auction, max bid and auto-pass function correctly
- [ x] Verify the Auto tab description updates appropriately for auction vs. non-auction games

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/assets/app/view/game/auto.rb
+++ b/assets/app/view/game/auto.rb
@@ -24,8 +24,11 @@ module View
         children = [
           h(:h2, 'What are Auto Actions?'),
           h(:p, "You can pre-program share purchases and passes during #{@game.stock_round_name}s."),
-          *(@game.available_programmed_actions.include?(Engine::Action::ProgramAuctionBid) ?
-            [h(:p, 'You can pre-program bids during auctions.')] : []),
+          *(if @game.available_programmed_actions.include?(Engine::Action::ProgramAuctionBid)
+              [h(:p, 'You can pre-program bids during auctions.')]
+            else
+              []
+            end),
           h(:p, 'Deactivates when players take actions that may affect you (e.g. shares are sold).'),
           h('p.bold', 'Your Auto Actions are not secret from other players.'),
           h(:p, render_wiki),
--- a/assets/app/view/game/auto.rb
+++ b/assets/app/view/game/auto.rb
@@ -24,6 +24,8 @@ module View
         children = [
           h(:h2, 'What are Auto Actions?'),
           h(:p, "You can pre-program share purchases and passes during #{@game.stock_round_name}s."),
+          *(@game.available_programmed_actions.include?(Engine::Action::ProgramAuctionBid) ?
+            [h(:p, 'You can pre-program bids during auctions.')] : []),
           h(:p, 'Deactivates when players take actions that may affect you (e.g. shares are sold).'),
           h('p.bold', 'Your Auto Actions are not secret from other players.'),
           h(:p, render_wiki),
--- a/assets/app/view/game/auto_action/auction_bid.rb
+++ b/assets/app/view/game/auto_action/auction_bid.rb
@@ -6,16 +6,19 @@ module View
   module Game
     module AutoAction
       class AuctionBid < Base
-        needs :bid_target, store: true, default: nil
-
         def name
           "Auto bid in Auction Round#{' (Enabled)' if @settings}"
         end
 
         def description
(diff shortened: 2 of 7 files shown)


ollybh:
@ddaybell It's not clear from the PR description what this autobid mechanism is going to do. Can you include some screenshots, or a description of the interface that players will see?

ddaybell:
You can see how it currently works in 18EU.  This just optimizes a couple of things as described above so this feature can more easily be extended to other titles.  

The way it works is once an asset (private, minor) goes up for auction, either by being selected or (in a waterfall) when the lowest available item has multiple bids on it, then every player who has a bid on the item can go to the Auto tab and enter a maximum bid they're willing to bid.  They will automatically increase their bid by the minimum bid until the designated max bid is reached.  The player can also choose to pass if their max bid gets outbid.  If they don't choose to pass, then they get their regular turn.  

And if two players both set max bids, then they will automatically bid each other up until one of them hits their max.  This has the potential to speed up async auctions significantly.  It'll probably be most effective in 1817/USA, once I implement that one, on liquidations/acquisitions of corporations.  Sometimes those auctions take a really long time.

A couple of screenshots of what the UI looks like, from 18EU where this all came from, and from 18MT which also already had this feature exposed, but not used.  The 18EU screen has a couple of additional entries, because that game uses a Reverse Dutch auction feature, in which the price can go down if there are no bids on an item.  The extra entries allow people to specify the price at which they'll buy the item if it drops that far.  The 18MT screenshot is for a more typical auction that would only go up in bid values.

18EU:

image

18MT:
image

PR #12667, opened 2026-05-23
[18OE] - Centered Double Towns - Layout and Routing

Witzman:
## Summary

Implements double-town tiles for 18OE and the engine infrastructure to support them.

 Original Tile: 
OE11

18xx tile:
grafik

### Engine changes (backward-compatible, all games)

- **`lib/engine/part/town.rb`** — adds `size:` DSL parameter (`@size`, `attr_reader`). Defaults to `1`; no behavioural change for existing tiles.
- **`lib/engine/tile.rb`** — passes `size: params['size']` to `Town.new`. No effect on any tile that doesn't declare `size:`.
- **`assets/app/view/game/part/town_dot.rb`** — when `town.size == 2`, renders two side-by-side dots inside a white oval instead of a single dot. Revenue labels rendered symmetrically left/right. Values other than `1` or `2` raise `NotImplementedError`. All existing single-dot rendering unchanged.

> **Design credit:** Oliver Burnett-Hall suggested extending `Part::Town` with a `size:` parameter and the `town=revenue:10,size:2;path=…` DSL form. Additional routing logic clarified in community discussion with Marek, TheCat, Anthony White, and Sebastian Schmitz.

### 18OE tile definitions

Six double-town tiles added, all using `size:2` center town:

| Tile | Color | Count | Exits |
|------|-------|-------|-------|
| OE9  | green | 3 | 0, 2, 3, 5 |
| OE10 | green | 3 | 1, 2, 3, 5 |
| OE11 | green | 3 | 0, 2, 3, 4 |
| OE20 | brown | 3 | 0, 1, 2, 5 |
| OE21 | brown | 2 | 0, 1, 2, 3, 5 |
| OE22 | brown | 6 | 0, 1, 2, 3, 4, 5 |

### 18OE routing logic

- **`game.rb` — `revenue_stops` override:** expands a `size == 2` town into two entries before `compute_stops` runs. This allows the engine to score 0, 1, or 2 town revenues from a double town based on the train's remaining town-pay capacity — no changes to `compute_stops` or `revenue_for` required.
- `visited_stops` is intentionally **not** overridden: that layer feeds `check_distance!` and the min-stop guard, so expanding there would cause a size-2 town to count as 2 distance units and break express train routes passing through one.
- Express and D-trains (`pay: 0` for towns in distance definition) are unaffected — neither expanded entry can fill a pay slot, so both are ignored automatically.
- Partial scoring (1 of 2) works naturally: if a train has 1 town slot remaining, `compute_stops` picks one of the two expanded entries.

### Compatibility

`size:` on a `town=` DSL part is unused in all existing games. The only existing `size:` usage in tile strings is on `upgrade=` parts (e.g. `g_1868_wy`) which is parsed separately and unaffected.

### DIFF
--- a/assets/app/view/game/part/town_dot.rb
+++ b/assets/app/view/game/part/town_dot.rb
@@ -153,16 +153,7 @@ module View
             raise NotImplementedError, "Unsupported town size: #{@town.size}"
           end
 
-          unless @town.solo?
-            if @town.size > 1
-              [-TWO_DOT_OFFSET, TWO_DOT_OFFSET].each do |dx|
-                children << h(HitBox, click: -> { touch_node(@town) },
-                               transform: "#{translate} translate(#{dx} 0)", r: 18)
-              end
-            else
-              children << h(HitBox, click: -> { touch_node(@town) }, transform: translate)
-            end
-          end
+          children << h(HitBox, click: -> { touch_node(@town) }, transform: translate) unless @town.solo?
           h(:g, { key: "#{@town.id}-d" }, children)
         end
 
--- a/assets/app/view/game/part/town_dot.rb
+++ b/assets/app/view/game/part/town_dot.rb
@@ -20,6 +20,7 @@ module View
         REVENUE_DISPLACEMENT = 42
         REVENUE_EDGE_DISPLACEMENT = 25
         REVENUE_ANGLE = -60
+        DOUBLE_DOT_OFFSET = 12
 
         REVENUE_REGIONS = {
           flat: [9, 16],
@@ -109,25 +110,68 @@ module View
 
         def render_part
           radius = 10 * (0.8 + (@width.to_i / 40))
-
-          dot_attrs = {
-            transform: translate.to_s,
-            r: radius,
-            fill: (@town.halt? ? 'gray' : @color),
-            stroke: (@town.halt? ? @color : 'white'),
-            'stroke-width': 4,
-          }
+          fill = @town.halt? ? 'gray' : @color
+          stroke = @town.halt? ? @color : 'white'
 
           children = []
 
-          children << h(:circle, attrs: dot_attrs)
-          children << render_boom if @town.boom
-
-          if @show_revenue && (rendered = render_revenue)
-            children << rendered
+          if @town.size > 1
+            children << h(:ellipse, attrs: {
+                            transform: translate.to_s,
+                            rx: DOUBLE_DOT_OFFSET + radius + 4,
+                            ry: radius + 3,
+                            fill: 'white',
+                            stroke: @color,
+                            'stroke-width': 2,
+                          })
+            [-DOUBLE_DOT_OFFSET, DOUBLE_DOT_OFFSET].each do |dx|
+              children << h(:circle, attrs: {
+                              transform: "#{translate} translate(#{dx} 0)",
+                              r: radius,
+                              fill: fill,
+                              stroke: stroke,
+                              'stroke-width': 4,
+                            })
+            end
+            children.concat(render_revenue_double) if @show_revenue
+          else
+            children << h(:circle, attrs: {
+                            transform: translate.to_s,
+                            r: radius,
+                            fill: fill,
+                            stroke: stroke,
+                            'stroke-width': 4,
+                          })
+            children << render_boom if @town.boom
+            if @show_revenue && (rendered = render_revenue)
+              children << rendered
+            end
           end
+
           children << h(HitBox, click: -> { touch_node(@town) }, transform: translate) unless @town.solo?
           h(:g, { key: "#{@town.id}-d" }, children)
+        end
+
+        def render_revenue_double
+          revenues = @town.uniq_revenues
+          return [] if revenues.size > 1
+
+          revenue = revenues.first
+          return [] if revenue.zero?
+
+          x = render_location[:x]
+          y = render_location[:y]
+
+          increment_weight_for_regions(REVENUE_REGIONS[layout])
+
+          [-REVENUE_DISPLACEMENT, REVENUE_DISPLACEMENT].map.with_index do |dx, i|
+            h(:g, {
+                key: "#{@town.id}-r#{i}",
+                attrs: { transform: "translate(#{(x + dx).round(2)} #{y.round(2)})" },
+              }, [
+              h(Part::SingleRevenue, revenue: revenue, transform: rotation_for_layout),
+            ])
+          end
         end
 
         def render_boom(transform: nil)
--- a/assets/app/view/game/part/town_dot.rb
+++ b/assets/app/view/game/part/town_dot.rb
@@ -153,7 +153,16 @@ module View
             raise NotImplementedError, "Unsupported town size: #{@town.size}"
           end
 
-          children << h(HitBox, click: -> { touch_node(@town) }, transform: translate) unless @town.solo?
+          unless @town.solo?
+            if @town.size > 1
+              [-TWO_DOT_OFFSET, TWO_DOT_OFFSET].each do |dx|
+                children << h(HitBox, click: -> { touch_node(@town) },
+                               transform: "#{translate} translate(#{dx} 0)", r: 18)
+              end
+            else
+              children << h(HitBox, click: -> { touch_node(@town) }, transform: translate)
+            end
+          end
           h(:g, { key: "#{@town.id}-d" }, children)
         end
 
--- a/assets/app/view/game/part/town_dot.rb
+++ b/assets/app/view/game/part/town_dot.rb
@@ -116,45 +116,56 @@ module View
           children = []
 
           case @town.size
-          when 2
-            children << h(:ellipse, attrs: {
-                            transform: translate.to_s,
-                            rx: TWO_DOT_OFFSET + radius + 4,
-                            ry: radius + 3,
-                            fill: 'white',
-                            stroke: @color,
-                            'stroke-width': 2,
-                          })
-            [-TWO_DOT_OFFSET, TWO_DOT_OFFSET].each do |dx|
-              children << h(:circle, attrs: {
-                              transform: "#{translate} translate(#{dx} 0)",
-                              r: radius,
-                              fill: fill,
-                              stroke: stroke,
-                              'stroke-width': 4,
-                            })
-            end
-            if @show_revenue && !(rendered = render_revenue_size2).empty?
-              children.concat(rendered)
-            end
-          when 1
-            children << h(:circle, attrs: {
-                            transform: translate.to_s,
-                            r: radius,
-                            fill: fill,
-                            stroke: stroke,
-                            'stroke-width': 4,
-                          })
-            children << render_boom if @town.boom
-            if @show_revenue && (rendered = render_revenue)
-              children << rendered
-            end
-          else
-            raise NotImplementedError, "Unsupported town size: #{@town.size}"
+          when 2 then children.concat(render_double_town(radius, fill, stroke))
+          when 1 then children.concat(render_single_town(radius, fill, stroke))
+          else raise NotImplementedError, "Unsupported town size: #{@town.size}"
           end
 
           children << h(HitBox, click: -> { touch_node(@town) }, transform: translate) unless @town.solo?
           h(:g, { key: "#{@town.id}-d" }, children)
+        end
+
+        def render_double_town(radius, fill, stroke)
+          result = []
+          result << h(:ellipse, attrs: {
+                        transform: translate.to_s,
+                        rx: TWO_DOT_OFFSET + radius + 4,
+                        ry: radius + 3,
+                        fill: 'white',
+                        stroke: @color,
+                        'stroke-width': 2,
+                      })
+          [-TWO_DOT_OFFSET, TWO_DOT_OFFSET].each do |dx|
+            result << h(:circle, attrs: {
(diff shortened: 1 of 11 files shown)


crericha:
Please include screenshots in the description 👍.

assets/app/view/game/part/town_dot.rb
```diff
@@ -109,27 +110,70 @@ def render_revenue
...
+            [-DOUBLE_DOT_OFFSET, DOUBLE_DOT_OFFSET].each do |dx|
+              children << h(:circle, attrs: {
+                              transform: "#{translate} translate(#{dx} 0)",
+                              r: radius,
+                              fill: fill,
+                              stroke: stroke,
+                              'stroke-width': 4,
+                            })
+            end
+            children.concat(render_revenue_double) if @show_revenue
```
crericha: Why is this logic different than line 146? (rendered = render_revenue_double)

assets/app/view/game/part/town_dot.rb
```diff
@@ -109,27 +110,70 @@ def render_revenue
...
+            if @show_revenue && (rendered = render_revenue)
+              children << rendered
+            end
           end
+
           children << h(HitBox, click: -> { touch_node(@town) }, transform: translate) unless @town.solo?
           h(:g, { key: "#{@town.id}-d" }, children)
         end
 
+        def render_revenue_double
```
crericha: The name says `_double`, but it is called from a logic case for towns > 1. That seems inconsistent.

assets/app/view/game/part/town_dot.rb
```diff
@@ -109,27 +110,70 @@ def render_revenue
...
+          if @town.size > 1
+            children << h(:ellipse, attrs: {
+                            transform: translate.to_s,
+                            rx: DOUBLE_DOT_OFFSET + radius + 4,
+                            ry: radius + 3,
+                            fill: 'white',
+                            stroke: @color,
+                            'stroke-width': 2,
+                          })
+            [-DOUBLE_DOT_OFFSET, DOUBLE_DOT_OFFSET].each do |dx|
```
crericha: Double dot? Isn't this logic for towns > 1? If it can't be done generically, maybe support town.size == 2 or 1 and if its another value, raise unsupported?

lib/engine/game/g_18_oe/game.rb
```diff
@@ -799,6 +841,10 @@ def upgrade_cost(tile, hex, entity, spender)
           cost
         end
 
+        def visited_stops(route)
+          super.flat_map { |stop| stop.town? && stop.size > 1 ? Array.new(stop.size, stop) : [stop] }
```
crericha: The multi town rendering logic was added in core. The logic to support it in the game code also needs to be there. 

Also, I'm not convinced yet that this naive approach will work. Lots of games count the route stops and burying the multiple stops inside of one stop isn't good.

Witzman: Good catch - the original `visited_stops` override was wrong. `route.rb` feeds `visited_stops` to both `check_distance!` and the min-stop guard, so expanding a size-2 town there would make it count as 2 distance units, breaking express train routes passing through one. Fixed in the latest push: the override is now on `revenue_stops` instead, which feeds only `compute_stops`. Distance checking, route-too-short validation, and groups checks all see the single unexpanded stop. `g_1880` uses `revenue_stops` for the same pattern.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -471,13 +482,28 @@ class Game < Game::Base
...
+            {
+              'count' => 3,
+              'color' => 'green',
+              'code' => 'town=revenue:10,size:2;path=a:0,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:5,b:_0',
+            },
+          'OE10' =>
             {
               'count' => 3,
               'color' => 'green',
+              'code' => 'town=revenue:10,size:2;path=a:1,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:5,b:_0',
```
ollybh: OE10 is a K-shape, not a trident. This definition is the same as you have for OE11, just rotated slightly.
```suggestion
              'code' => 'town=revenue:10,size:2;path=a:1,b:_0;path=a:2,b:_0;path=a:3,b:_0;path=a:4,b:_0',
```

lib/engine/game/g_18_oe/game.rb
ollybh, line 466: For consistency, you should update these tiles to use the double dot rendering.

lib/engine/game/g_18_oe/game.rb
Witzman: fixed

lib/engine/game/g_18_oe/game.rb
Witzman: you looked, while i drafted the PR - i worked on this yesterday :)

assets/app/view/game/part/town_dot.rb
```diff
@@ -109,27 +110,75 @@ def render_revenue
...
+                            transform: translate.to_s,
+                            r: radius,
+                            fill: fill,
+                            stroke: stroke,
+                            'stroke-width': 4,
+                          })
+            children << render_boom if @town.boom
+            if @show_revenue && (rendered = render_revenue)
+              children << rendered
+            end
```
crericha: These when cases got a little long. Can we move the logic for each when case into a separate helper function?

lib/engine/game/g_18_oe/meta.rb
```diff
@@ -11,7 +11,7 @@ module Meta
         DEV_STAGE = :prealpha
 
         GAME_DESIGNER = 'Edward T. Sindelar'
-        GAME_PUBLISHER = :DICE
+        GAME_PUBLISHER = :self_published
```
ollybh: This game was professionally published, so this is wrong.

Add the publisher to `lib/engine/publisher.rb`, though change `:DICE` to `:dice` for consistency. Their website is still live: http://designsice.com.

lib/engine/game/g_18_oe/game.rb
```diff
@@ -922,7 +922,7 @@ def new_consolidation_round
 
         def upgrades_to_correct_label?(from, to)
           chain = self.class::METROPOLIS_UPGRADE_CHAINS[from.hex&.coordinates]
-          return (chain[chain.index(from.name) + 1] == to.name) if chain&.include?(from.name)
+          return (chain[chain.find_index(from.name) + 1] == to.name) if chain&.include?(from.name)
```
ollybh: Please change this back. The preferred style for this project is `index`, not `find_index` (though you don't have to look too hard to find places where `find_index` is used.

lib/engine/game/g_18_oe/game.rb
Witzman: The expansion is entirely in 18OE's `game.rb` — no engine changes. `visited_stops` returns the stop twice so distance-checking correctly consumes 2 town slots from the train's capacity. `revenue_stops` returns it twice so it scores twice. The core stop object is unchanged; only the arrays returned by these two game-level overrides are expanded. Other games are unaffected.

assets/app/view/game/part/town_dot.rb
Witzman: The dispatch now uses explicit `when 2` / `when 1` / `else raise NotImplementedError` — unsupported sizes raise immediately at render time.

assets/app/view/game/part/town_dot.rb
Witzman: Renamed to `render_revenue_size2` and now called from an explicit `when 2` case, not a `size > 1` guard.

assets/app/view/game/part/town_dot.rb
Witzman: `render_revenue_double` has been renamed to `render_revenue_size2` and extracted as its own helper. The revenue logic for size 2 renders two symmetrical labels (left/right of the oval) rather than one displaced label, hence the separate method.

lib/engine/game/g_18_oe/game.rb
Witzman: Done — `:dice` added to `publisher.rb`, `GAME_PUBLISHER` updated in `meta.rb`.

Witzman: Done — OE1/OE2/OE3 all use `size:2` in this PR.

lib/engine/game/g_18_oe/game.rb
Witzman: Fixed.

assets/app/view/game/part/town_dot.rb
Witzman: Done — extracted `render_double_town` and `render_single_town` helpers; `render_part`'s case block is now 4 lines.

lib/engine/game/g_18_oe/game.rb
Witzman: Reverted to `index`.

lib/engine/game/g_18_oe/meta.rb
Witzman: Fixed — `:dice` added to `publisher.rb` (Designs in Computer Entertainment, `http://designsice.com`), `GAME_PUBLISHER` updated in `meta.rb`.

Witzman:
Addressing the routing concern raised by @crericha — want to be transparent about the design trade-off before the next review pass.

### Why one `Town` object with `size:2`, not two separate nodes

I tried the two-node approach (two `town=` entries with an internal path connecting them). It ran into two problems:

1. **Rendering**: the internal route segment between the two town nodes renders visibly on the tile and looks wrong (see tiles OE09-11 & OE 20-22) grafik. It also interacts badly when multiple trains run through the same tile via different exits.
2. **Routing complexity**: the internal connection between `town_0` and `town_1` needs an override to handle multiple trains — the walk logic would need to treat the internal path as non-blocking for subsequent trains, which is a non-trivial engine change.

### How stop-counting is handled

`visited_stops` and `revenue_stops` are both overridden in `G18OE::Game` to expand a `size:2` stop into two entries in the array. A 2+2 train running through a double town therefore consumes **2** town slots, not 1 — crericha's concern is addressed at the game level rather than in the core stop object.

### Known trade-off

Players cannot choose to visit only one of the two towns in a double town — it's always both or none. This only matters at **AE6** and **X79**, which are 20-revenue paying towns. A player theoretically might want to skip one half of a double town to preserve train capacity for the 20-revenue stop. In practice this edge case is rare - i will have a look on this, before moving to beta.

Happy to discuss further if this approach doesn't satisfy the project guidelines

crericha:
> `visited_stops` and `revenue_stops` are both overridden in `G18OE::Game` to expand a `size:2` stop into two entries in the array. A 2+2 train running through a double town therefore consumes **2** town slots, not 1 — crericha's concern is addressed at the game level rather than in the core stop object.

We can't add core code that relies on overriding in the game. I feel like the rendering was the hard part. If you make a two town part, you can maintain the rendering and add specific walk logic.

To be clear, my suggestion is to make a double town part, have that part composed of two towns + a path between them. Maintain the rendering code you already wrote. Add specific walk logic that if the part is entered, it visits both towns before exiting. Note: I'm not sure you even need the path. You'd have to see.

ollybh:
> To be clear, my suggestion is to make a double town part, have that part composed of two towns + a path between them. Maintain the rendering code you already wrote. Add specific walk logic that if the part is entered, it visits both towns before exiting. Note: I'm not sure you even need the path. You'd have to see.

I don't like this idea, particularly altering the walk code. That's already way too complicated.

These double towns are like a city with two slots. It should be a single stop on the route. The 18OE code can use `check_distance` or `visit_cost` or something similar to ensure that route length it correct.

Witzman:
reworked the PR based on your direction. replaced  + game-level overrides with a proper  engine class.

### what changed

**new:** `lib/engine/part/double_town.rb` — `DoubleTown < Town`
- holds two internal `Town` sub-stops (`@town_a`, `@town_b`) via `sub_stops`
- overrides `walk` to mark both sub-stops as visited on entry (no internal path object, so nothing extra renders)
- `size` returns 2, `rect?` returns false — rendering unchanged

**modified:** `lib/engine/tile.rb`
- `when 'town'`: instantiates `DoubleTown` instead of `Town` when `size:2`

**modified:** `lib/engine/game/base.rb` — one line in `visited_stops`:
```ruby
.flat_map { |s| s.respond_to?(:sub_stops) ? s.sub_stops : [s] }
```
expansion logic is owned by the Part via `sub_stops`, not game code. any game using `DoubleTown` gets correct stop counting automatically.

**deleted:** `G18OE::Game#visited_stops` and `#revenue_stops` overrides — gone entirely.

### why the base.rb line is needed

i tested the walk override alone first. `visited_stops` reads from `route.connection_data` (path endpoints), not from the walk's `visited` hash — so `DoubleTown` lands in `connection_data` once regardless of how many internal towns walk marks. the one-liner in `base.rb` asks the Part 'how many stops do you represent?' and the Part answers via `sub_stops`. same principle as your direction, just wired at the right layer.

### verified

- OE9–OE22 render correctly (oval + two dots, no internal segment)
- OE10 double-town scores £20 (2 × £10) via `sub_stops` expansion
- `visited_stops` returns 2 stops for any `DoubleTown`
- 14 IRB tests pass
- no game-level override anywhere

let me know if the `base.rb` line is still a concern — happy to discuss.

PR #12672, opened 2026-05-25
[18MO] Fixes half-price token with teleport token ability

### LINKED ISSUE #7490: [1852] Half-price token not working with teleport ability?

philcampeau:
Fixes #7490

## Before clicking "Create"

### Explanation of Change

Half price Token private wasn't working to discount the teleport tokens that different corporations have. 

This adds in a check for the half price token private in the process_place_token step, and inserts needed code in there to allow both abilities to work together.

### DIFF
--- a/lib/engine/step/tokener.rb
+++ b/lib/engine/step/tokener.rb
@@ -68,7 +68,7 @@ module Engine
         tokener = entity.name
         if ability
           tokener += " (#{ability.owner.sym})" if ability.owner != entity
-          entity.remove_ability(ability)
+          entity.remove_ability(ability) unless ability == special_ability
         end
 
         raise GameError, 'Token is already used' if token.used
@@ -126,11 +126,16 @@ module Engine
           return [token, special_ability]
         end
 
-        # TODO: special_ability token here
+        ability_found = false
+
         @game.abilities(entity, :token) do |ability, _|
+          discount = ability.discount
           next if ability.special_only && ability != special_ability
+          next if !discount.nil? && discount.positive? && discount < 1 && ability.hexes.empty? && ability != special_ability
           next if ability.hexes.any? && !ability.hexes.include?(hex.id)
           next if ability.city && ability.city != city.index
+
+          ability_found = true
 
           if ability.neutral
             neutral_corp = Corporation.new(
@@ -146,11 +151,29 @@ module Engine
           end
 
           token.price = ability.teleport_price if ability.teleport_price
-          token.price = ability.price(token) if @game.token_graph_for_entity(entity).reachable_hexes(entity)[hex]
+          token.price = ability.price(token).to_i if @game.token_graph_for_entity(entity).reachable_hexes(entity)[hex]
+
+          return [token, special_ability] if apply_token_discount!(token, special_ability, ability)
+
           return [token, ability]
         end
 
+        return [token, special_ability] if !ability_found && apply_token_discount!(token, special_ability, nil)
+
         [token, nil]
+      end
+
+      def apply_token_discount!(token, special_ability, ability)
+        if special_ability&.discount.is_a?(Float) &&
+          special_ability.discount < 1 &&
+          special_ability.hexes.empty? &&
+          special_ability != ability
+          token.price = (token.price * special_ability.discount).to_i
+          special_ability.use!
+          return true
+        end
+
+        false
       end
 
       def check_connected(entity, city, hex)
--- a/lib/engine/game/g_18_mo/step/special_token.rb
+++ b/lib/engine/game/g_18_mo/step/special_token.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../../../step/special_token'
+
 module Engine
   module Game
     module G18MO
--- a/lib/engine/game/g_18_mo/step/special_token.rb
+++ b/lib/engine/game/g_18_mo/step/special_token.rb
@@ -32,8 +32,12 @@ module Engine
           def available_hex(entity, hex)
             ability = ability(entity)
             return unless ability
-            return @game.token_graph_for_entity(entity.owner).reachable_hexes(entity.owner)[hex] if
-              ability.type == :token && ability.hexes.empty?
+
+            if ability.type == :token && ability.hexes.empty?
+              return true if entity.owner.all_abilities.any? { |a| a.type == :token && a.hexes.include?(hex.id) }
+
+              return @game.token_graph_for_entity(entity.owner).reachable_hexes(entity.owner)[hex]
+            end
 
             super
           end
@@ -42,7 +46,7 @@ module Engine
             return super unless entity&.company?
 
             @game.abilities(entity, :token) do |ability, _|
-              next if ability.owner_type == 'corporation'
+              next if ability.owner_type == 'corporation' && !ability.discount
 
               return ability
             end
--- a/lib/engine/game/g_18_mo/step/special_token.rb
+++ b/lib/engine/game/g_18_mo/step/special_token.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative '../../../step/special_token'
-
 module Engine
   module Game
     module G18MO
@@ -27,6 +26,31 @@ module Engine
 
             teleport_complete if @round.teleported
             @game.remove_teleport_destination(entity, action.city)
+          end
+
+          def available_hex(entity, hex)
+            ability = ability(entity)
+            return unless ability
+            return @game.token_graph_for_entity(entity.owner).reachable_hexes(entity.owner)[hex] if
+              ability.type == :token && ability.hexes.empty?
+
+            super
+          end
+
+          def ability(entity)
+            return super unless entity&.company?
+
+            @game.abilities(entity, :token) do |ability, _|
+              next if ability.owner_type == 'corporation'
+
+              return ability
+            end
+
+            @game.abilities(entity, :teleport) do |ability, _|
+              return ability if ability.used?
+            end
+
+            nil
           end
         end
       end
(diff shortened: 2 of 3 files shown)


ollybh:
If I've understood the linked issue correctly, the teleport token is being laid but the price not discounted. This PR seems to be making much bigger changes than just the token price – what else is being altered?

philcampeau:
It's not so much that I made bigger changes. It's the way the site wanted to apply two different abilities at once. It doesn't apply the token discount to the teleport token. 

It's possible I didn't a lot more work than was needed. I spent a good while on this, and this was the only configuration I found where it worked as expected.

ollybh (changes requested):
I've taken another look at this, but can't understand why you're changing `available_hex`. The linked issue is just talking about the token price being incorrect, but this method is controlling where a token can be placed, not its cost.

As for the changes in `process_place_token`, could you not override `adjust_token_price_ability!` instead? Or make changes to the base method if this combination is an oversight that should be fixed there.

philcampeau:
Just an update to say: You're definitely right about moving some of this to the base code. Predictably, that broke a handful of other games. I'm still testing things but I'm making progress.

philcampeau:
So yeah, you're right, `adjust_token_price_ability!` would be the correct place to fix this. The issue is that the base method didn't handle the case where a corporation has both a special hex ability with a teleport price and a separate discount token ability. It would find one or the other but never combine them.

The changes made to the base method are:

- Skip abilities that are fractional discounts with empty hexes in the outer loop, so they aren't mistakenly matched as the primary placement ability.
- After resolving the base price from the corp's special hex ability, check if `special_ability` is a fractional discount and apply it on top
- Handle the case where no corp ability matches at all (e.g. tokening a normal hex with HPTOK) by applying the discount post-loop
- Guard against double-consuming the ability when `special_ability == ability` (e.g. Weidenthal in 1847 AE), since `process_place_token` already calls use!
- Skip `entity.remove_ability` in `place_token` when `ability == special_ability` for the same reason

The `available_hex` change in SpecialToken is needed because when a company like HPTOK is the acting entity, the base available_hex calls `can_place_token?` which ends up calling `buying_power` on the company rather than the corporation, causing an error. The override substitutes current_entity and also ensures the corporation's special destination hex is shown as available even without a connection.

lib/engine/step/tokener.rb
```diff
@@ -146,13 +151,31 @@ def adjust_token_price_ability!(entity, token, hex, city, special_ability: nil)
...
+          special_ability.discount < 1 &&
+          special_ability.hexes.empty? &&
+          special_ability != ability
+          token.price = (token.price * special_ability.discount).to_i
+          special_ability.use!
+          return true
+        end
+
+        false
+      end
```
ollybh, line 177: Ruby methods with a `!` suffix shouldn't be returning a value.

lib/engine/step/tokener.rb
```diff
@@ -146,13 +151,31 @@ def adjust_token_price_ability!(entity, token, hex, city, special_ability: nil)
...
+          token.price = ability.price(token).to_i if @game.token_graph_for_entity(entity).reachable_hexes(entity)[hex]
+
+          return [token, special_ability] if apply_token_discount!(token, special_ability, ability)
+
           return [token, ability]
         end
 
+        return [token, special_ability] if !ability_found && apply_token_discount!(token, special_ability, nil)
+
         [token, nil]
```
ollybh, line 163: To avoid having a method `apply_token_discount!` that both modifies its arguments and returns a boolean, this could be structured something like this:
```ruby
  return [token, nil] unless !ability_found && token_discount?(token, special_ability)

  discount_token!(token, special_ability)
  [token, special_ability]
end

def token_discount?(token, special_ability, ability)
  special_ability&.discount.is_a?(Float) &&
    special_ability.discount < 1 &&
    special_ability.hexes.empty? &&
    special_ability != ability
end

def discount_token!(token, special_ability)
  token.price = (token.price * special_ability.discount).to_i
  special_ability.use!
end
```

lib/engine/step/tokener.rb
ollybh: Of course, this new method is called from a method which has an `!` name, and it returns an array 😖

ollybh (commented):
I attempted to review this PR, but I'm not able to follow the logic of `adjust_token_price_ability!`. It was already too long and too complicated, these additions do not make it simpler. It needs to be refactored, ideally split into several different functions for the different combinations of abilities that are being tested.

ollybh:
FWIW, I asked Deepseek to suggest a refactoring of `adjust_token_price_ability!`. This is what it came up with, including extensive Yardoc. It's certainly much clearer, but I have not tested any of this.

With a refactoring like this, would 18MO then just need to override `resolve_token_price`?

[The Yardoc highlighted something I had missed in your code: none of the existing code in `adjust_token_price_ability!` calls `ability.use!`, so your changes shouldn't do that either].

```ruby
  # Adjusts the token and its price based on the entity's abilities
  # and the destination hex/city.
  #
  # This method is called by #place_token before the token is actually
  # placed.  It may replace the token object (e.g. with a neutral token
  # or a newly-created token from a company ability) and may override
  # its price.
  #
  # The method composes three private helpers:
  #   1. #resolve_teleport_ability!  — handles teleport special abilities
  #   2. #select_token_ability       — finds the matching :token ability
  #   3. #build_token_for_ability!   — creates/reuses the correct token
  #
  # Game-specific sub-steps override this method to implement custom
  # token pricing (e.g. hex-distance-based pricing in 1867, 1837).
  #
  # @example Override for hex-distance pricing
  #   def adjust_token_price_ability!(entity, token, hex, _city,
  #                                    special_ability: nil)
  #     token.price = hex_distance(entity, hex) * token.price
  #     [token, nil]
  #   end
  #
  # @param entity [Engine::Corporation, Engine::Company] The entity
  #   placing the token.
  # @param token [Engine::Token] The default token object (usually
  #   +entity.next_token+).
  # @param hex [Engine::Hex] The hex where the token will be placed.
  # @param city [Engine::City] The specific city within the hex.
  # @param special_ability [Engine::Ability::Base, nil] An optional
  #   special ability that triggered this placement (e.g. from a
  #   company action like +SpecialToken+).
  #
  # @return [Array(Engine::Token, Engine::Ability::Base, nil)] A tuple
  #   of the resolved token and the ability that was used (or +nil+ if
  #   no ability was applied).  The caller is responsible for consuming
  #   the ability via #remove_ability.
  def adjust_token_price_ability!(entity, token, hex, city, special_ability: nil)
    return resolve_teleport_ability!(entity, token, special_ability) if
      special_ability&.type == :teleport

    ability = select_token_ability(entity, token, hex, city,
                                   special_ability: special_ability)
    return [token, nil] unless ability

    token = build_token_for_ability!(entity, token, ability)
    token.price = resolve_token_price(entity, token, hex, ability)

    [token, ability]
  end

  private

  # Handles teleport special abilities: creates or reuses a token and
  # sets its price to zero (free placement without connectivity).
  #
  # @param entity [Engine::Corporation, Engine::Company]
  # @param token [Engine::Token]
  # @param ability [Engine::Ability::Token]
  # @return [Array(Engine::Token, Engine::Ability::Token)]
  #
  def resolve_teleport_ability!(entity, token, ability)
    unless ability.from_owner
      token = Engine::Token.new(entity)
      entity.tokens << token
    end
    token.price = 0
    [token, ability]
  end

  # Finds the first +:token+ ability on +entity+ that matches the given
  # +hex+, +city+, and +special_ability+ constraints.
  #
  # Filtering rules:
  #   - Abilities marked +special_only+ are skipped unless they are the
  #     +special_ability+ being explicitly invoked.
  #   - Abilities restricted to specific hexes must include +hex+.
  #   - Abilities restricted to a specific city index must match +city+.
  #
  # @param entity [Engine::Corporation, Engine::Company]
  # @param token [Engine::Token]
  # @param hex [Engine::Hex]
  # @param city [Engine::City]
  # @param special_ability [Engine::Ability::Base, nil]
  # @return [Engine::Ability::Token, nil]
  #
  def select_token_ability(entity, _token, hex, city, special_ability: nil)
    @game.abilities(entity, :token) do |ability|
      next if ability.special_only && ability != special_ability
      next if ability.hexes.any? && !ability.hexes.include?(hex.id)
      next if ability.city && ability.city != city.index

      return ability
    end
    nil
  end

  # Creates or reuses a token based on the ability's properties.
  #
  # - +neutral+ abilities create a token owned by a dummy "Neutral"
  #   corporation.
  # - Company-owned abilities create a new token on the entity's charter
  #   (unless +from_owner+ is set, which uses the entity's existing next
  #   token).
  # - Otherwise the passed-in +token+ is returned unchanged.
  #
  # @param entity [Engine::Corporation, Engine::Company]
  # @param token [Engine::Token]
  # @param ability [Engine::Ability::Token]
  # @return [Engine::Token]
  #
  def build_token_for_ability!(entity, token, ability)
    if ability.neutral
      neutral_corp = Corporation.new(
        sym: 'N',
        name: 'Neutral',
        logo: 'open_city',
        tokens: [0],
      )
      return Engine::Token.new(neutral_corp, type: :neutral)
    end

    if ability.owner.company? && !ability.from_owner
      token = Engine::Token.new(entity)
      entity.tokens << token
    end

    token
  end

  # Computes the token price based on the ability and the destination's
  # reachability.
  #
  # Priority (first match wins):
  #   1. +teleport_price+ if the ability has one (bypasses connectivity).
  #   2. +price+ (with discount) if the destination hex is reachable from
  #      the entity's token network.
  #   3. Otherwise the token's default +price+ is preserved.
  #
  # @param entity [Engine::Corporation, Engine::Company]
  # @param token [Engine::Token]
  # @param hex [Engine::Hex]
  # @param ability [Engine::Ability::Token]
  # @return [Integer]
  #
  def resolve_token_price(entity, token, hex, ability)
    return ability.teleport_price if ability.teleport_price
    return ability.price(token) if
      @game.token_graph_for_entity(entity).reachable_hexes(entity)[hex]

    token.price
  end
```

PR #12680, opened 2026-05-25
[18GA] Jacksonville should be terminal except for ACL

### LINKED ISSUE #11206: [18GA] only ACL should be able to trace a route through Jacksonville
[Gist link](https://gist.github.com/philcampeau/8d85cfc90cb0ff0356d4a187071f797b) showing the issue.

**How to Reproduce the Bug?**

As per the rules: 

image

However, if you build a route to Jacksonville before ACL is floated, you can build past it and run trains through it:

image

philcampeau:
Fixes #11206

## Before clicking "Create"

### Explanation of Change

Per the rules, Jacksonville (J12) should be a terminal except for ACL. Currently, if ACL hasn't started, corps can run through that spot freely. 

This change puts a flipped ACL token in J12 prior to ACL floating. The reason I did it this way is so that ACL's home is still visible on the map from the start, and there's a visible cue as to why Jacksonville is blocked. 

I had to provide new fixtures, as all 3 broke this rule. Currently there's only one live game of 18GA and it's in the auction, so now's a perfect time to pop in this change!

### Screenshots

Rule:
image

Jacksonville with flipped token:
image


perwestling (approved):
LGTM

lib/engine/game/g_18_ga/game.rb
```diff
@@ -146,6 +153,11 @@ def float_corporation(corporation)
           @recently_floated << corporation
 
           super
+
+          return unless corporation.name == 'ACL'
```
ollybh:
```suggestion
          return unless corporation.id == 'ACL'
```

ollybh (commented):
The code for flipping ACL's token doesn't look like it will work. Shouldn't it be `corporation.id` not `corporation.name`?

lib/engine/game/g_18_ga/game.rb
philcampeau: yup, good point. Fixed.

ollybh (approved):

PR #12682, opened 2026-05-26
[assets] improves tile render with icons

### LINKED ISSUE #5167: 1856: CPR Destination overlaps Green Toronto's city value
image

philcampeau:
Fixes #5167

## Before clicking "Create"

### Explanation of Change

Currently, icons are rendered (and claim their regions in @region_use) after cities, so when a city renders its inline revenue circle it doesn't know the icon is coming and can pick the same spot. Moving icon rendering before cities means the icon claims its region first, and the city's revenue circle then picks a different spot that doesn't overlap.

The change is simply moving these two lines earlier in the render method, before Part::Cities is rendered:

```
large, normal = @tile.icons.partition(&:large)
render_tile_parts_by_loc(Part::Icons, parts: normal).each { |i| children << i }
```

### Screenshots

Before: 

image

after:

image

### DIFF
--- a/assets/app/view/game/tile.rb
+++ b/assets/app/view/game/tile.rb
@@ -82,6 +82,7 @@ module View
         large, normal = @tile.icons.partition(&:large)
         render_tile_parts_by_loc(Part::Icons, parts: normal).each { |i| children << i }
         children << render_tile_part(Part::Cities, show_revenue: !render_revenue) unless @tile.cities.empty?
+
         children << render_tile_part(Part::Towns, routes: @routes, show_revenue: !render_revenue) unless @tile.towns.empty?
 
         borders = render_tile_part(Part::Borders) if @tile.borders.any?(&:type)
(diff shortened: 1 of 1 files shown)


assets/app/view/game/tile.rb
```diff
@@ -79,8 +79,9 @@ def render
         if !@tile.paths.empty? || !@tile.stubs.empty? || !@tile.future_paths.empty?
           children << render_tile_part(Part::Track, routes: @routes)
         end
+        large, normal = @tile.icons.partition(&:large)
```
perwestling, line 82: Looking at the code it looks like it should be an empty line before large, normal

perwestling:
Looking fine to me but I am not familiar with this part of the code.

assets/app/view/game/tile.rb
philcampeau: sure, I restored it now.

ollybh:
@philcampeau This looks sensible but needs to be tested to make sure that nothing is made worse. Can you get your AI-assistant to load the various `/tiles/[game/all` pages (and the base `/tiles/all` page) before and after these changes and identify the tiles which are changed.

philcampeau:
I looked through a lot of tiles where the icon is in there from the TILES definition, and they all looked good. 

However, one thing I hadn't checked was preprinted map hexes, and I did come across a few issues there. I'm gonna need to think about that. I'll set this to Draft mode for now.

PR #12684, opened 2026-05-26
[1860] no subsidies on simplified insolvency

### LINKED ISSUE #11198: [1860] Simplified insolvency variant pays subsidy for halts
**What's your Game ID?**
https://18xx.games/game/174873

**How to Reproduce the Bug?**
See, e.g., https://18xx.games/game/174873?action=454 (OR 7.2, FYN operating)

0. Select "Simplified Insolvency" variant.
1. Operate an insolvent corporation.
2. Select a route for the leased train that includes halts.
3. Select a revenue that includes a subsidy.

Ideally, under the simplified insolvency variant, it should not be necessary to select any route. Since the corporation is insolvent, it necessarily has some route. Under the simplified insolvency variant, all routes regardless of length and stations/halts should pay the same (40 + 20 * leased train large station limit).

**Which Browser / Platform?**
Does not appear to be specific to browser. Do not know what browser other players are using. I am currently running Chrome 128.0.6613.120.

philcampeau:
Fixes #11198

## Before clicking "Create"

### Explanation of Change

It looks like I added a condition to `def loaner_new_rules?(route)` that shouldn't have been there when I coded this. This change fixes the issue, but may require some pins for games that used it incorrectly.

### DIFF
--- a/lib/engine/game/g_1860/game.rb
+++ b/lib/engine/game/g_1860/game.rb
@@ -1017,7 +1017,7 @@ module Engine
         end
 
         def loaner_new_rules?(route)
-          loaner?(route) && !option_original_insolvency? && !option_simplified_insolvency?
+          loaner?(route) && !option_original_insolvency?
         end
 
         def loaner_orig_rules?(route)


perwestling (approved):
LGTM

roseundy (approved):
Sorry for not responding

PR #12685, opened 2026-05-26
[1871] fixes EMR for Union Bank

### LINKED ISSUE #8831: Cannot Go Bankrupt when union bank is president
https://18xx.games/game/112613

Forced train buy for the MS with Union bank as president and acting player owner of the union bank. Acting player sold shares and did not have enough money. Game did not allow them to go bankrupt.

### LINKED ISSUE #11197: [1871] President was unable to sell share to contribute to Union Bank EMR
Player owning Union bank had to contribute to EMR to buy train. However game claims there are no sellable stocks. This appears to be in error. The stocks in So and A should be possible to sell, why MS and S can't be sold due to they not having operated.

**What's your Game ID?**

https://18xx.games/game/174177?action=498

**How to Reproduce the Bug?**

Unclear but it is possible that the evaluation of sellable stocks has an issue, as two companies in the president holding fails on the "have not operated yet" rule. However, the president should be able to sell in this situation So and A stock to get enough money to contribute the necessary to not have to declare bankruptcy. 

bild
bild

 
**Which Browser / Platform?**
Firefox 129.0.2

philcampeau:
Fixes #8831
Fixes #11197

## Before clicking "Create"

### Explanation of Change

Reworked the `can_go_bankrupt?` method so that it also checks the UB owner, and added a custom `total_emr_buying_power` method to factor in the UB owner's liquidity. 

As I'm continuing testing, I've discovered an issue if UB has $0. Will continue troubleshooting this and add a commit once i've corrected that. 

This PR could require pins for any games where the UB president couldn't afford a train, but those games would have been stuck anyway, so it's likely not an issue. 

EDIT: 

I added a change to Engine/Step/Train.rb It adds a  `if current_cash.positive?` gate onto two of the lines in the borrow_from section. 

The base code assumes that when borrow_from is set, the player always has some cash to contribute first. But that assumption breaks when:

- The player entity is the Union Bank with $0 (all shares sold, all cash exhausted)
- Any other game where a president legitimately reaches $0 before the train buy completes

In both cases, spend(0) and logging "X contributes $0" are meaningless operations that crash or pollute the log. The guards simply skip them when there's nothing to contribute, letting borrow_from.spend(extra_needed, entity) handle the full remaining amount cleanly. The end result, that the corporation gets the cash it needs, is identical.

### DIFF
--- a/lib/engine/game/g_1871/game.rb
+++ b/lib/engine/game/g_1871/game.rb
@@ -498,14 +498,30 @@ module Engine
         end
 
         def can_go_bankrupt?(player, corporation)
-          return false if player == @union_bank
+          # Normalize: player might be @union_bank itself or the human who owns it
+          acting = player == @union_bank ? company_by_id('UB').owner : player
+          ub_owner = company_by_id('UB')&.owner
 
-          if corporation.owner == @union_bank && company_by_id('UB').owner == player
-            total_emr_buying_power(player, corporation) +
-              total_emr_buying_power(@union_bank, corporation) < @depot.min_depot_price
+          if acting == ub_owner
+            player_power = liquidity(acting, emergency: true)
+            ub_power = liquidity(@union_bank, emergency: true)
+            corp_power = corporation.cash + emergency_issuable_cash(corporation)
+            player_power + ub_power + corp_power < @depot.min_depot_price
           else
             super
           end
+        end
+
+        def total_emr_buying_power(player, corporation)
+          # When UB is the acting entity, combine UB + owner liquidity
+          if player == @union_bank
+            ub_owner = company_by_id('UB')&.owner
+            owner_power = ub_owner ? liquidity(ub_owner, emergency: true) : 0
+            ub_power = liquidity(@union_bank, emergency: true)
+            corp_power = corporation ? corporation.cash + emergency_issuable_cash(corporation) : 0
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1871/game.rb
```diff
@@ -498,16 +498,32 @@ def must_buy_train?(entity)
...
-          return false if player == @union_bank
-
-          if corporation.owner == @union_bank && company_by_id('UB').owner == player
-            total_emr_buying_power(player, corporation) +
-              total_emr_buying_power(@union_bank, corporation) < @depot.min_depot_price
+          # Normalize: player might be @union_bank itself or the human who owns it
+          acting = player == @union_bank ? company_by_id('UB').owner : player
+          ub_owner = company_by_id('UB')&.owner
+
+          if acting == ub_owner
```
perwestling: Is this correct? Is not acting a boolean and ub_owner not a boolean?

philcampeau: Somehow I missed your comment. 

No, acting is not a boolean. This is a ternary, though I admit it's hard to read. Another way I could have written it would be 

```
if player == @union_bank
  acting = ub_owner
else
  acting = player
end
```

Either way, I think I have a way of cleaning it up even better..

PR #12687, opened 2026-05-26
[1871] Allow UB to split corps

### LINKED ISSUE #9944: [1871] Union Bank doesn't seem to be able to split a company
**What's your Game ID?**

https://18xx.games/game/141763?action=175

**How to Reproduce the Bug?**

Have the Union Bank be president of a company (with 4 shares). Ensure the company has two stations on the map. Ensure there's an open slot in the current tranche for a new company to be started. The bank should be able to split the company and form a branch company but does not appear to have the option to do so.

**Which Browser / Platform?**

Chrome Version 113.0.5672.92 (Official Build) (arm64).

philcampeau:
Fixes #9944

## Before clicking "Create"

### Explanation of Change

- `can_split?` was checking `corporation.owner == splitter`, which would never match a player when the UB is president. Fixed to resolve an effective_splitter: if the splitter is the UB's owner and the UB is the corporation's president, treat the UB as the effective splitter for ownership and percent checks.
- `can_split_any?` was only checking if the player directly satisfied `can_split?`. Fixed to also check if the UB satisfies `can_split?` when the entity is the UB's owner.
- `split_active_entities` was returning [@split_corporation.owner], which would be the UB (a corporation) rather than a player, so no one would get the choice prompt. Fixed to resolve the UB's owner as the active entity when the UB is the corporation's president.

- minor: corrected typo in method (splitter, not spliter)

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1871/game.rb
+++ b/lib/engine/game/g_1871/game.rb
@@ -580,16 +580,20 @@ module Engine
 
         # Allow a company to split, if the owner has >= 40 percent, at least two
         # tokens out and there is a tranch available
-        def can_split?(corporation, spliter)
+        def can_split?(corporation, splitter)
           return false unless tranch_available?
           return false if corporation.name == 'PEIR'
           return false unless corporation.ipoed
 
+          # Resolve the effective splitter — if the UB is president, its owner acts on its behalf
+          effective_splitter = splitter
+          effective_splitter = union_bank if splitter == company_by_id('UB')&.owner && union_bank == corporation.owner
+
           # Have to own the company in question
-          return false unless corporation.owner == spliter
+          return false unless corporation.owner == effective_splitter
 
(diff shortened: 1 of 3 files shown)


perwestling (dismissed):

---
### ✅ Strengths
- Fixes documented issue #9944
- Clear explanation of the problem and solution in PR description
- Tests pass (CI shows SUCCESS)
- Minimal, focused change (only 3 files, 14 lines)
- Also fixes typo (spliter → splitter)
- Well-structured code with clear comments

### ⚠️ Concerns
- None

### 💡 Suggestions
- Consider adding a fixture test for this scenario to prevent regression. While the existing tests pass, a specific test case for UB splitting corporations would provide additional confidence.

### ❓ Questions
- None

---
### ✅ Verdict: **Approve**

Ready to merge. Well-implemented bug fix.

PR #12688, opened 2026-05-26
[Front Page] Adds "Hide Full Games" checkbox to New Games section

### LINKED ISSUE #8658: Hide Full games in lobby
Add a checkbox above the New Games section:

"Hide Full Games" 

The result being only games with available seats will be displayed

philcampeau:
Fixes #8658

## Before clicking "Create"

## Implementation Notes

This adds a checkbox to the New Games section of the front page, allowing users to hide full new games that haven't started yet. It doesn't affect the display of active, hotseat, or finished games in any way. 

### Screenshots

Unchecked:
image

Checked:
image

### DIFF
--- a/assets/app/view/game_row.rb
+++ b/assets/app/view/game_row.rb
@@ -9,6 +9,7 @@ module View
     include GameManager
 
     needs :header
+    needs :header_extra, default: nil
     needs :game_row_games
     needs :user
     needs :type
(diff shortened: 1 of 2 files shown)


ollybh (changes requested):
This is a good idea, but:
- This should go with the rest of the filters controlling which games are shown.
- Changing this filter should change an URL parameter so people can bookmark a link to the page with full games hidden.
- It would be better to change the database query so that only non-full games are returned.

philcampeau:
All good ideas. I wasn't thinking about people who want to hide full games all the time. 

I'll need to start this one almost from scratch at that point. I'll close this and revisit later.

PR #12690, opened 2026-05-26
[1828] allow choice of odd share in merger even when only holding president's cert

### LINKED ISSUE #8160: [1828] When merging, it should allow me to keep the more expensive share, even if it is part of a President's share.
https://18xx.games/game/85892?action=678

The Erie 30% merged with the B&M 20%. I receive two shares of the system.

Expected: I am given a choice of B&M or Erie share to hold onto, for later trading up into a share of the system (and I choose the B&M because its share price is higher).

Observed: The President's shares are converted, and I am left with the remaining Erie share.

Per the rules 7.5.2 - Step 4.

philcampeau:
Fixes #8160 

## Before clicking "Create"

### Explanation of Change

When a player holds only a president's certificate of one of the merging corporations, they weren't being offered a choice of which share to keep as their odd share.

The original code fetched candidate odd shares using .reject(&:president), meaning a president's cert was invisible to the logic. If a player held only a president's cert of one corp and a regular share of the other, one of the two variables would be nil, and the elsif condition merger_share && target_share would fail, silently skipping the choice prompt and falling through to the else branch.

I changed the share fetching lines to fall back to the president's cert if no regular share exists:

```
merger_share = entity.shares_of(@merger).reject(&:president).first || entity.shares_of(@merger).find(&:president)
target_share = entity.shares_of(@target).reject(&:president).first || entity.shares_of(@target).find(&:president)
```

This means the choice is now correctly offered to the player even when one of their holdings is solely a president's cert.
Second fix: When the player's selection is resolved, if they chose the corp where their only holding is a president's cert, that cert counts as 2 shares and can't be the odd share. So the selection is redirected to the other corp's share instead:

```
if @player_selection
  chosen_corp = @player_selection.include?(@merger.name) ? @merger : @target
  chosen_share = entity.shares_of(chosen_corp).reject(&:president).first
  @odd_share = chosen_share || (chosen_corp == @merger ? target_share : merger_share)
  @player_selection = nil
```

If the player picked a corp where they only have a president's cert (chosen_share is nil), the odd share falls back to the other corp's share, correctly treating the cert as 2 shares that will form a pair in the exchange.

### DIFF
--- a/lib/engine/game/g_1828/step/merger.rb
+++ b/lib/engine/game/g_1828/step/merger.rb
@@ -321,11 +321,13 @@ module Engine
             total_shares = entity.num_shares_of(@merger) + entity.num_shares_of(@target)
             return unless total_shares.odd?
 
-            merger_share = entity.shares_of(@merger).reject(&:president).first
-            target_share = entity.shares_of(@target).reject(&:president).first
+            merger_share = entity.shares_of(@merger).reject(&:president).first || entity.shares_of(@merger).find(&:president)
+            target_share = entity.shares_of(@target).reject(&:president).first || entity.shares_of(@target).find(&:president)
 
             if @player_selection
-              @odd_share = @player_selection.include?(@merger.name) ? merger_share : target_share
+              chosen_corp = @player_selection.include?(@merger.name) ? @merger : @target
+              chosen_share = entity.shares_of(chosen_corp).reject(&:president).first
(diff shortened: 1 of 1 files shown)


crericha:
As discussed in slack, when the player's odd share is coming from a bare presidents cert,  I want the director share traded for two single shares; not hand waved.

PR #12691, opened 2026-05-27
Send notification when minimum player count is reached

### LINKED ISSUE #12445: Add Notification when people join your open game
**What's your idea?**

Sometimes signups take a while, and if you're not constantly checking you're potentially disgruntling the players, but you have other things to do than constantly watch your game during signup...

philcampeau:
Fixes #12445

## Before clicking "Create"

### Explanation of Change

Much like my PR that added a notification when your game is full (#11627), this one adds a notification when the set minimum player count has been reached. 

I realized that not everyone is necessarily waiting for the game to fill up before starting. Knowing that the minimum threshold has been reached has its usefulness as well.

### DIFF
--- a/routes/game.rb
+++ b/routes/game.rb
@@ -25,6 +25,14 @@ class Api
             halt(400, 'Cannot join game because it is full') if users.size >= game.max_players
             halt(400, 'Cannot join because game has started') unless game.status == 'new'
 
+            if users.size == game.min_players - 1
+              # Generate a message to the game owner
+              type = 'Minimum player count reached'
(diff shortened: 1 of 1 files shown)


ollybh (changes requested):
If `min_players` and `max_players` are the same then this is going to send two notifications at once.

Better would be:
```
if users.size == @game.max_players - 1
  …
elsif users.size == @game.min_players - 1
  …
end

philcampeau:
Solid point Olly! Sent a fix now.

ollybh (approved):

Issue #11242, opened 2024-09-28
[18India] GIPR doesn't float when exchanging Bond


Ticketautomat1:
GIPR company sold 20% and 30% were issued by exchanging Bonds, but GIPR wasn't floated yet.

https://18xx.games/game/180273?action=448

PR #12695, opened 2026-05-27
[18 India] Fix GIPR bond conversion: manager and float

### LINKED ISSUE #11289: [18 India] can't convert bonds to shares of the GIPR
When converting bonds to shares of the GIPR, following error message appears and the turn resets:
"undefined method 'percent_of' for nil".

https://18xx.games/game/179328

The GIPR hasn't been floated

ddaybell:
Fixes #11289, #11242

## Before clicking "Create"

### Explanation of Change

Bond shares in GIPR are created outside the normal IPO system, which caused two bugs when players converted Railroad Bonds.

**#11289 - Crash on bond conversion if nobody was GIPR manager yet (undefined method 'percent_of' for nil):**
convert_bond_to_gipr called check_manager_change, which accessed gipr.owner.percent_of(...). If no normal GIPR share had been purchased yet, gipr.owner is nil and the game crashes. Fixed by calling make_manager on the converting player when GIPR has no manager yet, mirroring what normal share purchases already do.

**#11242 - GIPR doesn't float when exchanging bonds:**
Bond shares bypass the IPO, so @ipo_owner.percent_of(self) never decreases from bond conversions and the standard floated? check never fires. Fixed with a new check_gipr_float method in game.rb that sums share_holders directly - capturing both normal IPO sales and bond conversions - and explicitly sets floated and calls float_corporation when the 30% threshold is reached. The percent_to_float display on the charter was also incorrect for the same reason; fixed with an override in Corporation using the same share_holders approach.

Home token placement after a bond-triggered float is handled in process_choose in the step, following the same pattern as normal share purchases.

### Any Assumptions / Hacks

### DIFF
--- a/lib/engine/game/g_18_india/corporation.rb
+++ b/lib/engine/game/g_18_india/corporation.rb
@@ -116,6 +116,13 @@ module Engine
         def mangaged_company?
           presidents_share.percent.zero?
         end
+
+        def percent_to_float
+          return 0 if floated?
+
+          non_ipo_pct = share_holders.sum { |holder, pct| holder == ipo_owner ? 0 : pct }
+          [@float_percent - non_ipo_pct, 0].max
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


perwestling:
There is a template for PRs. Should you not be using that for your PRs?

Also, decide if [18India] or [18 India] should be used in the titles.

You can also label all 18 India PRs with the 18 India label. If they also requires pin or archive, those are good labels as well to use. There are other labels that could possibly also be used.

ddaybell:
I don't think I can add labels to PRs.  It looks like you have to have Write access to do that, which I as a Contributor do not have.  

I updated the Summary above to use the template.  I'll be consistent in using [18 India] in the titles.  I'll update any open PRs that don't conform.

PR #12707, opened 2026-05-28
[1832] Update Cotton behavior and update both cotton and port tokens.

jemiahlee:
Fixes some graphical stuff with these two tokens, plus changes the coal tile from looking like it takes an actual token.

Screenshot 2026-05-28 at 3 55 45 AM

Screenshot 2026-05-28 at 3 56 19 AM

Screenshot 2026-05-28 at 3 59 42 AM


jemiahlee:
The Atlanta tile is super busy, and I'd love to be able to place the cotton token explicitly in the lower right side. Unfortunately, without modification of some of the base system, I don't think this is possible without quite a bit of code, which didn't feel worth it.

ollybh (approved):

PR #12709, opened 2026-05-28
[18 India] Keep Track step blocking when P2/P3 abilities are available (#11362)

### LINKED ISSUE #11362: [18 India] Skips Token Placement before using private company
**What's your Game ID?** [[185624](https://18xx.games/game/185624)](https://gist.github.com/fcboyd/b36a314a337d33178577d86daf66ceb4)

**How to Reproduce the Bug?** 
- PNS does not have access to an open station in its route
- fcboyd played tile 81 on B15 
- After tile placement, PNS still does not have access to open station spot which results in the "place station token" action being skipped.
- fcboyd then uses Portugese EIC to place an extra straight yellow track on C16 to connect routes.
- PNS would like to place a station token in Mumbai, but cannot because option to place token has already been skipped.

**Which Browser / Platform?**

Chrome 130.0.6723.116

ddaybell:
## Summary

Fixes #11362 - token placement was silently skipped before P2 (Portuguese EIC) or P3 (Dutch EIC) could be used to create station access.

**Root cause:** After a corporation's normal tile lay, Engine::Step::Track#process_lay_tile called pass! immediately via super. This caused skip_steps to evaluate the Token step before SpecialTrack had a chance to run. If no station was accessible yet, Token got skipped - even though P2/P3 could have been used to lay the connecting track first.

**How it worked before:** The Track step auto-passed the moment the tile was laid and can_lay_tile? returned false. The game then jumped straight to Token (or skipped it), with no opportunity to use P2/P3's tile-lay ability at the 	rack timing.

**Fix:** Two changes to G18India::Step::Track:

1. **process_lay_tile**: No longer delegates to super. Instead inlines lay_tile_action + the 18India-specific post-processing, then only calls pass! if both can_lay_tile? **and** special_track_ability_available? are false. This keeps Track active (not passed) when P2/P3 is still owned.

2. **ctions**: Returns %w[pass] when can_lay_tile? is false but a company tile-lay ability is still available. This makes Track *blocking* (not just active) so skip_steps stops there instead of skipping Token.

**Result:** When P2/P3 is owned, the player sees a "Done (Track)" button after their normal tile lay. The 	rack timing window stays open, so P2/P3 is still accessible through the SpecialTrack step. After using P2/P3 (and it closes), Track becomes non-blocking, skip_steps fires, and the Token step is evaluated against the updated board - which now has the connection P2/P3 just created.

If the player clicks "Done (Track)" without using P2/P3, token placement is still skipped (their deliberate choice).

**No regressions:** All 7429 existing fixture examples pass.

## Test plan

- [x] Verify in a live game: corporation owns P2, lays normal tile, sees "Done (Track)" button, uses P2 to lay adjacent tile creating station access, then places token
- [x] Verify P3 (Dutch EIC upgrade) works the same way
- [x] Verify that corporations without P2/P3 still auto-pass Track and behave as before

?? Generated with [Claude Code](https://claude.com/claude-code)


perwestling (dismissed):
LGTM

Reviewed with AI assistance (Mistral). Excellent PR. Root cause clearly identified (Track step auto-passing before SpecialTrack), minimal fix confined to 18 India variant, all 7429 tests pass, well-commented code. Ready to merge. 🚀

PR #12710, opened 2026-05-28
[1850] Gant private should also allow laying Chicago tile

### LINKED ISSUE #12708: [1850] Can't use Gant Brothers Private to lay Chicago tile
Cant use Gant Brothers Private to lay Chicago tile

Game ID: 254645

I am playing game https://18xx.games/game/254645.
During Operating Round 4.2 I laid the first yellow tile and bought Gant Brothers private with MP to lay the second yellow tile in Chicado using the private power.
It does not give any error, but it also doesn't let me do any action. 

If I do not try to place the tile in Chicago, and instead try to place any other tile there is no problem.

I do not see anything in the rules stating I can not lay Chicago using the Gant power. Is this possibly a bug?

- Browser Version
  - On Chrome: Version 148.0.7778.179 (Official Build) (64-bit)
- Platform: Desktop

Thanks,

philcampeau:
Fixes #12708

## Before clicking "Create"

### Explanation of Change

The Gant Brothers Construction Company allows the owning corp to lay an additional yellow tile each turn for $30 plus any terrain costs.

Looks like they forgot to include the Chicago yellow tile in the list of applicable tiles. I checked the rules and there's nothing forbidding its use there.

### DIFF
--- a/lib/engine/game/g_1850/entities.rb
+++ b/lib/engine/game/g_1850/entities.rb
@@ -82,7 +82,7 @@ module Engine
             special: false,
             when: 'track',
             hexes: [],
-            tiles: %w[1 2 3 4 5 6 7 8 9 55 56 57 58 69],
+            tiles: %w[1 2 3 4 5 6 7 8 9 55 56 57 58 69 128],
             count_per_or: 1,
             cost: 30,
           ],


perwestling (approved):
LGTM

PR #12712, opened 2026-05-29
[18 India] Fix connection bonus fires on visited but unpaid cities

### LINKED ISSUE #11512: [18 India] Route Bonus given even if Cities are not counted with 4E train
**What's your Game ID?** https://18xx.games/game/196032?action=617

**How to Reproduce the Bug?**

 Ran the 4E train with auto routing and happened with the next OR with a 4E×3 train https://18xx.games/game/196032?action=741 i included the screenshot for that one too

Image
Image
Image
Image

**Which Browser / Platform?**

 132.0.6834.163 (Version officielle) (64 bits) Mobile

ddaybell:
## Summary
- `connection_bonus` was using `route.visited_stops` instead of the `stops` (paid stops) argument
- This caused the R+ route bonus to fire when both endpoint cities of a bonus pair were merely *visited* rather than counted among the train's paid cities
- A 4E train visiting Karachi and Chennai but paying for 4 other cities would incorrectly collect the R+80 bonus; now both cities must be among the paid stops

Fixes #11512

This will very likely require pinning games,  as it changes the amounts awarded for many E train runs. 

## Test plan
- [x] 18India fixtures: 43/43 pass
- [x] Manual: run a 4E train that visits but does not pay for both cities of a bonus pair; confirm R+ does not appear
- [x] Manual: run a 4E that pays for both cities of a bonus pair; confirm R+ still appears

### DIFF
--- a/lib/engine/game/g_18_india/game.rb
+++ b/lib/engine/game/g_18_india/game.rb
@@ -1031,8 +1031,8 @@ module Engine
           variable_city_stops.count * [max_non_variable_value, 0].max * train_multiplier
         end
 
-        def connection_bonus(route, _stops)
-          visited_location_names = route.visited_stops.map { |stop| stop.tile.location_name }.compact
+        def connection_bonus(_route, stops)
+          visited_location_names = stops.map { |stop| stop.tile.location_name }.compact
           return 0 if visited_location_names.count < 2
 
           # Delhi, Kochi => 100 [G8, G36]
--- a/spec/lib/engine/game/g_18_india/game_spec.rb
+++ b/spec/lib/engine/game/g_18_india/game_spec.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Engine::Game::G18India::Game do
+  let(:players) { %w[a b] }
+
+  subject(:game) { Engine::Game::G18India::Game.new(players) }
+
+  def make_stop(location_name)
+    tile = double('tile', location_name: location_name)
+    double('stop', tile: tile)
+  end
+
+  describe '#connection_bonus' do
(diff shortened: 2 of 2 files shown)


lib/engine/game/g_18_india/game.rb
```diff
@@ -1032,8 +1031,8 @@ def variable_city_revenue(route, stops)
           variable_city_stops.count * [max_non_variable_value, 0].max * train_multiplier
         end
 
-        def connection_bonus(route, _stops)
-          visited_location_names = route.visited_stops.map { |stop| stop.tile.location_name }.compact
+        def connection_bonus(_route, stops)
+          visited_location_names = stops.map { |stop| stop.tile.location_name }.compact
```
perwestling, line 1035: **Strongly recommend**: Add state-based fixture test

Since 18India has existing fixtures, please add a test case that:
1. Runs a 4E train that visits but does not pay for both cities of a bonus pair
2. Confirms R+ does not appear
3. Runs a 4E that pays for both cities
4. Confirms R+ appears

**Alternative**: If setting up the game state is complex, a unit test for `connection_bonus` method verifying it uses the `stops` parameter (not `route.visited_stops`) would also work.

perwestling (dismissed):
## PR Review: #12712 - [18India] Fix connection bonus fires on visited but unpaid cities

---
### ✅ Strengths
- Fixes documented bug (#11512)
- Minimal, surgical change (2 lines)
- Correct root cause: was using `route.visited_stops` instead of paid `stops`
- All 18India fixtures pass (43/43)

---
### 💡 Strong Recommendation
Since 18India has existing fixtures, **please add a test case** for this fix. State-based fixture preferred, but a unit test for the `connection_bonus` method is also acceptable.

---
### ✅ Verdict: **Approve**

The fix is correct. Adding a test is strongly recommended but not blocking.

PR #12701, opened 2026-05-27
[1862 USA & Canada] PR 14: stock buyback + bond repayment (Step::StockBuyback, Step::RepayBond)

neutronc:
Fixes # (no issue — prealpha game implementation)

## Before clicking "Create"

- [ ] Branch is derived from the latest `master`
- [ ] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [x] Code passes linter with `docker compose exec rack rubocop -a`
- [x] Tests pass cleanly with `docker compose exec rack rake`

### Explanation of Change

Sprint 14 — Bond (Schuldschein) + Stock Buyback (Aktienrückkauf) full lifecycle.

- **feat: `Step::StockBuyback`** — Available in the OR's early slot from Phase 3 onward when the corporation has market shares and has not yet done a buyback. Pays each holder 50% of market price, halves IPO certs, issues the bond, reduces the director's cert limit by 1 permanently.
- **feat: `Step::RepayBond`** — Available after Dividend when a bond is outstanding. Deducts the bond amount from the corporation treasury and clears the debt flag (cert-limit penalty remains permanent).
- **fix: sell movement** — Shares sold below a full 10% lot (e.g. a 5% cert after halving) no longer trigger a price drop; `down_per_10` moves once per 10% sold.
- **fix: restore monopoly fee on director share buy above 60%** — Fee deducted at buy time in `Step::BuySellParShares`.

### Screenshots

N/A — logic changes only.

### Any Assumptions / Hacks

None.

PR #12713, opened 2026-05-29
[1862 USA & Canada] PR 15: TOR tile phase/color bypass + president share display

neutronc:
Fixes # (no issue — prealpha game implementation)

## Before clicking \"Create\"

- [ ] Branch is derived from the latest \`master\`
- [ ] Add the \`pins\` or \`archive_alpha_games\` label if this change will break existing games
- [x] Code passes linter with \`docker compose exec rack rubocop -a\`
- [x] Tests pass cleanly with \`docker compose exec rack rake\`

### Explanation of Change

Stacked on PR 14 (#12701). Three items in this PR:

**1. TOR tile phase/color bypass + president share display**
- Overrides \`upgrades_to_correct_color?\` and \`tile_valid_for_phase?\` so the TOR private can place its custom gray tile regardless of current phase or tile color restrictions.
- Adds \`status_str\` to display the president's share percentage on the corporation panel.

**2. Rules verification (confirmed from physical rulebook)**
- TP (Texas & Pacific) has no connection bonus targets — \`CORP_BONUSES['TP']\` entry removed. ORN and GMO also confirmed absent.
- Operating rounds: 2 per SR for all phases (was incorrectly 1 for phase 2 and 3 for phases 5–8).
- 6/6E trains confirmed never rust ✓. \`rusts_on: '6'\` for 3/3E correct — \`train.sym\` is set from the base name at initialisation and is not mutated when a variant is selected, so 6E purchases also trigger the rust ✓.

**3. BUG-014: V/P/S/L bonus icon placement on B2/D2/G3/I5**
Six corps share the four V/P/S/L bonus hexes. The default placement algorithm puts all six icons in a single row at \`PP_WIDE_TOP_CORNER\` (y=−65), which overflows the hex boundary on map-edge hexes (B2), gets covered by the adjacent pre-printed C1 gray tile (D2), or overlaps the city label (G3/I5).

Fix: split into two groups of three pinned to the right side of each hex:
- CP / NYH / NYC → \`loc:'3.5'\` (\`PP_WIDE_UPPER_RIGHT_CORNER\`, x:52 y:−25)
- SP / NP / CN → \`loc:'4.5'\` (\`PP_WIDE_BOTTOM_RIGHT_CORNER\`, x:52 y:+25)

### Any Assumptions / Hacks

None.


neutronc:
Closing to keep a clean one-PR-at-a-time review chain. Will resubmit once the preceding PR in the stack is merged.

PR #12714, opened 2026-05-30
[18 India] Fix hand share purchase price to use IPO price

### LINKED ISSUE #11114: [18India] Purchase from hand amount checking is wrong
**What's your Game ID?**
172417

**How to Reproduce the Bug?**
1. Run a company and pay out (BNR)
2. Sell off all cert
3. Buy from hand for director cert
4. Will not let you buy for 164 (BNR is 82 par), message shown not enough cash because BNR is at 112 now.
5. Normal rules is to treat hand same price as ipo this this is a bug with calculation for director cert. This also applied to normal cert on hand. Do not have issue with cert on ipo

ddaybell:
## Summary
- can_buy_from_hand? and can_buy_company? affordability checks for hand share certificates now use IPO/par price explicitly via a new hand_price(company) helper, rather than company.value which can reflect market price when the underlying share's ownership changes
- hand_price computes par_price * num_shares for :share and :president type certificates; falls back to company.value for other types or if par_price is unavailable

## Test plan
- [x] All 185 existing fixture tests pass
- [x] Verify a player can buy a hand share certificate at IPO price when the corporation's market price has moved above or below the original IPO price

image
image

Closes #11114

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/engine/game/g_18_india/step/sell_once_then_buy_certs.rb
+++ b/lib/engine/game/g_18_india/step/sell_once_then_buy_certs.rb
@@ -170,7 +170,8 @@ module Engine
             return false if @game.in_ipo?(company) && !can_buy_from_ipo?(entity, company)
             return false if current_entity.hand.include?(company) && !can_buy_from_hand?(entity, company)
 
-            (available_cash(entity) >= company.value)
+            price = current_entity.hand.include?(company) ? hand_price(company) : company.value
+            (available_cash(entity) >= price)
           end
 
           def can_buy_from_market?(entity, company)
@@ -196,8 +197,18 @@ module Engine
                 prior.name == company.name
               end
             else
-              (available_cash(entity) >= company.value)
+              (available_cash(entity) >= hand_price(company))
             end
+          end
+
+          def hand_price(company)
+            return company.value unless %i[share president].include?(company.type)
+
+            share = company.treasury
+            par = share.corporation.par_price
+            return company.value unless par
+
+            (par.price * share.num_shares(ceil: false)).ceil
           end
 
           def can_buy_from_ipo?(entity, company)
--- a/spec/lib/engine/game/g_18_india/game_spec.rb
+++ b/spec/lib/engine/game/g_18_india/game_spec.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Engine::Game::G18India::Game do
+  describe '18India_game_end_bank' do
+    # Action 403: Player 1 is in buying phase, holds EIR (share) in hand
+    # EIR par=100, market=126 (market is above IPO price)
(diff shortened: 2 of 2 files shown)


lib/engine/game/g_18_india/step/sell_once_then_buy_certs.rb
```diff
@@ -196,10 +197,20 @@ def can_buy_from_hand?(entity, company)
                 prior.name == company.name
               end
             else
-              (available_cash(entity) >= company.value)
+              (available_cash(entity) >= hand_price(company))
             end
           end
 
+          def hand_price(company)
```
perwestling, line 204: **Strongly recommend**: Add state-based fixture test

Since 18India has existing fixtures, please add a test that verifies:
1. A player can buy a hand share certificate at IPO price when corporation's market price is above IPO price
2. A player can buy a hand share certificate at IPO price when corporation's market price is below IPO price
3. Non-share/president certificates still use `company.value`

This prevents regression of the price calculation logic.

perwestling (dismissed):
## PR Review: #12714 - [18India] Fix hand share purchase price to use IPO price

---
### ✅ Strengths
- Fixes documented issue (#11114)
- Minimal, focused change (13 lines added, 2 removed)
- Correct root cause: was using `company.value` (market price) instead of IPO/par price for hand certificates
- All 18India fixtures pass (43/43)
- Clean helper method (`hand_price`) with proper fallbacks

---
### 💡 Strong Recommendation
Since 18India has existing fixtures, **please add a test case** verifying:
- Hand share certificates use IPO price, not market price
- Fallback to `company.value` for other certificate types
- Behavior is correct when market price differs from IPO price

---
### ❓ Questions
- Have you verified the manual test case mentioned in the PR description?

---
### ✅ Verdict: **Approve**

The fix is correct and addresses a real bug. Adding a test is strongly recommended but not blocking.

ddaybell:
Added `spec/lib/engine/game/g_18_india/game_spec.rb` with four state-based tests loaded from the existing `18India_game_end_bank.json` fixture.

**Market above IPO price (action 403)** - Player 1 holds an EIR 10% share cert in hand; EIR par=100, market=126:
- `hand_price` returns 100 (par), not 126 (market)
- A player with exactly 100 cash can buy from hand

**Market below IPO price (action 276)** - Player 2 holds a WIP 10% share cert in hand; WIP par=76, market=67:
- `hand_price` returns 76 (par), not 67 (market)
- A player with 76 cash (par price) can buy from hand
- A player with only 67 cash (market price) **cannot** buy - the IPO price is owed regardless of where the market has moved

The "market below IPO" case is the sharpest regression test: it verifies that `can_buy_from_hand?` requires the *par* price even when the market has fallen below it, so having just enough cash to afford the current market price is not sufficient.

PR #12717, opened 2026-05-30
[1862 USA & Canada] PR 16: rules verification — TP bonuses, operating rounds

neutronc:
Fixes # (no issue — prealpha game implementation)

## Before clicking "Create"

### Explanation of Change

Rules verified against physical rulebook. Two corrections:

**TP (Texas & Pacific) has no connection bonus targets.**
The `CORP_BONUSES` entry for TP (El Paso + V/P/S/L) was a placeholder pending rulebook confirmation. Confirmed no bonus targets — entry removed. ORN and GMO also confirmed absent (no change needed).

**Operating rounds: 2 per SR throughout all phases.**
All phases now set to `operating_rounds: 2`. Previously phase 2 had 1 OR and phases 5–8 had 3 ORs.

Also confirmed correct (no code change):
- 6/6E trains never rust ✓
- `rusts_on: '6'` for 3/3E correctly handles both `'6'` and `'6E'` purchases — `train.sym` is set from the base name at initialization and is not mutated when a variant is selected, so the rust check fires on either purchase ✓
- `SLC_FIRST_CONNECTION_BONUS = 20` already correct in prior PR ✓

### Screenshots

N/A

### Any Assumptions / Hacks

None.


neutronc:
Closing — these fixes should be committed to the preceding branch in the PR chain (pG), not submitted as a separate PR while the chain is still under review.

PR #12721, opened 2026-05-31
[18CZ] Rename kk's name to 'kaiserliche und königliche Staatsbahn'

tvedeane:
## Implementation Notes

The original name used the abbreviation. To be consistent with other companies I added full name.

### DIFF
--- a/lib/engine/game/g_18_cz/entities.rb
+++ b/lib/engine/game/g_18_cz/entities.rb
@@ -1803,7 +1803,7 @@ module Engine
               float_percent: 50,
               float_excludes_market: true,
               sym: 'kk',
-              name: 'kk Staatsbahn',
+              name: 'kaiserliche und königliche Staatsbahn',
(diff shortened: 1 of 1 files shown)


Zwergenpunk:
How does it look on the card?

tvedeane:
I followed this image https://boardgamegeek.com/image/2200617/18cz (upper-right corner)

Zwergenpunk:
This is from the rules. The image from you is not the original game. I do not have a copy at home. Doe it fit reasonable in the game card on 18xx?

tvedeane:
I don't have a copy at home unfortunately. The table in the rules seems also having the same mistake.

Zwergenpunk:
I found an unboxing video. The real cards also have kk Staatsbahn on it.

tvedeane:
Okay, thanks for checking. Then I guess I'll close the PR still wondering why it's like that :D

PR #12722, opened 2026-05-31
[1824/1858/1866/18Tokaido] Unlimited trains

### LINKED ISSUE #12719: [1824/1858/1866/18Tokaido] Unlimited trains for games with overridden `num_trains` method
> Several games already have their own versions of the `num_trains` method, typically looking up the train count in a constant, chosen according to game options. For these you can't just switch the train count to `unlimited`.
> 
> - ​1824.
> - 1858 family.
> - ​1866.
> - 18Tokaido.
> 
> Fixing these games probably just needs these two different cases for `num_trains` to be split into different methods.
> 1. Choosing the train count based on game options.
> 2. Determining the number of train objects to create for each train rank, as called from `init_train_handler`. 

 _Originally posted by @ollybh in [#12237](https://github.com/tobymao/18xx/issues/12237#issuecomment-4586260067)_

ollybh:
## Implementation Notes

Adds support for unlimited trains to 1824 (base game and Cisleithania variants), the 1858 family (Iberia, Switzerland and India), 1866 and 18Tokaido.

### Explanation of Change

Commit 287fcbae added support for unlimited trains. Part of this involved using the `num_trains` method to calculate how many train objects needed to be initially created for an unlimited rank of train, normally one, but more if there is an event that fires on the nth train being bought.

1824, 1858, 1866 and 18Tokaido already had their own versions of `num_trains`. They are using this method to allow the number of trains to be calculated from optional rules selected, or player counts.

These two different uses of `num_trains` don't work together, preventing these games from simply changing the train rank count to `unlimited` in their definitions.

This splits the uses, creating a new `train_rank_count` method that is called from `init_train_handler` to determine how many Train objects should be created for each type of train.

This also changes the Train initialiser to take `unlimited` as a parameter, instead of assuming that the `trains` hash will contain a `:num` key.

Finally, it changes 1848 to remove that game's custom `init_train_handler` method. To do this the `DEPOT_CLASS` constant is set for that game.

Fixes #12719. Works towards fixing #12237.

This supersedes PR #12595 and parts of #12593.

### Any Assumptions / Hacks

An alternate approach to adding the `unlimited` parameter to the Train class would be to call `num_trains` (or something similar) from the view classes and drop the `unlimited` attribute from a train object. That might make more sense (it's the rank of trains that's unlimited, not the individual train) but would have been a bigger change.

## Before clicking "Create"


perwestling:
I checked the 1824 code and it looked fine. (Although 1824::Depot probably could be refactored a little bit. But as it works I suppose it is fine. It is over 5 years since I created it.)

PR #12725, opened 2026-06-01
[1835] implement the formation of the Preußen

kepakiano:

### Explanation of Change

Implements the formation of the Preußen, including:
- optional formation the moment the first 4T is bought
- optional formation at the beginning of each OR and SR after the first 4T has been bought
- mandatory formation the moment the first 4+4T is bought
- optional folding in of minors and Prussian companies sorted by players
- mandatory folding in once the first 5T is bought
- adding the Preußen to the current OR operating order depending on the converted papers

The `MinorExchange` class is heavily influenced by the implementation of 1837 (although no conversion during an OR is possible there, only at the beginning unless I missed something).

This change did become part of a larger change that can be used for a first legal playthrough of 1835. This playthrough will become the first fixture (which I already have locally). I tried to keep everything out of this PR, but accidentally added some lines (e.g. number of tile lays), please excuse that.


lib/engine/game/g_1835/game.rb
```diff
@@ -134,15 +135,29 @@ def self.plus_train_distance(distance)
 
         TRAINS = [{ name: '2', distance: 2, price: 80, rusts_on: '4', num: 9 },
                   { name: '2+2', distance: plus_train_distance(2), price: 120, rusts_on: '4+4', num: 4 },
-                  { name: '3', distance: 3, price: 180, rusts_on: '6', num: 4 },
+                  { name: '3', distance: 3, price: 180, rusts_on: '6', num: 4, events: [{ 'type' => 'can_buy_trains' }] },
```
ollybh: Buying trains between companies is better implemented using a phase status instead of an event. Look at 1822 for an example of how this is done.

lib/engine/game/g_1835/round/operating.rb
```diff
@@ -0,0 +1,43 @@
...
+
+          def pending_tokens
+            @pending_tokens ||= []
+          end
+
+          def any_conversion_choice_available?
+            # Owner of 2 has the choice to form the PR
+            return true if @game.pr_can_form && !@game.corporation_by_id('PR').floated?
+            # PR has already been formed and not all minors/companies have been converted yet
+            if @game.corporation_by_id('PR').floated? && !(@game.minors + @game.prussian_companies).reject(&:closed?).empty?
```
ollybh: You've stored a reference to the PR, so you should use it here:
```suggestion
            return true if @game.pr_can_form && !@game.prussian.floated?
            # PR has already been formed and not all minors/companies have been converted yet
            if @game.prussian.floated? && !(@game.minors + @game.prussian_companies).reject(&:closed?).empty?
```

lib/engine/game/g_1835/round/operating.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require_relative '../../../round/operating'
+module Engine
+  module Game
+    module G1835
+      module Round
+        class Operating < Engine::Round::Operating
+          def setup
```
ollybh, line 9: This method is mostly cut'n'pasted from Engine::Round::Operating.setup. To avoid this duplication, could you add the `any_conversion_choice_available?` to `Game.payout_companies`? Then this method could just be calling `super` and setting the `@game.conversion_choice_during_or` variable.

ollybh (changes requested):
Only a few things to change, nothing major.

lib/engine/game/g_1835/game.rb
kepakiano: Ah, I was wondering what the preferred way was. Since it's not fully implemented in this PR anyway, I took out all the references to cross buying and will put it into its own PR.

lib/engine/game/g_1835/round/operating.rb
kepakiano: absolutely, done

lib/engine/game/g_1835/round/operating.rb
kepakiano: That makes sense. I guess I was tired when I implemented that :) Thanks!

lib/engine/game/g_1835/step/minor_exchange.rb
```diff
@@ -0,0 +1,203 @@
...
+            if @game.pr_can_form && !@game.prussian.floated? && !@round.declined.include?(@game.berlin_potsdamer_bahn)
+              return [@game.berlin_potsdamer_bahn]
+            end
+            return exchangeables_sorted_by_player if @game.prussian.floated?
+
+            []
+          end
+
+          def active?
+            active_entities.any?
```
ollybh:
```suggestion
            !active_entities.empty?
```

lib/engine/game/g_1835/step/minor_exchange.rb
```diff
@@ -0,0 +1,203 @@
...
+            end
+          end
+
+          # The rules forbid that anyone profits from a paper twice.
+          # Ergo: If 2 has already operated, PR cannot operate in the same OR or the (former) owner of 2 would profit
+          # from their paper twice. The only that let's the PR operate in the same OR it was formed is therefore:
+          # - 1 buys the train that triggers the conversion and
+          # - neither it nor the prussian companies HB and BB (since they paid out their revenue) are converted
+          def pr_should_operate_this_or?
+            no_prussian_company_or_minor_1_converted = (@round.converted & (@game.prussian_companies + [minor_1])).empty?
```
ollybh: I prefer using the intersect? method: `!a1.intersect?(a2)` instead of `(a1 | a2).empty?`.
```suggestion
            no_prussian_company_or_minor_1_converted = !@round.converted.intersect?(@game.prussian_companies + [minor_1]))
```

lib/engine/game/g_1835/game.rb
```diff
@@ -287,6 +316,130 @@ def hamburg_ferry?(route)
...
+          end
+
+          # Preußen already has a token in Berlin and the rules forbid having more than one token per hex
+          unless minor.id == '5'
+            token = minor.tokens.first
+
+            # make sure the first token (= home token) gets used or other methods might behave unexpectedly later, e.g.
+            # "maybe_place_home_token" called when buying shares
+            new_token = minor.id == '2' ? prussian.tokens.first : Token.new(prussian)
+            prussian.tokens << new_token
```
ollybh, line 412: This is creating new tokens for the Prussian. It starts with five tokens, will this not give it too many?

ollybh (commented):
Is the Prussian going to end up with too many tokens?

If it does get the right number of tokens after the exchanges, it might still be better to not create new tokens. Players might look at the Prussian's charter and see that it has five tokens, then be surprised that it can get more than this.

lib/engine/game/g_1835/game.rb
kepakiano: I didn't want to think about the tokens too much in this merge request yet, just had to make sure it works (which is why the home token is placed when the Prussian is formed). I figure I'll do that when I implement the cost of tokening. I'm also not yet sure that the correct number of tokens for the Prussian actually is. The German rules you on BGG say, the Prussian has one token. The English rules say seven. The game material itself has eight (minus one for the stock market), seen here: https://boardgamegeek.com/image/221486/1835
I vaguely remember when playing OTB that the Prussian had two tokens it could use freely, meaning the German rules on BGG are wrong. So I guess I'll do some digging.

regarding the question whether it's better to create new tokens here or just pace tokens the Prussian has: I guess I could also give the Prussian exchange tokens similar to 1822. Point being that they cannot be used for anything else. However, since each of these tokens has a specific minor they get exchanged for and cannot be transformed into regular tokens like in 1822, it does seem like overhead to me. Anybody who's read the rules and then sees only one or two tokens on the Prussian will understand that the tokens placed in exchange for the tokens of the minors are not taken away from the one or two tokens it has. What do you think?

ollybh: It's fine if you want to leave this to a later PR.

Like you say, exchange tokens might be a good way to implement this.

In general, I suggest that you try and reproduce what is in the physical copies of games. If the PR comes with seven (or eight) tokens on its charter, then give it the same number of tokens.

ollybh (approved):

PR #12727, opened 2026-06-02
[1835] fix percentages of OL and MS

kepakiano:

### Explanation of Change

`forced_share_percent` is not parsed from the corporation object, it's just something you set later, so it gets moved to the setup of the game here.


ollybh (approved):
It's probably just an oversight that the `forced_share_percent` attribute isn't set by the Corporation class's `initialize` method. You could submit a patch to add it.

PR #12734, opened 2026-06-05
[1835] properly label Berlin so it can be upgraded from yellow to green

kepakiano:

## Before clicking "Create"

### Explanation of Change

The yellow Berlin tile has two cities, the green one merges them into one city with three stations, so we have to allow that to happen.

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -265,6 +265,11 @@ module Engine
           floated.sort + not_floated + others
         end
 
+        def upgrade_ignore_num_cities(from)
+          # Two stations in Berlin get merged into one with three spots
+          from.hex.id == 'E19' && from.color == :yellow
+        end
+
         def revenue_for(route, stops)
           super + (hamburg_ferry?(route) ? -10 : 0)
         end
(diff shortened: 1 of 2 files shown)


neutronc:
why is this necessary ? or different between normal city tile upgrades where we move from 2 to 3 or 3 to four city slots ?

kepakiano:
> why is this necessary ? or different between normal city tile upgrades where we move from 2 to 3 or 3 to four city slots ?

Because of this line: https://github.com/tobymao/18xx/blob/db78c3ffdbf15a5bb540b0309db0c2f37e5ed2e8/lib/engine/game/base.rb#L1845

Berlin in yellow has no label and the number of cities goes down from 2 to 1. So in order for this line to not return false, the number of cities must be ignored.

neutronc:
Berlin has a label ptinted in the bordo.

See the photo attached 


First german edition

kepakiano:
> Berlin has a label ptinted in the bordo.
> 
> See the photo attached 
> 
> First german edition

Well, that simplifies things :) Thank you! I don't have a physical copy here, so I couldn't check there and I just worked with the map that was already implemented. This is much cleaner now.

perwestling (approved):
LGTM

ollybh (approved):

PR #12735, opened 2026-06-05
[1835] allow upgrading green XX tiles to brown X tiles

kepakiano:

## Before clicking "Create"

### Explanation of Change

The green tiles for twin-cities have an XX label, but the brown ones only one X, so we need to handle that.

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -265,12 +265,6 @@ module Engine
           floated.sort + not_floated + others
         end
 
-        def upgrades_to_correct_label?(from, to)
-          return to.label&.to_s == 'X' if from.color == :green && from.label&.to_s == 'XX'
-
-          super
-        end
-
         def revenue_for(route, stops)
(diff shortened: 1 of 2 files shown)


perwestling (approved):
LGTM

ollybh (commented):
Using a `future_label` on these tiles might be a better choice. That gives the players a visual reminder that the label will be changing for the upgrades.

kepakiano:
> Using a `future_label` on these tiles might be a better choice. That gives the players a visual reminder that the label will be changing for the upgrades.

I think you're right. It's a little more code, but it works:
image

There is no way to just augment the definition of an already existing tile, is there? I'd love to just append the `future_label`.

ollybh:
> There is no way to just augment the definition of an already existing tile, is there? I'd love to just append the `future_label`.

You could override `init_tiles` and do something to add the future labels to these tiles. But I think redefining them for this game is fine.

ollybh (approved):

PR #12744, opened 2026-06-06
[1835] let major corps lay two yellow tiles during yellow phase

kepakiano:

## Before clicking "Create"

### Explanation of Change
Nothing much to say, majors get to lay two yellow tiles during the yellow phase. If you're not a major during the yellow phase you lay or upgrade one tile.

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -165,8 +165,8 @@ module Engine
 
         CORPORATION_BLOCKS = [%w[BY SX], %w[BA WT HE PR], %w[MS OL]].freeze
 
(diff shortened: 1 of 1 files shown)


ollybh (commented):
Could you change the phase check to a phase status test? This would add information to the phase table, showing players which phase(s) the major companies can lay two tiles in.

kepakiano:
> Could you change the phase check to a phase status test? This would add information to the phase table, showing players which phase(s) the major companies can lay two tiles in.

done :) I also added the other tile lay to the status so that the user doesn't ask themself what happens once the phase with two tile lays is over.

ollybh (approved):

PR #12745, opened 2026-06-06
[core] read forced_share_percent from options

kepakiano:
refers to no issue, but it was mentioned in https://github.com/tobymao/18xx/pull/12727 that not reading `forced_share_percent` from the options was probably an oversight and multiple games seem to be assuming that it gets properly set. There is code setting this in the `entities.rb`for 1822 Africa, 1854, and 1847.

## Before clicking "Create"


lib/engine/corporation.rb
```diff
@@ -78,6 +78,7 @@ def initialize(sym:, name:, **opts)
       @reservation_color = opts[:reservation_color]
       @price_percent = opts[:price_percent] || @second_share&.percent || (@presidents_share.percent / 2)
       @price_multiplier = (@second_share&.percent || (@presidents_share.percent / 2)) / @price_percent
+      @forced_share_percent = opts[:@forced_share_percent]
```
perwestling: **Blocking issue**: Typo in option key

The key should be `:forced_share_percent` not `:@forced_share_percent`:

```ruby
@forced_share_percent = opts[:forced_share_percent]
```

The `@` prefix in the symbol key is incorrect. Compare with other options in the same method:
- `opts[:price_percent]`
- `opts[:reservation_color]`
- `opts[:treasury_as_holding]`

All use symbols without the `@` prefix. This typo will cause the option to never be read from entities.

perwestling (changes requested):
## PR Review: #12745 - [core] read forced_share_percent from options

---
### ✅ Strengths
- Fixes a real issue: `forced_share_percent` option wasn't being read from entity definitions
- Consolidates code: Removes hardcoded settings in g_1835/game.rb
- Aligns with existing patterns: 1822 Africa, 1854, 1847 already use this option
- All tests pass

---
### 🚨 Blocking Issue
- **Line 81 of corporation.rb**: Typo in option key (`:@forced_share_percent` → `:forced_share_percent`)

---
### ⚠️ Concerns
- **Missing label**: PR description says to add `pins` or `archive_alpha_games` label if this breaks existing games. Since this is a core change to Corporation initialization, please add the appropriate label.

---
### 💡 Suggestions
- Add tests verifying `forced_share_percent` is read correctly from options
- Consider if other options might have similar issues

---
### ❓ Questions
- Have you tested that corporations in 1835, 1822 Africa, 1854, and 1847 still work correctly with this change?

---
### ❌ Verdict: **Request changes**

The typo on line 81 of `corporation.rb` must be fixed before merge. Once corrected and the label is added, this will be ready to approve.

kepakiano:
@perwestling what do you think? I don't know any of the other affected games well. Could reading this flag lead to a change in their default share percentage (e.g. from 20 to 10) and thus in wrong amounts of money being transferred?
Ah, damn, I'd relied on the other games having tests. But I see now that 1854 has no tests, the 1822Africa tests don't cover money and the 1847 fixtures don't have explicit tests.

Nevermind any of that, I see the pipeline failing with 1822Africa tests being red. Do you know if Alexander Vansach is still active? He seems to be the original developer of 1822Africa and I'd like to talk to him or someone else about these fixtures.

PR #12746, opened 2026-06-06
[1835] allow buying trains from other corporations and fix ebuy

kepakiano:

## Before clicking "Create"

### Explanation of Change

As mentioned [in another PR](https://github.com/tobymao/18xx/pull/12725#discussion_r3343010676), allowing cross buy should be done via the status of the phase.

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -48,6 +48,8 @@ module Engine
 
         TOKEN_PLACEMENT_ON_TILE_LAY_ENTITY = :owner
 
+        EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false
+
         MARKET = [['', '', '', ''] + %w[132 148 166 186 208 232 258 286 316 348 382 418],
                   ['', ''] + %w[98 108 120 134 150 168 188 210 234 260 288 318 350 384],
                   %w[82 86 92p 100 110 122 136 152 170 190 212 236 262 290 320],
@@ -75,6 +77,7 @@ module Engine
             on: '3',
             train_limit: { minor: 2, major: 4 },
             tiles: %i[yellow green],
+            status: ['can_buy_trains'],
             operating_rounds: 2,
           },
           {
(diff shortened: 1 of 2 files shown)


lib/engine/game/g_1835/step/buy_train.rb
```diff
@@ -5,6 +5,12 @@ module Game
     module G1835
       module Step
         class BuyTrain < Engine::Step::BuyTrain
+          def buyable_trains(_entity)
+            return super if @game.phase.status.include?('can_buy_trains')
+
+            super.select(&:from_depot?)
+          end
+
```
ollybh: Instead of overriding this method, you should put this check in `Game.can_buy_train_from_others?`.

ollybh (changes requested):
The phase status is the right way to specify when trains can be bought across, but this should be tested in `Game.can_buy_trains_from_others?`, not `Step::BuyTrain.buyable_trains`.

lib/engine/game/g_1835/game.rb
ollybh, line 76: This is code you've inherited, but do you think these phases could be simplified to just phases 1, 2 and 3? Train events should be able to take care of all the things that happen within these phases.

This could be done in a future PR, it's just something I've noticed whilst looking at this one.

lib/engine/game/g_1835/step/buy_train.rb
kepakiano: Done! Since you mentioned it, I had done it the way 1822 did it. Is there general wisdom as to which game is implemented the "cleanest", so that it can serve as reference? So far I've looked at all kinds of games to see how they did things I needed.

lib/engine/game/g_1835/game.rb
kepakiano: That's a good idea. That would also resemble the phase overview better that comes with the game. I'll put it on my todo list.

ollybh (changes requested):

ollybh (changes requested):
You need to add `can_buy_trains` to the `STATUS_TEXT` hash. Without that this status is not shown in the phase table on the info page.

Other than that, everything looks good.

kepakiano:
> You need to add `can_buy_trains` to the `STATUS_TEXT` hash. Without that this status is not shown in the phase table on the info page.
> 
> Other than that, everything looks good.

Dang, I thought I had. Thanks. Must've gotten lost when switching branches and doing partial add of files. Thanks for catching that!

ollybh (approved):

PR #12752, opened 2026-06-08
[18ZOO] Display stock market metadata

### LINKED ISSUE #12723: 18ZOO stock market meta data fix
**What's your idea?**
Display dividend, appreciation threshold in stockmarket of 18ZOO

Let us know your proposed changes.
Add "info" text to 2D stock market (like zigzag), populate 18Zoo magic numbers here. 
**Are there related ideas?**

**Did you join the Slack?**

carlorusso1984:
Closes #12723

Change from:
image
To:
image

## Summary
- attach appreciation thresholds, dividends, and president bonuses to 18ZOO share-price metadata
- render metadata in enlarged 2D stock-market cells while preserving ledge arrows
- safely handle sparse 2D markets and retain their existing layout
- add regression coverage for 18ZOO market prices, types, and metadata

## Shared renderer audit
- audited the shared `grid_2d` renderer across 60 two-dimensional stock-market definitions; one-dimensional, zigzag, and hexagonal markets use separate renderers
- confirmed that 18ZOO is the only game assigning `SharePrice#info`, so only 18ZOO selects the enlarged metadata layout
- confirmed through DOM measurements that standard markets retain 50x50 cells while 18ZOO uses 70x70 cells
- identified 20 sparse 2D market definitions and verified that `flatten.compact` safely ignores empty cells, addressing the failure mode of the reverted 2023 implementation
- manually checked representative layouts in the browser:
  - 1880: sparse stepped market
  - 1822: large 15-row market with 39 empty cells
  - 1866: very wide 4x37 market
  - 18MS: compact market without empty cells
- all sampled markets preserved their existing dimensions, ledge arrows, colors, scrolling, and legends

## Testing
- `docker compose exec rack bundle exec rspec spec/lib/engine/game/g_18_zoo/game_spec.rb` (179 examples, 0 failures)
- `docker compose exec rack bundle exec rubocop lib/engine/game/g_18_zoo/game.rb assets/app/view/game/stock_market.rb spec/lib/engine/game/g_18_zoo/game_spec.rb`
- `git diff --check`
- manually verified the Market tab in local 18ZOO Map C
- manually verified 1880, 1822, 1866, and 18MS through the shared 2D renderer

### DIFF
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -82,6 +82,17 @@ module View
         width: "#{WIDTH_TOTAL - (2 * PAD) - (2 * BORDER)}px",
       }.freeze
 
+      PRICE_STYLE_2D_INFO = {
+        display: 'flex',
+        justifyContent: 'space-between',
+        fontSize: '65%',
+        lineHeight: '1',
+        position: 'absolute',
+        bottom: "#{PAD}px",
+        left: "#{PAD}px",
+        right: "#{PAD}px",
+      }.freeze
+
       def box_style_1d
         {
           position: 'relative',
@@ -399,6 +410,8 @@ module View
       end
 
       def grid_2d
+        has_price_info = @game.stock_market.market.flatten.compact.any?(&:info)
+
         # Need to peek at row below to know if sitting on ledge.
         @game.stock_market.market.push([]).each_cons(2).each_with_index.map do |rows, row_i|
           row_prices, next_row = rows
@@ -424,20 +437,44 @@ module View
                 arrow = ''
               end
 
+              if has_price_info && !arrow.empty?
+                align.delete(:bottom)
+                align[:top] = "#{PRICE_HEIGHT}px"
+              end
+
               first_price = false
 
-              h(:div, { style: cell_style(@box_style_2d, price.types) }, [
+              elements = [
                 h('div.xsmall_font', price.price),
                 h(:div, tokens),
                 h(:div, { style: { color: '#00000060', position: 'absolute', 'font-size': '170%' }.merge(align) }, arrow),
-              ])
+                render_2d_price_info(price.info),
+              ].compact
+
+              h(:div, { style: cell_style(has_price_info ? @box_style_2d_with_info : @box_style_2d, price.types) },
+                elements)
             else
-              h(:div, { style: @space_style_2d }, '')
+              h(:div, { style: has_price_info ? @space_style_2d_with_info : @space_style_2d }, '')
             end
           end
 
           h(:div, { style: { width: 'max-content' } }, row)
         end
+      end
+
+      def render_2d_price_info(info)
+        return unless info
+        return h(:div, { style: PRICE_STYLE_2D_INFO }, info) unless info.is_a?(Hash)
+
+        appreciation = info[:appreciation].zero? ? 'END' : info[:appreciation]
+        dividend = info[:dividend].to_s
+        dividend += "+#{info[:president]}" if info[:president].positive?
+        label = "Appreciation threshold: #{appreciation}; dividend per share: #{dividend}"
+
+        h(:div, { attrs: { title: label }, style: PRICE_STYLE_2D_INFO }, [
+          h(:span, appreciation),
+          h(:span, dividend),
+        ])
       end
 
       def logo_for_user(entity)
@@ -459,6 +496,14 @@ module View
 
         # For cells with prices
         @box_style_2d = @space_style_2d.merge(
+          border: "solid #{BORDER}px rgba(0,0,0,0.2)",
+          color: color_for(:font2),
+        )
+        @space_style_2d_with_info = @space_style_2d.merge(
+          width: "#{WIDTH_TOTAL + PRICE_HEIGHT - (2 * PAD) - (2 * BORDER)}px",
+          height: "#{HEIGHT_TOTAL + PRICE_HEIGHT - (2 * PAD) - (2 * BORDER)}px",
+        )
+        @box_style_2d_with_info = @space_style_2d_with_info.merge(
           border: "solid #{BORDER}px rgba(0,0,0,0.2)",
           color: color_for(:font2),
         )
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -426,14 +426,13 @@ module View
 
               first_price = false
 
-              cell_elements = [
-                h('div.xsmall_font', price.price),
-                h(:div, tokens),
-                h(:div, { style: { color: '#00000060', position: 'absolute', 'font-size': '170%' }.merge(align) }, arrow),
-                price.info ? h(:div, { style: PRICE_STYLE_INFO }, price.info) : nil,
-              ].compact
-
-              h(:div, { style: cell_style(@box_style_2d, price.types) }, cell_elements)
+              h(:div, { style: cell_style(@box_style_2d, price.types) },
+                [
+                  h('div.xsmall_font', price.price),
+                  h(:div, tokens),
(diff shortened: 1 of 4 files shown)


perwestling (dismissed):

---
### ✅ Strengths
- Implements requested feature (closes #12723)
- Comprehensive change: adds appreciation thresholds, dividends, president bonuses
- Clean UI implementation with enlarged 2D cells for metadata
- Extensive verification:
  - Audited shared `grid_2d` renderer across 60 markets
  - Confirmed 18ZOO is only game using `SharePrice#info`
  - Manual testing of 1880, 1822, 1866, 18MS (appropriate for frontend code)
- All backend tests pass (179 examples, 0 failures)
- RuboCop passes
- Adds backend regression tests for metadata

---
### ❓ Questions
- Have you considered adding a screenshot of the new UI to the PR description?

---
### ✅ Verdict: **Approve**

Well-crafted implementation. The frontend changes are appropriately verified through manual testing, and backend changes are well-tested.

carlorusso1984:
I checked whether the spec update is actually useful. My conclusion is that it is worth keeping.

The spec does not test the frontend rendering directly, but it protects the model/data change that feeds that rendering:

- it verifies that converting 18ZOO `MARKET` from string codes to hash entries did not change the share prices or important stock-market types such as `safe_par`, `endgame`, and `par_1`;
- it verifies that every 18ZOO `SharePrice` receives the expected `appreciation`, `dividend`, and `president` metadata;
- it covers the main backend risk introduced by this refactor: the UI depends on these generated `SharePrice#info` values being present and aligned with the existing stock-market constants.

So while the browser/manual checks are still the meaningful coverage for layout regressions, this spec is not redundant. It guards the data contract used by the shared 2D stock-market renderer.

assets/app/view/game/stock_market.rb
```diff
@@ -399,6 +410,8 @@ def grid_hex
       end
 
       def grid_2d
+        has_price_info = @game.stock_market.market.flatten.compact.any?(&:info)
```
crericha: Why do you need to know this ahead of time? Why can't the code add the info for the cells that have it as it comes across one?

assets/app/view/game/stock_market.rb
```diff
@@ -424,22 +437,46 @@ def grid_2d
                 arrow = ''
               end
 
+              if has_price_info && !arrow.empty?
+                align.delete(:bottom)
+                align[:top] = "#{PRICE_HEIGHT}px"
+              end
```
crericha: I'm not sure it makes sense to overwrite previous align settings completely. This may work for ZOO, but doesn't seem universal for common code.

assets/app/view/game/stock_market.rb
```diff
@@ -424,22 +437,46 @@ def grid_2d
...
+
               first_price = false
 
-              h(:div, { style: cell_style(@box_style_2d, price.types) }, [
+              elements = [
                 h('div.xsmall_font', price.price),
                 h(:div, tokens),
                 h(:div, { style: { color: '#00000060', position: 'absolute', 'font-size': '170%' }.merge(align) }, arrow),
-              ])
+                render_2d_price_info(price.info),
```
crericha: render is not the correct term here.

assets/app/view/game/stock_market.rb
```diff
@@ -424,22 +437,46 @@ def grid_2d
...
+              ].compact
+
+              h(:div, { style: cell_style(has_price_info ? @box_style_2d_with_info : @box_style_2d, price.types) },
```
crericha: This seems quite heavy handed. There isn't a cleaner way? The zigzag price info does not have these problems.

assets/app/view/game/stock_market.rb
```diff
@@ -424,22 +437,46 @@ def grid_2d
...
+
+        appreciation = info[:appreciation].zero? ? 'END' : info[:appreciation]
+        dividend = info[:dividend].to_s
+        dividend += "+#{info[:president]}" if info[:president].positive?
+        label = "Appreciation threshold: #{appreciation}; dividend per share: #{dividend}"
+
+        h(:div, { attrs: { title: label }, style: PRICE_STYLE_2D_INFO }, [
+          h(:span, appreciation),
+          h(:span, dividend),
+        ])
```
crericha: This is hard coded for 18ZOO. It needs to be generic.

lib/engine/game/g_18_zoo/game.rb
```diff
@@ -24,14 +24,49 @@ class Game < Game::Base
...
-        ].freeze
+        ].map.with_index do |row, row_index|
+          row.map.with_index do |code, column_index|
+            match = code.match(/(\d*)([a-zA-Z]*)/)
+            president = row_index.zero? ? STOCKMARKET_OWNER_GAIN[column_index] || 0 : 0
+
+            {
+              price: match[1].to_i,
+              types: match[2].chars.map { |char| Engine::SharePrice::TYPE_MAP[char] },
+              info: {
```
crericha: Info is a single text value, it is not a hash map.

crericha:
The stock market view code is way more complex than I was expecting, given how simple it is for the zigzag market. It also is going above what price info is meant for and creating the layout specifically for 18ZOO, which should not be in common code. P I spoke with some friends that play 18ZOO a lot and they feel "dividend, share price movement" is good enough. For example `200, 2`.

This would allow you to use the common code that is already there for zigzag markets with minimal changes.

carlorusso1984:
> The stock market view code is way more complex than I was expecting, given how simple it is for the zigzag market. It also is going above what price info is meant for and creating the layout specifically for 18ZOO, which should not be in common code. P I spoke with some friends that play 18ZOO a lot and they feel "dividend, share price movement" is good enough. For example `200, 2`.
> 
> This would allow you to use the common code that is already there for zigzag markets with minimal changes.

I agree with you - I checked the logic and simplify a lot based on your suggestions.

image

@serpentium what do you think?

serpentium:
> > The stock market view code is way more complex than I was expecting, given how simple it is for the zigzag market. It also is going above what price info is meant for and creating the layout specifically for 18ZOO, which should not be in common code. P I spoke with some friends that play 18ZOO a lot and they feel "dividend, share price movement" is good enough. For example `200, 2`.
> > This would allow you to use the common code that is already there for zigzag markets with minimal changes.
> 
> I agree with you - I checked the logic and simplify a lot based on your suggestions.
> 
> image
> @serpentium what do you think?

Good enough

carlorusso1984:
Done!!! @crericha I think I followed all the suggestions to get the best achievable with the smallest possible changes

assets/app/view/game/stock_market.rb
```diff
@@ -426,11 +426,14 @@ def grid_2d
...
-              h(:div, { style: cell_style(@box_style_2d, price.types) }, [
+              cell_elements = [
                 h('div.xsmall_font', price.price),
                 h(:div, tokens),
                 h(:div, { style: { color: '#00000060', position: 'absolute', 'font-size': '170%' }.merge(align) }, arrow),
-              ])
+                price.info ? h(:div, { style: PRICE_STYLE_INFO }, price.info) : nil,
+              ].compact
+
+              h(:div, { style: cell_style(@box_style_2d, price.types) }, cell_elements)
```
crericha: I don't think we need `cell_elements` anymore. It can be defined directly in the `h(:div,...)` like it used to be. Otherwise looks good.

crericha (changes requested):

assets/app/view/game/stock_market.rb
carlorusso1984: done

carlorusso1984: @crericha if you agree that is ok, can you please merge, so at least I'm able to close this bug? I have other two but they are less important than this

crericha (approved):

PR #12753, opened 2026-06-08
[1880 Romania] Game uses standard float

### LINKED ISSUE #12740: [1880 Romania] floating is strictly by percentage
In **1880 China**, corps float when the president owns a certain percentage. 1880 Romania just uses a standard "floats when x amount is sold"

philcampeau:
Fixes #12740

## Before clicking "Create"

### Explanation of Change

In the original 1880 China, corporations only floated if the president held a certain percentage of shares at the end of the SR where the corp was parred. 

1880 Romania does away with this rule, and now floating is strictly based on initially selling a certain percentage from the IPO. 

I didn't initially implement this because it wasn't mentioned in Lonny's rules-differences file he provided me. 

I put the archive tag because it's possible some games may not work after this change, if a player bought enough shares to float their corp and then sold down before the end of that SR. Otherwise, everything should be good.


perwestling (approved):
LGTM

PR #12754, opened 2026-06-09
[18ZOO] Clean up unlimited train variant

carlorusso1984:
Works towards #12237.

## Summary
- remove the obsolete finite train count from the 2J variant of the unlimited 4J rank
- add a regression spec confirming that removing a 2J replenishes the unlimited 4J/2J rank

## Testing
- `ruby -c lib/engine/game/g_18_zoo/game.rb`
- `ruby -c spec/lib/engine/game/g_18_zoo/game_spec.rb`
- `git diff --check`

RSpec could not be run locally because Bundler 2.5.13 is unavailable and the Docker daemon is not running.

### DIFF
--- a/lib/engine/game/g_18_zoo/game.rb
+++ b/lib/engine/game/g_18_zoo/game.rb
@@ -134,6 +134,7 @@ module Engine
                 distance: 2,
                 multiplier: 2,
                 price: 37,
+                num: 'unlimited',
               },
             ],
           },
--- a/lib/engine/game/g_18_zoo/game.rb
+++ b/lib/engine/game/g_18_zoo/game.rb
@@ -134,7 +134,6 @@ module Engine
                 distance: 2,
                 multiplier: 2,
                 price: 37,
-                num: 99,
               },
             ],
           },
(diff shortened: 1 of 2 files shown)


perwestling (dismissed):

---
### ✅ Strengths
- Works towards issue #12237
- Correct fix: removes obsolete `num: 99` so variant inherits parent's `num: 'unlimited'`
- Adds regression test confirming 4J/2J replenishment works correctly
- Syntax and style checks pass

---
### ✅ Verdict: **Approve**

The change is technically correct. Removing `num: 99` allows the 2J variant to properly inherit the unlimited property from its parent 4J train. Ready to merge.

ollybh (changes requested):
Don't rely on the behaviour if `num` isn't defined. If you want unlimited trains then that's what you need to say: `num: 'unlimited'`.

Also, what is the unit test actually testing?

carlorusso1984:
Agreed. I changed the 2J variant to declare `num: 'unlimited'` explicitly. The test was only exercising replenishment from the base 4J train’s unlimited flag, so it did not actually cover the variant `num` change; I removed it.

crericha (approved):

carlorusso1984:
@ollybh can you please take a look? it was approved, so I would like to have it merged too

ollybh (approved):

PR #12760, opened 2026-06-11
[System18] auto-pass users at cert limit during parliament rounds

### LINKED ISSUE #12594: [System18: Britain, Italy & Poland] Add new rule
Following an unusual case in a game of Sys18 Britain:
https://18xx.games/game/249891
where the player in the lead was able to drive up the price of the last company, to the detriment of the player coming last, I have decided to add the following rule:
"A player at certificate limit may not bid during a PR."

This applies to all maps with a Parliament Round - currently Britain (all versions), Italy, and Poland.

philcampeau:
Fixes #12594

## Before clicking "Create"

### Explanation of Change

Implements a forced-pass for players already at cert limit during parliamentary rounds, to prevent players who cannot acquire new certificates from bidding. Also adds a clear log message explaining the skip.

This may require pins for any games where players at cert limit placed bids on items in Parliamentary rounds. 

### Screenshots

image

### DIFF
--- a/lib/engine/game/g_system18/step/upwards_auction.rb
+++ b/lib/engine/game/g_system18/step/upwards_auction.rb
@@ -38,7 +38,8 @@ module Engine
               pass_auction(entity)
               resolve_bids
             else
-              @log << "#{entity.name} passes bidding"
+              reason = @game.num_certs(entity) >= @game.cert_limit(entity) ? 'is at cert limit and must pass' : 'passes'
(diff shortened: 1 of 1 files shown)


roseundy (approved):

ollybh:
No pins needed. This isn't changing the output of actions just adding an auto-action pass. So any illegal bids in existing games will be processed fine. You just won't be able to place these bids going forward.

PR #12761, opened 2026-06-13
[1832] Most everything to get merges and takeovers working.

jemiahlee:
## Before clicking "Create"

## Implementation Notes

This is pretty much everything I could find in my repo for getting mergers working. This game is weird in that it has VP certs that are double certificates, and they may be leftover from the merge if the merge is between two companies that have different presidents. So the algorithm here converts both presidencies to VP certificates and then compares percentages, and then trades in those VP certs first toward the president's share in the system. I'm not sure the post-merge VP cert thing works entirely (I will do a lot more testing again after my yet-to-come PR to make sure I didn't mess anything up in the process of splitting for review), but it does work well in the merge phase.

This is voluminous, and I'm sorry, and if there is a way you think I should split it up instead, I am open to that so that the review can be smaller. I just wanted to have everything out there for the merging step so that if there were a piece you were curious about, it would be represented here.

I did look at 1828 systems, but those are far more complex than the systems here in 1832, so I did not think it was worthwhile to extend those. This code just calculates whether a president may buy voluntarily buy a train for a system based on having train count <= (train_limit / 2), where the train limit for the system is doubled (it sounds weird, but the math is done in two different places that I _think_ make sense).

### DIFF
--- a/lib/engine/game/g_1832/corporation.rb
+++ b/lib/engine/game/g_1832/corporation.rb
@@ -6,7 +6,11 @@ module Engine
   module Game
     module G1832
       class Corporation < Engine::Corporation
-        attr_accessor :coal_token
+        attr_accessor :coal_token, :is_system, :system_shells
+
+        def system?
+          is_system
+        end
       end
     end
   end
--- a/lib/engine/game/g_1832/game.rb
+++ b/lib/engine/game/g_1832/game.rb
@@ -21,7 +21,7 @@ module Engine
         include G1832::Trains
 
         attr_accessor :sell_queue, :reissued, :coal_token_counter, :coal_company_sold_or_closed, :p4_invested_in,
-                      :miami_has_been_run
+                      :miami_has_been_run, :mid_or_resume_entities
 
         CORPORATION_CLASS = G1832::Corporation
         CORPORATE_BUY_SHARE_ALLOW_BUY_FROM_PRESIDENT = true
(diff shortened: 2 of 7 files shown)


perwestling (changes requested):
This was so large so I took help of an AI to suggest some splits. Here is the result:

---
### ⚠️ Major Concern: PR Size

This PR is **very large** (822 lines added, 4 deleted across 7 files). Large PRs are harder to review, harder to test, and have higher risk of introducing bugs.

---
### 💡 Suggested Split

This PR touches multiple independent features that could be separate PRs:

| **Suggested PR** | **Files** | **Focus** |
|-----------------|-----------|-----------|
| PR 1: Corporation changes | `corporation.rb` | Add `system?` method and `system_shells` |
| PR 2: Merger infrastructure | `game.rb` (partial), `merger.rb`, `round/operating.rb` | Merger round setup |
| PR 3: Merger logic | `game.rb` (partial), `step/merge.rb` | Core merger logic |
| PR 4: Takeover logic | `game.rb` (partial), `step/sell_shares_for_takeover.rb`, `step/buy_train.rb` | Takeover and train buying |
| PR 5: VP cert handling | `game.rb` (partial) | Vice president certificate handling |

**Benefits of splitting**: Easier to review, test, and merge independently.

---
### ✅ Strengths
- Comprehensive implementation of 1832 merger/takeover logic
- All tests pass
- Good use of existing 18xx patterns
- Author open to splitting

---
### ⚠️ Concerns
- **Size**: 822 lines across 7 files is too large for a single PR
- **Scope**: Touches multiple independent features
- **Testing**: Author mentions uncertainty about post-merge VP cert behavior

---
### ❓ Questions
- Have you considered splitting this into smaller PRs focused on specific features?
- Can the merger infrastructure be separated from the merger logic?
- Have you tested the post-merge VP certificate behavior thoroughly?

---
### ❌ Verdict: **Request changes**

This PR is too large and should be split into logical components. Please split as suggested above and add tests for the uncertain VP cert behavior.

PR #12764, opened 2026-06-15
[1832] Add system code to corporation, add logic for system train buys.

jemiahlee:
## Before clicking "Create"

### Explanation of Change

This PR adds a little bit of functionality to corporation to distinguish the system from other corps, and adds train buy functionality for Systems, which are allowed to voluntarily buy as long as the system has <= current train limit (basically one of the shells must be able to be configured with zero trains, though there is no such need for actually configuring them). The president is allowed to contribute cash for that voluntary buy.

### DIFF
--- a/lib/engine/game/g_1832/corporation.rb
+++ b/lib/engine/game/g_1832/corporation.rb
@@ -6,7 +6,11 @@ module Engine
   module Game
     module G1832
       class Corporation < Engine::Corporation
-        attr_accessor :coal_token
+        attr_accessor :coal_token, :is_system, :system_shells
+
+        def system?
+          is_system
+        end
       end
     end
   end
--- a/lib/engine/game/g_1832/corporation.rb
+++ b/lib/engine/game/g_1832/corporation.rb
@@ -6,11 +6,7 @@ module Engine
   module Game
     module G1832
       class Corporation < Engine::Corporation
-        attr_accessor :coal_token, :is_system, :system_shells
-
-        def system?
-          is_system
-        end
+        attr_accessor :coal_token, :system_shells
       end
     end
   end
--- a/lib/engine/game/g_1832/step/buy_train.rb
+++ b/lib/engine/game/g_1832/step/buy_train.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+require_relative '../../g_1870/step/buy_train'
+
+module Engine
+  module Game
+    module G1832
+      module Step
+        class BuyTrain < G1870::Step::BuyTrain
+          def actions(entity)
+            return super unless entity.system?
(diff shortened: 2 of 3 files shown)


lib/engine/game/g_1832/corporation.rb
```diff
@@ -6,7 +6,11 @@ module Engine
   module Game
     module G1832
       class Corporation < Engine::Corporation
-        attr_accessor :coal_token
+        attr_accessor :coal_token, :is_system, :system_shells
+
+        def system?
+          is_system
+        end
```
ollybh: This is almost certainly a bad idea. There are places in the code base that check `entity.system?` and will treat this like an 1828 System object if this returns true. You have a `G1832::Game.system?(entity)` method defined in your `game.rb` file, use that method to test whether a corporation has type `:system`.

lib/engine/game/g_1832/step/buy_train.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          # Redeemed shares (buyable: false) may not be sold to raise emergency train funds (§10.6).
+          def can_sell?(entity, bundle)
+            super && bundle.shares.all?(&:buyable)
+          end
+
+          private
+
+          def system_voluntary_buy?(system)
+            system.trains.count <= @game.train_limit(system) / 2
```
ollybh: This is a change that should not make any difference, but I'm always a little bit wary of division. Opal translates the Ruby code to Javascript, and numbers in Javascript are always floating point, so there's a chance that division could do something odd.
```suggestion
            system.trains.count * 2 < @game.train_limit(system)
```

lib/engine/game/g_1832/corporation.rb
```diff
@@ -6,7 +6,11 @@ module Engine
   module Game
     module G1832
       class Corporation < Engine::Corporation
-        attr_accessor :coal_token
+        attr_accessor :coal_token, :is_system, :system_shells
```
ollybh: Are you implementing system shells? Or just using the heuristic that the president can contribute if the system's train count is less than half the train limit?

ollybh (changes requested):
Don't have a corporation object with a `system?` method that could return true. That's almost certain to cause problems later. Drop the Corporation subclass and test `@game.system?(entity)` instead of `entity.system?`.

Everything else looks basically fine.

lib/engine/game/g_1832/corporation.rb
jemiahlee: Just using the heuristic. The system_shells here is basically to keep track of the shells in the event of a mid-OR merger (on the 6 train). We need some way to track which shells were involved so that we can avoid systems that should not run in the case one of the shells already ran when we pick the OR back up again.

PR #12767, opened 2026-06-19
[profile] username format validation for Slack & Discord webhooks

### LINKED ISSUE #12726: [18xxoggs] Some Slack notifications not parsed correctly - Slack no longer falls back when using Username instead of User ID
At some point in the last week or two, I stopped receiving Slack notifications but didn't notice as I was on holiday. This morning, I did some tests and the issue repeats in both the Android app and within a Chrome browser on a Windows laptop or Linux desktop. I tried changing my username to match the exact case of that of my profile, without impact.

The image below shows the current feed. One can see that some, but not all, notifications are recognising the initial username as a tag and some are not. My guess is Slack changed the API in some way.

Image

**How to Reproduce the Bug?**

Activate the "Send test notification" for an affected account.

**Which Browser / Platform?**

Chrome 148.0.7778.179
Android App 26.05.40.0

philcampeau:
Fixes #12726

## Before clicking "Create"

### Explanation of Change

Right now, users can setup webhooks for slack/discord and put pretty much anything in the User ID field. The problem is, without using the proper format, a message is posted to the channel but the user isn't properly tagged. 

This adds some checks that trigger flash_opts error messages if a user tries to signup or save their user profile with an invalid Slack or Discord user ID.

My only hesitation on this is that there could be people who don't care to be tagged, they just want to be notified - mostly people who've set up their own private discord/slack. If you think this might be too restrictive, I could limit it only to people setting up notifications on 18oggs, or even just abandon this altogether. Let me know!

### DIFF
--- a/assets/app/view/user.rb
+++ b/assets/app/view/user.rb
@@ -445,8 +445,14 @@ module View
       delete_user
     end
 
-    SLACK_MEMBER_ID = /\A[UW][A-Z0-9]{7,}\z/
-    DISCORD_USER_ID = /\A\d{17,19}\z/
+    WEBHOOK_CHECKS = {
+      # Slack member IDs start with U (or W), never a username.
+      # https://docs.slack.dev/messaging/formatting-message-text/
+      slack: { match: 'slack', id_regexp: /\A[UW][A-Z0-9]{7,}\z/ },
+      # Discord IDs are numeric, currently 17-19 digits.
+      # https://discord.com/developers/docs/reference#snowflakes
+      discord: { match: 'discord', id_regexp: /\A\d{17,19}\z/ },
+    }.freeze
 
     def submit
       case @type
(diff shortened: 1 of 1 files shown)


ollybh (changes requested):
Can you toggle the method names from `*_id_invalid?` to `*_id_valid?`. It doesn't matter much for the way you're testing the IDs, but if someone ever wants to check if an ID is valid it's weird to have to say `!…_invaid?`.

Please add comments pointing to the documentation of the legal user IDs. This will help anyone who later touches this code.

All the methods you're adding should be in a `private` section.

You have a method for testing whether this is a Slack webhook, but the test for a Discord webhook is embedded in the `discord_id_valid?` method. It would be better to be consistent.

If you're feeling more ambitious, this could be made more extensible by creating a data structure that stores the tests:
```ruby
WEBHOOK_CHECKS = {
  slack: {
    webhook?: Proc.new { … },
    id_regexp: /…/,
  },
  …
}
```

philcampeau:
Sure. I made some changes.

PR #12768, opened 2026-06-20
[Dev] Add hex map editor

JohnPDuong:
## High Level Summary
Change adds a map editor that generates ruby code to allow devs to easily construct a map.

Plans are to also add a selection of different hex types, adding locations for cities, towns, red zones, etc. and other customizability.

My story: I want to be able to make a map centered around Seattle but found making a map just through code is unruly. I thought other people may have the same issues so I wanted to make a map editor that allows myself and others to construct their maps with ease.

## Before clicking "Create"

## Implementation Notes

Add an in-app map editor at /map_editor/<game> for authoring a game's white/blue hexes, plus a standalone tools/map_editor.html
that needs no running server.

- Draws its own SVG hex grid reusing the engine's hex geometry (Lib::Hex::POINTS and the flat/pointy layout math) so coordinates line up with the real map.
- White/Blue/Erase tools, drag to paint, adjustable column/row canvas size, a session undo stack, and a Generate Ruby button that emits a paste-ready HEXES snippet for the game's map.rb.
- /map_editor/<game> auto-loads that game's map; bare /map_editor opens a small blank grid to start.

Wires up the /map_editor route in app.rb and api.rb.

### Screenshots
/map_editor pulls up empty blank canvas
image

/map_editor after making changes
Screenshot 2026-06-19 181421

/map_editor/20seattle when loading up a new map
Screenshot 2026-06-19 181538

/map_editor/20seattle when generating ruby code
image

/map_editor/1889 has special hexes, no blue hexes but also has special hexes so not compatible yet
image

/map_editor/1817 has special hexes so not compatible yet
image

### Any Assumptions / Hacks
N/A

### DIFF
--- a/api.rb
+++ b/api.rb
@@ -89,7 +89,7 @@ class Api < Roda
   ].freeze
 
   ROUTES_WITH_GAME_TITLES = %w[
-     map market fixture
+     map map_editor market fixture
   ].freeze
 
   Dir['./routes/*'].each { |file| require file }
@@ -129,7 +129,7 @@ class Api < Roda
     end
 
     r.on ROUTES_WITH_GAME_TITLES do
-      render(titles: request.path.split('/')[2].split('+'))
+      render(titles: request.path.split('/')[2]&.split('+') || [])
     end
 
     r.on 'profile' do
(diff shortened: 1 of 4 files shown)


crericha:
What scope are you planning? A blank map (e.g. all white and blue tiles)? The preprinted hexes? Do you plan to support all the options available?

Also, are you aware there use to be a way to import from 18xx.maker? It might be easier to fix whatever broke with that than add a whole new system.

tobymao:
thanks for this but we won’t accept this.

PR #12774, opened 2026-06-23
[1871] UB owner should see UB trains as crossbuy option

### LINKED ISSUE #12769: [1871 TOP] Union Bank must "show trains from other players" to buy trains across from other companies
See here:
https://18xx.games/game/256693?action=505

waaayback owns the UB.  The Union Bank owns the Murray River. It has no train and $2, and would like to buy a train across from the C (also owner by waaayback), but at first glance it is not given that option.

The player has to "show trains from other players" to see available trains, **even the trains owned by other companies owned by the Union Bank's owner**.  This is confusing and not intuitive.  Granted, having an entity like the UB is uncommon.

To resolve, I suggest that either:
1) Put up a note to say "click on 'show trains from other players' to possibly see trains owned by the same player"  --or--
2) Present those such trains as options on this screen

philcampeau:
Fixes #12769

## Before clicking "Create"

### Explanation of Change

User reports that when you're running the Union bank, you still have to click "show trains from other players" to see UB corp trains when operating one of your corps, and vice versa. They feel this is confusing and not intuitive, since you are operating all these corps in actual fact. 

This code adds in a check that displays trains held by UB corps to the UB's owner, and vice versa, without displaying them by default to others. 

### Screenshots

image

### DIFF
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -576,24 +576,9 @@ module View
       end
 
       def same_effective_owner?(corporation, other_corp)
-        buyer_owner = corp_owner(corporation)
-        seller_owner = other_owner(other_corp)
-
-        return true if buyer_owner == seller_owner
-
-        return false unless @game.respond_to?(:acting_for_player)
-
-        # code for The Old Prince 1871's Union Bank
-        ub_player = @game.instance_variable_get(:@union_bank)
-        return false unless ub_player
-
-        human = @game.acting_for_player(ub_player)
-        return false if !human || human == ub_player
-
-        effective_buyer = buyer_owner == ub_player ? human : buyer_owner
-        effective_seller = seller_owner == ub_player ? human : seller_owner
-
-        effective_buyer == effective_seller
+        player_a = corp_owner(corporation)
+        player_b = other_owner(other_corp)
+        player_a == player_b || @game.same_acting_player?(player_a, player_b)
       end
 
       # need to abstract due to corporations owning minors owning trains
--- a/assets/app/view/game/buy_trains.rb
+++ b/assets/app/view/game/buy_trains.rb
@@ -520,7 +520,7 @@ module View
               train_props[:style][:backgroundColor] = color
               train_props[:style][:color] = contrast_on(color)
             end
-            line = if @show_other_players || other_owner(other) == corp_owner(@corporation)
+            line = if @show_other_players || same_effective_owner?(corporation, other)
                      [h(:div, train_props, name),
                       h('div.nowrap', train_props,
                         "#{other.name} (#{count > 1 ? "#{count}, " : ''}#{corp_owner(other).name}#{real_name})"),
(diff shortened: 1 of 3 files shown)


chrstas:
hi @philcampeau,

Would a peer review be welcome? I got curious about this one and spent a bit time digging in. Functionally it resolves the issue but one architectural point I wanted to raise.

The change lives in the shared `assets/app/view/game/buy_trains.rb`, which every game renders, and reaches into 1871 internals from there:

```ruby
ub_player = @game.instance_variable_get(:@union_bank)
```

This leads to game specifics in the view layer. The shared view already routes game-specific policy through `@game` hooks - e.g. `@game.can_buy_train_from_others?`, which resolves a `Base` constant that a game can switch off - so the same separation would fit well here.

Two ways I looked at:

**Option A: reuse the existing `acting_for_player` hook**

```ruby
line = if @show_other_players ||
          @game.acting_for_player(corp_owner(@corporation)) == @game.acting_for_player(other_owner(other))
```

But I'd avoid this one. `acting_for_player` is also overridden by 18CZ (`@vaclav`) and 1858 Switzerland (`@robot`), so it would change their buy-train display as well.

**Option B: a dedicated hook with a `Base` default (my proposal)**

```ruby
# lib/engine/game/base.rb
def same_acting_player?(_owner_a, _owner_b)
  false
end

# lib/engine/game/g_1871/game.rb
def same_acting_player?(owner_a, owner_b)
  a = acting_for_player(owner_a)
  b = acting_for_player(owner_b)
  a && b && a == b
end
```

```ruby
# buy_trains.rb — view stays generic
line = if @show_other_players || same_effective_owner?(corporation, other)
# ...
def same_effective_owner?(corporation, other_corp)
  buyer = corp_owner(corporation)
  seller = other_owner(other_corp)
  buyer == seller || @game.same_acting_player?(buyer, seller)
end
```

(The game hook takes owners, the view helper takes corporations, hence the different names, so the call chain stays readable.)

All UB knowledge stays in `g_1871`, the view stays generic, and since only 1871 overrides the hook it's a no-op for every other game (18CZ/1858CH included). Same shape as the existing `can_buy_train_from_others?` hook.

The one downside of B is that it adds a hook only one game uses today.

**Additional Alternative C (not evaluated further)**

Alternatively, 1871's buy-train step could just override the existing `corp_owner`/`real_owner` helpers to go through `acting_for_entity`, which would fix this line with no shared-code change at all. The catch is that those two are used multiple times in the file, so it'd change consent handling and the cash display too - some of which you may well want, but it's a wider change than the hook.

I'd still lean towards B, since an explicit hook is greppable and matches `can_buy_train_from_others?` - but C fits existing precedent too.

Hope it helps...

chrstas:
Without digging deeper, one more thing that came into my mind looking high-level to the other 1871 PRs:

a couple of the open 1871 changes touch the same "the Union Bank acts as a player/president" problem, and 1871 already has `acting_for_player` / `acting_for_entity` for the identity part of it (`@union_bank → owner`).

Where a site just needs "who is the acting player", that seam covers it, e.g. `split_active_entities` in #12687 is `[@game.acting_for_entity(@split_corporation)].compact`, and the `acting = player == @union_bank ? … : …` line in #12685 is `acting_for_player(player)`. The hook in this PR builds on the same seam.

Just an idea...

philcampeau:
Hey @chrstas! Thanks for those suggestions. You're absolutely right that those are much cleaner suggestions. I'll look into implementing them now!

chrstas:
> Hey @chrstas! Thanks for those suggestions. You're absolutely right that those are much cleaner suggestions. I'll look into implementing them now!

Thanks for picking it up! I am a fan of peer reviews and in German we say "four eyes see more than two". Curious to hear the final word from the maintainer side too. Still learning myself here. 😊

PR #12775, opened 2026-06-23
[1858 India] Update tile roster - designer's request

### LINKED ISSUE #12773: [1858: India] Add extra tiles
Following more playtesting, I've decided to add some more tiles to the mix. Most are yellow tiles (which only affect f-t-f play, the online version has unlimited yellows), but the following affects the .games implementation:

Increase the number of #89, #116 and #X6 tiles from 2 to 3. (These are the green all-narrow gauge dot upgrades.)

philcampeau:
Fixes #12773

## Before clicking "Create"

### Explanation of Change

Ian Wilson asked that we modify the tile roster. Done per his request.


ollybh (approved):
This won't need any pins – it's adding extra tiles not removing them, so won't break any existing games.

PR #12776, opened 2026-06-23
[1822MX] unfloated corps shouldn't be able to bid on or buy NdeM tokens

### LINKED ISSUE #12759: [1822MX] - Unfloated company allowed to bid on NDEM tokens
In this game (https://18xx.games/game/250512?action=1023), the FCP was allowed to bid on and buy an NDEM token during the privatization auction, even though the company had not floated yet.

This seems contrary to the rules, but it's a tad ambiguous.  Rule 6.22.2 on NDEM token sales merely state that a player can bid on behalf of a company, without limiting it to floated companies.  And Rule 5.9 states that companies get their tokens when they are "formed" not when they are "floated".  So the FCP would have cash and tokens as soon as the president's share is purchased.  But, Rule 5.8.2 provides that directly formed companies (like the FCP here) "do not float and become operational until 50% of its shares are sold. And the NDEM privatization rule is in the section on operating rounds (rule 6), so this rule should be read to limit "companies" that can bid on tokens to those companies that are eligible to operate, which an unfloated company is not.  So I do not think this should be allowed.

philcampeau:
Fixes #12759

## Before clicking "Create"

### Explanation of Change

User reports:

> Rule 6.22.2 on NDEM token sales merely state that a player can bid on behalf of a company, without limiting it to floated companies. And Rule 5.9 states that companies get their tokens when they are "formed" not when they are "floated". So the FCP would have cash and tokens as soon as the president's share is purchased. But, Rule 5.8.2 provides that directly formed companies (like the FCP here) "do not float and become operational until 50% of its shares are sold. And the NDEM privatization rule is in the section on operating rounds (rule 6), so this rule should be read to limit "companies" that can bid on tokens to those companies that are eligible to operate, which an unfloated company is not. So I do not think this should be allowed.

I agree with that interpretation. 

I added gates for unfloated corps in 3 different methods in the auction_ndem_tokens.rb code. Here's why I felt all three were necessary: 

- Without the fix to `def max_bid_for_token`, the bid ceiling could be inflated by cash in an unfloated corp's treasury.

- Without the fix to `def corp_can_purchase_token?`, `player_can_purchase_token?` returns true for players whose only eligible corp is parred but unfloated, so they incorrectly enter the bidding rotation.

- Without the fix to `def mergeable`, the winner could assign the token to an unfloated corp, even if the other two methods are fixed.

### DIFF
--- a/lib/engine/game/g_1822_mx/step/auction_ndem_tokens.rb
+++ b/lib/engine/game/g_1822_mx/step/auction_ndem_tokens.rb
@@ -59,7 +59,7 @@ module Engine
           end
 
           def max_bid_for_token(player)
-            corps = player.presidencies.reject { |c| @game.exchange_tokens(c).zero? }
+            corps = player.presidencies.reject { |c| @game.exchange_tokens(c).zero? || !c.floated? }
             corp = corps.max_by(&:cash)
             corp ? corp.cash : 0
           end
@@ -75,6 +75,7 @@ module Engine
           end
 
           def corp_can_purchase_token?(corporation, token)
+            corporation.floated? &&
             corporation.cash >= @available_ndem_tokens[token] &&
             !@game.exchange_tokens(corporation).zero? &&
             token.hex.tile.cities.none? { |c| c.tokened_by?(corporation) }
@@ -239,7 +240,9 @@ module Engine
           end
 
           def mergeable(_corporation)
-            @auction_winner.presidencies.select { |c| !@game.exchange_tokens(c).zero? && c.cash >= @current_high_bid }
+            @auction_winner.presidencies.select do |c|
+              c.floated? && !@game.exchange_tokens(c).zero? && c.cash >= @current_high_bid
+            end
           end
 
           def show_other_players


lib/engine/game/g_1822_mx/step/auction_ndem_tokens.rb
```diff
@@ -239,7 +240,9 @@ def merge_name(_entity = nil)
           end
 
           def mergeable(_corporation)
-            @auction_winner.presidencies.select { |c| !@game.exchange_tokens(c).zero? && c.cash >= @current_high_bid }
+            @auction_winner.presidencies.select do |c|
+              c.floated? && !@game.exchange_tokens(c).zero? && c.cash >= @current_high_bid
```
chrstas, line 244: Should `mergeable` and `max_bid_for_token` run through `corp_can_purchase_token?` as well? Both skip the same-hex check on line 81; the gap predates the PR, but these are the lines you are rewriting. Live today: with a legally placed FCP token in the auctioned hex and its director also holding CHP ($322), `max_bid_for_token` returned FCP's $442 although line 81 bars FCP; above $322 `mergeable` returned FCP alone, and `actions` offers no pass. 6.22.3: "Companies may not place a second token in a hex."

chrstas (commented):
hi @philcampeau - another peer review, hope it is ok:  the `floated?` change behaves as described, and `max_bid_for_token` now returns 0 for a director whose only funded major is parred but unfloated. One non-blocking comment below, on a pre-existing asymmetry in lines this PR already touches.

Issue #12485, opened 2026-04-06
[18Cuba] 18Cuba Implementation Checklist


chrstas:
# 18Cuba Checklist

## Setup
- [x] Basic setup 
- [x] Starting cash by player count
- [ ] Further 2p, 3p player variants
- [x] Initialize stock market
- [x] Set certificate limits
- [x] Initialize tile groups & graphs (Minor = narrow, Major = broad)

## Commissioners
- [x] Auctioning
- [ ] Abilities

## Concessions
- [x] Drafting
- [ ] Special Abilities

## Minor Companies
- [x] Parring with concessions
- [x] Tile laying and upgrading
- [x] Home token logic
- [x] Calculate sugar production 
- [ ] No upgrade costs for sugar cane fields
- [ ] Narrow gauge trains buying and upgrading
- [ ] Narrow gauge train rusting 
- [ ] Sugar mill machines
- [ ] Issue / redeem share

## Major Companies
- [x] Parring 
- [x] Tile laying and upgrading
- [x] Home token logic
- [ ] Tile laying on sugare cane fields
- [ ] FEC home token (incl. edge case no free slot)
- [ ] Wagons
- [ ] Transport sugar logic and port restrictions
- [ ] Train rusting
- [ ] Issue / redeem share

## FC Company
- [x] FC icons
- [ ]  Train export 
- [ ]  Routes
- [ ]  Fees
- [ ]  Sugar collection

## Issues
- [ ] Mixed tiles upgrade seem to offer incorrect possibilities
- [ ] Emergency buy should distinguish minor and major trains

I am working on the implementation and reuse also the code that @bentziaxl has provided. Can you please assign this issue to me? Thanks

PR #12779, opened 2026-06-27
[18Cuba] Add sugar cube transport

chrstas:
Implements sugar cube visualization and manual transport to harbors for 18Cuba (rules VII.9–VII.11), modelled on 18Uruguay's goods mechanic (train-centric state, attach-and-load during the route step).

## Before clicking "Create"

### Explanation of Change

- **Wagons (1w/2w/3w)**: attach to a regular broad train (one per train); 4D and narrow-gauge trains cannot carry a wagon.
- **Loading & delivery**: a wagon train running through a sugar mill on its route may load cubes up to the wagon's capacity and deliver them to a harbor for $30 each. Loading is manual and optional.
- **Harbor extension**: a wagon may extend a route by one stop to reach a harbor; that added-on harbor scores zero revenue but can still receive cubes (VII.10).
- **4D**: completes VII.10 — its revenue locations are now doubled (`multiplier: 2`). 

Auto-loading (greedy pickup with auto-router valuation) is a planned follow-up PR — this one keeps loading manual, matching 18Uruguay.

### Screenshots

Sugar mills with produced sugar:

image

Attaching wagons to the trains:

image

Routing with extension to harbor (10$ revenue is not given as train distance does not cover harbor):

image

Loaded sugar:

image

### Any Assumptions / Hacks

- **Sugar logic is separated in `sugar.rb`** 
- **Harbor detection** uses the tile's anchor icon (`harbor?`) rather than a  hardcoded hex list, so new harbors are picked up automatically.
- **`current_routes` is recorded as a side effect in `check_distance`** so the route step can offer cube loading — this mirrors 18Uruguay (`game.rb`).
- **Out of scope (TODO-marked, follow-up PRs)**: Isla de Tesoros 20/40 harbor value, the La Habana→La Habana routing rule,  and  the FC undelivered-cube mechanic.


lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,133 @@ module Engine
...
+              next unless spare.positive?
+
+              @game.mill_corps_on_route(route).each do |corp|
+                next unless @game.unclaimed_cubes(corp).positive?
+
+                key = { 'type' => 'load', 'train' => train.id, 'corp' => corp.id }
+                result[key] = "Load sugar from #{corp.name} onto #{train.name}"
+              end
+            end
+            result
```
kepakiano, line 143: This is a thing of personal preference: For me this method is a little to long and creates too many things I need to remember. If I were to write it from scratch, I think I'd use less unless statements and more filters, roughly like this:
```
trains_with_wagons_stopping_at_harbors = entity.trains.select #the first two unless checks
trains_with_wagons_stopping_at_harbors.select { |t| @game.train_with_cubes(t) }.each # adding the unload action goes here
trains_with_sugar_capacity = trains_with_wagons_stopping_at_harbors.select # spare.positive?
trains_with_sugar_capacity.each # adding the load action goes here
```

chrstas: @kepakiano: thanks for the peer review - very much appreciated. Took the suggestion and yes it reads much better this way.

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
-            # Currently always empty; check_distance and check_route_combination fall back to super.
-            super.merge({ wagon_for_train: {} })
+            super.merge(current_routes: {}, wagon_for_train: {})
+          end
+
+          def actions(entity)
+            actions = super
+            return actions if actions.empty?
+
+            actions = actions.dup
```
ollybh: I don't think this line is necessary.

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+            actions << 'choose' if choosing?(entity)
+            actions
+          end
+
+          def choosing?(entity)
+            # Keep the choose panel mounted while the entity owns a wagon (route-independent), like 18Uruguay.
+            entity.trains.any? { |t| @game.wagon?(t) }
```
ollybh: I think that this, as currently coded, is going to break if a major company owns a wagon but no trains to attach it to. The player will be asked to attach a wagon, but will not have anything to attach it to.

This whole step should be skipped in this case. Implement a `route_trains` in your game class:
```ruby
def route_trains(entity)
  entity.runnable_trains.reject { |train| wagon?(train) }
end
```

To be extra defensive, this method could have guard clauses that return false if the entity is a minor company, or does not own any (non-wagon) trains.

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+          end
+
+          def choices
+            choices_for(current_entity)
+          end
+
+          # Show the attached wagon on the train in the route UI, e.g. "2+1w" (display only).
+          def train_name(_entity, train)
+            wagon = @round.wagon_for_train[train.id]
+            wagon ? "#{train.name}+#{wagon.name}" : train.name
```
ollybh: This might look a bit odd when you have plus trains, they're going to have two consecutive plus signs in this description, something like `3++1w`. I'm not sure what would be better.

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+
+          private
+
+          def choices_for(entity)
+            wagon_choices(entity).merge(cube_choices(entity))
+          end
+
+          def process_attach(entity, wagon_id, train_id)
+            wagon = entity.trains.find { |t| t.id == wagon_id }
+            train = entity.trains.find { |t| t.id == train_id }
```
ollybh: You can use the train cache. It doesn't check that the train is owned by the current company, but it shouldn't be possible for the train ownership to have changed between the choices being displayed and one chosen.
```suggestion
            wagon = @game.train_by_id(wagon_id }
            train = @game.train_by_id(train_id }
```

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+            raise GameError, 'Train cannot take this wagon' if !@game.wagon?(wagon) || !@game.wagon_attachable?(train)
+            raise GameError, 'Wagon or train already attached' if @round.wagon_for_train.key?(train.id) ||
+                                                                  @round.wagon_for_train.value?(wagon)
+
+            @round.wagon_for_train[train.id] = wagon
+            @log << "#{entity.name} attaches #{wagon.name} to #{train.name}"
+          end
+
+          def process_load(entity, train_id, corp_id)
+            train = entity.trains.find { |t| t.id == train_id }
```
ollybh:
```suggestion
            train = @game.train_by_id(train_id)
```

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+            raise GameError, 'Train has no wagon' unless @round.wagon_for_train.key?(train.id)
+            raise GameError, 'Wagon is full' unless @game.cubes_on_train(train).size < @game.wagon_capacity(train)
+            raise GameError, 'No sugar cube available' unless @game.unclaimed_cubes(corp).positive?
+
+            @game.attach_cube_to_train(train, corp)
+            @log << "#{entity.name} loads a sugar cube from #{corp.name} onto #{train.name}"
+          end
+
+          def process_unload(entity, train_id)
+            train = entity.trains.find { |t| t.id == train_id }
```
ollybh:
```suggestion
            train = @game.train_by_id(train_id)
```

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+            raise GameError, 'Invalid train' unless train
+
+            @game.unload_cubes(train)
+            @log << "#{entity.name} unloads the sugar from #{train.name}"
+          end
+
+          # Free wagons paired with trains that can still take one (not a wagon, not 4D, none yet).
+          def wagon_choices(entity)
```
ollybh, line 104: This could potentially create sixteen choice buttons, for a company that has four trains and four wagons in phase 2 or 3. Not likely, and only possible through crazy play. Is that too many?

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+            wagons = entity.trains.select { |t| @game.wagon?(t) && !@round.wagon_for_train.value?(t) }
+            trains = entity.trains.select { |t| @game.wagon_attachable?(t) && !@round.wagon_for_train.key?(t.id) }
+            wagons.each_with_object({}) do |wagon, result|
+              trains.each do |train|
+                key = { 'type' => 'attach', 'wagon' => wagon.id, 'train' => train.id }
+                result[key] = "Attach #{wagon.name} to #{train.name}"
+              end
+            end
```
ollybh: I'm not sure if this is any clearer, but this is how I might have written this.
```suggestion
            wagons.product(trains).to_h do |wagon, train|
              key = { 'type' => 'attach', 'wagon' => wagon.id, 'train' => train.id }
              val = "Attach #{wagon.name} to #{train.name}"
              [key, val]
            end
```

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,137 @@ module Engine
...
+            result = {}
+
+            wagon_trains_at_harbor = entity.trains.select do |train|
+              @round.wagon_for_train.key?(train.id) &&
+                @round.current_routes[train.id]&.visited_stops&.any? { |s| @game.harbor?(s) }
+            end
+
+            wagon_trains_at_harbor.select { |train| @game.train_with_cubes?(train) }.each do |train|
+              result[{ 'type' => 'unload', 'train' => train.id }] = "Unload sugar from #{train.name}"
+            end
```
ollybh: For readability, I'd either create another intermediate variable:
```ruby
            sugar_at_harbor = wagon_trains_at_harbor.select { |train| @game.train_with_cubes?(train) }
            sugar_at_harbor.each do |train|
              result[{ 'type' => 'unload', 'train' => train.id }] = "Unload sugar from #{train.name}"
            end
```

Or write the loop as:
```ruby
            wagon_trains_at_harbor.each do |train|
              next unless @game.train_with_cubes?(train)
              
              result[{ 'type' => 'unload', 'train' => train.id }] = "Unload sugar from #{train.name}"
            end
```

lib/engine/game/g_18_cuba/sugar.rb
```diff
@@ -0,0 +1,152 @@
...
+
+        private
+
+        def reset_cubes_on_train
+          @cubes_on_train = {}
+        end
+
+        def extended_harbor_revenue(route, stops)
+          return 0 unless @round.wagon_for_train.key?(route.train.id)
+          return 0 unless route.train.distance.is_a?(Numeric)
```
ollybh: What is this line for? All of the 18Cuba trains have numeric distances.

lib/engine/game/g_18_cuba/sugar.rb
```diff
@@ -0,0 +1,152 @@
...
+          cubes_on_train(route.train).size * CUBE_VALUE
+        end
+
+        # Delivers the cubes (return to supply); reads the cubes explicitly loaded onto each train.
+        def collect_wagon_cubes(routes)
+          routes.each do |route|
+            cubes = cubes_on_train(route.train)
+            next if cubes.empty?
+
+            tally = cubes.tally
```
ollybh, line 105: It shouldn't be possible to submit a route that doesn't include a harbour, but it might still be worth checking here and throwing an error if you've somehow got a train with cubes loaded but nowhere to take them.

lib/engine/game/g_18_cuba/step/route.rb
ollybh: Also, a major company could just own a 4D train and a wagon. You don't want to give a choice in that case.

lib/engine/game/g_18_cuba/game.rb
```diff
@@ -305,51 +311,53 @@ def token_cost_override(entity, city, token)
...
+          # Each delivering wagon train must deliver to a different harbor (rule VII.10).
+          # Filter on delivering, not merely attached: empty wagons don't compete for a harbor delivery.
           super
-          wagon_routes = routes.select { |r| @round.wagon_for_train[r.train] }
-          return if wagon_routes.size <= 1
+          delivering_routes = routes.select { |r| train_with_cubes?(r.train) }
+          return if delivering_routes.size <= 1
 
-          harbors = wagon_routes.map { |r| r.visited_stops.find(&:offboard?)&.hex }.compact
+          harbors = delivering_routes.map { |r| r.visited_stops.find { |s| harbor?(s) }&.hex }.compact
```
ollybh: This test isn't going to work for routes that include two harbours.

I'm not sure what the correct interpretation of the rules is for routes like this – do all harbours have to be unique to the route, or could a harbour be shared between to routes, as long as they both include a different harbour as well?

ollybh (changes requested):
There are only a couple of things that look incorrect in this PR:
1. The choices offered when a major company owns a wagon but no standard trains, or only 4D trains.
2. The check for unique harbours when routes include two harbours.

I've made some other suggestions for changes, but none are major.

Could you add some screenshots to the PR description? Bonus points for a screencast showing the sugar cube transport UI. I've not played 18Uruguay, and I'm not entirely sure how this is supposed to look.

lib/engine/game/g_18_cuba/step/route.rb
chrstas: swapped the dup-and-<< for a non-mutating actions += %w[choose]. Line's gone.

lib/engine/game/g_18_cuba/step/route.rb
chrstas: There's already a route_trains override in game.rb that rejects wagons, so a company owning only a wagon has its route step skipped and the wagon doesn't show as a runnable train. For the 4D+wagon case, I've tightened choosing? to return true only when there's an attachable wagon/train pair or a wagon already attached — no empty attach prompt.

lib/engine/game/g_18_cuba/step/route.rb
chrstas: Switched to parentheses:  3+ (1w) instead of 3++1w.

lib/engine/game/g_18_cuba/step/route.rb
chrstas: done

lib/engine/game/g_18_cuba/step/route.rb
chrstas: done

lib/engine/game/g_18_cuba/step/route.rb
chrstas: done

lib/engine/game/g_18_cuba/step/route.rb
chrstas: adopted...

lib/engine/game/g_18_cuba/sugar.rb
chrstas: you are right - not required. Removed it here and the identical guard in check_distance.

lib/engine/game/g_18_cuba/step/route.rb
chrstas: Went with the intermediate variable (sugar_at_harbor) to keep it consistent with the other named filters in the method.

lib/engine/game/g_18_cuba/sugar.rb
chrstas: Added the check

lib/engine/game/g_18_cuba/game.rb
chrstas: You're right, `find` only looks at the first harbour. My reading of "each must run to a different harbour" is that each delivering wagon delivers to one harbour, so a combination is fine as long as distinct harbours can be assigned — two routes could share one if each also reaches another. I've implemented that as a distinct-representatives check over every harbour on each route. Will try to ask it on BGG with Lonny if this is ok or if it should be interpreted stricter.

lib/engine/game/g_18_cuba/step/route.rb
```diff
@@ -6,11 +6,141 @@ module Engine
...
-            # Currently always empty; check_distance and check_route_combination fall back to super.
-            super.merge({ wagon_for_train: {} })
+            super.merge(current_routes: {}, wagon_for_train: {})
+          end
+
+          def actions(entity)
+            actions = super
+            return actions if actions.empty?
+
+            actions += %w[choose] if choosing?(entity)
```
ollybh, line 26: You don't need to worry about changing the array returned from `super`.
```suggestion
            actions << 'choose' if choosing?(entity)
```

ollybh (commented):
## Review of latest changes (commit `0b44836c8`)

I have reviewed the latest version of this PR after the significant updates made in `0b44836c8`. The two blocking issues from the previous review have been addressed:

1. **Choices for wagon-only/4D-only entities** — ✅ `choosing?` now correctly guards against minors and returns `false` when no attachable trains exist (e.g., only a 4D or only a wagon).

2. **Unique harbor check** — ✅ Replaced with a recursive `distinct_harbors?` matching algorithm that correctly handles routes with harbors at both ends.

Other improvements I noted:
- Train name display uses parentheses (`3+ (1w)`) instead of double plus signs.
- `@game.train_by_id` used for robustness.
- `wagon_choices` refactored to `wagons.product(trains).to_h`.
- Unnecessary guards removed from `check_distance` and `extended_harbor_revenue`.
- New validation in `check_other` and `collect_wagon_cubes`.

### Suggested improvements (non-blocking)

I saved a detailed review report to `deepseek-pr-reviews/PR12779-review-2026-07-02.md`. A few suggestions (all **medium** or **low** priority):

1. **M1** — Guard `wagon_capacity` against nil `@round` (sugar.rb:78) to prevent potential `NoMethodError` outside Route step context.
2. **M2** — Skip log message when `sugar_production` yields 0 cubes to reduce game log noise.
3. **L1** — Consider optimizing the SVG icon (800×800px will be scaled down heavily).
4. **L3** — Simplify `first([count, 0].max)` to `first(count)` since count is always non-negative.

None of these are blockers. The PR looks solid — nice work, @chrstas! 🚀

ollybh (approved):
This looks OK now. But can you please add the screenshots, I'm still not certain I've understood the interface?

lib/engine/game/g_18_cuba/step/route.rb
chrstas: Agree - this should be improved. I have added this point as additional improvement on the checkllist #12485

lib/engine/game/g_18_cuba/step/route.rb
chrstas: I will add this in the next PR - would prefer not to change the PR now after approval. hope this is fine...

chrstas:
> This looks OK now. But can you please add the screenshots, I'm still not certain I've understood the interface?

@ollybh Thanks a lot for the review and approval. I have added the screenshots above to the PR description.

PR #12785, opened 2026-06-28
[1713 Menorca] feat: add prealpha prototype (runnable minimum)

puattmas:
## Before clicking "Create"

### Explanation of Change

Adds the 1713 Menorca prototype game (prealpha): meta, entities
(private companies + corporations), map, market, phases, tiles,
trains, and corporation logos.

Scoped to a runnable minimum: everything is plain data on top of the
base engine, so it loads, runs, and passes CI with nothing depending
on engine code that doesn't exist yet.

- entities.rb: COMPANIES holds the private companies as plain
  value/revenue cards (no abilities yet); CORPORATIONS holds RNC,
  RCC, RAM and CLV. RNC and CLV share Mao (J11), which has two slots.
- map.rb: sea hexes are white so routes can be built and connected
  across them (blue hexes are impassable in the engine). Lighthouses
  and watchtowers are white town hexes with water terrain.
- tiles.rb: standard engine tiles only - town (3, 4, 58), broad gauge
  (7, 8, 9, 20, 23-29, 40) and narrow gauge (77-79, 677, 678,
  692-695, 699).
- market.rb: 2D stock market, 9 rows, consistent price steps per
  column.
- trains.rb / phases.rb: a single 3H-8H train roster with matching
  phases E1-E6.

### Any Assumptions / Hacks

The train roster assumes all four corporations are opened (players
are expected to open CLV). With fewer corporations floated, the depot
can run out of train slots before the 8H becomes available.

Private company abilities, event-driven mechanics (CLV becoming
available), separate wagon/ship fleets, lighthouse progression and
the fixed-tile / Cami d'en Kane build restrictions are intentionally
left out of this first PR. They'll come back in follow-ups once the
supporting engine code exists.

### DIFF
--- a/lib/engine/game/g_1713_menorca/entities.rb
+++ b/lib/engine/game/g_1713_menorca/entities.rb
@@ -1,0 +1,127 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1713Menorca
+      module Entities
+        # rubocop:disable Layout/LineLength
+        COMPANIES = [
+          {
+            sym: 'GSC',
+            name: '1. Gremi de Sabaters de Ciutadella',
+            value: 20,
+            revenue: 5,
+            desc: '+10R to any route that connects Ciutadella with Barcelona, Mallorca, or Marsella. '\
+                  'This bonus applies to all corporations using that route if the owner has transferred the ability.',
+          },
+          {
+            sym: 'RAG',
+            name: "2. Ramats de l'Albufera d'es Grau",
+            value: 20,
+            revenue: 5,
+            desc: "+15R to any route that starts in S'Albufera d'es Grau. The corporation that acquires this private is the only one "\
+                  "that can lay S'Albufera d'es Grau's initial tile by paying the cost shown on the map.",
+          },
+          {
+            sym: 'CRB',
+            name: '3. Cami Reial Britanic',
+            value: 25,
+            revenue: 5,
+            desc: "The owning corporation ignores the additional build cost on Cami d'en Kane land hexes during phases 1 and 2.",
+          },
+          {
+            sym: 'LBI',
+            name: "4. Lleves de Bestiar de l'Interior",
+            value: 30,
+            revenue: 5,
+            desc: 'If owned by RAM, the corporation ignores factory blocking from other corporations.  If owned by any other corporation, this ability is inactive.',
+          },
+          {
+            sym: 'FRN',
+            name: '5. Factura Pendent amb la Royal Navy',
+            value: 100,
+            revenue: 10,
+            desc: 'During phase 2, owner may exchange this private for one RNC share at current market price without using the stock action. Closes when phase 3 begins.',
+          },
+          {
+            sym: 'DRM',
+            name: '6. Drassanes Reials de Mao',
+            value: 40,
+            revenue: 10,
+            desc: "Owning corporation buys its first ship in any phase at a 20R discount and can build Torre de l'Illa de l'Aire without additional cost.",
+          },
+          {
+            sym: 'CSF',
+            name: '7. Castell de Sant Felip',
+            value: 60,
+            revenue: 15,
+            desc: 'Once per game, owner may block one route into or out of the port of Mao for one full OR. The block must be declared at the start of the OR.',
+          },
+          {
+            sym: 'VRC',
+            name: '8. Vinculació Familiar amb la Real Compania',
+            value: 100,
+            revenue: 0,
+            desc: 'Upon acquisition, owner immediately receives the RCC family trust share (20%) at no cost. Then this private closes.',
+          },
+          {
+            sym: 'IVR',
+            name: '9. Intel·ligència de Versalles',
+            value: 35,
+            revenue: 10,
+            desc: 'Owner sees event cards one round before they trigger. Relevant only when using hidden events variant.',
+          },
+        ].freeze
+        # rubocop:enable Layout/LineLength
+
+        CORPORATIONS = [
+          {
+            sym: 'RNC',
+            name: 'Royal Navy Company',
+            logo: '1713Menorca/RNC',
+            simple_logo: '1713Menorca/RNC.alt',
+            tokens: [0, 80],
+            coordinates: 'J11',
+            color: '#1A3A5C',
+            float_percent: 50,
+            shares: [20, 10, 10, 10, 10, 10, 10, 10, 10],
+          },
+          {
+            sym: 'RCC',
+            name: 'Real Compania de Comercio',
+            logo: '1713Menorca/RCC',
+            simple_logo: '1713Menorca/RCC.alt',
+            tokens: [0, 80],
+            coordinates: 'C8',
+            color: '#7B2D00',
+            float_percent: 50,
+            shares: [20, 10, 10, 10, 10, 10, 10, 10, 10],
+          },
+          {
+            sym: 'RAM',
+            name: 'Ramaders i Artesans de Menorca',
+            logo: '1713Menorca/RAM',
+            simple_logo: '1713Menorca/RAM.alt',
+            tokens: [0, 30, 60],
+            coordinates: 'G8',
+            color: '#2D5A1B',
+            float_percent: 50,
+            shares: [20, 10, 10, 10, 10, 10, 10, 10, 10],
+          },
+          {
+            sym: 'CLV',
+            name: 'Compagnie du Levant',
+            logo: '1713Menorca/CLV',
+            simple_logo: '1713Menorca/CLV.alt',
+            tokens: [0, 80],
+            coordinates: 'J11',
+            city: 1,
+            color: '#4A235A',
+            float_percent: 50,
+            shares: [20, 10, 10, 10, 10, 10, 10, 10, 10],
+          },
+        ].freeze
+      end
+    end
+  end
+end
--- a/lib/engine/game/g_1713_menorca/map.rb
+++ b/lib/engine/game/g_1713_menorca/map.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1713Menorca
+      module Map
+        LOCATION_NAMES = {
+          'C8' => 'Ciutadella',
+          'H5' => 'Fornells',
+          'G8' => 'Es Mercadal',
+          'J11' => 'Maó / Es Castell',
+          'J13' => 'Sant Lluís',
+          'H9' => 'Alaior',
+          'E8' => 'Ferreries',
+          'F9' => 'Es Migjorn Gran',
+          'I6' => "Port d'Addaia",
+          'C6' => 'Punta Nati',
+          'C10' => "Torre d'Artrutx",
+          'G4' => 'Torre de Cavalleria',
+          'J7' => 'Cap de Favàritx',
+          'J15' => "Illa de l'Aire",
+          'J9' => "S'Albufera d'es Grau",
+          'B3' => 'Barcelona',
+          'A12' => 'Mallorca',
+          'J3' => 'Marsella',
+          'K4' => 'Génova',
+          'K16' => 'Valencia',
+          'F15' => 'Alger',
+        }.freeze
+
+        # rubocop:disable Layout/LineLength
+        HEXES = {
+          yellow: {
+            # Ciutats i pobles inicials amb via preimpresa
+            ['C8'] => 'city=revenue:20;path=a:4,b:_0;path=a:5,b:_0;path=a:1,b:_0,track:narrow;path=a:2,b:_0,track:narrow;icon=image:port', # Ciutadella
+            ['H5'] => 'city=revenue:10;path=a:0,b:_0;path=a:1,b:_0;icon=image:port', # Fornells
+            ['G8'] => 'city=revenue:20;path=a:2,b:_0;path=a:5,b:_0', # Es Mercadal
+            ['J11'] => 'city=revenue:30;path=a:0,b:_0;path=a:2,b:_0;path=a:4,b:_0,track:narrow;path=a:5,b:_0,track:narrow;icon=image:port', # Maó / Es Castell
+            ['J13'] => 'town=revenue:10;path=a:2,b:_0;path=a:3,b:_0', # Sant Lluís
+          },
+          orange: {
+            # Fars i torres de guaita (fites, representades com a pobles)
+            ['C6'] => 'town=revenue:0;upgrade=cost:20,terrain:water', # Punta Nati
+            ['C10'] => 'town=revenue:0;upgrade=cost:20,terrain:water', # Torre d'Artrutx
+            ['G4'] => 'town=revenue:0;upgrade=cost:20,terrain:water',   # Torre de Cavalleria
+            ['J7'] => 'town=revenue:0;upgrade=cost:20,terrain:water',   # Cap de Favàritx
+            ['J15'] => 'town=revenue:0;upgrade=cost:20,terrain:water', # Illa de l'Aire
+          },
+          white: {
+            # Mar obert (buildable, blue impassable would block track connections)
+            %w[B5 B7 B9 B11
+               C4 C12
+               D5 D11 D13
+               E4 E12
+               F3 F11
+               G2 G12
+               H3 H13
+               I14
+               K6 K8 K10 K12 K14] =>
+              '',
+            # Entrades marítimes cap a Alger i Marsella
+            %w[E14 F13 G14 I4 J5] => 'upgrade=cost:40,terrain:water',
+            # Camí especial interior (cost 60)
+            %w[D7 F7] => 'upgrade=cost:60,terrain:hill;path=track:future,a:1,b:5',
+            ['I10'] => 'upgrade=cost:60,terrain:hill;path=track:future,a:2,b:5',
+            # Terreny accidentat (cost 15 per travessar)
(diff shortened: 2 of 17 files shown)


chrstas:
Hi @jordisalord , 

I'm reaching out in parallel via Slack, but wanted to leave a note here too.

The protoype sounds thematically very interesting. A couple of things that would make it easier for the maintainers to review:
Could you fill in the [PR template](https://github.com/tobymao/18xx/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and use this as basis fort he PR description?

For the scope itself, I'd suggest getting it down to a runnable minimum first: loading with green CI, and taking out the abilities and events that don't have engine logic behind them yet - those can come back in a follow-up.

One more thing: I assume including a prototype on the platform needs Toby's approval - did you already get the OK from him?

chrstas:
Hi @jordisalord, please find below my first findings from the peer review targeting to get this first PR down to a runnable minimum:

My suggestion: make this PR 1 the game that **loads, runs, and passes CI** — and nothing more. Everything else comes back in follow-ups, once the engine code that backs it exists.

### Fold `companies.rb` + `privates.rb` into one `entities.rb`

The names are currently swapped (`companies.rb` holds `CORPORATIONS`), and `privates.rb` isn't a filename this repo uses. In engine terms there are no "privates": what you call privates *are* `COMPANIES`. Propose to use a single `entities.rb`:

```ruby
module Entities
  COMPANIES    = [...].freeze   # currently privates.rb
  CORPORATIONS = [...].freeze   # currently companies.rb
end
```

### Take out the config that has no engine behind it

**`map.rb`** - these ten constants aren't referenced anywhere in the PR, and they aren't base-engine names either, so nothing outside can read them. 

`WATER_HEXES`, `LIGHTHOUSE_HEXES`, `KANE_HEXES`, `KANE_ENDPOINT_HEXES`, `KANE_BONUS`, `GSC_TARGET_OFFBOARD_HEXES`, `ALBUFERA_HEX`, `ILLA_AIRE_HEX`, `FIXED_TILES_BY_HEX`, `NON_ROTATABLE_TILE_PREFIXES`

**The private companies' abilities** - I would reduce them to plain value/revenue cards, removing any piece of ability for the moment

**`trains.rb`** - the V4 carries `events: [{ 'type' => 'clv_available' }]`, but there's no `event_clv_available` in `game.rb`. Propose to remove this for the moment

**`phases.rb`** - status strings `era_e1`…`era_e4` with no matching `STATUS_TEXT` entries in `game.rb`.

### What stays

Map hexes, tiles, market, phases and trains as pure data. 

Hope this helps...

chrstas:
### `tiles.rb` — most of these tiles already exist in the engine

The standard tile catalogue is in [`lib/engine/config/tile.rb`](https://github.com/tobymao/18xx/blob/master/lib/engine/config/tile.rb). Games reference them by name and count:

```ruby
TILES = {
  '7' => 6,
  '8' => 6,
  '9' => 6,
  '23' => 2,
  # ...
}.freeze
```

Looks like your plain-track tiles are re-creations of standard tiles. Rotation is free in the engine, so a rotated definition is the same tile:

| yours | standard | | yours | standard |
|---|---|---|---|---|
| `E1T3` | `7` | | `E1W3` | `77` |
| `E1T2` | `8` | | `E1W2` | `78` |
| `E1T1` | `9` | | `E1W1` | `79` |
| `E2T1` | `20` | | `E2W1` | `677` |
| `E2T2` | `23` | | `E2W2` | `678` |
| `E2T3` | `24` | | `E2W3` | `699` |
| `E2T4` | `25` | | `E2W4` | `692` |
| `E2T5` | `26` | | `E2W5` | `693` |
| `E2T6` | `27` | | `E2W6` | `695` |
| `E2T8` | `28` | | `E2W7` | `694` |
| `E2T7` | `29` | | | |
| `E3T1` | `40` | | | |

The engine has a full narrow-gauge family as well (`77`/`78`/`79`, `677`/`678`/`692`–`695`/`699`, `84`–`86`, `646`–`649`). Using the standard names gives you rendering, upgrade paths and rotation for free.

### What to do with the sea

Your sea hexes are pre-printed `blue:`, and blue hexes are dropped from the neighbour graph (`IMPASSABLE_HEX_COLORS`, `base.rb:2890`) — so nothing can be laid on them or connect to them, whatever colour the tile is.

1. `map.rb`: move the sea hexes from the `blue:` block to `white:`
2. `tiles.rb`: drop `E1W*` / `E2W*` and use the standard narrow-gauge tiles instead — `'77'`, `'78'`, `'79'` (yellow) and `'677'`, `'678'`, `'692'`–`'695'`, `'699'` (green), with your counts.
3. `phases.rb`: no change needed — those tiles are already yellow and green.

That gives you buildable sea routes with no new engine code. Keeping the sea blue is possible later, but it needs overrides (separate PR)

puattmas:
Thanks for the detailed review, @chrstas — really helpful. I've reworked the branch down to the runnable minimum you outlined and squashed it into a single commit so it's easy to re-review:

- **`entities.rb`**: merged `companies.rb` + `privates.rb`, fixed the swapped naming — `COMPANIES` now holds the private companies, `CORPORATIONS` the corporations.
- **Abilities removed**: the private companies are now plain value/revenue cards, no `abilities:` left.
- **`map.rb`**: dropped the ten unused constants (`WATER_HEXES`, `LIGHTHOUSE_HEXES`, `KANE_*`, `GSC_TARGET_OFFBOARD_HEXES`, `ALBUFERA_HEX`, `ILLA_AIRE_HEX`, `FIXED_TILES_BY_HEX`, `NON_ROTATABLE_TILE_PREFIXES`) — they belong to engine logic that isn't in this PR yet.
- **`trains.rb`**: removed the V4 `clv_available` event (no handler in this PR).
- **`phases.rb`**: removed the `era_e1`…`era_e4` status strings (no matching `STATUS_TEXT`), kept `can_buy_companies`.
- **Sea hexes**: moved from `blue:` to `white:` in `map.rb`, since blue is impassable in the engine regardless of tile.
- **Tiles**: replaced the custom plain-track/narrow tiles with the standard catalogue tiles you pointed to (`7,8,9,20,23-29,40,77-79,677,678,692-695,699`), same counts.

Verified locally: the game loads cleanly, `rubocop` passes with no offenses, and the engine specs (`config_spec`, `fixtures_required_spec`) pass.

Note: I force-pushed this branch to squash everything into the single commit above, since I wanted this first PR to start clean rather than carry the earlier back-and-forth. I won't force-push again from here on — happy to keep iterating with regular commits on top.

lib/engine/game/g_1713_menorca/entities.rb
```diff
@@ -0,0 +1,127 @@
...
+            float_percent: 50,
+            shares: [20, 10, 10, 10, 10, 10, 10, 10, 10],
+          },
+          {
+            sym: 'CLV',
+            name: 'Compagnie du Levant',
+            logo: '1713Menorca/CLV',
+            simple_logo: '1713Menorca/CLV.alt',
+            tokens: [0, 80],
+            coordinates: 'J11',
```
chrstas, line 117: RNC (entities.rb:84) and CLV (entities.rb:117-118) are both homed on J11, which has one city with one slot (map.rb:38). `city:` is a city index, not a slot count, and on a single-city tile it's forced to 0 (tile.rb:379) - so CLV competes for RNC's slot, gets none, and falls back to a tile-level reservation. That sends `place_home_token` into a `pending_tokens` the default operating round doesn't have, the first time CLV operates:

```
NoMethodError: undefined method `pending_tokens'
```

Are RNC and CLV meant to share Maó?

lib/engine/game/g_1713_menorca/market.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Engine
+  module Game
+    module G1713Menorca
+      module Market
+        MARKET = [
+          ['', '', '120', '135', '150', '170', '190', '215', '240', '270', '305', '345', '390'],
```
chrstas, line 8: ### Market grid: row 0 is offset by two columns

Row 0 of `MARKET` (`market.rb:8`) is indented by two cells, but its values are the ones that belong two columns further left — 120, 135, 150 are the steps above 90, 120, 135. `TwoDimensionalMovement` addresses cells as plain `[row, col]`, so column 2 of row 0 holds 120 while the cell below it holds 135. (1871 has the same two-cell indent in row 0 and stays consistent: 111 sits above 100.)

Every move across that boundary is inverted. And `log_share_price` reads the direction off the row index rather than the price, so being sold out at 135 at row 1 logs `share price moves up from 135R to 120R`.

```
up:    [1,2] 135 -> 120   [1,3] 150p -> 135   [1,4] 170 -> 150
       [1,5] 190 -> 170   [1,6] 215 -> 190    [1,7] 240 -> 215   [1,8] 270 -> 240
down:  [0,2] 120 -> 135   [0,3] 135 -> 150p   [0,4] 150 -> 170   ...
```

Dropping the two leading `''` lines the whole grid up, and gives 90 and 120 a cell above them as a side effect. Is that what row 0 was meant to be?

lib/engine/game/g_1713_menorca/map.rb
```diff
@@ -0,0 +1,92 @@
...
+        HEXES = {
+          yellow: {
+            # Ciutats i pobles inicials amb via preimpresa
+            ['C8'] => 'city=revenue:20;path=a:4,b:_0;path=a:5,b:_0;path=a:1,b:_0,track:narrow;path=a:2,b:_0,track:narrow;icon=image:port', # Ciutadella
+            ['H5'] => 'city=revenue:10;path=a:0,b:_0;path=a:1,b:_0;icon=image:port', # Fornells
+            ['G8'] => 'city=revenue:20;path=a:2,b:_0;path=a:5,b:_0', # Es Mercadal
+            ['J11'] => 'city=revenue:30;path=a:0,b:_0;path=a:2,b:_0;path=a:4,b:_0,track:narrow;path=a:5,b:_0,track:narrow;icon=image:port', # Maó / Es Castell
+            ['J13'] => 'town=revenue:10;path=a:2,b:_0;path=a:3,b:_0', # Sant Lluís
+          },
+          orange: {
```
chrstas, line 41: ### The five tower hexes can't be built on

C6, C10, G4, J7 and J15 sit in the `orange:` block (`map.rb:41-48`), while the tiles drawn for them are yellow through gray - and `upgrades_to_correct_color?` only allows a lay one colour-index further on, so orange → yellow is already illegal:

```
C6 -> []   C10 -> []   G4 -> []   J7 -> []   J15 -> []
```

Would it make sense to take the towers out of this PR? It looks like a fairly detailed mechanic to me. Could you say a bit more about how they work on the rules side? I'd be interested to read how you picture them.

chrstas (commented):
Thanks a lot for picking up the points...

Going now through the new commit I add my peer review points that I found. hope it helps.

### Phases E3 and E4 can't be reached: not enough train slots on the table

`Depot#available_upcoming_trains` (depot.rb:109) offers the head of `@upcoming` plus anything whose `available_on` phase has been reached, and a train only leaves the depot by being bought. The wagons sit in front of the ships, so V4 only becomes head once all 18 trains ahead of it are gone (C2 ×5, CM ×4, V2 ×5, V3 ×4) — and of those only the five C2 ever rust on the way, since V2 waits for V4 and V3 waits for VE. That leaves 13 to be held at once against `train_limit` 3 × 4 corporations = 12 (`lib/engine/step/train.rb:30`, per corporation).

Driving the depot with `buy_train` + `phase.buying_train!`, 17 of the 18 go through and it stalls at 12/12 in E2 with one V3 left. From there brown and gray tiles are never placeable, V2 and V3 never rust, the V4 and VE are dead stock, and MO3's second token slot never appears.

`available_on` isn't gating this: it's circular on V4 and VE, since the phase it names only arrives once that train has been bought, and the head-of-queue clause hands out the depot head regardless.

How did you picture the two fleets sharing the train limit?

lib/engine/game/g_1713_menorca/map.rb
puattmas: It can be changed to white, the same we did with blue tiles. The thing is: water+lighthouse only use 'narrow gauge' (which means ships). If I put everything on white tiles it doesn't look like and island hehe, but it's perfect for the alpha version.

puattmas: Would it be acceptable to solve all the water/lighthouses problems with this?

...            ['C6'] => 'town=revenue:0;upgrade=cost:20,terrain:water;label=PN', # Punta Nati
            ['C10'] => 'town=revenue:0;upgrade=cost:20,terrain:water;label=TA', # Torre d'Artrutx
            ['G4'] => 'town=revenue:0;upgrade=cost:20,terrain:water;label=TC',   # Torre de Cavalleria
            ['J7'] => 'town=revenue:0;upgrade=cost:20,terrain:water;label=CF',   # Cap de Favàritx
            ['J15'] => 'town=revenue:0;upgrade=cost:20,terrain:water;label=IA', # Illa de l'Aire
            ...
          '77' => { 'count' => 10, 'color' => 'yellow', 'code' => 'path=a:0,b:1,track:narrow;label=WA' },
          '78' => { 'count' => 10, 'color' => 'yellow', 'code' => 'path=a:0,b:2,track:narrow;label=WA' },
          '79' => { 'count' => 10, 'color' => 'yellow', 'code' => 'path=a:0,b:3,track:narrow;label=WA' },
          '677' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:3,track:narrow;path=a:0,b:4,track:narrow;label=WA' },
          '678' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:3,track:narrow;path=a:0,b:2,track:narrow;label=WA' },
          '692' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:3,track:narrow;path=a:0,b:5,track:narrow;label=WA' },
          '693' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:3,track:narrow;path=a:0,b:1,track:narrow;label=WA' },
          '694' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:4,track:narrow;path=a:0,b:5,track:narrow;label=WA' },
          '695' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:2,track:narrow;path=a:0,b:1,track:narrow;label=WA' },
          '699' => { 'count' => 4, 'color' => 'green', 'code' => 'path=a:0,b:2,track:narrow;path=a:0,b:4,track:narrow;label=WA' },
          ...


          I know I'm generating tiles that already exists, but I can't find another way to limit water tiles (now in the white section of map.rb)

lib/engine/game/g_1713_menorca/entities.rb
puattmas: I fixed it by changing the initial tile, which is now size 2. Thx!

lib/engine/game/g_1713_menorca/map.rb
chrstas: Two different problems as it turns out:

**Water.** The hole you've spotted is real. My suggestion, and it's only that: don't build it here. It means re-declaring tiles and labelling hexes, which pulls in the opposite direction from getting this PR small. The nicer answer is to go back to blue for the sea in a follow-up and make it buildable there, that gives you the restriction and the island look in one go, instead of lettering as a stand-in for both. For now the map is permissive, which it already is today, and nothing about it is wrong in a way that blocks play.

**Lighthouses.** The label doesn't reach them either way, because the colour is checked before the label. `orange -> yellow` still isn't a legal colour step, so C6 stays unbuildable. That one needs its own decision, and it's yours:

- **white** keeps the four upgrade stages, loses the orange fill. The `terrain:water` marker still renders, so they won't read as plain sea.
- **orange as a fixed landmark** pre-printed narrow paths and a fixed revenue, no upgrade stages, and the 20 tower tiles go. Orange isn't impassable here, so ships still reach them; that's how 21Moon and 18ESP use it.
- **orange with the progression** keeps both, but needs `upgrades_to_correct_color?` overridden in `game.rb`. Four games do that; it's the only option needing engine code, so I'd keep it for a follow-up.

Which fits depends on what a tower is in the rules: a stop that grows, or a fixed landmark. That I can't read off the data.

puattmas: Ups! I forgot to tell you that I put them in 'white' section. So the would be white, but with labels. Anyway, I'm leaving this topic for a next PR :)

lib/engine/game/g_1713_menorca/entities.rb
chrstas: Confirmed: with Maó at two slots, RNC and CLV both place their home token at J11 in either order, and the operating rounds after that run without errors. That was exactly the right fix, thanks!

lib/engine/game/g_1713_menorca/market.rb
chrstas: @puattmas: The new grid checks out: prices rise from left to right in every row and fall from top to bottom in every column, and none of the 332 moves (83 cells, four directions each) goes the wrong way. Writing the leading blank cells out explicitly makes the layout much easier to follow. Thanks...

lib/engine/game/g_1713_menorca/map.rb
chrstas: White works: all five tower hexes (C6, C10, G4, J7, J15) now take yellow tiles. Agreed on leaving the rest of the tower mechanic for a follow-up PR, that keeps this one small.

PR #12787, opened 2026-06-30
[1835] corps must buy trains

kepakiano:

### Explanation of Change

Pretty straight forward, forces corporations to own trains, but minors are exempt.

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -50,6 +50,8 @@ module Engine
 
         EBUY_DEPOT_TRAIN_MUST_BE_CHEAPEST = false
 
+        MUST_BUY_TRAIN = :always
+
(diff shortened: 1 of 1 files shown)


neutronc (commented):
Minors (the Preprussian) dont need to own a train. You can postpone buyying a train.

So why did you delete that exception ?

kepakiano:
> Minors (the Preprussian) dont need to own a train. You can postpone buyying a train.
> 
> So why did you delete that exception ?

I'm not following, what did I delete?

neutronc:
the check that Minor does not need to buy a train
your second commit deleted the check if the corp is a minor if i read this correctly ?

kepakiano:
> the check that Minor does not need to buy a train your second commit deleted the check if the corp is a minor if i read this correctly ?

yeah, because the first commit added it in the wrong spot when it was already there in the buy_trains step :) The total change of the PR is just one line.

ollybh (approved):

PR #12804, opened 2026-07-17
[1849] LoanChoice step doesn't trigger for last corp in OR set.

philcampeau:
## Before clicking "Create"

### Explanation of Change

User [here in slack](https://18xxgames.slack.com/archives/CV3R3HPUZ/p1784292714318009) reported that he wasn't getting the choice to take a loan after bankrupting. I noticed he was operating the last corp in the OR set, which means that the `finished?` method would trigger before the LoanChoice step could register the situation. 

Added a check to the `finished?` method in g_1849/round/operating.rb to prevent this from happening.

### DIFF
--- a/lib/engine/game/g_1849/round/operating.rb
+++ b/lib/engine/game/g_1849/round/operating.rb
@@ -12,6 +12,12 @@ module Engine
 
             super
           end
+
+          def finished?
+            return false if @game.loan_choice_player
+
+            super
+          end
         end
       end
     end


lib/engine/game/g_1849/round/operating.rb
```diff
@@ -12,6 +12,12 @@ def next_entity!
 
             super
           end
+
+          def finished?
+            return false if @game.loan_choice_player
```
chrstas, line 17: Would this need the `pins` label, or is it migratable? On master the round rolls into the stock round after the bankruptcy, so an affected game has stock round actions recorded next. Replaying such a game with this change, the first of those actions is rejected; the replay's message is `Blocking step Bankruptcy Choice cannot process action pass at 203`. The choice returns in the following OR, so the `choose` is in the log unless the game ended with that OR.

lib/engine/game/g_1849/round/operating.rb
```diff
@@ -12,6 +12,12 @@ def next_entity!
 
             super
           end
+
+          def finished?
```
chrstas, line 16: Is the trigger really the last slot in the OR set? The code looks right either way, this is about **the title and the description of the PR**. Moving the bankrupt corporation to the end on master, the round stays open and `active_step` is still `LoanChoice`. What does end the round is a different case: no open corporation is left in the set, so `any_to_act?` goes false although the blocking step is active. The guard sits in front of both, so it covers this too.

chrstas (commented):
hi @philcampeau, peer review with two findings. Code looks fine. The first is that an existing game that already hit this bug no longer replays after the change. The second is about the trigger condition, not the fix itself. 

Checked and fine: with the round held open the player is offered the choice and the round advances once it is answered, and the bond steps cannot interfere. I did not reconstruct how a real game runs out of open corporations in the set.

lib/engine/game/g_1849/round/operating.rb
philcampeau: yes, thanks, this was supposed to have the `pins` tag

lib/engine/game/g_1849/round/operating.rb
philcampeau: solid point. I updated the title and added an edit to the description

PR #12809, opened 2026-07-20
[core] Create new Setup Mode and Panel

GreatA1exander:
## Before clicking "Create"

- [ x ] Branch is derived from the latest `master`
- [ x ] Add the `pins` or `archive_alpha_games` label if this change will break existing games
- [ x ] Code passes linter with `docker compose exec rack rubocop -a`
- [ x ] Tests pass cleanly with `docker compose exec rack rake`

## Implementation Notes

**The core: a "god-move" action.** 18xx.games is event-sourced — a game is its title,
settings, players, and an ordered action log, replayed to derive state. So a preset is
expressed the only way state can be: as actions. This adds one privileged action,
`Engine::Action::Setup` (`lib/engine/action/setup.rb`), applied by a **non-blocking**
`Engine::Step::Setup` (`lib/engine/step/setup.rb`) that sits in every round via
`Round::Base::DEFAULT_STEPS`. Because `Setup#free?` is `true`, it bypasses the usual
step-legality and turn checks and mutates state directly — but it's still an ordinary logged
action, so it serializes, round-trips through export/import, undoes, and replays
deterministically like anything else.

**One directive key per facet.** The action carries any subset of directive fields
(`cash`, `par`, `market`, `shares`, `trains`, `rust`, `phase`, `tiles`, `remove_tiles`,
`tokens`, `companies`, `loans`, `advance`, `extensions`). `Step::Setup#process_setup`
applies them in a fixed order. Each handler **reuses the engine's own high-level methods**
rather than hand-rolling mutations, so derived inventory (share pool, depot, tile pool,
routing graph) stays consistent — e.g. `set_cash` for money (auto-balances the bank),
`stock_market.set_par`/`move`, `share_pool` transfers, `hex.lay` + `update_tile_lists` +
graph rebuild for tiles, `place_token`/`place_home_token` for tokens.

**Round positioning.** The `advance` directive drops the game at a clean boundary — the start
of a Stock or Operating round with a chosen turn/OR number, priority deal, and exact player
order — so a preset resumes at a real turn instead of mid-opening-auction.

**Frontend (Opal/Snabberb).** A new `SetupEditor` form component
(`assets/app/view/game/setup_editor.rb`) offers a panel per facet, plus radial tile
lay/delete directly on the Map tab. Every edit dispatches through the existing
`Actionable#process_action` path, so hotseat writes localStorage and multiplayer POSTs the
action — no new dispatch machinery. The editor is gated behind a "Setup Mode" toggle in the
Tools tab (like Master Mode) and is offered on hotseat games and owned, started multiplayer
games.

**Extensibility.** Titles with mechanics beyond the generic facets opt in by overriding
`Game::Base#process_setup_extension`, which receives the step (with public id-lookup helpers)
plus the directive key and payload. The default raises a clear error. Documented in
`docs/setup_extensions.md`.

**Testing.** `spec/lib/engine/game/setup_spec.rb` exercises every facet with both direct
replay and a JSON serialize/parse round-trip (the import path) across a representative sample
(1830, 1846, 1889, 1817, and the 1871 prototype). `spec/assets_spec.rb` renders the setup
game and the editor through the Opal-compiled engine under mini_racer to catch render/compat
issues. The full regression (`rake`: 9,312 examples + RuboCop on 2,514 files) passes — the
change is additive, so no existing fixture is affected.

### Explanation of Change
When playing a game with some friends a couple of weeks ago, we had to call it halfway through and really wanted to finish the game in .games, but found that there was no way to create an arbitrary state in the existing engine. I decided that I wanted to do something about that to have the option in the future. This will enable players to port their half finished games into .games and also to create fun new states and potentially puzzle games that would be interesting for community engagement.

This change introduces a new setup mode and panel that can be enabled from the tools panel section. 

Each edit is recorded as an ordinary `setup` action in the game log, so a preset position serializes, survives export/import, and replays deterministically like any other game — the feature is purely additive, and the full fixture suite (9,312 examples) still passes, so no existing game is affected.

We currently add support for:
1. Map Editing with arbitrary tiles and tile deletion
2. Setting cash for any corporation, player, or the bank (money is kept conserved automatically)
3. Paring a corporation at any valid price and assigning its president
4. Moving a corporation to any stock-market cell
5. Granting shares to players or into the market pool, at any percentage
6. Assigning specific trains to corporations
7. Rusting/retiring a train type everywhere (e.g. after advancing the phase)
8. Advancing the game to any phase
9. Placing tokens — home tokens, or on an explicit hex/city
10. Assigning private companies to players or corporations, and closing them
11. Taking loans for corporations (in loan games such as 1817)
12. Round positioning — dropping the game into a Stock or Operating round with a chosen turn/OR number, priority deal, and exact player order
13. A per-game extension hook (`process_setup_extension`) so individual titles can add their own bespoke setup directives

### Screenshots
Screenshot 2026-07-19 at 9 51 11 PM
All the options in the tab.

Screenshot 2026-07-19 at 9 51 29 PM
A recreated game with the ability to edit the map.

Screenshot 2026-07-19 at 9 51 41 PM
Showing where the option is in the tools menu.

Screenshot 2026-07-19 at 11 03 24 PM
Showing that we can place the straight tile in 1871, proving the extensibility demo that we have in 1871.

### Any Assumptions / Hacks
This covers the full set of capabilities needed for 1830 and 1830-like 18xx games. It won't fully handle titles with bespoke mechanics — e.g. 1880, whose building permits use special pseudo-shares and a par-location float process — since those need per-game setup logic. Rather than special-case every title, the design exposes a per-game extension hook (`Game::Base#process_setup_extension`, documented in `docs/setup_extensions.md`) so game authors can add their own setup directives. I'm hoping that lowers the bar enough to encourage designers to support setup for their games. I also plan on implementing some further games myself.


tobymao:
sorry, but this will not be accepted. it's too complicated

PR #12817, opened 2026-07-27
Use short stubs for terminal track on city hexes

philcampeau:
## Before clicking "Create"

### Explanation of Change

[In a discussion](https://groups.io/g/18xx/topic/120463190) on the 18xx groups.io page, it was pointed out that it's hard to see the tapered end of terminal track on hexes with Cities on them. 

I've introduced code here that shortens the taper if the hex has a City component on it, using the same dimensions (but not coloring) as the terminal: 2 taper.

### Screenshots

before: 

image

After: 

image

### DIFF
--- a/assets/app/view/game/part/track_node_path.rb
+++ b/assets/app/view/game/part/track_node_path.rb
@@ -442,21 +442,34 @@ module View
 
           # terminal tapered track only supported for centered city/town
           if @terminal
-            props[:attrs].merge!(
-              transform: "rotate(#{rotation})",
-              d: "M #{terminal_start_x} 70 L #{terminal_start_x} 87 L #{terminal_end_x} 87 "\
-                 "L #{terminal_end_x} 70 L #{point_x} 35 Z",
-              fill: color,
-              stroke: 'none',
-              'stroke-linecap': 'butt',
-              'stroke-linejoin': 'miter',
-              'stroke-width': (width.to_i * 0.75).to_s,
-              'stroke-dasharray': dash,
-            )
-            if @terminal == '2'
-              props[:attrs][:d] = "M #{terminal_start_x} 85 L #{terminal_start_x} 87 L #{terminal_end_x} 87 "\
-                                  "L #{terminal_end_x} 85 L #{point_x} 65 Z"
-              props[:attrs][:fill] = '#707070'
+            # Cities get the short stub automatically, regardless of terminal:1 vs terminal:2.
+            # Only terminal:2 (mountain/ignored dead-ends) keeps the grey fill.
+            short_taper = @terminal == '2' || @stop0&.city?
+
+            if short_taper
+              props[:attrs].merge!(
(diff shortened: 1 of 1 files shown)


assets/app/view/game/part/track_node_path.rb
```diff
@@ -442,21 +442,34 @@ def build_props(color, width, dash)
...
-            )
-            if @terminal == '2'
-              props[:attrs][:d] = "M #{terminal_start_x} 85 L #{terminal_start_x} 87 L #{terminal_end_x} 87 "\
-                                  "L #{terminal_end_x} 85 L #{point_x} 65 Z"
-              props[:attrs][:fill] = '#707070'
+            # Cities get the short stub automatically, regardless of terminal:1 vs terminal:2.
+            # Only terminal:2 (mountain/ignored dead-ends) keeps the grey fill.
+            short_taper = @terminal == '2' || @stop0&.city?
+
+            if short_taper
```
chrstas, line 449: The two branches repeat six of eight attributes; only `d` and `fill` differ, so they'd fold into a single hash with `taper_y`/`point_y` as locals.

Or: keep master's merge-then-overwrite, widen the condition, and leave the grey fill on `terminal: '2'` only. That one's a smaller diff but stays with the merge-then-overwrite shape.

No behaviour change either way, so it's your call.

assets/app/view/game/part/track_node_path.rb
```diff
@@ -442,21 +442,34 @@ def build_props(color, width, dash)
...
-              'stroke-width': (width.to_i * 0.75).to_s,
-              'stroke-dasharray': dash,
-            )
-            if @terminal == '2'
-              props[:attrs][:d] = "M #{terminal_start_x} 85 L #{terminal_start_x} 87 L #{terminal_end_x} 87 "\
-                                  "L #{terminal_end_x} 85 L #{point_x} 65 Z"
-              props[:attrs][:fill] = '#707070'
+            # Cities get the short stub automatically, regardless of terminal:1 vs terminal:2.
+            # Only terminal:2 (mountain/ignored dead-ends) keeps the grey fill.
+            short_taper = @terminal == '2' || @stop0&.city?
```
chrstas, line 447: I had a look at 1822PNW: here we have cities declared with `slots: 0` which draw no circle: the 1822PNW/1822MX port tiles P1/P2 are that case (`g_1822_pnw/map.rb:342,349`, `g_1822_mx/map.rb:348,355`). Every spike they can be laid on is a `terminal: '1'` junction today, drawing the long taper, so it shortens at the moment the port is placed - on a hex with nothing in the middle either way.

old:
Image

new:
Image

Hexes D9 and I10
Image

Not saying it's wrong; it works for me. More of an observation - were these on your list?

chrstas (commented):
hi @philcampeau - me again ;-) From a peer review perspective two observations, neither blocking - one on the port tiles, the other on the shape of the new `if`/`else`.

PR #12821, opened 2026-07-30
[Core] - Make game name on lobby tile clickable

### LINKED ISSUE #12741: make game name on the tile clickable
**What's your idea?**

Make game names clickable on the tiles: link to the wiki. on the running and especially on starting games tiles.

If I do not know the exact rules easier to figure it out this way.

Thanks.

ddaybell:
Fixes #12741

**Size:** 1 file changed, +13/-4 lines

## Before clicking "Create"

- [ ] Tests pass cleanly with `docker compose exec rack rake` (see note below - ran the fixture specs instead)

### Explanation of Change

The game title on each lobby tile ("Game: <title>") was plain text. Every game already
defines a `GAME_INFO_URL` on its `meta.rb` pointing at that game's wiki page (the same
constant `game/game_meta.rb`'s "More info" link already uses), so this makes just the
title portion a clickable, underlined link to that URL, opening in a new tab. The
"Game: " prefix and the dev-stage suffix stay as plain text. Games that don't define
`GAME_INFO_URL` (a minority) fall back to the current plain-text behavior with no link.

### Screenshots

image

### Any Assumptions / Hacks

None

### DIFF
--- a/assets/app/view/game_card.rb
+++ b/assets/app/view/game_card.rb
@@ -126,13 +126,22 @@ module View
       }
 
       dev_status = game.meta::DEV_STAGE
+      dev_suffix = dev_status != :production ? " (#{dev_status})" : ''
+
+      title_elm =
+        if game.meta::GAME_INFO_URL
+          link_props = {
+            attrs: { href: game.meta::GAME_INFO_URL, target: '_blank' },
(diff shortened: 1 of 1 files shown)


philcampeau (approved):
looks good to me!

PR #12823, opened 2026-07-30
[Market page] Add a large-cell toggle on 2D markets

### LINKED ISSUE #12801: 18zoo - Translucent token
I ask if possible to have token translucent instead of opaque so that the numbers on all cells are visible.
https://18xx.games/game/257595#market

Image

philcampeau:
Fixes #12801

## Before clicking "Create"

### Explanation of Change
The user was reporting that the tokens in the stock market in 18Zoo were blocking the view of some information. This is because 18Zoo includes more information in those cells than most games. 

That said, I thought it could be useful to have a button toggle that increases the dimensions of cells in 2D markets for all games. I could see it being useful when multiple tokens start clumping together in games, for example. 

### Screenshots

Normal size:
image

Large Cells toggled on:
image

### DIFF
--- a/assets/app/view/game/stock_market.rb
+++ b/assets/app/view/game/stock_market.rb
@@ -46,14 +46,49 @@ module View
 
       # 2D markets
       HEIGHT_TOTAL = 50
-      TOKEN_PAD = 3                               # left/right padding of tokens within box
-      BOX_WIDTH = WIDTH_TOTAL - (2 * BORDER)
-      LEFT_MARGIN = TOKEN_PAD                     # left edge of leftmost token
-      RIGHT_MARGIN = BOX_WIDTH - TOKEN_PAD        # right edge of rightmost token
-      LEFT_TOKEN_POS = LEFT_MARGIN
-      RIGHT_TOKEN_POS = RIGHT_MARGIN - TOKEN_SIZE # left edge of rightmost token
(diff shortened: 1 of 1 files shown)


ddaybell:
Looks nice.  Something else that would be nice would be a way to click on a cell in the stock market, and have it expand the giant stack of tokens in the cell so you knew which ones were there.  Thinking of games like 1822 where a ton of minors all float at the same time at the same price.  1866 sometimes has that problem with figuring out whose stock tokens are in a given cell.  If you don't want to combine that idea with this one, I could post this as a separate enhancement.  Heck, I could probably program that, too! :-)

philcampeau:
> Looks nice. Something else that would be nice would be a way to click on a cell in the stock market, and have it expand the giant stack of tokens in the cell so you knew which ones were there. Thinking of games like 1822 where a ton of minors all float at the same time at the same price. 1866 sometimes has that problem with figuring out whose stock tokens are in a given cell. If you don't want to combine that idea with this one, I could post this as a separate enhancement. Heck, I could probably program that, too! :-)

yeah that could be good, but I think it would be better as a separate PR!

crericha:
Can you help me understand why its a toggle and not fixed in the game config? imho it seems game specific, not user preference.

philcampeau:
Like I wrote in the description, I could see it being useful when multiple tokens start clumping together in games. Plus, that way it's not something that needs to be coded in on a game-by-game basis.

PR #12829, opened 2026-08-06
[1835][1847AE][core] adds test for vanilla draft, cert packages, and nationalization of 1835, add module for nationalization of shares

kepakiano:
refers to #3059

## Before clicking "Create"

### Explanation of Change

Starting to add tests to make collaborating on 1835 easier. For now, there are only tests written entirely by hand. Since we cheat in those tests, we can test stuff with very few actions, hopefully making these tests more resilient than actual game fixtures. 

There are also two fixes:
- the previous attempt for the nationalization code changed the price of the bundle that's getting nationalized. This sometimes was executed twice, so the price was wrong. Now, the code of 1847AE has been pulled into a module and reused
- after the draft round bumping the next entity counter would bump it one too many times. Which is why now we only do it if we're still in the draft round. I'm not very versed on when one should let the engine bump the counter and when one's implementation should do it, so I'd welcome any input on this

And one minor change: Storing whether the draft is complete is unnecessary, the game can just evaluate that itself.


lib/engine/game/g_1835/round/draft.rb
```diff
@@ -18,23 +18,16 @@ def after_process(_action)
           end
 
           def next_entity!
-            next_entity_index!
-            if finished?
-              @game.draft_finished = all_drafted?
-              return
-            end
+            next_entity_index! if active_step
```
chrstas: Can this condition ever be true? `after_process` four lines up calls `next_entity!` only when
`active_step` is nil, and the tail recursion on line 26 has the same guard, so both entry points
evaluate it as nil. (One more caller exists in `Game::Base#action_processed`, but it bails out
before reaching the round in 1835.) Deliberate cover for a future caller, or can it go?

spec/lib/engine/game/g_1835/game_spec.rb
```diff
@@ -0,0 +1,378 @@
...
+      buy(player_1, 'OBB')
+      buy(player_2, '5')
+      buy(player_3, '6')
+      buy(player_1, 'PB')
+
+      expect(player_1.percent_of(by)).to be 50
+      expect(player_3.percent_of(game.corporation_by_id('SX'))).to be 20
+
+      # player_2 has priority deal
+      expect(game.players.first).to eq(player_2)
```
chrstas, line 330: Does this hold at other player counts? The priority marker moves a fixed number of seats whatever the player count, so when the packet sells out 3 is the one where it lands right — at 4 to 7 the wrong player starts the next round, and 4 is one master gets right.

The clearest case to measure is the one where nobody buys at all, because XIV.3 pins the answer: "If no player buys a share during a share round then the loco stays with its previous owner." The distance from the previous holder should be 0, and it is 0, 3, 3, 3, 3 at 3-7 players here, 1, 0, 4, 4, 4 on master. Your change is a uniform −1 (mod n) in every case I tried, so the residue comes from the advances in `step/draft.rb` that you don't touch with this PR — which is what your question about bumping the counter is getting at.

Would this be a separate PR?

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
```diff
@@ -1,19 +1,25 @@
...
+          include Engine::Step::Nationalization
+
           def process_buy_shares(action)
             if action.bundle.owner.player?
-              raise GameError, 'Cannot nationalize this corporation' unless can_buy?(action.entity, action.bundle)
-
-              action.bundle.share_price = nationalization_price(action.bundle.corporation.share_price.price)
+              player = action.entity
+              bundle = action.bundle
+              nationalize(player, bundle)
```
chrstas, line 16: Was dropping the `can_buy?` gate intended? Master raised `'Cannot nationalize this corporation' unless can_buy?` here, and with `super` gone `check_legal_buy` doesn't run either — `nationalize` only checks `can_nationalize?` and cash, so the cert limit isn't enforced server-side on this path any more.

1847AE never gated it either, so the shared module lines 1835 up with it.

spec/lib/engine/game/g_1835/game_spec.rb
```diff
@@ -0,0 +1,378 @@
...
+      corp = game.corporation_by_id(corporation_id)
+      unless other_player
+        return game.process_action(Engine::Action::BuyShares.new(player,
+                                                                 shares: corp.shares.find(&:buyable)))
+      end
+
+      game.process_action(Engine::Action::BuyShares.new(player, shares: other_player.shares_of(corp).find do |share|
+                                                                          share.percent == percent
+                                                                        end))
+    end
```
chrstas: Is the double-price fix covered anywhere? I put master's `process_buy_shares` back and the file stays green — 7 examples, 0 failures. On that version the affordability check re-applies the multiplier to the price it just wrote: BY at 92 costs 138 to nationalize, but the check asks for 207, so anything from 138 to 206 bounces. The one example that nationalizes sets every player to 3000, well outside that band.

Part of why it slips past: `Game#process_action` rescues `StandardError` and parks it in `game.exception` (`game/base.rb:887`), so a rejected action is a silent no-op rather than a failure. `fixture_cache.rb:36` calls `maybe_raise!` after replaying a fixture, but no framework hook does that for actions a spec drives itself. `g_1836_jr56/game_spec.rb:52` chains `maybe_raise!` onto the call — worth doing in `buy`, `buy_shares` and `pass`?

lib/engine/game/g_1835/game.rb
```diff
@@ -285,6 +289,8 @@ def maybe_ipo_next_block(corporation)
           return if block == @corporation_blocks.last
 
           next_block = @corporation_blocks[@corporation_blocks.index(block) + 1]
+          return if next_block.any?(&:ipoed)
```
chrstas: `ipoed` isn't only set here - `SharePool#buy_shares` sets it too, whenever a president's share is bought or the corporation already has a player owner (`share_pool.rb:53`). So the guard can see it on a next-block corporation that was never announced, and the block never opens.

What keeps that out of reach is the UI: `BuySellShares` renders buy buttons only `if @corporation.ipoed` (`assets/app/view/game/buy_sell_shares.rb:43`), and the auto-buy programmer filters its corporation list on `ipoed` too.

Not from your PR: the buy path doesn't enforce the block order server-side. In the first stock round 1835's `can_buy?` returns true for all six corporations of the unopened blocks, and a `buy_shares` action on BA's president share is accepted.

Was making the call unconditional in `step/buy_sell_par_shares.rb:22` needed, or could the `unless owner == @game.share_pool` have stayed?

lib/engine/game/g_1835/game.rb
```diff
@@ -12,7 +12,7 @@ module Engine
   module Game
     module G1835
       class Game < Game::Base
-        attr_accessor :draft_finished, :pr_can_form, :conversion_choice_during_or
+        attr_accessor :pr_can_form, :conversion_choice_during_or
```
chrstas, line 15: `setup` still writes `@draft_finished = false` at line 199 - nothing reads it in 1835 since the accessor went. Leftover?

chrstas (commented):
Hi Sebastian, I've done a peer review - nothing that blocks the merge, and with the disclaimer that it's just a peer review and I'm still learning here myself.  Hope it helps...

Just thinking if it would be good to label it also with [1847 AE] and [CORE].

lib/engine/game/g_1835/round/draft.rb
kepakiano: Hey, no, it actually can't and really makes this function nonsensical, since - as @crericha stated - it should be a wrapper around `next_entity_index!` and not call it only conditionally. I have therefore made the call unconditional again and call `next_entity` from the step now instead of `next_entity_index!`.

spec/lib/engine/game/g_1835/game_spec.rb
kepakiano: Hey, as stated in our private conversation: First of all, thank you so much for diving so deeply into this, it is so very much appreciated! It gave me a lot of incentive to also dive deeper again and I think I found the culprit: Bumping the entity index in `skip!`. I have removed this (along with the changes mentioned above. Since I had already added tests for the PD after the draft, I added more for all allowed player counts to make sure it actually works. So no separate PR :) The tests make sure the PD is correct after either everyone passes or after the full sale of the starting packet.

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
kepakiano: It was not intentional, good catch! Added the check to the module so '47AE also benefits.

spec/lib/engine/game/g_1835/game_spec.rb
kepakiano: omg, `maybe_raise!` is so helpful. Thank you! I was getting annoyed by manually chasing down exceptions. Didn't know there was a way to raise them in the tests :)

To your actual point: No, no test had covered it, but it now does :)

lib/engine/game/g_1835/game.rb
kepakiano: yeah, I'm actually also not too happy about the tests checking `corporation_available`. The reason I removed the check on the call is that I wanted it to be inside the method, but I also wanted the method not to learn more about what was going on. I restored to how it was for now and will ponder on it more. Thanks!

lib/engine/game/g_1835/game.rb
kepakiano: good catch, removed.

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
chrstas: looks also good... 

Nitpick, not critical: Should the two raises share a message? Line 12 fires on any `can_buy?` rejection — cert limit, or the corporation already sold this round — and reports 'Cannot nationalize this corporation', which is the one thing that isn't the problem. In 1835 that text is master's (`can_buy?` was raised at the call site there), but 1847AE inherits it for the first time through the module. Line 13 looks unreachable on top: both steps start their `owner.player?` branch with `return false unless can_nationalize?`.

spec/lib/engine/game/g_1835/game_spec.rb
chrstas: @kepakiano, really happy if it helps - thanks for your words.. I believe a lot in the "4-eyes-see-more"...

Just one thing (based on the new commit): re-ran the priority deal over 25 one-buyer sequences, each seat as last purchaser at 3 to 7 players — 25 of 25 land where XIV.2 puts them now, against 4 of 25 on master. Well done :-)

On the new tests though: do the five 'should not move the PD if everyone passes' examples (`:43`, `:111`, `:145`, `:176`, `:206`) have a way to fail?  `player_1` is a lazy `let`, and nothing in the example touches it before the expectation, so it memoises `game.players.first` at that moment and compares it with itself. I swapped the `:after_last_to_act` branch of `reorder_players` for a plain `@players.rotate!(1)`: all five 'everyone passes' examples stayed green, while `:62`, `:148`, `:179` and `:209` failed.

lib/engine/game/g_1835/round/draft.rb
chrstas: Cool - looks good...

One small thing from that rework regarding the log: should the new `skip!` (`step/draft.rb:127`) still log once the round is over? Instrumenting `log_skip` over the 19 examples in `g_1835/game_spec.rb` gives 31 calls, 19 of them with `@round.finished?` true, one per example. `Engine::Step::Base#skip!` guards that with `if !@acted && current_entity` (`lib/engine/step/base.rb:61`), which covers 18 of the 19; the one it misses is `:46`, where b is logged as having no valid actions while holding 600 against a cheapest buyable company of 80.

kepakiano: Hmmm, okay, you're right, we log a little too many skips. I've changed the definition of `skip!` now, but I'm not sure I'm happy. I now call `super`, which calls `pass!`, but in order for all the tests to pass, I also need to call `current_entity.pass!`, which is confusing. `step` includes `Passer`, so `pass!` sets `@passed` of the step. `current_entity` is a Player, which also includes `Passser`, so that's a different `@passed` that gets set. I think the problem is that `round/draft` asks all the entities whether they've passed in order to find out whether the round has ended and calling `Engine::Step::Base::skip!` does not set that for the Entity, only for the entire Step. 
I guess setting `@passed` twice is OK though, because `Engine::Step::Base::buy_sell_par_shares` also does it.

spec/lib/engine/game/g_1835/game_spec.rb
kepakiano: Huh, very good point. Lazy Evaluation can be a pita :D I made it more robust by using the player IDs. Thanks!

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
kepakiano: Okay, I've changed the order and the messages. I know in the current two implementations the later calls include some of the earlier calls (`can_buy?` include `can_nationalize?`), but I think it has two advantages doing it this way: a) I don't want to force the current or coming implementations to act that way, I view the methods as opaque and kind of a coincidence that it is that way now and b) I can give more precise error messages.

chrstas: Agreed, and your precedent holds — `Engine::Step::BuySellParShares#pass!` does exactly that, `super` and then `current_entity&.pass!` (`lib/engine/step/buy_sell_par_shares.rb:203-211`). The two `@passed` answer different questions, so setting both is the honest thing here rather than a smell.

chrstas:
> Hi Sebastian, I've done a peer review - nothing that blocks the merge, and with the disclaimer that it's just a peer review and I'm still learning here myself. Hope it helps...
> 
> Just thinking if it would be good to label it also with [1847 AE] and [CORE].

@kepakiano Thanks for the quick turnarounds. The priority deal lands right at every player count now, and the tests catch it if that changes.

Nothing left from my side as a peer reviewer that should hold up a merge, I'd approve if that counted for anything here.

kepakiano:
Thanks again @chrstas! Splitting this PR up into https://github.com/tobymao/18xx/pull/12846, https://github.com/tobymao/18xx/pull/12847 and https://github.com/tobymao/18xx/pull/12848

PR #12832, opened 2026-08-10
[1835] implement distance based tokening cost

### LINKED ISSUE #3059: [1835] Implementation Map
Implementation Roadmap for 1835

- [x] Map
- [x] Privates
- [ ] Minor Companies - Vorpreussische Bahnen
- [ ] Major Companies
- [x] Stockmarket
- [ ] Bidding Rules
- [ ] Bidding Snake "Clemens" Variant
- [ ] Shares/Certificate Different Types (5,10,20,40% share certificates)
- [ ] Director Shares may not be sold partially "Strict Rules"
- [ ] Nationalisation - based on Phased with Voluntary Merge and Mandatory Merge in upcoming Phases
- [ ] +Trains 
- [ ] Hamburg Hex with Different Values per City Space if you cross a river
- [x] Tiles

kepakiano:
Fixes #3059

## Before clicking "Create"

### Explanation of Change

Implementing the dynamic token cost for 1835. Tokens cost 20 per hex distance from the home hex of a corp, which works, as you can see in the screenshot: Nürnberg-Fürth is three hexes away, so BY pays 60.

### Screenshots
image

However, the regular distance function does not take the game borders into account: For München - Dresden it yields 7, but it's actually 8, because when calculating the distance, we need to stay inside the borders. This combination is the only one where this must be taken into account, so it's hard-coded: we add another 20 to the cost and the resulting 160 is correct: 
image

If a corporation does not have enough money for a token, the step is skipped. In this screenshot BY only has 20M, observe the "skipped" message:

image

One OR (and some money added to the treasury) later, it has the money and is now given the option to token:

image


lib/engine/game/g_1835/step/token.rb
crericha, line 1: If you want to skip the token step if the corporation doesn't have any token locations it can afford, you can do something like https://github.com/tobymao/18xx/blob/master/lib/engine/game/g_1837/step/token.rb. (Ignore the multiple token...that is for dealing with one token minors)

crericha (approved):

lib/engine/game/g_1835/step/token.rb
kepakiano: yeah, good idea, done :)

crericha (approved):

PR #12833, opened 2026-08-11
[1880 Romania] fix tile L141 label

### LINKED ISSUE #12826: [1880Romania] Tile Lay L141 not possible
Game ID (w/ action): https://18xx.games/game/262579?action=359

Want to upgrade B18 to tile L141, but not present as an option.

Tile upgrade table in rules suggest this should be a legal upgrade.

philcampeau:
Fixes #12826

## Before clicking "Create"

### Explanation of Change

I accidentally put 00 instead of OO as the label on tile L141.


perwestling (approved):
LGTM

PR #12834, opened 2026-08-14
[1835] set train limit for Preußen and condense phases

### LINKED ISSUE #3059: [1835] Implementation Map

kepakiano:
Fixes #3059

## Before clicking "Create"

### Explanation of Change

As suggested in an earlier PR, I condensed the phases into the four distinct phases that are actually needed. And the PR correctly sets the train limit of the Preußen, which is always one higher than that of the other majors.

### Any Assumptions / Hacks

I'm leaving the class `prussian` in there, but it is only used to show the train limit in the Info tab. Splitting up the `major` corporation class into `major` and `prussian` seems much more complicated than the implementation in this PR, because the only difference between the PR and other majors is the train limit and that can be easily solved using an `ability`.


crericha:
> I'm leaving the class prussian in there, but it is only used to show the train limit in the Info tab. Splitting up the major corporation class into major and prussian seems much more complicated than the implementation in this PR, because the only difference between the PR and other majors is the train limit and that can be easily solved using an ability.

You can set the Prussian's corporation type to :national and then have a different train limit for it.
https://github.com/tobymao/18xx/blob/f7b2ce462052b10947a815a5adf94de6d2b125b1/lib/engine/game/g_1837/game.rb#L64

Or type: prussian would probably work too.

EDIT: Looks like you already have the type :prussian in the train limit code. I think all you have to do is make that the type in entities.rb and remove the train_limit ablity.

kepakiano:
> > I'm leaving the class prussian in there, but it is only used to show the train limit in the Info tab. Splitting up the major corporation class into major and prussian seems much more complicated than the implementation in this PR, because the only difference between the PR and other majors is the train limit and that can be easily solved using an ability.
> 
> You can set the Prussian's corporation type to :national and then have a different train limit for it.
> 
> https://github.com/tobymao/18xx/blob/f7b2ce462052b10947a815a5adf94de6d2b125b1/lib/engine/game/g_1837/game.rb#L64
> 
> Or type: prussian would probably work too.
> 
> EDIT: Looks like you already have the type :prussian in the train limit code. I think all you have to do is make that the type in entities.rb and remove the train_limit ablity.

yeah, I guess it's a little weird to have the "unused" type `prussian` dangling around in the phase. I thought having to double check for `:major` and `:prussian` everywhere would be cumbersome, but after moving that into a method it actually looks neat. Hence, I removed the ability.

lib/engine/game/g_1835/game.rb
```diff
@@ -365,6 +317,10 @@ def any_conversion_choice_available?
           prussian.floated? && !prussian_exchangeables.reject(&:closed?).empty?
         end
 
+        def major?(corporation)
+          corporation.type == :major || corporation.type == :prussian
```
crericha, line 321: Nit: its more idiomatic ruby to do `%i[major prussion].include?(corporation.type)`

crericha (approved):

PR #12842, opened 2026-08-20
[1833NE] initial commit

philcampeau:
## Before clicking "Create"

### Explanation of Change

This is just the initial commit - really just a skeletal structure. 

NOTE: GMT Games and Tom Lehmann have asked that the game not go live before December 1, 2026 at the earliest. I added a comment in the meta.rb file on the DEV_STAGE line noting as much. 

### Screenshots

image

image

### DIFF
--- a/lib/engine/game/g_1833_ne/step/takeover_corporation.rb
+++ b/lib/engine/game/g_1833_ne/step/takeover_corporation.rb
@@ -7,19 +7,21 @@ module Engine
     module G1833NE
       module Step
         class TakeoverCorporation < Engine::Step::Base
-          def actions(entity)
-            return [] if entity != current_entity
-            return [] unless entity.corporation?
-            return [] unless @round.view_merge_options
+          #  The code below is a placeholder for now
 
-            %w[takeover pass]
-          end
+          # def actions(entity)
+          #   return [] if entity != current_entity
+          #   return [] unless entity.corporation?
+          #   return [] unless @round.view_merge_options
 
-          def auto_actions(entity)
-            return if entity != current_entity || !@round.view_merge_options || !@game.takeover_candidates(entity).empty?
+          #   %w[takeover pass]
+          # end
 
-            [Engine::Action::Pass.new(entity)]
-          end
+          # def auto_actions(entity)
+          #   return if entity != current_entity || !@round.view_merge_options || !@game.takeover_candidates(entity).empty?
+
+          #   [Engine::Action::Pass.new(entity)]
+          # end
         end
       end
     end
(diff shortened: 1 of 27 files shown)


lib/engine/game/g_1833_ne/step/waterfall_auction.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require_relative '../../../step/waterfall_auction'
+
+module Engine
+  module Game
+    module G1833NE
+      module Step
+        class WaterfallAuction < Engine::Step::WaterfallAuction
+          def min_bid(company)
+            return unless company
```
crericha: why would company be nil?

lib/engine/game/g_1833_ne/step/takeover_corporation.rb
```diff
@@ -0,0 +1,27 @@
...
+          def actions(entity)
+            return [] if entity != current_entity
+            return [] unless entity.corporation?
+            return [] unless @round.view_merge_options
+
+            %w[takeover pass]
+          end
+
+          def auto_actions(entity)
+            return if entity != current_entity || !@round.view_merge_options || !@game.takeover_candidates(entity).empty?
```
crericha: Does auto_auctions get called for anyone but current_entity?

lib/engine/game/g_1833_ne/step/takeover_corporation.rb
```diff
@@ -0,0 +1,27 @@
...
+
+            [Engine::Action::Pass.new(entity)]
```
crericha: So if its current entity and there are view merge options and there are takeover candidates, you want to pass?

lib/engine/game/g_1833_ne/step/waterfall_auction.rb
philcampeau: I copied the `def min_bid` method from the base waterfall_auction.rb step and modified it so that the starting bid can be face value instead of $5 over. 

The base step used `return unless company`, so I left it in. This is just placeholder code for now; I hadn't started going over specifics. I can remove this line.

lib/engine/game/g_1833_ne/step/takeover_corporation.rb
philcampeau: Like what i wrote above, this is just placeholder code for now. I actually copied what's in here straight out of 18NY's AcquireCorporation step. It's very likely not the final code.

philcampeau: I commented out the code from that section for now. I'll revisit it when I really start getting things going.

lib/engine/game/g_1833_ne/step/takeover_corporation.rb
philcampeau: I commented out the code from that section for now. I'll revisit it when I really start getting things going.

crericha (approved):

PR #12848, opened 2026-08-28
[1835] fix nationalization

kepakiano:
refers to #3059

is blocked by https://github.com/tobymao/18xx/pull/12847

Check out the PR this PR was split from https://github.com/tobymao/18xx/pull/12829 lots of feedback has been given and applied there already.

## Before clicking "Create"

### Explanation of Change

There was a bug in the nationalization code of 1835 (the price was increased to 150% twice), which is fixed here by simply reusing the code of 1847AE.


lib/engine/game/g_1835/step/buy_sell_par_shares.rb
```diff
@@ -1,28 +1,34 @@
 # frozen_string_literal: true
 
+require_relative '../../nationalization'
+
 module Engine
   module Game
     module G1835
       module Step
         class BuySellParShares < Engine::Step::BuySellParShares
+          include Engine::Step::Nationalization
```
crericha: Can you explain why an entire step is being included? That seems wrong. I was expecting a module with shared code you wanted to use here.

lib/engine/game/nationalization.rb
crericha, line 1: There are a lot of games with nationalization. What makes this the correct generic logic?

lib/engine/game/g_1835/step/buy_sell_par_shares.rb
crericha: Reviewing further, I see it is a module. Closing.

lib/engine/game/nationalization.rb
crericha: This is a trivial amount of code. Unless this is truly the generic module nationalization in most games should be based on, I suggest duplicating the logic in the two games.

kepakiano: Hmm... I think the term "nationalize" has multiple meanings in the 18xx world. When I look for the word in the code base, I think only '47AE and '35 use it in the sense of one player forcedly buying shares of another player. Does this mechanic exist in other games and is called something else?
If not, then yes, for the two games that use the word in that sense, it seems to be the generic way. However, I do see your point that it's not much code. I'll split it up.

kepakiano: done!

crericha (approved):

PR #12856, opened 2026-09-03
[1835] implement Clemens variant

kepakiano:
refers to #3059

## Before clicking "Create"

### Explanation of Change

Implements the Clemens variant of the starter packet draft. A few months ago, we talked on Slack about which variants to implement and toby decided that Clemens will be the only one besides the vanilla draft.

### Any Assumptions / Hacks

- The round numbers are still out of whack, but that will be fixed in a later PR. I needed to set the round number for the draft here, though, because the very first draft round has the player order reversed on the first round
- If all players pass before each player has bought something (e.g. first player buys, then all pass until the round is over), the PD will be with Player 1. An argument could be made for the PD having to be with Player (n-1), but this is a very hypothetical edge case and I haven't found an answer to whose turn it should be, so I guess Player 1 is as good an answer as any.
- Only the very first round in the first draft round goes in reversed order. This is how I understand the Clemens opening and it's backed by some descriptions on line (e.g. https://home.snafu.de/jasiek/18varrul.html)

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -163,6 +163,10 @@ module Engine
           @corporation_blocks = CORPORATION_BLOCKS.map { |block| block.map { |c| corporation_by_id(c) } }
         end
 
+        def option_clemens?
+          @optional_rules&.include?(:clemens)
+        end
+
         def company_header(company)
           return 'MINOR' if '123456'.include?(company.sym)
           return 'SHARE' if company.sym == 'BY_D'
@@ -175,13 +179,13 @@ module Engine
         end
 
         def init_round
-          G1835::Round::Draft.new(self,
-                                  [G1835::Step::Draft])
+          new_draft_round
         end
 
         def new_draft_round
-          @log << "-- #{round_description('Draft')} --"
-          init_round
+          @log << "-- #{round_description('Draft', 0)} --"
+          G1835::Round::Draft.new(self,
+                                  [G1835::Step::Draft], round_num: @draft_round_num)
         end
 
         def next_round!
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -179,13 +179,13 @@ module Engine
         end
 
         def init_round
-          new_draft_round
+          G1835::Round::Draft.new(self,
+                                  [G1835::Step::Draft], round_num: @draft_round_num || 1)
         end
 
         def new_draft_round
-          @log << "-- #{round_description('Draft', 0)} --"
-          G1835::Round::Draft.new(self,
-                                  [G1835::Step::Draft], round_num: @draft_round_num)
+          @log << "-- #{round_description('Draft')} --"
+          init_round
         end
 
         def next_round!
--- a/lib/engine/game/g_1835/round/draft.rb
+++ b/lib/engine/game/g_1835/round/draft.rb
@@ -7,6 +7,17 @@ module Engine
     module G1835
       module Round
         class Draft < Engine::Round::Draft
+          def setup
+            return unless first_round_in_a_clemens_draft
+
+            @first_round_in_first_draft_round = true
+            @entity_index = @entities.size - 1
+          end
+
+          def first_round_in_a_clemens_draft
+            round_num == 1 && @game.option_clemens?
+          end
+
           def select_entities
             @game.players
           end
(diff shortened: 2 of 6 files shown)


lib/engine/game/g_1835/round/draft.rb
```diff
@@ -7,6 +7,17 @@ module Game
     module G1835
       module Round
         class Draft < Engine::Round::Draft
+          def setup
+            return unless first_round_in_a_clemens_draft
+
+            @first_round_in_first_draft_round = true
+            @entity_index = @entities.size - 1
```
chrstas: Which of the two answers to "who gets priority" is the intended one after the reversed round? `priority_deal_player` counts from `last_to_act`, while `reorder_players` takes `@players[@round.entity_index]` (`base.rb:3163`); those coincide in a forward round but not here. In a 4p `:clemens` game where d and c buy and everyone then passes, the marker shows d and `players.first` comes out as a. 

Neither the Clemens description nor XIV.2 settles the reversed case - the variant text is silent on priority and XIV.2 assumes a clockwise round - so this is your call... Your PD example passes because both land on the same player there.

spec/lib/engine/game/g_1835/game_spec.rb
```diff
@@ -309,6 +309,117 @@ def may_purchase?(company_id)
     end
   end
 
+  describe 'clemens_draft' do
+    let(:players) { %w[a b c] }
+    let(:game) { Engine::Game::G1835::Game.new(players, optional_rules: [:clemens]) }
+
+    it 'should be possible to buy any paper' do
```
chrstas: Does this one need an expectation? `actions` only asks whether *any* company is purchasable (`step/draft.rb:74`) and `process_bid` doesn't re-check, so the whole `clemens_draft` block still gives `6 examples, 0 failures` with `return true if @game.option_clemens?` deleted from line 37, or? The vanilla rule is asserted through the `may_purchase?` helper at `game_spec.rb:239`.

lib/engine/game/g_1835/game.rb
```diff
@@ -175,13 +179,13 @@ def init_share_pool
         end
 
         def init_round
```
chrstas, line 181: Cosmetic: Should there be a second `-- Draft Round 1 --` right before the stock round? `init_round` now logs, and `base.rb:2956` evaluates `when init_round.class`, so the case statement itself emits a header for a round that never runs. A 3p game without `:clemens` logs two Draft Round headers on this branch and none on master.

chrstas (commented):
hi @kepakiano, I have added 3 peer-review comments, none blocking. Two are player-visible (the priority deal after the reversed round, and a duplicated draft header in the log), one is a test that stays green without the rule it covers. All measurements are from 3p and 4p games; I did not look at the round numbering you parked in the description.

lib/engine/game/g_1835/game.rb
kepakiano: Ah, true, good catch. Should be fixed now.

spec/lib/engine/game/g_1835/game_spec.rb
kepakiano: I guess I'll add some assertions to `process_bid` :)

lib/engine/game/g_1835/round/draft.rb
kepakiano: I guess my intended answer is the one with the least amount of code, since the case is pretty academical :D Hence, Player 1 would be my answer.

kepakiano:
> hi @kepakiano, I have added 3 peer-review comments, none blocking. Two are player-visible (the priority deal after the reversed round, and a duplicated draft header in the log), one is a test that stays green without the rule it covers. All measurements are from 3p and 4p games; I did not look at the round numbering you parked in the description.

Thank you!

lib/engine/game/g_1835/round/draft.rb
kepakiano: Ah, I finally understood, what you mean: As long as we're in reverse order, the PD marker is rendered for the wrong player. I'll take a look.

kepakiano: Alright, I think I got it: By not setting `@round.last_to_act` in the very first Clemens round, the PD now stays with player 1. Also, the `entity_index` is now increased instead of decreased and the `entities` are reversed in the first round.

kepakiano: Sadly, I cannot test `priority_deal_player`, because it's a private method :(

PR #12859, opened 2026-09-05
[1835] forbid building over the Elbe west of Hamburg

kepakiano:
refers to #3059

## Before clicking "Create"

### Explanation of Change

Not much to say here: Before this PR one could build over the Elbe west of Hamburg. It's now `impassable` and the color has been set.

### Screenshots

image

Map before trying to build.

image

It's impossible to rotate the curve so that it points into C9. This was possible before this change.


chrstas (commented):
Peer Review: Looks good to me, checked against rule VI and against the ferry code. `hamburg_ferry?` puts C9 on the south side of C11 and B10 on the north, so the blocked edge is the right one. Ferry fee and border cost are unaffected.

PR #12860, opened 2026-09-06
[1835] make track laying permissive and forbid building into brown hexes

kepakiano:
refers to #3059

## Before clicking "Create"

### Explanation of Change

Just implementing the rules here:
- track laying is permissive (rule 5.5.1.14 in the German rulebook "Eine Gesellschaft darf nur in Feldern Plättchen austauschen, auf denen sie nach diesem Austausch mindestens eine minimale Strecke befahren kann.")
- brown hexes cannot be built into unless there is track on the other side (rule 5.5.1.21 in the German rulebook "Gleise dürfen nur dann in braune Felder hineingebaut werden, wenn sich an der anzuschließenden Seite ein Gleiszugang befindet.")

### DIFF
--- a/lib/engine/game/g_1835/game.rb
+++ b/lib/engine/game/g_1835/game.rb
@@ -52,6 +52,10 @@ module Engine
 
         MUST_BUY_TRAIN = :always
 
+        # brown tiles - in addition to the default - cannot be built into unless there is track on the other side
+        IMPASSABLE_HEX_COLORS = %i[blue brown gray red].freeze
+        TRACK_RESTRICTION = :permissive
+
         MARKET = [['', '', '', ''] + %w[132 148 166 186 208 232 258 286 316 348 382 418],
                   ['', ''] + %w[98 108 120 134 150 168 188 210 234 260 288 318 350 384],
                   %w[82 86 92p 100 110 122 136 152 170 190 212 236 262 290 320],


lib/engine/game/g_1835/game.rb
```diff
@@ -52,6 +52,10 @@ class Game < Game::Base
 
         MUST_BUY_TRAIN = :always
 
+        # brown tiles - in addition to the default - cannot be built into unless there is track on the other side
+        IMPASSABLE_HEX_COLORS = %i[blue brown gray red].freeze
```
chrstas, line 56: Does 5.5.1.21 also cover hexes that turn brown during the game? `IMPASSABLE_HEX_COLORS` is read once when the neighbour lists are built, so a hex promoted to brown in phase 3 keeps its neighbours and track can still be laid into its blank side; permanently, since 1835 has no gray tiles. 5.5.1.20 says "vorgedruckte" where it means the printed hexes and 5.5.1.21 does not, but 5.5.1.1 describes brown hexes as map scenery, so I could not settle it.

chrstas (commented):
hi @kepakiano, peer review: LGTM, nothing blocking. I checked the rest against the map: no pre-printed brown hex loses an existing connection, and the edges that close are exactly the blank ones.

One topic (rather question) but with disclaimer: I am not an 1835 expert and am going by the rulebook alone, so I left one question inline on the brown half.

lib/engine/game/g_1835/game.rb
kepakiano: Hey, I had the same question, see Donald's and my conversation on Slack. The answer to your question is "no": "Braune Felder" are only the pre-printed ones, players are allowed to build track into non-pre-printed brown tiles even if there's no track :)

chrstas: Thanks, that settles it. 5.5.1.21 leaving out the "vorgedruckte" from 5.5.1.20 is easy to misread, 5.5.1.1 is what ties it back. Nothing else from my side on this one.

chrstas:
> hi @kepakiano, peer review: LGTM, nothing blocking. I checked the rest against the map: no pre-printed brown hex loses an existing connection, and the edges that close are exactly the blank ones.
> 
> One topic (rather question) but with disclaimer: I am not an 1835 expert and am going by the rulebook alone, so I left one question inline on the brown half.

Thanks, that answers it. Nothing left from my side, the PR looks good from a peer review point of view.

Issue #12329, opened 2025-12-31
[18 India] If two companies start on an OO hex, the 2nd company's token can be deleted when the yellow OO tile is laid, preventing that 2nd company from ever operating.


jbswets:
This issue originally happened in a Multiplayer game but many actions were rolled back to allow us to continue playing and bypass the bug.  Multiplayer and Hotseat details included below:

https://18xx.games/game/237206?action=81
https://gist.github.com/jbswets/63758a6e1a140c83a8542536f1491b42

In Stock Round 1 (Action 46), GIPR was floated and placed a token in G8 (DELHI), one of two O-O hexs on the map.  In OR 1.1 when NWR was laying a tile on this hex (Action 81/82), NWR was not given the option to re-place the GIPR token on this new hex, effectively deleting the token.  Advancing the game confirmed that GIPR was unable to build or upgrade track or place additional tokens as it had no existing tokens placed on the map.

Issue appears on desktop and mobile, in multiple browsers.  Most testing was done on the following browser:

Chrome Version 143.0.7499.170 (Official Build) (64-bit)

I saw a similar issue when this version was newly published, I will try to find that bug report/game info to include as well.

Please let me know if you need any additional information.  Thanks!

jbswets:
This appears to be the same issue described in this issue thread:

https://github.com/tobymao/18xx/issues/11058

ollybh:
This isn't the same as #11058. That was for when there was a reserved token slot for another company in an OO hex. In this case we have an actual token, not a reservation.

I think this is a variation of #11699. PR #11860 fixed that bug, but is expecting the two companies in the Delhi hex to be NWR and SPD. In this game we have NWR and GIPR instead, so the second token is not being moved.

I don't know 18India, has something gone wrong to allow GIPR to start in Delhi instead of SPD? Or is that part correct, and the code needs to be generalised to work with different companies?

[CC: @MatthewDeKoning]

jbswets:
Good points ollybh - In this particular game SPD was not an available company so there was not a reserved token spot for that company.

GIPR is a unique company in 18 India that is always available to start, and picks its starting location when it floats. It shouldn't be possible for any other company other than GIPR or a company that starts there to have a token placed on a OO hex before it is yellow. So 5 total companies across the two OO hexes.

ddaybell:
[...]

Issue #11719, opened 2025-05-05
[18India] OO Tile does not allow tokening in yellow when only one OO company is in the game


fcboyd:
**What's your Game ID?**
https://gist.github.com/fcboyd/120c6abbaeb958bf357f9d10437fab4c

**How to Reproduce the Bug?**
Lay Yellow OO tile. Game will skip the option to place a token in OO. Since there is only a single OO Delhi company in the game, should be able to force NWR reserve spot to the other station in OO and token with PNS.
<img width="493" alt="Image" src="https://github.com/user-attachments/assets/2651680d-c9a7-4db9-b0a4-142f0313422b" />

**Which Browser / Platform?**
Chrome Version 135.0.7049.85 (Official Build) (x86_64)

ollybh:
Rules clarification for this, from the [18 India FAQ](https://boardgamegeek.com/thread/3208484/official-living-faq-for-18-india) on BGG (emphasis added):
> **Q: When placing the yellow (OO) tile in Delhi or Kolkata, who decides which city location (white circle) is assigned to which preexisting station token is already located on that hex?**
> **A:** In Delhi and Kolkata where two Companies exist on the same (OO) hexagon and one or both of those companies has already placed its station token, the Company laying the tile decides where the Companies’ stations go on the hexagon, even if the Company laying the tile is not one of the two Companies located in the hexagon. _Similarly, when only one of NWR / SPD is in the game, a different company may place the yellow 00 tile and place a token, forcing NWR/SPD into the other token spot._

ddaybell:
Fixed by #12533. The root cause was in HomeTrack#actions, which offered lay_tile to all pending entities rather than restricting it to the GIPR. A subsequent patch addressed the symptom by bypassing the engine's OO token placement mechanism entirely and replacing it with custom case logic - which introduced this crash. PR #12533 restores the engine's native behavior by fixing the one-line root cause, removing all the custom placement logic, and extending move_oo_reservations to also handle unfloated OO corps so they get an interactive city choice rather than auto-placing in a potentially unconnected city.

Issue #11308, opened 2024-10-29
[18 India] Can't take further legal actions after choosing home station in OO city


fanman80:
Game 183290

https://18xx.games/game/183290

Action https://18xx.games/game/183290?action=81

Here are the comments I made to alert the other players to the situation:
[20:50] Pawnpusher: @all, It's not letting me do anything after choosing where the NWR token goes.
[20:52] Pawnpusher: It is letting me upgrade Delhi like it's a dit, into a one station green city. Obviously, that's not what I want to do.
[20:56] Pawnpusher: There's no "Skip Track" button or "Done Track" button or "Skip Token" button or any other kind of action button.
[20:58] Pawnpusher: I can "Undo" or "Restart Turn", and that's it.

BillPeeck:
Game 189032 

https://18xx.games/game/189031


I suppose a 2nd version of the same problem does not help you.

MatthewDeKoning:
this should be fixed soon, duplicate of #11699

ddaybell:
Fixed by #12533. The root cause was in HomeTrack#actions, which offered lay_tile to all pending entities rather than restricting it to the GIPR. A subsequent patch addressed the symptom by bypassing the engine's OO token placement mechanism entirely and replacing it with custom case logic - which introduced this crash. PR #12533 restores the engine's native behavior by fixing the one-line root cause, removing all the custom placement logic, and extending move_oo_reservations to also handle unfloated OO corps so they get an interactive city choice rather than auto-placing in a potentially unconnected city.

PR #12866, opened 2026-09-12
[18 India] Fix OO corp home token crash and blocked home token placement for fixed-home corps

### LINKED ISSUE #11461: [18 India] WR player cannot lay home token
The game is https://18xx.games/game/194445. It was the second to last OR, even though there was space WR did not lay the home token when it was directed.

### LINKED ISSUE #12413: [18India] - Cannot Token a newly floated SPD on G* (Dehli)
**What's your Game ID?**
https://18xx.games/game/243370

**How to Reproduce the Bug?**

Image
When clicking the city circle that is empty in its starting tile, the error message "undefinied method 'tokenable?' for nil" appears

On Chrome 145.0.7632.117

ddaybell:
Fixes #12413, #12329, #11719, #11308, #11461

**Size:** 13 files changed, +188/-84 (`lib/engine/game/g_18_india/entities.rb`, `lib/engine/game/g_18_india/game.rb`, `lib/engine/game/g_18_india/step/home_track.rb`, `lib/engine/game/g_18_india/step/token.rb`, `lib/engine/game/g_18_india/step/track.rb`, `spec/lib/engine/game/g_18_india/step/home_track_spec.rb`, plus 6 fixture files under `public/fixtures/18India/`)

This PR combines and replaces #12533 (OO corp home token crash) and the original #12866 (WR home token blocked). They were originally separate, but both touch `home_track.rb`'s `actions`/`any_open_cities?`/`description` and a reviewer flagged that the two branches would collide, so this folds them into a single PR to review together.

## Before clicking "Create"

- [x] Tests pass cleanly (`docker compose exec rack bundle exec rspec spec/lib/engine/game/fixtures_spec.rb` — 7429 examples, 0 failures; plus the new `home_track_spec.rb`)

### Part 1: OO corp home token crash (originally #12533)

**Root cause:** The four OO corporations (NWR and SPD at G8/Delhi; EIR and EBR at P17/Kolkata) share a home hex. With `HOME_TOKEN_TIMING = :float`, each corp places its home token when it first operates. If one OO partner had already floated and the shared hex was since upgraded to yellow by the time the second partner floated, the second placement crashed.

**Design decision — tile-level reservations:** An earlier approach assigned `city: 0` / `city: 1` to each OO corp (NWR/EIR → city 0, SPD/EBR → city 1) so each would reserve a specific city slot. This caused problems: since tokens aren't placed until float time, the forced city pre-assignment created conflicts in the placement logic, requiring complex reservation-shuffling code. Reverting to **tile-level reservations** (removing `city:` from the corp definitions) is cleaner — the corp's initials appear as text labels on the tile until placement, and city assignment is deferred to the moment the token is actually placed. This let us delete `place_second_oo_token`, `swap_higher_value_oo_token`, `replace_oo_reservations`, and `move_oo_reservations` entirely.

**Fix — `game.rb`:**
- `place_home_token` now dispatches to separate OO/GIPR/default sub-methods (per reviewer feedback to stop mixing OO and GIPR logic in one method). On a white tile: places the token directly into the first untokened city (no pending queue needed). On a yellow tile (already upgraded before this corp floated): adds to `pending_tokens` for the tile-laying player to choose a city for each token already on the tile (conforms to FAQ 8.4.1.1, q:2).
- Removed the duplicate `Engine::Step::HomeToken` from `stock_round` and `operating_round`; `G18India::Step::HomeTrack` already handles all home-token work for this game.

**Fix — `step/track.rb`:**
- Added `update_token!` override. When a white OO tile is upgraded to yellow while it already holds placed token(s), the override lifts each token out of its city, logs a message, and queues it in `pending_tokens` for city re-assignment.

**Fix — `step/home_track.rb`:**
- `pending_token` skips OO corp entries whose home hex still has no paths (still a white tile), preventing those entries from blocking other steps prematurely.
- `actions` restricts `lay_tile` to GIPR only (OO corps don't upgrade their own home tile via this step).
- `any_open_cities?` handles the OO case by checking whether the home hex has paths before offering city placement.
- `process_lay_tile` uses `pending[:token]` and `delete` instead of `token` and `shift`, correctly targeting the specific pending entry (not just the first one).
- Added `auto_actions` to auto-place when only one untokened city remains.
- Removed: `place_second_oo_token`, `swap_higher_value_oo_token`, `process_place_token`, `replace_oo_reservations` — all replaced by the simpler approach above.

**Fix — `step/token.rb`:** Removed the `can_place_token?` override; no longer needed.

**Also fixed — Nationalization fixtures:** All 4 Nationalization test fixtures had a stale `pass` action (id=46), previously required by `Dividend` before it was changed to auto-skip at $0 revenue; it was being consumed by `SellBuyTrain` instead, blocking the subsequent `buy_train`. Removed from all 4 fixtures and renumbered subsequent action ids.

### Part 2: WR home token blocked when board has no other open city (originally #12866, fixes #11461)

**Root cause:** `HomeTrack#actions` gated the `place_token` action for *every* corporation (not just GIPR) on **board-wide** availability:

```ruby
actions << 'place_token' if any_open_cities?   # @game.open_city_hexes — every city on the ENTIRE map
```

For a fixed-home corp like WR, KGF, or TR, this is meaningless — `available_hex` already restricts placement to the corp's own home hex(es) via `pending_token[:hexes]`. If the rest of the board filled up late in the game, this board-wide check could fail even though the corp's own home hex was still open, leaving it with zero legal actions.

**Fix:** `any_open_cities?` now only applies the board-wide check to GIPR (whose home rule really is "pick any open city on the board"). For any other pending entity that isn't an OO corp, it returns `true` unconditionally — `available_hex` already does the real gating. `description` similarly falls back to the base `HomeToken#description` wording ("Place Home Token") for these corps instead of GIPR's "Lay home token in open city or upgrade town for X" phrasing, which implied a choice these corps don't actually have.

This also fixes the identical latent bug for every other fixed-home corporation in 18 India, not just WR.

A new fixture spec (`spec/lib/engine/game/g_18_india/step/home_track_spec.rb`) covers this directly, and the full 18 India fixture suite (including `OO-Fix`) was re-run against the combined branch to confirm the two fixes compose correctly — the OO-corp home-token dispatch and this fix touch the same methods, and a naive fix here could have broken the OO-corp path.

### Screenshots

image
image

The white OO tiles (G8/Delhi, P17/Kolkata) show both corps' initials as tile-level reservation labels before either corp has floated.

image

After the first corp floats, its token appears in a city circle while the other corp's label remains.

image

After the second corp floats, both tokens appear in the city circles.

image
image

After yellow is laid on the OO hex, the displaced token(s) re-enter pending placement and the tile-laying player chooses a city.

WR/KGF/TR home token placement (part 2) is engine-only and has no rendering change.

### Any Assumptions / Hacks

- On a white OO tile, `place_home_token` places into `tile.cities.find { |c| !c.tokened? }` — the first untokened city. Whichever OO partner floats first gets city 0; the second gets whatever city remains. The game rules do not prescribe which city belongs to which corp on the white tile, so this is acceptable — since the white tile has no paths, it is irrelevant which city each company occupies. Once the yellow OO tile is placed, the placing player relocates the OO tokens to the desired cities, which fixes their locations for the rest of the game.
- None for part 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/engine/game/g_18_india/step/home_track.rb
```diff
@@ -13,6 +13,7 @@ class HomeTrack < Engine::Step::HomeToken
 
           def actions(entity)
             return [] unless entity == pending_entity
+            return super unless entity.name == 'GIPR'
```
chrstas: Is #11461 the issue this PR fixes? The change itself looks right to me, but this step only runs when a corporation has to choose where its home token goes, and D17 has a single city, so WR is never asked. In game 253506 from the issue thread, the problem sits one step earlier: at action 503 the player becomes WR's first Manager after corporations already hold 30%, and neither the float nor a home token request happens, on master and on this branch. I checked that your #12868 fixes that: on its head, the token lands on D17 at that action.

Where this change does show is in issue #11308. In the OO-Fix fixture at action 103, SPD has to place NWR's token in Delhi. Master titles that step "Lay home token in open city or upgrade town for SPD" and offers a tile lay; this branch shows "Place NWR Home Token" and only the token.

Another observation: Your #12533 already lists #11308 and changes the same two methods. It also drops the generic HomeToken step from both rounds, and on its head, with the stubs from your spec, KGF gets no place_token at all and the round moves on to SellOnceThenBuyCerts. That matches the description here. Is this PR a piece split out of #12533, and if so, is it meant to land before it?

On #12533 the same check reads `entity == @game.gipr`; is `.name == 'GIPR'` here on purpose?

chrstas (commented):
Hi @ddaybell, I had a look from a peer-review perspective. One question on which issue this closes; the code change itself checked out against the rules (2.0, 2.4.3, 8.4.2) and replays identically on the 18India fixtures and on game 253506.

ddaybell:
@chrstas Thanks for catching this — you're right, and it was worse than just a collision.

This is now a piece split out of #12533, and it needed to land after it, not before: my fix here originally did a blunt `entity.name == 'GIPR'` check that bypassed `HomeTrack`'s whole override for any non-GIPR entity, which would have broken the OO-corp dispatch logic from #12533 for exactly the #11308 case you found (NWR/SPD at Delhi). I've rebased this branch onto #12533 and rewrote the fix to live inside `any_open_cities?` alongside the OO-aware check instead of short-circuiting past it, and re-ran the full 18India fixture suite (including `OO-Fix`) against the combined branch to confirm both fixes compose correctly.

Rather than try to keep these as two branches with a base-branch dependency (GitHub won't let a fork PR's base point at another branch that only exists in the fork), I've folded #12533's changes into this PR and closed #12533 — so this PR now covers both fixes. I also switched the `entity == @game.gipr` / `.name == 'GIPR'` discrepancy you flagged to consistently use `@game.gipr` (object identity), matching #12533's existing convention.

PR description above is updated to reflect the combined scope.

ddaybell:
Note for whoever merges this: #12705 (TR home token blocked when Nepal is full) depends on this PR — it patches `HomeTrack#process_place_token`/`place_second_oo_token`, both of which this PR removes. #12705 will need a rebase after this merges, and shouldn't be merged before it. Flagged on #12705 as well.

PR #12874, opened 2026-09-15
Home page: add Delete All button for hotseat games

GreatA1exander:
Adds a Delete All button to the Hotseat Games row header that removes every hotseat game from localStorage. Uses the same two-click Delete -> Confirm pattern as the per-game Delete button, sharing its confirm_delete state so only one deletion is pending at a time.

## Before clicking "Create"

### Explanation of Change

Clearing out old hotseat games currently means clicking Delete → Confirm on every card, one page of 12 at a time. This adds a single **Delete All** button beneath Hotseat Games row header.

- `GameManager#delete_all_hotseats` removes every `hs_`-prefixed key from localStorage (the same keys `View::Home` uses to build the Hotseat Games row), then resets `confirm_delete`, which re-renders the page. With no hotseats left, the row disappears as it normally does.
- `View::GameRow` renders the button only for the `:hotseat` row, on its own line directly under the heading (second grid row, first column), so the Prev/Next links keep their positions. The first click stores `:all_hotseats` in the existing `confirm_delete` store key and relabels the button to Confirm; the second click deletes.
- Reusing `confirm_delete` means clicking Delete on an individual card cancels a pending Delete All and vice versa, so there is never more than one armed delete on the page.

Deletion covers all hotseat games, not just the ones on the current page. No server, database, or engine changes.

### Screenshots

A bunch of hotseats 
1_delete_all_button
Primed to delete
2_confirm_state
Deleted
3_after_delete

### Any Assumptions / Hacks

- Assumes all of a hotseat game's data lives in its single `hs_*` localStorage key, which is how `create_hotseat` and the per-card `delete_game` already treat it. Other localStorage keys (settings, map zoom, spreadsheet prefs, master-mode flags for multiplayer games) are left untouched.
- No new confirmation UI: it deliberately mirrors the existing per-card two-click pattern rather than using a `confirm()` dialog.


phil-campeau (approved):
looks good

philcampeau (approved):

